Issue #264, opened 2020-10-26
Improve Span Rate Limiting


joe-elliott:
**Is your feature request related to a problem? Please describe.**
The span rate limiter in the distributor correctly rejects spans when they cross the rate limit threshold.  Grafana Agent/OpenTelemetry collector will then then batch and resend.  The problem is that the rate limit regularly trips due to spiky load while the rate as calculated over a minute or so is well below the configured threshold.  

The result is that spans are unnecessarily sent multiple times.

**Describe the solution you'd like**
Investigate tracking spans/second limit over a longer time frame.

**Describe alternatives you've considered**

**Additional context**
The limit on this environment is 200k spans/second, but spans are still refused when the average spans/second is well below as calculated over a minute+ time period.

![image](https://user-images.githubusercontent.com/2272392/97195318-969ca380-1781-11eb-915d-1b2adfec64ee.png)

Distributors are logging:  
```
err="rpc error: code = ResourceExhausted desc = ingestion rate limit (25000 spans) exceeded while adding 295 spans"
```
The 25k span rate limit is from 200k / 8 distributors.

calvernaz:
I was having this same error while investigating the memory leak issue in the compactors. This was preventing me from creating a trace with +100k spans.

Even though the rate-limiting strategy was local, not global, I had the same confusing message, the configuration below:

```
overrides:
  ingestion_rate_limit: 100000
  ingestion_max_batch_size: 1000
```

and the resulting message after trying to push 10k messages:

```
err="rpc error: code = ResourceExhausted desc = ingestion rate limit (100000 spans) exceeded while adding 10001 spans"
```

After some debugging, I fixed the configuration and stopped getting this problem:

```
overrides:
  ingestion_rate_limit: 1000
  ingestion_max_batch_size: 10001
```

Basically, the bucket size needs to be large enough to accommodate 10k spans (10000 + 1). In my case, the rate limit isn't that important.

joe-elliott:
I believe you're describing is a slightly different issue.  There are two different limits in play at the same time:

- spans/second
- spans/batch

You were hitting the spans/batch limit and, yes, the error message could definitely be improved.

[...]

PR #445, opened 2021-01-08
Standardize ingestion burst setting

mdisibio:
**What this PR does**:
This PR renames IngestionMaxBatchSize to IngestionBurstSize to bring it more in line with cortex and loki.   Instead of a max request size, it is more general purpose and in all cases the value is configured quite higher than the ingestion_rate_limit. 

Some links:
https://github.com/grafana/loki/blob/master/pkg/util/validation/limits.go#L64 (default is 1.5x of rate)
https://github.com/cortexproject/cortex/blob/master/pkg/util/validation/limits.go#L97 (default 2x of rate)
Internally, the burst size is even more generous (generally 10x of rate).

It doesn't appear that we have documented the override/limit settings.  There is good documentation in cortex here, that we should consider adopting:  https://github.com/cortexproject/cortex/blob/master/docs/configuration/arguments.md#ingester-distributor--querier-limits

n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added


joe-elliott (approved):

joe-elliott:
This Fixes #264

PR #446, opened 2021-01-11
Serve config at HTTP endpoint

annanay25:

**What this PR does**:
Adds a `/config` handler to serve the config that tempo is running with. Makes it easier to debug in non-k8s environments.

Long term, we should add an API module similar to Cortex. 

Fixes #nada

**Checklist**
- NA Tests updated
- NA Documentation added


joe-elliott (commented):
Tempo can include sensitive information in its config.  We either need to find a way to redact that or perhaps only serve this endpoint if it's configured.

annanay25:
Good point, I had not considered that. I was following suit with Cortex, which serves config by default I guess.

annanay25:
Okay, cortex is clever :) - https://github.com/cortexproject/cortex/blob/master/pkg/util/flagext/secret.go

I'll reuse this.

annanay25:
Results in

```
storage:
    trace:
        backend: s3
        local:
            path: ""
        gcs:
            bucket_name: ""
            chunk_buffer_size: 10485760
            endpoint: ""
            insecure: false
        s3:
            bucket: tempo
            endpoint: minio:9000
            access_key: '********'
            secret_key: '********'
            insecure: true
            part_size: 0
            signature_v2: false

```

joe-elliott (approved):

PR #447, opened 2021-01-11
Update release process instructions

annanay25:

**What this PR does**:
Updates release procedure to skip the goreleaser changelog generation and use CHANGELOG instead.
`Add the github handles of contributors at the end of each entry` is a lot of manual work though 😕 

Fixes #<issue number>

**Checklist**
- NA Tests updated
- [x] Documentation added

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -17,6 +17,7 @@ builds:
       - -X main.date={{ .CommitDate }}
     mod_timestamp: '{{ .CommitTimestamp }}'
 changelog:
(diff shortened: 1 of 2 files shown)


joe-elliott (approved):
Agree that this is a fair amount of manual work that I'm not excited about.  It appears that goreleaser has other options such as providing your own file, but I'm not sure we want to go down that path.

annanay25:
One workaround is to let the PR authors add their handle while adding the CHANGELOG entry.

PR #449, opened 2021-01-11
Fix Ingester Stutter

joe-elliott:
**What this PR does**:
Ingesters currently show unreasonably high p99's on read and write.  This is likely due to slow appends to the headblock.   Note that under the hood it is [inserting into a slice](https://github.com/grafana/tempo/blob/a7961951f136e323ada598b26f0ff38f34da2174/tempodb/encoding/appender.go#L30).

This PR reduces the time that we lock the active traces map by simply adding them to a linked list.  Then we drop the active traces lock and only lock the blocks while we are adding them to the append block.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/ingester/instance.go
+++ b/modules/ingester/instance.go
@@ -1,7 +1,6 @@
 package ingester
 
 import (
-	"container/list"
 	"context"
 	"fmt"
 	"hash"
@@ -119,13 +118,12 @@ func (i *instance) CutCompleteTraces(cutoff time.Duration, immediate bool) error
 	i.blocksMtx.Lock()
 	defer i.blocksMtx.Unlock()
 
-	for e := tracesToCut.Front(); e != nil; e = e.Next() {
-		trace := e.Value.(*trace)
-		out, err := proto.Marshal(trace.trace)
+	for _, t := range tracesToCut {
+		out, err := proto.Marshal(t.trace)
 		if err != nil {
 			return err
 		}
-		err = i.headBlock.Write(trace.traceID, out)
+		err = i.headBlock.Write(t.traceID, out)
 		if err != nil {
 			return err
 		}
@@ -274,6 +272,8 @@ func (i *instance) FindTraceByID(id []byte) (*tempopb.Trace, error) {
 	return nil, nil
 }
 
+// getOrCreateTrace will return a new trace object for the given request
+//  It must be called under the i.tracesMtx lock
 func (i *instance) getOrCreateTrace(req *tempopb.PushRequest) (*trace, error) {
 	traceID, err := pushRequestTraceID(req)
 	if err != nil {
@@ -314,18 +314,16 @@ func (i *instance) resetHeadBlock() error {
 	return err
 }
 
-// retrieves active traces that are ready to be cut and returns them.  these
-// traces are then removed from the map
-func (i *instance) tracesToCut(cutoff time.Duration, immediate bool) *list.List {
-	now := time.Now()
-	tracesToCut := list.New()
-
+func (i *instance) tracesToCut(cutoff time.Duration, immediate bool) []*trace {
 	i.tracesMtx.Lock()
 	defer i.tracesMtx.Unlock()
 
+	now := time.Now()
+	tracesToCut := make([]*trace, 0, len(i.traces))
+
 	for key, trace := range i.traces {
 		if now.Add(cutoff).After(trace.lastAppend) || immediate {
-			tracesToCut.PushBack(trace)
+			tracesToCut = append(tracesToCut, trace)
 			delete(i.traces, key)
 		}
 	}
--- a/modules/ingester/instance.go
+++ b/modules/ingester/instance.go
@@ -318,11 +318,11 @@ func (i *instance) tracesToCut(cutoff time.Duration, immediate bool) []*trace {
 	i.tracesMtx.Lock()
 	defer i.tracesMtx.Unlock()
 
-	now := time.Now()
+	cutoffTime := time.Now().Add(cutoff)
 	tracesToCut := make([]*trace, 0, len(i.traces))
 
 	for key, trace := range i.traces {
-		if now.Add(cutoff).After(trace.lastAppend) || immediate {
+		if cutoffTime.After(trace.lastAppend) || immediate {
 			tracesToCut = append(tracesToCut, trace)
 			delete(i.traces, key)
 		}
--- a/modules/ingester/instance.go
+++ b/modules/ingester/instance.go
@@ -1,6 +1,7 @@
 package ingester
 
 import (
+	"container/list"
 	"context"
 	"fmt"
 	"hash"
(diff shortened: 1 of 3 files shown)


joe-elliott:
Temporarily deploying this to our ops cluster is showing promising results on write:

image

Not seeing improvements to the read path.  Indicating latency on reads is from another source.  Perhaps we could use tracing to solve this problem :thinking:.

mdisibio:
Great improvement.  LGTM

modules/ingester/instance.go
```diff
@@ -113,26 +114,32 @@ func (i *instance) PushBytes(ctx context.Context, id tempodb_encoding.ID, object
 
 // Moves any complete traces out of the map to complete traces
 func (i *instance) CutCompleteTraces(cutoff time.Duration, immediate bool) error {
+	now := time.Now()
+	tracesToCut := list.New()
```
davecheney, line 118: what about using a slice here, which would avoid the boxing of values into interface{} and let you use a range statement on line 132?

modules/ingester/instance.go
```diff
@@ -113,26 +114,32 @@ func (i *instance) PushBytes(ctx context.Context, id tempodb_encoding.ID, object
...
+
 	i.tracesMtx.Lock()
-	defer i.tracesMtx.Unlock()
+	for key, trace := range i.traces {
```
davecheney, line 121: What about pulling this out into its own method that returns a tracesToCut slice, then you cam move the i.traceMtx locking into that method as well.

davecheney (changes requested):

modules/ingester/instance.go
annanay25: Agree with this, we can estimate slice capacity using `len(i.traces)` to avoid using `append()`

Thanks for the review @davecheney :)

modules/ingester/instance.go
joe-elliott: Yeah, this could help with readability.  A defer always feels like a cleaner/safer way to unlock.

modules/ingester/instance.go
joe-elliott: Unsure about this.  Each time CutCompleteTraces is called only a very small percentage of the traces in the map will actually end up in the list/slice.  It would be difficult to estimate the length of the slice.

Going to try some benchmarks.

joe-elliott: Predictably when a low number of traces is cut from the active traces map the linked list implementation performs better.  That advantage is slowly lost as more and more traces are cut.

```
$ go test -bench=BenchmarkCutTraces* -benchmem
goos: linux
goarch: amd64
pkg: github.com/grafana/tempo/modules/ingester
BenchmarkCutTraces1-8            	    5420	    229893 ns/op	     105 B/op	       2 allocs/op
BenchmarkCutTraces10-8           	    3825	    276236 ns/op	     549 B/op	      11 allocs/op
BenchmarkCutTraces100-8          	    3654	    288851 ns/op	    4871 B/op	     101 allocs/op
BenchmarkCutTraces1000-8         	    3024	    370372 ns/op	   48070 B/op	    1001 allocs/op
BenchmarkCutTraces5000-8         	    1792	    771957 ns/op	  240075 B/op	    5001 allocs/op
BenchmarkCutTraces10000-8        	    1376	   1003460 ns/op	  480084 B/op	   10001 allocs/op
BenchmarkCutTracesSlice1-8       	    3433	    319624 ns/op	   81934 B/op	       1 allocs/op
BenchmarkCutTracesSlice10-8      	    3636	    322590 ns/op	   81933 B/op	       1 allocs/op
BenchmarkCutTracesSlice100-8     	    3628	    309431 ns/op	   81933 B/op	       1 allocs/op
BenchmarkCutTracesSlice1000-8    	    3114	    326222 ns/op	   81936 B/op	       1 allocs/op
BenchmarkCutTracesSlice5000-8    	    3135	    345734 ns/op	   81935 B/op	       1 allocs/op
BenchmarkCutTracesSlice10000-8   	    3338	    328694 ns/op	   81934 B/op	       1 allocs/op
PASS
```

However, after reviewing the code I realized I am wrong above.  The _normal_ situation is for most traces to be cut from the active map.  Switching to a slice implementation.  Thanks for the review!

modules/ingester/instance.go
```diff
@@ -317,6 +314,23 @@ func (i *instance) resetHeadBlock() error {
...
 
+func (i *instance) tracesToCut(cutoff time.Duration, immediate bool) []*trace {
+	i.tracesMtx.Lock()
+	defer i.tracesMtx.Unlock()
+
+	now := time.Now()
+	tracesToCut := make([]*trace, 0, len(i.traces))
+
+	for key, trace := range i.traces {
+		if now.Add(cutoff).After(trace.lastAppend) || immediate {
```
davecheney, line 325: Maybe move the now.Add cutoff computation to the place where now is declared.

joe-elliott: Good call.  Thanks for the second set of eyes.

modules/ingester/instance.go
```diff
@@ -318,11 +318,11 @@ func (i *instance) tracesToCut(cutoff time.Duration, immediate bool) []*trace {
 	i.tracesMtx.Lock()
 	defer i.tracesMtx.Unlock()
 
-	now := time.Now()
+	cutoffTime := time.Now().Add(cutoff)
 	tracesToCut := make([]*trace, 0, len(i.traces))
 
 	for key, trace := range i.traces {
-		if now.Add(cutoff).After(trace.lastAppend) || immediate {
+		if cutoffTime.After(trace.lastAppend) || immediate {
```
davecheney, line 325: if you move immediate to the front, it will short circuit the cutoffTime calculation. this may reduce the amount of time tracesMtx is held

modules/ingester/instance_test.go
```diff
@@ -255,6 +256,90 @@ func TestInstanceLimits(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			instance := defaultInstance(t, tempDir)
+
+			for _, trace := range tc.input {
+				fp := instance.tokenForTraceID(trace.traceID)
+				instance.traces[fp] = trace
+			}
+
+			err := instance.CutCompleteTraces(tc.cutoff, tc.immediate)
+			assert.NoError(t, err)
```
davecheney, line 327: I think you should use testify/require here, if err is not nil, the test should stop at this point

modules/ingester/instance.go
joe-elliott: immediate is almost never true.  it only occurs when the ingester is shutting down to force flush traces from the active trace map to disk.  i think this ordering is fine.

modules/ingester/instance_test.go
joe-elliott: agreed.

modules/ingester/instance.go
davecheney: fair enough, thanks for considering it.

davecheney (approved):

annanay25 (approved):
lgtm!

PR #452, opened 2021-01-12
Downgraded the distributor alert to warning

joe-elliott:
**What this PR does**:
The distributor alert in the mixin does not actually indicate any kind of degraded functionality in Tempo, but rather that something perhaps should be investigated.  Unlike previously thought unhealthy distributors does _not_ impact global rate limiting.

**Checklist**
- [ ] Tests updated
- [x] Documentation added


mdisibio (approved):

joe-elliott:
Merging even though CI is not passing.  We need to fix or drop the "Check Docs" step.

PR #430, opened 2020-12-23
Switch distributor->ingester communication to more efficient PushBytes method

mdisibio:
**What this PR does**:
This PR switches distributor->ingester communication of trace data to a more efficient PushBytes method.  This api is different in that it contains a slice of byte slices with pre-marshaled Batches ([][]byte).  The distributor is marshalling the trace data to a byte slice only once (instead of per ingester), and all data is delivered to the ingester in 1 gRPC call (instead of per trace).   This has large improvements in both cpu and memory when ReplicationFactor >= 2, but also non-trivial improvement for replication factor = 1.

**Background**
The main driver for tempo TCO is compute as the actual object storage is very cost effective.  Compute is roughly 50/50 for the distributor and ingester layers.  As the distributor is mainly a proxy that replicates traffic to the ingesters according to the replication factor this was higher than expected and seemed to be a good area for improvement.  Pprof benchmarking the distributor showed that most cpu and memory processing was related to proto marshalling and compression.  From reviewing distributor.Push, the current implementation had several deficiencies. For example, when Replication Factor = 2, a gRPC call is made to each ingester for each belonging trace. This incurs:
* 2x marshaling per trace
* 2x compression per trace
* 2x gRPC calls per trace

The new API signature reduces it to the following in theory:
* 1x marshaling per trace
* 2x compression per trace
* 1x gRPC call per ingester

In practice there are larger than expected savings, as less memory churn means less garbage collection.

**Performance Analysis**
Performance before and after were measured locally and in a dev cluster.   A useful docker-compose setup that configures the necessary replication factor is ~located here:  https://github.com/mdisibio/tempo-load-test/tree/master~ added to the /integration/microservices/ folder. This setup includes cadvisor, grafana, and a dashboard.

The main metric of interest is the compute _efficiency_, which is _spans / s / core_ across all distributor and ingester pods.  This metric is computed with promQL of (simplifying) `rate(tempo_receiver_accepted_spans) / 
rate(container_cpu_usage_seconds_total)` 

The current e2e and benchmarking tools were not straightforward to measure this, hence the creation of the linked compose setup.

Local testing improvements:
* Spans/s/core went from 2700 -> 10-15K, 5x+
* In dev k8s cluster, went from 2200 -> 5K, 2.5x+

**Screenshots:**

**Before:  **

Screen Shot 2020-12-23 at 2 55 13 PM
  
**After**
Screen Shot 2020-12-23 at 3 06 10 PM

** K8s cluster**
[...]

PR #331, opened 2020-11-10
Add per tenant bytes counter

dgzlopes:
**What this PR does**:
This PR adds a `per tenant bytes counter` metric. I went with the distributor approach!

The metric is increased when we cut the complete traces.

**Which issue(s) this PR fixes**:
Fixes #223

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #453, opened 2021-01-12
Added bytes ingested metric per tenant

### LINKED ISSUE #410: Add per tenant bytes before limits
We need to count bytes pushed per tenant before limits are applied.

**Additional context**
We have added #331 bytes per tenant committed to the WAL, but we also need bytes received per tenant before limits are applied.

joe-elliott:
**What this PR does**:
Adds `tempo_distributor_bytes_received_total` per tenant.  This metric contains the uncompressed proto bytes received per tenant as a counter.  Previously I had been concerned about the performance impact of this addition, but after @mdisibio 's work in #430 I think this will be fine.

Fixes #410

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


annanay25 (approved):
Is the metric introduced in #331 still valuable after this?

joe-elliott:
> Is the metric introduced in #331 still valuable after this?

Maybe?  It's certainly less valuable.  Make an issue to remove?

PR #457, opened 2021-01-14
Add merging logic at Query Frontend for exhaustive search

annanay25:
**What this PR does**:
Merge responses from each of the sharded queries to return complete traces that might be split across blocks.
Continue returning the higher response code as is appropriately indicative of the failure (if any).

_(Small change unrelated to title: Add a make command to lint all go files in the repo)_

Fixes #na!

**Checklist**
- [x] Tests updated
- NA Documentation added


modules/frontend/querysharding.go
```diff
@@ -167,3 +154,41 @@ func DoRequests(ctx context.Context, downstream Handler, reqs []*http.Request) (
 
 	return resps, firstErr
 }
+
+func mergeResponses(rrs []RequestResponse) (*http.Response, error) {
```
joe-elliott, line 158: +1 to breaking this out and adding tests.

modules/frontend/querysharding.go
```diff
@@ -167,3 +154,41 @@ func DoRequests(ctx context.Context, downstream Handler, reqs []*http.Request) (
...
+	var errCode = http.StatusOK
+	var errBody io.ReadCloser
+	var combinedTrace []byte
+	for _, rr := range rrs {
+		if rr.Response.StatusCode == http.StatusOK {
+			body, err := ioutil.ReadAll(rr.Response.Body)
```
joe-elliott, line 164: do we need a `rr.Response.Body.Close()` or does `iotuil.ReadAll` handle that?

modules/frontend/querysharding.go
```diff
@@ -167,3 +154,41 @@ func DoRequests(ctx context.Context, downstream Handler, reqs []*http.Request) (
...
+			if err != nil {
+				return nil, errors.Wrap(err, "error ready response body at query frontend")
```
joe-elliott, line 166: nit:  "error reading..."

modules/frontend/querysharding.go
```diff
@@ -167,3 +154,41 @@ func DoRequests(ctx context.Context, downstream Handler, reqs []*http.Request) (
...
+			}
+
+			if len(combinedTrace) == 0 {
+				combinedTrace = body
+			} else {
+				combinedTrace = util.CombineTraces(combinedTrace, body)
+			}
+		}
+		if rr.Response.StatusCode > errCode {
```
joe-elliott, line 175: I struggle with this.  Wouldn't we rather return 400 then 500?   If both are true we'd rather them retry thinking it's on their side since the 500 is probably transient (and may have been caused by the bad request).

@mdisibio any ideas for the best ways to combine http status codes?

modules/frontend/querysharding.go
```diff
@@ -167,3 +154,41 @@ func DoRequests(ctx context.Context, downstream Handler, reqs []*http.Request) (
...
+			errCode = rr.Response.StatusCode
+			errBody = rr.Response.Body
+		}
+	}
+
+	if len(combinedTrace) == 0 {
```
joe-elliott, line 181: If one of the queries 500s, but another query returns some data we will return 200 OK along with whatever was returned from the one query.

I feel like we should 500 in that case.

joe-elliott (commented):
Added some questions/thoughts.  @mdisibio interested in your take as well.

@annanay25  How does this handle context timeouts?

modules/frontend/querysharding.go
annanay25: Nice catch!

modules/frontend/querysharding.go
annanay25: I feel like 5xx should always take precedence. From your other comment below, if we're return 5xx given one of the shards failed but another return 2xx, we should follow the same principle here for consistency right?

In the case below, I thought it would be good to atleast see a partial result from one of the successful shards even if some shards resulted in 5xx, but we can standardise on this.

mdisibio: Agree with 5xx taking precedence.  It seems better to propagate errors clearly, or else it could be hard to detect issues and not realize partial trace data was returned.    Regarding transient errors:   yes some 5xx errors are retryable and sounds ok to require caller to retry.

modules/frontend/querysharding.go
mdisibio: Agree, propagating errors clearly seems best, rather than masking over the error and returning partial trace.

modules/frontend/querysharding.go
```diff
@@ -68,29 +72,12 @@ func (s shardQuery) Do(r *http.Request) (*http.Response, error) {
...
-			errBody = rr.Response.Body
-		}
-	}
-
-	return &http.Response{
-		StatusCode: errCode,
-		Body:       errBody,
-		Header:     http.Header{},
-	}, nil
+	return mergeResponses(rrs)
```
mdisibio, line 80: If we decide that 5xx always takes precedence, it might be clearer to split that into a separate func like `checkForErrors(rrs)`.

annanay25: The function also combines traces that might be split across blocks, so `mergeResponses` seems more appropriate?

modules/frontend/querysharding.go
```diff
@@ -167,3 +154,43 @@ func DoRequests(ctx context.Context, downstream Handler, reqs []*http.Request) (
...
+			}
+		}
+		if rr.Response.StatusCode > errCode {
+			errCode = rr.Response.StatusCode
+			errBody = rr.Response.Body
+		}
+	}
+
+	// Always give precedence to 5xx errors and propagate them to the user so they can retry the query
+	if len(combinedTrace) == 0 || errCode > 499 {
```
joe-elliott, line 183: Should we translate all errorCodes > 400 (not 404) into 500?  Consider 400 (bad request), (401) unauthorized, etc...  Since we are building and executing the requests I think if we get 400 this should actually be a 500 externally.

This logic could use some ironing out for clarity:
```
errCode := 404
for _, rr := range rrs {
    if rr.statusCode = 200 {
      ...
    } else if rr.statusCode != 404 {
      errCode = rr.Response.StatusCode
      errBody = rr.Response.Body
    }
}

if errCode == 0 {
  // yay!
} else if errCode == 404 {
  //  404
} else {
  // 500 with whatever errBody is
}
```

modules/frontend/querysharding.go
```diff
@@ -167,3 +154,43 @@ func DoRequests(ctx context.Context, downstream Handler, reqs []*http.Request) (
...
+
+			if len(combinedTrace) == 0 {
+				combinedTrace = body
+			} else {
+				combinedTrace = util.CombineTraces(combinedTrace, body)
+			}
+		}
+		if rr.Response.StatusCode > errCode {
+			errCode = rr.Response.StatusCode
+			errBody = rr.Response.Body
```
joe-elliott, line 178: is this a resource leak b/c we're never closing this body?

modules/frontend/querysharding.go
annanay25: > Since we are building and executing the requests I think if we get 400 this should actually be a 500 externally

Hrm, this is not entirely true. At the moment, we pass the traceID as-is and verify only at the querier level. Should we move the traceID validation to the Query Frontend? We'll have to think some about the `/querier/api/traces/{traceID}` endpoint as well, that we want to keep for debugging.

400/401 behaviour will be consistent across all the shards/queriers. If one of them throws 400 all of them will. In that case we will still return 400 after merging responses.

modules/frontend/querysharding.go
annanay25: We're passing the body as-is to the tempo-query. It would be tempo-query's responsibility to handle the close correctly.

modules/frontend/querysharding.go
mdisibio: This logic makes sense, but one concern is losing the original error code which would make diagnosing a final 500 error harder.  Should there be logging or other here?
```
else if rr.statusCode != 404 {
      errCode = rr.Response.StatusCode
      errBody = rr.Response.Body
    }
```

modules/frontend/querysharding.go
joe-elliott: the body is remarshaled and then passed over GRPC.  it will never have access to that actual object.

modules/frontend/querysharding.go
annanay25: This was fixed as discussed offline. Also PTAL at the new test.

modules/frontend/querysharding.go
annanay25: The http request is constructed here and not over a client connection - https://github.com/cortexproject/cortex/blob/faf92e8f3dbedfa769fbd6a3ff6e0fbd127694f5/pkg/frontend/transport/roundtripper.go#L38

We should be good passing this down as is..

modules/frontend/querysharding.go
```diff
@@ -167,3 +155,49 @@ func DoRequests(ctx context.Context, downstream Handler, reqs []*http.Request) (
...
+				combinedTrace = util.CombineTraces(combinedTrace, body)
+			}
+		} else if rr.Response.StatusCode != http.StatusNotFound {
+			errCode = rr.Response.StatusCode
+			errBody = rr.Response.Body
+		}
+	}
+
+	if errCode == http.StatusOK {
+		if len(combinedTrace) > 0 {
```
joe-elliott, line 183: It's a little weird to conflate len == 0 with 404.  I don't think it's technically possible to push a 0 byte object into tempo but something to keep in mind.

Please add a short comment above this line that acknowledges this and says it's impossible to push a 0 byte trace so this logic is valid.

joe-elliott (approved):
one small nit.  let's go!

modules/frontend/querysharding.go
annanay25: Agree with this. I've added a counter for the 404s. Removing any ambiguity.

PR #460, opened 2021-01-15
Versioned Blocks: Backend Refactoring Finalish

joe-elliott:
**What this PR does**:
This is the last major backend refactoring PR to enable versioned blocks which will eventually allow us to support compressed blocks (and other future block changes).

- Moved block specific code to `encoding/v0`
- Moved "shared" objects to `encoding/index`
  - This primarily exists to break a dependency loop.  We can't put them in `./encoding` b/c `./encoding/v0` needs to reference them.  We could put them in `./encoding/v0`, but I like them having independence from the versions.
- wal directly references and relies on `encoding/v0`.  I think this is fine for now.  The wal may always write "v0" blocks.
- `encoding/versioned.go` provides a shim between other code and version specific blocks.  Not 100% satisfied with the shim methods like `writeBlockData`, but was unable to come up with anything I liked substantially better.

The current plan is to make a `encoding/v1` that references the v0 code and only wraps the necessary functionality to support compression.  This might just `BufferedAppender` and `BackendIterator`.  There is still a bit of cleanup to go on this PR, but I think the ideas are largely here and I'm interested in feedback.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


tempodb/encoding/index/record.go
```diff
@@ -0,0 +1 @@
+package index
```
mdisibio: Empty file?

cmd/tempo-cli/cmd-list-block.go
```diff
@@ -59,7 +59,15 @@ func dumpBlock(r tempodb_backend.Reader, c tempodb_backend.Compactor, tenantID s
 	if checkDupes {
 		fmt.Println("Searching for dupes ...")
 
-		iter, err := encoding.NewBackendIterator(tenantID, id, 10*1024*1024, r)
+		block, err := encoding.NewBackendBlock(&tempodb_backend.BlockMeta{
+			TenantID: tenantID,
+			BlockID:  id,
```
mdisibio, line 64: This is missing the version.  The original meta is available above, and it probably makes sense to add to the unifiedMeta and display it.

mdisibio:
As discussed, might make sense to rename /encoding/index to /encoding/common

tempodb/compactor_bookmark_test.go
```diff
@@ -73,7 +74,13 @@ func TestCurrentClear(t *testing.T) {
 	blockID = complete.BlockMeta().BlockID
 	rw := r.(*readerWriter)
 
-	iter, err := encoding.NewBackendIterator(testTenantID, blockID, 10, rw.r)
+	block, err := encoding.NewBackendBlock(&backend.BlockMeta{
```
mdisibio, line 77: I think this can use `complete.BlockMeta()`

cmd/tempo-cli/cmd-list-block.go
joe-elliott: good catch.

mdisibio (approved):
Good changes towards versioned blocks.

annanay25 (approved):
Lots of changes and looking good!

PR #461, opened 2021-01-15
Fix Ingester Query Latency Spikes

joe-elliott:
**What this PR does**:
Reduces the locking range on the block mutex during an active traces flush.  This significantly improves Ingester query latency:

image

I was surprised to not immediately see similar improvement at the query frontend.  However, I did find that the impact of querier scaling is now much more apparent with the ingester fix:

image

First annotation is the new ingester being deployed.  Second annotation is scaling the queriers to 10.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


annanay25 (approved):
lgtm!

Issue #418, opened 2020-12-16
Add WAL Checksums


joe-elliott:
**Is your feature request related to a problem? Please describe.**
Currently our only indication that there is corruption in the WAL is that proto fails to Unmarshal.  Add a checksum here so we can more robustly detect corruption issues.

**Describe the solution you'd like**
- Add per record checksums
- Verify checksums on replay.  Research whether a single corrupt record can be skipped
- Version the WAL file for future changes

**Additional context**
Unfortunately this may require versioning and abandoning previous WAL files on upgrade.  We need to determine if there's anyway to handle the old WAL file format and the new.   Given that we're still in 0.x I'm ok with the upgrade including a dropped WAL as long as this is documented as well as ways to preserve data.

fitzoh:
>Add per record checksums

Just to clarify, this would be serialized as part of an `Object`, right?

And the layout [here](https://github.com/grafana/tempo/blob/3703d8b25492c2f2bc1e875d40302664179e57e7/tempodb/encoding/object.go#L15) would maybe go from
`| total length (uint32) | id length (uint32) | id (variable) | object bytes (variable) |`
to something like
`| total length (uint32) | id length (uint32) | crc checksum (uint32) | id (variable) | object bytes (variable) |`



>We need to determine if there's anyway to handle the old WAL file format and the new.

This feels like it might be a little gross, but would it make sense to maintain multiple WAL format parser?
Thinking something like this:
```golang
var LatestObjectBinaryFormat = objectBinaryFormatV1{}

type ObjectBinaryFormat interface {
	MarshalObjectToWriter(id ID, b []byte, w io.Writer) (int, error)
	UnmarshalObjectFromReader(r io.Reader) (ID, []byte, error)
	UnmarshalAndAdvanceBuffer(buffer []byte) ([]byte, ID, []byte, error)
}

// | total length (uint32) | id length (uint32) | id (variable) | object bytes (variable) |
type objectBinaryFormatV0 struct {}
// | total length (uint32) | id length (uint32) | crc checksum (uint32) | id (variable) | object bytes (variable) |
type objectBinaryFormatV1 struct {}
```

Finally, any thoughts on how to manage the WAL version? Encode in the filename? Magic bytes at the start of the file? Something else?

joe-elliott:
> Just to clarify, this would be serialized as part of an Object, right?

Yes, I believe per object is the right way to go.

> This feels like it might be a little gross, but would it make sense to maintain multiple WAL format parser?

[...]

PR #463, opened 2021-01-19
Improve WAL Replay Performance

### LINKED ISSUE #409: Ingesters with many WAL files will concatenate them on restart
**Describe the bug**
If an ingester loses its ability to flush to object storage it will begin building a set of WAL files that it will never clear.  On restart these WAL file will all be replayed here:

https://github.com/grafana/tempo/blob/e9892bd70f24446c488c2a3169b8c764ec8ffa5a/modules/ingester/ingester.go#L252

During this replay the ingester is not cutting new blocks.  This means that all WAL files will be concatenated into one large WAL file which may be difficult to flush.

joe-elliott:
**What this PR does**:
For simplicity the current WAL replay code loads a file, iterates through all records and replays them into an ingester instance.  This creates two issues:

- We are doubling our disk needs for a brief time period while the WAL replays and rebuilds itself in a second file.  In our internal ops cluster we have seen this lead to failed WAL replays.
- There is a lot of overhead to reading an entire multi GB file and rewriting it back to disk.

This proposed fix iterates through the WAL as is and simply rebuilds the index in memory for the AppendBlock to use directly instead of rebuilding a new file.

One possible negative of this fix is that the current pattern would allow us to more easily skip or stop once records stopped replaying successfully perhaps due to corruption.  The method proposed in this PR would either need to throw away or keep the entire WAL.  (Perhaps in the future we could find a way to trunc?)

Fixes #409 

This PR fixes #409 by simply failing to start if there are two WAL files for a given instance.  In this case an operator would need to wipe out one of the files to allow the other to replay.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

**PR has been marked as draft as I still need to add tests/changelog**


mdisibio:
Agree with this change overall.   In reusing the wal file, should we implement some form of validation/data corruption handling?

joe-elliott:
> Agree with this change overall. In reusing the wal file, should we implement some form of validation/data corruption handling?

Yup.  Plan to per record add checksums:
https://github.com/grafana/tempo/issues/418

joe-elliott:
Closing in favor of #668

PR #476, opened 2021-01-26
Feat: s3 del bucket exists logic

swartz-k:
**What this PR does**:
Delete S3 Bucket Exists logic to support more storage like AlibabaClous OSS

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/backend/s3/s3.go
+++ b/tempodb/backend/s3/s3.go
@@ -101,14 +101,6 @@ func New(cfg *Config) (backend.Reader, backend.Writer, backend.Compactor, error)
 
 	// TODO: add custom transport with instrumentation.
 	//client.SetCustomTransport(minio.DefaultTransport(!cfg.Insecure))
-	exists, err := core.BucketExists(context.Background(), cfg.Bucket)
-	if err != nil {
-		return nil, nil, nil, fmt.Errorf("unexpected error from BucketExists on %s: %w", cfg.Bucket, err)
-	}
-
-	if !exists {
-		return nil, nil, nil, fmt.Errorf("s3 Bucket %s does not exist", cfg.Bucket)
-	}
 
 	// try listing objects
 	_, err = core.ListObjects(cfg.Bucket, "", "", "/", 0)


modules/ingester/instance.go
```diff
@@ -332,7 +332,17 @@ func (i *instance) tracesToCut(cutoff time.Duration, immediate bool) []*trace {
 func (i *instance) writeTraceToHeadBlock(id common.ID, b []byte) error {
 	i.blocksMtx.Lock()
 	defer i.blocksMtx.Unlock()
-
+	if i.headBlock == nil {
```
joe-elliott, line 335: Any particular reason you added this?  I would prefer keeping the PR focused on the change to remove BucketExists.

tempodb/wal/append_block.go
```diff
@@ -89,6 +90,9 @@ func (h *AppendBlock) Complete(w *WAL, combiner common.ObjectCombiner) (*encodin
 }
 
 func (h *AppendBlock) Find(id common.ID, combiner common.ObjectCombiner) ([]byte, error) {
+	if h.appender == nil {
```
joe-elliott, line 93: Same comment here.

joe-elliott (changes requested):

annanay25:
Agree with Joe on this. Let's focus on removing the `BucketExists` call in this PR, and we can review the other changes in a different PR.

modules/ingester/instance.go
swartz-k: Yes. I will meet some CI errors if those code not add.

joe-elliott: Ah, I think @mdisibio discovered and is working on this issue.  Thank you for also trying to address it.

swartz-k:
> Agree with Joe on this. Let's focus on removing the `BucketExists` call in this PR, and we can review the other changes in a different PR.

Right. I misunderstand it. Sometimes CI will fail.

joe-elliott (approved):

PR #477, opened 2021-01-26
Add support to cut an active block based on total bytes

### LINKED ISSUE #474: Add support to cut an active block based on total bytes
Currently Tempo allows for cutting the active block based on time or total traces.  However trace size can vary wildly.  This should allow for cutting the active block by total current bytes.

https://github.com/grafana/tempo/blob/master/modules/ingester/config.go#L20-L21

Add a config option (`max_block_bytes` ?) that cuts a block when the threshold is reached.  I think we should remove `traces_per_block` in favor of this setting.

mdisibio:
**What this PR does**:
Adds config value and logic to cut ingester headblock based on size.  Default is 1GB.  

Fixes #474 

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/tempo/website/configuration/_index.md
+++ b/docs/tempo/website/configuration/_index.md
@@ -53,6 +53,8 @@ ingester:
             replication_factor: 2   # number of replicas of each span to make while pushing to the backend
     trace_idle_period: 20s          # amount of time before considering a trace complete and flushing it to a block
     traces_per_block: 100000        # maximum number of traces in a block before cutting it
+    max_block_bytes: 1_000_000_000  # maximum block size before cutting it
+    max_block_duration: 1h          # maximum length of time before cutting a block
 ```
 
 ## Query Frontend
--- a/docs/tempo/website/configuration/_index.md
+++ b/docs/tempo/website/configuration/_index.md
@@ -52,7 +52,7 @@ ingester:
         ring:
             replication_factor: 2   # number of replicas of each span to make while pushing to the backend
     trace_idle_period: 20s          # amount of time before considering a trace complete and flushing it to a block
-    max_block_bytes: 1_000_000_000  # maximum block size before cutting it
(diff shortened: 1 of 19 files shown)


docs/tempo/website/configuration/_index.md
```diff
@@ -53,6 +53,8 @@ ingester:
             replication_factor: 2   # number of replicas of each span to make while pushing to the backend
     trace_idle_period: 20s          # amount of time before considering a trace complete and flushing it to a block
     traces_per_block: 100000        # maximum number of traces in a block before cutting it
+    max_block_bytes: 1_000_000_000  # maximum block size before cutting it
```
achatterjee-grafana, line 56:
```suggestion
    max_block_bytes: 1_000_000_000  # maximum size of a block before cutting it
```

achatterjee-grafana (approved):
Added one suggestion. Otherwise LGTM

tempodb/encoding/common/types.go
```diff
@@ -31,6 +31,7 @@ type Appender interface {
 	Complete()
 	Records() []*Record
 	Length() int
```
joe-elliott, line 33: We should be close to being able to remove the `Length()` function on this interface.  Double check and see how hard it would be? 

If nothing else the AppendBlock doesn't need the Length() method anymore and it can be removed.

joe-elliott (commented):
Nice.  This will make Tempo much easier to operate.

Have a couple of small questions, but other than that looks good.

tempodb/encoding/common/types.go
mdisibio: Yes it is mostly unused now.  The remaining use is for compactor max records per block.  As discussed, we should probably move that to a file size-based approach as well.

joe-elliott (approved):

PR #480, opened 2021-01-27
Correct the translation of protobuf to json at query frontend

annanay25:

**What this PR does**:
- This PR is a bugfix for #457, where we were incorrectly passing a json marshalled byte slice of traces to `combineTrace` which only accepts protobuf marshalled format.
- Changed signature of `combineTrace` to also return error, as it helps with bubbling it up as required
- Tests added accordingly

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- NA Documentation added

### DIFF
--- a/modules/compactor/compactor.go
+++ b/modules/compactor/compactor.go
@@ -149,7 +149,8 @@ func (c *Compactor) Owns(hash string) bool {
 }
 
 func (c *Compactor) Combine(objA []byte, objB []byte) []byte {
-	return tempo_util.CombineTraces(objA, objB)
+	combinedTrace, _ := tempo_util.CombineTraces(objA, objB)
+	return combinedTrace
 }
 
 func (c *Compactor) waitRingActive(ctx context.Context) error {
--- a/modules/compactor/compactor.go
+++ b/modules/compactor/compactor.go
@@ -149,7 +149,10 @@ func (c *Compactor) Owns(hash string) bool {
 }
 
 func (c *Compactor) Combine(objA []byte, objB []byte) []byte {
-	combinedTrace, _ := tempo_util.CombineTraces(objA, objB)
+	combinedTrace, err := tempo_util.CombineTraces(objA, objB)
+	if err != nil {
+		level.Error(util.Logger).Log("msg", "error combining trace protos", "err", err.Error())
+	}
 	return combinedTrace
 }
 
--- a/modules/frontend/frontend.go
+++ b/modules/frontend/frontend.go
@@ -13,6 +13,7 @@ import (
 	ot_log "github.com/opentracing/opentracing-go/log"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/weaveworks/common/middleware"
 	"github.com/weaveworks/common/user"
 
 	"github.com/grafana/tempo/pkg/util"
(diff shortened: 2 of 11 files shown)


modules/compactor/compactor.go
```diff
@@ -149,7 +149,8 @@ func (c *Compactor) Owns(hash string) bool {
 }
 
 func (c *Compactor) Combine(objA []byte, objB []byte) []byte {
-	return tempo_util.CombineTraces(objA, objB)
+	combinedTrace, _ := tempo_util.CombineTraces(objA, objB)
```
joe-elliott, line 152: let's log this error

modules/frontend/frontend.go
```diff
@@ -54,7 +55,14 @@ func NewTripperware(cfg Config, logger log.Logger, registerer prometheus.Registe
 			r = r.WithContext(ctx)
 			resp, err := rt.RoundTrip(r)
 
-			level.Info(logger).Log("method", r.Method, "url", r.URL.RequestURI(), "duration", time.Since(start).String(), "error", err == nil)
+			traceID, _ := middleware.ExtractTraceID(ctx)
+			var statusCode int
+			if resp == nil {
```
joe-elliott, line 60: definitely a weird situation where resp is nil.  for clarity:

```
statusCode := 500
if resp != nil {
  statusCode = resp.StatusCode
}
```

modules/frontend/querysharding.go
```diff
@@ -174,9 +196,31 @@ func mergeResponses(rrs []RequestResponse) (*http.Response, error) {
 	}
 
 	if errCode == http.StatusOK {
+		if marshallingFormat == util.ProtobufTypeHeaderValue {
```
joe-elliott, line 199: if there's a sane way to combine this with the querier code that also uses the Accept header to choose how to marshal, please do.  otherwise this is fine.

joe-elliott: Maybe could be cleaned up like:

```
var r Reader

if format == Proto {
 r := protoThing
} else {
  r := jsonThing
}

return &http.Response{  
  Body: Use r
 }
```

modules/ingester/instance.go
```diff
@@ -337,7 +337,8 @@ func (i *instance) writeTraceToHeadBlock(id common.ID, b []byte) error {
 }
 
 func (i *instance) Combine(objA []byte, objB []byte) []byte {
-	return util.CombineTraces(objA, objB)
+	combinedTrace, _ := util.CombineTraces(objA, objB)
```
joe-elliott, line 340: log the error if it's set

pkg/util/trace_test.go
```diff
@@ -71,7 +71,8 @@ func TestCombine(t *testing.T) {
 	}
```
joe-elliott, line 71: add a test that fails and check that err != nil on return.

joe-elliott (changes requested):
Couple of changes requested and then the linter is mad you didn't check err on some tests.

modules/compactor/compactor.go
annanay25: Hmm, as of now it will be logged at `CombineTrace` and in the parent fn, should we remove logging at `CombineTrace` altogether?

modules/ingester/instance.go
annanay25: https://github.com/grafana/tempo/pull/480/files#r565891495

modules/frontend/querysharding.go
annanay25: Have been thinking about a function like so - 

```
func MarshalTraceObject(trace *tempopb.Trace, marshallingFormat string) ([]byte, error)
```

But while this could be useful at the querier, at the frontend we receive the response in bytes, its not beneficial to unmarshal that into the object and then remarshal it at the function if we're going to return proto bytes anyway..

modules/compactor/compactor.go
joe-elliott: yeah, if we're going to return an error from CombineTraces(), let's log it outside that function.  honestly it's weird that a utility function like CombineTraces logs anything.

joe-elliott (approved):
One small request, but lgtm.

PR #482, opened 2021-01-27
Added compactor service

joe-elliott:
**What this PR does**:
Adds a compactor service.  This would useful if one were to want to serve the `/compactor/ring` page for instance.


annanay25 (approved):
lgtm!

PR #489, opened 2021-01-29
Query path - Add exhaustive search to combine traces split across blocks

### LINKED ISSUE #407: Appending spans to flushed traces
In a system in-place at our work, there can be a delay of seconds to minute or hours between successive spans being captured for a given trace. With the current setup of Tempo, it seems to be that once a trace is flushed to the backend storage, and new spans received for that trace overwrite the previous data.

We have initially set `trace_idle_period` to `10s`. I'm guessing we could bump this up to `1h` or more but I'm not sure what the consequences of that would be.

Would it be possible to add some functionality which can check for the existence of a previously flushed trace and append newly received spans to it instead of overwriting them?

I'm not sure how Tempo structures data in the storage backend, but maybe this could be done at the query layer by searching backwards for any additional spans for a given trace id?

annanay25:
**What this PR does**:
Add exhaustive search at the querier to go through all blocks and combine traces.
Involves changes 
- to `pool.RunJobs()` to not stop after the first hit
- and at the querier to search backend blocks even if a trace is found in an ingester

Fixes #407 

**Checklist**
- [x] Tests updated
- NA Documentation added


tempodb/tempodb.go
```diff
@@ -256,7 +258,22 @@ func (rw *readerWriter) Find(ctx context.Context, tenantID string, id common.ID,
...
+	if len(partialTraces) == 0 {
+		return nil, metrics, err
+	}
+
+	// merge all partial trace bytes into partialTraces[0]
+	for i := range partialTraces {
+		if i == 0 {
+			continue
+		}
+		partialTraces[0], err = tempo_util.CombineTraces(partialTraces[0], partialTraces[i])
```
mdisibio, line 270: This will unmarshal/marshal the trace repeatedly for every partial.  It would be more efficient use CombineTraceProtos.  Probably not significant in practice (low likelihood).

modules/querier/querier.go
```diff
@@ -184,37 +184,44 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
+	metricQueryReads.WithLabelValues("bloom").Observe(float64(metrics.BloomFilterReads.Load()))
+	metricQueryBytesRead.WithLabelValues("bloom").Observe(float64(metrics.BloomFilterBytesRead.Load()))
+	metricQueryReads.WithLabelValues("index").Observe(float64(metrics.IndexReads.Load()))
+	metricQueryBytesRead.WithLabelValues("index").Observe(float64(metrics.IndexBytesRead.Load()))
+	metricQueryReads.WithLabelValues("block").Observe(float64(metrics.BlockReads.Load()))
+	metricQueryBytesRead.WithLabelValues("block").Observe(float64(metrics.BlockBytesRead.Load()))
+
+	// combine out with completeTrace
+	var spanCountA, spanCountB, spanCountTotal int
+	completeTrace, spanCountA, spanCountB, spanCountTotal = tempo_util.CombineTraceProtos(completeTrace, storeTrace)
```
joe-elliott, line 219: rename `completeTrace` => `ingesterTrace` for clarity?

tempodb/tempodb.go
joe-elliott: Yeah, I agree with this.  Let's have `store.Find` return a slice of all objects it finds.  Then the querier can marshal each to proto once and combine.

tempodb/pool/pool.go
```diff
@@ -118,16 +118,18 @@ func (p *Pool) RunJobs(ctx context.Context, payloads []interface{}, fn JobFunc)
...
+	// read all from results channel
+	var msg [][]byte
+	for res := range resultsCh {
+		msg = append(msg, res)
 	}
 
 	// ignore err if msg != nil.  otherwise errors like "context cancelled"
 	//  will take precedence over the err
-	if msg != nil {
+	if len(msg) > 0 {
```
joe-elliott, line 132: i don't think this is true anymore.  we were taking into account the context cancelled errors b/c we were purposefully cancelling context if a single job completed successfully.   now that we are allowing all jobs to finish I think we should let errors take precedence.  WCGW? :)

joe-elliott (commented):
Added some thoughts.  Shouldn't there be other places with RunJobs?  I'm only seeing one place you changed.  

The places @mdisibio is currently removing in his PR?

tempodb/tempodb.go
mdisibio: Created a variadic version of CombineTraces which I think is what we need here:  https://github.com/mdisibio/tempo/blob/combinetracesv/pkg/util/trace.go#L17    This preserves all the error handling/logic of the current function, rather that doing it here in the Find method.

annanay25: I'm going to go ahead with the current function signature in this PR to avoid including too many changes, but I agree that we need this @mdisibio. I'll work on it in a follow up PR.

tempodb/pool/pool.go
annanay25: Yeah since we are no longer cancelling contexts on the individual jobs, I agree `err` should be given precedence here.

modules/querier/querier.go
```diff
@@ -184,36 +184,44 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
-		completeTrace = out
-		metricQueryReads.WithLabelValues("bloom").Observe(float64(metrics.BloomFilterReads.Load()))
-		metricQueryBytesRead.WithLabelValues("bloom").Observe(float64(metrics.BloomFilterBytesRead.Load()))
-		metricQueryReads.WithLabelValues("index").Observe(float64(metrics.IndexReads.Load()))
-		metricQueryBytesRead.WithLabelValues("index").Observe(float64(metrics.IndexBytesRead.Load()))
-		metricQueryReads.WithLabelValues("block").Observe(float64(metrics.BlockReads.Load()))
-		metricQueryBytesRead.WithLabelValues("block").Observe(float64(metrics.BlockBytesRead.Load()))
+		completeTrace, spanCountA, spanCountB, spanCountTotal = tempo_util.CombineTraceProtos(completeTrace, storeTrace)
 	}
+	span.LogFields(ot_log.String("msg", "combined trace protos from ingesters and store"),
```
joe-elliott, line 221: since we're iterating through a bunch of partial traces and combining them, this spansCountA/B/Total will only be the final iteration which isn't very useful.  I do like a span tag here to give information about the "size" of the combined trace.

maybe just do span total?  or bytes total by adding up the buffer lengths?

joe-elliott (approved):
one small comment, but looks good.  let's go!

modules/querier/querier.go
annanay25: The store returns marshalled traces but ingester client marshals it into trace object. I wouldn't want to add a marshalling step just to get this span tag, so going with total number of spans.

PR #497, opened 2021-02-02
OTEL Collector/Multitenant Example

joe-elliott:
**What this PR does**:
- Adds an example with the otel collector and multitenancy

### DIFF
--- a/docs/tempo/website/guides/multitenancy.md
+++ b/docs/tempo/website/guides/multitenancy.md
@@ -1,0 +1,51 @@
+---
+title: Multitenancy
+---
+
+Tempo is a multitenant distributed tracing backend. It supports multitenancy through the use
+of a header: `X-Scope-OrgID`. This guide details how to setup or disable multitenancy.
+
+## Multitenancy
+
+If you're interested in setting up multitenancy, please consult the [multitenant example](https://github.com/grafana/tempo/tree/master/example/docker-compose/docker-compose.multitenant.yaml)
+in the repo. This example uses the following settings to achieve multitenancy in Tempo:
+
+- Configure the OTEL Collector to attach the X-Scope-OrgID header on push:
+```
+exporters:
+  otlp:
+    headers:
+      x-scope-orgid: foo-bar-baz
+```
+- Configure the Tempo datasource in Grafana to pass the tenant as a bearer token. Yes, this is weird. It works b/c it is the only header that Jaeger can be configured to pass to its GRPC plugin.
+```
+- name: Tempo-Multitenant
+  jsonData:
+    httpHeaderName1: 'Authorization'
+  secureJsonData:
+    httpHeaderValue1: 'Bearer foo-bar-baz'
+```
+- Configure Jaeger Query to pass the bearer token to its backend.
+```
+--query.bearer-token-propagation=true
+```
+
+## Important Notes
+
+- Multitenancy on ingestion is currently [only working](https://github.com/grafana/tempo/issues/495) with GPRC and this may never change. It is strongly recommended to use the OpenTelemetry Collector to support multitenancy as described above.
+- The way the read path is configured is temporary and should be much more straightforward once the [tempo-query dependency is removed](https://github.com/grafana/tempo/issues/382).
+
+## Disabling Multitenancy
+Most Tempo installations will be single tenant. If this is desired simply set the following config
+value on all Tempo components:
+```
+auth_enabled: false
+```
+
+or from the command line:
+```
+--auth.enabled=false
+```
+
+This option will force all Tempo components to ignore the `X-Scope-OrgID` header and use the hardcoded
+value of `single-tenant`.
--- a/example/docker-compose/etc/otel-collector.yaml
+++ b/example/docker-compose/etc/otel-collector.yaml
@@ -1,0 +1,13 @@
+receivers:    
+  jaeger:     
+    protocols:
+      thrift_http:
+exporters:
+  otlp:
+    endpoint: tempo:55680
+    insecure: true
+service:
+  pipelines:
+    traces:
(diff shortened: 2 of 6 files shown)


example/docker-compose/etc/otel-collector.yaml
```diff
@@ -0,0 +1,15 @@
+receivers:    
+  jaeger:     
+    protocols:
+      thrift_http:
+exporters:
+  otlp:
+    endpoint: tempo:55680
+    insecure: true
+    headers:
+      x-scope-orgid: authed-tenant
```
mdisibio, line 10: Is there a value that might be clearer than `authed-tenant` that it is a placeholder for a real value?  Thinking something like `foo` or `123`

annanay25:
Much needed addition!

example/docker-compose/etc/otel-collector.yaml
```diff
@@ -0,0 +1,15 @@
...
+service:
+  pipelines:
```
annanay25, line 12: There is some ongoing discussion on label based routing. In that case, is it possible to create multiple pipelines, one for every tenant like so:

```
exporters:
  otlp-1:
    endpoint: tempo:55680
    headers:
      x-scope-orgid: authed-tenant-1
  otlp-2:
    endpoint: tempo:55680
    headers:
      x-scope-orgid: authed-tenant-2

processor:
  <label-based-routing-processor>

service:
  pipelines:
    traces-1:
      receivers: [jaeger]
      processors: <label-based-routing-processor>
      exporters: [otlp-1] 
    traces-2:
      receivers: [jaeger]
      processors: <label-based-routing-processor>
      exporters: [otlp-2] 
```

..that can give us real multitenant behaviour?

joe-elliott: You can already do it with the filter processor:  https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/filterprocessor

Although you'd have to be careful with your filters and tags so you didn't drop any spans.

annanay25: Can we document this option as well?

joe-elliott: did some digging and sadly the filter processor only applies to metrics.  perhaps it will be extended at some point and we can document.

docs/tempo/website/guides/multitenancy.md
```diff
@@ -0,0 +1,51 @@
...
+## Important Notes
+
+- Multitenancy on ingestion is currently [only working](https://github.com/grafana/tempo/issues/495) with GPRC and this may never change. It is strongly recommended to use the OpenTelemetry Collector to support multitenancy as described above.
+- The way the read path is configured is temporary and should be much more straightforward once the [tempo-query dependency is removed](https://github.com/grafana/tempo/issues/382).
+
+## Disabling Multitenancy
+Most Tempo installations will be single tenant. If this is desired simply set the following config
+value on all Tempo components:
+```
+auth_enabled: false
```
dgzlopes, line 42: If most Tempo installations will be single-tenant, perhabs it's interesting to set `auth_enabled: false` by default.

joe-elliott: yeah, i've thought about it before, as well as renaming that option to something makes more sense.  that particular config is inherited from Loki/Cortex.

dgzlopes: What do you have in mind for the rename?

I can create a new ticket. This looks like a `good first issue` :)

docs/tempo/website/guides/multitenancy.md
```diff
@@ -0,0 +1,51 @@
...
+
+Tempo is a multitenant distributed tracing backend. It supports multitenancy through the use
+of a header: `X-Scope-OrgID`. This guide details how to setup or disable multitenancy.
+
+## Multitenancy
+
+If you're interested in setting up multitenancy, please consult the [multitenant example](https://github.com/grafana/tempo/tree/master/example/docker-compose/docker-compose.multitenant.yaml)
+in the repo. This example uses the following settings to achieve multitenancy in Tempo:
+
+- Configure the OTEL Collector to attach the X-Scope-OrgID header on push:
```
achatterjee-grafana, line 13: Should these be numbered steps, not bullets? Or can a user configure these settings in any order? Thanks.

joe-elliott: There is no implied order here.  Bullets are fine.

docs/tempo/website/guides/multitenancy.md
joe-elliott: I'm not sure if we'd really want to rename this.  Maybe provide an alternative that maps to the same internal config like: `disable_multitenancy` and move all our docs over to it.

@annanay25 @mdisibio ?

annanay25: Yeah I'm not sure we'd want to rename this either, especially since this is consistent across Loki/Cortex. We could open a PR to Cortex and see if the change is accepted there, and meanwhile improve our docs to best explain what this option does

annanay25 (approved):

yashvardhan-zs:
Does this example enable grpc ingestion for multitennacy

kvrhdn:
I think the examples folder has been restructured a bit since this PR was merged. You can find a multitenancy example here: https://github.com/grafana/tempo/tree/main/example/docker-compose/otel-collector-multitenant

yashvardhan-zs:
This example is not working , unable to use tempo multitnnacy and also i want to use it without grafana datasource @kvrhdn @joe-elliott @annanay25

yashvardhan-zs:
is it possible to use multitennacy in tempo without using grafana?

PR #500, opened 2021-02-02
Skip workflow runs if the change only affects the docs folder

dgzlopes:
**What this PR does**:
If a PR only changes the contents of the `docs/` folder, there is no need to run the `benchmark` and the `ci` workflows.

I tried it on a private repo, and looks like it's working! Hopefully here will work too.

My tests:
- Change file inside `docs/` -> Workflow skiped.
- Change file outside `docs/` -> Workflow run.
- Change file inside `docs/` and file outside `docs/` -> Workflow run.

This was discussed with @annanay25 on Slack.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,5 +1,8 @@
 name: Benchmark
-on: [pull_request]
+on: 
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
 jobs:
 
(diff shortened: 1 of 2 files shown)


annanay25 (approved):
500th PR! Thanks @dgzlopes, lgtm.

PR #502, opened 2021-02-03
feat: update protobuf image version

### LINKED ISSUE #479: Upgrade to the latest protobuf docker image
OpenTelemetry is updating the protobuf image in the master repo to `0.2.1`, we should follow suit.

We're using `0.1.0` as of now - https://github.com/grafana/tempo/blob/3bd19f663f42c394e0b5ced8e1bd90a4f5798d75/Makefile#L124

swartz-k:
**What this PR does**:
Update protobuf image and opentelemetry collector version
Fixes #479

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


mdisibio:
Hi, thanks for looking into this.  We have looked into upgrading the open-telemetry-collector dependency as well and it looks like this is running into the same hurdles.   `make vendor-dependencies` fails with:
```
 github.com/grafana/tempo/modules/distributor/receiver imports
	go.opentelemetry.io/collector/receiver/opencensusreceiver imports
	github.com/soheilhy/cmux tested by
	github.com/soheilhy/cmux.test imports
	google.golang.org/grpc/examples/helloworld/helloworld: ambiguous import: found package google.golang.org/grpc/examples/helloworld/helloworld in multiple modules:
	google.golang.org/grpc v1.33.2 (/home/runner/go/pkg/mod/google.golang.org/grpc@v1.29.1/examples/helloworld/helloworld)
	google.golang.org/grpc/examples v0.0.0-20210129004707-0bc741730b81 (/home/runner/go/pkg/mod/google.golang.org/grpc/examples@v0.0.0-20210129004707-0bc741730b81/helloworld/helloworld)
Makefile:149: recipe for target 'vendor-dependencies' failed
```

The root cause is that a cmux test relies on grpc/examples, which conflicts with the grpc version required by cortex/etc.  One possible way to address this is to vendor cmux and delete the conflicting test.

Were you able to build and run this change locally?  If so, I would be interested to learn how that worked, since we ran into the same hurdles.

swartz-k:
> Were you able to build and run this change locally? If so, I would be interested to learn how that worked, since we ran into the same hurdles.

Sorry for the late. I tested it locally by `make vendor-check` but it's different with CI. I am working on it.

swartz-k:
> I think maybe I need to add a `make docker-vendor-check` in Makefile to test CI in locally.

In docker I got the same error. This problem can be solved by `rm -rf /go/pkg/mod/google.golang.org/grpc@v1.29.1/examples/helloworld/helloworld`
image

joe-elliott:
We have talked and @mdisibio is pursuing a more holistic method of fix this.  It will be a combination of fixing our vendor mangling, upgrade us to 1.16 and add direct support for OTEL proto into the distributors instead of relying on the shims.

Thank you for the research @swartz-k, but I think it's best to let him carry this one forward.

swartz-k:
You are welcome. Thanks.

PR #508, opened 2021-02-05
Add per-tenant block retention

### LINKED ISSUE #77: Add per tenant retention
Follow the same pattern as per tenant overrides.

mdisibio:
**What this PR does**:
Adds per-tenant block retention.  Uses existing overrides.yaml system, value is optional, defaults to global retention.  The callback design is done so that it continues to work with the overrides.yaml polling and use new limits without restarting.

This introduces a not-great situation where it is now possible to specify block retention in a third location:  tempo.yaml -> `overrides.block_retention`, which are the default limits, and would technically override the original value in tempo.yaml -> `compactor.compaction.block_retention`.  This isn't harmful and unlikely to ever be used, but wanted to mention it.

Fixes #77 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/compactor/compactor.go
```diff
@@ -34,11 +36,12 @@ type Compactor struct {
 	subservicesWatcher *services.FailureWatcher
 }
 
-// New makes a new Querier.
-func New(cfg Config, store storage.Store) (*Compactor, error) {
+// New makes a new Compactor.
```
joe-elliott, line 39: wups

modules/compactor/compactor.go
```diff
@@ -156,6 +161,11 @@ func (c *Compactor) Combine(objA []byte, objB []byte) []byte {
 	return combinedTrace
 }
 
+// BlockRentionForTenant implements CompactorOverrides
+func (c *Compactor) BlockRentionForTenant(tenantID string) time.Duration {
```
joe-elliott, line 165: nit: spelling:  `BlockRetentionForTenant`

joe-elliott (approved):
One nit.  I also agree with leaving the config in the compactor. Nearly all Tempo users outside of us will be running in single tenant mode and it doesn't make sense to force them to use overrides to set retention.

bputt-e:
@joe-elliott is this a part of tempo 1.4.1? The docs state "Currenly only ingestion limits can be overridden." which makes me think this was backed out or its undocumented

joe-elliott:
Yup, this feature has been in Tempo since 0.6.0. Thanks for the heads up. Pushing a fix to the docs now.

PR #509, opened 2021-02-06
Drop Vulture Loki dependency

### LINKED ISSUE #374: Drop Vulture's Loki Dependency
Currently the vulture requires Loki to find trace ids.  Instead we would like it to generate its own traces, push those into Tempo and then request them back for analysis.

Alternative ideas are welcome but what makes sense to me is something like:

1) Every n seconds wake up and generate a trace
2) Divide unixTime / n to get a stable value and then seed the random # generator with it
3) Generate and push a random trace to Tempo

Then a second goroutine can randomly select one of the time slots, use the same seed to generate the same trace, request it from Tempo and confirm it everything matches.

dgzlopes:
**What this PR does**:
This PR refactors Tempo Vulture to remove the actual Loki dep. 

Now, instead of searching for traces in Loki we: 
- Generate a new trace (with some random data/spans).
- Push the trace to Tempo (in batches).
- Query a "pushed" trace randomly.

Fixes #374 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


cmd/tempo-vulture/main.go
```diff
@@ -45,95 +42,148 @@ func init() {
...
 
-	ticker := time.NewTicker(15 * time.Second)
+	// Write
 	go func() {
 		for {
 			<-ticker.C
 
-			for _, duration := range testDurations {
+			iteration++
+			rand.Seed(startTime / iteration)
```
joe-elliott, line 66: i was thinking of seeding with something like:
```
time.Now().Unix() / 5 * 5
```
this will floor the unix timestamp to the nearest 5 seconds.  If we start a goroutine every 5 seconds with the seed passed in we can generate a trace using the seed.  5 seconds could be configurable.

cmd/tempo-vulture/main.go
```diff
@@ -45,95 +42,148 @@ func init() {
...
+				metricErrorTotal.Inc()
+				continue
 			}
 		}
 	}()
 
+	// Read
+	go func() {
+		for {
+			<-ticker.C
```
joe-elliott, line 88: on the read path we can now choose any 5 second interval in the past n days, seed a random number generator, generate a trace and query tempo for it.  

"n days" should be a mix of two parameters
- min-time             : we can just hardcode a datetime to never query before.  this will basically be the time we first start the vulture and will prevent it from querying for traces before it started.
- tempo-retention  : since we eventually drop traces we need to pass the retention to tempo-vulture so it won't query traces after we know tempo should have dropped them.

so the query path would be like:
1. randomly choose a time after max(min-time, tempo-retention) and before now
2. do the same floor operation above ((unix / 5) * 5)
3. generate trace
4. query for trace
5. confirm everything matches up.
6. back to 1

cmd/tempo-vulture/main.go
```diff
@@ -45,95 +42,148 @@ func init() {
...
+	if err != nil {
+		return nil, err
+	}
+	return jaeger_grpc.NewReporter(conn, nil, logger), err
+}
 
-	for _, id := range traceIDs {
-		time.Sleep(backoff)
+func makeThriftBatch() *thrift.Batch {
+	var spans []*thrift.Span
```
joe-elliott, line 140: oh man, let's make some beefy traces.  100s, 1000s, 10,000s of spans split across a bunch of batches.
randomize everything names, tags, logs.

joe-elliott (commented):
Great progress!

I tried to comment on what I was looking for w/r to the random # seeding.  If that's unclear please reach out.

cmd/tempo-vulture/main.go
dgzlopes: Makes sense, I'll work on this :)

cmd/tempo-vulture/main.go
dgzlopes: Thanks a lot for the explanation! I updated the PR. Hopefully, it's more in-line with your proposal. 

As you can see I re-used the `startTime` variable after retention, but I can create an internal go-routine variable if needed (that way startTime is constant).

Also, I used a different time for the ticker and the slot (slot=ticker*2). That's because I had some rounding problems... And preferred to give it some free room.

joe-elliott: nice.  yup, this is definitely what i was thinking.

> Also, I used a different time for the ticker and the slot (slot=ticker*2)

I think 2 seconds is just really tight for trace creation.  It could be as loose as 30 seconds and it would be fine.  Maybe  split how often to generate a trace and how often to query them back?

Also, don't be concerned if we see some failures at first start b/c it's querying a trace that does not yet exist.  This doesn't have to be perfect.  It just a tool to help find consistency issues if they creep into the codebase.

cmd/tempo-vulture/main.go
dgzlopes: Added the code required for creating **beefy traces**. As you can see, I've some hardcoded ranges. Probably it's interesting to let the user configure some of those (e.g. number of batches).

cmd/tempo-vulture/main.go
dgzlopes: Oh! My bad. I was using 2 seconds for testing. Bumped it to 30s, sounds like a good default.

Maybe splitting the how often we do each thing is a good idea, but I'm not sure if a lot of people would customize that.

cmd/tempo-vulture/main.go
```diff
@@ -45,95 +43,189 @@ func init() {
...
+			// query the trace
+			metrics, err := queryTempoAndAnalyze(tempoQueryURL, hexID)
+			if err != nil {
+				glog.Error("error querying Tempo ", err)
+				metricErrorTotal.Inc()
+				continue
+			}
+
+			metricTracesInspected.Add(float64(metrics.requested))
+			metricTracesErrors.WithLabelValues("notfound").Add(float64(metrics.notfound))
```
joe-elliott, line 114: let's add a "failed" metric for anything that's not 404 or 200.

cmd/tempo-vulture/main.go
```diff
@@ -45,95 +43,189 @@ func init() {
...
+		})
+	}
+	return logs
+}
+
+func makeThriftBatch(traceID int64) *thrift.Batch {
+	var spans []*thrift.Span
+	for i := int64(0); i < generateRandomInt(1, 5); i++ {
+		spans = append(spans, &thrift.Span{
+			TraceIdLow:    traceID,
```
joe-elliott, line 182: traceID is 64 bits, but traceIDLow is just 32.  traceID should be split across low and high.

cmd/tempo-vulture/main.go
dgzlopes: I'm re-using `util.QueryTrace`, so I added this new metric when that function returns an error and it's not `util.ErrTraceNotFound`.

I increment both `metricErrorTotal` and `metricTracesErrors{error="failed"}`, that's ok?

cmd/tempo-vulture/main.go
dgzlopes: Changed this. Now I generate both separately as we do here: https://github.com/grafana/tempo/blob/aa88d276fac6f696fc6149b3466eeb8aead578bb/integration/e2e/e2e_test.go#L243

By doing this, I simplified the read path a bit!

cmd/tempo-vulture/main.go
```diff
@@ -45,95 +43,189 @@ func init() {
...
+}
 
-		// iterate through
-		if hasMissingSpans(trace) {
-			glog.Error("has missing spans", id)
-			tm.missingSpans++
-		}
+func generateRandomLogs() []*thrift.Log {
+	var logs []*thrift.Log
+	for i := int64(0); i < generateRandomInt(1, 5); i++ {
```
joe-elliott, line 169: i think `generateRandomInt()` will be called on every loop evaluation.  this technically works but is kind of weird.  let's move it up a line:

```
count := generateRandomInt(1, 5)
for i := int64(0); i < count; i++ {
...
```

joe-elliott (approved):
one nit, then i'll merge.

cmd/tempo-vulture/main.go
dgzlopes: Didn't know that! Changed :)

cmd/tempo-vulture/main.go
```diff
@@ -45,95 +43,192 @@ func init() {
...
-		24 * time.Hour,
-		12 * time.Hour,
-		6 * time.Hour,
-		3 * time.Hour,
-		time.Hour,
-		30 * time.Minute,
-	}
+	startTime := time.Now().Unix()
+	ticker := time.NewTicker(tempoBackoffDuration)
+	slot := int64(tempoBackoffDuration/time.Second) * 2
```
joe-elliott, line 60: i'd like to drop this slot idea and just generate the directly on the timestamps.  i think this will make things easier to replicate outside the vulture.

i also think we should split the read/write tickers.  maybe we only want to generate a trace every 30 seconds, but execute a query every 5 seconds?

joe-elliott (commented):
this is really close.  one last comment, but i can be persuaded otherwise.

cmd/tempo-vulture/main.go
dgzlopes: > i'd like to drop this slot idea and just generate the directly on the timestamps. i think this will make things easier to replicate outside the vulture.

I removed the slot thing but from my POV the result was not great... Maybe I misunderstood something...

But, I think your interval approach is great, so I refactored the code to replicate that behavior. Now it's much cleaner and works flawlessly. Hopefully, it makes some sense!

> i also think we should split the read/write tickers. maybe we only want to generate a trace every 30 seconds, but execute a query every 5 seconds?

Done! :)

joe-elliott (approved):

Issue #395, opened 2020-12-07
Document the overrides configuration (global and per-tenant) with examples


herrbpl:
**Describe the bug**

Suddenly, distributor failed to accept traces with error messages:
"max live traces per tenant exceeded: per-user traces limit (local: 10000 global: 0 actual local: 10000) exceeded"

There seem to be no documentation about anything on it (and Tempo documentation is severly lacking, certainly not production-level documentation!). 

**To Reproduce**
1. Start Tempo (SHA ):

Used docker image grafana/tempo:latest

> ./tempo -version
> tempo, version  (branch: , revision: )
>   build user:       
>   build date:       
>   go version:       go1.15.3

2. Perform Operations (Read/Write/Others)
Just ingest certain amount of traces and it starts to happen.

**Expected behavior**
traces are being processed

**Environment:**
 - Infrastructure: Kubernetes
 - Deployment tool: Helm, improved version of available chart (which also does not really work)

**Additional Context**
Using jaeger http trift for this particular ingest, with default config. 

/ # cat /conf/tempo.yaml 
```yaml
auth_enabled: false
compactor: null
distributor:
  receivers:  
    jaeger:
      protocols:        
        thrift_http: 
        grpc: 
    opencensus:  
    zipkin: 
    otlp: 
      protocols: 
        http: 
        grpc: 
          endpoint: 0.0.0.0:55689
#          keepalive:
#            server_parameters:
#              max_connection_idle: 11s
#              max_connection_age: 12s
#              max_connection_age_grace: 13s
#              time: 30s
#              timeout: 5s
#            enforcement_policy:
#              min_time: 10s
#              permit_without_stream: true
ingester:
  lifecycler:
    ring:
      replication_factor: 2
memberlist:
  abort_if_cluster_join_fails: false
  bind_port: 7946
  join_members:
  - gossip-ring.tempo.svc.cluster.local:7946
overrides:
  per_tenant_override_config: /conf/overrides.yaml
server:
  http_listen_port: 3100
storage:
  trace:
    backend: s3
    s3:
      access_key: xxxxxxxxxxxxxxxxxxxxxx
      bucket: xxxxxxxxxxxxxxxxxxxxxx
      endpoint: xxxxxxxxxxxxxxxxxx
      insecure: false
      region: eu-central-1
      secret_key: xxxxxxxxxxxx
    blocklist_poll: 5m
    memcached:
      consistent_hash: true
      host: memcached
      service: memcached-client
      timeout: 500ms
[...]

PR #510, opened 2021-02-08
Updated doc for issue 395

achatterjee-grafana:
- Added new topic: "Override trace limit". 
- Updated topic "Troubleshooting" with a new section "Problem 4. Distributor is not accepting traces".

Issue: https://github.com/grafana/tempo/issues/395

### DIFF
--- a/docs/tempo/website/configuration/ingestion-limit.md
+++ b/docs/tempo/website/configuration/ingestion-limit.md
@@ -5,20 +5,16 @@ weight: 30
 
 # Ingestion limits
 
-The default per user or per tenant trace limits in Tempo may not be sufficient in high volume tracing environments. The following message, for example, indicates that your tenants are exceeding the trace limits imposed on them.
+The default ingestion limits in Tempo may not be sufficient in high volume tracing environments. The following message, for example, indicates that a tenant is exceeding the trace limits imposed on them.
  ```
     max live traces per tenant exceeded: per-user traces limit (local: 10000 global: 0 actual local: 10000) exceeded
 ```    
 
-In such environments, increase the limits either individually by tenants or globally across the cluster
-so that your traces are not limited by the default settings.
+The following sections describe the available options, per tenant overrides and global vs. local configurations.
 
-This following sections describe the parameters whose limits can be overridden, as well as the methods of
-overriding the default limits.
+## Configuration options
 
-## Override parameters
-
-You can override the default values of the following parameters:
+The following options can be used to limit ingestion:
 
    - `ingestion_burst_size` : Burst size used in span ingestion. Default is `100,000`.
    - `ingestion_rate_limit` : Per-user ingestion rate limit in spans per second. Default is `100,000`.
@@ -45,10 +41,10 @@ LIVE_TRACES_EXCEEDED: max live traces per tenant exceeded: per-user traces limit
 
 ## Standard overrides
 
-To set a new trace limit that applies to all tenants of the cluster:
+To configure new ingestion limits that applies to all tenants of the cluster:
 
-1. Create an `overrides` section at the bottom of the `/tempo/integration/bench/config.yaml` file.
-1. Add the `overrides` parameters under this section. For example:
+1. Create an `overrides` section at the bottom of your configuration file.
+1. Add the parameters under this section. For example:
 
 ```
     overrides:
@@ -58,16 +54,14 @@ To set a new trace limit that applies to all tenants of the cluster:
         max_traces_per_user: 10000 
 ``` 
 
-A snippet of a `config.yaml` file showing how the `overrides` section is [here](https://github.com/grafana/tempo/blob/a000a0d461221f439f585e7ed55575e7f51a0acd/integration/bench/config.yaml#L39-L40). A list of common override parameters and their description can  be found in [Override parameters](#override-parameters).
-
-Setting a maximum value for the `overrides` setting is a failsafe way to prevent one tenant from negatively impacting others within the cluster. See also, [Distributor is not accepting traces](../../troubleshooting/#problem-4-distributor-is-not-accepting-traces) for more information.
+A snippet of a `config.yaml` file showing how the `overrides` section is [here](https://github.com/grafana/tempo/blob/a000a0d461221f439f585e7ed55575e7f51a0acd/integration/bench/config.yaml#L39-L40). 
 
 ## Tenant-specific overrides
 
 Sometimes you don't want all tenants within the cluster to have the same settings. To add overrides that are specific to an individual tenant:
 
 1. Create an `overrides` section at the bottom of the `config.yaml` file.
-1.  Add an entry `per_tenant_override_config` to point to a separate file named `overrides.yaml` that contains tenant-specific overrides.
+1. Add an entry `per_tenant_override_config` to point to a separate file named `overrides.yaml` that contains tenant-specific overrides.
 
     ```
     overrides:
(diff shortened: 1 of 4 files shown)


docs/tempo/website/configuration/override.md
```diff
@@ -0,0 +1,63 @@
+---
+title: Override trace limit
+weight: 30
+---
+
+# Override trace limit
```
joe-elliott, line 6: This document opens up as if it's about one specific override and then generalizes.  I think a better layout would be:

1) List all 4 current override options.  Include details about what they do and sample log messages when they are tripped.

- ingestion_burst_size
- ingestion_rate_limit
- max_spans_per_trace
- max_traces_per_user

2) Show how to configure these using the standard overrides block:

```
overrides:
    ingestion_rate_limit:
    ...
```
3) Include details about global limit.  Global ingestion rate limit uses the distributor ring to attempt to 

```
overrides:
    ingestion_rate_strategy: global
```

4) Include details about how to do per tenant overrides:

```
overrides:
    per_tenant_override_config: /conf/overrides.yaml
```

file format:

```

File: docs/tempo/website/troubleshooting/_index.md [hunk 2]
```diff
@@ -4,10 +4,15 @@ weight: 470
 ---
 
 # Troubleshooting missing traces
-This topic helps with day zero operational issues that may come up when getting started with Tempo. It walks through debugging each part of the ingestion and query pipeline to drill down and diagnose issues. 
+This topic helps with day zero operational issues that may come up when getting started with Tempo. It walks through debugging each part of the ingestion and query pipeline to drill down and diagnose issues.
```
annanay25, line 7: This page is getting a bit too long and scary, can we also split out the individual problems into smaller pages?

achatterjee-grafana:
Updated topic based on Joe's review. Once this is approved/ merged, I will break down the Trounshooting as you suggested @annanay25.
I need some info to complete the topic, added NEED INFO sections to add the information.

achatterjee-grafana:
QUESTION: What happens if there is a separate override file with per-tenant settings as well as an overrides section within the main configuration file? Which setting has the precedence? Does it mean that the overrides parameter applies to all tenants of the cluster except to the ones that have their tenant-specific overrides? Thank you for any clarification.

joe-elliott:
> QUESTION: What happens if there is a separate override file with per-tenant settings as well as an overrides section within the main configuration file? Which setting has the precedence? Does it mean that the overrides parameter applies to all tenants of the cluster except to the ones that have their tenant-specific overrides? Thank you for any clarification.

Per tenant settings override the settings in the main configuration file.  The main configuration file is seen as a place to put defaults.  The per tenant settings can be used to adjust to limit each tenant individually.

docs/tempo/website/configuration/override.md
```diff
@@ -0,0 +1,91 @@
...
+so that your traces are not limited by the default settings. 
+
+This topic lists the parameters whose limits can be overridden, as well as describes various methods of
+overriding these default limits.
+
+## Override parameters
+
+The parameters whose default values can be overridden are:
+
+   - `ingestion_burst_size` : `NEED INFO`
```
joe-elliott, line 23: I can't say i know precisely what this means either.  It is passed here: https://pkg.go.dev/golang.org/x/time/rate#Limiter.Burst.  The `time/rate` is an implementation of a [token bucket](https://en.wikipedia.org/wiki/Token_bucket).  If you ctrl+f "burst" here:  https://cortexmetrics.io/docs/configuration/configuration-file/  you will find some descriptions of the same field in Cortex.

I think it's fine to say something like "Burst size used in span ingestion".   We currently run ingestion_burst_size == ingestion_rate_limit and I think that also would be fair guidance to provide.

docs/tempo/website/configuration/override.md
```diff
@@ -0,0 +1,91 @@
...
+   - `ingestion_max_batch_size` : Per-user allowed ingestion max batch size (in number of spans). Default is `1000`.
+   - `ingestion_rate_limit` : Per-user ingestion rate limit in spans per second. Default is `100,000`.
+   - `max_spans_per_trace` : Maximum number of spans per trace.  0 to disable. Default is `50,000`.
+   - `max_traces_per_user`: Maximum number of active traces per user, per ingester. 0 to disable. Default is `10,000`.
+
+`NEED INFO:The trigger points for each of the trace limits, and the error message to help the user identify which trace limit has been hit.`
```
joe-elliott, line 29: `ingestion_burst_size` and `ingestion_rate_limit` both control the rate limit.  A log message would contain:
```
RATE_LIMITED: ingestion rate limit (100000 spans) exceeded while adding 10 spans
```
`max_spans_per_trace`:
```
TRACE_TOO_LARGE: totalSpans (50000) exceeded while adding 2 spans
```
`max_traces_per_user`:
```
LIVE_TRACES_EXCEEDED: max live traces per tenant exceeded: per-user traces limit (local: 10000 global: 0 actual local: 1) exceeded
```

There is also a metric that tracks this information: `tempo_discarded_spans_total`.  It has a `reason` and a `tenant` label.  The `reason` label will contain either: `trace_too_large`, `live_traces_exceeded` or `rate_limited`.  `rate_limited` applies to both the rate limit and burst size.

docs/tempo/website/configuration/override.md
```diff
@@ -0,0 +1,91 @@
...
+
+This topic lists the parameters whose limits can be overridden, as well as describes various methods of
+overriding these default limits.
+
+## Override parameters
+
+The parameters whose default values can be overridden are:
+
+   - `ingestion_burst_size` : `NEED INFO`
+   - `ingestion_max_batch_size` : Per-user allowed ingestion max batch size (in number of spans). Default is `1000`.
```
joe-elliott, line 24: this needs to be removed.  no longer exists.

joe-elliott (approved):

achatterjee-grafana:
Let me know when you have finished @joe-elliott. Thanks for the review.

joe-elliott:
@achatterjee-grafana I am done.  I believe I addressed all of your "NEED INFO" tags.

achatterjee-grafana:
Added changes from Joe. Will break down Troubleshooting in separate PR.

docs/tempo/website/configuration/override.md
```diff
@@ -0,0 +1,107 @@
...
+## Specify overrides to the default settings
+
+To change the default settings for various trace limits, add an `overrides` section. You can:
+
+- Add the `overrides` setting directly in the configuration file.
+- Create a new yaml file containing the `overrides` settings and then reference it from the configuration file.
+
+You can set the `overrides` to apply to all tenants within the cluster, to specific tenants, or globally to the cluster. Per tenant settings override the settings in the main configuration file. Use the main configuration file as a place to add the default values. The per tenant settings can be used to adjust to limit each tenant individually. For more information, refer to the [Override strategies](#override-strategies) section.
+
+### Standard overrides
```
joe-elliott, line 56: Too much stuff going on in this section.  Just say:  To configure overrides add the following block to your confg yaml:

```
overrides:
  ingestion_burst_size: 50000
  ingestion_rate_limit: 50000
  max_spans_per_trace: 50000
  max_traces_per_user: 10000 
```

docs/tempo/website/configuration/override.md
```diff
@@ -0,0 +1,107 @@
...
+    overrides:
+        "<tenant id>":
+            ingestion_max_batch_size: 5000
+            ingestion_rate_limit: 400000
+            max_spans_per_trace: 250000
+            max_traces_per_user: 100000
+    ```
+## Override strategies
+
+The trace limits specified by the various parameters are, by default, applied as tenant-level limits. For example, a max_traces_per_user setting of 10000 means that each tenant within the cluster has a limit of 10000 traces per user. This is known as a `local` strategy in that the specified trace limits are local to each tenant.
```
joe-elliott, line 96: local means "local to the distributor", not local to the tenant.  global means averaged across all distributors by using a distributor ring.

basically with a local ingestion rate of 10000 then each distributor will independently enforce 10000 spans/second in that distributor.  if you set a global rate of 10000 then it will divide 10000 by the number of distributors (say 5) and each distributor will enforce 2000 spans/second.  if you then scale up distributors to 10 then each distributor will enforce 1000 spans/second.

docs/tempo/website/configuration/ingestion-limit.md
```diff
@@ -0,0 +1,106 @@
...
+```
+    TRACE_TOO_LARGE: totalSpans (50000) exceeded while adding 2 spans
+```
+
+Finally, when the limit for the `max_traces_per_user` parameter exceeds the following message is logged:
+
+```
+LIVE_TRACES_EXCEEDED: max live traces per tenant exceeded: per-user traces limit (local: 10000 global: 0 actual local: 1) exceeded
+```
+## Specify overrides to the default settings
```
joe-elliott, line 44: Combine the sections "Specify overrides to the default settings" and "Standard overrides".  These are basically the same thing: Telling the user how to configure non-default limits.

joe-elliott (changes requested):
One final change and we're good.  Thanks for this work!

docs/tempo/website/configuration/ingestion-limit.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Standard overrides
+
+To set a new trace limit that applies to all tenants of the cluster:
+
+1. Create an `overrides` section at the bottom of the `/tempo/integration/bench/config.yaml` file.
+1. Add the override parameters under this section. For example:
+
+```
+    overrides:
```
joe-elliott, line 54:
```suggestion
overrides:
```

joe-elliott (approved):

PR #512, opened 2021-02-08
Update cortex to v1.7.0-rc.0+(latest commit)

mdisibio:
**What this PR does**:
Update cortex to latest commit to bring in more ring membership improvements and fixes.  This is v1.7.0-rc.0 + some more commits.   It includes a new memberlist KV page which is hosted on /memberlist (for all targets, as it covers all rings).   Note - The bulk of the changes are from relocating `Logger` from `pkg/util` to `pkg/util/log`.  

n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/querier/config.go
```diff
@@ -25,13 +25,10 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
...
 		MaxConcurrentRequests: cfg.MaxConcurrentQueries,
 		Parallelism:           2,
-		GRPCClientConfig: grpcclient.ConfigWithTLS{
-			GRPC: grpcclient.Config{
-				MaxRecvMsgSize:     100 << 20,
-				MaxSendMsgSize:     16 << 20,
-				UseGzipCompression: false,
-				GRPCCompression:    "gzip",
-			},
+		GRPCClientConfig: grpcclient.Config{
```
joe-elliott, line 28: the previous settings are weird:

```
  UseGzipCompression: false,	
  GRPCCompression:    "gzip",
```

Does that mean we were or were not using gzip?  Does this PR change that?

joe-elliott (commented):
One question.  Also this is technically Cortex 1.6.1+.  They haven't cut 1.7.0 yet.

modules/querier/config.go
mdisibio: Here is the code in question:  https://github.com/cortexproject/cortex/blob/v1.6.0/pkg/util/grpcclient/grpcclient.go#L67

Afaict, we were using gzip.   `UseGzipCompression:true` would override the compression algorithm, but a `false` would not.

joe-elliott (approved):

PR #515, opened 2021-02-08
Record block data size in meta

mdisibio:
**What this PR does**:
This PR adds a new field to the block meta for the final data file size, and support for this and other recent fields to the tempo-cli. This PR is a precursor to switching compaction over to a target file size instead of target trace count.   

n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/backend/block_meta.go
```diff
@@ -22,6 +22,8 @@ type BlockMeta struct {
 	StartTime       time.Time `json:"startTime"`
 	EndTime         time.Time `json:"endTime"`
 	TotalObjects    int       `json:"totalObjects"`
+	TotalSize       uint64    `json:"totalSize"`
+	FinalSize       uint64    `json:"finalSize"`
```
annanay25, line 26: Why do we need two sizes in the block meta? Let's just use the compressed size?

tempodb/encoding/compactor_block.go
```diff
@@ -121,6 +123,7 @@ func (c *CompactorBlock) BlockMeta() *backend.BlockMeta {
 
 	meta.StartTime = c.inMetas[0].StartTime
 	meta.EndTime = c.inMetas[0].EndTime
+	meta.FinalSize = c.totalBytesFlushed
```
joe-elliott, line 126: Why not use `.DataLength()` in the compactor block like the complete block?

joe-elliott (commented):
An appender is used by both the compactor and complete blocks.  It already does some record keeping w/r to the amount of data that has been appended.  If we want to track "totalSize" does it make sense to just put the logic there?

Echo @annanay25 's question about whether or not we need both.  the compressed size seems like enough to me.  If we are just interested in compression ratio we may want to consider metricing it.  I know Loki does that.

tempodb/backend/block_meta.go
mdisibio: Sounds like the consensus is that storing the uncompressed data size isn't valuable.  Removed.

tempodb/encoding/compactor_block.go
mdisibio: Good catch.  Was thinking the appendBuffer is flushed by the compactor block, but the appender is not.

joe-elliott (approved):

PR #516, opened 2021-02-08
Increment discarded spans metric for all limit reasons

joe-elliott:
**What this PR does**:
`tempo_discarded_spans_total` is meant to be a metric to inform tenants as to why they may be getting refused spans.  Previously it only covered the most basic rate limiting case.   This PR looks to extend that to include all limit based reasons that a span might be refused.

There is unfortunately no 100% correct way to metric this.  For instance, in the provided solution we are only including spans from the one trace that was rejected however the distributor will refuse the entire batch.  

We could try to metric this in the distributor which would make the metric more authentic.  It would be gross to try to get the rejection reason.  Substring matching?  It is also true that if a batch is rejected in the ingester it has already probably succeeded on other traces.

No great answers here.  Suggestions welcome.

### DIFF
--- a/modules/distributor/distributor.go
+++ b/modules/distributor/distributor.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/cortexproject/cortex/pkg/ring"
@@ -18,6 +19,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/weaveworks/common/logging"
 	"github.com/weaveworks/common/user"
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/health/grpc_health_v1"
 
@@ -32,9 +34,14 @@ import (
 const (
 	discardReasonLabel = "reason"
 
-	// RateLimited is one of the values for the reason to discard samples.
-	// Declared here to avoid duplication in ingester and distributor.
-	rateLimited = "rate_limited"
+	// reasonRateLimited indicates that the tenants spans/second exceeded their limits
+	reasonRateLimited = "rate_limited"
+	// reasonTraceTooLarge indicates that a single trace has too many spans
+	reasonTraceTooLarge = "trace_too_large"
+	// reasonTooManyTraces indicates that tempo is already tracking too many live traces in the ingesters for this user
+	reasonTooManyTraces = "too_many_traces"
+	// reasonInternalError indicates that spans were rejected b/c of a failure in Tempo
+	reasonInternalError = "internal_error"
 )
 
 var (
@@ -196,14 +203,15 @@ func (d *Distributor) stopping(_ error) error {
 func (d *Distributor) Push(ctx context.Context, req *tempopb.PushRequest) (*tempopb.PushResponse, error) {
 	userID, err := user.ExtractOrgID(ctx)
 	if err != nil {
+		// can't record discarded spans here b/c there's no tenant
 		return nil, err
 	}
 
-	// calculate and metric size...
+	// metric size
 	size := req.Size()
 	metricBytesIngested.WithLabelValues(userID).Add(float64(size))
 
-	// ... and spans
+	// metric spans
 	if req.Batch == nil {
 		return &tempopb.PushResponse{}, nil
 	}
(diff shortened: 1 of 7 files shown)


joe-elliott:
Changed course and decided that counting all discarded spans inside the distributor.  This does require some string matching, but hoping that testing will keep things in sync.

Overall prefer this method as it keeps the discarded spans metricing in one place, allows us to metric other failures inside of tempo, and over counts dropped spans instead of undercounts (which I prefer).

modules/distributor/distributor.go
```diff
@@ -32,9 +34,14 @@ import (
...
 
-	// RateLimited is one of the values for the reason to discard samples.
-	// Declared here to avoid duplication in ingester and distributor.
-	rateLimited = "rate_limited"
+	// reasonRateLimited indicates that the tenants spans/second exceeded their limits
+	reasonRateLimited = "rate_limited"
+	// reasonTraceTooLarge indicates that a single trace has too many spans
+	reasonTraceTooLarge = "trace_too_large"
+	// reasonTooManyTraces indicates that tempo is already tracking too many live traces in the ingesters for this user
+	reasonTooManyTraces = "too_many_traces"
```
mdisibio, line 42: `too_many_traces` could also be interpreted as a rate limit.  Thoughts on an alternate string like `live_traces_exceeded` ?

modules/overrides/limits.go
```diff
@@ -6,11 +6,15 @@ import (
...
-	// Local ingestion rate strategy
+	// LocalIngestionRateStrategy indicates that this limit can be evaluated in local terms only
 	LocalIngestionRateStrategy = "local"
-
-	// Global ingestion rate strategy
+	// GlobalIngestionRateStrategy indicates that an attempt should be made to consider this limit across the entire Tempo cluster
 	GlobalIngestionRateStrategy = "global"
+
+	// ErrorPrefixTooManyTraces is used to flag batches from the ingester that were rejected b/c they had too many traces
+	ErrorPrefixTooManyTraces = "TOO_MANY_TRACES:"
```
mdisibio, line 15: `too_many_traces` could also be interpreted as a rate limit. Thoughts on tweaking it or an alternate string like `live_traces_exceeded` ?

joe-elliott: agree :+1:

joe-elliott:
- too_many_traces => live_traces_exceeded
- added test for spans/second rate limit
- removed "internal_error" reason.  Since we can't 100% accurately account for this I'd rather this metric only represent those spans that were discarded due to limits.

modules/distributor/distributor.go
```diff
@@ -347,3 +355,17 @@ func requestsByTraceID(req *tempopb.PushRequest, userID string, spanCount int) (
...
 }
+
+func recordDiscaredSpans(err error, userID string, spanCount int) {
+	s := status.Convert(err)
+	if s == nil {
+		return
+	}
+	desc := s.Message()
+
+	if strings.HasPrefix(desc, overrides.ErrorPrefixLiveTracesExceeded) {
```
annanay25, line 366: nit: Metric'ing based on an error message prefix from a different service seems fragile.. We could wrap the error or change it in many different ways in the future. Should we change this to a `strings.Contains()`? Or should we map different codes in the `codes` package to our errors?

joe-elliott: > Metric'ing based on an error message prefix from a different service seems fragile..
We have tests to detect drift.  I prefer prefix to contains as it's more performant and during an event with tons of errors we would want to this code path to be as efficient as possible. 

> Or should we map different codes in the codes package to our errors?
We need to return specific and meaningful GRPC calls so downstream will react appropriately.  Also if we hijack different grpc status codes to mean different failures due to limits we'll run out eventually.  We definitely need something more robust.

100% agree that it's not great, but I think the string error is the best way to pass custom error information back through a GRPC rpcerror.

mdisibio: If nested errors are preserved across the gRPC boundary we could use `errors.Is` and const values like `errLiveTracesExceeded`

joe-elliott: Sadly, it is not.  We had the same issues with the Jaeger GRPC plugin and decided to go with string matching.

mdisibio: In that case it seems like string prefix is the best option.

mdisibio (approved):

PR #520, opened 2021-02-10
Add new compaction `max_block_bytes` setting

mdisibio:
**What this PR does**:
This PR starts the transition moving away from the trace-oriented `max_compaction_objects` setting to a data-oriented `max_block_bytes` setting.  TimeWindowBlockSelector enforces this limit by simply summing the block data sizes added in #515 which is an estimate but accurate enough.  For compatibility, `max_compaction_objects` remains as existing blocks may not yet have the new Size, but is deprecated and removed from examples.

Additionally, the TimeWindowBlockSelector was refactored to a (hopefully) simpler design that will accommodate more changes to compaction behavior in the future. 

n/a

**Checklist**
- [x] Tests updated
- [x] Documentation added


modules/compactor/config.go
```diff
@@ -31,6 +31,7 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
 
 	f.DurationVar(&cfg.Compactor.BlockRetention, util.PrefixConfig(prefix, "compaction.block-retention"), 14*24*time.Hour, "Duration to keep blocks/traces.")
 	f.IntVar(&cfg.Compactor.MaxCompactionObjects, util.PrefixConfig(prefix, "compaction.max-objects-per-block"), 6000000, "Maximum number of traces in a compacted block.")
+	f.Uint64Var(&cfg.Compactor.MaxBlockSize, util.PrefixConfig(prefix, "compaction.max-block-size"), 100*1024*1024*1024 /* 100GB */, "Maximum size of a compacted block.")
```
joe-elliott, line 34: `max-block-bytes` ?  I think we've been ending all the new "bytes" config options with the unit.

tempodb/compaction_block_selector.go
```diff
@@ -58,168 +58,149 @@ func (sbs *simpleBlockSelector) BlocksToCompact() ([]*backend.BlockMeta, string)
...
+
+		// find everything from cursor forward that belongs to this group
+		// Gather contiguous blocks while staying within limits
+		i := 0
+		for ; i < len(twbs.entries); i++ {
+			for j := i + 1; j < len(twbs.entries); j++ {
+				stripe := twbs.entries[i : j+1]
+				if twbs.entries[i].group == twbs.entries[j].group &&
+					len(stripe) <= twbs.MaxInputBlocks &&
+					totalObjects(stripe) <= twbs.MaxCompactionObjects &&
```
joe-elliott, line 158: does this mean we have to set `max_compaction_objects` to a really high number (so this condition is always true) and then max block size to the value we actually want for it to work?

joe-elliott: after an offline disucssion agree that this is the behavior and it's fine since this is only a temporary transition period until we drop all blocks that don't have a size.

mdisibio: That's correct.  The trace limit is left as a safety guard while Size meta is rolled out.  Existing blocks will not have the Size meta and will be treated as Size 0.  During that time it is unlikely to hit the new size limit if the trace limit is not also increased.

modules/compactor/config.go
mdisibio: Sounds good, this also matches the setting added in #474

joe-elliott (approved):

PR #522, opened 2021-02-11
Switch default ingester_client grpc compression from gzip to snappy

mdisibio:
**What this PR does**:
Switches the default compression for ingester_client from gzip to snappy. This is primarily used for distributor->ingester traffic, but also for querier->ingester.   

Internal testing shows that snappy outperforms gzip:
* 53% reduction in distributor cpu
* 12% reduction in ingester cpu
* 10% reduction in Push latency
* Minor improvements to mem usage (working set)

n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/ingester/client/client.go
+++ b/modules/ingester/client/client.go
@@ -45,7 +45,7 @@ func New(addr string, cfg Config) (*Client, error) {
 	opts := []grpc.DialOption{
 		grpc.WithInsecure(),
 		grpc.WithDefaultCallOptions(
-			grpc.UseCompressor("gzip"),
+			grpc.UseCompressor("snappy"),
 		),
 	}
 
--- a/modules/ingester/client/client.go
+++ b/modules/ingester/client/client.go
@@ -44,9 +44,6 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 func New(addr string, cfg Config) (*Client, error) {
 	opts := []grpc.DialOption{
 		grpc.WithInsecure(),
-		grpc.WithDefaultCallOptions(
-			grpc.UseCompressor("snappy"),
-		),
 	}
 
 	instrumentationOpts, err := cfg.GRPCClientConfig.DialOption(instrumentation())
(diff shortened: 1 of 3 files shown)


modules/ingester/client/client.go
```diff
@@ -45,7 +45,7 @@ func New(addr string, cfg Config) (*Client, error) {
 	opts := []grpc.DialOption{
 		grpc.WithInsecure(),
 		grpc.WithDefaultCallOptions(
-			grpc.UseCompressor("gzip"),
+			grpc.UseCompressor("snappy"),
```
annanay25, line 48: Did you intend to use `c.IngesterClient.GRPCClientConfig.GRPCCompression` here?

mdisibio: No, as far as I can tell, this default is a backup for a mis-configured ingester_client setting.  In normal cases `c.IngesterClient.GRPCClientConfig.GRPCCompression` is present and overrides this.  However it is possible to enter yaml such as:
```
ingester_client:
   grpc_client_config:
```

that will cause `c.IngesterClient.GRPCClientConfig.GRPCCompression` to be blank, somewhat unexpectedly.

If we remove this default then it means situations like the above would not have any compression (default "")

annanay25: Hrm, ok agree with this change. Had one last question: If someone deliberately leaves the GRPCCompression field blank in the YAML like you mentioned, would we still enable snappy?

mdisibio: Upon further testing, the DefaultCallOption should be removed entirely. No compression is a lack of any UseCompressor call options, which is not possible while the default exists.  After removing the DefaultCallOption this does work to use no compression:
```
ingester_client:
   grpc_client_config:
        grpc_compression: "" 
```

Disregard my comment about mis-configured yaml doing this unexpectedly.  It's because in the case of:
```
ingester_client:
   grpc_client_config:
```
`grpc_client_config` is being treated as empty string instead of an object.  It breaks more than the compression and is not useable: min/max message size are 0 and nothing will work.

annanay25 (approved):
lgtm!

PR #524, opened 2021-02-12
Make sure counters have _total suffix

gouthamve:
Prompted by the metric: `tempo_receiver_accepted_spans` but that comes from OTel so not touching that. Found these two are missing _total so added it.


annanay25 (approved):
lgtm, thanks!

PR #526, opened 2021-02-15
Adding a Shutdown Handler

### LINKED ISSUE #355: Improve /flush handler
The `/flush` handler on the ingesters will correctly force flush all traces to the backend, but it will not stop the ingester from receiving anything new.  We should provide a parameter that allows the user to both flush and force the ingesters to stop accepting traffic.  This will be required when performing certain operational procedures like scaling down ingesters.

The flush handler should be extended to take a parameter that causes the ingester to:
- Remove itself from the ring
- Stop ingesting new traces
- Force flush everything to the backend

annanay25:
**What this PR does**:

Adds a shutdown handler that helps with moving all ingester data, from memory/WAL/CompleteBlocks to the backend. This will only be used in case of ingester scaledown.

Fixes #355 

**Checklist**
- [x] Tests updated
- NA Documentation added


modules/ingester/flush.go
```diff
@@ -32,10 +36,15 @@ var (
...
 )
 
+type opKind int
+
 const (
 	// Backoff for retrying 'immediate' flushes. Only counts for queue
 	// position, not wallclock time.
 	flushBackoff = 1 * time.Second
+
+	complete opKind = iota
```
joe-elliott, line 46: this will set both complete and flush to the same value i think.  i'd prefer a const block only for opkinds and it to look something like:
```
const (
opKindComplete = iota
opKindFlush
)
```

modules/ingester/flush.go
```diff
@@ -124,16 +179,46 @@ func (i *Ingester) flushLoop(j int) {
...
+			}
+			instance.completeBlocks = append(instance.completeBlocks, completeBlock)
+			instance.blocksMtx.Unlock()
+		} else {
+			level.Debug(log.Logger).Log("msg", "flushing block", "userid", op.userID, "fp")
+
+			err := i.flushUserTraces(op.userID)
+			if err != nil {
+				level.Error(log.WithUserID(op.userID, log.Logger)).Log("msg", "failed to flush user", "err", err)
+
```
dgzlopes, line 210: Perhaps it's interesting to increment metricFailedFlushes here.

modules/ingester/flush.go
annanay25: They both get different values in the different version as well :) But the rewrite makes it clearer, updated.

modules/ingester/flush.go
annanay25: Right, that is interesting. I thought about that but since we're re-queuing the op, it made sense to metric on the "bad" errors that cannot be retried.

modules/ingester/instance.go
```diff
@@ -80,7 +86,9 @@ func newInstance(instanceID string, limiter *Limiter, writer tempodb.Writer) (*i
 		limiter:            limiter,
 		writer:             writer,
 
-		hash: fnv.New32(),
+		waitForFlush: atomic.NewInt32(0),
+		flushChan: make(chan int, 10), // should not need more than 2
```
joe-elliott, line 90: i don't think we need this and am scared of tracking state like this.  we should be able to just check the length of the flush op queue to determine if additional work needs to be done.

modules/ingester/flush.go
```diff
@@ -124,16 +196,50 @@ func (i *Ingester) flushLoop(j int) {
...
+			if err != nil {
+				// this is a really bad error that results in data loss.  most likely due to disk full
+				_ = op.completingBlock.Clear()
+				metricFailedFlushes.Inc()
+				level.Error(log.Logger).Log("msg", "unable to complete block.  THIS BLOCK WAS LOST", "tenantID", op.userID, "err", err)
+				instance.blocksMtx.Unlock()
+				continue
+			}
+			instance.completeBlocks = append(instance.completeBlocks, completeBlock)
+			instance.blocksMtx.Unlock()
```
joe-elliott, line 224: only the instance should touch its own mutexes.  you could pass the block into the instance if necessary.

modules/ingester/flush.go
```diff
@@ -124,16 +196,50 @@ func (i *Ingester) flushLoop(j int) {
...
-			level.Error(log.WithUserID(op.userID, log.Logger)).Log("msg", "failed to flush user", "err", err)
+			completeBlock, err := instance.writer.CompleteBlock(op.completingBlock, instance)
+
+			if op.immediate {
+				instance.flushChan <- 1
+			}
+
+			instance.blocksMtx.Lock()
+			if err != nil {
+				// this is a really bad error that results in data loss.  most likely due to disk full
```
joe-elliott, line 216: if this fails we should just requeue.

modules/ingester/flush.go
```diff
@@ -52,16 +61,60 @@ func (i *Ingester) Flush() {
...
+		_, _ = w.Write([]byte("error marking ingester unavailable"))
+	}
+
+	// move all data into flushQueue
+	i.sweepAllInstances(true)
+
+	// wait for blocks to complete and flush
+	for i.isWaitingForFlush() {
+		time.Sleep(100 * time.Millisecond)
+	}
```
joe-elliott, line 81: elsewhere you are using chans/atomic ints to track state.  i believe it's so that you won't move past this part of the shutdown handler until everything is flushed.  we need to find a simpler way to do this by inspecting the state of the queue itself.  these state variables are possibly going to expose us to some weird and difficult to debug race conditions.

something like:
```
while <len of queue> > 0 
  sleep a bit
  sweepAllInstances(true)   // put more stuff in the queue? 
```

we need to find a way to examine existing state to determine if we should move on if at all possible.

modules/ingester/instance.go
annanay25: unfortunately in the time between de-queuing a "complete" operation and queuing a "flush" operation, the queue will appear empty as @mdisibio pointed out.. maybe there's some things we can do there

modules/ingester/flush.go
```diff
@@ -52,28 +60,52 @@ func (i *Ingester) Flush() {
...
+	// lifecycler should exit the ring on shutdown
+	i.lifecycler.SetUnregisterOnShutdown(true)
+
+	// stop accepting new writes
+	i.markUnavailable()
+
+	// move all data into flushQueue
+	i.sweepAllInstances(true)
+
+	for !i.flushQueues.IsEmpty() {
```
joe-elliott, line 76: i think we want to return immediately from the shutdownhandler. 

```
	if i.flushQueues != nil {
		i.flushQueues.Stop()
		i.flushQueuesDone.Wait()
	}
```
does this logic in stopping already do what we want?  prevent the ingester from shutting down while there is work in the queues?

modules/ingester/flush.go
```diff
@@ -124,16 +165,38 @@ func (i *Ingester) flushLoop(j int) {
...
+		if op.kind == opKindComplete {
+			level.Debug(log.Logger).Log("msg", "completing block", "userid", op.userID)
+			instance, exists := i.getInstanceByID(op.userID)
+			if !exists {
+				// instance no longer exists? that's bad, log and continue
+				level.Error(log.Logger).Log("msg", "instance not found", "tenantID", op.userID)
+				continue
+			}
+
+			isComplete := instance.CompleteBlock()
```
joe-elliott, line 177: after dequeue is called the caller must either call `Requeue()` or `Clear()` to signal success or failure.

we need to return an error from `instance.CompleteBlock()` so we know which one to do.

modules/ingester/flush.go
```diff
@@ -124,16 +165,38 @@ func (i *Ingester) flushLoop(j int) {
...
+					from:   time.Now().Unix(),
+					userID: instance.instanceID,
+				})
+			}
+
+		} else {
+			level.Debug(log.Logger).Log("msg", "flushing block", "userid", op.userID, "fp")
+
+			err := i.flushUserTraces(op.userID)
+			if err != nil {
```
joe-elliott, line 192: this if block needs to be moved down.  that way it will catch either errors from completing or flushing and call Requeue() for both operations.

modules/ingester/instance.go
```diff
@@ -133,59 +133,77 @@ func (i *instance) CutCompleteTraces(cutoff time.Duration, immediate bool) error
...
+
+	// potentially long running operation placed outside blocksMtx
+	completeBlock, err := i.writer.CompleteBlock(completingBlock, i)
+
+	i.blocksMtx.Lock()
+	if err != nil {
+		// this is a really bad error that results in data loss.  most likely due to disk full
+		_ = completingBlock.Clear()
+		metricFailedFlushes.Inc()
+		level.Error(log.Logger).Log("msg", "unable to complete block.  THIS BLOCK WAS LOST", "tenantID", i.instanceID, "err", err)
```
joe-elliott, line 183: this error needs to be returned so we can requeue on failure.  also the block is no longer lost :)

modules/ingester/instance.go
```diff
@@ -133,59 +133,77 @@ func (i *instance) CutCompleteTraces(cutoff time.Duration, immediate bool) error
...
-		}()
+		return true, nil
 	}
 
-	return nil
+	return false, nil
+}
+
+// CompleteBlock() moves a completingBlock to a completeBlock
+func (i *instance) CompleteBlock() bool {
```
joe-elliott, line 162: i think this method needs to complete all blocks in the completingBlocks slice and not just the first one.  the operation you are putting in the queue is "complete blocks for this instance" and so should cover all blocks.

if we changed the operations to identify a specific block then we would only do 1.

modules/ingester/flush.go
```diff
@@ -89,24 +121,33 @@ func (i *Ingester) sweepInstance(instance *instance, immediate bool) {
...
+	isReady, err := instance.CutBlockIfReady(i.cfg.MaxBlockDuration, i.cfg.MaxBlockBytes, immediate)
 	if err != nil {
 		level.Error(log.WithUserID(instance.instanceID, log.Logger)).Log("msg", "failed to cut block", "err", err)
 		return
 	}
 
+	if isReady {
+		i.flushQueues.Enqueue(&flushOp{
+			kind:   opKindComplete,
+			from:   math.MaxInt64,
```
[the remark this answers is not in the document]
annanay25: I chose `maxint64` because it made sense to give higher priority to `opKindComplete` operations over `opKindFlush`

modules/ingester/flush.go
annanay25: > i think we want to return immediately from the shutdownhandler.

ok. will make that change. 

> does this logic in stopping already do what we want? prevent the ingester from shutting down while there is work in the queues?

I think this doesn't work for us because it stops further "enqueues". So if a complete operation is going on, the following flush operation will never be enqueued..

modules/ingester/flush.go
annanay25: So about this, I was wondering if we should retry this at all. It seems like it could only do more harm... like use up all the space on disk. wdyt?

modules/ingester/flush.go
annanay25: See my other comment above. Should we retry errors from `instance.CompleteBlock()`?

modules/ingester/instance.go
annanay25: OK this made me think and uncover a totally new scenario. Consider this case:

- sweepAllInstances(immediate=false) is called from the regular flush cycle
- sweepAllInstances(immediate=true) is called immediate after from shutdownhandler
- the completingBlock from the first call will take a long time to complete (because it will be a large block) and during this time, the `enqueue(opKindComplete)` from the second call will not succeed. 

This is as good as saying "while there's one completingBlock no other can process".

I think we should add blockID to the flushOp. And use that as part of the `Key()` so multiple `opKindComplete`s can be enqueued.

modules/ingester/flush.go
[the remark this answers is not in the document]
joe-elliott: i would prefer time.  it will also make the backoff behave correctly.  with this code the backoff will make the `from` value go negative.

modules/ingester/flush.go
joe-elliott: yes, we should retry.  we just need to make sure it cleans up after itself correctly on failure.  for now at least this is going to be a very manual process that we watch after calling `/shutdown`.  so if there are errors we will be there to intervene.

modules/ingester/instance.go
joe-elliott: this is ok, we just need to make complete behave like flush.  the flush logic keeps flushing the instance until there is nothing left to do.

modules/ingester/flush.go
```diff
@@ -150,12 +213,7 @@ func (i *Ingester) flushUserTraces(userID string) error {
 		return fmt.Errorf("instance id %s not found", userID)
 	}
 
-	for {
-		block := instance.GetBlockToBeFlushed()
-		if block == nil {
-			break
-		}
-
+	for _, block := range instance.GetBlocksToBeFlushed() {
```
joe-elliott, line 216: `GetBlocksToBeFlushed()` will only be called once.  If another block is added to the instances flush list during this time period it will not be picked up.  we need to repeatedly call this method on every loop in case more blocks get added.

modules/ingester/instance.go
annanay25: Even if we change it to that, we will miss work in the shutdownhandler right? Consider this:

- complete op is queued and is iterating over the loop of completingBlocks
- it finishes this iteration, checks `len(completingBlocks)` is empty
- before it could quit that loop, "shutdownhandler" is called, this increases length of completingBlocks but the completeOp could not be enqueued because there's already an op in progress. 

That op would be lost

joe-elliott: you are correct.  this would be lost.  agree we should go to per block operations.

modules/ingester/flush.go
```diff
@@ -52,28 +62,55 @@ func (i *Ingester) Flush() {
 	}
 }
 
-// FlushHandler calls sweepUsers(true) which will force push all traces into the WAL and force
+// ShutdownHandler handles a graceful shutdown for an ingester. It does the following things in order
+// * Stop incoming writes by exiting from the ring
+// * Flush all blocks to backend
+func (i *Ingester) ShutdownHandler(w http.ResponseWriter, _ *http.Request) {
+	go func() {
+		// lifecycler should exit the ring on shutdown
```
mdisibio, line 70: Adding some log messages here to show shutdown started and completed would be helpful.

modules/ingester/flush.go
```diff
@@ -124,13 +162,38 @@ func (i *Ingester) flushLoop(j int) {
...
+			level.Debug(log.Logger).Log("msg", "completing block", "userid", op.userID)
+			instance, exists := i.getInstanceByID(op.userID)
+			if !exists {
+				// instance no longer exists? that's bad, log and continue
+				level.Error(log.Logger).Log("msg", "instance not found", "tenantID", op.userID)
+				continue
+			}
+
+			completeBlockID, err = instance.CompleteBlock(op.blockID)
+			if completeBlockID != uuid.Nil {
```
joe-elliott, line 177: this should never return nil, so let's metric/log when it does.  we would want to investigate if this ever occurred.

use the existing flush errors?

modules/ingester/flush.go
```diff
@@ -150,12 +213,7 @@ func (i *Ingester) flushUserTraces(userID string) error {
 		return fmt.Errorf("instance id %s not found", userID)
 	}
 
-	for {
-		block := instance.GetBlockToBeFlushed()
-		if block == nil {
-			break
-		}
-
+	if block := instance.GetBlockToBeFlushed(blockID); block != nil {
```
joe-elliott, line 216: same here.  block should never be nil.  metric/log.

modules/ingester/instance.go
```diff
@@ -133,54 +133,77 @@ func (i *instance) CutCompleteTraces(cutoff time.Duration, immediate bool) error
...
+	}
+	completeBlockID := completeBlock.BlockMeta().BlockID
+	i.completeBlocks = append(i.completeBlocks, completeBlock)
+	i.blocksMtx.Unlock()
+
+	return completeBlockID, nil
 }
 
-func (i *instance) GetBlockToBeFlushed() *encoding.CompleteBlock {
+// GetBlocksToBeFlushed gets a list of blocks that can be flushed to the backend
```
joe-elliott, line 200: nit: `GetBlockToBeFlushed`

joe-elliott (approved):

mdisibio (approved):

PR #544, opened 2021-02-22
Refactor in preparation for block v2

joe-elliott:
**What this PR does**:
Continues refactoring the encoding/block packages.  I continue to really like the direction this is headed.  Some notes:

- Added a v1test in prep for changes.
- versionedEncoding interface is feeling really clean.  The next PR will add a `newIndexWriter` method that will be used to read/write index pages.
- common is a dumping ground of crap and needs to be seriously cleaned up
- The encoding package "reaches in" to use the raw v0 marshal/unmarshal functions.  I think that in time we may need a "object reader/writer" that handles this lowest level of marshalling/unmarshalling.  Right now everything at the base is just v0, but I expect that to eventually change.


mdisibio (approved):
PageWriter is a good addition.  lgtm.

PR #546, opened 2021-02-22
Go 1.16 + OTLPProto/Collector upgrades

mdisibio:
**What this PR does**:
This PR updates several OTEL dependencies and addresses some of the pain points. This is in the same line as another attempt #502 but goes further.

Changes:
1. Go 1.16
2. Update `opentelemetry-collector` from v0.6.1 to v0.16.0.
3. Update `opentelemetry-proto` submodule from v0.3.0 to v0.7.0. 
3. Update `otel/build-protobuf` docker image from 0.1.0 to 0.2.1.
5. **Change the way otel proto is handled:** Previously Tempo edited the OpenTelemetryCollector vendored code to mesh with `opentelemetry-proto`.   Now Tempo leaves all the dependencies alone and creates its own derivative types in /pkg/tempopb.  Remarshal/unmarshal is used to swap types in `shim.go`. This is unfortunate for efficiency but a significant win by decoupling things.  The reverse swap also occurs in `tempo-query`.

Additional Changes/Notes:
1. The goal is that a downstream app (i.e. grafana) could vendor /pkg/tempopb/ to pull in everything needed to communicate with Tempo with no other dependencies, and also importantly: no conflicts with other usages of OTEL proto or collector.
2. Go 1.16 update includes new `go mod tidy -e` parameter to continue despite errors. This eliminates another hurdle with dependency updates which was the cmux -> test -> grpc hello world issue as stated in the previous PR. This error still occurs but the use of `-e` argument allows build process to continue.  Note that Tempo still builds and tests successfully because the grpc hello world reference are in cmux tests which Tempo CI does not execute. Note this change also allows cleanup of the `vendor-fix/prometheus` work-around. 
3. Since there is no longer a link between vendoring and protos, `Makefile` was simplified to remove `vendor-dependencies`.  Now there are only `vendor-check` and `gen-protos`.

Questions / Next Steps:
1. This PR could be somewhat broken down into smaller PRs if desired. 
2. Testing:
Local testing was performed to cover most cases, if anything else needed please mention.
2.1 Unit tests
2.2. integration tests
2.3. Jaeger collector (using synthetic-load-generator)
2.4. Zipkin collector (using curl)
2.5. OTLP gRPC and HTTP collectors (using my own test app)
2.6 Grafana/tempo-query lookup
3. The proto swap in `shim.go` converts `otlp.ExportTraceServiceRequest` to `tempopb.Trace` which are technically wire-compatible. But Tempo does have its own definition of ExportTraceServiceRequest which we could use but needs a few more lines in makefile, etc.  It seemed simpler to do it this way.

None but likely introduces a few

**Checklist**
- [ ] Tests updated
- [x] Documentation added


Makefile
```diff
@@ -119,44 +119,61 @@ ifndef COMPONENT
...
 vendor-dependencies:
 	go mod vendor
-	go mod tidy
-	# ignore log.go b/c the proto version used by v0.6.1 doesn't actually have logs proto.
-	find . | grep 'vendor/go.opentelemetry.io.*go$\' | grep -v -e 'log.go$\' | xargs -L 1 sed -i $(SED_OPTS) 's+go.opentelemetry.io/collector/internal/data/opentelemetry-proto-gen/+github.com/open-telemetry/opentelemetry-proto/gen/go/+g'
+	go mod tidy -e
 	$(MAKE) gen-proto
 
 
 .PHONE: clear-protos
```
joe-elliott, line 172: i think we should either remove this target and just make it part of gen-proto or actually clear all generated protos here (pkg/tempopb and the intermediate folder) and make gen-proto dependent on it.

Makefile
```diff
@@ -119,44 +119,61 @@ ifndef COMPONENT
...
+	@echo -- Gen proto -- 
+	@echo --
+	$(call PROTO_GEN,$(PROTO_INTERMEDIATE_DIR)/common/v1/common.proto,./pkg/tempopb/)
+	$(call PROTO_GEN,$(PROTO_INTERMEDIATE_DIR)/resource/v1/resource.proto,./pkg/tempopb/)
+	$(call PROTO_GEN,$(PROTO_INTERMEDIATE_DIR)/trace/v1/trace.proto,./pkg/tempopb/)
+	$(call PROTO_GEN,pkg/tempopb/tempo.proto,./)
+
+	rm -rf $(PROTO_INTERMEDIATE_DIR)
 
 .PHONY: vendor-dependencies
```
joe-elliott, line 165: this target can go away, the only reason it exists is b/c revendoring was tied in to our proto generation.  if we do keep it let's remove the gen-proto part.

also, if you remove it ctrl+shift+f for references to it b/c i think we mention it in contributing.md and possibly other places.

Makefile
```diff
@@ -119,44 +119,61 @@ ifndef COMPONENT
...
 
 
 .PHONE: clear-protos
 clear-protos:
 	rm -rf opentelemetry-proto
 
+
 ### Check vendored files
 .PHONY: vendor-check
 vendor-check: clear-protos vendor-dependencies
```
joe-elliott, line 179: add gen-protos here (b/c we're removing from vendor-dependencies)

joe-elliott (commented):
Usual smattering of questions and thoughts.  Also:

- does this PR contain native support for OTLP in the distributor?  i think we want to go ahead and add that.  no idea what the additional load will look like with the bytes marshalling/unmarshalling in the shim.
- ctrl+shift+f for 1.15.  it is referenced in a few places that need to be updated.
- contributing.md in particular probably needs review b/c it's where instructions related to these changes will be contained.
- if we do add native support for grpc/http for otlp to this PR we need some documentation that explains why the shims are convenience layers for low volume situations, but if you want a true tracing backend you need a trace pipeline and you should push otlp to tempo only (and not through the shim)

modules/distributor/distributor_test.go
```diff
@@ -13,9 +13,9 @@ import (
 	"github.com/cortexproject/cortex/pkg/ring/kv"
 	"github.com/cortexproject/cortex/pkg/util/flagext"
 	"github.com/gogo/status"
-	v1_common "github.com/open-telemetry/opentelemetry-proto/gen/go/common/v1"
-	v1_resource "github.com/open-telemetry/opentelemetry-proto/gen/go/resource/v1"
-	v1 "github.com/open-telemetry/opentelemetry-proto/gen/go/trace/v1"
+	v1_common "github.com/grafana/tempo/pkg/tempopb/common/v1"
```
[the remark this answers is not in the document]
mdisibio: The package structure was kept identical to opentelemetry-proto.

Makefile
mdisibio: Done, swapped `vendor-dependencies` for `gen-protos`

Makefile
mdisibio: Agree removed vendor-dependencies and updated contributing.md.  It actually mentioned vendor-check which is still a good idea for contributers, to just updated the verbiage some.

Makefile
mdisibio: Agree, merged clear-protos into gen-protos and made it delete more thoroughly.

mdisibio:
> * ctrl+shift+f for 1.15.  it is referenced in a few places that need to be updated.
> * contributing.md in particular probably needs review b/c it's where instructions related to these changes will be contained.
Good catch, addressed.

> * does this PR contain native support for OTLP in the distributor?  i think we want to go ahead and add that.  no idea what the additional load will look like with the bytes marshalling/unmarshalling in the shim.
> * if we do add native support for grpc/http for otlp to this PR we need some documentation that explains why the shims are convenience layers for low volume situations, but if you want a true tracing backend you need a trace pipeline and you should push otlp to tempo only (and not through the shim)
Yes agree, we definitely want to migrate to native support to reclaim the efficiency (and for other reasons), but think it makes to do in a followup PR:  this one is already decently large and risk of drift, and benchmarking locally and in internal environment isn't show any performance regression in cpu or mem usage.

joe-elliott (approved):
best PR since the first one.

i've been waiting months to fix all this.  excellent work.

Issue #38, opened 2020-02-12
Improve querier concurrency between backend and ingesters


joe-elliott:
The querier first attempts to ask ingesters for a trace id.  If an ingester is temporarily unavailable this will timeout.  Add a tight timer or, even better, execute it concurrently with a backend query.

annanay25:
This is fixed by #557

PR #557, opened 2021-02-25
Query all ingesters to mitigate 404s on rollout

### LINKED ISSUE #284: 404s Temporarily returned when rolling ingesters
**Describe the bug**
Rolling the ingesters causes a trace in the headblock to return 404 temporarily and intermittently.

**To Reproduce**
1. repeatedly curl a traceid known to be in a headblock
2. kc rollout restart statefulset/ingester
3. note that the call will mostly succeed but occassionally 404

**Expected behavior**
With a replication factor of 2 and write/read consistency of 1.  The trace should be returned even when one of the ingesters it is on is being restarted.

**Environment:**

**Additional Context**

annanay25:

**What this PR does**:
This is an experimental PR to check if querying all ingesters helps mitigate 404s on rollouts/scaleups. Will also be monitoring performance as it queries a lot more ingesters every time.
The PR also simplifies the sharding logic and adds a QueryMode at the Querier. This can be set to "ingesters" / "blocks" / "all" (default) and will only query the specified components.

Notez: 
- after testing it in our internal environment, we did not see signs of a performance penalty, ingesters respond queriers quickly
- we need minimum of 2 shards as one is reserved for querying ingesters

Attempts to fix #284 #38 

**Checklist**
- [x] Tests updated
- NA Documentation added


cmd/tempo/app/modules.go
```diff
@@ -158,8 +158,11 @@ func (t *App) initQuerier() (services.Service, error) {
 }
 
 func (t *App) initQueryFrontend() (services.Service, error) {
-	var err error
+	if t.cfg.Frontend.QueryShards < frontend.MinQueryShards || t.cfg.Frontend.QueryShards > frontend.MaxQueryShards {
```
joe-elliott, line 161: any particular reason we're capping at 256?

modules/frontend/querysharding.go
```diff
@@ -53,7 +56,7 @@ func (s shardQuery) Do(r *http.Request) (*http.Response, error) {
 
 	// only need to initialise boundaries once
 	if len(s.blockBoundaries) == 0 {
-		s.blockBoundaries = createBlockBoundaries(s.queryShards)
+		s.blockBoundaries = createBlockBoundaries(s.queryShards - 1) // one shard will be used to query ingesters
```
joe-elliott, line 59: if this is initialized once let's move it to the `ShardingWare` function above where the struct is created instead of relying on this nil check.

modules/querier/http.go
```diff
@@ -109,6 +106,14 @@ func validateAndSanitizeRequest(r *http.Request) (string, string, bool, error) {
 		return "", "", false, fmt.Errorf("invalid value for queryIngesters %s", q)
 	}
 
+	// no need to validate/sanitize other parameters if queryIngesters is true
+	if queryIngesters {
```
joe-elliott, line 110: the querier should not quit validating if queryIngesters is true.  Yes, the query frontend no longer constructs queries that have both queryIngesters and block ids, but it's still a valid query that the querier has to take.

modules/querier/querier.go
```diff
@@ -171,34 +171,35 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
-	if err != nil {
-		return nil, errors.Wrap(err, "error querying store in Querier.FindTraceByID")
-	}
-
-	span.LogFields(ot_log.String("msg", "done searching store"))
-	// combine partialTraces with completeTrace
-	for _, partialTrace := range partialTraces {
-		storeTrace := &tempopb.Trace{}
-		err = proto.Unmarshal(partialTrace, storeTrace)
+		span.LogFields(ot_log.String("msg", "done searching ingesters"),
```
joe-elliott, line 176: leave this logic the way it was.  even though we no longer are creating queries with both blocks and ingester the querier should still respond correctly if it gets that query.

modules/frontend/querysharding.go
```diff
@@ -63,15 +66,15 @@ func (s shardQuery) Do(r *http.Request) (*http.Response, error) {
 	}
 
 	reqs := make([]*http.Request, s.queryShards)
-	for i := 0; i < len(s.blockBoundaries)-1; i++ {
+	for i := 0; i < s.queryShards; i++ {
```
joe-elliott, line 69: instead of creating block boundaries, it would be a lot cleaner to make a method that created a slice of structs that represented the query.  like:
```
type queryParams struct {
  queryShards bool
  startBlock []byte
  endBlock []byte
}
```
that would simplify this logic quite a bit and would be much easier to test as well.

joe-elliott (changes requested):

modules/querier/querier.go
annanay25: There's no way to differentiate between a query from the Query Frontend that has `queryIngesters=true` and another fired directly to the querier with the flag.. Are you saying we need to perform different operations for each of these?

I would prefer we don't optimise for the debugging use-case.

modules/querier/http.go
annanay25: Right, but we wouldn't be using the other parameters right? If they're empty they're set to `blockIDMin` and `blockIDMax`.. and I personally like that behaviour

cmd/tempo/app/modules.go
annanay25: As of now we're using only the first 8 bits of the blockID for sharding. We could extend that but it seems like a reasonable limit for number of shards

modules/querier/http.go
joe-elliott: Right now if you curl a querier at `/querier/api/traces/{traceID}` it will do what I expect and check the ingesters and all blocks.  With these changes I think queryIngesters would be set to true and this would only query the ingesters.  I would prefer keeping the previous behavior.

pkg/tempopb/tempo.proto
```diff
@@ -19,7 +19,7 @@ message TraceByIDRequest {
   bytes traceID = 1;
   string blockStart = 2;
   string blockEnd = 3;
-  bool queryIngesters = 4;
+  string queryMode = 4;
```
joe-elliott, line 22: for backwards compatibility you need to use message 5 here.  frankly unsure what would happen here.

modules/querier/http.go
```diff
@@ -92,30 +95,37 @@ func (q *Querier) TraceByIDHandler(w http.ResponseWriter, r *http.Request) {
...
-	end := r.URL.Query().Get(BlockEndKey)
-
-	// validate queryIngesters. it should either be empty or one of (true|false)
-	var queryIngesters bool
-	if len(q) == 0 || q == "true" {
-		queryIngesters = true
-	} else if q == "false" {
-		queryIngesters = false
+// return values are (blockStart, blockEnd, queryMode, error)
+func validateAndSanitizeRequest(r *http.Request) (string, string, string, error) {
```
joe-elliott, line 99: mode is a breaking change and should be noted.  i believe the only impact will be some rockiness on rollout.

modules/querier/http.go
```diff
@@ -92,30 +95,37 @@ func (q *Querier) TraceByIDHandler(w http.ResponseWriter, r *http.Request) {
...
+	var queryMode string
+	if len(q) == 0 || q == QueryModeAll {
+		queryMode = QueryModeAll
+	} else if q == QueryModeIngesters {
+		queryMode = QueryModeIngesters
+	} else if q == QueryModeBlocks {
+		queryMode = QueryModeBlocks
 	} else {
-		return "", "", false, fmt.Errorf("invalid value for queryIngesters %s", q)
+		return "", "", "", fmt.Errorf("invalid value for queryMode %s", q)
```
joe-elliott, line 111: nit:  be clear here.  the actual parameter is "mode"

joe-elliott (approved):
Adding an approval b/c I think this is basically done, but there are a few nits.

pkg/tempopb/tempo.proto
annanay25: Nice catch. Updating.

PR #563, opened 2021-03-01
Upped docs and tests to repl factor 3

joe-elliott:
**What this PR does**:
We have stopped using repl factor 2 in ops/prod and while I think it can remain an experimental option it is not supported.  Updated docs, examples and test to use repl factor 3.


annanay25:
Can we wait for #557 to be merged before this one? Has some overlap in tests.

annanay25 (approved):
lgtm!

dgzlopes (approved):

PR #570, opened 2021-03-05
Close gaps where append/completeblock file handles would rely on GC to close

### LINKED ISSUE #569: Occassionally ingesters fail to clear a head block
**Describe the bug**
After successfully flushing sometimes ingesters fail to clear the head block like they should.  The sequence of events is normally:

- Fail to flush head block
- Run out of disk space filling the next head block
- OOM
- Restart  and replay WAL

Sometimes on restart the ingester replays the wal, flushes it and continues on its merry way.  Sometimes the ingester enters a crash loop attempting to replay the WAL and requires manual intervention.

image

When Tempo is in this state we can see that the filesystem considers the file deleted:
```
lsof -n | grep deleted
1	/tempo	/var/tempo/wal/7a5357b8-13fd-47b9-80c0-15c7279958a1:1 (deleted)
1	/tempo	/var/tempo/wal/38e10487-d4fa-4fc5-9c6a-63a01ae5ffa8:1 (deleted)
1	/tempo	/var/tempo/wal/completed/898aff05-a7e9-4ff5-ace2-28eee5070971:1 (deleted)
1	/tempo	/var/tempo/wal/completed/e3279e02-ff5d-4d50-b05f-fdf48d32e03b:1 (deleted)
```

Likely there is a code path in the ingester that accidentally leaves a handle open to the head block.

mdisibio:
**What this PR does**:
This PR addresses a few instances where Append/CompleteBlock file handles would be left open and rely on GC to cleanup. The split of the deletion of WAL (AppendBlock) from creation of CompleteBlock also helps to ensure that AppendBlock.Clear is called, closing all handles.

Fixes #569

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * [BUGFIX] Fixes permissions errors on startup in GCS. [#554](https://github.com/grafana/tempo/pull/554)
 * [BUGFIX] Fixes error where Dell ECS cannot list objects. [#561](https://github.com/grafana/tempo/pull/561)
 * [BUGFIX] Fixes listing blocks in S3 when the list is truncated. [#567](https://github.com/grafana/tempo/pull/567)
-* [BUGFIX] [#569](https://github.com/grafana/tempo/issues/569)
+* [BUGFIX] Fixes where ingester may leave file open [#569](https://github.com/grafana/tempo/issues/569)
 
 ## v0.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * [BUGFIX] Fixes permissions errors on startup in GCS. [#554](https://github.com/grafana/tempo/pull/554)
 * [BUGFIX] Fixes error where Dell ECS cannot list objects. [#561](https://github.com/grafana/tempo/pull/561)
 * [BUGFIX] Fixes listing blocks in S3 when the list is truncated. [#567](https://github.com/grafana/tempo/pull/567)
-* [BUGFIX] Fixes where ingester may leave file open [#569](https://github.com/grafana/tempo/issues/569)
+* [BUGFIX] Fixes where ingester may leave file open [#570](https://github.com/grafana/tempo/pull/570)
 
 ## v0.6.0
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * [BUGFIX] Fixes permissions errors on startup in GCS. [#554](https://github.com/grafana/tempo/pull/554)
 * [BUGFIX] Fixes error where Dell ECS cannot list objects. [#561](https://github.com/grafana/tempo/pull/561)
 * [BUGFIX] Fixes listing blocks in S3 when the list is truncated. [#567](https://github.com/grafana/tempo/pull/567)
+* [BUGFIX] Fixes where ingester may leave file open [#569](https://github.com/grafana/tempo/issues/569)
```
joe-elliott, line 10: we normally link to the PR here

modules/ingester/instance.go
```diff
@@ -187,13 +187,20 @@ func (i *instance) CompleteBlock(blockID uuid.UUID) (uuid.UUID, error) {
 		i.blocksMtx.Unlock()
 		return uuid.Nil, err
 	}
+
+	err = completingBlock.Clear()
```
joe-elliott, line 191: very nice cleanup.  i didn't even recognize that @annanay25 's PR allowed this.  this might also make my wal replay PR viable.

this is inside the lock.  should we do it after?  i don't know the kernel well enough to know why this call might hang, but i guess it might hang?

mdisibio: Actually that is a good question.  The Clear() must occur before removing from the list, else it risks the same situation:  file left around unexpectedly (either deleted or not).  

Without making any changes to the overall error handling, it would be safe to move the lock down after the deletion.  The current implementation actually seems to take the lock too early as if an error occurs it immediately unlocks and exits anyways.

mdisibio: Actually the delete cannot occur outside of the lock and before the blocklist is updated or else Find could attempt to search block and the file is missing.    It works to do the delete last outside of lock.

joe-elliott (approved):

modules/ingester/instance.go
annanay25: Nice cleanup in this function!

PR #583, opened 2021-03-10
Query compacted blocks that fall within 2 * BlocklistPoll

### LINKED ISSUE #565: Tempo occasionally 404s on existing traces
**Describe the bug**
We have very rarely (<.1% of queries)  seen a trace 404 that actually existed in the backend.  Some example logs:

```
level=info ts=2021-03-02T18:35:10.996468172Z caller=frontend.go:63 method=GET traceID=14369e2468612583 url=/api/traces/7ec7cc871f20688b duration=1.708811458s status=404
level=info ts=2021-03-02T20:01:51.099858089Z caller=frontend.go:63 method=GET traceID=26c59851cbe88622 url=/api/traces/00000000000000007ec7cc871f20688b duration=2.547382198s status=200
level=info ts=2021-03-02T20:03:09.096726326Z caller=frontend.go:63 method=GET traceID=2f9c2851e68eff69 url=/api/traces/00000000000000007ec7cc871f20688b duration=4.932888762s status=200
```

The trace was initially created at 15:04 UTC and the first failed query occurred ~3.5 hours later.  Based on our internal operations the trace was definitely in the backend at the time.  Perhaps related to a polling sync issue?

@annanay25 

**Expected behavior**
Tempo to always return a trace correctly that is in the backend (or 500).

annanay25:

**What this PR does**:
Search blocks that have been compacted within `2 * BlocklistPoll`.
This was added because we were inadvertently skipping some blocks from search due to the nature of query sharding and blocklist polling in our architecture.

For a complete explanation do check the comment in the issue linked below!

Fixes #565 

**Checklist**
- [x] Tests updated
- [x] Documentation added


tempodb/tempodb.go
```diff
@@ -242,6 +242,15 @@ func (rw *readerWriter) Find(ctx context.Context, tenantID string, id common.ID,
 			copiedBlocklist = append(copiedBlocklist, b)
 		}
 	}
+
+	// if block is compacted within lookback period, include it in search
+	lookback := time.Now().Add(-(2 * rw.cfg.BlocklistPoll))
+	compactedBlocklist := rw.compactedBlockLists[tenantID]
+	for _, c := range compactedBlocklist {
+		if c.CompactedTime.After(lookback) {
```
joe-elliott, line 250: this also needs to take into account the block id range.  please do something similar to the `includeBlock` method and test (or extend the includeBlock method if possible)

tempodb/tempodb.go
```diff
@@ -460,23 +467,21 @@ func (rw *readerWriter) cleanMissingTenants(tenants []string) {
 		tenantSet[tenantID] = struct{}{}
 	}
 
+	rw.blockListsMtx.Lock()
```
joe-elliott, line 470: this will probably be fast enough that the simplicity is an improvement.  @dgzlopes do you have any thoughts?

tempodb/tempodb.go
```diff
@@ -529,3 +534,12 @@ func includeBlock(b *backend.BlockMeta, id common.ID, blockStart []byte, blockEn
 
 	return true
 }
+
+// if block is compacted within lookback period, and is within shard ranges, include it in search
+func includeCompactedBlock(c *backend.CompactedBlockMeta, id common.ID, blockStart []byte, blockEnd []byte, poll time.Duration) bool {
+	lookback := time.Now().Add(-(2 * poll))
+	if c.CompactedTime.After(lookback) {
```
joe-elliott, line 541: nit:  I think this is more idiomatic:

```
if c.CompactedTime.Before(lookback) {
  return false
}

return includeBlock(&c.BlockMeta, id, blockStart, blockEnd)
```

joe-elliott (approved):

tempodb/tempodb.go
annanay25: we need it for more than that, the internal blocklist maps are unprotected on the very next line (this was causing a data race), and so we need to put the lock around them.

tempodb/tempodb.go
annanay25: updated.

PR #585, opened 2021-03-10
Vulture Fixes

joe-elliott:
**What this PR does**:
- Adds org id to push
- Parameterizes vulture in jsonnet

### DIFF
--- a/operations/jsonnet/microservices/config.libsonnet
+++ b/operations/jsonnet/microservices/config.libsonnet
@@ -35,6 +35,9 @@
     },
     vulture: {
       replicas: 0,
+      tempoPushUrl: 'http://distributor:14250',
+      tempoQueryUrl: 'http://query-frontend:3100',
+      tempoOrgId: '',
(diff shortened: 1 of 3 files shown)


operations/jsonnet/microservices/config.libsonnet
```diff
@@ -35,6 +35,9 @@
     },
     vulture: {
       replicas: 0,
+      tempoPushUrl: 'http://distributor:14250',
+      tempoQueryUrl: 'http://query-frontend:3100',
+      tempoOrgId: '',
```
dgzlopes, line 40: Double-checking. We had org-id=1 and now it's empty. That's ok?

dgzlopes (approved):
LGTM. Added one question.

operations/jsonnet/microservices/config.libsonnet
joe-elliott: yup :+1:

PR #586, opened 2021-03-11
Vulture fixes

dgzlopes:
**What this PR does**:
- Fix goroutine leak by re-using the Jaeger gRPC client (moving it outside of the write loop).
- Change bucket name (from `failed` to `notfound`)

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo-vulture/main.go
+++ b/cmd/tempo-vulture/main.go
@@ -116,7 +116,7 @@ func main() {
 			if err != nil {
 				glog.Error("error querying Tempo ", err)
 				metricErrorTotal.Inc()
-				metricTracesErrors.WithLabelValues("failed").Inc()
+				metricTracesErrors.WithLabelValues("notfound").Inc()
 				continue
 			}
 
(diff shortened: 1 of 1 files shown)


joe-elliott (approved):
might be easier to wrap this in function and use defer...but no big deal either way

cmd/tempo-vulture/main.go
```diff
@@ -65,17 +65,16 @@ func main() {
 
 	// Write
 	go func() {
+		c, err := newJaegerGRPCClient(tempoPushURL)
+		if err != nil {
+			glog.Error("error creating grpc client", err)
+			metricErrorTotal.Inc()
```
dgzlopes, line 71: Well, if this happens, Vulture won't work... Idk if this is the best way to handle this...

Actually, It's going to crash.

dgzlopes: For example, if the tempoPushURL is malformed.

dgzlopes: Replaced this with panic...

joe-elliott (approved):
thanks!

PR #589, opened 2021-03-11
Add wildcard tenant

joe-elliott:
**What this PR does**:
Adds the ability set a wildcard override in our override file.  This will allow us to feel comfortable that defaults are coming from within the file and that the override file is loaded.


annanay25 (approved):
lgtm, nice addition!

annanay25 (commented):
just remembered, can we also add this to the overrides documentation? can be easily missed.

PR #592, opened 2021-03-15
Added PushBytes to the race test

joe-elliott:
**What this PR does**:
Adds `PushBytes()` to the `TestInstanceDoesNotRace()` test.  This should have been part of the previous PR, but it was Saturday and I didn't think about it.

Also confirmed that reverting the previous fix causes this test to fail.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


annanay25 (approved):
Nice, lgtm!

PR #593, opened 2021-03-15
Several fixes and improvements to block flushing, wal replay, file deletion, and logging

### LINKED ISSUE #357: Research flush backoff behavior
**Describe the bug**
A client has reported that if an ingester flush fails then Tempo has no backoff and will repeatedly slam the backend attempting to push the block.

I have personally not seen this, but we should investigate the behavior and improve it if necessary.

### LINKED ISSUE #579: Tempo leaves 0 length wal files forever
**Describe the bug**
On startup if Tempo hits a 0 length WAL file it will replay it.  This propagates a non-existent tenant forever:

```
-rw-r--r--    1 root     root             0 Mar  9 13:44 2486a7e8-4ed1-4e53-8cad-2f5f7a421697:16816
-rw-r--r--    1 root     root             0 Mar  9 13:44 36481724-33ae-4d90-a2e2-d579625b4c68:10598
-rw-r--r--    1 root     root             0 Mar  9 13:44 46243ee3-9e90-4915-bafd-847021f7f042:93
-rw-r--r--    1 root     root             0 Mar  9 13:44 464d523e-ebab-452f-9dc0-7d92844e17be:21377
-rw-r--r--    1 root     root         14202 Mar  9 13:51 5ba21364-2991-464c-ab60-3816e35e5704:16503
-rw-r--r--    1 root     root          3343 Mar  9 13:44 66db6993-f7a4-468f-b7e1-cec462c14f78:22348
-rw-r--r--    1 root     root       3964513 Mar  9 13:51 6e6435e8-3a94-4e7f-ba54-30aa45b9cc7d:167
-rw-r--r--    1 root     root       4740741 Mar  9 13:52 8f8dca97-4831-496c-b3f7-87cb1d87dc01:78
```

**Expected behavior**
On startup the ingester should ignore 0 length files and not recreate those tenants.

mdisibio:
**What this PR does**:
This PR improves some stability, error handling, and logging for various things:

Main changes:
* Introduce exponential backoff for flushing blocks to backend (s3/gcs/etc), with a max backoff of 120s.  Flushes are attempted forever
* Introduce exponential backoff for converting wal->complete blocks. We have seen this error due to a corrupted wal file when the disk was full.  Backoff is from 30s->60s with 3 attempts, yielding around 120 seconds of retry time, after which the wal file is permanently deleted and data is (probably) lost.
* Fix regression where the WAL file was deleted as soon as the complete block was created, instead of only after a successful flush to the backend. This was broken unintentionally in the previous round of flush changes.
* Introduce jitter up to 10s for regular flushes which will help when flushing many tenants. Flushes initiated by /flush or /shutdown do not have jitter.

Minor changes:
* Fix "file already closed" errors by nil'ing out file handles as soon as they are closed
* Fix "enqueue on closed queue" panic by forking priority queues from cortex and changing method to return error instead of panic.  Additionally to cut down on noisy errors, a guard flag is kept on the state of the flush queues, when shutting down the flusher abandons the tasks (no data is lost because wal still exists)
* Completed blocks now use the same ID as the source WAL block, which should make troubleshooting easier (i.e. file in /wal/completed/ matches its source file in /wal/)
* WAL replay doesn't create the ingester instance for that tenant until data is encountered, which lets 0-length files be replayed and deleted like expected (instead of creating a new instance+head block thus perpetuating the problem)
* Increment `ingester_failed_flushes_total` metric if error occurs deleting WAL block
* Various log message changes

Fixes #357 #579 #598 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/ingester/flush.go
+++ b/modules/ingester/flush.go
@@ -301,7 +301,10 @@ func (i *Ingester) enqueue(op *flushOp, jitter bool) {
 			return
 		}
 
-		i.flushQueues.Enqueue(op)
+		err := i.flushQueues.Enqueue(op)
+		if err != nil {
+			handleFailedOp(op, err)
+		}
 	}()
 }
 
@@ -328,6 +331,9 @@ func (i *Ingester) requeue(op *flushOp) {
 			return
 		}
 
-		i.flushQueues.Requeue(op)
+		err := i.flushQueues.Requeue(op)
+		if err != nil {
+			handleFailedOp(op, err)
+		}
 	}()
 }
--- a/modules/ingester/flush.go
+++ b/modules/ingester/flush.go
@@ -3,6 +3,7 @@ package ingester
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"net/http"
 	"strconv"
 	"time"
@@ -38,6 +39,7 @@ var (
 
 const (
 	initialBackoff      = 1 * time.Second
+	flushJitter         = 10 * time.Second
 	maxBackoff          = time.Minute
 	maxCompleteAttempts = 10
 )
@@ -141,9 +143,9 @@ func (i *Ingester) sweepInstance(instance *instance, immediate bool) {
 	}
 
 	if blockID != uuid.Nil {
-		i.flushQueues.Enqueue(&flushOp{
+		// jitter to help when flushing many instances at the same time
+		i.enqueueWithJitter(&flushOp{
 			kind:    opKindComplete,
-			at:      time.Now(),
 			userID:  instance.instanceID,
 			blockID: blockID,
 		})
@@ -174,21 +176,25 @@ func (i *Ingester) flushLoop(j int) {
 
 		if op.kind == opKindComplete {
 			level.Debug(log.Logger).Log("msg", "completing block", "userid", op.userID)
-			instance, exists := i.getInstanceByID(op.userID)
-			if !exists {
-				// instance no longer exists? that's bad, log and continue
-				level.Error(log.Logger).Log("msg", "instance not found", "userID", op.userID, "block", op.blockID.String())
+			instance, err := i.getOrCreateInstance(op.userID)
+			if err != nil {
+				handleFlushError(op, err)
 				continue
 			}
 
-			err := instance.CompleteBlock(op.blockID)
+			err = instance.CompleteBlock(op.blockID)
 			if err != nil {
 				handleFlushError(op, err)
 
 				if op.attempts >= maxCompleteAttempts {
 					level.Error(log.WithUserID(op.userID, log.Logger)).Log("msg", "Block exceeded max completion errors. Deleting. POSSIBLE DATA LOSS",
 						"userID", op.userID, "attempts", op.attempts, "block", op.blockID.String())
-					instance.ClearCompletingBlock(op.blockID)
+
+					err = instance.ClearCompletingBlock(op.blockID)
+					if err != nil {
+						// Failure to delete the WAL doesn't prevent flushing the bloc
+						handleFlushError(op, err)
+					}
 				} else {
 					retry = true
 				}
@@ -255,6 +261,17 @@ func (i *Ingester) flushBlock(userID string, blockID uuid.UUID) error {
 	}
 
 	return nil
+}
+
+func (i *Ingester) enqueueWithJitter(op *flushOp) {
+	delay := time.Duration(rand.Float32() * float32(flushJitter))
+
+	op.at = time.Now().Add(delay)
+
+	go func() {
+		time.Sleep(delay)
+		i.flushQueues.Enqueue(op)
+	}()
 }
 
 func (i *Ingester) requeue(op *flushOp) {
--- a/modules/ingester/flush.go
+++ b/modules/ingester/flush.go
@@ -3,7 +3,6 @@ package ingester
 import (
 	"context"
 	"fmt"
-	"math"
 	"net/http"
 	"strconv"
 	"time"
@@ -38,9 +37,9 @@ var (
 )
 
 const (
-	// Backoff for retrying 'immediate' flushes. Only counts for queue
-	// position, not wallclock time.
-	flushBackoff = 1 * time.Second
+	initialBackoff      = 1 * time.Second
+	maxBackoff          = time.Minute
+	maxCompleteAttempts = 10
 )
 
 const (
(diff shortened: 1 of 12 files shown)


modules/ingester/flush.go
```diff
@@ -140,7 +143,7 @@ func (i *Ingester) sweepInstance(instance *instance, immediate bool) {
 	if blockID != uuid.Nil {
 		i.flushQueues.Enqueue(&flushOp{
 			kind:    opKindComplete,
-			from:    math.MaxInt64,
+			at:      time.Now(),
```
joe-elliott, line 146: should we add jitter to desync flushes across a large number of tenants?

modules/ingester/flush.go
```diff
@@ -165,49 +168,64 @@ func (i *Ingester) flushLoop(j int) {
...
 
-		var completeBlockID uuid.UUID
-		var err error
 		if op.kind == opKindComplete {
 			level.Debug(log.Logger).Log("msg", "completing block", "userid", op.userID)
 			instance, exists := i.getInstanceByID(op.userID)
 			if !exists {
 				// instance no longer exists? that's bad, log and continue
-				level.Error(log.Logger).Log("msg", "instance not found", "tenantID", op.userID)
+				level.Error(log.Logger).Log("msg", "instance not found", "userID", op.userID, "block", op.blockID.String())
```
joe-elliott, line 180: is this possible?

should we call `getOrCreateInstance`? if we did that it would just fail on the next step since the block wouldn't exist?

modules/ingester/flush.go
```diff
@@ -38,9 +37,9 @@ var (
 )
 
 const (
-	// Backoff for retrying 'immediate' flushes. Only counts for queue
-	// position, not wallclock time.
-	flushBackoff = 1 * time.Second
+	initialBackoff      = 1 * time.Second
+	maxBackoff          = time.Minute
+	maxCompleteAttempts = 10
```
joe-elliott, line 42: 10 seems like a lot.  what are our scenarios:

- corrupt wal
- some kind of temporary disk i/o error
- ??

my gut says more like 3 is good

modules/ingester/flush.go
```diff
@@ -165,49 +168,64 @@ func (i *Ingester) flushLoop(j int) {
...
+					at:      time.Now(),
 					userID:  instance.instanceID,
-					blockID: completeBlockID,
+					blockID: op.blockID,
 				})
 			}
 
 		} else {
-			level.Debug(log.Logger).Log("msg", "flushing block", "userid", op.userID, "fp")
+			level.Info(log.Logger).Log("msg", "flushing block", "userid", op.userID, "block", op.blockID.String())
```
joe-elliott, line 206: i what you've done is the right choice, but am wondering if we will eventually want max attempts here too.

modules/ingester/instance.go
```diff
@@ -174,35 +174,44 @@ func (i *instance) CompleteBlock(blockID uuid.UUID) (uuid.UUID, error) {
...
-	i.completeBlocks = append(i.completeBlocks, completeBlock)
 	i.blocksMtx.Unlock()
 
-	err = completingBlock.Clear()
-	if err != nil {
-		level.Error(log.Logger).Log("msg", "Error clearing wal", "tenantID", i.instanceID, "err", err)
+	if completingBlock != nil {
+		err := completingBlock.Clear()
+		if err != nil {
+			level.Error(log.Logger).Log("msg", "Error clearing wal", "tenantID", i.instanceID, "blockID", blockID.String(), "err", err)
```
joe-elliott, line 210: should we metric this somewhere? if we're unable to delete files for some reason and they start piling up would be nice to have some visibility.

maybe return err from this method and increment failed flushes in the calling code? (but don't actually fail the flush, just log and metric)

joe-elliott (commented):
Nice improvements. Thank you for catching that bug where we leave wal files hanging around for empty tenants.  That has been bugging me for awhile.

Some comments/thoughts but overall looks good.

modules/ingester/flush.go
mdisibio: Was thinking of disk full scenario: the wal is good but the new complete block can't be written.  Retrying over a span of a few minutes seems that it could allow time for other operations to free up space (remove completed blocks, blocks being replayed, etc).   We can increase the initial backoff and reduce the number of attempts, and still have similar retry window, does that sound better?

modules/ingester/flush.go
```diff
@@ -165,49 +170,68 @@ func (i *Ingester) flushLoop(j int) {
...
-				level.Error(log.Logger).Log("msg", "instance not found", "tenantID", op.userID)
+			instance, err := i.getOrCreateInstance(op.userID)
+			if err != nil {
+				handleFlushError(op, err)
 				continue
 			}
 
-			completeBlockID, err = instance.CompleteBlock(op.blockID)
-			if completeBlockID != uuid.Nil {
+			err = instance.CompleteBlock(op.blockID)
```
annanay25, line 185: I think the blockID for the `CompleteBlock` will not be the same as the completingBlock, we create a new uuid here - https://github.com/grafana/tempo/blob/fc4f477fb09507f54b39b9bdd75f594f0e9f2b42/tempodb/encoding/complete_block.go#L40

mdisibio: This PR changes that method to keep the same uuid.

modules/ingester/flush.go
mdisibio: Don't think it's possible, but went ahead and more properly handled the error.

modules/ingester/flush.go
```diff
@@ -229,10 +266,68 @@ func (i *Ingester) flushBlock(userID string, blockID uuid.UUID) error {
...
+		delay = time.Duration(rand.Float32() * float32(flushJitter))
+	}
+
+	op.at = time.Now().Add(delay)
+
+	go func() {
+		time.Sleep(delay)
+
+		// Check if shutdown initiated
+		if i.flushQueues.IsStopped() {
```
joe-elliott, line 299: technically still a race here which sucks. should we PR a fix upstream?

or (dare I say it) recover the panic?

joe-elliott (approved):
one minor concern, but you are welcome to merge as is.

modules/ingester/flush.go
mdisibio: Talked to @gouthamve and the PriorityQueue is basically part of the chunks storage engine and considered maintenance mode, so first instinct was to fork it to Tempo.   That is done, now returns error instead of panic.  Leaving in the guard bool because I think it is still cleaner, to know that work is being abandoned specifically versus some other error within the queue.

joe-elliott (approved):

PR #607, opened 2021-03-23
Memberlist: removed unncessary postfix

joe-elliott:
**What this PR does**:
Previously we were postfixing the memberlist name with a guid to force uniqueness on restarts. This functionality, however, has been added to memberlist directly so let's remove this confusing bit of code.

https://github.com/grafana/tempo/blob/master/vendor/github.com/cortexproject/cortex/pkg/ring/kv/memberlist/memberlist_client.go#L131

### DIFF
--- a/cmd/tempo/app/modules.go
+++ b/cmd/tempo/app/modules.go
@@ -3,7 +3,6 @@ package app
 import (
 	"fmt"
 	"net/http"
-	"os"
 
 	"github.com/cortexproject/cortex/pkg/cortex"
 	cortex_frontend "github.com/cortexproject/cortex/pkg/frontend"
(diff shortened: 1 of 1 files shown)


cmd/tempo/app/modules.go
```diff
@@ -229,13 +227,6 @@ func (t *App) initMemberlistKV() (services.Service, error) {
 		ring.GetCodec(),
 	}
 
-	hostname, err := os.Hostname()
```
dgzlopes, line 232: Double-checking :)

The package does this by default too? Seeing the commit history, I suppose you tried it!

dgzlopes (approved):
LGTM. Left one question.

cmd/tempo/app/modules.go
joe-elliott: yup, i removed it, added it back and removed it again.

the package does indeed already use hostname by default. good :eyes: !

PR #628, opened 2021-04-05
Ingester block lifetime changes

mdisibio:
**What this PR does**:
This PR redoes some of the block lifecycle on the ingester to improve stability and startup times.  

Previously the ingester would convert the WAL into a CompleteBlock which persisted the data on disk but kept the index and bloom filters in memory.  After a restart the blocks would have to be recreated from the WALs since memory was lost, which caused both heavy performance hit on startup, but also an inability to query data that had been flushed but no longer in the WAL.

Now the ingester uses a local backend mounted under /wal/blocks. The WAL is flushed to this local backend first, which fully persists the data/index/bloom, etc. Then it is flushed to the remote backend.    On restart any existing blocks are simply reloaded and the ingester resumes where it was. The flushed state of a local block is persisted with a timestamp in a dedicated `flushed` file.

No configuration or hosting changes are needed. The local backend is mounted under /wal/blocks/ by reusing the existing wal storage location, and it also uses the existing block encoding/downsampling/flushing options.

#633 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/tempodb.go
```diff
@@ -201,8 +204,70 @@ func (rw *readerWriter) WriteBlock(ctx context.Context, c WriteableBlock) error
...
+// CompleteBlock iterates the given WAL block but flushes it to the given backend instead of the default TempoDB backend. The
+// new block will have the same ID as the input block.
+func (rw *readerWriter) CompleteBlockWithBackend(ctx context.Context, block *wal.AppendBlock, combiner common.ObjectCombiner, r backend.Reader, w backend.Writer) (*encoding.BackendBlock, error) {
+	meta := block.Meta()
+	blockID := meta.BlockID
+	tenantID := meta.TenantID
+
+	// Default and nil check is primarily to make testing easier.
+	flushSize := DefaultFlushSizeBytes
+	if rw.compactorCfg != nil && rw.compactorCfg.FlushSizeBytes > 0 {
```
joe-elliott, line 221: will this ever be set? i think only the ingester will use this functionality which will not have a compactorCfg

tempodb/backend/local/local.go
```diff
@@ -14,30 +14,39 @@ import (
...
+func New(cfg *Config) (backend.Reader, backend.Writer, backend.Compactor, error) {
+	l, err := NewBackend(cfg)
+	return l, l, l, err
 }
 
 // Write implements backend.Writer
-func (rw *readerWriter) Write(ctx context.Context, name string, blockID uuid.UUID, tenantID string, buffer []byte) error {
-	return rw.WriteReader(ctx, name, blockID, tenantID, bytes.NewBuffer(buffer), int64(len(buffer)))
+func (rw *Backend) Write(ctx context.Context, name string, blockID uuid.UUID, tenantID string, buffer []byte) error {
+	return rw.WriteReader(ctx, name, blockID, tenantID, bytes.NewBuffer(buffer), 0)
```
joe-elliott, line 45: any particular reason to drop the `len(buffer)` here?

tempodb/encoding/backend_block.go
```diff
@@ -125,3 +126,50 @@ func (b *BackendBlock) NewIndexReader() (common.IndexReader, error) {
...
+	}
+
+	// Index
+	err = copy(nameIndex)
+	if err != nil {
+		return err
+	}
+
+	// Meta
+	err = w.WriteBlockMeta(ctx, b.meta)
```
joe-elliott, line 173: why use this backend block's meta and not copy from the reader?

tempodb/encoding/backend_block.go
```diff
@@ -125,3 +126,50 @@ func (b *BackendBlock) NewIndexReader() (common.IndexReader, error) {
...
+
+func (b *BackendBlock) BlockMeta() *backend.BlockMeta {
+	return b.meta
+}
+
+func (b *BackendBlock) Reader() backend.Reader {
+	return b.reader
+}
+
+func (b *BackendBlock) Write(ctx context.Context, r backend.ReaderReader, w backend.Writer) error {
```
joe-elliott, line 138: this method is copying from `ReaderReader` to `Writer` and uses the block meta and tenant id?

i think this will be confusing given that the "real" backend write methods are all in `block.go` and this is more like a copy from backend reader to writer method.

tempodb/encoding/streaming_block.go
```diff
@@ -25,17 +25,13 @@ type CompactorBlock struct {
...
 
-// NewCompactorBlock creates a ... new compactor block!
-func NewCompactorBlock(cfg *BlockConfig, id uuid.UUID, tenantID string, metas []*backend.BlockMeta, estimatedObjects int) (*CompactorBlock, error) {
+// NewStreamingBlock creates a ... new streaming block. Objects are appended one at a time to the backend.
+func NewStreamingBlock(cfg *BlockConfig, id uuid.UUID, tenantID string, metas []*backend.BlockMeta, estimatedObjects int) (*StreamingBlock, error) {
 	if len(metas) == 0 {
 		return nil, fmt.Errorf("empty block meta list")
 	}
 
-	if estimatedObjects <= 0 {
```
joe-elliott, line 34: any particular reason this was removed?

joe-elliott (commented):
Have some initial thoughts. It makes sense to me to move the "LocalBlock" and related functionality inside the `/tempodb/wal` package and provide methods for the ingester/instance to use.

tempodb/wal/wal.go
```diff
@@ -22,6 +23,7 @@ type WAL struct {
 type Config struct {
 	Filepath          string `yaml:"path"`
 	CompletedFilepath string
```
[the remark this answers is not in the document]
mdisibio: It seems necessary to perform the final wipe on startup after rolling out this change, which means it needs to stay. Agree we can use the same folder if we change the wipe behavior to only delete files and not folders. The local backend  `Tenants()` method already correctly considers folders only.

tempodb/tempodb.go
mdisibio: Single-binary deployments could have it, and also all unit/integration tests.

tempodb/encoding/backend_block.go
mdisibio: The metadata file name and location is actually internal to each backend and handled with a separate `WriteBlockMeta` method.  There is not a wider const for it.  This was discussed during previous refactorings of the backend, and the rationale at the time was that the block metadata is outside any encoding version, so it required a dedicated method.  I think we could create a const value for it and standardize the backends, but it might not be worth adding to this PR.

tempodb/encoding/streaming_block.go
mdisibio: There are quite a few tests that create empty WAL/complete blocks and expect it to work.  It used to because CompleteBlock did not enforce at least one trace to be present, even though it also created a bloom filter, because the bloom filter accounts for it here:  https://github.com/grafana/tempo/blob/master/tempodb/encoding/common/bloom.go#L23

Now that it is switched to Compactor/StreamingBlock under the hood, the tests do not pass this check.  Removing this check seems preferable, as the bloom filter will "work" regardless, and the behavior expected by the tests is preserved.

tempodb/encoding/backend_block.go
mdisibio: Agree, moved to a general purpose method in `block.go`

modules/ingester/local_block.go
```diff
@@ -0,0 +1,84 @@
+package ingester
```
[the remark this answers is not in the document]
mdisibio: Agree, moved to the wal package

modules/ingester/flush.go
```diff
@@ -200,20 +201,26 @@ func (i *Ingester) flushLoop(j int) {
 
 					err = instance.ClearCompletingBlock(op.blockID)
```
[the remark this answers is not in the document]
mdisibio: Agree, moved to a dedicated method.

tempodb/wal/local_block.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func NewLocalBlock(ctx context.Context, existingBlock *encoding.BackendBlock, l *local.Backend) (*LocalBlock, error) {
+
+	c := &LocalBlock{
+		BackendBlock: *existingBlock,
+		local:        l,
+	}
+
+	flushedBytes, err := existingBlock.Reader().Read(ctx, "flushed", c.BlockMeta().BlockID, c.BlockMeta().TenantID)
```
joe-elliott, line 32: I think there are two paths here that I would prefer over exposing the block Reader.

1) Providing a methods that reads/write generic files to a backend block and use that instead of exposing the reader. Something like:
```
ReadNamedObject(context, string) ([]byte, error)
WriteNamedObject(context, string, []byte) (error)
```

2) Just use `c.local.Read()` like you did with `c.local.Write()`

tempodb/wal/local_block.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (c *LocalBlock) SetFlushed(ctx context.Context) error {
+	flushedTime := time.Now()
+	flushedBytes, err := flushedTime.MarshalText()
+	if err != nil {
+		return errors.Wrap(err, "error marshalling flush time to text")
+	}
+
+	err = c.local.Write(ctx, "flushed", c.BlockMeta().BlockID, c.BlockMeta().TenantID, flushedBytes)
```
joe-elliott, line 65: nit: make "flushed" a const

Possibly replace this with `c.WriteNamedObject`. See comment above.

tempodb/encoding/backend_block.go
```diff
@@ -125,3 +125,11 @@ func (b *BackendBlock) NewIndexReader() (common.IndexReader, error) {
 
 	return reader, nil
 }
+
+func (b *BackendBlock) BlockMeta() *backend.BlockMeta {
+	return b.meta
+}
+
+func (b *BackendBlock) Reader() backend.Reader {
```
joe-elliott, line 133: I think we should remove this.  See comment below in `local_block.go`

joe-elliott (commented):
Thank you for the changes. This is definitely getting closer.

Some mild refactoring comments, but other than that I'm really liking where this is.

tempodb/tempodb.go
```diff
@@ -83,7 +85,8 @@ var (
 
 type Writer interface {
 	WriteBlock(ctx context.Context, block WriteableBlock) error
-	CompleteBlock(block *wal.AppendBlock, combiner common.ObjectCombiner) (*encoding.CompleteBlock, error)
+	CompleteBlock(block *wal.AppendBlock, combiner common.ObjectCombiner) (*encoding.BackendBlock, error)
```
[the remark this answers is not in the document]
mdisibio: `CompleteBlock` is used by a number of tests.

tempodb/wal/local_block.go
mdisibio: Went with option 2

tempodb/wal/wal_test.go
```diff
@@ -135,7 +134,7 @@ func TestAppend(t *testing.T) {
 	assert.Equal(t, numMsgs, i)
 }
 
-func TestAppendBlockComplete(t *testing.T) {
+/*func TestAppendBlockComplete(t *testing.T) {
```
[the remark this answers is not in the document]
mdisibio: Agree, reinstated this test under tempodb_test.go to match where the logic went.

modules/ingester/ingester.go
```diff
@@ -66,12 +67,19 @@ func New(cfg Config, store storage.Store, limits *overrides.Overrides) (*Ingeste
 		flushQueues: flushqueues.New(cfg.ConcurrentFlushes, metricFlushQueueLength),
 	}
 
+	l, err := local.NewBackend(&local.Config{
```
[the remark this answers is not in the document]
mdisibio: Moved the local backend creation from ingester to WAL.

joe-elliott (approved):

PR #638, opened 2021-04-09
Versioned WAL

### LINKED ISSUE #418: Add WAL Checksums

joe-elliott:
**What this PR does**:
Reuses all of the existing versioned code to enabled versioned WALs. This includes the ability to compress the WAL which addresses #418.

Benchmarks:
```
goos: linux
goarch: amd64
pkg: github.com/grafana/tempo/tempodb/wal
cpu: Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
BenchmarkWALNone-8     	     132	   8162877 ns/op	 2543406 B/op	   38124 allocs/op
BenchmarkWALSnappy-8   	     135	   8712571 ns/op	 2657195 B/op	   36621 allocs/op
BenchmarkWALLZ4-8      	      84	  13025362 ns/op	 5149115 B/op	   49653 allocs/op
BenchmarkWALGZIP-8     	      36	  34852004 ns/op	 3893758 B/op	   40209 allocs/op
BenchmarkWALZSTD-8     	      21	  47785689 ns/op	49718721 B/op	  112807 allocs/op
PASS
ok  	github.com/grafana/tempo/tempodb/wal	7.972s
```

The benchmarks do not reveal a clear winner here so I'm leaving `None` as the default and documenting the ability to turn this on. We will experiment with snappy internally.

Fixes #418

**Checklist**
- [x] Tests updated
- [x] Documentation added


tempodb/encoding/iterator_wal.go
```diff
@@ -0,0 +1,44 @@
+package encoding
+
+import (
+	"context"
+	"io"
+
+	"github.com/grafana/tempo/tempodb/encoding/common"
+)
+
+type walIterator struct {
```
mdisibio, line 10: Could/should this be in the wal package?

mdisibio (commented):
Left a few thoughts

example/docker-compose/etc/tempo-s3-minio.yaml
```diff
@@ -40,6 +40,7 @@ storage:
       encoding: zstd                   # block encoding/compression.  options: none, gzip, lz4-64k, lz4-256k, lz4-1M, lz4, snappy, zstd
     wal:
       path: /tmp/tempo/wal             # where to store the the wal locally
+      encoding: snappy                 # wal encoding/compression.  options: none, gzip, lz4-64k, lz4-256k, lz4-1M, lz4, snappy, zstd
```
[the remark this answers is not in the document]
joe-elliott: agree. i've found that people copy our example configs directly quite a bit. I think we should be more careful about just having them be all defaults.

tempodb/encoding/iterator_wal.go
joe-elliott: Renamed to "recordless" iterator. Also reviewing the iterators I think "paged" and "record" iterator could be combined fairly easily.

mdisibio (approved):

PR #641, opened 2021-04-12
Make all Iterators return io.EOF

joe-elliott:
**What this PR does**:
When working on the versioned WAL I noticed that the iterators inconsistently returned io.EOF. The deduping and standard iterator just returned all `nil` when complete. This is kind of a scary change, but important that all iterators behave the same.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/encoding/finder_paged.go
+++ b/tempodb/encoding/finder_paged.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"io"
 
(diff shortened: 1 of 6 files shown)


mapno (approved):
LGTM

mdisibio (approved):
Approved.  We discussed internally and returning `nil` for id would also work to signal the end of the iterators.  However, since neither option is preferred or more idiomatic, we are good with this.

PR #644, opened 2021-04-12
Add configurable bloom filters

### LINKED ISSUE #419: Make Bloom Filter Sharding Configurable
Currently bloom filters always default to 10 shards.  Make this configurable.  

I believe the best solution is to add a number of shards to meta.json.  If this value is not present we can just assume 10.

**Additional context**
Currently we have Bloom FP rate of 5% which is extremely high.  We'd love to push this down around 1%, but this will inflate our Bloom filter size which reduce query performance.  A configurable sharding rate will allow us to push our FP rate lower.

annanay25:

**What this PR does**:
Provides a way to shard out the bloom filter based on the false positive rate and shard size provided.

Fixes #419

**Checklist**
- [x] Tests updated
- NA Documentation added


tempodb/encoding/common/bloom.go
```diff
@@ -48,16 +81,20 @@ func (b *ShardedBloomFilter) WriteTo() ([][]byte, error) {
...
 }
 
-func GetShardNum() int {
-	return shardNum
+func ShardKeyForTraceID(traceID []byte, shardCount int) int {
+	return int(util.TokenForTraceID(traceID)) % ValidateShardCount(shardCount)
+}
+
+// For backward compatibility
+func ValidateShardCount(shardCount int) int {
```
joe-elliott, line 95: i think this functionality would be better placed in the marshalling code for BlockMeta. If the field does not exist then choose `legacyShardCount`.

tempodb/encoding/common/bloom.go
```diff
@@ -2,41 +2,64 @@ package common
```
annanay25, line 2: This is the main file that needs reviewing

tempodb/encoding/common/bloom.go
```diff
@@ -2,41 +2,64 @@ package common
...
+}
+
+// NewBloom creates a ShardedBloomFilter
+func NewBloom(fp float64, shardSize, estimatedObjects uint) *ShardedBloomFilter {
+	// estimate the number of shards needed. an approximate value is enough
+	var shardCount uint
+	var kPerBloom uint
+	for {
+		shardCount++
+		b := bloom.New(shardSize*8, uint(evaluateK(int(shardSize*8), int(estimatedObjects/shardCount))))
```
annanay25, line 36: Need to reconsider this. I could alternately write some math formulae that estimate it, its just easier this way but tests by creating new blooms

tempodb/encoding/common/bloom_test.go
```diff
@@ -52,9 +56,60 @@ func TestShardedBloom(t *testing.T) {
...
+			// parse byte representation into willf_bloom.Bloomfilter
+			var filters []*willf_bloom.BloomFilter
+			for i := 0; i < b.GetShardCount(); i++ {
+				filters = append(filters, &willf_bloom.BloomFilter{})
+			}
+
+			for i, singleBloom := range bloomBytes {
+				_, err = filters[i].ReadFrom(bytes.NewReader(singleBloom))
+				assert.NoError(t, err)
+				assert.LessOrEqual(t, filters[i].EstimateFalsePositiveRate(tt.estimatedObjects/uint(b.GetShardCount())), tt.bloomFP)
```
annanay25, line 111: Added as a separate test because `EstimateFalsePositiveRate` wipes the bloom entries

tempodb/encoding/common/bloom.go
annanay25: Swapped this out. Using the package function `bloom.EstimateParameters` instead.

annanay25:
something to note - with the following parameters:
```
bloomFP:          0.05,
shardSize:        250*1024,
estimatedObjects: 10_000_000,
```

the package creates 244 shards. we'll just have a million bloom files in object storage

tempodb/encoding/common/bloom.go
```diff
@@ -3,40 +3,53 @@ package common
...
 
-func NewWithEstimates(n uint, fp float64) *ShardedBloomFilter {
-	b := &ShardedBloomFilter{
-		blooms: make([]*bloom.BloomFilter, shardNum),
+// NewBloom creates a ShardedBloomFilter
+func NewBloom(fp float64, shardSize, estimatedObjects uint) *ShardedBloomFilter {
+	// estimate the number of shards needed. an approximate value is enough
+	var shardCount uint
+	var kPerBloom uint
+	for {
```
joe-elliott, line 22: https://github.com/bits-and-blooms/bloom/blob/master/bloom.go#L115

We should be able to calculate the shard size directly.  We don't need this loop.

tempodb/encoding/common/bloom.go
```diff
@@ -3,40 +3,53 @@ package common
...
 
 type ShardedBloomFilter struct {
 	blooms []*bloom.BloomFilter
 }
 
-func NewWithEstimates(n uint, fp float64) *ShardedBloomFilter {
-	b := &ShardedBloomFilter{
-		blooms: make([]*bloom.BloomFilter, shardNum),
+// NewBloom creates a ShardedBloomFilter
+func NewBloom(fp float64, shardSize, estimatedObjects uint) *ShardedBloomFilter {
```
joe-elliott, line 18: how reliable is estimatedObjects? I feel like we do a pretty good job of passing a good value in here, but not 100% sure.

tempodb/encoding/block.go
```diff
@@ -81,7 +81,7 @@ func CopyBlock(ctx context.Context, meta *backend.BlockMeta, src backend.Reader,
 	}
 
 	// Bloom
-	for i := 0; i < common.GetShardNum(); i++ {
+	for i := 0; i < common.ValidateShardCount(int(meta.BloomShardCount)); i++ {
```
joe-elliott, line 84: this is weird b/c there are two different values here. what's in the meta and what's in common.ShardedBloomFilter. I think this method should return error if these values are different.

tempodb/encoding/streaming_block_test.go
```diff
@@ -59,7 +59,8 @@ func TestStreamingBlockAddObject(t *testing.T) {
 
 	numObjects := (rand.Int() % 20) + 1
 	cb, err := NewStreamingBlock(&BlockConfig{
-		BloomFP:              .01,
+		BloomFP:              0.01,
+		BloomFilterShardSize: 100_000,
```
joe-elliott, line 63: let's make this much smaller. 1000? 10_000? just to make sure we're getting multiple shards.

tempodb/backend/block_meta.go
```diff
@@ -28,6 +28,7 @@ type BlockMeta struct {
 	IndexPageSize   uint32    `json:"indexPageSize"`   // Size of each index page in bytes
 	TotalRecords    uint32    `json:"totalRecords"`    // Total Records stored in the index file
 	DataEncoding    string    `json:"dataEncoding"`    // DataEncoding is a string provided externally, but tracked by tempodb that indicates the way the bytes are encoded
+	BloomShardCount uint16    `json:"bloomShardCount"` // Number of bloom filter shards
```
joe-elliott, line 31: perhaps just `bloomShards`. will match `totalRecords` and `totalObjects`

joe-elliott (commented):
Some thoughts.

I didn't dig super hard on the tests, but I think `testStreamingBlockToBackendBlock` will test that the bloom filters are functioning correctly.

> the package creates 244 shards. we'll just have that many bloom files in object storage
Yeah, I think this will be ok :). If not we're about to find out.

tempodb/encoding/block.go
annanay25: Here the bloom is present only on the `src` backend in the byte representation. There's no `ShardedBloomFilter` representation, not sure if I'm missing something

tempodb/encoding/common/bloom.go
annanay25: I am using exactly the same function `bloom.EstimateParameters`. Do you mean the following?
```
m, k := bloom.EstimateParameters(estimatedObjects, fp)

shardCount := m / (shardSize * 8)
```
I'm not sure if it works that way, does a smaller bloom with the same `k` have the same `fp`? I'll have to review that

tempodb/encoding/common/bloom.go
annanay25: This is accurate at the ingesters, but at the compactors the value is a sum of the object count present in the individual blocks.. so estimatedObjects will be > actualObjects

tempodb/encoding/common/bloom.go
annanay25: Agree on this but unfortunately, moving `ValidateShardCount` the block meta file results in a cyclic import.. 
```
package github.com/grafana/tempo/tempodb/encoding
	imports github.com/grafana/tempo/tempodb/backend
	imports github.com/grafana/tempo/tempodb/encoding/common
	imports github.com/grafana/tempo/tempodb/backend: import cycle not allowed
```
Will place it back here

tempodb/encoding/block.go
joe-elliott: yup, you're right.

tempodb/encoding/common/bloom.go
joe-elliott: I believe it's possible:

https://hur.st/bloomfilter/

tempodb/encoding/common/bloom.go
joe-elliott: i suppose that means we're going to be creating smaller bloom filters than expected and consequently more than we need.

easiest thing to do here may just be to push the change and see the impact in practice.

modules/storage/config.go
```diff
@@ -37,7 +38,8 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
 	cfg.Trace.WAL.Encoding = backend.EncNone
 
 	cfg.Trace.Block = &encoding.BlockConfig{}
-	f.Float64Var(&cfg.Trace.Block.BloomFP, util.PrefixConfig(prefix, "trace.block.bloom-filter-false-positive"), .05, "Bloom False Positive.")
+	f.Float64Var(&cfg.Trace.Block.BloomFP, util.PrefixConfig(prefix, "trace.block.bloom-filter-false-positive"), .05, "Bloom Filter False Positive.")
+	f.IntVar(&cfg.Trace.Block.BloomFilterShardSize, util.PrefixConfig(prefix, "trace.block.bloom-filter-shard-size"), 250*1024, "Bloom Filter Shard Size.")
```
mdisibio, line 42: Should probably be named `bloom-filter-shard-size-bytes` to follow convention

tempodb/encoding/common/bloom.go
mdisibio: The overage during typical compaction is probably a few percent, but when compacting level 0 blocks it will be much larger.  I.e. worst case is RF3 compacting 3 identical blocks, then it will be 3x needed.   Playing with a bloom filter calculator, keep constant filter size, but with 1/3rd the number of entries changes FP from 0.01 -> 0.004.

tempodb/encoding/common/bloom.go
mdisibio: It should err on the side of higher precision by rounding up the number of shards, i.e. `ceil(m / (shardSize * 8))`.  1/10th m with 1/10th objects is very close to the same FP, should be fine as long as shard count is rounded up.

tempodb/encoding/config.go
```diff
@@ -11,6 +11,7 @@ type BlockConfig struct {
 	IndexDownsampleBytes int              `yaml:"index_downsample_bytes"`
 	IndexPageSizeBytes   int              `yaml:"index_page_size_bytes"`
 	BloomFP              float64          `yaml:"bloom_filter_false_positive"`
+	BloomFilterShardSize int              `yaml:"bloom_filter_shard_size"`
```
mdisibio, line 14: See other comment, should probably be `bloom_filter_shard_size_bytes` to follow convention.

mdisibio (changes requested):
Nice work, this is much smoother than expected!  Left a few comments.

tempodb/encoding/common/bloom.go
annanay25: Yep, will round up 👍

tempodb/encoding/common/bloom.go
annanay25: Right. That's true. I think this is acceptable as well, level 0 blocks will be quickly crunched together, we can afford to have a lower fp and higher shards, should not affect performance.

joe-elliott (commented):
Two final thoughts and then I'm good.  @mdisibio anything else?

tempodb/wal/append_block.go
```diff
@@ -138,6 +138,7 @@ func newAppendBlockFromFile(filename string, path string) (*AppendBlock, error,
 
 	common.SortRecords(records)
 
+	b.meta.TotalObjects = len(records)
```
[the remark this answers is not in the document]
mdisibio: Correct, it wasn't set correctly but was masked by bloom filter logic that ensured object count was >= 1.

tempodb/encoding/common/bloom.go
```diff
@@ -2,41 +2,63 @@ package common
...
+	shardCount = uint(math.Ceil(float64(m) / (float64(shardSize) * 8.0)))
+
+	if shardCount < minShardCount {
+		shardCount = minShardCount
 	}
 
-	itemsPerBloom := n / shardNum
-	if itemsPerBloom == 0 {
-		itemsPerBloom = 1
+	if shardCount > maxShardCount {
```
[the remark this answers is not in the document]
mdisibio: Was thinking the same, but maybe 1000 which is a decent buffer over the 244 bloom filters example and other workloads that might be out there.

mdisibio:
LGTM after reducing `maxShardCount`.

mdisibio (approved):

tempodb/wal/append_block.go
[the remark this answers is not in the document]
annanay25: 👍  updated.

PR #654, opened 2021-04-19
Add kubernetes manifests for quick deployment

annanay25:

**What this PR does**:
- Adds k8s manifests for quick deployments
  - Also included the dummy jsonnet used to generate the same
- Edits a troubleshooting page to include more info and options

cc @bwplotka - I hope this solves your requirement!

Fixes #na!

**Checklist**
- NA Tests updated
- [x] Documentation added

### DIFF
--- a/operations/kube-manifests/ConfigMap-tempo-compactor.yaml
+++ b/operations/kube-manifests/ConfigMap-tempo-compactor.yaml
@@ -1,0 +1,58 @@
+apiVersion: v1
+data:
+  overrides.yaml: |
+    overrides: {}
+  tempo.yaml: |
+    compactor:
+        compaction:
+            block_retention: 144h
+            chunk_size_bytes: 1.048576e+07
+        ring:
+            kvstore:
+                store: memberlist
+    distributor:
+        receivers:
+            jaeger:
+                protocols:
+                    grpc:
+                        endpoint: 0.0.0.0:14250
+            otlp:
+                protocols:
+                    grpc:
+                        endpoint: 0.0.0.0:55680
+    ingester:
+        lifecycler:
+            ring:
+                replication_factor: 3
+    memberlist:
+        abort_if_cluster_join_fails: false
+        bind_port: 7946
+        join_members:
+          - gossip-ring.tracing.svc.cluster.local:7946
+    overrides:
+        per_tenant_override_config: /conf/overrides.yaml
+    server:
+        http_listen_port: 3100
+    storage:
+        trace:
+            backend: gcs
+            blocklist_poll: 10m
+            cache: memcached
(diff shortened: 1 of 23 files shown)


operations/kube-manifests/ConfigMap-tempo-compactor.yaml
```diff
@@ -0,0 +1,58 @@
+apiVersion: v1
```
bwplotka, line 1: Why this file is called Compactor configMap?

operations/kube-manifests/ConfigMap-tempo-querier.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: v1
```
bwplotka, line 1: Same here, this looks more than querier?

bwplotka (commented):
Nice! But I am bit confused with those files. Am I missing something?

operations/kube-manifests/ConfigMap-tempo-compactor.yaml
annanay25: I'm not sure I understand. This is indeed the configmap that the compactor uses..

operations/kube-manifests/ConfigMap-tempo-querier.yaml
annanay25: Oh right, at the moment we load almost the entire config into all components. This is actually the correct generated configmap.

annanay25: We should clean it up eventually. But notice that this doesn't have the compactor config block (its marked null)

operations/kube-manifests/util/generator.jsonnet
```diff
@@ -0,0 +1,67 @@
+// The jsonnet file used to generate these kube-manifests
```
joe-elliott, line 1: can we get a readme with the command to generate the yamls from this jsonnet?

annanay25: added!

operations/kube-manifests/ConfigMap-tempo-querier.yaml
annanay25: @bwplotka - are we good to proceed?

joe-elliott (approved):

operations/kube-manifests/ConfigMap-tempo-querier.yaml
bwplotka: Sorry for week lag, those are essentially different configs people can use? 🤔 How one can tell what we should use? (:

PR #668, opened 2021-04-23
Improve Wal Replay

### LINKED ISSUE #409: Ingesters with many WAL files will concatenate them on restart

joe-elliott:
**What this PR does**:
For simplicity the current WAL replay code loads a file, iterates through all records and replays them into an ingester instance.  This creates two issues:

- We are doubling our disk needs for a brief time period while the WAL replays and rebuilds itself in a second file.  In our internal ops cluster we have seen this lead to failed WAL replays.
- There is a lot of overhead to reading an entire multi GB file and rewriting it back to disk.

This proposed fix iterates through the WAL as is and simply rebuilds the index in memory for the AppendBlock to use directly instead of rebuilding a new file.

Fixes #409 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/wal/wal.go
```diff
@@ -73,27 +73,35 @@ func New(c *Config) (*WAL, error) {
 	}, nil
 }
 
-func (w *WAL) AllBlocks() ([]*ReplayBlock, error) {
+func (w *WAL) AllBlocks() ([]*AppendBlock, []error, error) { // jpe test failure modes. delete file, warnings
```
mdisibio, line 76: AllBlocks previously was a fairly light-weight operation, but is now quite intense (parsing entire contents of all wal files before returning). This design also prevents ability to log progress well from the ingester when replaying.   Maybe instead it could return metadata or a new kind of wal-specific info, and the ingester can call  `newAppendBlockFromFile` or an equivalent new thing, and log progress again.

tempodb/wal/append_block.go
```diff
@@ -52,6 +60,78 @@ func newAppendBlock(id uuid.UUID, tenantID string, filepath string, e backend.En
...
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+
+		reader := bytes.NewReader(buffer)
+		id, _, err := objectReader.UnmarshalObjectFromReader(reader)
+		if err != nil {
+			return nil, err
```
mdisibio, line 110: It looks like a proto unmarshal failure here will cause the entire WAL file to be skipped and deleted up the call chain in `AllBlocks1.  I.e. if the last record wasn't flushed properly then the whole block will be lost ?   Previously I think it would have preserved the records that it could.    Maybe breaking out of the loop and continuing would have that effect here.  Returning it as a warning up the call-chain too.

tempodb/wal/wal.go
joe-elliott: Passed in a logger for visibilty and renamed.

tempodb/wal/append_block.go
joe-elliott: Good catch. Partial replay support added + tests.

mdisibio (approved):

PR #671, opened 2021-04-26
Compression buffer

joe-elliott:
**What this PR does**:
PProf is showing that ~20% of ingester allocations during a CompleteBlock operation occur in `v1.dataReader => tempo_io.ReadAllWithEstimate`. This change adds a reusable buffer slice to the `v1.dataReader` to reduce allocations.

Also moved the +1 bytes into `ReadAllWithEstimate()`

### DIFF
--- a/tempodb/encoding/streaming_block_test.go
+++ b/tempodb/encoding/streaming_block_test.go
@@ -173,7 +173,7 @@ func testStreamingBlockToBackendBlock(t *testing.T, cfg *BlockConfig) {
 	}
 	sort.Slice(ids, func(i int, j int) bool { return bytes.Compare(ids[i], ids[j]) == -1 })
 
-	iterator, err := backendBlock.Iterator(10 * 1024 * 1024)
+	iterator, err := backendBlock.Iterator(50 * 1024)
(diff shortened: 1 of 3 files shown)


tempodb/encoding/streaming_block_test.go
```diff
@@ -173,7 +173,7 @@ func testStreamingBlockToBackendBlock(t *testing.T, cfg *BlockConfig) {
 	}
 	sort.Slice(ids, func(i int, j int) bool { return bytes.Compare(ids[i], ids[j]) == -1 })
 
-	iterator, err := backendBlock.Iterator(10 * 1024 * 1024)
+	iterator, err := backendBlock.Iterator(50 * 1024)
```
mdisibio, line 176: This is a fairly large change in buffer size.  Was the previous 10MB very wrong ?

joe-elliott: It just force pulled every record all at once a single `Read` call. This changes it so the test actually tests the reuse of the buffer correctly.

mdisibio (approved):

PR #674, opened 2021-04-26
Expand buffer to its capacity before appending more slices

mapno:
**What this PR does**:
Expand buffer to its capacity before appending more slices to make up for the difference to the length of compressed pages to read. Otherwise we're first filling already allocated elements before actually expanding the buffer.

None

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott (approved):
wups, lgtm!

PR #676, opened 2021-04-27
Enable mutex profiling. Fix load shedding.

joe-elliott:
**What this PR does**:
- Adds a cli parameter `mutex-profile-fraction` allowing us to dynamically enable mutex profiling
- Moves max traces check before grabbing the traces mutex and uses an atomic.Int32. This will relieve pressure on the traces mutex and allow more effective load shedding.

### DIFF
--- a/cmd/tempo/main.go
+++ b/cmd/tempo/main.go
@@ -40,10 +40,9 @@ func init() {
 }
 
 func main() {
-	runtime.SetMutexProfileFraction(5)
-
 	printVersion := flag.Bool("version", false, "Print this builds version information")
 	ballastMBs := flag.Int("mem-ballast-size-mbs", 0, "Size of memory ballast to allocate in MBs.")
+	mutexProfileFraction := flag.Int("mutex-profile-fraction", 0, "Enable mutex profiling.")
 
 	config, err := loadConfig()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/tempo/main.go
```diff
@@ -73,6 +74,10 @@ func main() {
 		}
 	}()
 
+	if *mutexProfileFraction > 0 {
```
annanay25, line 77:
```suggestion
	if *mutexProfileFraction > 0  && *mutexProfileFraction <= 1 {
```
protect against bad fractions?

modules/ingester/instance.go
```diff
@@ -337,15 +345,11 @@ func (i *instance) getOrCreateTrace(req *tempopb.PushRequest) (*trace, error) {
...
 
-	err = i.limiter.AssertMaxTracesPerUser(i.instanceID, len(i.traces))
-	if err != nil {
-		return nil, status.Errorf(codes.FailedPrecondition, "%s max live traces per tenant exceeded: %v", overrides.ErrorPrefixLiveTracesExceeded, err)
-	}
-
 	maxBytes := i.limiter.limits.MaxBytesPerTrace(i.instanceID)
 	trace = newTrace(maxBytes, fp, traceID)
 	i.traces[fp] = trace
 	i.tracesCreatedTotal.Inc()
```
annanay25, line 351: nit: can we metric this off the atomic int?

annanay25 (approved):
lgtm, left minor comments.

cmd/tempo/main.go
joe-elliott: `mutexProfileFraction` is an int and is the denominator of the fraction. i.e. `5` is acceptable input and means that 1/5 mutex interactions are profiled.

modules/ingester/instance.go
joe-elliott: i think this is fine. this is metric'ed by the distributor as refused.

cmd/tempo/main.go
annanay25: Mmm, that's confusing, would be more intuitive if it had `-denominator` in the flag name

PR #678, opened 2021-04-28
Tempodb support for block level data encoding

joe-elliott:
**What this PR does**:
Proposes a structure to track data encoding at the block level. This PR is incomplete and still needs additional tests, but communicates the basic idea.

- Add DataEncoding to the BlockMeta. This string is tracked by tempodb but is otherwise opaque. Tempodb just hands this back to calling code to make decisions with.
- Move trace utilities to `/pkg/model`. As we add other backend data formats we will start organizing code in this package to combine/sort/translate between them.
- Restrict the compactor to only compact blocks of the same encoding.
- Remove simple block strategy. It's not used and I didn't want to add support for dataEncodings.

**Alternative**:
Make no changes to tempodb and prefix every single object in the backend with a few bytes to indicate its encoding.
Pros:
- No changes to tempodb/simpler

Cons:
- Tempo proper would have to do byte level manipulation to extract the encoding. Not terrible but would be a rearrangement of complexity.
- There is no current encoding so we'd have to feel comfortable we wouldn't collide with the start of any existing proto object.
- Minor additional storage overhead. With our current 28 billion traces we would see an additional 26GiB in our backend. This seems non-horrible. This is only a .04% increase in storage.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


joe-elliott:
Dropped draft status and looking for reviews @grafana/tempo. After internal discussion we agreed on the path this PR is taking which is to track a `dataEncoding` value with each block and return it to the ingester/querier/compactor's as needed.

There are still two todos, but I intend to tackle them as part of the PR adding a new format.

pkg/model/combiner.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"github.com/go-kit/kit/log/level"
+
+	"github.com/cortexproject/cortex/pkg/util/log"
+)
+
+type objectCombiner struct{}
+
+var ObjectCombiner = objectCombiner{}
```
mapno, line 11: Any particular reason to have the struct as an exported variable?

pkg/model/encoding.go
```diff
@@ -0,0 +1,21 @@
+package model
+
+import (
+	"github.com/grafana/tempo/pkg/tempopb"
+
+	"github.com/gogo/protobuf/proto"
+)
+
+// CurrentEncoding is a string representing the encoding that all new blocks should be created with
+const CurrentEncoding = ""
```
mapno, line 10: Shouldn't this encoding be set via config? How is the encoding management expected to be done?

pkg/model/encoding.go
```diff
@@ -0,0 +1,21 @@
...
+
+// TracePBEncoding is a string that represents the original TracePBEncoding. Pass this if you know that the
+// bytes are encoded *tracepb.Trace
+const TracePBEncoding = ""
```
mapno, line 14: Should we use a different value? e.g. `const TracePBEncoding = "pb"`

mapno (commented):
Overall I like the changes, although i'm hesitant to approve as I feel I lack context of the bigger picture at the moment.

pkg/model/combiner.go
joe-elliott: There's no state so it seems unnecessary to reinstantiate it constantly.  Seemed it easier to do it once and use it whereever.

pkg/model/encoding.go
joe-elliott: I don't think we'll ever set this via config. Like the block encodings we will always just use the latest encoding available. i.e. `CurrentEncoding`.

pkg/model/encoding.go
joe-elliott: This has to be an empty string b/c it represents the encoding that is running in Tempo now which will loads as an empty string:

https://github.com/grafana/tempo/pull/678/files#diff-a427eb1e5a0cc55b603ffc38455586344d3b33c778a17d201f01ea6c6241fed9R259

mapno: Maybe we could return by default that encoding if there were less than 5 segments to the file name. It's a total nitpick anyway.

pkg/model/encoding.go
mapno: is not something people should care to configure right?  sounds good 👍

pkg/model/encoding.go
joe-elliott: It's a good point and was considered. The issue is that when we go to delete the file we call `fullFilename()` to get the file to delete and it would have to undo this logic or it won't find the file to delete correctly. We made this mistake when we first rolled out wal compression.

You are correct it would be nicer to have an actual string (instead of empty) to identify the original encoding. I'll think about it some.

pkg/model/encoding.go
annanay25, line 10: Should this be configurable?

tempodb/tempodb_test.go
```diff
@@ -202,7 +202,7 @@ func TestNilOnUnknownTenantID(t *testing.T) {
 	}, log.NewNopLogger())
 	assert.NoError(t, err)
 
-	buff, err := r.Find(context.Background(), "unknown", []byte{0x01}, BlockIDMin, BlockIDMax)
+	buff, _, err := r.Find(context.Background(), "unknown", []byte{0x01}, BlockIDMin, BlockIDMax)
```
annanay25, line 205: We need a real test around `Find` to ensure it returns the right encodings

pkg/model/trace.go
```diff
@@ -15,18 +15,18 @@ import (
 	v1 "github.com/grafana/tempo/pkg/tempopb/trace/v1"
 )
 
-func CombineTraces(objA []byte, objB []byte) (_ []byte, wasCombined bool, _ error) {
+// todo(jpe):
+// - add cross data encoding tests
+
+func CombineTraceBytes(objA []byte, objB []byte, dataEncodingA string, dataEncodingB string) (_ []byte, wasCombined bool, _ error) {
```
annanay25, line 21: Its unclear what encoding this function returns the combined bytes in.. can we add a comment before this?

modules/querier/querier.go
```diff
@@ -167,40 +166,54 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
-				return nil, err
+		if len(partialTraces) != 0 {
+			traceCountTotal = 0
+			spanCountTotal = 0
+			// combine partialTraces
+			var allBytes []byte
+			baseEncoding := dataEncodings[0] // just arbitrarily choose an encoding. generally they will all be the same
+			for i, partialTrace := range partialTraces {
+				dataEncoding := dataEncodings[i]
+				allBytes, _, err = model.CombineTraceBytes(allBytes, partialTrace, baseEncoding, dataEncoding)
```
annanay25, line 197: IIUC, the encoding of `allBytes` is unpredictable here. If the two traces were equal in the first call to combine, it will remain in `baseEncoding`. Else it might be proto marshalled bytes

tempodb/compactor.go
```diff
@@ -145,9 +144,11 @@ func (rw *readerWriter) compact(blockMetas []*backend.BlockMeta, tenantID string
 	}()
 
 	var totalRecords int
+	var dataEncoding string
 	for _, blockMeta := range blockMetas {
 		level.Info(rw.logger).Log("msg", "compacting block", "block", fmt.Sprintf("%+v", blockMeta))
 		totalRecords += blockMeta.TotalObjects
+		dataEncoding = blockMeta.DataEncoding // compacted blocks always have the same data encoding
```
annanay25, line 151: nit: `// blocks chosen for compaction always have the same data encoding`

annanay25 (commented):
Added comments, changes looking good overall!

pkg/model/encoding.go
joe-elliott: nope, see above.

modules/querier/querier.go
joe-elliott: Yes, it will be, but in the end we call:

`storeTrace, err := model.Unmarshal(allBytes, baseEncoding)` which will marshal whatever the encoding is into a *tracepb.Trace. So it will work.

pkg/model/trace.go
joe-elliott: good catch, we were always marshalling using the original encoding. This has been updated to support multiple encodings. data is always returned in dataEncodingA.

tempodb/tempodb_test.go
joe-elliott: good call. added.

annanay25 (approved):

PR #679, opened 2021-04-28
Preallocate byte slices from prom pool

annanay25:

**What this PR does**:
Uses a custom protobuf type `PreallocRequest` which reuses byte slices from a pool.
The pool implementation is taken from prometheus.
Byte slices are "donated" back to the pool in the `func (i *Ingester) PushBytes` function once the request is used.

Fixes #na!

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/tempopb/tempo.proto
```diff
@@ -37,5 +38,5 @@ message PushResponse {
 
 message PushBytesRequest {
   // pre-serialized PushRequests
-  repeated bytes requests = 1;
```
joe-elliott, line 40: don't replace requests, mark it deprecated and add a new proto field.  distributors always use the new field, ingesters use whichever one they happen to get.

we will have to rollout our ingesters first to support both fields and then follow it up with distributors.

modules/distributor/distributor.go
```diff
@@ -268,11 +268,11 @@ func (d *Distributor) sendToIngestersViaBytes(ctx context.Context, userID string
 		localCtx = user.InjectOrgID(localCtx, userID)
 
 		req := tempopb.PushBytesRequest{
-			Requests: make([][]byte, len(indexes)),
+			Requests: make([]tempopb.PreallocRequest, len(indexes)),
 		}
 
 		for i, j := range indexes {
-			req.Requests[i] = rawRequests[j][0:]
+			req.Requests[i].Request = rawRequests[j][0:]
```
joe-elliott, line 275: is `[0:]` necessary? I think it makes a new slice pointing at the same backing array. unsure if that's necessary.

pkg/tempopb/tempo.proto
joe-elliott: Nevermind, I understand now.

pkg/tempopb/prealloc.go
```diff
@@ -0,0 +1,45 @@
+package tempopb
+
+import (
+	"github.com/prometheus/prometheus/pkg/pool"
+)
+
+var (
+	// buckets: [1KiB, 2KiB, 4KiB, 8KiB, 16KiB]
+	bytePool = pool.New(1_000, 16_000, 2, func(size int) interface{} { return make([]byte, 0, size) })
```
joe-elliott, line 9: are these numbers backed up with metrics?

pkg/tempopb/prealloc.go
```diff
@@ -0,0 +1,45 @@
...
+	r.Request = bytePool.Get(len(dAtA)).([]byte)
+	r.Request = r.Request[:len(dAtA)]
+	copy(r.Request, dAtA)
+	return nil
+}
+
+// MarshalTo implements proto.Marshaller.
+func (r *PreallocRequest) MarshalTo(dAtA []byte) (int, error) {
+	copy(dAtA[:], r.Request[:])
+	return 0, nil // returned int is not used
```
joe-elliott, line 28: even if it's not used we should aim to fulfill the expectations of the interface

joe-elliott (commented):
Some comments. Good work!

pkg/tempopb/prealloc.go
annanay25: To some extent yes. We don't have the exact batch level metrics on ingesters, but noting down what I had in mind -

Internally, we are seeing ~180MiB/sec and ~1250 batches/sec. So roughly 144KiB per batch.
50th percentile of `tempo_distributor_traces_per_batch_bucket` is 20.
That's roughly 7.2KiB per trace per batch.

modules/distributor/distributor.go
annanay25: OK I didn't actually edit that array, but now that I think about it, I think `[0:]` is correct because it just copies over the slice, otherwise it would make a copy of the entire array

annanay25: Let me know if I'm missing something!

pkg/tempopb/prealloc.go
joe-elliott: Ha, this is hard.  Ok I'm getting:
- 156 MB/s to the distribs * 3 RF = 470 MB/s of proto to the ingesters
   `(sum(rate(tempo_distributor_bytes_received_total{cluster="ops-us-east-0"}[10m])) / 1024 / 1024) * 3`
- 140k ingester pushes/second
   `sum(rate(tempo_distributor_ingester_appends_total{cluster="ops-us-east-0"}[30m]))`
   `sum(rate(tempo_request_duration_seconds_count{cluster="ops-us-east-0", route="/tempopb.Pusher/PushBytes"}[30m]))`

470 MB/s / 140k pushes/s = 3.3 MB/push

- p50 of traces/batch = 20
   `histogram_quantile(.5, sum(rate(tempo_distributor_traces_per_batch_bucket{cluster="ops-us-east-0"}[30m])) by (le))`

3.3 MB / 20 = 168KB?  

Close enough for me!

joe-elliott (approved):
I think this is ready to go. Going to approve so you can merge as convenient. Two things to look at before merging

- Double check whether `[:0]` is necessary. Perhaps @mapno our resident slice expert can help :)
- Add some tests as discussed.

modules/distributor/distributor.go
annanay25: I think that linked program will always be true because it copies the entire slice over. 

Check this one - https://play.golang.org/p/OFRbw97uZa3

I'm printing the address of the backing array after copying in two different ways

cc @mapno :)

pkg/tempopb/prealloc.go
annanay25: Whoa, if your calculation is right then I'll have to change my buckets because I don't have a bucket that can store 168KB traces and we'd basically be creating a new one each time = no improvement.

I agree with the first half of your calculation, we should be calculating ingester level bytes received. However, in the second half, I think there are two changes needed:


1. > 470 MB/s / 140k pushes/s = 3.3 MB/push

this would be  470MB/s / 140000 pushes/s ~ 3.3KB/push


2.  > p50 of traces/batch = 20
histogram_quantile(.5, sum(rate(tempo_distributor_traces_per_batch_bucket{cluster="ops-us-east-0"}[30m])) by (le))

This metric is for incoming batches at the distributor right? Not at the ingester. 

```

                                   (smaller batches)  -- ingester-0
                                 /
agent -- (batches) --> distributor 
                                 \
                                   (smaller batches)  -- ingester-1

```

20 traces/batch at distributor = 20 * 3 (RF) / 40 (# of ingesters) = 1.5 traces/batch at ingester

3.3KB/1.5 ~2KB per trace per batch at the ingester. which sounds closer to our avg trace size (3KB)

I think I should adjust my buckets to [0.25KB, 0.5KiB, 1KiB, 2KiB, 4KiB, 8KiB]. WDYT?

modules/distributor/distributor.go
annanay25: OK now I'm plain confused - https://play.golang.org/p/V1WGO_jh88Z

mapno: There is a bit of trickery in that example since the initial variable is an array, rather than a slice. Although they look similar, they behave differently in this case. Arrays are values, so when copied, all the elements are copied over to the new variable. That is why we see two different pointers.

On the contrary, slices are pointers to where the elements are stored. Copying slices in go just copies the pointer, len and cap of the original slice. To actually duplicate the elements you need to use `copy()`. Unless I'm missing something, there is no difference between using `[0:]` or not (see https://play.golang.org/p/g2PAnWl-I2P).

As for the code, that line is copying the pointer of `rawRequests[j]` to `req.Requests[i].Request`. If `rawRequests[j]` is modified, it'll be shown in the _copy_. So it's important to know what happens to `rawRequests[j]` afterwards.

annanay25: Ah, thanks for explaining @mapno. That makes a lot of sense. I didn't realise that `[4]int{1, 2, 3, 4}` is the array representation.

In conclusion, in this case we should be ok using either, but since we were using `[0:]` earlier, I suggest we continue to use the same

PR #684, opened 2021-05-03
Update required s3 permissions

joe-elliott:
**What this PR does**:
Adds "s3:PutObjectTagging" and "s3:GetObjectTagging" to the list of required permissions for s3. If an object has tags s3 copy will fail (which we use to move meta.json -> meta.compacted.json) in the compactors.

https://medium.com/collaborne-engineering/s3-copyobject-access-denied-5f7a6fe0393e


kvrhdn (approved):
Looks good to me but I'm not yet a code owner of course 🙂

annanay25 (approved):

PR #687, opened 2021-05-04
Add span deduper to support zipkin client/server kind spans

annanay25:

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/deduper.go
+++ b/modules/frontend/deduper.go
@@ -1,0 +1,122 @@
+package frontend
+
+import (
+	"context"
+	"encoding/binary"
+	"fmt"
+
+	"github.com/opentracing/opentracing-go"
+
+	"github.com/grafana/tempo/pkg/tempopb"
+	v1 "github.com/grafana/tempo/pkg/tempopb/trace/v1"
+)
+
+const (
+	warningTooManySpans = "cannot assign unique span ID, too many spans in the trace"
+)
+
+var (
+	maxSpanID uint64 = 0xffffffffffffffff
+)
+
+// This is copied over from Jaeger and modified to work for OpenTelemetry Trace data structure
+// https://github.com/jaegertracing/jaeger/blob/12bba8c9b91cf4a29d314934bc08f4a80e43c042/model/adjuster/span_id_deduper.go
+type spanIDDeduper struct {
+	trace     *tempopb.Trace
+	spansByID map[uint64][]*v1.Span
+	maxUsedID uint64
+}
+
+
+func DedupeSpanIDs(ctx context.Context, trace *tempopb.Trace) (*tempopb.Trace, error) {
+	span, _ := opentracing.StartSpanFromContext(ctx, "frontend.DedupeSpanIDs")
+	defer span.Finish()
+
+	deduper := &spanIDDeduper{trace: trace}
+	deduper.groupSpansByID()
+	deduper.dedupeSpanIDs()
+	return deduper.trace, nil
+}
+
+// groupSpansByID groups spans with the same ID returning a map id -> []Span
+func (d *spanIDDeduper) groupSpansByID() {
+	spansByID := make(map[uint64][]*v1.Span)
+	for _, batch := range d.trace.Batches {
+		for _, ils := range batch.InstrumentationLibrarySpans {
+			for _, span := range ils.Spans {
+				id := binary.BigEndian.Uint64(span.SpanId)
+				if spans, ok := spansByID[id]; ok {
+					// TODO maybe return an error if more than 2 spans found
+					spansByID[id] = append(spans, span)
+				} else {
+					spansByID[id] = []*v1.Span{span}
+				}
+			}
+		}
+	}
+	d.spansByID = spansByID
+}
+
+func (d *spanIDDeduper) isSharedWithClientSpan(spanID uint64) bool {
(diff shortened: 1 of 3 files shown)


modules/frontend/deduper_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/grafana/tempo/pkg/tempopb"
+	v1 "github.com/grafana/tempo/pkg/tempopb/trace/v1"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDedupeSpanIDs(t *testing.T) {
+	firstSpanId := []byte{0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01}
+	secondSpanId := []byte{0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02}
+	thirdSpanId := []byte{0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x03}
+	tests := []struct {
```
joe-elliott, line 16: can we expand the test to check for the actual expected output?

joe-elliott (commented):
Two comments but otherwise looks good so far.  Besides the comments two other things that would be nice to see:
- a comment to indicate that this is being done to support zipkin
- benchmarking of various trace sizes: 100, 1k, 10k, 100k spans

I'm a little nervous about the impact on correctness but with @kvrhdn 's improvements to the vulture I think we will feel better.

modules/frontend/querysharding.go
```diff
@@ -194,6 +194,16 @@ func mergeResponses(ctx context.Context, marshallingFormat string, rrs []Request
 		}, nil
 	}
 
+	traceObject := &tempopb.Trace{}
+	err := proto.Unmarshal(combinedTrace, traceObject)
+	if err != nil {
+		return nil, err
+	}
+	traceObject, err = DedupeSpanIDs(ctx, traceObject)
```
joe-elliott, line 202: should this be injected as a new middleware pipeline thing instead of snuck into the end of querysharding?

annanay25: Yep was a little lazy hehe but you're right, going to implement as a middleware.

annanay25:
Just need to add benchmarks now.

modules/frontend/deduper.go
```diff
@@ -0,0 +1,176 @@
...
+
+	if resp.StatusCode == http.StatusOK {
+		body, err := ioutil.ReadAll(resp.Body)
+		defer resp.Body.Close()
+		if err != nil {
+			return nil, err
+		}
+
+		traceObject := &tempopb.Trace{}
+		err = proto.Unmarshal(body, traceObject)
```
joe-elliott, line 64: dang, sad about this. this will cost us on the big traces, but i guess there's no other way to do this using the http pipeline stuff? I wonder if Cortex/Loki repeatedly marshal/unmarshal?

is there a way to sanely combine this and the sharding pipeline processor so we only have to unmarshal/marshal once?

we want to keep the logic isolated, but it's ok if we technically use the same pipeline processor if it saves us on latency.

modules/frontend/deduper_test.go
```diff
@@ -0,0 +1,117 @@
...
+func BenchmarkDeduper10000(b *testing.B) {
+	benchmarkDeduper(b, 10000)
+}
+
+func BenchmarkDeduper100000(b *testing.B) {
+	benchmarkDeduper(b, 100000)
+}
+
+func benchmarkDeduper(b *testing.B, traceSpanCount int) {
+	s := &spanIDDeduper{
```
joe-elliott, line 109: since the deduper rewrites the trace in place i don't think this benchmark is accurate since it will be deduped after the first loop.

modules/frontend/deduper_test.go
```diff
@@ -0,0 +1,117 @@
...
+		name        string
+		trace       *tempopb.Trace
+		checkParent bool
+	}{
+		{
+			name: "no duplicates",
+			trace: &tempopb.Trace{
+				Batches: []*v1.ResourceSpans{
+					{
+						InstrumentationLibrarySpans: []*v1.InstrumentationLibrarySpans{
```
joe-elliott, line 24: would be better to have the input spans and expected output in each test instead of two tests that happen to result in the same output.

joe-elliott (commented):
Some comments. I think you need to make a change to your benchmark but I don't expect it to appreciably change the results.

Agree if the results stay roughly the same that we can just always run the deduper.

modules/frontend/deduper.go
annanay25: A round of unmarshalling => marshalling is needed anyway in either case at the frontend regardless of where the processor sits. If we were to combine it with the querysharder, the same logic would go here - https://github.com/grafana/tempo/pull/687/commits/4bd976a4563f618a40d81a80d2ffe635ade21f4a#diff-b79ae30491380fa96485ab56b95197e2ea7f39b0f1460b7ff9d4fdeadb370ed1R198

The current approach of using the deduper as a middleware should not have a significant performance impact. 

A potentially important thing to discuss would be whether we want to make the whole module configurable so it can be turned on/off per-tenant.

joe-elliott (approved):

Issue #682, opened 2021-05-03
Support Zipkin Client/Server Spans


joe-elliott:
**Is your feature request related to a problem? Please describe.**
According to the Zipkin spec the client and server spans should share the same id:
https://zipkin.io/zipkin/2.11.3/zipkin/zipkin2/Span.html#shared--

> When an RPC trace is client-originated, it will be sampled and the same span ID is used for the server side. However, the server shouldn't set span.timestamp or duration since it didn't start the span.

This breaks the Grafana UI b/c it does not expect two spans to have the same ID.

**Describe the solution you'd like**
Dedupe span IDs on query.  Here is sample code from Jaeger:

https://github.com/jaegertracing/jaeger/blob/master/model/adjuster/span_id_deduper.go#L24

PR #688, opened 2021-05-04
Support for Zipkin: CombineTraces

joe-elliott:
**What this PR does**:
Uses span.Kind to differentiate between zipkin client/server spans in combine.

Related to #682 and #687 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/util/trace.go
+++ b/pkg/util/trace.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"bytes"
+	"encoding/binary"
 	"hash"
 	"hash/fnv"
 	"sort"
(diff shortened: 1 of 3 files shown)


pkg/util/trace.go
```diff
@@ -68,12 +69,13 @@ func CombineTraceProtos(traceA, traceB *tempopb.Trace) (*tempopb.Trace, int, int
 	spanCountTotal := 0
 
 	h := fnv.New32()
+	buffer := make([]byte, 4)
```
annanay25, line 72: Just mentioning for reference (was wondering why we need 4bytes), the span kind field is a int32 
https://github.com/open-telemetry/opentelemetry-collector/blob/51661b8400d8a90e1bded85e2770fd56865761be/internal/data/protogen/trace/v1/trace.pb.go#L34

annanay25 (approved):
lgtm, nice!

pkg/util/trace.go
joe-elliott: 32 / 8 = 4?

annanay25: lol i meant i didn't think we need more than 1 byte

PR #689, opened 2021-05-06
Vulture: add metric label for requests that failed

kvrhdn:
**What this PR does**:
Query requests from vulture can fail for multiple reasons, not only because Tempo returned a 404.  This PR introduces a new label `requestfailed` which will count the amount of errors that are not 404.

Additionally, I believe there was a mistake in how `error_total` is counted (`metricErrorTotal` in Go): if `queryTempoAndAnalyze` returns an error, this increments `error_total` but we should also still increase `trace_total`.
We currently do not do this and because of this `trace_total` will drop a bit when requests fail. This is misleading since we did attempt to query this trace.

Screenshot 2021-05-06 at 13 34 08

Other dashboards calculate a percentage of "not found traces" using `trace_error_total / trace_total`:
```promql
sum(rate(tempo_vulture_trace_error_total{cluster="$cluster", namespace="$namespace", error="notfound"}[1h])) / sum(rate(tempo_vulture_trace_total{cluster="$cluster", namespace="$namespace"}[1h]))
```

So this strengthens my believe that `trace_total` should also included failed requests.

~~Fixes #<issue number>~~

**Checklist**
- [ ] ~~Tests updated~~
- [ ] ~~Documentation added~~

### DIFF
--- a/cmd/tempo-vulture/main.go
+++ b/cmd/tempo-vulture/main.go
@@ -39,9 +39,10 @@ var (
 )
 
 type traceMetrics struct {
-	requested    int
-	notfound     int
-	missingSpans int
+	requested     int
+	requestFailed int
+	notFound      int
+	missingSpans  int
 }
 
 func init() {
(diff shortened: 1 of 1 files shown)


dgzlopes (approved):
Thanks for changing this! LGTM :)

mapno (approved):
lgtm, nice!

PR #693, opened 2021-05-12
Minor improvements to GCS driver and Query Path

annanay25:
**What this PR does**:
Bunch of small improvements:
- replace the use of `ioutil.ReadAll()` with `ReadAllWithEstimate` with a predefined buffer size, using the returned object attributes in tempodb/backend/gcs and tempodb/backend/s3
- add `ContentLength` header in span id deduper middleware
- update gcs driver version from v1.12.0 => v1.15.0. Primarily to catch the following change
```
// configureHTTP2 configures the ReadIdleTimeout HTTP/2 option for the
// transport. This allows broken idle connections to be pruned more quickly,
// preventing the client from attempting to re-use connections that will no
// longer work.
```
- fix a panic in query frontend by checking for nil response

Fixes #na

**Checklist**
- [x] Tests updated
- NA Documentation added


tempodb/backend/gcs/gcs.go
```diff
@@ -279,7 +278,15 @@ func (rw *readerWriter) readAll(ctx context.Context, name string) ([]byte, error
 	}
 	defer r.Close()
 
-	return ioutil.ReadAll(r)
+	// Preallocate the buffer with the exact size so we don't waste allocations
+	// while progressively growing an initial small buffer. The buffer capacity
+	// is increased by MinRead to avoid extra allocations due to how ReadFrom()
+	// internally works.
+	buf := bytes.NewBuffer(make([]byte, 0, r.Attrs.Size+bytes.MinRead))
+	if _, err := buf.ReadFrom(r); err != nil {
```
mdisibio, line 286: Thoughts on using the new method pkg/io/ReadAllWithBuffer?  Would that avoid needing to take bytes.MinRead into account?

modules/frontend/frontend.go
```diff
@@ -73,7 +73,7 @@ func NewTripperware(cfg Config, logger log.Logger, registerer prometheus.Registe
 
 			resp, err := rt.RoundTrip(r)
 
-			if resp.StatusCode == http.StatusOK && marshallingFormat == util.JSONTypeHeaderValue {
+			if resp != nil && resp.StatusCode == http.StatusOK && marshallingFormat == util.JSONTypeHeaderValue {
```
joe-elliott, line 76: in what situations is resp nil? have we ever seen this panic?

tempodb/backend/gcs/gcs.go
joe-elliott: yeah, i'd prefer `ReadAllWithBuffer()`. It will also cover cases where the size is incorrectly calculated for some reason. 

would prefer the same for all below.

annanay25: Cool will change. `ReadAllWithBuffer()` starts off with a 512 byte slice which is exactly what `bytes.MinRead` is

modules/frontend/frontend.go
annanay25: Yep, just panic'ed today. `resp` is nil for a 5xx error.

tempodb/backend/gcs/gcs.go
joe-elliott: If you pass in estimatedBytes that you can start with a buffer that will require far less allocations. You should be able to do that for most if not all of these changes.

joe-elliott (approved):

PR #694, opened 2021-05-12
Reduce Ingester Allocations

### LINKED ISSUE #658: Performance Improvement: Don't marshal to proto before the WAL
Currently our biggest performance hit is repeatedly marshalling bytes <=> proto. There is technically no reason to unmarshal from bytes to proto before writing a trace to the WAL. We really would only need to do this if the trace were queried, but we do it to every single trace. 

Let's remove this marshal/unmarshal for performance. I believe the complexity of this change will come from here:
https://github.com/grafana/tempo/blob/main/modules/ingester/trace.go#L14.

This is the object that we are currently writing and reading from the WAL. To do this we would need to write a different object that contained all of the unmarshalled byte slices.

joe-elliott:
**What this PR does**:
Adds two new fields to tempopb.PushBytesRequest "traces" and "ids" and deprecated the "requests" field. traces are pre-marshalled byte slices containing trace which the ingester can now take directly to the wal without additional marshalling. Also added a `TraceBytes` object that the ingesters now write to the wal and eventually to the blocks.

Other changes:
- Many test improvements/additions
- Improve records check in append_block Find to prevent unnecessary lookups
- PushBytes supports both the old and new methods of pushing to the ingesters to recud
- Propagate context through instance.FindTraceByID()
- Added/extended functionality in the `pkg/model` package to support the new data encoding.
- Adjusted PreallocRequest to handle the new byte slices

**Breaking Change!**
Distributors will immediately attempt to start pushing to the ingesters with the new proto which will cause the new ingesters to silently ignore the new fields. To prevent any dropped spans roll out ingesters first and then distributors once all ingesters have upgraded.

Fixes #658 

**Performance Before Changes**
cpu usage seconds:
image

go gc duration seconds:
image

**Performance After Changes**
cpu usage seconds:
image

go gc duration seconds:
image

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/model/sort.go
```diff
@@ -0,0 +1,75 @@
...
+	// Sort by start time, then id
+	if a.StartTimeUnixNano == b.StartTimeUnixNano {
+		return bytes.Compare(a.SpanId, b.SpanId) == -1
+	}
+
+	return a.StartTimeUnixNano < b.StartTimeUnixNano
+}
+
+// SortTraceBytes sorts a *tempopb.TraceBytes
+func SortTraceBytes(t *tempopb.TraceBytes) {
```
mdisibio, line 54: Have a few thoughts/questions.  The purpose of this is to deterministically sort data written by the ingesters so that the compactor can correctly determine identical content and skip recombination (replaces usage of SortTrace).   In this case maybe `bytes.Compare == -1` would work just as well and be simpler.    Either way, sort.Slice takes a `less` func, so I think the comparisons for `i > j` are backwards?

joe-elliott: Good catch. :+1:

mdisibio:
Ran locally and stability and performance improvements all check out.

mdisibio (approved):

PR #696, opened 2021-05-12
Remove ReuseTraceBytes

joe-elliott:
**What this PR does**:
Removes usage of the bytepool.  We are using this incorrectly and, as written, will only cause panics.

This is complicated but at least two things are occurring which has me remove this for now:
- The prometheus slice pool is not guaranteed to return a slice of the size requested. The way we have written this code it is easy to panic with too small slice.
- gogoproto is not compatible with GRPC latest. I don't believe the custom classes are being used like we think they are.

All around this just requires more research. Removing it for now.


mdisibio:
> * The prometheus slice pool is not guaranteed to return a slice of the size requested. The way we have written this code it is easy to panic with too small slice.
https://github.com/prometheus/prometheus/blob/main/pkg/pool/pool.go#L66  It looks like the slice returned is the size of the matching bucket.  The result could be resized to the needed length.

annanay25:
Yeah, a subtle difference is that it will create a slice of _atleast_ minSize

joe-elliott:
Yup, if it were that simple I'd go for that change, but I don't think we even really understand when this code is currently being run. I think we need to step back and really understand what it's doing before bringing it back in.

annanay25 (approved):

joe-elliott:
Closing in favor of #698 for now.

PR #700, opened 2021-05-13
Added trace id logging

joe-elliott:
**What this PR does**:
Added the ability to turn on trace id logging in the distributors to help with debugging.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/tempo/website/configuration/manifest.md
+++ b/docs/tempo/website/configuration/manifest.md
@@ -95,6 +95,7 @@ distributor:
     receivers: {}
     override_ring_key: distributor
     extend_writes: true
+    log_received_traces: false
 ingester_client:
     pool_config:
         checkinterval: 15s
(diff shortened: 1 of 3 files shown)


mdisibio (approved):
Nice, will be very helpful.   (Non-blocking) I wonder if there is a slightly better spot to log.  Debating earlier up on every received input regardless of distributor rate checks, or after successfully pushing to the ingesters.

annanay25 (approved):
> Nice, will be very helpful. (Non-blocking) I wonder if there is a slightly better spot to log. Debating earlier up on every received input regardless of distributor rate checks, or after successfully pushing to the ingesters.

I vote for after a successful push. It would otherwise be hard to differentiate between what I can query tempo for and what was dropped.

joe-elliott:
Moved before everything. This is entirely meant to be a debugging tool and I think it's wisest to just dump everything that ever comes in.

Issue #428, opened 2020-12-18
Document all APIs


joe-elliott:
**Is your feature request related to a problem? Please describe.**
There is no concise place where all apis are documented for Tempo.  We should add this.  For the various ingestion protocols we should just link to external docs.  There's no reason to replicate it.

annanay25:
Cortex API docs for reference: https://github.com/cortexproject/cortex/blob/master/docs/api/_index.md

cc @achatterjee-grafana

swartz-k:
Does this api docs are mostly about Querier and Query-Frontend?
Show API in [here](https://github.com/grafana/tempo/blob/master/cmd/tempo/app/modules.go) Such as in Querier `/querier/api/traces/{traceID}` to FindTraceByID

joe-elliott:
This does not need to cover the APIs of the receiver shims. However, it should link to appropriate documentation where possible.

PR #714, opened 2021-05-27
Add HTTP endpoint /api/echo to query-frontend

### LINKED ISSUE #624: Create a way for a Grafana datasource to be able to confirm it is correctly configured for Tempo
We need a way for Grafana to query Tempo when setting up a datasource so that it knows for sure everything is configured correctly.  Currently it queries `/api/traces/` and if it receives a 404 then it assumes things are correct.  This, however, does not catch situations where you point the datasource at a non-Tempo http service or just the wrong path in Tempo

Add a way for Grafana to query Tempo and confirm that the datasource is setup correctly.

Possible ideas:
- add an echo endpoint that always 200s: `/api/echo`
- add a special trace id that always 200s `/api/traces/0` or `/api/traces/test` or something.
- key off the 404 body that confirms it did indeed hit the `/api/traces` endpoint.

kvrhdn:
**What this PR does**:
Adds a new endpoint `/api/echo` that always returns with 200  and a body `echo`.

Closes #624

**Checklist**
- [x] Tests updated
- [ ] Documentation added --> should be captured in #428

### DIFF
--- a/cmd/tempo/app/modules.go
+++ b/cmd/tempo/app/modules.go
@@ -3,6 +3,7 @@ package app
 import (
 	"fmt"
 	"net/http"
+	"path"
 
 	"github.com/cortexproject/cortex/pkg/cortex"
 	cortex_frontend "github.com/cortexproject/cortex/pkg/frontend"
@@ -43,6 +44,11 @@ const (
 	Store         string = "store"
 	MemberlistKV  string = "memberlist-kv"
 	All           string = "all"
+)
+
+const (
+	apiPathTraces string = "/api/traces/{traceID}"
+	apiPathEcho   string = "/api/echo"
 )
 
 func (t *App) initServer() (services.Service, error) {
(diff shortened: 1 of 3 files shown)


mapno (approved):
LGTM!

cmd/tempo/app/modules.go
```diff
@@ -185,6 +185,8 @@ func (t *App) initQueryFrontend() (services.Service, error) {
 	// http query endpoint
 	t.server.HTTP.Handle(queryEndpoint(&t.cfg), tracesHandler)
 
+	t.server.HTTP.Handle("/api/echo", echoHandler())
```
joe-elliott, line 188: this should use `queryEndpoint` like the endpoint above it. it will need to support the prefix so we can read it through our gateway

kvrhdn: Oh right, good catch!

cmd/tempo/app/modules.go
```diff
@@ -274,3 +276,13 @@ func (t *App) setupModuleManager() error {
 func queryEndpoint(cfg *Config) string {
 	return cfg.HTTPAPIPrefix + "/api/traces/{traceID}"
 }
+
```
joe-elliott, line 279: Let me nit you just a bit more :)!

let's make a single method with some really basic tests:

```
func  applyHTTPPrefix(cfg, path) {
}
```

Also let's use `path.Join` instead of string concat.

Thanks for the work, we're almost there!

kvrhdn: Done! I didn't add a unit test for `addHTTPAPIPrefix` because the function is really small and not very error-prone (feel free to disagree on this!).
If we change this logic at some point, the e2e tests should break.

I did notice whoever that the e2e tests don't configure a prefix... What do you think, should I adapt one of the configs to use a prefix?

PR #720, opened 2021-05-31
Clean up examples

### LINKED ISSUE #707: Remove tempo-query from examples
The following examples continue to use tempo-query. Remove this and query directly from Grafana.

./example/jsonnet
./example/helm

Also remove mention of tempo-query from the main readme and look for any other places it can be removed. It should only be mentioned in the "Querying with Grafana" Docs.

kvrhdn:
**What this PR does**:
Update / clean up the examples: remove stray mentions of tempo-query and verify they work.

Rendered preview: https://github.com/kvrhdn/tempo/tree/707-update-docs/example

Closes #707

**Checklist**
- [ ] ~~Tests updated~~
- [x] Documentation added

### DIFF
--- a/example/docker-compose/readme.md
+++ b/example/docker-compose/readme.md
@@ -30,25 +30,26 @@ Tempo can be run with local storage, S3, Azure, or GCS backends. See below for e
 In this example all data is stored locally in the example-data/tempo folder.
 
 1. First start up the local stack. 
-```
+```console
 docker-compose up -d
 ```
 
 At this point, the following containers should be spun up -
 
 ```console
-$ docker-compose -f docker-compose.local.yaml ps
+docker-compose ps
+```
+```
                   Name                                 Command               State                         Ports
 --------------------------------------------------------------------------------------------------------------------------------------
 docker-compose_grafana_1                    /run.sh                          Up      0.0.0.0:3000->3000/tcp
 docker-compose_prometheus_1                 /bin/prometheus --config.f ...   Up      0.0.0.0:9090->9090/tcp
 docker-compose_synthetic-load-generator_1   ./start.sh                       Up
-docker-compose_tempo-query_1                /go/bin/query-linux --grpc ...   Up      0.0.0.0:16686->16686/tcp
 docker-compose_tempo_1                      /tempo -storage.trace.back ...   Up      0.0.0.0:32772->14268/tcp, 0.0.0.0:32773->3100/tcp
 ```
 
 2. If you're interested you can see the wal/blocks as they are being created.
-```
+```console
 ls ./example-data/tempo
 ```
 
@@ -56,18 +57,19 @@ ls ./example-data/tempo
 
 ```console
 docker-compose logs -f synthetic-load-generator
-.
-.
+```
+```
 synthetic-load-generator_1  | 20/10/24 08:27:09 INFO ScheduledTraceGenerator: Emitted traceId 57aedb829f352625 for service frontend route /product
 synthetic-load-generator_1  | 20/10/24 08:27:09 INFO ScheduledTraceGenerator: Emitted traceId 25fa96b9da24b23f for service frontend route /cart
 synthetic-load-generator_1  | 20/10/24 08:27:09 INFO ScheduledTraceGenerator: Emitted traceId 15b3ad814b77b779 for service frontend route /shipping
 synthetic-load-generator_1  | 20/10/24 08:27:09 INFO ScheduledTraceGenerator: Emitted traceId 3803db7d7d848a1a for service frontend route /checkout
-.
 ```
 
 Logs are in the form
 
-`Emitted traceId <traceid> for service frontend route /cart`
+```
+Emitted traceId <traceid> for service frontend route /cart
+```
 
 Copy one of these trace ids.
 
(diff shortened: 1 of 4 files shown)


example/tk/readme.md
```diff
@@ -8,16 +8,21 @@ better at demonstrating trace discovery flows using Loki and other tools.
...
 The Jsonnet is meant to be applied to with [tanka](https://github.com/grafana/tanka).  To test the jsonnet locally requires:
 
 - k3d > v3.2.0
 - tanka > v0.12.0
 
-Create a cluster
+Create a cluster with 3 nodes
+
+```console
+k3d cluster create tempo --servers 3 --api-port 6443 --port "16686:80@loadbalancer"
```
kvrhdn, line 19: I had to add `--servers 3` otherwise it was not possible to query Tempo in the microservices config.

Reason:
- due to anti-affinity rules, it's not possible to run 3 ingesters on a single node
- if there are not enough ingesters online queries fail in the query-frontend with
  ```
  error finding ingesters in Querier.FindTraceByID: too many unhealthy instances in the ring
  ```

This is not an issue in the helm example... so we maybe we can avoid this by changing some jsonnet config instead 🤔

kvrhdn:
I still want to figure out why viewing traces in the tk example does not work (I always get a 404 in Jaeger UI), but we can already merge this PR.

example/tk/readme.md
dgzlopes: I think that the >=3 servers requirement is reasonable! But if we can do some Jsonnet magic to support these <3 node deployments, it could be interesting, agree.

dgzlopes (commented):
Thanks for the cleanup :rocket: 

Small question: Removing tempo-query from `operations/jsonnet/*` and the related configs on `example/tk/*` is out-of-scope for the PR?

example/docker-compose/readme.md
```diff
@@ -30,44 +30,46 @@ Tempo can be run with local storage, S3, Azure, or GCS backends. See below for e
...
 1. First start up the local stack. 
-```
+```console
 docker-compose up -d
 ```
 
 At this point, the following containers should be spun up -
 
 ```console
-$ docker-compose -f docker-compose.local.yaml ps
```
annanay25, line 40: I think we need this because `docker-compose ps` would look up the services in the file `docker-compose.yaml` by default.

annanay25: honestly it might be better to just remove the `ps` command to make this file leaner

example/helm/readme.md
```diff
@@ -20,11 +19,16 @@ Create a cluster
...
+If you wish to use a local image, you can import these into k3d
+
+```console
+k3d image import grafana/tempo:latest --cluster tempo
+```
+
 Next either deploy the microservices or the single binary.
 
 ### Microservices
-The microservices deploy of Tempo is fault tolerant, high volume, independently scalable.  This jsonnet is in use by
```
annanay25, line 26: I like this sentence, any particular reason behind removing it? :)

example/tk/readme.md
annanay25: Can we run with replication factor 1? Does it still fail?

example/tk/readme.md
```diff
@@ -8,16 +8,21 @@ better at demonstrating trace discovery flows using Loki and other tools.
...
+```
+
+If you wish to use a local image, you can import these into k3d
```
annanay25, line 22: Nice, this is very useful!

annanay25 (commented):
Thanks for the clean up @kvrhdn; I've left some small comments.

kvrhdn:
> Small question: Removing tempo-query from `operations/jsonnet/*` and the related configs on `example/tk/*` is out-of-scope for the PR?

Good question and I'm not sure. Folks might still have setups out there with Grafana 7.4 so I don't want to break their setup completely.

We can change the defaults to not deploy tempo-query unless explicitly enabled.

example/docker-compose/readme.md
kvrhdn: This example doesn't need the `-f` flag because it is using the default `docker-compose.yaml`. But maybe this is confusing?

example/helm/readme.md
kvrhdn: This is about the helm config, we don't use this internally. I think it's a copy-paste error.

example/tk/readme.md
kvrhdn: Oh! I can try to override the replication config, I doesn't work if I just override `replicas`.

example/docker-compose/readme.md
dgzlopes: We had `docker-compose.local.yaml`: https://github.com/grafana/tempo/blob/v0.2.0/example/docker-compose/docker-compose.local.yaml

But we no longer have that file... so I think it's alright to remove the flag.

dgzlopes:
> We can change the defaults to not deploy tempo-query unless explicitly enabled.

Sounds good! If we're removing all the references to tempo-query, probably we don't want to deploy it by default :smile:

example/docker-compose/readme.md
```diff
@@ -30,44 +30,46 @@ Tempo can be run with local storage, S3, Azure, or GCS backends. See below for e
...
+docker-compose ps
+```
+```
                   Name                                 Command               State                         Ports
 --------------------------------------------------------------------------------------------------------------------------------------
 docker-compose_grafana_1                    /run.sh                          Up      0.0.0.0:3000->3000/tcp
```
joe-elliott, line 45: i've always found these sections that list every pod to be too much detail and cluttering the doc. what do you think?

should we remove?

example/tk/readme.md
joe-elliott: the "jsonnet magic" path to just remove the anti affinity rules is the most attractive to me.  

however, setting "--servers=3" isn't that big of a deal so if this is a lot of work please feel free to just leave this. since this is meant to be a semi production setup i'd like to keep RF=3 and have 3 ingesters.

kvrhdn: I have been able to remove the affinity settings using this jsonnet: https://github.com/kvrhdn/tempo/blob/707-update-docs/example/tk/tempo-microservices/main.jsonnet#L71-L80

example/docker-compose/readme.md
kvrhdn: I personally don't mind it because it's a bit easier to follow along without executing all the commands. I more worried about the duplication on this page, every example follows the same format but has a slightly different command to check the storage.

kvrhdn:
I can swap Jaeger UI for Grafana (for the tk example) in another PR.

example/docker-compose/readme.md
joe-elliott: yeah, perhaps if we broke this into one .md for each example it would feel better?

joe-elliott (approved):

PR #740, opened 2021-06-04
Reorganize docker-compose examples into subfolders

### LINKED ISSUE #731: Docker-compose examples cleanup
The docker-compose examples were recently split out to have individual readmes which helps with readability, but all of the files are just jammed in the same folder: ./example/docker-compose.

Reorganize the examples into individual subfolders for clarity. For instance move:
```
example/docker-compose/docker-compose.agent.yaml
example/docker-compose/etc/agent.yaml
example/docker-compose/readme.agent.md
```
to
```
example/docker-compose/agent/docker-compose.yaml
example/docker-compose/agent/agent.yaml
example/docker-compose/agent/readme.md
```

Note that there are shared config files and relative paths that will need to be cleaned up.

kvrhdn:
**What this PR does**:
Split all docker-compose examples into separate subfolders. I've also updated examples and made them more consistent with each other.

Browse through the new structure here: https://github.com/kvrhdn/tempo/tree/731-split-examples/example/docker-compose

Fixes #731

**TODO**

- [x] Update links in `docs/`


mdisibio:
This is looking really good so far, will keep reviewing.  Agree with mounting the `grafana-datasources.yaml` file instead of the folder, seems simpler to me.

I think some (all?) of the examples are missing the dashboards now.  Any example that runs tempo/prometheus/grafana should have the tempo dashboards mounted too.

One more thing:  this isn't related to the changes in this PR, but I would like to reconsider the order of examples described in the readme.  The first example in the list is the Loki example, but I'm not sure that is the best for new users.  The local-storage example is probably the simplest and ideal to list first.

mapno (commented):
Overall I really like the changes, great work!

It's a shame we can't use the dashboards because it won't work with single binary deployments. I wonder if we could craft a compatible version with jsonnet.

joe-elliott:
> It's a shame we can't use the dashboards because it won't work with single binary deployments. I wonder if we could craft a compatible version with jsonnet.  

Cortex does this:
https://github.com/grafana/cortex-jsonnet/blob/main/cortex-mixin/config.libsonnet#L24

I don't think we should worry about fixing that as part of this PR though.

kvrhdn:
I'll create an issue to keep track of the dashboards :+1:

joe-elliott (approved):

joe-elliott:
>  The first example in the list is the Loki example, but I'm not sure that is the best for new users. The local-storage example is probably the simplest and ideal to list first.

I did this on purpose b/c it should drive people to the "most complete" example so they can see how to use Loki to create links into Tempo. All of the other examples require copy/pasting trace ids and i didn't want people to think that was required to use Tempo.

PR #748, opened 2021-06-09
Add the specific datasource with loki and not the shared

kaflake:
Add the specific datasource file to loki example instead the shared


kvrhdn (approved):
Good catch! This went wrong in #740.

mapno (approved):
LGTM, thanks!

kvrhdn:
The same change has been applied here: https://github.com/grafana/tempo/pull/873

PR #750, opened 2021-06-09
Hedgin Bets and Requests

joe-elliott:
**What this PR does**:

Hedges GCS/S3/Azure requests using [this library](https://github.com/cristalhq/hedgedhttp). ~This currently only supports s3 and gcs. Azure does not seem to give access to http transport so I'm not sure if we can use this solution there.  This PR is ready for review but I'm still evaluating options for adding tests.~
Azure support added!

Using a `hedge_requests_at` value of `500ms` seeing the following impact in ops:

**frontend latency:**
- p99 from 9.8s -> 2.5s
- p50/p90 down 10s of ms
image

**gcs requests/second:**
- slightly elevated? almost too little to tell. this is unsurprising since 500ms is above our p99

image

Thx to @cristaloleg and @storozhukBM for their help in getting hedgedhttp to support GCS/HTTP2.

**Checklist**
- [ ] Tests updated
- [x] Documentation added


cristaloleg (approved):
SGTM. Minor thing: 2nd param to the `NewClient` can be extracted to a named const, will be easier to get what `2` means.

cristaloleg:
Made a minor change with params validation, no other changes https://github.com/cristalhq/hedgedhttp/releases/tag/v0.5.0

Also @storozhukBM suggested that `hedgedhttp` can be easily abstracted to any other hedging-like work, like adding same feature to Azure client even without access to `http.Transport`.

No sooner said than done - https://github.com/cristalhq/synx/blob/main/hedged.go but not tested and well designed yet. Will be polished next week(s).

tempodb/backend/gcs/gcs.go
```diff
@@ -319,3 +300,47 @@ func (rw *readerWriter) readRange(ctx context.Context, name string, offset int64
...
+		transportOptions = append(transportOptions, option.WithoutAuthentication())
+		customTransport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+	}
+	transport, err := google_http.NewTransport(ctx, customTransport, transportOptions...)
+	if err != nil {
+		return nil, errors.Wrap(err, "creating google http transport")
+	}
+
+	// add instrumentation
+	transport = newInstrumentedTransport(transport)
```
annanay25, line 322: Thinking about this some - should we move instrumentation to after hedging in the middleware pipeline?

tempodb/backend/gcs/gcs.go
```diff
@@ -298,7 +279,7 @@ func (rw *readerWriter) readAllWithModTime(ctx context.Context, name string) ([]
 }
 
 func (rw *readerWriter) readRange(ctx context.Context, name string, offset int64, buffer []byte) error {
-	r, err := rw.bucket.Object(name).NewRangeReader(ctx, offset, int64(len(buffer)))
+	r, err := rw.hedgedBucket.Object(name).NewRangeReader(ctx, offset, int64(len(buffer)))
```
annanay25, line 282: I wonder if 500ms timeouts on readRange would affect the compactors? Since we pull a few MB at a time

annanay25 (commented):
This is amazing 🚀

One more thought:
- Should we add a metric on number of hedged requests that we can track?

tempodb/backend/gcs/gcs.go
mapno: We could get a higher rate of errors than we actually have because of all the retried and cancelled requests. I would instrument the hedging itself maybe.

tempodb/backend/gcs/gcs.go
```diff
@@ -24,56 +25,36 @@ import (
 	"github.com/grafana/tempo/tempodb/backend/util"
 )
 
+const uptoHedgedRequests = 2
```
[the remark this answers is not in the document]
kvrhdn: maxRetries might be confusing since these aren't real retries. I'd still include the word hedge in there somewhere.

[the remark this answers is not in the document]
joe-elliott: agree with @kvrhdn it's not call `maxRetries` b/c 2 means only 1 retry.

tempodb/backend/gcs/gcs.go
joe-elliott: this should instrument both requests independently. i think this is what we want. it shows us our "true" gcs requests.

annanay25: OK. Do correct me if I'm wrong, but what I'm concerned about is that at the moment we don't have a lot of visibility of the number of hedged requests we're making in our deployment. As an example, if our gcs p99 increases to say 700ms, maybe due to some bloom filter/index changes we make in the future, the current instrumentation may "hide" it -

```
without hedged request:
gcs get request -> 700ms
```

```
with hedged:
gcs get request first try -> 500ms -> timed out -> second try -> 50ms
```

While this is exactly the behaviour that we want, this will cause our gcs p99 to flatline at 500ms. we wouldn't know what value we should set for the `hedge_requests_at` config parameter. A retry metric might we useful to track in that case

kvrhdn: This is not entirely correct: in my understanding the first request will not time out when the hedged request starts, it will not be cancelled unless the second request completes. So it would be like this:

```
1st request starts ->                 ...               -> 1st request is cancelled because 2nd completed
                   -> after 500 ms 2nd request starts -> 2nd request completes after 50 ms
```

In this case the first request had a latency of 550 ms.

If the second request is slower, the first one will not be cancelled:

```
1st request starts ->                 ...             -> 1st request completes after 700 ms
                   -> after 500 ms 2nd request starts   -> 2nd request is cancelled
```

So the duration of the first request should still show us the total duration of the operation.

---

Maybe it's interesting to track how often the hedged request wins? This can help us when tuning the system: if the hedged requests is never faster, we shouldn't spend resources launching it.

joe-elliott: The maintainers would need to accept a PR where hedgedghttp takes a prometheus registry and then we do some internal metrics. Personally I'd prefer moving forward as is. Although I agree there will be a small lack of visibility here.

cristaloleg: Feel free to raise a feature request. I think something simple like counters can be added. However I would like to keep library without dependencies (especially Prometheus :)

Maybe it possible to solve with a custom `RoundTripper`, which can track distinguish 1st and any other consecutive request (probably by url or even `*req` address) and update proper metrics?

(also my 5c: agree with @joe-elliott it's a good feature for the future)

joe-elliott:
@grafana/tempo Please re-review
- tests added
- azure support added

This is ready to go from my perspective.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## main / unreleased
 
+* [FEATURE] Hedged requests for S3 and GCS [#750](https://github.com/grafana/tempo/pull/750) (@joe-elliott)
```
kvrhdn, line 3: Don't forget to update this to include Azure 😉

tempodb/backend/azure/azure_helpers.go
```diff
@@ -7,12 +7,17 @@ import (
...
 
+	"github.com/Azure/azure-pipeline-go/pipeline"
 	blob "github.com/Azure/azure-storage-blob-go/azblob"
+	"github.com/cristalhq/hedgedhttp"
 )
 
-const maxRetries = 3
+const (
+	maxRetries        = 1
+	maxHedgedRequests = 2
```
kvrhdn, line 17: The `gcs` and `s3` package have similar properties (`uptoHedgedRequests`). Do we want to keep these separate or combine them somewhere? It might be nice to have them somewhere central so we are sure they stay aligned. But moving them out of the package also removes a lot of the surrounding context though...

tempodb/backend/azure/config.go
```diff
@@ -9,4 +13,5 @@ type Config struct {
 	Endpoint           string         `yaml:"endpoint-suffix"`
 	MaxBuffers         int            `yaml:"max-buffers"`
 	BufferSize         int            `yaml:"buffer-size"`
+	HedgeRequestsAt    time.Duration  `yaml:"hedge-requests-at"`
```
kvrhdn, line 16: Not specific to this PR: I noticed this config uses kebab-case instead of snake_case like the other configs. Is this intentional? Is this just some debt we can't get rid of anymore now?

tempodb/backend/s3/s3.go
```diff
@@ -67,59 +70,27 @@ func (s *overrideSignatureVersion) IsExpired() bool {
...
 
-	core, err := minio.NewCore(cfg.Endpoint, opts)
+	hedgedCore, err := createCore(cfg, true)
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, nil, fmt.Errorf("unexpected error creating hedgedCore: %w", err)
 	}
 
-	// TODO: add custom transport with instrumentation.
-	//client.SetCustomTransport(minio.DefaultTransport(!cfg.Insecure))
```
kvrhdn, line 111: Is this something we want to track in an issue? I noticed only GCS has instrumentation for now (which is not ideal, but probably fine).

kvrhdn (approved):
LGTM! :+1: Left some remarks, nothing blocking.

tempodb/backend/azure/azure_helpers.go
joe-elliott: i think it's fine to leave them internal even though they have the same value. one day they may become independent settings anyway.

tempodb/backend/azure/config.go
joe-elliott: yeah, i noticed this for the first time adding this config option. i think everything is snake except for this azure config. we should consider a breaking change PR where we move azure to the same standard as everything else.

tempodb/backend/s3/s3.go
joe-elliott: yup:

https://github.com/grafana/tempo/issues/758
https://github.com/grafana/tempo/issues/759

storozhukBM:
@joe-elliott 
FYI: in new version https://github.com/cristalhq/hedgedhttp/releases/tag/v0.6.0
We have stats that you can use for metrics

annanay25 (approved):
Thanks @cristaloleg @storozhukBM for the super useful library and great collaboration 🚀 

I agree we can move ahead with this, I'll open an issue for the metrics that we can work on in the future.

PR #754, opened 2021-06-10
Compaction multi-threading

mdisibio:
**What this PR does**:
Compaction is single-threaded and solidly limited to using a single core.  This PR introduces some basic multi-threading by moving iteration to a separate goroutine, which improves performance.

Benchmark was added and shows ~12% improvement:
```
pkg: github.com/grafana/tempo/tempodb
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkCompaction-12     	    3594	    329848 ns/op
BenchmarkCompaction2-12    	    4042	    290434 ns/op
```

Testing in a real env is better:
objects and bytes written up ~20%
image
image

Some notes:
- New tunable: the channel buffer size.  Unfortunately it is in traces (not bytes). Defaulted to 1000 based on testing.
- ID and data slices must be copied to escape the iterators which increases GC and memory pressure slightly.  Interested in trying pooling, but the current `Iterator` interface doesn't expose the ability return slices from `Next()`.   Will explore this in a later PR unless there is interest in trying it now.

n/a

**Checklist**
- [x] Tests updated
- [x] Documentation added


mdisibio:
> Ha! you deleted compactor bookmarks. This is my new favorite PR.

Sorry then to say they were actually just moved into iterator_multiblock.go. Although there was a slight change to retain currentError.

tempodb/encoding/iterator_multiblock.go
```diff
@@ -0,0 +1,194 @@
...
+	for _, b := range i.bookmarks {
+		if !b.done(i.ctx) {
+			return false
+		}
+	}
+	return true
+}
+
+func (i *multiblockIterator) getErr() error {
+	i.errMtx.RLock()
```
joe-elliott, line 86: might be easier to use `uber.atomic` for the error:
https://github.com/uber-go/atomic

tempodb/encoding/iterator_multiblock.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+}
+
+func (i *multiblockIterator) Next(ctx context.Context) (common.ID, []byte, error) {
+	if err := i.getErr(); err != nil {
+		return nil, nil, err
+	}
+
+	select {
+	case <-i.ctx.Done():
```
joe-elliott, line 64: i believe this should use the passed in context and we shouldn't store a context in the struct.

tempodb/encoding/iterator_multiblock.go
```diff
@@ -0,0 +1,194 @@
...
+		if len(lowestID) == 0 || len(lowestObject) == 0 || lowestBookmark == nil {
+			i.setErr(errors.New("failed to find a lowest object in compaction"))
+			return
+		}
+
+		// Copy slices allows data to escape the iterators
+		res := iteratorResult{
+			ID:   append([]byte(nil), lowestID...),
+			Body: append([]byte(nil), lowestObject...),
+		}
```
joe-elliott, line 136: why does the object need to be copied? it didn't before

tempodb/encoding/iterator_multiblock.go
```diff
@@ -0,0 +1,194 @@
...
+	resultsCh    chan iteratorResult
+	quitCh       chan bool
+
+	errMtx sync.RWMutex
+	err    error
+}
+
+type iteratorResult struct {
+	ID   common.ID
+	Body []byte
```
joe-elliott, line 27: `Body` => `Object` ?

can we make this hidden fields as well?

joe-elliott (changes requested):
Comments included.

I'm finding the logic of the iterators and channels to be a bit confusing to follow. I think it's mostly b/c you directly translated the gross bookmark logic into a goroutine. The error path (actual error vs EOF) is I think the part I find most confusing.

tempodb/encoding/iterator_multiblock_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	inner := &testIterator{}
+	for i := 0; i < recordCount; i++ {
+		inner.Add(make([]byte, i), make([]byte, i), nil)
+	}
+
+	// Create iterator and close it after 100ms
+	iter := NewMultiblockIterator(context.TODO(), []Iterator{inner}, recordCount/2, nil, "")
+	time.Sleep(100 * time.Millisecond)
+	iter.Close()
```
joe-elliott, line 114: can we confirm that we can cancel works through `context.Cancel()` as well?

mdisibio: Added test.

tempodb/encoding/iterator_multiblock.go
mdisibio: Previously, there was an implicit copy via `currentBlock.AddObject(id, body)` -> `appender.writer.Write(id, b)`, which worked because it was lock-step with the iterator.  Now it has to be copied earlier so it can be buffered in the channel.

joe-elliott (approved):
Nice!

annanay25:
Nice! Couldn't give this a thorough review but the decoupling of reads and writes using a buffered message queue looks great

PR #755, opened 2021-06-10
replace ioutil.ReadAll with io.ReadAll

wperron:
**What this PR does**:

Update usage of `ioutil.ReadAll` to `io.ReadAll` since the former was marked as deprecated in Go v1.16. From discussion on Slack with @mdisibio I gather that `ReadAllWithEstimate` is typically the preferred way now, however looking at the places where that function is used, there's usually a way to estimate the size of the buffer beforehand. The remaining uses of `ioutil.ReadAll` on the other hand are used to read the `Response` from an HTTP request; while it's entirely possible that the response has a `Content-Length` header, it's also not a guarantee.

_none_

**Checklist**
- [ N/A ] Tests updated
- [ N/A ] Documentation added

### DIFF
--- a/cmd/tempo-query/tempo/plugin.go
+++ b/cmd/tempo-query/tempo/plugin.go
@@ -3,7 +3,7 @@ package tempo
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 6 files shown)


joe-elliott (approved):
Nice. Thanks for the contribution!

PR #762, opened 2021-06-14
jsonnet: add connection_limit to $._config.memcached - increase default to 4096

kvrhdn:
**What this PR does**:
This makes overriding the connection limit more consistent with other configurations. Before you had to override connection limit at `$.memcached.connection_limit`, now you can use `$._config.memcached.connection_limit`

### DIFF
--- a/operations/jsonnet/microservices/config.libsonnet
+++ b/operations/jsonnet/microservices/config.libsonnet
@@ -29,6 +29,7 @@
     },
     memcached: {
       replicas: 3,
+      connection_limit: 1024,
     },
     jaeger_ui: {
       base_path: '/',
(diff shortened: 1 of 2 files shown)


bboreham:
Given that we know several installations of Cortex and Tempo that have increased the limit, should the default be higher than 1024?
(I guess that could be a different PR, just asking here since the idea came to me)

kvrhdn:
> Given that we know several installations of Cortex and Tempo that have increased the limit, should the default be higher than 1024?
> (I guess that could be a different PR, just asking here since the idea came to me)

This is a good idea, there is no drawback to setting the limit to 4096 (AFAIK) so we might as well make this the default.

joe-elliott (approved):

PR #496, opened 2021-02-01
A few dashboard tweaks

mdisibio:
**What this PR does**:
A few tempo operational dashboard tweaks:
1. goroutines panel use {{pod}} instead of {{job}}.  Job is a constant value like "ingester". Pod is what we want "ingester-nn"
2. Working set panel ignore container=POD, which yields not useful metrics like memory = 3MB.

**Which issue(s) this PR fixes**:
n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #764, opened 2021-06-14
Tweak CPU and memory display on operational dashboard

bboreham:
**What this PR does**:

Add container name to Memory and CPU metrics. They are reported per cgroup, so pods with more than one container have several lines. This change allows them to be distinguished.

Also filter out 'pause' container - these metrics are typically very low because the "pause" container is a piece of Kubernetes internals that sits and does nothing. Filtering them out of the display aids readability.

Note in some Kubernetes installations the pause container is labled "POD", which was previously filtered out from the dashboard in #496. See containerd/cri#922 for background on the name.

None

**Checklist**
- [ ] N/A Tests updated
- [ ] N/A Documentation added
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## main / unreleased
 
 * [FEATURE] Added the ability to hedge requests with all backends [#750](https://github.com/grafana/tempo/pull/750) (@joe-elliott)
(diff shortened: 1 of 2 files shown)


joe-elliott:
@bboreham 
The operational dashboard has needed some love for awhile now. Thank you!

Please sign the CLA and we'll get this merged.

joe-elliott (approved):

PR #772, opened 2021-06-17
Docs: add operations/caching page

kvrhdn:
**What this PR does**:
This adds a _Caching_ section under _Deployment and Operations_. Intention is to gather recommendations and troubleshooting info about running proper caching with Tempo.

I guess this is also kind of a runbook? So maybe we should tackle this differently... I'm open for suggestions!


docs/tempo/website/operations/caching.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Caching
+weight: 6
+---
+
+# Caching
+
+Tempo can use an external cache to improve query performance.
+The supported implementations are [Memcached](https://memcached.org/) and [Redis](https://redis.io/). 
+Caching is used by the queriers and compactors.
```
achatterjee-grafana, line 10:
```suggestion
Caching is used by the queriers and compactors.
Tempo can use an external cache to improve query performance.
The supported implementations are [Memcached](https://memcached.org/) and [Redis](https://redis.io/). 
```

docs/tempo/website/operations/caching.md
```diff
@@ -0,0 +1,18 @@
...
+
+### Memcached
```
achatterjee-grafana, line 12: Add information to this section.

docs/tempo/website/operations/caching.md
```diff
@@ -0,0 +1,18 @@
...
+
+#### Connection limit
+
+As a cluster grows in size, the amount of instances of Tempo connecting to the cache servers will also increase.
+By default Memcached has a connection limit of 1024 and if this limit is exceeded new connections will be refused (resulting in HTTP 500 errors).
```
joe-elliott, line 17: resulting in Tempo HTTP 500 errors?  Is there a log line or metric we could share here that would help people know when they're overwhelming memcached?

docs/tempo/website/operations/caching.md
```diff
@@ -0,0 +1,18 @@
...
+This is resolved by increasing the connection limit of Memcached.
```
achatterjee-grafana, line 18:
```suggestion
As a cluster grows in size, the number of instances of Tempo connecting to the cache servers also increases.
By default, Memcached has a connection limit of 1024. If this limit is surpassed new connections are refused (resulting in HTTP 500 errors).
This is resolved by increasing the connection limit of Memcached.
```

achatterjee-grafana (changes requested):
Added doc change request.

docs/tempo/website/operations/caching.md
kvrhdn: Agree, we can share way more information to help out folks :+1:

kvrhdn:
Thanks for the reviews! I've added a bit more detail.

I'm still not sure about the location of this section, it feels a bit like a troubleshooting page? It's very specific to caching though, so I don't want to drop it under troubleshooting/.

WDYT?

docs/tempo/website/operations/caching.md
```diff
@@ -0,0 +1,44 @@
+---
+title: Caching
+weight: 6
+---
+
+# Caching
+
+Caching is used by the queriers and compactors.
+Tempo can use an external cache to improve query performance.
```
annanay25, line 9:
```suggestion
Caching is mainly used to improve query performance by storing bloom filters of all backend blocks which are accessed on every query.

Tempo uses an external cache to improve query performance.
```

annanay25:
> I'm still not sure about the location of this section, it feels a bit like a troubleshooting page? It's very specific to caching though, so I don't want to drop it under troubleshooting/.


I vote for its current location (under Operations).

docs/tempo/website/operations/caching.md
kvrhdn: I've added a small introduction text.

joe-elliott (approved):

PR #775, opened 2021-06-18
Jsonnet: use dedicated configmaps for distributors and ingesters

### LINKED ISSUE #742: [Jsonnet] Ingester ConfigMap changes
These are the different configmaps we run at the moment
```
tempo                    2      76d
tempo-compactor          2      76d
tempo-querier            1      76d
tempo-query              1      76d
tempo-query-frontend     1      76d
```

There are two issues to fix:
- [ ] The ingester (and distributor) should get its own configmap which allows modification of ingester specific values without rolling out other components (in this case - the distributor, which shares the tempo configmap with the ingester)
- [ ] The ingester statefulset should get a new annotation with the config hash of the ingester configmap so that it is auto-rolled when the config changes.

kvrhdn:
**What this PR does**:
Split the `tempo` configmap into `tempo-distributor` and `tempo-ingester`. This will allow both services to be updated independently (changing the config of the ingesters should not impact the distributor and vice versa).

I've also added `config_hash` annotation to both so they get replaced when the config changes.

Fixes #742 

**Checklist**
- [ ] ~~Tests updated~~
- [ ] ~~Documentation added~~


annanay25 (approved):
lgtm!

kvrhdn:
Thanks for updating kube-manifests 🙏🏻

PR #777, opened 2021-06-21
Add histogram for flush size in bytes

bboreham:
**What this PR does**:

Add a histogram for size paralleling `tempo_ingester_flush_duration_seconds`.
This improves the observability of the ingester write path.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [x] `CHANGELOG.md` updated

### DIFF
--- a/modules/ingester/flush.go
+++ b/modules/ingester/flush.go
@@ -36,6 +36,12 @@ var (
 		Help:      "Records the amount of time to flush a complete block.",
 		Buckets:   prometheus.ExponentialBuckets(1, 2, 10),
 	})
+	metricFlushSize = promauto.NewHistogram(prometheus.HistogramOpts{
+		Namespace: "tempo",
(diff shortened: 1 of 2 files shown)


mapno (approved):
LGTM!

modules/ingester/flush.go
```diff
@@ -275,6 +281,7 @@ func (i *Ingester) handleFlush(userID string, blockID uuid.UUID) (retry bool, er
 		start := time.Now()
 		err = i.store.WriteBlock(ctx, block)
 		metricFlushDuration.Observe(time.Since(start).Seconds())
+		metricFlushSize.Observe(float64(block.BlockMeta().Size))
```
joe-elliott, line 284: this is fine, but keep in mind that this is the trace data only. it does not take into account the bloom filters, meta or index.

these other objects are small in comparison, but non-zero.

joe-elliott (approved):

PR #779, opened 2021-06-21
Extend tempodb_gcs_request_duration_seconds histogram

bboreham:
**What this PR does**:

 - Record duration of requests that error - requests that cannot get an http response should be included in timings, particularly if the reason for failure is a timeout.
 - Extend GCS histogram buckets to 80s - the previous cap of 5 seconds is too short to see patterns when writing large blocks.

**Checklist**
- NA Tests updated
- NA Documentation added
- [x] `CHANGELOG.md` updated

### DIFF
--- a/tempodb/backend/gcs/instrumentation.go
+++ b/tempodb/backend/gcs/instrumentation.go
@@ -36,8 +36,12 @@ func newInstrumentedTransport(next http.RoundTripper) http.RoundTripper {
 func (i instrumentedTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	start := time.Now()
 	resp, err := i.next.RoundTrip(req)
+	var status string
 	if err == nil {
-		i.observer.WithLabelValues(req.Method, strconv.Itoa(resp.StatusCode)).Observe(time.Since(start).Seconds())
+		status = strconv.Itoa(resp.StatusCode)
+	} else {
+		status = "500"
 	}
+	i.observer.WithLabelValues(req.Method, status).Observe(time.Since(start).Seconds())
 	return resp, err
 }
--- a/tempodb/backend/gcs/instrumentation.go
+++ b/tempodb/backend/gcs/instrumentation.go
@@ -15,9 +15,8 @@ var (
 		Name:      "gcs_request_duration_seconds",
 		Help:      "Time spent doing GCS requests.",
 
-		// GCS latency seems to range from a few ms to a few secs and is
-		// important.  So use 6 buckets from 5ms to 5s.
-		Buckets: prometheus.ExponentialBuckets(0.005, 4, 6),
+		// We often write large blocks to GCS, so use buckets from 5ms to 80s.
+		Buckets: prometheus.ExponentialBuckets(0.005, 4, 8),
 	}, []string{"operation", "status_code"})
 )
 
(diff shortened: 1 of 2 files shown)


mapno (approved):
LGTM

tempodb/backend/gcs/instrumentation.go
```diff
@@ -36,8 +35,12 @@ func newInstrumentedTransport(next http.RoundTripper) http.RoundTripper {
 func (i instrumentedTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	start := time.Now()
 	resp, err := i.next.RoundTrip(req)
+	var status string
 	if err == nil {
-		i.observer.WithLabelValues(req.Method, strconv.Itoa(resp.StatusCode)).Observe(time.Since(start).Seconds())
+		status = strconv.Itoa(resp.StatusCode)
+	} else {
+		status = "500"
```
joe-elliott, line 42: do we want this change? Recording latency metrics on err return could be misleading b/c err is returned mainly in situations where it was impossible to establish the request. Do we want to record latency on something like a TCP connect timeout?

tempodb/backend/gcs/instrumentation.go
```diff
@@ -15,9 +15,8 @@ var (
 		Name:      "gcs_request_duration_seconds",
 		Help:      "Time spent doing GCS requests.",
 
-		// GCS latency seems to range from a few ms to a few secs and is
-		// important.  So use 6 buckets from 5ms to 5s.
-		Buckets: prometheus.ExponentialBuckets(0.005, 4, 6),
+		// We often write large blocks to GCS, so use buckets from 5ms to 80s.
```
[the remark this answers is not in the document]
bboreham: I believe a single flush will be hundreds of MB.
We ask the API to send that in 10MB chunks but the flush code doesn't see that.

tempodb/backend/gcs/instrumentation.go
bboreham: Well I think we don't have much data on "mainly". I was aiming at the situation where the whole request times out.

Cortex (and therefore Loki) does it - e.g. https://github.com/cortexproject/cortex/blob/8e84b22ce77163f8f0a859067aa344ba18d42c36/pkg/util/middleware/grpc.go#L56

We could break down the error into different kinds - can detect a context cancelled error and assign it 499, for instance.

bboreham: Sorry I think I'm wrong about Loki doing it - not any longer post-NoSQL

joe-elliott: Your link is to GRPC instrumentation. Chunks GCS/HTTP instrumentation is here which does not record latencies on error:

https://github.com/cortexproject/cortex/blob/8e84b22ce77163f8f0a859067aa344ba18d42c36/pkg/chunk/gcp/instrumentation.go#L83

I'm fairly certain that in GRPC unsuccessful status codes are translated into go `error`s and so it makes sense to instrument these. With HTTP unsuccessful status codes are returned with a nil error and the caller is responsible for checking the HTTP status code. I would guess an error here basically boils down to TCP/TLS failures (no response, port refused, etc) which I don't think we should instrument as latencies.

In our case the overall context is being cancelled due to the overall [FlushOpTimeout](https://github.com/grafana/tempo/blob/main/modules/ingester/flush.go#L278) which could occur at any time for a single HTTP request. The flush context could be cancelled in the first millisecond or 10 seconds into any particular HTTP call that was part of the broader effort to ship a block to GCS.

bboreham: We know from Goroutine dumps that a single http call gets blocked for minutes.
Yes it could be that in other cases it gets interrupted at a different time.

bboreham:
Could you say why this PR was closed?

mdisibio:
Hmm I think this was a mistake, sorry.  I was cleaning out old branches.  Can you repush the branch or reopen from a personal fork?

PR #781, opened 2021-06-21
refactor(jsonnet/single-binary): Remove kausal from the root element

jvrplmlmn:
**What this PR does**: Remove the assumption that kausal is always available in the root element, instead we declare it explicitly, because it is an implementation detail that is leaking to the consumers of the libraries.

N/A

**Checklist**
- [ ] ~~Tests updated~~
- [ ] ~~Documentation added~~

### DIFF
--- a/operations/jsonnet/single-binary/tempo.libsonnet
+++ b/operations/jsonnet/single-binary/tempo.libsonnet
@@ -1,14 +1,14 @@
-(import 'ksonnet-util/kausal.libsonnet') +
 (import 'configmap.libsonnet') +
 (import 'config.libsonnet') + 
 {
-  local container = $.core.v1.container,
-  local containerPort = $.core.v1.containerPort,
-  local volumeMount = $.core.v1.volumeMount,
-  local pvc = $.core.v1.persistentVolumeClaim,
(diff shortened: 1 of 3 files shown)


kvrhdn (approved):
Tested this with the `tempo-single-binary` example and works as expected :+1:

I see the resource settings actually didn't work before, so thanks for fixing that 🙂

operations/jsonnet/single-binary/tempo.libsonnet
```diff
@@ -43,7 +43,9 @@
     container.withVolumeMounts([
       volumeMount.new(tempo_config_volume, '/conf'),
       volumeMount.new(tempo_data_volume, '/var/tempo'),
-    ]),
+    ]) +
+    k.util.resourcesRequests('3', '3Gi') +
```
joe-elliott, line 47: i'd rather not set limits/requests in the vendored jsonnet. these values are so specific to a given installation that i think they should always be set by the caller.

joe-elliott: heh, nevermind, i see we were doing it before :P

joe-elliott (approved):
thanks!

PR #782, opened 2021-06-21
refactor(jsonnet/microservices): Remove kausal from the root element

jvrplmlmn:
**What this PR does**: Remove the assumption that kausal is always available in the root element, instead we declare it explicitly, because it is an implementation detail that is leaking to the consumers of the libraries.

N/A

**Checklist**
- [ ] ~~Tests updated~~
- [ ] ~~Documentation added~~

### DIFF
--- a/operations/jsonnet/microservices/common.libsonnet
+++ b/operations/jsonnet/microservices/common.libsonnet
@@ -1,6 +1,7 @@
 {
   util+:: {
-    local container = $.core.v1.container,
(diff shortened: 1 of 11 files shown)


kvrhdn (approved):
Tested this with the `tempo-microservices` example and works as expected as well 👍
There is no diff after applying these changes.

joe-elliott (approved):
thanks!

PR #785, opened 2021-06-22
refactor(example/tk): Remove kausal from the root element

kvrhdn:
**What this PR does**:
Pretty much the same thing as #781 and #782 but for the tanka examples instead.

> Remove the assumption that kausal is always available in the root element, instead we declare it explicitly, because it is an implementation detail that is leaking to the consumers of the libraries.

~~Fixes #<issue number>~~

**Checklist**
- [ ] ~~Tests updated~~
- [ ] ~~Documentation added~~

### DIFF
--- a/example/tk/lib/grafana/main.libsonnet
+++ b/example/tk/lib/grafana/main.libsonnet
@@ -1,12 +1,11 @@
-local k = import 'ksonnet-util/kausal.libsonnet';
-
 {
-  local configMap = $.core.v1.configMap,
-  local container = $.core.v1.container,
-  local volumeMount = $.core.v1.volumeMount,
-  local deployment = $.apps.v1.deployment,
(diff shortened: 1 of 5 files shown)


jvrplmlmn (commented):
Reviewed the files in their entirety and LG2M modulo one ocurrence:

Can't give inline comments because they are out of the diff, so take a look at `local ingress = $.networking.v1beta1.ingress,` in `example/tk/tempo-microservices/main.jsonnet`

joe-elliott (approved):

kvrhdn:
> Can't give inline comments because they are out of the diff, so take a look at local ingress = $.networking.v1beta1.ingress, in example/tk/tempo-microservices/main.jsonnet

I'll fix this in a later PR :+1:

PR #789, opened 2021-06-23
Refactor Blocklist Polling

joe-elliott:
**What this PR does**:
Refactors blocklist polling code by moving it into its own package and adding tests. This is in preparation for adding a bucket index.

### DIFF
--- a/tempodb/blocklist/poller.go
+++ b/tempodb/blocklist/poller.go
@@ -1,0 +1,162 @@
+package blocklist
+
+import (
+	"context"
+	"sort"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/grafana/tempo/pkg/boundedwaitgroup"
+	"github.com/grafana/tempo/tempodb/backend"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"go.uber.org/atomic"
+)
+
+var (
+	metricBlocklistErrors = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "tempodb",
+		Name:      "blocklist_poll_errors_total",
+		Help:      "Total number of times an error occurred while polling the blocklist.",
+	}, []string{"tenant"})
+	metricBlocklistPollDuration = promauto.NewHistogram(prometheus.HistogramOpts{
+		Namespace: "tempodb",
+		Name:      "blocklist_poll_duration_seconds",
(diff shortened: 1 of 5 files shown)


mapno (approved):
LGTM

tempodb/tempodb.go
```diff
@@ -389,101 +375,18 @@ func (rw *readerWriter) maintenanceLoop() {
 }
 
 func (rw *readerWriter) pollBlocklist() {
-	start := time.Now()
-	defer func() { metricBlocklistPollDuration.Observe(time.Since(start).Seconds()) }()
+	blocklist, compactedBlocklist, err := rw.blocklistPoller.Do()
```
annanay25, line 378: super minor nit: `readerWriter` is doing the polling so its more of a block lister than a block poller.. :P

annanay25 (approved):
lgtm!

tempodb/blocklist/poller.go
```diff
@@ -0,0 +1,162 @@
...
+	if err = anyError.Load(); err != nil {
+		return nil, nil, err
+	}
+
+	newBlockList := make([]*backend.BlockMeta, 0, len(blockIDs))
+	for m := range chMeta {
+		newBlockList = append(newBlockList, m)
+	}
+	sort.Slice(newBlockList, func(i, j int) bool {
+		return newBlockList[i].StartTime.Before(newBlockList[j].StartTime)
```
zalegrala, line 127: @joe-elliott Do you recall why we sort by time?  Is this to reduce the work necessary to find recent blocks?  I'm wondering we can sort by ID instead, which might enable faster searching on the blocklist.

joe-elliott: the only thing i can think of is that in the query frontend this will make it consistently push jobs for certain blocks first which would make the search more likely to return results on one end or another of the range.

the way this is written we are ordering by starttime ascending which makes it most likely to return results at the beginning of the range. if anything i think we'd prefer the opposite and return the most recent results satisfying a particular query.

sorting 10k+ blocks might take a bit and shouldn't be in the query path. i think we should keep this in an order that the query path prefers. however, sorting 10k+ blocks in the polling cycle would be more acceptable.

zalegrala: I was thinking of pushing this sort into the blocklist.List struct, perhaps on the ApplyPollingResults ordered by ID.  Per your comment above, are you saying that would impact the query path here negatively?

PR #790, opened 2021-06-23
Consolidate backend metrics and add hedged request metric

### LINKED ISSUE #760: Add metrics for better visibility around hedged requests
#750 added Hedged requests to GCS/S3/Azure backends.

There was a discussion on the PR to add metrics for tracking the performance of hedged requests, and the corresponding stats have been introduced in the hedgedhttp library in the latest release - https://github.com/cristalhq/hedgedhttp/releases/tag/v0.6.0

Use these stats to generate prometheus metrics for each backend.

JosephWoodward:
**What this PR does**:

This PR does two things:

- Consolidates the backend metrics into a more general purpose metrics
- Adds a new metrics to track the number of hedged requests

Fixes #760

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott:
Thanks for digging into this. I was thinking more like consolidating the three instrumentation.go files we have in gcs/s3/azure and then rolling this new metric into that.

It's a shame that these metrics names are different but we should keep the metrics backwards compatible:
`tempodb_s3_request_duration_seconds`
`tempodb_azure_request_duration_seconds`
`tempodb_gcs_request_duration_seconds`

All new metrics should not contain the name of the backend and should be the same no matter what backend you choose.

JosephWoodward:
> Thanks for digging into this. I was thinking more like consolidating the three instrumentation.go files we have in gcs/s3/azure and then rolling this new metric into that.
> 
> It's a shame that these metrics names are different but we should keep the metrics backwards compatible:
> `tempodb_s3_request_duration_seconds`
> `tempodb_azure_request_duration_seconds`
> `tempodb_gcs_request_duration_seconds`
> 
> All new metrics should not contain the name of the backend and should be the same no matter what backend you choose.

I see, and grabbing the metrics via the `stats.Snapshot()` method?

joe-elliott:
Yup, that's what I was thinking.

JosephWoodward:
Hi @joe-elliott, I've been working on this over the past few evenings and feel it's getting to a point where it'd be valuable to get some feedback on the approach. When you have moment would you mind having a quick look?

Thanks

tempodb/backend/instrumentation/backend_transports.go
```diff
@@ -0,0 +1,80 @@
...
+	}, []string{"operation", "status_code"})
+
+	azureRequestDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: "tempodb",
+		Name:      "azure_request_duration_seconds",
+		Help:      "Time spent doing Azure requests.",
+		Buckets:   prometheus.ExponentialBuckets(0.005, 4, 6),
+	}, []string{"operation", "status_code"})
+
+	requestDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
```
joe-elliott, line 34: thank you for providing an upgrade path. let's note in the changelog that the old metrics are deprecated and this should be used instead.

joe-elliott (commented):
A well thought out addition. Thank you.

Please add a changelog entry mentioning the new metrics and the deprecation of the old. Other than that this look ready to go.

tempodb/backend/instrumentation/backend_transports.go
JosephWoodward: Done. Is it worth adding a note that the older metrics are deprecated in the metrics help?

JosephWoodward:
@joe-elliott This PR is good to go if you're happy with it.

tempodb/backend/instrumentation/backend_transports.go
joe-elliott: Agree. Let's add "DEPRECATED: See tempodb_backend_request_duration_seconds"

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## main / unreleased
 
+* [ENHANCEMENT] Added hedged request metric `backend_hedged_roundtrips_total` and general purpose `backend_request_duration_seconds` metric that will supersede soon to be deprecated storage specific metrics (eg `request_duration_seconds`). [#790](https://github.com/grafana/tempo/pull/790) (@JosephWoodward)
```
joe-elliott, line 3: Can you list the deprecated 3 explicitly? 

Also all the metrics are prefixed by `tempodb_`

JosephWoodward: > Can you list the deprecated 3 explicitly?
Yeah, no problem

> Also all the metrics are prefixed by `tempodb_`
Good catch! I forgot about the namespace. That's done now.

tempodb/backend/instrumentation/backend_transports.go
JosephWoodward: Sounds good, that's done.

joe-elliott (approved):
Excellent! Thank you for the contribution.

JosephWoodward:
@joe-elliott No problem! I'm pretty new to Go so this has been a great learning opportunity on a reasonably large, real-world Go product. I'll look at picking up another issue now that this has been merged.

Issue #701, opened 2021-05-14
Microservice jsonnet library ignores resource requests and limits


PatTheSilent:
**Describe the bug**
Because of import ordering in `tempo.libsonnet` resouce requirements and limits in `config.libsonnet` are ignored.

`config.libsonnet` defines resource limits and requirements [operations/jsonnet/microservices/config.libsonnet#L59](https://github.com/grafana/tempo/blob/main/operations/jsonnet/microservices/tempo.libsonnet#L5)

`tempo.libsonnet` imports `config.libsonnet` before all other relevant modules [operations/jsonnet/microservices/tempo.libsonnet#L4](https://github.com/grafana/tempo/blob/main/operations/jsonnet/microservices/tempo.libsonnet#L4)

`tempo_compactor_container` from config.libsonnet is overwritten by `tempo_compactor_container` from `compactor.libsonnet`

It's not a big deal and one can easily set those in ones tanka environment but it's confusing because when looking only at config.libsonnet I'd expect those resource constraints to be present in PodSpecs. 

I'd suggest changing `config.libsonnet` to 

```jsonnet
compactor: {
  replicas: 1,
  resources: {
    limits: {
      cpu: '',
      memory: ''
    },
    requests: {
      cpu: '',
      memory: ''
    }
  }
},
```

And then handle that in `compactor.libsonnet`

```jsonnet
+ util.resourcesRequests($._config.compactor.resources.requests.cpu, $._config.compactor.resources.requests.mem)
```

I'm using compactor as an example because it's spiky resource consuption caused some throttling on my VMs without those requests and limits but this is true for all Tempo components.

**To Reproduce**
Use https://github.com/grafana/tempo/blob/main/operations/jsonnet/microservices/tempo.libsonnet

**Expected behavior**
Microservices have resource requirements and limits set

**Environment:**
 - Infrastructure: Kubernetes
 - Deployment tool: jsonnet

**Additional Context**
None

joe-elliott:
I think this would be a nice addition.

Currently we set limits directly on the container:

```
  tempo_querier_container+::
    k.util.resourcesRequests('500m', '2Gi') +
    k.util.resourcesLimits('3', '4Gi')
```

but I agree adding some structure around them could be good. If you are able to submit a PR please do!

PatTheSilent:
Haven't yet had the time to look at that. I'm hoping to find some time next week.

kvrhdn:
I can look into this, I was working on the tanka config anyway 🙂

[...]

PR #793, opened 2021-06-24
Move resource requests and limits into $._config

kvrhdn:
**What this PR does**:
Move all Kubernetes resource settings into `$._config`, this way users only need to override `$._config` and don't need to know about the internals (`tempo_compactor_container` etc.).

Fixes https://github.com/grafana/tempo/issues/701: resource settings were not applied correctly due to import ordering.

**Checklist**
- [ ] ~~Tests updated~~
- [ ] ~~Documentation added~~

### DIFF
--- a/operations/jsonnet/microservices/config.libsonnet
+++ b/operations/jsonnet/microservices/config.libsonnet
@@ -11,21 +11,71 @@
     gossip_member_label: 'tempo-gossip-member',
     compactor: {
       replicas: 1,
+      resources: {
+        requests: {
+          cpu: '500m',
+          memory: '3Gi',
+        },
+        limits: {
+          cpu: '1',
+          memory: '5Gi',
+        },
+      },
     },
     query_frontend: {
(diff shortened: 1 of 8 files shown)


operations/jsonnet/microservices/config.libsonnet
```diff
@@ -11,21 +11,71 @@
...
       replicas: 1,
+      resources: {
+        requests: {
+          cpu: '500m',
+          memory: '3Gi',
+        },
+        limits: {
+          cpu: '1',
+          memory: '5Gi',
+        },
```
kvrhdn, line 22: Since the compactor is multi-threaded now, should we bump these resource settings?

joe-elliott: I don't think it really matters. None of these limits will work for a high scale deployment.

joe-elliott (approved):
lgtm!

PR #798, opened 2021-06-30
Additional Guard code while parsing the header

joe-elliott:
**What this PR does**:
Adds guard code to catch corrupt blocks and return an error instead of panic'ing.


annanay25 (approved):
LGTM.

Related: I remember a discussion around moving these to proto eventually. Might be worth opening an issue for later?

PR #805, opened 2021-07-07
Add support for caching based on compaction level and block age

annanay25:
**What this PR does**:
Adds support for caching based on compaction level and max block age. Ideally, we want to cache bloom filters of blocks that will be around longer - so blocks which reach higher levels of compaction soon.

Fixes #na!

**Checklist**
- [x] Tests updated
- [x] Documentation added


annanay25:
There's one last thing left to implement: remove any caching while writing objects at the ingester. At the compactor we can have a similar shouldCache function as implemented for querier.

joe-elliott:
Don't we need to merge #818 first before this?

> remove any caching while writing objects at the ingester

Why wouldn't we cache objects written by the ingester?

annanay25:
> Don't we need to merge #818 first before this?

That's right. 818 is cleaned up and ready to review.

> Why wouldn't we cache objects written by the ingester?

My bad, not remove caching, the write methods on ingester/compactor will use `ShouldCache()` to decide whether the objects should be cached.

docs/tempo/website/operations/caching.md
```diff
@@ -46,3 +46,33 @@ Too many open connections
...
+        [cache_max_block_age: <duration>]
+```
+
+Using a combination of these config options, we can narrow down on which bloom filters are cached, thereby reducing our
+cache eviction rate, and increasing our cache hit rate. Nice!
+
+So how do we decide the values of these config parameters? We have added a new command to [tempo-cli](tempo_cli) that
+prints a summary of bloom filter shards per day and per compaction level. The result looks something like this:
+
+<p align="center"><img src="../cache-summary.png" alt="Cache summary"></p>
```
kvrhdn, line 75: I think this image is missing? 😅

tempodb/tempodb.go
```diff
@@ -466,6 +478,36 @@ func (rw *readerWriter) updateBlocklist(tenantID string, add []*backend.BlockMet
...
 }
 
+func (rw *readerWriter) shouldCache(meta *backend.BlockMeta, curTime time.Time) bool {
+	// compaction level is _atleast_ CacheMinCompactionLevel
+	if rw.cfg.CacheMinCompactionLevel > 0 && meta.CompactionLevel < rw.cfg.CacheMinCompactionLevel {
+		return false
+	}
+
+	// block is not older than CacheMaxBlockAge
+	if rw.cfg.CacheMaxBlockAge > 0 && curTime.Sub(meta.StartTime) > rw.cfg.CacheMaxBlockAge {
```
kvrhdn, line 488: ⚠️  _This comment is about the declaration of `BlockMeta.StartTime`, but I can't add a comment there in GitHub._

The field declaration of BlockMeta.StartTime has a comment 
```
// Currently mostly meaningless but roughly matches to the time the first obj was written to this block
```
Can we update this? Since we use this to make caching decisions it's not meaningless anymore.

tempodb/tempodb.go
```diff
@@ -100,6 +101,9 @@ type readerWriter struct {
 	w backend.Writer
 	c backend.Compactor
 
+	uncachedReader backend.Reader
+	uncachedWriter backend.Writer
```
kvrhdn, line 105: It would be nice if we could shield off direct access to `r`, `w`, `uncachedReader` and `uncachedWriter`. So you are always forced to call `getReaderForBlock` and `getWriterForBlock`.
Unfortunately we can't shield private fields because everything is in the same package.

I think we should either add a comment to warn you or somehow combine these fields into a new struct that knows what to cache.

When I search for usage of `r` and `w` I still find some code using them directly (so without `getXxxForBlock`). I don't know if this is okay.

`r`:
- compactor.go:154 (`compact`)
- compactor.go:160 (`compact`)
- tempodb.go:208 (`CompleteBlock`)

`w`:
- compactor.go:251 (`appendBlock`)
- tempodb.go:208 (`CompleteBlock`)

kvrhdn:
Overall changes look good and are easy to follow. It's fine for me to merge now and address the comments later.

kvrhdn (approved):
Overall changes look good and are easy to follow. It's fine for me to merge now and address the comments later.

tempodb/tempodb.go
mdisibio: Some cases are unavoidable, such as compactor.go:154 to the block meta from the backend, so the fields will have to remain accessible.  compactor:160 and :251 are ok because they are iterating or flushing data, which does not involve the bloom filter.    tempodb.go:208 actually looks like obsolete code only used by a querier test.  The ingester used to call this when iterating the wal, but now it calls CompleteBlockWithBackend.

kvrhdn: Why is `compactor.go:154` unavoidable? Using `r` directly means we read using the cache, what's the difference with using the reader we get from `getReaderForBlock`?

docs/tempo/website/operations/caching.md
mdisibio: Good catch, the image is there but this link is wrong. Will fix.

tempodb/tempodb.go
mdisibio: You're right.  I was thinking that there will be cases where the meta is not available, i.e. a call to get the meta, but in this case the compactor does already have it.

mdisibio (approved):

Issue #471, opened 2021-01-25
Support FindService && FindOperations


swartz-k:
**Is your feature request related to a problem? Please describe.**
While using 16686 for query traces, Serive && Operations are not querable.

**Describe the solution you'd like**
Usr 16686 Jaeger with tempo grpc plugin as Jaeger total experience.

**Describe alternatives you've considered**
Support Memory Index Label & Boltdb shipper for multi info query.
I have done this like below.
![image](https://user-images.githubusercontent.com/49771587/105661193-ccc7fb80-5f07-11eb-8f8d-8a2b6345425c.png)

**Additional context**

1. In `ingester instance` add `github.com/cortexproject/cortex/pkg/ingester/index` to cache Service and Operation in method `modules/ingester/instance.go:284 getOrCreateTrace`
2. In `tempo.proto` add struct such as `GetServicesRequest` `GetOperationsRequest`

joe-elliott:
Thanks for the feature request.  I'm unsure how I feel about this.  It's kind of a clever shortcut to allow for querying this information from Tempo.  For clarity, we intend to eventually add search to Tempo, but we intend to do it by augmenting the blocks in a way that they become searchable.

I feel like there's limited value in returning services and operations if broader search capabilities are not present, but I'm certainly open to be persuaded otherwise.  Also, I would be very hesitant to persist this information to the backend since it may complicate future attempts to add search, but perhaps an in-memory/ephemeral option would be feasible.

swartz-k:
Yes. I just do it as a demo to research wheter it possible to support query traces in a broader way.
I had tested it by add an app which has about 500qps and total 999999 requests, per request create about 5 spans, those will cost memory about 10Gi. Also in memory way, search traces with multi ingester pods may have some problems.
As you said, Add search to Tempo is a better way to users and also make Tempo suitable for more situation.
Should we do some work about design docs for Tempo Searching like Loki does?

joe-elliott:
Internally we have only the seed of a "design doc".  Really just a scratchpad with ideas/challenges/initial thoughts on adding search to Tempo.  We will discuss at our next meeting about whether we wan to make this one public.  It's less of a design doc and more of a scratchpad.

swartz-k:
Thanks a lot. I will do some works about that too. Hope to make Tempo Better together.

joe-elliott:
So if the question we are trying to answer is:  What services and operations are pushing to Tempo I think I would prefer using metrics.

Once this is merged:
[...]

Issue #216, opened 2020-10-09
Compactors Have a Memory Leak


joe-elliott:
Occasionally compactor memory blows up and OOMs:

![image](https://user-images.githubusercontent.com/2272392/95597607-dbec7180-0a1c-11eb-88cb-04c36a768470.png)

Compactors come back online quickly and tend to recover so the system continues working, but we still need to address.  I believe this related to very large traces (100k+ spans) going through the `CombineTraces` method.

PR #806, opened 2021-07-09
Search v0.1.0 (ingester only)

mdisibio:
**What this PR does**:
This PR adds basic search capabilities of traces in the ingesters.  The approach should be decently fast and powerful enough for a first pass.  Flatbuffer-encoded metadata for each trace is stored in a new file alongside the block data.  When searching, the flatbuffer metadata is evaluated and matching traces are returned.  The OTLP/protobuf block data is not involved at all.   A new search api is exposed in the query-frontend/querier/ingester, and it can be called directly (by a Grafana experimental UI that is in the works), or via tempo-query which is also updated to translate the jaeger search conditions. There are also apis to lookup attribute names and values (for autocomplete and jaeger dropdowns).

Flatbuffers are quite fast and this basic implementation can search 3+GB/s of files, based on the benchmark in /tempodb/search/.  With additional tuning I believe we could increase it further, but this is already enough to saturate most disk and network i/o, therefore a better direction might be to look at indexing or compression. 

Functionality must be enabled by setting `search_enabled : true` at the root yaml config for query-frontend, querier, and distributor.  This causes the search apis to be registered and the distributor to start capturing search data of traces as they are received. Ingester is backwards compatible and tolerates data present or not.

**Query Details**
This approach is tag-oriented.  The search metadata effectively flattens a trace down to all unique key/value pairs for span and resource attributes, and the min/max start and end times.  All attributes are coerced to strings.  Therefore this first version is quite basic and can answer questions about hits anywhere within a trace, or the overall trace duration.  But it cannot satisfy complex searches on individual spans except the root.  

Examples:  
- service.name="myservice" // Find traces where any span had resource attribute service.name
- root.service.name = "myservice" // Find traces where root span had etc
- http.status_code="500" // Find traces where any span had attribute http.status_code=500

Combining conditions is matching multiple hits anywhere in the trace:
- http.url="/api/thing" http.status_code="500" // Find traces where there exists both of these values, but not necessarily on the same span.

**Implementation Details**
Search data is extracted in the distributor since this is the only location where the trace is available in "cleartext".  Flatbuffer metadata is built and byte slices are sent from distributor->ingester.  Ingester stores this data alongside live traces, and it is eventually flushed to the WAL (separate files in /wal), and completed to the local backend (new files in /wal/blocks/<tenant>/<blockid>/search).  When searching, the ingester checks in all 3 locations.

Flatbuffers are a compiled schema. There is a new pkg/tempofb with this and `make gen-flat` to compile it.  One key detail is the use of `flatbuffers.CreateSharedString` which interns a string in the block of data, leading to efficient storage for common tags and values.

Search results are trace headers and not entire trace bodies. It includes basic details like id, duration, service, operation.  Also included in the api response is some basic metrics for how many traces, blocks, and bytes were inspected. This will let us quickly gauge the performance of various queries.

**Large remaining gaps / next steps**
Consensus is that these are not needed for this first pass and will be addressed in the future.
- All search data is abandoned on each restart. This will be resolved when versioning for search files is introduced, but not included now so that we can continue to experiment with the file format. Versioning is a requirement if/when search data is pushed to the full backend (s3, gcs, etc).
- Explore extracting common/well-known tags to their own fields. This will be even more efficient than the interned strings, and also faster to evaluate.  When not present there should be no overhead because of how flatbuffers skips encoding for non-present fields.  This is not done yet because it gets messy/repetitive quickly, if we want to extract dozens or hundreds of well-known tags (which I think is a good direction).

**Small remaining gaps / next steps. Not necessarily required in this PR but open to feedback**
- When searching live traces don't lock the mutex the whole time. Maybe unlock/relock between each trace?  
- Change signature of search api?  Maybe POST instead of GET.
- Instrumentation: tracing, metrics, logging for the search itself.

Fixes: #471 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


annanay25 (commented):
I had a few observations reading this PR (can be addressed in follow-ups, open to feedback):
- [x] we need limits on the size of search data created for a given trace
- [x] test on distributor extract search data

Created issues for the following:
- integrations test
- api docs

modules/ingester/instance.go
```diff
@@ -488,12 +548,223 @@ func (i *instance) rediscoverLocalBlocks(ctx context.Context) error {
...
 
 	return nil
 }
+
+func (i *instance) Search(ctx context.Context, req *tempopb.SearchRequest) ([]*tempopb.TraceSearchMetadata, error) {
+
+	done := make(chan struct{})
+	defer close(done)
+
+	var resultsCh []<-chan *tempopb.TraceSearchMetadata
```
annanay25, line 569: Nice addition. Should we add an `errorCh` as well? Each of the individual search functions could fail silently.

modules/search/search_results.go
```diff
@@ -0,0 +1,101 @@
...
+	// Closing done channel makes all subsequent and blocked calls to AddResult return
+	// quit immediately.
+	close(sr.doneCh)
+	sr.quit.Store(true)
+}
+
+// SetWorkerCount sets the number of workers that will be started
+func (sr *SearchResults) SetWorkerCount(count int32) {
+	sr.workerCount.Add(count)
+}
```
annanay25, line 73: @mdisibio fyi changed this because of -

```
tempo_1                     | panic: close of closed channel
tempo_1                     |
tempo_1                     | goroutine 1685 [running]:
tempo_1                     | github.com/grafana/tempo/modules/search.(*SearchResults).FinishWorker(0xc0010ec930)
tempo_1                     | 	/Users/annanay/Desktop/git/go/src/github.com/grafana/tempo/modules/search/search_results.go:83 +0x48
tempo_1                     | github.com/grafana/tempo/modules/ingester.(*instance).searchWAL.func1(0xc0010ec930, 0x29a06a8, 0xc0010ec8d0, 0xc000d18970, 0x1, 0x1, 0x295c0e0, 0xc000d1a720)
tempo_1                     | 	/Users/annanay/Desktop/git/go/src/github.com/grafana/tempo/modules/ingester/instance_search.go:98 +0xb9
tempo_1                     | created by github.com/grafana/tempo/modules/ingester.(*instance).searchWAL
tempo_1                     | 	/Users/annanay/Desktop/git/go/src/github.com/grafana/tempo/modules/ingester/instance_search.go:95 +0x165
```

Can move this logic around more if needed.

modules/search/search_results.go
```diff
@@ -0,0 +1,101 @@
...
+	resultsCh   chan *tempopb.TraceSearchMetadata
+	doneCh      chan struct{}
+	quit        atomic.Bool
+	workerCount atomic.Int32
+
+	tracesInspected atomic.Uint32
+	bytesInspected  atomic.Uint64
+}
+
+func NewSearchResults() *SearchResults {
```
joe-elliott, line 23: Unsure if this makes sense but we already have the worker pool code in `/tempodb/pool`. Perhaps it can be reused?

mdisibio: pool.RunJobs waits for jobs to complete and returns the results synchronously.  That is not the direction I was going for, which is that the instance can end the search at any time because it has received enough results, or a deadline (timeout), and not wait for all tasks to complete.  For instance, a query of `name=xyz` could find enough results quickly and there is no reason to keep executing against the rest of the data.

modules/ingester/instance.go
mdisibio: True.  We could bubble errors up and log them, but not sure where that is most appropriate.   The current instance search method doesn't wait for tasks to complete, and may not be around to log the errors.

modules/search/search_results.go
joe-elliott: +1

modules/ingester/instance_search_test.go
```diff
@@ -159,3 +160,84 @@ func TestInstanceSearchNoData(t *testing.T) {
 	assert.NoError(t, err)
 	require.Len(t, sr.Traces, 0)
 }
+
+func TestInstanceSearchDoesNotRace(t *testing.T) {
```
annanay25, line 164: This currently detects a race:
```
WARNING: DATA RACE
Write at 0x00c000111d40 by goroutine 470:
  runtime.mapassign_fast64()
      /usr/local/go/src/runtime/map_fast64.go:92 +0x0
  github.com/grafana/tempo/tempodb/encoding.(*appender).Append()
      /Users/annanay/Desktop/git/go/src/github.com/grafana/tempo/tempodb/encoding/appender.go:60 +0x3cf
  github.com/grafana/tempo/modules/search.(*StreamingSearchBlock).Append()
      /Users/annanay/Desktop/git/go/src/github.com/grafana/tempo/modules/search/streaming_search_block.go:90 +0x5b3
  github.com/grafana/tempo/modules/ingester.(*instance).writeTraceToHeadBlock()
      /Users/annanay/Desktop/git/go/src/github.com/grafana/tempo/modules/ingester/instance.go:498 +0x26b
  github.com/grafana/tempo/modules/ingester.(*instance).CutCompleteTraces()
      /Users/annanay/Desktop/git/go/src/github.com/grafana/tempo/modules/ingester/instance.go:185 +0x2a4
  github.com/grafana/tempo/modules/ingester.TestInstanceSearchDoesNotRace.func3()
      /Users/annanay/Desktop/git/go/src/github.com/grafana/tempo/modules/ingester/instance_search_test.go:217 +0x5c
  github.com/grafana/tempo/modules/ingester.TestInstanceSearchDoesNotRace.func1()
      /Users/annanay/Desktop/git/go/src/github.com/grafana/tempo/modules/ingester/instance_search_test.go:190 +0x4b

Previous read at 0x00c000111d40 by goroutine 474:
  github.com/grafana/tempo/tempodb/encoding.(*appender).Records()
      /Users/annanay/Desktop/git/go/src/github.com/grafana/tempo/tempodb/encoding/appender.go:67 +0x8a
  github.com/grafana/tempo/modules/search.(*StreamingSearchBlock).Search()
      /Users/annanay/Desktop/git/go/src/github.com/grafana/tempo/modules/search/streaming_search_block.go:96 +0x70
  github.com/grafana/tempo/modules/ingester.(*instance).searchWAL.func1()
      /Users/annanay/Desktop/git/go/src/github.com/grafana/tempo/modules/ingester/instance_search.go:98 +0xe6
```

modules/ingester/instance.go
annanay25: I'm thinking in the `instance.Search()` method, we could `select` on a `resultsCh` and `errorCh`. [At line 32](https://github.com/grafana/tempo/pull/806/files#diff-8c9aac72ec16ad39eb093753942cde7f893be0b5950578ffe10ef003d2ff5237R32), once all the workers have been fired off, instead of
```
	for result := range sr.Results() {
```
we can have something like:

```
for {
  select {
    case <- sr.Results():
      // handle results
    case <- sr.Errors():
      // handle errors
  }
}
```

modules/ingester/instance_search_test.go
annanay25: I think we discussed this one already @mdisibio 

Some searchBlock specific mutexes need to be set up I guess.

tempodb/encoding/appender.go
```diff
@@ -22,6 +23,7 @@ type appender struct {
 	records       map[uint64][]common.Record
 	hash          hash.Hash64
 	currentOffset uint64
+	mtx           sync.RWMutex
```
annanay25, line 26: So the "search headblock" is the only place where there is concurrent access between Append() and Records()

How about we separate the search headblock into a new field and lock that instead of adding the mtx to the appender?

annanay25: Added something that works to my personal branch - https://github.com/annanay25/tempo/commit/e2636556a7863f353ea25d89c6316c0926ad1ab9

Will push here if it looks good to you

annanay25: Pushed.

annanay25 (commented):
First high level review of the query path.

modules/frontend/frontend.go
```diff
@@ -97,76 +180,62 @@ func NewTripperware(cfg Config, logger log.Logger, registerer prometheus.Registe
...
-}
+			// check marshalling format
+			marshallingFormat := util.JSONTypeHeaderValue
+			if r.Header.Get(util.AcceptHeaderKey) == util.ProtobufTypeHeaderValue {
+				marshallingFormat = util.ProtobufTypeHeaderValue
+			}
 
-// MiddlewareFunc is like http.HandlerFunc, but for Middleware.
-type MiddlewareFunc func(Handler) Handler
+			// TODO GRPC requests aren't handled properply yet
```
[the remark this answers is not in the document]
kvrhdn: Yeah, I'm kind of annoyed at this tbh. Having support for json is nice to view stuff in your browser/curl, but if we want to support json while using protobuf internally we would have to convert between them here. That's the commented out code underneath `rt.RoundTrip(r)`.

The difference between `NewTracesTripperware` and `NewSearchTripperware` is that search can return multiple data structs, so we would have to check what path we are on to determine which struct to unmarshal the data into.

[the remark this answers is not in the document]
kvrhdn: Personally I'd like to keep support for json in the query-frontend, just because it's a good for debugging issues when Grafana isn't working.

modules/frontend/frontend.go
```diff
@@ -25,15 +25,107 @@ import (
...
+	req = req.WithContext(ctx)
+
+	// route the request to the appropriate RoundTripper
+	switch op := getOperation(req.URL.Path); op {
+	case TracesOp:
+		resp, err = r.traces.RoundTrip(req)
+	case SearchOp:
+		resp, err = r.search.RoundTrip(req)
+	default:
+		resp, err = r.next.RoundTrip(req)
```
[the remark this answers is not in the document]
kvrhdn: This is a bit convoluted: in `NewTripperware` (at the top of the file) we are wrapping `next` with traces and search tripperware. So if we call `r.traces.RoundTrip` this will execute the traces stuff and then call `next.RoundTrip`.
The default case skips the traces and search tripperware and directly calls `next.RoundTrip`.
So all calls end up executing `next.RoundTrip`, but when the path matches `/api/traces` or `/api/search` we perform extra validation and conversions.

This is basically the same code [as Loki](https://github.com/grafana/loki/blob/17f4a73a24c8afc36bcb4e2795c6b1e06d4b95a1/pkg/querier/queryrange/roundtrip.go#L144) btw but a lot simpler.

modules/frontend/frontend.go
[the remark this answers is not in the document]
annanay25: There's also the option of adding search support to the CLI (that way we could only support protobufs)

But agree that having json is nice

> The difference between NewTracesTripperware and NewSearchTripperware is that search can return multiple data structs, so we would have to check what path we are on to determine which struct to unmarshal the data into.

Ah I see. In that case it might even be better to support only json :)

modules/querier/querier.go
```diff
@@ -246,3 +252,168 @@ func (q *Querier) forGivenIngesters(ctx context.Context, replicationSet ring.Rep
 
 	return responses, err
 }
+
+func (q *Querier) Search(ctx context.Context, req *tempopb.SearchRequest) (*tempopb.SearchResponse, error) {
+	//userID, err := user.ExtractOrgID(ctx)
+	//if err != nil {
+	//	return nil, errors.Wrap(err, "error extracting org id in Querier.FindTraceByID")
+	//}
```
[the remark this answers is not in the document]
kvrhdn: I think this was copied from `Querier.FindTraceByID`. Since search doesn't need this, we probably commented it out to avoid compilation errors.
But now I'm wondering why we don't need `userID` in search 🤔

modules/frontend/frontend.go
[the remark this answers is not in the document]
annanay25: For the traces path we only moved to proto because of the marshalling/unmarshalling costs. But those might be small in the search path.

modules/querier/querier.go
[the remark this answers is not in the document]
annanay25: Yeah we will need userID in search as well.

modules/frontend/frontend.go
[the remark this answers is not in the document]
kvrhdn: Only json would be the path of least resistance, it seems like a good tradeoff to leave it as is for now and add protobuf support later on.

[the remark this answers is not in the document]
annanay25: 👍  agree. Let's drop the TODO in the comment. We can revisit it later if we see performance issues.

modules/querier/querier.go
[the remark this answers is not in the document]
kvrhdn: Nah, we don't need this since we only query ingesters for now. The ingester can extract the userID out of `ctx` by itself.

modules/frontend/frontend.go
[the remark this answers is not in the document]
annanay25: Per offline discussion, we agreed that the "default case" should actually never be called. I will be adding a warning log & a comment here.

modules/querier/querier.go
```diff
@@ -227,7 +227,7 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
 }
 
 // forGivenIngesters runs f, in parallel, for given ingesters
-func (q *Querier) forGivenIngesters(ctx context.Context, replicationSet ring.ReplicationSet, f func(client tempopb.QuerierClient) (*tempopb.TraceByIDResponse, error)) ([]responseFromIngesters, error) {
+func (q *Querier) forGivenIngesters(ctx context.Context, replicationSet ring.ReplicationSet, f func(client tempopb.QuerierClient) (interface{}, error)) ([]responseFromIngesters, error) {
```
annanay25, line 230: Modified `forGivenIngesters` to make it generic, @kvrhdn please review the latest commit!

modules/querier/querier.go
```diff
@@ -254,6 +254,11 @@ func (q *Querier) forGivenIngesters(ctx context.Context, replicationSet ring.Rep
 }
 
 func (q *Querier) Search(ctx context.Context, req *tempopb.SearchRequest) (*tempopb.SearchResponse, error) {
+	_, err := user.ExtractOrgID(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "error extracting org id in Querier.Search")
+	}
```
annanay25, line 260: re-added this because we may want to check orgID at the queriers before firing search queries for all ingesters.

modules/querier/querier.go
kvrhdn: Looks good to me, good idea to combine all of them :+1:

vikrambe:
@annanay25 Interesting work. It would be good to add config to accept a list of tags that need to be added to flatbuffer. I have not dug deep into flatbuffer, do we have plans to support wildcard search

modules/ingester/instance.go
```diff
@@ -66,6 +67,11 @@ type instance struct {
 	completingBlocks []*wal.AppendBlock
 	completeBlocks   []*wal.LocalBlock
 
+	searchHeadBlock      *searchStreamingBlockEntry
+	searchAppendBlocks   map[*wal.AppendBlock]*searchStreamingBlockEntry
```
joe-elliott, line 71: is it the long term plan to maintain a set of search blocks separate from existing tempo blocks or to pull this functionality together?

modules/ingester/instance.go
```diff
@@ -239,16 +262,39 @@ func (i *instance) CompleteBlock(blockID uuid.UUID) error {
...
 	}
 
+	// Search data (optional)
 	i.blocksMtx.Lock()
-	i.completeBlocks = append(i.completeBlocks, ingesterBlock)
+	oldSearch := i.searchAppendBlocks[completingBlock]
 	i.blocksMtx.Unlock()
 
+	var newSearch search.SearchBlock
+	if oldSearch != nil {
```
joe-elliott, line 271: seems weird that oldSearch would be nil. should that ever be true? is this an error?

modules/ingester/instance.go
```diff
@@ -81,9 +87,22 @@ type instance struct {
 	hash hash.Hash32
 }
 
+type searchStreamingBlockEntry struct {
+	b   *search.StreamingSearchBlock
```
joe-elliott, line 91: do StreamingSearchBlock and SearchBlock follow the same pattern as AppendBlock and BackendBlock?

the first is meant to just append data to as it comes in and the second is for storing in some kind of completed format ready for upload? maybe i just need to go read the block code :)

modules/ingester/instance_search.go
```diff
@@ -0,0 +1,167 @@
...
+			err := e.b.Search(ctx, p, sr)
+			if err != nil {
+				fmt.Println("error searching local block", b.BlockMeta().BlockID.String(), err)
+			}
+		}(b, e)
+	}
+}
+
+func dedupeResults(results []*tempopb.TraceSearchMetadata) []*tempopb.TraceSearchMetadata {
+	for i := range results {
```
joe-elliott, line 141: how large do these results slices get? should we create a map of existing trace ids instead of double looping?

modules/ingester/instance.go
mdisibio: TBD.  TempoDB is historically an agnostic KV store and integrating search would change that.  It was considered, but starting out as separate seems the best either way.

modules/ingester/instance.go
mdisibio: Yes it will be nil for any blocks replayed/reloaded on startup that didn't have search data. This can be true for WAL or locally completed blocks from a previous version, or the same version in this case as search is intentionally not reloaded on startup.

modules/ingester/instance_search.go
mdisibio: We don't have a good feel for what is a reasonable number of search results yet.  20 feels too low, but 1000 feels like too many.  In either case a map is fine, will fix.

modules/ingester/trace.go
```diff
@@ -42,5 +51,18 @@ func (t *trace) Push(_ context.Context, trace []byte) error {
 
 	t.traceBytes.Traces = append(t.traceBytes.Traces, trace)
 
+	if searchData != nil {
+		searchDataSize := len(searchData)
+		if t.maxSearchBytes != 0 { // disable limit if set to 0
+			if t.currentSearchBytes+searchDataSize > t.maxSearchBytes {
+				// todo: info level since we are not expecting this limit to be hit, but calibrate accordingly in the future
+				level.Info(cortex_util.Logger).Log("msg", "size of search data exceeded max search bytes limit", "maxSearchBytes", t.maxSearchBytes)
```
joe-elliott, line 59: we will need to metric this so we can track it ourselves as well as eventually expose it to people in Grafana Cloud

modules/search/backend_search_block.go
```diff
@@ -0,0 +1,187 @@
+package search
```
joe-elliott, line 1: i would like to find a better place to put this code. all other modules are controlled through modules.go and are managed via the Cortex service manager.

modules/search/pipeline.go
```diff
@@ -0,0 +1,88 @@
...
+	"strings"
+	"time"
+
+	"github.com/grafana/tempo/pkg/tempofb"
+	"github.com/grafana/tempo/pkg/tempopb"
+)
+
+type tracefilter func(header *tempofb.SearchData) (matches bool)
+
+type Pipeline struct {
```
joe-elliott, line 13: nit: name

Predicate? Matcher? ComparerThing?

modules/search/search_block_meta.go
```diff
@@ -0,0 +1,51 @@
...
+	"bytes"
+	"context"
+	"encoding/json"
+
+	"github.com/google/uuid"
+	tempo_io "github.com/grafana/tempo/pkg/io"
+	"github.com/grafana/tempo/tempodb/backend"
+)
+
+type SearchBlockMeta struct {
```
joe-elliott: i agree with the linter on this one. `search.BlockMeta` would be better.

it does feel like this and the traditional blocks are diverging. We need some internal discussion about what the long term vision is here. I don't mind there being a separate set of "search blocks" if that's what makes sense, but we just need to discuss what data is stored reduce storage costs.

pkg/util/http.go
```diff
@@ -66,3 +66,11 @@ func HexStringToTraceID(id string) ([]byte, error) {
 
 	return byteID, nil
 }
+
+// TraceIDToHexString converts a trace ID to its string representation and removes any leading zeros.
```
joe-elliott, line 70: any particular reason to trim 0s? We just use `hex.EncodeToString()` throughout the codebase to go from bytes => string for trace ids

pkg/util/timers.go
```diff
@@ -0,0 +1,29 @@
+package util
+
+import (
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var metricCompactionSeconds = promauto.NewCounterVec(prometheus.CounterOpts{
```
joe-elliott, line 10: can't find any references to this. test code?

modules/search/backend_search_block_writer.go
```diff
@@ -0,0 +1,121 @@
...
+		pageBuf:  make([]byte, 0, 1024*1024),
+		finalBuf: finalBuf,
+		builder:  flatbuffers.NewBuilder(1024),
+		dw:       dw,
+	}, nil
+}
+
+// Write the data to the flatbuffer builder. Input must be a SearchDataMutable. Returns
+// the number of bytes written, which is determined from the current object in the builder.
+func (w *backendSearchBlockWriter) Write(ctx context.Context, ID common.ID, i interface{}) (int, error) {
```
joe-elliott, line 57: Nice. This is a clever integration of the existing Writer interfaces and the flatbuffer paging.

I wonder if there's something we can do to reduce the complexity of all these Writers/Readers. Can we make the existing things generic enough to support this case as well as the current one? or would we just make everything more complex?

Even if we never merge this code with the existing set of writers/readers I think we need to come up with a code organization structure that works going forward. Should we push all this functionality into tempodb? I honestly don't know. Let's discuss.

tempodb/encoding/v0/data_reader.go
```diff
@@ -38,16 +38,26 @@ func (r *dataReader) Read(ctx context.Context, records []common.Record, buffer [
 		length += record.Length
 	}
 
-	buffer = make([]byte, length)
+	// Reset/resize buffer
+	if cap(buffer) < int(length) {
```
joe-elliott, line 42: eep, was this just a bug? we weren't correctly reusing the buffer?

tempodb/encoding/v0/data_reader.go
```diff
@@ -38,16 +38,26 @@ func (r *dataReader) Read(ctx context.Context, records []common.Record, buffer [
...
+	buffer = buffer[:length]
+
 	_, err := r.r.ReadAt(ctx, buffer, int64(start))
 	if err != nil {
 		return nil, nil, err
 	}
 
-	slicePages := make([][]byte, 0, len(records))
+	//Reset/resize buffer
+	if cap(pagesBuffer) < len(records) {
```
joe-elliott, line 53: does pagesBuffer have any performance impact? the "real" cost should be allocating the underlying bytes buffer.

pagesBuffer would just prevent us from reallocating a few slice structs?

modules/search/pipeline.go
mdisibio: Naming was inspired by Loki:  https://github.com/grafana/loki/blob/main/pkg/logql/log/pipeline.go  Might be worth considering as more complex search capabilities are added.

modules/search/backend_search_block_writer.go
mdisibio: > Can we make the existing things generic enough to support this case as well as the current one? or would we just make everything more complex?

Right, was considering if AppenderBuffered should be a byte-oriented shim around the new generic one (just one small part). Agree, let's discuss.

modules/search/backend_search_block.go
mdisibio: I guess we need to determine if this belongs in tempodb or not.  If not, /pkg/search might be ok.

tempodb/encoding/v0/data_reader.go
mdisibio: Right, the main savings are already present in `buffer`.  However, in benchmarks there is about a 3% performance increase in BackendSearchBlock.Search when reusing the pagesBuffer too. Thoughts?

joe-elliott: Surprised at the 3%. Under normal operation reusing the buffer should save MBs of allocation while the slices are only a few bytes. My thought was it would be negligible and not worth the code complexity, but if it's really makes iterating through a block 3% faster than i'm ok with it.

mdisibio: To clarify:  reusing `buffer` is ~30% improvement.   Reusing `pagesBuffer` is an additional ~3%.

pkg/tempofb/tempo.fbs
```diff
@@ -0,0 +1,24 @@
...
+}
+
+table SearchData {
+    id : string; // Converted to []byte
+    tags : [KeyValues];
+    start_time_unix_nano: uint64;
+    end_time_unix_nano: uint64;
+}
+
+table BatchSearchData {
```
annanay25, line 15: A little confused by the naming, should we make this `BlockSearchData`?

mdisibio: A batch is actually a page.  How about BatchSearchData -> SearchPage, SearchData -> SearchEntry ?

modules/search/backend_search_block.go
mdisibio: Discussed offline and decided to move it to /tempodb/search.

modules/search/backend_search_block_writer.go
mdisibio: Tried out making AppenderBuffered a byte-oriented shim and it didn't seem worth the additional complexity. Will leave the implementations separate for now.

pkg/util/http.go
mdisibio: @kvrhdn Thoughts?

pkg/tempofb/tempo.fbs
annanay25: Yup these two renames make sense to me.

tempodb/search/backend_search_block_writer.go
```diff
@@ -0,0 +1,101 @@
...
+	finalBuf *bytes.Buffer
+	dw       common.DataWriter
+}
+
+var _ common.DataWriterGeneric = (*backendSearchBlockWriter)(nil)
+
+func newBackendSearchBlockWriter(blockID uuid.UUID, tenantID string, w backend.RawWriter, enc backend.Encoding) (*backendSearchBlockWriter, error) {
+	finalBuf := &bytes.Buffer{}
+
+	dw, err := encoding.LatestEncoding().NewDataWriter(finalBuf, enc)
```
annanay25, line 36: I feel like this is "risky" because the latest encoding is a moving target, should we pin this to v2? Similarly for the index writers

modules/ingester/trace.go
```diff
@@ -17,20 +30,26 @@ type trace struct {
...
+	searchData         [][]byte
+	maxSearchBytes     int
+	currentSearchBytes int
 }
 
-func newTrace(maxBytes int, traceID []byte) *trace {
+func newTrace(traceID []byte, maxBytes int, maxSearchBytes int) *trace {
 	return &trace{
 		traceBytes: &tempopb.TraceBytes{
 			Traces: make([][]byte, 0, 10), // 10 for luck
```
annanay25, line 43: should we preallocate `searchData` for luck as well? :)

pkg/util/http.go
mdisibio: Just adding a thought:  It might be to trim the leading zeroes from a 8-byte trace ID stored in a 16-byte slice, to make the trace ID simpler.

kvrhdn: I added this function because it mirrors nicely with `HexStringToTraceID`. It encapsulates how we stringify trace IDs which is nice for people newer to the codebase. It's not immediately obvious that `hex.EncodeToString()` is the best way to print a trace ID so they'll have to search around for other places where we do this (at least that's what I had to do).
Even better would be to have a `TraceID` type with some associated functions to convert it (I think OTel has this), though I'm not sure if this is worth it as this will add some boilerplate.

Regarding trim 0s: I added this to avoid seeing a ton of zeroes when you are working with 64-bit trace IDs. I believe this is one of the first places where we show trace IDs as output to users, so I thought this would be a bit friendlier. Though we could also argue Grafana should do this for us and not bother.
The current implementation is a bit rough though... a trace ID `0` will be converted into an empty string 🙃 A better implementation would be to attempt to detect whether our trace ID is 64-bit and if so, only encode the last 8 bytes.

joe-elliott: +1 to @kvrhdn
Agree with all points made.

modules/ingester/trace.go
mdisibio: Luck is important :)  But in this case search data is optional, and the ingester may not be receiving it from the distributor.  I think it would be good to pool resources better here in the future. The TraceBytes struct functionally makes sense (alloc up front, return to pool when flushing trace), but is too specific-named.

joe-elliott (approved):
EL GEE TEE EM

modules/ingester/trace.go
annanay25: Yeah agree. Let's take these optimisations up in a later PR

annanay25 (approved):
LGTM 🚀

mdisibio:
We determined that this also fixed #216

PR #810, opened 2021-07-13
Support returning compressed responses from query-frontend

kvrhdn:
**What this PR does**:
The query-frontend will now gzip compress its response if the client requests this.

You can test this with curl by setting `--compressed`.

_But... I thought this was already supported?_

Despite what this config option might make you believe...

https://github.com/grafana/tempo/blob/026cff3910becf3ff9855f769149ce753cd2805f/modules/frontend/config.go#L16

...Tempo does not compress responses yet

While updating Cortex to v1.9 (#808) I noticed this flag has been deprecated and removed. When I investigated how this was being used, I noticed it isn't part of the code Tempo uses.
https://github.com/cortexproject/cortex/blob/4afaa357469fb37fd92cb1e2a0c1d10cdddc5ecd/pkg/cortex/modules.go#L557-L559

### DIFF
--- a/cmd/tempo/app/modules.go
+++ b/cmd/tempo/app/modules.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"path"
 
+	"github.com/NYTimes/gziphandler"
 	"github.com/cortexproject/cortex/pkg/cortex"
 	cortex_frontend "github.com/cortexproject/cortex/pkg/frontend"
 	cortex_transport "github.com/cortexproject/cortex/pkg/frontend/transport"
@@ -185,6 +186,8 @@ func (t *App) initQueryFrontend() (services.Service, error) {
 	tracesHandler := middleware.Merge(
 		t.httpAuthMiddleware,
 	).Wrap(cortexHandler)
+
+	tracesHandler = gziphandler.GzipHandler(tracesHandler)
 
 	// register grpc server for queriers to connect to
 	cortex_frontend_v1pb.RegisterFrontendServer(t.Server.GRPC, t.frontend)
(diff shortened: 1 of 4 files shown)


kvrhdn:
So this works when I test it manually, but it doesn't in the e2e test. The HTTP server created by the Cortex e2e framework don't suppor this or filter it out.

joe-elliott:
I think the correct place to put functionality like this might be in weaveworks/common. @bboreham do you have any insight on this?

bboreham:
No strong feelings. If you put it in `Server` on by default it will help people who didn’t realise they needed it; it will be worse for people who know for sure they don’t want it (maybe they serve payloads that are always short, or always already compressed). 

If it’s not on by default it won’t help much. 

I note that the Go standard library http client will _uncompress_ by default.

joe-elliott:
@kvrhdn it's fine with me to add it to just Tempo, but it does feel weird to just explicitly add this to the query frontend. I suppose all internal Tempo communication is through GRPC so maybe it is ok?

Can you investigate adding this when initializing the server so it impacts all components? Perhaps with one of these options:

https://github.com/weaveworks/common/blob/master/server/server.go#L79-L80

kvrhdn:
Sure, I'll see what is possible 🙂 

fyi, Cortex moved this code into their api package and made it a helper function: https://github.com/cortexproject/cortex/blob/20f8931014bc9f3271380884f8eadee8de83ae96/pkg/api/api.go#L141-L143
That's also a possible solution.

joe-elliott:
The Cortex route probably has a lot of thought behind it. Perhaps some endpoints are better off not having this middleware?

Fine with either route. Pitch me something!

PR #812, opened 2021-07-14
Emit traces for ingester flush operations

bboreham:
**What this PR does**:
Create a new span for each flush, log the trace ID, and any error that comes back.

Passing a context in to `handleFlush()` seems more natural than having it assume it is only called in background context.

**Checklist**
- NA Tests updated
- NA Documentation added
- [x] `CHANGELOG.md` updated

### DIFF
--- a/modules/ingester/flush.go
+++ b/modules/ingester/flush.go
@@ -8,8 +8,13 @@ import (
 	"strconv"
 	"time"
 
+	gklog "github.com/go-kit/kit/log"
 	"github.com/google/uuid"
+	ot "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
+	otlog "github.com/opentracing/opentracing-go/log"
 	"github.com/pkg/errors"
+	"github.com/uber/jaeger-client-go"
 
 	"github.com/cortexproject/cortex/pkg/util/log"
 	"github.com/cortexproject/cortex/pkg/util/services"
@@ -187,7 +192,7 @@ func (i *Ingester) flushLoop(j int) {
 		if op.kind == opKindComplete {
 			retry, err = i.handleComplete(op)
 		} else {
-			retry, err = i.handleFlush(op.userID, op.blockID)
+			retry, err = i.handleFlush(context.Background(), op.userID, op.blockID)
 		}
 
 		if err != nil {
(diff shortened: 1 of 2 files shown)


modules/ingester/flush.go
```diff
@@ -261,8 +266,25 @@ func (i *Ingester) handleComplete(op *flushOp) (retry bool, err error) {
 	return false, nil
 }
 
-func (i *Ingester) handleFlush(userID string, blockID uuid.UUID) (retry bool, err error) {
-	level.Info(log.Logger).Log("msg", "flushing block", "userid", userID, "block", blockID.String())
+// withSpan adds traceID to a logger, if span is sampled
+// TODO: move into some central trace/log package
```
annanay25, line 270: Cortex has this - https://github.com/cortexproject/cortex/blob/master/pkg/util/spanlogger/spanlogger.go. Might be worth just using the same

modules/ingester/flush.go
```diff
@@ -261,8 +266,25 @@ func (i *Ingester) handleComplete(op *flushOp) (retry bool, err error) {
...
+		return logger
+	}
+
+	return gklog.With(logger, "traceID", sctx.TraceID().String())
+}
+
+func (i *Ingester) handleFlush(ctx context.Context, userID string, blockID uuid.UUID) (retry bool, err error) {
+	sp := ot.GlobalTracer().StartSpan("flush", ot.Tag{Key: "organization", Value: userID}, ot.Tag{Key: "blockID", Value: blockID.String()})
+	defer sp.Finish()
+	ctx = ot.ContextWithSpan(ctx, sp)
```
mapno, line 286: Why is attaching the span to the context necessary?

mapno (approved):
LGTM, just left a minor question

modules/ingester/flush.go
```diff
@@ -283,6 +305,8 @@ func (i *Ingester) handleFlush(userID string, blockID uuid.UUID) (retry bool, er
 		metricFlushDuration.Observe(time.Since(start).Seconds())
 		metricFlushSize.Observe(float64(block.BlockMeta().Size))
 		if err != nil {
+			ext.Error.Set(sp, true)
+			sp.LogFields(otlog.Error(err))
```
annanay25, line 309: should we put this under a `defer()` block and check for err?

annanay25 (approved):

modules/ingester/flush.go
bboreham: So that any tracing info emitted from things we call will be attached to this span.
However I realise I should just call `StartSpanFromContext`.

modules/ingester/flush.go
bboreham: That requires extreme care with scoping of `err`; it's not something I would naturally do.

modules/ingester/flush.go
mapno: I see, thanks

bboreham:
Re-pushed with simpler span creation and changelog.

I realise this could be a common util with Cortex; I specifically dislike the current implementation in Cortex so will need to fix that before making them common.

Issue #761, opened 2021-06-14
Tempo throws 500s when rolling queriers


annanay25:
**Describe the bug**

Tempo Query Frontend throws 500s when the querier deployment is rolled. 

**To Reproduce**
1. Start Tempo (SHA or version): `latest`
2. Perform Operations (Read/Write/Others): 

```
kubectl rollout restart deployment querier
```
and then watch the Query Frontend RED metrics.

**Expected behavior**

No errors on the query frontend. I would expect the querier to bail only after completing any in-flight requests.

**Environment:**
 - Infrastructure: [e.g., Kubernetes, bare-metal, laptop]: Kubernetes
 - Deployment tool: [e.g., helm, jsonnet]: 

**Additional Context**

Without digging very deep, these are some guesses of where things could be going wrong:
- misconfigured defaults on the querier worker
- Stopping the pool in parallel with the querier worker (could be that the pool needs to be shut down after the worker) https://github.com/grafana/tempo/blob/3b9f309a9e9a1e0e943ee8b7a3d8c4f458962445/modules/querier/querier.go#L92

joe-elliott:
Queriers can also incorrectly 404 on startup. This appears to be b/c they will answer queries before they have completely polled the blocklist.

kvrhdn:
Okay, I'm seeing a couple of issues related to queriers stopping and starting:

#### A new querier receives a request while its ring is still empty

```
2021-07-09T12:03:00.080690887Z stderr F level=warn ts=2021-07-09T12:03:00.078940012Z caller=querier.go:105 msg=Querier.starting
2021-07-09T12:03:00.080693554Z stderr F level=warn ts=2021-07-09T12:03:00.07954222Z caller=querier.go:107 msg="Querier.starting - exiting"
2021-07-09T12:03:00.080695137Z stderr F level=warn ts=2021-07-09T12:03:00.079652137Z caller=querier.go:120 msg=Querier.running
[...]

PR #814, opened 2021-07-14
Add retry middleware in query-frontend

kvrhdn:
**What this PR does**:
Requests from the query-frontend to the querier might fail when a querier is starting or stopping. Simply retrying this request usually solves this.

This PR adds a retry middleware which will relaunch requests (up to 5 times by default) if the request failed (http 500 or error).

Related to #761 

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/frontend/retry.go
+++ b/modules/frontend/retry.go
@@ -52,6 +52,5 @@ func (r retryWare) Do(req *http.Request) (*http.Response, error) {
 		}
 
 		span.LogFields(ot_log.String("msg", "error processing request"), ot_log.Int("try", triesLeft), ot_log.Error(err))
-		continue
 	}
 }
--- a/modules/frontend/retry.go
+++ b/modules/frontend/retry.go
@@ -1,0 +1,57 @@
+package frontend
+
+import (
+	"net/http"
+
+	"github.com/go-kit/kit/log"
+	"github.com/opentracing/opentracing-go"
+	ot_log "github.com/opentracing/opentracing-go/log"
+)
+
+func RetryWare(maxRetries int, logger log.Logger) Middleware {
+	return MiddlewareFunc(func(next Handler) Handler {
+		return retryWare{
+			next:       next,
+			logger:     logger,
+			maxRetries: maxRetries,
+		}
+	})
+}
+
+type retryWare struct {
+	next       Handler
+	logger     log.Logger
+	maxRetries int
+}
+
+// Do implements Handler
+func (r retryWare) Do(req *http.Request) (*http.Response, error) {
+	ctx := req.Context()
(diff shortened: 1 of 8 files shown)


kvrhdn:
This PR also fixes context propagation in the other middleware functions.

Before and after:

Screenshot 2021-07-14 at 18 02 30

Screenshot 2021-07-14 at 18 03 17

mapno (commented):
Changes look good.

My concern is if this change could lead to hammering down a slow or overloaded querier, making the issue worse. It might be worth considering using a backoff for the retries or other mitigation alternatives.

kvrhdn:
> My concern is if this change could lead to hammering down a slow or overloaded querier, making the issue worse. It might be worth considering using a backoff for the retries or other mitigation alternatives.

In Cortex the querier pulls work from the queue in query-frontend. The amount of queries that are processed concurrently by a querier is limited by [MaxConcurrentQueries](https://github.com/grafana/tempo/blob/main/modules/querier/config.go#L16).
Retrying a request will put this request on the queue again so another querier can pick this up.

BTW, this PR is pretty much the same code as Cortex uses in their queryrange processor which shards requests by day.
https://github.com/cortexproject/cortex/blob/master/pkg/querier/queryrange/retry.go

modules/frontend/retry.go
```diff
@@ -0,0 +1,57 @@
...
+		if triesLeft == 0 {
+			return resp, err
+		}
+
+		if err == nil && resp.StatusCode/100 != 5 {
+			return resp, nil
+		}
+
+		span.LogFields(ot_log.String("msg", "error processing request"), ot_log.Int("try", triesLeft), ot_log.Error(err))
+		continue
```
joe-elliott, line 55: continue?

does this do anything?

joe-elliott (approved):
one nit. nice addition.

modules/frontend/retry.go
kvrhdn: Well, it continues the for loop 😛 

I've removed it, I refactored my loop a couple of times for clarity.

mapno:
> > My concern is if this change could lead to hammering down a slow or overloaded querier, making the issue worse. It might be worth considering using a backoff for the retries or other mitigation alternatives.
> 
> In Cortex the querier pulls work from the queue in query-frontend. The amount of queries that are processed concurrently by a querier is limited by [MaxConcurrentQueries](https://github.com/grafana/tempo/blob/main/modules/querier/config.go#L16).
> Retrying a request will put this request on the queue again so another querier can pick this up.
> 
> BTW, this PR is pretty much the same code as Cortex uses in their queryrange processor which shards requests by day.
> https://github.com/cortexproject/cortex/blob/master/pkg/querier/queryrange/retry.go

Thanks for the explanation

PR #816, opened 2021-07-15
Remove -o pipefail from image-tag script

### LINKED ISSUE #815: tools/image-tag: "5: set: Illegal option -o pipefail"
**Describe the bug**
`image-tag` is using a bash-ism but requests to run sh.

**To Reproduce**
```
$ ./tools/image-tag 
./tools/image-tag: 5: set: Illegal option -o pipefail
```

**Expected behavior**
No diagnostic

**Environment:**
Ubuntu 20.04.2 LTS

**Additional Context**
I get this error simply by cloning the repo and running `make`.

bboreham:
**What this PR does**:
Remove -o pipefail from image-tag script

It is a bash-ism so not supported in some environments

Script doesn't have any pipelines where we could get a failure protected by pipefail

Fixes #815 

**Checklist**
- NA Tests updated
- NA Documentation added
- NA `CHANGELOG.md` updated - this is not a user-visible change

### DIFF
--- a/tools/image-tag
+++ b/tools/image-tag
@@ -2,7 +2,6 @@
 
 set -o errexit
 set -o nounset
-set -o pipefail
 
 SHA="$(git rev-parse --short HEAD)"
 


dgzlopes (approved):

annanay25:
I think the right thing to do here is to use bash in the script. I don't remember why I swapped over to sh, but here is the loki image tag script for reference - https://github.com/grafana/loki/blob/main/tools/image-tag

To test that the changes work in drone, you can edit the triggers defined here - https://github.com/grafana/tempo/blob/2ad27e7cbd3f265ed93e66382d25026cdcb2b0e0/.drone/drone.yml#L188-L191 - to include this PR (so it will trigger on every commit to this PR) and then remove the trigger before merging.

bboreham:
Perhaps a case for a common repository of tools, so we don't have the opportunity to edit the same script in multiple places?
This one originated in https://github.com/weaveworks/build-tools/, but is significantly different here.

annanay25:
TIL :)

And yes 100% agree.

joe-elliott:
Thanks! This has been annoying me forever but I never tried to track it down.

PR #818, opened 2021-07-15
Add shouldCache param to backend interface read/write methods

annanay25:
**What this PR does**:
This PR adds a new `shouldCache` parameter to the Backend interface on the Read/Write functions.

Pros:
- Cleaner interface, follow up of #804 
- Ability to cache based on some custom parameters. Will be useful in #805 

Cons:
- Lots of `ioutil.NopCloser(bytes.NewReader(buffer))` usage while translating byte slices <> io.ReadCloser 

_Note: I'm still working on tests and CI, opened the PR for early review of the general direction._

Fixes #na!

**Checklist**
- [x] Tests updated
- NA Documentation added

### DIFF
--- a/pkg/io/read.go
+++ b/pkg/io/read.go
@@ -1,17 +1,13 @@
 package io
 
 import (
-	"fmt"
 	"io"
 )
 
 // ReadAllWithEstimate is a fork of https://go.googlesource.com/go/+/go1.16.3/src/io/io.go#626
 //  with a starting buffer size. if none is provided it uses the existing default of 512
 func ReadAllWithEstimate(r io.Reader, estimatedBytes int64) ([]byte, error) {
-	if estimatedBytes < 0 {
-		return nil, fmt.Errorf("negative value received for estimatedBytes")
-	}
-	if estimatedBytes == 0 {
+	if estimatedBytes <= 0 {
 		estimatedBytes = 512
 	}
 
--- a/pkg/io/read.go
+++ b/pkg/io/read.go
@@ -1,10 +1,16 @@
 package io
 
-import "io"
+import (
+	"fmt"
+	"io"
+)
 
 // ReadAllWithEstimate is a fork of https://go.googlesource.com/go/+/go1.16.3/src/io/io.go#626
 //  with a starting buffer size. if none is provided it uses the existing default of 512
 func ReadAllWithEstimate(r io.Reader, estimatedBytes int64) ([]byte, error) {
+	if estimatedBytes < 0 {
+		return nil, fmt.Errorf("negative value received for estimatedBytes")
+	}
 	if estimatedBytes == 0 {
 		estimatedBytes = 512
 	}
--- a/tempodb/backend/cache/cache.go
+++ b/tempodb/backend/cache/cache.go
@@ -1,11 +1,15 @@
 package cache
 
 import (
+	"bytes"
 	"context"
 	"io"
+	"io/ioutil"
 	"strings"
 
 	cortex_cache "github.com/cortexproject/cortex/pkg/chunk/cache"
+
+	tempo_io "github.com/grafana/tempo/pkg/io"
 	"github.com/grafana/tempo/tempodb/backend"
 )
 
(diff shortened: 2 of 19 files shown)


joe-elliott:
> Lots of ioutil.NopCloser(bytes.NewReader(buffer)) usage while translating byte slices <> io.ReadCloser

These could be consolidated in the backend.Reader interface while still simplifying the backend.RawReader interface. Similar I suppose for whatever was done for the Writer side.

tempodb/backend/raw.go
```diff
@@ -22,9 +26,7 @@ type KeyPath []string
 // RawWriter is a collection of methods to write data to tempodb backends
 type RawWriter interface {
 	// Write is for in memory data.  It is expected that this data will be cached.
-	Write(ctx context.Context, name string, keypath KeyPath, buffer []byte) error
-	// StreamWriter is for larger data payloads streamed through an io.Reader.  It is expected this will _not_ be cached.
-	StreamWriter(ctx context.Context, name string, keypath KeyPath, data io.Reader, size int64) error
+	Write(ctx context.Context, name string, keypath KeyPath, data io.Reader, size int64) error
```
annanay25, line 29: This is the most important file to review in this change, all the other changes are about adapting the new interface definitions.

annanay25:
OK. After some back and forth, I have finally cleaned everything up and I feel this is in a good place. Specifically I wanted to:
- Clean and simplify the backend.RawWriter interface (since its implemented by multiple backends - GCS/S3/Azure/Cache)
- Expose shouldCache from the backend package so we can control that better. This will be better utilised in an upcoming PR.

tempodb/backend/cache/cache.go
```diff
@@ -25,57 +29,57 @@ func NewCache(nextReader backend.RawReader, nextWriter backend.RawWriter, cache
...
-// Read implements backend.Reader
-func (r *readerWriter) Read(ctx context.Context, name string, keypath backend.KeyPath) ([]byte, error) {
+// Read implements backend.RawReader
+func (r *readerWriter) Read(ctx context.Context, name string, keypath backend.KeyPath, shouldCache bool) (io.ReadCloser, int64, error) {
 	var k string
-	if shouldCache(name) {
+	if shouldCache {
 		k = key(keypath, name)
 		found, vals, _ := r.cache.Fetch(ctx, []string{k})
 		if len(found) > 0 {
```
mapno, line 43: If this is a partial hit, shouldn't it go to the storage to fetch the misses?

mapno (commented):
LGTM overall, nice changes

tempodb/backend/cache/cache.go
annanay25: The Cache interfaces are adopted from Cortex, but in Tempo there is no partial hit at the moment - its either a hit or a miss.

So if `len(found) == 0` the function will fetch object from the backend (line 48)

tempodb/backend/raw.go
```diff
@@ -1,19 +1,23 @@
...
 	"github.com/google/uuid"
+
+	tempo_io "github.com/grafana/tempo/pkg/io"
 )
 
 const (
 	MetaName          = "meta.json"
 	CompactedMetaName = "meta.compacted.json"
 	BlockIndexName    = "blockindex.json.gz"
+	BloomName         = "bloom"
```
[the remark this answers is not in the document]
annanay25: This should be removed. Will make the change.

tempodb/backend/raw.go
```diff
@@ -1,19 +1,23 @@
...
 	"io"
 	"path"
 
 	"github.com/google/uuid"
+
+	tempo_io "github.com/grafana/tempo/pkg/io"
 )
 
 const (
 	MetaName          = "meta.json"
```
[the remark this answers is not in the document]
annanay25: I think there's a reference in the local backend. Will make them internal if they are not used.

tempodb/backend/raw.go
```diff
@@ -178,3 +187,7 @@ func CompactedMetaFileName(blockID uuid.UUID, tenantID string) string {
 func RootPath(blockID uuid.UUID, tenantID string) string {
 	return path.Join(tenantID, blockID.String())
 }
+
+func ShouldCache(name string) bool {
+	return name != MetaName && name != CompactedMetaName && name != BlockIndexName
```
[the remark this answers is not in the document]
annanay25: I would like to expose `shouldCache bool` on the Reader/Writer interfaces already, I thought it would be neat to consolidate the caching logic in one place. But if its preferred to pass true/false I'll make that change.

tempodb/backend/readerat.go
```diff
@@ -41,7 +41,7 @@ func (b *backendReader) ReadAt(ctx context.Context, p []byte, off int64) (int, e
 
 // ReadAll implements ContextReader
 func (b *backendReader) ReadAll(ctx context.Context) ([]byte, error) {
-	return b.r.Read(ctx, b.name, b.meta.BlockID, b.meta.TenantID)
+	return b.r.Read(ctx, b.name, b.meta.BlockID, b.meta.TenantID, ShouldCache(b.name))
```
annanay25, line 44: @joe-elliott  fyi - ContextReader still needs `shouldCache()` unless we modify its initialisation function to 
```
// NewContextReader creates a ReaderAt for the given BlockMeta
func NewContextReader(meta *BlockMeta, name string, r Reader, shouldCache bool) ContextReader

```

annanay25: ^ made that change.

tempodb/backend/raw.go
```diff
@@ -134,7 +137,12 @@ func (r *reader) Blocks(ctx context.Context, tenantID string) ([]uuid.UUID, erro
 }
 
 func (r *reader) BlockMeta(ctx context.Context, blockID uuid.UUID, tenantID string) (*BlockMeta, error) {
-	bytes, err := r.r.Read(ctx, MetaName, KeyPathForBlock(blockID, tenantID))
+	reader, size, err := r.r.Read(ctx, MetaName, KeyPathForBlock(blockID, tenantID), false)
+	if err != nil {
+		return nil, err
+	}
+
+	bytes, err := tempo_io.ReadAllWithEstimate(reader, size)
```
joe-elliott, line 145: what happens if size is a negative value?  i think we'd panic.

can any of the backends return negative?

tempodb/backend/raw.go
```diff
@@ -134,7 +137,12 @@ func (r *reader) Blocks(ctx context.Context, tenantID string) ([]uuid.UUID, erro
 }
 
 func (r *reader) BlockMeta(ctx context.Context, blockID uuid.UUID, tenantID string) (*BlockMeta, error) {
-	bytes, err := r.r.Read(ctx, MetaName, KeyPathForBlock(blockID, tenantID))
+	reader, size, err := r.r.Read(ctx, MetaName, KeyPathForBlock(blockID, tenantID), false)
```
joe-elliott, line 140: close reader?

tempodb/backend/raw.go
```diff
@@ -95,12 +94,16 @@ func NewReader(r RawReader) Reader {
 	}
 }
 
-func (r *reader) Read(ctx context.Context, name string, blockID uuid.UUID, tenantID string) ([]byte, error) {
-	return r.r.Read(ctx, name, KeyPathForBlock(blockID, tenantID))
+func (r *reader) Read(ctx context.Context, name string, blockID uuid.UUID, tenantID string, shouldCache bool) ([]byte, error) {
+	objReader, size, err := r.r.Read(ctx, name, KeyPathForBlock(blockID, tenantID), shouldCache)
```
joe-elliott, line 98: close reader/same size comment as below.

tempodb/backend/raw.go
annanay25: 👍    handled in `ReadAllWithEstimate()`

pkg/io/read.go
```diff
@@ -1,10 +1,16 @@
...
+import (
+	"fmt"
+	"io"
+)
 
 // ReadAllWithEstimate is a fork of https://go.googlesource.com/go/+/go1.16.3/src/io/io.go#626
 //  with a starting buffer size. if none is provided it uses the existing default of 512
 func ReadAllWithEstimate(r io.Reader, estimatedBytes int64) ([]byte, error) {
+	if estimatedBytes < 0 {
+		return nil, fmt.Errorf("negative value received for estimatedBytes")
```
joe-elliott, line 12: might be better just to use the 512?

annanay25: Other than the local backend which can return a negative value from the `stat()` no other backend/cache would return a negative, I _think_ an error here might be fine. But I wouldn't mind changing this to a warning log + just using the default 512 bytes. Let me know!

mdisibio: +1 with using default instead of erroring, as the parameter is a performance hint not a strict requirement.

mdisibio (approved):

PR #828, opened 2021-07-20
Prepare CI for our internal release train

jvrplmlmn:
This PR lays the groundwork to adopt an internal weekly release cycle. This simply adds an additional trigger to CI so it can build and push artifacts on `r?` and `r??` branches, which are later consumed by our internal automation.

### DIFF
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -36,6 +36,8 @@ trigger:
   ref:
   - refs/heads/main
   - refs/tags/**
+  - refs/heads/r? # For weekly release
+  - refs/heads/r?? # For weekly release
 
 ---
 ##  AMD64  ##
@@ -90,6 +92,8 @@ trigger:
   ref:
   - refs/heads/main
   - refs/tags/v*
+  - refs/heads/r? # For weekly release
+  - refs/heads/r?? # For weekly release
 
 ---
 ##  ARM64  ##
@@ -144,6 +148,8 @@ trigger:
   ref:
   - refs/heads/main
   - refs/tags/v*
+  - refs/heads/r? # For weekly release
+  - refs/heads/r?? # For weekly release
 
 ---
 ##  MANIFEST  ##
@@ -189,6 +195,8 @@ trigger:
   ref:
   - refs/heads/main
   - refs/tags/v*
+  - refs/heads/r? # For weekly release
+  - refs/heads/r?? # For weekly release
 
 ---
 kind: secret
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -11,6 +11,7 @@ steps:
 - name: image-tag
   image: alpine/git
   commands:
+  - apk --update --no-cache add bash
   - git fetch origin --tags
   - echo $(./tools/image-tag) > .tags
 
@@ -52,6 +53,7 @@ steps:
 - name: image-tag
   image: alpine/git
   commands:
+  - apk --update --no-cache add bash
   - git fetch origin --tags
   - echo $(./tools/image-tag)-amd64 > .tags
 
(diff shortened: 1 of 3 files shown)


tools/image-tag
```diff
@@ -2,14 +2,20 @@
```
annanay25, line 2: nit: Can you set this to bash please? Refer #816

tools/image-tag
```diff
@@ -2,14 +2,20 @@
 
 set -o errexit
 set -o nounset
+set -o pipefail
 
-SHA="$(git rev-parse --short HEAD)"
+WIP=$(git diff --quiet || echo '-WIP')
```
annanay25, line 7: Just to double check, this will create images for every commit to the branch? Not sure if we need that, but if it helps standardise this script across all our projects let's do it!

annanay25 (approved):
LGTM, but I guess @kvrhdn can give the signoff on this one since he has been working with you!

tools/image-tag
kvrhdn: With this PR, the script is exactly the same as [Loki's image-tag](https://github.com/grafana/loki/blob/main/tools/image-tag) except for the `#!/usr/bin/env bash`. Seems reasonable to also use bash then 🙂

.drone/drone.yml
```diff
@@ -36,6 +36,8 @@ trigger:
   ref:
   - refs/heads/main
   - refs/tags/**
+  - refs/heads/r? # For weekly release
+  - refs/heads/r?? # For weekly release
```
kvrhdn, line 40: How does the `r?` and `r??` work and how is this different from `**`? I can't find this in their docs.

tools/image-tag
kvrhdn: I think `-WIP` will only be added if you run it locally. Drone will only run on new commits to the `main` and `r??` branches or new tags.

@jvrplmlmn am I correct that this is intended to add `-WIP` to your version if you build an image locally while you have changes?

tools/image-tag
```diff
@@ -2,14 +2,20 @@
...
+
+# If not a tag, use branch-hash else use tag
+TAG=$((git describe --exact-match 2> /dev/null || echo "") | sed 's/v//g')
 
-# If tag, use tag
-TAG=$( (git describe --exact-match 2> /dev/null || echo "") | sed 's/v//g')
 if [ -z "$TAG" ]
 then
-      echo ${SHA}
+      echo ${BRANCH}-${SHA}${WIP}
```
kvrhdn, line 18: This will change our docker tags from e.g. `99217c5e` to `main-99217c5e`. Should we add a CHANGELOG entry for this?

kvrhdn (approved):
LGTM 🙂  I have some minor questions.

tools/image-tag
jvrplmlmn: @kvrhdn you are correct 👍

.drone/drone.yml
jvrplmlmn: From https://docs.drone.io/pipeline/triggers/

> You can use wildcard matching in your triggers. Note that triggers use glob pattern matching, not regular expressions.

So it will match on `r1`, `r33`, etc. We could change this to `refs/heads/r*`, which will build every branch that starts with `r`.

tools/image-tag
jvrplmlmn: Addressed in 46a61fa

tools/image-tag
jvrplmlmn: Addressed in b314bc7

kvrhdn (approved):
Thanks for the edits 🙂  This is looking good, can someone with the necessary permission merge this for us? @annanay25 maybe?

PR #831, opened 2021-07-22
rename legacy datasource

the-it:
Operations removed a legacy cluster. The old name shouldn't be used anymore. Therefore we also renamed the data source.


joe-elliott (approved):
thx!

PR #834, opened 2021-07-23
Tenant Index

### LINKED ISSUE #575: Reduce pressure on the backend caused by blocklist polling
Currently compactors and queriers each individually repeatedly poll the backend to get up to date block lists.  

One component should regularly poll the backend and build an index for all other components to load.  This could be as simple as a compressed json file that contains all blocks.

### LINKED ISSUE #825: Queriers can 404 when searching blocklist after a rollout
**Describe the bug**
While rolling the queriers we are seeing a slight increase in queries 404'ing. This was detected by running vulture.

Theory: when the querier starts it needs some time to poll the blocklist. If the querier receives a query before this has happened, it will not find the query.

This issue is split off from #761, https://github.com/grafana/tempo/issues/761#issuecomment-863253132

**To Reproduce**
1. Run tempo and tempo-vulture (make sure they run long enough that blocks are getting flushed and vulture also queries 'old' traces)
2. Roll the querier deployment (`kubectl rollout restart deployment querier`)
3. Vulture will report an increased amount of 404s

**Expected behavior**
If the trace exists in the backend the querier should not 404 (at least if it is searching the right block ID space).

joe-elliott:
**What this PR does**:
Adds support for a tenant index with the following new parameters:
```
storage:
  trace:
    blocklist_poll_fallback: true
    blocklist_poll_tenant_index_builders: 2
```
By default all queriers and most compactors will attempt to download a tenant index located at `/<tenant id>/index.json.gzip` instead of repeatedly listing the bucket. This gzipped json file will contain a list of all metas and compacted metas for this tenant.

`blocklist_poll_fallback` controls whether or not the various components fall back to scanning the bucket if the index fails. 
`blocklist_poll_tenant_index_builders` is the _max_ number of compactors that attempt to build the index. it could be less depending on sharding.

This PR includes all changes necessary to do the above, as well as related metrics, alerts, and runbook entries.  Additionally #825 is fixed because the querier and compactors will now require a polling cycle to complete before they raise their readiness probes.

Fixes #575 
Fixes #825 

**Checklist**
- [x] Tests updated
- [x] Documentation added


kvrhdn (commented):
Running this locally, I can confirm it fixes #825. The PR in general seems fine, though I'm not super familiar with the compactor.

tempodb/blocklist/poller.go
```diff
@@ -85,13 +121,37 @@ func (p *Poller) Do() (PerTenant, PerTenantCompacted, error) {
...
+
+		metricTenantIndexErrors.WithLabelValues(tenantID).Inc()
+
+		// we had an error, do we bail?
+		if !p.cfg.PollFallback {
+			return nil, nil, err
+		}
+
+		// log error and keep on keeping on
+		level.Error(p.logger).Log("msg", "failed to pull bucket index for tenant. falling back to polling", "tenant", tenantID, "err", err)
```
annanay25, line 143: I think the very first time the tenant index is being created, this log line will be hit. The tenant index will not be generated, and all components will be required to poll like they always do.. Should this be an "error" level log?

tempodb/blocklist/poller.go
```diff
@@ -85,13 +121,37 @@ func (p *Poller) Do() (PerTenant, PerTenantCompacted, error) {
...
+		i, err := p.reader.TenantIndex(ctx, tenantID)
+		if err == nil {
+			metricTenantIndexAgeSeconds.WithLabelValues(tenantID).Set(float64(time.Since(i.CreatedAt) / time.Second))
+			level.Info(p.logger).Log("msg", "successfully pulled tenant index", "tenant", tenantID, "createdAt", i.CreatedAt, "metas", len(i.Meta), "compactedMetas", len(i.CompactedMeta))
+			return i.Meta, i.CompactedMeta, nil
+		}
+
+		metricTenantIndexErrors.WithLabelValues(tenantID).Inc()
+
+		// we had an error, do we bail?
```
annanay25, line 137: What happens if compactors are rolled? That could now impact the state of the blocklist. As a safety catch, should we also fallback to polling if the tenantIndex is older than `2 * BlocklistPoll`?

cmd/tempo/app/modules.go
```diff
@@ -146,8 +146,11 @@ func (t *App) initQuerier() (services.Service, error) {
 		}
 	}
 
+	// do not enable polling if this is the single binary. in that case the compactor will take care of polling
```
annanay25, line 149: Slightly confused by this. Enabling polling will enable tenant index creation as well. Should we always allow the compactors to create the tenant index? If the queriers also try to create the index they can overwrite the index created by the compactors?

annanay25 (commented):
Very well thought out change. Really like the tenant index creation idea. Added comments.

tempodb/blocklist/poller.go
```diff
@@ -154,9 +223,19 @@ func (p *Poller) pollBlock(ctx context.Context, tenantID string, blockID uuid.UU
...
 		return nil, nil, err
 	}
 
 	return blockMeta, compactedBlockMeta, nil
 }
+
+func (p *Poller) buildTenantIndex() bool {
+	for i := 0; i < p.cfg.TenantIndexBuilders; i++ {
+		job := jobPrefix + strconv.Itoa(i)
+		if p.sharder.Owns(job) {
```
mdisibio, line 235: This is more of a question on the intended behavior: it looks all index builders poll and write the index for all tenants continuously.  That seems ok from a redundancy standpoint (i.e. if one builder stops working, the system still has fresh data).  Sharding by tenant here seems beneficial for scaling up to many tenants, but an unhealthy compactor ring entry would also impact polling.

tempodb/blocklist/list.go
```diff
@@ -0,0 +1,163 @@
...
+	if tenantID == "" {
+		return
+	}
+
+	l.mtx.Lock()
+	defer l.mtx.Unlock()
+
+	l.updateInternal(tenantID, add, remove, compactedAdd)
+
+	// save off
```
mdisibio, line 111:
```suggestion
	// save off, they are retained for an additional polling cycle
```

modules/storage/config.go
```diff
@@ -23,15 +22,15 @@ type Config struct {
...
 // RegisterFlagsAndApplyDefaults registers the flags.
 func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet) {
 
+	cfg.Trace.BlocklistPollFallback = true
 	cfg.Trace.BlocklistPollConcurrency = tempodb.DefaultBlocklistPollConcurrency
+	cfg.Trace.BlocklistPollTenantIndexBuilders = tempodb.DefaultTenantIndexBuilders
 
 	f.StringVar(&cfg.Trace.Backend, util.PrefixConfig(prefix, "trace.backend"), "", "Trace backend (s3, azure, gcs, local)")
-	f.DurationVar(&cfg.Trace.BlocklistPoll, util.PrefixConfig(prefix, "trace.maintenance-cycle"), DefaultBlocklistPoll, "Period at which to run the maintenance cycle.")
+	f.DurationVar(&cfg.Trace.BlocklistPoll, util.PrefixConfig(prefix, "trace.maintenance-cycle"), tempodb.DefaultBlocklistPoll, "Period at which to run the maintenance cycle.")
```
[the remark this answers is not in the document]
joe-elliott: It is called `blocklist_poll` in the yaml. An unfortunate name mismatch. We should make the cli option match the yaml and mark it a breaking change (not in this PR).

modules/storage/config.go
```diff
@@ -23,15 +22,15 @@ type Config struct {
...
 }
 
-const DefaultBlocklistPoll = 5 * time.Minute
-
 // RegisterFlagsAndApplyDefaults registers the flags.
 func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet) {
 
+	cfg.Trace.BlocklistPollFallback = true
 	cfg.Trace.BlocklistPollConcurrency = tempodb.DefaultBlocklistPollConcurrency
+	cfg.Trace.BlocklistPollTenantIndexBuilders = tempodb.DefaultTenantIndexBuilders
```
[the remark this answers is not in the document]
joe-elliott: In the past we've been picky about what we register as flags to prevent "option overload". When a user runs `./tempo --help` we want them to only see the most common options required to get Tempo running and leave most options as settable in yaml only.

I feel like these options should not be CLI params, but am open to discussion.

tempodb/blocklist/list.go
```diff
@@ -0,0 +1,163 @@
...
+	return tenants
+}
+
+func (l *List) Metas(tenantID string) []*backend.BlockMeta {
+	l.mtx.Lock()
+	defer l.mtx.Unlock()
+
+	if tenantID == "" {
+		return nil
+	}
```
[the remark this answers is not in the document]
joe-elliott: It should never happen, but agree it is the preferred order regardless.

tempodb/blocklist/poller.go
joe-elliott: Agree there will be a rush of this error the first time that you upgrade to a version that supports the tenant index, but during normal operation I would consider this an error and think it should remain at this level.

tempodb/blocklist/poller.go
joe-elliott: For the moment I would prefer responding to the alert/metric `tempodb_blocklist_tenant_index_age_seconds` and treating it like an operational issue.

I expect this will not happen often, but if it does we can add an option in a future PR to fallback if the index age > some value.

cmd/tempo/app/modules.go
joe-elliott: > Would it be a better idea to only allow compactors to create the tenant index?
Currently only compactors create the index no matter the mode. The querier always just pulls the tenant index.

In the single binary if both the compactor and querier enable polling I think it will just duplicate work. There will be 2 goroutines in the single binary: One creating the tenant index and one pulling it. It honestly might not be worth this code to get the querier component to not poll.

tempodb/blocklist/poller.go
joe-elliott: As discussed offline this approach is simpler operationally and we will go with it for now. The mentioned per tenant sharding is preferable, however, in the long term. I will make an issue to add it after this PR is merged.

docs/tempo/website/operations/_index.md
```diff
@@ -11,4 +11,7 @@ Resources for deploying and running Tempo:
 - [Monitoring](monitoring)
 - [Ingester PVCs](ingester_pvcs)
 - [Caching](caching)
-- [Tempo CLI](tempo_cli)
\ No newline at end of file
+- [Tempo CLI](tempo_cli)
+
+
+# jpe tenant index
```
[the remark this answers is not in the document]
joe-elliott: todone!

cmd/tempo/app/modules.go
annanay25: > Currently only compactors create the index no matter the mode.

Umm, correct me if I'm wrong, but I feel even the queriers are attempting to create the tenant index:
- Queriers will call [pollTenantAndCreateIndex](https://github.com/grafana/tempo/blob/17867a66e5488f63671b75d4d41fd439507ee8c1/tempodb/blocklist/poller.go#L123)
- which in-turn calls [buildTenantIndex](https://github.com/grafana/tempo/blob/17867a66e5488f63671b75d4d41fd439507ee8c1/tempodb/blocklist/poller.go#L241).
- and since the poller config will be shared between all components, [TenantIndexBuilders](https://github.com/grafana/tempo/blob/17867a66e5488f63671b75d4d41fd439507ee8c1/tempodb/blocklist/poller.go#L242) will be > 0
- the queriers will then proceed to build the tenant index

Not sure what I'm missing

modules/storage/config.go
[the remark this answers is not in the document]
kvrhdn: Okay, didn't know about this but makes sense :+1:

docs/tempo/website/configuration/polling.md
```diff
@@ -0,0 +1,47 @@
...
+        # By default components will pull the blocklist from the tenant index. If that fails the component can
+        # fallback to scanning the entire bucket. Set to false to disable this behavior. Default is true.
+        [blocklist_poll_fallback: <bool>]
+
+        # Maximum number of compactors that should build the tenant index. All other components will download 
+        # the index.  Default 2.
+        [blocklist_poll_tenant_index_builders: <int>]
+```
+
+Due to the mechanics of the [tenant index]({{< relref "../operations/polling" >}}) the blocklist will stale by
```
kvrhdn, line 25: I think it's _to be stale_? (but I'm not a native speaker so might be wrong )
```suggestion
Due to the mechanics of the [tenant index]({{< relref "../operations/polling" >}}) the blocklist will be stale by
```

docs/tempo/website/configuration/polling.md
```diff
@@ -0,0 +1,47 @@
...
+at most 2x the configured `blockist_poll` duration. There are two configuration options that need to be balanced
```
kvrhdn, line 26: I'm not sure I correctly understand why it's 2x `blocklist_poll`: is it because a compactor will create a new index every poll interval and then it takes the querier up to this interval again to fetch it?
Maybe we should quickly describe this in operations/polling.md as well.

modules/storage/config.go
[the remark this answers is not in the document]
kvrhdn: I made an issue for this so we can track it https://github.com/grafana/tempo/issues/862

cmd/tempo/app/modules.go
joe-elliott: `buildTenantIndex` will always return false b/c the querier returns false here:

https://github.com/grafana/tempo/pull/834/files#diff-a1fec687cd87c7167aee355f831ebae6beb368066e3bfbf092ef0aaf5e216aa2R261

The compactor has logic to check the ring to determine if it should return true or false.

docs/tempo/website/configuration/polling.md
joe-elliott: you are correct!

docs/tempo/website/configuration/polling.md
joe-elliott: Yup. This is exactly why. I will add more details in operations/polling.md

kvrhdn: Thanks, this is clearer  :+1:

cmd/tempo/app/modules.go
annanay25: Got it. 👍

kvrhdn (approved):

annanay25 (approved):

PR #839, opened 2021-07-26
Log status_code of failed requests in retry middleware

kvrhdn:
**What this PR does**:

Log status_code of failed requests in retry middleware

### DIFF
--- a/modules/frontend/retry.go
+++ b/modules/frontend/retry.go
@@ -51,9 +51,14 @@ func (r retryWare) Do(req *http.Request) (*http.Response, error) {
 			return resp, nil
 		}
 
+		statusCode := 0
+		if resp != nil {
+			statusCode = resp.StatusCode
+		}
+
 		span.LogFields(
 			ot_log.String("msg", "error processing request"),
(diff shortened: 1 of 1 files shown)


modules/frontend/retry.go
```diff
@@ -51,6 +51,11 @@ func (r retryWare) Do(req *http.Request) (*http.Response, error) {
 			return resp, nil
 		}
 
-		span.LogFields(ot_log.String("msg", "error processing request"), ot_log.Int("try", triesLeft), ot_log.Error(err))
+		span.LogFields(
+			ot_log.String("msg", "error processing request"),
+			ot_log.Int("status_code", resp.StatusCode),
```
annanay25, line 56: CI caught this as well, but `resp` could be `nil` here

kvrhdn: Oh right, I wrote those tests myself 🤦

kvrhdn: Should be good now 👍🏻

mapno (approved):

PR #842, opened 2021-07-28
Use OpenTelemetry tracer, bridge OpenTracing and OpenCensus

kvrhdn:
**What this PR does**:
Adds a new feature flag `use_otel_tracer` which replaces the OpenTracing tracer with the OpenTelemetry tracer. The OTel tracer will install two bridges (OpenTracing and OpenCensus) so all existing spans will be collected and sent.

Why?
- This will allows us to gather OpenCensus spans, these are for instance emitted by the GCS SDK.
- OpenTracing has merged into OpenTelemetry

The feature flag defaults to false, i.e. keep using the OpenTracing tracer as before. There are a couple of issue we still need to iron out and not all features match (i.e. remote sampling is missing from the Jaeger exporter).

~~Fixes #<issue number>~~

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


kvrhdn:
Issues:

- context is not propagated between instances https://github.com/open-telemetry/opentelemetry-go/issues/2137
- we are getting misconfiguration warnings:
  ```
  SDK should have deferred the context setup, see the documentation of go.opentelemetry.io/otel/bridge/opentracing/migration
  ```

kvrhdn:
Example trace with spans from the GCS SDK
Screenshot 2021-07-29 at 16 08 36

kvrhdn:
Process tags have changed slightly:

### OpenTracing

Screenshot 2021-07-30 at 12 45 10

### OpenTelemetry

Screenshot 2021-07-30 at 12 46 14

cmd/tempo/main.go
```diff
@@ -171,3 +185,110 @@ func loadConfig() (*app.Config, error) {
...
+	}
+
+	tp := tracesdk.NewTracerProvider(
+		tracesdk.WithBatcher(exp),
+		tracesdk.WithResource(resources),
+	)
+	otel.SetTracerProvider(tp)
+
+	shutdown := func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
mdisibio, line 234: What is the rationale for 5 second timeout on shutdown?

cmd/tempo/main.go
```diff
@@ -171,3 +185,110 @@ func loadConfig() (*app.Config, error) {
...
+		"JAEGER_USER":       "OTEL_EXPORTER_JAEGER_USER",
+		"JAEGER_PASSWORD":   "OTEL_EXPORTER_JAEGER_PASSWORD",
+		"JAEGER_TAGS":       "OTEL_RESOURCE_ATTRIBUTES",
+	}
+	for jaegerKey, otelKey := range jaegerToOtel {
+		value, jaegerOk := os.LookupEnv(jaegerKey)
+		_, otelOk := os.LookupEnv(otelKey)
+
+		if jaegerOk && !otelOk {
+			level.Warn(log.Logger).Log("msg", "migrating Jaeger environment variable, consider using native OpenTelemetry variables", "variable", jaegerKey)
```
mdisibio, line 279: Also log `otelKey` ?

modules/frontend/frontend.go
```diff
@@ -100,7 +99,8 @@ func NewTripperware(cfg Config, logger log.Logger, registerer prometheus.Registe
 
 			span.SetTag("response marshalling format", marshallingFormat)
 
-			traceID, _ := middleware.ExtractTraceID(ctx)
+			otelspan := trace.SpanFromContext(ctx)
```
mdisibio, line 102: Is this backwards compatible when not using otel tracing?  Do we need to try to get a trace ID from both methods, or maybe main.go can setup a callback and provide it to this method somehow?

cmd/tempo/main.go
kvrhdn: This used in the Jaeger example as well: https://github.com/open-telemetry/opentelemetry-go/blob/23cb9396117ee0be80981ec9a60d1d1620688e29/example/jaeger/main.go#L78-L83
I'm not sure if this is really needed, other examples just use `context.Background()`.

modules/frontend/frontend.go
kvrhdn: Nope, it's not 🙃  This is still code from before I added the feature flag.

kvrhdn: Got this working now 🙂  I tried to test it but when you initialise one of the tracers they stay active for all the tests. I.e. the tests influence each other.

mdisibio (approved):

Issue #843, opened 2021-07-30
Query-frontend will occasionally panic when under sustained load


kvrhdn:
**Describe the bug**

Query-frontend will occasionally panic:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x47ad05]
goroutine 302 [running]:
errors.(*errorString).Error(0x0, 0x209c7e0, 0x0)
	/usr/local/go/src/errors/errors.go:68 +0x5
github.com/opentracing/opentracing-go/log.Field.Marshal(0x2485b37, 0xc, 0x9, 0x0, 0x0, 0x0, 0x209c7e0, 0x0, 0x2995038, 0xc06a0b00d8)
	/drone/src/vendor/github.com/opentracing/opentracing-go/log/field.go:231 +0x309
github.com/uber/jaeger-client-go.ConvertLogsToJaegerTags(0xc000a68780, 0x3, 0x3, 0xc06984e008, 0x39c7248, 0xc000cfbd78)
	/drone/src/vendor/github.com/uber/jaeger-client-go/jaeger_tag.go:31 +0x13f
github.com/uber/jaeger-client-go.buildLogs(0xc06975e480, 0x1, 0x1, 0x100, 0x39c7248, 0x0)
	/drone/src/vendor/github.com/uber/jaeger-client-go/jaeger_thrift_span.go:82 +0x145
github.com/uber/jaeger-client-go.BuildJaegerThrift(0xc0697586c0, 0x0)
	/drone/src/vendor/github.com/uber/jaeger-client-go/jaeger_thrift_span.go:43 +0x16c
github.com/uber/jaeger-client-go.(*udpSender).Append(0xc000ccad80, 0xc0697586c0, 0x0, 0x0, 0x2)
	/drone/src/vendor/github.com/uber/jaeger-client-go/transport_udp.go:129 +0x45
github.com/uber/jaeger-client-go.(*remoteReporter).processQueue(0xc000533f80)
	/drone/src/vendor/github.com/uber/jaeger-client-go/reporter.go:304 +0x1c2
created by github.com/uber/jaeger-client-go.NewRemoteReporter
	/drone/src/vendor/github.com/uber/jaeger-client-go/reporter.go:237 +0x1a5
```

We observed this while receiving a constant load of about 23 queries per second.

**To Reproduce**
1. Run Tempo
2. Send a ton of queries

**Expected behavior**
[...]

PR #849, opened 2021-08-02
Properly handle max_retries set to 0 or lower

kvrhdn:
**What this PR does**:
If `max_retries` was set to 0 or lower, this would lead to up to infinite retries. This PR fixes this.

Bad code: https://github.com/grafana/tempo/blob/d50c64ba9df22e20fb895562aec3e04c107bf6a2/modules/frontend/retry.go#L45-L48

`triesLeft` starts at the value of `max_retries`, so if its 0 we never match `triesLeft == 0`.

Spotted while investigating #843.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/retry.go
+++ b/modules/frontend/retry.go
@@ -33,7 +33,7 @@ func (r retryWare) Do(req *http.Request) (*http.Response, error) {
 	// context propagation
 	req = req.WithContext(ctx)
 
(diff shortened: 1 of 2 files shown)


kvrhdn:
Closing in favour of #850

PR #850, opened 2021-08-02
query-frontend/retry: properly handle grpc errors; lower default max_retries

### LINKED ISSUE #824: Query-frontend retries queries that timed out in querier
**Describe the bug**
After merging #814 I noticed the query-frontend is retrying request that timed out in the querier.

The good: after retrying the request, it usually succeeds quickly
The bad: if we keep retrying a query that timed out, the query might hang for up `max_retries * query_timeout` seconds (50s by default).

An example of such a trace:

Screenshot 2021-07-16 at 14 52 32

**To Reproduce**
1. Start Tempo (2ad27e7c or more recent)
2. Perform a query
3. If the querier times out, observe retries in the query-frontend

**Expected behavior**
I'm not sure to be honest. Some options:

- do not retry a query that timed out
- enforce a timeout in the query-frontend
- leave it as is (but a request that OOMs the querier could be sent 5 times this way)

### LINKED ISSUE #843: Query-frontend will occasionally panic when under sustained load

### LINKED ISSUE #844: Retry tripperware will retry 429 errors wrapped in RPC errors
**Describe the bug**
I've spotted the retry tripperware retrying HTTP 429 errors (= too many requests).

Screenshot 2021-07-30 at 14 38 46

**To Reproduce**
1. Run Tempo
2. Perform queries
3. Wait untli the querier returns with an HTTP 4xx error (I guess?)

**Expected behavior**
We should not retry 4xx errors, these are caused by the caller. In fact, retrying a 429 error will make matters worse 😬 

**Environment:**
 - Infrastructure: Kubernetes
 - Deployment tool: jsonnet

**Additional Context**

kvrhdn:
**What this PR does**:
Improves the retry middleware in query-frontend:
- when err is a grpc error, extract http response and check status code (fixes #844)
- lower default retry to 2
- add a prometheus metric `query_frontend_retries`
- don't log errors (avoids #843)

Closes #844
Closes #843
Supersedes #849
Closes #824

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/frontend/retry.go
+++ b/modules/frontend/retry.go
@@ -52,13 +52,13 @@ func (r retryWare) Do(req *http.Request) (*http.Response, error) {
 		resp, err := r.next.Do(req)
 
 		// do not retry if no error and response is not HTTP 5xx
-		if err == nil && resp.StatusCode/100 != 5 {
+		if err == nil && !shouldRetry(resp.StatusCode) {
 			return resp, nil
 		}
 
 		// do not retry if GRPC error contains response that is not HTTP 5xx
 		httpResp, ok := httpgrpc.HTTPResponseFromError(err)
-		if ok && httpResp.Code/100 != 5 {
+		if ok && !shouldRetry(int(httpResp.Code)) {
 			return resp, err
 		}
 
@@ -87,4 +87,8 @@ func (r retryWare) Do(req *http.Request) (*http.Response, error) {
 			ot_log.String("errMsg", errMsg),
 		)
 	}
+}
+
+func shouldRetry(statusCode int) bool {
+	return statusCode/100 == 5
 }
--- a/modules/frontend/retry.go
+++ b/modules/frontend/retry.go
@@ -3,25 +3,32 @@ package frontend
 import (
 	"net/http"
 
-	"github.com/go-kit/kit/log"
 	"github.com/opentracing/opentracing-go"
 	ot_log "github.com/opentracing/opentracing-go/log"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/weaveworks/common/httpgrpc"
 )
 
-func RetryWare(maxRetries int, logger log.Logger) Middleware {
+func RetryWare(maxRetries int, registerer prometheus.Registerer) Middleware {
 	return MiddlewareFunc(func(next Handler) Handler {
 		return retryWare{
 			next:       next,
-			logger:     logger,
 			maxRetries: maxRetries,
+			retriesCount: promauto.With(registerer).NewHistogram(prometheus.HistogramOpts{
+				Namespace: "tempo",
+				Name:      "query_frontend_retries",
+				Help:      "Number of times a request is retried.",
+				Buckets:   []float64{0, 1, 2, 3, 4, 5},
+			}),
 		}
 	})
 }
 
 type retryWare struct {
-	next       Handler
-	logger     log.Logger
-	maxRetries int
+	next         Handler
+	maxRetries   int
+	retriesCount prometheus.Histogram
 }
 
 // Do implements Handler
@@ -33,7 +40,8 @@ func (r retryWare) Do(req *http.Request) (*http.Response, error) {
 	// context propagation
 	req = req.WithContext(ctx)
 
-	triesLeft := r.maxRetries
+	tries := 0
+	defer func() { r.retriesCount.Observe(float64(tries)) }()
 
 	for {
 		if ctx.Err() != nil {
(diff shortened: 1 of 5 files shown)


achatterjee-grafana (approved):

modules/frontend/retry.go
```diff
@@ -42,25 +50,23 @@ func (r retryWare) Do(req *http.Request) (*http.Response, error) {
...
+			return resp, err
 		}
 
-		span.LogFields(
-			ot_log.String("msg", "error processing request"),
-			ot_log.Int("status_code", statusCode),
-			ot_log.Error(err),
-			ot_log.Int("triesLeft", triesLeft),
-		)
+		span.LogFields(ot_log.String("msg", "error processing request. retrying"))
```
mdisibio, line 70: What is the thinking to remove the extra data from the span? It seems helpful.

modules/frontend/retry.go
```diff
@@ -42,25 +50,23 @@ func (r retryWare) Do(req *http.Request) (*http.Response, error) {
...
 		if err == nil && resp.StatusCode/100 != 5 {
 			return resp, nil
 		}
 
-		statusCode := 0
-		if resp != nil {
-			statusCode = resp.StatusCode
+		// do not retry if GRPC error contains response that is not HTTP 5xx
+		httpResp, ok := httpgrpc.HTTPResponseFromError(err)
+		if ok && httpResp.Code/100 != 5 {
```
mdisibio, line 60: Researching on which http status codes are retry-able, it seems maybe we shouldn't retry 501 (Not Implemented), but should also retry 429 (Too Many Requests).  What do you think?

mdisibio (approved):
LGTM, left a few comments but not blocking. This solves the main issue of not retrying on 4xx.

modules/frontend/retry.go
kvrhdn: I removed `ot_log.Error(err)` because this triggers https://github.com/grafana/tempo/issues/843
Agree it's valuable info, but for some reason calling `err.Error()` can cause panics 🙃 

I'll re-add the tries.

modules/frontend/retry.go
kvrhdn: 429 is intended as a kind of rate-limiting I believe, if you get this you should send less requests. So I don't want to retry this to avoid piling on even more.

If we can ensure retries are sent to a different querier, I agree it would be good to retry this as well. #856

kvrhdn: I think it's fine to ignore 501 since this shouldn't happen within Tempo. If we hit an endpoint that is not enabled (for instance with search) Tempo returns 404 not found.

modules/frontend/retry.go
kvrhdn: Found a safe way to log the error 🙃

modules/frontend/retry.go
kvrhdn: Yeah grpc returns 429 errors: I've seen the queriers return 429 when Tempo was getting a lot of request.

mdisibio: Oh I just realized that retrying on 429 prompted this fix in #844.  Yep agree that might be good to explore later with better logic: a different querier or a backoff algorithm.

PR #854, opened 2021-08-04
Allow only valid trace ID characters when decoding

### LINKED ISSUE #846: Trace ID with extra space triggers complex error message
**Describe the bug**
I copy-pasted a trace ID and added a space at the end by accident.
The resulting error was:
```
failed to get trace with id: 1f34e43692d1f5d8 Status: 400 Bad Request Body: encoding/hex: invalid byte: U+0020 ' '
```

This error is quite difficult to understand, and also not helpful in pointing out the problem.

**Expected behavior**
Something like this would be easier to use:
```
failed to get trace with id: "1f34e43692d1f5d8 "; invalid character: U+0020 ' ' at position 17
```

zalegrala:
Without this change, invalid characters received from the user input causes a bad error message to the user.  Here we check that all of the characters that are to be decoded are valid hex characters and improve the error to the user in the event the trace ID submitted has invalid characters.

Fixes #846

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/util/http.go
+++ b/pkg/util/http.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/gorilla/mux"
 )
@@ -32,6 +33,18 @@ func ParseTraceID(r *http.Request) ([]byte, error) {
 }
 
 func hexStringToTraceID(id string) ([]byte, error) {
+	// The encoding/hex package does not handle non-hex characters.
+	// Ensure the ID has only the proper characters
+	for pos, idChar := range strings.Split(id, "") {
+		if (idChar >= "a" && idChar <= "f") ||
+			(idChar >= "A" && idChar <= "F") ||
+			(idChar >= "0" && idChar <= "9") {
+			continue
+		} else {
+			return nil, fmt.Errorf("trace IDs can only contain hex characters: invalid character '%s' at position %d", idChar, pos+1)
+		}
+	}
+
 	// the encoding/hex package does not like odd length strings.
 	// just append a bit here
 	if len(id)%2 == 1 {
@@ -45,7 +58,7 @@ func hexStringToTraceID(id string) ([]byte, error) {
 
 	size := len(byteID)
 	if size > 16 {
-		return nil, errors.New("trace ids can't be larger than 128 bits")
+		return nil, errors.New("trace IDs can't be larger than 128 bits")
 	}
 	if size < 16 {
 		byteID = append(make([]byte, 16-size), byteID...)
--- a/pkg/util/http_test.go
+++ b/pkg/util/http_test.go
@@ -1,6 +1,7 @@
 package util
 
 import (
+	"errors"
 	"testing"
 
 	"github.com/grafana/tempo/cmd/tempo-query/tempo"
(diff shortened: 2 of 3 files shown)


dgzlopes (approved):

pkg/util/http_test.go
```diff
@@ -36,6 +36,11 @@ func TestHexStringToTraceID(t *testing.T) {
 			id:       "234567890abcdef", // odd length
 			expected: []byte{0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef},
 		},
+		{
+			id:          "1234567890abcdef ", // trailing space
+			expected:    nil,
+			expectError: true,
```
kvrhdn, line 42: Can you check we get the correct error message and not just an error in general?

kvrhdn (commented):
Besides the small note, this looks okay to me 🙂 

I'm trying to find the most common invalid characters, but I'm not sure we log this somewhere...

pkg/util/http_test.go
zalegrala: Thanks for the review.  Good suggestion on checking the error.  I've updated the commit to check the specific errors in the util tests.

kvrhdn (approved):

pkg/util/http.go
```diff
@@ -32,6 +33,15 @@ func ParseTraceID(r *http.Request) ([]byte, error) {
 }
 
 func hexStringToTraceID(id string) ([]byte, error) {
+	// The encoding/hex package does not handle non-hex characters.
+	// Catch a few common characters early.
+	nonHexChars := []string{" ", "-", "/", "_", "\\", "(", ")", "[", "]"}
```
mdisibio, line 38: This changes the error message, but it doesn't introduce any new information, so I'm not sure it will help the user resolve the issue better.  I.e. the space at the end may still be easily overlooked, without showing the position of the invalid char.  

I was on board with doing a simple trim for unambiguous cases (like a trailing space).  If we want to go the route of erroring still, then let's switch from enumerating invalid chars, to ensuring the input only has valid chars: 0-9, a-f, A-F, and print the position.

mdisibio (changes requested):

pkg/util/http.go
zalegrala: Thanks for the review.  Good points about catching only valid characters.  I've updated the PR to reflect this desire and point at the string positions of any bad characters if they are found in the trace ID.

pkg/util/http_test.go
```diff
@@ -30,20 +31,25 @@ func TestHexStringToTraceID(t *testing.T) {
 		{
 			id:          "121234567890abcdef1234567890abcdef", // value too long
 			expected:    []byte{0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef, 0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef},
-			expectError: true,
+			expectError: errors.New("trace ids can't be larger than 128 bits"),
```
JosephWoodward, line 34: Should `ids` be uppercased to match existing error messages?

zalegrala: Sure, I can update.

mdisibio (approved):

Issue #802, opened 2021-07-07
Dashboards for distributed deployments require undocumented config


kvrhdn:
The [dashboards in tempo-mixin](https://github.com/grafana/tempo/tree/main/operations/tempo-mixin) use a cluster and job matcher: https://github.com/grafana/tempo/blob/29c6e25f5c0f9750b16fca673d437a490a6e3069/operations/tempo-mixin/dashboard-utils.libsonnet#L31-L32

This requires a cluster and job label to be present in the time series, which is not documented anywhere.

We should either:

- document what is expected to use this dashboard
- adapt the dashboard to not require this

Ideally, we also update the tk and helm examples to use these dashboards out of the box.

kvrhdn:
fyi, adding an external label worked for me.

Using the Grafana Agent:
```yaml
prometheus:
  global:
    external_labels:
      cluster: k3d-local-dev
  # ...
```

B3DTech:
This seems like it should have been taken care of by now, but I'm still experiencing the issue of the missing cluster label.  Any ideas?

kvrhdn:
This should be fixed by https://github.com/grafana/tempo/pull/1913

PR #859, opened 2021-08-09
Include distributed docker-compose example

### LINKED ISSUE #746: Add a tempo-distributed docker-compose
It would be nice to have a docker-compose setup that has all services. This would be useful for testing the distributed mode of different components - for ex: routing by traceID on distributor with a replication factor > 1, sharding on the query path, rolling out different images for different services, testing a breaking config/schema change, etc.

We currently have one for performance testing of the write path - https://github.com/grafana/tempo/blob/main/integration/microservices/docker-compose.yaml. We could copy this over and add the query path components.

zalegrala:
Without this change, there is not an example of how a user might proceed with
local testing and development using docker-compose.  Here we copy the local
docker-compose example and extend it to include the distributed configuration
pieces necessary to get a functional example.

Fixes #746

**Checklist**
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 * [ENHANCEMENT] Add retry middleware in query-frontend. [#814](https://github.com/grafana/tempo/pull/814) (@kvrhdn)
 * [ENHANCEMENT] Add `-use-otel-tracer` to use the OpenTelemetry tracer, this will also capture traces emitted by the gcs sdk. Experimental: not all features are supported (i.e. remote sampling). [#842](https://github.com/grafana/tempo/pull/842) (@kvrhdn)
 * [ENHANCEMENT] Add `/services` endpoint. [#863](https://github.com/grafana/tempo/pull/863) (@kvrhdn)
+* [ENHANCEMENT] Include distributed docker-compose example [#859](https://github.com/grafana/tempo/pull/859) (@zalegrala)
 
 ## v1.0.1
 
--- a/example/docker-compose/distributed/prometheus.yaml
+++ b/example/docker-compose/distributed/prometheus.yaml
@@ -1,0 +1,18 @@
+global:
+  scrape_interval:     15s
+  evaluation_interval: 15s
+
+scrape_configs:
+  - job_name: 'prometheus'
+    static_configs:
+      - targets: [ 'localhost:9090' ]
+  - job_name: 'tempo'
+    static_configs:
+      - targets:
+        - 'compactor:3200'
+        - 'distributor:3200'
+        - 'ingester-0:3200'
+        - 'ingester-1:3200'
+        - 'ingester-2:3200'
+        - 'querier:3200'
+        - 'query-frontend:3200'
--- a/example/docker-compose/distributed/readme.md
+++ b/example/docker-compose/distributed/readme.md
@@ -1,0 +1,64 @@
+## Distributed
+
+In this example tempo is configured in a distributed manner, where modules are
+running as separate components and configured to write data to S3 via MinIO,
+which presents an S3 compatible API.
+
+1. First start up the distributed stack.
+
+```console
+docker-compose up -d
+```
+
+At this point, the following containers should be spun up -
+
+```console
+docker-compose ps
+```
+```
+                 Name                               Command               State                         Ports                       
+------------------------------------------------------------------------------------------------------------------------------------
+distributed_compactor_1                  /tempo -target=compactor - ...   Up      0.0.0.0:49662->3200/tcp,:::49652->3200/tcp        
+distributed_distributor_1                /tempo -target=distributor ...   Up      0.0.0.0:49659->3200/tcp,:::49649->3200/tcp        
+distributed_grafana_1                    /run.sh                          Up      0.0.0.0:3000->3000/tcp,:::3000->3000/tcp          
+distributed_ingester-0_1                 /tempo -target=ingester -c ...   Up      0.0.0.0:49663->3200/tcp,:::49653->3200/tcp        
+distributed_ingester-1_1                 /tempo -target=ingester -c ...   Up      0.0.0.0:49660->3200/tcp,:::49650->3200/tcp        
+distributed_ingester-2_1                 /tempo -target=ingester -c ...   Up      0.0.0.0:49665->3200/tcp,:::49655->3200/tcp        
+distributed_minio_1                      sh -euc mkdir -p /data/tem ...   Up      9000/tcp, 0.0.0.0:9001->9001/tcp,:::9001->9001/tcp
+distributed_prometheus_1                 /bin/prometheus --config.f ...   Up      0.0.0.0:9090->9090/tcp,:::9090->9090/tcp          
+distributed_querier_1                    /tempo -target=querier -co ...   Up      0.0.0.0:49664->3200/tcp,:::49654->3200/tcp        
+distributed_query-frontend_1             /tempo -target=query-front ...   Up      0.0.0.0:49661->3200/tcp,:::49651->3200/tcp        
+distributed_synthetic-load-generator_1   ./start.sh                       Up                                                        
+```
+
+2. If you're interested you can see the wal/blocks as they are being created.  Navigate to minio at
+http://localhost:9001 and use the username/password of `tempo`/`supersecret`.
+
+3. The synthetic-load-generator is now printing out trace ids it's flushing into Tempo.  To view its logs use -
+
+```console
+docker-compose logs -f synthetic-load-generator
+```
+```
+synthetic-load-generator_1  | 20/10/24 08:27:09 INFO ScheduledTraceGenerator: Emitted traceId 57aedb829f352625 for service frontend route /product
+synthetic-load-generator_1  | 20/10/24 08:27:09 INFO ScheduledTraceGenerator: Emitted traceId 25fa96b9da24b23f for service frontend route /cart
+synthetic-load-generator_1  | 20/10/24 08:27:09 INFO ScheduledTraceGenerator: Emitted traceId 15b3ad814b77b779 for service frontend route /shipping
+synthetic-load-generator_1  | 20/10/24 08:27:09 INFO ScheduledTraceGenerator: Emitted traceId 3803db7d7d848a1a for service frontend route /checkout
+```
+
+Logs are in the form
+
+```
+Emitted traceId <traceid> for service frontend route /cart
+```
+
+Copy one of these trace ids.
+
+4. Navigate to [Grafana](http://localhost:3000/explore) and paste the trace id to request it from Tempo.
+Also notice that you can query Tempo metrics from the Prometheus data source setup in Grafana.
+
+5. To stop the setup use -
+
+```console
+docker-compose down -v
+```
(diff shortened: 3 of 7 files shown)


example/docker-compose/local-distributed/tempo-distributed.yaml
```diff
@@ -0,0 +1,62 @@
...
+    block_retention: 1h
+    compacted_block_retention: 10m
+
+query_frontend:
+    max_retries: 99
+    query_shards: 2
+
+querier:
+  frontend_worker:
+    frontend_address: tempo:9095
```
annanay25, line 46:
```suggestion
    frontend_address: query-frontend:9095
```

example/docker-compose/local-distributed/docker-compose.yaml
```diff
@@ -0,0 +1,122 @@
...
+      - ingester-0
+      - ingester-1
+      - ingester-2
+
+  querier:
+    image: tempo:latest
+    command: "-target=querier -config.file=/etc/tempo.yaml"
+    volumes:
+      - ./tempo-distributed.yaml:/etc/tempo.yaml
+      - ./tempo-data:/tmp/tempo
```
annanay25, line 81: Can remove.

annanay25 (commented):
Nice work @zalegrala! Added a few comments.

example/docker-compose/local-distributed/docker-compose.yaml
```diff
@@ -0,0 +1,122 @@
...
+
+  prometheus:
+    image: prom/prometheus:latest
+    command: [ "--config.file=/etc/prometheus.yaml" ]
+    volumes:
+      - ../shared/prometheus.yaml:/etc/prometheus.yaml
+    ports:
+      - "9090:9090"
+
+  grafana:
```
kvrhdn, line 113: This might be something for a second PR, but it would be really nice if we could test our dashboards with this setup.
We have dashboards that only work with a distributed setup, to work properly I believe they need a `cluster` and `namespace` label: https://github.com/grafana/tempo/tree/main/operations/tempo-mixin/yamls

example/docker-compose/local-distributed/docker-compose.yaml
```diff
@@ -0,0 +1,122 @@
...
+    image: grafana/grafana:7.5.7
```
kvrhdn, line 114: We can bump this!
```suggestion
    image: grafana/grafana:8.1.1
```

example/docker-compose/local-distributed/tempo-distributed.yaml
```diff
@@ -0,0 +1,62 @@
...
+  - ingester-1:7946
+  - ingester-2:7946
+
+compactor:
+  compaction:
+    block_retention: 1h
+    compacted_block_retention: 10m
+
+query_frontend:
+    max_retries: 99
```
kvrhdn, line 41: This is really high, the default is 2. We retry after query sharding, so with `query_shards: 2` this could send up to 198 requests to the querier 😅 
If the query fails because some component isn't ready, the user should just retry later.

example/docker-compose/local-distributed/docker-compose.yaml
```diff
@@ -0,0 +1,122 @@
...
+      - ./tempo-distributed.yaml:/etc/tempo.yaml
+      - ./tempo-data2:/tmp/tempo
+    ports:
+      - "14268"  # jaeger ingest
+      - "3200"   # tempo
+      - "7946"
+    depends_on:
+      - distributor
+
+  tempo:
```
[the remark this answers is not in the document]
kvrhdn: fyi, if you change this you will also have to change the configured datasources of Grafana.

example/docker-compose/local-distributed/docker-compose.yaml
kvrhdn: Doesn't the querier need access to the storage to query blocks flushed by the ingesters? But I'm not even sure how the ingesters will flush their blocks in this setup 🤔  I have never run the local backend in distributed mode.
It might be easier to use a containerised backend like fake (gcs) or minio (s3).

annanay25: Ah that's right. We need this one.

example/docker-compose/local-distributed/docker-compose.yaml
zalegrala: Let's follow up on that in another PR.  Sounds pretty useful to have a way to test those dashboards.

kvrhdn: Yep, agree :+1:

This issue is related, getting them working with the docker-compose examples will be a good example of how to configure them https://github.com/grafana/tempo/issues/802

zalegrala:
Thanks for the review!  I believe I addressed the comments, except for the distributed dashboards testing.

kvrhdn (commented):
I just noticed we are missing the compactor 😳 
The compactor is a required component to run Tempo long-term: https://grafana.com/docs/tempo/latest/operations/architecture/

And I querying the local backend doesn't seem to work well. I can successfully query a trace ID for about 10 minutes and then I get "trace not found". So I believe this matches with when the ingesters flush their blocks to the backend.
As long as the block lives on the ingester, the querier can query the ingesters to find the trace. But once the block is flushed and deleted, the querier has to find it in the backend. I guess something is going wrong while flushing...

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## main / unreleased
 
+* [ENHANCEMENT] Include distributed docker-compose example [#859](https://github.com/grafana/tempo/pull/859) (@zalegrala)
```
joe-elliott, line 3: please put this enhancement with the others in the changelog.

example/docker-compose/local-distributed/docker-compose.yaml
```diff
@@ -0,0 +1,107 @@
...
+    image: minio/minio:latest
+    environment:
+      - MINIO_ACCESS_KEY=tempo
+      - MINIO_SECRET_KEY=supersecret
+    ports:
+      - "9000:9000"
+    entrypoint:
+      - sh
+      - -euc
+      - mkdir -p /data/tempo && /usr/bin/minio server /data
```
joe-elliott, line 80: Add the `--console-address` parameter and set it to something besides 9000 (e.g. 9001) so we can access the minio console.  See the `../s3/docker-compose.yaml` file.

example/docker-compose/local-distributed/docker-compose.yaml
```diff
@@ -0,0 +1,107 @@
...
+    image: tempo:latest
+    command: "-target=querier -config.file=/etc/tempo.yaml"
+    volumes:
+      - ./tempo-distributed.yaml:/etc/tempo.yaml
+    ports:
+      - "3200"   # tempo
+
+  compactor:
+    image: tempo:latest
+    command: "-target=compactor -config.file=/etc/tempo.yaml"
```
joe-elliott, line 63: add `restart: always`

zalegrala:
Thanks for the review.  This is looking pretty stable now that minio is in use.

example/docker-compose/local-distributed/docker-compose.yaml
```diff
@@ -0,0 +1,108 @@
+version: "3"
+services:
+
+  distributor:
+    image: tempo:latest
+    command: "-target=distributor -config.file=/etc/tempo.yaml"
+    volumes:
+      - ./tempo-distributed.yaml:/etc/tempo.yaml
+    ports:
+      - "14268"  # jaeger ingest
```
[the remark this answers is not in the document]
zalegrala: I'll drop the extra ingest ports, since it sounds like maybe that isn't something we want to test.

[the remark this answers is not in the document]
mdisibio: That's fair.  Any concern about still dropping 7946?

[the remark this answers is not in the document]
zalegrala: Not from me, I'll drop those too.

annanay25 (approved):
LGTM. Nice work @zalegrala 🚀

example/docker-compose/distributed/prometheus.yaml
```diff
@@ -0,0 +1,17 @@
...
+      - targets: [ 'localhost:9090' ]
+  - job_name: 'tempo'
+    static_configs:
+      - targets:
+        - 'distributor:3200'
+        - 'ingester-0:3200'
+        - 'ingester-1:3200'
+        - 'ingester-2:3200'
+        - 'query-frontend:3200'
+        - 'querier:3200'
```
kvrhdn, line 17: We are missing the compactor here

example/docker-compose/distributed/readme.md
```diff
@@ -0,0 +1,66 @@
+## Local Storage
+
+In this example all data is stored locally in the `tempo-data*` directories for each of the ingester instances. Local storage is fine for experimenting with Tempo
+or when using the single binary, and each instance mounts its own data directory to keep the cluster resources separated.
```
kvrhdn, line 4: I believe this is still copied from the local example. You can probably copy some text from the s3 example and mention that with this setup all components run independently, which is more scalable and fault-tolerant.

example/docker-compose/distributed/readme.md
```diff
@@ -0,0 +1,66 @@
...
+distributed_querier_1                    /tempo -target=querier -co ...   Up      0.0.0.0:49664->3200/tcp,:::49654->3200/tcp        
+distributed_query-frontend_1             /tempo -target=query-front ...   Up      0.0.0.0:49661->3200/tcp,:::49651->3200/tcp        
+distributed_synthetic-load-generator_1   ./start.sh                       Up                                                        
+```
+
+2. If you're interested you can see the wal/blocks as they are being created.
+
+```console
+ls tempo-data*/
+```
```
kvrhdn, line 37: This is also copied from the local example. Instead, you can now copy `2.` from the s3 example 🙂

kvrhdn (commented):
LGTM 🙂  Just some small details left to straighten out. The example works as expected and it's a welcome addition!

example/docker-compose/distributed/prometheus.yaml
zalegrala: Ah yes, good catch.

example/docker-compose/distributed/readme.md
```diff
@@ -0,0 +1,62 @@
+## Local Storage
```
kvrhdn, line 1:
```suggestion
## Distributed
```

zalegrala: Oh no I thought I got them all!  :)

example/docker-compose/distributed/readme.md
kvrhdn: I'd also mention this is the distributed setup, otherwise it's almost the same as the s3 example.

> _In this example tempo is configured to write data to S3 via MinIO which presents an S3 compatible API._

⬇️ 

> _In this example Tempo is deployed in distributed mode and writes data to S3 via MinIO, in distributed mode the Tempo components are deployed as microservices. The microservices deploy of Tempo is fault tolerant, high volume, independently scalable._

What do you think?

zalegrala: That sounds good to me.  I've made the update to include some verbiage about being distributed.

kvrhdn (approved):
LGTM, nice work. This will be very useful 🙂

PR #863, opened 2021-08-11
Add /services endpoint

kvrhdn:
**What this PR does**:
Add `/services` which prints a plaintext list of all the services in the service map with their status and optional failure case.

```
$ curl  localhost:3201/services
memberlist-kv: Running
querier: Running
ring: Running
server: Running
store: Running
```

Cortex has a similar endpoint which returns HTML, but I decided to keep this simple and plaintext.
https://github.com/cortexproject/cortex/blob/master/pkg/cortex/status.go

~~Fixes #<issue number>~~

**Checklist**
- [ ] Tests updated
- [x] Documentation added


joe-elliott (approved):
nice!

i was wanting this yesterday when the compactors were failing to start for some reason.

PR #868, opened 2021-08-11
Update release workflow to ensure go 1.16

mdisibio:
**What this PR does**:
Github release workflow is failing with error.  This adds a step to install go 1.16 similar to ci and bench workflows. This solves the issue as best I can tell locally using https://github.com/nektos/act  
```
# github.com/grafana/tempo/tempodb/backend
tempodb/backend/readerat.go:79:9: undefined: "io".ReadAll
note: module requires Go 1.16
```

n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,13 +10,22 @@ jobs:
   release:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - name: Set up Go 1.16
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.16
+        id: go
+
+      - name: Check out code
+        uses: actions/checkout@v2
+        
       - name: Release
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -10,13 +10,22 @@ jobs:
   release:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - name: Set up Go 1.16
+        uses: actions/setup-go@v1
```
kvrhdn, line 14: setup-go has a v2 release, don't know if it is worth upgrading: https://github.com/actions/setup-go#v2

kvrhdn (approved):
This looks good, I don't know why it doesn't use Go 1.16 by default. It should be installed according to this: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#go

PR #876, opened 2021-08-13
Added query blocks cli command

joe-elliott:
**What this PR does**:
Adds the ability to query the backend blocks directly and dump all info found for a trace id. This also changes the existing query option to better distinguish it in the cli.

`./tempo-cli query` => `./tempo-cli query api` or `./tempo-cli query blocks`

The command outputs each block found, its contents and the combined trace.

**Checklist**
- [ ] Tests updated
- [x] Documentation added


mdisibio:
Nice!  In the documentation let's put the warning that it can be intense, as it might download quite a bit of data while checking bloom filters, etc.

mdisibio (approved):

PR #887, opened 2021-08-18
Rename and clarify TempoFlushesFailing -> TempoIngesterFlushesFailing

kvrhdn:
**What this PR does**:
Clarifies the `TempoFlushesFailing` alert and renames it to `TempoIngesterFlushesFailing`.

- The original description mentions failed compactions, but this alert can only be triggered by ingesters. It uses the `tempo_ingester_failed_flushes_total` metric that is created in `modules/ingester/flush.go`
- The section about the compactor creating orphaned blocks seems useful, maybe we should move it to another alert?

~~Fixes #<issue number>~~

### DIFF
--- a/operations/tempo-mixin/alerts.libsonnet
+++ b/operations/tempo-mixin/alerts.libsonnet
@@ -82,7 +82,7 @@
             },
           },
           {
-            alert: 'TempoFlushesFailing',
+            alert: 'TempoIngesterFlushesFailing',
             expr: |||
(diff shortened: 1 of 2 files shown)


annanay25 (approved):
LGTM. Wonder why we had the compactor stuff in the alert runbook before 🤔 

This is the only place the metric is used - https://github.com/grafana/tempo/blob/addfa703eb5db8b44cd4c3929daa3e1b42bd3103/modules/ingester/flush.go#L213

PR #891, opened 2021-08-18
Add gzip middleware to HTTP endpoints

kvrhdn:
**What this PR does**:
Adds gzip middleware to the query endpoints. See the description of https://github.com/grafana/tempo/pull/810 for why this needed.

The gzip middleware will compress responses if

1. the clients requests this by setting `Accept-Encoding: gzip`
2. the response is larger than 1500 bytes

This uses [nytimes/gziphandler](https://github.com/nytimes/gziphandler), which is also used by Cortex.

To test this, you can use `curl` with the `--compressed` switch.
The following trick prints the actual download size. `size_download` is _the total amount of bytes that were downloaded._

- Get trace without asking for compression
```
$ curl -sv -o /dev/null localhost:3200/api/traces/7295a9c7bfcb3ace40d6d55fd082ef78 -w '%{size_download}'

> GET /api/traces/7295a9c7bfcb3ace40d6d55fd082ef78 HTTP/1.1
> Host: localhost:3200
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Vary: Accept-Encoding
< Date: Wed, 18 Aug 2021 15:49:42 GMT
< Content-Type: text/plain; charset=utf-8
< Transfer-Encoding: chunked
<

79975
```
- Ask for compression by setting `--compressed`
```
$ curl --compressed -sv -o /dev/null localhost:3200/api/traces/7295a9c7bfcb3ace40d6d55fd082ef78 -w '%{size_download}'

> GET /api/traces/7295a9c7bfcb3ace40d6d55fd082ef78 HTTP/1.1
> Host: localhost:3200
> User-Agent: curl/7.64.1
> Accept: */*
> Accept-Encoding: deflate, gzip
>
< HTTP/1.1 200 OK
< Content-Encoding: gzip
< Content-Type: text/plain; charset=utf-8
< Vary: Accept-Encoding
< Date: Wed, 18 Aug 2021 15:50:29 GMT
< Transfer-Encoding: chunked
<

22562
```

To test this make sure you have large traces, I'm using traces generated by vulture.

Fixes #

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo/app/http.go
+++ b/cmd/tempo/app/http.go
@@ -1,0 +1,23 @@
+package app
+
+import (
+	"net/http"
+
+	"github.com/NYTimes/gziphandler"
+	"github.com/weaveworks/common/middleware"
+)
+
+func httpGzipMiddleware() middleware.Interface {
+	return middleware.Func(func(handler http.Handler) http.Handler {
+		gzipHandler := gziphandler.GzipHandler(handler)
+
+		return http.HandlerFunc(func(writer http.ResponseWriter, r *http.Request) {
+			// do not gzip the response when requesting protobuf as this will mess up the encoding
+			if r.Header.Get("Accept") == "application/protobuf" {
+				handler.ServeHTTP(writer, r)
+			} else {
+				gzipHandler.ServeHTTP(writer, r)
+			}
+		})
+	})
+}
--- a/cmd/tempo/app/modules.go
+++ b/cmd/tempo/app/modules.go
@@ -158,6 +158,7 @@ func (t *App) initQuerier() (services.Service, error) {
 
 	tracesHandler := middleware.Merge(
 		t.httpAuthMiddleware,
+		httpGzipMiddleware(),
 	).Wrap(http.HandlerFunc(t.querier.TraceByIDHandler))
 
 	t.Server.HTTP.Handle(path.Join("/querier", addHTTPAPIPrefix(&t.cfg, apiPathTraces)), tracesHandler)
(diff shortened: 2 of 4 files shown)


kvrhdn:
This I still want to do (but I got to go now 🏃🏻 )

- [ ] Update CHANGELOG
- [ ] Add e2e test if possible (this will require large traces)
- [ ] The response size logged in frontend.go is not correct anymore as gzip compression is applied after the logging

cmd/tempo/app/http.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/NYTimes/gziphandler"
+	"github.com/weaveworks/common/middleware"
+)
+
+func httpGzipMiddleware() middleware.Interface {
+	return middleware.Func(func(handler http.Handler) http.Handler {
+		gzipHandler := gziphandler.GzipHandler(handler)
+
+		return http.HandlerFunc(func(writer http.ResponseWriter, r *http.Request) {
+			// do not gzip the response when requesting protobuf as this will mess up the encoding
```
joe-elliott, line 15: that's weird, why?

integration/e2e/util_test.go
```diff
@@ -0,0 +1,55 @@
...
+		})
+	}
+	return &thrift.Batch{Spans: spans}
+}
+
+func extractHexID(batch *thrift.Batch) string {
+	return fmt.Sprintf("%016x%016x", batch.Spans[0].TraceIdHigh, batch.Spans[0].TraceIdLow)
+}
+
+func assertTrace(t *testing.T, reader io.Reader, expectedBatches int, expectedName string) {
```
zalegrala, line 49: To compare traces, we have https://github.com/grafana/tempo/blob/main/cmd/tempo-vulture/main.go#L317, but perhaps that could move somewhere like `util/` and be common for comparing trace?

PR #896, opened 2021-08-19
Use single configmap for overrides

### LINKED ISSUE #875: jsonnet/microservices: create separate configmap for overrides
The current microservices Jsonnet library creates a single configmap per component which contains both the Tempo config and the overrides. This spreads out the overrides configuration, which is updated more often.

We'd like to combine all the overrides in a single configmap which is mounted by the components that need it (distributors, ingesters and compactors). This way only this configmap needs to be managed.

Create a new configmap (e.g. `tempo-overrides`?) that contains all the overrides information.

We can also not do this ofc.

**Additional context**

For example: this if how the distributor configmap is created right now:
https://github.com/grafana/tempo/blob/f822caa9b14a017b788e4c10bb7e9d13646ade38/operations/jsonnet/microservices/configmap.libsonnet#L102-L111

And this is how it is mounted:
https://github.com/grafana/tempo/blob/f822caa9b14a017b788e4c10bb7e9d13646ade38/operations/jsonnet/microservices/distributor.libsonnet#L47-L49

If we change how the overrides.yaml is mounted, we'll have to update the Tempo config as well:
https://github.com/grafana/tempo/blob/f822caa9b14a017b788e4c10bb7e9d13646ade38/operations/jsonnet/microservices/configmap.libsonnet#L44-L46

kavirajk:
**What this PR does**:
Add support to tempo workloads to read `overrides` values from single configmap in microservice mode.

Fixes #875

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/operations/jsonnet/microservices/config.libsonnet
+++ b/operations/jsonnet/microservices/config.libsonnet
@@ -97,6 +97,7 @@
     backend: error 'Must specify a backend',  // gcs|s3
     bucket: error 'Must specify a bucket',
 
+    // `use_overrides_configmap` enables the tempo workloads read `overrides.yaml` from single configmap
     // Eventually, once we enable it everywhere, we remove this and default will be true.
     use_overrides_configmap: false,
     overrides_configmap_name: 'overrides',
--- a/operations/jsonnet/microservices/config.libsonnet
+++ b/operations/jsonnet/microservices/config.libsonnet
@@ -97,6 +97,10 @@
     backend: error 'Must specify a backend',  // gcs|s3
     bucket: error 'Must specify a bucket',
 
+    // Eventually, once we enable it everywhere, we remove this and default will be true.
+    use_overrides_configmap: false,
+    overrides_configmap_name: 'overrides',
+
     overrides+:: {
       super_user:: {
         max_traces_per_user: 100000,
--- a/operations/jsonnet/microservices/config.libsonnet
+++ b/operations/jsonnet/microservices/config.libsonnet
@@ -97,11 +97,7 @@
     backend: error 'Must specify a backend',  // gcs|s3
     bucket: error 'Must specify a bucket',
 
-    // `use_overrides_configmap` enables the tempo workloads read `overrides.yaml` from single configmap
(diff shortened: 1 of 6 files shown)


operations/jsonnet/microservices/config.libsonnet
```diff
@@ -97,6 +97,11 @@
     backend: error 'Must specify a backend',  // gcs|s3
     bucket: error 'Must specify a bucket',
 
+    // `use_overrides_configmap` enables the tempo workloads read `overrides.yaml` from single configmap
+    // Eventually, once we enable it everywhere, we remove this and default will be true.
```
kvrhdn, line 101: We don't need a feature flag to roll this out gradually. I'm actually fine with making this the new default. This change doesn't cause more disruption than a normal update does (we roll all components anyway) and using a separate configmap does not have any disadvantage afaik.

kavirajk: I would highly recommend using the flag, Just to avoid any surprises!. It's very straight forward and easy to cleanup once we are confident and changes are in all envs :)

joe-elliott: +1 to @kvrhdn 's thoughts. Let's just swap over instead of all the if/else logic in jsonnet.

kavirajk: Updated. No conditional flag! :)

kvrhdn (commented):
Thanks for this @kavirajk, splitting up the overrides yaml works well 🙂 

There is only an issue when you modify `$.config.overrides_configmap_name`, this change isn't propagated correctly and some code relies on the default value.

operations/jsonnet/microservices/config.libsonnet
```diff
@@ -97,6 +97,7 @@
     backend: error 'Must specify a backend',  // gcs|s3
     bucket: error 'Must specify a bucket',
 
+    overrides_configmap_name: 'overrides',
```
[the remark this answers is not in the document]
kavirajk: Sure. I don't mind :)

kvrhdn (approved):
LGTM! Thanks again 🙂

PR #898, opened 2021-08-19
Vulture Improvements

### LINKED ISSUE #791: Extend vulture to generate long-running traces
Vulture currently generates random traces and sends them in a single batch to Tempo. To improve our test coverage we'd also like to send long-running traces, i.e. traces that will span multiple blocks in Tempo.

The ingester considers a trace complete after a configurable idle period ([by default 30s](https://grafana.com/docs/tempo/latest/configuration/#ingester)). If vulture can send spans with a longer delay between them they should end up in different blocks and Tempo will have to do additional work to combine all spans into a single trace when requested.

Vulture is currently stateless and ideally it stays stateless.

The following should probably be configurable:
- a setting to enable/disable long-running traces
- the delay between send requests

**Additional context**

zalegrala:
**What this PR does**:
Refactor vulture to send traces using the oltp exporter.
Add option to send traces over a longer period of time, in addition to short traces.
Begin comparing the trace received from a query to the trace we sent.

Fixes #791

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


cmd/tempo-vulture/main.go
```diff
@@ -226,43 +104,66 @@ func generateRandomInt(min int64, max int64) int64 {
...
 
 	// iterate through
 	if hasMissingSpans(trace) {
-		logger.Error("trace has missing spans")
+		log.Error("trace has missing spans")
 		tm.missingSpans++
 	}
 
+	count := spanCount(trace)
+	if count != complete.spanCount {
```
joe-elliott, line 143: excellent improvement and we will merge as is, but the long term vision is to regenerate the entire trace from the seed and check all events/attributes/spans.  i.e. revalidate the entire trace.

cmd/tempo-vulture/vulture.go
```diff
@@ -0,0 +1,187 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/trace"
+	"go.uber.org/zap"
+)
+
+// Vulture is used to send traces to Tempo, and then read those traces back out to verify that the service is operating correctly.
+type Vulture struct {
+	completeChan     chan completeTrace
```
joe-elliott, line 17: the original design was to seed the random number generator with known values anchored on timestamps. this way vulture can query for any trace in the past 2 weeks and know exactly what it should look like without having to keep anything in memory. i'd prefer keeping that design unless there are strong reasons to do otherwise.

cmd/tempo-vulture/main.go
zalegrala: One thing I couldn't figure out was why attributes weren't being sent.  There is this notion of "recording" or not, and the otlp code seems to skip doing anything with attributes if the exporter isn't recording.  If we can figure out how to enable that bit, then checking the attributes should be extensible from the structure that is in place now.

cmd/tempo-vulture/vulture.go
zalegrala: It seemed like there was a race happening where we'd try to discover a past time stamp that was never sent and then check for it, and get a 404 back.  I like Vulture being stateless, in that nothing gets persisted between restarts.  Is there a middle ground here between trying to rediscover an old timestamp and keeping track of every trace id?  We could perhaps continue to query traces that had been sent in a sampled way if checking for older traces is important.  I definitely didn't factor that in.  Though, the long running traces should ensure that traces pull from multiple blocks.

cmd/tempo-vulture/main.go
zalegrala: There is another way to get the attributes on the trace.  I'll see if I can get something in place to include more detail that we can check for in the trace.

zalegrala: I've updated this to inject some randomness into the attributes and ensure they get validated on the query.

cmd/tempo-vulture/vulture.go
joe-elliott: Generating and then validating a trace from a past timestamp is stateless :) which is one reason I like it so much. It's easy to validate a complete trace that was pushed last week without having to store it somewhere. I'd really like to keep it. 

I'm willing to endure a small number of 404s to keep this design. 

> We could perhaps continue to query traces that had been sent in a sampled way if checking for older traces is important.

Querying older traces is very important b/c it ensures that nothing changes as the trace is cut, flushed to the backend and repeatedly compacted over 2 weeks.

zalegrala:
I'll close this out and open from my fork.

PR #903, opened 2021-08-23
[cli] Add command to generate index/bloom from a data file

### LINKED ISSUE #894: Add a cli command to generate bloom & index for a block from data file
As part of the Production Readiness Review of Tempo, we would like to add a cli command that can generate the bloom filter & index from the data file of a block in case of GCS file corruptions/deletions.

A single missing bloom filter for a block can 5xx _all_ queries, so this makes it all the more important.

annanay25:
**What this PR does**:
- Adds a command to generate the index from a data file. To test, run:
```
go run ./cmd/tempo-cli gen index --backend=local --bucket=./cmd/tempo-cli/test-data/ single-tenant b18beca6-4d7f-4464-9f72-f343e688a4a0
```

The new index file will be created in the `test-data/single-tenant/b18beca6-4d7f-4464-9f72-f343e688a4a0` folder.

Diff the newly generated index with the `index-copy` present in the same folder. There should be no differences.

- This PR also adds some diagrams that help with understanding the encoding package. The drawio XML file is also included. The expectation is that the xml and png file will be version controlled but it is not on the official website docs and so it doesn't have to be updated with every PR. 

Tempo encoding

Fixes #894 

**Checklist**
- [x] Tests updated
  - [x] Needs some form of runtime validation as well
- [x] Documentation added


cmd/tempo-cli/cmd-gen-bloom-index.go
```diff
@@ -0,0 +1,122 @@
...
+		return err
+	}
+
+	meta, err := r.BlockMeta(context.TODO(), blockID, cmd.TenantID)
+	if err != nil {
+		return err
+	}
+
+	// replay file to extract records
+	records, warning, err := ReplayBlockAndGetRecords(meta, "./cmd/tempo-cli/test-data/" + cmd.TenantID + "/" + cmd.BlockID + "/data")
```
mdisibio, line 95: Can this use `cmd.backendOptions.Bucket` for the local path instead of `./cmd/tempo-cli/test-data/` ?

cmd/tempo-cli/main.go
```diff
@@ -44,6 +44,10 @@ var cli struct {
 		Index viewIndexCmd `cmd:"" help:"View contents of block index"`
 	} `cmd:""`
 
+	Gen struct {
+		GenIndex genIndexCmd `cmd:"" help:"Generate bloom/index for a block"`
```
mdisibio, line 48: Let's rename this to just `Index` to remove the repetition of the command line.  Then it will be `tempo-cli gen index`

docs/tempo/website/operations/tempo_cli.md
```diff
@@ -199,3 +199,12 @@ Arguments:
...
 tempo-cli view index -c ./tempo.yaml single-tenant ca314fba-efec-4852-ba3f-8d2b0bbf69f1
 ```
+
+## Generate Index
+
+To generate the index/bloom for a block if the files were deleted/corrupted.
+
+**Example:**
+```bash
+tempo-cli gen gen-index -c ./tempo.yaml single-tenant b18beca6-4d7f-4464-9f72-f343e688a4a0
```
mdisibio, line 209: See other comment about just `tempo-cli gen index`

cmd/tempo-cli/cmd-gen-bloom-index.go
```diff
@@ -0,0 +1,122 @@
...
+		recordID := append([]byte(nil), id...)
+		records = append(records, common.Record{
+			ID:     recordID,
+			Start:  currentOffset,
+			Length: pageLen,
+		})
+		currentOffset += uint64(pageLen)
+	}
+
+	common.SortRecords(records)
```
mdisibio, line 73: Is sorting necessary?  They should already be in order from the data file, and if they aren't, wondering if that needs more consideration for side effects.

cmd/tempo-cli/cmd-gen-bloom-index.go
```diff
@@ -0,0 +1,122 @@
...
+		id, _, err := objectReader.UnmarshalObjectFromReader(reader)
+		if err != nil {
+			warning = err
+			break
+		}
+
+		// make a copy so we don't hold onto the iterator buffer
+		recordID := append([]byte(nil), id...)
+		records = append(records, common.Record{
+			ID:     recordID,
```
mdisibio, line 66: Need another set of 👀 to think through this.   This is using the first trace ID from the page for the index entry, but I think AppenderBuffered uses the last ID per page:  https://github.com/grafana/tempo/blob/main/tempodb/encoding/appender_buffered.go#L52   Thoughts?

cmd/tempo-cli/cmd-gen-bloom-index.go
annanay25: I agree with you. We shouldn't sort here. Just return in-order. Records should be in order already.

cmd/tempo-cli/cmd-gen-bloom-index.go
annanay25: In the process of implementing an Iterator for the page. Will update tomorrow.

cmd/tempo-cli/cmd-gen-bloom-index.go
```diff
@@ -0,0 +1,122 @@
...
+			warning = err
+			break
+		}
+
+		reader := bytes.NewReader(buffer)
+		id, _, err := objectReader.UnmarshalObjectFromReader(reader)
+		if err != nil {
+			warning = err
+			break
+		}
```
joe-elliott, line 61: i believe this needs to loop through all objects in the page? it seems to only get the first. basically loop until `UnmarshalObjectFromReader` returns io.EOF.

To test we should create a block using the StreamingBlock, replay it using this code and then compare the index/blooms.

cmd/tempo-cli/cmd-gen-bloom-index.go
joe-elliott: Yes, agree with Marty. The index entry is the largest id on the page.

cmd/tempo-cli/cmd-gen-bloom-index.go
annanay25: > To test we should create a block using the StreamingBlock, replay it using this code and then compare the index/blooms.

I don't want to do this because it will require that we download the block from the remote backend. It should be easy to write a verification utility that iterates through the index and ensures the records are lined up accordingly. WDYT?

joe-elliott: No, I mean to just create use a StreamingBlock to create a block, write it to a local backend and then use this code to recreate the index/bloom and make sure it's correct.

This test does that but then uses a BackendBlock to read the streaming block:
https://github.com/grafana/tempo/blob/main/tempodb/encoding/streaming_block_test.go#L150

annanay25:
Added a very basic index verifier and I think the first stage of should include just the local backend. As follow ups, we can implement a remote backend reader which will involve implementing an `io.Reader` for the S3 backend.

joe-elliott:
Also, let's improve these docs:

https://github.com/grafana/tempo/blob/main/docs/tempo/website/troubleshooting/bad-blocks.md

now that your PR adds a way to fix blocks!

joe-elliott (approved):

PR #904, opened 2021-08-23
Refactor Vulture for query verification

zalegrala:
This chance modifies the randomization to allow for the query response attributes to be verified against the deterministic pseudo-random attribute generation.

**Checklist**
- [x] Tests updated


cmd/tempo-vulture/main.go
```diff
@@ -127,7 +127,7 @@ func main() {
 			}
 
 			// pick past interval and re-generate trace
-			seed := (generateRandomInt(startTime, currentTime) / interval) * interval
+			seed := (generateRandomInt(startTime, currentTime, rand.New(rand.NewSource(1))) / interval) * interval
 			rand.Seed(seed)
```
mdisibio, line 131: Should this be changed to not use the global rand as well?

zalegrala: I've updated the approach here.  The global rand is still used, but only to select from a list of known time slots, so I think it is acceptable.

cmd/tempo-vulture/main.go
```diff
@@ -162,79 +198,84 @@ func newJaegerGRPCClient(endpoint string) (*jaeger_grpc.Reporter, error) {
...
 		})
 	}
 	return &thrift.Batch{Spans: spans}
 }
 
-func generateRandomInt(min int64, max int64) int64 {
-	number := min + rand.Int63n(max-min)
+func generateRandomInt(min int64, max int64, r *rand.Rand) int64 {
+	number := min + r.Int63n(max-min)
 	if number == min {
```
joe-elliott, line 263: this is funny, can we clean this up by just adding 1 to the requested min?

cmd/tempo-vulture/main.go
```diff
@@ -148,12 +143,53 @@ func main() {
 	log.Fatal(http.ListenAndServe(prometheusListenAddress, nil))
 }
 
+func intervalsBetween(start, stop time.Time, interval time.Duration) []time.Time {
+	if stop.Before(start) {
+		return nil
+	}
+
+	intervals := []time.Time{start}
+	next := start.Add(interval)
```
joe-elliott, line 152: let's align the times to multiples of the interval.

cmd/tempo-vulture/main.go
```diff
@@ -148,12 +143,53 @@ func main() {
...
+
+	for next.Before(stop) {
+		intervals = append(intervals, next)
+		next = next.Add(interval)
+	}
+
+	return intervals
+}
+
+func trimOutdatedIntervals(intervals []time.Time, lookBehind time.Duration) []time.Time {
```
joe-elliott, line 162: would be clearer to just choose the starttime and use the above method. i don't think we need this method.

cmd/tempo-vulture/main.go
```diff
@@ -141,19 +142,53 @@ func main() {
...
 	log.Fatal(http.ListenAndServe(prometheusListenAddress, nil))
 }
 
+func intervalsBetween(start, stop time.Time, interval time.Duration, retention time.Duration) []time.Time {
+	if stop.Before(start) {
+		return nil
+	}
+
+	intervals := []time.Time{start}
+	next := start.Round(interval)
```
mdisibio, line 159: time.Round will round down for the first half of the interval.  I think this means that it can choose a `next` that is before `start`.  Thoughts?

cmd/tempo-vulture/main.go
```diff
@@ -299,3 +360,44 @@ func hasMissingSpans(t *tempopb.Trace) bool {
...
+
+		// Due to the several transforms above, some manual mangling is required to
+		// get the parentSpanID to match.  In the case of an empty []byte in place
+		// for the ParentSpanId, we set to nil here to ensure that the final result
+		// matches the json.Unmarshal value when tempo is queried.
+		for ib, b := range t.Batches {
+			for il, l := range b.InstrumentationLibrarySpans {
+				for is, s := range l.Spans {
+					if len(s.GetParentSpanId()) == 0 {
+						t.Batches[ib].InstrumentationLibrarySpans[il].Spans[is].ParentSpanId = nil
```
mdisibio, line 393: Not sure I understand the need for array indexing here.  Could this be replaced with just `s.ParentSpanId = nil` ?

cmd/tempo-vulture/main.go
```diff
@@ -141,19 +142,53 @@ func main() {
...
 			metricTracesErrors.WithLabelValues("missingspans").Add(float64(metrics.missingSpans))
+			metricTracesErrors.WithLabelValues("incorrectresult").Add(float64(metrics.incorrectResult))
 		}
 	}()
 
 	http.Handle(prometheusPath, promhttp.Handler())
 	log.Fatal(http.ListenAndServe(prometheusListenAddress, nil))
 }
 
+func intervalsBetween(start, stop time.Time, interval time.Duration, retention time.Duration) []time.Time {
```
mdisibio, line 153: It doesn't look like there is a need for this function to return all of the possible intervals, since only 1 is chosen at random (and the rest thrown away).   It seems more readable to go ahead and have this method do that encapsulate that functionality, change the name/signature to something like 
```go
choosePastTimestamp(start, stop time.Time, interval, retention time.Duration) (newStart time.Time, ts time.Time)
```   

Call it like 
```go
startTime, seed := intervals := intervalsBetween(startTime, now, interval, tempoRetentionDuration)
````

File: cmd/tempo-vulture/main.go (hunk 5 above)
Comment (zalegrala) [reply cmd/tempo-vulture/main.go:159]: Ah yes, I've just updated to include only time windows after the start.  How's that look to you now?

File: cmd/tempo-vulture/main.go (hunk 6 above)
Comment (zalegrala) [reply cmd/tempo-vulture/main.go:395]: I think if you are in the loop, you want to avoid referencing variables inside the loop as well when you intent to modify the variable outisde the loop, so here we take the index of each of the loop and make the change there.

File: cmd/tempo-vulture/main.go (hunk 7 above)
Comment (zalegrala) [reply cmd/tempo-vulture/main.go:153]: Sure, I like that.  Return both items that we need.  I'll make an update.

Comment (zalegrala) [reply cmd/tempo-vulture/main.go:156]: I've moved the logic into a separate method, but if you'd prefer a single method I can adjust for that as well.  I'll need to update the tests to make sure this is consistent if we want to make it a single method.

**Review by joe-elliott — APPROVED** [review]
nice!

File: cmd/tempo-vulture/main.go [hunk 8]
```diff
@@ -252,17 +285,42 @@ func queryTempoAndAnalyze(baseURL string, seed int64, traceID string) (traceMetr
...
+				)
+			}
+		}
+		return tm, nil
 	}
 
 	return tm, nil
 }
 
+func equalTraces(a, b *tempopb.Trace) bool {
```
mdisibio, line 317: Can we add a unit test for this?  Just to know for sure that it's reliable.

cmd/tempo-vulture/main.go
zalegrala: Ah good point.  I usually just modify the original out of habit.  I'll make the change.

mdisibio (approved):

PR #905, opened 2021-08-24
Enhancement: make s3 backend readError logic more robust

wei840222:
for some s3 compatible storage backend, it may has differnt error messsage when object not exist.
I think using minio.error code for judge the object is not exist is more robust.

### DIFF
--- a/tempodb/backend/s3/s3.go
+++ b/tempodb/backend/s3/s3.go
@@ -26,8 +26,9 @@ import (
 )
 
 const (
-	s3KeyDoesNotExist  = "The specified key does not exist."
-	uptoHedgedRequests = 2
+	s3KeyDoesNotExist     = "The specified key does not exist."
+	s3KeyDoesNotExistCode = "NoSuchKey"
+	uptoHedgedRequests    = 2
 )
 
 // readerWriter can read/write from an s3 backend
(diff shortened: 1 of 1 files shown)


tempodb/backend/s3/s3.go
```diff
@@ -264,6 +265,8 @@ func (rw *readerWriter) readAllWithObjInfo(ctx context.Context, name string) ([]
 	reader, info, _, err := rw.hedgedCore.GetObject(ctx, rw.cfg.Bucket, name, minio.GetObjectOptions{})
 	if err != nil && err.Error() == s3KeyDoesNotExist {
 		return nil, minio.ObjectInfo{}, backend.ErrDoesNotExist
+	} else if err != nil && minio.ToErrorResponse(err).Code == s3KeyDoesNotExistCode {
```
mapno, line 268: Isn't line 266 making the same check?

```go
	if err != nil && err.Error() == s3KeyDoesNotExist {
```

mapno (commented):
~Hi! Thanks for the contribution. Is there any difference between the current check `err.Error() == s3KeyDoesNotExist` and checking the error code for the same 'no key' error? It's being checked in both modified methods.~

mapno:
Sorry, I re-read your description and saw the explanation `for some s3 compatible storage backend`. Then, would it make sense to just check the code?

wei840222:
I think it is make sense to just check the code.
If possible, I can submit another commit to make corrections.

wei840222:
changed

tempodb/backend/s3/s3.go
```diff
@@ -26,8 +26,8 @@ import (
 )
 
 const (
-	s3KeyDoesNotExist  = "The specified key does not exist."
-	uptoHedgedRequests = 2
+	s3KeyDoesNotExistCode = "NoSuchKey"
```
joe-elliott, line 29: instead of redefining this can we reference the exported error code in the s3 package?

https://github.com/aws/aws-sdk-go/blob/main/service/s3/errors.go#L40

wei840222: @joe-elliott hi, changed

PR #906, opened 2021-08-24
Make overrides configmap block name consistent

kavirajk:
**What this PR does**:
It edit tempo microservices jsonnet and makes the `overrides_config` block name consistent with Loki and Cortex

1. Loki - https://github.com/grafana/loki/blob/main/production/ksonnet/loki/overrides.libsonnet#L21
2. Cortex - https://github.com/grafana/cortex-jsonnet/blob/main/cortex/config.libsonnet#L469

Fixes #NA

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


annanay25 (commented):
Is the name going to be used in a script/automation setting? Tempo has a consistent naming convention in the jsonnets at the moment, all configmaps have the `_configmap` suffix. Would be nice if we could retain that, but its ok to make the change if needed

kavirajk:
@annanay25 Yea unfortunately I think we have to keep it `_config`. It's just that entitlement system expects this block name inside the jsonnet. Also NOTE: the name of final configmap on the k8 will still be `tempo-overrides` (no problem with that). 

But I totally understand and agree with your naming consistency concern within tempo. I can happily change it to `_configmap` everywhere(including LOki and Cortex) once the migration is complete. :)

annanay25 (approved):

PR #919, opened 2021-08-30
Drop support for v0 and v1 blocks

joe-elliott:
**What this PR does**:
Drops support for v0 and v1 blocks and consolidates all code into v2.

**TODO**
- [x] Investigate Performance regression indicated by benchmarks
- [x] Figure out what to do with buffers

**Checklist**
- [x] Tests updated
- [ ] Documentation added


joe-elliott:
This is ready to go, but leaving draft status up b/c I want to merge #806 first. There will be some small conflicts.

tempodb/encoding/v2/data_reader.go
```diff
@@ -23,41 +27,97 @@ var constDataHeader = &dataHeader{}
...
+		if err != nil {
+			return nil, nil, err
+		}
 	}
 
 	return pagesBuffer, buffer, nil
 }
 
 func (r *dataReader) Close() {
-	r.dataReader.Close()
```
zalegrala, line 60: Do we still need to do something here to close this out?

joe-elliott: Nope, b/c there is no longer an encapsulated data reader.

zalegrala (approved):

tempodb/encoding/v2/data_reader.go
```diff
@@ -23,41 +27,97 @@ var constDataHeader = &dataHeader{}
...
+		}
+
+		compressedPages = append(compressedPages, page.data)
+	}
+
+	// prepare pagesBuffer
+	if cap(pagesBuffer) < len(compressedPages) {
+		// extend pagesBuffer
+		diff := len(compressedPages) - cap(pagesBuffer)
+		pagesBuffer = append(pagesBuffer[:cap(pagesBuffer)], make([][]byte, diff)...)
```
annanay25, line 96: `append()` will copy over all the items from old to new array right? and the resultant pagesBuffer array might have a cap larger than `len(compressedPages)`, worth just creating a new array with `cap(pagesBuffer) = len(compressedPages)` ? The reuse is primarily useful when `cap(pagesBuffer) > len(compressedPages)` I guess

annanay25 (approved):
LGTM, just a minor non-blocking nit

tempodb/encoding/v2/data_reader.go
joe-elliott: > and the resultant pagesBuffer array might have a cap larger than len(compressedPages)
I believe after this line len(pagesBuffer) = cap(pagesBuffer) = len(compressedPages)

> worth just creating a new array with cap(pagesBuffer) = len(compressedPages)
by using whatever currently exists in pagesBuffer we reduce byte slice allocations. after this line the first N entries in pagesBuffer already have allocated byte slices that can be reused in `ReadAllWIthBuffer`. if we recreated the slice we would lose that.

PR #936, opened 2021-09-07
Add runtime config handler

### LINKED ISSUE #893: Add a runtime config handler

I often want to access the tenant limits configured on a tempo instance through an HTTP API.

Add a runtime config handler that provides information from the runtimeCfg manager.

Currently the runtimeCfg manager is implemented as a subservice of the Overrides module. 
https://github.com/grafana/tempo/blob/5ad92c9c227147ac63f66b487b2a89172318eb77/modules/overrides/overrides.go#L61-L71 

We should split the runtimeCfg manager as a separate service and register an API endpoint for it. Cortex implementation for reference:
https://github.com/cortexproject/cortex/blob/fbf067776f612ec33e7228430d4670ef566b57ec/pkg/cortex/modules.go#L150-L178 

-

**Additional context**

-

mapno:
**What this PR does**:

Adds a runtime config handler at `/runtime_config`

Fixes #893

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/overrides/overrides.go
+++ b/modules/overrides/overrides.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net/http"
 	"time"
 
+	"github.com/cortexproject/cortex/pkg/util"
 	"github.com/cortexproject/cortex/pkg/util/log"
 	"github.com/grafana/dskit/runtimeconfig"
 	"github.com/grafana/dskit/services"
@@ -37,13 +39,18 @@ func loadPerTenantOverrides(r io.Reader) (interface{}, error) {
 	return overrides, nil
 }
 
+type Config struct {
+	Defaults           *Limits             `yaml:"defaults"`
+	PerTenantOverrides *perTenantOverrides `yaml:"overrides,omitempty"`
+}
+
 // Overrides periodically fetch a set of per-user overrides, and provides convenience
 // functions for fetching the correct value.
 type Overrides struct {
 	services.Service
 
-	defaultLimits *Limits
-	tenantLimits  TenantLimits
+	config       *Config
+	tenantLimits TenantLimits
 
 	// Manager for subservices
 	subservices        *services.Manager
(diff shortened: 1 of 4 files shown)


modules/overrides/overrides.go
```diff
@@ -36,13 +38,18 @@ func loadPerTenantOverrides(r io.Reader) (interface{}, error) {
 	return overrides, nil
 }
 
+type Config struct {
+	Defaults           *Limits             `yaml:"defaults"`
+	PerTenantOverrides *perTenantOverrides `yaml:"overrides,omitempty"`
```
annanay25, line 43: Not a detailed review, but this seems to be holding information that's present in `tenantLimits` already. Can we club the two somehow? Maybe something like this:
```
type Overrides struct {
	services.Service

	defaultLimits *Limits
	tenantLimits  perTenantOverrides

	// Manager for subservices
	subservices        *services.Manager
	subservicesWatcher *services.FailureWatcher
}
```

mapno: That's a good point, addressed in 47c4af1cc5a74580d498de08ec8fb26ee3adfffb. Thanks

modules/overrides/overrides.go
```diff
@@ -123,6 +136,65 @@ func (o *Overrides) stopping(_ error) error {
...
+			Defaults:           o.defaultLimits,
+			PerTenantOverrides: tenantOverrides,
+		}
+		switch r.URL.Query().Get("mode") {
+		case "diff":
+			// Default runtime config is just empty struct, but to make diff work,
+			// we set defaultLimits for every tenant that exists in runtime config.
+			defaultCfg := perTenantOverrides{TenantLimits: map[string]*Limits{}}
+			defaultCfg.TenantLimits = map[string]*Limits{}
+			for k, v := range tenantOverrides.TenantLimits {
```
annanay25, line 168: I think this will result in null pointer dereference if `o.tenantOverrides == nil`

annanay25 (approved):
Looking good, nice work @mapno! Left a minor comment.

modules/overrides/overrides.go
mapno: I think it's fine since we're instantiating a zero value in line 153 and only assign a value if `o.tenantOverrides() != nil`

PR #939, opened 2021-09-07
Delete completed local blocks when replaying wal

### LINKED ISSUE #937: "Magic number error"s after changing compression format on the ingesters
**Describe the bug**
Tempo was running stable for some time on zstd compression.  We decided to investigate snappy compression, switched the storage/trace/block/encoding field to 'snappy' and restarted ingesters.  Afterwards, tempo started encountering many 'magic number' errors. This error is from a compactor but other components had the same error:
```
level=error ts=2021-09-07T15:57:10.601766281Z caller=compactor.go:110 msg="error during compaction cycle" err="error iterating input blocks: error reading objects for records, blockID: 4644fe19-1bc4-495f-99da-be0d1831543a, err: invalid input: magic number mismatch"
```

Inspecting the block, it seems like the `data` file was written with snappy as expected, but the meta.json still contains `zstd`.  This lines up with the error message, which seems to be from zstd checking and failing for its magic number prefix.   

A couple more things to note about the cli output below:
(1) Our ingesters cut blocks fairly consistently, and on zstd were ~350MB, and this block is ~600MB.  This leads us to believe that the `data` file is indeed snappy.
(2) The `size` is also wrong.  It is the ~350MB size from zstd and not the true data file size anymore.

```
$ gsutil ls -l gs://<mybucket>/1/4644fe19-1bc4-495f-99da-be0d1831543a/
    100024  2021-09-07T14:38:40Z  gs://<mybucket>/1/4644fe19-1bc4-495f-99da-be0d1831543a/bloom-0
    100024  2021-09-07T14:38:40Z  gs://<mybucket>/1/4644fe19-1bc4-495f-99da-be0d1831543a/bloom-1
    100024  2021-09-07T14:38:40Z  gs://<mybucket>/1/4644fe19-1bc4-495f-99da-be0d1831543a/bloom-2
    100024  2021-09-07T14:38:41Z  gs://<mybucket>/1/4644fe19-1bc4-495f-99da-be0d1831543a/bloom-3
 611997809  2021-09-07T14:38:40Z  gs://<mybucket>/1/4644fe19-1bc4-495f-99da-be0d1831543a/data
    256000  2021-09-07T14:38:41Z  gs://<mybucket>/1/4644fe19-1bc4-495f-99da-be0d1831543a/index
       393  2021-09-07T14:38:41Z  gs://<mybucket>/1/4644fe19-1bc4-495f-99da-be0d1831543a/meta.json
TOTAL: 7 objects, 612654298 bytes (584.27 MiB)

$ gsutil cp gs://<mybucket>/1/4644fe19-1bc4-495f-99da-be0d1831543a/meta.json .
Copying gs://<mybucket>/1/4644fe19-1bc4-495f-99da-be0d1831543a/meta.json...
/ [1 files][  393.0 B/  393.0 B]                                                
Operation completed over 1 objects/393.0 B.       
                               
$ cat meta.json 
{"format":"v2","blockID":"4644fe19-1bc4-495f-99da-be0d1831543a","minID":"AAAAAAAAAAAAAAXhfU3CWw==","maxID":"7V8evRot612xwrcHbnhCcQ==",
"tenantID":"1","startTime":"2021-09-07T14:34:36.182817282Z","endTime":"2021-09-07T14:37:06.427198961Z",
"totalObjects":268962,"size":349737066,"compactionLevel":0,
"encoding":"zstd","indexPageSize":256000,"totalRecords":2329,"dataEncoding":"v1","bloomShards":4}
```

**To Reproduce**
1. Start Tempo r7
2. Create blocks with zstd encoding
3. Change configmap to snappy
4. Deploy r8 (which includes an ingester update and restart)
5. Perform queries or let compactors run

**Expected behavior**
It should be fine to switch an ingester's encoding between restarts. Blocks written prior should keep the previous encoding, and new blocks should be written with the new encoding.

**Environment:**
 - Infrastructure: Kubernetes
 - Deployment tool: jsonnet

**Additional context**
Here are ingester logs for this block.  After the restart it both replayed the WAL and rediscovered the complete block.  It doesn't seem like it should have done both... only one needs to be done. Maybe the error is an edge case in this replay/rediscover logic.

```
2021-09-07 14:37:06 | level=info ts=2021-09-07T14:37:06.427374977Z caller=flush.go:157 msg="head block cut. enqueueing flush op" userid=1 block=4644fe19-1bc4-495f-99da-be0d1831543a
2021-09-07 14:37:08 | level=info ts=2021-09-07T14:37:08.833570107Z caller=flush.go:232 msg="completing block" userid=1 blockID=4644fe19-1bc4-495f-99da-be0d1831543a
2021-09-07 14:38:17 | level=info ts=2021-09-07T14:38:17.626684172Z caller=wal.go:93 msg="beginning replay" file=4644fe19-1bc4-495f-99da-be0d1831543a:1:v2:snappy:v1 size=748082090
2021-09-07 14:38:21 | level=info ts=2021-09-07T14:38:21.974282518Z caller=wal.go:120 msg="replay complete" file=4644fe19-1bc4-495f-99da-be0d1831543a:1:v2:snappy:v1 duration=4.347593985s
2021-09-07 14:38:21 | level=info ts=2021-09-07T14:38:21.987396049Z caller=instance.go:595 msg="reloaded local block" tenantID=1 block=4644fe19-1bc4-495f-99da-be0d1831543a flushed=0001-01-01T00:00:00Z
2021-09-07 14:38:24 | level=info ts=2021-09-07T14:38:24.280352444Z caller=flush.go:232 msg="completing block" userid=1 blockID=4644fe19-1bc4-495f-99da-be0d1831543a
2021-09-07 14:38:31 | level=info ts=2021-09-07T14:38:31.560074186Z caller=flush.go:288 msg="flushing block" userid=1 block=4644fe19-1bc4-495f-99da-be0d1831543a
2021-09-07 14:38:39 | level=info ts=2021-09-07T14:38:39.082371529Z caller=flush.go:239 msg="block completed" userid=1 blockID=4644fe19-1bc4-495f-99da-be0d1831543a duration=14.802010543s
```

mdisibio:
**What this PR does**:
This PR fixes #937 by deleting locally complete blocks during wal replay, and another race condition on ingester startup.  Here is a description of the situation and why this should fix it.

1. An ingester may restart or crash while completing the wal file. It can happen after creating the new local block, but before deleting the wal file.
2. On restart the ingester replays wal, and rediscovers local blocks.  Both exist it ends up with 2 queue entries instead of 1.  It will both recomplete the wal and reflush the local block.  These occur independently and there are no consistency checks while processing the queue entry.
3. This is a race condition in multiple ways. 
  a. The wal complete is in progress, writing over top of `data` when it is flushed to the backend. The flush will reach EOF and assume all went well. This creates a partially written block in the backend.
  b. The wal complete can entirely finish before the flush.   It appends another entry into `completedBlocks`.  When the flush occurs it reads the first entry from `completedBlocks` and saves it as the meta data. This is the actual condition for `magic number` errors and incorrect meta. It uses the meta as rediscovered in step 2, which may be a different encoding, etc.
  c. The wal complete can start on a new block before rediscoverLocalBlocks begins. rediscoverLocalBlocks sees a bad block (missing meta) and deletes it.  I think this ends up in a situation similar to (a) but different causes. This is handled by not rediscovering the local block is there is still a wal for it.

A couple more notes:
* Enabling search increased the risk of these errors because it added more work between creating the local block and deleting the wal in step 1.
* We've also seen issues with partially flushed blocks like 3a and 3b, but afaik there isn't an issue tracking them.  

Fixes #937 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/ingester/ingester.go
+++ b/modules/ingester/ingester.go
@@ -343,6 +343,17 @@ func (i *Ingester) replayWal() error {
 			return err
 		}
 
+		// Delete anything remaining for the completed version of this
+		// wal block. This handles the case where a wal file is partially
+		// or fully completed to the local store, but the wal file wasn't
+		// deleted (because it wa rescanned above). This can happen for reasons
+		// such as a crash or restart. In this situation we err on the side of
+		// caution and replay the wal block.
+		err = instance.local.ClearBlock(b.Meta().BlockID, tenantID)
+		if err != nil {
+			return err
+		}
+
 		instance.AddCompletingBlock(b)
 
 		i.enqueue(&flushOp{
--- a/modules/ingester/ingester.go
+++ b/modules/ingester/ingester.go
@@ -346,7 +346,7 @@ func (i *Ingester) replayWal() error {
 		// Delete anything remaining for the completed version of this
 		// wal block. This handles the case where a wal file is partially
 		// or fully completed to the local store, but the wal file wasn't
-		// deleted (because it wa rescanned above). This can happen for reasons
+		// deleted (because it was rescanned above). This can happen for reasons
 		// such as a crash or restart. In this situation we err on the side of
 		// caution and replay the wal block.
 		err = instance.local.ClearBlock(b.Meta().BlockID, tenantID)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * [FEATURE] Add ability to search ingesters for traces [#806](https://github.com/grafana/tempo/pull/806) (@mdisibio)
 * [BUGFIX] Update port spec for GCS docker-compose example [#869](https://github.com/grafana/tempo/pull/869) (@zalegrala)
 * [BUGFIX] Cortex upgrade to fix an issue where unhealthy compactors can't be forgotten [#878](https://github.com/grafana/tempo/pull/878) (@joe-elliott)
+* [BUGFIX] Fix "magic number" errors and other block mishandling when an ingester forcefully shuts down [#937](https://github.com/grafana/tempo/issues/937) (@mdisibio)
 * [ENHANCEMENT] Added "query blocks" cli option. [#876](https://github.com/grafana/tempo/pull/876) (@joe-elliott)
 * [ENHANCEMENT] Added traceid to `trace too large message`. [#888](https://github.com/grafana/tempo/pull/888) (@mritunjaysharma394)
 * [ENHANCEMENT] Add support to tempo workloads to `overrides` from single configmap in microservice mode. [#896](https://github.com/grafana/tempo/pull/896) (@kavirajk)
--- a/modules/ingester/ingester_test.go
+++ b/modules/ingester/ingester_test.go
@@ -238,7 +238,8 @@ func TestWalReplayDeletesLocalBlocks(t *testing.T) {
 	// Simulate a restart.
 	inst.completeBlocks = []*wal.LocalBlock{}
 	inst.completingBlocks = []*wal.AppendBlock{}
-	i.starting(ctx)
+	err = i.starting(ctx)
+	require.NoError(t, err)
 
 	// After restart we only have the 1 wal block
 	require.Len(t, inst.completingBlocks, 1)
--- a/modules/ingester/ingester_test.go
+++ b/modules/ingester/ingester_test.go
@@ -212,6 +212,39 @@ func TestWal(t *testing.T) {
 	}
 }
 
+// TestWalReplayDeletesLocalBlocks simulates the condition where an ingester restarts after a wal is completed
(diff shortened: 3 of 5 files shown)


modules/ingester/ingester.go
```diff
@@ -343,6 +343,17 @@ func (i *Ingester) replayWal() error {
 			return err
 		}
 
+		// Delete anything remaining for the completed version of this
+		// wal block. This handles the case where a wal file is partially
+		// or fully completed to the local store, but the wal file wasn't
+		// deleted (because it wa rescanned above). This can happen for reasons
```
joe-elliott, line 349: nit: was

joe-elliott (approved):

annanay25 (approved):

Issue #932, opened 2021-09-02
Ingester search production readiness


mdisibio:
Ingester search was recently merged #806 but it is marked experimental. This issue is tracking the list of additional work to consider it production-ready:

- [x] Config: Ingester: Make search block encoding and page size configurable (https://github.com/grafana/tempo/pull/990)
- [x] Config: Querier: separate search timeout from trace lookup timeout (https://github.com/grafana/tempo/pull/984)
- [ ] Querier: tolerate disappearing ingesters and return partial results
  - Related to https://github.com/grafana/tempo/issues/899
- [x] Querier/Ingester: correctly enforce `limit`, make the default configurable (#1022)
- [x] Ingester: Add block headers to eliminate blocks from search faster. Maybe also use blockheaders for autocomplete lookups instead of in-memory LRU cache which is somewhat inefficient. (https://github.com/grafana/tempo/pull/943)
- [x] Ingester: Fix duration filtering (sometimes incorrect results are returned) (https://github.com/grafana/tempo/pull/933)
- [x] Ingester: Dedupe WAL records when converting StreamingSearchBlock -> to BackendSearchBlock (https://github.com/grafana/tempo/pull/940)
- [x] Ingester: Reload search blocks on startup (requires gracefully handling blocks without search data) (https://github.com/grafana/tempo/pull/1000)
- [x] Distributor: Extract span status and make available for searching (error=y?) (https://github.com/grafana/tempo/pull/1093)
- [ ] Operational: alerts for search apis
- [x] Operational: dashboard panels for search apis (https://github.com/grafana/tempo/pull/930)
- [x] Documentation: config options, apis, alerts, troubleshooting (https://github.com/grafana/tempo/pull/1003)

joe-elliott:
We are seeing traceheaders use quite a bit of disk space in the ingesters. More than previously estimated. Consider the following tasks as well:

- [ ] Global denylist of tags that should not be added to traceheaders
- [ ] Per tenant allowlist/denylist of tags
- [ ] Max tag value length? Just truncate the value if it's > then some configurable number of bytes/runes

PR #940, opened 2021-09-08
Dedupe search records while replaying WAL

annanay25:
**What this PR does**:
Dedupe WAL records when moving from StreamingSearchBlock -> BackendSearchBlock
- Redefines the Combiner function to be variadic and adds a Search data combiner 
- Uses Deduping iterator to replay WAL records and combine using above combiner

Part of the ingester production readiness checklist https://github.com/grafana/tempo/issues/932

**Checklist**
- [x] Tests updated
- NA Documentation added


pkg/model/combine.go
```diff
@@ -18,12 +20,29 @@ type objectCombiner struct{}
...
+
+	if len(objs) == 1 {
+		return objs[0], false
 	}
+
+	combinedTrace := objs[0]
+	var wasCombined bool
+	var err error
+	for _, obj := range objs[1:] {
+		combinedTrace, wasCombined, err = CombineTraceBytes(combinedTrace, obj, dataEncoding, dataEncoding)
```
joe-elliott, line 39: CombineTraceBytes is just a wrapper for CombineTraceProtos. It would be more efficient to unmarshal to proto using the dataencoding here and then call CombineTraceProtos directly. Otherwise you're repeatedly marshalling/unmarshalling if len(objs) > 2

tempodb/search/backend_search_block.go
```diff
@@ -23,6 +26,101 @@ type BackendSearchBlock struct {
 	l        *local.Backend
 }
 
+//nolint:golint
+type SearchDataCombiner struct{}
+
+func (*SearchDataCombiner) Combine(_ string, searchData ...[]byte) ([]byte, bool) {
```
joe-elliott, line 32: it feels weird to mimic the Combiner interface and ignore the format string. Should the format string indicate that the thing being combined is marshalled search data and not object data?

Does it make sense to mimic the interface at all? It's an "ObjectCombiner" and these aren't objects. Are we conflating two things that happen to match in the function signature but are semantically different?

mdisibio: Reuse of DedupingIterator is the main thing.  If we want to split this from object combination, I would lean towards making DedupingIterator take a func callback, instead of an interface thing. That seems simpler.

pkg/model/combine.go
mdisibio: Yep we definitely want to do this optimization, but nothing is combining more than 2 entries yet.  Places where that's needed are DedupingIterator (wal), and MultiblockIterator (compaction).   Didn't want to tackle them in this PR. Thoughts?

tempodb/search/backend_search_block.go
joe-elliott: This makes sense. I'm fine with either path. I think we should take into consideration the broader goal is likely to deeply integrate these pieces with the existing backend block. Your all's call.

pkg/model/combine.go
joe-elliott: That's fine. Let's add a todo to make this improvement in the future.

annanay25: Yeah there's no changes to the way objects are currently being combined, will add this todo

tempodb/search/streaming_search_block.go
```diff
@@ -112,6 +86,10 @@ func (s *StreamingSearchBlock) Search(ctx context.Context, p Pipeline, sr *Resul
 			return nil
 		}
 
+		if r.Length == 0 {
+			continue
+		}
```
annanay25, line 91: Tempted to say "how did it work before", haven't we always had zero length records in the appender 🤔

tempodb/search/backend_search_block.go
```diff
@@ -23,6 +26,70 @@ type BackendSearchBlock struct {
...
+	file         *os.File
+}
+
+func (s *SearchDataIterator) Next(_ context.Context) (common.ID, []byte, error) {
+	if s.currentIndex >= len(s.records) {
+		return nil, nil, io.EOF
+	}
+
+	currentRecord := s.records[s.currentIndex]
+	buffer := make([]byte, currentRecord.Length)
```
mdisibio, line 76: Should try to resuse the buffer here.  Not blocking, I can follow up in another PR.

mdisibio (approved):

PR #943, opened 2021-09-09
Add search block headers

mdisibio:
**What this PR does**:
Adds block-level search data which speeds up some searches by eliminating blocks faster.  The block header contains the superset of unique tags and duration range within the block.  Data is stored in a new file `search-header` with a dedicated flatbuffer struct.

* Other notes:
  * Minor refactoring of search pipeline to be more structured around blocks/pages/traces. 
  * Previously it was possible to perform exhaustive searches and benchmarks using unmatching condition like `x=y`. This is no longer possible since it eliminates the block immediately due to tag not present in the header.  Since we still want these abilities it was readded with a secret tag `x-dbg-exhaustive`. If this tag is present in the search then it will evaluate all data again.

~This is likely to conflict with #940 so let's merge that first.~

Fixes part of #932 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/search/backend_search_block.go
```diff
@@ -76,20 +76,17 @@ func (s *SearchDataIterator) Next(_ context.Context) (common.ID, []byte, error)
...
-	// resize/extend buffer
-	if cap(s.buffer) < int(currentRecord.Length) {
-		s.buffer = make([]byte, currentRecord.Length)
-	}
-	s.buffer = s.buffer[:currentRecord.Length]
-
-	_, err := s.file.ReadAt(s.buffer, int64(currentRecord.Start))
+	// Use unique buffer that can be returned to the caller.
+	// This is primarily for DedupingIterator which uses 2 buffers at once.
+	buffer := make([]byte, currentRecord.Length)
```
annanay25: Should we add a test around this? Might be easy enough to do?

tempodb/search/backend_search_block.go
```diff
@@ -228,6 +160,27 @@ func (s *BackendSearchBlock) Search(ctx context.Context, p Pipeline, sr *Results
...
+
+	hb, err := tempo_io.ReadAllWithEstimate(hbr, hbrlen)
+	if err != nil {
+		return err
+	}
+
+	header := tempofb.GetRootAsSearchBlockHeader(hb, 0)
+	if !p.MatchesBlock(header) {
+		// Block filtered out
+		// TODO - metrics ?
```
annanay25, line 180: Yeah I think we should add a metric here to track how effective the block headers are.

tempodb/search/streaming_search_block.go
```diff
@@ -120,3 +122,48 @@ func (s *StreamingSearchBlock) Search(ctx context.Context, p Pipeline, sr *Resul
...
+		file:    s.file,
+	}
+
+	combiner := &DataCombiner{}
+
+	// Streaming (wal) blocks have to be deduped.
+	return encoding.NewDedupingIterator(iter, combiner, "")
+}
+
+//nolint:golint
```
annanay25, line 138: nit: nolint can be removed

tempodb/search/backend_search_block.go
mdisibio: Some measurement is already possible by comparing the inspectedBytes/Traces/Blocks in the response vs. an exhaustive search with `x-dbg-exhaustive`.  This comment was more about if there is anything missing, like maybe we should count the block header in inspectedBytes so that it more accurately measures i/o performed during the search.

What metric did you have in mind?

tempodb/search/backend_search_block.go
mdisibio: Yes tried, but unfortunately wasn't able to reproduce the issue in a go test, even after changing `newBackendSearchBlockWithTraces` to purposefully write duplicate data.  I think there are other factors at play, like maybe the 2nd buffer read has to be smaller than the 1st, so that it only overwrites part of the buffer.

annanay25: oh :( alright yeah we can skip this, the comment was a good idea.

tempodb/search/backend_search_block.go
annanay25: I was thinking something along the lines of exported metrics (to prometheus) that show how many requests are filtered out by the headers, that would be really useful to track. 

But also thinking if we can make this generic enough and instrument each stage of the pipeline.

tempodb/search/backend_search_block_test.go
```diff
@@ -28,15 +28,16 @@ func genSearchData(traceID []byte, i int) [][]byte {
...
 	f, err := os.OpenFile(path.Join(t.TempDir(), "searchdata"), os.O_CREATE|os.O_RDWR, 0644)
 	require.NoError(t, err)
 
 	b1, err := NewStreamingSearchBlockForFile(f)
 	require.NoError(t, err)
+
 	for i := 0; i < traceCount; i++ {
-		assert.NoError(t, b1.Append(context.Background(), id, genSearchData(id, i)))
+		id := make([]byte, 16)
+		binary.LittleEndian.PutUint32(id, uint32(i))
```
annanay25, line 39: Making a note to self: Need to add a new test for deduping WAL segments if the traceIDs are going to become unique here

annanay25 (approved):
lgtm!

tempodb/search/backend_search_block_test.go
annanay25: This is done here - https://github.com/grafana/tempo/pull/968

PR #944, opened 2021-09-09
Implement search in vulture

### LINKED ISSUE #922: Update the Vulture to support search
Search is going to be merged soon, but we don't have a way to do live e2e testing.

Extend the vulture to also exercise the search API looking for traces it has recently pushed.

zalegrala:
**What this PR does**:

Here we include a Vulture update to exercise the search API in Tempo and report on the results.

Fixes #922

**Checklist**
- [x] Tests updated
- [x] Documentation added


cmd/tempo-vulture/main.go
```diff
@@ -137,27 +144,72 @@ func main() {
...
+				log.Error("query for metrics failed",
+					zap.Error(err),
+				)
 			}
+			pushMetrics(queryMetrics)
+		}
+	}()
+
+	// Search
+	go func() {
```
zalegrala, line 166: Should we just smoosh the readers together?  We could still keep them on a different ticker, but just running out of the same routine.

cmd/tempo-vulture/main.go
```diff
@@ -127,8 +133,9 @@ func main() {
 			startTime, seed = selectPastTimestamp(startTime, now, interval, tempoRetentionDuration)
 
 			// Don't attempt to read on the first itteration if we can't reasonably
-			// expect the write loop to have fired yet.
-			if seed.Before(actualStartTime.Add(tempoWriteBackoffDuration)) {
+			// expect the write loop to have fired yet.  Double the duration here to
+			// avoid a race.
+			if seed.Before(actualStartTime.Add(tempoWriteBackoffDuration).Add(tempoWriteBackoffDuration)) {
```
joe-elliott, line 138: nit: I think this would be more readable
```
actualStartTime.Add(2 * tempoWriteBackoffDuration)
```

cmd/tempo-vulture/main.go
joe-elliott: I think the way you have it is fine.

cmd/tempo-vulture/main.go
```diff
@@ -137,27 +144,72 @@ func main() {
...
 
 	http.Handle(prometheusPath, promhttp.Handler())
 	log.Fatal(http.ListenAndServe(prometheusListenAddress, nil))
 }
 
+func pushMetrics(metrics traceMetrics) {
+	metricTracesInspected.Add(float64(metrics.requested))
+	metricTracesErrors.WithLabelValues("incorrectresult").Add(float64(metrics.incorrectResult))
+	metricTracesErrors.WithLabelValues("missingspans").Add(float64(metrics.missingSpans))
+	metricTracesErrors.WithLabelValues("tracemissingfromtagsearch").Add(float64(metrics.traceMissingFromTagSearch))
```
joe-elliott, line 208: `tracemissingfromtagsearch` is quite long winded. we are definitely in personal preference territory here, but maybe we could do something like:

`notfound_byid`
`notfound_search`

cmd/tempo-vulture/main.go
```diff
@@ -258,20 +310,89 @@ func generateRandomInt(min int64, max int64, r *rand.Rand) int64 {
...
+
+	logger := logger.With(
+		zap.Int64("seed", seed.Unix()),
+		zap.String("hexID", hexID),
+		zap.Duration("ago", time.Since(seed)),
+	)
+	logger.Info("searching Tempo")
+
+	// Use the search API to find details about the expected trace
+	for _, expectedBatch := range expected.Batches {
```
joe-elliott, line 355: is this loop going to search for every attribute in the trace? maybe we should just choose one randomly and use that?

the two `backoffDuration` config params above are used to control the load that vulture puts on Tempo. if we execute a bunch of queries on each pass we're kind of betraying this config.

pkg/util/client.go
```diff
@@ -0,0 +1,90 @@
...
+func NewClient(baseURL, orgID string, log *zap.Logger) *Client {
+	return &Client{
+		BaseURL: baseURL,
+		OrgID:   orgID,
+		client:  http.DefaultClient,
+		logger:  log,
+	}
+}
+
+func (c *Client) GetFor(url string, m proto.Message) error {
```
joe-elliott, line 33: make private since it's a helper method designed to be used internally?

pkg/util/http.go
```diff
@@ -74,3 +74,29 @@ func TraceIDToHexString(byteID []byte) string {
...
+	aa, err := HexStringToTraceID(a)
+	if err != nil {
+		return false, err
+	}
+	bb, err := HexStringToTraceID(b)
+	if err != nil {
+		return false, err
+	}
+
+	if len(aa) != len(bb) {
```
joe-elliott, line 91: use `bytes.Equal`

joe-elliott (changes requested):

cmd/tempo-vulture/main.go
zalegrala: Shorter and more readable sounds good to me.

cmd/tempo-vulture/main.go
zalegrala: That's fair.  There are now three backoff config params, but I can see just selecting one of the attributes, so that we know that for each tick of the loops, one call is made to the service and we can make some assumptions about the load characteristics that are added with vulture.

zalegrala:
Thanks for the review and the good suggestions.  I believe I addressed all the feedback.

cmd/tempo-vulture/main.go
```diff
@@ -127,8 +134,14 @@ func main() {
...
-			// expect the write loop to have fired yet.
-			if seed.Before(actualStartTime.Add(tempoWriteBackoffDuration)) {
+			// expect the write loop to have fired yet.  Double the duration here to
+			// avoid a race.
+			if seed.Before(actualStartTime.Add(2 * tempoWriteBackoffDuration)) {
+				continue
+			}
+
+			// Do the same for the startTime we've just selected
+			if seed.Before(startTime.Add(tempoWriteBackoffDuration)) {
```
joe-elliott, line 144: does this matter? if startTime > actualStartTime it means that vulture has been running longer than the retention period which means there shouldn't be a race condition?

zalegrala: I've adjusted the search and read loops to use the simpler logic.  Thanks for the suggestion.

cmd/tempo-vulture/main.go
```diff
@@ -258,23 +313,89 @@ func generateRandomInt(min int64, max int64, r *rand.Rand) int64 {
...
+
+			}
+		}
+
+		return false
+	}
+
+	attr := randomAttrFromTrace(expected, r)
+	if attr == nil {
+		return tm, fmt.Errorf("no search attr selected from trace")
```
mapno, line 352: Should we add this failure as a metric?

mapno (approved):
Overall LGTM, great work

cmd/tempo-vulture/main.go
zalegrala: I think so, good point.

joe-elliott (approved):

PR #945, opened 2021-09-10
Make tempo use runtimeconfig from `dskit` instead of `cortex`

kavirajk:
**What this PR does**:
Make tempo use runtimeconfig from `dskit` instead of `cortex` (similar to Loki)

Fixes # NA

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 * [ENHANCEMENT] Add support to tempo workloads to `overrides` from single configmap in microservice mode. [#896](https://github.com/grafana/tempo/pull/896) (@kavirajk)
 * [ENHANCEMENT] Make `overrides_config` block name consistent with Loki and Cortex in microservice mode. [#906](https://github.com/grafana/tempo/pull/906) (@kavirajk)
 * [ENHANCEMENT] Updated config defaults to reflect better capture operational knowledge. [#913](https://github.com/grafana/tempo/pull/913) (@joe-elliott)
-* [ENHANCEMENT] Make tempo use `runtimeconfig` package from `dskit` module. [#945](https://github.com/grafana/tempo/pull/945) (@kavirajk)
+* [ENHANCEMENT] Make tempo use `runtimeconfig` package from `dskit` module. This also changes the metrics name from `cortex_runtime_config_last_reload_successful` to `tempo_runtime_config_last_reload_successful` [#945](https://github.com/grafana/tempo/pull/945) (@kavirajk)
   ```
   ingester:
     trace_idle_period: 30s => 10s  # reduce ingester memory requirements with little impact on querying
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 * [ENHANCEMENT] Add support to tempo workloads to `overrides` from single configmap in microservice mode. [#896](https://github.com/grafana/tempo/pull/896) (@kavirajk)
 * [ENHANCEMENT] Make `overrides_config` block name consistent with Loki and Cortex in microservice mode. [#906](https://github.com/grafana/tempo/pull/906) (@kavirajk)
 * [ENHANCEMENT] Updated config defaults to reflect better capture operational knowledge. [#913](https://github.com/grafana/tempo/pull/913) (@joe-elliott)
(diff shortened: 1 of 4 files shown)


aknuds1 (approved):

modules/overrides/overrides.go
```diff
@@ -58,12 +59,12 @@ func NewOverrides(defaults Limits) (*Overrides, error) {
...
 
 	if defaults.PerTenantOverrideConfig != "" {
-		runtimeCfg := runtimeconfig.ManagerConfig{
+		runtimeCfg := runtimeconfig.Config{
 			LoadPath:     defaults.PerTenantOverrideConfig,
 			ReloadPeriod: time.Duration(defaults.PerTenantOverridePeriod),
 			Loader:       loadPerTenantOverrides,
 		}
-		runtimeCfgMgr, err := runtimeconfig.NewRuntimeConfigManager(runtimeCfg, prometheus.DefaultRegisterer)
+		runtimeCfgMgr, err := runtimeconfig.New(runtimeCfg, prometheus.WrapRegistererWithPrefix("cortex_", prometheus.DefaultRegisterer), log.Logger)
```
kavirajk, line 67: NOTE: Keeping the prefix `cortex_` as I find this metric `cortex_runtime_config_last_reload_successful` used in one of the tempo alerts.

```
ksonnet/lib/tempo-cloud-mixin/alerts.libsonnet
36:                cortex_runtime_config_last_reload_successful{namespace=~"%s"} == 0
```

annanay25: Can we update that alert instead?

annanay25 (commented):
Thanks for the changes Kavi.

Pinging @mapno because he is currently making changes to this module as well.

modules/overrides/overrides.go
kavirajk: Hmm. But we couldn't change the alert until this change reaches the particular cell correct?

What we can do is, I can change prefix to `tempo_` and once this reaches the prod, someone can change the alert name. WDYT?

kavirajk: Ah. the alert is from vendor. I can directly change it in here!. 👍

kavirajk: I changed the prefix to `tempo_`. and created TODO https://github.com/grafana/deployment_tools/pull/15791
for the reminder!

mapno (approved):
LGTM

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 * [ENHANCEMENT] Add support to tempo workloads to `overrides` from single configmap in microservice mode. [#896](https://github.com/grafana/tempo/pull/896) (@kavirajk)
 * [ENHANCEMENT] Make `overrides_config` block name consistent with Loki and Cortex in microservice mode. [#906](https://github.com/grafana/tempo/pull/906) (@kavirajk)
 * [ENHANCEMENT] Updated config defaults to reflect better capture operational knowledge. [#913](https://github.com/grafana/tempo/pull/913) (@joe-elliott)
+* [ENHANCEMENT] Make tempo use `runtimeconfig` package from `dskit` module. This also changes the metrics name from `cortex_runtime_config_last_reload_successful` to `tempo_runtime_config_last_reload_successful` [#945](https://github.com/grafana/tempo/pull/945) (@kavirajk)
```
pstibrany, line 13: Nit: not sure if user cares about dskit. I would classify this as a change of metric name only.

modules/overrides/overrides.go
```diff
@@ -6,7 +6,8 @@ import (
 	"io"
 	"time"
 
-	"github.com/cortexproject/cortex/pkg/util/runtimeconfig"
+	"github.com/cortexproject/cortex/pkg/util/log"
```
pstibrany, line 9: does Tempo have it's own global logger? :)

pstibrany (approved):
Thank you!

modules/overrides/overrides.go
kavirajk: I tried finding one. AFAIK they use this global logger in other places as well.

modules/overrides/overrides.go
annanay25: Hmm, shouldn't that alert be in the public mixin because its not specific to Grafana Cloud?

annanay25: But this is fine, can be picked in another PR :)

PR #947, opened 2021-09-10
Chore: Update to latest Cortex

chaudum:
**What this PR does**:

Upgrades the Cortex dependency to the latest commit.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


cmd/tempo/app/modules.go
```diff
@@ -248,13 +249,23 @@ func (t *App) initStore() (services.Service, error) {
...
+	reg := prometheus.DefaultRegisterer
+	t.cfg.MemberlistKV.MetricsRegisterer = reg
 	t.cfg.MemberlistKV.MetricsNamespace = metricsNamespace
 	t.cfg.MemberlistKV.Codecs = []codec.Codec{
 		ring.GetCodec(),
 	}
 
-	t.MemberlistKV = memberlist.NewKVInitService(&t.cfg.MemberlistKV, log.Logger)
+	dnsProviderReg := prometheus.WrapRegistererWithPrefix(
+		"cortex_",
```
joe-elliott, line 260: probably don't want to use the `cortex_` provider here. `tempo_`?

mdisibio (approved):

cmd/tempo/app/modules.go
MichelHollands: Good catch. I'll update that.

joe-elliott (approved):
looks like you have some conflicts with main, but other than that lgtm

MichelHollands:
> looks like you have some conflicts with main, but other than that lgtm

We can't keep up with you :) It's fixed now.

PR #948, opened 2021-09-13
Zstd decoder

joe-elliott:
**What this PR does**:
- Use the zstd decoder instead of the streaming io.Reader for reads. Seeing a 10-20% increase in throughput when iterating a block and 10-20% decrease in compactor CPU with this change.
- Update klauspost/compress dependency to latest

I also attempted the same change for the encoder/writing, but saw a performance regression so I left it out.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


tempodb/encoding/v2/data_reader.go
```diff
@@ -105,7 +108,13 @@ func (r *dataReader) Read(ctx context.Context, records []common.Record, pagesBuf
 			return nil, nil, err
 		}
 
-		pagesBuffer[i], err = tempo_io.ReadAllWithBuffer(reader, len(page), pagesBuffer[i])
+		// zstd decoder is ~10-20% faster then the streaming io.Reader interface so prefer that
+		decoder, ok := reader.(*zstd.Decoder)
+		if ok {
+			pagesBuffer[i], err = decoder.DecodeAll(page, pagesBuffer[i][:0])
```
mdisibio, line 114: I was thinking about this too when benchmarking in search.  I think some of the speedup is from not wrapping the `[]byte` in a buffer read.   I think this is a use case we could support directly in the pool, i.e. de/encode slice to slice with no buffer shuffling.    What do you think about something like a new method in the pools, `pool.Decode([]byte,[]byte) ([]byte, []byte,err)`   The zstd pool would do exactly like what you did here (call DecodeAll).  The other pools that require readers would still do the same shuffling with `tempo_io.ReadAllWithBuffer`.

joe-elliott: The only issue with doing it directly in the pool is it would require allocating a new reader each time. Right now we can allocate one reader and call reset between pages. I've not benchmarked the difference but docs indicate this is preferred for performance.

mdisibio (approved):
lg (enough) tm !

PR #951, opened 2021-09-13
Add support for vulture sending long running traces

### LINKED ISSUE #791: Extend vulture to generate long-running traces

zalegrala:
**What this PR does**:

Here we implement the ability for the vulture to send long-running traces, which will enable the vulture verify additional infrastructure components required to resolve the trace.

Fixes #791

**Checklist**
- [x] Tests updated
- [ ] Documentation added


zalegrala:
This will require https://github.com/grafana/tempo/pull/944

zalegrala:
This is working pretty good locally.  There are two conditions that could result in false positives that I know about currently.

The first is that the number of bytes on a trace may exceed the allowed limit of Tempo, which is hard to know unless we query the API to know the size.  But perhaps some work on that API should precede that so we can work with json.  In lieu of working on the API and reading the configured value from Tempo, I've reduced the chance of extending a trace to include future amendments to a given trace.  This seems like maybe a reasonable approach, since for each extension of a trace, the chance of extending it yet again continues to diminish. 

The second is that when we go to read what we expect is the result, we are effectively counting the number of writes that we expect has taken place.  There is a margin of error, where if the timing worked out just perfect, the `incorrectresult` ticker would rise falsely.  I've added a dirty little sleep that should reduce the chance of it happening if it seems like we are within that margin of error, but I don't feel great about it, but it probably works fine.

cmd/tempo-vulture/main.go
```diff
@@ -205,6 +207,54 @@ func main() {
...
+}
+
+func queueFutureBatches(client *jaeger_grpc.Reporter, info *traceInfo) {
+	log := logger.With(
+		zap.String("org_id", tempoOrgID),
+		zap.String("write_trace_id", fmt.Sprintf("%016x%016x", info.traceIDHigh, info.traceIDLow)),
+		zap.Int64("seed", info.timestamp.Unix()),
+	)
+
+	if maybe(info.r) {
```
joe-elliott, line 244: i'd like to look at a better way to do this. perhaps we could generate a "total batches" int from the random number generator and store it in traceinfo. then we could decrement that value each time we emitted a batch?

zalegrala: Good call, I think that cleans up the readability a bit.  I've made the update.

cmd/tempo-vulture/main.go
```diff
@@ -507,40 +558,70 @@ func hasMissingSpans(t *tempopb.Trace) bool {
...
+
+	for info.longWritesRemaining > 0 {
+		info.longWritesRemaining--
+
+		lastWrite = lastWrite.Add(tempoLongWriteBackoffDuration)
+
+		// If the last write has happened very recently, we'll wait a bit to make
+		// sure the write has taken place, and then add the batches that would have
+		// just been written.
+		if time.Since(lastWrite) < (1 * time.Second) {
```
joe-elliott, line 617: this is pretty janky.

from a seed we should be able to determine when a trace was completed and either query or not query the trace if we feel it is not complete.

putting sleeps here is going to weirdly impact the timings of other loops.

zalegrala: Alright, I've washed off some of that jank.  No more sleeps to construct the expected trace, and we only search/query when we expect that all writes have taken place.

cmd/tempo-vulture/main.go
```diff
@@ -34,16 +34,25 @@ var (
...
+	tempoRetentionDuration        time.Duration
+	tempoSearchRetentionDuration  time.Duration
 
 	logger *zap.Logger
+
+	// batchHighWaterMark is used when writing and reading, and needs to match so
+	// that we get the expected number of batches on a trace.  A value larger
+	// than 25 here results in vulture writing traces that exceed the maximum
+	// trace size.
+	batchHighWaterMark int64 = 25
```
joe-elliott, line 53: rename:
```
maxBatchesPerWrite
maxLongWritesPerTrace
```

joe-elliott (approved):

PR #952, opened 2021-09-13
Consolidate status information onto /status endpoint

### LINKED ISSUE #949: Consolidate status API information on `/status`
Currently there are a few endpoints that convey information about the configuration and status of the running instance.  As part of #938 and #632, we now have a `/status` end point that is ready to accept a query parameter to organize all the data about status.

The following may serve as a guide as to what gets moved to the `/status` endpoint.

```
/status?config
/status?runtime_config
/status?services
/status?endpoints <-- currently implemented
```

This could mean we reduce the number of endpoints we have to convey status information, and still allow users to ask for only the portion of the status that is of interest.

zalegrala:
**What this PR does**:
Here we move a few status endpoints onto the `/status` endpoint.  Optionally, users can pass a few query parameters that will reduce the amount of output.

Fixes #949

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@
 * [CHANGE] Renamed CLI flag from `--storage.trace.maintenance-cycle` to `--storage.trace.blocklist_poll`. This is a **breaking change**  [#897](https://github.com/grafana/tempo/pull/897) (@mritunjaysharma394)
 * [CHANGE] update jsonnet alerts and recording rules to use `job_selectors` and `cluster_selectors` for configurable unique identifier labels [#935](https://github.com/grafana/tempo/pull/935) (@kevinschoonover)
 * [CHANGE] Modify generated tag keys in Vulture for easier filtering [#934](https://github.com/grafana/tempo/pull/934) (@zalegrala)
+* [CHANGE] Consolidate status information onto /status endpoint [ #952 ](https://github.com/grafana/tempo/pull/952) @zalegrala)
 * [FEATURE] Add runtime config handler  [#936](https://github.com/grafana/tempo/pull/936) (@mapno)
 
 ## v1.1.0 / 2021-08-26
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -38,6 +38,7 @@
 * [CHANGE] Renamed CLI flag from `--storage.trace.maintenance-cycle` to `--storage.trace.blocklist_poll`. This is a **breaking change**  [#897](https://github.com/grafana/tempo/pull/897) (@mritunjaysharma394)
 * [CHANGE] update jsonnet alerts and recording rules to use `job_selectors` and `cluster_selectors` for configurable unique identifier labels [#935](https://github.com/grafana/tempo/pull/935) (@kevinschoonover)
 * [CHANGE] Modify generated tag keys in Vulture for easier filtering [#934](https://github.com/grafana/tempo/pull/934) (@zalegrala)
+* [CHANGE] Consolidate status information onto /status endpoint [ #952 ](https://github.com/grafana/tempo/pull/952) @zalegrala)
```
joe-elliott, line 41: This is a breaking change. We need to list that and also list all the endpoints that moved and where they moved to.

See below for examples.

joe-elliott (commented):
Generally heading the right way but let's make a few changes as discussed offline:

1) switch to path based i.e. `/status/config` instead of `/status?config`
2) `/status` should continue to list everything
3) at the top of each section in `/status` list the endpoint that needs to be used to request only that section
4) add `/status/version`
5) order of values in `/status` should be version, services, endpoints, runtime_config and then config

joe-elliott:
See here:
https://github.com/grafana/tempo/blob/main/cmd/tempo/main.go#L67
for an example of how we print the version.

zalegrala:
I've updated the approach here, and I hope it is still pretty readable.  I wasn't quite sure about the error handling on that status method, but let me know what you think.

zalegrala:
Oh that might not work due to the linter.  Hmm.  I'll be less clever.

joe-elliott (approved):

PR #960, opened 2021-09-15
Support global denylist and per-tenant allowlist of tags for search data

annanay25:

**What this PR does**:
- Support global denylist of tags to be dropped from search data. The denylist will be applied at the distributors and only filtered tags will be added to the flatbuffers.
- Support per-tenant allowlist of tags that will override the global denylist.

Fixes part of #932 

**Checklist**
- [x] Tests updated
- [x] Documentation added


mapno (approved):
LGTM

modules/distributor/distributor.go
```diff
@@ -155,6 +158,7 @@ func New(cfg Config, clientCfg ingester_client.Config, ingestersRing ring.ReadRi
 		DistributorRing:      distributorRing,
 		ingestionRateLimiter: limiter.NewRateLimiter(ingestionRateStrategy, 10*time.Second),
 		searchEnabled:        searchEnabled,
+		tagManager:           NewTagManager(o, cfg.SearchTagsDenyList),
```
mapno, line 161: Should the tag manager be made a `services.Service`?

modules/distributor/tag_manager.go
```diff
@@ -0,0 +1,116 @@
...
+				match := true
+				for _, tag := range newList {
+					if _, ok := currentMap[tag]; !ok {
+						match = false
+					}
+				}
+				if len(currentMap) != len(newList) {
+					match = false
+				}
+				if match == false {
```
mapno, line 68:
```suggestion
				if !match {
```

modules/distributor/tag_manager.go
```diff
@@ -0,0 +1,116 @@
...
+					tm.ch <- tenant
```
mapno, line 69: I think this could cause a deadlock if the chan is already full. Something like this https://play.golang.org/p/T40V9_EHosO

modules/distributor/tag_manager.go
```diff
@@ -0,0 +1,116 @@
...
+	tm.mtx.RLock()
+	defer tm.mtx.RUnlock()
+	if tenantAllowList, ok := tm.tenantAllowList[tenant]; ok {
+		if _, exists := tenantAllowList[tag]; exists {
+			// exists in tenant allow list, allow
+			return true
+		}
+	} else if allowList := tm.o.SearchTagsAllowList(tenant); len(allowList) > 0 { // fetch the allowList from overrides
+		// we want to take only a read lock here, pass it to the update loop
+		tm.ch <- tenant
```
joe-elliott, line 99: i don't understand what this channel is doing or why we are checking two allowlists for a tenant? is the first not sufficient?

annanay25:
Thanks @joe-elliott, addressed both.

PR #963, opened 2021-09-15
Add search block headers for wal blocks

mdisibio:
**What this PR does**:
Maintains a search blocker header for wal blocks like #943 did for completed blocks. This helps eliminate blocks from the search faster for non-existent or sparse values.

The wal requires a mutable header, so some refactoring was done to better fit the search pipeline and the various trace/page/block structs together, i.e. the pipeline can now match directly against a SearchBlockHeaderMutable.

Fixes some of #932 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


annanay25 (approved):
Looking really good. Changelog entry and we should merge this.

PR #972, opened 2021-09-21
[Tempo CLI] Additional Features

joe-elliott:
**What this PR does**:
Adds two new features to the Tempo CLI that were created while experimenting with block iteration.
- Extends the list --scan feature to print out all label/value pairs along with their frequency of occurrence
- Adds a `search blocks` command that searches a time range for a key/value pair
- Adds a PrefetchIterator to asynchronously pull data from the backend while parsing/search

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


cmd/tempo-cli/cmd-list-block.go
```diff
@@ -85,13 +100,16 @@ func dumpBlock(r tempodb_backend.Reader, c tempodb_backend.Compactor, tenantID s
...
 		minObjSize := 0
 
+		allKVP := kvPairs{}
 		printStats := func() {
 			fmt.Println()
 			fmt.Println("Scanning results:")
 			fmt.Println("Objects scanned : ", i)
 			fmt.Println("Duplicates      : ", dupe)
 			fmt.Println("Smallest object : ", humanize.Bytes(uint64(minObjSize)))
 			fmt.Println("Largest object  : ", humanize.Bytes(uint64(maxObjSize)))
```
annanay25, line 110: While you're in here - can we print the ID for the largest object? Currently trying to debug some OOMing compactors and it would help to know if they are trying to combine a super large trace across multiple blocks

joe-elliott: done. good call.

annanay25 (approved):
lgtm!

PR #975, opened 2021-09-22
Set span.kind to client in query-frontend

mapno:
**What this PR does**:

It sets `span.kind` to `client` in query-frontend's retry middleware spans. This allows for query-frontend client spans to be paired with querier server spans.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/frontend/retry.go
```diff
@@ -37,6 +37,7 @@ func (r retryWare) Do(req *http.Request) (*http.Response, error) {
 	ctx := req.Context()
 	span, ctx := opentracing.StartSpanFromContext(ctx, "frontend.Retry")
 	defer span.Finish()
+	span.SetTag("span.kind", "client")
```
annanay25, line 40: I think it might be better to use the `opentracing-go/ext` which has the right set of tags to add 
Ex: https://github.com/weaveworks/common/blob/284892f3c53779d1a5956460b7fc8843b02ae7a9/instrument/instrument.go#L157

annanay25 (approved):
Approving to unblock, left a minor comment

modules/frontend/retry.go
joe-elliott: Doesn't this already work with the cortex/loki? any idea why those are working? Don't we need a matching span.kind = server? or does that already exist?

mapno: > Doesn't this already work with the cortex/loki? any idea why those are working?
Yes, it works for them. Cortex is using this [instrumentation](https://github.com/cortexproject/cortex/blob/master/pkg/querier/queryrange/instrumentation.go#L26-L30) for its middlewares, which instruments requests that go through them. I can explore that too.

> Don't we need a matching span.kind = server? or does that already exist?
Yeah, the globaltracer that Tempo uses correctly observes the incoming request and tags it accordingly.

joe-elliott (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 * [CHANGE] **BREAKING CHANGE** Change ingester metric `ingester_bytes_metric_total` in favor of `ingester_bytes_received_total` [#979](https://github.com/grafana/tempo/pull/979) (@mapno)
 * [FEATURE] Add ability to search ingesters for traces [#806](https://github.com/grafana/tempo/pull/806) (@mdisibio)
 * [FEATURE] Add runtime config handler  [#936](https://github.com/grafana/tempo/pull/936) (@mapno)
+* [BUGFIX] Update port spec for GCS docker-compose example [#869](https://github.com/grafana/tempo/pull/869) (@zalegrala)
+* [BUGFIX] Fix "magic number" errors and other block mishandling when an ingester forcefully shuts down [#937](https://github.com/grafana/tempo/issues/937) (@mdisibio)
+* [BUGFIX] Fix compactor memory leak [#806](https://github.com/grafana/tempo/pull/806) (@mdisibio)
+* [BUGFIX] Set span's tag `span.kind` to `client` in query-frontend [#975](https://github.com/grafana/tempo/pull/975) (@mapno)
```
annanay25, line 18: Bugfixes have been moved to the bottom of this list, let's append there

mapno: Yeah, that was a bad rebase. Fixed, thanks

PR #979, opened 2021-09-22
Add received and ingested bytes metrics in ingesters

mapno:
**What this PR does**:

Measures received and ingested bytes in the ingesters.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


CHANGELOG.md
```diff
@@ -40,6 +40,7 @@
 * [ENHANCEMENT] Compression updates: Added s2, improved snappy performance [#961](https://github.com/grafana/tempo/pull/961) (@joe-elliott)
 * [ENHANCEMENT] Add search block headers [#943](https://github.com/grafana/tempo/pull/943) (@mdisibio)
 * [ENHANCEMENT] Add search block headers for wal blocks [#963](https://github.com/grafana/tempo/pull/963) (@mdisibio)
+* [ENHANCEMENT] Add received bytes metric in ingesters [#979](https://github.com/grafana/tempo/pull/979) (@mapno)
```
joe-elliott, line 43: Please put the new metric name in the changelog.

modules/ingester/instance.go
```diff
@@ -159,6 +166,12 @@ func (i *instance) Push(ctx context.Context, req *tempopb.PushRequest) error {
 
 // PushBytes is used to push an unmarshalled tempopb.Trace to the instance
 func (i *instance) PushBytes(ctx context.Context, id []byte, traceBytes []byte, searchData []byte) error {
+	// measure received bytes as sum of slice capacities
+	// type byte is guaranteed to be 1 byte in size
+	// ref: https://golang.org/ref/spec#Size_and_alignment_guarantees
+	size := cap(id) + cap(traceBytes) + cap(searchData)
```
joe-elliott, line 172: use len() (not cap) and don't count the id, just the traceBytes/searchData

joe-elliott (commented):
A good start. I think we should:
- remove the bytes_written and replace it with this
- use a label on the counter to distinguish between tracedata and searchdata

CHANGELOG.md
mapno: Added, thanks

modules/ingester/instance.go
mapno: I wasn't sure which one to use. Changed, thanks

joe-elliott (approved):

PR #980, opened 2021-09-22
Compression Fix #3

joe-elliott:
**What this PR does**:
Ok, this one is going to stick.

- Reverts snappy to continue using `golang/snappy`. Although `klauspost/snappy` outperformed on benchmarks in a real world scenario there is either no difference or `golang/snappy` is better. This is likely b/c our workload is to compress a large number of small byte slices. I believe with tuning klauspost/s2 could outperform golang/snappy.
- Keeps klauspost/s2. This allows us to experiment with a potentially better compression for the wal while not risking stability by swapping the snappy implementation out from under us.
- Explicitly .Close() snappy/s2 writes on Put
- Found and fixed the unclosed appender.


mapno (approved):

annanay25 (approved):
lgtm.

PR #982, opened 2021-09-23
Register `/ingester/ring` endpoint when running ingester target

chaudum:
**What this PR does**:

This commit fixes the problem that the ring endpoint for the ingester
(`/ingester/ring`) was only registered when running the `all` target, but not
when running the `ingester` target.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@
 * [BUGFIX] Fix "magic number" errors and other block mishandling when an ingester forcefully shuts down [#937](https://github.com/grafana/tempo/issues/937) (@mdisibio)
 * [BUGFIX] Fix compactor memory leak [#806](https://github.com/grafana/tempo/pull/806) (@mdisibio)
 * [BUGFIX] Fix an issue with WAL replay of zero-length search data when search is disabled. [#968](https://github.com/grafana/tempo/pull/968) (@annanay25)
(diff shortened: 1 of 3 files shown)


chaudum:
I don't really know why CI fails, because locally the test passes.

joe-elliott:
Thanks for the contribution. We have a rather flakey test which is why CI is failing: https://github.com/grafana/tempo/issues/983

Regarding the change: The reason that the `/ingester/ring` endpoint does not work currently is because the ring _client_ exposes that endpoint and an ingester has no need for the ingester ring client. This endpoint is available on both queriers and distributors which both do have need for the ring client.  However, I do see some value in reducing confusion because a lot of people check the ingesters to see the ring state.

However, I believe this will introduce unnecessary overhead b/c every ingester will now start polling etcd/consul and place extra load on the kv store used to propagate the ring. @bboreham is an expert on the ring and may have more insight on whether this is appropriate or not.

chaudum:
> Regarding the change: The reason that the `/ingester/ring` endpoint does not work currently is because the ring _client_ exposes that endpoint and an ingester has no need for the ingester ring client.

Ah I see, got it.

> This endpoint is available on both queriers and distributors which both do have need for the ring client.

Should not be a problem, in the gateway I will proxy the `/ingester/ring` path to the `distributor` then.

> However, I do see some value in reducing confusion because a lot of people check the ingesters to see the ring state.

Yes, me 🙈 

Thanks for clarifying!

PR #988, opened 2021-09-28
WIP: Blocks partial results

### LINKED ISSUE #899: Support Returning Partial Results
Occasionally we see intermittent failures to query traces due to a single block having issues. These issues could be data corruption or occassionally just GCS misbehaving. Currently this results in a 500 even though it's quite possible we have correctly retrieved the trace.

In the above case let's return whatever we happen to have retrieved along with a simple warning. Something like:
```
Potential partial results. Failed to query block <block id>. 
```

Note: Will require some coordination with the Grafana team to add support for the warning.

**Additional context**

mapno:
**What this PR does**:

It supports partial results when block queries fail.

The general idea is to let queriers fail when querying blocks and decide the query frontend if the errors are too many or if the partial result is worth returning.

Partial results are indicated with http 206 (partial content).

Fixes #899 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/querier/querier.go
```diff
@@ -185,13 +185,19 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
 			ot_log.Int("combinedTraces", traceCountTotal))
 	}
 
+	var blockErrsCount uint32
 	if req.QueryMode == QueryModeBlocks || req.QueryMode == QueryModeAll {
 		span.LogFields(ot_log.String("msg", "searching store"))
-		partialTraces, dataEncodings, err := q.store.Find(opentracing.ContextWithSpan(ctx, span), userID, req.TraceID, req.BlockStart, req.BlockEnd)
+		partialTraces, dataEncodings, blockErrs, err := q.store.Find(opentracing.ContextWithSpan(ctx, span), userID, req.TraceID, req.BlockStart, req.BlockEnd)
```
kvrhdn, line 191: We don't use blockErrs in a meaningful way (we only use the count, not the errors itself). Maybe we should only return an int instead of `[]error`?

modules/querier/querier.go
```diff
@@ -185,13 +185,19 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
+		// err contains unrecoverable errors
+		// errs querying blocks are contained in blockErrs
```
kvrhdn, line 193: Can you add these comments to `Reader.Find` itself?

modules/frontend/querysharding.go
```diff
@@ -181,6 +198,14 @@ func mergeResponses(ctx context.Context, rrs []RequestResponse) (*http.Response,
...
 
+	if combinedTrace != nil {
+		var err error
+		combinedTraceBytes, err = combinedTrace.Marshal()
+		if err != nil {
+			return nil, errors.Wrap(err, "error marshaling combined trace at query frontend")
+		}
+	}
+
 	if shardMissCount == len(rrs) {
```
kvrhdn, line 209: We can move this block above `if combinedTrace != nil`: if we exit the loop and no querier returned something, just exit immediately.

modules/frontend/querysharding.go
```diff
@@ -190,12 +215,18 @@ func mergeResponses(ctx context.Context, rrs []RequestResponse) (*http.Response,
...
+		if totalBlockErrCount > 0 {
+			// If there are failed blocks, and we haven't returned with an error,
+			// signal the upstream that the result contains partial results.
+			statusCode = http.StatusPartialContent
+		}
 		return &http.Response{
-			StatusCode: http.StatusOK,
-			Body:       ioutil.NopCloser(bytes.NewReader(combinedTrace)),
+			StatusCode: statusCode,
+			Body:       ioutil.NopCloser(bytes.NewReader(combinedTraceBytes)),
```
kvrhdn, line 226:
```suggestion
			Body:       io.NopCloser(bytes.NewReader(combinedTraceBytes)),
```
`ioutil.NopCloser` has been moved to `io.NopCloser`, see https://golang.org/doc/go1.16#ioutil
We just removed all occurrences of `ioutil` here #998 🙂

modules/querier/http.go
```diff
@@ -94,7 +97,7 @@ func (q *Querier) TraceByIDHandler(w http.ResponseWriter, r *http.Request) {
 
 	span.SetTag("response marshalling format", util.JSONTypeHeaderValue)
 	marshaller := &jsonpb.Marshaler{}
-	err = marshaller.Marshal(w, resp.Trace)
+	err = marshaller.Marshal(w, resp)
```
kvrhdn, line 100: There is a difference between our responses depending on whether the caller accept protobuf or not: if the caller accepts protobuf we return `HTTP 206` and `resp.Trace` (line 82).

But if the caller requests something else (aka JSON) we marshal the entire `TraceByIDResponse`. Which would be something like:
```json
{
  "trace": ...,
  "blockErrCount": ...
}
```
Why not also return `HTTP 206`? This change breaks our API I think.

modules/frontend/querysharding.go
```diff
@@ -154,25 +160,36 @@ func mergeResponses(ctx context.Context, rrs []RequestResponse) (*http.Response,
...
 
-			if len(combinedTrace) == 0 {
-				combinedTrace = body
-			} else {
-				combinedTrace, _, err = model.CombineTraceBytes(combinedTrace, body, model.TracePBEncoding, model.TracePBEncoding)
-				if err != nil {
-					// will result in a 500 internal server error
-					return nil, errors.Wrap(err, "error combining traces at query frontend")
+			var resp tempopb.TraceByIDResponse
+			err = proto.Unmarshal(body, &resp)
```
kvrhdn, line 176: The querier only seems to marshal the trace part of `TraceByIDResponse` (modules/querier/http.go:82):
```
b, err := proto.Marshal(resp.Trace)
```
How can we unmarshal a full `TraceByIDResponse` here?

modules/frontend/querysharding.go
```diff
@@ -154,25 +160,36 @@ func mergeResponses(ctx context.Context, rrs []RequestResponse) (*http.Response,
...
+	var shardMissCount, totalBlockErrCount int
 	for _, rr := range rrs {
-		if rr.Response.StatusCode == http.StatusOK {
+		partialContent := rr.Response.StatusCode == http.StatusPartialContent
+		if rr.Response.StatusCode == http.StatusOK || partialContent {
 			body, err := io.ReadAll(rr.Response.Body)
 			rr.Response.Body.Close()
 			if err != nil {
 				return nil, errors.Wrap(err, "error reading response body at query frontend")
 			}
```
kvrhdn, line 173: I'm wondering if we should fail the entire request here. If we can't read the body from one of the requests, isn't this also a partial result?
Same for unmarshaling the body.

modules/frontend/querysharding.go
```diff
@@ -154,25 +160,36 @@ func mergeResponses(ctx context.Context, rrs []RequestResponse) (*http.Response,
...
+			err = proto.Unmarshal(body, &resp)
+			if err != nil {
+				return nil, errors.Wrap(err, "error reading response body at query frontend")
+			}
+
+			if partialContent {
+				totalBlockErrCount += int(resp.BlockErrCount)
+				if totalBlockErrCount > maxBlockErrCount {
+					return nil, fmt.Errorf("too many block queries failed (max %d)", maxBlockErrCount)
 				}
```
kvrhdn, line 185: I'm wondering if it's useful to fail on `maxBlockErrCount`. We already did the work: all the queriers returned something. Instead of throwing away the results we can still return whatever we got.

kvrhdn (commented):
Nice work already! Implementing partial results is pretty tricky it seems and it touches a lot of code 😅 

A general concern: passing the amount of blocks that failed seems similar in intention to the [`SearchMetrics`](https://github.com/grafana/tempo/blob/5ab55dbaa53650e6dee6b0a1d8345f80e30b2286/pkg/tempopb/tempo.proto#L53-L58), that is: we want to capture some statistics about what happened in the ingester or querier and pass it all the way up.
Maybe instead of a `uint32 blockErrCount` we could work with a struct that specific for custom metrics about the query. In the future we could also add the amount of blocks we scanned and the GBs or whatever.

mapno:
https://github.com/grafana/tempo/pull/1002 created too many conflicts and a lot of changes had to be done anyway, so moved the PR to https://github.com/grafana/tempo/pull/1007

PR #994, opened 2021-09-29
Nil check overrides module in /status endpoint

mapno:
**What this PR does**:

Nil check overrides module in /status endpoint

Querier and query-frontend do not run the overrides module. The status handler now checks in that module is being run and skips it in case it isn't.

It also modifies the distributed docker-compose to run minio with `minio` command, instead of executing the binary.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott (changes requested):
We should just check for nil and print a message like "overrides not loaded". There's no reason to scan for overrides in these services.

joe-elliott (approved):

PR #953, opened 2021-09-13
Update to Go 1.17

zalegrala:
**What this PR does**:

**Which issue(s) this PR fixes**:
Fixes #885

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #998, opened 2021-09-30
refactor: move from io/ioutil to io and os package

Juneezee:
**What this PR does**:
The `io/ioutil` package has been deprecated in Go 1.16 (See https://golang.org/doc/go1.16#ioutil). Since tempo has upgraded to Go 1.17 (#953), this PR replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages. This change is transparent to end users.

N/A

**Checklist**
- ~~[ ] Tests updated~~
- ~~[ ] Documentation added~~
- ~~[ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`~~

### DIFF
--- a/tempodb/wal/wal.go
+++ b/tempodb/wal/wal.go
@@ -2,7 +2,6 @@ package wal
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"time"
@@ -78,7 +77,7 @@ func New(c *Config) (*WAL, error) {
 
 // RescanBlocks returns a slice of append blocks from the wal folder
 func (w *WAL) RescanBlocks(log log.Logger) ([]*AppendBlock, error) {
-	files, err := ioutil.ReadDir(w.c.Filepath)
+	files, err := os.ReadDir(w.c.Filepath)
 	if err != nil {
 		return nil, err
 	}
@@ -90,7 +89,12 @@ func (w *WAL) RescanBlocks(log log.Logger) ([]*AppendBlock, error) {
 		}
 
 		start := time.Now()
-		level.Info(log).Log("msg", "beginning replay", "file", f.Name(), "size", f.Size())
+		fileInfo, err := f.Info()
+		if err != nil {
+			return nil, err
+		}
+
+		level.Info(log).Log("msg", "beginning replay", "file", f.Name(), "size", fileInfo.Size())
 		b, warning, err := newAppendBlockFromFile(f.Name(), w.c.Filepath)
 
 		remove := false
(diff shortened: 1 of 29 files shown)


kvrhdn:
Hi 👋🏻  Thank you for this, I'll let the CI run.

Did you do this all manually or is there some kind of tool out there?

Juneezee:
> Hi 👋🏻 Thank you for this! I'll let the CI run.
> 
> Did you do this all manually or is there some kind of tool out there?

Hi, thank you! I used the Search function in VSCode

kvrhdn (approved):
Thanks for this, this is a great _first contribution_! Besides my comment below it all looks good 👍🏻

tempodb/wal/wal.go
```diff
@@ -90,7 +89,7 @@ func (w *WAL) RescanBlocks(log log.Logger) ([]*AppendBlock, error) {
 		}
 
 		start := time.Now()
-		level.Info(log).Log("msg", "beginning replay", "file", f.Name(), "size", f.Size())
+		level.Info(log).Log("msg", "beginning replay", "file", f.Name())
```
kvrhdn, line 92: You can grab the file size using [`DirEntry.Info`](https://pkg.go.dev/io/fs#DirEntry) and [`FileInfo.Size`](https://pkg.go.dev/io/fs#FileInfo). Can you add the `size` label again? Then this PR will not change anything 🙂

kvrhdn (changes requested):

tempodb/wal/wal.go
```diff
@@ -90,7 +89,7 @@ func (w *WAL) RescanBlocks(log log.Logger) ([]*AppendBlock, error) {
 		}
 
 		start := time.Now()
-		level.Info(log).Log("msg", "beginning replay", "file", f.Name(), "size", f.Size())
```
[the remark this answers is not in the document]
Juneezee: Yeah, the `os.DirEntry` interface does not have the `Size()` method 😞. If we want to keep the `size` in the log, we would need to do something like this:
```
fileInfo, err := f.Info()
if err != nil {
    return nil, err
}
level.Info(log).Log("msg", "beginning replay", "file", f.Name(), "size", fileInfo.Size())
```

tempodb/wal/wal.go
```diff
@@ -90,7 +89,12 @@ func (w *WAL) RescanBlocks(log log.Logger) ([]*AppendBlock, error) {
...
+		fileInfo, err := f.Info()
+		if err != nil {
+			return nil, err
+		}
+
+		level.Info(log).Log("msg", "beginning replay", "file", f.Name(), "size", fileInfo.Size())
```
Juneezee, line 97: @kvrhdn Done 😃, please take a look again. The `err` here is not `nil` when the file `f` is remove after `os.ReadDir` is called, it's a rare error but still possible sometimes.

PR #999, opened 2021-09-30
Search no root attrs

mdisibio:
**What this PR does**:
This PR removes capturing of `root.<xyz>` search attributes.  This reduces the size of the search data by ~30%, and seems worth it as in practice these attributes were not very useful to search on.  `root.name` and `root.service.name` are still captured because they are used for display.    

Fixes part of #932 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


kvrhdn (approved):
LGTM 🙂

PR #1000, opened 2021-09-30
Reload search blocks and replay search WAL

annanay25:
**What this PR does**:
- Rescan search wal folder and replay all available files
- Handle blocks with no search file
- Add support for encoding & dataEncoding as part of search WAL file name

also
- dibs on PR 1000!

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/ingester/ingester_test.go
+++ b/modules/ingester/ingester_test.go
@@ -212,6 +212,42 @@ func TestWal(t *testing.T) {
 	}
 }
 
+//func TestSearchWAL(t *testing.T) {
+//	tmpDir := t.TempDir()
+//
+//	i, _, _ := defaultIngester(t, tmpDir)
+//	inst, ok := i.getInstanceByID("test")
+//	require.True(t, ok)
+//
+//	// Write wal
+//	err := inst.CutCompleteTraces(0, true)
+//	require.NoError(t, err)
+//	_, err = inst.CutBlockIfReady(0, 0, true)
+//	require.NoError(t, err)
+//
+//	// assert that search WAL is being searched
+//	ctx := user.InjectOrgID(context.Background(), "test")
+//	searchReq := &tempopb.SearchRequest{Tags: map[string]string{
+//		search.SecretExhaustiveSearchTag: "",
+//	}}
+//	results, err := inst.Search(ctx, searchReq)
+//	assert.NoError(t, err)
+//	assert.Greater(t, results.Metrics.InspectedTraces, 0)
+//
+//	// Shutdown
+//	err = i.stopping(nil)
+//	require.NoError(t, err)
+//
+//	// replay wal
+//	i, _, _ = defaultIngester(t, tmpDir)
+//	inst, ok = i.getInstanceByID("test")
+//	require.True(t, ok)
+//
+//	results, err = inst.Search(ctx, searchReq)
+//	assert.NoError(t, err)
+//	assert.Greater(t, results.Metrics.InspectedTraces, 0)
+//}
+
 // TestWalReplayDeletesLocalBlocks simulates the condition where an ingester restarts after a wal is completed
 // to the local disk, but before the wal is deleted. On startup both blocks exist, and the ingester now errs
 // on the side of caution and chooses to replay the wal instead of rediscovering the local block.
--- a/modules/ingester/instance.go
+++ b/modules/ingester/instance.go
@@ -629,7 +629,7 @@ func (i *instance) rediscoverLocalBlocks(ctx context.Context) error {
 
 		i.blocksMtx.Lock()
 		i.completeBlocks = append(i.completeBlocks, ib)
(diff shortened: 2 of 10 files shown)


joe-elliott:
PR #1000!

tempodb/wal/wal.go
```diff
@@ -135,7 +135,14 @@ func (w *WAL) NewFile(blockid uuid.UUID, tenantid string, dir string, name strin
 	if err != nil {
 		return nil, err
 	}
-	return os.OpenFile(filepath.Join(p, fmt.Sprintf("%v:%v:%v", blockid, tenantid, name)), os.O_CREATE|os.O_RDWR, 0644)
+
+	// blockID, tenantID, version, encoding (compression), dataEncoding, searchFileSuffix
+	filename := fmt.Sprintf("%v:%v:%v:%v:%v:%v", blockid, tenantid, "v2", backend.EncNone, "", name)
```
mdisibio, line 140: It might be better to move the filename parsing code from rescan_blocks.go to here, and in the future this can cover the file name generation and parsing for the regular wal files too.

annanay25: I agree. At the moment, the regular wal files already have a `parseFilename()` in [append_block.go](https://github.com/grafana/tempo/blob/6e94c1ffc10c4c817d53f0feb5c23afad11063aa/tempodb/wal/append_block.go#L263) that has some logic around checking the number of splits etc, for backward compatibility reasons when we added new versions. We could simplify that function now that we have dropped support for those older versions.

annanay25: btw, thoughts on dropping the "searchdata" suffix as the sixth split in the filename? we have a separate directory for search data so we shouldn't need that anymore

annanay25: comment addressed. also, dropped the "searchdata" suffix.

modules/ingester/ingester_test.go
```diff
@@ -211,6 +211,42 @@ func TestWal(t *testing.T) {
 	}
 }
 
+//func TestSearchWAL(t *testing.T) {
```
mdisibio, line 214: Do we want to delete this commented out test?

annanay25: Hmm, I had to make a copy of `defaultIngester()` with some search data but lost track of it

tempodb/search/streaming_search_block.go
```diff
@@ -112,15 +114,20 @@ func (s *StreamingSearchBlock) Search(ctx context.Context, p Pipeline, sr *Resul
 		}
 		buf = buf[:r.Length]
 
-		_, err := s.file.ReadAt(buf, int64(r.Start))
+		pagesBuffer, buffer, err = dr.Read(ctx, []common.Record{r}, pagesBuffer, buffer)
```
mdisibio, line 117: This looks like similar functionality to `StreamingSearchBlock.Iterator()`, could we reuse the iterator here?  One downside might be buffer reuse, as the iterator returns a new buffer each time.

tempodb/wal/wal.go
```diff
@@ -139,7 +142,57 @@ func (w *WAL) NewFile(blockid uuid.UUID, tenantid string, dir string, name strin
 	if err != nil {
 		return nil, err
 	}
-	return os.OpenFile(filepath.Join(p, fmt.Sprintf("%v:%v:%v", blockid, tenantid, name)), os.O_CREATE|os.O_RDWR, 0644)
+
+	// blockID, tenantID, version, encoding (compression), dataEncoding
+	filename := fmt.Sprintf("%v:%v:%v:%v:%v", blockid, tenantid, "v2", backend.EncNone, "")
```
mdisibio, line 147: We can delete the unused `name` parameter now?

modules/ingester/instance.go
```diff
@@ -496,7 +502,7 @@ func (i *instance) resetHeadBlock() error {
 		return err
```
mdisibio, line 502: We can remove the unused `"searchdata"` param now?

mdisibio (commented):
Very close to LGTM. Left a few comments.

mdisibio (approved):

tempodb/search/streaming_search_block.go
```diff
@@ -166,15 +168,21 @@ func (s *streamingSearchBlockIterator) Next(_ context.Context) (common.ID, []byt
 
 	// Use unique buffer that can be returned to the caller.
 	// This is primarily for DedupingIterator which uses 2 buffers at once.
-	buffer := make([]byte, currentRecord.Length)
-	_, err := s.file.ReadAt(buffer, int64(currentRecord.Start))
+	var buffer []byte
+	s.pagesBuffer[0] = make([]byte, currentRecord.Length)
```
zalegrala, line 172: Is the `[0]` here to reuse the memory?

PR #1004, opened 2021-10-01
Ha operational run mode

zalegrala:
**What this PR does**:
WIP

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


cmd/tempo/app/modules.go
```diff
@@ -45,6 +45,7 @@ const (
 	Store         string = "store"
 	MemberlistKV  string = "memberlist-kv"
 	All           string = "all"
+	HA            string = "ha"
```
joe-elliott, line 48: ha-single-binary?
scaleable-single-binary?

just `ha` is not very descriptive.

cmd/tempo/app/modules.go
```diff
@@ -138,16 +139,21 @@ func (t *App) initIngester() (services.Service, error) {
 }
 
 func (t *App) initQuerier() (services.Service, error) {
+
 	// validate worker config
-	// if we're not in single binary mode and worker address is not specified - bail
-	if t.cfg.Target != All && t.cfg.Querier.Worker.FrontendAddress == "" {
-		return nil, fmt.Errorf("frontend worker address not specified")
-	} else if t.cfg.Target == All {
+	switch t.cfg.Target {
```
joe-elliott, line 144: no reason to change this. we could probably improve the comment since there's a new mode, but the logic is fine.

cmd/tempo/app/modules.go
```diff
@@ -290,6 +301,7 @@ func (t *App) setupModuleManager() error {
 	mm.RegisterModule(Compactor, t.initCompactor)
 	mm.RegisterModule(Store, t.initStore, modules.UserInvisibleModule)
 	mm.RegisterModule(All, nil)
+	mm.RegisterModule(HA, t.initHA)
```
joe-elliott, line 304: why not follow the same pattern as `All` and pass nil here? Is there a need for the init function?

cmd/tempo/main.go
```diff
@@ -176,10 +176,14 @@ func loadConfig() (*app.Config, error) {
...
 	// if we're in single binary mode we're going to force some settings b/c nothing else makes sense
-	if config.Target == app.All {
+	switch config.Target {
+	case app.All:
 		config.Ingester.LifecyclerConfig.RingConfig.KVStore.Store = "inmemory"
 		config.Ingester.LifecyclerConfig.RingConfig.ReplicationFactor = 1
 		config.Ingester.LifecyclerConfig.Addr = "127.0.0.1"
+	case app.HA:
+		// config.Ingester.LifecyclerConfig.RingConfig.KVStore.Store = "memberlist"
+		config.Ingester.LifecyclerConfig.Addr = "localhost"
```
joe-elliott, line 186: we don't want anything here. previous logic is fine.

cmd/tempo/app/modules.go
zalegrala: I agree, I was mostly exploring.

cmd/tempo/app/modules.go
zalegrala: What do you think about just "scaleable"?

joe-elliott: I would like to communicate that it's a "single binary" or "all in one" kind of deployment.

zalegrala:
What other documentation do we want to include here?  We don't currently say much about the `target=all`, so I'm wondering how much needs to be said to differentiate.

zalegrala:
I've included some brief docs about our run modes.

integration/microservices/README.md
```diff
@@ -1,35 +1,39 @@
 # tempo-load-test
```
zalegrala, line 1: I used `prettier` to format the file.  How do we feel about the markdown style change here?

docs/tempo/website/operations/deployment.md
```diff
@@ -32,4 +32,45 @@ Helm charts are available in the grafana/helm-charts repo:
...
+
+* Single Binary
+* Scalable Single Binary
+* Microservices
+
+These modes are controlled by the runtime configuration `target`, or using the CLI flag `-target`.  The default target is `all`, which we refer to as "Single Binary".
+
+## Single Binary
+
+This mode is the simplest to get started, in which all of the top-level components, referenced in the [Architecture](architecture) section, are run within the same instance of Tempo.
```
joe-elliott, line 49: does this link work? I've been using that "relref" tag. Something like:

`[tenant index]({{< relref "../operations/polling" >}})`

docs/tempo/website/operations/deployment.md
```diff
@@ -32,4 +32,45 @@ Helm charts are available in the grafana/helm-charts repo:
...
+querier:
+  frontend_worker:
+    frontend_address: tempo.lab.example.com:9095
+```
+
+Each of the `queriers` will perform a DNS lookup for the `frontend_address` and connect to addresses found within the DNS record.
+
+## Microservices
+
+In this mode, each component is deployed in only one instance.  For example, to deploy a `querier`, the Tempo configuration would either contain `target: querier` or the binary is to be executed with `-target=querier`.  Each of the comonents referenced in the [Architecture](architecture) must be deployed in order to get a working Tempo environment.
```
joe-elliott, line 76: link to our dockercompose example that does a microservice deployment.

docs/tempo/website/operations/deployment.md
```diff
@@ -32,4 +32,45 @@ Helm charts are available in the grafana/helm-charts repo:
...
+
+These modes are controlled by the runtime configuration `target`, or using the CLI flag `-target`.  The default target is `all`, which we refer to as "Single Binary".
+
+## Single Binary
+
+This mode is the simplest to get started, in which all of the top-level components, referenced in the [Architecture](architecture) section, are run within the same instance of Tempo.
+
+## Scalable Single Binary
+
+This mode allows the simplicity of the Single Binary mode, but utilizes `memberlist` in order to cluster the components.  The following configuration is required to use the memberlist.
```
joe-elliott, line 53: i don't think we should mention memberlist explicitly. there are a lot of ways to cluster tempo. let's just talk about how it packages all tempo components into one package that can be scaled horizontally.

also create and link to a docker-compose example that runs in this mode.

integration/e2e/e2e_test.go
```diff
@@ -240,6 +241,87 @@ func TestMicroservices(t *testing.T) {
...
+	}
+
+	apiClient1 := tempoUtil.NewClient("http://"+tempo1.Endpoint(3200), "")
+
+	queryAndAssertTrace(t, apiClient1, info)
+
+	err = tempo1.Stop()
+	require.NoError(t, err)
+
+	err = tempo2.Stop()
```
joe-elliott, line 318: after you stop one of the replicas confirm that you can still push and query traces. RF=3 should allow for this.

integration/microservices/README.md
joe-elliott: Possibly useful if we built it into CI, but I don't think it's a big deal.

joe-elliott (commented):
So far nice work. Have a few cleanup requests on the doc and then let's add a docker-compose example that uses this new mode so people can see a simple way to set it up.

example/docker-compose/scalable-single-binary/docker-compose.yaml
```diff
@@ -0,0 +1,98 @@
+version: "3"
+services:
+
+  tempo1:
+    image: grafana/tempo:latest
+    command: "-target=ha -config.file=/etc/tempo.yaml"
```
mdisibio, line 6: Need to update target command line arg (several places)

PR #1006, opened 2021-10-04
Reduce traceInfo maxBatchesPerWrite

zalegrala:
Here we reduce the number of batches that could get sent to Tempo by default.
Without this change, we increase the potential of sending too large of traces
for the defaults of the Tempo service.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/util/trace_info.go
```diff
@@ -19,7 +19,7 @@ var (
 	// that we get the expected number of batches on a trace.  A value larger
 	// than 25 here results in vulture writing traces that exceed the maximum
```
mapno, line 20:
```suggestion
	// than 10 here results in vulture writing traces that exceed the maximum
```

mapno (commented):
Can this be configured or calculated somehow? Instead of a hardcoded value.

zalegrala:
One way I was thinking about retrieving this value would be to query the API of Tempo to know what the configured value is and then determine something appropriate here.

joe-elliott:
> One way I was thinking about retrieving this value would be to query the API of Tempo to know what the configured value is and then determine something appropriate here.

Feels too complex to me. Fine with just reducing the default.

joe-elliott (approved):

PR #1013, opened 2021-10-06
drone: generate drone.yml from jsonnet

kvrhdn:
**What this PR does**:
Add `drone.jsonnet` to generate `drone.yml`. This should be easier to manage as we can work with for-loops and if-cases to aviod repetition.

I've tried to minimise changes in `drone.yml`, all values should be the same, they have just changed order a bit.

~~Fixes #<issue number>~~

**Checklist**
- [ ] ~~Tests updated~~
- [ ] ~~Documentation added~~

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -1,0 +1,128 @@
+local apps = ['tempo', 'tempo-vulture', 'tempo-query'];
+local archs = ['amd64', 'arm64'];
+
+## Building blocks ##
+
+local pipeline(name, arch = 'amd64') = {
+  kind: 'pipeline',
+  name: name,
+  platform: {
+    os: 'linux',
+    arch: arch,
+  },
+  steps: [],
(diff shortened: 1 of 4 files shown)


kvrhdn:
I feel you 😅  I tried to keep the Jsonnet as simple as possible, but it's still difficult to read at first. In this case you can at least easily see what you're generating and you can choose to read either the Jsonnet or the YAML

joe-elliott:
Worst case we just kill the jsonnet and still have the yaml. Let's give this a shot.

kvrhdn:
> Worst case we just kill the jsonnet and still have the yaml. Let's give this a shot.

Good point :+1:

> Hard to say what kind of further improvements are possible, unless there is a drone jsonnet lib available.

I'm hoping this will avoid the risk of updating one of the two build pipelines. This should also make it easier to move more of CI into Drone.

PR #1040, opened 2021-10-14
Run Release GitHub Action workflow only on tag pushes

yvrhdn:
**What this PR does**:
I noticed we are running the Release action on every push to main. This is needless as the steps that do something are disabled with a `if` clause.
List of workflow runs: https://github.com/grafana/tempo/actions/workflows/release.yml

This modifies the `on` condition so the workflow only runs when a tag matching `v*` is pushed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,32 +2,28 @@ name: Release
 on:
   push:
     tags:
-      - '**'
-    branches:
-      - main
+      - v*
 
 jobs:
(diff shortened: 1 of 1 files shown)


kvrhdn:
Btw, I based it on this doc on how the `on.push` syntax works: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

PR #1046, opened 2021-10-15
Only fire TempoNoTenantIndexBuilders if blocklist is not empty

kvrhdn:
**What this PR does**:
Tweak the expression of `TempoNoTenantIndexBuilders` so it only fires if there a no tenant index builders _and_ the blocklist is not empty.

In a Tempo cluster that has not received any data yet, there will also not be any tenant index builders.

### DIFF
--- a/operations/tempo-mixin/alerts.libsonnet
+++ b/operations/tempo-mixin/alerts.libsonnet
@@ -126,7 +126,8 @@
           {
             alert: 'TempoNoTenantIndexBuilders',
             expr: |||
-              sum by (%(group_by_tenant)s) (tempodb_blocklist_tenant_index_builder{}) == 0
+              sum by (%(group_by_tenant)s) (tempodb_blocklist_tenant_index_builder{}) == 0 and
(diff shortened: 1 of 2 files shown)


joe-elliott:
> Concern: how does tempodb_blocklist_length get updated, is it by the tenant index builder itself? The blocklist poller updates tempodb_blocklist_length, but if there are no tenant index builders will there also be no poller?

All components that poll the blocklist set `tempodb_blocklist_length`. Only the few components that actually do a full listing of the backend raise `tempodb_blocklist_tenant_index_builder`. 

Unsure about this change. It will technically work, but I wonder if there's a more direct way to test it. In any case I think this would be better:
```
max by (%(group_by_cluster)s) (tempodb_blocklist_length{}) > 0
```
So we only ignore tenant index builders if there are 0 blocks across all tenants in the entire cluster.

joe-elliott (approved):

kvrhdn:
Okay, I've confirmed this works while setting up a new Tempo cluster.

PR #1068, opened 2021-10-21
Search: drop use of TagCache, extract tags and tag values on-demand

yvrhdn:
**What this PR does**:
Instead of storing tags and tag values in the TagCache we can extract them directly from live traces, WAL and local blocks when requested. This should lower the amount of work spent on updating the tagcache during ingest.

Since tag values are extracted on-demand, we are not limited anymore by the size of the tag cache. So this makes it possible to return thousands of values when needed.

~~Fixes #<issue number>~~

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,8 +109,14 @@ jobs:
 
       - name: Install jsonnet, jsonnet-bundler & tanka
         run: |
-          brew install jsonnet tanka
+          curl -fSL -o jsonnet.tar.gz https://github.com/google/jsonnet/releases/download/v0.17.0/jsonnet-bin-v0.17.0-linux.tar.gz
+          tar -xvf jsonnet.tar.gz -C /usr/local/bin/
+          chmod a+x /usr/local/bin/jsonnet
+          
           go install github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb@v0.4.0
+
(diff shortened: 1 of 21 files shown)


kvrhdn:
Still to address: when fetching tag values with very high cardinality (for instance `http.url`) requests fail/time out when being sent between components. I.e. the ingester is fine digging up all the values, but the GRPC connection between the ingester, querier and query-frontend was hitting the max bytes size...
We should add some limits (optionally configurable) to protect against this happening. Ideally, we return early when we have 'a lot of' values.

Future improvements (not for this PR): we can add filter capabilities to the tag values endpoint, so if someone starts typing `http.url=/myService/` we only return values containing `/myService/`. This would make high cardinality tags more useable.

joe-elliott (approved):
Remove the `TagCache` struct?

kvrhdn:
> Remove the `TagCache` struct?

Yep :+1:, I've removed `tempodb/search/tag_cache.go`, it's all the way at the bottom of the diff
https://github.com/grafana/tempo/pull/1068/files#diff-b29b479cf004a8bf08fb55114702b94f6d818911fca0c83c985e771db2fbe4bb

mdisibio (approved):
Thanks for getting the limit check in.  Good to merge.

PR #1089, opened 2021-10-28
Include metrics for configured limit overrides

zalegrala:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * [ENHANCEMENT] Add middleware to compress frontend HTTP responses with gzip if requested [#1080](https://github.com/grafana/tempo/pull/1080) (@kvrhdn, @zalegrala)
 * [ENHANCEMENT] Allow query disablement in vulture [#1117](https://github.com/grafana/tempo/pull/1117) (@zalegrala)
 * [ENHANCEMENT] Improve memory efficiency of compaction and block cutting. [#1121](https://github.com/grafana/tempo/pull/1121) (@joe-elliott)
+* [ENHANCEMENT] Include metrics for configured limit overrides: tempo_limits_overrides [#1089](https://github.com/grafana/tempo/pull/1089) (@zalegrala)
 * [BUGFIX] Fix defaults for MaxBytesPerTrace (ingester.max-bytes-per-trace) and MaxSearchBytesPerTrace (ingester.max-search-bytes-per-trace) (@bitprocessor)
 * [BUGFIX] Ignore empty objects during compaction [#1113](https://github.com/grafana/tempo/pull/1113) (@mdisibio)
 
@@ -83,7 +84,6 @@
 * [ENHANCEMENT] Include tempo-cli in the release [#1086](https://github.com/grafana/tempo/pull/1086) (@zalegrala)
 * [ENHANCEMENT] Add search on span status [#1093](https://github.com/grafana/tempo/pull/1093) (@mdisibio)
 * [ENHANCEMENT] Slightly improved compression performance [#1094](https://github.com/grafana/tempo/pull/1094) (@bboreham)
-* [ENHANCEMENT] Include metrics for configured limit overrides [#1089](https://github.com/grafana/tempo/pull/1089) (@zalegrala)
 * [BUGFIX] Update port spec for GCS docker-compose example [#869](https://github.com/grafana/tempo/pull/869) (@zalegrala)
 * [BUGFIX] Fix "magic number" errors and other block mishandling when an ingester forcefully shuts down [#937](https://github.com/grafana/tempo/issues/937) (@mdisibio)
 * [BUGFIX] Fix compactor memory leak [#806](https://github.com/grafana/tempo/pull/806) (@mdisibio)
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -78,6 +78,7 @@
 * [ENHANCEMENT] Include tempo-cli in the release [#1086](https://github.com/grafana/tempo/pull/1086) (@zalegrala)
 * [ENHANCEMENT] Add search on span status [#1093](https://github.com/grafana/tempo/pull/1093) (@mdisibio)
 * [ENHANCEMENT] Slightly improved compression performance [#1094](https://github.com/grafana/tempo/pull/1094) (@bboreham)
+* [ENHANCEMENT] Include metrics for configured limit overrides [#1089](https://github.com/grafana/tempo/pull/1089) (@zalegrala)
```
annanay25, line 81: Might be good to mention the metric name here as well.

modules/overrides/overrides.go
```diff
@@ -214,43 +223,57 @@ func (o *Overrides) IngestionRateStrategy() string {
 // MaxLocalTracesPerUser returns the maximum number of traces a user is allowed to store
 // in a single ingester.
 func (o *Overrides) MaxLocalTracesPerUser(userID string) int {
-	return o.getOverridesForUser(userID).MaxLocalTracesPerUser
+	x := o.getOverridesForUser(userID).MaxLocalTracesPerUser
+	metricOverridesLimits.WithLabelValues("max_local_traces_per_user", userID).Set(float64(x))
```
annanay25, line 227: I think it would be better if we recorded these values when they are periodically loaded from the overrides file (I think in the `loadPerTenantOverrides` function above)  since that's the only time those values can change, rather than on each retrieval as its done currently.

modules/overrides/overrides.go
```diff
@@ -12,11 +12,20 @@ import (
...
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"gopkg.in/yaml.v2"
 )
 
 const wildcardTenant = "*"
 
+var (
+	metricOverridesLimits = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "tempo",
+		Name:      "limits",
```
annanay25, line 24: bikeshedding on naming but should we name this `overrides` so it's consistent with cortex?
https://github.com/cortexproject/cortex/blob/c7845aedc3f045a83aa9c27d029391fc4b244d30/pkg/util/validation/exporter.go#L19

CHANGELOG.md
joe-elliott: This needs to move up a section as well. it's still in 1.2.

modules/overrides/overrides.go
joe-elliott: Here is how cortex is doing it:
https://github.com/cortexproject/cortex/blob/master/pkg/util/validation/exporter.go

modules/overrides/overrides.go
zalegrala: Can do.

modules/overrides/overrides.go
zalegrala: Updating the metrics on file load makes sense to me.  Do we want to introduce a channel for this to copy cortext as well?

zalegrala:
I've added a metric for default limits as well.

modules/overrides/overrides.go
joe-elliott: > Do we want to introduce a channel for this to copy context as well?

Unsure what you're asking, but I would like to use the exporter pattern like the linked cortex code.

zalegrala:
I've made an update to implement the Overrides module as a Collector.  We could adjust to have the exporter be its own module /package like is done in the cortex code, but I think that might be a little more than we need for this.  What do you think?

modules/overrides/overrides.go
```diff
@@ -71,6 +87,14 @@ func NewOverrides(defaults Limits) (*Overrides, error) {
 	var manager *runtimeconfig.Manager
 	subservices := []services.Service(nil)
 
+	metricDefaultsLimits.WithLabelValues("max_local_traces_per_user").Set(float64(defaults.MaxLocalTracesPerUser))
```
joe-elliott, line 90: make `max_local_traces_per_user`, etc. constants

zalegrala: This is addressed in the last commit.

joe-elliott: I'm still seeing individual static strings.

zalegrala: Oh sorry, I thought you meant `prometheus.MustNewConstMetric` to make them constants.  You mean make the metric name string a constant to be referenced by both the defaults and the overrides, right?

joe-elliott: yup. I just meant to make a constant string that they both use. I have no idea what `MustNewConstMetric` is but let's just follow whatever Cortex does here.

zalegrala: Yep that's what they are doing.  This `promehteus.Collector` pattern is new to me, but the const bit is document here: https://prometheus.io/docs/instrumenting/writing_exporters/#collectors

zalegrala: In Cortex the exporter is also a module, but as it is now we aren't doing that.  If we want to do that, I think we just need to make a `limits/exporter` package and glue it up, but not sure how far to go with modules.  It would be our first "exporter" module, but it could be a pattern if we want to structure the exporters as modules.  Seems to me that they belong with the module they support, rather than a module in themselves.  What do you think?

joe-elliott (approved):
nice! i think we're there!

PR #1093, opened 2021-11-01
Support error and status.code searches

mdisibio:
**What this PR does**:
This PR adds support for searching on span status.   Raw span status in OTLP is an integer, and this is recorded as the new tag `status.code=<int>` (this matches the way Grafana already displays this data).   This can be searched directly, or via the following more user-friendly methods:
1.  `status.code=unset|ok|error` - The search pipeline will rewrite the text versions into a lookup for the corresponding integer. The tag name/value lookup apis will return these user-friendly texts and not the integers.
2. `error=true|false` - A new virtual "error" tag can be searched that is rewritten into a lookup on status.code integer. Similarly, the tag name/value lookup apis will return this tag and the two values.

Fixes part of #932 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/search_data.go
+++ b/modules/distributor/search_data.go
@@ -67,7 +67,7 @@ func extractSearchData(trace *tempopb.Trace, id []byte, extractTag extractTagFun
 				// Collect for any spans
 				data.AddTag(search.SpanNameTag, s.Name)
 				if s.Status != nil {
-					data.AddTag(search.StatusCodeTag, strconv.FormatInt(int64(s.Status.Code), 10))
+					data.AddTag(search.StatusCodeTag, strconv.Itoa(int(s.Status.Code)))
 				}
 				data.SetStartTimeUnixNano(s.StartTimeUnixNano)
 				data.SetEndTimeUnixNano(s.EndTimeUnixNano)
--- a/modules/distributor/search_data.go
+++ b/modules/distributor/search_data.go
@@ -66,6 +66,9 @@ func extractSearchData(trace *tempopb.Trace, id []byte, extractTag extractTagFun
 
 				// Collect for any spans
 				data.AddTag(search.SpanNameTag, s.Name)
+				if s.Status != nil {
+					data.AddTag(search.StatusCodeTag, strconv.FormatInt(int64(s.Status.Code), 10))
+				}
 				data.SetStartTimeUnixNano(s.StartTimeUnixNano)
 				data.SetEndTimeUnixNano(s.EndTimeUnixNano)
 
--- a/modules/querier/querier.go
+++ b/modules/querier/querier.go
@@ -18,6 +18,7 @@ import (
 	"github.com/grafana/tempo/pkg/model"
 	"github.com/grafana/tempo/pkg/tempopb"
 	"github.com/grafana/tempo/pkg/validation"
+	"github.com/grafana/tempo/tempodb/search"
 	"github.com/opentracing/opentracing-go"
 	ot_log "github.com/opentracing/opentracing-go/log"
 	"github.com/pkg/errors"
@@ -309,6 +310,11 @@ func (q *Querier) SearchTags(ctx context.Context, req *tempopb.SearchTagsRequest
 		}
 	}
 
+	// Extra hard-coded values (for now)
+	for _, k := range search.GetStaticTags() {
+		uniqueMap[k] = struct{}{}
+	}
+
 	// Final response (sorted)
 	resp := &tempopb.SearchTagsResponse{
 		TagNames: make([]string, 0, len(uniqueMap)),
@@ -325,6 +331,14 @@ func (q *Querier) SearchTagValues(ctx context.Context, req *tempopb.SearchTagVal
 	_, err := user.ExtractOrgID(ctx)
 	if err != nil {
 		return nil, errors.Wrap(err, "error extracting org id in Querier.SearchTagValues")
+	}
+
+	// Hard-coded values?
+	staticValues := search.GetStaticTagValues(req.TagName)
+	if len(staticValues) > 0 {
+		return &tempopb.SearchTagValuesResponse{
+			TagValues: staticValues,
+		}, nil
 	}
 
 	replicationSet, err := q.ring.GetReplicationSetForOperation(ring.Read)
--- a/modules/querier/querier.go
+++ b/modules/querier/querier.go
@@ -310,8 +310,8 @@ func (q *Querier) SearchTags(ctx context.Context, req *tempopb.SearchTagsRequest
 		}
 	}
 
-	// Extra hard-coded values (for now)
-	for _, k := range search.GetStaticTags() {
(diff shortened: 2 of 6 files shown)


modules/distributor/search_data.go
```diff
@@ -66,6 +66,9 @@ func extractSearchData(trace *tempopb.Trace, id []byte, extractTag extractTagFun
 
 				// Collect for any spans
 				data.AddTag(search.SpanNameTag, s.Name)
+				if s.Status != nil {
+					data.AddTag(search.StatusCodeTag, strconv.FormatInt(int64(s.Status.Code), 10))
```
kvrhdn, line 70: Nit: this can be simplified a tiny bit:

```suggestion
					data.AddTag(search.StatusCodeTag, strconv.Itoa(int(s.Status.Code)))
```

Though I'm not sure it's worth it, under the hood this will cast back the value to `int64` before calling `FormatInt`.

modules/querier/querier.go
```diff
@@ -309,6 +310,11 @@ func (q *Querier) SearchTags(ctx context.Context, req *tempopb.SearchTagsRequest
 		}
 	}
 
+	// Extra hard-coded values (for now)
+	for _, k := range search.GetStaticTags() {
+		uniqueMap[k] = struct{}{}
+	}
```
kvrhdn, line 316: This should only be necessary for the virtual tags right? I'd leave out `status.code` since it will be indexed correctly and actually exists in the ingesters.

WDYT about renaming `GetStaticTags` -> `GetVirtualTags`? That should make it more clear why this is needed here.

modules/querier/querier.go
```diff
@@ -327,6 +333,14 @@ func (q *Querier) SearchTagValues(ctx context.Context, req *tempopb.SearchTagVal
 		return nil, errors.Wrap(err, "error extracting org id in Querier.SearchTagValues")
 	}
 
+	// Hard-coded values?
+	staticValues := search.GetStaticTagValues(req.TagName)
+	if len(staticValues) > 0 {
+		return &tempopb.SearchTagValuesResponse{
+			TagValues: staticValues,
+		}, nil
+	}
```
kvrhdn, line 342: Instead of returning early, what do you think about still querying the ingesters and combining both lists?
I can imagine a case in which someone uses the `error` tag to capture the error message (which is perfectly fine since this is not reserved in OTel). In this case we should still show all tag values for `error` imo.

tempodb/search/pipeline.go
```diff
@@ -91,6 +87,54 @@ func NewSearchPipeline(req *tempopb.SearchRequest) Pipeline {
...
+			// Not a bool string = fall-through
+			break
+		}
+
+		if b {
+			// Error = true
+			return false, StatusCodeTag, strconv.Itoa(int(v1.Status_STATUS_CODE_ERROR))
+		} else {
+			// Error = false
+			return false, StatusCodeTag, strconv.Itoa(int(v1.Status_STATUS_CODE_OK))
```
kvrhdn, line 121: I'm worried this will be confusing: OTel only sets STATUS_CODE_OK if someone explicitly marks the operation as successful. By default every operation that is not an error gets STATUS_CODE_UNSET.
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status

I'd expect `error=false` to mean STATUS_CODE_UNSET _or_ STATUS_CODE_OK, but we can't express that yet 😞  But this means that searching `error=false` will most likely not do what you expect it to do.

Not sure what the best alternative is though...

kvrhdn (commented):
Left some comments/concerns, but overall this is really nice work. I like the concept of virtual tags to match the model Grafana uses when displaying the trace.

modules/querier/querier.go
mdisibio: I did originally implement it to call the ingesters and combine both lists (like `SearchTags`), but in the case of `status.code` it seemed better to do a full value replacement (else the autocomplete lists also contains '0', '2'...), and hence the ingester calls were unnecessary and it moved to exit early.  We could split the functionality up where some tags exit early and some merge (maybe names like `GetStaticTagValues` and `GetExtraTagValues`), or just have `status.code` merge and show all values in the autocomplete. Thoughts?

modules/querier/querier.go
mdisibio: I did consider naming it `GetVirtualTags` but left it matching `GetxxxTagValues` because they are related.  Based on how we address the other [comment](https://github.com/grafana/tempo/pull/1093#discussion_r740927391) yes I'm on board with renaming/changes in this area.  Also why I left it populating `status.code` (for symmetry).

tempodb/search/pipeline.go
mdisibio: Ah ok thanks for clarifying this behavior. I noticed it in our test data but didn't realize the convention.  Agree we can't express it yet 😞 Let's keep thinking and discussing.

modules/querier/querier.go
kvrhdn: Oh yeah, I didn't think about the `status.code` tag.

I'm favouring returning both the actual values (`0`, `1`, `2`) and the virtual values (`unset`, `ok`,...) because that is most transparent:
- the trace view currently shows `span.status` with their numeric value, so people might be more used to seeing these.
- if we don't show the labels, they might not know it's possible to use these as well.

I just hope seeing both numbers and labels isn't confusing: will it be clear that they are aliases of each other?

modules/querier/querier.go
mdisibio: Renamed according to discussion.

modules/querier/querier.go
mdisibio: Updated to merge all virtual values according to discussion. Even though the list will contain all values and might be a little confusing, it is also the most transparent and flexible.

tempodb/search/pipeline.go
```diff
@@ -91,6 +87,52 @@ func NewSearchPipeline(req *tempopb.SearchRequest) Pipeline {
...
+		if err != nil {
+			// Not a bool string = fall-through
+			break
+		}
+
+		if b {
+			// Error = true
+			return false, StatusCodeTag, strconv.Itoa(int(v1.Status_STATUS_CODE_ERROR))
+		}
+		// Else fall-through
```
kvrhdn, line 120: I think we can simplify this a little bit:
```go
case ErrorTag:
	// Convert error=true into status.code=2
	if v == "true" {
		return false, StatusCodeTag, strconv.Itoa(int(v1.Status_STATUS_CODE_ERROR))
	}
	// Else fall-through
```

This will be a bit more restrictive because `strconv.ParseBool` also accepts `t` and `1` as `true`, but I don't think that's a bad thing. WDYT?

kvrhdn (approved):
Besides a possible simplification this looks good to me :+1:

I like how this turned out, with the virtual tags and values we don't have to muck with the actual ingested data 😌

PR #1094, opened 2021-11-02
Call go-kit/log instead of go-kit/kit/log to fix spurious debug logs

bboreham:
Because some upstream code is now using `go-kit/log` instead of `go-kit/kit/log`, some debug logging from that code will pass through the level filter. `go-kit/kit` v0.12.0 addresses this by making `go-kit/kit/log` just a pass-through to `go-kit/log`, but we need to update all the calls in Tempo otherwise golangci-lint fails because the passthrough is also marked deprecated.

Updates a number of dependencies, including the following:
* github.com/aws/aws-sdk-go from v1.38.68 to v1.40.45
* github.com/go-kit/log from v0.1.0 to v0.2.0
* github.com/json-iterator/go from  v1.1.11 to v1.1.12
* github.com/hashicorp/go-hclog from v0.14.0 to v0.16.2
* golang.org/x/net from ad29c8ab022f to 978cfadd31cf

None of these changes looks material to Tempo.

`github.com/klauspost/compress` updates from v1.13.5 to v1.13.6, [advertising](https://github.com/klauspost/compress/releases/tag/v1.13.6) some performance improvements.

**Checklist**
- NA Tests updated
- NA Documentation added
- [ ] `CHANGELOG.md` updated


mdisibio (approved):
Thanks for the update.  Basic testing is looking good (log level filtering, s3 backend)

PR #1105, opened 2021-11-08
Update golangci linters for deprecation warnings

zalegrala:
**What this PR does**:

Here we make the necessary updates to address the deprecations warning.  Replace `golint` with `revive` and remove `interfacer` due to non-maintenance and limited functionality.  Also sort.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -47,19 +47,18 @@ linters:
     - goconst
     - gofmt
     - goimports
-    - golint
     - gosimple
+    - govet
     - ineffassign
     - megacheck
     - misspell
+    - revive
     - structcheck
+    - typecheck
     - unconvert
     - unparam
-    - varcheck
-    - govet
     - unused
-    - interfacer
-    - typecheck
+    - varcheck
 
 linters-settings:
   errcheck:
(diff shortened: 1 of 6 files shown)


.golangci.yml
```diff
@@ -47,19 +47,18 @@ linters:
     - goconst
     - gofmt
     - goimports
-    - golint
     - gosimple
```
joe-elliott, line 50: was golint the only removed linter? the others seem like they were just moved around?

what checks are lost by removing golint?

joe-elliott (changes requested):

.golangci.yml
zalegrala: Revive is added to replace the deprecation on golint.  According to the docs, revive has more rules than golint.

https://github.com/mgechev/revive

`interfacer` was dropped due to deprecation and doesn't have a replacement.

And then I sorted the list.

joe-elliott (approved):

.golangci.yml
joe-elliott: Ok, I'm good on this. Can you remove the all of the 
```
// nolint:interfacer
```
comments sprinkled around the codebase?

joe-elliott (approved):

joe-elliott:
Nice cleanup. Thanks!

PR #1114, opened 2021-11-12
Include prettier CI action

zalegrala:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


zalegrala:
@KMiller-Grafana The team discussed including `prettier` markdown formater recently and there was some agreement as long as we included it in the CI system as well, but I realize now I didn't discuss this with you.  What do you think about this change?  All the changes to the markdown files were made with the new `make lint-fix` make target.  There are some options with prettier too, so if there is any particular syntax that is distasteful, we might be able to adjust.
https://prettier.io/docs/en/options.html

zalegrala:
Too, there are editor plugins https://prettier.io/docs/en/editors.html

joe-elliott:
@grafana/tempo Thoughts on this? I don't like that it requires installing prettier on my laptop directly, but that's already what we're doing with golangci-lint.

Makefile
```diff
@@ -89,6 +89,11 @@ fmt:
 .PHONY: lint
 lint:
 	$(LINT) run
+	prettier --check $(ALL_MD)
+
+.PHONY: lint-fix
+lint-fix:
```
joe-elliott, line 95: golangci-lint also has an auto-fix option. go ahead and add it here.

zalegrala:
I hear you about installing more tools, but as an attempt to assuage any concerns about it being a toy project, it has over 500 contributors and 41k stars on github.  https://github.com/prettier/prettier

Makefile
zalegrala: Good call, I'll add that.

mattmendick:
This seems like a win to me. A standard on formatting seems to make sense as it does for the go code already.

KMiller-Grafana:
I believe y'all asked for my opinion.  I don't like or want to run any tool that makes formatting decisions on markdown source. I'm following style guidelines set by the Grafana Labs docs team.  If prettier deviates from that style in any way, it makes my job take more time, and makes the Tempo docs into a unicorn that I have to treat differently than others.  

Selfishly, I also don't have or want to take the time to configure prettier.

Here's a question for all involved: what problem does using this tool on markdown source solve?  
I don't see a problem, so I don't yet understand why we'd want to add an extra CI step and check.

mattmendick:
Good point @KMiller-Grafana, I don’t think we should go forward with this change if it will take more time to wrestle with it to make the output acceptable to the docs team standards, especially since there’s probably no easy way to flag something as ignored. 

I think if we could agree on the settings that it uses though, we could apply it across the board and it may have its intended effect of saving the team time because you wouldn’t have to think about enforcing those rules. Perhaps the onus is on the tempo team to prove that we’ve come up with the right settings that match the grafana standards so you can be confident this tool is saving you time, not wasting it.

IMO the reason we like the idea of this so much is because we’ve seen the golangci-lint tool free us from discussions and back and forth in PRs discussing formatting decisions and allow us to just move past that. Has been really freeing.

KMiller-Grafana:
Thanks for the detailed consideration, @mattmendick . Running a linter on code is super.  If there are issues that the team is having with markdown formatting, I'd like to hear the issues.  Maybe a targeted linter (prettier) can help.  Maybe not.  Depends on the issue.

PR #1111, opened 2021-11-11
Edit ObjectCombiner interface to return error

annanay25:
Signed-off-by: Annanay <annanayagarwal@gmail.com>

In an effort to track down a bug related to zero length traces https://github.com/grafana/tempo/issues/1107

PR #1121, opened 2021-11-16
Improve Combine Efficiency

joe-elliott:
**What this PR does**:
- Improves combine efficiency when combining more than 2 objects with the same id by only marshalling to/from proto once
- Adds an error return from the ObjectCombiner.Combine() method and bubbles it up. (replaces #1111)

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/model/combine.go
+++ b/pkg/model/combine.go
@@ -20,7 +20,7 @@ var ObjectCombiner = objectCombiner{}
 var _ common.ObjectCombiner = (*objectCombiner)(nil)
 
 // Combine implements tempodb/encoding/common.ObjectCombiner
-func (o objectCombiner) Combine(dataEncoding string, objs ...[]byte) ([]byte, bool, error) { // jpe test additional cases
+func (o objectCombiner) Combine(dataEncoding string, objs ...[]byte) ([]byte, bool, error) {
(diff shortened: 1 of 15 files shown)


pkg/model/combine.go
```diff
@@ -23,29 +20,48 @@ var ObjectCombiner = objectCombiner{}
 var _ common.ObjectCombiner = (*objectCombiner)(nil)
 
 // Combine implements tempodb/encoding/common.ObjectCombiner
-func (o objectCombiner) Combine(dataEncoding string, objs ...[]byte) ([]byte, bool) {
+func (o objectCombiner) Combine(dataEncoding string, objs ...[]byte) ([]byte, bool, error) {
 	if len(objs) <= 0 {
-		return nil, false
+		return nil, false, errors.New("no objects provided")
```
zalegrala, line 25: Just thinking something about including the method or action we're erroring on.  Something like "unable to combine zero objects" or something.  Unless the caller has the context, in which case perhaps it doesn't matter.

annanay25 (approved):
lgtm, fewer proto <> bytes round trips and also nice tests on deduping iterator.

PR #1130, opened 2021-11-19
Compactor Memory Improvements

joe-elliott:
**What this PR does**:
- Improves memory usage 
- Possibly fixes a bug with calling `b.clear()` on bookmarks that did not represent the lowest id

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/encoding/iterator_multiblock.go
+++ b/tempodb/encoding/iterator_multiblock.go
@@ -100,8 +100,8 @@ func (i *multiblockIterator) iterate(ctx context.Context) {
 
 	for !i.allDone(ctx) {
 		var lowestID []byte
-		var lowestObject []byte
-		var lowestBookmark *bookmark
+		var lowestObjects [][]byte
+		var lowestBookmarks []*bookmark
 
 		// find lowest ID of the new object
 		for _, b := range i.bookmarks {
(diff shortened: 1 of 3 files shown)


tempodb/encoding/iterator_multiblock.go
```diff
@@ -116,20 +116,25 @@ func (i *multiblockIterator) iterate(ctx context.Context) {
...
 		}
 
-		if len(lowestID) == 0 || len(lowestObject) == 0 || lowestBookmark == nil {
+		lowestObject, _, err := i.combiner.Combine(i.dataEncoding, lowestObjects...)
+		if err != nil {
+			i.err.Store(fmt.Errorf("error combining while Nexting: %w", err))
+			return
+		}
+		for _, b := range lowestBookmarks {
+			b.clear()
```
mapno, line 134: I don't understand the logic here. Why are we appending to `lowestBookmarks` if we clear them before using them in the log? Could we use a counter instead?

mapno (commented):
I have a question about the iterator's logic, but LGTM overall.

tempodb/encoding/iterator_multiblock.go
joe-elliott: `lowestBookmarks` is a slice that contains bookmark structs that "point to" the next lowest object for compaction. Clearing them just tells them to move past their current object. The reason we do it here is b/c if the condition below hits and we `continue` then we need to be sure that we've moved these bookmarks to the next thing or we will just get in an infinite loop.

Honestly unsure about the quality of that log. Might make more sense to just log `len(lowestBookmarks)`. The only thing I really wanted was info about why the condition evaluated to true.

mdisibio: The log message previously printed `lowestBookmark`, which mainly was to distinguish between nil or not (when not nil it just prints `unsupported object type`).  It would be similar here, printing nil/empty slice/nil/non-nil entries quite messily.... but quite frankly since we haven't determined root cause for the empty objects yet, I am ok with that and prefer it over a simple `len()`.

mdisibio (approved):
Great fix and improvement, lgtm.

PR #1131, opened 2021-11-19
Backend Search Serverless

joe-elliott:
**What this PR does**:
Adds code to start fleshing out a serverless option for Tempo search. This is the last garbage PR I intend to submit. Next, I will consolidate the two search paths and add appropriate tests, documentation and a backend search feature flag.


tempodb/encoding/iterator_paged.go
```diff
@@ -4,11 +4,13 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"math"
 
 	"github.com/grafana/tempo/tempodb/encoding/common"
 )
 
+const intSize = 32 << (^uint(0) >> 63) // 32 or 64
```
joe-elliott, line 11: This is pulled out of the math package. We had to do this because Google Cloud Functions supports 1.16 as the latest runtime and these constants were defined after that.

modules/frontend/searchsharding.go
```diff
@@ -70,7 +70,7 @@ func (r *searchResponse) addResponse(res *tempopb.SearchResponse) {
 	defer r.mtx.Unlock()
 
 	r.results.Traces = append(r.results.Traces, res.Traces...)
-	r.results.Metrics.InspectedBlocks += res.Metrics.InspectedBlocks
+	// purposefully ignoring InspectedBlocks as that value is set by the sharder
```
annanay25, line 73: I expected to see something like this in the frontend code
```
if queryparams contains "backend_search" {
  // call the cloud function here
  .
  http.get(frontend.cfg.cloud_function_endpoint)
  .
}
```

is that integrated yet or will be done next?

cmd/tempo-serverless/Makefile
```diff
@@ -0,0 +1,25 @@
...
+#   https://github.com/buildpacks/pack
+build-docker: 
+	go mod vendor
+
+	pack build \
+	  --builder gcr.io/buildpacks/builder:v1 \
+	  --env GOOGLE_RUNTIME=go \
+	  --env GOOGLE_FUNCTION_SIGNATURE_TYPE=http \
+	  --env GOOGLE_FUNCTION_TARGET=Handler \
+	  tempo-serverless
```
kvrhdn, line 13: Something to look into later: according to [their installation docs](https://buildpacks.io/docs/tools/pack/), pack can also be run from a container. It would be nice to replace this with a docker run command, so you don't have to install pack if you don't want to (also nice for CI).

cmd/tempo-serverless/handler.go
```diff
@@ -0,0 +1,208 @@
...
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	// load local backend config file. by convention this must be in the same folder at ./config.json
+	reader, err := loadBackend()
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
```
kvrhdn, line 86: Something to look into later: this code is not specific to the incoming request, would it make sense to move this into some kind of init function? I'm not sure how Google Functions work, but with Lambda's you can run expensive initialisation code once and reuse it for multiple invocations.

tempodb/encoding/iterator_paged.go
kvrhdn: Can we add a comment here with something like "use math.MaxInt once Google Cloud Functions supports Go 1.17"?

pkg/model/matches.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+const RootSpanNotYetReceivedText = "<root span not yet received>"
+
+// Matches determines if the passed object encoded using dataEncoding matches the tempopb.SearchRequest.
+//  If the object matches the request then a non-nil tempopb.TraceSearchMetaData is returned. Otherwise
+//  nil is returned.
+func Matches(id []byte, obj []byte, dataEncoding string, reqStart, reqEnd uint32, req *tempopb.SearchRequest) (*tempopb.TraceSearchMetadata, error) {
+	start := uint64(math.MaxUint64)
+	end := uint64(0)
```
kvrhdn, line 21: On a first read I was confusing these two parameters with start and end from the search request, maybe consider renaming them to `traceStart` and `traceEnd` to clarify these are timestamps?

cmd/tempo-serverless/Makefile
```diff
@@ -0,0 +1,25 @@
...
+
+	rm -rf vendor
```
kvrhdn, line 15: Is there a particular reason to not check in the vendor folder? We do it for our top-level modules.

kvrhdn (commented):
Cool stuff! I only have a couple of comments/questions and a few suggestions for future PRs.

tempodb/encoding/iterator_paged.go
joe-elliott: - Changed go.mod to be 1.16
- Added requested comment
- Also had to drop use of time.UnixMicro()

cmd/tempo-serverless/handler.go
```diff
@@ -0,0 +1,208 @@
...
+	marshaller := &jsonpb.Marshaler{}
+	err = marshaller.Marshal(w, resp)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
+
+func loadBackend() (backend.Reader, error) {
+	// horrible viper dance since it won't unmarshal to a struct from env: https://github.com/spf13/viper/issues/188
```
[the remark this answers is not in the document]
joe-elliott: I can't find a way in Google Cloud Functions to deliver a config file alongside the binary/code. It might be impossible. However, setting environment variables is supported across all serverless cloud functions which is why we are supporting that.

cmd/tempo-serverless/handler.go
```diff
@@ -0,0 +1,208 @@
...
+	cfg := &tempodb.Config{}
+	v.Unmarshal(cfg, setTagName)
+
+	var r backend.RawReader
+
+	switch cfg.Backend {
+	case "local":
+		err = fmt.Errorf("local backend not supported for serverless functions")
+	case "gcs":
+		r, _, _, err = gcs.NewNoConfirm(cfg.GCS)
```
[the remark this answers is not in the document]
joe-elliott: Comment added!

cmd/tempo-serverless/readme.md
```diff
@@ -0,0 +1,19 @@
+# tempo-servless
```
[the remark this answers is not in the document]
joe-elliott: Environment variables. I believe all major cloud providers support this. They will look like this:

```
TEMPO_GCS_BUCKET_NAME=some-random-gcs-bucket
TEMPO_GCS_CHUNK_BUFFER_SIZE=10485760
TEMPO_GCS_HEDGE_REQUESTS_AT=400ms
TEMPO_BACKEND=gcs
```

Adding to the readme!

modules/frontend/searchsharding.go
joe-elliott: I will be doing that next. The next PR is going to focus on:
- adding all config options (including option to add a serverless endpoint)
- testing
- collapsing the backend search query path into the standard query path

cmd/tempo-serverless/Makefile
joe-elliott: I spent some time trying this and was getting permissions errors. Agree that this needs some love before it's integrated into CI.

cmd/tempo-serverless/handler.go
joe-elliott: Dang. Good catch. Used `sync.Once()` to prevent reinitialization.

pkg/model/matches.go
joe-elliott: +1

cmd/tempo-serverless/Makefile
joe-elliott: I'm honestly not sure the best way to handle this. The vendored folder actually includes tempo files b/c this module vendors Tempo. For now I think I'd rather leave them out while we figure out best practices with nested go modules.

kvrhdn (commented):
I'm good with merging this :+1:
I'll wait with approving to give @annanay25 a chance to review

cmd/tempo-serverless/go.mod
```diff
@@ -0,0 +1,148 @@
...
+	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
+	google.golang.org/api v0.58.0 // indirect
+	google.golang.org/appengine v1.6.7 // indirect
+	google.golang.org/genproto v0.0.0-20211016002631-37fc39342514 // indirect
+	google.golang.org/grpc v1.40.0 // indirect
+	google.golang.org/protobuf v1.27.1 // indirect
+	gopkg.in/ini.v1 v1.57.0 // indirect
+)
+
+replace github.com/grafana/tempo => ../../
```
annanay25, line 135: maybe for a future PR: Should we version this somehow?

annanay25 (approved):

PR #1137, opened 2021-11-25
tempo-mixin: add panel for Envoy Proxy sidecar, update tanka example …

kvrhdn:
**What this PR does**:

Changes:

### Tempo / Writes dashboard

Add a panel with Envoy metrics. This panel is to be used when the gateway is run with an Envoy Proxy sidecar (for HTTP2 load balancing).
The metrics from Envoy expose the gRPC status codes, which are essential pieces of information when debugging ingress issues. The gateway only reports HTTP statuses, which is deceptive since a 200 OK HTTP packet might contain a gRPC message with an error.
Example: if a request is denied by the ingester because it hits the limit, the distributor will return a HTTP 200 OK message containing a gRPC message with status 9 FAILED PRECONDITION.

I've also added a text panel listing the gRPC status codes since these aren't common knowledge.
It links to this doc: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

### tanka example

- update prometheus config to always add `cluster` label, the tempo-mixin dashboards depend on this label
- enable search in Grafana and Tempo
- use a relative link to tempo-mixin, this makes it easier to quickly test changes

I've also run `jb update` in tempo-mixin and tk examples, hence the large amount of vendor file being changed.

~~Fixes #<issue number>~~

**Checklist**
- [ ] ~~Tests updated~~
- [ ] ~~Documentation added~~


kvrhdn:
Btw, the end result looks a little bit like this:

Screenshot 2021-11-26 at 00 57 52

Interesting things to note here:

- the distributor has a small bump in refused spans starting around 23:00
- the gateway does not report this, it's all 200s there
- the gRPC status codes from Envoy expose this as an increase in 9 FAILED PRECONDITION

operations/tempo-mixin/dashboards.libsonnet
```diff
@@ -102,7 +102,59 @@ dashboard_utils {
...
+        )
+        .addPanel(
+          $.textPanel(
+            'gRPC status codes',
+            |||
+              Visit [Status codes and their use in gRPC](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md)
+
+              Code | Number | Description
+              ---|---|---
+              OK | 0 | Not an error; returned on success.
```
annanay25, line 136: Nice, very useful addition.

annanay25 (approved):
LGTM.

Issue #1102, opened 2021-11-05
Update Tempo's open-telemetry/collector dependency


kvrhdn:
**Is your feature request related to a problem? Please describe.**

Tempo depends on [go.opentelemetry.io/collector](https://pkg.go.dev/go.opentelemetry.io/collector) for its receivers. We are currently still using v0.21 while the latest version is already v0.38...

https://github.com/grafana/tempo/blob/c5d007da27071358fa9b12730c048f0eddbeb82c/go.mod#L57

The receivers are used by the distributor: https://github.com/grafana/tempo/search?q=go.opentelemetry.io%2Fcollector

**Describe the solution you'd like**

Update the dependency.
Update the documentation if this changes any defaults (I'm thinking about https://github.com/grafana/tempo/issues/637).

**Describe alternatives you've considered**

If we don't update we are going to miss out on security and performance improvements.

**Additional context**
N/A

kvrhdn:
We should probably also look into updating our opentelemetry-proto submodule:
https://github.com/grafana/tempo/blob/c5d007da27071358fa9b12730c048f0eddbeb82c/.gitmodules#L1-L3

I don't expect any changes in the traces format, but it's good to stay fairly up-to-date.

Diff: https://github.com/open-telemetry/opentelemetry-proto/compare/a17f202fdae65e828ac29fb663b5bc5b64b13290...main

trexx:
I take it that upgrading the collector dependency will also allow enabling TLS on the otlp receiver?

annanay25:
The collector was updated in the following PRs
- https://github.com/grafana/tempo/pull/1142
- https://github.com/grafana/tempo/pull/1214
Should we rename this issue (to only update the proto) and move it into the next milestone?

kvrhdn:
I'm not sure if we want to bother updating the proto. OTLP traces is declared stable so not much has changed.
These are changes between our version and main: https://github.com/open-telemetry/opentelemetry-proto/compare/a17f202fdae65e828ac29fb663b5bc5b64b13290...main

Maybe we can look into this when we work on new backend formats?

PR #1142, opened 2021-11-28
Update opentelemetry collector to v0.40.0

tete17:
**What this PR does**:
This PR updates the opentelemetry-collector dependency to 0.40.0

We can consider it a sibling of this PR  https://github.com/grafana/helm-charts/pull/855. The reason is the old version still had the wrong default port number for otel http endpoint. By upgrading the dependency we make sure to always use the port 4318 and avoid future confusion.

The changes are fairly minor, most of my work was going through the changelog and see the renaming of symbols they did. The biggest hurdles were the change they did in config parsing and the fact that the metrics for collectors are now hidden due under an internal package, see https://github.com/open-telemetry/opentelemetry-collector/pull/3253.

Both changes are mostly reflected in the file `modules/distributor/receiver/shim.go`. That can be a good point of focus for anyone willing to review this change. Ignoring the vendor folder it shouldn't be that big.

**Checklist**
- [x] Tests updated
- [x] Documentation added


zalegrala:
Relates to https://github.com/grafana/tempo/issues/1102

modules/distributor/receiver/shim.go
```diff
@@ -248,22 +250,16 @@ func newLogger(level logging.Level) *zap.Logger {
 	return logger
 }
 
-func newMetricViews() ([]*view.View, error) {
-	views := obsreport.Configure(configtelemetry.LevelNormal)
-	err := view.Register(views...)
```
mapno, line 253: We need to register views in order to get metrics from the OTel components. Otherwise, the registered exporter below won't have views to export.

Unfortunately, the `obsmetrics` pkg has been made internal, so we can't import it here. We should find a replacement or solution to it.

mapno (commented):
Hi, thanks for the contribution! Overall the upgrade looks good.

Left a comment about the registered metrics.

modules/distributor/receiver/shim.go
tete17: Hi @mapno, thanks for the feedback

I am not sure if keeping the old metrics from the opentelemetry collector is still an option, they mention they want to expose opentelemetry metrics instead of opencesus but I believe that is not an option yet. 

Do you suggest to scrap the registered exporter or implementing the metrics ourrselves?

joe-elliott: @mapno should we use Grafana's otel collector fork for this?

mapno: Right, I don't think there is a direct solution at the moment that doesn't involve a fork, implementing the metrics ourselves or a similar approach.

At Grafana we currently host a [fork](https://github.com/grafana/opentelemetry-collector/) of the collector in which `/internal` folders are exported, so we can circumvent this issue.

You can see the Agent importing this fork: [go.mod](https://github.com/grafana/agent/blob/main/go.mod#L140-L143). To get PR out, I'm fine with using this updating and using the fork for v0.40.0, while a solution is found. Does that sound good to you @tete17
?

mapno: This [branch](https://github.com/grafana/opentelemetry-collector/tree/0.40-grafana) is updated to v0.40.0 with internal folders exported

tete17: Sounds like an awesome plan.

tete17: This feels like a much deeper conversation that what I intended to go with a quick and dirty dependency update :laughing:

mapno (approved):
All looks good, many thanks for the contribution!

mapno:
Hey @tete17, it seems that the CI failed at two checks `vendor-check` and `lint`.

1. For `vendor-check`: Can you run `go mod tidy`? There is a tiny diff in the modules.
2. For `Lint`: `consumererror.Combine(errs)` appears to be deprecated. You can replace it with `go.uber.org/multierr.Combine(errs...)`

Thanks again!

tete17:
Hey @mapno sorry for the silly mistakes the vendor issue must have slipped me with the rebase and the deprecated function I saw in the first ci run and did nothing my bad.

Should be fixed now :smile:

tete17:
I tried running the e2e tests locally and found no error. Are they run against what would be the result of the merge or the branch of the MR?

Could it be just a fluke? can we run them again?

joe-elliott (approved):
Awesome improvement! I was privately dreading this upgrade. Thanks!

PR #1144, opened 2021-12-01
Compaction load metric

mapno:
**What this PR does**:

Adds new metric to measure compaction load. It counts the number of blocks that are yet to be compacted before the next maintenance cycle. It's higher, the greater the compaction load is.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/compactor.go
+++ b/tempodb/compactor.go
@@ -51,8 +51,6 @@ var (
 		Name:      "compaction_outstanding_blocks",
 		Help:      "Number of blocks remaining to be compacted before next maintenance cycle",
 	}, []string{"tenant"})
-
-	outstandingCompactionBlocks = make(map[string]int)
 )
 
 const (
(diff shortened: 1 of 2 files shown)


mdisibio (approved):
Nice addition, should be very helpful for compactor scaling and troubleshooting.

tempodb/compactor.go
```diff
@@ -113,6 +118,17 @@ func (rw *readerWriter) doCompaction() {
...
 		if start.Add(rw.cfg.BlocklistPoll).Before(time.Now()) {
+			// count number of outstanding blocks remaining before next maintenance cycle
+			for {
+				leftToBeCompacted, _ := blockSelector.BlocksToCompact()
+				if len(leftToBeCompacted) == 0 {
+					break
+				}
+				for _, block := range leftToBeCompacted {
+					metricCompactionOutstandingBlocks.WithLabelValues(block.TenantID).Inc()
+				}
```
joe-elliott, line 129: We could:

Use counter.Add(len(leftToBeCompacted)) to remove this loop.

But:

If we're going to make this per tenant I'd prefer using a gauge as it would show is exactly how many blocks remaining per tenant.

mapno: Changed to a gauge 👍

tempodb/compactor.go
```diff
@@ -308,6 +317,19 @@ func markCompacted(rw *readerWriter, tenantID string, oldBlocks []*backend.Block
 	rw.blocklist.Update(tenantID, newBlocks, oldBlocks, newCompactions)
 }
 
+func measureOutstandingBlocks(tenantID string, blockSelector CompactionBlockSelector) {
```
joe-elliott, line 320: +1 moving this into its own method. cleans up doCompaction() nicely

joe-elliott (approved):

PR #1145, opened 2021-12-01
Improve compaction efficiency

joe-elliott:
**What this PR does**:
In multitenant environments tenants with long blocklists can be starved of work from the compactors. This is in part due to the compactionCycle wait time. Reducing this hardcoded value to improve the rate at which a compactor will move through all tenants.

- Reduce `compactionCycle` to delay just enough to prevent a tight loop that eats CPU for no reason.
- Increase `waitOnStartup` to account for the loss of 30s in the compactionCycle. This might not be necessary but making the change to reduce the impact of this PR.
- Added a new config parameter `max_compaction_cycle` that allows the polling cycle and the compaction cycle to be independently configured.
- Regenned manifest

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


KMiller-Grafana (approved):
Docs portion of this PR looks good to me.

modules/storage/config.go
```diff
@@ -28,6 +28,7 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
 	cfg.Trace.BlocklistPollFallback = true
 	cfg.Trace.BlocklistPollConcurrency = tempodb.DefaultBlocklistPollConcurrency
 	cfg.Trace.BlocklistPollTenantIndexBuilders = tempodb.DefaultTenantIndexBuilders
+	cfg.Trace.MaxCompactionCycle = tempodb.DefaultBlocklistPoll
```
mapno, line 31: Should this have it's own default, or we want it tied to blocklist poll cycle?

mapno (approved):
lgtm

modules/storage/config.go
joe-elliott: yeah, this is a good idea to prevent confusion.

PR #1150, opened 2021-12-02
Pin version of tanka used in CI, update kube-manifests

kvrhdn:
**What this PR does**:
CI has been failing since Tanka has released v0.19: https://github.com/grafana/tanka/blob/main/CHANGELOG.md#0190-2021-11-22
This pins the version to 0.19 and updates the kube-manifests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,7 +109,7 @@ jobs:
 
       - name: Install jsonnet, jsonnet-bundler & tanka
         run: |
-          brew install jsonnet tanka@0.19.0
+          brew install jsonnet tanka@0.19
           go install github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb@v0.4.0
 
       - name: Check out code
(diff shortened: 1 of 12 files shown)


annanay25:
Hmm, this link says `0.19.0` should work https://formulae.brew.sh/api/bottle/tanka.json

Is the architecture used in CI mentioned on the brew page?

kvrhdn:
Apparently you can only use the `@<version>` syntax if the formula explicitly supports this.

For example, postgresql does:
```
$ brew search postgresql
==> Formulae
postgresql           postgresql@10        postgresql@11        postgresql@12        postgresql@13        postgresql@9.4       postgresql@9.5       postgresql@9.6       qt-postgresql        postgrest

==> Casks
navicat-for-postgresql
```

Tanka does not:
```
$ brew search tanka
==> Formulae
tanka ✔

==> Casks
font-anka-coder
```

I'm opting to download the right formula directly from Github instead, this is also described here: https://github.com/Homebrew/brew/issues/8791#issuecomment-696652589

annanay25 (approved):
sad that we need to do this, but it is what it is, lgtm.

PR #1173, opened 2021-12-13
Remove Push from tempopb.Pusher

kvrhdn:
**What this PR does**:
Remove the `Push` method from `tempopb.Pusher`.

I noticed `Pusher` is implemented by both the distributor and the ingester, but:
- the distributor does not support `PushBytes`
- `Push` is deprecated for the ingester. This change was introduced in January (#430). It was a breaking change back then so I don't think this is a breaking change again now.

I also noticed none of the `Pusher` methods in distributor were used. The OTel Receiver shim defines its own interface `BatchPusher`. So I deleted this implementation.


joe-elliott (approved):
Nice cleanup

PR #1178, opened 2021-12-15
Switch compactor to new BasicLifecycler with AutoForgetDelegate

### LINKED ISSUE #1081: Unhealthy compactors do not leave the ring after rollout
**Describe the bug**
When rolling out a new deployment of the compactors, some old instances will remain in the ring as Unhealthy. The only fix seems to be to port-forward one of the compactors and use the `/compactor/ring` page to "Forget" all the unhealthy instances.

**To Reproduce**
1. Start Tempo in kubernetes (we have tried the 1.1 release but the issue persists with `af34e132a1b8`)
2. Perform a rollout of the compactors

**Expected behavior**
The compactors from the previous deployment leave the ring correctly

**Environment:**
 - Infrastructure: kubernetes
 - Deployment tool: `kubectl apply`

**Additional Context**
We do not see this happen all the time. On one of our similarly sized but less busy clusters, old compactors rarely stay in the ring after a rollout. On the busier cluster, we had 14 unhealthy compactors from a previous deployment still in the ring, out of 30 in the deployment.

Our tempo config for memberlist:
```
memberlist:
      abort_if_cluster_join_fails: false
      join_members:
        - tempo-gossip-ring
      dead_node_reclaim_time: 15s
      bind_addr: ["${POD_IP}"]
```

Sample logs from a compactor that stayed in the ring as unhealthy, from the moment where shutdown was requested:
```
level=info ts=2021-10-26T15:26:54.628754652Z caller=signals.go:55 msg=""=== received SIGINT/SIGTERM ===\n*** exiting""level=info ts=2021-10-26T15:26:54.629120545Z caller=lifecycler.go:457 msg=""lifecycler loop() exited gracefully"" ring=compactor
level=info ts=2021-10-26T15:26:54.629162701Z caller=lifecycler.go:768 msg=""changing instance state from"" old_state=ACTIVE new_state=LEAVING ring=compactor
level=info ts=2021-10-26T15:26:55.632049563Z caller=lifecycler.go:509 msg=""instance removed from the KV store"" ring=compactor
level=info ts=2021-10-26T15:26:55.63212068Z caller=module_service.go:96 msg=""module stopped"" module=compactor
level=info ts=2021-10-26T15:26:55.632206675Z caller=module_service.go:96 msg=""module stopped"" module=overrides
level=info ts=2021-10-26T15:26:55.632206876Z caller=memberlist_client.go:572 msg=""leaving memberlist cluster""
level=info ts=2021-10-26T15:26:55.632249778Z caller=module_service.go:96 msg=""module stopped"" module=store
level=warn ts=2021-10-26T15:27:05.735678769Z caller=memberlist_client.go:587 msg=""broadcast messages left in queue"" count=16 nodes=146
level=info ts=2021-10-26T15:27:07.192768676Z caller=module_service.go:96 msg=""module stopped"" module=memberlist-kv
level=info ts=2021-10-26T15:27:07.194383366Z caller=server_service.go:50 msg=""server stopped""
level=info ts=2021-10-26T15:27:07.194466883Z caller=module_service.go:96 msg=""module stopped"" module=server
level=info ts=2021-10-26T15:27:07.19452497Z caller=app.go:271 msg=""Tempo stopped""
level=info ts=2021-10-26T15:27:07.194539895Z caller=main.go:135 msg=""Tempo running""
```
I was confused by that last `Tempo running` line but looking at the code in main.go, this seems normal.

mdisibio:
**What this PR does**:
This PR switches the compactor ring from `Lifecycler` to the new pluggable `BasicLifeCycler`, and adds `AutoForgetDelegate` to automatically forget unhealthy entries. 

Notes:
* Autoforget defaults to 2xHeartbeatTimeout=2M, which is a middle ground between other components with auto-forget [loki ingester (1x)](https://github.com/grafana/loki/blob/main/pkg/ingester/ingester.go#L305) and [cortex storegateway (10x)](https://github.com/cortexproject/cortex/blob/master/pkg/storegateway/gateway.go#L41).   Would like feedback on if this should be exposed as a config option, and preferences on different default behavior, if any.

* There is slightly more leg-work required to use `BasicLifecycler` such creating the KV client, and the new `OnRingInstance*` callbacks.  Considered locating parts of this in `tempo/pkg/ring` for reuse, but the numerous small differences in how each component interacts with the ring (ingester vs compactor, for example), make it seem not very beneficial.  Can revisit later if/when more components are switched to `BasicLifecycler`.

* There are a few other changes to make the compactor startup leverage existing ring functionality, for waiting for ring activation and topology stabilization.

Fixes #1081 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/compactor/compactor.go
```diff
@@ -19,7 +20,17 @@ import (
...
+	// in the ring will be automatically removed.
+	ringAutoForgetUnhealthyPeriods = 2
+
+	// We use a safe default instead of exposing to config option to the user
+	// in order to simplify the config.
+	ringNumTokens = 512
+)
+
+var (
+	ringOp = ring.NewOp([]ring.InstanceState{ring.ACTIVE}, nil)
```
joe-elliott, line 33: nice +1 to this ringOp over the old one.

modules/compactor/compactor.go
```diff
@@ -19,7 +20,17 @@ import (
...
 
 const (
-	waitOnStartup = 90 * time.Second
+	// ringAutoForgetUnhealthyPeriods is how many consecutive timeout periods an unhealthy instance
+	// in the ring will be automatically removed.
+	ringAutoForgetUnhealthyPeriods = 2
+
+	// We use a safe default instead of exposing to config option to the user
+	// in order to simplify the config.
+	ringNumTokens = 512
```
joe-elliott, line 29: I would personally go with a smaller value like 128 or 256. Since most Tempo users are using gossip a smaller value would reduce the chatter between components. I think that getting a super even distribution is less important in this component than others.

No data to back that up. Just a gut feeling.

modules/compactor/compactor.go
```diff
@@ -45,50 +56,94 @@ func New(cfg Config, store storage.Store, overrides *overrides.Overrides, reg pr
...
 
-	subservices := []services.Service(nil)
 	if c.isSharded() {
-		lifecyclerCfg := c.cfg.ShardingRing.ToLifecyclerConfig()
-		lifecycler, err := ring.NewLifecycler(lifecyclerCfg, ring.NewNoopFlushTransferer(), "compactor", cfg.OverrideRingKey, false, log.Logger, prometheus.WrapRegistererWithPrefix("cortex_", reg))
+		reg = prometheus.WrapRegistererWithPrefix("cortex_", reg)
+
+		lifecyclerStore, err := kv.NewClient(
+			cfg.ShardingRing.KVStore,
+			ring.GetCodec(),
```
joe-elliott, line 64: This will match up with the codec we are currently using? i.e. on rollout all components will still gossip successfully?

modules/compactor/compactor.go
```diff
@@ -45,50 +56,94 @@ func New(cfg Config, store storage.Store, overrides *overrides.Overrides, reg pr
...
 }
 
-func (c *Compactor) starting(ctx context.Context) error {
-	if c.subservices != nil {
+func (c *Compactor) starting(ctx context.Context) (err error) {
+	// In case this function will return error we want to unregister the instance
+	// from the ring. We do it ensuring dependencies are gracefully stopped if they
+	// were already started.
+	defer func() {
+		if err == nil || c.subservices == nil {
```
joe-elliott, line 102: i had never though of checking err in a defer. this is cool

modules/compactor/compactor.go
```diff
@@ -166,27 +219,41 @@ func (c *Compactor) BlockRetentionForTenant(tenantID string) time.Duration {
...
+	// When we initialize the compactor instance in the ring we want to start from
+	// a clean situation, so whatever is the state we set it ACTIVE, while we keep existing
+	// tokens (if any) or the ones loaded from file.
+	var tokens []uint32
+	if instanceExists {
+		tokens = instanceDesc.GetTokens()
 	}
+
+	takenTokens := ringDesc.GetTokens()
+	newTokens := ring.GenerateTokens(512-len(tokens), takenTokens)
```
joe-elliott, line 239: should 512 be `ringNumTokens`?

joe-elliott (commented):
Agree some of this boilerplate is awkward. Will follow your lead on making any of this shared :+1: 

A few comments/questions.

joe-elliott:
Oh, one more question, if a compactor is auto-forgot and then reappears what happens? Does it cleanly reinsert itself into the ring?

modules/compactor/compactor.go
mdisibio: Yes this is a copy of the behavior that was in [NewLifecycler()](https://github.com/grafana/tempo/blob/v1.2.1/vendor/github.com/cortexproject/cortex/pkg/ring/lifecycler.go#L155) that is no longer in `NewBasicLifecycler`.

modules/compactor/compactor.go
mdisibio: Shamelessly taken from cortex :)

modules/compactor/compactor.go
mdisibio: Shamelessly taken from cortex :)

modules/compactor/compactor.go
mdisibio: This matches the previous hardcoded value that the compactor was using via [lifecycler config conversion](https://github.com/grafana/tempo/blob/v1.2.1/vendor/github.com/cortexproject/cortex/pkg/compactor/compactor_ring.go#L104).   Not opposed to defaulting to a smaller value, but then wondering if it needs to be configurable.

mdisibio:
> Oh, one more question, if a compactor is auto-forgot and then reappears what happens? Does it cleanly reinsert itself into the ring?

It depends on what you mean by cleanly.    If the compactor was restarted since being auto-forget (i.e. a new pod), since there is no token persistence, it will rejoin with new tokens (in this case there is no difference between a restarted compactor or a new compactor).      If the compactor remained running (but temporarily lost ability to heartbeat to the ring), I think it should rejoin with the same tokens.

joe-elliott (approved):

PR #1190, opened 2021-12-27
instrumentation: fix trace context propagation distributor -> ingester

kvrhdn:
**What this PR does**:
Ensure the trace context is propagated to all ingesters and add a couple of spans.

Result:

Screenshot 2021-12-27 at 14 15 19


mapno (approved):
LGTM

PR #1191, opened 2021-12-27
Log the tenant ID in the live traces exceeded error

kvrhdn:
**What this PR does**:
Log the tenant ID that is being rate limited by live traces.


mapno (approved):
LGTM

Issue #1133, opened 2021-11-23
Increase protection from large traces


joe-elliott:
**Is your feature request related to a problem? Please describe.**
It is relatively easy to push traces larger than the configured limit into Tempo. These very large traces can break compactors, the query path, and sometimes ingesters.

**Describe the solution you'd like**
It is impossible to completely protect ourselves from traces exceeding the configured limit, but there is room for improvement. Currently as soon as a trace is removed from the active traces map we no longer track trace size and it is possible to start pushing to the trace again. Proposed:
- Maintain a `map[uint32]struct{}` of trace hashes that had refused spans due to trace too large
- Check the map on write and reject any hashes found with TRACE_TOO_LARGE
- Clear the map when the head block is cut

**Describe alternatives you've considered**

**Additional context**
@annanay25 I think you were considering similar functionality. If you have any thoughts please include them here.

annanay25:
We were tracking some options here: https://github.com/grafana/tempo/issues/976 

But that was focused on protecting compactors, good call on "moving it up" to the ingesters - If we can reject the trace at ingestion it will protect all downstream components

mdisibio:
More options:
* Compactors delete traces larger than configurable threshold.  These mega/gigatraces aren't queryable anyway... acceptable data loss?
* Compactors skip over/don't combine traces larger than configurable threshold. Some small code structure changes needed as several things require a single output `[]byte` (`ObjectCombiner` interface, deduping iterator, etc).  Traces still aren't queryable, so the problem is not fully solved.
* Performance improvements to `model.CombineObjects`. I.e. the overall design remains, just improve stability/performance around large traces.  A couple changes:  only scan span IDs once per input, only sort final result once.

joe-elliott:
This was fixed in #1199.

Issue #976, opened 2021-09-22
Compactors require high memory as traces combine and grow in the backend


annanay25:
**Describe the bug**

Compactors require high memory as traces combine and grow in the backend. This opens the possibility of crafting a long running trace with low spans-per-second that can grow and eventually OOM compactors.

**To Reproduce**
1. Start Tempo (SHA or version): all versions till e5f7ded
2. Perform Operations (Read/Write/Others): Carefully craft super-long running traces with a few spans every second that will eventually be combined by the compactors into a MEGA trace (the largest we are seeing so far are 1.3GB)

**Expected behavior**

Compactors do not keep OOMing.

**Environment:**
 - Infrastructure: [e.g., Kubernetes, bare-metal, laptop]
 - Deployment tool: [e.g., helm, jsonnet]

**Additional Context**

Some possibilities considered:
- write multiple splits of a trace into the same block (might be harder than it sounds)
- Do not compact blocks that have very large traces

mdisibio:
> a MEGA trace (the largest we are seeing so far are 1.3GB)

For both possibilities listed, a trace of this size would still be trouble for the queriers.  I.e. if the compactor is able to write multiple splits of the trace (or ignore the block entirely), the expected behavior is still the querier recombines all segments.    Maybe a solution for the querier is limit the amount of data returned in a single call, add a new paged API to retrieve all the splits.  Quick ideas on the limits per call, 100MiB?  Even a 100MiB trace is quite large and hard to utilize.

Also, I propose to call any trace over 1GB a `GIGA trace` :)

PR #1199, opened 2022-01-04
Ingester prevent writes to large traces even after flushing

mdisibio:
**What this PR does**:
This PR updates the ingester to keep track of traces that exceeded max size and reject further writes to them, even after they are flushed to disk. The list of traces is reset when cutting the head block. 

Fixes part of #1133 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,12 +27,12 @@
 * [ENHANCEMENT] Reduce search data file sizes by optimizing contents [#1165](https://github.com/grafana/tempo/pull/1165) (@mdisibio)
 * [ENHANCEMENT] Add `tempo_ingester_live_traces` metric [#1170](https://github.com/grafana/tempo/pull/1170) (@mdisibio)
 * [ENHANCEMENT] Update compactor ring to automatically forget unhealthy entries [#1178](https://github.com/grafana/tempo/pull/1178) (@mdisibio)
+* [ENHANCEMENT] Prevent writes to large traces even after flushing to disk [#1199](https://github.com/grafana/tempo/pull/1199) (@mdisibio)
 * [BUGFIX] Add process name to vulture traces to work around display issues [#1127](https://github.com/grafana/tempo/pull/1127) (@mdisibio)
 * [BUGFIX] Fixed issue where compaction sometimes dropped spans. [#1130](https://github.com/grafana/tempo/pull/1130) (@joe-elliott)
 * [BUGFIX] Ensure that the admin client jsonnet has correct S3 bucket property. (@hedss)
 * [BUGFIX] Publish tenant index age correctly for tenant index writers. [#1146](https://github.com/grafana/tempo/pull/1146) (@joe-elliott)
 * [BUGFIX] Ingester startup panic `slice bounds out of range` [#1195](https://github.com/grafana/tempo/issues/1195) (@mdisibio)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -32,6 +32,7 @@
 * [BUGFIX] Ensure that the admin client jsonnet has correct S3 bucket property. (@hedss)
 * [BUGFIX] Publish tenant index age correctly for tenant index writers. [#1146](https://github.com/grafana/tempo/pull/1146) (@joe-elliott)
 * [BUGFIX] Ingester startup panic `slice bounds out of range` [#1195](https://github.com/grafana/tempo/issues/1195) (@mdisibio)
+* [BUGFIX] Prevent writes to large traces even after flushing to disk [#1199](https://github.com/grafana/tempo/pull/1199) (@mdisibio)
```
joe-elliott, line 35: enhancement? i think this is just an improvement over what already existed and we'll never get it perfect.

modules/ingester/instance.go
```diff
@@ -140,9 +161,6 @@ func (i *instance) Push(ctx context.Context, req *tempopb.PushRequest) error {
 		return status.Errorf(codes.FailedPrecondition, "%s max live traces exceeded for tenant %s: %v", overrides.ErrorPrefixLiveTracesExceeded, i.instanceID, err)
 	}
 
-	i.tracesMtx.Lock()
-	defer i.tracesMtx.Unlock()
-
 	id, err := pushRequestTraceID(req)
```
joe-elliott, line 164: i wonder if we should just delete `Push()`? it hasn't been used for 3 or 4 releases.

modules/ingester/trace.go
```diff
@@ -54,7 +50,8 @@ func (t *trace) Push(_ context.Context, instanceID string, trace []byte, searchD
 	if t.maxBytes != 0 {
 		reqSize := len(trace)
 		if t.currentBytes+reqSize > t.maxBytes {
-			return status.Errorf(codes.FailedPrecondition, "%s max size of trace (%d) exceeded while adding %d bytes to trace %s", overrides.ErrorPrefixTraceTooLarge, t.maxBytes, reqSize, hex.EncodeToString(t.traceID))
+			//return status.Errorf(codes.FailedPrecondition, "%s max size of trace (%d) exceeded while adding %d bytes to trace %s", overrides.ErrorPrefixTraceTooLarge, t.maxBytes, reqSize, hex.EncodeToString(t.traceID))
```
joe-elliott, line 53: remove comment?

joe-elliott (approved):
minor thoughts. lgtm

CHANGELOG.md
annanay25, line 35: I think another solution that might cover more cases is returning an error from `CombineTraces()` as soon as the pieces sum up to > maxbytes. That will help fix compaction as well as querying.

modules/ingester/instance.go
mdisibio: Agree but didn't seem in-scope for this PR.

CHANGELOG.md
mdisibio: Agree, we can explore that in a separate PR.  Let's continue collecting thoughts in #1133 and #976.

PR #1205, opened 2022-01-05
Add design-proposals directory

kvrhdn:
**What this PR does**:
Add a new directory `design-proposals` in which we can capture future design proposals. I've also added a README suggesting an initial process. This is just an initial suggestion that we can improve incrementally as we gain experience with this process.


kvrhdn:
I've added a first design proposal using this structure: https://github.com/grafana/tempo/pull/1206
I can obviously adapt that PR if we want to change the suggested structure.

joe-elliott (approved):

PR #1211, opened 2022-01-06
Refactor: Data Encoding

joe-elliott:
**What this PR does**:
Adds some structure to the model package and includes a new abstraction to encapsulate new encodings. This PR lays the groundwork for a new encoding that will improve search times.


pkg/model/v1/decoder.go
```diff
@@ -0,0 +1,80 @@
...
+func (d *Encoder) Marshal(t *tempopb.Trace) ([]byte, error) {
+	traceBytes := &tempopb.TraceBytes{}
+	bytes, err := proto.Marshal(t)
+	if err != nil {
+		return nil, err
+	}
+
+	traceBytes.Traces = append(traceBytes.Traces, bytes)
+
+	return proto.Marshal(traceBytes)
```
mdisibio, line 79: Does this need to marshal to a buffer from `tempopb.SliceFromBytePool` ?  I don't think it does because the output buffer is not returned to the pool by the ingester, but double checking.

mdisibio (changes requested):
Overall this is good cleanup, definitely appreciate the rename to `liveTrace` and the distinct encoding packages.  Left a few comments/questions.

pkg/model/v0/decoder.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"fmt"
+
+	"github.com/gogo/protobuf/proto"
+	"github.com/grafana/tempo/pkg/model/trace"
+	"github.com/grafana/tempo/pkg/tempopb"
+)
+
+const Encoding = ""
```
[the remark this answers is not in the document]
joe-elliott: I split the interface to hide Marshal from the outside world and now the interface is called "Decoder". Agree that calling everything "Encoding" was getting weird. This may help with this const name. LMK

pkg/model/v1/decoder.go
joe-elliott: tempopb.Trace doesn't use the `PreallocBytes` b/c it's just a thin wrapper around otel proto. `PushBytesRequest` is our only proto message tthat uses `PreallocBytes`.

mdisibio (approved):

PR #1224, opened 2022-01-13
vulture: reduce cardinality of random generated span names

kvrhdn:
**What this PR does**:
If you send traces from vulture to a tenant containing other traces, the random span names are hard to distinguish from normal span names. This is annoying if you want to search based upon the span names.
This change adds a `vulture-` prefix to random generated strings (similar to what we do for the tags) so all span names end up next to each other.

Screenshot 2022-01-13 at 13 07 13


pkg/util/trace_info.go
```diff
@@ -161,7 +162,7 @@ func (t *TraceInfo) generateRandomString() string {
 	for i := range s {
 		s[i] = letters[t.r.Intn(len(letters))]
 	}
-	return string(s)
+	return "vulture-" + string(s)
```
annanay25, line 165: While we're here should we take a minute to think about whether we need completely random strings or if we should cap them to some cardinality?

We could set it to a 1000. And the span names could be `vulture-0`, `vulture-1`, ..., `vulture-999` (again this is similar to what we do with tags)

kvrhdn: Yeah, that's fair. Span names aren't expected to be super high cardinality. Depends a bit on whether we like testing search with this abnormal high cardinality or not.
`generateRandomString` is also used to generate tag values, so we might have to split them up.

kvrhdn:
Updated the PR so span names will always follow the format `vulture-#`, in which # is a number between 0 and 100. So this reduces the cardinality of span names to 100.

annanay25 (approved):

Issue #1175, opened 2021-12-13
Backend Search: Serverless Implementations


joe-elliott:
- [ ] Code improvements/bug fixes
  - [x] Support pools of external search endpoints
  - [x] Build/Test serverless folder as part of CI
  - [x] tempo-cli command to do backend search
  - [x] Have queriers prefer themselves to the external search endpoints
  - [x] Fix min/max time range on blocks
  - [x] Support searching on fields besides strings
  - [x] Add ingestion time range limits to spans
  - [ ] Improve querier/query-frontend relationship. Can we have per endpoint `max_parrallelism`?
- [ ] Build tooling to create code packages for all major cloud providers
  - [x] GCP
  - [x] AWS
  - [ ] Azure
- [ ] Create guides for deploying to the major cloud providers
  - [x] GCP
  - [x] AWS
  - [ ] Azure

kvrhdn:
> Determine why coldstart is faster than reusing existing function instances

I wonder if this might be a garbage collection issue? After the execution the function is frozen and garbage collection can not run. So if we are allocating a lot, it will probably have to run during the next execution.

joe-elliott:
Removed

> Determine why coldstart is faster than reusing existing function instances

b/c I no longer think this is an issue with Cloud Functions. I believe I was exhausting a resource in the querier that was causing timeouts and retries.

joe-elliott:
This was completed except for Azure. Splitting off a new issue and closing this.  

See #1898

PR #1227, opened 2022-01-13
Expansion of Encoding Interfaces and Addition of V2

joe-elliott:
**What this PR does**:
Expands the encoding interfaces added in #1211 and adds a `v2` interface which is capable of efficiently retrieving its range to improve search speeds. This PR also puts everything into place to fix the start/end time range of blocks referenced in #1175 .

Marked as draft b/c this shouldn't be merged in 1.3. ~Also it needs some cleanup.~

UPDATE:
This PR is ready for review but I'm going to leave it draft until 1.3 is cut. 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/tempopb/tempo.proto
```diff
@@ -6,7 +6,9 @@ import "trace/v1/trace.proto";
 import "github.com/gogo/protobuf/gogoproto/gogo.proto";
 
 service Pusher {
-  rpc PushBytes(PushBytesRequest) returns (PushResponse) {};
+  // different versions of PushBytes expect the trace data to be pushed in different formats
+  rpc PushBytes(PushBytesRequest) returns (PushResponse) {};   // ./pkg/model/v1
+  rpc PushBytesV2(PushBytesRequest) returns (PushResponse) {}; // ./pkg/model/v2
```
zalegrala, line 11: Are you adding both versions here to allow handling of both versions at the same time?

modules/distributor/distributor.go
```diff
@@ -87,6 +89,14 @@ var (
 	}, []string{discardReasonLabel, "tenant"})
 )
 
+// rebatchedTrace is used to more cleanly pass the set of data
+type rebatchedTrace struct {
+	id    []byte
```
zalegrala, line 94: Is `id` used?  I see `trace.id` used, but maybe missed where `id` is used.

pkg/tempopb/tempo.proto
joe-elliott: yup. while the distributors transition to the new version the ingesters will need to be able to take both v1 and v2.

modules/distributor/distributor.go
joe-elliott: Yup, it's used on line 333 where the distributor creates the batches it sends to the ingesters.

pkg/model/v2/segment_decoder.go
```diff
@@ -0,0 +1,115 @@
...
+	const uint32Size = 4
+
+	sz := proto.Size(pb)
+	buff := make([]byte, 0, sz+uint32Size*2) // proto buff size + start/end uint32s
+
+	buffer := proto.NewBuffer(buff)
+
+	_ = buffer.EncodeFixed32(uint64(start)) // EncodeFixed32 can't return an error
+	_ = buffer.EncodeFixed32(uint64(end))
+	err := buffer.Marshal(pb)
```
zalegrala, line 89: Is this marshal appending to the buffer?  So the two uint64s are put in place, and then the variable-length trace bytes?

mdisibio (commented):
Overall this is looking good and should make it easier to support more data encodings in the future.  Do you have any numbers on how much the new per-trace start/end improve search times?

I'm not sure I am understanding the difference between ObjectDecoder and SegmentDecoder.  In V2 they have the same underlying format (uint32/uint32/[]byte), and similarities in PrepareForRead and Combine/ToObject.  Could/should these be combined, or maybe some more comments/clarification why they are not?   This might be a terminology/definition difference (see the other comment for more specifics)

pkg/model/v2/segment_decoder.go
joe-elliott: technically two uint32s, but yes. for some reason `EncodeFixed32` takes a uint64.

pkg/model/segment_decoder.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"fmt"
+
+	v1 "github.com/grafana/tempo/pkg/model/v1"
+	v2 "github.com/grafana/tempo/pkg/model/v2"
+	"github.com/grafana/tempo/pkg/tempopb"
+)
+
+// SegmentDecoder is used by the distributor/ingester to aggregate and pass batches of traces
```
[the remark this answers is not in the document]
joe-elliott: Yup, I tried to move everything to segment but obviously missed a few. Will review and correct.

modules/ingester/ingester.go
```diff
@@ -164,8 +167,41 @@ func (i *Ingester) markUnavailable() {
...
+	}
+
+	for i, t := range req.Traces {
+		trace, err := v1Decoder.PrepareForRead([][]byte{t.Slice})
+		if err != nil {
+			return nil, fmt.Errorf("error calling v1.PrepareForRead %w", err)
+		}
+
+		now := uint32(time.Now().Unix())
+		v2Slice, err := v2Decoder.PrepareForWrite(trace, now, now)
```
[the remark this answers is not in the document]
joe-elliott: Right. Since this is only called during migration I didn't want to add any more complexity here then necessary.

Unfortunately, it doesn't quite mimic past behavior. If you wrote a trace from a minute after it was actually generated then it's possible that now/now would exclude it incorrectly from a search. I'm unsure how much we want to invest in this though given how the tiny percentage of traces this will impact.

pkg/model/v2/segment_decoder.go
zalegrala: Right sorry, `uint32`.

mdisibio (approved):

PR #1228, opened 2022-01-17
Replace use of cortex util directory with tempo

annanay25:

**What this PR does**:
- Replaces the use of cortex util directory by copying relevant files over and using `dskit` wherever necessary
- Shuffle some files around in tempo util directory to better match use
- ⚠️  Drop `PrometheusLogger` -> This no longer tries to register the metric `log_messages_total`. The cortex util package will still register it but it will always show a value of zero. :sadface:

Fixes N/A

**Checklist**
- N/A Tests updated
- N/A Documentation added
- N/A `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`


pkg/util/http.go
```diff
@@ -0,0 +1,10 @@
+package util
+
+import (
+	"strings"
+)
+
+// IsRequestBodyTooLarge returns true if the error is "http: request body too large".
+func IsRequestBodyTooLarge(err error) bool {
```
joe-elliott, line 8: does this belong in /pkg/api?

honestly it only seems to bei n use in `/modules/frontend`. should we just put it there?

pkg/util/validation.go
```diff
@@ -0,0 +1,8 @@
+package util
+
+// Supported values for ingestion rate strategies
+// TODO: check if these can be moved to dskit
+const (
+	LocalIngestionRateStrategy  = "local"
```
joe-elliott, line 6: these are already defined in `/modules/overrides`. should we use those?

joe-elliott (commented):
Couple o' questions

pkg/util/http.go
annanay25: makes sense. moved to `frontend/handler.go`

pkg/util/validation.go
annanay25: totally missed those, yeah makes sense to use those.

mapno (approved):

PR #1234, opened 2022-01-20
Correctly exclude traces where only one tag matches

joe-elliott:
**What this PR does**:
Updates the trace matches code to correctly exclude traces where either a resource or span attribute fails to match even when another matches.

We definitely want this one in 1.3.


pkg/model/trace/matches.go
```diff
@@ -14,21 +14,28 @@ import (
...
 
 	var rootSpan *v1.Span
 	var rootBatch *v1.ResourceSpans
 	// todo: is it possible to shortcircuit this loop?
 	for _, b := range trace.Batches {
-		if !tagFound && searchAttributes(req.Tags, b.Resource.Attributes) {
-			tagFound = true
+		batchMatch := attributesNotFound
+		if traceMatch != attributesPassedMatch {
+			batchMatch = searchAttributes(req.Tags, b.Resource.Attributes)
```
annanay25, line 38: this logic is slightly hard to follow because matching is split between resource attributes and trace attributes. Would it be easier to have a `map[string]bool` initialised with the query parameters, and we can check the state of the map after going through each batch?

pkg/model/matches_test.go
```diff
@@ -170,98 +170,112 @@ func TestMatches(t *testing.T) {
...
+		{
+			name:  "one includes/resource tag excludes",
+			trace: testTrace,
+			req: &tempopb.SearchRequest{
+				Start: 12,
+				End:   15,
+				Tags:  map[string]string{"foo": "bar", "service.name": "blerg"},
+			},
+			expected: nil,
+		},
```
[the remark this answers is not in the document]
joe-elliott: good call. done.

pkg/model/trace/matches.go
joe-elliott: yeah, agree that this is more complex than i was hoping. having to compare across resource and span as well as having to still iterate spans even if the batch fails a match really adds some complexity.

unsure on your suggestion though. can you provide more details?

annanay25 (approved):

PR #1240, opened 2022-01-21
Tempo Backend Search: Actual Fixes

joe-elliott:
**What this PR does**:
The previous PR (#1234) incorrectly tested for all conditions to be true on a single span which would make backend search behave differently then ingester search. This PR

- correctly tests to make sure query conditions are true across the entire trace
- adds support for `name`, `status.code` and `error` tags
- removes an errant .zip file from the repo

Respect to @annanay25 for his earlier suggestion. My approach of asserting conditions per span was wrong from the start! Using a map to track conditions across the entire trace was the correct approach.

### DIFF
--- a/pkg/model/trace/matches.go
+++ b/pkg/model/trace/matches.go
@@ -14,31 +14,23 @@ import (
 
 const RootSpanNotYetReceivedText = "<root span not yet received>"
 
-const (
-	notFound    = 1
-	failedMatch = 2
-	passedMatch = 3
-)
-
 func MatchesProto(id []byte, trace *tempopb.Trace, req *tempopb.SearchRequest) (*tempopb.TraceSearchMetadata, error) {
 	traceStart := uint64(math.MaxUint64)
 	traceEnd := uint64(0)
 
-	traceMatch := notFound
-	if len(req.Tags) == 0 {
-		traceMatch = passedMatch
+	// copy the tags from tempopb.SearchRequest. This map is then mutated by the two match* functions below
+	// if, at the end of this loop, there are no more entries in the map then we matched all tags
+	tagsToFind := map[string]string{}
+	for k, v := range req.Tags {
+		tagsToFind[k] = v
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/model/trace/matches.go
```diff
@@ -14,31 +14,23 @@ import (
...
 	// todo: is it possible to shortcircuit this loop?
 	for _, b := range trace.Batches {
-		// check if the batch matches. we will use the value of batchMatch later to determine if
-		//  - we need to bother checking if the spans match
-		//  - we can mark the entire trace as matching
-		batchMatch := attributesNotFound
-		if traceMatch != attributesPassedMatch {
-			batchMatch = searchAttributes(req.Tags, b.Resource.Attributes)
+		if !allTagsFound(tagsToFind) {
+			matchAttributes(tagsToFind, b.Resource.Attributes)
```
mdisibio, line 33: Probably good to add a nil check around `b.Resource`.  I think it mostly crops up in testing.

pkg/model/trace/matches.go
```diff
@@ -114,12 +99,37 @@ func MatchesProto(id []byte, trace *tempopb.Trace, req *tempopb.SearchRequest) (
...
+func allTagsFound(tagsToFind map[string]string) bool {
+	return len(tagsToFind) == 0
+}
+
+// matchSpan searches for reserved tag names and maps them to actual
+// properties of the span. it removes any matches it finds from
+// the provided tags map
+func matchSpan(tags map[string]string, s *v1.Span) {
+	if name, ok := tags[search.SpanNameTag]; ok {
+		if name == s.Name {
```
mdisibio, line 111: I think this should be case-insensitive substring match, if we want to match ingester search exactly.  Don't feel strongly, as this is more accurate and seems fine.  Same note applies to `matchAttributes`, case-insensitivity.

mdisibio (approved):
Left a couple thoughts, not-blocking.

PR #1241, opened 2022-01-23
Fix Query-frontend doesn't log request details when request is cancelled

### LINKED ISSUE #1136: Query-frontend doesn't log request details when request is cancelled
**Describe the bug**

The query-frontend logs some details about the query when it completes, for example:

```
level=info ts=2021-11-25T11:32:38.59660534Z caller=handler.go:96 tenant=1 method=GET traceID=57060d6b589d1903 url="/tempo/api/search?x-dbg-exhaustive=true&limit=20" duration=7.079269575s response_size=96 status=200
```

If you cancel the request before it completes, this log does not appear.

**To Reproduce**

1. Run Tempo (single binary or distributed, shouldn't matter)
2. Monitor logs from the query-frontend
3. Perform a query that is slow (so you have enough time to cancel it)
4. Cancel the request
5. Nothing is logged by the query-frontend

_Note: it's surprisingly difficult to get a slow query in a normal sized Tempo. To makes this easier to test you might want to introduce a `time.Sleep(5 * time.Second)` somewhere in the query path._

**Expected behavior**

I'd expect to see the usual logline, but with `status=499`.

**Additional Context**

The log line is printed here:

https://github.com/grafana/tempo/blob/ba93a400a4968c303cdbff9f5ccf88f6194489ca/modules/frontend/handler.go#L96-L104

I believe we should not return early when the request fails:

https://github.com/grafana/tempo/blob/ba93a400a4968c303cdbff9f5ccf88f6194489ca/modules/frontend/handler.go#L67-L76

adityapwr:
**What this PR does**:
I've added a logline before returning. When the request is canceled a usual logline(info) is printed but with status 499

Fixes #1136

**Checklist**
- [x] Tests updated
- [ ] Documentation added


adityapwr:
I added 5sec sleep in `FindTraceByID` to test this, here are some logs from my test.
```
grafana_1                   | logger=context userId=0 orgId=1 uname= t=2022-01-23T10:47:14.27+0000 lvl=eror msg="Query data error" error="failed to query data: failed get to tempo: Get \"http://tempo:3200/api/traces/bee1dd5237c8e117\": context canceled" remote_addr=172.20.0.1
grafana_1                   | logger=context userId=0 orgId=1 uname= t=2022-01-23T10:47:14.27+0000 lvl=eror msg="Request Completed" method=POST path=/api/ds/query status=500 remote_addr=172.20.0.1 time_ms=415 size=30 referer="http://localhost:3000/explore?orgId=1&left=%5B%22now-1h%22,%22now%22,%22Tempo%22,%7B%22query%22:%2249f4616bce4a7c98%22,%22queryType%22:%22nativeSearch%22,%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22tempo%22,%22uid%22:%22tempo-query%22%7D,%22serviceName%22:%22checkoutservice%22%7D%5D&right=%5B%221642931011494%22,%221642934611494%22,%22Tempo%22,%7B%22query%22:%22bee1dd5237c8e117%22,%22queryType%22:%22traceId%22,%22refId%22:%22A%22%7D%5D"
tempo_1                     | level=error ts=2022-01-23T10:47:14.280004272Z caller=frontend_processor.go:70 msg="error processing requests" address=127.0.0.1:9095 err="rpc error: code = Unknown desc = context canceled"
tempo_1                     | level=error ts=2022-01-23T10:47:14.28024218Z caller=frontend_processor.go:70 msg="error processing requests" address=127.0.0.1:9095 err="rpc error: code = Unknown desc = context canceled"
tempo_1                     | level=info ts=2022-01-23T10:47:14.280543766Z caller=handler.go:72 tenant=single-tenant method=GET traceID=463711e40d6ecc8f url=/api/traces/bee1dd5237c8e117 duration=413.432985ms response_size=0 status=499
tempo_1                     | level=error ts=2022-01-23T10:47:14.280731384Z caller=frontend_processor.go:70 msg="error processing requests" address=127.0.0.1:9095 err="rpc error: code = Unknown desc = context canceled"
tempo_1                     | level=error ts=2022-01-23T10:47:14.280885505Z caller=frontend_processor.go:70 msg="error processing requests" address=127.0.0.1:9095 err="rpc error: code = Unknown desc = context canceled"
tempo_1                     | level=error ts=2022-01-23T10:47:14.281478586Z caller=frontend_processor.go:70 msg="error processing requests" address=127.0.0.1:9095 err="rpc error: code = Unknown desc = context canceled"
```

modules/frontend/handler.go
```diff
@@ -67,6 +68,17 @@ func (f *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		writeError(w, err)
+		if err == context.Canceled {
+			level.Info(f.logger).Log(
+				"tenant", orgID,
+				"method", r.Method,
+				"traceID", traceID,
+				"url", r.URL.RequestURI(),
+				"duration", time.Since(start).String(),
+				"response_size", 0,
+				"status", StatusClientClosedRequest,
```
joe-elliott, line 79: is this necessarily true? if the query-frontend itself cancels the context won't it return the same error?

joe-elliott:
Thanks for the contribution! I have wanted something like this in the past. Had one question regarding the status returned.

modules/frontend/handler.go
adityapwr: thank you for the review. I agree, is there a way to differentiate between client and query-frontend cancellation? Will this work`if err == ctx.Err()`?

joe-elliott: ok, apologies. I didn't look closely at this PR/code before I responded. 

Let's do this:

- Instead of only logging an error if context is canceled, let's log every error. 
- For status let's just use a generic 500 since we don't know what the cause of the error is. 
- Add this log line to both the `resp == nil` and `err != nil` blocks
- change `writeError` to return an error instead of calling `server.WriteError` directly. That way we can also add the error string. 
- Log a new field after "status": 
  `"err", err.String()`

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@
 * [BUGFIX] Ingester startup panic `slice bounds out of range` [#1195](https://github.com/grafana/tempo/issues/1195) (@mdisibio)
 * [BUGFIX] Update goreleaser install method to `go install`. [#](https://github.com/grafana/tempo/) (@mapno)
 * [BUGFIX] tempo-mixin: remove TempoDB Access panel from `Tempo / Reads`, metrics don't exist anymore [#1218](https://github.com/grafana/tempo/issues/1218) (@kvrhdn)
+* [BUGFIX] Fixed issue when query-frontend doesn't log request details when request is cancelled [#1136](https://github.com/grafana/tempo/issues/1136) (@adityapwr)
```
joe-elliott, line 43: move this up to "unreleased" please

joe-elliott (changes requested):

CHANGELOG.md
adityapwr: thank you, I've moved it to unreleased

modules/frontend/handler.go
adityapwr: thank you, I've made changes.

joe-elliott:
I think this PR is ready to go. It just needs a small change for the linter:

>  Error: `reponse` is a misspelling of `response` (misspell)

adityapwr:
apologies, I missed lint. I'll take care from now on.

joe-elliott:
> apologies, I missed lint. I'll take care from now on.

no worries! I forget to lint before pushing all the time :+1:

joe-elliott (approved):
awesome! thanks for the contribution!

PR #1249, opened 2022-01-25
Copy `cortex/pkg/distributor` requirements into Tempo

zalegrala:
**What this PR does**:

As part of the migration away from `cortex`, here we copy what we need
to satisfy the distributor.

### DIFF
--- a/modules/distributor/config.go
+++ b/modules/distributor/config.go
@@ -4,7 +4,6 @@ import (
 	"flag"
 	"time"
 
-	cortex_distributor "github.com/cortexproject/cortex/pkg/distributor"
 	"github.com/grafana/dskit/flagext"
 	"github.com/grafana/dskit/ring"
 	ring_client "github.com/grafana/dskit/ring/client"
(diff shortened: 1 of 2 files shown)


zalegrala:
That cleans it up quite a bit. :)

modules/distributor/config.go
```diff
@@ -4,10 +4,10 @@ import (
 	"flag"
 	"time"
 
-	cortex_distributor "github.com/cortexproject/cortex/pkg/distributor"
 	"github.com/grafana/dskit/flagext"
 	"github.com/grafana/dskit/ring"
 	ring_client "github.com/grafana/dskit/ring/client"
+	tempo_distributor "github.com/grafana/tempo/pkg/distributor"
```
joe-elliott, line 10: no reason to give this an alias. just leave it `distributor`

joe-elliott (commented):
one nit. looks good.

modules/distributor/config.go
zalegrala: I just didn't like shadowing the package name.

joe-elliott: I didn't realize that you had copied these to `/pkg`. See comment here: https://github.com/grafana/tempo/pull/1251#discussion_r792660048

joe-elliott (approved):

PR #1251, opened 2022-01-25
Copy `cortex/pkg/compactor` requirements into Tempo

zalegrala:
**What this PR does**:

As part of the migration away from cortex, here we copy what we need
to satisfy the compactor.

### DIFF
--- a/modules/compactor/config.go
+++ b/modules/compactor/config.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"time"
 
-	cortex_compactor "github.com/cortexproject/cortex/pkg/compactor"
 	"github.com/go-kit/log"
 	"github.com/grafana/dskit/flagext"
 	"github.com/grafana/dskit/ring"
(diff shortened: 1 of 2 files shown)


modules/compactor/config.go
```diff
@@ -5,7 +5,8 @@ import (
 	"fmt"
 	"time"
 
-	cortex_compactor "github.com/cortexproject/cortex/pkg/compactor"
+	tempo_compactor "github.com/grafana/tempo/pkg/compactor"
```
joe-elliott, line 8: remove alias. `compactor` is fine.

joe-elliott (commented):
one nit. looks good.

mapno (approved):

modules/compactor/config.go
joe-elliott: Oh, I see, you moved it to `/pkg/compactor``. Let' s just move this code directly into `/modules/compactor`. It's only used by that module and we'd prefer keeping it isolated there to making a new module.

Generally we only put things in `/pkg` that will be shared by multiple components of Tempo.

zalegrala: Sure thing, I've updated to keep it all in the `modules/`.  For the distributor PR also.

joe-elliott (approved):

PR #1254, opened 2022-01-26
Copy `cortex/pkg/*` package dependencies into Tempo

zalegrala:
**What this PR does**:

As part of the cortex dependency removal, here we copy the parts of cortex that we need.

    - `github.com/cortexproject/cortex/pkg/frontend`
    - `github.com/cortexproject/cortex/pkg/frontend/v1`
    - `github.com/cortexproject/cortex/pkg/frontend/v1/frontendv1pb`
    - `github.com/cortexproject/cortex/pkg/frontend/v2`
    - `github.com/cortexproject/cortex/pkg/frontend/v2/frontendv1pb`
    - `github.com/cortexproject/cortex/pkg/querier/stats`
    - More of `github.com/cortexproject/cortex/pkg/util/*`
    - `github.com/cortexproject/cortex/pkg/tenant`
    - `github.com/cortexproject/cortex/pkg/scheduler`
    - Include proto build for frontend packages


zalegrala:
Needs https://github.com/grafana/tempo/pull/1253

modules/querier/stats/stats.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto3";
+
+package stats;
+
+option go_package = "github.com/grafana/tempo/modules/querier/stats";
```
zalegrala, line 5: I think maybe it has been a while since the protos had been generated, but more recent versions of protoc require that the go_package name contain more than just the string, so this has been updated here to reflect the name of the package.  Same goes for the two frontend proto packages.

zalegrala:
I'm confused about the CI failure here.  It seems I've updated all the references, but cortexproject/cortex/pkg/frontend/v1 still looks to get imported during the benchmark and e2e tests.

zalegrala:
Hmm, maybe the protos still need an update for the frontend packages.

Makefile
```diff
@@ -168,6 +168,8 @@ gen-proto:
 	$(call PROTO_GEN,$(PROTO_INTERMEDIATE_DIR)/resource/v1/resource.proto,./pkg/tempopb/)
 	$(call PROTO_GEN,$(PROTO_INTERMEDIATE_DIR)/trace/v1/trace.proto,./pkg/tempopb/)
 	$(call PROTO_GEN,pkg/tempopb/tempo.proto,./)
+	# protoc -I vendor -I . --go_out=./ --go_opt=paths=source_relative --go-grpc_out=./ --go-grpc_opt=paths=source_relative modules/frontend/v1/frontendv1pb/frontend.proto
```
joe-elliott, line 171: is there a reason not to generate these?

modules/frontend/v1/frontendv1pb/frontend.proto
```diff
@@ -0,0 +1,49 @@
+syntax = "proto3";
+
+// Protobuf package should not be changed when moving around go packages
+// in order to not break backward compatibility.
+package frontend;
+
+option go_package = "modules/frontend/v1/frontendv1pb";
```
mdisibio, line 7: Does this need the pathing `modules/....`?  Just comparing to Loki which also [forked the cortex frontend recently](  https://github.com/grafana/loki/blob/main/pkg/lokifrontend/frontend/v1/frontendv1pb/frontend.proto) and wondering if that is related to the discussion about not gen'ing these in Makefile.

zalegrala: I think the path just needs to be a depiction of where the module is located from the perspective of the `protoc` command.  In the note I left in the `Makefile`, you can see that we're including two `-I` flags to use as import paths.  In the case of the public packages, we pull them out of our vendor directory.  Perhaps we should make this the full public path for the package instead of the local relative path.  But if I leave the `go_package` name as a simple string, such as `frontendv1pb` I get a message like the following.

```
❯ protoc -I vendor -I . --go_out=./ --go_opt=paths=source_relative --go-grpc_out=./ --go-grpc_opt=paths=source_relative modules/frontend/v1/frontendv1pb/frontend.proto
protoc-gen-go: invalid Go import path "frontendv1pb" for "modules/frontend/v1/frontendv1pb/frontend.proto"

The import path must contain at least one period ('.') or forward slash ('/') character.

See https://developers.google.com/protocol-buffers/docs/reference/go-generated#package for more information.
```

This situation started happening maybe a year or 18 months ago when newer versions of `protoc` started requiring a more complete package name.  There are probably a few ways around it, like compiling an older version of `proto` or executing the build from a different directory.

One thing I'm not clear on after reviewing the Loki change you pointed out, is why there isn't a version header in the generated pb.gp.  In ours after generation we have the following: 

```
// Code generated by protoc-gen-go. DO NOT EDIT.
// versions:
//      protoc-gen-go v1.27.1
//      protoc        v3.19.2
// source: modules/frontend/v1/frontendv1pb/frontend.proto
```

This might be related to building with a new set of tooling, or perhaps there is a flag the Loki folks are using to skip including the header.  In either case, at first glance I wasn't able to figure out what proto version they were using to generate the pb.go files.

zalegrala: I'll follow up with the Loki team to see if there is an easier change we can make here.

zalegrala: This might be more complicated than I'd originally thought.  I wasn't aware of the gogo/protobuf project but it seems that both cortex and loki are using this to generate the pb.go files.  Should we also think about this?

Also, both cortex and loki are using a build image which is pinned to a specific package version of the gogo/protobuf package, but perhaps this is not compatible with newer protobufs.

https://github.com/cortexproject/cortex/blob/master/build-image/Dockerfile#L36
https://github.com/grafana/loki/blob/main/loki-build-image/Dockerfile#L75

I also I see the gogo/protobuf project is looking for new ownership, and hasn't had much contribution recently, so I don't know where we should stand here, since this seems like the most we'll get of a deprecation warning if new ownership doesn't arrive to update the project for newer protobufs.

https://github.com/gogo/protobuf

I'm also not clear the benefits of this over google/protobuf.

Thoughts on how to move forward?

zalegrala: Doing the gogoproto thing now.

pkg/gogocodec/gogocodec.go
```diff
@@ -18,10 +18,11 @@ import (
 )
 
 const (
-	tempoProtoGenPkgPath  = "github.com/grafana/tempo/pkg/tempopb"
-	jaegerProtoGenPkgPath = "github.com/jaegertracing/jaeger/proto-gen"
-	jaegerModelPkgPath    = "github.com/jaegertracing/jaeger/model"
-	otelProtoPkgPath      = "go.opentelemetry.io/collector"
+	frontendProtoGenPkgPath = "github.com/grafana/tempo/modules/frontend"
```
joe-elliott, line 21: is this necessary? i'd prefer not doing this unless we have to.

zalegrala: I got some marshaling error in the e2e tests without it.  Which I think makes sense since we are using the gogoproto generator to write the pb.go files.

zalegrala: https://github.com/grafana/tempo/runs/5028797385?check_suite_focus=true

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * [BUGFIX]: Remove unnecessary PersistentVolumeClaim [#1245](https://github.com/grafana/tempo/issues/1245)
 * [BUGFIX] Fixed issue when query-frontend doesn't log request details when request is cancelled [#1136](https://github.com/grafana/tempo/issues/1136) (@adityapwr)
 * [BUGFIX] Fixed panic when using etcd as ring's kvstore [#1260](https://github.com/grafana/tempo/pull/1260) (@mapno)
+* [CHANGE] Copy cortex dependencies into Tempo [#1254](https://github.com/grafana/tempo/pull/1254) (@zalegrala)
```
joe-elliott, line 11: i'm inclined to not list this since it should be a no-op from a user's perspective.

cmd/tempo-serverless/go.mod
```diff
@@ -33,3 +33,6 @@ replace (
 // This function is used by jeagertracing/jaeger, but we can't update jaeger
 // without updating the open-telemetry/collector as well
 replace google.golang.org/grpc => google.golang.org/grpc v1.38.0
+
+// these have been added due to cortex incompatibilies. they can be removed once we drop cortex as a dependency
```
joe-elliott, line 37: i thought this PR dropped all remaining cortex deps? or are there some still out there?

joe-elliott (commented):
Great work. Couple of Q's

cmd/tempo-serverless/go.mod
zalegrala: Well, yes good catch. :)  There are two references currently.  Here I was thinking that we change the dependencies after the removal, but if you would prefer to update here, I can do that also.

Two of these just look like version pinings, but the other two look like forks we might want to trawl through the code for also.

CHANGELOG.md
zalegrala: Fair enough.

zalegrala: Dropped this commit.

cmd/tempo-serverless/go.mod
joe-elliott: This PR is already insane. I agree. Let's get this in and then we can do a separate cleanup on the go.mod file.

joe-elliott (approved):
my browser lags when i go to the "Files changed" tab. i can barely type this message

:see_no_evil:

PR #1255, opened 2022-01-27
Add licensing boiler plate in case we ever need it

RichiH:
It's a minuscle change and currently not needed, but it makes my life tracking changes across the various projects easier.


RichiH:
Trying to figure out why CLA assistant does not use my existing agreement.

RichiH:
Signed this version as well. Following up with torkel on Slack on why this happened.

kvrhdn (approved):

PR #1256, opened 2022-01-27
Enterprise jsonnet: add config to create tokengen job explicitly

kvrhdn:
**What this PR does**:
The jsonnet for GET creates a job to generate the token from a GET license (`tokengen`). This job should be created once when installing GET and the job can not be modified afterwards. To make this easier to manage, this PR adds a config option `run_tokengen_job` which is disabled by default and can be set to explicitly create the tokengen job.

_How it works:_

The first time a GET cluster is created the user should set:

```jsonnet
  _config+:: {
    run_tokengen_job: true,
  },
```

During `tk apply`, tanka will create the tokengen job (just like before). After the job has run, the user should remove this line and the tokengen job will not be modified anymore by tanka.

~~Fixes #<issue number>~~

**Checklist**
- [ ] ~~Tests updated~~
- [ ] Documentation added: update https://grafana.com/docs/enterprise-traces/next/setup/tanka/


joe-elliott (approved):

hedss (approved):
Perfect! Thank you for doing this!

PR #1259, opened 2022-01-31
Docs: add a new troubleshooting section for tag search issue

jleloup:
Signed-off-by: Julien Leloup <jleloup@talend.com>

**What this PR does**:

This PR adds a new troubleshooting case related to Tempo traces tag search failing to return values.

Does not fixes opened issues.

**Checklist**
- [ ] Tests updated
- [X] Documentation added

### DIFF
--- a/docs/tempo/website/troubleshooting/search-tag.md
+++ b/docs/tempo/website/troubleshooting/search-tag.md
@@ -5,23 +5,22 @@ weight: 476
 
 # Tag search
 
-While searching for traces in Grafana Explore UI: `Service Name` and `Span Name` dropdown lists are empty with a *No options found* message.
+An issue occurs while searching for traces in Grafana Explore. The **Service Name** and **Span Name** drop down lists are empty, and there is a `No options found` message.
 
-HTTP calls to Tempo Query Frontend on `/api/search/tag/service.name/values` would answer an empty set.
+HTTP requests to Tempo query frontend endpoint at `/api/search/tag/service.name/values` would respond with an empty set.
 
 Note: this happens on Grafana Tempo 1.3 or higher.
 
 ## Root cause
 
-Grafana Tempo 1.3 has introduced another way of looking for tags to be used later in Grafana Explore UI, especially `service.name` and `name` tags that are used in the `Service Name` & `Span Name` dropdown lists respectively.
+The introduction of a cap on the size of tags causes this issue.
 
-Queries used to fetch those tags are now submitted to a size limit set with the `max_bytes_per_tag_values_query` parameter (documented here: [Tempo configuration overrides](https://grafana.com/docs/tempo/latest/configuration/#overrides)
+Configuration parameter `max_bytes_per_tag_values_query` causes the return of an empty result
+when a query exceeds the configured value.
 
-If by any chance this query goes over the size configured there: it will return an empty result.
+## Solutions
 
-## Resolution
+There are two main solutions to this issue:
 
-There are two main course of actions to solve this issue:
-
-* reduce the cardinality of tags pushed to Tempo: reducing the number of unique tag values will reduce the size returned by the tag search query later on.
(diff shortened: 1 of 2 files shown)


joe-elliott (approved):
this looks great! going to let this rest for a bit before merging in case @KMiller-Grafana would like to comment.

thanks for the addition.

annanay25 (commented):
Thanks @jleloup. And sorry that you faced this. We are soon going to work on a way to pass warnings from Tempo -> Grafana that can bubble up issues like this and display them in Grafana.

docs/tempo/website/troubleshooting/search-tag.md
```diff
@@ -0,0 +1,27 @@
...
+Queries used to fetch those tags are now submitted to a size limit set with the `max_bytes_per_tag_values_query` parameter (documented here: [Tempo configuration overrides](https://grafana.com/docs/tempo/latest/configuration/#overrides)
+
+If by any chance this query goes over the size configured there: it will return an empty result.
+
+## Resolution
+
+There are two main course of actions to solve this issue:
+
+* reduce the cardinality of tags pushed to Tempo: reducing the number of unique tag values will reduce the size returned by the tag search query later on.
+* increase the `max_bytes_per_tag_values_query` in the override section of your Tempo configuration: there is no rule of thumb though 10Mb or even 50Mb is not unheard of.
```
KMiller-Grafana, line 27: Stating that there's an issue with unreleased, future versions of Tempo does not make sense. 

```suggestion
An issue occurs while searching for traces in Grafana Explore. The **Service Name** and **Span Name** drop down lists are empty, and there is a `No options found` message.

HTTP requests to Tempo query frontend endpoint at `/api/search/tag/service.name/values` would respond with an empty set.

Note: this happens on Grafana Tempo 1.3 or higher.

## Root cause

The introduction of a cap on the size of tags causes this issue.
Configuration parameter `max_bytes_per_tag_values_query` causes the return of an empty result
when a query exceeds the configured value.

## Solutions

There are two main solutions to this issue:

* Reduce the cardinality of tags pushed to Tempo. Reducing the number of unique tag values will reduce the size returned by a tag search query.
* Increase the `max_bytes_per_tag_values_query` parameter in the [overrides]({{< relref "../configuration/#overrides" >}}) block of your Tempo configuration to a value as high as 50MB.
```

KMiller-Grafana (changes requested):
Thanks for this. I've provided rewrites.

docs/tempo/website/troubleshooting/search-tag.md
jleloup: Allright so I updated the files according to your changes.

docs/tempo/website/troubleshooting/search-tag.md
```diff
@@ -0,0 +1,26 @@
...
+weight: 476
+---
+
+# Tag search
+
+An issue occurs while searching for traces in Grafana Explore. The **Service Name** and **Span Name** drop down lists are empty, and there is a `No options found` message.
+
+HTTP requests to Tempo query frontend endpoint at `/api/search/tag/service.name/values` would respond with an empty set.
+
+Note: this happens on Grafana Tempo 1.3 or higher.
```
KMiller-Grafana, line 12: I had hoped that you would reword this, since it doesn't make sense to state that future versions of Tempo would have this issue.  They might not.

For now, let's just remove the reference.  I a reader is looking at the docs for a version of the product that has this issue, they'll see this section.
```suggestion
```

KMiller-Grafana (changes requested):
One more change that is needed.  Note that next time, if you click on the commit suggestion button, it'll be easier to re-review.

jleloup:
> One more change that is needed. Note that next time, if you click on the commit suggestion button, it'll be easier to re-review.

Done. I actually was not aware of this "Commit suggestion" feature, quite useful indeed :)

KMiller-Grafana (approved):
Thanks.

PR #1261, opened 2022-02-01
Fork the Prometheus Byte Pool

joe-elliott:
**What this PR does**:
Forks the [prometheus byte pool](https://github.com/prometheus/prometheus/tree/c954cd9d1d4e3530be2939d39d8633c38b70913f/util/pool) in order to make a few small changes:

- Updated to work with `[]byte` only
- `Put` changed to only accept byte slices whose lengths align with the created buckets. This prevents panics when byte slices are added to the pool that did not originate from it. Given our use of the byte pool it is very difficult to guarantee that those slices we add back to the pool came from it. 

This PR is required for #1227.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/tempopb/pool/pool.go
+++ b/pkg/tempopb/pool/pool.go
@@ -1,0 +1,74 @@
+// Forked with love from: https://github.com/prometheus/prometheus/tree/c954cd9d1d4e3530be2939d39d8633c38b70913f/util/pool
+// This package was forked to provide better protection against putting byte slices back into the pool that
+// did not originate from it.
+
+package pool
+
+import (
+	"sync"
+)
+
+// Pool is a bucketed pool for variably sized byte slices.
+type Pool struct {
+	buckets []sync.Pool
+	sizes   []int
+	// make is the function used to create an empty slice when none exist yet.
+	make func(int) []byte
+}
+
+// New returns a new Pool with size buckets for minSize to maxSize
+// increasing by the given factor.
+func New(minSize, maxSize int, factor float64, makeFunc func(int) []byte) *Pool {
+	if minSize < 1 {
+		panic("invalid minimum pool size")
+	}
+	if maxSize < 1 {
+		panic("invalid maximum pool size")
+	}
+	if factor < 1 {
+		panic("invalid factor")
+	}
+
+	var sizes []int
+
+	for s := minSize; s <= maxSize; s = int(float64(s) * factor) {
+		sizes = append(sizes, s)
+	}
+
+	p := &Pool{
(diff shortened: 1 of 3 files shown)


pkg/tempopb/pool/pool.go
```diff
@@ -0,0 +1,74 @@
...
+// Put adds a slice to the right bucket in the pool. This method has been adjusted from its initial
+// implementation to ignore byte slices that dont have the correct size
+func (p *Pool) Put(s []byte) {
+	c := cap(s)
+	for i, size := range p.sizes {
+		if c == size {
+			p.buckets[i].Put(s) // nolint: staticcheck
+		}
+		if c < size {
+			return
```
annanay25, line 71: Should we resize the buffer and `Put()` it into the smaller bucket in this case?

pkg/tempopb/pool/pool_test.go
```diff
@@ -0,0 +1,74 @@
...
+		},
+	}
+	for _, c := range cases {
+		ret := testPool.Get(c.size)
+		require.Equal(t, c.expectedCap, cap(ret))
+		testPool.Put(ret)
+	}
+}
+
+func TestPoolIgnoresRandomCapSlices(t *testing.T) {
```
annanay25, line 42: IIUC, this doesn't test if the pool ignores random cap sizes because of the `>=` in `require.True(t, cap(ret) >= size)`, but this test is still valuable.

Maybe we can add another test that tries to add a single buffer with a random (predetermined) size and checks if that exists in the pool?

pkg/tempopb/pool/pool.go
joe-elliott: I went down this path for a bit but decided against it because it complicates the code quite a bit and it should only happen in exceptional cases. It happens in #1227 only during the transition from the old format to the next. 

A possible improvement in the future? I was wondering what our code would look like if every time we needed a byte slice anywhere we used the pool. A change like that would make it possible.

pkg/tempopb/pool/pool_test.go
joe-elliott: You can't really check if something exists in the pool. It's rather opaque. I think this code is about as close as you could get to what you're suggesting although we could do it more explicitly. I do like that this tests to make sure that the pool will not return a too small buffer if you put and get a bunch of random sized slices to/from it.

I'll change the name for clarity.

annanay25: Oh by "check if that exists in the pool" I meant just requesting back the same size buffer, but yeah fair

annanay25 (approved):

PR #1263, opened 2022-02-02
Tempo Mixin: TempoRequestLatency improvements

joe-elliott:
**What this PR does**:
- Made latency and exclude list configurable
- Added `debug_pprof` to the exclude list


joe-elliott:
`TestPoolReusesSlice` is already flaking. Removing.

zalegrala (approved):

PR #1267, opened 2022-02-03
Makefile: add update-mod to tidy go.mod's

mapno:
**What this PR does**:

Break vendor-check into a new command update-mod, which is responsible
for keeping tempo's and tempo serverless' go.mod and tempo vendor file tidy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,7 +179,7 @@ gen-flat:
 ### Check vendored files and generated proto
 .PHONY: vendor-check
 vendor-check: gen-proto gen-flat update-mod
-	git diff --exit-code -- go.sum go.mod vendor/ pkg/tempopb/ pkg/tempofb/
+	git diff --exit-code -- **/go.sum **/go.mod vendor/ pkg/tempopb/ pkg/tempofb/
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -178,10 +178,15 @@ gen-flat:
 
 ### Check vendored files and generated proto
 .PHONY: vendor-check
-vendor-check: gen-proto gen-flat
+vendor-check: gen-proto gen-flat update-mod
+	git diff --exit-code -- go.sum go.mod vendor/ pkg/tempopb/ pkg/tempofb/
```
kvrhdn, line 182: I think we should include `cmd/tempo-serverless/go.sum` and `go.mod` in here as well. Otherwise they aren't included.

kvrhdn: Or maybe we could do
```suggestion
	git diff --exit-code -- **/go.sum **/go.mod vendor/ pkg/tempopb/ pkg/tempofb/
```

mapno: Applied the suggestion. Nice catch, thanks

kvrhdn (approved):

zalegrala (approved):

PR #1271, opened 2022-02-03
Remove patch version from release branch

mapno:
**What this PR does**:

Remove patch version from release branch in release guide.


KMiller-Grafana (approved):
The location of release branch creation within the release process is likely to change in the future.  For now, this is better.  Thanks for updating the procedure.

joe-elliott (changes requested):
If we do this we should go back and rename old release branches to match this pattern. I don't think that will break anything, but unsure.

Let's discuss as a team and decide what path to move forward with.

joe-elliott (approved):
We're good here. I'm going to push branches named using this standard for all of the old releases. For now we'll leave the old ones as well.

PR #1278, opened 2022-02-08
GCP Drone Serverless 

joe-elliott:
**What this PR does**:
Adds steps to build the Google Cloud Functions code package and push it to a GCS bucket.

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -18,9 +18,7 @@ local pipeline(name, arch = 'amd64') = {
         'refs/tags/v*',
         // weekly release branches
         'refs/heads/r?',
-        'refs/heads/r??',  // jpe name r* builds correctly. what branches do we execute on?
-        // jpe remove
-        'drone-serverless',
+        'refs/heads/r??',
     ],
   },
 };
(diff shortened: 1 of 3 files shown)


.drone/drone.jsonnet
```diff
@@ -35,12 +35,23 @@ local secret(name, vault_path, vault_key) = {
...
 // secret needed for dep-tools
 local gh_token_secret = secret('gh_token', 'infra/data/ci/github/grafanabot', 'pat');
 
+# gcs buckets to copy serverless functions to
+local gcp_serverless_deployments = [ 
+  { 
+    bucket: 'ops-tools-tempo-function-source',
+    secret: fn_upload_ops_tools_secret.name,
+  },
+];
```
annanay25, line 53: Does this array need to be updated every time we add a new cluster/region?

annanay25 (approved):
This is great 🚀

.drone/drone.jsonnet
joe-elliott: This will need to be updated everytime we add a new bucket. I'd like as many clusters as make sense to share the same buckets. Going to play with that some today.

PR #1282, opened 2022-02-11
[Metrics generator] Initial implementation

mapno:
**What this PR does**:

This PR contributes the first implementation of a new component: the metrics-generator. This implementation is based on the design proposal: [metrics generator](https://github.com/grafana/tempo/pull/1206)

The metrics-generator is a new optional component, that sits in the write path and derives metrics from traces. If present, the distributor will write received spans to both the ingester and the metrics-generator. The metrics-generator processes spans and writes metrics to a Prometheus datasource using the Prometheus remote write protocol.

You can see an architectural diagram below:

```
                                                                      │
                                                                      │
                                                                   Ingress
                                                                      │
                                                                      ▼
                                                          ┌──────────────────────┐
                                                          │                      │
                                                          │     Distributor      │
                                                          │                      │
                                                          └──────────────────────┘
                                                                    2│ │1
                                                                     │ │
                                                  ┌──────────────────┘ └────────┐
┌ ─ ─ ─ ─ ─ ─ ─ ─                     ┏━━━━━━━━━━━━━━━━━━━━━━┓      ┌──────────────────────┐
                 │                    ┃                      ┃      │                      │
│   Prometheus    ◀────Prometheus ────┃  Metrics-generator   ┃      │       Ingester       │◀───Queries────
                 │    Remote Write    ┃                      ┃      │                      │
└ ─ ─ ─ ─ ─ ─ ─ ─                     ┗━━━━━━━━━━━━━━━━━━━━━━┛      └──────────────────────┘
                                                                                │
                                                                                │
                                                                                │
                                                                                ▼
                                                                       ┌─────────────────┐
                                                                       │                 │
                                                                       │     Backend     │
                                                                       │                 │
                                                                       └─────────────────┘
```

Initially, the metrics-generator will be able to derive two different sets of metrics: span-metrics and service graphs.

* Span-metrics: aggregates request, error and duration metrics (RED) from span data. This processor computes the total count and the duration of spans for every unique combination of dimensions. Dimensions can be the service name, the operation, the span kind, the status code and any tag or attribute present in the span.
* Service graphs: A service graph is a visual representation of the interrelationships between various services. The service graph processor will analyse trace data and generate metrics describing the relationship between the services. These metrics can be used by e.g. Grafana to draw a service graph. For a more in-depth look, check in [here](https://grafana.com/docs/tempo/next/grafana-agent/service-graphs/)

We encourage everyone to take a look. All feedback is highly appreciated. Feel free to do so in this PR, or by pinging us (@kvrhdn and @mapno) at [Slack](grafana.slack.com).

**Known issues / TODOs**

A list of open tasks that we will tackle after this PR is merged and before the next Tempo release. Once this PR is merged we can create dedicated issues fort them.

- The distributor spawns a new goroutine for every push, we should optimize this by for instance introducing a queue. We want a way to limit the amount of pending requests to the metrics-generator in case they are having issues and requests start backing up.

- When generating metrics, we should take the timestamp of the received span into account. Spans can take a long time to reach Tempo (requests can be retried multiple times etc.). The timestamp could even be in the future. We should either figure out a way to correct the metrics that have been sent or refuse too old/new spans.

- We should add controls to limit the amount of series we generate/remote write.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


cmd/tempo/app/app.go
```diff
@@ -161,16 +168,18 @@ func newDefaultConfig() *Config {
...
-	distributor  *distributor.Distributor
-	querier      *querier.Querier
-	frontend     *frontend_v1.Frontend
-	compactor    *compactor.Compactor
-	ingester     *ingester.Ingester
-	store        storage.Store
-	MemberlistKV *memberlist.KVInitService
+	Server        *server.Server
+	ring          *ring.Ring // TODO should we rename this to ingesterRing?
+	generatorRing *ring.Ring
```
joe-elliott, line 173: i believe that the ingester ring is shared here is b/c it's used by both distribs and queriers. this ring can probably remain local to the initGenerator() method.

modules/distributor/config.go
```diff
@@ -37,6 +37,9 @@ type Config struct {
 	//  note that setting these two config values reduces tolerance to failures on rollout b/c there is always one guaranteed to be failing replica
 	ExtendWrites bool `yaml:"extend_writes"`
 
+	// Enables writes to metric-generator instances
+	MetricsGeneratorEnabled bool `yaml:"metrics_generator_enabled"`
```
joe-elliott, line 41: Wondering if we should consolidate this and `search_enabled` under some kind of features config?
```
features:
  search_enabled:
  metrics_generator_enabled:
```

modules/distributor/distributor.go
```diff
@@ -300,6 +338,17 @@ func (d *Distributor) PushBatches(ctx context.Context, batches []*v1.ResourceSpa
 		recordDiscaredSpans(err, userID, spanCount)
 	}
 
+	if d.cfg.MetricsGeneratorEnabled && len(d.overrides.MetricsGeneratorProcessors(userID)) > 0 && err == nil {
+		// Handle requests sent to the metrics-generator in a separate goroutine, this way we don't
+		// influence the overall write
+		go func() {
```
joe-elliott, line 344: Already discussed but want to note the concerns about this goroutine. We can definitely merge as is, but I think long term we want a queue here. We will definitely want a way to limit the amount of pending requests to the metrics generators in case they are having issues and requests start backing up.

cmd/tempo/app/app.go
kvrhdn: Well spotted! I'll remove it from the struct.

modules/distributor/config.go
kvrhdn: Sounds good to me. These flags are unlikely to change a lot after Tempo is set up, so having them in component specific config blocks isn't that important. We just have to caution with adding too much here.

I'm fine with moving `metrics_generator_enabled` to the global config in this PR but introducing the `features` section and moving `search_enabled` should probably be done in a separate PR. This move will require some migration code/docs.

modules/distributor/distributor.go
kvrhdn: Yep :+1: I'll add a section to the PR description to list these tasks. Once the PR is merged we can create dedicated issues for them.

modules/generator/instance.go
```diff
@@ -0,0 +1,279 @@
...
+	defer i.processorsMtx.Unlock()
+
+	deletedProcessor, ok := i.processors[processorName]
+	if !ok {
+		return
+	}
+
+	delete(i.processors, processorName)
+
+	err := deletedProcessor.Shutdown(context.TODO(), i.registry)
```
joe-elliott, line 196: doing this under lock could be dangerous if shutdown includes flushing any data. can we do this outside the lock?

or maybe just establishing the expectation that `Shutdown` has to be synchronous and fast is enough.

modules/generator/generator.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+func (g *Generator) running(ctx context.Context) error {
+	collectMetricsTicker := time.NewTicker(g.cfg.CollectionInterval)
+	defer collectMetricsTicker.Stop()
+
+	for {
+		select {
+		case <-collectMetricsTicker.C:
+			g.collectMetrics()
```
joe-elliott, line 160: if the timeout on the RW client is > then the collection interval we could start to see some backup here. Should we force the client timeout to be equal to the collection interval?

joe-elliott (commented):
Additional thoughts. Things are looking good. I will say that e2e tests would be nice to have, but not necessary for merging this PR.

modules/generator/processor/registry.go
```diff
@@ -0,0 +1,153 @@
...
+				bucketLabels := withLabel(labels, "__name__", fmt.Sprintf("%s_bucket", metricFamily.GetName()))
+				for _, bucket := range histogram.GetBucket() {
+
+					if bucket.GetUpperBound() == math.Inf(1) {
+						addedInfBucket = true
+					}
+
+					// TODO make a complete copy of this slice, not sure why this is needed
+					//  it works without in registry_test.go but fails in servicegraphs_test.go 🙃
+					//  in the servicegraphs test the le labels gets overwritten with the latest value 🤷
```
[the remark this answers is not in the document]
kvrhdn: Yeah, it is. I can take a deeper look at this again. I think the issue is that we take a shallow copy of the slice and append a new label to each copy. Since the shallow copies share the same underlying buffer, the appends overwrite each other.

modules/generator/generator.go
kvrhdn: We enforce the deadline in `g.collectMetrics()`. So each tenant can use up to collection interval seconds. But agree this doesn't avoid backup issues entirely. I.e. if every tenant takes 10 seconds and we have 4 tenants this call would still hang for way too long.
I'll add a global context enforcing the deadline across all tenants.

We will also want to revisit this code when we add per-tenant collection intervals.

modules/generator/instance.go
kvrhdn: Shutting down the processor should not flush any data. This is only meant to release any resources created by the processor (e.g. goroutines). If we want to flush the data before shutdown we should call `i.collectAndPushMetrics` first (like we do in `instance.shutdown`). `processor.Shutdown` will unregister all the counters from the registry, so this data is deleted immediately.

Agree there is no reason to do this while holding the lock, besides that the code is simpler. Once the processor is removed from the map we can release the lock.

Let me rethink this whole lock situation a bit, we should be able to make this a bit simpler / easier to follow 🤔

modules/distributor/config.go
kvrhdn: I've created a separate issue for this: https://github.com/grafana/tempo/issues/1287

cmd/tempo/app/app.go
kvrhdn: Okay, I've tried some options here and it's not super clear cut what the best option is. `generatorRing` is actually only used by the distributor, the distributor uses it as a `ring.ReadRing`.

First issue: `generatorRing` is a service, by returning it in `initGeneratorRing` it becomes its own module which updates the service. If we create the ring within `initDistributor` we have to use a `services.Manager` to wrap both services. This makes stuff more complicated.

Second issue: we could move the ring creation into `distributor.New`. This should work, but we would have to pass the `metrics_generator.ring` config in there as well because we need this to create the ring. So this would create a dependency from the `distributor` to the `generator` package.

We could avoid this by creating a new top-level config block `metrics_generator_ring`. We already have `ingester_client` and `metrics_generator_client`.
I think this would make it more clear that this config is shared by multiple components (the distributor and the generator), but it's yet another top-level config block. Thoughts?

modules/generator/generator.go
kvrhdn: Checking out this code a bit more, the current implementation already considers this. `g.collectMetrics` creates a context with a timeout of `g.cfg.CollectionInterval`. So if collecting metrics for all tenants exceeds the collection interval, we abort the operation.

Would it be clearer if I created this context in `running`? I added it in `collectMetrics` because then we can do `defer cancel()` to clean up resources.

Code: https://github.com/grafana/tempo/blob/4089b5299ac1ba6e4a70d7fe5eb10c83890067b9/modules/generator/generator.go#L253-L266

modules/distributor/config.go
kvrhdn: I've moved the flag. I noticed we are now passing both `searchEnabled` and `metricsGeneratorEnabled` to the distributor, introducing a struct here with all features might make sense.

modules/generator/instance.go
```diff
@@ -0,0 +1,279 @@
...
+
+	// remove processors that are not in desiredProcessors
+	for processorName := range i.processors {
+		_, ok := desiredProcessors[processorName]
+		if ok {
+			// processor is also in the desired list
+			continue
+		}
+
+		i.processorsMtx.RUnlock()
```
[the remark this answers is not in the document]
kvrhdn: Okay, I have simplified this section a lot, please take a look again :)

`updateProcessors` is now divided into two sections now:
- diff `desiredProcessors` with the existing processors
- if there is a diff, add/remove the processors

For the first part we only need a read lock and we can return early if there is no diff. So this shouldn't stall `pushSpans`.

modules/generator/processor/registry.go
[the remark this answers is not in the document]
kvrhdn: Okay, tracked down the issue. The other tests fail when they add 2 additional labels, the registry tests only adds at most 1. If you add 2 labels the capacity of the underlying slice grows and appending the `le` label does not allocate a new buffer -> memory is shared between different buckets and the value of the `le` label gets overwritten.
Summary: this is a bug caused by poor slice handling and it works sometimes because of how the slice is reallocated when it grows 😬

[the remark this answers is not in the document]
kvrhdn: Fixed this now 🙂

kvrhdn:
I think we've addressed all review comments now, another review would be welcome 🙂 

Agree adding an e2e test would be really nice, this is on our todo list. Not sure yet if we can create one now and that we will PR it later.

modules/generator/generator.go
joe-elliott: I see now. Nope, I'm good as is. Let's go!

joe-elliott (approved):
Let's go!

modules/generator/remotewrite/appender.go
```diff
@@ -0,0 +1,156 @@
...
+	a.exemplars = append(a.exemplars, prompb.Exemplar{
+		Labels:    labelsToLabelsProto(e.Labels),
+		Value:     e.Value,
+		Timestamp: e.Ts,
+	})
+	return 0, nil
+}
+
+func (a *remoteWriteAppender) Commit() error {
+	level.Debug(a.logger).Log("msg", "writing samples to remote_write target", "tenant", a.userID, "target", a.remoteWriter.Endpoint(), "count", len(a.samples))
```
annanay25, line 56: nit: Not sure if we should log `.Endpoint()` here as it can contain user/pass credentials? Is this true?

Thinking if we should use `.Name()` but I saw you're using the same name for all clients `metrics_generator`. I don't think we can keep the name unique because it seems to be used as a label in some metrics.

Same comment to the `.Endpoint()` logging done a few lines below.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -0,0 +1,128 @@
...
+	return &processor{
+		cfg: cfg,
+		now: time.Now,
+	}
+}
+
+func (p *processor) Name() string { return Name }
+
+func (p *processor) RegisterMetrics(reg prometheus.Registerer) error {
+	labelNames := []string{"service", "span_name", "span_kind", "span_status"}
```
annanay25, line 37: here: `span_status` is going to be an integer with values 0/1/2. That might be slightly confusing? do we want to turn this into an `error=true/false` at some point?

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -0,0 +1,128 @@
...
+	if len(p.cfg.Dimensions) > 0 {
+		labelNames = append(labelNames, p.cfg.Dimensions...)
```
annanay25, line 39: I definitely think we need some limits here to keep a check on the number of active series, if we're going to allow custom dimensions. Either here or in `appender.go` where we are appending labels/samples. Maybe for a future PR though?

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	p.spanMetricsCallsTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "traces",
```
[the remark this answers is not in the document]
kvrhdn: We are using the `traces` namespace here to match with the metrics generated by the Grafana Agent. If we generate the same metrics, it will be easier for people to reuse queries/dashboards for both.
The service graphs view in Grafana for instance expects metrics with the name `traces_service_graph_...`. We could make this configurable in Grafana, but for now we want to keep it as simple as possible.

modules/generator/processor/spanmetrics/spanmetrics.go
kvrhdn: Agree we need to limit the amount of series we create because this will impact the memory used by the generators. I wouldn't bother limiting it here (for now) since `dimensions` is controlled by the operator of Tempo, so it's easy to lower this when the cluster is getting in trouble.
At this point I'm more concerned about the series created by labels like `span_name` because these are dependent on the cardinality of the traces we are ingesting and we can't control that.

Limiting the active series in the appender makes a lot of sense, both to protect Tempo and to protect the downstream TSDB. I'll add it as a TODO to this PR and will make a tracking issue with all remaining tasks once this PR is merged.

modules/generator/processor/spanmetrics/spanmetrics.go
kvrhdn: We call `.String()` on `StatusCode` and the generated proto code uses an enum lookup to get the string label. So we are adding label values like `STATUS_CODE_UNSET` here.

- https://github.com/grafana/tempo/blob/metrics-generator-dev/modules/generator/processor/spanmetrics/spanmetrics.go#L110
- https://github.com/grafana/tempo/blob/metrics-generator-dev/pkg/tempopb/trace/v1/trace.pb.go#L177-L179

Example of the metrics that are generated:

Screenshot 2022-02-18 at 15 05 36

---

Regarding using `error=true/false` vs `span_status=STATUS_CODE_...`: I'm not sure. We already have this mix in various places:
- in Tempo search we allow searching on `error=true`, even though `error` is not necessarily a real attribute
- Grafana also shows `error=true` in the trace view because the trace view is based upon Jaeger

This processor is based upon [the spanmetrics processor from OTel Collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/spanmetricsprocessor) which exclusively works on OTLP. So that's why we only added the `span_status` label.

So for now I'd say we keep it simple and only add `span_status` (which aligns most closely to the underlying data) and if this is confusing/annoying we can look into adding an `error` label to the metrics.

modules/generator/remotewrite/appender.go
kvrhdn: Good spotting, I'm not entirely sure tbh 🤔  Looking at the Prometheus code we use, this _seems_ fine.

I think it's always possible that someone configures a URL like `http://<username>:<password>@example.com/`, which would get logged in this case.

This is not really how it is intended to be used though: the remote write config struct has dedicated fields to configure your credentials. `metrics_generator.remote_write.client` is just this Prometheus struct: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write
So there are fields like `basic_auth`, `authorization` and `oauth2` to configure credentials.

We pass this config to a Prometheus `WriteClient`. The `WriteClient` interface only exposes `Name` and `Endpoint` which makes it seem like this would be safe to print: https://github.com/prometheus/prometheus/blob/e239e3ee8b13b51b0f791a199813a14f74600a7e/storage/remote/queue_manager.go#L327-L336

Another data point: when sending the request fails, this is logged:
```
level=error ts=2022-02-17T14:28:59.380752142Z caller=appender.go:70 msg="error sending remote-write requests" tenant=116095 target=https://xxx.grafana.net/api/prom/push err="Post \"https://xxx.grafana.net/api/prom/push\": EOF"
```
Note that the error that is returned by the write client also contains the full endpoint.

annanay25: Tracked this down. No credentials should be published in the metric label: https://github.com/prometheus/prometheus/pull/9383

modules/generator/processor/spanmetrics/spanmetrics.go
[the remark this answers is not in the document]
annanay25: Hmm. Yeah I wish we could make this `tempo` but I don't see a clear path to making that change. I guess its ok to leave it as is.

modules/generator/processor/spanmetrics/spanmetrics.go
annanay25: Ok, sounds good. 👍 

Thanks for the explanation!

modules/generator/remotewrite/appender.go
kvrhdn: Nice find!

PR #1291, opened 2022-02-17
Replace CombineTraceProtos with new Combiner

mdisibio:
**What this PR does**:
This PR introduces a new Combiner which is similar to CombineTraceProtos but more efficient when combining more than two inputs.  The previous pairwise usage of CombineTraceProtos had a couple inefficiencies when combining more than two inputs: (a) the intermediate result was sorted every time (b) the hash of span tokens was rebuilt every time.  Combiner is stateful and improves this, which leads to significant reduction in cpu and memory.  Performance is identical when combining just 2 inputs.

Additionally, this PR changes the span/token hashing to 64-bit to reduce the collision rate.  Experimentally the collision rate of fnv32 approached 1 in 10,000 spans, which is significant because any collision results in a dropped span. 64-bit has no collisions up to the tested limit of a trace with 1M spans. Performance is still good.

**Feedback**
In order to maintain identical performance against 2 segments, Combiner must not save the span tokens for the second input, like how CombineTraceProtos did not. This can be generalized in that we never need to save the span tokens for the last input. These savings are significant enough to where it's worth accounting for, and there are many cases where we do know the length.  I would like feedback on the chosen ergonomics/naming/style and see if there is a better pattern.  For example:

```go
// Unknown number of inputs
c := NewCombiner()
c.Consume(tr)
c.Consume(tr)
    ...
c.Result()

// Known number of inputs
c := NewCombiner()
for i, tr := range traces {
    c.ConsumeWithFinal(tr, i == len(traces)-1)
}
c.Result()
```

**Benchmarks**
This benchmark combines trace 2 to 8 segments of 100K spans each. Improvements are greater as more segments are combined.
```
BenchmarkCombine/2/CombineTraceProtos-12         	      27	  44842636 ns/op	 6316197 B/op	  115510 allocs/op
BenchmarkCombine/2/Combiner-12                   	      25	  44691228 ns/op	 6319784 B/op	  115615 allocs/op

BenchmarkCombine/3/CombineTraceProtos-12         	       9	 125415228 ns/op	17512674 B/op	  291023 allocs/op
BenchmarkCombine/3/Combiner-12                   	      15	  72986714 ns/op	11202546 B/op	  175560 allocs/op

BenchmarkCombine/4/CombineTraceProtos-12         	       5	 241490759 ns/op	35993108 B/op	  524033 allocs/op
BenchmarkCombine/4/Combiner-12                   	      10	 108902778 ns/op	18498868 B/op	  233082 allocs/op

BenchmarkCombine/8/CombineTraceProtos-12         	       1	1119084275 ns/op	164185168 B/op	 2061117 allocs/op
BenchmarkCombine/8/Combiner-12                   	       5	 242659052 ns/op	 37625163 B/op	  469882 allocs/op
```

Should help #976 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/frontend/tracebyidsharding.go
```diff
@@ -65,7 +65,8 @@ func (s shardQuery) RoundTrip(r *http.Request) (*http.Response, error) {
 
 	var overallError error
 	var totalFailedBlocks uint32
-	overallTrace := &tempopb.Trace{}
+	combiner := trace.NewCombiner()
+	combiner.Consume(&tempopb.Trace{}) // The query path returns a non-nil result even if no inputs (which is different than other paths which return nil for no inputs)
```
mdisibio, line 69: This is another fiddly bit.  Some paths expect nil if no inputs, but this read path defaults to an empty trace.  I think it might be cleaner to make this an option in `NewCombiner` to control the default result:  either nil or empty trace.

pkg/model/trace/combine_test.go
```diff
@@ -1,17 +1,32 @@
 package trace
 
 import (
-	"hash/fnv"
+	"bytes"
+	"crypto/rand"
```
joe-elliott, line 5: nit: `math/rand`

pkg/model/trace/combine.go
```diff
@@ -70,15 +72,122 @@ func CombineTraceProtos(traceA, traceB *tempopb.Trace) (*tempopb.Trace, int) {
 	return traceA, spanCountTotal
 }
 
-// tokenForID returns a uint32 token for use in a hash map given a span id and span kind
+// token is unint64 to reduce hash collision rates.  Experimentally, it was observed
```
joe-elliott, line 75: nit: `uint64`

pkg/model/trace/combine.go
```diff
@@ -8,6 +8,8 @@ import (
 	"github.com/grafana/tempo/pkg/tempopb"
 )
 
+// nolint: staticcheck
+// Deprecated: This is only capable of pairwise combination. It is replaced by Combiner.
```
joe-elliott, line 12: it seems like you've caught almost all of the uses of `CombineTraceProtos`. Any reason not to finish the job and remove this method?

If we do keep this method, should we just make it a wrapper for the Combiner so we only have the logic once?

joe-elliott (commented):
Drooling over some of these benchmarks. We should definitely see some improvements in resource usage in both compaction and querying. A couple of baby nits and one real question.

Nice!

modules/frontend/tracebyidsharding.go
mdisibio: Downstream from this is the logic to determine 2xx/4xx/5xx from the responses, wasn't clear if nil could even be interpreted differently.  Therefore agree this seems better than introducing a change there, at least for now and we can always come back to it if not.

pkg/model/trace/combine.go
mdisibio: Was thinking it is useful to keep it for benchmark and logic comparisons, until the new Combiner has been more thoroughly tested, i.e. deprecate and delete in separate PRs.    The only remaining usage is in `func CombineForRead(obj []byte, dataEncoding string, t *tempopb.Trace) (*tempopb.Trace, error)` which only combines 2 segments so there's no benefit other than code cleanup.

joe-elliott: I would prefer just cleaning it up completely. I already feel like there's too many ways to combine objects and it's a bit confusing

If you feel strongly about keeping it though it's not a blocker. You've marked it deprecated which is a good indicator that no one should go near it unless they know what they're doing.

mdisibio: Sounds good, removed completely.

mdisibio:
Pushed an optimization to alloc the span map using the first input size, which saves a few more MB per call.  Thanks @tanner-bruce !

```
BenchmarkCombine/2/Combiner-12         	      27	  41954908 ns/op	 4650348 B/op	  113443 allocs/op
BenchmarkCombine/3/Combiner-12         	      15	  72924713 ns/op	 9526314 B/op	  173208 allocs/op
BenchmarkCombine/4/Combiner-12         	      10	 105436694 ns/op	16823280 B/op	  230750 allocs/op
BenchmarkCombine/8/Combiner-12         	       5	 237391203 ns/op	35946148 B/op	  467689 allocs/op
```

PR #1306, opened 2022-02-23
Add Content-Type header on query-frontend paths

### LINKED ISSUE #1185: Endpoints are missing a content-type header
Some of the API endpoints are returning a content-type of `text/plain` even if the `Accept` header on the request is set to `application/json` or `application/protobuf`. I noticed it on `/api/traces/{traceID}` and `/api/search` at least, there could be others as well.

wperron:
This commit adds the `Content-Type` header on the read paths of the Query Frontend.

Fixes #1185

**Checklist**
- [X] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/tracebyidsharding.go
+++ b/modules/frontend/tracebyidsharding.go
@@ -15,6 +15,7 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/golang/protobuf/proto"
 	"github.com/grafana/tempo/modules/querier"
+	"github.com/grafana/tempo/pkg/api"
 	"github.com/grafana/tempo/pkg/model/trace"
 	"github.com/grafana/tempo/pkg/tempopb"
 	"github.com/opentracing/opentracing-go"
@@ -175,8 +176,10 @@ func (s shardQuery) RoundTrip(r *http.Request) (*http.Response, error) {
 	}
 
 	return &http.Response{
-		StatusCode:    http.StatusOK,
-		Header:        http.Header{},
+		StatusCode: http.StatusOK,
+		Header: http.Header{
+			api.HeaderContentType: {api.HeaderAcceptProtobuf},
+		},
 		Body:          io.NopCloser(bytes.NewReader(buff)),
 		ContentLength: int64(len(buff)),
 	}, nil
--- a/modules/querier/http.go
+++ b/modules/querier/http.go
@@ -193,6 +193,7 @@ func (q *Querier) SearchHandler(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+	w.Header().Set(api.HeaderContentType, api.HeaderAcceptJSON)
 }
 
 func (q *Querier) SearchTagsHandler(w http.ResponseWriter, r *http.Request) {
@@ -217,6 +218,7 @@ func (q *Querier) SearchTagsHandler(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+	w.Header().Set(api.HeaderContentType, api.HeaderAcceptJSON)
 }
 
 func (q *Querier) SearchTagValuesHandler(w http.ResponseWriter, r *http.Request) {
@@ -249,4 +251,5 @@ func (q *Querier) SearchTagValuesHandler(w http.ResponseWriter, r *http.Request)
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+	w.Header().Set(api.HeaderContentType, api.HeaderAcceptJSON)
 }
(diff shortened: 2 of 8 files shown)


modules/frontend/tracebyidsharding.go
```diff
@@ -175,8 +176,10 @@ func (s shardQuery) RoundTrip(r *http.Request) (*http.Response, error) {
 	}
 
 	return &http.Response{
-		StatusCode:    http.StatusOK,
-		Header:        http.Header{},
+		StatusCode: http.StatusOK,
+		Header: http.Header{
+			api.HeaderContentType: {api.HeaderAcceptProtobuf},
```
joe-elliott, line 181: it's weird that we assume protobuf here b/c of a header that was set in frontend.go.

modules/querier/http.go
```diff
@@ -193,6 +193,7 @@ func (q *Querier) SearchHandler(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+	w.Header().Set(api.HeaderContentType, api.HeaderAcceptJSON)
```
joe-elliott, line 196: set content type in `TraceByIDHandler` as well?

joe-elliott (commented):
Thanks for taking a swing at this. "Content Negotation" is pretty rough in Tempo and I'm impressed with all the things you were able to find and correct.

I think the part that most concerns me is the way the the trace by id path works.
- frontend.go - force sets Accepts to protobuf
- tracebyidsharding.go - assumes that whatever it gets is protobuf b/c of something that happened in frontend.go

I know you didn't create this problem, but perhaps we could iron it out just a little bit? Thoughts on how we could clean this up?

modules/querier/http.go
wperron: Mhm, that's a good point 🤔  The TraceByID path is a bit odd because eventually it all reaches the `newTraceByIdMiddleware` in `frontend.go` where the payload is serialized according to the `Accept` header. I had to modify the `tracebyidsharding.go` because the tests weren't passing, but as it is, the tests are all passing. So while it would make sense to also add it to the `TraceByIDHandler` I also don't know if it'll make any practical difference.

modules/frontend/tracebyidsharding.go
wperron: It's not, the Trace Sharder [_only_ ever sends a protobuf message](https://github.com/grafana/tempo/blob/main/modules/frontend/tracebyidsharding.go#L166-L171), it's then unmarshalled and re-marshalled back into either protobuf or json in the `frontend.go` in the `newTraceByIDMiddleware` function [here](https://github.com/grafana/tempo/blob/main/modules/frontend/frontend.go#L137-L151)

modules/querier/http.go
joe-elliott: > where the payload is serialized according to the Accept header
Sadly I think `tracebyidsharding.go` makes no checks and [always assumes](https://github.com/grafana/tempo/blob/7d26ee26986411cf096f4a93ad1dd7fa422fa972/modules/frontend/tracebyidsharding.go#L120-L121) that the returned payload is proto. I don't expect a fix for that in this PR.

It then uses the `Accept` header in `frontend.go` to determine how to marshal the response for the caller.

Since we are adding the `Content-Type` header to other responses here I would like to just slap it in the `TraceByIDHandler` as well, but don't feel obligated to adjust the frontend.

wperron: Done 👍

modules/frontend/tracebyidsharding.go
joe-elliott: My bad, I meant it's weird we assume that the data received in the trace by id sharder (from the querier) [is proto](https://github.com/grafana/tempo/blob/7d26ee26986411cf096f4a93ad1dd7fa422fa972/modules/frontend/tracebyidsharding.go#L120-L121), but the only reason it is proto is because a [header that's set](https://github.com/grafana/tempo/blob/7d26ee26986411cf096f4a93ad1dd7fa422fa972/modules/frontend/frontend.go#L116) in frontend.go. 

I think we should set the `Accept` for proto in `tracebyidsharding.go` since that code also assumes that the returned data is proto. But these gross corners have been around for long before you made this improvement.

joe-elliott:
CI seems to be stuck? If you can resolve the changelog diff and push it will trigger it again.

modules/frontend/tracebyidsharding.go
wperron: Oh I see! I assumed this was intentional after reading the comment on that same line you linked, I guess it kinda make sense to keep content negotiation as close the edge as possible? Then again if internal communication is expected to be only protobuf, then setting the Accept header might just be superfluous 🤷

joe-elliott (approved):
This looks good. We've been having trouble with Github Actions CI lately. So once that passes we will merge. 

Thanks!

PR #1307, opened 2022-02-23
Tempo Serverless: Queriers Prefer Self

joe-elliott:
**What this PR does**:
- Adds a configuration option to allow the queriers to perform a configurable number of searches locally before reaching out to the external endpoints
- Adds docs for `max_concurrent_queries`

Ticks a box on #1175 

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/tempo/website/configuration/_index.md
+++ b/docs/tempo/website/configuration/_index.md
@@ -254,6 +254,12 @@ querier:
     # The default value of "" disables this feature.
     [search_external_endpoints: <list of strings> | default = <empty list>]
 
+    # If search_external_endpoints is set then the querier will primarily act as a proxy for whatever serverless backend
+    # you have configured. This setting allows the operator to have the querier prefer itself for a configurable
+    # number of queries. In the default case of 2 the querier will process up to 2 search requests before starting
+    # to reach out to search_external_endpoints.
+    [search_prefer_self: <int> | default = 2 ]
+
     # The query frontend turns both trace by id (/api/traces/<id>) and search (/api/search?<params>) requests
     # into subqueries that are then pulled and serviced by the queriers.
     # This value controls the overall number of simultaneous subqueries that the querier will service at once. It does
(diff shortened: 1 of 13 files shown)


docs/tempo/website/configuration/_index.md
```diff
@@ -254,6 +254,18 @@ querier:
     # The default value of "" disables this feature.
     [search_external_endpoints: <list of strings> | default = <empty list>]
 
+    # If search_external_endpoints is set then the querier will primarily act as a proxy for whatever serverless backend
+    # you have configured. This setting allows the operator to have the querier prefer itself for a configurable
+    # number of queries. In the default case of 2 the querier will process up to 2 search requests before starting
```
mdisibio, line 259: This is more of a question to check my understanding:  `2 search requests` sounds like the external endpoint won't be hit unless there are concurrent search apis being called, but I think it is actually 2 search request shards? i.e. part of a whole search from the frontend as divided up among blocks/pages.

modules/querier/querier.go
```diff
@@ -391,13 +395,23 @@ func (q *Querier) SearchTagValues(ctx context.Context, req *tempopb.SearchTagVal
...
-	if len(q.cfg.SearchExternalEndpoints) != 0 {
-		endpoint := q.cfg.SearchExternalEndpoints[rand.Intn(len(q.cfg.SearchExternalEndpoints))]
-		return searchExternalEndpoint(ctx, endpoint, req)
+	// if we have no external configuration always search in the querier
+	if len(q.cfg.SearchExternalEndpoints) == 0 {
+		return q.internalSearchBlock(ctx, req)
+	}
+
+	// if we have external configuration but there's an open slot locally then search in the querier
+	if q.searchPreferSelf.TryAcquire(1) {
```
mdisibio, line 404: It looks like setting preferSelf=0 will work as expected to disable this.  Can you confirm, and also it might be good to note that in the config help text.

docs/tempo/website/configuration/_index.md
joe-elliott: > I think it is actually 2 search request shards?

This is correct. I will clarify.

modules/querier/querier.go
joe-elliott: Yup, will clarify :+1:

mdisibio (approved):

PR #1311, opened 2022-02-28
Return value of error tooManySpansError instead of reference

mapno:
**What this PR does**:

* Return value of error tooManySpansError instead of reference. As a consequence, `errors.Is` was failing to detect it.
*  Remove error return in PushSpans method
    * We don't want to stop processing due to an error in one of the processor. Errors can be handled and logging inside the processors.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/generator.go
+++ b/modules/generator/generator.go
@@ -214,10 +214,7 @@ func (g *Generator) PushSpans(ctx context.Context, req *tempopb.PushSpansRequest
 		return nil, err
 	}
 
-	err = instance.pushSpans(ctx, req)
-	if err != nil {
-		return nil, err
-	}
+	instance.pushSpans(ctx, req)
 
 	return &tempopb.PushResponse{}, nil
 }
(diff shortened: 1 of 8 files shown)


kvrhdn (approved):
This change looks good to me :+1:

A more general question about error handling in the processors: I noticed `servicegraphs.PushSpans` never returns an error:

https://github.com/grafana/tempo/blob/4b26046226b99646411a436de21ebfb10eeff910/modules/generator/processor/servicegraphs/servicegraphs.go#L181

I'm guessing we don't do this because it stops processing for all processors:

https://github.com/grafana/tempo/blob/4b26046226b99646411a436de21ebfb10eeff910/modules/generator/instance.go#L222-L226

Should we remove `error` from the method signature? We don't use it anyway right now.
Or should we not `return err` in `instance.pushSpans` and continue processing when one processor fails?

kvrhdn (approved):

PR #1312, opened 2022-02-28
Add backport workflow

mdisibio:
**What this PR does**:
This PR adds a new GHA for backporting.  It checks merged PRs for new labels like `backport release-x.y` and creates a matching PR with the same changes against the `release-x.y` release branch.

n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,33 @@
+# This checks merged PRs for labels like "backport release-x.y" 
+# and opens a new PR backporting the same commit to the release branch.
+# This workflow also runs when the PR is labeled or opened, but will
+# will only check a few things and detect that the PR is not yet merged. 
+# At this time only squashed PRs are supported since the cherry-pick 
+# command does not include "-m <n>" arg required for merge commits.
+name: Backport
+on:
+  pull_request_target:
+    types:
+      - closed
+      - labeled
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Actions
+        uses: actions/checkout@v2
+        with:
+          repository: "grafana/grafana-github-actions"
+          path: ./actions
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -0,0 +1,33 @@
...
+          path: ./actions
+          ref: main
+
+      - name: Install Actions
+        run: npm install --production --prefix ./actions
+
+      - name: Run backport
+        uses: ./actions/backport
+        with:
+          token: ${{secrets.GITHUB_TOKEN}}
```
kvrhdn, line 31: Will `GITHUB_TOKEN` have sufficient permissions to push a commit and open a PR? According to this doc, it only has limited read permissions: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

Loki seems to use another token, not sure if this is necessary: https://github.com/grafana/loki/blob/main/.github/workflows/backport.yml#L24

mdisibio: Yep saw that Loki uses the other token, but I lean towards defaults/built-ins if we can.  In testing on my [personal fork](https://github.com/mdisibio/tempo/blob/main2/.github/workflows/backport.yml) it was sufficient, [resultant PR](https://github.com/mdisibio/tempo/pull/6).   

I believe it is related to the repo settings for [workflow permissions](https://github.com/grafana/tempo/settings/actions)
image



One other thing I noticed is that the author is different `github actions (bot)` vs `grafanabot`, so if we have a preference there, it could also be a deciding factor.

kvrhdn: I'm also in favour of using the built in options. That is one token less we have to maintain 🙂

kvrhdn (approved):
🚀

zalegrala (approved):

PR #1320, opened 2022-03-02
Avoid extending the block meta time window if a span is too old

zalegrala:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/backend/block_meta.go
+++ b/tempodb/backend/block_meta.go
@@ -14,21 +14,22 @@ type CompactedBlockMeta struct {
 }
 
 type BlockMeta struct {
-	Version         string    `json:"format"`          // Version indicates the block format version. This includes specifics of how the indexes and data is stored
-	BlockID         uuid.UUID `json:"blockID"`         // Unique block id
-	MinID           []byte    `json:"minID"`           // Minimum object id stored in this block
-	MaxID           []byte    `json:"maxID"`           // Maximum object id stored in this block
-	TenantID        string    `json:"tenantID"`        // ID of tehant to which this block belongs
-	StartTime       time.Time `json:"startTime"`       // Roughly matches when the first obj was written to this block. Used to determine block age for different purposes (cacheing, etc)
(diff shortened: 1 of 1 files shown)


tempodb/backend/block_meta.go
```diff
@@ -14,21 +14,22 @@ type CompactedBlockMeta struct {
...
+	EndTime         time.Time     `json:"endTime"`            // Currently mostly meaningless but roughly matches to the time the last obj was written to this block
+	TotalObjects    int           `json:"totalObjects"`       // Total objects in this block
+	Size            uint64        `json:"size"`               // Total size in bytes of the data object
+	CompactionLevel uint8         `json:"compactionLevel"`    // Kind of the number of times this block has been compacted
+	Encoding        Encoding      `json:"encoding"`           // Encoding/compression format
+	IndexPageSize   uint32        `json:"indexPageSize"`      // Size of each index page in bytes
+	TotalRecords    uint32        `json:"totalRecords"`       // Total Records stored in the index file
+	DataEncoding    string        `json:"dataEncoding"`       // DataEncoding is a string provided externally, but tracked by tempodb that indicates the way the bytes are encoded
+	BloomShardCount uint16        `json:"bloomShards"`        // Number of bloom filter shards
+	MaxSpanIndexAge time.Duration `json:"max_span_index_age"` // Maximum span index age that will adjust the StartTime of the BlockMeta
```
joe-elliott, line 32: this should not be on the block meta. this needs to be a config option somewhere. I'm thinking this option should exist in the wal config? The two places where objects are added and we need to be conscious of the time range are both in the wal:

Replay:
https://github.com/grafana/tempo/blob/main/tempodb/wal/append_block.go#L99

And normal append (which you found below):
https://github.com/grafana/tempo/blob/main/tempodb/wal/append_block.go#L126

zalegrala: Ah yes, I see now on https://github.com/grafana/tempo/blob/main/tempodb/wal/append_block.go#L118

zalegrala:
Closing this to make way for others to solve this.

Issue #1303, opened 2022-02-23
Metrics-generator production readiness


mapno:
A new component `metrics-generator` has been recently added to Tempo in https://github.com/grafana/tempo/pull/1282. But it's not considered production ready yet. This issue aims to list the additional work that is needed before its official release.

The first implementation is based on design document https://github.com/grafana/tempo/pull/1206. The design document is, at the time of writing, open and we encourage everyone to take a look.

All feedback is highly appreciated. Feel free to do so in this issue, by pinging us (@kvrhdn and @mapno) at [Slack](https://github.com/grafana/tempo/pull/grafana.slack.com), or by opening PRs with changes and proposals.

## TODOs

The following list contains all tasks that we aim to complete before the component's official release. Tasks are divided by topic or area within Tempo.

### Distributor

- [ ] Include metrics-generator in distributor tests
- [ ] Improve how we stream data from distributor -> metrics-generator
  - Use a workers/channels mechanism to push traces to metrics generators instead of spawning goroutines. Discussed in https://github.com/grafana/tempo/pull/1282#discussion_r804943641
  - Design queueing and retries mechanisms

### Trace format

- [ ] Use a custom trace format (so not OTLP), but something more lightweight that contains minimal info. We should benchmark and compare performance.

### Metrics generator

- [ ] Make processors configurable per tenant and reload config at run-time (using overrides)
- [ ] Deal with inactive/stale series: if a metrics-generator instance is running for multiple days, the amount of samples it will write per request will continuously grow. We should trim this now and then by removing time series that have not been updated recently.

#### Spanmetrics processor

- [ ] Verify correctness, compare with agent / OTel processor

#### Service graphs processor

- [ ] Verify correctness, compare with agent / OTel processor
- [ ] Investigate code reusability between Tempo and the Agent.

### Remote write

- [ ] Parallelise write requests
- [ ] Make remote write requests max size configurable.
- [ ] Add limits per tenant
  - Goal: control the amount of active series we create in the downstream database.
  - Possible solutions:
    - Track the series we have sent the last ~30 minutes and drop new series if limit is exceeded.
    - Since we know we don't trim active series right now, active series will be more or less equal to the amount of samples we send per scrape. We could limit active series by setting a per tenant limit "max_series: 50000"
- [ ] Use a queue to store samples in the appender, if we add too many samples discard oldest

Operations

- [ ] Add a dedicated operational dashboard to tempo-mixin
- [ ] Add a docker-compose example

Documentation

[...]

PR #1323, opened 2022-03-04
metrics-generator: use Prometheus Agent WAL and remote storage

kvrhdn:
**What this PR does**:
Replaces our homegrown remote write implementation with the WAL and remote storage implementation from the Prometheus Agent.

We use two components from Prometheus:
- [`agent.DB`](https://github.com/prometheus/prometheus/blob/v2.34.0-rc.0/tsdb/agent/db.go#L214-L215): this is the WAL optimised for remote writing metrics only. It's a Prometheus TSDB without the querying, alerting,... capabilities. Whenever we scrape/collect metrics we append them to `agent.DB` which will store the samples on disk.
- [`remote.Storage`](https://github.com/prometheus/prometheus/blob/v2.34.0-rc.0/storage/remote/storage.go#L51-L53) (only the remote write functionality): this tails the WAL and writes data to the configured remote write endpoint(s). It has retry logic and can scale up queues as necessary. It supports a wide range of authorization options and can rewrite labels, see https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write

**Multi-tenancy**: Prometheus is not multi-tenant, to support multi-tenancy we create a WAL and remote-writer for every tenant sending data. Each WAL will be stored in `<WAL path>/<tenant ID>`.

**Resilience**: even though the WAL stores samples on disk, it does not make the metrics-generator resilient to crashes. After a restart the remote writer will start from the end of the WAL, even if older data was sent yet. See https://github.com/prometheus/prometheus/issues/8809.
But it will make the metrics-generator more resilient against an outage of the downstream TSDB. Pending samples is stored on disk so we do not lose them or risk running out of memory. This should allow the metrics-generator to overcome short outages.

Related to #1303 

**Checklist**
- [x] Tests updated
- [ ] ~~Documentation added~~ _Will be done in a later PR_


kvrhdn:
TODO
- [ ] Fix existing tests
- [ ] Add e2e test

modules/generator/generator_test.go
```diff
@@ -1,217 +0,0 @@
...
-)
-
-const localhost = "localhost"
-
-type metric struct {
-	name string
-	val  float64
-}
-
-func TestGenerator(t *testing.T) {
```
kvrhdn, line 44: I've deleted these tests in favour of the e2e tests. These tests have to do a lot of config management/mocking to get the generator running and verify the metrics emitted are correctly. This is exactly the same as what the e2e tests do, but their code is a bit simpler.

If anyone feels strong about keeping them, I don't mind reinstating them.

integration/e2e/metrics-generator/metrics_generator_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+	util "github.com/grafana/tempo/integration"
+)
+
+const (
+	configMetricsGenerator = "config-metrics-generator.yaml"
+	prometheusImage        = "prom/prometheus:latest"
+)
+
+func TestMetricsGenerator(t *testing.T) {
```
mapno, line 30: This e2e test is skipped with the current makefile targets, and the CI. We should add it there.

modules/generator/storage/instance.go
```diff
@@ -0,0 +1,111 @@
...
+
+	walDir := filepath.Join(cfg.Path, tenant)
+
+	level.Info(logger).Log("msg", "creating WAL", "dir", walDir)
+
+	// Create WAL directory with necessary permissions
+	err := os.MkdirAll(walDir, 0o755)
+	if err != nil {
+		level.Warn(logger).Log("msg", "could not create directory for metrics WAL", "err", err)
+	}
```
mapno, line 51: Isn't this redundant with lines 54-57 creating the same directory + `/wal`?

mapno (commented):
Nice! Love this changes. Left just two nits. LGTM

integration/e2e/metrics-generator/metrics_generator_test.go
kvrhdn: Doh 🤦🏻

modules/generator/storage/instance.go
kvrhdn: Yes, that's correct I think. I started out with the first `MkdirAll` and added the second one later to avoid the error from remote storage trying to read `<walDir>/<tenant>/wal`.

MkdirAll says:

> _The permission bits perm (before umask) are used for all directories that MkdirAll creates._

So the second `MkdirAll` should also create the tenant directory with the right permission bits.

mapno (approved):

kvrhdn:
Hmm, `TestInstance_multiTenancy` seems to be flaky in CI. I'll see if I can make it more robust.

PR #1329, opened 2022-03-08
Update a few error messages and doc strings for readability

zalegrala:
**What this PR does**:

Here we update some error strings to improve readability in determining from a single error string when one error stops and when another begins.  Also a correction to a docstring.


joe-elliott (approved):
oof good catch on us dropping the error on those FindTraceByID calls

zalegrala:
:+1:   Debugging yesterday left me a little confused on the reason when I saw that error.

PR #1332, opened 2022-03-08
TempoDB: Restrict Ingestion time range

joe-elliott:
**What this PR does**:
Adds a new config option to restrict the time range around now that Tempo will honor when updating the current block start and end times. The new option defaults to `2m` meaning that if a span is received outside of this time range it will be stored and findable by trace id, but the block's time range it is stored in will not reflect this. This will make it less likely to be found using a trace search query.

One gross corner is that during WAL replay we have to assume that we should adjust the block's time range based on the maximum time that a WAL file will be kept since we don't know when it was created.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/ingester/ingester.go
+++ b/modules/ingester/ingester.go
@@ -342,7 +342,7 @@ func (i *Ingester) replayWal() error {
 			return 0, 0, err
 		}
 		return start, end, nil
-	}, log.Logger)
+	}, i.cfg.MaxBlockDuration, log.Logger)
 	if err != nil {
 		return fmt.Errorf("fatal error replaying wal %w", err)
 	}
--- a/modules/ingester/ingester.go
+++ b/modules/ingester/ingester.go
@@ -327,6 +327,10 @@ func (i *Ingester) TransferOut(ctx context.Context) error {
 func (i *Ingester) replayWal() error {
 	level.Info(log.Logger).Log("msg", "beginning wal replay")
 
+	// pass i.cfg.MaxBlockDuration into RescanBlocks to make an attempt to set the start time
+	// of the blocks correctly. as we are scanning traces in the blocks we read their start/end times
+	// and attempt to set start/end times appropriately. we use now - max_block_duration - ingestion_slack
+	// as the minimum acceptable start time for a replayed block.
 	blocks, err := i.store.WAL().RescanBlocks(func(b []byte, dataEncoding string) (uint32, uint32, error) {
 		d, err := model.NewObjectDecoder(dataEncoding)
 		if err != nil {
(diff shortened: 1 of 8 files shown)


zalegrala (approved):
Nice, this looks reasonable to me.  The 2 minute time window seems a little small to me as a default, but configuration is there to be adjusted.

joe-elliott:
>The 2 minute time window seems a little small to me as a default, but configuration is there to be adjusted.

I'm concerned about a wide time window b/c the larger the window the more unnecessary data you're pulling in during search. Also, if something is just outside the window (say 30s) it will likely still be picked up during a search b/c the whole block will be scanned.

tempodb/wal/append_block.go
```diff
@@ -238,3 +243,20 @@ func (a *AppendBlock) file() (*os.File, error) {
...
 }
+
+func (a *AppendBlock) adjustTimeRangeForSlack(start uint32, end uint32, additionalStartSlack time.Duration) (uint32, uint32) {
+	now := time.Now()
+	startOfRange := uint32(now.Add(-a.ingestionSlack).Add(-additionalStartSlack).Unix())
+	endOfRange := uint32(now.Add(a.ingestionSlack).Unix())
+
+	if start < startOfRange {
+		metricWarnings.WithLabelValues(a.meta.TenantID, reasonOutsideIngestionSlack).Inc()
+		start = uint32(now.Unix())
```
mdisibio, line 254: Should these clip to the range instead of now?

tempodb/wal/append_block.go
```diff
@@ -238,3 +243,20 @@ func (a *AppendBlock) file() (*os.File, error) {
...
+	}
+	if end > endOfRange {
+		metricWarnings.WithLabelValues(a.meta.TenantID, reasonOutsideIngestionSlack).Inc()
```
mdisibio, line 257: Not necessarily a problem, but noting that a single trace can trigger both start and end warnings.  Probably worth rearranging this logic to fix.

tempodb/wal/append_block.go
joe-elliott: I originally did that and moved away from it. Now that I'm thinking through it again I am in agreement.

modules/ingester/ingester.go
```diff
@@ -342,7 +342,7 @@ func (i *Ingester) replayWal() error {
 			return 0, 0, err
 		}
 		return start, end, nil
-	}, log.Logger)
+	}, i.cfg.MaxBlockDuration, log.Logger)
```
mdisibio, line 345: Yeah agree this is a hard spot but understand the reasons, and this is reasonable.  Maybe a comment ?

mdisibio (approved):

tempodb/wal/append_block.go
joe-elliott: Oh, so the reason I decided to do use `time.Now()` is b/c we know we are regularly receiving backdated spans from multiple weeks ago. If we use `start` instead of `time.Now()` then every block's time range will always start `ingestion_time_range_slack` earlier than it actually does. Also, every replayed block's start time will always be replay time - `max_block_duration` - `ingestion_time_range_slack`.

PR #1335, opened 2022-03-09
Make Compaction Cycle configurable

### LINKED ISSUE #1232: Make `compactionCycle` configurable

As of now, the `compactionCycle` defaults to 500ms. https://github.com/grafana/tempo/blob/2674b233d8663852f2ccc65cc47edfee36caba23/tempodb/compactor.go#L60

This gets tough to work with on local development setups where logs are spammed with compactor cycle logs
```
tempo_1                     | level=info ts=2022-01-19T10:37:14.0120974Z caller=compactor.go:98 msg="starting compaction cycle" tenantID=single-tenant offset=0
tempo_1                     | level=info ts=2022-01-19T10:37:14.0122882Z caller=compactor.go:104 msg="compaction cycle complete. No more blocks to compact" tenantID=single-tenant
tempo_1                     | level=info ts=2022-01-19T10:37:14.5105278Z caller=compactor.go:98 msg="starting compaction cycle" tenantID=single-tenant offset=0
tempo_1                     | level=info ts=2022-01-19T10:37:14.5108502Z caller=compactor.go:104 msg="compaction cycle complete. No more blocks to compact" tenantID=single-tenant
tempo_1                     | level=info ts=2022-01-19T10:37:15.0127511Z caller=compactor.go:98 msg="starting compaction cycle" tenantID=single-tenant offset=0
tempo_1                     | level=info ts=2022-01-19T10:37:15.0129207Z caller=compactor.go:104 msg="compaction cycle complete. No more blocks to compact" tenantID=single-tenant
tempo_1                     | level=info ts=2022-01-19T10:37:15.510125Z caller=compactor.go:98 msg="starting compaction cycle" tenantID=single-tenant offset=0
tempo_1                     | level=info ts=2022-01-19T10:37:15.5102975Z caller=compactor.go:104 msg="compaction cycle complete. No more blocks to compact" tenantID=single-tenant

```

Make the compactionCycle configurable (but continue to default to 500ms) so we can configure that on local dev setups.

N/A

**Additional context**

N/A

willdot:
**What this PR does**:
This PR allows for the Compaction Cycle to be configurable, setting the default value to be 30 seconds (it was this originally)

Fixes #1232 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


willdot:
Unsure how to write a test for this, since it's testing a time config without writing some crazy spy logic. However I've tested this locally by configuring it, running it using the docker compose example, and using good ole logging to check the value that gets set from config.

tempodb/compactor.go
```diff
@@ -66,6 +66,11 @@ const (
 
 // todo: pass a context/chan in to cancel this cleanly
 func (rw *readerWriter) compactionLoop() {
+	compactionCycle := DefaultCompactionCycle
```
zalegrala, line 69: We may set the default value in `RegisterFlagsAndApplyDefaults`, and then we can use the value here.

zalegrala (commented):
Thanks for the PR.  This seems reasonable to me.  Just one comment about where we set the default.

tempodb/compactor.go
willdot: I'm not sure I follow what you mean, but isn't the default  already being set in `RegisterFlagsAndApplyDefaults` ?

https://github.com/willdot/tempo/blob/6b83318b934f4c961fc5e962ac448e3140202254/modules/compactor/config.go#L30

zalegrala: Oh you're right, I read to fast.

zalegrala: Could we reference the `rw.compactorCfg.CompactionCycle` below in place of the local?  Does a `0` compaction cycle mean something, or I suppose in that case a config value of `0` would be actually `30`, which could be confusing.

willdot: If 0 is passed into a new ticker, you get a panic `non-positive interval for NewTicker`

Which is why I use the default if the config is set to 0. 

Thinking about it, this either means it's not been configured (valid) or it's been configured as 0 (also valid). Is there somewhere we can document to users that it can't be 0 and that the default of 30 seconds will be used if it is?

To make the code clearer I could either drop a comment, or extract it out into a helper function, something like 

`getCompactionCycleFromConfigOrDefault(rw.compactorCfg.CompactionCycle, DefaultCompactionCycle)`

annanay25: I'd vote for using the default of 30 seconds even if the user manually configures a value of 0 (because that doesn't make sense). We can make a note of it in our configuration docs, which are over here:
https://github.com/grafana/tempo/blob/main/docs/tempo/website/configuration/_index.md#compactor

annanay25 (approved):
LGTM except for the documentation updates to the compactor configuration section.

tempodb/compactor.go
willdot: Added something to the configuration docs.

annanay25:
@willdot we made some changes to the e2e tests in main, can you please rebase this branch? Thanks!

willdot:
> @willdot we made some changes to the e2e tests in main, can you please rebase this branch? Thanks!

Done.

PR #1337, opened 2022-03-09
metrics-generator: verify WAL is writeable

kvrhdn:
**What this PR does**:
When creating a metrics-generator `storage.Instance`, verify we can write to the given path before creating any resources. If we create Prometheus' `remote.Storage` twice, it will panic as metrics are registered twice.

~~Fixes #<issue number>~~

**Checklist**
- [x] Tests updated
- [ ] ~~Documentation added~~

### DIFF
--- a/modules/generator/storage/instance.go
+++ b/modules/generator/storage/instance.go
@@ -50,7 +50,6 @@ func New(cfg *Config, tenant string, reg prometheus.Registerer, logger log.Logge
 	// subdirectory remote storage logs a scary error.
 	err := os.MkdirAll(filepath.Join(walDir, "wal"), 0o755)
 	if err != nil {
-		level.Error(logger).Log("msg", "could not create directory for metrics WAL", "err", err)
 		return nil, fmt.Errorf("could not create directory for metrics WAL: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


modules/generator/storage/instance.go
```diff
@@ -49,7 +50,8 @@ func New(cfg *Config, tenant string, reg prometheus.Registerer, logger log.Logge
 	// subdirectory remote storage logs a scary error.
 	err := os.MkdirAll(filepath.Join(walDir, "wal"), 0o755)
 	if err != nil {
-		level.Warn(logger).Log("msg", "could not create directory for metrics WAL", "err", err)
+		level.Error(logger).Log("msg", "could not create directory for metrics WAL", "err", err)
```
joe-elliott, line 53: i'm fine with this if you want it, but does it make sense to log and return the error? or should we just rely on the caller to log the error?

kvrhdn: Yeah that's a good point. This error will get returned by the GRPC call so this error should show in the logs of the distributor and the metrics-generator. It will be a bit harder to parse as it gets wrapped in some more stuff but this shouldn't be an issue if you read it properly.

PR #1338, opened 2022-03-09
Serverless aws deploy

zalegrala:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


.drone/drone.yml
```diff
@@ -7,63 +7,75 @@ platform:
...
+      build_args:
+        - TARGETARCH=amd64
+      dockerfile: cmd/tempo-query/Dockerfile
+      password:
+        from_secret: docker_password
+      repo: grafana/tempo-query
+      username:
+        from_secret: docker_username
+
+  - name: build-tempo-serverless-lambda
```
zalegrala, line 61: Here lies the addition for `amd64`.  Perhaps we want `arm` also, but given that we're cloud, not sure it is needed.

.drone/drone.yml
```diff
@@ -233,30 +246,36 @@ platform:
...
+    image: google/cloud-sdk
+    commands:
+      - cd ./cmd/tempo-serverless/cloud-functions
+      - printf "%s" "$ops_tools_fn_upload" > ./creds.json && gcloud auth activate-service-account --key-file ./creds.json && gsutil cp tempo-serverless*.zip gs://ops-tools-tempo-function-source
+      - printf "%s" "$ops_tools_fn_upload" > ./creds.json && gcloud auth activate-service-account --key-file ./creds.json && gsutil cp tempo-serverless*.zip gs://grafanalabs-global-tempo-function-source
+    environment:
+      ops_tools_fn_upload:
+        from_secret: ops_tools_fn_upload
+
+  - name: deploy-tempo-serverless-aws
```
zalegrala, line 267: Breadcumbs for deploying to s3.  Will need to work out credentials.

cmd/tempo-serverless/lambda/Dockerfile
```diff
@@ -34,12 +34,11 @@ COPY --from=build /lambda-local-proxy /lambda-local-proxy
 RUN chmod +x /lambda-local-proxy
 
 RUN echo -e '\
-#!/bin/bash \n\
-AWS_REGION=dummy AWS_ACCESS_KEY_ID=dummy AWS_SECRET_ACCESS_KEY=dummy /lambda-local-proxy -e http://localhost:8080 -p 9000 -f function & \n\
-exec /lambda-entrypoint.sh lambda \n\
-' > /run.sh
+  #!/bin/bash \n\
```
zalegrala, line 37: This diverges somewhat from the golang example here: https://docs.aws.amazon.com/lambda/latest/dg/go-image.html

But the local proxy could be useful in e2e testing and local development.

cmd/tempo-serverless/lambda/Dockerfile.proxy
```diff
@@ -0,0 +1,44 @@
+#
```
joe-elliott, line 1: if this is intended to be a proxy image only we could clean it up quite a bit. there's no need to put the lambda binary in here.

also, i'm not seeing changes to the e2e tests?

cmd/tempo-serverless/lambda/Dockerfile
```diff
@@ -1,45 +1,5 @@
...
-AWS_REGION=dummy AWS_ACCESS_KEY_ID=dummy AWS_SECRET_ACCESS_KEY=dummy /lambda-local-proxy -e http://localhost:8080 -p 9000 -f function & \n\
-exec /lambda-entrypoint.sh lambda \n\
-' > /run.sh
-RUN chmod +x /run.sh
-
-# run lambda-local-proxy and the actual lambda runtime itself.
-ENTRYPOINT /run.sh
-           
\ No newline at end of file
+FROM alpine:3.9 as certs
```
joe-elliott, line 1: does this work? doesn't a custom docker image require specific base images? or some kind of lambda runtime?

zalegrala: I think this will come up in the testing.  I believe you might be right though if we want to test outside of AWS, we need to have the proper environment variables in place.

cmd/tempo-serverless/lambda/Dockerfile.proxy
zalegrala: I'll be looking at cleaning up the proxy file, and maybe relocating it.  I'm thinking we want to produce a proxy docker image for the e2e environment.

cmd/tempo-serverless/Makefile
```diff
@@ -25,18 +25,6 @@ build-gcf-zip:
 	$(IN_CLOUD_FUNCTIONS) zip tempo-serverless-$(VERSION).zip ./* -r
 	$(IN_CLOUD_FUNCTIONS) rm -rf vendor
 
-#
```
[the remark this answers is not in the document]
zalegrala: The docker build for lambda has been moved to the root Makefile, but I was leaving `build-lambda-zip` in place while the #1347 is in flight.

joe-elliott:
Also note that you'll need to update the docs merged here:
https://github.com/grafana/tempo/pull/1347

zalegrala:
I prefer https://github.com/grafana/tempo/pull/1357 for the e2e parts, and for deploy, I'll follow up with the s3 approach.  Building out the docker image as I attempted here doesn't gain us anything over the s3 approach in that it won't be able to provide an image for the community to use, since users will need to copy this to their own ecr repo, which has slightly more complexity than the bucket + access approach we can take with s3.

PR #1339, opened 2022-03-10
metrics-generator: consistently pass log.Logger around

kvrhdn:
**What this PR does**:
Small refactor to pass `log.Logger` to every component that logs something instead of relying on the global logger variable. This makes it easier to add common labels to logs and allows us to control logging better in tests.


annanay25 (approved):
LGTM.

PR #1340, opened 2022-03-10
metrics-generator: add custom registry

kvrhdn:
**What this PR does**:
This introduces a custom registry to store metrics generated in the metrics-generator. The registry can scrape itself and enforce limits.

So far the metrics-generator has relied on the `prometheus.Registry` from [prometheus/client_golang](https://github.com/prometheus/client_golang). This has a number of issues:
- it is not possible to remove series: if the instance is running for a long time, it will keep resending metrics that haven't been updated in a long time
  - -> this increases the amount of active series in the downstream TSDB and has little additional value
- it is not possible to enforce a maximum amount of series per tenant
   - -> this can be useful to protect Tempo and the downstream TSDB against a cardinality spike from the ingested traces

This PR introduces a new `Registry` which kind of mirrors the `prometheus.Registerer`. It supports counters and histograms. There are two implementations:
- `ManagedRegistry`: a registry that can scrape itself and write data into an `storage.Appender`. It will also enforce limits and remove stale series.
- `TestRegistry`: a simple implementation to verify the correctness of processors using `Registry`.

The config will look like this:

```yaml
metrics_generator:
    registry:
        collection_interval: 15s
        stale_duration: 30m
        external_labels:
          - source: tempo
```

The overrides will look like this:

```yaml
metrics_generator_registry_scrape_interval: 1m
metrics_generator_registry_max_active_series: 10000
```

To see the `ManagedRegistry` in practice: this is the amount of active series when we are not removing stale series. As new data shows up, the amount of metrics series that are being tracked constantly increases, until the instance is restarted.

Screenshot 2022-03-10 at 13 50 05

With the `ManagedRegistry` stale series are removed after 15 minutes, resulting in a stable system. This will ensure the amount of active series in the downstream TSDB remains fairly constant over time.

Screenshot 2022-03-10 at 13 49 11

Related to #1303 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -50,26 +62,30 @@ type processor struct {
...
-	serviceGraphRequestServerHistogram *prometheus.HistogramVec
-	serviceGraphRequestClientHistogram *prometheus.HistogramVec
-	serviceGraphUnpairedSpansTotal     *prometheus.CounterVec
-	serviceGraphDroppedSpansTotal      *prometheus.CounterVec
+	serviceGraphRequestTotal                  registry.Counter
+	serviceGraphRequestFailedTotal            registry.Counter
+	serviceGraphRequestServerSecondsHistogram registry.Histogram
+	serviceGraphRequestClientSecondsHistogram registry.Histogram
+	//serviceGraphUnpairedSpansTotal            registry.Counter
+	//serviceGraphDroppedSpansTotal             registry.Counter
```
kvrhdn, line 70: These metrics weren't updated before, so I don't know if we should keep them around. I'll check if Grafana actually uses these in the service graphs view.

We already expose these metrics as metrics from Tempo:
 - `metrics_generator_processor_service_graphs_dropped_spans`
 - `metrics_generator_processor_service_graphs_unpaired_edges`

So a Tempo operator can see these stats already.

kvrhdn:
I think I made a design mistake implementing this: looking at some profiling data this implementation spends a lot of time in `labels.Builder` allocating and growing slices.

Issues:
- we are doing a lot more work to update metrics than to scrape them, even though updating is much more common (it's correlated to spans ingested/sec) while scraping is once every 15s or 1m.
  - updating metrics should be as cheap as possible and allocate as little as possible
- `labels.Labels` is inefficient since we often grow slices as we add additional labels (like `__name__`) and then have to sort them again
  - I think we should avoid using `labels.Labels` until necessary (i.e. when appending data)

joe-elliott (commented):
This generally looks fine to me. As discussed offline there are a few performance improvements to consider:

- Precalculate the hash for a set of label values
- Don't bother incrementing 0 on histogram buckets
- Consider where pointers could be dropped in favor of label values.

modules/generator/processor/servicegraphs/servicegraphs.go
joe-elliott: We probably need to have these per tenant so that individual tenants can have a feeling for the quality of their metrics. Alternatively, instead of exposing these directly, it might be better to use the new `tempo_warnings_total` metric with a specific `reason` label for these situations.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -107,20 +74,15 @@ func (p *processor) aggregateMetricsForSpan(svcName string, span *v1_trace.Span)
...
-		// Build additional dimensions
-		for _, d := range p.cfg.Dimensions {
-			for _, attr := range span.Attributes {
-				if d == attr.Key {
-					labelValues = append(labelValues, attr.GetValue().GetStringValue())
-				}
+	for _, d := range p.cfg.Dimensions {
+		for _, attr := range span.Attributes {
+			if d == attr.Key {
+				labelValues = append(labelValues, attr.GetValue().GetStringValue())
```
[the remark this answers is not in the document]
kvrhdn: Right, good catch :+1: Fixed this now

kvrhdn:
Switched design a bit and this can now do 40MB/s or 200k spans/sec per instance, which is on par with performance before this PR 🎉 

- introduced a `LabelValues` struct that can cache the hash of label values. This means we only calculate the hash once, even if 2-4 metrics are updated.
- split up `metric` into a dedicated `counter` and `histogram` type: this results in some duplicated code but each type is optimised for its own kind of series. So for histograms this mean we only manage 1 series for the count, sum and buckets metrics.

Some stuff I tried that didn't work out:

- changing `map[uint64]*serie` by `map[uint64]serie` lowers memory usage a tad (~100 MB), but makes CPU and P99 latency spiky. This is kind of expected as we need to lock the entire map to update a serie:

```go
	hash := labelValues.getHash()

	m.seriesMtx.Lock()
	defer m.seriesMtx.Unlock()

	s, ok := m.series[hash]
	if ok {
		s.value += value
		s.lastUpdated = time.Now().UnixMilli()
		m.series[hash] = s
		return
	}

	if !m.onAddSerie() {
		return
	}

	labelValuesCopy := make([]string, len(labelValues.values))
	copy(labelValuesCopy, labelValues.values)

	m.series[hash] = serie{
		labelValues: labelValuesCopy,
		value:       value,
		lastUpdated: time.Now().UnixMilli(),
	}
```
  by using `*serie` we only need to lock the map to look up the serie, all other updates to the serie can happen concurrently

modules/generator/registry/histogram.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	for _, s := range h.series {
+		// set series-specific labels
+		for i, name := range h.labels {
+			lb.Set(name, s.labelValues[i])
+		}
+
+		// sum
+		lb.Set("__name__", h.nameSum)
```
mdisibio, line 149: *nit*: This constant is exposed as `labels.MetricName`, since we are already using that for `le`.

kvrhdn: Done!

modules/generator/registry/histogram.go
```diff
@@ -0,0 +1,206 @@
...
+		} else {
+			newSerie.buckets = append(newSerie.buckets, atomic.NewFloat64(0))
+		}
+	}
+
+	h.seriesMtx.Lock()
+	defer h.seriesMtx.Unlock()
+
+	s, ok = h.series[hash]
+	if ok {
```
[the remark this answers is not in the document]
kvrhdn: Yep, it's probably fairly unlikely but another goroutine might have added the new series in the time between releasing the read lock and grabbing the write lock.

kvrhdn:
Trying to get GitHub Actions to run...

modules/generator/registry/counter.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func (c *counter) setCallbacks(onAddSeries func(uint32) bool, onRemoveSeries func(count uint32)) {
+	c.onAddSeries = onAddSeries
+	c.onRemoveSeries = onRemoveSeries
+}
+
+func (c *counter) Inc(labelValues *LabelValues, value float64) {
+	if value < 0 {
+		panic("counter can only increase")
```
zalegrala, line 50: Should we panic here?  Would an error log and early return be sufficient?

mdisibio: I think it's ok, the Prometheus client [does the same](https://github.com/prometheus/client_golang/blob/main/prometheus/counter.go#L107), although the method is more general `Add` vs `Inc` which takes no params.

kvrhdn: Yeah, we could handle this a bit more elegantly, but I didn't want to bother adding a `log.Logger` field to the counter or returning an error just in case someone misuses the counter.
A panic isn't nice, but it should ensure whoever tries to decrease the counter detects it before it ships.

Prometheus client_golang also just panics btw: https://github.com/prometheus/client_golang/blob/main/prometheus/counter.go#L108-L110

modules/generator/registry/counter.go
```diff
@@ -0,0 +1,140 @@
...
+
+type counter struct {
+	name   string
+	labels []string
+
+	// seriesMtx is used to sync modifications to the map, not to the data in series
+	seriesMtx sync.RWMutex
+	series    map[uint64]*counterSeries
+
+	onAddSeries    func(count uint32) bool
```
mapno, line 21: I think calling these callbacks can be a bit confusing since they intervine in the creation of the metrics (i.e. the result of `onAddSeries` will determine if the series is created or not).

modules/generator/registry/histogram.go
```diff
@@ -0,0 +1,207 @@
...
+	h.seriesMtx.RLock()
+	s, ok := h.series[hash]
+	h.seriesMtx.RUnlock()
+
+	if ok {
+		h.updateSeries(s, value)
+		return
+	}
+
+	if !h.onAddSerie(h.activeSeriesPerHistogramSerie()) {
```
mapno, line 81: Maybe a nil check here?

modules/generator/registry/histogram.go
```diff
@@ -0,0 +1,207 @@
...
+		series:       make(map[uint64]*histogramSeries),
+	}
+}
+
+func (h *histogram) setCallbacks(onAddSeries func(count uint32) bool, onRemoveSeries func(count uint32)) {
+	h.onAddSerie = onAddSeries
+	h.onRemoveSerie = onRemoveSeries
+}
+
+func (h *histogram) Observe(labelValues *LabelValues, value float64) {
```
mapno, line 65: Should it check that the value is not negative too?

modules/generator/registry/counter.go
```diff
@@ -0,0 +1,140 @@
...
+
+func newCounter(name string, labels []string) *counter {
+	return &counter{
+		name:   name,
+		labels: labels,
+		series: make(map[uint64]*counterSeries),
+	}
+}
+
+func (c *counter) setCallbacks(onAddSeries func(uint32) bool, onRemoveSeries func(count uint32)) {
```
mapno, line 43: Any reason not to pass these functions in the constructor? I didn't see these fields changed other than when the metric is created.

mapno (commented):
Overall LGTM, great improvements. Limiting active series and pruning inactive ones will be very useful. Left a few comments.

modules/generator/registry/counter.go
kvrhdn: I think I did this so I could make `setCallbacks` part of the `metric` interface. It doesn't save much code though and setting it in the `newCounter`/`newHistogram` function is just as easy.

modules/generator/registry/histogram.go
kvrhdn: I do a nil check in `newCounter`/`newHistogram` now.

modules/generator/registry/histogram.go
kvrhdn: I'm using Prometheus client_golang library as reference and they seem to allow negative values: https://github.com/prometheus/client_golang/blob/main/prometheus/histogram.go#L307-L309

I've never seen a histogram with negative values though, so I'm not sure if it works as intended 🤷🏻

modules/generator/registry/counter.go
kvrhdn: I agree these things are confusing. I don't like we have to use anonymous functions to couple the counters/histograms to the registry, but I'm not sure what would work better. I wanted to keep this functions somewhat generic so it's easier to test. Injecting the full `managedRegistry` means we have to test them together all the time.

Part of the complexity is that `onAddSeries` is used for both checking whether we can add series and to keep count of them. We could split this up into 2 callbacks: `canAddSeries` and `incActiveSeries` but this is 1) more work and 2) might introduce a race condition between asking whether you can add series and reporting you added them.

modules/generator/processor/servicegraphs/servicegraphs.go
kvrhdn: We currently keep track of dropped spans and unpaired edges per tenant. So we already collect the data, I think it will just be a matter of exposing them to the end-user.
`tempo_warnings_total` looks interesting, I'll take a look at what warnings we want to share from the metrics-generator in general.

mdisibio (approved):
Nice work!

PR #1341, opened 2022-03-10
e2e: reduce flakiness

kvrhdn:
**What this PR does**:

Changes:

- Always check `tempo_ingester_traces_created_total` is _at least_ 1.  Some tests expect exactly 1 trace was created, others just expected more than 0. Let's always check traces created is >0. I saw tests fail because 2 traces were created while 1 was expected.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/integration/e2e/backend/backend.go
+++ b/integration/e2e/backend/backend.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/grafana/e2e"
 	e2e_db "github.com/grafana/e2e/db"
+
 	"github.com/grafana/tempo/cmd/tempo/app"
 	util "github.com/grafana/tempo/integration"
 	"github.com/grafana/tempo/tempodb/backend/azure"
@@ -15,7 +16,7 @@ import (
 
 const (
 	azuriteImage = "mcr.microsoft.com/azure-storage/azurite"
-	gcsImage     = "fsouza/fake-gcs-server"
+	gcsImage     = "fsouza/fake-gcs-server:1.36.3"
 )
 
 func parsePort(endpoint string) (int, error) {
--- a/integration/e2e/config-microservices.tmpl.yaml
+++ b/integration/e2e/config-microservices.tmpl.yaml
@@ -15,7 +15,7 @@ ingester:
       kvstore: {{ .KVStore }}
       replication_factor: 3
     heartbeat_period: 100ms
-  trace_idle_period: 1s
+  trace_idle_period: 2s
   max_block_duration: 2s
   complete_block_timeout: 5s
   flush_check_period: 1s
--- a/integration/e2e/config-microservices.tmpl.yaml
+++ b/integration/e2e/config-microservices.tmpl.yaml
@@ -23,6 +23,7 @@ ingester:
 storage:
   trace:
     blocklist_poll: 2s
+    blocklist_poll_stale_tenant_index: 15s
     backend: s3
     s3:
       bucket: tempo
--- a/integration/e2e/config-microservices.tmpl.yaml
+++ b/integration/e2e/config-microservices.tmpl.yaml
@@ -23,7 +23,7 @@ ingester:
 storage:
   trace:
     blocklist_poll: 2s
-    blocklist_poll_stale_tenant_index: 15s
+    blocklist_poll_stale_tenant_index: 1s # force components to always fall back to polling
     backend: s3
     s3:
       bucket: tempo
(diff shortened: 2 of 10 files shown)


integration/e2e/e2e_test.go
```diff
@@ -92,7 +93,7 @@ func TestAllInOne(t *testing.T) {
 			assertEcho(t, "http://"+tempo.Endpoint(3200)+"/api/echo")
 
 			// ensure trace is created in ingester (trace_idle_time has passed)
-			require.NoError(t, tempo.WaitSumMetrics(e2e.Equals(1), "tempo_ingester_traces_created_total"))
+			require.NoError(t, tempo.WaitSumMetrics(e2e.Greater(0), "tempo_ingester_traces_created_total"))
```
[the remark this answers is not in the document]
annanay25: I think what we should do instead is increase the `trace_idle_period` in all tests to atleast 1s.
https://github.com/grafana/tempo/blob/15bf6ecdecf143fa38e075073202f65758d8b43a/integration/e2e/config-all-in-one-s3.yaml#L27

That way all spans for the same trace will reach Tempo within a single window and we won't create 2 traces.

integration/e2e/backend/backend.go
```diff
@@ -8,14 +8,15 @@ import (
...
+
 	"github.com/grafana/tempo/cmd/tempo/app"
 	util "github.com/grafana/tempo/integration"
 	"github.com/grafana/tempo/tempodb/backend/azure"
 )
 
 const (
 	azuriteImage = "mcr.microsoft.com/azure-storage/azurite"
-	gcsImage     = "fsouza/fake-gcs-server"
+	gcsImage     = "fsouza/fake-gcs-server:1.36.3"
```
annanay25, line 19: Why are we pinning the image? If there's a compelling reason should we pin azurite as well?

integration/e2e/config-microservices.tmpl.yaml
```diff
@@ -23,6 +23,7 @@ ingester:
 storage:
   trace:
     blocklist_poll: 2s
+    blocklist_poll_stale_tenant_index: 15s
```
annanay25, line 26: What kind of errors does this setting resolve?

kvrhdn: Since there are no compactors, no component is responsible for writing the tenant index. So the tenant index only gets created once when fetching the tenant index fails (the component falls back to polling and writes the index). If the component making the tenant index does not have a complete view of the blocklist, this tenant index will be wrong and never updated.
Setting the stale duration ensures the query-frontend and querier reject the tenant index after a while.

An alternative solution would be to add a compactor to the system but this might introduce other race conditions as the blocklist length will change due to compactions.

I've added some logs in the description: https://github.com/grafana/tempo/pull/1341/commits/990db134e8275fe2c14da3c20ad15e7905815e01

integration/e2e/backend/backend.go
kvrhdn: When I use the latest image (1.37) the test fails very consistently (even locally). [This version was released yesterday](https://github.com/fsouza/fake-gcs-server/releases/tag/v1.37.0) and I think there might be a bug in some of the calls we use to list the bucket.

I've added some logs here (https://github.com/grafana/tempo/pull/1341/commits/37843d678bb9ff98cf4d943e8b8e1d59b888fe66) that show the block is flushed correctly but the compactor never sees it.

kvrhdn: Regarding pinning the azurite image: I don't feel particular strong about either option. These apis don't change much, so it should be fairly uncommon that an update to the image breaks the tests (though it just happened of course 🙈 )

integration/e2e/e2e_test.go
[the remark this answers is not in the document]
kvrhdn: Good idea! I've increased the `trace_idle_period` for all tests that check `tempo_ingester_traces_created_total`.

integration/e2e/backend/backend.go
annanay25: Haha, alright. In that case its ok to pin the gcs one since its breaking I guess. We'll pin the azurite one in the future if needed.

Side note: Is it concerning that the gcs tests are breaking? I don't think so because the google cloud APIs are stable, but just checking.

integration/e2e/config-microservices.tmpl.yaml
annanay25: Hmm, this is fine. If there's an option to disable tenant index creation (to just enforce blocklist polling always) then I would prefer that in the e2e tests.

annanay25 (approved):
No more _blocking_ comments so approving, lgtm.

integration/e2e/config-microservices.tmpl.yaml
kvrhdn: Don't think you can disable it, but setting stale at 1s forces components to always fall back to polling.

PR #1342, opened 2022-03-11
Updated this page to refer to 'monolithic' instead of 'single binary'

09jvilla:
We've standardized on this as the terminology we want to use for all of our backends, and have already made the changes in [Loki](https://grafana.com/docs/loki/next/fundamentals/architecture/deployment-modes/) and Mimir. See related slack [thread](https://raintank-corp.slack.com/archives/C02RJ13QB4K/p1645818793120839).

The rationale for moving away from 'single binary' is that Tempo (and Loki, and Mimir) are always a 'single binary', whether or not you are deploying as microservices.(In a microservices deployment, each component is the same 'single binary', but it just has a different value of the target flag.)

### DIFF
--- a/docs/tempo/website/api_docs/_index.md
+++ b/docs/tempo/website/api_docs/_index.md
@@ -8,7 +8,7 @@ weight: 350
 Tempo exposes an API for pushing and querying traces, and operating the cluster itself.
 
 For the sake of clarity, in this document we have grouped API endpoints by service, but keep in mind that they're exposed both when running Tempo in microservices and monolithic mode:
-- **Microservices**: each service exposes its own endpoints
+- **microservices**: each service exposes its own endpoints
 - **monolithic**: the Tempo process exposes all API endpoints for the services running internally
 
 ## Endpoints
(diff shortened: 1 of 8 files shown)


joe-elliott (commented):
The term "single binary" occurs in a number of other places in the docs. should we go ahead and clean those up as well?

09jvilla:
Good point @joe-elliott . I took another crack at it. The helm charts, the target=scalable-single-binary, and a few URLs in the docs will still mention 'single binary', but I cleaned up the docs as much as possible and made a note that there may still be some legacy references.

joe-elliott:
@09jvilla I was considering updating all of our examples (helm,tk,docker-compose) to use the new language but I was concerned that if I started changing the paths I would break some links. Instead I've added some notes here:

https://github.com/grafana/tempo/pull/1342/commits/dd33a6a42c6f9160e2ed78c1f6710d30726af1d2

Does this work? Should we update all of the examples terminology but leave the paths?

09jvilla:
> Does this work? 

Yea that would be my suggestion rather than the **Note** approach you took. I'm scared to change the name of the stuff pointed to by specific commands, like `single-binary-grafana-values.yaml` and the `tk apply tempo-single-binary` -- not sure if this will break folks, but updating the content of the readme itself seems fine.

09jvilla:
Feel free to just merge your commits on top of my PR.

joe-elliott (approved):

PR #1346, opened 2022-03-14
Refactor for alternative block formats

mdisibio:
**What this PR does**:
This PR has a variety of refactors to support the goal of non-byte-slice (i.e. columnar) block formats.  Overview of changes:
* BackendBlock is oriented around new `FindByTraceID` and `Search` which use `tempopb` data types (`Trace` and `SearchRequest/SearchResponse`.  These methods internalize all the work of searching the block, and callers were all updated (querier, ingester, serverless, tempodb, cli).  Offline we discussed this is as good interface for blocks in general, going forward.
* Some of the overlap between `pkg/model`, `pkg/model/trace`, and `tempodb/search` was unwound up by moving `ObjectCombiner` and other search constants to `pkg`.  The goal is to have `tempodb` use `pkg` and not the other way around.
* For the above `FindTraceByID` and `Search` methods, new interfaces are introduced `Finder` and `Searcher`. These aren't super necessary in this PR, but will make more sense in future refactoring as we cover more block types or tweak the `readerWriter` interface.

Future refactoring:  There is more refactoring needed but this PR is already a good size, it seems better to visit separately.  TODOs:
* Compaction - What does compaction look like? TBD
* BackendBlock still exposes some things like `Iterate` which returns byte slices. This is only used by compaction.
* Append/Streaming blocks - Need similar changes for non-byte slice content. StreamingBlock changes might depend on decisions for compaction.

This PR is draft so we can check CI. 

n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/model/combine.go
+++ b/pkg/model/combine.go
@@ -4,8 +4,6 @@ import (
 	"bytes"
 	"fmt"
 
-	"github.com/grafana/tempo/tempodb/encoding/common"
-
 	"github.com/grafana/tempo/pkg/model/trace"
 	"github.com/grafana/tempo/pkg/tempopb"
 	"github.com/pkg/errors"
@@ -13,9 +11,11 @@ import (
 
 type objectCombiner struct{}
 
-var ObjectCombiner = objectCombiner{}
+type ObjectCombiner interface {
+	Combine(dataEncoding string, objs ...[]byte) ([]byte, bool, error)
+}
 
-var _ common.ObjectCombiner = (*objectCombiner)(nil)
+var StaticCombiner = objectCombiner{}
 
 // Combine implements tempodb/encoding/common.ObjectCombiner
 func (o objectCombiner) Combine(dataEncoding string, objs ...[]byte) ([]byte, bool, error) {
--- a/pkg/model/trace/matches.go
+++ b/pkg/model/trace/matches.go
@@ -9,10 +9,26 @@ import (
 	v1common "github.com/grafana/tempo/pkg/tempopb/common/v1"
 	v1 "github.com/grafana/tempo/pkg/tempopb/trace/v1"
 	"github.com/grafana/tempo/pkg/util"
-	"github.com/grafana/tempo/tempodb/search"
 )
 
-const RootSpanNotYetReceivedText = "<root span not yet received>"
+const (
+	RootSpanNotYetReceivedText = "<root span not yet received>"
+	RootServiceNameTag         = "root.service.name"
+	ServiceNameTag             = "service.name"
+	RootSpanNameTag            = "root.name"
+	SpanNameTag                = "name"
+	ErrorTag                   = "error"
+	StatusCodeTag              = "status.code"
+	StatusCodeUnset            = "unset"
+	StatusCodeOK               = "ok"
+	StatusCodeError            = "error"
+)
+
+var StatusCodeMapping = map[string]int{
+	StatusCodeUnset: int(v1.Status_STATUS_CODE_UNSET),
+	StatusCodeOK:    int(v1.Status_STATUS_CODE_OK),
+	StatusCodeError: int(v1.Status_STATUS_CODE_ERROR),
+}
 
 func MatchesProto(id []byte, trace *tempopb.Trace, req *tempopb.SearchRequest) (*tempopb.TraceSearchMetadata, error) {
 	traceStart := uint64(math.MaxUint64)
@@ -83,7 +99,7 @@ func MatchesProto(id []byte, trace *tempopb.Trace, req *tempopb.SearchRequest) (
 		rootSpanName = rootSpan.Name
 
 		for _, a := range rootBatch.Resource.Attributes {
-			if a.Key == search.ServiceNameTag {
+			if a.Key == ServiceNameTag {
 				rootServiceName = a.Value.GetStringValue()
 				break
 			}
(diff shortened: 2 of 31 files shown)


tempodb/encoding/blocker.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/grafana/tempo/tempodb/encoding/common"
+)
+
+type Finder interface {
+	FindTraceByID(ctx context.Context, id common.ID) (*tempopb.Trace, error)
+}
+
+type Searcher interface {
+	Search(ctx context.Context, req *tempopb.SearchRequest, opts ...SearchOption) (*tempopb.SearchResponse, error)
+}
```
annanay25, line 16: Nice additions. Curious if these could go into a `BackendBlock` interface, that would be really neat

```
type BackendBlock interface {
    Finder
    Searcher
    Iterator(?)
}
```

But unsure if we can achieve that level of abstraction outside of tempodb

annanay25 (approved):
Nice changes!

tempodb/encoding/blocker.go
mdisibio: I did originally start with a single interface for all block reads, but realized that Tempo never performs them at the same time.  It's either trace lookup or a search.  Therefore the split interface makes sense, and should be improve readability and testability.

pkg/model/trace/matches.go
```diff
@@ -9,10 +9,26 @@ import (
 	v1common "github.com/grafana/tempo/pkg/tempopb/common/v1"
 	v1 "github.com/grafana/tempo/pkg/tempopb/trace/v1"
 	"github.com/grafana/tempo/pkg/util"
-	"github.com/grafana/tempo/tempodb/search"
 )
 
-const RootSpanNotYetReceivedText = "<root span not yet received>"
+const (
```
joe-elliott, line 14: i feel like ultimately these will end up somewhere else. other block types like columnar will need access to the same constants?

tempodb/encoding/blocker.go
```diff
@@ -0,0 +1,51 @@
...
+	chunkSizeBytes     uint32
+	startPage          int
+	totalPages         int
+	maxBytes           int
+	prefetchTraceCount int
+}
+
+type SearchOption func(*SearchOptions)
+
+func WithPages(startPage, totalPages int) SearchOption {
```
joe-elliott, line 28: I, personally, am not a fan of the functional options pattern. I think it makes it harder to interact with the struct, but I'm also not adamantly opposed to it. 

I don't believe we use this anywhere else, but if the team wants to move this direction I am ok with it.

pkg/model/combine.go
```diff
@@ -4,18 +4,18 @@ import (
...
-
 	"github.com/grafana/tempo/pkg/model/trace"
 	"github.com/grafana/tempo/pkg/tempopb"
 	"github.com/pkg/errors"
 )
 
 type objectCombiner struct{}
 
-var ObjectCombiner = objectCombiner{}
+type ObjectCombiner interface {
```
joe-elliott, line 14: Since all of this model code is explicitly written for the "byte slice" style blocks I'm wondering if eventually this should all be consolidated under tempodb somewhere next to  the block types that use it. I suppose, however, that if we do move to a columnar format the point is moot b/c this basically all gets deleted anyway. 

Way too early to call now, but I wonder if there will be any reason to support both block styles?

tempodb/encoding/backend_block.go
```diff
@@ -139,3 +148,126 @@ func (b *BackendBlock) NewIndexReader() (common.IndexReader, error) {
...
+	}
+	defer iter.Close()
+
+	respMtx := sync.Mutex{}
+	resp = &tempopb.SearchResponse{
+		Metrics: &tempopb.SearchMetrics{},
+	}
+
+	var searchErr error
+	wg := boundedwaitgroup.New(5)
```
joe-elliott, line 205: I'm concerned about this b/c it will impact the amount of resources needed by the serverless functions which do not concurrently iterate through the block. it would be interesting to see if this ups throughput but I would like to do a test before merging this.

my instinct is to remove this concurrency and require the caller to do more in parallel.

tempodb/pool/pool.go
```diff
@@ -29,11 +29,10 @@ var (
 	})
 )
 
-type JobFunc func(ctx context.Context, payload interface{}) ([]byte, string, error)
+type JobFunc func(ctx context.Context, payload interface{}) (interface{}, error)
```
joe-elliott, line 32: Not something to address now, but thoughts on killing the job pool in favor of just using the boundedwaitgroup and goroutines?

The only reason I can think to keep this code long term is b/c it preserves order of jobs. So if multiple queries are hitting at the same time this should reduce long tail latency by completely finishing one trace by id query before moving to the next.

tempodb/pool/pool.go
```diff
@@ -29,11 +29,10 @@ var (
...
 
 type result struct {
-	data []byte
-	enc  string
+	data interface{}
```
joe-elliott, line 35: is there any reason to use an empty interface over *tempopb.Trace?

joe-elliott (approved):
A lot of questions that merit discussion, but no blockers.

For those of you following along at home, tempodb traditionally was meant to be an agnostic kv store built on top of object storage. This has served the project well so far, but we now need to expand our block types to make room for other formats such as columnar. Unfortunately, the only sane way to get this one was to migrate tempodb to specifically be a trace database built on top of object storage.

pkg/model/trace/matches.go
mdisibio: `pkg/model/trace` seems like an ok location, and I'd expect other block types to reference it for common operations like translation/inspection of proto. Were you thinking a different .go in this folder since, I agree, other block types wouldn't be using `matches.go` ?

tempodb/pool/pool.go
mdisibio: Yes had the same thoughts exactly (while editing this I had fond memories of tuning pool size for large block lists).  But leaned toward making that change in a separate PR.

tempodb/pool/pool.go
mdisibio: If we wanted to keep the pool around for any other activities.   Following comments on above, let's remove it in a next PR.

tempodb/encoding/backend_block.go
mdisibio: If we made this a parameter and serverless passed `1`, would that negatively impact it also?   Considering what a boundedwaitgroup of 1 looks like, seems so and inefficient.  Will remove.   

The remaining impacted caller is the querier for backend search, when it prefers self for a search shard. Since there's already good parallelization over the whole request, thinking it's ok as-is if an individual shard goes from concurrency=5 to 1. Thoughts?

pkg/model/trace/matches.go
joe-elliott: I guess in my mind `pkg/model` is eventually going away since it works with the "byte slice" tempodb style. Although maybe it lives on for constants like this? Also, it's possible that the "SegmentDecoder" functionality survives new block formats as well.

0 issues with moving this here now. I think this is all in flux and I'm sure everything will land somewhere that makes sense.

tempodb/encoding/backend_block.go
joe-elliott: Agree. Let's just remove the boundedwaitgroup/concurrency here and we can increase `search_prefer_self` if we want to increase the amount of work a querier will do.

tempodb/encoding/blocker.go
mdisibio: Team slightly leaned against this pattern, so replaced it with a basic struct.

PR #1328, opened 2022-03-07
docs: add metrics-generator configuration

yvrhdn:
**What this PR does**:
Metrics-generator configuration block is stabilising, time to document it!

**Which issue(s) this PR fixes**:
Related to #1303 

**Checklist**
- [ ] ~~Tests updated~~
- [x] Documentation added
- [ ] ~~`CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`~~

PR #1349, opened 2022-03-23
metrics-generator: use shuffle sharding to allow tenants to use smaller rings

kvrhdn:
**What this PR does**:
In large multi-tenant clusters we might run Tempo with a lot of replicas. In this case tenants will spread out the load across all instances as well. With the metrics-generator this is not ideal as each replica in the ring will increase the amount of active series generated in the downstream TSDB. With shuffle sharding it's possible to use a smaller ring per tenant.

This PR adds a new limit: `metrics_generator_ring_size` which allows you to configure the ring size per tenant. With this configured, the ring size of a tenant will not grow, even if more replicas are added.
If it's 0 or larger than the actual ring, the actual ring size is used.

Related to #1303 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


mapno (approved):
LGTM, nice addition.

kvrhdn:
Yeah fully agree. I was thinking of adding a page under [Deployment and Operations](https://grafana.com/docs/tempo/latest/operations/) specifically about the metrics-generator and how to deal with scaling and active series.
I'll update the limits doc in https://github.com/grafana/tempo/pull/1328 as well.

PR #1351, opened 2022-03-25
Refactor for alternative block formats 2

mdisibio:
**What this PR does**:
This PR continues building on #1346 to move towards alternative block formats (i.e. parquet).  It is quite a hefty refactor, changes listed below.  

* Make compaction block-agnostic by introducing the `encoding/common/Compactor` interface and moving the low-level logic into `encoding/v2`. TempoDB compactor calls this to create the new block, but still does the surrounding bookkeeping like selecting the input blocks and marking them as compacted.
* Move the bulk of `encoding` primitives into `encoding/v2`. This includes things like Append/Backend/StreamingBlocks, iterators, etc.  These are all tied closely to the v2 encoding and aren't needed for parquet. Common things like bloom filter shards and file names _are_ needed and were instead moved to `encoding/common`.
* Move some metrics and error constants around to break dependencies. 

The final folder layout is like this:
```
* tempodb
   * backend       <- Raw/reads writes 
   * encoding      <- Block agnostic entry point
       * common    <- stuff used by multiple encodings (interfaces, bloom filters, etc).   
       * v2        <- v2 stuff, internal
       * vparquet  <- Parquet stuff will go here 
   * metrics       <- Tempodb metrics used across things (i.e. only the compactor ones for now)
``` 

This move makes it more obvious where the other parts of Tempo are using v2 primitives and need to be addressed such as tempo-cli (regen bloom/index commands), ingester (completing blocks and pushing them to the backend), querier, serverless.

I think one more PR will be needed after this to tighten up the top-level `encoding` and `readerWriter` interfaces and remove most (all?) usages of `encoding/v2` outside of `encoding`.   And figure out where things meet in the middle:  I.e. when the serverless function wants to search a block, do we still want something like `encoding.NewBackendBlock(meta...).Search(...)` or something like `tempodb.Search(meta...)`.

Fixes n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added


annanay25 (commented):
Overall this refactor is making perfect sense. It's great to see how all the proto specific details have been abstracted away into "v2" and we can start afresh in "vparquet".  Left a couple of comments, will add as I think about this further..

cmd/tempo-cli/cmd-list-block.go
```diff
@@ -85,7 +85,7 @@ func dumpBlock(r tempodb_backend.Reader, c tempodb_backend.Compactor, tenantID s
 	if scan {
 		fmt.Println("Scanning block contents.  Press CRTL+C to quit ...")
 
-		block, err := encoding.NewBackendBlock(&unifiedMeta.BlockMeta, r)
+		block, err := v2.NewBackendBlock(&unifiedMeta.BlockMeta, r)
```
joe-elliott, line 88: based on the PR comment I think we're on the same page here. Ultimately this will call `encoding.NewBackendBlock()` and that code will use the meta to determine which object to return?

cmd/tempo-cli/cmd-gen-index.go
```diff
@@ -53,7 +54,7 @@ func ReplayBlockAndGetRecords(meta *backend.BlockMeta, filepath string) ([]commo
 			break
 		}
 
-		iter := encoding.NewIterator(bytes.NewReader(buffer), objectRW)
+		iter := v2.NewIterator(bytes.NewReader(buffer), objectRW)
```
joe-elliott, line 57: can the concept of an iterator even exist with the parquet formats? cli is going to require some real thought as we move to parquet. a lot of these commands won't even make sense anymore and we'll probably want entirely new ones.

easiest thing may be to just drop all cli options that no longer work with parquet, cut tempo 2.0 to justify the breaking changes, and then dust our hands off convincingly.

modules/ingester/ingester_test.go
```diff
@@ -320,7 +320,7 @@ func defaultIngesterModule(t *testing.T, tmpDir string) *Ingester {
 			Local: &local.Config{
 				Path: tmpDir,
 			},
-			Block: &encoding.BlockConfig{
+			Block: &common.BlockConfig{
```
joe-elliott, line 323: this is a tough one. block config is _very_ opinionated for current block and should probably end up in v2?

noted elsewhere, but details like this are making me think a cutover to 2.0 is necessary. tempo 2.0 could presumably read the old blocks but never create new ones. then we could ditch config like this.

tempodb/compactor.go
```diff
@@ -119,7 +81,7 @@ func (rw *readerWriter) doCompaction() {
 			level.Warn(rw.logger).Log("msg", "unable to find meta during compaction.  trying again on this block list", "err", err)
 		} else if err != nil {
 			level.Error(rw.logger).Log("msg", "error during compaction cycle", "err", err)
-			metricCompactionErrors.Inc()
+			metrics.MetricCompactionErrors.Inc()
```
joe-elliott, line 84: i can go either way on this change, but it is a pretty big shift. do you intend to move all tempodb metrics here eventually?

tempodb/encoding/common/block.go
```diff
@@ -0,0 +1,17 @@
+package common
+
+import "strconv"
+
+const (
```
joe-elliott, line 5: similar thoughts here as for BlockConfig. this will all end up in v2 eventually?

tempodb/encoding/common/interfaces.go
```diff
@@ -0,0 +1,53 @@
+package common
```
joe-elliott, line 1: +1 to this. let's go! this is what we need to make the big shift to a new block format.

tempodb/encoding/common/interfaces.go
```diff
@@ -0,0 +1,53 @@
...
+	ChunkSizeBytes     uint32 // Buffer size to read from backend storage.
+	StartPage          int    // Controls searching only a subset of the block. Which page to begin searching at.
+	TotalPages         int    // Controls searching only a subset of the block. How many pages to search.
+	MaxBytes           int    // Max allowable trace size in bytes. Traces exceeding this are not searched.
+	PrefetchTraceCount int    // How many traces to prefetch async.
+}
+
+func DefaultSearchOptions() SearchOptions {
+	return SearchOptions{
+		ChunkSizeBytes: 1_000_000,
```
joe-elliott, line 30: i think we're just going to have to have a period in which there are settings and options that don't make sense for parquet.

tempodb/metrics/metrics.go
```diff
@@ -0,0 +1,39 @@
+package metrics
```
joe-elliott, line 1: discussed elsewhere, but i do like the tight coupling of metrics with the code that uses them in the same file.

i consider this a minor preference though and can go either way.

tempodb/wal/local_block.go
```diff
@@ -19,16 +19,16 @@ const nameFlushed = "flushed"
 // LocalBlock is a block stored in a local storage.  It can be searched and flushed to a remote backend, and
 // permanently tracks the flushed time with a special file in the block
 type LocalBlock struct {
```
joe-elliott, line 21: ultimately i think LocalBlock will need to be an interface like BackendBlock and be either parquet or the current format or whatever so it can be directly shipped to object storage.

The AppendBlock can remain in the current format for the foreseeable future and we can do the translation when we "cut" the head block.

joe-elliott (commented):
:heart: 

some thoughts. i'd approve but it's still marked draft :). overall 100% on board with this trajectory. in my mind i was already moving all the encoding stuff below v2 as well.

tempodb/metrics/metrics.go
mdisibio: Yeah agree about the tight coupling, but didn't come up with a convincing refactor that kept it.  The way I chose is for the proto and future (parquet) compactor implementations to have direct access to these metrics, but this also means moving the metrics to avoid a circular dependency.  I.e. tempodb compactor (metrics originally here) -> encoding -> v2 -> metrics.   Otherwise, if we keep the metrics in tempodb/compactor.go, then we have to devise something like callbacks for each time an object is combined, or bytes written.  That seemed less readable than moving the metrics.

tempodb/encoding/common/interfaces.go
mdisibio: Agree, and I think that's ok.  I'm sure there will be new parquet settings which don't make sense for the existing blocks also.

cmd/tempo-cli/cmd-list-block.go
mdisibio: I think there are several options here, basically cosmetic.  

1. We can do `encoding.NewBackendBlock` which returns a new `BackendBlock` interface, but that creates the do-it-all interface that I was trying to avoid.  
2. `encoding.SearchBlock` returns `Searcher`, etc. The smaller interfaces here seem better for readability and testability, but maybe cluttered if there are too many.
3. Or maybe this should all be in `tempodb`.  I don't think `encoding` was intended to be the entry-point to the backend, but it's trending that way.

Was leaving these decisions for the next PR but happy to tackle some now if we can.

cmd/tempo-cli/cmd-gen-index.go
mdisibio: Yes we could make an iterator for `tempopb.Trace`. It would work for every block type, but much slower too.  It wouldn't be used for any normal functions, but it is somewhat useful as a CLI command to dump the contents of a block.  This specific command `gen index` isn't relevant to parquet at all.

I don't think we _have_ to drop the cli commands.  One possibility to leave the existing commands working against a proto (v2) block but error out on a parquet block.

modules/ingester/ingester_test.go
mdisibio: Wanted to avoid breaking changes to the yaml, and the module layout makes it a little slow to reason about.      Here's an idea:  we could move these fields to a new `v2/BlockConfig` and then move this back and compose the v2 struct into it. That way the yaml is unchanged, but the fields are still versioned.

tempodb/encoding/common/block.go
mdisibio: Some of it is expected to remain common across encodings, like the meta and bloom filters.  The rest, like copying the `index` file does need to move.  Wasn't in a hurry to move this yet as I think it depends on what we decide for the "entrypoint" to the backend (some options listed in the other comment)

tempodb/encoding/versioned.go
```diff
@@ -24,52 +22,28 @@ type VersionedEncoding interface {
 	NewIndexReader(ra backend.ContextReader, pageSizeBytes int, totalPages int) (common.IndexReader, error)
```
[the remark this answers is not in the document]
mdisibio: Yep, was leaving those for the next refactor after we make some decisions about what the "entrypoint" should look like.

cmd/tempo-cli/cmd-list-block.go
joe-elliott: Nah, feel free to move forward with this PR as is. I think I'd favor #2, but that's just a gut call w/o digging into the details.

modules/ingester/ingester_test.go
joe-elliott: i'm not super concerned about it's current location, so feel free to leave it as is while we slowly moving things. i think we need to sync on bigger plans which will make things like "what do we do with BlockConfig" more apparent. i.e. once a new block format exists are we going to only create the new type? will the user even be able to create the old type?

tempodb/encoding/versioned.go
```diff
@@ -24,52 +22,28 @@ type VersionedEncoding interface {
...
 
 	NewObjectReaderWriter() common.ObjectReaderWriter
-	NewRecordReaderWriter() common.RecordReaderWriter
+
+	NewCompactor() common.Compactor
```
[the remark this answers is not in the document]
mdisibio: Agree, wasn't 100% on the name but couldn't think of anything better.  Technically the new `Compactor` interface only has one method. Would it be more readable to have the new `Compact` method directly in `VersionedEncoding`?

joe-elliott (approved):

PR #1354, opened 2022-03-31
Update mixin for TempoIngesterFlushes thresholds

zalegrala:
**What this PR does**:

* replace the exiting critical alert for TempoIngesterFlushes with a warning, and lengthen the critical threshold duration
* replace the metric used for TempoIngesterFlushes to only measure retries

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/operations/tempo-mixin/alerts.libsonnet
+++ b/operations/tempo-mixin/alerts.libsonnet
@@ -86,8 +86,8 @@
             // wait 5m for failed flushes to self-heal using retries
             alert: 'TempoIngesterFlushesUnhealthy',
             expr: |||
-              sum by (%s) (increase(tempo_ingester_flush_failed_retries_total{}[1h])) > %s and
-              sum by (%s) (increase(tempo_ingester_flush_failed_retries_total{}[5m])) > 0
+              sum by (%s) (increase(tempo_ingester_failed_flushes_total{}[1h])) > %s and
+              sum by (%s) (increase(tempo_ingester_failed_flushes_total{}[5m])) > 0
             ||| % [$._config.group_by_cluster, $._config.alerts.flushes_per_hour_failed, $._config.group_by_cluster],
             'for': '5m',
             labels: {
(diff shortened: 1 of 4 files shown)


operations/tempo-mixin/alerts.libsonnet
```diff
@@ -84,12 +84,28 @@
...
-              sum by (%s) (increase(tempo_ingester_failed_flushes_total{}[5m])) > 0
+              sum by (%s) (increase(tempo_ingester_flush_failed_retries_total{}[1h])) > %s and
+              sum by (%s) (increase(tempo_ingester_flush_failed_retries_total{}[5m])) > 0
             ||| % [$._config.group_by_cluster, $._config.alerts.flushes_per_hour_failed, $._config.group_by_cluster],
             'for': '5m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: 'Greater than %s flushes have failed in the past hour.' % $._config.alerts.flushes_per_hour_failed,
```
mdisibio, line 97: I think it would be good to tweak the message so the warning and critical are different.  Don't have a recommendation though... any ideas?

zalegrala: Good idea, I'll wordsmith around a little.

zalegrala: Updated a little.  How's that read to you now?

operations/tempo-mixin/alerts.libsonnet
```diff
@@ -84,17 +84,33 @@
           },
           {
             // wait 5m for failed flushes to self-heal using retries
-            alert: 'TempoIngesterFlushesFailing',
+            alert: 'TempoIngesterFlushesUnhealthy',
             expr: |||
-              sum by (%s) (increase(tempo_ingester_failed_flushes_total{}[1h])) > %s and
-              sum by (%s) (increase(tempo_ingester_failed_flushes_total{}[5m])) > 0
+              sum by (%s) (increase(tempo_ingester_flush_failed_retries_total{}[1h])) > %s and
```
joe-elliott, line 89: what do we think about keying the "unhealthy" warning alert  on the old metric and the "failing" critical alert on the retries metric?

then for consistency keep the `for: 5m` on both?

zalegrala: That could be reasonable.  The goal being to always know when a flush failed at a warning level, but only get paged when a retry fails for the `5m` duration?

joe-elliott: Yeah, that's what i was thinking. :+1:

zalegrala: Okay, I've made the update.

joe-elliott (approved):

PR #1356, opened 2022-04-01
Update to latest dskit (new flagext.Secret and move of pkg/tenant)

simonswine:
- Update dskit
- Refactor to make use of new secret flag type
- Use tenant package from dskit


tempodb/backend/s3/config.go
```diff
@@ -10,7 +10,7 @@ type Config struct {
 	Bucket            string         `yaml:"bucket"`
 	Endpoint          string         `yaml:"endpoint"`
 	Region            string         `yaml:"region"`
-	AccessKey         flagext.Secret `yaml:"access_key"`
+	AccessKey         string         `yaml:"access_key"`
```
annanay25, line 13: I believe this change should be mentioned in the changelog as its a user facing change. The YAML marshaller will no longer print `AccessKey` as `********`

go.mod
```diff
@@ -89,6 +89,8 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 )
 
+require github.com/cespare/xxhash/v2 v2.1.2
```
annanay25, line 92: nit: Can we club this require statement with the above block?

annanay25 (approved):
LGTM, thanks @simonswine. Left a few minor comments.

go.mod
simonswine: Unsure why that happened, now moved to the first require block

tempodb/backend/s3/config.go
simonswine: Good point added a changelog entry

simonswine:
A bit unsure what's wrong with CI, but it seems to be in other PRs as well: (e.g. https://github.com/grafana/tempo/runs/5822436282?check_suite_focus=true)

annanay25 (approved):
Thanks! LGTM.

PR #1359, opened 2022-04-04
Update single-binary-extras yaml api changes

manishmaharjan:
**What this PR does**:
extensions/v1beta1 has been deprecated and no longer works from v1.22 onwards. This changes reflects the necessary changes required for this to work...


mdisibio:
Thanks for the update. This will also move the minimum supported k8s version to 1.19 which should be ok.  Will review.

zalegrala:
I was able to test this locally and it appears to work as desired.  Marty points out in https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122 that the `backend` term becomes `defaultBackend`.  Do we need to make this change as well?

```json
❯ k get ingress ingress -o json | jq '{apiVersion,spec}'
{
  "apiVersion": "networking.k8s.io/v1",
  "spec": {
    "rules": [
      {
        "http": {
          "paths": [
            {
              "backend": {
                "service": {
                  "name": "grafana",
                  "port": {
                    "number": 80
                  }
                }
              },
              "path": "/",
              "pathType": "Prefix"
            }
          ]
        }
      }
    ]
  }
}
```

manishmaharjan:
I earlier tried adding `defaultBackend`, but got some error. Therefore, I had to change this way to make it work.

mdisibio:
Researched some more, and maybe just misinterpreting the migration guide.  `defaultBackend` is optional, so this seems ok.

zalegrala:
The lint failure here looks unrelated to your change, but we did recently update the golangci version in `main`.  Would you mind rebasing to bring in that change?

manishmaharjan:
Alright, Hope those lint failure will go away this time. :)

zalegrala (approved):

zalegrala:
Looks good to me, thanks for the contribution.

PR #1360, opened 2022-04-04
update golint-ci to latest

mdisibio:
**What this PR does**:
Update golangci-lint to latest which includes support for 1.18 (even though Tempo isn't using it directly). This PR was having issues:  https://github.com/grafana/tempo/runs/5818491537?check_suite_focus=true  and I think it's related and something changed recently.  Notably the previous step sets up 1.17, but then golangci-lint sets up 1.18 anyway.   The docker images are using 1.18 now too:

```
$ docker run -it golangci/golangci-lint:v1.45.2-alpine go version
go version go1.18 linux/amd64
```

~This means there is more cleanup to do:  golangci-lint has a parameter `skip-go-installation` which can be set to `true` that we aren't using. Options:~

~1. Turn on `skip-go-installation` and stick with 1.17.~
~2. Remove the first `setup-go` step and let golangci-lint do it~

~I lean towards option 2.~

See comment below about v3:

Fixes n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Lint
-        uses: golangci/golangci-lint-action@v2
+        uses: golangci/golangci-lint-action@v3
         with:
           version: v1.45.2
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Lint
         uses: golangci/golangci-lint-action@v2
         with:
-          version: v1.41.1
+          version: v1.45.2
 
   unit-tests:
     name: Test packages


zalegrala:
Good find here.  I think I lean towards option 1, but mostly so we have some more control over which go version is being used.  Perhaps this is less importantly because it is only being used for the lint, but if differences in go version might mean a difference in outcome, that could be somewhat harder to troubleshoot locally with two go versions.  But I also see your point about just letting golangci do what it wants.  Perhaps we go that route and see if there are any issues with the difference between the build and the lint that cause dev headaches?

mapno (commented):
Nice find! I was having the same issues with a different PR.

I'm fine with option 2. I think it'll be easier to maintain if we leave to golangci to set up itself. We can control the go version used via golangci's own version too.

mdisibio:
Oops, golangci-lint has [moved away from doing go setup in v3](https://github.com/golangci/golangci-lint-action#compatibility) so option 1 it is.  Let's try using v3 too.

mapno (approved):

PR #1361, opened 2022-04-04
Service graphs: add dimensions and rename expired edges metric

mapno:
Add dimensions to the service graphs processor config, which attaches
additional labels to the metrics from spans attributes.

It also renames the expired edges internal metric.

### DIFF
--- a/cmd/tempo-vulture/main.go
+++ b/cmd/tempo-vulture/main.go
@@ -332,7 +332,7 @@ func searchTag(client *util.Client, seed time.Time) (traceMetrics, error) {
 		zap.String("hexID", hexID),
 		zap.Duration("ago", time.Since(seed)),
 		zap.String("key", attr.Key),
-		zap.String("value", attr.Value.GetStringValue()),
+		zap.String("value", util.StringifyAnyValue(attr.Value)),
 	)
 	logger.Info("searching Tempo")
 
--- a/cmd/tempo-vulture/main.go
+++ b/cmd/tempo-vulture/main.go
@@ -340,7 +340,7 @@ func searchTag(client *util.Client, seed time.Time) (traceMetrics, error) {
 	//  around the seed.
 	start := seed.Add(-30 * time.Minute).Unix()
 	end := seed.Add(30 * time.Minute).Unix()
-	resp, err := client.SearchWithRange(fmt.Sprintf("%s=%s", attr.Key, attr.Value.GetStringValue()), start, end)
+	resp, err := client.SearchWithRange(fmt.Sprintf("%s=%s", attr.Key, util.StringifyAnyValue(attr.Value)), start, end)
 	if err != nil {
 		logger.Error(fmt.Sprintf("failed to search traces with tag %s: %s", attr.Key, err.Error()))
 		tm.requestFailed++
--- a/modules/generator/processor/servicegraphs/servicegraphs.go
+++ b/modules/generator/processor/servicegraphs/servicegraphs.go
@@ -12,6 +12,7 @@ import (
 	"github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/prometheus/prometheus/util/strutil"
 
 	gen "github.com/grafana/tempo/modules/generator/processor"
 	"github.com/grafana/tempo/modules/generator/processor/servicegraphs/store"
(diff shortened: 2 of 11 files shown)


modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -188,29 +196,51 @@ func (p *processor) consume(resourceSpans []*v1.ResourceSpans) error {
...
 
 // collectEdge records the metrics for the given edge.
 // Returns true if the edge is completed or expired and should be deleted.
 func (p *processor) collectEdge(e *store.Edge) {
 	if e.IsCompleted() {
-		clientServerLabelValues := registry.NewLabelValues([]string{e.ClientService, e.ServerService})
+		values := []string{e.ClientService, e.ServerService}
+		for _, dimension := range p.cfg.Dimensions {
+			values = append(values, e.Dimensions[dimension])
+		}
```
kvrhdn, line 210: We can preallocate this slice with the expected length, this saves some memory allocations as we append the dimensions.

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -188,29 +196,51 @@ func (p *processor) consume(resourceSpans []*v1.ResourceSpans) error {
...
+	}
+}
+
+func (p *processor) upsertDimensions(m map[string]string, resourceAttr []*v1common.KeyValue, attr []*v1common.KeyValue) {
+	for _, dim := range p.cfg.Dimensions {
+		for _, kv := range resourceAttr {
+			key := kv.Key
+			if key == dim {
+				m[key] = commonUtil.StringifyAnyValue(kv.Value)
+				continue
```
kvrhdn, line 231: This `continue` should break to the outer loop since we found the dimension, right?
```go
for _, dim := range p.cfg.Dimensions {
```
Maybe add a label to clarify what for-loop we want to continue.

pkg/util/attributes.go
```diff
@@ -0,0 +1,44 @@
...
+		return anyValue.GetStringValue()
+	}
+
+	return ""
+}
+
+var invalidLabelCharRE = regexp.MustCompile(`[^a-zA-Z0-9_]`)
+
+// SanitizeLabelName sanitizes a label name for Prometheus.
+func SanitizeLabelName(name string) string {
```
kvrhdn, line 42: WDYT about calling this `SanitizePrometheusLabelName`? Just to clarify this method is only intended to handle Prometheus requirements.

kvrhdn: Oh I see Prometheus has a similar function, can we just use their function instead? It's not private or internal.

https://github.com/prometheus/prometheus/blob/main/util/strutil/strconv.go#L39-L43

cmd/tempo-vulture/main.go
```diff
@@ -332,7 +332,7 @@ func searchTag(client *util.Client, seed time.Time) (traceMetrics, error) {
 		zap.String("hexID", hexID),
 		zap.Duration("ago", time.Since(seed)),
 		zap.String("key", attr.Key),
-		zap.String("value", attr.Value.GetStringValue()),
+		zap.String("value", util.StringifyAnyValue(attr.Value)),
```
kvrhdn, line 335: Is there a reason why you changed this `GetStringValue` but not the one 8 lines further?

kvrhdn (commented):
This will be really neat! Service graphs per cluster or namespace 🤩 

Left some comments and questions but looks good overall.

modules/generator/processor/servicegraphs/servicegraphs.go
mapno: Addressed in 309793e3fe14e9817d916b0c89d46c77f40508d5

modules/generator/processor/servicegraphs/servicegraphs.go
mapno: Refactored in 309793e3fe14e9817d916b0c89d46c77f40508d5

pkg/util/attributes.go
mapno: Changed in 309793e3fe14e9817d916b0c89d46c77f40508d5

cmd/tempo-vulture/main.go
mapno: I missed it. Added in 309793e3fe14e9817d916b0c89d46c77f40508d5

kvrhdn (approved):
Nice work!

PR #1364, opened 2022-04-05
metrics-generator: emit exemplars

kvrhdn:
**What this PR does**:
Follow up from #1340 and adds the capability to capture and remote write exemplar.

Related to #1303 

**Checklist**
- [x] Tests updated
- [ ] ~~Documentation added~~

### DIFF
--- a/modules/generator/registry/interface.go
+++ b/modules/generator/registry/interface.go
@@ -15,9 +15,8 @@ type Counter interface {
 // Histogram
 // https://prometheus.io/docs/concepts/metric_types/#histogram
 type Histogram interface {
-	Observe(values *LabelValues, value float64)
-	// TODO support exemplars
-	//ObserveWithExemplar(lbls labels.Labels, value float64, exemplarLbls labels.Labels)
+	// Observe a datapoint with the given values. traceID will be added as exemplar.
+	Observe(values *LabelValues, value float64, traceID string)
 }
 
 // LabelValues is a wrapper around a slice of label values. It has the ability to cache the hash of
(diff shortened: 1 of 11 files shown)


kvrhdn:
Service graphs with exemplars 🎉 

Screenshot 2022-04-05 at 19 40 30

Span metrics with exemplars 🎉 

Screenshot 2022-04-05 at 19 41 47

modules/generator/registry/interface.go
```diff
@@ -15,9 +15,8 @@ type Counter interface {
 // Histogram
 // https://prometheus.io/docs/concepts/metric_types/#histogram
 type Histogram interface {
-	Observe(values *LabelValues, value float64)
-	// TODO support exemplars
-	//ObserveWithExemplar(lbls labels.Labels, value float64, exemplarLbls labels.Labels)
+	// Observe a datapoint with the given values. traceID will be added as exemplar.
+	Observe(values *LabelValues, value float64, traceID string)
```
mapno, line 19: Why not use similar interface as Prometheus: `Observe` and `ObserveWithExemplar`?

mapno (approved):
LGTM

modules/generator/registry/interface.go
kvrhdn: I originally had both functions but then noticed that we never use `Observe`. If we don't use it, why bother supporting it? 🤷🏻 
I'm guessing that we will always have a traceID available since every metric is based upon trace data. If at some point this isn't the case anymore we can add the original observe function again.

PR #1368, opened 2022-04-07
Allow GCS config for setting attributes on new objects

### LINKED ISSUE #1348: Configuring GCS Metadata
When storing objects in GCS, I often have to deal with setting object-level metadata, particularly the caching-related metadata. Currently, Tempo does not provide a way to do this. 

A config option to set metadata. If I could define metadata in the tempo config and it gets attached to all the objects that tempo stores in the GCS, that would be the ultimate solution. 

**Additional context**
I have been benchmarking Tempo on deciding an ultimate tool for my different use cases. During the benchmarks, I wanted to configure the GCS caching behavior that is often configured using object-level metadata. Unfortunately, Tempo does not provide a way to configure any kind of Cloud Storage metadata.

zalegrala:
**What this PR does**:

Here we add two options to configure GCS object attributes.  One to control the aspects of object caching, and the other to allow users to specify metadata on the object.

Fixes #1348

**Checklist**
- [x] Tests updated
- [x] Documentation added


tempodb/backend/gcs/gcs.go
```diff
@@ -168,7 +173,15 @@ func (rw *readerWriter) Shutdown() {
 }
 
 func (rw *readerWriter) writer(ctx context.Context, name string) *storage.Writer {
-	w := rw.bucket.Object(name).NewWriter(ctx)
+	o := rw.bucket.Object(name)
+	if configuredObjectAttributes(rw.cfg.ObjAttrs) {
+		_, err := o.Update(ctx, rw.cfg.ObjAttrs)
+		if err != nil {
+			level.Error(rw.logger).Log("msg", "failed to update object attributes", "objectName", name, "err", err)
```
mdisibio, line 180: I think this could lead to a nil reference/panic in `.Append()` and `.Write()`.  Thoughts on returning the error and doing normal error checking?

cmd/tempo-serverless/handler.go
```diff
@@ -196,6 +198,7 @@ func setDecodeHooks(c *mapstructure.DecoderConfig) {
 		mapstructure.StringToTimeDurationHookFunc(),
 		mapstructure.StringToSliceHookFunc(","),
 		stringToFlagExt(),
+		mapstructure.StringToTimeHookFunc(time.RFC3339),
```
mdisibio, line 201: Do you have an example of a config that this allows to be parsed?  Would it be to set timestamps on the GCS objects?

zalegrala: I don't yet, but I will be working on integration tests soon that should surface this elsewhere.  Though, I don't think we're using viper elsewhere so this specifically shouldn't be an issue after this change.  This came up because of the marshal<>unmarshal round tripping that the viper `loadConfig()` does for the serverless stuff gave a default value for the `time.Time`, which was then confusing the mapstructure pkg.

tempodb/backend/gcs/gcs.go
zalegrala: Good call, I didn't check those methods, but I like returning an error better than a nil that the caller has to know something about.  That said, should we also update the `.Append()` and `.Write()` methods to avoid a nil pointer should one get returned, or just rely on the error value to tell us we aren't in a good state?

mdisibio: I think relying on the error is idiomatic go and ok.

cmd/tempo-serverless/handler.go
joe-elliott: Do we have any config fields that are of type time? I'm not sure we need this.

tempodb/backend/gcs/config.go
```diff
@@ -1,12 +1,19 @@
...
+)
 
 type Config struct {
-	BucketName        string        `yaml:"bucket_name"`
-	ChunkBufferSize   int           `yaml:"chunk_buffer_size"`
-	Endpoint          string        `yaml:"endpoint"`
-	Insecure          bool          `yaml:"insecure"`
-	HedgeRequestsAt   time.Duration `yaml:"hedge_requests_at"`
-	HedgeRequestsUpTo int           `yaml:"hedge_requests_up_to"`
+	BucketName string `yaml:"bucket_name"`
```
joe-elliott, line 10: i prefer the ordering/organization of this struct before.

tempodb/backend/gcs/gcs.go
```diff
@@ -167,10 +175,14 @@ func (rw *readerWriter) ReadRange(ctx context.Context, name string, keypath back
 func (rw *readerWriter) Shutdown() {
 }
 
-func (rw *readerWriter) writer(ctx context.Context, name string) *storage.Writer {
-	w := rw.bucket.Object(name).NewWriter(ctx)
+func (rw *readerWriter) writer(ctx context.Context, name string) (*storage.Writer, error) {
```
joe-elliott, line 178: why did we add an error return here? you don't seem to be using it and it only complicates code above.

integration/e2e/config-all-in-one-gcs.yaml
```diff
@@ -38,9 +38,11 @@ storage:
       bucket_name: tempo
       endpoint: https://tempo_e2e-gcs:4443/storage/v1/
       insecure: true
+      obj_attrs:
+        cachecontrol: "no-cache"
```
joe-elliott, line 42: does this work? will it set this field:

https://github.com/googleapis/google-cloud-go/blob/main/storage/storage.go#L1263

I wonder if we can only expose the ability to configure this:

https://github.com/googleapis/google-cloud-go/blob/main/storage/storage.go#L1326

and then if the user specifies a key of `Cache-Control` it will be magically translated into the Fixed-key metadata:

https://cloud.google.com/storage/docs/metadata

tempodb/backend/gcs/config.go
zalegrala: This update was to try an satisfy the maligned struct lint check, but wasn't complete enough to fix it.  Still in progress.

tempodb/backend/gcs/gcs.go
zalegrala: There was an update call in a previous commit that returned an error.  If this doesn't end up returning an error, I agree we can drop the signature change.

cmd/tempo-serverless/handler.go
zalegrala: https://github.com/googleapis/google-cloud-go/blob/main/storage/storage.go#L1391 for reference.

tempodb/backend/gcs/gcs.go
zalegrala: The `err` is back now with the use of `mergo`.

joe-elliott: i'll admit i really don't like the use of mergo here. likely everything would be fine, but this could also be a source of future head scratching bugs. I'd prefer either 

- manually copying over a set of static fields we support
OR
- only setting the ObjectAttrs.Metadata map. It's possible that setting a "Cache-Control" = "no-cache" in this map is translated by the API into the fixed attribute.

zalegrala: If the ObjectAttrs.Metadata map is enough, that seems fine just to allow a `map[string]string` to be set in our config and then only update that.  I'm not sure without testing, but I wonder why there would be two methods to update the same value control.  Perhaps one is read only?  Reading here was what clued me in to where the attributes need to be set.

https://github.com/googleapis/google-cloud-go/blob/main/storage/storage.go#L1126

But I also see that the Metadata documentation mentions "user provided" fields.

zalegrala: This doc makes me think `metadata` is only used for custom key-value pairs.

https://cloud.google.com/storage/docs/viewing-editing-metadata

The use of "standard" vs "custom" in the docs seems to indicate that to set a standard field like Cache-Control, you use the top level field, and for custom bits you use the metadata field.

I'll work up a test to see what it looks like.

zalegrala: Setting the metadata doesn't appear to update the `CacheControl` field on the object.

https://gist.github.com/zalegrala/9d86cbeefa1ebb43bee90afd6e6d4600

The first test file was created with the `metadata` field set, and the second was created with the `CacheControl` field on the object set.

```
❯ gsutil ls -L gs://zalegrala-test/test1
gs://zalegrala-test/test1:
    Creation time:          Mon, 11 Apr 2022 20:43:16 GMT
    Update time:            Mon, 11 Apr 2022 20:43:16 GMT
    Storage class:          STANDARD
    Content-Length:         13
    Content-Type:           text/plain; charset=utf-8
    Metadata:               
        Cache-Control:      no-cache
    Hash (crc32c):          sqwNTw==
    Hash (md5):             KVH1pxihX0ylP/tHoGXnAw==
    ETag:                   COSVp/PvjPcCEAE=
    Generation:             1649709796674276
    Metageneration:         1
    ACL:                    []
TOTAL: 1 objects, 13 bytes (13 B)
```
```
❯ gsutil ls -L gs://zalegrala-test/test2
gs://zalegrala-test/test2:
    Creation time:          Mon, 11 Apr 2022 20:44:59 GMT
    Update time:            Mon, 11 Apr 2022 20:44:59 GMT
    Storage class:          STANDARD
    Cache-Control:          no-cache
    Content-Length:         13
    Content-Type:           text/plain; charset=utf-8
    Hash (crc32c):          sqwNTw==
    Hash (md5):             KVH1pxihX0ylP/tHoGXnAw==
    ETag:                   CMXJsaTwjPcCEAE=
    Generation:             1649709899605189
    Metageneration:         1
    ACL:                    []
TOTAL: 1 objects, 13 bytes (13 B)
```

Note the output only contains the cache control on the second.

Since it sounds like we are leaning to avoiding `mergo` and bringing in all possible attributes, which ones do we want to focus on supporting?  The issue we're aiming to fix here only mentions cache-control, but perhaps there are other common uses.

joe-elliott: Hmm, `CacheControl` and `Metadata` ? Do any other fields make sense to you?

zalegrala: Those sound like a good start to me.  We can increment if someone else wants something later.

tempodb/backend/gcs/config.go
```diff
@@ -1,12 +1,16 @@
...
 
 type Config struct {
-	BucketName        string        `yaml:"bucket_name"`
-	ChunkBufferSize   int           `yaml:"chunk_buffer_size"`
-	Endpoint          string        `yaml:"endpoint"`
-	Insecure          bool          `yaml:"insecure"`
-	HedgeRequestsAt   time.Duration `yaml:"hedge_requests_at"`
-	HedgeRequestsUpTo int           `yaml:"hedge_requests_up_to"`
+	BucketName        string            `yaml:"bucket_name"`
+	CacheControl      string            `yaml:"cache_control"`
```
joe-elliott, line 9: i think we should either make these a subject object or prefix them to clarify to the user that they are related to the objects.

Something like?
```
object_cache_control
object_metadata
```

docs/tempo/website/configuration/_index.md
```diff
@@ -374,6 +374,11 @@ storage:
             # Example: "bucket_name: tempo"
             [bucket_name: <string>]
 
+            # Optional
+            # Example: "cache_control: "no-cache""
+            # See the GCS documentation for more detail: https://cloud.google.com/storage/docs/metadata
```
joe-elliott, line 379: Pointing the reader to the GCS documentation is good, but we should include a line here that briefly explains that the metadata is written to the object.

Same comment for below config setting.

PR #1370, opened 2022-04-08
Metrics generator docker compose

mapno:
**What this PR does**:

* Add metrics-generator to `local` docker-compose example
* Remove unnecessary ring from metrics-generator. The metrics-generator's ring is instantiated as an independent module and used only by the distributor. There is no need for the metrics-generator to run it as well.

Contributes to https://github.com/grafana/tempo/issues/1303

### DIFF
--- a/example/docker-compose/local/tempo-local.yaml
+++ b/example/docker-compose/local/tempo-local.yaml
@@ -1,3 +1,5 @@
+metrics_generator_enabled: true
+
 server:
   http_listen_port: 3200
 
@@ -28,6 +30,15 @@ compactor:
     block_retention: 1h
     compacted_block_retention: 10m
 
+metrics_generator:
+  registry:
+    external_labels:
+      source: docker-compose
+  storage:
+    path: /tmp/tempo/generator/wal
+    remote_write:
+      - url: http://prometheus:9090/api/v1/write
+
 storage:
   trace:
     backend: local                     # backend configuration to use
@@ -43,3 +54,6 @@ storage:
     pool:
       max_workers: 100                 # worker pool determines the number of parallel requests to the object store backend
       queue_depth: 10000
+
+overrides:
+  metrics_generator_processors: [service-graphs, span-metrics]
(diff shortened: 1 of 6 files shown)


modules/generator/generator.go
```diff
@@ -98,12 +97,6 @@ func New(cfg *Config, overrides metricsGeneratorOverrides, reg prometheus.Regist
 		return nil, fmt.Errorf("create ring lifecycler: %w", err)
 	}
 
-	ringCfg := cfg.Ring.ToRingConfig()
-	g.ring, err = ring.NewWithStoreClientAndStrategy(ringCfg, ringNameForServer, RingKey, ringStore, ring.NewIgnoreUnhealthyInstancesReplicationStrategy(), prometheus.WrapRegistererWithPrefix("cortex_", reg), g.logger)
```
kvrhdn, line 102: Nice find! 🙌🏻  I did not know just the lifecycler was enough.

example/docker-compose/local/tempo-local.yaml
```diff
@@ -28,6 +30,15 @@ compactor:
     block_retention: 1h
     compacted_block_retention: 10m
 
+metrics_generator:
+  registry:
+    external_labels:
+      source: docker-compose
```
kvrhdn, line 36: I'd consider setting this to `tempo` instead. Source imo is about the process generating the metrics and this makes it seem like docker-compose is emitting these metrics.

We could also do:
```
source: tempo
cluster: docker-compose
```

kvrhdn (approved):
Left some comments but non are blocking. LGTM!

Also about exemplars, in the distributed example I configured links from Prometheus to Tempo:

https://github.com/grafana/tempo/blob/e2f7b9c83d78cd55241cf6c1a77cd8fd739a033b/example/docker-compose/distributed/grafana-datasources.yaml#L13-L16

I think we can simply copy this over to the `shared/grafana0-datasources.yaml`.

PR #1371, opened 2022-04-08
Refactor for alternative block formats 3

mdisibio:
**What this PR does**:
This PR is another step towards alternative block formats. See previous PRs #1351 and #1346.   In this PR we really pare down the VersionedEncoding interface and I think it's looking really good.  Adding parquet support should be straightforward now.

List of discussion points:
* Removed all of the primitives from VersionedEncoding like index/data readers/writers.  Instead we have high-level functions like `OpenBlock`, `CreateBlock`, `CopyBlock`.  I think this interface is looking really good.
* Reintroduced an interface `common.BackendBlock` for all of the block read methods.  It just simplified things a lot better.  Maybe we don't need the separate `Finder` or `Searcher` interfaces after all.  
* This is the biggest change:  The core logic to convert WAL -> complete block was moved from `tempodb.CompleteBlockWithBackend` to the encoding's `CreateBlock`.  It is still []byte-based to maintain efficiency for the current proto encoding.  Future encodings will used the passed `ObjectDecoder` to extract a `tempopb.Trace` from the bytes.
  * There is some noise from metadata handling and a few impacted tests that were actually generating bad blocks but it didn't matter.
  * CompleteBlockWithBackend was using the compactor's flush size, but now it's just a 30MB default inside v2 encoding.  Seems fine but open to feedback.
* The WAL and search blocks are fully stuck on v2 encoding, so was able to remove a few struct fields. 
* The tempo-cli parts that were v2-specific were salvaged by checking the block version and failing if anything else. The cli uses the v2 primitives directly to do things like dump the index. This seems better than leaving them in the `VersionedEncoding` interface and returning `errUnsupported` for parquet but open to feedback.

n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/ingester/instance.go
```diff
@@ -517,12 +517,12 @@ func (i *instance) resetHeadBlock() error {
 	i.lastBlockCut = time.Now()
 
 	// Create search data wal file
-	f, version, enc, err := i.writer.WAL().NewFile(i.headBlock.BlockID(), i.instanceID, searchDir)
+	f, _, enc, err := i.writer.WAL().NewFile(i.headBlock.BlockID(), i.instanceID, searchDir)
```
joe-elliott, line 520: looks like nothing else uses the function. just drop the version return value?

tempodb/backend/block_meta.go
```diff
@@ -48,14 +48,19 @@ func NewBlockMeta(tenantID string, blockID uuid.UUID, version string, encoding E
 // ObjectAdded updates the block meta appropriately based on information about an added record
 //  start/end are unix epoch seconds
 func (b *BlockMeta) ObjectAdded(id []byte, start uint32, end uint32) {
-	startTime := time.Unix(int64(start), 0)
-	endTime := time.Unix(int64(end), 0)
 
-	if b.StartTime.IsZero() || startTime.Before(b.StartTime) {
-		b.StartTime = startTime
+	if start > 0 {
```
joe-elliott, line 52: is this a bug in the current implementation?
add a quick test?

tempodb/encoding/v2/create_block.go
```diff
@@ -0,0 +1,107 @@
...
+
+const DefaultFlushSizeBytes int = 30 * 1024 * 1024 // 30 MiB
+
+func CreateBlock(ctx context.Context, cfg *common.BlockConfig, tenantID string, blockID uuid.UUID, encoding backend.Encoding,
+	dataEncoding string, estimatedTotalObjects int, i common.TraceIterator, to backend.Writer) (*backend.BlockMeta, error) {
+	defer i.Close()
+
+	newMeta := backend.NewBlockMeta(tenantID, blockID, VersionString, encoding, model_v2.Encoding)
+
+	dec := model_v2.NewSegmentDecoder()
```
joe-elliott, line 25: should be ok to pin this? i don't think this will cause any upgrade issues 1.3 -> 1.4.

joe-elliott, line 25: should we use `model.CurrentEncoding`? i'm honestly not sure. we've definitely made it quite difficult to add a v3 encoding, but i think that's ok. once a completely new block format hits all this code will be retired.

joe-elliott (commented):
Looks good. Some thoughts and questions in comments.

> CompleteBlockWithBackend was using the compactor's flush size, but now it's just a 30MB default inside v2 encoding. Seems fine but open to feedback.

I'm good.

> The WAL and search blocks are fully stuck on v2 encoding, so was able to remove a few struct fields.

There will likely be a period in which we support reading old style blocks. After that maybe we can just copy all relevant v2 code into the wal package?

I suppose the old search code will likely also be retired with a new block format. So this doesn't feel like an issue.

> This is a step backwards in efficiency because we're no longer copying raw bytes from the wal to the new block. I have some ideas on how to fix this, but mainly gorpy and not worth it imo. Creating parquet blocks won't be able to avoid the work in the future anyway.

This concerns me. we should definitely hold on merging this until after 1.4 is cut, but I'd hate to announce that 1.5 is less performant in anticipation of future features. Maybe that's the best choice though to move this project forward? At what point should we start maintaining a parquet fork for changes like this that are performance regressions. Maybe 1.4 is the last 1.x release? and we march to parquet 2.0 next?

tempodb/tempodb.go
```diff
@@ -202,66 +202,32 @@ func (rw *readerWriter) WriteBlock(ctx context.Context, c WriteableBlock) error
...
+	meta := block.Meta()
 
-		if newBlock.CurrentBufferLength() > int(flushSize) {
-			tracker, _, err = newBlock.FlushBuffer(ctx, tracker, w)
-			if err != nil {
-				return nil, errors.Wrap(err, "error flushing compactor block")
-			}
-		}
-	}
+	// Pinning to v2 here
```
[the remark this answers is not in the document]
mdisibio: Good catch, this function had several mistakes.  Updated to use latest as suggested, but also fixed to use the latest dataEncoding and honor the backend block encoding config instead of reusing the wal's encoding.

tempodb/encoding/v2/create_block.go
mdisibio: Good catch, this function wasn't honoring the passed in `dataEncoding` at all. Fixed.

tempodb/backend/block_meta.go
joe-elliott: > it gets the timestamps from the spans while iterating the TraceIterator, so it can't create an input meta with timestamps

my concern here is that we constrain setting the min/max block times based on:
https://github.com/grafana/tempo/blob/main/modules/storage/config.go#L40

if compaction just uses the span start/stop times then the compacted blocks time ranges will start exceeding this constraint.

mdisibio: Good catch, those were definitely not being preserved.   Ok updated `TraceIterator` to include the trace start/end times, and for `encoding.CompleteBlock` to take and honor them and the block-level start/end times from the wal block.

mdisibio:
> There will likely be a period in which we support reading old style blocks. After that maybe we can just copy all relevant v2 code into the wal package?

Sure, I hadn't given too much thought that far down the line, but that's a good idea for cleanup.

> This concerns me. we should definitely hold on merging this until after 1.4 is cut, but I'd hate to announce that 1.5 is less performant in anticipation of future features. 

Some promising findings:  I was speaking theoretically about the performance regression and hadn't actually measured it.  It looks like it's not a huge impact to the ingester overall.  Here are the results from some local e2e benchmarking:
* Input:  4000 spans/s (the most my system could handle stably)
* Ingesters: 3x in RF3
* Backend: s3
* Max block lifetime: 1m (trying to match how often our internal cluster flushes frequently)
* Everything else defaults (encoding, etc)

Looking at the minimum performance:  1.3.2 was 2.26K, and this PR was 2.20K.  ~3% decrease.   Promising but not conclusive.  

Created a go benchmark and it's as expected: way slower.
```
This PR
BenchmarkCompleteBlock-12    	       1	16685445461 ns/op

1.3.2
BenchmarkCompleteBlock-12    	      12	   95091364 ns/op
```

Screen Shot 2022-04-15 at 8 34 38 AM

tempodb/encoding/versioned.go
```diff
@@ -15,15 +15,26 @@ import (
 type VersionedEncoding interface {
 	Version() string
 
-	NewDataWriter(writer io.Writer, encoding backend.Encoding) (common.DataWriter, error)
-	NewIndexWriter(pageSizeBytes int) common.IndexWriter
+	// OpenBlock for reading
+	OpenBlock(meta *backend.BlockMeta, r backend.Reader) (common.BackendBlock, error)
```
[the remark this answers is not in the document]
mdisibio: I like that it matches with `CreateBlock`, `CopyBlock`, and backend blocks are generally considered immutable and read-only?  But we can rename if strong opinion.

tempodb/encoding/versioned.go
```diff
@@ -15,15 +15,26 @@ import (
...
 
-	NewDataReader(ra backend.ContextReader, encoding backend.Encoding) (common.DataReader, error)
-	NewIndexReader(ra backend.ContextReader, pageSizeBytes int, totalPages int) (common.IndexReader, error)
+	// NewCompactor creates a Compactor that can be used to combine blocks of this
+	// encoding. It is expected to use internal details for efficiency.
+	NewCompactor() common.Compactor
```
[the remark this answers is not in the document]
mdisibio: I lean towards option 1, to only compact blocks of the same encoding, as it is the most performant since encodings can make use of internal details.  Added a note to our implementation checklist.

Also considered option 2, use TraceIterator, because there are similarities between WAL -> complete block and compaction (both are creating a new block with known trace contents).  The flexibility is much better, and allows upgrading of existing blocks to new formats.  However, performance would be reduced for normal compaction which is the majority of work.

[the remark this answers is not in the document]
joe-elliott: +1 to option 1. Currently we don't compact across data encodings:

https://github.com/grafana/tempo/blob/main/tempodb/compaction_block_selector.go#L126 

As a concept "dataEncoding" goes away with parquet blocks. Perhaps it means something else? or we drop the field entirely?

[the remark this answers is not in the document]
mdisibio: Oops we are mixing up terminology which is clouding the discussion.  The compactor needs to select blocks with the same `BlockMeta.Version` and `BlockMeta.DataEncoding`, i.e. v2->v2 and parquet->parquet.  `BlockMeta.Encoding` is actually the compression which doesn't matter.   Agree that the first parquet format won't make use of DataEncoding to start, so we can either leave it as an empty string for further use, or remove it.

tempodb/encoding/common/interfaces.go
```diff
@@ -51,3 +51,17 @@ func DefaultCompactionOptions() CompactionOptions {
 		OutputBlocks:       1,
 	}
 }
+
+type TraceIterator interface {
+	// Next returns the next trace and optionally the start and stop times
+	// for the trace that may have been adjusted.
+	Next(ctx context.Context) (ID, *tempopb.Trace, uint32, uint32, error)
```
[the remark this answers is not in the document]
mdisibio: > move the []byte -> *Trace unmarshalling within the vparquet

To decode the []byte it would also need an `ObjectDecoder`, and move `Iterator` to `encoding/common`.  Agree this should be fine.

> what percentage of WAL traces actually go through the model/combiner

It looks like we don't have good metrics for this, hard to say and likely workload dependent (if spans arrive irregularly or after trace flush).

annanay25 (approved):
LGTM.

PR #1375, opened 2022-04-12
metrics-generator: add override to disable metrics collection

kvrhdn:
**What this PR does**:
Add an override `metrics_generator_disable_collection` which tells the registry to not collect and write samples for this tenant. This will break the link between the registry and the storage layer (which is responsible for remote writing metrics).

The metrics-generator will still generate metrics and keep track of all the counters and histograms. If you don't want to generate metrics at all, you should clear `metrics_generator_processors` for this tenant.

This setting is useful if you want to ingest data into the metrics-generator, observe resource usage and active series, but don't want to write real metrics. This should be especially handy if you are concerned about the generated active series overwhelming a downstream TSDB.

Related to #1303 

**Checklist**
- [x] Tests updated
- [ ] Documentation added, will do in #1328 

### DIFF
--- a/modules/generator/registry/registry.go
+++ b/modules/generator/registry/registry.go
@@ -158,6 +158,10 @@ func (r *ManagedRegistry) onRemoveMetricSeries(count uint32) {
 }
 
 func (r *ManagedRegistry) collectMetrics(ctx context.Context) {
+	if r.overrides.MetricsGeneratorDisableCollection(r.tenant) {
+		return
+	}
+
 	r.metricsMtx.RLock()
 	defer r.metricsMtx.RUnlock()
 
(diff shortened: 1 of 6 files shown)


mapno (commented):
Great idea. Left a comment regarding the approach.

modules/generator/registry/registry.go
```diff
@@ -158,6 +158,10 @@ func (r *ManagedRegistry) onRemoveMetricSeries(count uint32) {
 }
 
 func (r *ManagedRegistry) collectMetrics(ctx context.Context) {
+	if r.overrides.MetricsGeneratorDisableCollection(r.tenant) {
```
joe-elliott, line 161: Is there a reason to do this in the metrics-generator and not the distributor?

joe-elliott (commented):
This PR generally looks fine to me, but I have some questions:

- Why do this in the metrics-generator and not the distributor?
- The distributor already has a per tenant "switch" to send data to the metrics generator. How is this different?

joe-elliott (approved):
Ok, I just read your PR description which answers my questions. :)

PR #1378, opened 2022-04-12
Design Proposal: TraceQL Concepts

joe-elliott:
**What this PR does**:

Adds a design proposal that outlines the concepts and basic syntax of TraceQL, the upcoming Tempo query language. The full specification for this language is still a WIP, but we feel the core concepts are ready to be shared.  We invite anyone to provide comments and criticisms to better create a language that serves the needs of the tracing community!

Special thanks to @yurishkuro, @tomwilkie, @mdisibio and @cyriltovena for helping build this language over the past months.

### DIFF
--- a/docs/design-proposals/2022-04 TraceQL Concepts.md
+++ b/docs/design-proposals/2022-04 TraceQL Concepts.md
@@ -43,7 +43,7 @@ The TraceQL language is designed to use the same syntax and semantics as PromQL 
 A query is an expression that is evaluated on one trace at a time. The query is structured as a set of chained
 expressions (a pipeline). Each expression in the pipeline selects or discards spansets from being included in the results set. E.g.
 
-`{ http.status = 200 } | by(namespace) | count() > 3`
+`{ .http.status = 200 } | by(.namespace) | count() > 3`
 
 Queries select sets of spans and filter them through a pipeline of aggregators and conditions. If a spanset is produced after evaluation
 on a trace then this spanset (and by extension the trace) is included in the resultset of the query.
@@ -52,7 +52,7 @@ on a trace then this spanset (and by extension the trace) is included in the res
 
 In TraceQL the curly brackets `{}` always select a set of spans from the current trace. They will commonly be paired with a condition to reduce the spans being passed in.
 
-`{ http.status = 200 }`
+`{ .http.status = 200 }`
 
 This simple query will be evaluated on every span of every trace, one at a time. 
 
@@ -66,11 +66,9 @@ Every span has certain intrinsic fields that can be referenced when selecting sp
 
 | field name | description                                 |
 | ---------- | ------------------------------------------- |
-| `start`    | start time of the span                      |
-| `end`      | end time of the span                        |
 | `duration` | end - start time of the span                |
 | `name`     | operation or span name                      |
-| `status`   | status values are "error", "ok", or "unset" |
+| `status`   | status values are error, ok, or unset       |
 | `parent`   | the parent of this span                     |
 
 **Examples**
@@ -87,13 +85,21 @@ We can refer to dynamic attributes (also known as tags) on the span or the span'
 **Examples**
 
 Find traces with the GET http method:  
-`{ http.method = "GET" }`
+`{ .http.method = "GET" }`
 
 Find traces that passed through the prod namespace:  
-`{ namespace = "prod" }`
+`{ .namespace = "prod" }`
 
 Find traces that cross service boundaries:  
-`{ parent.service.name != service.name }`
+`{ parent.service.name != .service.name }`
+
+### Scoped attribute fields
+
+Attributes can be specifically scoped to either "span" or "resource". Specifying "span" or "resource"
+can result in significant performance benefits.
+
+`{ span.http.status = 200 }`
+`{ resource.namespace = "prod" }`
 
 ### Field expressions
 
@@ -105,17 +111,17 @@ Find traces where the difference of the duration of a set of parent/child spans 
 `{ parent.duration - duration > 500ms }`
 
 Find traces with "success" http status codes:  
-`{ http.status >= 200 && http.status < 300 }`
+`{ .http.status >= 200 && .http.status < 300 }`
 
 Note that the second expression requires both conditions to be true on the same span. The entire expression inside of `{}` must be evaluated as true on a single span for it to be included in the resultset.
 
 ## Combining Spansets
 
 Logical operators combine sets of spans. For example, if we wanted to find a trace that went through two specific regions:  
-`{ region = "eu-west-0" } && { region = "eu-west-1" } `
+`{ .region = "eu-west-0" } && { .region = "eu-west-1" } `
 
 Note the difference between the above and the following:  
-`{ region = "eu-west-0"  && region = "eu-west-1" } `
+`{ .region = "eu-west-0"  && .region = "eu-west-1" } `
 
 The second expression will return no traces because it's impossible for both conditions to be simultaneously true on the same span.
 
@@ -151,7 +157,7 @@ Pipelining allows us to "pipe" a set of spans from one expression to the next. T
 **Examples**
 
 Find traces that have more than 3 spans with an attribute `http.status` with a value of `200`:  
-`{ http.status = 200 } | count() > 3`
+`{ .http.status = 200 } | count() > 3`
 
 ## Grouping
 
@@ -160,18 +166,18 @@ Grouping allows us to take a trace and break it down into sets of spans that are
 **Examples**
 
 Find traces that have more than 5 spans in any region:  
-`by(region) | count() > 5 `
+`by(.region) | count() > 5 `
 
 ## Examples
 
 Any span matches an attribute:  
-`{ namespace = "prod" }`
+`{ .namespace = "prod" }`
 
 Two attributes appear on the same span:  
-`{ namespace = "prod" && http.status = 200 }`
+`{ .namespace = "prod" && .http.status = 200 }`
 
 Two attributes appear anywhere within a trace:  
-`{ namespace = "prod" } && { http.status = 200 }`
+`{ .namespace = "prod" } && { .http.status = 200 }`
 
 Any span has a duration over one second:  
 `{ duration > 1s }`
@@ -180,24 +186,13 @@ The trace as a whole has a duration of over one second:
 `max(end) - min(start) > 1s`
 
 The average duration of spans exceeds 1s and any span has a specific attribute:
-`avg(duration) > 1s && { namespace = "prod" }`
+`avg(duration) > 1s && { .namespace = "prod" }`
 
 A trace has over 5 spans with http.status = 200 in any given namespace:  
-`{ http.status = 200 } | by(namespace) | count() > 5`
+`{ .http.status = 200 } | by(.namespace) | count() > 5`
 
 A trace passed through two regions in a specific order:  
-`{ region = "eu-west-0" } >> { region = "eu-west-1" }`
+`{ .region = "eu-west-0" } >> { .region = "eu-west-1" }`
 
 A trace sees network latency > 1s when passing between any two services:  
-`{ parent.service.name != service.name } | max(parent.duration - duration) > 1s`
-
-### Shorthand
-
-The following rules are proposed to make using TraceQL easy to use for simple cases.
-
-1. A bare condition is wrapped in `{}`  
-  `http.status = 200` -> `{ http.status = 200 }`
-
-2. Multiple bare conditions are individually wrapped in `{}`    
-  `http.status = 200 && namespace = "prod"` ->
-   `{ http.status = 200 } && { namespace = "prod" }`
+`{ parent.service.name != .service.name } | max(parent.duration - duration) > 1s`
--- a/docs/design-proposals/2022-04 TraceQL Concepts.md
+++ b/docs/design-proposals/2022-04 TraceQL Concepts.md
@@ -1,0 +1,204 @@
+---
+Authors: Yuri Shkuro (@yurishkuro), Tom Wilkie (@tomwilkie), Cyril Tovena (@cyriltovena), Martin Disibio (@mdisibio), Joe Elliott (@joe-elliott)
+Created: 2021 October - 2022 March
+Last updated: 2022-04-12
+---
+
+# TraceQL Concepts
+
+## Summary
+
+This design document describes a language for selecting traces that Tempo will implement. This language is currently only focused on trace selection. It will be extended to derive metrics from traces in the future.
+
+This document is **not** meant to be a complete language specification. Rather it is meant to invite the community to comment on the basic structure and concepts of the language.
+
+### Table of Contents
+
+- [Capabilities](#capabilities)
+- [Structure](#structure)
+- [Selecting spans](#selecting-spans)
+  - [Intrinsic fields](#intrinsic-fields)
+  - [Attribute fields](#attribute-fields)
+  - [Field expressions](#field-expressions)
+- [Combining Spansets](#combining-spansets)
+  - [Structural Operators](#structural-operators)
+- [Aggregators](#aggregators)
+- [Pipelining](#pipelining)
+- [Grouping](#grouping)
+- [Examples](#examples)
+  - [Shorthand](#shorthand)
+
+## Capabilities
+
+A TraceQL query can select traces based on 
+
+- span attributes, timing and duration
+- structural relationships between spans
+- aggregated data from the spans in a trace
+
+The TraceQL language is designed to use the same syntax and semantics as PromQL and LogQL where possible. The syntax and semantics differ due to the specialized needs of querying traces.
+
+## Structure
+
+A query is an expression that is evaluated on one trace at a time, and iterates, evaluating each trace. The expression is evaluated on each trace as a whole, one at a time. If the evaluation produces a set of spans, then the trace and the set of spans is part of the result set.
+
+The structure of a query is an expression or set of chained expressions (a pipeline). Chained expressions are evaluated left to right.  Each expression selects or discards this trace from being included in the results set.
+
+## Selecting spans
+
+In TraceQL the curly brackets `{}` always select a set of spans from the current trace. They will commonly be paired with a condition to reduce the spans being passed in.
+
+`{ http.status = 200 }`
+
+This simple query will be evaluated on every span of every trace, one at a time. 
+
+If the trace being evaluated contains *no* spans with an attribute`http.status` with the value `200` then no spans will be selected and this trace will not appear in our resultset.
+If the trace does contain spans with an attribute`http.status` with the value `200` then *only* those spans will be returned. The trace is reduced to only the set of spans that match the condition inside the `{}`. This set of spans (and therefore trace) will be included in our resultset.
+
+### Intrinsic fields
+
+Every span has certain intrinsic fields that can be referenced when selecting spans
+
+| field name | description |
+| —--------- | ----------- |
+| `start`      | start time of the span |
+| `end`        | end time of the span |
+| `duration`   | end - start time of the span |
+| `name`       | operation or span name |
+| `status`     | status values are "error", "ok", or "unset" |
+| `parent`     | the parent of this span | 
+
+**Examples**
+Find traces that contain spans whose duration is greater than 2 seconds:
+`{ duration > 2s }`
+
+Find traces that contain a span named "HTTP POST":
+`{ name = "HTTP POST"  }`
+
+Find traces that cross service boundaries:
+`{ parent.service.name != service.name }`
+
+### Attribute fields
+
+We can refer to dynamic attributes (also known as tags) on the span or the span's resource.
+
+** Examples **
+
+Find traces with the GET http method:
+`{ http.method = "GET" }`
+
+Find traces that passed through the prod namespace:
+`{ namespace = "prod" }`
+
+### Field expressions
+
+Fields can also be combined in various expected ways.
+
+** Examples **
+
+Find traces where the difference of the duration of a set of parent/child spans exceeds 500ms.
+`{ parent.duration - duration > 500ms }`
+
+Find traces with "success" http status codes:
+`{ http.status >= 200 && http.status < 300 }`
+
+Note that the second expression requires both conditions to be true on the same span. The entire expression inside of `{}` must be evaluated as true on a single span for it to be included in the resultset.
+
+## Combining Spansets
+
+Logical operators combine sets of spans. For example, if we wanted to find a trace that went through two specific regions:
+
+`{ region = "eu-west-0" } && { region = "eu-west-1" } `
+
+Note the difference between the above and the following. 
+
+`{ region = "eu-west-0"  && region = "eu-west-1" } `
+
+The second expression will return no traces because it's impossible for both conditions to be simultaneously true on the same span.
+
+### Structural operators
+
+Structural operators evaluate traces based on the span relationships within the span tree. 
+
+`{ } >> { }`
+  This is the descendant operator. The spans returned from this operator will match the right hand side conditions while also being descendants of spans that match the left hand side conditions.
+
+`{ } > { }`
+  This is the child operator. The spans returned from this operator will match the right hand side conditions while also being children of spans that match the left hand side conditions.
+
+`{ } ~ { }`
+  This is the sibling operator. The spans returned from this operator will match the right hand side conditions while also being siblings of spans that match the left hand side conditions.
+
+## Aggregators
+
+All of the above expressions involve asking questions about individual spans. However, sometimes we want to ask questions about a set of spans. For that we have aggregate functions.
+
+** Examples **
+
+Find traces where the total number of spans is greater than 10:
+
+`count() > 10`
+
+Find traces where the average duration is greater than 1s:
+
+`avg(duration) > 1s`
+
+## Expression Pipelining
+
+Pipelining allows us to "pipe" a set of spans from one expression to the next. This is particularly useful if you want to perform an aggregate over a subset of a trace.
+
+** Examples **
+
+Find traces where that have more than 3 spans with an attribute`http.status` with a value of `200`:
+
+`{ http.status = 200 } | count() > 3`
+
+## Grouping
+
+Grouping allows us to take a trace and break it down into sets of spans that are evaluated by future pipeline entries. Each set of spans created by the group is _individually_ evaluated by downstream expressions.
+
+** Examples **
+
+Find traces that have more than 5 spans in any region:
+
+`by(region) | count() > 5 `
+
+## Examples
+
+Any span matches an attribute
+`{ namespace = "prod" }`
+
+Two attributes appear on the same span
+`{ namespace = "prod" && http.status = 200 }`
+
+Two attributes appear anywhere within a trace. 
+`{ namespace = "prod" } && { http.status = 200 }`
+
+Any span has a duration over one second.
+`{ duration > 1s }`
+
+The trace as a whole has a duration of over one second.
+`max(end) - min(start) > 1s`
+
+Any span has a duration of over one second and any span has a specific attribute. These spans can be the same or different.
+`max(duration) > 1s && { namespace = "prod" }`
+
+A trace has over 5 spans with http.status = 200 in any given namespace.
+`by(namespace) | { http.status = 200 } | count() > 5`
+
+A trace passed through two regions in a specific order.
+`{ region = "eu-west-0" } >> { region = "eu-west-1" }`
+
+A trace sees network latency > 1s when passing between any two services.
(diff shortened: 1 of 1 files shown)


docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,198 @@
...
+
+Find traces where the difference of the duration of a set of parent/child spans exceeds 500ms:  
+`{ parent.duration - duration > 500ms }`
+
+Find traces with "success" http status codes:  
+`{ http.status >= 200 && http.status < 300 }`
+
+Note that the second expression requires both conditions to be true on the same span. The entire expression inside of `{}` must be evaluated as true on a single span for it to be included in the resultset.
+
+## Combining Spansets
```
yurishkuro, line 107: maybe add a section introducing the result set data model and define `spanset` there

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Selecting spans
+
+In TraceQL the curly brackets `{}` always select a set of spans from the current trace. They will commonly be paired with a condition to reduce the spans being passed in.
+
+`{ http.status = 200 }`
+
+This simple query will be evaluated on every span of every trace, one at a time. 
+
+If the trace being evaluated contains *no* spans with an attribute`http.status` with the value `200` then no spans will be selected and this trace will not appear in our resultset.
```
annanay25, line 55: nit
```suggestion
If the trace being evaluated contains *no* spans with an attribute `http.status` with the value `200` then no spans will be selected and this trace will not appear in our resultset.
```

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,198 @@
...
+- structural relationships between spans
+- aggregated data from the spans in a trace
+
+The TraceQL language is designed to use the same syntax and semantics as PromQL and LogQL where possible. The syntax and semantics differ due to the specialized needs of querying traces.
+
+## Structure
+
+A query is an expression that is evaluated on one trace at a time, and iterates, evaluating each trace. The expression is evaluated on each trace as a whole, one at a time. If the evaluation produces a set of spans, then the trace and the set of spans is part of the result set.
+
+The structure of a query is an expression or set of chained expressions (a pipeline). Chained expressions are evaluated left to right.  Each expression selects or discards this trace from being included in the results set.
```
annanay25, line 45: There are a few occurrences of  "resultset"  in this document. What does that signify? Is it the set of results displayed in the search result table? Why does it contains only those spans that match the conditions and not the entire trace?

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,198 @@
...
+`max(end) - min(start) > 1s`
+
+Any span has a duration of over one second and any span has a specific attribute. These spans can be the same or different:  
+`max(duration) > 1s && { namespace = "prod" }`
+
+A trace has over 5 spans with http.status = 200 in any given namespace:  
+`by(namespace) | { http.status = 200 } | count() > 5`
+
+A trace passed through two regions in a specific order:  
+`{ region = "eu-west-0" } >> { region = "eu-west-1" }`
```
annanay25, line 184: Nice. This seems very powerful.

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,198 @@
...
+Find traces that have more than 5 spans in any region:  
+`by(region) | count() > 5 `
+
+## Examples
+
+Any span matches an attribute:  
+`{ namespace = "prod" }`
+
+Two attributes appear on the same span:  
+`{ namespace = "prod" && http.status = 200 }`
```
annanay25, line 166: Hmm, I wonder if we need the `&&` operator for multiple conditions on the same span. Wonder if we could use `,` that would be more aligned with PromQL/LogQL:

`{ namespace = "prod", http.status = 200 }`

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,198 @@
...
+`{ namespace = "prod" } && { http.status = 200 }`
+
+Any span has a duration over one second:  
+`{ duration > 1s }`
+
+The trace as a whole has a duration of over one second:  
+`max(end) - min(start) > 1s`
+
+Any span has a duration of over one second and any span has a specific attribute. These spans can be the same or different:  
+`max(duration) > 1s && { namespace = "prod" }`
```
annanay25, line 178: How would we represent aggregators and conditions on the same span?

`{ max(duration) > 1s &&  service.name = "querier" }` ?

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,198 @@
...
+
+**Examples**
+Find traces that contain spans whose duration is greater than 2 seconds:  
+`{ duration > 2s }`
+
+Find traces that contain a span named "HTTP POST":  
+`{ name = "HTTP POST"  }`
+
+Find traces that cross service boundaries:  
+`{ parent.service.name != service.name }`
```
pavolloffay, line 79: The service is not defined in the `Intrinsic fields`. 

How are the resource attributes queried? Should there be `resource`  defined in  `Intrinsic fields` ?

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Attribute fields
+
+We can refer to dynamic attributes (also known as tags) on the span or the span's resource.
```
pavolloffay, line 83: Does this imply that there is no way to distinguish between resource attributes and span attributes?

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,198 @@
...
+`{ duration > 1s }`
+
+The trace as a whole has a duration of over one second:  
+`max(end) - min(start) > 1s`
+
+Any span has a duration of over one second and any span has a specific attribute. These spans can be the same or different:  
+`max(duration) > 1s && { namespace = "prod" }`
+
+A trace has over 5 spans with http.status = 200 in any given namespace:  
+`by(namespace) | { http.status = 200 } | count() > 5`
```
mdisibio, line 181:
```suggestion
`{ http.status = 200 } | by(namespace) | count() > 5`
```
How about switching the order - reduce then group.  This seems like it should be more efficient, but also is clearer conceptually.

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,198 @@
...
+`{ namespace = "prod" && http.status = 200 }`
+
+Two attributes appear anywhere within a trace:  
+`{ namespace = "prod" } && { http.status = 200 }`
+
+Any span has a duration over one second:  
+`{ duration > 1s }`
+
+The trace as a whole has a duration of over one second:  
+`max(end) - min(start) > 1s`
```
mdisibio, line 175: I expect Tempo to store the final trace duration determined exactly the same way, so could this simplified?  Thinking that this is an inefficient precedence to set, as it requires scanning all spans for two values instead of the single value per trace.
```suggestion
`duration > 1s`
```

I realize this kind of conflicts with the span-level duration, so maybe another intrinsic property like 'totalDuration`.

docs/design-proposals/2022-04 TraceQL Concepts.md
mdisibio, line 178: I think this is the same query, but doesn't imply that we have to scan all spans and aggregate a value.  If we want to show an example of `max(attribute)`, then maybe a different example is better.  

```suggestion
`duration > 1s && { namespace = "prod" }`
```

joe-elliott: `{ service.name = "querier" && duration > 1s }`

since these conditions are evaluated on spans individually there is nothing really to aggregate.

joe-elliott: Good point. Will change to avg.

docs/design-proposals/2022-04 TraceQL Concepts.md
joe-elliott: There is currently no way to reference the "trace" as a thing in traceql. All of the operations are selecting and filtering sets of spans. There are a few "trace level" attributes that I wonder if it would make sense to include or not, but we would have to think quite carefully about we would integrate it into our current "pipeline of spansets" approach.

Examples of trace level attributes
- duration
- rootSpanName
- ??

To help with the second one we have considered adding a "root" intrinsic so you could ask something like:

`{ name = "do the thing" && root = true }`

but that's still up in the air.

docs/design-proposals/2022-04 TraceQL Concepts.md
joe-elliott: In our internal doc we have suggested using a prefix like the below to explicitly select resource or span attributes. If "span" or "resource" is not present then we would look on both the resource and the span. The exact details here are still under  discussion. Input is appreciated.

`{ span.key = "val" }`
`{ resource.key = "val" }`

I kind of hate the term "resource" and prefer the Jaeger terminology "process", but we will likely align with OTel on naming conventions.

docs/design-proposals/2022-04 TraceQL Concepts.md
joe-elliott: That's fair. I would like to support a variety of operators here so I prefer the `&&`

`{ namespace = "prod" || http.status = 200 }`

of course that's equivalent to: 

`{ namespace = "prod" } || { http.status = 200 }`

so perhaps comma makes sense? interested in others' thoughts.

yurishkuro: -1 on comma, it's an additional mental barrier while `&&` is universally understood

docs/design-proposals/2022-04 TraceQL Concepts.md
joe-elliott: I will add some clarification on spanset/resultset. (Requested by Yuri as well above). I really like only returning those spans that match the criteria specified in the pipeline. If you were to write

`{ http.status = 200 }`

and only returned the matching spans. Then a UI could provide links to both the matching trace as a whole as well as deep links to the few spans that matched the requested criteria.

docs/design-proposals/2022-04 TraceQL Concepts.md
joe-elliott: Good point. I am going to reword the "structure" section above to better highlight how integral spansets are to the language and add some details about the resultset.

till:
Does your AGPL apply to this language as well?

docs/design-proposals/2022-04 TraceQL Concepts.md
mdisibio: > If "span" or "resource" is not present then we would look on both the resource and the span

I would love the ability to specify span-level or resource-level, if no reason other than improved search times and efficiency.  From my standpoint, I think we're waiting for the right syntax.   `{ key=val }` is great for usability but not precise... And ` {span.key=val}` feels wordy, and could conflict if another attribute is named `span.key`.

joe-elliott:
> Does your AGPL apply to this language as well?

@till This is a good question. There are at least few components here: the language spec itself, documents such as this that describe the language and the implementation of the language. I am working with people internally and will provide you with an honest answer as soon as I can.

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,203 @@
...
+`{ parent.service.name != service.name } | max(parent.duration - duration) > 1s`
+
+### Shorthand
+
+The following rules are proposed to make using TraceQL easy to use for simple cases.
+
+1. A bare condition is wrapped in `{}`  
+  `http.status = 200` -> `{ http.status = 200 }`
+
+2. Multiple bare conditions are individually wrapped in `{}`
```
electron0zero, line 201: Like the idea of shorthand but I also think that shorthand can become source of confusion.

We have `&&` in field expressions where we can write a query like `{ http.status >= 200 && http.status < 300 }`, and `&&` is also being used to write a query like `{ namespace = "prod" } && { http.status = 200 }`

let's say I write two queries
1. `{ namespace = "prod" } && { http.status = 200 }`
2. `{ namespace = "prod" &&  http.status = 200 }`

are these two queries same?

yurishkuro: they are not the same. (1) selects traces where _some_ spans match 1st and 2nd condition independently. (2) selects traces where both conditions hold on the same spans.

electron0zero: I see, thanks.

In that case, I believe suggested shorthand 2 with multiple bare conditions might lead to confusion / unexpected results.

If a user writes `http.status = 200 && namespace = "prod"`, they might expect it return results for `{ http.status = 200 && namespace = "prod" }` (whole query gets wrapped into `{}`, like how it happend for shorthand 1)

But it will return results for `{ http.status = 200 } && { namespace = "prod" }` (wrapping each part into `{}`, as sugessted into shorthand 2)

One way to avoid this is to always wrap whole query into `{}` (only keep shorthand 1) for both cases.

metalmatze:
Hey, very nice work! 
Looking at the provided example labels like `http.status` reminds me more of hierarchical structures than Prometheus labels. Is this a limitation that came from the existing labels in OpenTracing/OpenTelemtery? Is this a limitation to begin with? Does the dot mean anything? 
What would this look like with grpc for example? Will the `grpc.method` and `grpc.service` show up in the spans and then be able to be queried just the same?

docs/design-proposals/2022-04 TraceQL Concepts.md
joe-elliott: I very flexible on the shorthand idea. I mainly wanted the user to type something very quick, easy and intuitive and get an meaningful response. Thanks for this feedback!

joe-elliott:
@metalmatze Good questions.

> Looking at the provided example labels like http.status reminds me more of hierarchical structures than Prometheus labels. Is this a limitation that came from the existing labels in OpenTracing/OpenTelemtery? Is this a limitation to begin with? Does the dot mean anything?

These examples include `http.status` simply b/c it is listed in the [OpenTelemetry semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions) and is easily understood. These are really arbitrary attribute key/value pairs that appear on your spans. The dot is only meaningful in terms of the linked conventions where it kind of "namespaces" http properties.

> What would this look like with grpc for example? Will the grpc.method and grpc.service show up in the spans and then be able to be queried just the same?

Since these tags are arbitrary it can look like whatever you want for grpc. You are welcome to name the attributes anything you'd like. However, like http, there are some published conventions for [rpc as well](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md).

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,203 @@
...
+| `duration` | end - start time of the span |
+| `name`     | operation or span name |
+| `status`   | status values are "error", "ok", or "unset" |
+| `parent`   | the parent of this span | 
+
+**Examples**
+Find traces that contain spans whose duration is greater than 2 seconds:  
+`{ duration > 2s }`
+
+Find traces that contain a span named "HTTP POST":
```
devrimdemiroz, line 80: So far I have seen only these "contain" wording. I searched for:
- wildcard ,  *
- regex , regular, expression
- starts with, ends with
- match
- like , %

I would presume a regex usage as already exist in promql. Is it addressed or intentionally excluded?

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,203 @@
...
+
+`{ } > { }`  
+  This is the child operator. The spans returned from this operator will match the right hand side conditions while also being children of spans that match the left hand side conditions.
+
+`{ } ~ { }`  
+  This is the sibling operator. The spans returned from this operator will match the right hand side conditions while also being siblings of spans that match the left hand side conditions.
+
+## Aggregators
+
+All of the above expressions involve asking questions about individual spans. However, sometimes we want to ask questions about a set of spans. For that we have aggregate functions.
```
devrimdemiroz, line 137: Aggregation functionality is something I am missing from APM world as traces/spans seem to be burst data flow. However, when it comes to usage scenarios, it was not about always about attributes of captured span/method/class/etc. In an HTTP scenario, the aggregation key can be query parameters like '?&operation=getItems' or an attribute inside json post body. For even more complex ones, one might need to skip the root or/and a few spans and extract key aggregation parameter in a couple of levels below. 

The good news is, you generally do not need such aggregation for categorization over all nodes transaction visits, rather it gets generally needed on the very first contact point of an application or system.

I do not think  TraceQL should cover all such aspects at once. Just sharing my previous needs in the last decade regarding aggregation.

docs/design-proposals/2022-04 TraceQL Concepts.md
joe-elliott: Yes, we intend to support regex using the same syntax as promql.

`name =~ ".*test.*"`

docs/design-proposals/2022-04 TraceQL Concepts.md
joe-elliott: Thanks for the thoughts. Since TraceQL can operate on custom attributes you could extract query params and add them as attributes to the span. Then you would be able to aggregate on them easily. 

We did our best to cover the use cases we found the most important, but of course we will find users with new ideas and needs as the project grows. After we are all using TraceQL please keep bringing these ideas back to the table with real world examples so we can continue to improve the language.

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,203 @@
...
+
+If the trace being evaluated contains no spans with an attribute `http.status` with the value `200` then no spans will be selected and this trace will not appear in our resultset.
+
+If the trace does contain spans with an attribute `http.status` with the value `200` then only those spans will be returned. The trace is reduced to only the set of spans that match the condition inside the `{}`. This set of spans (and therefore trace) will be included in our resultset.
+
+### Intrinsic fields
+
+Every span has certain intrinsic fields that can be referenced when selecting spans
+
+| field name | description |
```
jpkrohling, line 67: This table deserves to be properly formatted :-)

```
| field name | description                                 |
| ---------- | ------------------------------------------- |
| `start`    | start time of the span                      |
| `end`      | end time of the span                        |
| `duration` | end - start time of the span                |
| `name`     | operation or span name                      |
| `status`   | status values are "error", "ok", or "unset" |
| `parent`   | the parent of this span                     |
```

docs/design-proposals/2022-04 TraceQL Concepts.md
jpkrohling: Note: there's a new OTEP for a new "scope attribute".

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,203 @@
...
+`avg(duration) > 1s`
+
+## Expression Pipelining
+
+Pipelining allows us to "pipe" a set of spans from one expression to the next. This is particularly useful if you want to perform an aggregate over a subset of a trace.
+
+**Examples**
+
+Find traces that have more than 3 spans with an attribute `http.status` with a value of `200`:  
+`{ http.status = 200 } | count() > 3`
```
jpkrohling, line 154: That answers my question about the mysql example :-) I'll leave it there, just to confirm that this was indeed the logical outcome.

docs/design-proposals/2022-04 TraceQL Concepts.md
jpkrohling: Trace duration is absolutely an attribute that people would query often. Perhaps having a `root` level would be already sufficient in terms of performance? The trace duration would then be `max(end) - root.start`?

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,203 @@
...
+Every span has certain intrinsic fields that can be referenced when selecting spans
+
+| field name | description |
+| ---------- | ----------- |
+| `start`    | start time of the span |
+| `end`      | end time of the span |
+| `duration` | end - start time of the span |
+| `name`     | operation or span name |
+| `status`   | status values are "error", "ok", or "unset" |
+| `parent`   | the parent of this span |
```
mdisibio, line 74: The `parent` intrinsic I feel has some hidden complexity and wonder if it is a worthwhile element of the language.  A basic spanset operator such as `{ A && B}` is evaluated on an individual span which is a straightforward concept and allows for simple storage mechanisms and algorithms.  However `{ parent.A && B }` evaluates across multiple spans.  The implied downstream implementation must now theoretically have both spans present in-memory and evaluated at the same time.  For a worst case example consider the possibility of `parent.parent.parent...`.


The `parent` intrinsic could be replaced with the existing child operator:  `{ parent.A && B }` is equal to ` {A} > {B}`, which is more consistent with the rest of the language, and also the implementation is clear.

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,203 @@
...
+  This is the sibling operator. The spans returned from this operator will match the right hand side conditions while also being siblings of spans that match the left hand side conditions.
+
+## Aggregators
+
+All of the above expressions involve asking questions about individual spans. However, sometimes we want to ask questions about a set of spans. For that we have aggregate functions.
+
+**Examples**
+
+Find traces where the total number of spans is greater than 10:  
+`count() > 10`
```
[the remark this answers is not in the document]
joe-elliott: as you discovered below `{ peer.name = "mysql" } | count() > 10`. also the otel convention is `db.system` i believe if you were specifically referencing that as well:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#connection-level-attributes

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Combining Spansets
+
+Logical operators combine sets of spans. For example, if we wanted to find a trace that went through two specific regions:  
+`{ region = "eu-west-0" } && { region = "eu-west-1" } `
+
+Note the difference between the above and the following:  
+`{ region = "eu-west-0"  && region = "eu-west-1" } `
+
+The second expression will return no traces because it's impossible for both conditions to be simultaneously true on the same span.
```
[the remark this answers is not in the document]
joe-elliott: This is a good point I had no considered. Technically the proto does not prevent it, but the specification requires that for a set of attributes the key name is unique: 

https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/common#attribute

So span attributes _should_ be unique, but technically that attribute could exist on both the resource and the span.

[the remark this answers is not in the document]
mdisibio: I agree the wire-format may allow it, but the spec does [require uniqueness]( https://opentelemetry.io/docs/reference/specification/common/common/#attribute-collections) if I am reading it right.

docs/design-proposals/2022-04 TraceQL Concepts.md
joe-elliott: I do like parent for expressions such as:
`{ parent.duration - duration > 1s }` 
or 
`{ parent.service.name != service.name }`

Also, I"m fine with restricting parent to only being allowed to have depth of 1. Let's dig into the details as we work toward an internal POC. This is still an evolving spec and we can make adjustments.

docs/design-proposals/2022-04 TraceQL Concepts.md
mdisibio: > of course that's equivalent to
@joe-elliott Technically those are not equivalent, maybe edit this comment to prevent any confusion?

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,203 @@
...
+- span attributes, timing and duration
+- structural relationships between spans
+- aggregated data from the spans in a trace
+
+The TraceQL language is designed to use the same syntax and semantics as PromQL and LogQL where possible. The syntax and semantics differ due to the specialized needs of querying traces.
+
+## Structure
+
+A query is an expression that is evaluated on one trace at a time. The query is structured as a set of chained
+expressions (a pipeline). Each expression in the pipeline selects or discards spansets from being included in the results set. E.g.
```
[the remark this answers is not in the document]
joe-elliott: Yup. When you write a traceql expression you are building a pipeline of operations on a spanset. A trace enters the pipeline whole and if any subset of it "survives" the filters and selectors applied in the pipeline then it is considered part of the resultset.

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,203 @@
...
+
+In TraceQL the curly brackets `{}` always select a set of spans from the current trace. They will commonly be paired with a condition to reduce the spans being passed in.
+
+`{ http.status = 200 }`
+
+This simple query will be evaluated on every span of every trace, one at a time. 
+
+If the trace being evaluated contains no spans with an attribute `http.status` with the value `200` then no spans will be selected and this trace will not appear in our resultset.
+
+If the trace does contain spans with an attribute `http.status` with the value `200` then only those spans will be returned. The trace is reduced to only the set of spans that match the condition inside the `{}`. This set of spans (and therefore trace) will be included in our resultset.
```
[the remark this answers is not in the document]
joe-elliott: That is correct. In the general case the results of a traceql expression will be a collection of spansets. Each spanset will be a subset of spans from a given trace that "survive" the pipeline. I will add a bit more detail here. Thanks

docs/design-proposals/2022-04 TraceQL Concepts.md
yurishkuro: @mdisibio your complexity argument doesn't sit well with me. The whole point of a dedicated trace QL to me is its ability to express queries across spans of the trace; if it didn't do that I could just as well stick with SQL. So any sensible implementation of TraceQL would have to deal with conditions not defined on individual spans, this complexity is unavoidable.

The language purity argument makes sense, I think it's a matter of how difficult it is to express the same clauses that Joe showed above using the `{A} > {B}` notation.

docs/design-proposals/2022-04 TraceQL Concepts.md
joe-elliott: :)

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,203 @@
...
+The second expression will return no traces because it's impossible for both conditions to be simultaneously true on the same span.
+
+### Structural operators
+
+Structural operators evaluate traces based on the span relationships within the span tree. 
+
+`{ } >> { }`  
+  This is the descendant operator. The spans returned from this operator will match the right hand side conditions while also being descendants of spans that match the left hand side conditions.
+
+`{ } > { }`
```
mmanciop, line 129: While `>` for direct parent-child relation looks intuitive when thinking of spans, I can imagine that overloading the `>` character will confuse users with less familiarity with the trace model.

docs/design-proposals/2022-04 TraceQL Concepts.md
```diff
@@ -0,0 +1,203 @@
...
+
+`{ http.status = 200 }`
+
+This simple query will be evaluated on every span of every trace, one at a time. 
+
+If the trace being evaluated contains no spans with an attribute `http.status` with the value `200` then no spans will be selected and this trace will not appear in our resultset.
+
+If the trace does contain spans with an attribute `http.status` with the value `200` then only those spans will be returned. The trace is reduced to only the set of spans that match the condition inside the `{}`. The resultset will contain only this subset of spans matching the condition.
+
+### Intrinsic fields
```
kvrhdn, line 63: I think span kind is missing from this list. Span kind should be especially useful with the descendant operator.

joe-elliott:
Updated with the latest
- Shorthand removed.
- Updated all attributes to use the latest syntax. 
- Added a section on resource vs span attributes
- Removed start/end intrinsics.

joe-elliott:
@till Apologies for the late response. After discussing this internally we have decided that all code related to TraceQL will be licensed under AGPLv3.

joe-elliott:
Here we go!

mdisibio (approved):

Issue #1374, opened 2022-04-12
msg="failed to cut traces" err="snappy: Writer is closed"


jvilhuber:
**Describe the bug**
After some time, the ingesters seem to lose connection to something, and I start seeing 100+/sec of these errors:
```level=error ts=2022-04-12T03:33:13.326708706Z caller=flush.go:155 org_id=single-tenant msg="failed to cut traces" err="snappy: Writer is closed"```

When I look at the "Tempo Operational" Dashboard (can't remember where I found that), I see that tempo claims to still be functioning normally (at least nothing jumps out at me; see image in comment below): Traces created shows a normal graph. And yet searching for traces by ID in tempo results in "not found" for most traces. The only symptom I noticed that might indicate that things are not working as intended are panels related to blocks like "Blocks Flushed" and "Blocks Cleared" which remain at 0: No activity.

When I restart the ingesters, the errors disappear for a while, searches for traces are successful again, and the blocks-panels show activity.

I'm not sure what else to look for, so please ask and I'll see what I can gather.

**To Reproduce**
I wish I knew. Let tempo in our deployment run for a while and this starts happening.

**Expected behavior**
If some connection problem occurs, ingesters should close the connection and reconnect. An error might be useful to alert us to the connection problem source, so it can be fixed.

**Environment:**
tempoVersion='1.3.2'

 - Infrastructure: Kubernetes, microservices
 - Deployment tool: manual jsonnet, derived from the helm output.

**Additional Context**
Compactor pods: 10, 10Gigs Ram
Distributor pods: 8, 2gigs Ram
Ingester pods (stateful set): 10, 7Gigs Ram, 5Gigs Storage via PVC
Querier pods: 3, 1.5Gigs Ram
Query Frontends: 2, 2Gigs Ram

```apiVersion: v1
data:
  config.yaml: |-
    "compactor":
      "compaction":
        "block_retention": "144h"
        "chunk_size_bytes": 10485760
      "ring":
        "kvstore":
          "store": "memberlist"
    "distributor":
      "receivers":
        "jaeger":
          "protocols":
            "grpc": null
            "thrift_binary": null
            "thrift_compact": null
            "thrift_http": null
        "opencensus": null
        "otlp":
          "protocols":
            "grpc": null
            "http": null
        "zipkin": null
    "http_api_prefix": ""
    "ingester":
      "lifecycler":
        "ring":
          "kvstore":
            "store": "memberlist"
          "replication_factor": 3
[...]

PR #1379, opened 2022-04-13
Fix closed writer panic by resetting buffers/writer correctly on error path

annanay25:

**What this PR does**:
Refer to a longer explanation in this comment: https://github.com/grafana/tempo/issues/1374#issuecomment-1096358727

There is a small section of code in our data writer where the compression writer is closed but not reset. If there are any errors like writing to disk / transient errors with kubernetes pvcs, the function could return without resetting the compression writer. This PR changes that to put the reset logic in a defer func so it triggers whenever the return is triggered.

I understand that working with defer funcs in this manner makes it tricky to follow the logic, so I'm open to alternate implementation methods.

Edit: Offline discussions confirmed my fears as mentioned above :) and so went ahead with a different strategy.

cc @jvilhuber

Fixes the red herring error message in https://github.com/grafana/tempo/issues/1374. But we still need to fix the fact that we drop traces without warning when there are issues writing to disk.

**Checklist**
- N/A Tests updated
- N/A Documentation added


annanay25:
Nope theres no panic. We just want to make sure ingesters function normally after a transient failure like disk full.

joe-elliott:
Frankly if the disk is full it might make sense to actually panic. This is a pretty dire failure mode for Tempo as ingested data will be lost. 

With this change how will the ingester behave on disk full?

annanay25:
Ok. In addition to this cleanup does it then make sense to panic if there's a error in these lines:
https://github.com/grafana/tempo/pull/1379/files#diff-5925637787bab0225dab0a4ab2bcf81bfdd3a6c841d051759b11eae011e37067R75-R78

I believe that should do the trick, WDYT?

joe-elliott (approved):
Thanks for the updates! Looks good.

PR #1382, opened 2022-04-13
Begin adding AWS serverless deployment to drone

zalegrala:
**What this PR does**:

Here we include changes to push the lambda artifacts to s3 as part of a serverless deployment.

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -60,6 +60,17 @@ local gcp_serverless_deployments = [
 local aws_dev_access_key_id = secret('AWS_ACCESS_KEY_ID-dev', 'secret/tempo-dev/aws-credentials-drone', 'access_key_id');
 local aws_dev_secret_access_key = secret('AWS_SECRET_ACCESS_KEY-dev', 'secret/tempo-dev/aws-credentials-drone', 'secret_access_key');
 
+
+local aws_serverless_deployments = [
+  {
+    env: 'dev',
+    bucket: 'tempo-dev-fn-source',
+    access_key_id: aws_dev_access_key_id.name,
+    secret_access_key: aws_dev_secret_access_key.name,
+  },
+];
+
+
 //# Steps ##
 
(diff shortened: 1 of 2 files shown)


.drone/drone.jsonnet
```diff
@@ -36,58 +36,70 @@ local docker_username_secret = secret('docker_username', 'infra/data/ci/docker_h
...
   },
 ];
 
-## Steps ##
+local aws_serverless_deployments = [
+  {
+    local cluster = 'dev-us-east-0',
+    local namespace = 'tempo-dev-02',
+    bucket: '%s-%s-fn-source' % [cluster, namespace],
+    secrets: [
```
joe-elliott, line 65: i don't see why this is a slice. should this not be an object of some kind:

```
{
  access_key: <>,
  secret_key: <>
}
```

Technically this could work as a slice I guess, but these are known/named fields so an object would probably be a clearer way to implement this.

.drone/drone.yml
```diff
@@ -251,6 +251,17 @@ steps:
     ops_tools_fn_upload:
       from_secret: ops_tools_fn_upload
 
+- name: deploy-tempo-serverless-lambda
+  image: amazon/aws-cli
+  commands:
+  - cd ./cmd/tempo-serverless/lambda
+  - aws s3 cp tempo-serverless*.zip s3://dev-us-east-0-tempo-dev-02-fn-source
+  environment:
+    AWS_ACCESS_KEY_ID-tempo-dev-02:
```
joe-elliott, line 260: will this work? doesn't this just need to be `AWS_ACCESS_KEY_ID` ?

joe-elliott (commented):
Two questions/concerns.

Also, do you have some kind of formatter running? It has made a lot of changes to this file that make this PR more difficult to review.

.drone/drone.yml
zalegrala: Ah yes, thank you.

zalegrala: Updated with some help from the standard lib.

joe-elliott (approved):

PR #1388, opened 2022-04-15
Add a startTime and endTime parameter to the trace by id query API to improve query performance

### LINKED ISSUE #1373: Add a startTime and endTime parameter to the trace by id query API to improve query performance
The query trace by id feature in tempo searches all blocks in blocklist to find relevant blocks and merge trace objects found in these blocks. However, in environments that have high loads, even with compactors, the blocklist can extend upto thousands like 18000 etc, blocks and parsing through each of them affects performance. The current problem that we see is, each block has minId and maxId spanning across the maximum ranges which makes all blocks in blocklist an eligible candidate for parsing for a given trace id. 

We have come up with a time based search approach, that will include blocks for search based on the start time and end time of the blocks. The start time for the query is retrieved based on tags search. Default behaviour remains same if the time window is not specified.
Currently, in order to use the trace by id query, we have an elastic search deployed that gives us the trace ids based on specific tags or attributes. These objects also return the time window for the traces. We use the same window to query trace object by id so that the blocks outside the window are excluded and hence, the latency decreases and performance is better. 

We have tried experimenting with compactor to merge only contiguous blocks so that we can restrict the minId and maxId of compacted blocks, as opposed to the current implementation of sorting by block age but we have not been able to scale using this approach. So, we think the alternate way to improve the query performance is to restrict the search for given time windows.

**Additional context**
The API spec would now be /api/traces/{traceID}?timeEnd={end}&timeStart={start} 
A snapshot of the code changes would be : https://github.com/grafana/tempo/compare/v1.3.2...sagarwala:v1.3.2-branch?expand=1

sagarwala:
**What this PR does**:
This PR adds a startTime and endTime parameter to the Trace By ID API call, to restrict the number of blocks parsed while querying based on the time of the block. Default behaviour remains same if the time window is not specified. 
The API spec would now be /api/traces/{traceID}?timeEnd={end}&timeStart={start}

Fixes #1373

cc : @bikashmishra100 @ashwinidulams

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/querier/http.go
```diff
@@ -22,6 +23,8 @@ const (
 	BlockStartKey = "blockStart"
 	BlockEndKey   = "blockEnd"
 	QueryModeKey  = "mode"
+	TimeStartKey  = "timeStart"
```
joe-elliott, line 26: i'd prefer these params be called `start` and `end`. that would match the same parameters on the `/api/search` endpoint.

modules/querier/http.go
```diff
@@ -99,36 +104,40 @@ func (q *Querier) TraceByIDHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 // return values are (blockStart, blockEnd, queryMode, error)
-func validateAndSanitizeRequest(r *http.Request) (string, string, string, error) {
+func validateAndSanitizeRequest(r *http.Request) (string, string, string, int64, int64, error) {
```
joe-elliott, line 107: These new params also need to be checked on the query frontend and rejected in the event that they are invalid. 

We are currently checking the trace id here: https://github.com/grafana/tempo/blob/main/modules/frontend/frontend.go#L10

Perhaps we should move `validateAndSanitizeRequest` to the `/pkg/api` package? Then we could use the same method to check the params in both places?

joe-elliott, line 107: These new params also need to be checked on the query frontend and rejected in the event that they are invalid. 

We are currently checking the trace id here: https://github.com/grafana/tempo/blob/main/modules/frontend/frontend.go#L10

Perhaps we should move `validateAndSanitizeRequest` to the `/pkg/api` package? Then we could use the same method to check the params in both places?

tempodb/tempodb_test.go
```diff
@@ -371,7 +385,7 @@ func TestIncludeBlock(t *testing.T) {
 			e, err := tc.blockEnd.MarshalBinary()
 			require.NoError(t, err)
 
-			assert.Equal(t, tc.expected, includeBlock(tc.meta, tc.searchID, s, e))
+			assert.Equal(t, tc.expected, includeBlock(tc.meta, tc.searchID, s, e, 0, 0))
```
joe-elliott, line 388: this should use tc.start and tc.end

joe-elliott, line 388: this should use tc.start and tc.end

tempodb/tempodb_test.go
```diff
@@ -440,7 +462,7 @@ func TestIncludeCompactedBlock(t *testing.T) {
 			e, err := tc.blockEnd.MarshalBinary()
 			require.NoError(t, err)
 
-			assert.Equal(t, tc.expected, includeCompactedBlock(tc.meta, tc.searchID, s, e, blocklistPoll))
+			assert.Equal(t, tc.expected, includeCompactedBlock(tc.meta, tc.searchID, s, e, blocklistPoll, 0, 0))
```
joe-elliott, line 465: this should use tc.start and tc.end

joe-elliott, line 465: this should use tc.start and tc.end

joe-elliott (commented):
Overall this is pretty close. I appreciate the work that's done here. We are close to cutting 1.4 and if we hustle we can get this feature in that release (which I would like).

In addition to the requested changes we also need to add these new params to our docs:

https://github.com/grafana/tempo/blob/main/docs/tempo/website/api_docs/_index.md?plain=1#L100

It's also very important that while we describe these new fields carefully in the docs. If the start/end parameters are provided it's quite possible that the trace will not be found or returned partially. 

We should also mention the interaction between these new params and this config option:

https://github.com/grafana/tempo/blob/main/docs/tempo/website/configuration/_index.md?plain=1#L776

If a span is received that is outside that slack configuration value then the block start/end times are not updated. This currently has no impact on trace by id search b/c it searches everywhere. 

However, after this change, if you push a trace with dates from last week and then search for that trace using `/api/traces/<id>?start=<val>&end=<val>` then it will not be returned even if start/end correctly overlap with the trace.

Thanks for this feature addition. Let's clean this up some and try to get it merged for 1.4!

sagarwala:
Thank you for the review comments @joe-elliott. 

I have updated the documentation to include start/end parameters and added details on _ingestion_time_range_slack_ configuration. The `ValidateAndSanitizeRequest` method is now added to `package/api` and I am using the same for frontend as well as querier request validation. The test cases are also updated to include the start and end parameters.

docs/tempo/website/api_docs/_index.md
```diff
@@ -97,7 +97,7 @@ GET /api/traces/<traceid>
 The following query API is also provided on the querier service for _debugging_ purposes.
 
 ```
-GET /querier/api/traces/<traceid>?mode=xxxx&blockStart=0000&blockEnd=FFFF
+GET /querier/api/traces/<traceid>?mode=xxxx&blockStart=0000&blockEnd=FFFF&start=<start>&end=<end>
```
joe-elliott, line 100: It's good that you added these params to the querier docs (leave them), but we also need to add them to the query frontend endpoint a few lines above:

https://github.com/grafana/tempo/blob/main/docs/tempo/website/api_docs/_index.md?plain=1#L94

joe-elliott (commented):
a few more nits and we'll be good to merge. thanks!

sagarwala:
Updated the documentation and tests as per PR comments. Thanks much @joe-elliott for your time and review.

docs/tempo/website/api_docs/_index.md
```diff
@@ -91,13 +91,18 @@ Tempo's Query API is simple. The following request is used to retrieve a trace f
...
 
 ```
-GET /api/traces/<traceid>
+GET /api/traces/<traceid>?start=<start>&end=<end>
 ```
+Parameters:
+- `start = (unix epoch seconds)`
+  Optional.  Along with `end` define a time range from which traces should be returned. 
+- `end = (unix epoch seconds)`
+  Optional.  Along with `start` define a time range from which traces should be returned. Providing both `start` and `end` will include traces for the specified time range only.
```
joe-elliott, line 100: Can we move the extended warning and description below here? This is considered the external endpoint and it's the one we really need that warning on.

joe-elliott (commented):
Two small changes and we'll merge!

- Move the warning in the docs. (I added a note on the exact line I'm concerned about).
- Add a change log entry

sagarwala:
Done! Thank you!

joe-elliott (approved):
Really cool enhancement. Great addition @sagarwala!

PR #1393, opened 2022-04-18
Upgrade OTel to v0.46.0

mapno:
**What this PR does**:

Follow up from https://github.com/grafana/tempo/pull/1383

Tempo uses a replace directive to use an internal fork of the OTel collector. To update OTel, we also need to upgrade the fork.


joe-elliott (approved):
integration tests are passing. quick, merge before they break!

PR #1407, opened 2022-04-22
Design proposal: Parquet blocks

mdisibio:
**What this PR does**:
This PR adds a design proposal for Parquet blocks. Rendered markdown file can be viewed [here](https://github.com/mdisibio/tempo/blob/design-proposal-parquet/docs/design-proposals/2022-04%20Parquet.md)

n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/design-proposals/2022-04 Parquet.md
+++ b/docs/design-proposals/2022-04 Parquet.md
@@ -21,7 +21,7 @@ Over the last few months, the Tempo Team has invested a lot of effort in impleme
 The main design goals in choosing a new block format are to increase the speed and efficiency of Tempo search, and power the upcoming TraceQL language for querying and extracting metrics from traces.
 
 We also had the following requirements of the new format:
-* Roundtrippable with OTLP - A new block format must support full trace read/write, so it must be able to return be converted from OTLP and back again.
+* Roundtrippable with OTLP - A new block format must support full trace read/write, so it must be able to be converted from OTLP and back again.
 * More efficient search - Reduce i/o
 * Faster search - There's no point in having a new format that is slower
 * Similar or better block size - No significant increase in storage requirements for the same data
@@ -159,7 +159,7 @@ Span-level
 
 
 ### "Any"-type Attributes
-OTLP attributes have variable data types, which is easy accomplish in formats like protocol-buffers, but does not translate directly to Parquet.  Each column must have a concrete type.
+OTLP attributes have variable data types, which is easy to accomplish in formats like protocol-buffers, but does not translate directly to Parquet.  Each column must have a concrete type.
 
 There are several possibilities here but we chose to have an optional values for each concrete type.  We aren't 100% sold on this approach and open to feedback.  Array and KeyValueList types are stored as JSON-encoded strings. The data is portable and still searchable at extremely basic levels.
 
@@ -184,7 +184,7 @@ Pros:
 
 Cons:
 
-* Many nulls - as each attribute only populates 1 column, the other 5 are guaranteed to be null. This has a non-trivial increase in storage size.
+* Many nulls - as each attribute only populates 1 column, the other 5 are guaranteed to be null. This has a non-trivial impact on storage size.
 
 ### Event Attributes
 Span event attributes are stored as JSON-encoded strings in a generic key/value map. This is by far the most space-efficient encoding and the trade-off of decreased searchability seems worthwhile.  Storing event attributes this way reduces the block size ~16% for our dataset, which is huge.  There are currently no use cases to search event attributes, but we can revisit this in the future if needed.
@@ -239,7 +239,7 @@ Reads  : 290 6.65 MB
 ```
 
 ## Implementation
-We have been refactoring Tempo recently in anticipation of this, therefore Parquet-formatted blocks should be as easy as creating a new folder `/tempodb/encoding/vparquet` and implementing the `VersionedEncoding` interface. Tempo's parquet support will be based on the new library https://github.com/segmentio/parquet-go which we have used for all prototyping and have had execellent success with.
+We have been refactoring Tempo recently in anticipation of this, therefore Parquet-formatted blocks should be as easy as creating a new folder `/tempodb/encoding/vparquet` and implementing the `VersionedEncoding` interface. Tempo's parquet support will be based on the new library https://github.com/segmentio/parquet-go which we have used for all prototyping and have had excellent success with.
 
 ### Write path
 
--- a/docs/design-proposals/2022-04 Parquet.md
+++ b/docs/design-proposals/2022-04 Parquet.md
@@ -10,16 +10,23 @@ Last updated: 2022-04-18
 
 This design document describes adding a new columnar block format to Tempo based on Apache Parquet.  A columnar block format has many advantages within Tempo such as enabling faster search but also downstream by enabling a large ecosystem of tools access to the underlying trace data.
 
+## Context
+Over the last few months, the Tempo Team has invested a lot of effort in implementing efficient search queries over trace data spanning long time ranges. We mainly worked on two major initiatives to enable search and summarise our findings below:
+
+1. **Scaling up the existing Protobuf-based blocks**.  Tempo currently leverages massively-parallel serverless functions to perform full backend-search.  However even with thousands of functions, Tempo search is not able to achieve the search speeds that we would like on larger datasets.  Each trace is stored as an individual proto message and must be unmarshaled to look for things like service names and tags.  Even with scanning speeds of 60 gb/s from object storage, the large I/O makes it infeasible to query >24h periods over larger datasets.
+
+2. **Implementing a Flatbuffer-based storage format**.  Tempo currently uses flatbuffer-based data to power ingester search.  Flatbuffers improve search speeds by incurring no deserialization cost as the on-disk format matches the in-memory format.  However, the current ingester data is very limited to the current application and stores only a subset of searchable data.  Storing a 100% roundtrippable version of traces proved to be ~50% larger than the existing block size and is not feasible. This is due to the design of flatbuffers for alignment of various values, padding, and inability to use space-efficient data encodings such as variable-width integers.
+
 ## Design Goals and Requirements
 The main design goals in choosing a new block format are to increase the speed and efficiency of Tempo search, and power the upcoming TraceQL language for querying and extracting metrics from traces.
 
-We also had the following requirements:
+We also had the following requirements of the new format:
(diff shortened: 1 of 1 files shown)


docs/design-proposals/2022-04 Parquet.md
```diff
@@ -0,0 +1,256 @@
...
+
+2. **Implementing a Flatbuffer-based storage format**.  Tempo currently uses flatbuffer-based data to power ingester search.  Flatbuffers improve search speeds by incurring no deserialization cost as the on-disk format matches the in-memory format.  However, the current ingester data is very limited to the current application and stores only a subset of searchable data.  Storing a 100% roundtrippable version of traces proved to be ~50% larger than the existing block size and is not feasible. This is due to the design of flatbuffers for alignment of various values, padding, and inability to use space-efficient data encodings such as variable-width integers.
+
+## Design Goals and Requirements
+The main design goals in choosing a new block format are to increase the speed and efficiency of Tempo search, and power the upcoming TraceQL language for querying and extracting metrics from traces.
+
+We also had the following requirements of the new format:
+* Roundtrippable with OTLP - A new block format must support full trace read/write, so it must be able to return be converted from OTLP and back again.
```
joe-elliott, line 24:
```suggestion
* Roundtrippable with OTLP - A new block format must support full trace read/write, so it must be able to be converted from OTLP and back again.
```

docs/design-proposals/2022-04 Parquet.md
```diff
@@ -0,0 +1,256 @@
...
+* More efficient search - Reduce i/o
+* Faster search - There's no point in having a new format that is slower
```
joe-elliott, line 26: :rocket:

docs/design-proposals/2022-04 Parquet.md
```diff
@@ -0,0 +1,256 @@
...
+* DurationNanos - The total trace duration, computed at ingest time. This powers the min/max duration filtering in the current Tempo search and is more efficient than scanning the spans duration column. However, it may go away with TraceQL or we could decide to change it to span-level duration filtering too.
+
+
+### Dedicated columns
+Projecting attributes to their own columns has massive benefits for speed and size, and these are too good to pass up on.  Therefore we are taking an opinionated approach and projecting some common attributes to their own columns.  All other attributes are stored in the generic key/value maps and are still searchable, but not as quickly.  We chose these attributes based on what we commonly use ourselves (scratching our own itch), but we think they will be useful to most workloads.  There is an associated cost for unused attributes as it will still store a column of nulls, however the cost should be minimal.  We can be more generous with unused resource-level attributes because their overhead is the smallest of all, we are finding it to be ~.05% (0.0005) of the total block size.
+
+Resource-level:
```
joe-elliott, line 148: If we made these configurable would we have to store it in the block meta somewhere? or would it be dynamically detected during querying?

I feel like it would blow up the block meta too much to store these for every block, but I like the idea of having a configurable list of "dedicated columns".

docs/design-proposals/2022-04 Parquet.md
```diff
@@ -0,0 +1,256 @@
...
+* `container` and `k8s.container.name`
+
+Span-level
+* `http.method`
+* `http.url`
+* `http.status_code` (int)
+
+
+### "Any"-type Attributes
+OTLP attributes have variable data types, which is easy accomplish in formats like protocol-buffers, but does not translate directly to Parquet.  Each column must have a concrete type.
```
joe-elliott, line 162:
```suggestion
OTLP attributes have variable data types, which is easy to accomplish in formats like protocol-buffers, but does not translate directly to Parquet.  Each column must have a concrete type.
```

docs/design-proposals/2022-04 Parquet.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+Pros:
+
+* Preserves the functionality of parquet column min/max statistics which can be leveraged during search
+* Native data types should be more efficient to store and process
+
+Cons:
+
+* Many nulls - as each attribute only populates 1 column, the other 5 are guaranteed to be null. This has a non-trivial increase in storage size.
```
joe-elliott, line 187:
```suggestion
* Many nulls - as each attribute only populates 1 column, the other 5 are guaranteed to be null. This has a non-trivial impact on storage size.
```

docs/design-proposals/2022-04 Parquet.md
```diff
@@ -0,0 +1,256 @@
...
+    optional binary  Value (STRING);
+    optional boolean ValueBool;
+    optional double  ValueDouble;
+    optional int64   ValueInt (INT(64,true));
+    optional binary  ValueArray (STRING);
+    optional binary  ValueKVList (STRING);
+}
+```
+
+Pros:
```
joe-elliott, line 180: with 14 day retention object storage is ~30% of our TCO. Above it sounds like we are within ~5% of current block size. If this schema is improving query times I think it's worth it.

docs/design-proposals/2022-04 Parquet.md
```diff
@@ -0,0 +1,256 @@
...
+
+* Preserves the functionality of parquet column min/max statistics which can be leveraged during search
+* Native data types should be more efficient to store and process
+
+Cons:
+
+* Many nulls - as each attribute only populates 1 column, the other 5 are guaranteed to be null. This has a non-trivial increase in storage size.
+
+### Event Attributes
+Span event attributes are stored as JSON-encoded strings in a generic key/value map. This is by far the most space-efficient encoding and the trade-off of decreased searchability seems worthwhile.  Storing event attributes this way reduces the block size ~16% for our dataset, which is huge.  There are currently no use cases to search event attributes, but we can revisit this in the future if needed.
```
joe-elliott, line 190: I think this is a smart trade off for getting started. As we look at adding queryable events to TraceQL we can revisit.

docs/design-proposals/2022-04 Parquet.md
```diff
@@ -0,0 +1,256 @@
...
+column.Trace.ResourceSpans.InstrumentationLibrarySpans.Spans.ID             10428415 values 82.37 MB  # this is never used!
+```
+
+Some super early benchmarks on local SSDs:
+
+* Searching protobuf data for a simple query `cluster=ops and minDuration=1s`
+
+```
+=== RUN   TestSearchProto
+Traces : 55
```
joe-elliott, line 226: I think "Traces" means "Traces Matched"?

docs/design-proposals/2022-04 Parquet.md
```diff
@@ -0,0 +1,256 @@
...
+```
+=== RUN   TestSearchPipelineStatic
+Traces : 55
+Traces inspected: 154414
+Reads  : 290 6.65 MB
+--- PASS: TestSearchPipelineStatic (0.18s)
+```
+
+## Implementation
+We have been refactoring Tempo recently in anticipation of this, therefore Parquet-formatted blocks should be as easy as creating a new folder `/tempodb/encoding/vparquet` and implementing the `VersionedEncoding` interface. Tempo's parquet support will be based on the new library https://github.com/segmentio/parquet-go which we have used for all prototyping and have had execellent success with.
```
joe-elliott, line 242:
```suggestion
We have been refactoring Tempo recently in anticipation of this, therefore Parquet-formatted blocks should be as easy as creating a new folder `/tempodb/encoding/vparquet` and implementing the `VersionedEncoding` interface. Tempo's parquet support will be based on the new library https://github.com/segmentio/parquet-go which we have used for all prototyping and have had excellent success with.
```

docs/design-proposals/2022-04 Parquet.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Write path
+
+* Creating blocks: The ingester will write and flush parquet-formatted blocks. The compactor will only compact like-encoded-blocks, i.e. parquet to parquet. This seems better than having the ingester continue to flush proto and the compactor to convert them.
+
+*  One tunable is the row group size.  Our index pages target something like 256K, but row groups are typically much larger, 50-100MB.  Not sure about the best value here, will need to experiment.
```
joe-elliott, line 248: I'd say this difference is concerning. The smaller pages allows for heavy parallelization across the same block. My guess is that the significantly reduced i/o will more than make up for it. Agree experimentation is the best path to finding a good value.

joe-elliott (commented):
Very exciting work. Let's go!

Only other thought is that perhaps we should link to the dremel paper that parquet is based on. The nested columnar approach is such a nice match to trace data.

docs/design-proposals/2022-04 Parquet.md
annanay25: Even if we added a configurable list of dedicated columns, the schema together with all columns would be part of the parquet metadata, and we are looking at ways to cache that to prevent pulling it from the object store on every read. So I think it will be "fine" until it blows up to a million columns, but I'm just speculating.

docs/design-proposals/2022-04 Parquet.md
annanay25: Yes. Correct. We added that metric in our local tests to confirm both proto/parquet are returning the same number of hits. We can leave this line out in this doc though.

docs/design-proposals/2022-04 Parquet.md
annanay25: While we could reduce the rowgroup size further and get a lot more workers active on a single block, I agree with you that the reduced I/O makes up for the reduced parallelism. A lot of our local benchmarking uses sequential scanning of all rowgroups and the results are pretty good already. We'll continue to experiment.

annanay25: Also, the columnar format allows adding a new dimension to parallelism -- which is to search on multiple columns simultaneously.

docs/design-proposals/2022-04 Parquet.md
mdisibio: Making the columns configurable is a good idea and we had the same thought to be "less opinionated", but it ends up as a dynamic schema still, which is non-trivial and presents several challenges that we were avoiding in this first pass.  (1) read/write using reflection since a static struct is not available (2) attribute name -> column name translation (3) type-handling (unless name + type is part of the configuration).  

Yes the plan for when we get to dynamic schema is to detect the columns at runtime during querying.

docs/design-proposals/2022-04 Parquet.md
mdisibio: Good thoughts here.  Adding a note that we default to 10MB per search shard, which seems like a reasonable lower bound for row group size experimentation.  Also our default index downsampling is 1MB not 256KB, will fix.

PR #1409, opened 2022-04-25
metrics-generator: refactor service graphs processor

kvrhdn:
**What this PR does**:
This started out as me trying to add a metric to track the total amount of edges (`tempo_metrics_generator_processor_service_graphs_edges`). I then noticed we don't necessarily need to send edges into a collect channel as we can complete them during upsert. This 1) reduces memory pressure as we remove edges quicker and 2) improves correctness of the metrics as we update them when we receive the second span instead of when the edge expires.

Changes:
- create separate callbacks for on completed and on expired edge events
- during `UpsertEdge`, if the edge is completed call `onCompleted` and immediately remove the edge
- add tests for maxItems and to shake out data races
- rename `metrics_generator_processor_service_graphs_expired_spans` -> `metrics_generator_processor_service_graphs_edges`

Part of #1303 

**Checklist**
- [x] Tests updated
- [ ] ~~Documentation added~~


modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -177,14 +174,12 @@ func (p *processor) consume(resourceSpans []*v1_trace.ResourceSpans) error {
...
-				// upsertEdge will only return this errTooManyItems
+				// UpsertEdge will only return ErrTooManyItems
 				if err != nil {
 					return err
 				}
 
-				if edge.IsCompleted() {
-					p.collectCh <- k
-				}
+				p.metricTotalEdges.Inc()
```
mapno, line 182: Is this metric correct here? It looks like it should be called in `onComplete`.

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -30,10 +30,15 @@ var (
 		Name:      "metrics_generator_processor_service_graphs_dropped_spans",
 		Help:      "Number of dropped spans.",
 	}, []string{"tenant"})
-	metricExpiredSpans = promauto.NewCounterVec(prometheus.CounterOpts{
+	metricTotalEdges = promauto.NewCounterVec(prometheus.CounterOpts{
 		Namespace: "tempo",
-		Name:      "metrics_generator_processor_service_graphs_expired_spans",
-		Help:      "Number of spans that expired before finding their pair",
+		Name:      "metrics_generator_processor_service_graphs_edges",
+		Help:      "Total number of edges detected",
```
mapno, line 36: This metric is a bit confusing to me. Are these total number of edges recorded or only unique edges?

mapno, line 36: This metric is a bit confusing to me. Are these total number of edges recorded or only unique edges?

modules/generator/processor/servicegraphs/store/store.go
```diff
@@ -15,120 +15,103 @@ var _ Store = (*store)(nil)
...
+
+	ttl      time.Duration
+	maxItems int
 }
 
-func NewStore(ttl time.Duration, maxItems int, evictCallback Callback) Store {
+// NewStore creates a Store to build service graphs. The store caches edges, each representing a
+// request between two services. Once an edge is complete its metrics can be collected. Edges that
+// have not found their pair are deleted after ttl time.
+func NewStore(ttl time.Duration, maxItems int, onComplete, onExpired Callback) Store {
```
mapno, line 31: Nit: can we unify naming on callbacks? i.e. `onCompleted` or `onExpire`

mapno, line 31: Nit: can we unify naming on callbacks? i.e. `onCompleted` or `onExpire`

modules/generator/processor/servicegraphs/store/store.go
```diff
@@ -15,120 +15,103 @@ var _ Store = (*store)(nil)
...
 	s := &store{
-		l:   list.New(),
-		mtx: &sync.RWMutex{},
```
mapno, line 29: Is the new mutex set anywhere?

mapno (commented):
Nice refactor, left a few comments.

modules/generator/processor/servicegraphs/servicegraphs.go
kvrhdn: My intention is keep track of the total amount of unique edges. That should give more context around the amount of expired edges. With this metric we can calculate %age of exipred edges as `total / expired`.

That said, I see the code is not correct and it's currently counting the amount of spans that contribute to an edge... Let me fix that...

modules/generator/processor/servicegraphs/servicegraphs.go
kvrhdn: Okay, I've fixed this now by only incrementing it when a new edge is added. Returning `isNew` from `UpsertEdge` is a bit tedious but it keeps the Prometheus metrics out of the store implementation. We could alternatively also just move all metrics into store.

modules/generator/processor/servicegraphs/store/store.go
kvrhdn: Done :+1:

mapno (approved):
LGTM

PR #1414, opened 2022-04-26
metrics-generator: drop traces_spanmetrics_calls_total

kvrhdn:
**What this PR does**:
The value of `traces_spanmetrics_calls_total` always matches with `traces_spanmetrics_duration_seconds_count`. So to reduce cardinality we can just drop these.

Related to #1303 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


kvrhdn:
Closing this PR: it's better to keep the counter and stay consistent with the span-metrics processor from the OpenTelemetry Collector.

PR #1420, opened 2022-05-03
Fix compactor missing compaction_objects_combined_total metric and not honoring max_bytes_per_trace 

mdisibio:
**What this PR does**:
This PR fixes some things that were broken by refactor #1351.  (1) the `compaction_objects_combined_total` metric is now populated again.  (2) the per-tenant override `max_bytes_per_trace` is honored again, and the corresponding metric `tempo_discarded_spans_total{reason="trace_too_large_to_compact"}` populated.  

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## main / unreleased
 
 * [BUGFIX] metrics-generator: don't inject X-Scope-OrgID header for single-tenant setups [1417](https://github.com/grafana/tempo/pull/1417) (@kvrhdn)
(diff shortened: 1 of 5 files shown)


joe-elliott (approved):
This compaction combine/logic has gotten quite difficult to a follow. This was an easy mistake to make.

This feels like a consequence of Tempo being between two different design patterns? The old "data agnostic" tempodb and the new trace specific tempodb.

The good news is that all of this slowly unwinds:
1) Introduce Parquet blocks
2) Drop support for writing v2 blocks and only read them
3) Replace v2 format in wal
4) Replace v2 format in distributor <-> ingester communication
5) Delete v2 
6) :taco:

As we work through the above we can slowly push this mess into smaller and smaller corners and replace everything with the new block interfaces and formats. I think `/pkg/model` will always live as a way to translate the backend formats to OTLP, but eventually we will be able to gut it as well.

The future is bright!

PR #1421, opened 2022-05-04
Per tenant max search duration

joe-elliott:
**What this PR does**:
Adds the ability to have a per tenant max search duration. By default, if there is no per tenant configuration, the original config value is used. This PR also updates the jsonnet under `./operations` to mount the overrides yaml onto the frontend.

Additional changes
- Documentation as well as catching the missing `block_retention` override
- Updated the jsonnet example to use full backend search

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/tempo/website/configuration/_index.md
+++ b/docs/tempo/website/configuration/_index.md
@@ -988,12 +988,12 @@ overrides:
     # actually writing these metrics.
     [metrics_generator_disable_collection: <bool> | default = false]
 
-    # Per-user block retention. If this value is set to 0 (default) then block_retention
-    #  in the compactor config is used.
+    # Per-user block retention. If this value is set to 0 (default), then block_retention
+    #  in the compactor configuration is used.
     [block_retention: <duration> | default = 0s]
 
-    # Per-user max search duration. If this value is set to 0 (default) then max_duration
-    #  in the frontend config is used.
+    # Per-user max search duration. If this value is set to 0 (default), then max_duration
+    #  in the front-end configuration is used.
     [max_search_duration: <duration> | default = 0s]
 
     # Tenant-specific overrides settings configuration file. The empty string (default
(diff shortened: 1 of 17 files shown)


docs/tempo/website/configuration/_index.md
```diff
@@ -987,7 +987,15 @@ overrides:
...
     # actually writing these metrics.
     [metrics_generator_disable_collection: <bool> | default = false]
-      
+
+    # Per-user block retention. If this value is set to 0 (default) then block_retention
+    #  in the compactor config is used.
+    [block_retention: <duration> | default = 0s]
+
+    # Per-user max search duration. If this value is set to 0 (default) then max_duration
+    #  in the frontend config is used.
```
knylander-grafana, line 996:
```suggestion
    # Per-user block retention. If this value is set to 0 (default), then block_retention
    #  in the compactor configuration is used.
    [block_retention: <duration> | default = 0s]

    # Per-user max search duration. If this value is set to 0 (default), then max_duration
    #  in the front-end configuration is used.
```

knylander-grafana (changes requested):
Reviewed the doc files and have a few minor suggestions. "Config" was changed to configuration so it's spelled out. Front-end component is a compound adjective so should be hyphenated.

knylander-grafana (approved):
Thank you for making the suggested changes. I'll use Comment instead next time.

modules/overrides/limits.go
```diff
@@ -70,6 +70,9 @@ type Limits struct {
 	// Querier enforced limits.
 	MaxBytesPerTagValuesQuery int `yaml:"max_bytes_per_tag_values_query" json:"max_bytes_per_tag_values_query"`
 
+	// QueryFrontend enforced limits
+	MaxSearchDuration model.Duration `yaml:"max_search_duration" json:"max_search_duration"`
```
mdisibio, line 74: Is there a reason for `model.Duration`?  Just comparing to `MetricsGeneratorCollectionInterval time.Duration` above.

mdisibio (approved):
LGTM just one question.

PR #1433, opened 2022-05-17
Encoding use callbacks instead of direct compactor metrics

mdisibio:
**What this PR does**:
This PR moves the tempodb compaction metrics back closer to their usage, and the encodings use callbacks instead of direct access. If we like this pattern we can extend it a bit further.  I'm thinking of `compaction_objects_combined_total` which is measured by instrumenting an ObjectCombiner, however the future parquet encoding won't use that but a callback would work well.

Fixes n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/encoding/v2/compactor.go
```diff
@@ -92,7 +91,7 @@ func (*Compactor) Compact(ctx context.Context, l log.Logger, r backend.Reader, w
 
 		// make a new block if necessary
 		if currentBlock == nil {
-			currentBlock, err = NewStreamingBlock(&opts.BlockConfig, uuid.New(), tenantID, inputs, recordsPerBlock)
+			currentBlock, err = NewStreamingBlock(&c.opts.BlockConfig, uuid.New(), tenantID, inputs, recordsPerBlock)
```
joe-elliott, line 94: `&`? do we need to dereference here?

tempodb/compactor.go
```diff
@@ -25,6 +27,39 @@ const (
 	DefaultIteratorBufferSize = 1000
 )
 
+var (
+	MetricCompactionBlocks = promauto.NewCounterVec(prometheus.CounterOpts{
```
joe-elliott, line 31: make metrics internal?

joe-elliott (approved):
two minor/non-blocking comments. looks good!

PR #1434, opened 2022-05-18
metrics-generator: support per-tenant processor configuration

yvrhdn:
**What this PR does**:
Allow per-tenant configuration of the processor histogram buckets and dimensions.

The following overrides are added:
```yaml
metrics_generator_processor_service_graphs_histogram_buckets
metrics_generator_processor_service_graphs_dimensions
metrics_generator_processor_span_metrics_histogram_buckets
metrics_generator_processor_span_metrics_dimensions
```

Related to #1303 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/tempo/website/configuration/_index.md
+++ b/docs/tempo/website/configuration/_index.md
@@ -968,7 +968,7 @@ overrides:
     [metrics_generator_processors: <list of strings>]
 
     # Per-user configuration of the metrics-generator processors. The following configuration
-    # will override settings in the global config.
+    # overrides settings in the global configuration.
     [metrics_generator_processor_service_graphs_histogram_buckets: <list of float>]
     [metrics_generator_processor_service_graphs_dimensions: <list of string>]
     [metrics_generator_processor_span_metrics_histogram_buckets: <<list of float>]


modules/generator/instance.go
```diff
@@ -113,60 +116,92 @@ func (i *instance) watchOverrides() {
 	}
 }
 
-func (i *instance) updateProcessors(desiredProcessors map[string]struct{}) error {
+func (i *instance) updateProcessors() error {
+	desiredProcessors := i.overrides.MetricsGeneratorProcessors(i.instanceID)
```
joe-elliott, line 120: should this use some kind of default value if an instance has no overrides? are overrides required to turn on processors for a tenant?

modules/generator/instance.go
```diff
@@ -113,60 +116,92 @@ func (i *instance) watchOverrides() {
...
 			toAdd = append(toAdd, processorName)
 		}
 	}
-	for processorName := range i.processors {
+	for processorName, processor := range i.processors {
 		if _, ok := desiredProcessors[processorName]; !ok {
 			toRemove = append(toRemove, processorName)
+			continue
+		}
+		switch processorName {
```
joe-elliott, line 173: instead of adding `Config() interface{}` to the processor interface i'd consider just doing a type switch here on `processor` and accessing the processors config directly once you know the type.

modules/generator/instance.go
```diff
@@ -113,60 +116,92 @@ func (i *instance) watchOverrides() {
...
 			return err
 		}
 	}
 	for _, processorName := range toRemove {
 		i.removeProcessor(processorName)
 	}
+	for _, processorName := range toReplace {
+		i.removeProcessor(processorName)
+
+		err := i.addProcessor(processorName, desiredCfg)
```
joe-elliott, line 149: i feel like this is a little dangerous b/c we could remove a tenants processor and then fail to add it back during an overrides update. as long as we have sufficient alerts around `metricActiveProcessorsUpdateFailed` it should be ok?

joe-elliott (commented):
overall looks good. some thoughts in the comments.

per offline discussion there's one more update to this to prevent multiple series from being registered? after that's pushed i will do a final review.

modules/generator/instance.go
kvrhdn: Yeah, that's definitely a concern. This can already happen btw if you have a typo in your processor name.
I.e. the following config will fail because the processor name is incorrect:
```diff
        metrics_generator_processors:
            - servicegraphs
```
We have an alert on `tempo_metrics_generator_active_processors_update_failed_total` to catch this 🙂

modules/generator/registry/histogram.go
```diff
@@ -147,7 +149,11 @@ func (h *histogram) updateSeries(s *histogramSeries, value float64, traceID stri
 	s.lastUpdated.Store(time.Now().UnixMilli())
 }
 
-func (h *histogram) collectMetrics(appender storage.Appender, timeMs int64, externalLabels map[string]string) (activeSeries int, err error) {
+func (h *histogram) Name() string {
+	return h.name
+}
+
+func (h *histogram) CollectMetrics(appender storage.Appender, timeMs int64, externalLabels map[string]string) (activeSeries int, err error) {
```
joe-elliott, line 156: any particular reason these were exported from the package?

modules/generator/registry/registry.go
```diff
@@ -132,7 +134,10 @@ func (r *ManagedRegistry) registerMetric(m metric) {
 	r.metricsMtx.Lock()
 	defer r.metricsMtx.Unlock()
 
-	r.metrics = append(r.metrics, m)
+	if _, ok := r.metrics[m.Name()]; ok {
+		level.Debug(r.logger).Log("msg", "replacing metric, counters will be reset", "metric", m.Name())
```
joe-elliott, line 138: nit: i would make this info level. it won't happen often and it would be good to have a record of it when it does.

joe-elliott (approved):
a nit and a question. looks good!

modules/generator/registry/histogram.go
kvrhdn: Err, for consistency and to avoid naming conflicts 😅  The interface and the structs these methods are linked to aren't exported, so I didn't think it matters much.

But basically:
- I introduced the method `name()` to the `metric` interface
- this conflicts with the field `name`, so I decided to rename the function to `Name`
- for consistency, I've also written the other functions with a capital letter

https://github.com/kvrhdn/tempo/blob/23ebaf43bd3dfce386a23f965ca42ab9d09ec457/modules/generator/registry/registry.go#L74-L78

I can see if there is a cleaner solution...

knylander-grafana (approved):
Minor suggestion to change future tense (will ...) to present tense.

docs/tempo/website/configuration/_index.md
```diff
@@ -967,6 +967,13 @@ overrides:
     #  - span-metrics
     [metrics_generator_processors: <list of strings>]
 
+    # Per-user configuration of the metrics-generator processors. The following configuration
+    # will override settings in the global config.
```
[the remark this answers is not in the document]
kvrhdn: Thanks!

modules/generator/registry/histogram.go
kvrhdn: Alright, reverted back the name change.

PR #1436, opened 2022-05-19
TraceQL Parser

joe-elliott:
**What this PR does**:
Adds an unused TraceQL Parser to the codebase. This parser converts a string into a TraceQL AST to eventually be executed against the upcoming Parquet backend.

Follow up to #1378 

This PR is ready for review. Some notes:

- There are large number of values that need to be exported from this package. Currently saving this for a later PR when we actually begin implementation.
- There are a set of empty functions in ast.go that begin with double underscore: `__`. These functions exist only to implement interfaces that give structure to the way the ast objects are related. 
- Nested parents are currently not allowed. `{ parent.parent.foo }` currently parses as a reference to an attribute named `parent.foo` on the current span's parent. Unsure if we want to support deeply nesting this or not.
- Scalar pipeline expressions do not allow static expressions or statics on the LHS. This seems odd but it's to remove conflicts with scalar filters. These are currently not allowed:
  - `(by(namespace) | count()) > 2 * 2`
  - `(by(namespace) | count()) * 2 > 2`
  - `2 < (by(namespace) | count())`
- There is one "shift/reduce conflict" on a query like this: `({true}) && ({true})` This could be parsed as two separate spanset pipelines that have the results &&ed together or as two spanset expressions that have been &&ed together. It's parsed as the latter which is preferred.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


mdisibio:
This is awesome.  Is `expr.y.go` the only auto-generated file?

pkg/traceql/test_examples.yaml
```diff
@@ -0,0 +1,229 @@
...
+  - '{ true } || { true }'
+  - '{ true } >> { true }'
+  - '{ true } > { true }'
+  - '{ true } ~ { true }'
+  # scalar filters
+  - 'avg(.field) > 1'
+  - 'sum(3) = 2'
+  - 'min(childCount) < 2'
+  - 'max(duration) >= 1s'
+  - 'avg(.field) > avg(1)'
```
mdisibio, line 75: I think I understand why `aggregate(static)` is valid (fieldExpression definition includes all attribute lookups/intrinsics/status).  It doesn't seem useful and would prefer if it weren't parseable, and probably not worth changing the grammar for.  But wondering if these statements should be included in the "valid" list?

joe-elliott:
> This is awesome. Is expr.y.go the only auto-generated file?
yup!

pkg/traceql/test_examples.yaml
```diff
@@ -0,0 +1,229 @@
...
+  - '{ resource.a != 3 }'
+  - '{ span.a != 3 }'
+  - '{ !("test" != .c || ((true && .b) || 3 < .a)) }'
+  - '{ parent = nil }'
+  - '{ status = ok }'
+  - '{ status = unset }'
+  - '{ status = error }'
+  - '{ status != error }'
+  - '{ duration > 1s }'
+  - '{ duration > 1s * 2s }'
```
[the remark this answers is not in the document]
joe-elliott: My thought was just to default to go which would i believe multiply `2s * 1s` as nanoseconds which give a completely nonsensical answer. The reason I figured we'd go this way is due to the complexity of resolving whether or not this particular combination of floats, ints, durations and operators makes sense.

My guess is that it would be better to be open and let the use type nonsensical things then be strict and miss some cases that made sense? Perhaps it's not as hard to I think to come up with all correct and incorrect cases. Currently all combinations parse and validate.

Some examples that make sense:
```
1s + 2s
1s = 2s
1s / 2
1s * 2
```

Some examples that don't make sense:
```
1s * 2s
1s = 2
```

pkg/traceql/test_examples.yaml
joe-elliott: It would be difficult to remove from the grammar, but not particularly difficult to add to the validate logic. We could descend the field expression and make sure that at least one element was an attribute or intrinsic. 

Will make this change.

joe-elliott:
> Could you clarify more about what is invalid? It looks like MUL has higher precedence, so I would expect this to be equivalent to (by(namespace) | count()) > 4 which seems ok?

There is currently no way to have statics in scalar pipeline expressions. It's surprisingly difficult to do without conflicting with scalar expressions and scalar filters.

The following not being allowed is unfortunate, but technically you could still express it:
`(by(namespace) | count()) * 2 > 2` -> `(by(namespace) | count()) > 1`

Not allowing this is also unfortunate and also related to the parsing of scalar expressions:
`2 < (by(namespace) | count()) `

mdisibio (approved):

PR #1437, opened 2022-05-22
[Feature] Add error span logging processor

faustodavid:
**What this PR does**:

I would like to keep a small discussion that I started on the Tempo Slack channel, where I wonder what would you think about having an error-span logging processor, similar to the one that Grafana Agent has but this one just log span with status error.

The motivation is to have a way to do aggregation over high cardinality span tags like for example `http.url`  without overwhelming the logging tool and an easy way to discover old traces that have errors while Tempo search performance keep being optimized.

Because the metrics and the logs use the same label sanitizers will be easy to jump between metrics to logs using data links.

Example usage:

Create a Grafana panel using the span generated metrics and group it by  (span_name, http_status_code and http_method)

The `span_name` is the URL template, that the libraries use to keep the cardinality down.

We get a graph like the following

image

From it, we can create data links that use the same label set of the series that have a high error percentage and query our logging tool.

From the logs now we can aggregate the tags to find correlations with the failure, sometimes developers add more tags to enrich the context of some spans and those tags can be very useful to understand the error correlation.

image

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/example/docker-compose/error-span-logging/docker-compose.yaml
+++ b/example/docker-compose/error-span-logging/docker-compose.yaml
@@ -1,0 +1,71 @@
+version: "3"
+services:
+
+  tempo:
+    image: grafana/tempo:latest
+    command: [ "-config.file=/etc/tempo.yaml" ]
+    volumes:
+      - ./tempo.yaml:/etc/tempo.yaml
+      - ./tempo-data/:/tmp/tempo
+    ports:
+      - "3200:3200"   # tempo
+      - "14268"  # jaeger ingest
+    depends_on:
+      - loki
+    logging:
+      driver: loki
+      options:
+        loki-url: 'http://localhost:3100/loki/api/v1/push'
+
+  synthetic-load-generator:
+    image: omnition/synthetic-load-generator:1.0.25
+    volumes:
+      - ../shared/load-generator.json:/etc/load-generator.json
+    environment:
+      - TOPOLOGY_FILE=/etc/load-generator.json
+      - JAEGER_COLLECTOR_URL=http://tempo:14268
+    depends_on:
+      - tempo
+
+  prometheus:
+    image: prom/prometheus:latest
+    volumes:
+      - ../shared/prometheus.yaml:/etc/prometheus.yaml
+    ports:
+      - "9090:9090"
+    command: [
+      "--config.file=/etc/prometheus/prometheus.yml",
+      "--storage.tsdb.path=/prometheus",
+      "--web.console.libraries=/usr/share/prometheus/console_libraries",
+      "--web.console.templates=/usr/share/prometheus/consoles",
+      "--enable-feature=exemplar-storage",
+      "--web.enable-remote-write-receiver"
+    ]
+
+  grafana:
+    image: grafana/grafana:latest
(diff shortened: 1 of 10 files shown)


kvrhdn:
It's really cool you could set this up so quickly with the existing metrics-generator code 🙂 

We discussed this with the team and a major drawback we see here is that the logs generator is not multi-tenant. Tempo is natively multi-tenant (but can be run in single-tenant mode without issue) and any feature we add should support that as well. In this case, the logs the logs processor emits are sent to the standard log stream, together with all the other logs from Tempo.
Ideally we create separate log streams for every tenant and support a robust logging protocol. In the metrics part we achieve this be writing metrics using the Prometheus remote write protocol, isolating them from the operational metrics that are scraped from Tempo.

While I believe this is just a technical hurdle to figure out, the added complexity of managing log streams and dealing with log protocols isn't something we will be able to work on and maintain. Our priority for the next half year (at least) is maturing the metrics-generator and TraceQL / Parquet. 

We suggest the following alternatives:
- we can move this logic into the distributor as a configurable logging option. We already have a config option to log all received spans, we can extend this to only log spans with an error and log more attributes.
https://github.com/grafana/tempo/blob/a54bf2d9156377630636906a4efcf8d79f2959ea/modules/distributor/distributor.go#L245-L247
- the automatic logging processor in the Grafana Agent already covers most of your requirements, I think we could add a config option here to only log spans with an error. That should be a quick change. The agent also has strong support for sending logs, so that's an already solved issue there. That said, this is only suitable if you are able to run the Grafana Agent in your infrastructure.
  You can find that code here: https://github.com/grafana/agent/tree/main/pkg/traces/automaticloggingprocessor

faustodavid:
hey @kvrhdn , yeah, you are right, it makes total sense all that you mentioned. I will close the Draft PR, and I will create a new one updating the distributor.

But I would like to have some guidance on how the distributor's configuration file should look.

At the moment it looks like the following:

```yaml
distributor:
    #optional
    log_received_traces: bool
```

Should we add another flag to don't introduce breaking changes in the configuration file, or should we update the existing one adding the option to just log the spans with errors and include the attributes?

This is an idea if we add a new flag.

```yaml
distributor:
    #optional
    log_received_spans_with_errors: 
        enabled: bool
        include_attributes: bool      
```

or something more simple like: 

```yaml
distributor:
    #optional
    [log_received_spans_with_errors: bool]
```

faustodavid:
Hey @kvrhdn, I will create a new draft PR with the following configuration in the distributor, and ask for feedback from the team.

```yaml
  log_received_traces:
    enabled: <bool> | default = false
    include_attributes: <bool> | default = false
    filter_by_status: <list of string> | default = <empty list> # Valid values: "ok", "unset", "error"
```

kvrhdn:
Hi @faustodavid, sorry for not getting back to you earlier. The suggested config looks really nice! Will it be possible to make this backwards-compatible? I.e. support both:

```yaml
  log_received_traces: true
```
and
```yaml
  log_received_traces:
    enabled: true
```

But feel free to create a PR, we can discuss details in there 🙂

PR #1439, opened 2022-05-23
Change label name for metrics generator instance

joe-elliott:
**What this PR does**:
The metrics generators add an `instance` label to distinguish between metrics series. This prevents out of order issues with remote write. Unfortunately `instance` is a very common label name and may conflict with custom dimensions on service graphs or span metrics.

This PR proposes a much more unique label name that is far less likely to conflict `__metrics_gen_instance`. Feel free to suggest other options if desired.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## main / unreleased
 
-* [CHANGE] metrics-generator: Changed added metric label `instance` to `__metrics_gen_instance` to prevent reduce collisions with custom dimensions. [#1439](https://github.com/grafana/tempo/pull/1439) (@joe-elliott)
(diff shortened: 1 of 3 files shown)


kvrhdn (approved):
Back when I added the `instance` label I was trying to figure out what would be most appropriate. The only guidance I found about this was:

> _Label names may contain ASCII letters, numbers, as well as underscores. They must match the regex `[a-zA-Z_][a-zA-Z0-9_]*`. Label names beginning with __ are reserved for internal use._

https://prometheus.io/docs/concepts/data_model/

So this technically isn't allowed, but tbh I think this document is mostly aimed towards clients that are scraped by Prometheus?
An internal label conflicting with `__metrics_gen_instance` seems _very_ unlikely.

PR #1443, opened 2022-05-24
Migrate all Operational dashboard panels

zalegrala:
**What this PR does**:

Here we update the panels on the Operational dashboard, which should be slightly more performant, and has some newer features we may want to use.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/operations/tempo-mixin/dashboards/tempo-operational.json
+++ b/operations/tempo-mixin/dashboards/tempo-operational.json
@@ -3,7 +3,10 @@
     "list": [
       {
         "builtIn": 1,
-        "datasource": "-- Grafana --",
+        "datasource": {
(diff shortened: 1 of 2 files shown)


joe-elliott:
Nice! Can we get a screenshot or list of updates?

joe-elliott (approved):

zalegrala:
image
It seemed to help the exemplars in the read portion.  This was just me clicking "migrate" on all the panels, and maybe one color choice, which we can argue about. :)

joe-elliott:
exemplars!

under the write tab did you aggregate away tenants? right now i think bytes/sec and spans/sec are both per tenant per instance. would be a nice cleanup.

PR #1447, opened 2022-05-25
Limit of 0 disables max_bytes_per_tag_values_query

joe-elliott:
**What this PR does**:
- Disables the `max_bytes_per_tag_values_query` limit when set to 0.
- Returns a 500 instead of 200. (Screenshot of error in Grafana below)
- Adds instance SearchTag/TagValues test. This tests the scenario where `max_bytes_per_tag_values_query` is set to 0.

image

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


CHANGELOG.md
```diff
@@ -3,7 +3,8 @@
 * [CHANGE] metrics-generator: Changed added metric label `instance` to `__metrics_gen_instance` to reduce collisions with custom dimensions. [#1439](https://github.com/grafana/tempo/pull/1439) (@joe-elliott)
 * [FEATURE] metrics-generator: support per-tenant processor configuration [#1434](https://github.com/grafana/tempo/pull/1434) (@kvrhdn)
 * [ENHANCEMENT] Added the ability to have a per tenant max search duration. [#1421](https://github.com/grafana/tempo/pull/1421) (@joe-elliott)
-* [BUGFIX] Fix nil pointer panic when the trace by id path errors. [1441](https://github.com/grafana/tempo/pull/1441) (@joe-elliott)
+* [BUGFIX] Fix nil pointer panic when the trace by id path errors. [#1441](https://github.com/grafana/tempo/pull/1441) (@joe-elliott)
+* [BUGFIX] Don't enforce `max_bytes_per_tag_values_query` when set to 0. [#1447](https://github.com/grafana/tempo/pull/1447) (@joe-elliott)
```
annanay25, line 7: I don't really think this is a bugfix: we just have mixed behaviour throughout Tempo when it comes to limits, maybe we can move this to `CHANGE`

modules/ingester/instance_search_test.go
```diff
@@ -145,6 +147,94 @@ func TestInstanceSearch(t *testing.T) {
 	checkEqual(t, ids, sr)
 }
 
+func TestInstanceSearchTags(t *testing.T) {
```
annanay25, line 150: Can we club this test with the above?

annanay25, line 150: Was wondering if we could write a test where we return 500 once the tag-values limit is exceeded

joe-elliott: added.

annanay25 (approved):
lgtm!

PR #1451, opened 2022-05-26
Publish compiled mixin like mimir

pracucci:
**What this PR does**:
I'm leading an effort to improve and uniform Mimir, Loki and Tempo monitoring. This includes publishing the compiled mixin and runbooks. I see Tempo already does both (good job!). As a further step, I would like to align the location of the published mixin to the Mimir one.

Do you have any concern if we move it to `operations/tempo-mixin-compiled` and move dashboards to a dedicated folder?

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott:
No issues here with moving the location. I feel like it's easier to find in the new folder. 

Currently we also compile `./operations/jsonnet` to `./operations/kube-manifests`. Thoughts on changing that folder name to align with this change? `./operations/jsonnet` => `./operations/jsonnet-compiled`? WDYT?

pracucci:
> Currently we also compile ./operations/jsonnet to ./operations/kube-manifests. Thoughts on changing that folder name to align with this change? ./operations/jsonnet => ./operations/jsonnet-compiled? WDYT?

I think it makes sense. If you don't mind, I will do it in a separate PR, to keep changes separated.

pracucci:
> > Currently we also compile ./operations/jsonnet to ./operations/kube-manifests. Thoughts on changing that folder name to align with this change? ./operations/jsonnet => ./operations/jsonnet-compiled? WDYT?
> 
> I think it makes sense. If you don't mind, I will do it in a separate PR, to keep changes separated.

See: https://github.com/grafana/tempo/pull/1454

annanay25 (approved):
This LGTM. Thanks.

One comment: Since the mixin is auto-vendored into our internal environment, we will have to update the new location so our vendor bots can continue to find the mixin.

pracucci:
> One comment: Since the mixin is auto-vendored into our internal environment, we will have to update the new location so our vendor bots can continue to find the mixin.

I don't understand this. Our internal environment vendor the jsonnet, not the compiled one. This PR only moves the location of the compiled `.yaml` and `.json`, so how does it affect our internal environment?

PR #1455, opened 2022-05-31
Snyk workflow add

SadFaceSmith:
What this PR does / why we need it:
This PR adds a short GH Action to generate and upload SBOM reports via Snyk.

Which issue(s) this PR fixes:

Fixes #

Special notes for your reviewer:

FYSA @grafana/security-team

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -1,0 +1,60 @@
+name: Snyk SBOM CI
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  snyk_scans:
+
+    runs-on: ubuntu-latest
+
+    steps:
+
+      - uses: actions/checkout@v3
+      - name: Run Snyk to check for vulnerabilities - sarif output
+        continue-on-error: true
+        uses: snyk/actions/node@master
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+        with:
+          command: test
+          args:  --all-projects --sarif-file-output=${{ github.event.repository.name }}.sarif --strict-out-of-sync=false
+
+      - name: Run Snyk to check for vulnerabilities - json output
+        continue-on-error: true
+        uses: snyk/actions/node@master
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+        with:
+          command: test
+          args: --json --all-projects --json-file-output=${{ github.event.repository.name }}.json  --strict-out-of-sync=false
+
+      - name: Use Node.js 18.x
+        uses: actions/setup-node@v2
+        with:
+          node-version: 18.x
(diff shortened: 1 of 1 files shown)


joe-elliott:
What are SBOM reports? What is Snyk? Why do we want this change?

SadFaceSmith:
@joe-elliott SBOM is a 'Software Bill of Materials'. Essentially a list of dependencies/libraries/etc used within a codebase/product. Snyk is a security tool that we are trialing to provide our customers with SBOM and SCA (software composition analysis) reports. 

We have multiple customers now asking for these reports from our enterprise offerings. We are integrating some of our public repos with Snyk now to get a good idea if Snyk will work across all of our codebases.

.github/workflows/snyk.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - uses: actions/checkout@v3
+      - name: Run Snyk to check for vulnerabilities - sarif output
+        continue-on-error: true
+        uses: snyk/actions/golang@master
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+        with:
+          command: test
+          args:  --all-projects --sarif-file-output=${{ github.event.repository.name }}.sarif --strict-out-of-sync=false
```
joe-elliott, line 24: is there a reason we're running this twice with different output formats? (sarif and json)? 

is a single pass not sufficient?

.github/workflows/snyk.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Run Snyk to check for vulnerabilities - json output
+        continue-on-error: true
+        uses: snyk/actions/golang@master
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+        with:
+          command: test
+          args: --json --all-projects --json-file-output=${{ github.event.repository.name }}.json  --strict-out-of-sync=false
+
+      - name: install snyk-to-html
```
joe-elliott, line 35: i suppose a similar question here as before. any particular reason to generate all these different formats?

joe-elliott (commented):
A few questions in the review. 

Also,
- Will this block PRs if it fails?
- Are the resulting reports public?

.github/workflows/snyk.yml
SadFaceSmith: In my testing, you can only have one format type per run of Snyk unfortunately. We've had asks for both SARIF and JSON formats.

.github/workflows/snyk.yml
SadFaceSmith: Per customer requests mostly. Some customers want the SARIF format, some SPDX, some HTML reports.

SadFaceSmith:
@joe-elliott As of right now, yes the results are public. They are uploaded to this repo by the last Action task. They are in a zip file located on the 'Summary' page for the Action.

joe-elliott (approved):

PR #1456, opened 2022-05-31
Include rollout dashboard, borrowed from Mimir

zalegrala:
**What this PR does**:

* Include rollout dashboard
* Adjust Grafana label to `tempo-mixin`

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/operations/tempo-mixin-compiled/dashboards/tempo-reads.json
+++ b/operations/tempo-mixin-compiled/dashboards/tempo-reads.json
@@ -1498,7 +1498,7 @@
  "schemaVersion": 14,
  "style": "dark",
  "tags": [
-  "tempo"
+  "tempo-mixin"
  ],
  "templating": {
   "list": [
--- a/operations/tempo-mixin-compiled/dashboards/tempo-reads.json
+++ b/operations/tempo-mixin-compiled/dashboards/tempo-reads.json
@@ -1498,7 +1498,7 @@
  "schemaVersion": 14,
  "style": "dark",
  "tags": [
-  "tempo-mixin"
+  "tempo"
  ],
  "templating": {
   "list": [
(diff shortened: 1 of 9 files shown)


operations/tempo-mixin/dashboards/rollout-progress.libsonnet
```diff
@@ -0,0 +1,318 @@
...
+          id: 1,
+          gridPos: { h: 8, w: 10, x: 0, y: 0 },
+        },
+
+        //
+        // Writes
+        //
+        $.panel('Writes - 2xx') +
+        $.newStatPanel(|||
+          sum(rate(cortex_request_duration_seconds_count{%(gateway_job_matcher)s, route=~"%(gateway_write_routes_regex)s",status_code=~"2.+"}[$__rate_interval])) /
```
joe-elliott, line 96: I think all these `cortex_` metrics should be `tempo_` metrics? Double check by importing the dashboard into an env that has tempo to make sure these graphs are populated.

operations/tempo-mixin-compiled/dashboards/tempo-reads.json
```diff
@@ -1498,7 +1498,7 @@
  "schemaVersion": 14,
  "style": "dark",
  "tags": [
-  "tempo"
+  "tempo-mixin"
```
annanay25, line 1501: I would prefer to keep this "tempo" as it shows up nicely in the Grafana Dashboard explore. 
image

but no strong opinion!

annanay25 (approved):
Nice one Zach 🚀

operations/tempo-mixin-compiled/dashboards/tempo-reads.json
zalegrala: Once nice thing about changing it is that we'd be able to tell that those dashboards came from the mixin.  I was thinking about this in the context when other dashboards get created that aren't part of the mixin, knowing where to make changes requires some head scratching.

zalegrala:
image

PR #1459, opened 2022-05-31
Add metric to track feature enablement

zalegrala:
**What this PR does**:

Add metric to track which features are enabled in the configuration.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo/app/app.go
+++ b/cmd/tempo/app/app.go
@@ -180,13 +180,13 @@ func (c *Config) Describe(ch chan<- *prometheus.Desc) {
 func (c *Config) Collect(ch chan<- prometheus.Metric) {
 
 	features := map[string]int{
-		"external_endpoints": 0,
-		"search":             0,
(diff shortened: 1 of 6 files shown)


cmd/tempo/app/app.go
```diff
@@ -163,6 +173,35 @@ func (c *Config) CheckConfig() {
...
 }
 
+func (c *Config) Describe(ch chan<- *prometheus.Desc) {
+	ch <- metricConfigFeatDesc
+}
+
+func (c *Config) Collect(ch chan<- prometheus.Metric) {
+
+	features := map[string]int{
+		"external_endpoints": 0,
```
kvrhdn, line 183: I think `search_external_endpoints` would be more clear, WDYT?

zalegrala: Good call, that sounds good to me.

PR #1465, opened 2022-06-01
Extend distributor trace logger with optional features to include span attributes and filter by error status

### LINKED ISSUE #1485: Enrich log_received_traces with attributes and error filter
Tempo metrics generator is awesome to understand the rate and latency of the generated spans, but the only thing I am missing is a way to do error distribution analysis with high cardinality attributes.

For example:

We enabled Metrics Generator with the custom dimensions `http.status_code` and `http.method`.

Then use the following query to see the error percentages in a GRPC service.

```
sum(rate(traces_spanmetrics_duration_seconds_count{service="$service", span_kind="SPAN_KIND_SERVER", http_status_code=~"$httpErrorPattern"}[$interval])) by (span_name, http_status_code, http_method) / 
ignoring(http_status_code) group_left sum(rate(traces_spanmetrics_duration_seconds_count{service="$service", span_kind="SPAN_KIND_SERVER"}[$interval])) by (span_name, http_method)
```

If we found a spike of errors like the following graph, we can use exemplars or Tempo backend search to find traces and try to comprehend why we are getting these errors, but because we have to look at the traces one by one, sometimes it is hard to find common attributes among the spans with an error that could lead us to find the root cause of the issue.

image

Enrich `log_received_traces` feature to include all attributes and add a filter to just log spans that have errors. 

Then we can have data links from the metrics to the logs using the same tags that our metric is using

image

Then in the logs we can do aggregations by attributes to understand its distribution. For example:

image

Now, I am confident that all issues are related to an exception in the mapping, and from this log, I can go to the trace to see who is the caller and why now we are getting this issue.

Another example, we are getting spikes of 404 in GET Requests, we will be able to see the URL and understand exactly which resources are not found.

Additionally, it is important to have the Error filter to don't overload the logs. Assuming that a service has an availability of 99.9% we will be logging ~0.1% of all spans.

faustodavid:
**What this PR does**:
Extend distributor trace logger with optional features to include span attributes and filter by error status

This can help to do error correlation analysis by attribute.

Fixes #1485

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/tempo/website/configuration/_index.md
+++ b/docs/tempo/website/configuration/_index.md
@@ -130,7 +130,7 @@ distributor:
     # Enable to log every received span to help debug ingestion or calculate span error distributions using the logs
     log_received_spans:
         [enabled: <boolean> | default = false]
-        [include_attributes: <boolean> | default = false]
+        [include_all_attributes: <boolean> | default = false]
         [filter_by_status_error: <boolean> | default = false]
 
     # Optional.
--- a/modules/distributor/config.go
+++ b/modules/distributor/config.go
@@ -46,9 +46,9 @@ type Config struct {
 }
 
 type LogReceivedSpansConfig struct {
-	Enabled             bool `yaml:"enabled"`
-	IncludeAttributes   bool `yaml:"include_attributes"`
-	FilterByStatusError bool `yaml:"filter_by_status_error"`
+	Enabled              bool `yaml:"enabled"`
+	IncludeAllAttributes bool `yaml:"include_all_attributes"`
+	FilterByStatusError  bool `yaml:"filter_by_status_error"`
 }
 
 // RegisterFlagsAndApplyDefaults registers flags and applies defaults
@@ -62,6 +62,6 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
 
 	f.BoolVar(&cfg.LogReceivedTraces, util.PrefixConfig(prefix, "log-received-traces"), false, "Enable to log every received trace id to help debug ingestion.")
 	f.BoolVar(&cfg.LogReceivedSpans.Enabled, util.PrefixConfig(prefix, "log-received-spans.enabled"), false, "Enable to log every received span to help debug ingestion or calculate span error distributions using the logs.")
-	f.BoolVar(&cfg.LogReceivedSpans.IncludeAttributes, util.PrefixConfig(prefix, "log-received-spans.include-attributes"), false, "Enable to include span attributes in the logs.")
+	f.BoolVar(&cfg.LogReceivedSpans.IncludeAllAttributes, util.PrefixConfig(prefix, "log-received-spans.include-attributes"), false, "Enable to include span attributes in the logs.")
 	f.BoolVar(&cfg.LogReceivedSpans.FilterByStatusError, util.PrefixConfig(prefix, "log-received-spans.filter-by-status-error"), false, "Enable to filter out spans without status error.")
 }
--- a/modules/distributor/distributor.go
+++ b/modules/distributor/distributor.go
@@ -285,8 +285,8 @@ func (d *Distributor) PushBatches(ctx context.Context, batches []*v1.ResourceSpa
 	}
 
 	if d.cfg.LogReceivedSpans.Enabled || d.cfg.LogReceivedTraces {
-		if d.cfg.LogReceivedSpans.IncludeAttributes {
-			logSpansWithAttributes(batches, d.cfg.LogReceivedSpans.FilterByStatusError, d.logger)
+		if d.cfg.LogReceivedSpans.IncludeAllAttributes {
+			logSpansWithAllAttributes(batches, d.cfg.LogReceivedSpans.FilterByStatusError, d.logger)
 		} else {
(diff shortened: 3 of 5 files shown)


faustodavid:
Hey @kvrhdn, in my last draft PR, you mentioned making `log_received_traces` backward compatible. Do you have an example of how I can parse the YAML optionally as a bool or struct? 

Also the name `log_received_traces`, I think it should be `log_received_spans`, what do you think?

modules/distributor/config.go
```diff
@@ -43,6 +45,12 @@ type Config struct {
 	factory func(addr string) (ring_client.PoolClient, error) `yaml:"-"`
 }
 
+type LogReceivedSpansConfig struct {
+	Enabled             bool `yaml:"enabled"`
+	IncludeAttributes   bool `yaml:"include_attributes"`
```
mapno, line 50: This can become incredibly verbose, as cardinality in traces tends to be high. What do you think of changing it to a `[]string` so specific attributes can be selected instead? It can have a `all` option to just print everything.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## main / unreleased
 
+* [FEATURE] Extend distributor trace logger with optional features to include span attributes and filter by error status. [#1465](https://github.com/grafana/tempo/pull/1465) (@faustodavid)
```
mapno, line 3: We should note the breaking change/deprecation here.

kvrhdn (commented):
This looks good to me. Nice work on the tests! Works as expected in my local environment :)

modules/distributor/config.go
kvrhdn: If we make it a boolean now and later want to change into `[]string` that would be a breaking change (or some nasty yaml handling). So perhaps it's better to add it now.
On the other hand, I'm in favour of keeping it simple and reducing the amount of configuration options. Cardinality isn't a concern for log fields but the volume of logs will be.

modules/distributor/distributor.go
```diff
@@ -531,11 +540,54 @@ func recordDiscaredSpans(err error, userID string, spanCount int) {
...
+				}
+				level.Info(logger).Log("msg", "received", "spanid", hex.EncodeToString(s.SpanId), "traceid", hex.EncodeToString(s.TraceId))
+			}
+		}
+	}
+}
+
+func logSpansWithAttributes(batch []*v1.ResourceSpans, filterByStatusError bool, logger log.Logger) {
+	for _, b := range batch {
+		logger := logger // copy logger so we can modify it per resource
```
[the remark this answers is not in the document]
faustodavid: done, can you review again, please?

CHANGELOG.md
faustodavid: done. 

Should I add a warning log if the deprecated flag is used?
https://github.com/grafana/tempo/blob/main/cmd/tempo/app/app.go#L146

modules/distributor/config.go
faustodavid: @mapno should I modify it like this?

```yaml
log_received_spans:
    [enabled: <boolean> | default = false]
    [include_attributes: <list of string> | default = ] // To include all attributes use `all`]
    [filter_by_status_error: <boolean> | default = false]
```

I can also rename `include_attributes` to `include_all_attributes` so if in the future is decided to have the option to select just a few attributes we could use `include_attributes` as the list of attributes.

@kvrhdn indeed, if this feature is used in production without the error filter it can overwhelm the log engine.

docs/tempo/website/configuration/_index.md
```diff
@@ -123,7 +123,15 @@ distributor:
...
+
+    # Optional.
+    # Enable to log every received span to help debug ingestion or calculate span error distributions using the logs
+    log_received_spans:
+        [enabled: <boolean> | default = false]
+        [include_attributes: <boolean> | default = false]
+        [filter_by_status_error: <boolean> | default = false]
 
     # Optional.
     # disables write extension with inactive ingesters. Use this along with ingester.lifecycler.unregister_on_shutdown = true
```
knylander-grafana, line 137: "Disables" should be capitalized since it is the first word in a sentence. 
```suggestion
    # Disables write extension with inactive ingesters. Use this along with ingester.lifecycler.unregister_on_shutdown = true
```

knylander-grafana (approved):
Doc updates look good. I fixed a capitalization error on another line.

modules/distributor/config.go
mapno: I like the proposal of using `include_all_attributes`. It's explicit and keeps the door open for extension in the future.

CHANGELOG.md
mapno: That'd be ideal yes. Thanks :)

mapno (commented):
Looks good to me. I'm in favor of merging when the last two minor threads are closed 👍

CHANGELOG.md
faustodavid: done :)

modules/distributor/config.go
faustodavid: done :)

faustodavid:
Hey @mapno @kvrhdn , I have linked an issue to the PR to describe a bit better the intention of this feature.

mapno (approved):
Great work 😄. Thanks for the contribution!

PR #1468, opened 2022-06-02
Procect forwarder watchOverrides from race condition

mapno:
**What this PR does**:

Reworks update logic in watchOverrides loop.
The race condition comes from the unpredictable behaviour of ranging over a map.

This could cause that a newly created queueManagers was added to the ranged map,
and be evaluated for update.
Since the queueManager was _just_ created, it wouldn't have enough running workers,
which would trigger another update.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/forwarder.go
+++ b/modules/distributor/forwarder.go
@@ -138,22 +138,43 @@ func (f *forwarder) watchOverrides() {
 		select {
 		case <-time.After(f.overridesInterval):
 			f.mutex.Lock()
+
+			var (
+				queueManagersToDelete []*queueManager
+				queueManagersToAdd    []struct {
+					tenantID               string
+					queueSize, workerCount int
+				}
+			)
+
 			for tenantID, tm := range f.queueManagers {
 				queueSize, workerCount := f.getQueueManagerConfig(tenantID)
-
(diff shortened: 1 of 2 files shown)


modules/distributor/forwarder.go
```diff
@@ -138,22 +138,43 @@ func (f *forwarder) watchOverrides() {
...
-					go func() {
-						// shutdown the queue manager
-						// this will block until all workers have finished and the queue is drained
-						if err := tm.shutdown(); err != nil {
-							level.Error(log.Logger).Log("msg", "error shutting down queue manager", "tenant", tenantID, "err", err)
-						}
-					}()
-					delete(f.queueManagers, tenantID)
-					f.queueManagers[tenantID] = newQueueManager(tenantID, queueSize, workerCount, f.forwardFunc)
+					queueManagersToDelete = append(queueManagersToDelete, tm)
```
joe-elliott, line 154: let's add an "info" log message here that indicates we are reloading a queue manager. this will be useful for understanding some of these behaviors if we continue to have issues in this area.

mapno: Good point, added!

PR #1470, opened 2022-06-02
Add s3 config option insecure_skip_verify

### LINKED ISSUE #1466: s3: add insecure_skip_verify to enable TLS without verifying the certificate
The s3 config block supports an `insecure` parameter which disables TLS completely. For environments with self-signed certificates it would be nice to support `insecure_skip_verify` as well. This would enable TLS but not verify the certificate.

Add `insecure_skip_verify` to not verify certificates.

You can probably work around this by adding the self-signed certificates to your authority store, but this isn't always straight-forward.

**Additional context**
Example code to achieve this using minio-go: https://github.com/minio/minio-go/issues/1019#issuecomment-415917603

fyi, Loki has this option: https://github.com/grafana/loki/blob/main/pkg/storage/bucket/http/config.go#L12

zalegrala:
**What this PR does**:

Here we include an S3 config option to disable TLS verification on HTTPs endpoints.

Fixes #1466

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * [ENHANCEMENT] metrics-generator: expose max_active_series as a metric [#1471](https://github.com/grafana/tempo/pull/1471) (@kvrhdn)
 * [ENHANCEMENT] Azure Backend: Add support for authentication with Managed Identities. [#1457](https://github.com/grafana/tempo/pull/1457) (@joe-elliott)
 * [ENHANCEMENT] Add metric to track feature enablement [#1459](https://github.com/grafana/tempo/pull/1459) (@zalegrala)
+* [ENHANCEMENT] Add s3 config option `insecure_skip_verify` [#1470](https://github.com/grafana/tempo/pull/1470) (@zalegrala)
 * [BUGFIX] Fix nil pointer panic when the trace by id path errors. [#1441](https://github.com/grafana/tempo/pull/1441) (@joe-elliott)
 * [BUGFIX] Update tempo microservices Helm values example which missed the 'enabled' key for thriftHttp. [#1472](https://github.com/grafana/tempo/pull/1472) (@hajowieland)
 * [BUGFIX] Fix race condition in forwarder overrides loop. [1468](https://github.com/grafana/tempo/pull/1468) (@mapno)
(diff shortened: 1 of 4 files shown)


tempodb/backend/s3/s3.go
```diff
@@ -343,6 +343,10 @@ func createCore(cfg *Config, hedge bool) (*minio.Core, error) {
 		return nil, errors.Wrap(err, "create minio.DefaultTransport")
 	}
 
+	if cfg.InsecureSkipVerify {
```
joe-elliott, line 346: i believe this will work but may be worth testing locally. you could probably spin up a minio with https and set this parameter pretty easily.

Insecure would need to be false.

joe-elliott (approved):

docs/tempo/website/configuration/_index.md
```diff
@@ -532,7 +532,11 @@ storage:
 
             # optional.
             # enable if endpoint is http
-            [insecure: <bool>]          
+            [insecure: <bool>]
+
+            # optional.
+            # enable to skip verification of an HTTPs endpoint
+            [insecure_skip_verify: <bool>]
```
KMiller-Grafana, line 539: I know that this definition looks similar to those near it within the file.  It could be improved by stating that a value of true causes the verification to be skipped,  and that the default is false. (Did I get the implications right?)

When I read this, I have to make a mental jump to associating "enable" with the boolean value "true." Easy for me, but perhaps more difficult for others.

knylander-grafana: Learned something by reading your comment. :) I would not have caught the same thing.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * [ENHANCEMENT] Added the ability to have a per tenant max search duration. [#1421](https://github.com/grafana/tempo/pull/1421) (@joe-elliott)
 * [ENHANCEMENT] Azure Backend: Add support for authentication with Managed Identities. [#1457](https://github.com/grafana/tempo/pull/1457) (@joe-elliott)
 * [ENHANCEMENT] metrics-generator: expose max_active_series as a metric [#1471](https://github.com/grafana/tempo/pull/1471) (@kvrhdn)
+* [ENHANCEMENT] Add s3 config option `insecure_skip_verify` [#1470](https://github.com/grafana/tempo/pull/1470) (@zalegrala)
```
knylander-grafana, line 10: Consider changing Add to Added to it matches other entries in the changelog. If s3 is Amazon S3, then the S should probably be capitalized.
```suggestion
* [ENHANCEMENT] Added S3 configuration option `insecure_skip_verify` [#1470](https://github.com/grafana/tempo/pull/1470) (@zalegrala)
```

knylander-grafana (commented):
I have one minor suggestion on the changelog file.

tempodb/backend/s3/s3.go
zalegrala: Alright, I've been able to test this and confirm that it does work.

CHANGELOG.md
zalegrala: Thanks for the review.  I've updated, though I do see a split in the changelog.

PR #1483, opened 2022-06-09
Google Cloud Run is the new Cloud Functions

joe-elliott:
**What this PR does**:
Changes Tempo's GCP serverless technology from Function to Run. Google Cloud Run supports generic container images and allows us to use Go 1.18 for Parquet support. 

Changes made:
- In drone upload a docker image instead of a zip file to gcs
- Remove cloud functions code/makefile instructions and replace with cloud run
- Updated e2e tests to use the new cloud run image
- Dropped a few places we were using custom code instead of 1.17 features (time.UnixMicro(), math.MaxInt)
- Changed Cloud Functions documentation to be Cloud Run focused
- Updated both Lambda and Cloud run to use Go 1.17 to match the rest of the codebase.

**Checklist**
- [ ] Tests updated
- [x] Documentation added


pkg/util/trace_info.go
```diff
@@ -141,7 +141,7 @@ func (t *TraceInfo) makeThriftBatch(TraceIDHigh int64, TraceIDLow int64) *thrift
 			OperationName: fmt.Sprintf("vulture-%d", t.generateRandomInt(0, 100)),
 			References:    nil,
 			Flags:         0,
-			StartTime:     unixMicro(t.timestamp),
+			StartTime:     time.Now().UnixMicro(),
```
mdisibio, line 144: Does this impact vulture's consistency checks?

joe-elliott: This should be the exact same value:

https://github.com/grafana/tempo/pull/1483/files#diff-64607fbb74bbb31f1d923cc9c1b4916f90ba904eb09a21320dbf7f2cf2dc59fdL13-L16

joe-elliott: I say this and I just noticed that tons of tests related to these timestamps are failing. Will dig deeper.

joe-elliott: yup, i see what i did now. good catch!

cmd/tempo-serverless/cloud-run/go.mod
```diff
@@ -0,0 +1,147 @@
+module github.com/grafana/tempo/cmd/tempo-serverless/cloud-run
+
+go 1.18
```
zalegrala, line 3: Time to party.

cmd/tempo-serverless/cloud-run/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+#
+#  to exercise the function
+#    curl http://localhost:8080/?start=1000&end=1001&...
+#
+
+#
+# build the serverless container image
+#
+#  todo: FROM scratch saves ~5MB which could be meaningful in a serverless setting, but using scratch gave strange errors on query.
+FROM alpine:3.15 as certs
```
zalegrala, line 12: I think 3.16 is out if you want to upgrade.

zalegrala (approved):
This looks good to me.  Thanks for putting it together.

cmd/tempo-serverless/cloud-run/go.mod
joe-elliott: I'm actually going to walk this back to 1.17 to match the rest of the codebase. We should move this forward together.

cmd/tempo-serverless/cloud-run/Dockerfile
joe-elliott: Not a bad idea, but all other images are on 3.15. I'm going to leave this on 3.15 as well. We should open a separate PR that moves all of them forward as one.

PR #1484, opened 2022-06-10
Implement configuration option for the querier to only hit the ingesters based on the trace id hashes instead of all healthy ingesters

### LINKED ISSUE #1477: Implement configuration option for the querier to only hit the ingesters based on the trace id hashes instead of all healthy ingesters.
This is a feature request for Trace by ID API call on Query Frontend. We understand that each call gets sharded and parallel requests are sent to Querier and Ingester to fetch the trace based on the ID, but as per the [code](https://github.com/grafana/tempo/blob/main/modules/querier/querier.go#L194) , are we making a call to all healthy ingesters in the ring. 

This can lead to issues with fetch, even though one of the ingesters are unhealthy(for RF=1), we will not be able to query the ingesters as it will fail with _Too many unhealthy replicas_ error. We watch to fetch the ingester instance based on the hash of trace id (how we find out the ingester instance during distributor writes by using the filter on [ringstrategy](https://github.com/grafana/tempo/blob/35d7e668d7c8466e372f99edaa99007be5eb2b03/vendor/github.com/grafana/dskit/ring/batch.go#L58)

We can increase the RF to increase failure tolerance but that has an associated cost.

cc: @bikashmishra100 , @ashwinidulams

sagarwala:
**What this PR does**:
Implement configuration option for the querier to only hit the ingesters based on the trace id hashes instead of all healthy ingesters

Fixes #1477

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/querier/config.go
```diff
@@ -17,6 +17,7 @@ type Config struct {
 	ExtraQueryDelay         time.Duration `yaml:"extra_query_delay,omitempty"`
 	MaxConcurrentQueries    int           `yaml:"max_concurrent_queries"`
 	Worker                  worker.Config `yaml:"frontend_worker"`
+	QuerySingleIngester     bool          `yaml:"query_single_ingester"`
```
joe-elliott, line 20: Let's change this parameter name. When using RF3 your change would actually be querying 3 ingesters, not 1.

I'm not coming up with a good name. If you have a good idea, please use it over :).
`query_ingesters_from_ring`? `query_relevant_ingesters`?

docs/tempo/website/configuration/_index.md
```diff
@@ -341,6 +341,12 @@ querier:
     # not distinguish between the types of queries.
     [max_concurrent_queries: <int> | default = 5]
 
+    # The query frontend sents sharded requests to ingesters and querier (/api/traces/<id>)
+    # By default, healthy ingesters are queried for the trace id.
```
joe-elliott, line 345:
```suggestion
    # By default, all healthy ingesters are queried for the trace id.
```

docs/tempo/website/configuration/_index.md
```diff
@@ -341,6 +341,12 @@ querier:
...
+    # This value controls if single ingester is to be retrieved based on the trace id hash.
```
joe-elliott, line 346:
```suggestion
    # When true the querier will hash the trace id in the same way that distributors do.
    # The querier will then only query those ingesters who own the trace id hash as determined by the ring as opposed to querying all ingesters.
```

joe-elliott (commented):
Change looks good. Had a few comments and suggestions.

sagarwala:
Thank you @joe-elliott . I have updated the PR as per the comments. I am using `query_relevant_ingesters` for the config name :)

joe-elliott (approved):
Nice addition. Thanks!

PR #1492, opened 2022-06-15
Add `write trace` CLI command

zalegrala:
**What this PR does**:

Here we include a `tempo-cli write trace` command that was useful in producing a large trace for debugging limits.

**Checklist**
- [ ] Tests updated
- [x] Documentation added


knylander-grafana:
Should write trace be added to the CLI page? https://github.com/grafana/tempo/blob/main/docs/tempo/website/operations/tempo_cli.md

zalegrala:
I believe this is going to require a bump of the `opentelementry-proto` submodule before the `make vendor-check` will pass.  The alternative is to use an older version of the exporter, but I like that idea less.

PR #1510, opened 2022-06-23
Build and release Tempo for all GOOS/GOARCH combinations

julienduchesne:
We are building a company-wide deb and RPM repository and we want to have all of our OSS products in it. Before I start working on this for Tempo, I saw that only linux-amd64 is built and released. This PR widens that, with the goal being of standardizing our offering across products (mimir, loki and grafana already offer binaries for all OSes/archs)

In this PR:
- Build all three tools (tempo-query, tempo-cli and tempo) across more archs and os combinations
- Keep the same archive format as now (all three tools in the same tar.gz file), let me know if that would be better to change. Though, changing the format could break automation that users have created
- Add a pull request run for `make release-snapshot`. Testing the whole releasing process. This will be very useful once I add rpm and deb packaging


zalegrala (commented):
Nice PR.  I'm now reminded that I miss having snapshot releases.  Thanks for your efforts with standards also.

.goreleaser.jsonnet
```diff
@@ -0,0 +1,63 @@
+local tools = ['tempo', 'tempo-query', 'tempo-cli'];
```
joe-elliott, line 1: i'd prefer leaving this yaml. it's easier to read.

.drone/drone.jsonnet
```diff
@@ -266,6 +266,41 @@ local deploy_to_dev() = {
               for d in aws_serverless_deployments
             ],
   },
+  // Build and deploy serverless code packages
```
joe-elliott, line 269: is there a reason to move this to drone? it was working fine in GHA.

joe-elliott (commented):
Thank you for taking on the addition of these architectures to our release binaries, but I'd prefer leaving our old CI patterns intact.

.goreleaser.jsonnet
julienduchesne: Templating. Lots of repetition with all three tools. And it would get even worse with OS packaging. I don’t mind moving it back to YAML if maintaining a big YAML file isn’t an issue

.drone/drone.jsonnet
joe-elliott: That's my bad. I think I just saw jsonnet and became unnecessarily enraged. Let me give this a real review.

.drone/drone.yml
```diff
@@ -274,6 +274,37 @@ trigger:
...
+platform:
+  arch: amd64
+  os: linux
+steps:
+- commands:
+  - go install github.com/google/go-jsonnet/cmd/jsonnet@latest
+  - make release-snapshot
+  image: golang:1.18
+  name: test release
+  when:
```
joe-elliott, line 289: Were we running the release snapshot before? If we weren't I'd prefer removing it instead of running it on every PR.

.drone/drone.yml
```diff
@@ -274,6 +274,37 @@ trigger:
...
+  - go install github.com/google/go-jsonnet/cmd/jsonnet@latest
+  - make release
+  env:
+    GITHUB_TOKEN:
+      from_secret: gh_token
+  image: golang:1.18
+  name: release
+  when:
+    event:
+    - tag
```
joe-elliott, line 302: Will this run on every tag? I think we'd prefer the `v*` tags like before.

.goreleaser.jsonnet
joe-elliott: I prefer the simplicity of the yaml file to the additional toolchain dependencies of jsonnet and having this file in repo. perhaps when i see the full file i will change my mind? I didn't find the repetition of the previous config to be that terrible.

.drone/drone.yml
julienduchesne: I'm setting up the stage to build OS package (deb and rpm). This will involve GPG signing and probably some OS-specific scripts. I think it would be best to have some kind of test as it would be a bit brittle/hard to maintain otherwise

As nfpm, the packaging tool, is integrated in goreleaser, the test I'm planning is:
1. Build snapshot release (creating the packages)
2. Install the packages on bare debian and centos/fedora images

.drone/drone.jsonnet
julienduchesne: Sorry for enraging you 😃. I might change my mind on Drone before this is merged (I believe it would be an even harder sell for the mimir team, they don't even have a config yet). To be continued next week

.drone/drone.yml
joe-elliott: Good call :+1:

.goreleaser.jsonnet
julienduchesne: I removed the jsonnet part

.drone/drone.jsonnet
julienduchesne: I put it back in Github Actions. Least friction for you, indeed. But it is unbearably slow. 90 secs in Drone vs 13 minutes in Github actions

julienduchesne: Done!

joe-elliott (approved):
Thanks!

PR #1514, opened 2022-06-24
Sync docs to website on release branches

mdisibio:
**What this PR does**:
This PR adds a new workflow to sync docs to the website from release branches.  This will have no effect on existing branches unless we copy the workflow into the branch.  Added as a separate workflow from the existing main->next sync because it is quite different.  The main->next workflow as renamed for clarity.

n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


zalegrala (approved):
Nice work.

PR #1518, opened 2022-06-28
Add `blocklist_poll_jitter_ms` config value

### LINKED ISSUE #1462: Azure DNS Lookup Failures
**Describe the bug**
In Azure all components can register DNS errors while trying to work with meta.compacted.json. This issue does not occur in any other environments. Unsure if this is an Azure issue or a Tempo issue. The error itself is a failed tcp connection to a DNS server which suggests this is some issue with Azure infrastructure. However, the fact that the error (almost?) always occurs on `meta.compacted.json` suggests something about the way we handle that file is different and is causing this issue.

The failures look like:

```
reading storage container: Head "https://tempoe**************.blob.core.windows.net/tempo/single-tenant/d8aafc48-5796-4221-ac0b-58e001d18515/meta.compacted.json?timeout=61": dial tcp: lookup tempoe**************.blob.core.windows.net on 10.0.0.10:53: dial udp 10.0.0.10:53: operation was canceled
```

or
 
```
error deleting blob, name: single-tenant/*******************/data: Delete "https://tempoe******.blob.core.windows.net/tempo/single-tenant/5b1ab746-fee7-409c-944d-1c1d5ba7a70e/data?timeout=61": dial tcp: lookup tempoe******.blob.core.windows.net on 10.0.0.10:53: dial udp 10.0.0.10:53: operation was canceled
```

We have seen this issue internally. Also reported here: #1372.

joe-elliott:
**What this PR does**:
Adds a configurable number of maximum milliseconds to sleep before pulling a block in the poller. This is useful when hosting on Azure to reduce pressure on DNS.

Fixes #1462 

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Additionally, default label `span_status` is renamed to `status_code`.
 * [ENHANCEMENT] Azure Backend: Add support for authentication with Managed Identities. [#1457](https://github.com/grafana/tempo/pull/1457) (@joe-elliott)
 * [ENHANCEMENT] Add metric to track feature enablement [#1459](https://github.com/grafana/tempo/pull/1459) (@zalegrala)
 * [ENHANCEMENT] Added s3 config option `insecure_skip_verify` [#1470](https://github.com/grafana/tempo/pull/1470) (@zalegrala)
+* [ENHANCEMENT] Added polling option to reduce issues in Azure `blocklist_poll_jitter_ms` [#1518](https://github.com/grafana/tempo/pull/1518) (@joe-elliott)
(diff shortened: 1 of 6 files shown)


knylander-grafana (approved):
Thank you for the doc updates.

zalegrala (approved):
Nice work chasing this down.  Looks good to me.

joe-elliott:
Thanks for the reviews. I'm going to hold off on merging this for a bit. The environment that was showing these errors is now, inexplicably, not showing these errors. Even without the fix. 

I want to confirm this actually fixes something before merging.

PR #1519, opened 2022-06-28
Adding Blocklist backend metrics

### LINKED ISSUE #1502: Add more blocklist metrics
The tempo blocklist is polled in the background by the poller, and the poller is interesting because it is the "source of truth" for the rest of the system, and the place with the most accurate view of the blocks in the backend.  It currently records one metric for the number of live blocks: [tempodb_blocklist_length](https://github.com/grafana/tempo/blob/main/tempodb/blocklist/poller.go#L125)

Since we have the entire lists of live and compacted blocks right there, let's record some other useful and easy to calculate metrics.  I propose these new metrics:

* tempodb_backend_objects
  * The total number of objects (traces) in the backend, defined as the sum(meta.TotalObjects).
  * Gauge, int64
  * Labels:  tenant, block status ("live", "compacted")
* tempodb_backend_bytes
  * The total number of bytes in the backend.  Just summing the data files `sum(meta.Size)` would fairly accureate.  We could include index files by also adding `sum(meta.IndexPageSize * meta.TotalRecords)`.  Bloom filters are not straightforward, because while the meta includes BloomShardCount, it doesn't include the shard size.  I have some ideas on this, but I think it's ok to skip. The data file is typically >99% of the block size anyway.
  * Gauge, int64
  * Labels: tenant, block status ("live", "compacted")

ie-pham:
**What this PR does**:
This adds additional blocklist backend metrics as mentioned in #1502 which are total backend objects and total backend bytes. 

Fixes #1502

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ Additionally, default label `span_status` is renamed to `status_code`.
 * [BUGFIX] Update tempo microservices Helm values example which missed the 'enabled' key for thriftHttp. [#1472](https://github.com/grafana/tempo/pull/1472) (@hajowieland)
 * [BUGFIX] Fix race condition in forwarder overrides loop. [1468](https://github.com/grafana/tempo/pull/1468) (@mapno)
 * [ENHANCEMENT] Add a config to query single ingester instance based on trace id hash for Trace By ID API. (1484)[https://github.com/grafana/tempo/pull/1484] (@sagarwala, @bikashmishra100, @ashwinidulams)
+* [ENHANCEMENT] Add blocklist metrics for total backend objects and total backend bytes [#1519](https://github.com/grafana/tempo/pull/1519) (@ie-pham)
 
 ## v1.4.1 / 2022-05-05
 
--- a/tempodb/blocklist/poller.go
+++ b/tempodb/blocklist/poller.go
@@ -300,8 +300,8 @@ func (p *Poller) tenantIndexPollError(idx *backend.TenantIndex, err error) error
 type backendMetaMetrics struct {
 	blockMetaTotalObjects          int
 	compactedBlockMetaTotalObjects int
-	blockMetaTotalBytes            int
-	compactedBlockMetaTotalBytes   int
+	blockMetaTotalBytes            uint64
+	compactedBlockMetaTotalBytes   uint64
 }
 
 func sumTotalBackendMetaMetrics(blockMeta []*backend.BlockMeta, compactedBlockMeta []*backend.CompactedBlockMeta) backendMetaMetrics {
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ Additionally, default label `span_status` is renamed to `status_code`.
 * [BUGFIX] Update tempo microservices Helm values example which missed the 'enabled' key for thriftHttp. [#1472](https://github.com/grafana/tempo/pull/1472) (@hajowieland)
 * [BUGFIX] Fix race condition in forwarder overrides loop. [1468](https://github.com/grafana/tempo/pull/1468) (@mapno)
 * [ENHANCEMENT] Add a config to query single ingester instance based on trace id hash for Trace By ID API. (1484)[https://github.com/grafana/tempo/pull/1484] (@sagarwala, @bikashmishra100, @ashwinidulams)
+* [ENHANCEMENT] Add blocklist metrics for total backend objects and total backend bytes [#1502](https://github.com/grafana/tempo/pull/1519) (@ie-pham)
```
annanay25, line 33:
```suggestion
* [ENHANCEMENT] Add blocklist metrics for total backend objects and total backend bytes [#1519](https://github.com/grafana/tempo/pull/1519) (@ie-pham)
```
We generally link the PR number here

tempodb/blocklist/poller_test.go
```diff
@@ -381,6 +381,132 @@ func TestTenantIndexPollError(t *testing.T) {
...
 }
 
+func TestBlockListBackendMetrics(t *testing.T) {
+	tests := []struct {
+		name                        string
+		list                        PerTenant
+		compactedList               PerTenantCompacted
+		testType                    string
+		expectedBackendObjectsTotal int
+		expectedBackendBytesTotal   uint64
```
annanay25, line 391: I think we should add similar fields for compacted objects meta just to make it explicit that we're testing those as well -- in the current scenario the total counts of objects/sizes are coincidentally lining up which can be confusing if the tests change in the future

tempodb/blocklist/poller.go
```diff
@@ -126,6 +141,14 @@ func (p *Poller) Do() (PerTenant, PerTenantCompacted, error) {
 
 		blocklist[tenantID] = newBlockList
 		compactedBlocklist[tenantID] = newCompactedBlockList
+
+		totalObjectsBlockMeta, totalObjectsCompactedBlockMeta := p.sumTotalObjects(newBlockList, newCompactedBlockList)
```
annanay25, line 145: Just a couple of questions to think about, but non blocking - 
- Should metric counts occur on the main polling function? Small but unexpected side effects (1) Longer polling time in highly multitenant envs with large blocklists, (2) will the `metricBlocklistPollDuration` drift? I guess not
- Should all components metric this? Maybe it only needs to be done by the component that creates a tenant index

I believe @joe-elliott has written most of the `Poller` code so maybe he can provide some more insight on these

annanay25 (commented):
Nice work @ie-pham! Happy to see your PR 🚀

tempodb/blocklist/poller.go
```diff
@@ -275,3 +296,34 @@ func (p *Poller) tenantIndexPollError(idx *backend.TenantIndex, err error) error
 
 	return nil
 }
+
+type backendMetaMetrics struct {
+	blockMetaTotalObjects          int
+	compactedBlockMetaTotalObjects int
+	blockMetaTotalBytes            int
```
joe-elliott, line 303: it's not difficult to exceed a 32 bit signed int's max value in total bytes stored. these days most things are running on 64 bit systems but if we're going to use an "int" data type we should consider the impact of it being 32 bits wide.

Maybe leave these as uint64 until they are cast to floats for the prom metric?

joe-elliott (approved):
let's go!

PR #1530, opened 2022-06-29
Parquet: Expose ReadBufferSize and ReadBufferCount to be configurable

joe-elliott:
**What this PR does**:
Exposes two major parameters that control Parquet search to be configurable. This was a little awkward b/c the SearchOptions is a combination struct of configurable fields and http params. Also, some solutions create circular dependencies. Did what I could, suggestions welcome.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/config.go
+++ b/tempodb/config.go
@@ -24,8 +24,11 @@ const (
 	DefaultBlocklistPollConcurrency = uint(50)
 	DefaultRetentionConcurrency     = uint(10)
 	DefaultTenantIndexBuilders      = 2
-	DefaultPrefetchTraceCount       = 1000
-	DefaultSearchChunkSizeBytes     = 1_000_000
+
+	DefaultPrefetchTraceCount   = 1000
+	DefaultSearchChunkSizeBytes = 1_000_000
+	DefaultReadBufferCount      = 8
+	DefaultReadBufferSize       = 4 * 1024 * 1024
 )
 
 // Config holds the entirety of tempodb configuration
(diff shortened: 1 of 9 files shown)


tempodb/config.go
```diff
@@ -60,8 +63,33 @@ type Config struct {
 }
 
 type SearchConfig struct {
+	// v2 blocks
 	ChunkSizeBytes     uint32 `yaml:"chunk_size_bytes"`
 	PrefetchTraceCount int    `yaml:"prefetch_trace_count"`
```
annanay25, line 68: `PrefetchTraceCount` is used in `vparquet` as well

tempodb/config.go
```diff
@@ -60,8 +63,33 @@ type Config struct {
...
+
+	// vParquet blocks
+	ReadBufferCount int `yaml:"read_buffer_count"`
+	ReadBufferSize  int `yaml:"read_buffer_size"`
```
annanay25, line 72: One other option is to have a `Config` in `encoding/vparquet` and embed that struct into the `SearchConfig` here, kind of like - 

```

type SearchConfig struct {
	ChunkSizeBytes     uint32 `yaml:"chunk_size_bytes"`

	ParquetConfig      vparquet.SearchConfig    `yaml:","`
}
```

annanay25 (commented):
Agree, the current location of the search params seem to be most practical. Can't think of better ideas other than the one suggested below -

tempodb/config.go
mdisibio: Might consider adjusting the field names.  I think our convention is to include `...Bytes` for clarity, like `ChunkSizeBytes`.

tempodb/config.go
joe-elliott: Seems like it's only used during compaction which defaults to 1000 and does not seem configurable.

tempodb/config.go
joe-elliott: That's not a bad idea but we'd need to figure out a way to have that roll forward with future schemas.

tempodb/config.go
annanay25: I'm not sure I understand what you mean - what is not configurable?

joe-elliott: I could be wrong, but the only place I can find us using a "PrefetchTraceCount" in parquet was during compaction. And that is set here and I don't think there's anyway to configure that using yaml.

https://github.com/grafana/tempo/blob/a7a308550b149835cd3dc206c8e00b7bce1d4a0e/tempodb/encoding/common/interfaces.go#L57

docs/tempo/website/configuration/_index.md
```diff
@@ -666,6 +666,16 @@ storage:
             # Example: "prefetch_trace_count: 10000"
             [prefetch_trace_count: <int>]
 
+            # Size of read buffers used when performing search on a vparquet block. This times read_buffer_count
```
knylander-grafana, line 669: Is vparquet a typo?
```suggestion
            # Size of read buffers used when performing search on a parquet block. This times read_buffer_count
```

docs/tempo/website/configuration/_index.md
```diff
@@ -666,6 +666,16 @@ storage:
...
+            # is the total amount of bytes used for buffering when performing search on a parquet block.
```
knylander-grafana, line 670:
```suggestion
            # Size of read buffers used when performing search on a vparquet block. This value times the read_buffer_count
            # is the total amount of bytes used for buffering when performing search on a parquet block.
```

docs/tempo/website/configuration/_index.md
```diff
@@ -666,6 +666,16 @@ storage:
...
+            # Default: 4194304
+            [read_buffer_size_bytes: <int>]
+
+            # Number of read buffers used when performing search on a vparquet block. This times read_buffer_size_bytes
+            # is the total amount of bytes used for buffering when performing search on a parquet block.
```
knylander-grafana, line 675:
```suggestion
            # Number of read buffers used when performing search on a vparquet block. This value times the  read_buffer_size_bytes
            # is the total amount of bytes used for buffering when performing search on a parquet block.
```

knylander-grafana (approved):
Doc updates look good. Minor wording suggestions.

annanay25 (approved):

Issue #1480, opened 2022-06-09
Parquet Blocks Production Readiness


annanay25:
Meta issue to track production readiness of the Apache Parquet block format.

### References

- Design proposal #1407 
- Initial PR adding parquet support #1479 

### Tracking production readiness

- [ ] Fix FindTraceByID [@mdisibio]
  - [x] Fix panic when scanning rowgroups with multiple pages
  - [x] Current bug with find by ID logic? (https://github.com/grafana/tempo/pull/1531)
- [x] Serverless build
  - [x] Make a decision on Google Cloud Run vs. Cloud Functions
  - [x] Swap serverless querying to `vParquet`
  - [x] Rework our CI pipeline
- [ ] Caching  [@annanay25]
  - [ ] Put metadata/footer in real cache.
- [ ] Fine tuning [@joe-elliott]
  - [ ] Row group size (100MB? 10000 traces?)
  - [ ] Read buffer size/count (1MB x 16 currently)
  - [ ] Search shard target bytes per job (30MB, 100MB?)
  - [ ] Sort inner trace content: Sort batches by span name - does this significantly improve the common case of searching for span by name?
- [ ] Compaction [@mdisibio]
  - [x] Pre-fetch _before_ combining?
  - [ ] Fix combiner metrics -- move combiner into versioned encodings
- [ ] Performance/Stability testing in highly multitenant environments
- [ ] Upstream issues: [@mdisibio]
  - [x] https://github.com/segmentio/parquet-go/issues/250 
  - [x] https://github.com/segmentio/parquet-go/issues/254 

#### Nice to haves (non-blocking)
- [ ] CLI (prints schema & column sizes already) [@mdisibio]
  - [x] iterate through a column and debug print all values https://github.com/grafana/tempo/pull/1531

### Future
The following are out of scope for production readiness of parquet blocks, rather its a rough roadmap for the parquet block format. 

- [ ] Cache specific resource attribute columns? `cluster` / `namespace` / `service.name` ..
- [ ] Switch WAL over to `vParquet`
- [ ] Dynamic column support
- [ ] Implement TraceQL over Parquet blocks.

mdisibio:
I would say at this point parquet is production ready :)  And we implemented almost everything on this list, including all of the roadmap laid out here.  Going to go ahead and close.

PR #1531, opened 2022-06-29
Migrate TraceID column from string to byte slice

annanay25:

**What this PR does**:

- Migrates the `TraceID` column from string to byte slice. This is important to ensure the sort order of traces within a parquet file and to ensure the rowgroup bound checks work as expected.
- A new string column `TraceIDHexString` has been added with snappy encoding to allow debugging TraceID bytes in the parquet file.
- Introduces a new cli command to debug print all the values in a rowgroup together with the min/max bounds. The tenantid/blockid/column is configurable, but at the moment it is hardcoded to convert the value into bytes before printing (this is useful to debug print the new TraceID byte slice column). Example run:
```
➜  tempo git:(find-by-id-bug) ✗ go run ./cmd/tempo-cli/... -c tempo-local.yaml list column single-tenant 39499fe3-30d8-407b-8a86-6c4ab0c1b40b TraceID

***************       rowgroup 0      ********************

Min Value of rowgroup [0 0 0 0 0 0 0 0 3 8 223 115 22 133 236 62]
Max Value of rowgroup [0 0 0 0 0 0 0 0 254 175 250 151 237 203 86 123]
[0 0 0 0 0 0 0 0 3 8 223 115 22 133 236 62]
[0 0 0 0 0 0 0 0 4 164 113 64 184 100 128 109]
[0 0 0 0 0 0 0 0 5 83 136 112 106 83 253 201]
[0 0 0 0 0 0 0 0 9 93 224 32 180 88 236 237]
.
.
[0 0 0 0 0 0 0 0 249 44 82 236 206 122 4 202]
[0 0 0 0 0 0 0 0 249 111 113 25 89 198 85 108]
[0 0 0 0 0 0 0 0 249 174 241 192 97 108 129 15]
[0 0 0 0 0 0 0 0 254 175 250 151 237 203 86 123]
```
- ⚠️ There is no migration path provided for blocks created with the old schema as we are still in the iteration phase. Users are expected to delete any `vparquet` blocks created with the old schema ⚠️

Part of #1480 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ Additionally, default label `span_status` is renamed to `status_code`.
 * [BUGFIX] Fix nil pointer panic when the trace by id path errors. [#1441](https://github.com/grafana/tempo/pull/1441) (@joe-elliott)
 * [BUGFIX] Update tempo microservices Helm values example which missed the 'enabled' key for thriftHttp. [#1472](https://github.com/grafana/tempo/pull/1472) (@hajowieland)
 * [BUGFIX] Fix race condition in forwarder overrides loop. [1468](https://github.com/grafana/tempo/pull/1468) (@mapno)
+* [BUGFIX] Fix bug with FindTraceByID logic due to lexicographic ordering of TraceID string. Migrate TraceID column in parquet schema to byte slices. [#1531](https://github.com/grafana/tempo/pull/1531) (@annanay25)
```
joe-elliott, line 34: Until parquet is in a release I don't think we should update the changelog for all the bugfixes/improvements to get it ready. Open to other opinions here.

pkg/util/traceid.go
```diff
@@ -66,3 +66,14 @@ func EqualHexStringTraceIDs(a, b string) (bool, error) {
 
 	return bytes.Equal(aa, bb), nil
 }
+
+func PadTraceIDTo16Bytes(traceID []byte) []byte {
+	if len(traceID) == 16 {
+		return traceID
+	}
+
+	padded := make([]byte, 16-len(traceID))
```
joe-elliott, line 75: likely faster to make a slice of 16 bytes and copy the traceID into it.

joe-elliott (commented):
Some minor thoughts. I think we should let @mdisibio take a look before merging.

It would be more complex, but we could also write custom a `Compare` function that takes slice length into account.

joe-elliott:
oh, i see, there is the min/max information in the parquet file itself that we need these fields to be "naturally" sortable on. So we can't really get tricky and do custom sorting taking into account the length of the string or byte slice.

annanay25:
Yup exactly

pkg/util/traceid.go
joe-elliott: oh, and i think this will panic if len > 16. we should probably just take the least significant 16 bytes if something > 16 shows up here.

CHANGELOG.md
annanay25: Appended to the parquet feature in the CHANGELOG - let me know if this looks ok

mdisibio (approved):

PR #1538, opened 2022-06-30
Fix a few inconsistencies in backend search between block formats

mdisibio:
**What this PR does**:
This PR fixes a few inconsistencies in backend search between block formats.  But also this adds a set of search tests at the tempodb-level, that will help keep things consistent as we make more changes.   Hint:  I will be making more changes.

Fixes:
* v2 check on span name was exact instead of substring
* v2 request start/end times weren't optional
* vParquet was using the wrong tag to check "status.code"

Fixes n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


annanay25:
LGTM.  Great to see parameterised tests on both versions, more to come

PR #1546, opened 2022-07-05
Adjust docs for backend search hedge configuration

### LINKED ISSUE #1543: query-frontend fails to start with the recomended search options
**Describe the bug**
Using the recomended search options described in the [docs](https://grafana.com/docs/tempo/latest/operations/backend_search/#query-frontend) causes the query frontend service to fail to start with the following log:
```
failed parsing config: failed to parse configFile /conf/tempo.yaml: yaml: unmarshal errors:
  line 27: field hedge_requests_at not found in type frontend.SearchConfig
  line 29: field hedge_requests_up_to not found in type frontend.SearchConfig
```

**To Reproduce**
1. Configure the query-frontend with the `hedge_requests_at` and `hedge_requests_up_to` flags.
2. Start the query-frontend with version 1.4.1

**Expected behavior**
Should not crash the container

**Environment:**
 - Infrastructure: kubernetes
 - Deployment tool: yaml

**Additional Context**
query_frontend config:
```yaml
query_frontend:
  # When increasing concurrent_jobs, also increase the queue size per tenant,
  # or search requests will be cause 429 errors.
  max_outstanding_per_tenant: 2000
  search:
    # At larger scales, increase the number of jobs attempted simultaneously,
    # per search query.
    concurrent_jobs: 2000
    # If set to a non-zero value a second request will be issued at the provided duration. Recommended to
    # be set to p99 of search requests to reduce long tail latency.
    hedge_requests_at: 5s
    # The maximum number of requests to execute when hedging. Requires hedge_requests_at to be set.
    hedge_requests_up_to: 3
```

zalegrala:
**What this PR does**:

Here we move the configuration docs for the `hedge_requests_up_to` and `hedge_requests_at` options.  This appears to be an oversight in a previous release.

Fixes #1543

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/tempo/website/operations/backend_search.md
+++ b/docs/tempo/website/operations/backend_search.md
@@ -16,7 +16,7 @@ unless you make some the of the changes detailed here.
 ## Configuration
 
 Queriers and query frontends have additional configuration related
-to search of the backend datastore. 
+to search of the backend datastore.
 Some defaults are currently tuned for a search by trace ID.
 
 ### All components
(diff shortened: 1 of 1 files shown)


zalegrala:
The yaml currently specifies `external_` when the options don't seem to be specific to external endpoints, but apply to all searches generally.  We could take the time now to remove `external_` from the options and document the breaking change in the config.  Thoughts?

mdisibio:
> The yaml currently specifies `external_` when the options don't seem to be specific to external endpoints, but apply to all searches generally. We could take the time now to remove `external_` from the options and document the breaking change in the config. Thoughts?

It looks like the options currently are only used when calling the external endpoints and I think that was the intent.  Tempo can control hedging there, whereas other work goes through the query-frontend job queue which does retries different.   I think the yaml is ok, and to keep these settings specific to external endpoints.

PR #1551, opened 2022-07-06
Dependabot permissions and run interval

zalegrala:
**What this PR does**:

Here we attempt to grant the correct permissions necessary to allow the extra commit action to run the workflow after committing the serverless updates.

Also here is to reduce the run interval of the dependabot jobs.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@ updates:
   - package-ecosystem: 'gomod'
     directory: '/'
     schedule:
-      interval: 'daily'
+      interval: 'weekly'
 
   - package-ecosystem: 'docker'
     directory: '/'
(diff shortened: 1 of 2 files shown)


mdisibio:
Adding a note here:  CI isn't running as expected after the dependabot PRs are updated with the serverless go mod changes.  We think we figured it out and it's:  https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

> When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. For example, if a workflow run pushes code using the repository's GITHUB_TOKEN, a new workflow will not run even when the repository contains a workflow configured to run when push events occur.

Going to try using the other bot token.

mdisibio (approved):

PR #1552, opened 2022-07-06
Break app/config out to config.go

zalegrala:
**What this PR does**:

Here we move the app.Config to its own file.  This makes it slightly easier to find by name and matches many of the other packages of the project.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo/app/app.go
+++ b/cmd/tempo/app/app.go
@@ -3,16 +3,13 @@ package app
 import (
 	"bytes"
 	"context"
-	"flag"
 	"fmt"
 	"io"
 	"net/http"
 	"sort"
-	"time"
 
 	"github.com/go-kit/log/level"
 	"github.com/gorilla/mux"
-	"github.com/grafana/dskit/flagext"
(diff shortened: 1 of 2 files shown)


cmd/tempo/app/app.go
```diff
@@ -58,154 +52,6 @@ var (
...
-	if c.MetricsGeneratorEnabled {
-		features["metrics_generator"] = 1
-	}
-
-	for label, value := range features {
-		ch <- prometheus.MustNewConstMetric(metricConfigFeatDesc, prometheus.GaugeValue, float64(value), label)
-	}
-}
-
 func newDefaultConfig() *Config {
```
kvrhdn, line 55: I'd consider moving this function as well, this is kind of the constructor of `Config`.

zalegrala: Oh good call.  I'll add that.

kvrhdn (approved):
LGTM!

Issue #1448, opened 2022-05-25
Allow set custom X-Scope-OrgID header in remote-write of metrics-generator


luistilingue:
**Describe the bug**
When using remote-write with a custom `X-Scope-OrgID` value in `metrics-generator`, Tempo change the header value to the current one in it. 

This code https://github.com/grafana/tempo/blob/main/modules/generator/storage/config_util.go is the responsable of this changing.

**To Reproduce**
1. Start Tempo 1.4.0+ with metrics-generator enabled and storage config with remote-write + custom `X-Scope-OrgID` header value.
2. Start metric ingestion and check WAL folders - there isn't any with custom header value. The custom header value you've inserted isn't used and the following warning is logged:

`caller=config_util.go:30 tenant=tenantA msg="discarding X-Scope-OrgId header" key=X-Scope-OrgID value=TenantAB`

**Expected behavior**
It'll very great have possibility to use a real custom `X-Scope-OrgID` value due nature of multitenancy and its granularity.

**Environment:**
 - Infrastructure: Kubernetes
 - Deployment tool: Helm

**Additional Context**
Slack conversation: https://grafana.slack.com/archives/C01D981PEE5/p1653420879734839
Configuration:
```
  metrics_generator:
    ring:
      kvstore:
        store: memberlist
    processor:
      span_metrics:
        histogram_buckets: [0.001, 0.002, 0.004, 0.006, 0.008, 0.010, 0.050, 0.100, 0.200, 0.400, 0.800, 1, 1.4, 2, 5, 10, 15, 20, 40, 100, 500, 1000, 10000]
        dimensions:
          - http.status_code
          - db.system
          - service.version
          - telemetry.auto.version
    storage:
      path: /var/tempo/wal
      wal:
        wal_compression: true
      remote_write:
        - url: http://mimir-staging-distributor.mimir-system.svc.cluster.local:8080/api/v1/push
          headers:
            X-Scope-OrgID: tenantAB
          send_exemplars: true
```

kvrhdn:
Agree that Tempo is stricter than necessary. This check was added originally to avoid overwriting the X-Scope-OrgID set internally. Instead I think just logging a warning should be sufficient.

So the behaviour would be:

- if run in single-tenant mode: do not inject X-Scope-OrgID, leave the config alone
- if run in multi-tenant mode: inject the X-Scope-OrgID if not set already; if already set, log a warning that the internal tenant ID will be overwritten

Does this seem reasonable?

[...]

PR #1554, opened 2022-07-07
metrics-generator: do not remove x-scope-orgid header in single tenant modus

kvrhdn:
**What this PR does**:
When metrics-generator is run in single-tenant modus and an `x-scope-orgid` header is set in the remote write config, do not remove it. This is a protection added to multitenant system, in which you don't want each tenant to use the same remote write config. This doesn't make sense in single-tenant setups.

Fixes https://github.com/grafana/tempo/issues/1448

**Checklist**
- [x] Tests updated
- [ ] ~~Documentation added~~


mapno (approved):
LGTM

PR #1556, opened 2022-07-07
Add ephemeral storage to generator for wal storage

zalegrala:
**What this PR does**:

Without this change, the generator uses node storage and is not constrianed, and could lead to resource contention on the node.

Here we include a request/limit resource on the generator deployment and an emptydir to constrian the use of local storage.. A side effect of this change is that metrics generator instances will be asked to terminate if the limit is exceeeded, and a new instance will be created.  Storage values should be sufficiently high in order to prevent unnecessary churn on the deployment and memberlist.  If the requested size is not available, the pod will fail to schedule and could lead to data loss.

Fixes grafana/tempo-squad#91

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/operations/jsonnet/microservices/config.libsonnet
+++ b/operations/jsonnet/microservices/config.libsonnet
@@ -80,8 +80,8 @@
       },
     },
     metrics_generator: {
-      pvc_size: error 'Must specify an ingester pvc size',
-      pvc_storage_class: error 'Must specify an ingester pvc storage class',
+      ephemeral_storage_request: error 'Must specify a generator ephemeral_storage_request size',
+      ephemeral_storage_limit: error 'Must specify a metrics generator ephemeral_storage_limit size',
       replicas: 0,
       resources: {
         requests: {
--- a/operations/jsonnet/microservices/config.libsonnet
+++ b/operations/jsonnet/microservices/config.libsonnet
@@ -80,8 +80,8 @@
       },
     },
     metrics_generator: {
-      ephemeral_storage_request: error 'Must specify a generator ephemeral_storage_request size',
-      ephemeral_storage_limit: error 'Must specify a metrics generator ephemeral_storage_limit size',
+      ephemeral_storage_request_size: error 'Must specify a generator ephemeral_storage_request size',
+      ephemeral_storage_limit_size: error 'Must specify a metrics generator ephemeral_storage_limit size',
       replicas: 0,
       resources: {
         requests: {
--- a/operations/jsonnet/microservices/config.libsonnet
+++ b/operations/jsonnet/microservices/config.libsonnet
@@ -80,11 +80,13 @@
       },
     },
     metrics_generator: {
+      pvc_size: error 'Must specify an ingester pvc size',
+      pvc_storage_class: error 'Must specify an ingester pvc storage class',
       replicas: 0,
       resources: {
-        requests:{
+        requests: {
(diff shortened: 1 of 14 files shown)


operations/jsonnet/microservices/generator.libsonnet
```diff
@@ -1,15 +1,27 @@
...
   local tempo_overrides_config_volume = 'overrides',
 
+  tempo_metrics_generator_pvc::
+    pvc.new()
+    + pvc.mixin.metadata.withName(tempo_data_volume)
+    + pvc.mixin.spec.resources.withRequests({ storage: $._config.generator.pvc_size })
+    + pvc.mixin.spec.withAccessModes(['ReadWriteOnce'])
+    + pvc.mixin.spec.withStorageClassName($._config.generator.pvc_storage_class)
+    + pvc.mixin.metadata.withLabels({ app: target_name })
+    + pvc.mixin.metadata.withNamespace($._config.namespace),
```
annanay25, line 23: Is there a config to specify that it can be deleted after the pod is deleted? Because
- we won't be replaying the metrics WAL, we only need to prevent filling host disk.
- if not, then the pvc will hang around forever even after the pods are rotated

zalegrala: I think there is, but I'll do some research.  Good point.

zalegrala: It looks like the reclaim policy is set on the StorageClass, and we aren't able to override it through the PVC, so I think this is already happening for us.

annanay25: I'm not sure I understand. `StorageClass` is the type of disk attached - SSD/etc - and that has a reclaim policy? Then how does the ingester PVC stay around when the pods are rotated? I'm not sure, but if you have checked and confirmed this, I'm good to move forward

.github/workflows/backport.yml
```diff
@@ -1,33 +0,0 @@
-# This checks merged PRs for labels like "backport release-x.y"
```
annanay25, line 1: Why are all these workflow files being deleted?

zalegrala: Sorry, this was a mistake.

operations/jsonnet/microservices/generator.libsonnet
zalegrala: To answer your question though, the PVC sticks around so that the PV is not deleted.  Depending on the storage class reclaim policy is what happens when the PVC no longer has a claim on the PV.  The `sts` or qdeploy` can be removed independent of the PVC.

zalegrala:
I've made an update here to include an emptyDir and request limits.  Let me know your thoughts.

operations/jsonnet/microservices/config.libsonnet
```diff
@@ -80,11 +80,13 @@
       },
     },
     metrics_generator: {
+      ephemeral_storage_request: error 'Must specify a generator ephemeral_storage_request size',
```
zalegrala, line 83: Do we consider this a breaking change?

annanay25: I guess. Safe to call it out.

annanay25 (approved):
Nice one! LGTM.

annanay25 (approved):

PR #1557, opened 2022-07-08
Extend ReadRange interface to accept shouldCache parameter

annanay25:

**What this PR does**:
Adds a (currently unused) `shouldCache` parameter to the ReadRange interface. This will only be needed for the cache implementation and all other backends can ignore it. PR is in preparation to introduce metadata caching for parquet blocks to speed up queries.

Part of #1480 

**Checklist**
- N/A Tests updated
- N/A Documentation added
- N/A `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/tempodb/backend/backend.go
+++ b/tempodb/backend/backend.go
@@ -39,8 +39,9 @@ type Reader interface {
 	Read(ctx context.Context, name string, blockID uuid.UUID, tenantID string, shouldCache bool) ([]byte, error)
 	// StreamReader is for streaming entire objects from the backend.  It is expected this will _not_ be cached.
 	StreamReader(ctx context.Context, name string, blockID uuid.UUID, tenantID string) (io.ReadCloser, int64, error)
-	// ReadRange is for reading parts of large objects from the backend.  It is expected this will _not_ be cached.
-	ReadRange(ctx context.Context, name string, blockID uuid.UUID, tenantID string, offset uint64, buffer []byte) error
+	// ReadRange is for reading parts of large objects from the backend.
+	// There will be an attempt to retrieve this from cache if shouldCache is true. Cache key will be tenantID:blockID:offset:bufferLength
+	ReadRange(ctx context.Context, name string, blockID uuid.UUID, tenantID string, offset uint64, buffer []byte, shouldCache bool) error
 	// Tenants returns a list of all tenants in a backend
 	Tenants(ctx context.Context) ([]string, error)
 	// Blocks returns returns a list of block UUIDs given a tenant
--- a/tempodb/backend/cache/cache.go
+++ b/tempodb/backend/cache/cache.go
@@ -74,9 +74,7 @@ func (r *readerWriter) ReadRange(ctx context.Context, name string, keypath backe
 
 	err := r.nextReader.ReadRange(ctx, name, keypath, offset, buffer, false)
 	if err == nil && shouldCache {
-		bufCopy := make([]byte, len(buffer))
-		copy(bufCopy, buffer)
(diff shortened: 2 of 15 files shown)


tempodb/backend/backend.go
```diff
@@ -39,8 +39,9 @@ type Reader interface {
 	Read(ctx context.Context, name string, blockID uuid.UUID, tenantID string, shouldCache bool) ([]byte, error)
 	// StreamReader is for streaming entire objects from the backend.  It is expected this will _not_ be cached.
 	StreamReader(ctx context.Context, name string, blockID uuid.UUID, tenantID string) (io.ReadCloser, int64, error)
-	// ReadRange is for reading parts of large objects from the backend.  It is expected this will _not_ be cached.
-	ReadRange(ctx context.Context, name string, blockID uuid.UUID, tenantID string, offset uint64, buffer []byte) error
+	// ReadRange is for reading parts of large objects from the backend.
+	// There will be an attempt to retrieve this from cache if shouldCache is true. Cache key will be tenantID:blockID:offset:bufferLength
```
mapno, line 43: > There will be an attempt to retrieve this from cache if shouldCache is true

I think this could come up as unexpected behaviour from reading the param name. I'd expect to read always try to read from the cache, but only cache in certain scenarios.

I feel like using a wrapper for the cases in which using a cache is possible is a bit clearer (e.g. `NewCachedReader(r Reader)`

mapno (commented):
Left a thought about the approach

tempodb/backend/backend.go
annanay25: Yes, there is a design choice in place in here about whether we want the application to handle storing data in cache (similar to the approach you have mentioned `NewCachedReader`) vs. passing a `shouldCache` parameter and having the underlying reader worry about how to handle caching those objects. 

We have opted for the latter - where Tempo can just specify which objects should be cached, and the underlying backend reader will handle that. If there are strong opinions against this approach we can discuss how the alternate could work. Note that we are already using the `shouldCache` design with `Read(..)` so that will have to be changed as well - https://github.com/grafana/tempo/blob/29ee11386fcd49fa3deb6cf8772d45f22a5fdf91/tempodb/backend/backend.go#L38-L39

mapno: If it's been discussed and in use already, I'm fine with it. I only find the name of the param and its behaviour slightly confusing, but it's not a big deal IMO.

Thanks for clarifying.

annanay25: If the naming is confusing we could change it, open to suggestions :)

tempodb/backend/cache/cache.go
```diff
@@ -58,8 +59,27 @@ func (r *readerWriter) Read(ctx context.Context, name string, keypath backend.Ke
...
+		found, vals, _ := r.cache.Fetch(ctx, []string{k})
+		if len(found) > 0 {
+			copy(buffer, vals[0])
+		}
+	}
+
+	err := r.nextReader.ReadRange(ctx, name, keypath, offset, buffer, false)
+	if err == nil && shouldCache {
+		bufCopy := make([]byte, len(buffer))
+		copy(bufCopy, buffer)
```
joe-elliott, line 78: any particular reason to make a copy?

annanay25: Hmm, I wrote this to protect from any upstream mangling of the buffer, but on second thought I don't think ReadRange results are ever modified in place

joe-elliott (approved):

PR #1560, opened 2022-07-09
Refactor: Remove unused code

joe-elliott:
Cleans up some dead code.

@kvrhdn @mapno I believe the `metrics_helper.go` file was used in metrics generator. Was this left on purpose or just an oversight?


mdisibio (approved):
Approving map_size.go which was obsoleted by #1517

kvrhdn (approved):
`metrics_helper.go` isn't related to the metrics-generator. I think it might have been part of the e2e tests? We have code in there to scrape metrics and do assertions on them.

Issue #1435, opened 2022-05-18
Include spans with span kind producer/consumer in service graphs


kvrhdn:
**Is your feature request related to a problem? Please describe.**
If your architecture contains a queue, it's convention that spans interacting with the queue have span kind producer and consumer (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#span-kind). The current implementation of service graphs only looks for span kind client and server, effectively ommitting all interaction between services that use a queue.

**Describe the solution you'd like**
Service graphs should also track requests between services when they are traversing a queue. Using OTel semantics, it should automatically detect such an edge and visualise it in Grafana.

So:
- the service graphs processor should pick up spans with span.kind producer or consumer
- these edges have to be recorded in the metrics somehow
- Grafana's service graphs page should visualize queues

**Describe alternatives you've considered**
We could drift from OTel conventions and keep using span.kind `client` and `server`, but this doesn't feel right.

**Additional context**

PR #1576, opened 2022-07-14
metrics-generator: include messaging systems and databases in service graphs

kvrhdn:
**What this PR does**:
Include databases and requests across messaging systems in service graphs metrics. We rely on the OpenTelemetry semantic conventions to detect when a span is involved in one of these systems.

Links:
- messaging system: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md
- databases: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md

**Messaging system**: span kind will be `consumer` and `producer`. We add `connection_type="messaging_system"` to the metrics.

**Database**: span kind will be `client` but there will be no matching `server` span (the database is most likely not instrumented). We copy the values from the client side and use that to populate the server latency. Downside: you can't track the client and server-side latency of a request.
We add `connection_type="database"` to the metrics.

Fixes https://github.com/grafana/tempo/issues/1435

**Checklist**
- [x] Tests updated
- [ ] Documentation added


kvrhdn:
For testing: you can run this Tempo image in the intro-to-mlt demo https://github.com/grafana/intro-to-mlt/tree/kvrhdn/add-rabbitmq
I'll add proper test cases soon.

This will generate the following connections:
Screenshot 2022-07-14 at 19 38 49

modules/generator/processor/servicegraphs/store/store.go
```diff
@@ -93,16 +93,22 @@ func (s *store) UpsertEdge(key string, update Callback) (bool, error) {
...
 	}
 
+	edge := newEdge(key, s.ttl)
+	update(edge)
+
+	if edge.isComplete() {
+		s.onComplete(edge)
+		return true, nil
+	}
+
```
mapno, line 103: Nice improvement 👍

mapno (commented):
Don't have a lot to comment. It looks good 👍

kvrhdn:
I'll update the corresponding docs in a separate PR since I've a couple of changes pending and it's just going to clutter this PR.

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -149,20 +149,41 @@ func (p *Processor) consume(resourceSpans []*v1_trace.ResourceSpans) (err error)
 
 		for _, ils := range rs.InstrumentationLibrarySpans {
 			for _, span := range ils.Spans {
+				connectionType := store.Unknown
+
 				switch span.Kind {
+				case v1_trace.Span_SPAN_KIND_PRODUCER:
+					connectionType = store.MessagingSystem
+					fallthrough
```
annanay25, line 157: I think we should add a comment here to emphasise why the producer case was placed before client and similarly for consumer <> server.

annanay25 (approved):

modules/generator/processor/servicegraphs/servicegraphs.go
zalegrala: Whoa, I've never seen `fallthrough` before.  Neat.

zalegrala (approved):

PR #1589, opened 2022-07-19
Added root.name and root.service.name support to v2 and parquet

joe-elliott:
**What this PR does**:
Adds support for two special tags to both parquet and v2. Unsure how this is currently not supported in v2. Both of these special tags are supported in ingester/flatbuffer search.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


mdisibio (approved):
Nice catch

PR #1597, opened 2022-07-22
Test compactor ring for SSB target

### LINKED ISSUE #1594: Scalable Single Binary: Compactor ring not created
The scalable single binary does not correctly create the compactor ring. The consequence of this is that the different instances of the Tempo process will attempt to compact the same blocks repeatedly.

Steps to reproduce:
- start the scalable single binary
- note that `/compator/ring` returns 404

zalegrala:
**What this PR does**:

Without this change, the examples for SSB (Scalabe Single Binary) are missing the configuration necessary for the compactor module to establish cluster membership and coordinate the work of compacting.  Here we ensure that the configuration in the examples works as desired.

Fixes #1594

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/example/docker-compose/scalable-single-binary/tempo-scalable-single-binary.yaml
+++ b/example/docker-compose/scalable-single-binary/tempo-scalable-single-binary.yaml
@@ -30,6 +30,9 @@ ingester:
     heartbeat_period: 100ms
 
 compactor:
+  ring:
+    kvstore:
+      store: memberlist
   compaction:
     compaction_window: 1h              # blocks in this time window will be compacted together
     max_block_bytes: 100_000_000       # maximum size of compacted blocks
--- a/example/docker-compose/scalable-single-binary/docker-compose.yaml
+++ b/example/docker-compose/scalable-single-binary/docker-compose.yaml
@@ -13,6 +13,8 @@ services:
           - tempo
     depends_on:
       - minio
+    ports:
+      - 3200:3200
 
   tempo2:
     image: grafana/tempo:latest
--- a/integration/e2e/config-scalable-single-binary.yaml
+++ b/integration/e2e/config-scalable-single-binary.yaml
@@ -7,6 +7,11 @@ distributor:
       protocols:
         grpc:
 
+compactor:
+  ring:
+    kvstore:
+      store: memberlist
+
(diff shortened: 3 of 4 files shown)


operations/jsonnet/scalable-single-binary/config.libsonnet
```diff
@@ -0,0 +1,45 @@
+{
```
joe-elliott, line 1: i'm concerned about adding more untested jsonnet. we will not keep this up to date as Tempo moves forward.

How do we have good deployment examples like this that we can also make sure remain maintained? If they are not maintained they will end up doing more harm then good.

example/docker-compose/scalable-single-binary/tempo-scalable-single-binary.yaml
```diff
@@ -30,6 +30,9 @@ ingester:
     heartbeat_period: 100ms
 
 compactor:
+  ring:
```
joe-elliott, line 33: ah, is the "fix" just having a correct example?

i thought we defaulted to memberlist if no config was present? maybe we should do that?

operations/jsonnet/scalable-single-binary/config.libsonnet
zalegrala: I'm currently running this in a lab environment, so I know that it does work.  There will be a fast follow PR to include a headless service for member joining.  I can split this out into a different PR if preferred, or if we don't care to maintain another deployment strategy that's fine too.  We also have some Helm to think about if we want to keep all ecosystems updated with all deployment scenarios.

example/docker-compose/scalable-single-binary/tempo-scalable-single-binary.yaml
zalegrala: #1613 has been filed from our recent meeting.  The fix was indeed just to include the ring config in the examples.

operations/jsonnet/scalable-single-binary/config.libsonnet
zalegrala: I've dropped the jsonnet.

joe-elliott (approved):

PR #1608, opened 2022-07-29
Parquet compaction improvements

mdisibio:
**What this PR does**:
This PR improves parquet compaction performance by changing it to operate on raw `parquet.Row`s instead of the unmarshalled `Trace`s where possible.  This is akin to how Tempo already handles v2 proto: proto-bytes are copied between blocks efficiently and only unmarshalled when dedupe is needed.

Benchmarks:
```
name                    old time/op    new time/op    delta
Compaction/v2-12           1.87s ± 5%     1.85s ± 1%     ~     (p=0.829 n=10+8)
Compaction/vParquet-12     1.87s ± 3%     1.14s ± 9%  -39.01%  (p=0.000 n=9+9)

name                    old alloc/op   new alloc/op   delta
Compaction/v2-12           281MB ± 4%     280MB ± 2%     ~     (p=0.730 n=9+9)
Compaction/vParquet-12    2.88GB ± 1%    2.52GB ± 1%  -12.71%  (p=0.000 n=9+7)

name                    old allocs/op  new allocs/op  delta
Compaction/v2-12           1.40M ± 0%     1.41M ± 0%   +0.33%  (p=0.000 n=10+10)
Compaction/vParquet-12     11.0M ± 0%      1.6M ± 0%  -85.48%  (p=0.000 n=10+10)
```

Additionally, the way vParquet cuts row groups has changed to cut them more consistently.  This has been a challenge so far because it's hard to estimate the size of a row group while it's being created (both because of various encodings, but also the sizes of in-memory buffers not exposed by `parquet-go`), so we had settled on a number of traces instead.  This lead to large variability in the size of row groups, but more critically memory usage was also highly unpredictable.  Now it targets a total number of column values, i.e. cut a row group after writing 20M values (any columns, any types).  This drives more consistent behavior across small and large traces, and more predictable memory usage, and is also easy to calculate by `len(row)`.   Luckily there is a close approximation between values and bytes such that we can reuse the existing `FlushSizeBytes` and `RowGroupSizeBytes` config as the number of values, and in my measurement it's between 75-110% accurate.

More notes:
* Add callback for ObjectsCombined metric to fix it for vParquet encoding
* Add tracing instrumentation to compaction.  Very basic, but includes iterators and backend reads/writes.
* Parquet flushes the row group before it reaches the limit, instead of after adding the next trace. This ensures that the setting is a max instead of a min.
* Prefetch around parquet multiblock to allow dedupe to happen concurrently.
* Parquet's main iterator is the new `RawIterator` which returns `(id, row, err)`, but left the existing `Iterator` because it's extremely useful for testing.
* Introduce pooling for when a `Trace` must be un/marshalled, but only pool small-ish buffers.

Fixes n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/compactor.go
```diff
@@ -22,7 +24,7 @@ const (
 
 	DefaultCompactionCycle = 30 * time.Second
 
-	DefaultFlushSizeBytes uint32 = 30 * 1024 * 1024 // 30 MiB
+	DefaultFlushSizeBytes uint32 = 20 * 1024 * 1024 // 30 MiB
```
joe-elliott, line 27: nit: update comment // 20 MiB

tempodb/encoding/vparquet/compactor.go
```diff
@@ -206,3 +262,37 @@ func (b *bookmark) clear() {
...
+
+func (r *rowPool) Put(row parquet.Row) {
+	if len(row) > r.maxSize {
+		return
+	}
+
+	// Clear before putting into the pool.
+	// This is important so that pool entries don't hang
+	// onto the underlying buffers.
+	for i := range row {
```
joe-elliott, line 294: does line 297 already do this by reducing the slice length to 0?

tempodb/encoding/vparquet/create.go
```diff
@@ -57,7 +57,7 @@ func CreateBlock(ctx context.Context, cfg *common.BlockConfig, meta *backend.Blo
 			return nil, err
 		}
 
-		if s.CurrentBufferLength() > cfg.RowGroupSizeBytes || s.CurrentBufferedObjects() > 10_000 {
+		if s.CurrentBufferedValues() > cfg.RowGroupSizeBytes {
```
joe-elliott, line 60: RowGroupSizeBytes...Ish :)

joe-elliott (approved):

tempodb/encoding/vparquet/compactor.go
mdisibio: I don't believe so.  String and []byte `parquet.Value`s contain a pointer to another buffer:
```go
type Value struct {
	ptr *byte
```

This loop erases the pointers to allow those buffers to be freed.  I believe the reference remains if only setting the slice len to 0.

tempodb/encoding/vparquet/create.go
mdisibio: Added a comment which helps a bit, and aligned it with the comment in compactor.go.

PR #1612, opened 2022-08-01
filter out stale spans from metrics generator

ie-pham:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/generator/instance.go
```diff
@@ -44,6 +45,11 @@ var (
 		Name:      "metrics_generator_bytes_received_total",
 		Help:      "The total number of proto bytes received per tenant",
 	}, []string{"tenant"})
+	metricSpansDiscarded = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "tempo",
+		Name:      "metrics_generator_spans_discarded_total",
```
kvrhdn, line 50: I'd consider renaming to `metrics_generator_discarded_spans_total` just to make it similar to this other metric: https://github.com/grafana/tempo/blob/main/modules/overrides/discarded_spans.go#L12 Not a big deal though, both should show up in grafana 🤷🏻 

I think it's good to have separate metrics since a span discarded in the metrics-generator is very different from a span discarded in the ingester/compactor.

modules/generator/config.go
```diff
@@ -31,6 +32,8 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
 	cfg.Processor.RegisterFlagsAndApplyDefaults(prefix, f)
 	cfg.Registry.RegisterFlagsAndApplyDefaults(prefix, f)
 	cfg.Storage.RegisterFlagsAndApplyDefaults(prefix, f)
+	// setting default for max span age before discarding to 30 sec
+	cfg.MaxSpanAge = 30
```
kvrhdn, line 36: I'm curious how this default behaves in practices. I honestly have no clue what a typical latency is between the span creation and ingestion by Tempo.

modules/generator/config.go
```diff
@@ -19,10 +19,11 @@ const (
...
 type Config struct {
-	Ring      RingConfig      `yaml:"ring"`
-	Processor ProcessorConfig `yaml:"processor"`
-	Registry  registry.Config `yaml:"registry"`
-	Storage   storage.Config  `yaml:"storage"`
+	Ring       RingConfig      `yaml:"ring"`
+	Processor  ProcessorConfig `yaml:"processor"`
+	Registry   registry.Config `yaml:"registry"`
+	Storage    storage.Config  `yaml:"storage"`
+	MaxSpanAge int64           `yaml:"max_span_age_sec"`
```
kvrhdn, line 26: We can make this typed
```suggestion
	MaxSpanAge time.Duration   `yaml:"max_span_age_sec"`
```
This allows users to specifcy e.g. `30s` or `5m`. It will also be easier to work with in `updatePushMetrics`.

modules/generator/instance.go
```diff
@@ -261,14 +266,27 @@ func (i *instance) pushSpans(ctx context.Context, req *tempopb.PushSpansRequest)
...
+				} else {
+					expiredSpans++
+				}
+			}
+			ils.Spans = newSpansArr
 		}
 	}
 	metricBytesIngested.WithLabelValues(i.instanceID).Add(float64(size))
 	metricSpansIngested.WithLabelValues(i.instanceID).Add(float64(spanCount))
+	metricSpansDiscarded.WithLabelValues(i.instanceID, "max_age_reached").Add(float64(expiredSpans))
```
kvrhdn, line 289: Also just a nitpick: consider making `max_age_reached` a constant. If we have multiple reasons, we can put them together and add some documentation about each.

kvrhdn (commented):
This looks good! I've left a bunch of comments, no major stuff just some suggestions/nitpicks.

modules/generator/instance.go
```diff
@@ -261,14 +266,27 @@ func (i *instance) pushSpans(ctx context.Context, req *tempopb.PushSpansRequest)
...
+	expiredSpans := 0
 	for _, b := range req.Batches {
 		size += b.Size()
 		for _, ils := range b.InstrumentationLibrarySpans {
 			spanCount += len(ils.Spans)
+			// filter spans that have end time > max_age
+			var newSpansArr []*v1.Span
+			timeNow := time.Now().UnixNano()
+			for _, span := range ils.Spans {
+				if span.EndTimeUnixNano >= uint64(timeNow-i.cfg.MaxSpanAge*1000000000) {
```
joe-elliott, line 278: both sides of this time range need to be checked. If the user sends a span that's 5 days in the future it should not impact metrics.

we have similar code in the wal:
https://github.com/grafana/tempo/blob/5c885f37f1a01b07db1be1f6d6ca04ce1a58113c/tempodb/wal/append_block.go#L235

modules/generator/config.go
joe-elliott: +1 to using a duration

regarding the name. perhaps we could choose something similar to:

https://github.com/grafana/tempo/blob/5c885f37f1a01b07db1be1f6d6ca04ce1a58113c/tempodb/wal/wal.go#L51

modules/generator/instance.go
```diff
@@ -261,14 +266,27 @@ func (i *instance) pushSpans(ctx context.Context, req *tempopb.PushSpansRequest)
 func (i *instance) updatePushMetrics(req *tempopb.PushSpansRequest) {
```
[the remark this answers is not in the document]
joe-elliott: agree that the function needs to be renamed if it's going to mutate the contents of the push request.

Side question. Currently the processor interface takes a complete PushSpansRequest. Can we change that to taking individual spans? Then we could avoid the potentially costly slice reallocations being done here.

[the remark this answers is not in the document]
kvrhdn: > Side question. Currently the processor interface takes a complete PushSpansRequest. Can we change that to taking individual spans? Then we could avoid the potentially costly slice reallocations being done here.

Yeah, that should be possible. Both the span metrics and the service graphs processor loop through the batches anyway and they process spans one by one.
It will be a bit tricky how we deal with the resource attributes though. We currently extract some data out of the resource attributes before looping through the instrumentation library spans individually.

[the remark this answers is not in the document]
joe-elliott: Oh, that is kind of gross. We could nil out the span pointers and make it an expectation of the processors that some spans may be nil.

I think ideally we leave the processors alone and do some clever in place manipulation of the slices to remove the "bad" spans. This logic could get rather gross though.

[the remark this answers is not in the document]
ie-pham: I could filter out the outdated spans inside the `aggregateMetricsForSpan` function and the `consume` function for service graph. But then we wouldn't be keep track of the numbers of spans dropped in this situation

modules/generator/instance.go
ie-pham: Hmm should we make it similar to the other discarded span name or should we keep it similar to the other metrics in the same space?
https://github.com/grafana/tempo/blob/main/modules/generator/instance.go#L39

modules/generator/instance.go
ie-pham: Is there a reason why we pick 5 days?

modules/generator/instance.go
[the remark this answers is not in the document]
kvrhdn: But that would require every processor to implement the same logic which will lead to duplicated work and code. I'd be interested to see how many spans we drop in practice, if only a small amount of batches have to be reallocated the impact will be fine. If we are constantly dropping spans that are too old we might have to re-evaluate.

If we need to get rid of these reallocations we could change the interface of `Processor` so it passes the resource attributes of the batch next to each span.

modules/generator/instance.go
kvrhdn: Oh I see 🙃 Err, either is fine I guess? Maybe a slight preference for keeping it consistent with the other `tempo_metrics_generator_` metrics then.

Naming is hard 😅

modules/generator/instance.go
kvrhdn: I think 5 days was just an example. I think we can start with a symmetrical time range, i.e. use the same duration before and after `time.Now()`. If this doesn't work well we could still break it out into two config options.

knylander-grafana (commented):
Doc updates look good.

modules/generator/config.go
ie-pham: Screen Shot 2022-08-24 at 10 38 14 AM
Screen Shot 2022-08-24 at 10 39 00 AM
These are the ingestion latency numbers on ops for a few days. Do we think setting it at 30s is a right or is it too aggressive? @kvrhdn @joe-elliott

kvrhdn: 30s as a default looks good to me. It seems this should be include 99% of the data while excluding 1% that is lagging behind.
It's also configurable, so other people can switch it up.

kvrhdn (approved):
Nice work!

PR #1625, opened 2022-08-08
Bug fix relating to Cluster Seed File

ie-pham:
**What this PR does**: This PR fixes an issue for GCS when anonymous usage stats is enabled. Due to the cluster seed file being in the same folder, an incorrect tenant id of an empty string gets returned in the List() call. 

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/backend/raw.go
+++ b/tempodb/backend/raw.go
@@ -132,6 +132,8 @@ func (r *reader) ReadRange(ctx context.Context, name string, blockID uuid.UUID, 
 
 func (r *reader) Tenants(ctx context.Context) ([]string, error) {
 	list, err := r.r.List(ctx, nil)
+
+	// this filter is added to fix a GCS usage stats issue that would result in ""
 	filteredList := make([]string, 0)
 	for _, tenant := range list {
 		if tenant != "" && tenant != ClusterSeedFileName {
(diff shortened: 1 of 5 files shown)


tempodb/backend/raw.go
```diff
@@ -129,7 +131,17 @@ func (r *reader) ReadRange(ctx context.Context, name string, blockID uuid.UUID,
 }
 
 func (r *reader) Tenants(ctx context.Context) ([]string, error) {
-	return r.r.List(ctx, nil)
+	list, err := r.r.List(ctx, nil)
+
+	// this filter is added to fix a GCS usage stats issue that would result in ""
+	filteredList := make([]string, 0)
```
mdisibio, line 137: nit: `make(..., 0)` isn't common.  We can alloc with the capacity = `len(list)` or if going with empty, then prefer `var filteredList []string`

mdisibio (approved):

zalegrala (approved):

tempodb/backend/raw.go
```diff
@@ -129,7 +131,17 @@ func (r *reader) ReadRange(ctx context.Context, name string, blockID uuid.UUID,
 }
 
 func (r *reader) Tenants(ctx context.Context) ([]string, error) {
-	return r.r.List(ctx, nil)
+	list, err := r.r.List(ctx, nil)
+
+	// this filter is added to fix a GCS usage stats issue that would result in ""
+	var filteredList []string
```
joe-elliott, line 137: @zalegrala @ie-pham 

minor but we should have preallocated this slice to match the tenants length.

a test on this behavior also would have been nice.

PR #1629, opened 2022-08-10
Parquet improvements

mdisibio:
**What this PR does**:
This PR has several improvements for parquet:

* Honor the `max_bytes_per_trace` limit during compaction and increment the discarded_spans metric.  This limit is interpreted as the length of the trace when protobuf encoded.  Since the exact protobuf isn't available we use an estimate.
* Reduce memory pressure during ingester flush, by adding pooling to parquet StreamingBlock and allocating with proper length in `traceToParquet`
* Reduce memory pressure on compaction by:
  * Reducing the max number of input blocks from 8 to 4. 
  * Removing prefetch. This is very impactful when very large traces (1M spans) are present as it limits the number of them that can be present in memory concurrently.
  * Tuning the rowPool to start larger and not limit the size of pooled objects
  * SortTrace pass by address
* Update parquet-go library
* Add useful tidbits to `compaction complete` message

Fixes n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/encoding/vparquet/compactor.go
```diff
@@ -36,8 +36,7 @@ func (c *Compactor) Compact(ctx context.Context, l log.Logger, r backend.Reader,
 		minBlockStart   time.Time
 		maxBlockEnd     time.Time
 		bookmarks       = make([]*bookmark, 0, len(inputs))
-		pool            = newRowPool(100_000)
-		prefetchSize    = c.opts.IteratorBufferSize / (len(inputs) + 1) // 1 for each input plus multiblock iterator
+		pool            = newRowPool(int(c.opts.FlushSizeBytes))
```
joe-elliott, line 39: this defaults to ~31M and we don't override it anywhere as far as I can tell. it feels weird to chuck this config in here.

does the per tenant max_bytes_per_trace make more sense?

tempodb/encoding/vparquet/compactor.go
```diff
@@ -82,6 +80,21 @@ func (c *Compactor) Compact(ctx context.Context, l log.Logger, r backend.Reader,
 			return rows[0], nil
 		}
 
+		// Total
+		if c.opts.MaxBytesPerTrace > 0 {
```
joe-elliott, line 84: this ditches traces a bit more aggressively than the v2 logic. let's keep an eye on spans discarded as we roll this out.

tempodb/encoding/vparquet/compactor.go
mdisibio: I think either makes sense, as long as it's large enough to prevent allocations for the largest traces.   Now that we have `max_bytes_per_trace` it does seem more appropriate. In the case that there is no limit, it would be zero.  Will change it.

tempodb/encoding/vparquet/compactor.go
mdisibio: Yes that is true, it is comparing `sum(len(inputs))` whereas v2 compared against `len(output)`.  However, in this case the mem usage from parsing and combining a large trace is significant (possibly a few GB).  If we can think of a way to make this more efficient/precise, let's do it.

joe-elliott (approved):

PR #1650, opened 2022-08-17
Parquet: Generic Writer

joe-elliott:
**What this PR does**:
Uses the generic writer for streaming block `Add()` instead of deconstructing and writing. This results in ~20% CPU reduction and ~40% memory reduction in ingesters.

A really gorpy "estimate size" method was added to make guesses about the size of the trace as it is added to the block. This is used for deciding to cut a rowgroup (much like compaction). I'm tempted to remove the current implementation in favor of something quite a bit more simple such as just adding a static value for every span attribute in the trace.

Marking as draft because, as is, this PR trips a GC crash https://github.com/segmentio/parquet-go/issues/299. We will merge this after 1.5 is cut.


tempodb/encoding/vparquet/create.go
```diff
@@ -139,19 +132,29 @@ func (b *streamingBlock) AddRaw(id []byte, row parquet.Row, start, end uint32) e
...
 }
 
 func (b *streamingBlock) CurrentBufferedObjects() int {
 	return b.currentBufferedTraces
 }
 
 func (b *streamingBlock) Flush() (int, error) {
+	// batch write traces
+	if len(b.bufferedTraces) > 0 {
+		_, err := b.pw.Write(b.bufferedTraces)
```
mdisibio, line 151: It's very interesting that this results in ~40% less memory.  I would have expected an increase, since at this point the entire page of traces is resident in memory twice:  `b.bufferedTraces` and within `b.pw`.   Do you think it's worth zeroing out `b.bufferedTraces` to allow GC?  Otherwise it seems possible for Traces to hang around until the streaming block is collected.

mdisibio:
Left a few comments.  Otherwise this PR looks good to me, the numbers are fantastic.

tempodb/encoding/vparquet/create.go
```diff
@@ -215,3 +227,89 @@ func (b *streamingBlock) Complete() (int, error) {
 
 	return n, writeBlockMeta(b.ctx, b.to, b.meta, b.bloom)
 }
+
+// estimateTraceSize attempts to estimate the size of trace in bytes. This is used to make choose
+// when to cut a row group during block creation.
+// TODO: This function regularly estimates lower values then estimateProtoSize() and the size
```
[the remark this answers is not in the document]
joe-elliott: Yeah, throwing in an extra byte per field does bring it a lot closer. I suppose that consistency is as important as accuracy here. As long as it's consistent it will respond in expected ways to adjusting the row sizes.

Currently having both of these methods is a little tedious though. Hopefully we will find ways to get the compaction cycle to use GenericReader/Writer effectively and we can just have one of these methods.

tempodb/encoding/vparquet/create.go
joe-elliott: The memory savings come from simply exerting way less pressure on the GC. Agree we should zero out traces. I will add that.

joe-elliott:
The Parquet library recently fixed the issue blocking this PR: https://github.com/segmentio/parquet-go/pull/323

Dropping draft.

mdisibio (approved):

PR #1662, opened 2022-08-18
Add span size as a new metrics to metrics generator

### LINKED ISSUE #1615: Include size of the spans in span metrics

In the scenario that Tempo is used by multiple teams and that the costs of each team should be charged back, it is essential to know how much volume each team is generating.
Using the span metrics generated by the metrics-generator, it's possible to know the amount of spans per service, cluster, etc. But not the actual size of those spans.

The span metrics processor should (optionally) also omit a metric `traces_spanmetrics_bytes_total` with the size of each received span. This metric has the same label as the existing span metrics.
This metric will behave very similar to `traces_spanmetrics_calls_total`, but instead of counting the spans it just adds up the sizes.

You can currently estimate the volume of each team using the average span size. This is probably close enough, but doesn't take into account that some teams might have heavier spans than others.

**Additional context**

ie-pham:
**What this PR does**: This PR adds span size as a new metrics to metrics generator. 

Fixes #1615

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## main / unreleased
 
+* [ENHANCEMENT] metrics-generator: expose span size as a metric [#1662](https://github.com/grafana/tempo/pull/1662) (@ie-pham)
+
 ## v1.5.0 / 2022-08-17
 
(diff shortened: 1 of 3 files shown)


zalegrala (commented):
Nice this looks good to me.  Re running the failing test.

zalegrala (approved):

PR #1663, opened 2022-08-19
Improve configuration warnings testing and presentation

### LINKED ISSUE #1575: Improve Warnings testing and presentation
Currently we try to print warnings for the user if their config seems like it might be incorrect:

https://github.com/grafana/tempo/blob/701739d14082396c4a304a4c0c818aa320478523/cmd/tempo/app/config.go#L119

Instead of directly logging, let's adjust this function to return a slice of strings that the caller can then log. Then we can:

- Add tests to confirm it returns the expected strings for a given configuration
- Improve visibility by logging lines like "-- CONFIGURATION WARNINGS --" before we spit out the warnings.

Currently warnings can disappear into the logs:

image

frzifus:
~Hey, would be nice to get some feedback if [this](https://github.com/grafana/tempo/pull/1663/commits/2556b099b05f72da76190a915eff9da1c424e088) roughly looks like what you aimed for. If so, i will continue with persisting the change with some tests. :)~

~cc @joe-elliott @zalegrala~

---

**What this PR does**:
This PR changes the behavior of the `cmd/tempo/app/config.CheckConfig` method. Previously, suspicious values were logged individually. Now a list of warnings and descriptions is returned. The caller can then decide whether and how to use this information. In this PR we decided to reproduce the existing logging behavior.

Fixes #1575

**Checklist**
- [x] Tests updated
- [ ] ~Documentation added~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## main / unreleased
 
+* [CHANGE] tempo: check configuration returns now a list of warnings [#1663](https://github.com/grafana/tempo/pull/1663) (@frzifus)
 * [ENHANCEMENT] metrics-generator: expose span size as a metric [#1662](https://github.com/grafana/tempo/pull/1662) (@ie-pham)
 * [ENHANCEMENT] Set Max Idle connections to 100 for Azure, should reduce DNS errors in Azure [#1632](https://github.com/grafana/tempo/pull/1632) (@electron0zero)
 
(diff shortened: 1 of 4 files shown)


joe-elliott (commented):
Love the direction you're taking this. I think a basic test that confirms warnings are returned for various configs is all we need.

frzifus:
perfect, then i will continue. Thanks for feedback @joe-elliott !

frzifus:
mh.. linting should be fixed and hope the other changes did not cause `CI / Test integration e2e suite` timeout ([panic: test timed out after 16m0s](https://github.com/grafana/tempo/runs/7955947843?check_suite_focus=true#step:4:2458)).

frzifus:
Seems my linter fix actually broke the test. I reversed the change and tested it locally. :+1:

PR #1678, opened 2022-08-22
Parquet: Complete Block Search In Ingesters

joe-elliott:
**What this PR does**:
Primarily this PR removes flatbuffer code for complete blocks in exchange for performing search, tag search and tag value search with parquet blocks. This is the first step towards a complete "parquet ingester". This PR also changes the default block version to parquet which required some test updates throughout the codebase.

A very important change to point out is that it breaks local ingester search for the `v2` blocks. In the migration to 2.0 this is something of a "point of no return". Ultimately I'd like to whittle down v2 only capable of trace by id search and require parquet for search by tags/traceql.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/encoding/vparquet/block_search.go
```diff
@@ -317,14 +422,121 @@ func rawToResults(ctx context.Context, pf *parquet.File, rgs []parquet.RowGroup,
...
-	return &tempopb.SearchMetrics{
-		//InspectedTraces: uint32(p.pTraceID.InspectedValues.Load()),
+func makeIterFunc(ctx context.Context, rgs []parquet.RowGroup, pf *parquet.File) func(name string, predicate pq.Predicate, selectAs string) pq.Iterator {
+	return func(name string, predicate pq.Predicate, selectAs string) pq.Iterator {
+		index, _ := pq.GetColumnIndexByPath(pf, name)
+		if index == -1 {
+			// TODO - don't panic, error instead
+			panic("column not found in parquet file:" + name)
+		}
+		if selectAs == "" {
```
mdisibio, line 536: This change increases the amount of cpu work done.  The iterator uses an empty selectAs to avoid saving the values in the IteratorResults where we filter the column but don't need to return the values upstream.

tempodb/encoding/vparquet/block_search.go
```diff
@@ -317,14 +422,121 @@ func rawToResults(ctx context.Context, pf *parquet.File, rgs []parquet.RowGroup,
...
+		}
+	}
+	iter.Close()
+
+	return nil
+}
+
+// searchSpecialTagValues searches a parquet file for all values for the provided column. It first attempts
+// to only pull all values from the column's dictionary. If this fails it falls back to scanning the entire path.
+func searchSpecialTagValues(ctx context.Context, column string, pf *parquet.File, cb common.TagCallback) error {
```
mdisibio, line 470: Does it make sense build this on top of the existing iterator?  A predicate could be used to collect the distinct string values and still make use of the dictionary when present. The logic would be something like:

* KeepColumnChunk: return true
* KeepColumnPage: 
  * if dictionary present: read it, return false (done with this page)
  * else: return true (must read raw values)
* KeepValue: read string, return false

tempodb/encoding/vparquet/block_search.go
joe-elliott: nice catch. removed and `rawToResults` has been updated to explicitly name its columns.

tempodb/encoding/vparquet/block_search.go
joe-elliott: Good call. I swapped to a new predicate `reportValuesPredicate` that works as described. Since it's not a "real" predicate I left it in the `vparquet` folder, but if you think it belongs in `parquetquery` I'd be glad to move it.

tempodb/encoding/vparquet/block_search.go
```diff
@@ -73,71 +84,172 @@ func (b *backendBlock) Search(ctx context.Context, req *tempopb.SearchRequest, o
...
+		if k == LabelStatusCode {
 			code := StatusCodeMapping[v]
-			resourceIters = append(resourceIters, makeIter("rs.ils.Spans.StatusCode", pq.NewIntBetweenPredicate(int64(code), int64(code)), ""))
-		default:
-			otherAttrConditions[k] = v
+			resourceIters = append(resourceIters, makeIter(column, pq.NewIntBetweenPredicate(int64(code), int64(code)), ""))
+			continue
 		}
+
+		traceIters = append(traceIters, makeIter(column, pq.NewSubstringPredicate(v), ""))
```
mdisibio, line 252: Previously, only `RootServiceName` and `RootSpanName` were trace-level iters.  Trying to think how this changes the search behavior... it would have meant that all of the resource-level conditions were asserted on the same batch, but now can match any batch anywhere in the trace.  Thoughts?

tempodb/encoding/vparquet/block_search.go
```diff
@@ -317,14 +422,75 @@ func rawToResults(ctx context.Context, pf *parquet.File, rgs []parquet.RowGroup,
...
+	iter := pq.NewJoinIterator(DefinitionLevelResourceAttrs, []pq.Iterator{
+		makeIter(FieldResourceAttrKey, keyPred, "keys"),
+		makeIter(FieldResourceAttrVal, nil, "values"),
+	}, nil)
+	for {
+		match := iter.Next()
+		if match == nil {
+			break
+		}
+		m := match.ToMap()
```
mdisibio, line 442: It's possible to iterate `match.Entries` looking for `.Key=="values"`, which should be more efficient than calling `.ToMap`.   Interestingly, I don't see that done anywhere in the current code base, but it is a common pattern in the upcoming traceql implementation (so I would consider it an acceptable approach).

modules/ingester/instance_search.go
```diff
@@ -161,25 +165,38 @@ func (i *instance) searchWAL(ctx context.Context, p search.Pipeline, sr *search.
 }
 
 // searchLocalBlocks starts a search task for every local block. Must be called under lock.
-func (i *instance) searchLocalBlocks(ctx context.Context, p search.Pipeline, sr *search.Results) {
-	for _, e := range i.searchCompleteBlocks {
+func (i *instance) searchLocalBlocks(ctx context.Context, req *tempopb.SearchRequest, sr *search.Results) {
+	for _, e := range i.completeBlocks {
+		if e.BlockMeta().Version == v2.Encoding {
```
mdisibio, line 170: Would lean towards removing the encoding-specific logic from ingesters, and instead catching errUnsupported somewhere.  I see that this avoids starting up a bunch of goroutines destined to fail.  Wondering if there is a decent way to push the errUnsupported further up to `tempodb/encoding/versioned.go` instead of the block.XXX methods.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 * [ENHANCEMENT] metrics-generator: expose span size as a metric [#1662](https://github.com/grafana/tempo/pull/1662) (@ie-pham)
 * [ENHANCEMENT] Set Max Idle connections to 100 for Azure, should reduce DNS errors in Azure [#1632](https://github.com/grafana/tempo/pull/1632) (@electron0zero)
+* [CHANGE] Use Parquet for local block search, tag search and tag value search instead of flatbuffers.
+  - Makes Parquet the default encoding.
+  - *BREAKING CHANGE* Local search for traces, tags and tag values no longer works with v2 blocks. It is recommended to manually 
+  clean out the search folder to remove no longer used files. By default these files are located at `/var/tempo/wal/search`.
```
mdisibio, line 8: Is this manual delete actually needed?  I think the search files will be deleted alongside their parent block, should be ok, unless I am missing something.

tempodb/encoding/vparquet/block_search_test.go
```diff
@@ -218,8 +219,42 @@ func TestBackendBlockSearch(t *testing.T) {
...
+		_, exists := attrs[s]
+		require.True(t, exists, s)
+		delete(attrs, s)
+	}
+
+	ctx := context.Background()
+	err := block.SearchTags(ctx, cb, defaultSearchOptions())
+	require.NoError(t, err)
+
+	require.Empty(t, attrs)
```
mdisibio, line 236: I'm finding the inverted nature of this test affects readability.  It seems like we want to assert that every tag is found, and only found once.   Could we collect the unique tags into another map and `require.Equal(t, attrs, foundAttrs)` ?

tempodb/encoding/vparquet/block_search.go
joe-elliott: good point. Added a check to only add the two root attributes to traceIters and the rest to resourceIters

tempodb/encoding/vparquet/block_search.go
joe-elliott: I think I got it from the method above `rawToResults`.

modules/ingester/instance_search.go
joe-elliott: we have to decide how we are going to deprecate v2 blocks. if we release 2.0 w/ no support for searching v2 blocks then we can just make Search return err.Unsupported and remove this code.

If we want to release 2.0 with support for backend search on v2 blocks then we need to leave that code and do something like this. I was thinking we would support search on the backend v2 blocks for at least a minor rev which is why I went this direction.

CHANGELOG.md
joe-elliott: I think so. I'm pretty sure I removed the code that tracks and removes these files completely.

tempodb/encoding/vparquet/block_search_test.go
joe-elliott: reorganized the test

CHANGELOG.md
joe-elliott: Definitely misunderstood the way this works. No the delete was not necessary. Updated changelog.

modules/ingester/instance_search.go
mdisibio: Yes agree on the transition period for v2.  And good with the new flag to control flatbuffer search and startup warnings instead of pushing the logic to the encodings.

mdisibio (approved):
Ran locally and it's working well.  Will be interesting to see the resource savings also.

PR #1710, opened 2022-09-01
Document that local storage backend is supported only in monolithic mode

pavolloffay:
Signed-off-by: Pavol Loffay <p.loffay@gmail.com>

The fact is mentioned here https://github.com/grafana/tempo/blob/97941594d2d4f51da8a45921936483c5e6677ffe/example/docker-compose/local/readme.md#L3 in the example, but it would be great to have it directly in the docs.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/tempo/website/configuration/_index.md
+++ b/docs/tempo/website/configuration/_index.md
@@ -473,7 +473,7 @@ storage:
     trace:
 
         # The storage backend to use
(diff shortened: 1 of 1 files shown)


joe-elliott (approved):
Nice, thx for the clarification

PR #1714, opened 2022-09-02
Add support for storage class configuration

### LINKED ISSUE #1684: Allow configuring S3 StorageClass
Currently I can’t configure grafana tempo to use the SATA S3 StorageClass

Would want to add this as a configurable feature

amitsetty:
**What this PR does**: Adds support for storage class configuration

Fixes #1684

**Checklist**
- [x] Tests updated


tempodb/backend/s3/s3.go
```diff
@@ -138,9 +144,7 @@ func (rw *readerWriter) Append(ctx context.Context, name string, keypath backend
 	var a appendTracker
 	objectName := backend.ObjectFileName(keypath, name)
 
-	options := minio.PutObjectOptions{
-		PartSize: rw.cfg.PartSize,
-	}
+	options := getTempoObjectOptions(rw)
```
joe-elliott, line 147: This does add UserTags here which feels like a bug fix. Can you think of any reason we wouldn't want that?

joe-elliott (commented):
Looks good! One comment and it looks like there's a conflict in the changelog

tempodb/backend/s3/s3_test.go
```diff
@@ -19,6 +19,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const TAGHEADER = "X-Amz-Tagging"
```
mdisibio, line 22: nit:  Please rename these consts to be cased `tagHeader` and `storageClassHeader`

tempodb/backend/s3/s3.go
```diff
@@ -103,14 +103,20 @@ func internalNew(cfg *Config, confirm bool) (backend.RawReader, backend.RawWrite
 	return rw, rw, rw, nil
 }
 
+func getTempoObjectOptions(rw *readerWriter) minio.PutObjectOptions {
```
mdisibio, line 106: nit: Let's rename this method to `getPutObjectOptions`

tempodb/backend/s3/s3_test.go
```diff
@@ -118,17 +121,17 @@ func TestReadError(t *testing.T) {
 	assert.Equal(t, wups, errB)
 }
 
-func fakeServerWithTags(t *testing.T, obj *url.Values) *httptest.Server {
+func fakeServerWithHeader(testedHeaderName string, t *testing.T, obj *url.Values) *httptest.Server {
```
mdisibio, line 124: nit:  It's convention to keep `t *testing.T` as the first argument, can we swap this around?

tempodb/backend/s3/s3.go
amitsetty: Thought about it a little. 
This will just set the user tags to the new user tags during append. 
Sounds to me like something that leads to a negligible performance impact and will only lead to a behavior change if someone changes the user tags (which in this case will lead to the new user tags when appending) 

Seems to me that if anything this is a good thing unless there is a performance change. Do you think that this could lead to a performance change? If so I'll be more than happy to change it

tempodb/backend/s3/s3_test.go
```diff
@@ -185,3 +188,39 @@ func TestObjectBlockTags(t *testing.T) {
...
+				SecretKey:    flagext.SecretWithValue("test"),
+				Bucket:       "blerg",
+				Insecure:     true,
+				Endpoint:     server.URL[7:], // [7:] -> strip http://
+				StorageClass: tc.StorageClass,
+			})
+			require.NoError(t, err)
+
+			ctx := context.Background()
+			_ = w.Write(ctx, "object", backend.KeyPath{"test"}, bytes.NewReader([]byte{}), 0, false)
```
mdisibio, line 222: Does this test check the storage class was actually used?  Comparing to `TestObjectBlockTags` which looks at the results in `obj`.

tempodb/backend/s3/config.go
```diff
@@ -21,4 +21,6 @@ type Config struct {
 	SignatureV2    bool              `yaml:"signature_v2"`
 	ForcePathStyle bool              `yaml:"forcepathstyle"`
 	Tags           map[string]string `yaml:"tags"`
+	StorageClass   string            `yaml:"storage_class"`
+	UserMetadata   map[string]string `yaml:"user_metadata"`
```
mdisibio, line 25: nit: I'd like to remove `user_` and make the yaml just `metadata`, to match tags and also it looks like the underlying s3 api doesn't describe it as "user"-level, maybe it's implied.   Thoughts?

mdisibio (changes requested):
Hi, thanks for the contribution!  Left some comments on naming, and a couple questions.

tempodb/backend/s3/s3.go
mdisibio: Agree it seems like a great bug fix.  It doesn't look like the headers are supported on [UploadPart](https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPart.html) but they are supported on [CreateMultipartUpload](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateMultipartUpload.html).  Since it's working then either the extra headers don't matter or the minio client is handling it.   Have you ran this against real S3 and seen any issues?

amitsetty: Haven't ran it against a real S3. Feel free to check and I'll make any changes necessary :)

tempodb/backend/s3/config.go
amitsetty: Good idea

tempodb/backend/s3/s3_test.go
amitsetty: Good point

mdisibio (approved):
Tested against real S3 and it's working well. Thanks!

PR #1729, opened 2022-09-09
Add query parameter to search API for traceQL queries

kvrhdn:
**What this PR does**:
Extends the search API to accept a query parameter `q` which can optionally contain a traceQL query. Query is already validated but not yet implemented (it will return `501 Not Implemented`).

~~Fixes #<issue number>~~

**Checklist**
- [x] Tests updated
- [ ] ~~Documentation added~~ I'll add docs once traceQL queries are implemented


pkg/api/http.go
```diff
@@ -21,6 +23,7 @@ import (
 const (
 	URLParamTraceID = "traceID"
 	// search
+	urlParamQuery       = "q"
```
kvrhdn, line 26: I noticed Loki and Prometheus both use `query` instead. I think we opted for `q` since this is least likely to conflict with tag lookups using the old top-level query parameters.

Loki's API: https://grafana.com/docs/loki/latest/api/#query-loki
Prometheus' API: https://prometheus.io/docs/prometheus/latest/querying/api/#instant-queries

pkg/tempopb/tempo.proto
```diff
@@ -42,13 +42,15 @@ message TraceByIDMetrics {
 
 // SearchRequest takes no block parameters and implies a "recent traces" search
 message SearchRequest {
+  // TraceQL query
```
joe-elliott, line 45: changing proto field #s is backwards incompatible. just add the query to the end as field 8

kvrhdn: Right, I wasn't sure but forgot to check.

pkg/api/http.go
joe-elliott: I'm good on `q`

joe-elliott (approved):

PR #1731, opened 2022-09-12
Fixing Data Races in Unit Tests

ie-pham:
**What this PR does**: 
There are four tests that fail consistently: 

- TestCompleteBlock
- Test_Memberlist
- TestBackendBlockFindTraceByID
- TestForwarder_pushesQueued

This PR will only fix "TestForwarder_pushesQueued" and "Test_Memberlist" tests since the issues are presented in the unit tests and not the actual code. 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/forwarder_test.go
+++ b/modules/distributor/forwarder_test.go
@@ -94,14 +94,15 @@ func TestForwarder_pushesQueued(t *testing.T) {
 
 	// 10 pushes are buffered, 1 is picked up by the worker
 
-	for i := 0; i < 10; i++ {
+	for i := 0; i < 11; i++ {
 		f.SendTraces(context.Background(), tenantID, keys, rebatchedTraces)
 		fmt.Println("sent trace #: ", i + 1)
 		fmt.Println("Length: ", len(f.queueManagers[tenantID].reqChan))
 	}
 
 	// queue is full with 10 items
-	assert.Equal(t, 11, len(f.queueManagers[tenantID].reqChan))
+	fmt.Println("Length after all sent: ", len(f.queueManagers[tenantID].reqChan))
+	assert.Equal(t, 10, len(f.queueManagers[tenantID].reqChan))
 }
 
 func TestForwarder_shutdown(t *testing.T) {
--- a/modules/distributor/forwarder_test.go
+++ b/modules/distributor/forwarder_test.go
@@ -101,7 +101,7 @@ func TestForwarder_pushesQueued(t *testing.T) {
 
 	wg.Wait()
 	time.Sleep(time.Second)
-	
+
 	// queue is full with 10 items
 	assert.Equal(t, 10, len(f.queueManagers[tenantID].reqChan))
 }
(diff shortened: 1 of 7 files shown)


ie-pham:
Screen Shot 2022-09-12 at 4 58 03 PM
Screen Shot 2022-09-13 at 12 38 31 PM
Screen Shot 2022-09-13 at 12 56 37 PM
Only failed once out of 12 runs (failed one not shown in the screen shot)

modules/distributor/forwarder_test.go
```diff
@@ -136,6 +136,8 @@ func TestForwarder_shutdown(t *testing.T) {
 		assert.Equal(t, 0, len(f.queueManagers[tenantID].reqChan))
 	}()
 
+	time.Sleep(time.Second)
```
electron0zero, line 139: Is this sleep required because we have sleep on line 132?

can we use a waitgroup here, and remove sleep from both places (here and line 132)??

modules/distributor/forwarder_test.go
```diff
@@ -91,11 +91,16 @@ func TestForwarder_pushesQueued(t *testing.T) {
 		assert.Equal(t, 0, len(f.queueManagers[tenantID].reqChan))
 	}()
 
+	wg := sync.WaitGroup{}
+	wg.Add(11)
```
joe-elliott, line 95: move this to line 98 and change it to `wg.Add(1)`. that way if someone changes the for loop length the waitgroup still works.

modules/distributor/forwarder_test.go
```diff
@@ -91,11 +91,16 @@ func TestForwarder_pushesQueued(t *testing.T) {
 		assert.Equal(t, 0, len(f.queueManagers[tenantID].reqChan))
 	}()
 
+	wg := sync.WaitGroup{}
 	// 10 pushes are buffered, 1 is picked up by the worker
 	for i := 0; i < 11; i++ {
+		wg.Add(1)
 		f.SendTraces(context.Background(), tenantID, keys, rebatchedTraces)
+		wg.Done()
```
joe-elliott, line 99: there's no concurrency here. i don't see how using a waitgroup here does anything?

pkg/usagestats/seed_test.go
```diff
@@ -70,6 +71,8 @@ func Test_Memberlist(t *testing.T) {
 	// create a first memberlist to get a valid listening port.
 	initMKV := createMemberlist(t, 0, -1)
 
+	var wg sync.WaitGroup
+	wg.Add(10)
```
zalegrala, line 75: Since the `10` constant here is mentioned twice, perhaps this add should be inside the `for` loop and then just add `1` per iteration.

ie-pham:
Screen Shot 2022-09-23 at 10 07 42 AM
Each of those successful runs has 4 test jobs so it successfully ran 16 times!

zalegrala (approved):
This looks good to me.  The cluster seed still looks stable in the backend after this change.

PR #1732, opened 2022-09-12
Traceql storage implementation

mdisibio:
**What this PR does**:
This PR implements basic functionality for fetching spansets from a parquet block for a given traceql query.  Things remaining include support for the `status`, `duration`, and parent intrinsics, various comparison operators, but the basics are there.  It is currently unused and not connected to the query path, which is actually the impetus for merging it now so we can start iterating on the traceql engine proper and query path.  

Fixes n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/encoding/vparquet/block_traceql.go
```diff
@@ -0,0 +1,1018 @@
...
+// internal consistencies:  operand count matches the operation, all operands in each condition are identical
+// types, and the operand type is compatible with the operation.
+func (b *backendBlock) Fetch(ctx context.Context, req traceql.FetchSpansRequest) (traceql.FetchSpansResponse, error) {
+
+	err := checkConditions(req.Conditions)
+	if err != nil {
+		return traceql.FetchSpansResponse{}, errors.Wrap(err, "conditions invalid")
+	}
+
+	rr := NewBackendReaderAt(ctx, b.r, DataFileName, b.meta.BlockID, b.meta.TenantID)
```
joe-elliott, line 92: use `openForSearch()`? We've consolidate the reader building logic nicely there.

tempodb/encoding/vparquet/block_traceql.go
```diff
@@ -0,0 +1,1018 @@
...
+	// Span-level columns
+	LabelHTTPStatusCode: {columnPathSpanHttpStatusCode, traceql.AttributeScopeSpan, traceql.TypeInt},
+	LabelHTTPMethod:     {columnPathSpanHttpMethod, traceql.AttributeScopeSpan, traceql.TypeString},
+	LabelHTTPUrl:        {columnPathSpanHttpUrl, traceql.AttributeScopeSpan, traceql.TypeString},
+}
+
+// Fetch spansets from the block for the given TraceQL FetchSpansRequest. The request is checked for
+// internal consistencies:  operand count matches the operation, all operands in each condition are identical
+// types, and the operand type is compatible with the operation.
+func (b *backendBlock) Fetch(ctx context.Context, req traceql.FetchSpansRequest) (traceql.FetchSpansResponse, error) {
```
joe-elliott, line 85: Pass in start/total row groups for partial searches? Or save that for when we wire up the query path?

tempodb/encoding/vparquet/block_traceql.go
```diff
@@ -0,0 +1,1018 @@
...
+	"github.com/grafana/tempo/pkg/parquetquery"
+	"github.com/grafana/tempo/pkg/traceql"
+)
+
+// Helper function to create an iterator, that abstracts away
+// context like file and rowgroups.
+type makeIterFunc func(columnName string, predicate parquetquery.Predicate, selectAs string) parquetquery.Iterator
+
+const (
+	columnPathTraceID                  = "TraceID"
```
joe-elliott, line 23: we are doing similar things here and this PR: 

https://github.com/grafana/tempo/pull/1678/files#diff-aae971dafc9cc2527a4b439bb75059ea397e6c72de2292b26d875c456f1742c0R39

We should merge one and consolidate this if possible.

mdisibio: Yep, the structure here contains some additional information needed for traceql including scope and data type.  I propose that we merge #1678 first and then consolidate on this larger structure since it will work for both purposes. The only concern is if there is a more generic way to represent the data type.  Here is uses traceql types which makes sense, but less so for tag-based search.

tempodb/encoding/vparquet/block_traceql.go
mdisibio: Yes was delaying that until deciding a good struct layout because I didn't want to clutter the traceql interfaces with internal details like that and other things (buffer counts/sizes etc).  One possibility is a new `InternalFetchSpansRequest` that composes `traceql.FetchSpansRequest`.   Good with saving that for the next step.

joe-elliott: sounds good :+1:

tempodb/encoding/vparquet/block_traceql.go
mdisibio: I think for now we should leave the structures separate and eventually retire the tag-based one.  The main sticking point is the ability of traceql to distinguish between attributes and intrinsics with the same label.  The query `{ .name = 'foo' }` looks for `name` in the generic key/value columns, but tag-based search directs `name=foo` to the span name column.  Couldn't think of a non-hacky way to make one lookup table work for both.  `{ .status.code=x }` has the same problem.   A minor concern was also making tag-based search use traceql constants.

joe-elliott (approved):

PR #1734, opened 2022-09-13
Remove unused frontend scheduler code

mapno:
**What this PR does**:

Although you could configure the `query-frontend` to work with the query scheduler from Cortex, it was actually never used and failed to work with that setup. This PR cleans up all that code.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


mapno:
v1 could now be renamed to something else, since it's the only supported version.

joe-elliott:
I like this. If we ever need the extended functionality of the scheduler we would go see what mimir/loki are doing now instead of using this old code.

joe-elliott (approved):

PR #1735, opened 2022-09-13
Add hedged request to trace by ID

mapno:
**What this PR does**:

Adds hedging to trace by ID lookups in the `query-frontend`.

Also, deprecates `query-frontend.query_shards` in favor of `query_frontend.trace_by_id.query_shards`

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/frontend/config.go
+++ b/modules/frontend/config.go
@@ -19,19 +19,24 @@ var (
 )
 
 type Config struct {
-	Config               v1.Config    `yaml:",inline"`
-	MaxRetries           int          `yaml:"max_retries,omitempty"`
-	QueryShards          int          `yaml:"query_shards,omitempty"`
-	TolerateFailedBlocks int          `yaml:"tolerate_failed_blocks,omitempty"`
-	Search               SearchConfig `yaml:"search"`
+	Config               v1.Config       `yaml:",inline"`
+	MaxRetries           int             `yaml:"max_retries,omitempty"`
+	QueryShards          int             `yaml:"query_shards,omitempty"`
+	TolerateFailedBlocks int             `yaml:"tolerate_failed_blocks,omitempty"`
+	Search               SearchConfig    `yaml:"search"`
+	TraceByID            TraceByIDConfig `yaml:"trace_by_id"`
 }
 
 type SearchConfig struct {
 	Sharder SearchSharderConfig `yaml:",inline"`
 }
 
+type TraceByIDConfig struct {
+	HedgeRequestsAt   time.Duration `yaml:"hedge_requests_at,omitempty"`
+	HedgeRequestsUpTo int           `yaml:"hedge_requests_up_to,omitempty"`
+}
+
 func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
-	cfg.Config.MaxOutstandingPerTenant = 100
 	cfg.MaxRetries = 2
 	cfg.QueryShards = 20
 	cfg.TolerateFailedBlocks = 0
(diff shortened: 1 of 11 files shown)


knylander-grafana (approved):
Doc updates look good. Thank you for updating the docs.

modules/frontend/config.go
```diff
@@ -19,17 +19,27 @@ var (
...
 
 type Config struct {
-	Config               v1.Config    `yaml:",inline"`
-	MaxRetries           int          `yaml:"max_retries,omitempty"`
-	QueryShards          int          `yaml:"query_shards,omitempty"`
-	TolerateFailedBlocks int          `yaml:"tolerate_failed_blocks,omitempty"`
-	Search               SearchConfig `yaml:"search"`
+	Config               v1.Config       `yaml:",inline"`
+	MaxRetries           int             `yaml:"max_retries,omitempty"`
+	QueryShards          int             `yaml:"query_shards,omitempty"`
```
joe-elliott, line 24: we should move these under `trace_by_id` for v2.

modules/frontend/hedged_requests.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+func newHedgedRequestWare(cfg HedgingConfig) Middleware {
+	return MiddlewareFunc(func(next http.RoundTripper) http.RoundTripper {
+		if cfg.HedgeRequestsAt == 0 {
+			return next
+		}
+		ret, stats, err := hedgedhttp.NewRoundTripperAndStats(cfg.HedgeRequestsAt, cfg.HedgeRequestsUpTo, next)
+		if err != nil {
+			panic(err)
```
joe-elliott, line 33: should we just 500 the call?

modules/frontend/hedged_requests.go
```diff
@@ -0,0 +1,53 @@
...
+		},
+	)
+)
+
+func newHedgedRequestWare(cfg HedgingConfig) Middleware {
+	return MiddlewareFunc(func(next http.RoundTripper) http.RoundTripper {
+		if cfg.HedgeRequestsAt == 0 {
+			return next
+		}
+		ret, stats, err := hedgedhttp.NewRoundTripperAndStats(cfg.HedgeRequestsAt, cfg.HedgeRequestsUpTo, next)
```
joe-elliott, line 31: should we construct this once on line 27  and then use it repeatedly?

mapno: I had the same doubt while coding it, but this is only called once when instantiating the trace by ID round tripper if I'm not mistaken.

modules/frontend/hedged_requests.go
mapno: Hmm, this is a invalid config, I'd rather that Tempo doesn't start.

modules/frontend/config.go
mapno: Done!

modules/frontend/config.go
```diff
@@ -46,6 +56,12 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
 			TargetBytesPerRequest: defaultTargetBytesPerRequest,
 		},
 	}
+	cfg.TraceByID = TraceByIDConfig{
+		Hedging: HedgingConfig{
+			HedgeRequestsAt:   5 * time.Second,
```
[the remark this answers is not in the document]
mapno: Sounds good, updated!

modules/frontend/hedged_requests.go
joe-elliott: you are correct on both this and the below comment. i was confused by the multiple function nesting :+1:

joe-elliott (approved):

PR #1742, opened 2022-09-15
Exit early from sharded search requests

electron0zero:
Exit early when we are done with a search request, we can be done with a search request when:
- we hit the search request limits
- when search request errors out

Before this change, we used to process atleast `concurrent_jobs` sharded requests even when we are done.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/frontend/searchsharding.go
```diff
@@ -232,41 +143,49 @@ func (s searchSharder) RoundTrip(r *http.Request) (*http.Response, error) {
...
+	// in a goroutine and call subCancel() to exit early??
+	// that goroutine will act as watchdog, and cancel the work as early as possible??
+	// as a side-note: it can keep hammering the lock so maybe not a good idea??
+
+	executedReqs := 0
 	for _, req := range reqs {
+		// if shouldQuit is true, terminate and abandon requests
 		if overallResponse.shouldQuit() {
+			// cancel currently running goroutines
+			subCancel()
```
joe-elliott, line 156: instead of doing this. what do you think about passing the `subCancel` func into the searchResponse object? Then, everytime `addRequest` or `setError` or `setStatus` is called you can check at the end of that function and call `subCancel` if necessary.

i.e. the searchResponse object can be responsible for cancelling the context. it would also cancel slightly quicker when the search limit is reached.

electron0zero: makes sense, passing around cancelFunc, and cancelling it in each event on searchResponse.

joe-elliott (approved):
Let's go!

modules/frontend/searchsharding.go
```diff
@@ -283,12 +197,22 @@ func (s searchSharder) RoundTrip(r *http.Request) (*http.Response, error) {
...
 	wg.Wait()
 
+	// print out request metrics
+	cancelledReqs := startedReqs - overallResponse.finishedRequests
+	_ = level.Info(s.logger).Log(fmt.Sprintf(
+		"sharded search request stats, raw_query: %s, total: %d, started: %d, finished: %d, cancelled: %d",
+		r.URL.RawQuery, len(reqs), startedReqs, overallResponse.finishedRequests, cancelledReqs))
+
 	// all goroutines have finished, we can safely access searchResults fields directly now
 	span.SetTag("inspectedBlocks", overallResponse.resultsMetrics.InspectedBlocks)
```
joe-elliott, line 211: We are accessing the fields of overallResponse here, but there are still goroutines out that can modify it. We can either have the searchResponse object stop taking updates once `internalShouldQuit()` returns true. We can add accessors with appropriate locking for the fields that we need. This already partially exists with `result()`.

joe-elliott: wups! nevermind. I forgot we stuck with the `wg.Wait()` style. I was thinking about the channel selection we had discussed. this is good!

PR #1744, opened 2022-09-16
Update readme.md

t00350320:
in latest version, docker-compose's container_name is built by "-" instead of "_".

### DIFF
--- a/example/docker-compose/loki/readme.md
+++ b/example/docker-compose/loki/readme.md
@@ -15,7 +15,7 @@ docker-compose up -d
 ```
 
 At this point, the following containers should be spun up -  
-If your docker-compose shows column "Name" like "loki-loki-1", you'd better modify the "container_name" correspondingly in step 3's query words.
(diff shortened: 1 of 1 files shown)


example/docker-compose/loki/readme.md
```diff
@@ -14,7 +14,8 @@ docker plugin install grafana/loki-docker-driver:latest --alias loki --grant-all
 docker-compose up -d
 ```
 
-At this point, the following containers should be spun up -
+At this point, the following containers should be spun up -  
+If your docker-compose shows column "Name" like "loki-loki-1", you'd better modify the "container_name" correspondingly in step 3's query words.
```
joe-elliott, line 18:
```suggestion
Note: if docker-compose shows a column "Name" like "loki-loki-1", you will need to modify the "container_name" correspondingly in step 3.
```

joe-elliott (commented):
Thanks for noticing and PRing the improvement! I added a small suggestion on the wording.

joe-elliott (approved):
Thanks for the improvement!

PR #1700, opened 2022-08-29
fixed the issue that windows cannot contain filenames with colons (replaced ':' with '+')

kilian-kier:
**What this PR does**:
It saves wal files with '+' filename separator, instead of ':' , because Windows does not allow it

**Checklist**
- [x] Tests updated
- [ ] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #1745, opened 2022-09-19
Refactor: wal.AppendBlock -> interface

joe-elliott:
**What this PR does**:
This PR prepares the codebase to add a Parquet wal by doing the below.

- Changes wal.AppendBlock to be an interface.
- Adjusts the wal.AppendBlock interface to remove the need to pass a model.ObjectCombiner

**Checklist**
- [x] Tests updated
- [ ] Documentation added


joe-elliott:
> Do we want to go ahead and move the v2 wal implementation into `tempodb/encoding/v2`? Or rather, how do you see introducing parquet wal in this structure and making use of `tempodb/encoding/vparquet`?

I'm currently inclined to leave them both in `/tempodb/wal` since for the foreseeable future we will need to create parquet blocks from a v2 wal and vice versa. (This will be true as long as we support v2 overall).

My thought there was to provide 2 separate iteration funcs in the interface. Something like:
```
  Iterator() (common.Iterator, error)
  SomeParquetIterator() (blerg, err)
```
Then when creating the block we can call the preferred method of the two and the AppendBlock is responsible for making the conversion if it's not the native format. There will be overhead when rolling ingesters with a new format (and it has to convert the existing wals).

> Regarding file name parsing: I was thinking that could already accommodate new wal formats, since the filename encodes version/dataencoding, etc. If we work with the tempodb/encoding setup it might let the wal be fairly agnostic to the block format. Also, for these changes seeing a conflict with #1700

My current thought on this is to have the parquet wal user directories like. Something like:
```
<wal folder>/wal+<block-guid>+<tenant id>/000001
                                                        /000002
                                                        ...
```
and store each "segment" of the wal independently instead of trying to force it into the old v2 format.

joe-elliott:
Closed in favor of #1755

Issue #1746, opened 2022-09-20
OTLP HTTP Collector fails with error 'unknown field scopeSpans'


mdisibio:
**Describe the bug**
Tempo's OTLP HTTP Collector is failing to ingest data from apps using the latest OpenTelemetry instrumentation. The error is `{"code":3,"message":"unknown field \"scopeSpans\" in v1.ResourceSpans"}`.   This is due to the [renaming efforts within opentelemetry-proto](https://github.com/open-telemetry/opentelemetry-proto/pull/362).  OTLP gRPC is unaffected because the protobuf is backwards-compatible (same field ids and layout).

**To Reproduce**
1. Start Tempo (SHA or version)
2. Perform Operations (Read/Write/Others)

**Expected behavior**

**Environment:**
 - Infrastructure: [e.g., Kubernetes, bare-metal, laptop]
 - Deployment tool: [e.g., helm, jsonnet]

**Additional Context**

PR #1754, opened 2022-09-22
Update opentelemetry-proto to v0.18.0

mapno:
**What this PR does**:

Updates `opentelemetry-proto` submodule from v0.8.0 to v0.18.0. v0.18.0 is the latest version that still contains the deprecated fields `instrumentation_library...`, which were replaced by `scope...`. All internal types have been updated to use scope fields.

For Tempo, this means that the query by ID API will return the new format (scope-based naming), both in JSON and protobuf. This is a breaking change for JSON.

Contributes to https://github.com/grafana/tempo/issues/1746

**Checklist**
- [x] Tests updated
- [ ] Documentation added


zalegrala (approved):
Looks good to me.  Are there any breaking changes to call out?

joe-elliott:
Nice work. Per @zalegrala's comment please highlight that the breaking change impacts the trace by id endpoint if json is requested. (and whatever else it breaks)

PR #1757, opened 2022-09-22
Update OTel collector to v0.57.2

### LINKED ISSUE #1746: OTLP HTTP Collector fails with error 'unknown field scopeSpans'

mapno:
**What this PR does**:

Updates the OTel collector to v0.57.2. That is the latest version that is backwards compatible with the OTel proto changes introduced in https://github.com/grafana/tempo/pull/1754.

In order to remove the use of our internal OTel collector fork, some parts of the [obsmetrics](https://pkg.go.dev/go.opentelemetry.io/collector/internal/obsreportconfig/obsmetrics) pkg have been copied over.

Fixes #1746 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/distributor/receiver/metrics.go
```diff
@@ -0,0 +1,92 @@
+package receiver
```
mapno, line 1: Should we indicate where this comes from or import it in a different way? I did it this way b/c we only need some bits from the entire pkg and all this wasn't in a single file

mapno:
This PR breaks propagation of the OTel tracer as we drop the opentracing bridge fork.

modules/distributor/receiver/metrics.go
joe-elliott: If we just copied this code from somewhere else then we need to check the license of the repo it came from. If the license requires attribution then, yeah, we should link back to where it was copied from.

modules/distributor/receiver/shim.go
```diff
@@ -118,11 +129,36 @@ func New(receiverCfg map[string]interface{}, pusher BatchPusher, middleware Midd
...
 
-	p := config.NewMapFromStringMap(map[string]interface{}{
-		"receivers": receiverCfg,
+	receivers := make([]string, 0, len(receiverCfg))
+	for k := range receiverCfg {
+		receivers = append(receivers, k)
+	}
+	fmt.Println("receivers", receivers)
+
+	pro, err := service.NewConfigProvider(service.ConfigProviderSettings{
```
joe-elliott, line 138: whoa, this intense. can we get a comment about what all this config does?

cmd/tempo-query/tempo/plugin.go
```diff
@@ -89,7 +89,7 @@ func (b *Backend) GetTrace(ctx context.Context, traceID jaeger.TraceID) (*jaeger
 		return nil, fmt.Errorf("%s", body)
 	}
 
-	otTrace, err := otlp.NewProtobufTracesUnmarshaler().UnmarshalTraces(body)
+	otTrace, err := ptrace.NewProtoUnmarshaler().UnmarshalTraces(body)
```
joe-elliott, line 92: i'm pretty sure i know what the answer is, but i have to ask: does this upgrade provide anyway to get around this nonsense?

joe-elliott (approved):
some comments, but overall lgtm.

enough is changing in the distributor that we should watch performance as this rolls out.

modules/distributor/receiver/metrics.go
mapno: Good point. Added the same license entry and specified where the code codes from.

modules/distributor/receiver/shim.go
mapno: Added a few comments, but I agree is a bit overwhelming. The process of generating the actual config struct has become cumbersome and requires all that scaffolding.

cmd/tempo-query/tempo/plugin.go
mapno: Right, I didn't find a way around it :(

PR #1768, opened 2022-09-28
Parquet Ingester/Compactor memory improvements

joe-elliott:
**What this PR does**:
Refactoring to reduce allocations in `traceToParquet` and `CreateBlock`

- `traceToParquet` use pooled slices. ~levelled pooling would be preferred but we are still seeing a performance improvement with this change only.~
- `v2.DedupingIterator` changed to support both `common.Iterator` and `BytesIterator` allowing it to be used directly when completing a block. Previously, if a trace needed to deduped, we would unmarshal the individual byte slices to `*tempopb.Trace`, deupe, marshal back to a byte slice and finall back to `*tempopb.Trace` to be written to the complete block. This change allows the deduping iterator to directly pass the results of the combine to the caller which reduces allocations significantly.
- Added a leveled pooling object for pooling generic slices.

Previous In Use Space:
image

Previous alloc objects:
image

New In Use Space:
image

New alloc objects:
image

Average old memory usage vs average of this branch:
image

### DIFF
--- a/tempodb/encoding/v2/iterator_deduping.go
+++ b/tempodb/encoding/v2/iterator_deduping.go
@@ -50,29 +50,9 @@ func NewDedupingIterator(iter BytesIterator, combiner model.ObjectCombiner, data
 
 // Next implements BytesIterator
 func (i *dedupingIterator) NextBytes(ctx context.Context) (common.ID, []byte, error) {
-	if i.currentID == nil {
-		return nil, nil, io.EOF
-	}
-
-	var dedupedID []byte
-	currentObjects := [][]byte{i.currentObject}
-
-	for {
-		id, obj, err := i.iter.NextBytes(ctx)
-		if err != nil && err != io.EOF {
-			return nil, nil, err
-		}
-
-		if !bytes.Equal(i.currentID, id) {
-			dedupedID = i.currentID
-
-			i.currentID = id
-			i.currentObject = obj
-			break
-		}
-
-		i.currentID = id
-		currentObjects = append(currentObjects, obj)
+	dedupedID, currentObjects, err := i.next(ctx)
+	if err != nil {
+		return nil, nil, err
 	}
 
 	if len(currentObjects) == 1 {
@@ -93,29 +73,9 @@ func (i *dedupingIterator) Next(ctx context.Context) (common.ID, *tempopb.Trace,
 		return nil, nil, fmt.Errorf("dedupingIterator.Next() called but no decoder set")
 	}
 
-	if i.currentID == nil {
-		return nil, nil, io.EOF
-	}
-
-	var dedupedID []byte
-	currentObjects := [][]byte{i.currentObject}
-
-	for {
-		id, obj, err := i.iter.NextBytes(ctx)
-		if err != nil && err != io.EOF {
-			return nil, nil, err
-		}
-
-		if !bytes.Equal(i.currentID, id) {
-			dedupedID = i.currentID
-
-			i.currentID = id
-			i.currentObject = obj
-			break
-		}
-
-		i.currentID = id
-		currentObjects = append(currentObjects, obj)
+	dedupedID, currentObjects, err := i.next(ctx)
+	if err != nil {
+		return nil, nil, err
 	}
 
 	if len(currentObjects) == 1 {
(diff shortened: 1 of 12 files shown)


tempodb/encoding/vparquet/schema.go
```diff
@@ -563,3 +560,13 @@ func parquetTraceToTempopbTrace(parquetTrace *Trace) *tempopb.Trace {
 
 	return protoTrace
 }
+
+func extendReuseSlice[T any](sz int, in []T) []T {
+	// jpe nil out slice to avoid leaking memory
+	if cap(in) >= sz {
+		// slice is large enough
+		return in[:sz]
```
zalegrala, line 568: Does `len(in)` end up being equal to `len(sz)`?  If so, doesn't that leave the original data at the positions?

tempodb/encoding/vparquet/schema.go
```diff
@@ -193,30 +198,29 @@ func attrToParquet(a *v1.KeyValue) Attribute {
...
 	}
-	return p
 }
 
-func traceToParquet(id common.ID, tr *tempopb.Trace) Trace {
-
-	ot := Trace{
-		TraceIDText: util.TraceIDToHexString(id),
-		TraceID:     util.PadTraceIDTo16Bytes(id),
+func traceToParquet(id common.ID, tr *tempopb.Trace, ot *Trace) *Trace {
```
zalegrala, line 203: This is a hard method to read currently, but I see why its necessary.  The `ot` signature extension looks like it is only used during `CreateBlock()` correct?  Could the calls to `extendReuseSlice` be hoisted into the `CreaetBlock()` method to avoid the extension?

tempodb/encoding/vparquet/schema.go
```diff
@@ -312,17 +317,16 @@ func traceToParquet(id common.ID, tr *tempopb.Trace) Trace {
...
+						attrCount++
 					}
 				}
-
-				oils.Spans = append(oils.Spans, ss)
+				ss.Attrs = ss.Attrs[:attrCount]
 			}
 
-			ob.ScopeSpans = append(ob.ScopeSpans, oils)
+			// jpe is this needed? or is it covered above?
```
zalegrala, line 327: Looks like it isn't needed.

zalegrala (commented):
Still reviewing, but a couple of notes.

tempodb/encoding/vparquet/schema.go
joe-elliott: yup, good thought. the assumption is that the calling code will overwrite every single field on the structs in the slice. if every field is not overwritten then, yes, data from a different trace could "leak" through.

tempodb/encoding/vparquet/schema.go
joe-elliott: it's not :), thank you

tempodb/encoding/vparquet/schema.go
joe-elliott: unsure how we would do that, but if you have a suggestion please lay it on me. an improvement on this code that I don't think would be too difficult would be to pool the slices. currently we are relying on this one trace object which we repeatedly overwrite to "hold" our buffers.

i'm currently working on a "leveled" pooling PR for the upstream parquet library. I could likely do something similar here which would make this an easier read and more effectively use memory.

joe-elliott: Leveled pool added and used.

tempodb/encoding/vparquet/schema.go
joe-elliott: Good call. There were definitely some fields that were not zero'ed out. Added.

tempodb/encoding/vparquet/schema_test.go
```diff
@@ -22,8 +24,90 @@ func TestProtoParquetRoundTrip(t *testing.T) {
...
+		expectedTrace := test.MakeTrace(batches, id)
+
+		parqTr := traceToParquet(id, expectedTrace, trp)
+		actualTrace := parquetTraceToTempopbTrace(parqTr)
+		require.Equal(t, expectedTrace, actualTrace)
+	}
+}
+
+func TestFieldsAreCleared(t *testing.T) {
+	traceID := []byte{0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0E, 0x0F}
```
mdisibio, line 58: Thinking about ways to ensure this test stays updated as we add fields to the `vparquet.Trace` struct.  There are similar fully-populated traces [here](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet/block_traceql_test.go#L395), and [here](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet/block_search_test.go#L31) wonder if we can reuse by converting to tempopb.Trace.  That fully-populated traces should be kept up to date with new columns naturally by the search tests.

mdisibio (commented):
LGTM, but defer to @zalegrala.  Left a few minor thoughts.

tempodb/encoding/v2/iterator_deduping.go
```diff
@@ -75,6 +87,61 @@ func (i *dedupingIterator) NextBytes(ctx context.Context) (common.ID, []byte, er
...
+	}
+
+	if i.currentID == nil {
+		return nil, nil, io.EOF
+	}
+
+	var dedupedID []byte
+	currentObjects := [][]byte{i.currentObject}
+
+	for {
```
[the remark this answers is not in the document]
joe-elliott: good call. done.

zalegrala (approved):
Nice work, this looks good to me.

tempodb/encoding/vparquet/schema_test.go
joe-elliott: - swapped the schema_tests to use `fullyPopulatedTestTrace()`
- removed allFieldsTrace
- extended the trace in `fullyPopulatedTestTrace()` to include more fields
- fixed a few latent bugs in `traceToParquet`

Issue #1769, opened 2022-09-28
Small Multipart Errors with Parquet


amitsetty:
**Describe the bug**

Last block created by compactor is statistically small since the final block is shipped to the backend regardless of its size. 

This leads to final blocks statistically being under 5MB because the compact cycle remainder is shipped regardless of its size.
The outcome is that when using s3 and parquet the following error happens statistically
```
msg="error during compaction cycle" err="error shipping block to backend: error completing multipart upload, object: single-tenant/***, obj etag: : Your proposed upload is smaller than the minimum allowed object size."
```

**To Reproduce**
1. Start Tempo using parquet
2. Use an S3 backend

**Expected behavior**
The compactor should not have statistical errors when using parquet

**Additional Context**
Related to Issue #306
My understanding is that the impact of this is that the final blocks are not being compacted in these cases but also all blocks compacted in the compaction cycle are not marked as compacted even though they have already been flushed.

Sure you guys would have a better understanding how to handle this but some things that come to mind are 
1) Flush last block as final multipart (not sure what the side effects of this are)
2) Leave the final block for the next compaction cycle (not sure what this entails)
3) Always hold 2 blocks in order to flush the last 2 together each taking half the buffer (seems less elegant)

In addition I think we should probably rethink the error situation so that an error with one upload won’t lead to not marking compacted blocks as compacted

Seems like this is where the issue is occurring
[https://github.com/grafana/tempo/blob/97dbd7ccd8dc253991738ae33a0721dbb87edce7/tempodb/encoding/vparquet/compactor.go#L203](url)

joe-elliott:
When compacting parquet blocks we flush a segment to the backend everytime we cut a row group. I'm surprised we're not seeing this internally. It does make sense that the final rowgroup would occasionally be < 5MB.

Our current setting is:

```
storage:
    trace:
        block:
            row_group_size_bytes: 2e+08
```

The tip of main logic is a bit wonky in that the actual row group size it will cut is actually quite a bit smaller than that 30-40MB. I do believe the row_group_size_bytes default is 30MB which would occasionally cut row groups < 5MB.

amitsetty:
Occasionally meaning 1/6 of the time at default setting. #Meaning 1/6 of the time all blocks are flushed twice aside from the last since they aren’t flagged as compacted

Strange to hear you guys haven’t been experiencing this also 

Also our average block is cut around 35MB which seems a bit above the 30MB rather than below in our case

ie-pham:
[...]

PR #1772, opened 2022-09-30
Mark compacted blocks as compacted after errors within maintanence cycle

amitsetty:
**What this PR does**:
Since the compactor first compacts many blocks and then marks them for compaction it is important that if the compactor stops compacting blocks during a maintenance cycle due to an error the code will mark compacted blocks as compacted regardless of whether or not an error occurs after they are shipped 

Partially addresses  #1769

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Old config will still work but will be removed in a future release. [#1735](http
 * [ENHANCEMENT] metrics-generator: filter out older spans before metrics are aggregated [#1612](https://github.com/grafana/tempo/pull/1612) (@ie-pham)
 * * [ENHANCEMENT] Add hedging to trace by ID lookups created by the frontend. [#1735](https://github.com/grafana/tempo/pull/1735) (@mapno)
     New config options and defaults:
+* [BUGFIX] Mark compacted blocks as compacted after errors within maintanence cycle. [#1772](https://github.com/grafana/tempo/pull/1772) @amitsetty
 ```
 query_frontend:
   trace_by_id:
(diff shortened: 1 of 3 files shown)


joe-elliott (commented):
Thanks for the contribution. I am concerned about a few things here and would like to get an understanding of what the goal of this change is.

First, generally if a function returns an error I'm concerned about trusting the other returned parameters. This would impose a very unusual and easy to forget requirement on `compactor.Compact`.

Second, as written if the compactor is halfway through creating a block and fails it will add that block to its internal list to be compacted. I don't think we want that. Attempting to compact a partial block with no meta will definitely fail.

Perhaps if you share some details about what you're attempting to fix we can discuss a change that we feel addresses it.

joe-elliott:
> The reason this seems necessary because certain blocks are being added fully within the for loop and then later there is an error this leading to blocks being fully compacted but not marked for compaction.

I don't believe this is happening. If `compactor.Compact()` returns an error it means that the blocks were not successfully compacted and we shouldn't mark them that way.

A compactor chooses 2-4 eligible blocks, reads, combines their data and writes one block out. If successful it then "marks" the 2-4 input blocks as compacted by renaming `meta.json` to `meta.compacted.json`. A configurable amount of time later (1h by default) it will then delete those blocks permanently.

The issue at hand is multipart uploads are failing on s3 which we believe is related to a bug in the row group size. (@ie-pham is currently researching this) If one of those multipart uploads fails then compaction failed and we don't want to mark the blocks compacted.

amitsetty:
Hey seems like I misunderstood the code. Sorry about that! Seems like this PR is really not solving this issue

joe-elliott:
> Hey seems like I misunderstood the code. Sorry about that! Seems like this PR is really not solving this issue

No worries! We always appreciate the effort. I do think @ie-pham is close to understanding the problem and hopefully will be able to provide a fix soon.

PR #1775, opened 2022-10-03
Implement generic forwarder

Blinkuu:
**What this PR does**:

This PR introduces a new subcomponent to `distributor` named `forwarder`. This generic forwarder serves as a facade for forwarding batches of spans in "fire-and-forget" way. The feature is designed in such a way that it's configurable on per-tenant basis. 

Example configuration:
```yaml
# config.yaml
distributor:
  forwarders:
    - name: "otel-forwarder"
      backend: "otlpgrpc"
      otlpgrpc:
        endpoints: ['otelcol:4317']
        tls:
          insecure: true

# overrides.yaml
overrides:
  "example-tenant-1":
    forwarders: ['otel-forwarder']
  "example-tenant-2":
    forwarders: ['otel-forwarder']
```

**Checklist**
- [x] Tests updated
- [ ] Documentation added


mapno (commented):
Gave it a quick review. I will come back to it later, it's a big PR. Overall I like the direction of the PR.

modules/distributor/queue/queue.go
```diff
@@ -0,0 +1,225 @@
...
+// If the queue is full, the data is dropped
+func (m *Queue[T]) Push(ctx context.Context, data T) error {
+	if m.readOnly.Load() {
+		return fmt.Errorf("queue is read-only")
+	}
+
+	m.pushesTotalMetrics.WithLabelValues(m.name, m.tenantID).Inc()
+
+	reqCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
```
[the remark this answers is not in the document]
Blinkuu: The core idea was to remove `context.Background` that was previously used in queue manager, like this:

```go
m.forwardRequest(context.Background(), req)
```

and add at least some timeout. I've put it here because it's now easy to derive this context from the context that we receive in Push() as an argument if need be. I don't have strong opinions on this, though.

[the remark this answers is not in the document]
Blinkuu: I know understand what you meant when you wrote "context is canceled". You're right, that's not the intended behaviour. I fixed it and moved the context back where it was before (to the queue itself). However, I kept the 10s timeout :)

mapno (commented):
Overall, looks very good to me. I only find the manager logic to be complicated and difficult to follow. Left a comment.

Will wait to see if anyone else wants to chip in.

modules/distributor/forwarder/otlpgrpc/forwarder.go
```diff
@@ -0,0 +1,173 @@
...
+	unaryClientInterceptor, streamClientInterceptor := instrumentation()
+
+	opts = append(
+		opts,
+		grpc.WithTransportCredentials(creds),
+		grpc.WithUnaryInterceptor(grpcmw.ChainUnaryClient(unaryClientInterceptor...)),
+		grpc.WithStreamInterceptor(grpcmw.ChainStreamClient(streamClientInterceptor...)),
+	)
+
+	grpcClientConn, err := grpc.DialContext(ctx, endpoint, opts...)
```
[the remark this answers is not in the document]
Blinkuu: Yeah, I was wondering the same. Tbh, I didn't include WithBlock() as the default option, since I couldn't find it anywhere in the codebase and thought that this is the default you prefer. I personally prefer WithBlock(), as I'm a strong believer that APIs in go should be blocked by default :P Will add this :)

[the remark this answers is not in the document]
Blinkuu: Added in [c946376](https://github.com/grafana/tempo/pull/1775/commits/c946376c740cc37bec99a9c5f484ca5aad12d7c0)

modules/distributor/distributor.go
```diff
@@ -351,6 +362,12 @@ func (d *Distributor) PushBatches(ctx context.Context, batches []*v1.ResourceSpa
 		d.generatorForwarder.SendTraces(ctx, userID, keys, rebatchedTraces)
 	}
 
+	if len(d.cfg.Forwarders) > 0 && len(d.overrides.Forwarders(userID)) > 0 {
+		if err := d.forwardersManager.ForTenant(userID).ForwardBatches(ctx, tempopb.Trace{Batches: batches}); err != nil {
+			_ = level.Warn(d.logger).Log("msg", "failed to forward batches for tenant=%s: %w", userID, err)
```
joe-elliott, line 367: we should metric this in a way we can expose to grafana cloud tracing customers. `tempo_warnings_total` is already piped back to the user and was meant to be a place for generic warnings.

modules/distributor/distributor.go
```diff
@@ -351,6 +362,12 @@ func (d *Distributor) PushBatches(ctx context.Context, batches []*v1.ResourceSpa
 		d.generatorForwarder.SendTraces(ctx, userID, keys, rebatchedTraces)
 	}
 
+	if len(d.cfg.Forwarders) > 0 && len(d.overrides.Forwarders(userID)) > 0 {
```
joe-elliott, line 365: nit: seems unnecessary to check len of `d.cfg.Forwarders`

modules/distributor/forwarder.go
```diff
@@ -77,25 +78,25 @@ func newForwarder(fn forwardFunc, o *overrides.Overrides) *forwarder {
...
+func (f *generatorForwarder) SendTraces(ctx context.Context, tenantID string, keys []uint32, traces []*rebatchedTrace) {
 	select {
 	case <-f.shutdown:
 		return
 	default:
 	}
 
-	qm := f.getOrCreateQueueManager(tenantID)
-	err := qm.pushToQueue(ctx, &request{keys: keys, traces: traces})
+	q := f.getOrCreateQueue(tenantID)
```
joe-elliott, line 88: this would be far easier to review if we had split the refactor from the addition of the new forwarder. is that still possible?

joe-elliott (commented):
This is a large PR that will require several hours to give a proper review to. Is there anyway to break the refactor up from the functionality addition?

modules/distributor/distributor.go
Blinkuu: len(d.cfg.Forwarder) can be 0 in a situation where there is a forwarder defined for a particular tenant in the `overrides.yaml` config, but forwarders are defined in the distributor's `config.yaml`.

Also, this check serves as an additional safety net, it ensures that the feature is completely toggled off when there are no forwarders defined in distributor's `config.yaml`. If there were a bug or anything, it guarantees a super simple way to shut down this feature.

modules/distributor/distributor.go
Blinkuu: Frankly, I don't understand. Do you mean that I cannot use warn-level logs because there is some random side-effect Could you provide some more context?

Blinkuu:
This PR is now rebased on top of #1796. That PR *must* be merged before this one.

modules/distributor/forwarder.go
Blinkuu: Done.

1. Refactor: #1796
2. I've rebased this PR on top of #1796. It will become easier to review after we merge #1796 to origin/main.

modules/distributor/forwarder/manager.go
```diff
@@ -0,0 +1,370 @@
...
+
+func (m *Manager) start(_ context.Context) error {
+	tenantToForwarderNames := m.readOverrides()
+	m.updateTenantToForwarderList(tenantToForwarderNames)
+
+	return nil
+}
+
+func (m *Manager) run(ctx context.Context) error {
+	ticker := time.NewTicker(1 * time.Second)
```
joe-elliott, line 89: let's back this up a bit. 10 seconds? 15 seconds?

modules/distributor/forwarder/manager.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+
+	return result
+}
+
+func (m *Manager) updateTenantToForwarderList(tenantToForwarderNames map[string][]string) {
+	m.tenantToQueueListMu.Lock()
+	defer m.tenantToQueueListMu.Unlock()
+
+	oldTenantIDs := make([]string, 0, len(m.tenantToQueueList))
```
joe-elliott, line 140: i feel like this is unnecessarily complex. why don't we just iterate through `m.tenantToQueueList`. if the tenant is not in `tenantToForwaderNames` then call remove?

we can then call update on anything remaining.

modules/distributor/forwarder/manager.go
```diff
@@ -0,0 +1,370 @@
...
+	queueList, found := m.tenantToQueueList[tenantID]
+	if !found {
+		queueList = newQueueList(m.logger, m.reg, tenantID)
+		m.tenantToQueueList[tenantID] = queueList
+	}
+
+	queueList.update(forwarderNames, m.forwarderNameToForwarder)
+}
+
+func (m *Manager) getQueueList(tenantID string) (*queueList, bool) {
```
joe-elliott, line 193: if this method is only called from `ForTenant()` then why don't we remove it and place the contents in `ForTenant()`?

modules/distributor/forwarder/manager.go
```diff
@@ -0,0 +1,370 @@
...
+	m.tenantToQueueListMu.Lock()
+	defer m.tenantToQueueListMu.Unlock()
+
+	var errs []error
+	for tenantID, ql := range m.tenantToQueueList {
+		if err := ql.shutdown(ctx); err != nil {
+			errs = append(errs, fmt.Errorf("failed to shutdown queuelist for tenantID=%s: %w", tenantID, err))
+		}
+
+		delete(m.tenantToQueueList, tenantID)
```
joe-elliott, line 215: instead of individually deleting create a new, empty map after the loop?

modules/distributor/forwarder/manager.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+
+	for forwarderName, forwarder := range m.forwarderNameToForwarder {
+		if err := forwarder.Shutdown(ctx); err != nil {
+			errs = append(errs, fmt.Errorf("failed to shutdown forwarder with name=%s: %w", forwarderName, err))
+		}
+
+		delete(m.forwarderNameToForwarder, forwarderName)
```
joe-elliott, line 223: same as above

modules/distributor/forwarder/manager.go
```diff
@@ -0,0 +1,370 @@
...
+
+type Overrides interface {
+	TenantIDs() []string
+	Forwarders(tenantID string) []string
+}
+
+type Manager struct {
+	services.Service
+	logger    log.Logger
+	reg       prometheus.Registerer
```
joe-elliott, line 33: is this used? if not remove

modules/distributor/forwarder/manager.go
```diff
@@ -0,0 +1,370 @@
...
+}
+
+type queueListDiff[T any] struct {
+	added   []T
+	removed []T
+}
+
+type queueList struct {
+	logger               log.Logger
+	reg                  prometheus.Registerer
```
joe-elliott, line 236: is this used? if not remove

modules/distributor/forwarder/manager.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+
+	return result
+}
+
+func (l *queueList) update(forwarderNames []string, forwarderNameToForwarder map[string]Forwarder) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+
+	oldForwarderNames := make([]string, 0, len(l.forwarderNameToQueue))
```
joe-elliott, line 268: similar to above there is likely a more straightforward way to write this.

modules/distributor/forwarder/otlpgrpc/forwarder.go
```diff
@@ -0,0 +1,174 @@
...
+	f.connections = connections
+	f.clients = clients
+	f.initialized = true
+
+	return nil
+}
+
+func (f *Forwarder) ForwardBatches(ctx context.Context, trace tempopb.Trace) error {
+	// tempopb.Trace is wire-compatible with OTLP, so we convert to bytes, and back.
+	m, err := trace.Marshal()
```
joe-elliott, line 79: we really need to avoid this marshal/unmarshal dance. we already do this once in the distributor and i've been trying to find a way to remove that for years.

since our trace is wire compatible with otlp we can just marshal it once when we send it and the other side should able to unmarshal.

modules/distributor/forwarder/otlpgrpc/forwarder.go
```diff
@@ -0,0 +1,174 @@
...
+		return fmt.Errorf("failed to unmarshal traces: %w", err)
+	}
+	req := ptraceotlp.NewRequestFromTraces(traces)
+
+	awaitErrs := make(chan error, len(f.clients))
+	wg := sync.WaitGroup{}
+	f.mu.RLock()
+	for _, client := range f.clients {
+		wg.Add(1)
+		go func(client ptraceotlp.Client) {
```
joe-elliott, line 95: this is a lot of goroutines at scale. this needs to send synchronously and rely on the queue to handle slowdowns

joe-elliott (commented):
I think we can reduce the complexity of this PR significantly. Let me know if some of these comments are unclear.

modules/distributor/distributor.go
zalegrala: If we don't add a metric to this PR, it might be handy to have an issue to make sure we come back to it.  A distributor metric might be handy for us internally too.

modules/distributor/forwarder/config.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+type Config struct {
+	Name     string          `yaml:"name"`
+	Backend  string          `yaml:"backend"`
+	OTLPGRPC otlpgrpc.Config `yaml:"otlpgrpc"`
+}
+
+// RegisterFlagsAndApplyDefaults registers flags and applies defaults
+func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet) {
```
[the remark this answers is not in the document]
Blinkuu: Removed

modules/distributor/forwarder/forwarder.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/grafana/tempo/modules/distributor/forwarder/otlpgrpc"
+	"github.com/grafana/tempo/pkg/tempopb"
+)
+
+type Forwarder interface {
+	ForwardBatches(ctx context.Context, trace tempopb.Trace) error
+	Shutdown(ctx context.Context) error
+}
+
+type List []Forwarder
```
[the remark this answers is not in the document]
Blinkuu: I don't think the `List` type should implement the `Forwarder` interface in its entirety. This type is just for convenience.

The underlying forwarders are basically queues (`[]queue.Queue` wrapped in a `queueAdapter` to implement a `Forwarder` interface). The shutdown method is called on them in a couple of scenarios:

1.  `forwarder.Manager` is stopped. This triggers a call to `Manager.shutdown()` method which calls shutdown on all underlying queues via a `queueList.shutdown()`. Additionally, `Manager.shutdown()` will also call shutdown on the real forwarders.
2. queueList gets deleted when a configuration for a particular tenant is removed. This happens in `Manager.removeQueueListForTenantUnderLock()`.
3. queueList gets updated when new configuration removes a reference to a forwarder for a particular tenant. This happens in `queueList.update()` while iterating over `diff.removed` slice.

modules/distributor/forwarder/manager.go
Blinkuu: Removed, was used before.

modules/distributor/forwarder/manager.go
Blinkuu: Changed to 10 seconds.

modules/distributor/forwarder/manager.go
Blinkuu: Removed

modules/distributor/forwarder/otlpgrpc/config.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/grafana/tempo/pkg/util"
+)
+
+type Config struct {
+	Endpoints flagext.StringSlice `yaml:"endpoints"`
+	TLS       TLSConfig           `yaml:"tls"`
+}
+
+// RegisterFlagsAndApplyDefaults registers flags and applies defaults
+func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet) {
```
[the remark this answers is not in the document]
Blinkuu: Removed

modules/distributor/forwarder/otlpgrpc/forwarder.go
Blinkuu: Made is synchronous

modules/distributor/forwarder/otlpgrpc/forwarder.go
Blinkuu: The problem is that the `Export()` method of `ptraceotlp.Client` accepts a `ptraceotlp.Request` as an argument, and I don't think there is a simple way to just convert `tempopb.Trace` type directly into the request type. I'd rather not do some reinterpret cast magic.

I don't know enough about Go, but in C++ such stuff would most likely be optimized away :D

modules/distributor/forwarder/manager.go
```diff
@@ -0,0 +1,370 @@
...
+	return m, nil
+}
+
+func (m *Manager) ForTenant(tenantID string) List {
+	list, found := m.getQueueList(tenantID)
+	if !found {
+		return nil
+	}
+
+	return list.copy()
```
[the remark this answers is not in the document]
Blinkuu: We cannot return the original slice. First of all, there is no "original" slice. `list.copy()` constructs a slice from a map. Secondly, the underlying map can (and will) change on overrides configuration updates.

If we want to get rid of those allocations, perhaps the `ForTenant()` could be changed to `ForwardBatchesForTenant()` and we would just call forward batches on each queue under a lock.

Let me know your thoughts.

modules/distributor/forwarder/manager.go
Blinkuu: I don't think it's that simple. This solution does not cover new tenant IDs showing up in tenantToForwarderNames. To be honest, I like the "diff" solution. Perhaps it's a little more code than if I were to just iterate over those two maps, but IMO it's very descriptive.

modules/distributor/forwarder/manager.go
Blinkuu: I see no value in removing this function

modules/distributor/forwarder/manager.go
Blinkuu: Done

modules/distributor/forwarder/manager.go
Blinkuu: Done

modules/distributor/forwarder/manager.go
Blinkuu: Yeah, similar comment here.

modules/distributor/forwarder/otlpgrpc/forwarder.go
mapno: You can build the request with `ptraceotlp.NewRequestFromTraces(td)` if the conversion is moved up from the [shim](https://github.com/grafana/tempo/blob/main/modules/distributor/receiver/shim.go#L270-L283) to the distributor's `PushBatches`. The downside is that we would leak OTel to the distributor, but IMO it's worth for avoiding the conversion.

Blinkuu: I don't have a strong opinion on this. @joe-elliott which option would you prefer?

modules/distributor/forwarder/otlpgrpc/forwarder.go
mapno: Optionally, we could use a [boundedwaitgroup](https://github.com/grafana/tempo/tree/main/pkg/boundedwaitgroup) per forwarder

Blinkuu: I think keeping it synchronous is good enough. On top of that, I think this feature would be primarily used in "single forwarder-single endpoint" manner. Defining multiple endpoints is kind of an edge case.

modules/distributor/forwarder/otlpgrpc/forwarder.go
```diff
@@ -0,0 +1,164 @@
...
+			errs = append(errs, fmt.Errorf("failed to close grpc connection for endpoint=%s: %w", endpoint, err))
+		}
+
+		delete(f.connections, endpoint)
+	}
+
+	return multierr.Combine(errs...)
+}
+
+func (f *Forwarder) exportTracesUsingClient(ctx context.Context, client ptraceotlp.Client, req ptraceotlp.Request) error {
```
mapno, line 118: Now that sending traces is sync we could remove this fn for simplicity

modules/distributor/forwarder/forwarder.go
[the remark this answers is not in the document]
joe-elliott: i'm fine with this. just noticed it partially implemented the interface and wondered if you saw value in doing the whole thing.

modules/distributor/forwarder/manager.go
joe-elliott: less code?

modules/distributor/forwarder/manager.go
joe-elliott: I propose this instead:

- queueList is immutable. it cannot be updated (cleans up all that code)
- on update of forwarder manager
  - loop through all existing queueLists. remove all that are to be deleted and all that need to be updated
  - call shutdown on all of these queueLists (removed and updated). this will cause them to properly drain async.
  - readd new queueLists for all updated ones that will be used by future pushes

this will clean up all the diffing code and all of the queueList update code. It will also simplify this code significantly. Since queue lists become immutable you can likely also remove all locking code there.

modules/distributor/forwarder/manager.go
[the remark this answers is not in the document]
joe-elliott: Making the queueList immutable as described below opens up the possibility to return a slice owned by the queueList for reduced allocations. I'll admit I'm not honestly sure what the impact will be here. It will certainly be far less then the proto marshalling discussion below.

modules/distributor/forwarder/otlpgrpc/forwarder.go
```diff
@@ -0,0 +1,164 @@
...
+		clients:     make(map[string]ptraceotlp.Client),
+		initialized: false,
+		mu:          &sync.RWMutex{},
+	}, nil
+}
+
+// Dial creates client connections and clients based on config.
+// Dial is expected to be called only once.
+func (f *Forwarder) Dial(ctx context.Context, opts ...grpc.DialOption) error {
+	f.mu.Lock()
```
joe-elliott, line 51: i would really like to get rid of the locking code in this object. i think we can do that using the concurrency controls in the forwarder manager one package up. basically, making sure that lifecycle events Dial, ForwardBatches, and Shutdown are never called concurrently. it may already be true.

modules/distributor/forwarder/otlpgrpc/forwarder.go
joe-elliott: > I'd rather not do some reinterpret cast magic.
100% agreed

I like @mapno's solution here.

modules/distributor/forwarder/manager.go
mapno: I like this proposal 👍 It's very close to the logic from the original forwarder, so it makes sense to me.

modules/distributor/forwarder/otlpgrpc/forwarder.go
Blinkuu: I wouldn't change this. The idea is that many tenants can "attach" to a single Forwarder instance. The flow is kind of like this:

Distributor -> Queue (per tenant) -> Forwarder

Many tenants share the same forwarder. Also, the Dial is just called once, during Tempo startup, and `ForwardBatches` only acquires a read-lock and does so on Queue's thread (in workers, essentially). I don't think we have much to gain with this micro-optimization.

modules/distributor/forwarder/otlpgrpc/forwarder.go
Blinkuu: Done in [2de3a5c](https://github.com/grafana/tempo/pull/1775/commits/2de3a5c157ebb23eec7d1f21af567751ace2b6ef).

modules/distributor/forwarder/manager.go
Blinkuu: Done in [1871143](https://github.com/grafana/tempo/pull/1775/commits/187114373c152431a9ecdda8f4cf9136d3e81cca).

modules/distributor/forwarder/manager.go
Blinkuu: Done in [57fde51](https://github.com/grafana/tempo/pull/1775/commits/57fde5127c307c25414c4c5eb2bfe6afeeb1f818).

`queueList` is now immutable

mapno (commented):
I think it's almost there. Great work 👏

modules/distributor/forwarder/manager.go
```diff
@@ -0,0 +1,283 @@
...
+
+			delete(m.tenantToQueueList, tenantID)
+
+			continue
+		}
+
+		if ql.shouldUpdate(forwarderNames) {
+			if err := ql.shutdown(context.Background()); err != nil {
+				_ = level.Warn(m.logger).Log("msg", "failed to shutdown queue list", "tenantID", tenantID)
+			}
```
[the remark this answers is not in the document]
Blinkuu: Done in [e94a896](https://github.com/grafana/tempo/pull/1775/commits/e94a896103fdae5d7a4d24b9d0c5511c3f237eef).

modules/distributor/distributor.go
```diff
@@ -223,10 +227,18 @@ func New(cfg Config, clientCfg ingester_client.Config, ingestersRing ring.ReadRi
 
 		subservices = append(subservices, d.generatorsPool)
 
-		d.generatorForwarder = newGeneratorForwarder(logger, reg, d.sendToGenerators, o)
+		d.generatorForwarder = newGeneratorForwarder(logger, d.sendToGenerators, o)
 		subservices = append(subservices, d.generatorForwarder)
 	}
 
+	forwardersManager, err := forwarder.NewManager(d.cfg.Forwarders, logger, o)
```
joe-elliott, line 234: ultimately this calls `grpc.DialContext()` which (i believe) attempts an asynchronous tcp connection. if so, then the code is fine as is.

However, if `grpc.DialContext()` blocks and returns an error on an unsuccessful connection then we need to make changes. i.e. If a forwarder endpoint is down then it is important that a distributor will start successfully.

Can you confirm the behavior either way?

modules/distributor/forwarder/otlpgrpc/forwarder.go
joe-elliott: For me, it's less about optimization and more about only using locks where necessary to reduce code complexity. However, I consider this minor am fine with leaving it in.

joe-elliott (commented):
One final question and then I'm good.

modules/distributor/distributor.go
Blinkuu: Done in [28e63a1](https://github.com/grafana/tempo/pull/1775/commits/28e63a15e5861ff2f40c09d7f48aacca158abec7).

mapno (approved):
LGTM

PR #1776, opened 2022-10-03
Parquet Go Update

joe-elliott:
Catches a few updates to parquet go. In particular it brings in a memory improvement for ingesters and compactors.


mapno (approved):
LGTM

PR #1777, opened 2022-10-04
tempo/modules/overrides: Preallocate memory

florianl:
**What this PR does**:

Preallocate memory instead of enforcing an incremental growth. This will result in less work for the garbage collector.

Trivial fix.

**Checklist**
- [ ] ~Tests updated~
Trivial change.
- [ ] ~Documentation added~
Trivial change.
- [ ] ~`CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`~
Trivial change.

### DIFF
--- a/modules/overrides/list_to_map.go
+++ b/modules/overrides/list_to_map.go
@@ -15,7 +15,7 @@ var _ json.Unmarshaler = (*ListToMap)(nil)
 
 // MarshalYAML implements the Marshal interface of the yaml pkg.
 func (l ListToMap) MarshalYAML() (interface{}, error) {
-	list := make([]string, 0)
+	list := make([]string, 0, len(l))
 	for k := range l {
 		list = append(list, k)
 	}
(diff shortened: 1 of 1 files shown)


modules/overrides/list_to_map.go
```diff
@@ -15,7 +15,7 @@ var _ json.Unmarshaler = (*ListToMap)(nil)
 
 // MarshalYAML implements the Marshal interface of the yaml pkg.
 func (l ListToMap) MarshalYAML() (interface{}, error) {
-	list := make([]string, 0)
+	list := make([]string, 0, len(l))
 	for k := range l {
 		list = append(list, k)
```
zalegrala, line 20: I think if we have the slice of size len(l), then we'll also want to use the index to set the list values rather than append.  Append will continue to grow the size of the slice by starting at the end.

joe-elliott: this change sets the capacity, not the length. it will be fine.

joe-elliott (approved):
Appreciate the improvement!

zalegrala (approved):

PR #1785, opened 2022-10-05
Util `client` uses protobuf when querying traces by ID

mapno:
**What this PR does**:

Util `client` uses protobuf when querying traces by ID. This prevents a breaking change when we upgraded `opentelemetry-proto` to v0.18.0.

### DIFF
--- a/pkg/util/client.go
+++ b/pkg/util/client.go
@@ -8,8 +8,8 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/gogo/protobuf/jsonpb"  //nolint:all
-	"github.com/golang/protobuf/proto" //nolint:all
+	"github.com/golang/protobuf/jsonpb" //nolint:all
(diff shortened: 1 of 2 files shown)


pkg/util/client.go
```diff
@@ -18,6 +18,10 @@ const (
 	orgIDHeader = "X-Scope-OrgID"
 
 	QueryTraceEndpoint = "/api/traces"
+
+	AcceptHeader        = "Accept"
```
joe-elliott, line 22: i think these are already defined in `/pkg/http`

joe-elliott (approved):
one suggestion, but otherwise lgtm.

i'm surprised integration tests didn't catch this.

pkg/util/client.go
mapno: Yes, but that causes an import cycle. They could be moved here, but I think they're better suited in `pkg/http.go`. What I can do is un-export them.

PR #1786, opened 2022-10-06
Metrics generator: extract status_message field from spans

### LINKED ISSUE #1764: Ability to access status_message standard field from `span_metrics`
I'm trying to extract the status.message attribute from a span using span_metrics processor, but it doesn't seem to work. I can see in my traces that this attribute is set, but Tempo metric-generator will not pick it up. I wonder if it is the same issue as in this OpenTelementry bug: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/14175

I should be able to do:
```
          span_metrics:
            dimensions:
            - status.message
```
and pull the status.message out from Span attributes.

I tried to use OpenTelemetry Collector to do the same (with the attribute processor), but it is not feasible there too.

**Additional context**
I had a slack conversation with Koenraad, who acknowledge the issue:
"So it seems to be a field set on the Status object of a span (so next to status.code). The way the span metrics processor is designed we only look for field names in the span and resource attributes. status.message isn’t an attribute but a built-in field and can not be looked up this way. If you want you can open an issue for this, we can consider adding it so that if you specify status.message we now to look in the status object."

stoewer:
**What this PR does**:
When generating span metrics, extract the non attribute field `status_message` or `status.message`, when it is provided in the span_metrics dimensions.

Fixes #1764

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/tempo/website/configuration/_index.md
+++ b/docs/tempo/website/configuration/_index.md
@@ -249,7 +249,7 @@ metrics_generator:
 
             # Additional dimensions to add to the metrics along with the default dimensions
             # (service, span_name, span_kind and span_status). Dimensions are searched for in the
-            # resource and span attributes as well as the span status. Listed dimensions are 
+            # resource and span attributes as well as `status.message`. Listed dimensions are 
             # added to the metrics if present.
             [dimensions: <list of string>]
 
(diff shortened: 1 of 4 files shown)


mapno (approved):

docs/tempo/website/configuration/_index.md
```diff
@@ -249,7 +249,8 @@ metrics_generator:
 
             # Additional dimensions to add to the metrics along with the default dimensions
             # (service, span_name, span_kind and span_status). Dimensions are searched for in the
-            # resource and span attributes and are added to the metrics if present.
+            # resource and span attributes as well as the span status. Listed dimensions are
```
mapno, line 252:
```suggestion
            # resource and span attributes as well as `span.status`. Listed dimensions are 
```

More explicit, WDYT?

PR #1795, opened 2022-10-12
Add jsonnetfmt make target and format existing files.

pstibrany:
**What this PR does**:
This PR adds `jsonnetfmt` Makefile target and formats existing files. This PR does NOT add any CI check for formatted files, although I think that would be a valuable check to have. If you agree, I can send it in separate PR.

**Checklist**
- [na] Tests updated
- [na] Documentation added
- [na] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -116,6 +116,14 @@ make fmt
 
 However, do note that the above command requires the `gofmt` and `goimports` binaries accessible via `$PATH`.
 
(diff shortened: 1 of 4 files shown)


joe-elliott (approved):
Thanks @pstibrany!

If we are going to have this, then agree we should enforce it in CI. if you have time to put that together we would appreciate it.

PR #1796, opened 2022-10-12
Refactor queueManager into generic queue.Queue

Blinkuu:
**What this PR does**:
This PR refactors queueManager into a generic queue.Queue. It is a prerequisite for #1775.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/forwarder.go
+++ b/modules/distributor/forwarder.go
@@ -25,12 +25,12 @@ var (
 	metricForwarderPushes = promauto.NewCounterVec(prometheus.CounterOpts{
 		Namespace: "tempo",
 		Name:      "distributor_forwarder_pushes_total",
-		Help:      "Total number of successful requests queued up for a tenant to the generatorForwarder",
+		Help:      "Total number of successful requests queued up for a tenant to the generatorForwarder. This metric is now deprecated in favor of tempo_distributor_queue_pushes_total.",
 	}, []string{"tenant"})
 	metricForwarderPushesFailures = promauto.NewCounterVec(prometheus.CounterOpts{
 		Namespace: "tempo",
 		Name:      "distributor_forwarder_pushes_failures_total",
-		Help:      "Total number of failed pushes to the queue for a tenant to the generatorForwarder",
+		Help:      "Total number of failed pushes to the queue for a tenant to the generatorForwarder. This metric is now deprecated in favor of tempo_distributor_queue_pushes_failures_total.",
 	}, []string{"tenant"})
 )
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 * [ENHANCEMENT] Refactor queueManager into generic queue.Queue [#1796](https://github.com/grafana/tempo/pull/1796) (@Blinkuu)
   - **BREAKING CHANGE** Rename `tempo_distributor_forwarder_queue_length` metric to `tempo_distributor_queue_length`. New metric has two custom labels: `name` and  `tenant`.
+  - Deprecated `tempo_distributor_forwarder_pushes_total` metric in favor of `tempo_distributor_queue_pushes_total`.
+  - Deprecated `tempo_distributor_forwarder_pushes_failures_total` metric in favor of `tempo_distributor_queue_pushes_failures_total`.
 * [ENHANCEMENT] Filter namespace by cluster in tempo dashboards variables [#1771](https://github.com/grafana/tempo/pull/1771) (@electron0zero)
 * [ENHANCEMENT] Exit early from sharded search requests [#1742](https://github.com/grafana/tempo/pull/1742) (@electron0zero)
 * [CHANGE] Identify bloom that could not be retrieved from backend block [#1737](https://github.com/grafana/tempo/pull/1737) (@AlexDHoffer)


modules/distributor/forwarder.go
```diff
@@ -165,20 +190,38 @@ func (f *forwarder) watchOverrides() {
...
 			}()
 
 			// Synchronously update queue managers
-			for _, qm := range queueManagersToAdd {
-				level.Info(log.Logger).Log("msg", "Updating queue manager", "tenant", qm.tenantID)
-				f.queueManagers[qm.tenantID] = newQueueManager(qm.tenantID, qm.queueSize, qm.workerCount, f.forwardFunc)
+			for _, q := range queuesToAdd {
+				_ = level.Info(f.logger).Log("msg", "Updating queue manager", "tenant", q.tenantID)
+
+				processFunc := func(ctx context.Context, data *request) error {
```
joe-elliott, line 207: this function exists twice. instead of declaring and using it locally maybe just create a normal function they both reference?

honestly this whole block (lines 207 -> 224) is the same as above. maybe consolidate the whole thing into a function to improve readability?

modules/distributor/forwarder.go
```diff
@@ -165,20 +190,38 @@ func (f *forwarder) watchOverrides() {
...
+						Name:        "metrics-generator",
+						TenantID:    q.tenantID,
+						Size:        q.queueSize,
+						WorkerCount: q.workerCount,
+					},
+					f.logger,
+					f.reg,
+					processFunc,
+				)
+				newQueue.StartWorkers()
```
joe-elliott, line 222: the old queue implementation called `StartWorkers` on creation. no strong feelings here, but wondering why you moved it out?

modules/distributor/forwarder.go
```diff
@@ -189,156 +232,24 @@ func (f *forwarder) watchOverrides() {
...
 
 	return nil
 
 }
 
-func (f *forwarder) stop(_ error) error {
+func (f *generatorForwarder) stop(_ error) error {
 	close(f.shutdown)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
```
joe-elliott, line 245: does the services.Manager already have a timeout? I'm honestly not super familiar with that code. if it does i'd rather just rely on it to cancel context.

modules/distributor/queue/queue.go
```diff
@@ -0,0 +1,224 @@
...
+	// wg is used to wait for the workers to drain the queue while stopping
+	wg sync.WaitGroup
+
+	logger log.Logger
+
+	name           string
+	tenantID       string
+	workerCount    int
+	size           int
+	reqChan        chan *request[T]
```
joe-elliott, line 34: any reason to not make this `chan T` ?

modules/distributor/queue/queue.go
```diff
@@ -0,0 +1,224 @@
...
+
+	pushesTotalMetrics         *prometheus.CounterVec
+	pushesFailuresTotalMetrics *prometheus.CounterVec
+	lengthMetric               *prometheus.GaugeVec
+
+	readOnly *atomic.Bool
+}
+
+func New[T any](cfg Config, logger log.Logger, reg prometheus.Registerer, fn ProcessFunc[T]) *Queue[T] {
+	pushesTotalMetrics := prometheus.NewCounterVec(prometheus.CounterOpts{
```
joe-elliott, line 46: we already have a lot of metrics here. i think we either need to keep these and remove the external metrics that are basically the same thing or expect individual users of this queue to handle their own metrics.

if we do remove the metrics in forwader.go in favor of these we need to list a breaking change in the changelog.

modules/distributor/queue/queue.go
```diff
@@ -0,0 +1,224 @@
...
+	lengthMetric := prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "tempo",
+		Subsystem: "distributor",
+		Name:      "queue_length",
+		Help:      "Number of queued requests for a tenant",
+	}, []string{"name", "tenant"})
+
+	var alreadyRegisteredErr prometheus.AlreadyRegisteredError
+	err := reg.Register(pushesTotalMetrics)
+	if err != nil && !errors.As(err, &alreadyRegisteredErr) {
```
joe-elliott, line 67: does this work? if we fail to register this metric a second time due to it being already registered then we lose a reference to the original metric that is registered and these metrics stop updating? this will happen the first time the config is reloaded?

just define them with promauto?

modules/distributor/queue/queue.go
```diff
@@ -0,0 +1,224 @@
...
+		_ = level.Warn(logger).Log("msg", "failed to register queue_pushes_failures_total metric", "err", err)
+	}
+
+	err = reg.Register(lengthMetric)
+	if err != nil && !errors.As(err, &alreadyRegisteredErr) {
+		_ = level.Warn(logger).Log("msg", "failed to register queue_length metric", "err", err)
+	}
+
+	return &Queue[T]{
```
joe-elliott, line 80: we lost `workerAliveCount` from the previous implementation. it's not clear to me what the purpose of that var was. @kvrhdn do you see any issues with removing it?

modules/distributor/queue/queue.go
```diff
@@ -0,0 +1,224 @@
...
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/weaveworks/common/user"
+	"go.uber.org/atomic"
```
joe-elliott, line 14: would prefer moving to the std lib implementation unless there's a reason to use this one.

modules/distributor/queue/queue.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	m.pushesTotalMetrics.WithLabelValues(m.name, m.tenantID).Inc()
+
+	req := &request[T]{
+		data: data,
+	}
+
+	select {
+	case <-ctx.Done():
```
joe-elliott, line 110: why did we split this select? if the chan is blocked and context is cancelled this will no longer exit immediately.

modules/distributor/queue/queue.go
```diff
@@ -0,0 +1,224 @@
...
+func (m *Queue[T]) StartWorkers() {
+	for i := 0; i < m.workerCount; i++ {
+		m.wg.Add(1)
+
+		go m.worker()
+	}
+}
+
+// Name returns queue name.
+func (m *Queue[T]) Name() string {
```
joe-elliott, line 138: nit: if these are used then keep them, otherwise delete them.

no reason to have accessors that are unused.

modules/distributor/queue/queue.go
```diff
@@ -0,0 +1,224 @@
...
+	for {
+		select {
+		case req := <-m.reqChan:
+			m.lengthMetric.WithLabelValues(m.name, m.tenantID).Dec()
+			m.forwardRequest(req)
+		case <-m.workersCloseCh:
+			// Forces to always trying to pull from the queue before exiting
+			// This is important during shutdown to ensure that the queue is drained
+			select {
+			case req, ok := <-m.reqChan:
```
joe-elliott, line 183: is this equivalent to the previous logic? does anything close m.reqChan?

modules/distributor/queue/queue.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+func (m *Queue[T]) forwardRequest(req *request[T]) {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	ctx = user.InjectOrgID(ctx, m.tenantID)
+
+	if err := m.fn(ctx, req.data); err != nil {
+		_ = level.Error(m.logger).Log("msg", "pushing with forwarder failed", "err", err)
```
joe-elliott, line 204: nit: i'd prefer that whatever passes in the forward function handles its own errors. i don't see a ton of value in returning an error and just logging it.

joe-elliott (commented):
Appreciate splitting this PR! Some thoughts/changes, but overall this is looking good!

modules/distributor/queue/queue.go
Blinkuu: I think the only easy way to fix this is to define those metrics on the package level to make initialize them only once. It's a bummer, but I see no other way around this without an abstraction around Prometheus. Therefore, I moved them outside of this function and they are now "global".

modules/distributor/queue/queue.go
mapno: Workers can panic and that wouldn't get propagated since they're running in goroutines, so we could end up in a situation where workers are exhausted but they're not marked as such. The idea was to detect this and spawn new workers or reset the queue.

Blinkuu: Hmm, interesting. Interesting solution, however, I think that there were a couple of concurrency issues with it (hence the TODO there). Maybe we could create a separate issue for this one and leave it as is for now? I am also fine with bringing back the `workerAliveCount`, so if you prefer that, let me know.

modules/distributor/queue/queue.go
joe-elliott: yup. agree with this change. i misunderstood what was happening here. thx for the clear explanation!

modules/distributor/queue/queue.go
joe-elliott: understood, thank you!

modules/distributor/queue/queue.go
mapno: I'm fine with removing it too 👍

modules/distributor/forwarder.go
```diff
@@ -77,25 +78,25 @@ func newForwarder(fn forwardFunc, o *overrides.Overrides) *forwarder {
...
 	select {
 	case <-f.shutdown:
 		return
 	default:
 	}
 
-	qm := f.getOrCreateQueueManager(tenantID)
-	err := qm.pushToQueue(ctx, &request{keys: keys, traces: traces})
+	q := f.getOrCreateQueue(tenantID)
+	err := q.Push(ctx, &request{tenantID: tenantID, keys: keys, traces: traces})
```
mapno, line 89: Nit: I don't think the tenantID is really needed. The queue already knows its tenant.

modules/distributor/queue/queue.go
joe-elliott: I would prefer that we start using these metrics in the queue and remove the old forwarder metrics that are the exact same thing. However, this would require changes to our alerts and runbooks due to renaming these two metrics.

In order to move this forward let's do this:
- Keep both sets of metrics.
- Mark the old ones deprecated 
  - add a comment in code that they are deprecated
  - add it to the prom metric descrption
  - call out the deprecation in the changelog and indicate the new metric.
- In a future change we can remove the old ones and update all of our related operational code.

joe-elliott (commented):
Getting close! There were some subtle changes to the way channels were handled in here and I appreciate the patient explanations on the new behavior. Thanks :pray:

mapno (commented):
LGTM. Will leave for other to comment.

modules/distributor/forwarder.go
Blinkuu: Yeah, but perhaps not every queue user is interested in having the `tenantID` injected. I'd just keep it abstract. Also, if I were to remove it now, I'd have to either refactor the queue.ProcessFunc so that it accepts `tenantID` as a parameter or inject & extract it from context, which will only bloat the code IMO.

modules/distributor/queue/queue.go
Blinkuu: Done in [f995cac](https://github.com/grafana/tempo/pull/1796/commits/f995cace7f4a12e4b34f81f25e17dec60d456508).

modules/distributor/queue/queue.go
kvrhdn: We used `workerAliveCount` to detect when a worker go routine panics and crashes. `shouldUpdate` will check `workerAliveCount` with the desired count and recreate the workers if one or more is missing.
We had to keep a separate counter because `WaitGroup` doesn't say how many workers are active.

So the risk here is that multiple workers can crash, we won't detect this and the queue will struggle to keep up.

That said, I'm not sure if we ever saw this happen... I'm fine with dropping the code, if we end up having performance issues we should see the panics in the logs and can add extra safeguards.

kvrhdn (approved):
LGTM! Nice work 🙂

PR #1802, opened 2022-10-14
Cancel running workflows on new pushes.

pstibrany:
**What this PR does**:
This PR configures Github actions to cancel running workflows if new push is done on a branch or PR. Previous workflow has little value at that point. (Copied from Mimir)

**Checklist**
- [na] Tests updated
- [na] Documentation added
- [na] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,14 @@ on:
 #      - 'docs/**'
 #      - 'example/**'
 #      - 'tools/**'
+
+concurrency:
+  # Cancel any running workflow for the same branch when new commits are pushed.
+  # We group both by ref_name (available when CI is triggered by a push to a branch/tag)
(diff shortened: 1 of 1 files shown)


mdisibio (approved):
Hi, thanks for this change.  Linking the matching [Mimir PR](https://github.com/grafana/mimir/pull/627).  I will say it's a little unexpected for the CI workflow to cancel other unrelated workflows for the same branch, but it makes sense and doesn't seem to cause issues as Mimir was changed quite a while ago.  (There does exist a [{{ github.workflow }}](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow) variable if we want CI to only cancel CI in the future)

PR #1806, opened 2022-10-17
Traceql more functionality

mdisibio:
**What this PR does**:
This PR adds more TraceQL functionality at the storage layer for parquet format.

1. Re-add duration filtering with original implementation as End-Start (it was removed in #1724). This is not optimal and we still want to add a dedicated column, but introducing new columns still has some things to research, and better to have the traceql functionality.
2. Add remaining operators (!=, >=, <=, etc)
3. Add support for `status` intrinsic

Fixes n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/parquetquery/predicates.go
+++ b/pkg/parquetquery/predicates.go
@@ -261,6 +261,92 @@ func (p *IntBetweenPredicate) KeepPage(page pq.Page) bool {
 	return true
 }
 
+// Generic predicate with callbacks to evalulate data of type T
+// Fn evalulates a single data point and is required. Optionally,
+// a RangeFn can evalulate a min/max range and is used to
+// skip column chunks and pages when RangeFn is supplied and
+// the column chunk or page also include bounds metadata.
+type GenericPredicate[T any] struct {
+	fn      func(T) bool
+	rangeFn func(min, max T) bool
+	extract func(pq.Value) T
+}
+
+func NewGenericPredicate[T any](fn func(T) bool, rangeFn func(T, T) bool, extract func(pq.Value) T) *GenericPredicate[T] {
+	return &GenericPredicate[T]{fn, rangeFn, extract}
+}
+
+func (p *GenericPredicate[T]) KeepColumnChunk(c pq.ColumnChunk) bool {
+	if p.rangeFn == nil {
+		return true
+	}
+
+	if ci := c.ColumnIndex(); ci != nil {
+		for i := 0; i < ci.NumPages(); i++ {
+			min := p.extract(ci.MinValue(i))
+			max := p.extract(ci.MaxValue(i))
+			if p.rangeFn(min, max) {
+				return true
+			}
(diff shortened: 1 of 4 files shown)


tempodb/encoding/vparquet/block_traceql.go
```diff
@@ -575,29 +593,41 @@ func createStringPredicate(op traceql.Operator, operands traceql.Operands) (parq
...
-	vals := make([]string, 0, len(operands))
-
 	for _, op := range operands {
 		if op.Type != traceql.TypeString {
 			return nil, fmt.Errorf("operand is not string: %+v", op)
 		}
-		vals = append(vals, op.S)
 	}
 
+	s := operands[0].S
```
mapno, line 602: Logic here and above is a bit brittle. Why validate operands if only one is taken? Shouldn't the length be checked here not to panic?

mdisibio: Agree, these parts are a little unclear.  All operations have a slice of operands, but each operation expects a different number.  The number of operands is checked for correctness earlier in this file in [checkConditions](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet/block_traceql.go#L108)  In the following block, all operations take the one argument, and it is guaranteed to exist since the count was checked already.  OpNone exits early at the top.

pkg/parquetquery/predicates.go
```diff
@@ -261,6 +261,94 @@ func (p *IntBetweenPredicate) KeepPage(page pq.Page) bool {
...
+// the column chunk or page also include bounds metadata.
+type GenericPredicate[T any] struct {
+	Fn      func(T) bool
+	RangeFn func(min, max T) bool
+	Extract func(pq.Value) T
+}
+
+var _ Predicate = (*GenericPredicate[int64])(nil)
+
+func NewGenericPredicate[T any](fn func(T) bool, rangeFn func(T, T) bool, extract func(pq.Value) T) *GenericPredicate[T] {
```
joe-elliott, line 277: nit: Should extract return a bool `ok`? That way the calling code can signal that the `pq.Value` cannot be translated to whatever type is being attempted. 

Depending on how this is intended to be used might not matter at all.

joe-elliott (approved):
things look basically fine. one question that might be meaningless.

approving to unblock. gogo traceql

pkg/parquetquery/predicates.go
mdisibio: I think an `ok` is not needed because all values in a column have the same type.  The case where some values convert but others can't seems more like T isn't defined appropriately to the column.   The intent isn't to support any T, ideally only the same data types as parquet (int64, etc), but wasn't sure the best way to implement that constraint in go generics (is it possible?).  Further down this line of thinking, we could eliminate `Extract` altogether and automatically get the right data type based on T.

PR #1808, opened 2022-10-18
traceql: initial engine implementation

kvrhdn:
**What this PR does**:
This adds an initial implementation of the TraceQL engine. It connects the API handler with the storage layer. It's main tasks are:
- parse incoming requests and extract flattened conditions the storage layer can work with
- pull spansets from the storage layer and revalidate the query matches each span
- return the search response

It's capabilities are limited for now:
- we can only work on queries with a single span set filter (i.e. a single pair of `{ }`)
- arithmetics are not implemented yet

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/traceql/ast_execute.go
```diff
@@ -0,0 +1,117 @@
...
+		return NewStaticBool(lhs.asFloat() < rhs.asFloat()), nil
+	case OpLessEqual:
+		return NewStaticBool(lhs.asFloat() <= rhs.asFloat()), nil
+	case OpPower:
+	case OpEqual:
+		return NewStaticBool(lhs.Equals(rhs)), nil
+	case OpNotEqual:
+		return NewStaticBool(!lhs.Equals(rhs)), nil
+	case OpRegex:
+		matched, err := regexp.MatchString(rhs.S, lhs.S)
```
mdisibio, line 59: Should this return false when both arguments aren't strings? There is a failing test that covers this, `3 =~ 5` which returns true because both `.S` are empty.  

Same applies to `OpNotRegex`.

pkg/traceql/ast.go
```diff
@@ -303,6 +306,23 @@ func (s Static) impliedType() StaticType {
 	return s.Type
 }
 
+func (s Static) Equals(other Static) bool {
+	return s.Type == other.Type && s.N == other.N && s.F == other.F && s.S == other.S && s.B == other.B && s.D == other.D && s.Status == other.Status
```
mdisibio, line 310: This could be `s == other` unless I am missing something.  However there is likely special logic needed here for span `status` that came up in #1806.  We have `TypeStatus` to represent `unset, ok, error` special values, but it can also be addressed as an int `status = 42`, which impacts the equality logic here.

pkg/traceql/ast_execute.go
kvrhdn: Yeah, it should. I removed some code right before pushing that validates `lhs` and `rhs` have the same type. It's the same code as `validate()` but filling in the the types we got from the storage layer. Once I get those back in the test should be okay again.

pkg/traceql/ast.go
kvrhdn: So If I'm understanding you correctly, it should also be valid to compare a status with an int? So we can do a simple `s == other` unless type is `TypeStatus` in which we also check if `s.Status = other.N`.

pkg/tempopb/tempo.proto
```diff
@@ -80,6 +81,19 @@ message TraceSearchMetadata {
...
+
+message SpanSet {
+  repeated Span spans = 1;
+  uint32 matched = 2;
+}
+
+message Span {
+  string spanID = 1;
+  uint64 startTimeUnixNano = 2;
+  uint32 durationMs = 3;
```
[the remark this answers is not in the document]
kvrhdn: Sure agreed, just wanted to be consistent with the `TraceSearchMetadata` but I think we kind of regret using ms there already.

pkg/traceql/ast_execute.go
kvrhdn: Latest commit has extra logic to validate the types after they have been resolved with span data. This is the same logic as `validate()`, just using data from the storage layer.

pkg/traceql/ast_execute.go
```diff
@@ -0,0 +1,117 @@
...
+	panic("UnaryOperation has Op different from Not and Sub")
+}
+
+func (s Static) execute(span Span) (Static, error) {
+	return s, nil
+}
+
+func (a Attribute) execute(span Span) (Static, error) {
+	// TODO improve this code
+	for attribute, static := range span.Attributes {
```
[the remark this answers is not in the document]
kvrhdn: Alright, I've implemented the logic to take the scope into account. It's easier than I expected 😅

PR #1810, opened 2022-10-18
Delete TempoRequestErrors alert from mixin

zalegrala:
**What this PR does**:

Here I propose delete the undocumented `TempoRequestErrors` alert from the `tempo-mixin`.  This alert has limited utility in a generic sense due to the thresholds here. 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/operations/tempo-mixin/runbook.md
+++ b/operations/tempo-mixin/runbook.md
@@ -1,8 +1,7 @@
 # Runbook
 
-This document should help with remediating operational issues in Tempo.
+This document should help with remediation of operational issues in Tempo.
 
-## TempoRequestErrors
 ## TempoRequestLatency
 
 Aside from obvious errors in the logs the only real lever you can pull here is scaling.  Use the Reads or Writes dashboard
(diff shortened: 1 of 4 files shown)


zalegrala:
Does this warrant a breaking change note in the CHANGELOG?

operations/tempo-mixin/runbook.md
```diff
@@ -1,8 +1,7 @@
 # Runbook
 
-This document should help with remediating operational issues in Tempo.
+This document should help with remediation of operational issues in Tempo.
 
-## TempoRequestErrors
```
joe-elliott, line 5: i think this might be linked by internal docs in deployment tools, please double check

joe-elliott (commented):
torn on this. i think it's a simple alert that others can build off of, but it's almost too generic to be useful.

i'm fine with this PR as is. alternatively we could make the alert threshold configurable and just not create the alert if set to 0? either way is fine, up to the team.

the number of people who use are jsonnet directly is likely small, but they exist and it does need to be listed as breaking change if we remove.

operations/tempo-mixin/runbook.md
zalegrala: No mentions outside of the mixin that I see.

joe-elliott (approved):

PR #1812, opened 2022-10-19
Set omitempty on yaml serialization config

rubenvp8510:
Signed-off-by: Ruben Vargas <ruben.vp8510@gmail.com>

**What this PR does**:

This PR sets `omitempty` on the config fields of different components. This allow us to serialized the config structures into a YAML  and omit fields that has empty values (empty string, 0s or nil), In this way we can use the structures to generate a valid config YAML with only the fields we want to specify, and use default config values for the fields that are not appear in the serialized configuration.

If we don't specify `omitempty` the config structure will generate a YAML with all the fields with empty strings, 0s etc.., this will override the default values for the different fields.

The concrete use case for this is to use the same tempo config structures to generate the configuration for the tempo operator.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo-serverless/handler.go
+++ b/cmd/tempo-serverless/handler.go
@@ -154,7 +154,7 @@ func loadBackend() (backend.Reader, *tempodb.Config, error) {
 
 func loadConfig() (*tempodb.Config, error) {
 	defaultConfig := &tempodb.Config{
-		Search: &tempodb.SearchConfig{
+		Search: tempodb.SearchConfig{
 			ChunkSizeBytes:     tempodb.DefaultSearchChunkSizeBytes,
 			PrefetchTraceCount: tempodb.DefaultPrefetchTraceCount,
 		},
@@ -176,8 +176,14 @@ func loadConfig() (*tempodb.Config, error) {
 	}
 
 	v.AutomaticEnv()
+
 	v.SetEnvPrefix(envConfigPrefix)
 	v.SetEnvKeyReplacer(strings.NewReplacer(".", "_", "-", "_"))
+	err = bindEnvs(v, defaultConfig)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to bind env vars from config")
+
+	}
 
 	cfg := &tempodb.Config{}
 	err = v.Unmarshal(cfg, setTagName, setDecodeHooks)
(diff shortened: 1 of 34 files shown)


joe-elliott (commented):
So, I'm generally ok with this change if it lets you move forward with the operator. CI does seem to be mad at you so that needs to be cleaned up.

I am a bit concerned about swapping all of the value types to pointers. Mainly that we might introduce some edge case nil pointer panics.

rubenvp8510:
I'm still not sure if this is the right approach, the main advantage is the re-use of structures and it will be easier to maintain.

The concern about introducing pointers is valid I think.

There are also another couple of things that makes me feel no t100% comfortable doing this.

 - I see the config uses other external structures imported from other libraries, for those are not possible to do the modifications and add `omitempty`.
 -How to handle the upgrade process on the operator when the config changed if only one dependency version of tempo can be used,  (I'm not sure if in go we can support having dependencies for multiple versions of the same module)
 
Thoughts @pavolloffay @frzifus ?

frzifus:
> How to handle the upgrade process on the operator when the config changed if only one dependency version of tempo can be used, (I'm not sure if in go we can support having dependencies for multiple versions of the same module)

1. My first question would be how stable the configuration interface remains.
2. Next, do we need any information from the old configuration?
 
As far as I can see, the operator generates the configuration from the CR configuration. A new Tempo release would lead to a new operator version. Which would then generate the new configuration based on the CR. So no configuration migration is needed or? (in both directions upgrade/downgrade)

What do you think?

rubenvp8510:
> > How to handle the upgrade process on the operator when the config changed if only one dependency version of tempo can be used, (I'm not sure if in go we can support having dependencies for multiple versions of the same module)
> 
>     1. My first question would be how stable the configuration interface remains.
> 
>     2. Next, do we need any information from the old configuration?
> 
> 
> As far as I can see, the operator generates the configuration from the CR configuration. A new Tempo release would lead to a new operator version. Which would then generate the new configuration based on the CR. So no configuration migration is needed or? (in both directions upgrade/downgrade)
> 
> What do you think?

Yes you're right, good point, I think we can proceed here, I'm fixing the CI first :D

PR #1813, opened 2022-10-19
Fix failing SearchTagValues endpoint after startup

### LINKED ISSUE #1792: Ingesters not registering RPC endpoints on startup
**Describe the bug**
Ingesters seem to not be registering RPC endpoints until well after startup. Sometimes several minutes. In particular we have seen failures on `/tempopb.Querier/SearchTagValues` and `/tempopb.Querier/SearchTags` although likely other endpoints are not being registered as well.

Ingesters log:
image

Queriers log:
image

**To Reproduce**
1. Restart an ingester and use the Grafana UI to hit the named endpoints above.

stoewer:
**What this PR does**:
The bug described in #1792 happened because `rediscoverLocalBlocks` created `BackendSearchBlock` objects for WAL blocks, even when they didn't contain the files required for flatbuffer search (`search-header` etc.). Later calls of `TagValues()` on the search blocks resulted in the logged error message and caused the `/tempopb.Querier/SearchTagValues` endpoint to fail.

This PR skips the creation of `BackendSearchBlock` for WAL blocks that do not contain the flatbuffer search files.

Fixes #1792

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/backend/local/local.go
+++ b/tempodb/backend/local/local.go
@@ -104,14 +104,17 @@ func (rw *Backend) CloseAppend(ctx context.Context, tracker backend.AppendTracke
 // List implements backend.Reader
 func (rw *Backend) List(ctx context.Context, keypath backend.KeyPath) ([]string, error) {
 	path := rw.rootPath(keypath)
-	dirEntries, err := os.ReadDir(path)
+	folders, err := os.ReadDir(path)
 	if err != nil {
 		return nil, err
 	}
 
-	objects := make([]string, 0, len(dirEntries))
-	for _, d := range dirEntries {
-		objects = append(objects, d.Name())
+	objects := make([]string, 0, len(folders))
+	for _, f := range folders {
+		if !f.IsDir() {
+			continue
+		}
+		objects = append(objects, f.Name())
 	}
 
 	return objects, nil
--- a/tempodb/backend/local/local.go
+++ b/tempodb/backend/local/local.go
@@ -104,17 +104,14 @@ func (rw *Backend) CloseAppend(ctx context.Context, tracker backend.AppendTracke
 // List implements backend.Reader
 func (rw *Backend) List(ctx context.Context, keypath backend.KeyPath) ([]string, error) {
 	path := rw.rootPath(keypath)
-	folders, err := os.ReadDir(path)
+	dirEntries, err := os.ReadDir(path)
 	if err != nil {
 		return nil, err
 	}
 
-	objects := make([]string, 0, len(folders))
-	for _, f := range folders {
-		if !f.IsDir() {
-			continue
-		}
-		objects = append(objects, f.Name())
+	objects := make([]string, 0, len(dirEntries))
+	for _, d := range dirEntries {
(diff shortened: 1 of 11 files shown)


tempodb/backend/local/local.go
```diff
@@ -104,17 +104,14 @@ func (rw *Backend) CloseAppend(ctx context.Context, tracker backend.AppendTracke
...
 
-	objects := make([]string, 0, len(folders))
-	for _, f := range folders {
-		if !f.IsDir() {
-			continue
-		}
-		objects = append(objects, f.Name())
+	objects := make([]string, 0, len(dirEntries))
+	for _, d := range dirEntries {
+		objects = append(objects, d.Name())
```
joe-elliott, line 114: This was changed to support the `Has()` method? I'm concerned about the impact on other uses of `List()`

tempodb/backend/raw.go
```diff
@@ -113,6 +113,19 @@ func NewReader(r RawReader) Reader {
 	}
 }
 
+func (r *reader) Has(ctx context.Context, name string, blockID uuid.UUID, tenantID string) (bool, error) {
+	objects, err := r.r.List(ctx, KeyPathForBlock(blockID, tenantID))
```
joe-elliott, line 117: No idea if this works for all backends :). We only really need it to work for local, but that's not ideal.

joe-elliott (commented):
I'm torn on this. This is a very correct way to fix this problem, but making changes to the backend interfaces for something I view as temporary is tough. In 2.0 we will drop support for search over the old block format ('v2') and all of this code goes away.

In `OpenBackendSearchBlock()` we could attempt to call `ReadRange()` for 1 byte on the header. If that returns "does not exist" then we return `ErrSearchNotSupported`. I know this is far "hackier" then then the proposed solution, but it will be nicely cleaned up when we delete the rest of this code.

mdisibio:
> making changes to the backend interfaces

I also lean towards not changing the backend interface.  Tempo is run on a many backends, both the popular clouds but also api-compatible vendor solutions that we have no way to test. Traditionally we have kept Tempo's usage as basic as possible to improve compatibility.  I don't necessarily see an issue with adding `Has` here since it is constrained to the ingester and local backend, but would have concerns if it had wider usage throughout Tempo. Therefore I would be ok with @joe-elliott 's alternate method (reading 1 byte).

tempodb/backend/local/local.go
stoewer: Yes. From the signature alone it is not clear whether `List` is supposed to return only directory names. The code where `List` is used ([raw.go#L158](https://github.com/grafana/tempo/blob/main/tempodb/backend/raw.go#L158) or [raw.go#L139](https://github.com/grafana/tempo/blob/main/tempodb/backend/raw.go#L139)) seems to anticipate that filenames are listed too. However, the s3 backend also only returns directories / prefixes.

stoewer:
Thanks for the review. I removed `List` method and implemented the check using `ReadRange` as suggested by @joe-elliott.

PR #1821, opened 2022-10-21
[TraceQL] Implement Pipeline.evaluate

kvrhdn:
**What this PR does**:
Implement `Pipeline.evaluate`.

~~Fixes #<issue number>~~

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/ast_test.go
+++ b/pkg/traceql/ast_test.go
@@ -50,6 +50,55 @@ func TestStatic_Equals(t *testing.T) {
 	}
 }
 
+func TestPipelineEvaluate(t *testing.T) {
+	testCases := []struct {
+		query  string
+		input  []Spanset
+		output []Spanset
+	}{
+		{
+			"{ true } | { true } | { true }",
+			[]Spanset{
+				{Spans: []Span{{}}},
+			},
+			[]Spanset{
+				{Spans: []Span{{}}},
+			},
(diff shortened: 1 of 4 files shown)


mdisibio (approved):

pkg/traceql/ast_test.go
```diff
@@ -50,6 +50,55 @@ func TestStatic_Equals(t *testing.T) {
...
+		},
+		{
+			"{ true } | { false } | { true }",
+			[]Spanset{
+				{Spans: []Span{{}}},
+			},
+			[]Spanset{},
+		},
+		{
+			"{ .foo1 = `a` } | { .foo2 = `b` }",
```
kvrhdn, line 76: This test fails right now with
```
binary operations must operate on the same type: .foo2 = `b`
```
This happens because `.foo2` doesn't exist and has type `Nil`. I think we should handle errors differently in `SpansetFilter.evaluate` and not interrupt the for loop.

@mdisibio, thoughts?

kvrhdn: Addressed by https://github.com/grafana/tempo/pull/1821/commits/1e7fa462b90444b7030d7f8b1bb61b7e70cd9d7f

mdisibio (approved):

PR #1823, opened 2022-10-21
[TraceQL] Extract storage layer conditions from all pipeline elements

kvrhdn:
**What this PR does**:
Changes:
- extract conditions from all pipeline elements
- use pipeline.evaluate to process the span set once retrieved

~~Fixes #<issue number>~~

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/traceql/engine.go
```diff
@@ -61,59 +58,46 @@ func (e *Engine) Execute(ctx context.Context, searchReq *tempopb.SearchRequest,
 			break
 		}
 
-		span.LogKV("msg", "iterator.Next", "rootSpanName", spanSet.RootSpanName, "rootServiceName", spanSet.RootServiceName, "spans", len(spanSet.Spans))
-
-		spanSet = e.validateSpanSet(spanSetFilter, spanSet)
-		if spanSet == nil {
+		ss, err := rootExpr.Pipeline.evaluate([]Spanset{*spanSet})
+		if err != nil {
+			span.LogKV("msg", "pipeline.evaluate", "err", err)
```
mdisibio, line 63: Wondering if it makes sense to fail the query on engine instead, to make failures more apparent.   In this log message let's go ahead and include some other help info like the pipeline itself (stringer will reprint the query).

pkg/traceql/engine.go
```diff
@@ -192,52 +149,52 @@ func (e *Engine) asTraceSearchMetadata(spanset *Spanset) (*tempopb.TraceSearchMe
...
 		return &common_v1.AnyValue{
 			Value: &common_v1.AnyValue_StringValue{
-				StringValue: static.Status.String(),
+				StringValue: s.Status.String(),
 			},
-		}, nil
+		}
 	default:
-		return nil, fmt.Errorf("static has unexpected type %v", static.Type)
+		panic(fmt.Errorf("static has unexpected type %v", s.Type))
```
mdisibio, line 198: Do we need to add TypeNil?

mdisibio (approved):
LGTM, left a few non-blocking comments.

pkg/traceql/engine.go
kvrhdn: What do you mean with fail the query on engine instead? The query will fail here if data from the storage layer doesn't agree with the query (e.g. attribute has the wrong type), we can't handle these earlier.
I do agree we should improve the validation functions a bit and block some queries in the query-frontend.

pkg/traceql/engine.go
kvrhdn: I'm not entirely sure since I don't think we can return an attribute with value Nil. I've added it just in case, this function should never panic now.

pkg/traceql/engine.go
mdisibio: I mean to exit and return the error instead of logging and continuing.   Can you clarify a little what you mean the case where an attribute type doesn't agree with the query?  TraceQL is designed with no type casting, and only matches when the types are compatible:  i.e. ` { http.status_code >= 500 }` should simply ignore non-numeric values for `http.status_code` (BinaryOperation return false).   Would that generate an error?    The benefit of returning the error immediately is that it will be clearer to identify undefined or improperly defined engine logic.

kvrhdn: I was confusing this with the type check we do in the validation step. During validation `{ "foo"  = 5 }` will return an error, but when evaluating a span mismatching types will just return `false`. I'll change the logic to fail fast here: if evaluation errors, just return that to the caller. This way will be able to see issues faster.

PR #1830, opened 2022-10-24
[TraceQL] Support for basic aggregates count/avg and scalar filters

mdisibio:
**What this PR does**:
This PR adds support for TraceQL queries that use basic aggregate + scalar filter. This overlaps with and should be merged after #1823.   Count and average are working, and scalar filter is only supported if LHS is an aggregate and RHS is a static.

Some examples:
* `{ .foo = "a" } | count() > 10`
* `{ .foo = "a" } | avg(duration) > 10ms`

Fixes n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added


kvrhdn (approved):
Nice!

Do you think we can sharpen the `validate()` functions to only allow scalar filters that are simple enough? (Doesn't have to be in this PR)

PR #1833, opened 2022-10-25
[TraceQL] Add Name field to response span set; filter out duration from attributes

yvrhdn:
**What this PR does**:
Changes:
- adds a field `name` to each returned span
- filters out `name` and `duration` from attributes since they are set on the spans object

~~Fixes #<issue number>~~

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/engine.go
+++ b/pkg/traceql/engine.go
@@ -176,7 +176,7 @@ func (e *Engine) asTraceSearchMetadata(spanset *Spanset) (*tempopb.TraceSearchMe
 		}
 
 		for attribute, static := range span.Attributes {
-			if attribute.Name == IntrinsicName.String() || attribute.Name == IntrinsicDuration.String() {
+			if attribute.Intrinsic == IntrinsicName || attribute.Intrinsic == IntrinsicDuration {
 				continue
 			}
 


pkg/tempopb/tempo.proto
```diff
@@ -91,9 +91,10 @@ message SpanSet {
 
 message Span {
   string spanID = 1;
-  uint64 startTimeUnixNano = 2;
-  uint64 durationNanos = 3;
-  repeated tempopb.common.v1.KeyValue attributes = 4;
+  string name = 2;
```
kvrhdn, line 94: I'm aware this is a breaking change, but AFAIK no one is consuming this API yet as protobuf. This is not a breaking change for JSON.

pkg/traceql/engine.go
```diff
@@ -171,7 +171,15 @@ func (e *Engine) asTraceSearchMetadata(spanset *Spanset) (*tempopb.TraceSearchMe
 			Attributes:        nil,
 		}
 
+		if name, ok := span.Attributes[NewIntrinsic(IntrinsicName)]; ok {
+			tempopbSpan.Name = name.S
+		}
```
kvrhdn, line 176: Currently, the name attribute will only be fetched if the query needs it. Should we always fetch it?

pkg/traceql/engine.go
```diff
@@ -171,7 +171,15 @@ func (e *Engine) asTraceSearchMetadata(spanset *Spanset) (*tempopb.TraceSearchMe
...
+
 		for attribute, static := range span.Attributes {
+			if attribute.Name == IntrinsicName.String() || attribute.Name == IntrinsicDuration.String() {
```
mdisibio, line 179: This could be `attribute.Intrinsic == ...`

mdisibio (approved):

pkg/traceql/engine.go
mdisibio: I think this is ok for now.

pkg/traceql/engine.go
kvrhdn: Yeah, good catch!

mdisibio (approved):

Issue #1392, opened 2022-04-18
Query limits don't work or work unexpectedly 


symmetriac:
**Describe the bug**
Query limits don't work or work unexpectedly  

**To Reproduce**
Make a query in grafana with tempo datasource with limit parameter, for example 1000. Looking at query inspector in response object there is an array with more than 1000 elements.
Also max_result_limit not respected. 

query frontend settings:
```
query_frontend:
  max_outstanding_per_tenant: 2000
  search:
    default_result_limit: 100
    max_result_limit: 1500
    concurrent_jobs: 300
    max_duration: 168h
```

Tested on tempo-querier traces with the same time range of 2 days:

with limit set to 1000 (less than max_result_limit)

<img width="1149" alt="Снимок экрана 2022-04-18 в 13 12 17" src="https://user-images.githubusercontent.com/46437352/163793978-8a99307a-3a4e-49c0-b1a8-8aa2c4e00e27.png">

with limit greater than max_result_limit

<img width="1149" alt="Снимок экрана 2022-04-18 в 13 16 28" src="https://user-images.githubusercontent.com/46437352/163794394-8c5f4bbf-6735-4e45-9744-629c50ae8357.png">

Same issue when querying via search api.
Is this behavior expected or am I missing some settings for this limits?

**Expected behavior**
When limit is set greater than max_result_limit response objects count is equal to max_result_limit.

**Environment:**
 - Infrastructure: Kubernetes, Grafana (v8.4.4)
 - Deployment tool: helm

**Additional Context**

joe-elliott:
As the query frontend is aggregating results from the queriers it collects them here:

https://github.com/grafana/tempo/blob/da37526479bf08c917a801a08c2fcf8ede3ee4ba/modules/frontend/searchsharding.go#L74

Each querier, when it's performing a subrequest, should respect the limit parameter of 1000 but when we re-aggregate I purposefully ignored this limit. Basically if we have _more_ than the limit we cancel all current jobs and return whatever we happen have:

https://github.com/grafana/tempo/blob/da37526479bf08c917a801a08c2fcf8ede3ee4ba/modules/frontend/searchsharding.go#L101

Due to random timings in how the subrequests return this can exceed the limit specified in the url.

[...]

PR #1839, opened 2022-10-26
searchsharding: ensure limit is never exceeded

kvrhdn:
**What this PR does**:
Ensure the search response from Tempo always contain at most `limit` traces. We were missing a check when appending results to the response object, this way additional traces could sneak into the response object.

Example: you already have 5 results, limit is 6. If a subquery returns 2 more results we would end up with 7 results. By checking the limit every time a trace is appended this won't happen anymore.

~~Fixes #<issue number>~~

**Checklist**
- [x] Tests updated
- [ ] Documentation added


joe-elliott:
This was originally written this way intentionally. If we've already done the work to pull additional traces why throw them away? However, we've gotten so many questions about this that it probably makes sense to go ahead with this change.

mdisibio (approved):

kvrhdn:
> This was originally written this way intentionally. If we've already done the work to pull additional traces why throw them away? However, we've gotten so many questions about this that it probably makes sense to go ahead with this change.

I see 😅 Yeah, I agree with that reasoning, we already fetched this data from the queriers so why not just return it. I created the PR because we got reports of traceQL queries returning more than the limit.

I guess it depends a bit on how we view `limit`: is this a value to protect the amount of data the caller receives or is this a value to protect the backend? We mostly see it as the latter, but I guess most APIs use limit in combination with pagination to limit response size. Hence the confusion.

joe-elliott:
Fixes #1392

PR #1850, opened 2022-10-31
Allow compactor disablement

zalegrala:
Without this change, there isn't a good way to disable compaction when running in single binary or scalable single binary mode.  Normally, when running in microservices mode, operators can simply scale down the compactors to 0 to effectively disable compaction, though this will have other consequences, since the compactors are doing a little more than compaction.

Here we include a config option to disable compaction at the compactor level.  This allows the compactor module to run, but disable compaction, allowing blocks to remain in the backend for debug or investigation purposes.

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/tempo/website/configuration/_index.md
+++ b/docs/tempo/website/configuration/_index.md
@@ -435,7 +435,6 @@ compactor:
     # Note: This should only be used in a non-prodution context for debugging purposes.  This will allow blocks to say in the backend for further investigation if desired.
     [disabled: <bool>]
 
-
     ring:
 
         kvstore:
--- a/docs/tempo/website/configuration/_index.md
+++ b/docs/tempo/website/configuration/_index.md
@@ -431,6 +431,11 @@ Compactors stream blocks from the storage backend, combine them and write them b
 ```yaml
 compactor:
 
+    # Optional. Disables backend compaction.  Default is false.
+    # Note: This should only be used in a non-prodution context for debugging purposes.  This will allow blocks to say in the backend for further investigation if desired.
+    [disabled: <bool>]
+
+
     ring:
 
         kvstore:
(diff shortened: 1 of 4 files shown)


mapno (commented):
This should be in the docs, with an explanation on when to use it. It's a very niche use case that will confuse more users than it'll be used by.

mapno (approved):

modules/compactor/config.go
```diff
@@ -37,6 +38,7 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
 	f.IntVar(&cfg.Compactor.MaxCompactionObjects, util.PrefixConfig(prefix, "compaction.max-objects-per-block"), 6000000, "Maximum number of traces in a compacted block.")
 	f.Uint64Var(&cfg.Compactor.MaxBlockBytes, util.PrefixConfig(prefix, "compaction.max-block-bytes"), 100*1024*1024*1024 /* 100GB */, "Maximum size of a compacted block.")
 	f.DurationVar(&cfg.Compactor.MaxCompactionRange, util.PrefixConfig(prefix, "compaction.compaction-window"), time.Hour, "Maximum time window across which to compact blocks.")
+	f.BoolVar(&cfg.Disabled, util.PrefixConfig(prefix, "compaction.disabled"), false, "Disable compaction.")
```
zalegrala, line 41: This should just be "disabled", not nested under `compaction` to match the config object.

docs/tempo/website/configuration/_index.md
```diff
@@ -431,6 +431,11 @@ Compactors stream blocks from the storage backend, combine them and write them b
 ```yaml
 compactor:
 
+    # Optional. Disables backend compaction.  Default is false.
+    # Note: This should only be used in a non-prodution context for debugging purposes.  This will allow blocks to say in the backend for further investigation if desired.
+    [disabled: <bool>]
+
```
zalegrala, line 437: Whitespace.

PR #1852, opened 2022-10-31
Avoid running pool jobs with a cancelled context

zalegrala:
Without this change, jobs will continue to get scheduled, even if the context has been cancelled.  Here we include a check and avoid doing work if we have been asked to stop.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


mapno (approved):
Changelog entry?

PR #1855, opened 2022-11-01
Update several dependencies

zalegrala:
Signed-off-by: Zach Leslie <zach.leslie@grafana.com>

**What this PR does**:

This PR updates several depenencies as part of an update for `weaveworks/common` to allow TLS specifications.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


go.mod
```diff
@@ -29,68 +29,68 @@ require (
...
 	github.com/segmentio/parquet-go v0.0.0-20221019213850-8b6e2b21a101
-	github.com/sirupsen/logrus v1.8.1
-	github.com/sony/gobreaker v0.4.1
+	github.com/sirupsen/logrus v1.9.0
+	github.com/sony/gobreaker v0.5.0
 	github.com/spf13/viper v1.12.0
-	github.com/stretchr/testify v1.8.0
-	github.com/thanos-io/thanos v0.24.0
+	github.com/stretchr/testify v1.8.1
+	github.com/thanos-io/thanos v0.29.0-rc.0
```
zalegrala, line 70: This is a big one.

zalegrala: How do we feel about an rc here?  We could wait, or take the rc and quickly update when 29.0 comes.

mdisibio: What is the motivation for upgrading to `0.29.0-rc.0` vs the latest stable which I assume is `0.28.0`?

zalegrala: There was a recent breaking change in one of the dependencies that a bunch of projects share.

https://github.com/thanos-io/thanos/blob/v0.28.1/go.mod#L19

This caused a bunch of headache for updates, which has been updated in 29.0-rc.0 to point to the new location.

zalegrala:
Let's hold on this.  I might be having local issues that prevented me from making a smaller update.

bboreham:
Note Tempo already has the TLS settings; the current depenency includes https://github.com/weaveworks/common/pull/256.
The only diff between that and latest `weaveworks/common` is in the wording of the help text.

zalegrala:
Let's close this one out.  I've raised https://github.com/grafana/tempo/pull/1857 as a replacement update.

PR #1871, opened 2022-11-08
Support global and wildcard overrides in Generic Forwarding

Blinkuu:
**What this PR does**:

This PR adds support for global and wildcard overrides. Before, generic forwarding could only be enabled on a per-tenant basis. To make implementation behave in a similar manner to the rest of the code, I have added support for both wildcard and global overrides. From now on, users are not required to explicitly define forwarders on a per-tenant basis, but can also do so using a wildcard character in per-tenant overrides config ("*") or even define forwarders globally in the `overrides` section of `tempo.yaml`.

Example:

Let's imagine we have overrides defined in three ways: globally, using wildcard and on a per-tenant basis.

```yaml
# tempo.yaml
overrides:
  forwarders: ["my-forwarder-1"]
  per_tenant_override_config: /overrides.yaml
```

```yaml
# overrides.yaml
overrides:
  "tenant-0":
    forwarders: [ "my-forwarder-2" ]
  "*":
    forwarders: [ "my-forwarder-3" ]
```

Given above configuration:

- `tenant-0` will forward to forwarder named `my-forwarder-2`
- Any other tenant will forward to forwarder named `my-forwarder-3`

If we were to remove the wildcard config from the `overrides.yaml`:

```yaml
# overrides.yaml
overrides:
  "tenant-0":
    forwarders: [ "my-forwarder-2" ]
```

- `tenant-0` will still forward to forwarder named `my-forwarder-2`
- Any other tenant will forward to forwarder named `my-forwarder-1`

This PR is a continuation of #1775.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/forwarder/manager.go
+++ b/modules/distributor/forwarder/manager.go
@@ -66,6 +66,10 @@ func NewManager(cfgs ConfigList, logger log.Logger, overrides Overrides) (*Manag
 }
 
 func (m *Manager) ForTenant(tenantID string) List {
+	if len(m.forwarderNameToForwarder) < 1 || len(m.overrides.Forwarders(tenantID)) < 1 {
+		return nil
+	}
+
 	ql, ok := m.getOrCreateQueueList(tenantID)
 	if !ok {
 		return nil
@@ -84,10 +88,6 @@ func (m *Manager) getOrCreateQueueList(tenantID string) (*queueList, bool) {
 	defer m.tenantToQueueListMu.Unlock()
 
 	forwarderNames := m.overrides.Forwarders(tenantID)
-	if len(forwarderNames) < 1 {
-		return nil, false
-	}
-
 	ql, err := newQueueList(m.logger, tenantID, forwarderNames, m.forwarderNameToForwarder)
 	if err != nil {
 		_ = level.Warn(m.logger).Log("msg", "failed to create queue list", "err", err)
--- a/modules/distributor/distributor.go
+++ b/modules/distributor/distributor.go
@@ -379,10 +379,8 @@ func (d *Distributor) PushTraces(ctx context.Context, traces ptrace.Traces) (*te
 		d.generatorForwarder.SendTraces(ctx, userID, keys, rebatchedTraces)
 	}
 
-	if len(d.cfg.Forwarders) > 0 && len(d.overrides.Forwarders(userID)) > 0 {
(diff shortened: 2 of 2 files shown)


modules/distributor/forwarder/manager.go
```diff
@@ -66,21 +66,49 @@ func NewManager(cfgs ConfigList, logger log.Logger, overrides Overrides) (*Manag
...
+func (m *Manager) getOrCreateQueueList(tenantID string) (*queueList, bool) {
+	ql, found := m.getQueueList(tenantID)
+	if found {
+		return ql, true
+	}
+
+	m.tenantToQueueListMu.Lock()
+	defer m.tenantToQueueListMu.Unlock()
+
+	forwarderNames := m.overrides.Forwarders(tenantID)
```
joe-elliott, line 86: it seems like for tenants w/o a configured forwarder this lock will hit repeatedly b/c it will not be found in the list. should we store a `nil` value in `m.tenantToQueueList` to prevent this?

joe-elliott: It seems like this is the core of the change? By repeatedly asking for the forwarder we can catch the wildcard scenarios? I'm fine with that idea, but I'd like to find a way to "cache" the nil to prevent constant lookups here.

Other than that looks good.

joe-elliott: I see now. I'll admit I don't like the logic to get the queue is split between this function and  `distributor.go:382`. It's difficult to see all the logic related to how we request a forwarder for a tenant. Likely it was put there because it mirrors the metric generator lines above it?

Personally, I think it would be cleaner to move this logic:
```
	if len(d.cfg.Forwarders) > 0 && len(d.overrides.Forwarders(userID)) > 0 {
```

into the manager b/c there would be one place to look to see exactly how a forwarder is requested for a tenant. This is not a blocker though. WDYT?

Blinkuu: Moved in [ce63eeb](https://github.com/grafana/tempo/pull/1871/commits/ce63eebee0fd54436bcaf1f16fa5c2d1c862f067).

joe-elliott (approved):
lgtm @mapno did you want to take a look?

clean up the changelog conflict and we'll merge

mapno (approved):
LGTM

Blinkuu:
@mapno I've rebased on top of origin/main and fixed CHANGELOG.md conflicts.

PR #1873, opened 2022-11-08
"Fix" row group size bytes

joe-elliott:
**What this PR does**:
Currently there are two methods that estimate trace size when building row groups in parquet blocks. Both of these methods do a decent job of estimating the uncompressed proto size of a trace, but are poor at estimating a compressed and encoded version in parquet. The result of this is that the actual row group created has been observed to be anywhere between 5->30% of the configured row_group_bytes.

This PR
- Adjusts the two methods to be lower bounds on row_group_size bytes. In practice on real data the created row groups are 100->150% of the configured size.
- Roughly aligns both methods around the concept of counting the total number of parquet values pushed with the assumption that across enough traces we approach 1 byte per value stored.
- Renames both methods for clarity.
- Sets the default to 100MB.
- Adds documentation for `row_group_size_bytes`.

### DIFF
--- a/tempodb/encoding/vparquet/compactor.go
+++ b/tempodb/encoding/vparquet/compactor.go
@@ -97,7 +97,7 @@ func (c *Compactor) Compact(ctx context.Context, l log.Logger, r backend.Reader,
 		if c.opts.MaxBytesPerTrace > 0 {
 			sum := 0
 			for _, row := range rows {
-				sum += esimateMarshalledSizeFromParquetRow(row)
+				sum += estimateProtoSizeFromParquetRow(row)
 			}
 			if sum > c.opts.MaxBytesPerTrace {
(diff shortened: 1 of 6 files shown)


tempodb/encoding/vparquet/compactor.go
```diff
@@ -328,10 +328,10 @@ func (r *rowPool) Put(row parquet.Row) {
 	r.pool.Put(row[:0]) //nolint:all //SA6002
 }
 
-// estimateProtoSize estimates the byte-length of the corresponding
+// estimateMarshalledSizeFromParquetRow estimates the byte-length of the corresponding
```
mdisibio, line 331: Typo in comment

mdisibio (approved):
Lgtm and think the accuracy is acceptable. Noticed one typo.

PR #1874, opened 2022-11-08
Add TLS support to vulture

zalegrala:
**What this PR does**:

Add a `tempo-tls` flag to enable TLS negotiation, but not verification, on the gRPC connection to Tempo.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


zalegrala:
Note that this is only the write path.  The query path should be simple, but I don't have a test up right now.  Perhaps we can take that as another PR later?

mapno (approved):

PR #1878, opened 2022-11-09
Parquet WAL

mdisibio:
**What this PR does**:
This PR introduces support for a parquet-formatted WAL with the goal to support TraceQL search on WALs (which is currently infeasible on v2).  The WAL format is controlled by a new `storage.Trace.WAL.Version` config which is similar to the one used for backend blocks `storage.Trace.Block.Version`.  Setting it to `vParquet` enables to the new WAL format, default is still `v2`.  All combinations of WAL and Block versions are supported, but performance is best when they match.

**Design**
Parquet WALs break from the existing design where the WAL is an open file to which new data is appended as it is received.  Now a parquet WAL is a folder with a series of sub-files, like this:

* path/wal/`<blockID>+<tenantID>+vParquet`
** meta.json
** 00001
** 00002
** 00003
** 00004

Each file is a standalone parquet file, created each time the ingester flushes idle traces from memory to disk (every few seconds).  From there the sub-files are identical in format to existing parquet blocks and all the same functionality is reused:  search, find trace by id, tag lookup, and soon: traceql.   There is also an operational benefit that the standalone files are easy to inspect and troubleshoot with existing parquet tooling.

**Refactor**
This PR includes a refactor of the `tempodb/wal` module to support different encodings and is similar to what was done for the backend.  A generic interface for a WAL block was introduced in `tempodb/encoding/common` and now both `v2` and `vParquet` encodings implement it.

**Performance Impact**
As expected there is an increase in resources for this format, ~2x cpu and ~2x memory.  I don't think it can ever be as efficient as `v2` which was basically copying bytes from memory to disk, but there are several areas we can explore to mitigate this - especially the case of parquet wal -> parquet block.

**TODOs**
This PR is stable and usable, and worth merging now since it is already quite large.  However there are several TODOs which make sense to do in follow-up PRs:
* Wire up the TraceQL read path to the ingesters (and these new WAL blocks)
* Test replay with partially-written/corrupted files
* Improve performance of `parquet wal -> parquet block` step. Maybe we can do the same as compaction and copy raw rows between blocks instead of `Trace` objects. (And use buffer pooling to further reduce memory overhead) Also change the multi-block iterator to not simultaneously buffer the next `Trace` from all inputs. (i.e. 100 wal sub-files means 100 traces resident in memory at the same time during completion process)

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/encoding/common/types.go
```diff
@@ -1,7 +1,44 @@
...
+	"hash/fnv"
+)
+
 // This file contains types that need to be referenced by both the ./encoding and ./encoding/vX packages.
 // It primarily exists here to break dependency loops.
 
 // ID in TempoDB
 type ID []byte
+
+// IDMap is a helper for recording and checking for IDs. Not safe for concurrent use.
```
joe-elliott, line 14: not seeing this used anywhere. remove?

tempodb/encoding/vparquet/block_traceql.go
```diff
@@ -181,6 +181,34 @@ func (i *spansetIterator) Next(ctx context.Context) (*traceql.Spanset, error) {
...
+// mergeSpansetIterator iterates through a slice of spansetIterators exhausting them
+// in order
+type mergeSpansetIterator struct {
+	iters []*spansetIterator
+	cur   int
+}
+
+var _ traceql.SpansetIterator = (*mergeSpansetIterator)(nil)
+
+// jpe test
```
joe-elliott, line 193: remove todo

tempodb/encoding/vparquet/interfaces.go
```diff
@@ -7,9 +7,10 @@ import (
 	"github.com/segmentio/parquet-go"
 )
 
-type Iterator interface {
-	Next(context.Context) (*Trace, error)
+type TraceIterator interface {
+	NextTrace(context.Context) (common.ID, *Trace, error)
 	Close()
+	// __vparquet() jpe add a hidden method to prevent external implementations?
```
joe-elliott, line 13: remove todo

tempodb/encoding/common/types.go
mdisibio: It's used in parquet wal block FindTraceByID, to only read from pages containing the trace. The map is maintained on append/flush and rebuilt on rescan.

joe-elliott: whoops. didn't expand that diff :P

mapno (approved):
LGTM 👍 

Left only a comment, everything else looks ok. But, I may not be the best to review this code, so take it with a grain of salt 😅

joe-elliott (approved):

tempodb/encoding/vparquet/multiblock_iterator.go
```diff
@@ -78,17 +79,64 @@ func (m *MultiBlockIterator) Next(ctx context.Context) (common.ID, parquet.Row,
...
+func (b *bookmark[T]) clear() {
+	b.currentID = nil
+	b.currentObject = nil
+}
+
+func (b *bookmark[T]) close() {
+	b.iter.Close()
+}
+
+type genericIterator[T iteratable] interface { // jpe make this name not terrible
```
joe-elliott, line 139: i suppose either make the name not terrible or remove the todo :)

tempodb/encoding/vparquet/wal_block_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			actual := tc.b.walPath()
+			assert.Equal(t, tc.expected, actual)
+		})
+	}
+}
+
+// jpe - restore this when what partial failure conditions exist
```
joe-elliott, line 47: this test was meant to test that partial replays work. may make sense to move into v2?

tempodb/wal/wal.go
```diff
@@ -126,7 +139,7 @@ func (w *WAL) RescanBlocks(additionalStartSlack time.Duration, log log.Logger) (
 		}
 
 		if remove {
-			err = os.Remove(filepath.Join(w.c.Filepath, f.Name()))
+			err = os.RemoveAll(filepath.Join(w.c.Filepath, f.Name())) // jpe: handle vparquet wal folders
```
joe-elliott, line 142: i think this will work for `vparquet`. unsure why i flagged it.

tempodb/encoding/vparquet/wal_block.go
```diff
@@ -0,0 +1,577 @@
...
+	tr := traceToParquet(id, trace, getPooledTrace())
+
+	start, end = b.adjustTimeRangeForSlack(start, end, 0)
+
+	b.meta.ObjectAdded(id, start, end)
+
+	// add to current
+	b.traces = append(b.traces, tr)
+	b.ids.Set(id)
+
```
joe-elliott, line 228: i think we need to return `tr` to the pool here. profiling this branch in ops is showing 57% of all allocated memory in the `traceToParquet` conversion code.

electron0zero (approved):

tempodb/encoding/vparquet/wal_block_test.go
mdisibio: It had already been copied to the v2 folder so good there. Agree this is valuable so attempted a parquet version [here](https://github.com/grafana/tempo/pull/1878/commits/8b0464b9f1e2b66194fd7fed3eb15e34ff2cf3bb).  In this case we can only replay whole flushes, not individual traces like in v2.

PR #1882, opened 2022-11-10
use per_cluster_label in dashboard-utils.libsonnet

obliadp:
**What this PR does**:

Changes `tempo-mixin/dashboards/dashboard-utils.libsonnet` to respect `per_cluster_label` from config.libsonnet

Also adds a configurable `namespace_selector_separator` to config.libsonnet. The reason for this is that serviceMonitors from the helm chart don't have job names matching `($namespace)/%s` , but rather `($namespace)-%s`. The config value defaults to current `/`.

### DIFF
--- a/operations/tempo-mixin/config.libsonnet
+++ b/operations/tempo-mixin/config.libsonnet
@@ -26,6 +26,7 @@
     },
 
     per_cluster_label: 'cluster',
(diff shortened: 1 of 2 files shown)


joe-elliott (approved):
Thanks for the improvement!

PR #1883, opened 2022-11-10
Use snake case on Azure Storage config

### LINKED ISSUE #1879: Snake case for Azure backend configuration
Azure backend naming convention is with "-" and the other backends are snake case. This might make it tricky when writing or reading documentation.

For example, the following sample code won't work with azure.
https://github.com/grafana/tempo/blob/main/docs/tempo/website/configuration/parquet.md?plain=1#L47

Use snake case in the azure backend configuration file
https://github.com/grafana/tempo/blob/main/tempodb/backend/azure/config.go

**Additional context**
This will be a breaking change for current Tempo users that uses Azure as the backend. If the team is OK with this, I can create a PR to update the docs and the configuration file.

faustodavid:
**What this PR does**: 
Use snake case on Azure Storage config

Fixes #1879

**Checklist**
- [x] Tests updated
- [x] Documentation added

Should I update the docker-compose azure example now or after Tempo 2.0 is released?


joe-elliott (commented):
Thank you for this fix. I've been wanting to do it for awhile, but haven't found the time.

There is also a file we use for a local azure example that will need to be updated:
`/example/docker-compose/azure/tempo-azure.yaml`

It also looks like you need to run `make jsonnet` to rebuild the jsonnet files? This requires some local tooling setup that can be kind of annoying if you never use jsonnet. I can run this and push to your branch if it's easier.

Thanks @faustodavid!

faustodavid:
> Thank you for this fix. I've been wanting to do it for awhile, but haven't found the time.
> 
> There is also a file we use for a local azure example that will need to be updated: `/example/docker-compose/azure/tempo-azure.yaml`
> 
> It also looks like you need to run `make jsonnet` to rebuild the jsonnet files? This requires some local tooling setup that can be kind of annoying if you never use jsonnet. I can run this and push to your branch if it's easier.
> 
> Thanks @faustodavid!

Happy to contribute! :D

@joe-elliott I wasn't sure if I should update the docker-compose example because if somebody tries to run it without compiling Tempo it will fail.

joe-elliott:
> @joe-elliott I wasn't sure if I should update the docker-compose example because if somebody tries to run it without compiling Tempo it will fail.

This is a good point, but I think we have to keep the examples consistent with tip of main. I'd prefer it fail if you have an older `tempo:latest` than if you have the most recent `tempo:latest`

joe-elliott (approved):
Thanks!

PR #1887, opened 2022-11-10
Stop Distributor on fatal error

rdooley:
**What this PR does**:
* occasionally otel receivers will report a fatal error, and expected behavior is that the Host is stopped
* match this behavior by stopping the receiver shim service and letting the distributor stop itself

If you think we need a test added to prevent regression, let me know

No Issue

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/receiver/shim.go
+++ b/modules/distributor/receiver/shim.go
@@ -279,6 +279,7 @@ func (r *receiversShim) ConsumeTraces(ctx context.Context, td ptrace.Traces) err
 // ReportFatalError implements component.Host
 func (r *receiversShim) ReportFatalError(err error) {
 	_ = level.Error(log.Logger).Log("msg", "fatal error reported", "err", err)
+	r.StopAsync()
 }
 
 // GetFactory implements component.Host
--- a/modules/distributor/receiver/shim.go
+++ b/modules/distributor/receiver/shim.go
@@ -69,6 +69,7 @@ type receiversShim struct {
 	pusher      TracesPusher
 	logger      *log.RateLimitedLogger
 	metricViews []*view.View
+	fatal       chan error
 }
 
 func (r *receiversShim) Capabilities() consumer.Capabilities {
@@ -94,6 +95,7 @@ func New(receiverCfg map[string]interface{}, pusher TracesPusher, middleware Mid
 	shim := &receiversShim{
 		pusher: pusher,
 		logger: log.NewRateLimitedLogger(logsPerSecond, level.Error(log.Logger)),
+		fatal:  make(chan error),
 	}
 
 	// shim otel observability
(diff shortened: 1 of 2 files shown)


modules/distributor/receiver/shim.go
```diff
@@ -279,6 +279,7 @@ func (r *receiversShim) ConsumeTraces(ctx context.Context, td ptrace.Traces) err
 // ReportFatalError implements component.Host
 func (r *receiversShim) ReportFatalError(err error) {
 	_ = level.Error(log.Logger).Log("msg", "fatal error reported", "err", err)
+	r.StopAsync()
```
joe-elliott, line 282: have you tested this? i honestly don't know the services code very well. this may simply stop the receiver, but not cause the distributor to exit.

joe-elliott (commented):
Thanks for the fix. Can we get a changelog entry?

> If you think we need a test added to prevent regression, let me know

If you can think of a meaningful way to test this, we would appreciate it.

modules/distributor/receiver/shim.go
rdooley: have not yet, writing an otel receiver mock to do so

rdooley: (it doesnt work, uh one sec)

rdooley: got this working by switching the shim to be a basicservice
Will update this PR with some testing results on Monday (fed holiday tomorrow I'm off)

rdooley:
Some context before I get into the testing:
- Shopify is using a custom pubsublite receiver and exporter in our fork of opentelemetry-collector
- Occasionally the pubsublite client returns an unretryable error, for the purposes of our test we will simulate this by deleting the pubsublite subscription


After deleting this subscription observed
```
level=error ts=2022-11-14T16:53:11.711264932Z caller=shim.go:221 msg="fatal error reported" err="assigner(projects/${PROJECT_NAME}/locations/us-east1-c/subscriptions/rees-test-again): rpc error: code = NotFound desc = Resource projects/${PROJECT_ID}/locations/us-east1-c/subscriptions/rees-test-again of type SUBSCRIPTION does not exist."
```

And saw pods restart after the logged fatal error
```
kubectl get pods | grep distributor
tempo-distributor-us-central1-a-999446849-tzvnz    1/1     Running   1 (31s ago)   22m
tempo-distributor-us-central1-a-999446849-vh7nn    1/1     Running   1 (31s ago)   22m
tempo-distributor-us-central1-a-999446849-x6vvj    1/1     Running   1 (29s ago)   22m
tempo-distributor-us-central1-c-77cf7f7696-cw6hv   1/1     Running   1 (31s ago)   22m
tempo-distributor-us-central1-c-77cf7f7696-g7257   1/1     Running   1 (29s ago)   22m
tempo-distributor-us-central1-c-77cf7f7696-xfwjj   1/1     Running   1 (29s ago)   22m
```

modules/distributor/receiver/shim.go
```diff
@@ -279,6 +291,7 @@ func (r *receiversShim) ConsumeTraces(ctx context.Context, td ptrace.Traces) err
...
```
mapno, line 294: I think we need to `AwaitTerminated()` here to gracefully shutdown the rest of the receivers of the shim.

mapno, line 294: I think you need to pass the err to `fatal` channel.

rdooley: Mea Culpa
I added the err to fatal channel logic to our fork which is on an older version and completely failed at updating the shim here correctly

mapno (approved):

Issue #1854, opened 2022-11-01
[TraceQL] 2.0 to do


kvrhdn:
Tracking issue for all the changes we want to do to TraceQL to make it release ready!

**Parsing/Correctness**

- [x] Correctly handle empty queries like `{}` and `{ }` https://github.com/grafana/tempo/pull/1993
- [x] Time range isn't handled correctly (startTime doesn't seem to work)
- [ ] Add additional validation to block unsupported queries, e.g. we can't handle arithmetics and nested pipelines yet

**Performance**

- [x] Pass `SearchOptions` down to storage layer to do block sharding as well
- [ ] ~~Load data in two phases? I.e. don't read span ID/name/etc unless needed~~

**Autocomplete**

- [x] Return values with their type, e.g. `{ .http.status_code = ` should autocomplete to `200` and not `"200"` #1956

**Clean up**

- [ ] `TraceSearchMetadata`: refactor `durationMs` to `durationNs` to be more consistent with the spans object

**Stretch / nice to have**

- [ ] Implement arithmetics
- [x] Expressions without spaces aren't parsed correctly, e.g. `.foo="bar"` is treated as one attribute instead of a binary expression. https://github.com/grafana/tempo/pull/1941

joe-elliott:
I would like to close this in favor of individual issues to clean up some of the "unsupported" traceql options. e.g.

https://github.com/grafana/tempo/issues/1949
https://github.com/grafana/tempo/issues/2015

Everything else in this issue is done for 2.0.

PR #1902, opened 2022-11-21
traceql: fix how timestamps are converted from search request

kvrhdn:
**What this PR does**:
We had a (pretty dumb) bug in the conversion from `SearchRequest.StartTime`/`EndTime` to `FetchSpansRequest.StartTimeUnixNanos`/`EndTimeUnixNanos`. The time range in `SearchRequest` is in unix seconds, not milliseconds.

Part of https://github.com/grafana/tempo/issues/1854

**Checklist**
- [x] Tests updated
- [ ] ~~Documentation added~~

### DIFF
--- a/pkg/traceql/engine.go
+++ b/pkg/traceql/engine.go
@@ -106,8 +106,8 @@ func (e *Engine) createFetchSpansRequest(searchReq *tempopb.SearchRequest, pipel
 	// TODO handle SearchRequest.MinDurationMs and MaxDurationMs, this refers to the trace level duration which is not the same as the intrinsic duration
 
 	req := FetchSpansRequest{
-		StartTimeUnixNanos: unixMilliToNano(searchReq.Start),
-		EndTimeUnixNanos:   unixMilliToNano(searchReq.End),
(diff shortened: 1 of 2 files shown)


pkg/traceql/engine.go
```diff
@@ -165,8 +165,8 @@ func (e *Engine) asTraceSearchMetadata(spanset Spanset) *tempopb.TraceSearchMeta
 	return metadata
 }
 
-func unixMilliToNano(ts uint32) uint64 {
-	return uint64(ts) * 1000
+func unixSecToNano(ts uint32) uint64 {
+	return uint64(ts) * 1_000_000_000
```
joe-elliott, line 169: nit: there are constants in the `time` package to add some semantics to conversions like this.

I believe it would be `ts * time.Second / time.Nanosecond` since you are translating seconds -> nanoseconds. Dunno if that reds better or worse :).

joe-elliott (approved):

pkg/traceql/engine.go
kvrhdn: Yeah, that's fair. It's not super compact due to type casting but it's more intent revealing.

PR #1909, opened 2022-11-23
tempo-mixin: don't pull in entire jsonnet-libs repository as dependency

kvrhdn:
**What this PR does**:
https://github.com/grafana/tempo/pull/1901 added a dependency on the entire jsonnet-libs repository, pulling it into `vendor/`. Instead we should add dependencies on subfolders (like we do for `grafana-builder`).
This PR removes the dependency and the vendored repository.

~~Fixes #<issue number>~~

**Checklist**
- [ ] ~~Tests updated~~
- [ ] ~~Documentation added~~


joe-elliott (approved):
Is this why jsonnet PRs are always so ridiculously large?

PR #1918, opened 2022-11-29
Parquet Ingester: Perf Improvements

joe-elliott:
**What this PR does**:
A collection of performance improvements for a Parquet ingester. Specifically targeting the parquet wal.

- Adjust prealloc bytes pool to catch a larger percentage of write sizes
- Rewrite the complete block logic to use rows instead of Trace objects for better pooling
- Improve extendReuseSlice() to keep existing values on extension

This is observed to improve:
- Heap Bytes by 25%
- Working set by 15%
- CPU by 20%

### DIFF
--- a/tempodb/encoding/vparquet/schema.go
+++ b/tempodb/encoding/vparquet/schema.go
@@ -619,5 +619,7 @@ func extendReuseSlice[T any](sz int, in []T) []T {
 		return in[:sz]
 	}
 
-	return make([]T, sz)
+	// append until we're large enough
+	in = in[:cap(in)]
+	return append(in, make([]T, sz-len(in))...)
 }
--- a/tempodb/encoding/vparquet/wal_block.go
+++ b/tempodb/encoding/vparquet/wal_block.go
@@ -30,6 +30,9 @@ var _ common.WALBlock = (*walBlock)(nil)
 // likely the best candidate is some fraction of max trace size per tenant.
 const defaultRowPoolSize = 100000
 
+// completeBlockRowPool is used by the wal iterators and complete block logic to pool rows
+var completeBlockRowPool = newRowPool(defaultRowPoolSize)
+
 // path + filename = folder to create
 //   path/folder/00001
 //   	        /00002
@@ -339,13 +342,12 @@ func (b *walBlock) DataLength() uint64 {
 }
 
 func (b *walBlock) Iterator() (common.Iterator, error) {
-	pool := newRowPool(defaultRowPoolSize)
 	bookmarks := make([]*bookmark[parquet.Row], 0, len(b.flushed))
 
 	for _, page := range b.flushed {
 		idx, _ := parquetquery.GetColumnIndexByPath(page.file, TraceIDColumnName)
 		r := parquet.NewReader(page.file)
-		iter := newRowIterator(r, page.ids.ValuesSortedByID(), pool, idx)
+		iter := newRowIterator(r, page.ids.ValuesSortedByID(), idx)
 
 		bookmarks = append(bookmarks, newBookmark[parquet.Row](iter))
 	}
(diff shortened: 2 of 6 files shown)


zalegrala:
Interesting, nice work.  Is this worth getting some benchmarks on?

joe-elliott:
Unsure why this is showing deltas of ~:

```
name                                old time/op    new time/op    delta
CompleteBlock/vParquet->vParquet-8     38.7s ± 2%     31.2s ± 2%   ~     (p=0.100 n=3+3)

name                                old alloc/op   new alloc/op   delta
CompleteBlock/vParquet->vParquet-8    54.1GB ± 0%    50.8GB ± 0%   ~     (p=0.100 n=3+3)

name                                old allocs/op  new allocs/op  delta
CompleteBlock/vParquet->vParquet-8     56.3M ± 0%     25.0M ± 0%   ~     (p=0.100 n=3+3)
```
You can see we halved the allocs and shaved time off. Small improvement in total memory alloc'ed as well.

mdisibio:
> Unsure why this is showing deltas of ~:
I believe this is because `n=3`.  I get the same, delta is not computed unless `n>=4`.

tempodb/encoding/vparquet/schema.go
```diff
@@ -619,5 +619,7 @@ func extendReuseSlice[T any](sz int, in []T) []T {
 		return in[:sz]
 	}
 
-	return make([]T, sz)
+	// append until we're large enough
+	in = in[:cap(in)]
+	return append(in, make([]T, sz-len(in))...)
```
mdisibio, line 624: In the surface this seems like it would incur 2 allocations:  one in `make(...)` for the new segment, then again inside `append` because we already verified that the capacity is too small.  Is this more efficient than the original, which is 1 alloc?  Wonder if there is any compiler/runtime magic happening.

tempodb/encoding/vparquet/wal_block.go
```diff
@@ -534,43 +549,64 @@ func (i *traceIterator) Next(ctx context.Context) (common.ID, *Trace, error) {
...
 	if err != nil {
 		return nil, nil, err
 	}
 
-	return tr.TraceID, tr, nil
+	row := rows[0]
+	var id common.ID
+	for _, v := range row {
+		if v.Column() == i.traceIDIndex {
+			id = v.ByteArray()
```
mdisibio, line 562: Need to add `break`

mdisibio (approved):
Lgtm, left one question and one small fix.

tempodb/encoding/vparquet/schema.go
mdisibio: Adding another thought:  `append` will over-allocate mem (2*size?) and that would be more efficient when growing the slices than allocing the exact size.  Perhaps `return make([]T, sz*2)` would be the best third option.

joe-elliott: nice catch. a benchmark is showing only one alloc:

```
func BenchmarkExtendReuseSlice(b *testing.B) {
	in := []int{1, 2, 3}
	for i := 0; i < b.N; i++ {
		_ = extendReuseSlice(100, in)
	}
}
```

```
BenchmarkExtendReuseSlice-8   	 3197289	       397.9 ns/op	     896 B/op	       1 allocs/op
```

I imagine that this pattern is common enough that the compiler optimizes it into 1 alloc. Generally I've found that all of the built in slice functions have these kinds of hidden optimizations.

joe-elliott: > Adding another thought: append will over-allocate mem (2*size?)

a quick test is showing worse performance with 2x here. I imagine it's overallocating quite a bit of space for huge traces which is then going unused.

PR #1940, opened 2022-12-07
Parquet WAL: Reduce prefetch to bring down memory requirements

joe-elliott:
**What this PR does**:
Adjusts the multiblock iterator code to expect a new method `peekNextID()`. If this method is supported by the underlying iterator it is used to only call `current()` on those bookmarks that have the lowest ID thereby reducing total memory usage of the ingesters while completing a parquet wal.

Also:
- Adds tests for the multiblock iterator and row iterator
- Removes the unused prefetchIterator

This PR comes with substantial memory improvements over tip of main (all graphs rooted on 0):
image


mapno (approved):
LGTM

PR #1941, opened 2022-12-07
Tweak traceql parsing to correctly parse binary operations without space

### LINKED ISSUE #1939: TraceQL not finding matches in some circumstances
**Describe the bug**
There are likely more, but there are some cases in which TraceQL is not returning matches that the normal search is. For instance looking for traces from the namespace "tempo-ops" and with an attribute bloom="bloom-0":

Using search many results are quickly returned (this should be a commonly found combination internally):
image

However all of the following TraceQL queries over the same time period return nothing:
```
{ .namespace = "tempo-ops" } && { .bloom="bloom-0" }
{ resource.namespace = "tempo-ops" } && { .bloom="bloom-0" }
{ .namespace = "tempo-ops" && .bloom="bloom-0" }
{ resource.namespace = "tempo-ops" && .bloom="bloom-0" }
```

mdisibio:
**What this PR does**:
This PR changes parsing of most binary operations to work without spacing.  Example:  `{.foo="bar"}`.  Previously this wasn't handled as expected  and the entire `foo="bar"` was considered the attribute name.  This was done purposefully to accommodate searching on attribute names with syntaxy characters like `= ~ - +` etc, but the negative impact to every day usage isn't worth it.

Characters in the changed operations can no longer be included in an attribute name so querying on them is impossible.  Some operations are not changed because the likelihood of those characters in attribute names is enough to  weigh the decision the other way.

Fixed: `= != =~ != > >= < <= ^ && ||`
Unchanged: `+ - / % *`

**Overall**
Currently this isn't in a good spot because we are trading parsing flexibility for attribute name support. TraceQL needs some escape hatches to support the full range of [valid attribute names](https://opentelemetry.io/docs/reference/specification/common/attribute-naming/) which is any valid unicode sequence and could include things like newline, space, quotes, etc. There's no way to unambiguously fit this into the current language spec. 

Fixes #1939 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,7 @@ Internal types are updated to use `scope` instead of `instrumentation_library`. 
 * [BUGFIX] tempo-mixin: tweak dashboards to support metrics without `cluster` label present [#1913](https://github.com/grafana/tempo/pull/1913) (@kvrhdn)
 * [ENHANCEMENT] New tenant dashboard [#1901](https://github.com/grafana/tempo/pull/1901) (@mapno)
 * [BUGFIX] Fix docker-compose examples not running on Apple M1 hardware [#1920](https://github.com/grafana/tempo/pull/1920) (@stoewer)
+* [BUGFIX] Fix traceql parsing of most binary operations to not require spacing [#1939](https://github.com/grafana/tempo/pull/1941) (@mdisibio)
 
 ## v1.5.0 / 2022-08-17
 
--- a/pkg/traceql/lexer.go
+++ b/pkg/traceql/lexer.go
@@ -224,12 +224,16 @@ func isDurationRune(r rune) bool {
 }
 
 func isAttributeRune(r rune) bool {
-	return !unicode.IsSpace(r) &&
-		r != scanner.EOF &&
-		r != '(' &&
-		r != ')' &&
-		r != '}' &&
-		r != '{'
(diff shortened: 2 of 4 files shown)


joe-elliott (commented):
I really want to add all of the operators `+ - % * /`, but I recognize that `-` in particular is likely to be somewhere. I suppose there's some consistency in that the arithmetic operators require spaces but the binary operators don't?

wdyt?

mdisibio:
> I really want to add all of the operators `+ - % * /`, but I recognize that `-` in particular is likely to be somewhere. I suppose there's some consistency in that the arithmetic operators require spaces but the binary operators don't?
> 
> wdyt?

Not opposed.  I suppose restricting behavior at the beginning is usually the best because it's easier to relax specifications later (thinking about if we were revert this change down the road and restore support for `-` it would silently break existing queries).   In that sense adding all the operators does seem the best?

joe-elliott:
Checking internal data as expected most of this list does not show up tag names. 
`/` shows up in 2:
image

and `-` predictably shows up in a large number of attributes. I would guess nearly 100 total. Most of which are http header attributes:
image

In my mind it's more likely that someone will query an attribute with `-` in the name then use the arithmetic operators and I want the best experience out of the gate. I think we go with this as is and figure it out on the other side.

PR #1947, opened 2022-12-13
Remove tenant when last block is removed from local backend

joe-elliott:
**What this PR does**:
This fixes a long running issue where tenants were persisted even if they had a zero length wal. The issue was that when rediscovering local blocks a tenant was still found even if there were no blocks. This PR fixes the issue with two changes:

- When rediscovering local blocks don't include tenants without blocks
- When clearing a block in the local backend remove the tenant as well

Technically only the first of the two is needed and would be a safer change. Having the second is nice because it will clear up folders in our ingester local backends. Should we remove the second change and err towards safety? Especially with the 2.0 release coming up?

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/backend/local/compactor.go
+++ b/tempodb/backend/local/compactor.go
@@ -2,6 +2,7 @@ package local
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path"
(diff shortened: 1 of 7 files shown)


joe-elliott:
Another note:

Currently when rediscovering local blocks this PR will nicely skip tenants that don't have blocks. However, it will leave those folders forever. Only when a block is cleared from the local backend do we make the check on whether or not to remove the tenant. This PR, as is, will not clean up already existing empty tenants in the local backend.

tempodb/backend/local/compactor.go
```diff
@@ -20,14 +21,37 @@ func (rw *Backend) MarkBlockCompacted(blockID uuid.UUID, tenantID string) error
...
+	folders, err := os.ReadDir(path)
+	if errors.Is(err, os.ErrNotExist) {
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("failed to list keypath for tenant %s: %w", path, err)
+	}
+
+	if len(folders) == 0 {
+		err = os.RemoveAll(path)
```
mdisibio, line 48: There shouldn't be a race condition here between blocks because this is always done under mutex lock correct?

mdisibio:
> Only when a block is cleared from the local backend do we make the check on whether or not to remove the tenant. This PR, as is, will not clean up already existing empty tenant folders in the local backend.

Does it make sense to do local folder cleanup at ingester shutdown, which could also handle this scenario?

tempodb/backend/local/compactor.go
joe-elliott: I actually don't know. Generally I've considered the backends concurrency safe b/c they are protected by the file system or object storage's own ordering of events. I think this is a point in favor of removing it.

joe-elliott:
Giving this PR more thought and considering your comment I think we should change this PR to only the change in `rediscoverLocalBlocks()`. Thoughts?

> Does it make sense to do local folder cleanup at ingester shutdown, which could also handle this scenario?

This is likely the better approach. Let me see if I can put something together :+1:

tempodb/backend/local/local.go
```diff
@@ -161,9 +161,27 @@ func (rw *Backend) ReadRange(ctx context.Context, name string, keypath backend.K
 	return nil
 }
 
-// Shutdown implements backend.Reader
+// Shutdown implements backend.Reader. It attempts to clear all tenants
+// that do not have blocks.
 func (rw *Backend) Shutdown() {
```
mdisibio, line 166: Forgot this existed.  Excellent.

mdisibio (approved):

tempodb/backend/local/local.go
joe-elliott: Yeah. It's weird that `Shutdown()` on the Reader interface does the job of cleaning up, but meh. We should combine these interfaces anyway. I don't think much is gained from the split up.

(or just add Shutdown to the writer?)

Issue #1134, opened 2021-11-24
Do not add service.name tag values to search data


tarvip:
**Is your feature request related to a problem? Please describe.**

I am trying to replace our current Jaeger and Elasticsearch setup with Grafana Tempo. So far it has been quite promising.

However, there is a small annoyance with search functionality.
We have tracing enabled in Traefik and it sets `service.name` tag, see [here](https://github.com/traefik/traefik/blob/95dc43ce4a0b9a3cdfa1ce33e606af5ef0b09f5b/pkg/middlewares/tracing/forwarder.go#L46)

This `service.name` tag value is also picked up by Tempo for search data and used in Service Name drop-down in UI see attached pictures.

<img width="753" alt="Screen Shot 2021-11-24 at 10 43 19" src="https://user-images.githubusercontent.com/1782293/143213041-c4709ddc-a9eb-4972-ac99-e0fb764a1ad9.png">

<img width="718" alt="Screen Shot 2021-11-24 at 10 43 13" src="https://user-images.githubusercontent.com/1782293/143213064-27b92bb6-ed7e-4162-95b3-7f66fffc7590.png">

On the other hand Jaeger has no such behavior. 

**Describe the solution you'd like**

Would like to avoid having `service.name` tag values in Service Name drop-down.

**Describe alternatives you've considered**
At the moment I am renaming `service.name` tag in opentelemetry-collector, but it feels a bit wrong.

**Additional context**

There is [similar case](https://github.com/traefik/traefik/issues/7578#issuecomment-814953961) with DataDog

kvrhdn:
Hi, glad you're trying out Tempo!

This is an unfortunate conflict between conventions from OpenTelemetry and OpenTracing and also a consequence of how Tempo search is implemented...

1. Why we use `service.name`:

Traces ingested by Tempo are converted into the OpenTelemetry format (even if you are ingesting Jaeger, Zipkin, etc.), internally we use the same receivers as the OpenTelemetry Collector which works with the OpenTelemetry format.

In OpenTelemetry `service.name` is a special tag: by convention it should contain the name of the service this span is coming from. See the OpenTelemetry spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service
So this is why we use `service.name` in the search dropdown.

2. Why search shows `service.name: ping@internal`:

[...]

PR #1956, opened 2022-12-15
[TraceQL] Add API SearchTagValues V2

mdisibio:
**What this PR does**:
This PR adds a new V2 SearchTagValues API that addresses some limitations with the existing api.  

* This new API returns data type information for each value (string,int,keyword,etc), so that the UI can make better decisions on formatting and UX, mainly around adding quotes or not.
* Instead of taking a tag name like `service.name` it takes a TraceQL identifier like `.service.name` and supports intrinsics and scoped attributes.  This fixes ambiguity between builtin keywords and attributes: i.e. `/api/search/tag/name/values` only looks at the span name, and there is no way to lookup the attribute called `name`, and similarly `/api/search/tag/service.name/values` at resource-level `service.name` attribute and there is no way to lookup the span-level value.
* It is V2 because we deemed it too cumbersome to make backwards-compatible with the current API response, which is just a list of strings.

**Example API output**
```
curl http://localhost:3200/api/v2/search/tag/name/values
{
  "tagValues": [
    {
      "type": "string",
      "value": "HTTP GET /"
    },
    {
      "type": "string",
      "value": "HTTP GET /customer"
    }
  ]
}

curl http://localhost:3200/api/v2/search/tag/.http.status_code/values
{
  "tagValues": [
    {
      "type": "int",
      "value": "200"
    }
  ]
}

```

**Status**
~Marking as draft because there are a couple parts that could use a bit more discussion.  Will mark them with comments.~

Related to #1134

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/tempopb/tempo.proto
```diff
@@ -121,6 +122,15 @@ message SearchTagValuesResponse {
   repeated string tagValues = 1;
 }
 
+message TagValue {
```
mdisibio, line 125: Would prefer to leverage existing structs, but think that isn't possible here. The OpenTelemetry attribute struct is not sufficient to convey the type information because it can't distinguish between string and keyword (status "ok", etc), and there is actually no wire format for any TraceQL structs yet.  So this very basic struct was created and we need to settle on the possible values for the type strings.

pkg/api/http.go
```diff
@@ -60,6 +60,8 @@ const (
 	PathEcho            = "/api/echo"
 	PathUsageStats      = "/status/usage-stats"
 
+	PathSearchTagValuesV2 = "/api/v2/search/tag/{tagName}/values"
```
mdisibio, line 63: Since this is our first v2 api, let's make sure we like the versioning style.

tempodb/encoding/vparquet/block_search_tags.go
```diff
@@ -0,0 +1,312 @@
...
+		att = traceql.NewAttribute(tag)
+	}
+
+	// Wrap to v2-style
+	cb2 := func(v traceql.Static) bool {
+		cb(strings.Trim(v.String(), "`"))
+		return false
+	}
+
+	return b.SearchTagValuesV2(ctx, att, cb2, opts)
```
mdisibio, line 162: SearchTagsV1 is now internally a wrapper around the V2 logic. This is good overall but led to a few quirky bits because the root span-name/service-name columns are addressable in V1 (ex: `/search/tag/root.name/values`) but technically don't exist in TraceQL.   So the `nonTraceQLAttributes` map catches these and is checked below.

tempodb/search/util.go
```diff
@@ -24,6 +25,26 @@ func GetVirtualTagValues(tagName string) []string {
 	return nil
 }
 
+func GetVirtualTagValuesV2(tagName string) []tempopb.TagValue {
+
+	switch tagName {
+	case traceql.IntrinsicStatus.String():
+		return []tempopb.TagValue{
+			{Type: "keyword", Value: traceql.StatusOk.String()},
```
mdisibio, line 33: I think these are the only keywords returned so far. Need to settle on the type string here and above.

tempodb/encoding/vparquet/block_search.go
```diff
@@ -641,8 +416,10 @@ func (r *reportValuesPredicate) KeepColumnChunk(cc parquet.ColumnChunk) bool {
 func (r *reportValuesPredicate) KeepPage(pg parquet.Page) bool {
 	if dict := pg.Dictionary(); dict != nil {
 		for i := 0; i < dict.Len(); i++ {
-			s := dict.Index(int32(i)).String()
-			r.cb(s)
+			v := dict.Index(int32(i))
```
joe-elliott, line 419: i don't recall if i fixed this when we discovered the single page per row group improvement in the other predicates. worth double checking.

joe-elliott (commented):
Looking good! I would approve, but it's still marked draft. I'll give it a second pass once you're ready.

tempodb/encoding/vparquet/block_search_tags.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	// now search all row groups
+	var err error
+	rgs := pf.RowGroups()
+	for _, rg := range rgs {
+		// search all special attributes
+		for idx, lbl := range specialAttrIdxs {
+			cc := rg.ColumnChunks()[idx]
+			err = func() error {
```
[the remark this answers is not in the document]
mdisibio: Good catch, fixed.

tempodb/encoding/vparquet/block_search.go
mdisibio: Good catch, fixed.

tempodb/search/util.go
joe-elliott: is `status.code` an old way to refer to `status`?

tempodb/encoding/vparquet/wal_block.go
```diff
@@ -514,6 +514,16 @@ func (b *walBlock) SearchTags(ctx context.Context, cb common.TagCallback, opts c
 }
 
 func (b *walBlock) SearchTagValues(ctx context.Context, tag string, cb common.TagCallback, opts common.SearchOptions) error {
+	// Wrap to v2-style
+	cb2 := func(v traceql.Static) bool {
+		cb(strings.Trim(v.String(), "`"))
```
joe-elliott, line 519: any particular reason we're stripping backticks?

mdisibio: This is wrapper that casts the new v2 `traceql.Static` back to v1 `string`.  Static stringer [includes backticks](https://github.com/grafana/tempo/blob/main/pkg/traceql/ast_stringer.go#L67) for AST roundtrip printing.  This seemed better than reproducing the switch statement, but open to other possibilities.  This is also exists in `backendBlock.SearchTagValues`

tempodb/search/util.go
mdisibio: I think this was from a previous line of thinking.  Removed.

joe-elliott (approved):

tempodb/encoding/vparquet/wal_block.go
mdisibio: Good point.  Added a helper method `Static.EncodeToString(quotes bool)` and refactored.

docs/sources/api_docs/_index.md
```diff
@@ -258,6 +259,48 @@ $ curl -G -s http://localhost:3200/api/search/tag/service.name/values  | jq
 }
 ```
 
+### Search tag values V2
+
+This endpoint retrieves all discovered values and their data types for the given TraceQL identifier.  The endpoint is available in the query frontend service in
+a microservices deployment, or the Tempo endpoint in a monolithic mode deployment. This endpoint is similar to `/api/search/tag/<tag>/values` but operates on TraceQL identifiers and types.  See [TraceQL](../traceql/) documention for more information. The following request will return all discovered service names.
```
knylander-grafana, line 265: Clear explanation of what this does and it has link! Best practice to avoid using future tense, when possible. 
```suggestion
a microservices deployment, or the Tempo endpoint in a monolithic mode deployment. This endpoint is similar to `/api/search/tag/<tag>/values` but operates on TraceQL identifiers and types. See [TraceQL](../traceql/) documention for more information. The following request returns all discovered service names.
```

docs/sources/api_docs/_index.md
```diff
@@ -258,6 +259,48 @@ $ curl -G -s http://localhost:3200/api/search/tag/service.name/values  | jq
...
+
+```
+GET /api/search/tag/.service.name/values
+```
+
+#### Example
+
+Example of how to query Tempo using curl.
+This query will return all discovered values for the tag `service.name`.
```
knylander-grafana, line 274:
```suggestion
This query returns all discovered values for the tag `service.name`.
```

knylander-grafana, line 274:
```suggestion
This example queries Tempo using curl and returns all discovered values for the tag `service.name`.
```

knylander-grafana (commented):
Doc approved. Thank you for adding API doc!

PR #1960, opened 2022-12-16
Make intrinsic dimensions configurable and disable status_message by default

### LINKED ISSUE #1932: Ability to remove dimension from intrinsicDimensions
We had a lot of prometheus OOM crash lately now i've checked and the reason is status_message field and as far as i see there is no option to disable intrinsicDimensions
response from tsdb analyze:
```
Highest cardinality labels:
41449 status_message
```

I'd like to be able to change intrinsicDimensions

stoewer:
**What this PR does**:

The metrics generator always adds the intrinsic dimensions `service`, `span_name`, `span_kind`, `status_code`, and `status_message`. The `status_message` may contain arbitrary strings and can have a very high cardinality (see #1932). Therefore it is safer not to add this dimension by default.

With this PR:
* All above mentioned intrinsic dimensions are configurable via configuration file and overrides.
* The `status_message` is not added by default.

Fixes #1932

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,7 @@ Internal types are updated to use `scope` instead of `instrumentation_library`. 
 * [BUGFIX] New wal file separator '+' for the NTFS filesystem and backward compatibility with the old separator ':' [#1700](https://github.com/grafana/tempo/pull/1700) (@kilian-kier)
 * [ENHANCEMENT] metrics-generator: extract `status_message` field from spans [#1786](https://github.com/grafana/tempo/pull/1786), [#1794](https://github.com/grafana/tempo/pull/1794) (@stoewer)
 * [ENHANCEMENT] metrics-generator: handle collisions between user defined and default dimensions [#1794](https://github.com/grafana/tempo/pull/1794) (@stoewer)
+* [ENHANCEMENT] metrics-generator: make intrinsic dimensions configurable and disable `status_message` by default [#1960](https://github.com/grafana/tempo/pull/1960) (@stoewer)
 * [ENHANCEMENT] distributor: Log span names when `distributor.log_received_spans.include_all_attributes` is on [#1790](https://github.com/grafana/tempo/pull/1790) (@suraciii)
 * [ENHANCEMENT] metrics-generator: truncate label names and values exceeding a configurable length [#1897](https://github.com/grafana/tempo/pull/1897) (@kvrhdn)
 * [ENHANCEMENT] Add parquet WAL [#1878](https://github.com/grafana/tempo/pull/1878) (@joe-elliott, @mdisibio)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,7 +97,7 @@ Internal types are updated to use `scope` instead of `instrumentation_library`. 
 * [BUGFIX] New wal file separator '+' for the NTFS filesystem and backward compatibility with the old separator ':' [#1700](https://github.com/grafana/tempo/pull/1700) (@kilian-kier)
 * [ENHANCEMENT] metrics-generator: extract `status_message` field from spans [#1786](https://github.com/grafana/tempo/pull/1786), [#1794](https://github.com/grafana/tempo/pull/1794) (@stoewer)
 * [ENHANCEMENT] metrics-generator: handle collisions between user defined and default dimensions [#1794](https://github.com/grafana/tempo/pull/1794) (@stoewer)
-  **BREAKING CHANGE** Custom dimensions colliding with intrinsic dimensions will be prefixed with `--`.
+  **BREAKING CHANGE** Custom dimensions colliding with intrinsic dimensions will be prefixed with `__`.
 * [ENHANCEMENT] metrics-generator: make intrinsic dimensions configurable and disable `status_message` by default [#1960](https://github.com/grafana/tempo/pull/1960) (@stoewer)
 * [ENHANCEMENT] distributor: Log span names when `distributor.log_received_spans.include_all_attributes` is on [#1790](https://github.com/grafana/tempo/pull/1790) (@suraciii)
 * [ENHANCEMENT] metrics-generator: truncate label names and values exceeding a configurable length [#1897](https://github.com/grafana/tempo/pull/1897) (@kvrhdn)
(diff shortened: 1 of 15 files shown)


mapno (commented):
LGTM. If you can re-generate `docs/tempo/website/configuration/manifest.md`, that'd be great. It's due an update.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -40,10 +36,27 @@ type Processor struct {
...
+	if cfg.IntrinsicDimensions.SpanKind {
+		labels = append(labels, dimSpanKind)
+	}
+	if cfg.IntrinsicDimensions.StatusCode {
+		labels = append(labels, dimStatusCode)
+	}
+	if cfg.IntrinsicDimensions.StatusMessage {
+		labels = append(labels, dimStatusMessage)
+	}
+	intrinsicDimensions := labels[:]
```
joe-elliott, line 56: why copy the slice here? it's pointing to the same underlying array.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -108,7 +128,7 @@ func (p *Processor) aggregateMetricsForSpan(svcName string, rs *v1.Resource, spa
 	p.spanMetricsDurationSeconds.ObserveWithExemplar(registryLabelValues, latencySeconds, tempo_util.TraceIDToHexString(span.TraceId))
 }
 
-func sanitizeLabelNameWithCollisions(name string) string {
+func sanitizeLabelNameWithCollisions(name string, intrinsicDimensions []string) string {
```
joe-elliott, line 131: sanitizing attribute names only if they conflict with _configured_ intrinsics has some weird properties. if someone configures `status_message` as a normal attribute and then later adds the matching intrinsic then the name of the normal attribute label changes. this may break dashboards or queries b/c it changes the meaning of their metric labels. similar situations can occur when disabling intrinsic dimensions.

my instinct is to sanitize against all intrinsics even if they are not configured.

CHANGELOG.md
```diff
@@ -97,6 +97,7 @@ Internal types are updated to use `scope` instead of `instrumentation_library`.
 * [BUGFIX] New wal file separator '+' for the NTFS filesystem and backward compatibility with the old separator ':' [#1700](https://github.com/grafana/tempo/pull/1700) (@kilian-kier)
 * [ENHANCEMENT] metrics-generator: extract `status_message` field from spans [#1786](https://github.com/grafana/tempo/pull/1786), [#1794](https://github.com/grafana/tempo/pull/1794) (@stoewer)
 * [ENHANCEMENT] metrics-generator: handle collisions between user defined and default dimensions [#1794](https://github.com/grafana/tempo/pull/1794) (@stoewer)
+* [ENHANCEMENT] metrics-generator: make intrinsic dimensions configurable and disable `status_message` by default [#1960](https://github.com/grafana/tempo/pull/1960) (@stoewer)
```
joe-elliott, line 100: i believe there's a breaking change here (or perhaps one in #1786) that should be mentioned. If the user has a configured normal attribute dimension called `status_message` then it's name will change to `__status_message`.

See below for comments on the renaming behavior.

modules/overrides/limits.go
```diff
@@ -60,17 +60,18 @@ type Limits struct {
...
+	MetricsGeneratorForwarderWorkers                        int             `yaml:"metrics_generator_forwarder_workers" json:"metrics_generator_forwarder_workers"`
+	MetricsGeneratorProcessorServiceGraphsHistogramBuckets  []float64       `yaml:"metrics_generator_processor_service_graphs_histogram_buckets" json:"metrics_generator_processor_service_graphs_histogram_buckets"`
+	MetricsGeneratorProcessorServiceGraphsDimensions        []string        `yaml:"metrics_generator_processor_service_graphs_dimensions" json:"metrics_generator_processor_service_graphs_dimensions"`
+	MetricsGeneratorProcessorSpanMetricsHistogramBuckets    []float64       `yaml:"metrics_generator_processor_span_metrics_histogram_buckets" json:"metrics_generator_processor_span_metrics_histogram_buckets"`
+	MetricsGeneratorProcessorSpanMetricsDimensions          []string        `yaml:"metrics_generator_processor_span_metrics_dimensions" json:"metrics_generator_processor_span_metrics_dimensions"`
+	MetricsGeneratorProcessorSpanMetricsIntrinsicDimensions map[string]bool `yaml:"metrics_generator_processor_span_metrics_intrinsic_dimensions" json:"metrics_generator_processor_span_metrics_intrinsic_dimensions"`
```
joe-elliott, line 74: should we fail to load overrides if a value is specified here that is not an intrinsic dimension?
```
metrics_generator_processor_span_metrics_intrinsic_dimensions:
  status_message: true
  foo: false
```

joe-elliott (commented):
Some thoughts.

stoewer:
> If you can re-generate docs/tempo/website/configuration/manifest.md, that'd be great. It's due an update.

Good idea. I regenerated it.

modules/generator/processor/spanmetrics/spanmetrics.go
stoewer: This was needed in order to detect collisions with configured intrinsic dimensions. This is no longer needed as dimensions are now also treated as collisions when the colliding intrinsic dimension is absent.

modules/generator/processor/spanmetrics/spanmetrics.go
stoewer: Done

CHANGELOG.md
stoewer: Agreed. I think the actual breaking change was in #1794. I marked the change as breaking in the changelog.

modules/generator/overrides_test.go
```diff
@@ -43,3 +44,7 @@ func (m *mockOverrides) MetricsGeneratorProcessorSpanMetricsHistogramBuckets(use
 func (m *mockOverrides) MetricsGeneratorProcessorSpanMetricsDimensions(userID string) []string {
 	return m.spanMetricsDimensions
 }
+
+func (m *mockOverrides) MetricsGeneratorProcessorSpanMetricsIntrinsicDimensions(userID string) map[string]bool {
```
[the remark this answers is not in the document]
stoewer: Sound good. I added both tests.

modules/overrides/limits.go
stoewer: I could not think of a way to fail while loading the overrides without running into import cycles or repleating the intrinsic dimension definition in the overrides package.

I solved it by checking for valid keys when the overrides are applied.

mapno (approved):
LGTM, nice work

PR #1962, opened 2022-12-16
fix: complete writing batch even if some spans refused

### LINKED ISSUE #1957: Complete writing a batch even if some spans refused
**Describe the bug**
Currently if an ingester refuses traces b/c a trace is too large it will stop writing the current batch and return an error. The end result is that the first part of the batch was written but everything after the "trace too large" error is dropped. Let's continue writing the batch even if we receive a trace too large error.

Loop to write traces:
https://github.com/grafana/tempo/blob/909b487d4bac42797a786a10796664b4cfcd0746/modules/ingester/instance.go#L167

Trace too large returned here:
https://github.com/grafana/tempo/blob/909b487d4bac42797a786a10796664b4cfcd0746/modules/ingester/instance.go#L210

97amarnathk:
**What this PR does**:
* Currently if an ingester refuses traces b/c a trace is too large it will stop writing the current batch and return an error. The end result is that the first part of the batch was written but everything after the "trace too large" error is dropped.
* This tries to ingest all spans. Incase of failures, errors are collected, and returned at end after processing entire batch.

Fixes #1957 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * [ENHANCEMENT] Add synchronous read mode to vParquet and vParquet2 optionally enabled by env vars  [#2165](https://github.com/grafana/tempo/pull/2165) (@mdisibio)
 * [ENHANCEMENT] Add option to override metrics-generator ring port  [#2399](https://github.com/grafana/tempo/pull/2399) (@mdisibio)
 * [BUGFIX] tempodb integer divide by zero error [#2167](https://github.com/grafana/tempo/issues/2167) (@kroksys)
+* [BUGFIX] Complete writing a batch even if some spans refused. [#1962](https://github.com/grafana/tempo/pull/1962)
 
 ## v2.1.1 / 2023-04-28
 * [BUGFIX] Fix issue where Tempo sometimes flips booleans from false->true at storage time. [#2400](https://github.com/grafana/tempo/issues/2400) (@joe-elliott)
--- a/modules/ingester/instance.go
+++ b/modules/ingester/instance.go
@@ -19,6 +19,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"go.uber.org/atomic"
+	"go.uber.org/multierr"
 	"google.golang.org/grpc/codes"
 
 	"github.com/grafana/tempo/modules/overrides"
@@ -143,13 +144,14 @@ func newInstance(instanceID string, limiter *Limiter, writer tempodb.Writer, l *
 }
 
 func (i *instance) PushBytesRequest(ctx context.Context, req *tempopb.PushBytesRequest) error {
+	var errs []error
 	for j := range req.Traces {
 		err := i.PushBytes(ctx, req.Ids[j].Slice, req.Traces[j].Slice)
 		if err != nil {
-			return err
+			errs = append(errs, err)
 		}
 	}
-	return nil
+	return multierr.Combine(errs...)
 }
 
 // PushBytes is used to push an unmarshalled tempopb.Trace to the instance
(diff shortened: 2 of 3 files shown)


mapno (commented):
Can you add an entry in the CHANGELOG?

97amarnathk:
Done.

mapno (commented):
Sorry, I should have left this in the previous review: can you add tests to cover the new expected behaviour? There are a few candidate places, like [instance_test.go](https://github.com/grafana/tempo/blob/main/modules/ingester/instance_test.go) or the [e2e tests](https://github.com/grafana/tempo/tree/main/integration/e2e).

Otherwise, it looks good to me. Thanks for the contribution.

joe-elliott (commented):
Thanks for taking this on! We do have one thing we need to iron out. As written I'm not sure what will happen as this error object is sent over GRPC to the distributors. There are two concerns that come out of this:

- Currently the error is propagated back through the distributor to the sender as is and the error code (InvalidArgument, FailedPrecondition, etc.) is meaningful to the sender. The OTel Collector/Grafana Agent uses these [error codes](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) to determine if they [should resend](https://github.com/open-telemetry/opentelemetry-collector/blob/150c1ede2f7fb5607035d3c7a10bfcab61c39afe/exporter/otlpexporter/otlp.go#L166). While aggregating the errors to return I believe we need to make sure that if ANY error code is a not retryable then the whole error code is not retryable. If all error codes are retryable then we can return a retryable error code to the sender.

- Secondly we use the string prefix in the distributor to count different metrics: https://github.com/grafana/tempo/blob/main/modules/distributor/distributor.go#L559. I believe this change would break that behavior.

Working these issues out cleanly is going to require a bit of thought and maybe some digging on GRPC behavior.

joe-elliott:
I'm hoping I have not blocked @97amarnathk on this change with my comment. We do need to confirm that the correct GRPC codes are returned from the distributor when the user hits errors like "trace too large" or "live traces exceeded", but this is an excellent improvement that we all want to see in the codebase. 

I think with sufficient finagling we could write some nice integration tests that confirm the correct grpc codes are returned. These tests are close:

https://github.com/grafana/tempo/blob/main/integration/e2e/limits_test.go#L34

but only check to see that specific substrings are present. We can likely improve them to also check the GRPC error.

97amarnathk:
Hey @mapno I have updated the instance tests.

CHANGELOG.md
```diff
@@ -698,4 +699,4 @@ Jsonnet users will now need to specify a storage request and limit for the gener
 * [BUGFIX] S3 multi-part upload errors [#306](https://github.com/grafana/tempo/pull/325)
 * [BUGFIX] Increase Prometheus `notfound` metric on tempo-vulture. [#301](https://github.com/grafana/tempo/pull/301)
 * [BUGFIX] Return 404 if searching for a tenant id that does not exist in the backend. [#321](https://github.com/grafana/tempo/pull/321)
-* [BUGFIX] Prune in-memory blocks from missing tenants. [#314](https://github.com/grafana/tempo/pull/314)
+* [BUGFIX] Prune in-memory blocks from missing tenants. [#314](https://github.com/grafana/tempo/pull/314)
```
joe-elliott, line 702: unsure what the diff is here, but let's remove it? maybe an errant newline?

joe-elliott (commented):
Thanks for reupping this! So the big concern is that when we refuse traces we need to make sure we are returning the correct GRPC error code.

When we return an error like ["too many traces"](https://github.com/grafana/tempo/blob/main/modules/ingester/instance.go#L166) or ["trace too large"](https://github.com/grafana/tempo/blob/main/modules/ingester/instance.go#L192) we purposefully return `FailedPrecondition` so that it is [not retried](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/otlpexporter/otlp.go#L163). B/c we are returning a GRPC error it is correctly propagated back to the caller. I'm not honestly not sure what GRPC would do with a "multierr".

One thing we can do is adjust this test:
https://github.com/grafana/tempo/blob/main/integration/e2e/limits_test.go#L18

Currently it uses Jaeger thrift to push data and confirm that various limits are hit. It would be nice to update this test to use GRPC instead and verify the error codes are as expected.

PR #1963, opened 2022-12-19
Remove mentions of synthetic-load-generator from example README

stoewer:
**What this PR does**:

Example README files still mentioned the synthetic-load-generator which is no longer used in most examples. I also regenerated the output of `docker-compose ps` as it also looks quite different now in some cases.

None, I just noticed this while working on the debug example.

**Checklist**
- [ ] Tests updated
- [x] Documentation added


mapno (approved):
Nice 👍

PR #1966, opened 2022-12-26
Add support for TraceQL in Parquet WAL and Local Blocks

electron0zero:
**What this PR does**:
Add support for TraceQL in Parquet WAL. We added Parquet WAL in #1878, but TraceQL Search is not supported in Parquet WAL. 

**How?**
In ingesters, a trace can be live, in headBlock, in the completing blocks and complete blocks.
we search in this order: HeadBlock, CompletingBlocks, and completeBlocks.
search is done by Iterating over blocks, executing TraceQL, and finally collecting results.

note: searching liveTraces is not supported yet

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/ingester/instance_search.go
```diff
@@ -32,11 +34,47 @@ func (i *instance) Search(ctx context.Context, req *tempopb.SearchRequest) (*tem
...
-		// TODO actually implement recent traceQL search
+		// A trace can be live, in headBlock, completing blocks and complete blocks.
+		// search in this order: HeadBlock, CompletingBlocks, and then finally in completeBlocks.
+		// Iterate over blocks, execute TraceQL, collect results, dedupe, sort and return traces.
+		// return early if we hit maxResults during search.
+		traces, metrics, blockErrs := i.searchWALWithTraceQL(ctx, req, maxResults)
+		err := multierr.Combine(blockErrs...)
+		if err != nil {
+			span.LogFields(ot_log.Error(fmt.Errorf("blockErrors: %w", err)))
+		}
```
mapno, line 47: Not sure about the error handling in search in general. We always return successful, even if we got no results and everything failed. I think we should bubble it up more than a log in a span, that will stay hidden 99% of the time.

modules/ingester/instance_search.go
```diff
@@ -205,6 +243,106 @@ func (i *instance) searchWAL(ctx context.Context, req *tempopb.SearchRequest, p
...
+	defer i.blocksMtx.RUnlock()
+	span.LogFields(ot_log.Event("blocksMtx acquired"))
+
+	// search headBlock
+	res, err := engine.Execute(ctx, req, traceql.NewSpansetFetcherWrapper(func(ctx context.Context, req traceql.FetchSpansRequest) (traceql.FetchSpansResponse, error) {
+		// span.SetTag("blockID", i.headBlock.BlockMeta().BlockID.String())
+		return i.headBlock.Fetch(ctx, req, opts)
+	}))
+	if err != nil {
+		blockErrs = append(blockErrs, err)
```
mapno, line 274: Log this error?

modules/ingester/instance_search.go
```diff
@@ -205,6 +243,106 @@ func (i *instance) searchWAL(ctx context.Context, req *tempopb.SearchRequest, p
...
+		ot_log.Int("traces_size", len(traces)))
+
+	if len(traces) >= maxResults {
+		// hit limit, return early
+		return traces, metrics, blockErrs
+	}
+
+	// search completingBlocks
+	for _, block := range i.completingBlocks {
+		if len(traces) >= maxResults {
```
mapno, line 292: This logic can fail if there are duplicates, no?

modules/ingester/instance_search.go
```diff
@@ -205,6 +243,106 @@ func (i *instance) searchWAL(ctx context.Context, req *tempopb.SearchRequest, p
...
+		}
+
+		// TODO: Execute will parse the query again and again... maybe fix that
+		// TODO: we are building NewSpansetFetcherWrapper for each block, feels bit wasteful :)
+		r, err := engine.Execute(ctx, req, traceql.NewSpansetFetcherWrapper(func(ctx context.Context, req traceql.FetchSpansRequest) (traceql.FetchSpansResponse, error) {
+			return block.Fetch(ctx, req, opts)
+		}))
+
+		if err != nil {
+			blockErrs = append(blockErrs, err)
```
mapno, line 303: I think we should at least be logging these

modules/ingester/instance_search.go
```diff
@@ -205,6 +243,106 @@ func (i *instance) searchWAL(ctx context.Context, req *tempopb.SearchRequest, p
...
+			continue // collect error and continue to next block
+		}
+		traces = append(traces, r.Traces...)
+		metrics = mergeSearchMetrics(metrics, res.Metrics)
+	}
+	span.LogFields(
+		ot_log.String("msg", "done searching completeBlocks"),
+		ot_log.Int("traces_size", len(traces)))
```
mapno, line 330: Something like `total_traces_size` would be more clear that this is also includes traces from the `headBlock` and `completingBlocks`.

```suggestion
		ot_log.Int("total_traces_size", len(traces)))
```

mapno (commented):
Left a few comments, but I like the changes 👍

modules/ingester/instance_search.go
electron0zero: it can return less then max results if we have duplicates, will remove limits logic from block search.

modules/ingester/instance_search.go
electron0zero: done

modules/ingester/instance_search.go
electron0zero: done

modules/ingester/instance_search.go
electron0zero: removed this logic, we will end up doing extra work when searching blocks but we will always get correct results.

modules/ingester/instance_search.go
```diff
@@ -32,11 +35,67 @@ func (i *instance) Search(ctx context.Context, req *tempopb.SearchRequest) (*tem
 		maxResults = 20
 	}
 
+	span.LogFields(ot_log.String("SearchRequest", req.String()))
 	if api.IsTraceQLQuery(req) {
-		// TODO actually implement recent traceQL search
+		// A trace can be live, in headBlock, completing blocks and complete blocks.
```
joe-elliott, line 40: instead of writing all this new code here to can't we just add lines in these two places:

https://github.com/grafana/tempo/blob/main/modules/ingester/instance_search.go#L172
and 
https://github.com/grafana/tempo/blob/main/modules/ingester/instance_search.go#L258

This would hook into the existing goroutine/channel/pipeline thing we already have built?

joe-elliott: Basically if it's traceql do something like:

```
	res, err := engine.Execute(ctx, req, traceql.NewSpansetFetcherWrapper(func(ctx context.Context, req traceql.FetchSpansRequest) (traceql.FetchSpansResponse, error) {
		return i.headBlock.Fetch(ctx, req, opts)
	}))
```

else do what it's currently doing?

joe-elliott:
I'm currently in the middle of a config refactor that has the potential to strongly conflict with this PR. Please ignore the review comments about adjusting tests. I will handle it on my side.

modules/ingester/ingester_test.go
```diff
@@ -83,7 +84,7 @@ func TestPushQueryAllEncodings(t *testing.T) {
 
 func TestFullTraceReturned(t *testing.T) {
 	ctx := user.InjectOrgID(context.Background(), "test")
-	ingester, _, _ := defaultIngester(t, t.TempDir())
+	ingester, _, _ := defaultIngester(t, t.TempDir(), v2.VersionString)
```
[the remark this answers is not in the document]
joe-elliott: Please ignore. I've come across some issues while working on a config change PR that will waste a lot of your time. Let me solve these issues and just focus on the other comments.

modules/ingester/instance_search.go
electron0zero: updated.

modules/ingester/instance_search.go
```diff
@@ -163,15 +160,36 @@ func (i *instance) searchWAL(ctx context.Context, req *tempopb.SearchRequest, p
...
+				return b.Fetch(ctx, req, opts)
+			}))
+		} else {
+			resp, err = b.Search(ctx, req, opts)
+		}
+
 		if err != nil {
-			level.Error(log.Logger).Log("msg", "error searching wal block", "blockID", b.BlockMeta().BlockID.String(), "err", err)
+			msg := "error searching local block"
+			if isTraceQL && errors.Is(err, common.ErrUnsupported) {
```
joe-elliott, line 188: do we need special processing here for traceql and unsupported? i'm fine with just logging the error regardless

modules/ingester/instance_search.go
```diff
@@ -250,14 +269,35 @@ func (i *instance) searchLocalBlocks(ctx context.Context, req *tempopb.SearchReq
...
+					return e.Fetch(ctx, req, opts)
+				}))
+			} else {
+				resp, err = e.Search(ctx, req, opts)
+			}
+
 			if err != nil {
-				level.Error(log.Logger).Log("msg", "error searching local block", "blockID", blockID, "err", err)
+				msg := "error searching local block"
+				if errors.Is(err, common.ErrUnsupported) && isTraceQL {
```
joe-elliott, line 296: same as above

joe-elliott (approved):

modules/ingester/instance_search.go
electron0zero: not really, I did this to make this error more verbose. will just log the error and the tarceql flag.

PR #1975, opened 2023-01-05
Collect inspectedBytes from SearchMetrics

electron0zero:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/searchsharding.go
+++ b/modules/frontend/searchsharding.go
@@ -32,7 +32,7 @@ const (
 var (
 	metricInspectedBytes = promauto.NewHistogram(prometheus.HistogramOpts{
 		Namespace: "tempo",
-		Name:      "frontend_query_results_metrics_inspected_bytes",
+		Name:      "query_frontend_result_metrics_inspected_bytes",
 		Help:      "Inspected Bytes in a search query",
 		Buckets:   prometheus.ExponentialBuckets(1024*1024, 2, 10), // from 1MB up to 1GB
 	})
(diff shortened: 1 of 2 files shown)


modules/frontend/searchsharding.go
```diff
@@ -27,6 +29,15 @@ const (
 	defaultConcurrentRequests    = 50
 )
 
+var (
+	metricInspectedBytes = promauto.NewHistogram(prometheus.HistogramOpts{
+		Namespace: "tempo",
+		Name:      "frontend_query_results_metrics_inspected_bytes",
```
joe-elliott, line 35: other metrics from this module are named "query_frontend...

e.g.  https://github.com/grafana/tempo/blob/main/modules/frontend/retry.go#L18

joe-elliott (approved):
do we need the metric? the changes to the log line will make it easy to calculate this metric using loki

modules/frontend/searchsharding.go
electron0zero: I am leaning on the side of keeping the metric for now, and testing it out in dev. we can remove if it's not useful.

will rename the metric.

electron0zero:
> do we need the metric? the changes to the log line will make it easy to calculate this metric using loki

I am leaning on the side of keeping the metric for now, and testing it out in dev. we can remove if it's not useful.  also OSS users might not have Loki for logs.

joe-elliott (approved):

PR #1984, opened 2023-01-09
[release-v1.3] Update publishing workflows to use PATs with fine-grained access control

jdbaldry:


mdisibio:
The CI errors don't look related to the changes in this PR, probably the `release-1.3` branch is stale.  However, it seems good to address the token handling even if we don't expect to build 1.3 any time soon, so I will force merge this. Thanks!

PR #1987, opened 2023-01-09
TraceQL: Fix bug trying to optimize queries using only && and repeated attribute name

### LINKED ISSUE #1948: TraceQL: span scope not working with ranges
`{ span.http.status_code > 200 && span.http.status_code < 300 }` returns an empty resultset:

image

`{ .http.status_code > 200 && .http.status_code < 300 }` returns hits:

image

`{ span.http.status_code > 200 }` returns hits:

image

mdisibio:
**What this PR does**:
Fixes an issue when trying to optimize some queries using only `&&` which turns into the `AllConditions=true` storage layer hint, and including repeated attribute names.

Fixes #1948 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/storage.go
+++ b/pkg/traceql/storage.go
@@ -74,11 +74,13 @@ func MustExtractFetchSpansRequest(query string) FetchSpansRequest {
 // ExtractFetchSpansRequest parses the given traceql query and returns
 // the storage layer conditions. Returns an error if the query fails to parse.
 func ExtractFetchSpansRequest(query string) (FetchSpansRequest, error) {
-	req := FetchSpansRequest{}
-
 	ast, err := Parse(query)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/traceql/storage.go
```diff
@@ -61,26 +61,30 @@ type SpansetFetcher interface {
...
 	if err != nil {
-		return cond, err
+		return FetchSpansRequest{}, err
 	}
-	req := &FetchSpansRequest{}
-	ast.Pipeline.extractConditions(req)
-	return req.Conditions, nil
+
+	req := FetchSpansRequest{
+		AllConditions: true,
```
mapno, line 83: I would add TODO here to remove when we support more types of queries.

mapno (approved):

PR #2003, opened 2023-01-18
Return more consistent search results by combining partial traces

### LINKED ISSUE #2001: [TraceQL] Duration reported in UI is occasionally wrong
**Describe the bug**
Occasionally the duration reported by the UI is not correct. Example:
image

My guess is that this is due to the search matching a partial trace which has the search duration listed. Then, when the full trace is returned via trace by id, the total duration is larger. If true then recent traces and large traces would be the most susceptible to this issue.

A perfect fix may be impossible in an RF=3 environment, but one possible way to alleviate it would be to adjust this code:

https://github.com/grafana/tempo/blob/972f630afc655433c384e8362d065d9f3e72b30e/modules/frontend/search_response.go#L73

Currently the first trace that returns establishes the start time and duration. Instead we could coalesce start time by taking the min and coalesce duration by taking the max.

mapno:
**What this PR does**:

Makes search results by combining partial results and taking the earliest start time and longest duration.

Fixes #2001

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -171,6 +171,7 @@ Internal types are updated to use `scope` instead of `instrumentation_library`. 
 * [BUGFIX] Don't persist tenants without blocks in the ingester[#1947](https://github.com/grafana/tempo/pull/1947) (@joe-elliott)
 * [BUGFIX] TraceQL: span scope not working with ranges [#1948](https://github.com/grafana/tempo/issues/1948) (@mdisibio)
 * [BUGFIX] TraceQL: skip live traces search [#1997](https://github.com/grafana/tempo/pull/1997) (@mapno)
+* [BUGFIX] Return more consistent search results by combining partial traces [#](https://github.com/tempo/pull/) (@mapno)
 
 ## v1.5.0 / 2022-08-17
 
--- a/modules/frontend/search_response.go
+++ b/modules/frontend/search_response.go
@@ -68,9 +68,10 @@ func (r *searchResponse) addResponse(res *tempopb.SearchResponse) {
 	defer r.mtx.Unlock()
 
 	for _, t := range res.Traces {
-		// todo: determine a better way to combine?
 		if _, ok := r.resultsMap[t.TraceID]; !ok {
 			r.resultsMap[t.TraceID] = t
+		} else {
+			coalesce(r.resultsMap[t.TraceID], t)
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


modules/frontend/search_response.go
```diff
@@ -139,3 +140,15 @@ func (r *searchResponse) result() *tempopb.SearchResponse {
...
+// coalesce combines two trace search metadata objects into one by taking the earliest start time and longest duration
+func coalesce(a, b *tempopb.TraceSearchMetadata) {
+	// Take min start time and max duration
+	if a.StartTimeUnixNano > b.StartTimeUnixNano {
+		a.StartTimeUnixNano = b.StartTimeUnixNano
+	}
+	if a.DurationMs < b.DurationMs {
+		a.DurationMs = b.DurationMs
+	}
+	// TODO: Merge spansets?
```
joe-elliott, line 153: there are no spansets in the metadata

modules/frontend/search_response.go
```diff
@@ -139,3 +140,15 @@ func (r *searchResponse) result() *tempopb.SearchResponse {
 
 	return res
 }
+
+// coalesce combines two trace search metadata objects into one by taking the earliest start time and longest duration
+func coalesce(a, b *tempopb.TraceSearchMetadata) {
```
joe-elliott, line 145: looks like this function already exists: https://github.com/grafana/tempo/blob/main/tempodb/search/util.go#L55

mapno: Changed 0245a3eb9

modules/frontend/search_response.go
mapno: Removed! 0245a3eb9

joe-elliott (approved):

CHANGELOG.md
```diff
@@ -171,6 +171,7 @@ Internal types are updated to use `scope` instead of `instrumentation_library`.
 * [BUGFIX] Don't persist tenants without blocks in the ingester[#1947](https://github.com/grafana/tempo/pull/1947) (@joe-elliott)
 * [BUGFIX] TraceQL: span scope not working with ranges [#1948](https://github.com/grafana/tempo/issues/1948) (@mdisibio)
 * [BUGFIX] TraceQL: skip live traces search [#1997](https://github.com/grafana/tempo/pull/1997) (@mapno)
+* [BUGFIX] Return more consistent search results by combining partial traces [#](https://github.com/tempo/pull/) (@mapno)
```
electron0zero, line 174: @mapno url broken?

PR #2004, opened 2023-01-18
Remove feature flags and default to on

joe-elliott:
**What this PR does**:
Removes two feature flags `search_enabled` and `metrics_generator_enabled` and defaults both of them to on. Additional logic needed to be added into the metrics generator for single binary installs that did not configure the metrics generator. 

This PR also updates and cleans up all of the docker-compose and jsonnet examples.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [x] Documentation added


cmd/tempo/app/app.go
```diff
@@ -96,21 +94,11 @@ func New(cfg Config) (*App, error) {
...
 	}
 
-	statFeatureEnabledMetricsGenerator.Set(0)
-	if cfg.MetricsGeneratorEnabled {
-		statFeatureEnabledMetricsGenerator.Set(1)
-	}
-
 	statFeatureEnabledMultitenancy.Set(0)
-	if cfg.SearchEnabled {
+	if cfg.MultitenancyEnabled {
```
kvrhdn, line 98: Was this a bug we never noticed? It's a weird change

kvrhdn (approved):
Nice work! Makes Tempo a bit simpler to use again. Also good work tidying up the examples.

cmd/tempo/app/app.go
joe-elliott: yup, if you had search enabled then it raised the "mutlitenancy enabled" metric

PR #2008, opened 2023-01-20
Add Throughput and SLO Metrics with SLOConfig in Query Frontend

electron0zero:
**What this PR does**:
Expose Throughput (for search only) and SLO Metrics, and add config in Query Frontend to configure SLOs

Note: this doesn't work with TraceQL yet because we are not sending SearchMetrics from TraceQL query execution, it will work after we start returning SearchMetrics from TraceQL query execution

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/frontend/config.go
+++ b/modules/frontend/config.go
@@ -42,14 +42,14 @@ type HedgingConfig struct {
 }
 
 type SLOConfig struct {
-	DurationSLO   time.Duration `yaml:"duration_slo,omitempty"`
-	ThroughputSLO float64       `yaml:"throughput_slo,omitempty"`
+	DurationSLO        time.Duration `yaml:"duration_slo,omitempty"`
+	ThroughputBytesSLO float64       `yaml:"throughput_bytes_slo,omitempty"`
 }
 
 func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
 	slo := SLOConfig{
-		DurationSLO:   0,
-		ThroughputSLO: 0,
+		DurationSLO:        0,
+		ThroughputBytesSLO: 0,
 	}
 
 	cfg.Config.MaxOutstandingPerTenant = 2000
--- a/modules/frontend/config.go
+++ b/modules/frontend/config.go
@@ -48,8 +48,8 @@ type SLOConfig struct {
 
 func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
 	slo := SLOConfig{
-		DurationSLO:   5 * time.Second,
-		ThroughputSLO: 100 * 1024 * 1024, // 100 MB
+		DurationSLO:   0,
+		ThroughputSLO: 0,
 	}
 
 	cfg.Config.MaxOutstandingPerTenant = 2000
--- a/modules/frontend/config.go
+++ b/modules/frontend/config.go
@@ -2,9 +2,8 @@ package frontend
 
 import (
 	"flag"
-	"time"
-
 	"net/http"
+	"time"
 
 	"github.com/go-kit/log"
 	"github.com/prometheus/client_golang/prometheus"
@@ -28,11 +27,13 @@ type Config struct {
 
 type SearchConfig struct {
 	Sharder SearchSharderConfig `yaml:",inline"`
+	SLO     SLOConfig           `yaml:",inline"`
 }
 
 type TraceByIDConfig struct {
 	QueryShards int           `yaml:"query_shards,omitempty"`
 	Hedging     HedgingConfig `yaml:",inline"`
+	SLO         SLOConfig     `yaml:",inline"`
 }
 
 type HedgingConfig struct {
@@ -40,7 +41,17 @@ type HedgingConfig struct {
 	HedgeRequestsUpTo int           `yaml:"hedge_requests_up_to"`
 }
 
+type SLOConfig struct {
+	DurationSLO   time.Duration `yaml:"duration_slo,omitempty"`
+	ThroughputSLO float64       `yaml:"throughput_slo,omitempty"`
+}
+
 func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
+	slo := SLOConfig{
+		DurationSLO:   5 * time.Second,
+		ThroughputSLO: 100 * 1024 * 1024, // 100 MB
+	}
+
 	cfg.Config.MaxOutstandingPerTenant = 2000
 	cfg.MaxRetries = 2
 	cfg.TolerateFailedBlocks = 0
@@ -54,9 +65,11 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
 			ConcurrentRequests:    defaultConcurrentRequests,
 			TargetBytesPerRequest: defaultTargetBytesPerRequest,
 		},
+		SLO: slo,
 	}
 	cfg.TraceByID = TraceByIDConfig{
 		QueryShards: 50,
+		SLO:         slo,
 		Hedging: HedgingConfig{
 			HedgeRequestsAt:   2 * time.Second,
 			HedgeRequestsUpTo: 2,
--- a/modules/frontend/frontend.go
+++ b/modules/frontend/frontend.go
@@ -79,6 +79,7 @@ func New(cfg Config, next http.RoundTripper, o *overrides.Overrides, store stora
 
 	traces := traceByIDMiddleware.Wrap(next)
 	search := searchMiddleware.Wrap(next)
+	// TODO: pass the prom metric here??
 	return &QueryFrontend{
 		TraceByID:        newHandler(traces, traceByIDCounter, logger),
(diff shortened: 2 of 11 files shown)


docs/sources/configuration/_index.md
```diff
@@ -389,6 +389,21 @@ query_frontend:
         # (default: 30m)
         [query_ingesters_until: <duration>]
 
+        # If set to a non-zero value, it's value will be used to decide if query is within SLO or not
+        # tempo_query_frontend_search_queries_within_slo_total is incremented if query 
+        # completes in duration less than value of duration_slo
+        # This works in boolean OR with throughput_slo config,
+        # Query is within SLO if it returned within duration_slo seconds OR processed throughput_slo bytes/s data
```
knylander-grafana, line 396:
```suggestion
        # If set to a non-zero value, it's value will be used to decide if query is within SLO or not.
        # tempo_query_frontend_search_queries_within_slo_total is incremented if query 
        # completes in duration less than value of duration_slo.
        # This works in boolean OR with throughput_slo config,
        # Query is within SLO if it returned within duration_slo seconds OR processed throughput_slo bytes/s data.
```

docs/sources/configuration/_index.md
```diff
@@ -389,6 +389,21 @@ query_frontend:
...
+        [duration_slo: <duration> | default = 5s ]
+
+        # If set to a non-zero value, it's value will be used to decide if query is within SLO or not
+        # tempo_query_frontend_search_queries_within_slo_total is incremented if query 
+        # throughput (data processed per second) more than value of throughput_slo
+        # This works in boolean OR with throughput_slo config,
+        # Query is within SLO if it returned within duration_slo seconds OR processed throughput_slo bytes/s data
```
knylander-grafana, line 403:
```suggestion
        # If set to a non-zero value, it's value will be used to decide if query is within SLO or not.
        # tempo_query_frontend_search_queries_within_slo_total is incremented if query 
        # throughput (data processed per second) more than value of throughput_slo.
        # This works in boolean OR with throughput_slo config,
        # Query is within SLO if it returned within duration_slo seconds OR processed throughput_slo bytes/s data.
```

docs/sources/configuration/_index.md
```diff
@@ -402,6 +417,12 @@ query_frontend:
         # The maximum number of requests to execute when hedging.
         # Requires hedge_requests_at to be set. Must be greater than 0.
         [hedge_requests_up_to: <int> | default = 2 ]
+
+        # If set to a non-zero value, it's value will be used to decide if query is within SLO or not
+        # tempo_query_frontend_tracebyid_queries_within_slo_total is incremented if query 
+        # completes in duration less than value of duration_slo
+        # Query is within SLO if it returned within duration_slo seconds
```
knylander-grafana, line 424:
```suggestion
        # If set to a non-zero value, it's value will be used to decide if query is within SLO or not.
        # tempo_query_frontend_tracebyid_queries_within_slo_total is incremented if query 
        # completes in duration less than value of duration_slo.
        # Query is within SLO if it returned within duration_slo seconds.
```

knylander-grafana (commented):
Thank you for adding documentation for these changes. My only comment was to add some missing periods.

modules/frontend/searchsharding.go
```diff
@@ -276,6 +299,17 @@ func (s searchSharder) RoundTrip(r *http.Request) (*http.Response, error) {
...
 }
 
+// recordSearchMetrics records throughput and SLO metrics
+func (s *searchSharder) recordMetrics(tenantID string, status int, throughput float64, reqTime time.Duration) {
+	sts := strconv.Itoa(status)
+	searchThroughput.WithLabelValues(tenantID, sts).Observe(throughput)
+
+	// query is within SLO if query returned within DurationSLO seconds OR processed ThroughputSLO bytes/s data
+	if reqTime < s.sloCfg.DurationSLO || throughput > s.sloCfg.ThroughputSLO {
+		sloSearchPerTenant.WithLabelValues(tenantID, sts).Inc()
```
joe-elliott, line 309: i find this metric a little confusing. the name is `query_frontend_search_queries_within_slo_total` but
- if the status >= 500 then the that is not a succesful query
- this metric is used for both search and find trace by id

maybe there's a more accurate name? i'd also consider removing status and adding the endpoint hit.

modules/frontend/tracebyidsharding.go
```diff
@@ -191,6 +221,15 @@ func (s shardQuery) RoundTrip(r *http.Request) (*http.Response, error) {
 	}, nil
 }
 
+// recordTraceByIDMetrics records throughput and SLO metrics
+func (s shardQuery) recordMetrics(tenantID string, status int, reqTime time.Duration) {
+	// we are within SLO if query returned within DurationSLO seconds
+	// TODO: we don't have throughput metrics for TraceByID yet
+	if reqTime < s.sloCfg.DurationSLO {
+		sloTraceByIDPerTenant.WithLabelValues(tenantID, strconv.Itoa(status)).Inc()
```
joe-elliott, line 229: oh, i see. you have two different metrics here. i was confused by the similar structures. my personal .02 is that this SLO metric should mirror this metric:

https://github.com/grafana/tempo/blob/main/modules/frontend/frontend.go#L61

it should be named similarly so they are discoverable together and have basically the same labels so it's to combine them in promql math queries.

joe-elliott (commented):
overall nice work! handful of thoughts that i'm looking for your opinion on.

modules/frontend/searchsharding.go
electron0zero: only recording this in case of 200, and removed status from the metric, metric is renamed and added op label for search query types.

modules/frontend/frontend.go
```diff
@@ -34,7 +34,6 @@ const (
 type QueryFrontend struct {
 	TraceByID, Search http.Handler
 	logger            log.Logger
-	queriesPerTenant  *prometheus.CounterVec
```
electron0zero, line 37: this was not used, removed it

modules/frontend/config.go
```diff
@@ -28,19 +27,31 @@ type Config struct {
 
 type SearchConfig struct {
 	Sharder SearchSharderConfig `yaml:",inline"`
+	SLO     SLOConfig           `yaml:",inline"`
 }
 
 type TraceByIDConfig struct {
 	QueryShards int           `yaml:"query_shards,omitempty"`
 	Hedging     HedgingConfig `yaml:",inline"`
+	SLO         SLOConfig     `yaml:",inline"`
```
mdisibio, line 36: It's possible to configure `throughput_slo` for traces by ID, but it's unused.  Should we make the slo logic consistent between endpoints by moving this block to a shared function?  Also thinking to not require both settings, allow either or both to be set.
```go
// only record metric when it's enabled and within slo
 	if s.sloCfg.DurationSLO != 0 && s.sloCfg.ThroughputSLO != 0 {
 		if reqTime < s.sloCfg.DurationSLO || throughput > s.sloCfg.ThroughputSLO {
 			// query is within SLO if query returned 200 within DurationSLO seconds OR
 			// processed ThroughputSLO bytes/s data
 			sloSearchCounter.WithLabelValues(tenantID).Inc()
 		}
 	}
```

docs/sources/configuration/_index.md
```diff
@@ -389,6 +389,15 @@ query_frontend:
         # (default: 30m)
         [query_ingesters_until: <duration>]
 
+        # If set to a non-zero value, it's value will be used to decide if query is within SLO or not.
+        # Query is within SLO if it returned 200 within duration_slo seconds OR processed throughput_slo bytes/s data.
+        [duration_slo: <duration> | default = 0s ]
```
mdisibio, line 394: Let's add a note that both must be configured to work.

docs/sources/configuration/_index.md
```diff
@@ -389,6 +389,15 @@ query_frontend:
...
+
+        # If set to a non-zero value, it's value will be used to decide if query is within SLO or not.
+        # Query is within SLO if it returned 200 within duration_slo seconds OR processed throughput_slo bytes/s data.
+        [throughput_slo: <float> | default = 0 ]
```
mdisibio, line 398: Rename to `throughput_bytes_slo`

modules/frontend/searchsharding.go
```diff
@@ -30,12 +31,23 @@ const (
 )
 
 var (
-	metricInspectedBytes = promauto.NewHistogram(prometheus.HistogramOpts{
+	queryThroughput = promauto.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace: "tempo",
-		Name:      "query_frontend_result_metrics_inspected_bytes",
-		Help:      "Inspected Bytes in a search query",
+		Name:      "query_frontend_bytes_processed_per_seconds",
```
joe-elliott, line 36:
```suggestion
		Name:      "query_frontend_bytes_processed_per_second",
```

joe-elliott (commented):
two small changes, but looks great!

joe-elliott (approved):
nice work!

PR #2011, opened 2023-01-23
Parquet improvements

mdisibio:
**What this PR does**:
This PR contains a variety of stability and performance improvements, which came about after experiencing contention/performance regressions internally.

(1) Fixed ingester read/write contention by removing block locks in searchWAL, and moving the lock into the parquet wal block around the minimal area (the slice of flushes).  I.e. the ingester assumes/requires all implementations of `common.WALBlock` to be safe for concurrent read/write and now vParquet is.
(2) Fix iterator error handling and exit correctly after being closed or on a context error (eg: deadline exceeded)
(3) Fix performance of LeftJoinIterator by seeking correctly
(4) Fix performance of ReportValuesPredicate in that it didn't skip subsequent pages after evaluating the first dictionary.
(5) Restore original behavior (and snappy performance) of special attributes in v1 tag lookup, e.g. looking up `service.name` translates to v2 lookup `resource.service.name`.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/encoding/vparquet/wal_block.go
```diff
@@ -251,6 +252,19 @@ type walBlock struct {
 	// Flushed data
 	flushed     []*walBlockFlush
 	flushedSize int64
+	mtx         sync.Mutex
+}
+
+func (b *walBlock) readFlushes() []*walBlockFlush {
+	b.mtx.Lock()
+	defer b.mtx.Unlock()
+	return b.flushed
```
joe-elliott, line 261: does this need to create a new slice and copy the values over?

joe-elliott (approved):
a lot of subtle changes in here. I'm fine with moving forward so we can cut an rc.0. if there are issues we will flush them out as we go.

one question!

tempodb/encoding/vparquet/wal_block.go
mdisibio: In this case it is ok.  There is an implicit copy of the slice struct (pointer/len), and because `b.flushed` is only appended to it will remain valid.

Issue #2017, opened 2023-01-24
[TraceQL] TraceQL queries with nil return successfully


joe-elliott:
**Describe the bug**

This query returns successfully

```
{ resource.blerg = nil }
```

but doesn't actually evaluate this condition. Presumably it would return traces that have a resource without the `blerg` attribute.

joe-elliott:
Fixed in #2019

Issue #2016, opened 2023-01-24
[TraceQL] Parent intrinsic returns 200


joe-elliott:
**Describe the bug**

The `parent` intrinsic is not yet implemented but still returns 200:
```
{ parent.resource.cluster = "prod" }
```
This should fail to parse and return 400, but instead spins for awhile and then returns no results.

joe-elliott:
Fixed in #2019

Issue #2015, opened 2023-01-24
[TraceQL] Arithmetic Operators panic ingesters


joe-elliott:
**Describe the bug**
Executing a query with arithmetic operators such as :
`{ span.http.status_code > ( 200 + 1 ) }`

panics ingesters with:

```
panic: operator + is not yet implemented

goroutine 104720405 [running]:
github.com/grafana/tempo/pkg/traceql.BinaryOperation.execute({0x1, {0x276fed0, 0xc07a5d1f00}, {0x276fed0, 0xc07a5d1f40}}, {{0xc0c63469e0, 0x8, 0x8}, 0x173d59baf87037a1, 0x173d59bba9c21e69, ...})
	/Users/marty/src/tempo/pkg/traceql/ast_execute.go:193 +0x122d
github.com/grafana/tempo/pkg/traceql.BinaryOperation.execute({0xa, {0x276fe30, 0xc066427320}, {0x276fe80, 0xc066427350}}, {{0xc0c63469e0, 0x8, 0x8}, 0x173d59baf87037a1, 0x173d59bba9c21e69, ...})
	/Users/marty/src/tempo/pkg/traceql/ast_execute.go:143 +0x207
github.com/grafana/tempo/pkg/traceql.SpansetFilter.evaluate({{0x276fe80?, 0xc066427380?}}, {0xc06afb0160?, 0x1, 0xb0?})
	/Users/marty/src/tempo/pkg/traceql/ast.go:269 +0x38a
github.com/grafana/tempo/pkg/traceql.Pipeline.evaluate({{0xc0643e1a90?, 0x2762228?, 0xc0664272f0?}}, {0xc06afb0160?, 0x173d59e03812e000?, 0xc0664236d0?})
	/Users/marty/src/tempo/pkg/traceql/ast.go:91 +0x82
github.com/grafana/tempo/pkg/traceql.(*Engine).Execute(0x2762180?, {0x2762228?, 0xc0664272c0?}, 0xc066426fc0, {0x274eee0, 0xc0c9c08360})
	/Users/marty/src/tempo/pkg/traceql/engine.go:65 +0x50c
github.com/grafana/tempo/modules/ingester.(*instance).searchWAL.func2({0x27781d0, 0xc077a6e580})
	/Users/marty/src/tempo/modules/ingester/instance_search.go:179 +0x3a7
created by github.com/grafana/tempo/modules/ingester.(*instance).searchWAL
[...]

PR #2019, opened 2023-01-25
Add unsupported error to traceql validation

### LINKED ISSUE #1949: TraceQL: `by()` returns successfully
Even though `by()` is not supported it returns 200:

image

Same with `coalesce()`:

image

joe-elliott:
**What this PR does**:
Correctly fails validation for a wide array of TraceQL queries. 
- removes all panics from the execute/validate step for safety. 
- adds a call to "validate" in the engine parseQuery function.
- validates traceql in the query frontend

Fixes #1949, #2015, #2016, #2017 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/ast_validate_test.go
+++ b/pkg/traceql/ast_validate_test.go
@@ -18,6 +18,8 @@ type TestQueries struct {
 	Dump          []string `yaml:"dump"`
 }
 
+// jpe - confirm ast_execute never panics
+
 func TestExamples(t *testing.T) {
 	b, err := os.ReadFile(testExamplesFile)
 	require.NoError(t, err)
@@ -48,7 +50,7 @@ func TestExamples(t *testing.T) {
 			require.NoError(t, err)
 			err = p.validate()
 			require.Error(t, err)
-			require.False(t, errors.As(err, &unsupportedError{})) // jpe confirm this works
+			require.False(t, errors.As(err, &unsupportedError{}))
 		})
 	}
 
@@ -58,7 +60,7 @@ func TestExamples(t *testing.T) {
 			require.NoError(t, err)
 			err = p.validate()
 			require.Error(t, err)
-			require.True(t, errors.As(err, &unsupportedError{})) // jpe confirm this works
+			require.True(t, errors.As(err, &unsupportedError{}))
 		})
 	}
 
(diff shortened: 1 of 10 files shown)


pkg/traceql/ast_validate_test.go
```diff
@@ -46,6 +48,17 @@ func TestExamples(t *testing.T) {
...
+			require.False(t, errors.As(err, &unsupportedError{}))
+		})
+	}
+
+	for _, q := range queries.Unsupported {
+		t.Run("unsupported - "+q, func(t *testing.T) {
+			p, err := Parse(q)
+			require.NoError(t, err)
+			err = p.validate()
+			require.Error(t, err)
```
zalegrala, line 60: Do we want to make assertions about the type of error being returned here?

zalegrala (approved):
Looks pretty good to me.  Approving, but I've not been in this code, so perhaps a second pair of eyes is warranted.

pkg/traceql/ast_validate_test.go
joe-elliott: type is asserted on the next line. i suppose we technically don't need this line

zalegrala: Oh indeed yes.  I was thinking require.ErrorAs iirc, but that works too.

pkg/traceql/engine.go
```diff
@@ -205,7 +209,11 @@ func (s Static) asAnyValue() *common_v1.AnyValue {
 				StringValue: "nil",
 			},
 		}
-	default:
-		panic(fmt.Errorf("static has unexpected type %v", s.Type))
+	}
+
+	return &common_v1.AnyValue{
+		Value: &common_v1.AnyValue_StringValue{
+			StringValue: fmt.Sprintf("error formatting val: static has unexpected type %v", s.Type),
```
mdisibio, line 216: I think this will make the error message show up in the results table, which is probably ok.

mdisibio (approved):

SimenB:
(drive by comment, but merging this PR only closed one of the linked issue - they all need their own `Fixes` for GH to auto close on merge)

joe-elliott:
Thanks @SimenB !

PR #2023, opened 2023-01-27
Bubble up ingesters search errors

### LINKED ISSUE #2002: Ingesters don't propagate search errors correctly
**Describe the bug**
There are a few places where ingesters don't correctly propagate error messages. Here's an example:

https://github.com/grafana/tempo/blob/972f630afc655433c384e8362d065d9f3e72b30e/modules/ingester/instance_search.go#L160

This error needs to be correctly passed up the chain instead of swallowed here. Be aware that these blocks can return `common.Unsupported` as an error that _should_ be ignored.

electron0zero:
**What this PR does**:
Collects and combines errors from Ingester search

Fixes #2002 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/ingester/instance_search.go
```diff
@@ -55,9 +55,17 @@ func (i *instance) Search(ctx context.Context, req *tempopb.SearchRequest) (*tem
...
 	sr.AllWorkersStarted()
 
+	// read and combine search results
 	resultsMap := map[string]*tempopb.TraceSearchMetadata{}
 
+	// collect results from all the goroutines via sr.Results channel.
+	// range loop will exit when sr.Results channel is closed.
 	for result := range sr.Results() {
+		// exit early and Propagate error upstream
+		if sr.Error() != nil {
```
joe-elliott, line 65: if a query returns no results and hits an error will this line be executed? Do we need a check after the for loop to see if `sr.Error()` equals anything?

modules/ingester/instance_search.go
```diff
@@ -158,6 +167,7 @@ func (i *instance) searchWAL(ctx context.Context, req *tempopb.SearchRequest, p
 		err := e.b.Search(ctx, p, sr)
 		if err != nil {
 			level.Error(log.Logger).Log("msg", "error searching wal block", "blockID", e.b.BlockID().String(), "err", err)
```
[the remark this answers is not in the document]
electron0zero: I am leaning towards leaving it in, I want to be verbose when things are failing.

modules/ingester/instance_search.go
electron0zero: done

joe-elliott (approved):

PR #2025, opened 2023-01-27
Fix example call to search_tag_values_v2

09jvilla:
I think there should be a v2 in here too but perhaps I am misunderstanding.


joe-elliott (approved):
thx jen :)

Issue #2044, opened 2023-02-01
Add support for custom internal server


rubenvp8510:
**Is your feature request related to a problem? Please describe.**
. The key purpose for this one is to serve the Tempo HTTP API via mTLS while serving the readiness endpoints without (i.e. kubernetes readiness/liveness probes do not support authentication)

**Describe the solution you'd like**
adds support for a custom internal server listener dedicated for the start to serve the ready handler on a separate port and optionally TLS configuration. 

Similat to what was done on https://github.com/grafana/loki/pull/7069

PR #2045, opened 2023-02-01
add support for custom internal server

rubenvp8510:
Signed-off-by: Ruben Vargas <ruben.vp8510@gmail.com>

**What this PR does**:
The following PR adds support for a custom internal server listener dedicated for the start to serve the ready handler on a separate port and optionally TLS configuration. The key purpose for this one is to serve the Tempo HTTP API via mTLS while serving the readiness endpoints without (i.e. kubernetes readiness/liveness probes do not support authentication)

Fixes https://github.com/grafana/tempo/issues/2044

This is based on https://github.com/grafana/loki/pull/7069
**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo/app/modules.go
+++ b/cmd/tempo/app/modules.go
@@ -102,6 +102,11 @@ func (t *App) initServer() (services.Service, error) {
 }
 
 func (t *App) initInternalServer() (services.Service, error) {
+
+	if !t.cfg.InternalServer.Enable {
+		return services.NewIdleService(nil, nil), nil
+	}
+
 	DisableSignalHandling(&t.cfg.InternalServer.Config)
 	serv, err := server.New(t.cfg.InternalServer.Config)
 	if err != nil {
@@ -111,8 +116,8 @@ func (t *App) initInternalServer() (services.Service, error) {
 	servicesToWaitFor := func() []services.Service {
 		svs := []services.Service(nil)
 		for m, s := range t.serviceMap {
-			// Server should not wait for itself.
-			if m != InternalServer {
+			// Server should not wait for itself or the server
+			if m != InternalServer && m != Server {
 				svs = append(svs, s)
 			}
 		}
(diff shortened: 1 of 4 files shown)


cmd/tempo/app/modules.go
```diff
@@ -413,13 +441,34 @@ func (t *App) setupModuleManager() error {
 		UsageReport:          {MemberlistKV},
 	}
 
+	if t.cfg.InternalServer.Enable {
```
joe-elliott, line 444: what is this code for?

cmd/tempo/app/modules.go
```diff
@@ -90,6 +91,30 @@ func (t *App) initServer() (services.Service, error) {
...
+	serv, err := server.New(t.cfg.InternalServer.Config)
+	if err != nil {
+		return nil, err
+	}
+
+	servicesToWaitFor := func() []services.Service {
+		svs := []services.Service(nil)
+		for m, s := range t.serviceMap {
+			// Server should not wait for itself.
+			if m != InternalServer {
```
joe-elliott, line 105: as written will the internal server wait on the server and vice versa? do we want that behavior?

joe-elliott (commented):
Thanks for this contribution. I definitely like this feature addition, but think we need a little cleanup.

I've added a few questions to get started, but we may also want to refactor to reduce duplicate code between Server and InternalServer.

Let me know what you think

cmd/tempo/app/modules.go
rubenvp8510: This adds the internal server as a dependency of the other services, only it it is enabled.

joe-elliott: this is kind of tough to parse. can we instead :
- make the Server always dependent on InternalServer
- in initInternalServer make the choice to start it or not
  - if it's disabled you can just return a dummy service like this: https://github.com/grafana/tempo/blob/main/cmd/tempo/app/modules.go#L181

cmd/tempo/app/modules.go
```diff
@@ -390,6 +420,9 @@ func (t *App) setupModuleManager() error {
 	mm := modules.NewManager(log.Logger)
 
 	mm.RegisterModule(Server, t.initServer, modules.UserInvisibleModule)
+	if t.cfg.InternalServer.Enable {
```
joe-elliott, line 423: This if statement needs to be removed. Even if the internal server is disabled we still need to register the module b/c of the changes just made.

joe-elliott (approved):
nice addition. thx!

PR #2052, opened 2023-02-01
Fix search tag endpoint return code (#2021)

### LINKED ISSUE #2021: Search Tags Endpoint should return 400 but returns 500
**Describe the bug**
The search tags endpoint returns 500 in cases where it should return 400. This occurs when you curl the endpoint with an invalid tag name such as `tagname`:

```
$ curl -v  http://localhost:3100/api/v2/search/tag/tagname/values 
...
< HTTP/1.1 500 Internal Server Error
< Content-Type: text/plain; charset=utf-8
< Vary: Accept-Encoding
< X-Content-Type-Options: nosniff
< Date: Thu, 26 Jan 2023 16:40:35 GMT
< Content-Length: 229
< 
error querying ingesters in Querier.SearchTagValues: failed to get response from ingesters: failed to execute f() for 10.138.213.73:9095: rpc error: code = Unknown desc = tag name is not valid intrinsic or scoped attribute: asdf
* Connection #0 to host localhost left intact
```

The reason this fails is b/c this tagname needs a scope. Either: `.tagname`, `resource.tagname` or `span.tagname`. 

```
$ curl -v http://localhost:3100/tempo/api/v2/search/tag/.tagname/values 
...
< HTTP/1.1 200 OK
< Content-Type: application/json
< Vary: Accept-Encoding
< Date: Thu, 26 Jan 2023 16:43:40 GMT
< Content-Length: 2
< 
* Connection #0 to host localhost left intact
{}
```

thmshmm:
**What this PR does**: Returns http status code 400 for failed tag search requests

Fixes #2021

### DIFF
--- a/modules/querier/http.go
+++ b/modules/querier/http.go
@@ -11,9 +11,11 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/opentracing/opentracing-go"
 	ot_log "github.com/opentracing/opentracing-go/log"
+	"github.com/pkg/errors"
 
 	"github.com/grafana/tempo/pkg/api"
 	"github.com/grafana/tempo/pkg/tempopb"
+	"github.com/grafana/tempo/pkg/traceql"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


joe-elliott:
Thanks for taking a look at this. I would like to be more selective about what we 400 and what we 500. The issue is that when this method fails in the ingester:

https://github.com/grafana/tempo/blob/main/modules/ingester/instance_search.go#L469

we should really return a 400 b/c the user specified an invalid tag name. For this endpoint likely all other errors should be 500s.

In the other HTTP endpoints we do some validation up front like this:

https://github.com/grafana/tempo/blob/main/modules/querier/http.go#L38

Hopefully we can find similar logic to execute in this method to return 400 before we even attempt the search:

https://github.com/grafana/tempo/blob/main/modules/querier/http.go#L179

joe-elliott (approved):
Thanks for the change! This is definitely more along the lines of what I was thinking of.

PR #2053, opened 2023-02-01
Update helm microservices example for Kubernetes v1.22 deprecations

zalegrala:
**What this PR does**:

Using later versions of k3d bundles newer kubernetes, which includes some deprecations.  This change updates the mentioned resources to allow the `kubectl create` call to succeed in the `microservices` example

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/example/helm/microservices-extras.yaml
+++ b/example/helm/microservices-extras.yaml
@@ -1,3 +1,21 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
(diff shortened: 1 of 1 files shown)


zalegrala:
Only tested for local local apply against k3d.

mapno (approved):
LGTM

PR #2062, opened 2023-02-03
span id to hex

### LINKED ISSUE #2060: Stop trimming leading 0s on span ids
**Describe the bug**
After some discussion with Grafana on this issue:
https://github.com/grafana/grafana/issues/62725

We decided that the best fix is for Tempo to start returning span ids with leading 0s.

ie-pham:
**What this PR does**:

Fixes #2060

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/util/traceid_test.go
```diff
@@ -89,6 +89,34 @@ func TestTraceIDToHexString(t *testing.T) {
 	}
 }
 
+func TestSpanIDToHexString(t *testing.T) {
```
joe-elliott, line 92: thx for adding tests :pray:

pkg/util/traceid.go
```diff
@@ -51,6 +51,14 @@ func TraceIDToHexString(byteID []byte) string {
 	return id
 }
 
+// SpanIDToHexString converts a span ID to its string representation and WITHOUT removing any leading zeros.
+// Assuming everything passed in here will be 64 bit or less
+func SpanIDToHexString(byteID []byte) string {
+	id := hex.EncodeToString(byteID)
+	return id[len(id)-16:]
```
joe-elliott, line 58: so even though a span should never be less than 16 characters this function should not assume that. it should protect itself from invalid input. 

as written with a little bit of bad luck or a bad assumption somewhere else this would panic tempo components.

pkg/util/traceid.go
```diff
@@ -51,6 +51,28 @@ func TraceIDToHexString(byteID []byte) string {
 	return id
 }
 
+// SpanIDToHexString converts a span ID to its string representation and WITHOUT removing any leading zeros.
+// Assuming everything passed in here will be 64 bit or less
+func SpanIDToHexString(byteID []byte) string {
+	id := hex.EncodeToString(byteID)
+	length := len(id)
```
joe-elliott, line 58: lines 58-72 can be written fairly succinctly using a fmt.Sprintf and a format modifier like `"%016s"`. this will likely be a trade off between simplicity and performance. wdyt?

joe-elliott: lines 59-64 are still a mystery to me. if the length is > 16 and starts with a 0 then take the 16th right most characters in the string?

i think just the `fmt.Sprintf` should handle what we want w/o any of this extra logic.

ie-pham: Yes. I just realized that as well.

pkg/util/traceid.go
```diff
@@ -51,6 +51,14 @@ func TraceIDToHexString(byteID []byte) string {
 	return id
 }
 
+// SpanIDToHexString converts a span ID to its string representation and WITHOUT removing any leading zeros.
+// If the id is < 16, left pad with 0s
+func SpanIDToHexString(byteID []byte) string {
+	id := hex.EncodeToString(byteID)
+	id = strings.TrimLeft(id, "0")
```
joe-elliott, line 58: is this needed?

ie-pham: You're right. I was trying to solve for an invalid test case I added.

ie-pham: Yes. Without trimming it, padding it 
```
0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x12
```
would return
```
00000000000000000000000000000012
```

ie-pham:
One last question, extra logic would have to be added to this I believe but what happens if the span id has leading zeroes and it is longer than 16 characters? Still no trimming?

joe-elliott:
> @joe-elliott One last question, extra logic would have to be added to this I believe but what happens if the span id has leading zeroes and it is longer than 16 characters? Still no trimming?

since we're in "shouldn't happen" territory i'm fine with anything that:
- doesn't panic
- makes decent sense
- doesn't overcomplicate the code

joe-elliott (approved):
looks good!

PR #2068, opened 2023-02-03
[generator] Insert leading zero value for new counter series

### LINKED ISSUE #2006: Initialise stale counter series at 0
The metrics-generator has a mechanism to remove series that have become stale—ie. have not received a new value in some time. The main objectives are keeping resources down and make room for more frequent series.

This is a problem for series that are removed frequently, because `traces_spanmetrics_calls_total` gets initialised at the value of the first scrape once the series gets registered, and Prometheus skips the jump from `null` to a given value. In order not to lose that first value, we need to push a first value of `0` when a new series is added to the counter.

zalegrala:
**What this PR does**:

Here we allow new values on counter series to be used by Prometehus by inserting a leading 0 into new counter series.

Fixes #2006 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/registry/counter.go
+++ b/modules/generator/registry/counter.go
@@ -27,10 +27,34 @@ type counterSeries struct {
 	labelValues []string
 	value       *atomic.Float64
 	lastUpdated *atomic.Int64
+	// firstSeries is used to track if this series is new to the counter.  This
+	// is used to ensure that new counters being with 0, and then are incremented
+	// to the desired value.  This avoids Prometheus throwing away the first
+	// value in the series, due to the transition from null -> x.
+	firstSeriesMtx sync.RWMutex
+	firstSeries    bool
 }
 
 var _ Counter = (*counter)(nil)
 var _ metric = (*counter)(nil)
+
+func (co *counterSeries) isNew() bool {
+	co.firstSeriesMtx.RLock()
+	defer co.firstSeriesMtx.RUnlock()
+	return co.firstSeries
+}
+
+func (co *counterSeries) registerFirstSeries() {
+	co.firstSeriesMtx.Lock()
+	defer co.firstSeriesMtx.Unlock()
+	co.firstSeries = true
+}
+
+func (co *counterSeries) registerSeenSeries() {
+	co.firstSeriesMtx.Lock()
+	defer co.firstSeriesMtx.Unlock()
+	co.firstSeries = false
+}
 
 func newCounter(name string, labels []string, onAddSeries func(uint32) bool, onRemoveSeries func(count uint32)) *counter {
 	if onAddSeries == nil {
@@ -84,6 +108,7 @@ func (c *counter) Inc(labelValues *LabelValues, value float64) {
 		c.updateSeries(s, value)
 		return
 	}
+	newSeries.registerFirstSeries()
 	c.series[hash] = newSeries
 }
 
@@ -126,6 +151,17 @@ func (c *counter) collectMetrics(appender storage.Appender, timeMs int64, extern
 			lb.Set(name, s.labelValues[i])
 		}
 
+		// If we are about to call Append for the first time on a series, we need
+		// to first insert a 0 value to allow Prometheus to start from a non-null
+		// value.
+		if s.isNew() {
+			_, err = appender.Append(0, lb.Labels(nil), timeMs, 0)
+			if err != nil {
+				return
+			}
+		}
+
+		s.registerSeenSeries()
 		_, err = appender.Append(0, lb.Labels(nil), timeMs, s.value.Load())
 		if err != nil {
 			return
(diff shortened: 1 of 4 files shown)


modules/generator/registry/counter.go
```diff
@@ -126,6 +151,17 @@ func (c *counter) collectMetrics(appender storage.Appender, timeMs int64, extern
 			lb.Set(name, s.labelValues[i])
 		}
 
+		// If we are about to call Append for the first time on a series, we need
+		// to first insert a 0 value to allow Prometheus to start from a non-null
+		// value.
+		if s.isNew() {
+			_, err = appender.Append(0, lb.Labels(nil), timeMs, 0)
```
kvrhdn, line 158: I'm wondering if we need to add a tiny offset to timeMs here. If the 0 sample and the actual value have the same timestamp, Prometheus might reject the second sample because they would overwrite each other.

modules/generator/registry/counter.go
```diff
@@ -27,11 +27,35 @@ type counterSeries struct {
 	labelValues []string
 	value       *atomic.Float64
 	lastUpdated *atomic.Int64
+	// firstSeries is used to track if this series is new to the counter.  This
+	// is used to ensure that new counters being with 0, and then are incremented
+	// to the desired value.  This avoids Prometheus throwing away the first
+	// value in the series, due to the transition from null -> x.
+	firstSeriesMtx sync.RWMutex
+	firstSeries    bool
```
kvrhdn, line 35: Is there a specific reason to use `sync.RWMutex` over `atomic.Bool`? I'm not entirely sure which solution would be more performant, but if it's equal I'd say `atomic.Bool` is easier to work with.

An alternative approach would be to store a `created` timestamp. `isNew()` could then be implemented as `created == lastUpdated.load()`.
Since `created` would never be modified we don't need any concurrency handling for it.

modules/generator/registry/counter.go
zalegrala: I was also wondering this.  Perhaps just subtracting 100 from the `timeMs` here would be sufficient.

modules/generator/registry/counter.go
zalegrala: I'd considered the alternative you mentioned here, which seems a little cleaner than any locking.  I'll revisit this.

modules/generator/registry/counter.go
zalegrala: I added with adding a 1 second offset to the initial desired value and updated the tests a bit for this.

modules/generator/registry/counter.go
zalegrala: Seems a little cleaner after the update.  I think we still want to be able to turn off the isNew, and so kept the bool in place of the timestamp approach.

modules/generator/registry/counter.go
```diff
@@ -121,12 +137,26 @@ func (c *counter) collectMetrics(appender storage.Appender, timeMs int64, extern
...
+		if s.isNew() {
+			_, err = appender.Append(0, lb.Labels(nil), timeMs, 0)
+			if err != nil {
+				return
+			}
+			// Increment timeMs to ensure that the next value is not at the same time.
+			t = t.Add(insertOffsetDuration)
+		}
+
+		s.registerSeenSeries()
```
kvrhdn, line 158: Nit: if we move this within the if case we only execute this when we actually have to change the value. I can't imagine the overhead of this will be big but still, no need to keep writing false.

modules/generator/registry/counter.go
```diff
@@ -121,12 +137,26 @@ func (c *counter) collectMetrics(appender storage.Appender, timeMs int64, extern
...
+		// If we are about to call Append for the first time on a series, we need
+		// to first insert a 0 value to allow Prometheus to start from a non-null
+		// value.
+		if s.isNew() {
+			_, err = appender.Append(0, lb.Labels(nil), timeMs, 0)
+			if err != nil {
+				return
+			}
+			// Increment timeMs to ensure that the next value is not at the same time.
+			t = t.Add(insertOffsetDuration)
```
kvrhdn, line 155: I'm wondering if we could just do `timeMs - 1000`. Instead of delaying our sample, write a zero sample before it. This might trigger out-of-order writes though... I'm not sure if Prometheus checks for out-of-order writes within the whole batch or just for the serie.

kvrhdn (commented):
Looks good overall, just left some comments. Nothing blocking from me.

modules/generator/registry/counter.go
zalegrala: Good call, I'll move.

modules/generator/registry/counter.go
zalegrala: I wondered about this also, but avoided going negative offset because I wasn't sure about the interplay with Prometheus and was somewhat hard to test.  Do you think this will impact the usage having a delay on our sample?

kvrhdn: It adds latency between when the span arrived and when we report it has arrived, so the data becomes more inaccurate. But since we already aggregate data in the collection interval (which is typically 15s) I don't think this will be significant. So I'm fine with going for the safe options and just adding a 1s offset.

kvrhdn (approved):

PR #2070, opened 2023-02-03
Allow toggling of latency/count metrics for metrics-generator

### LINKED ISSUE #2013: [Metrics generator] Ability to toggle off latency metrics entirely

We have lots of instrumentation in our stack but like the metrics-generator for getting request rates for specific endpoints with added dimensionality. However, we don't have much interest in the latency metrics. But, because we have such high traffic and high metric flows, the latency metrics add an intense degree of cardinality that can tank our remote write instance. 

One can disable most of the buckets (it still leaves +Inf and 0 buckets) by setting the histograms to [null]. But I'd like to turn them off entirely to make administration of this tool in a high volume environment simpler.

This, for example, is not ideal when I don't care about latency metrics:

image

Turn off latency histograms but keep request rate data.

Setting buckets to [null] allows you to reduce the latency cardinality, but not remove it.

AlexDHoffer:
**What this PR does**:

Allows user to toggle off latency and count metrics in the metrics-generator. It does this by introducing the concept of a subprocessor. The two current subprocessors are ``span-metrics-count`` and ``span-metrics-latency``. These subprocessors are toggled on by default, i.e. when "span-metrics" is indicated in the overrides. A user can choose the ``span-metrics-count`` subprocessor by indicating it in the overrides, likewise with the ``-latency`` one. If a subprocessor that doesn't exist is indicated, it won't be parsed by subprocessor logic but instead will be flagged as a bad processor. Under the hood indicating the ``span-metrics`` processor alone enables both subprocessors.

Fixes #2013 

**Checklist**
- [ x ] Tests updated
- [ ] Documentation added
- [ x ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

Happy to update documentation if this particular approach for this problem is accepted by maintainers


modules/generator/processor/spanmetrics/config.go
```diff
@@ -27,6 +29,9 @@ type Config struct {
 	// Additional dimensions (labels) to be added to the metric. The dimensions are generated
 	// from span attributes and are created along with the intrinsic dimensions.
 	Dimensions []string `yaml:"dimensions"`
+	// Subprocessor options for this Processor include Latency, Counts
+	// These are metrics categories that exist under the umbrella of Span Metrics
+	Subprocessors map[string]bool
```
zalegrala, line 34: I'm wondering if this could be `map[SubProcessor]bool` to force us to be specific about the key space.

AlexDHoffer: @zalegrala Sure, I could make type ``Subprocessor`` with field ``Name``. If we wanted to enforce specificity about the keys e.g. Latency and Count we could make a constructor for the type that does a check to make sure the value matches one of the choices.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -58,14 +58,18 @@ func New(cfg Config, registry registry.Registry) gen.Processor {
...
 
-	return &Processor{
-		Cfg:                        cfg,
-		registry:                   registry,
-		spanMetricsCallsTotal:      registry.NewCounter(metricCallsTotal, labels),
-		spanMetricsDurationSeconds: registry.NewHistogram(metricDurationSeconds, labels, cfg.HistogramBuckets),
-		spanMetricsSizeTotal:       registry.NewCounter(metricSizeTotal, labels),
-		now:                        time.Now,
+	p := &Processor{}
+	if cfg.Subprocessors["Latency"] {
```
mapno, line 62: Can you move `"Latency"` and `"Count"` to constants? Even maybe using the `SubProcessor` type suggested by @zalegrala

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## main / unreleased
 
+* [ENHANCEMENT] Ability to toggle off latency or count metrics in metrics-generator [#2070](https://github.com/grafana/tempo/pull/2070) (@AlexDHoffer)
```
mapno, line 3: Why not include size as well? Either as part of `count` or its own type.

modules/generator/instance.go
```diff
@@ -124,13 +124,47 @@ func (i *instance) watchOverrides() {
...
 }
 
+// Look at the processors defined and see if any are actually span-metrics subprocessors
+// If they are, set the appropriate flags in the spanmetrics struct
+func (i *instance) updateSubprocessors(desiredProcessors map[string]struct{}, desiredCfg ProcessorConfig) (map[string]struct{}, ProcessorConfig) {
+	_, allOk := desiredProcessors["span-metrics"]
+	_, countOk := desiredProcessors["span-metrics-count"]
+	_, latencyOk := desiredProcessors["span-metrics-latency"]
+
+	if allOk {
```
mapno, line 134: What's the difference between `allOk` and `countOk && latencyOk`? Are they not equivalent?

mapno (commented):
I like the approach. Left a few comments.

modules/generator/processor/spanmetrics/spanmetrics.go
AlexDHoffer: @mapno I will push up something that has the Subprocessor type

CHANGELOG.md
AlexDHoffer: Mostly was interested in getting a copy up to verify the approach made sense to maintainers before I finished polishing

modules/generator/instance.go
AlexDHoffer: @mapno In the case of the overrides, you could have the values "span-metrics", "span-metrics-latency", or "span-metrics-count". In this conditional block, ``allOk`` is the case where the user indicated "span-metrics". ``countOk`` and ``latencyOk`` is the case where the user indicated "span-metrics-latency" and "span-metrics-count" separately.

AlexDHoffer: I changed the logic here a little bit. Since we know we want to collapse "span-metrics-latency" and "-count" into "span-metrics", we don't need the ``allOk``, and we don't need individual delete statements since we can delete them agnostic to the different conditions

modules/generator/processor/spanmetrics/config.go
zalegrala: Sure, a struct or an int iota sort of approach could probably work also.

AlexDHoffer:
@zalegrala @mapno ready for another look, I've switched to using a stable key space for subprocessors with the iota pattern and added support for the size metric

mapno (commented):
Do you think this PR could be simplified if it used booleans (eg. `export_count`, `export_latency` and `export_size`) instead of the subprocessors approach? The current approach leaks much of the logic to `instance.updateProcessors`, and I think it could be mostly kept to the processor instead.

AlexDHoffer:
> Do you think this PR could be simplified if it used booleans (eg. `export_count`, `export_latency` and `export_size`) instead of the subprocessors approach? The current approach leaks much of the logic to `instance.updateProcessors`, and I think it could be mostly kept to the processor instead.

I like it the way it is but I can change it to whatever your team wants as I would like the feature.

AlexDHoffer:
@mapno FYI in order to rectify your preferred approach with the strategy laid out by Joe in the issue itself: If the idea is to indicate processor in the ``metrics_generator_processor`` (e.g. ``span-metrics-latency``, ``span-metrics-count`` listed in the ``overrides``) we will no matter what have to have some logic in instance.go to determine which processors are specified so that we can boil them down into one ``span-metrics`` processor that contains the right ``export`` booleans. I figure we can add logic where the switch statement of ``addProcessor`` is that can determine which span metrics processors are indicated, then supply that result to the object creation ``New`` function of spanmetrics

mapno:
Apologies, this fell through the cracks and I didn't see a notification until https://github.com/grafana/tempo/issues/2013#issuecomment-1493474822. I'm sorry.

> @mapno FYI in order to rectify your preferred approach with the strategy laid out by Joe in the issue itself: If the idea is to indicate processor in the `metrics_generator_processor` (e.g. `span-metrics-latency`, `span-metrics-count` listed in the `overrides`) we will no matter what have to have some logic in instance.go to determine which processors are specified so that we can boil them down into one `span-metrics` processor that contains the right `export` booleans. Since each processor in that overrides field is added separately by `addProcessor` there is still complexity needed in `instance.go`. I don't think this can be done mostly in spanmetrics.go without doing something in the overrides code or adding an overrides field specific to these subprocessors

Don't you think we can move much of the logic in https://github.com/grafana/tempo/pull/2070/files#diff-2b2bf6091e8c6446ffe339e21fd2f917d4ba0c172a76a226153f53121181b202R127-R171 to the processor? The instance can just pass all the processors named `span-metrics.*`, and the processor handles which processors it wants running.

Anyway, I don't block this much longer after not replaying so long.

AlexDHoffer:
@mapno I've rebased

AlexDHoffer:
@mapno Sorry, is that last comment supposed to mean an approval is imminent? That's why I said I rebased 😆

mapno:
Sorry, I wasn't clear. I was hoping we could move much of the logic in https://github.com/grafana/tempo/pull/2070/files#diff-2b2bf6091e8c6446ffe339e21fd2f917d4ba0c172a76a226153f53121181b202R127-R171 to the processor. Spanmetrics can take the processors that are enabled and work out which subprocessors need to be running.

Nonetheless, I don't want to block this further if you're not convinced that approach is any better. Let me know and we can see to approving and merging.

AlexDHoffer:
@mapno As I said above, you will still need logic in ``instance.go`` so I don't see how you'd even do the other way. These "subprocessors" are considered normal processors, so "span-metrics-latency", "span-metrics-count" are received in ``instance.go`` and you can't create a separate processor for both, so ``instance.go`` will have to parse out which subprocessors are there in order to get one processor with the right configs out of it. So either way the logic is not totally contained in the processor, so we should just keep things this way

mapno:
I was away for a few days, I'm ok to merge. Can you solve the conflict? Thanks

mapno (approved):

AlexDHoffer:
@mapno Rebased, thanks

mapno:
Merging, thanks for the patience.

PR #2087, opened 2023-02-08
Capture and update search metrics for TraceQL

### LINKED ISSUE #2086: SearchMetrics is always empty for TraceQL queries
We are using InspectedBytes search metric to compute query Throughput in #2008, but InspectedBytes is not set for TraceQL queries

See the TODO at:
https://github.com/grafana/tempo/blob/f9000562d5606f1615af799375e89ce0ee246fd3/pkg/traceql/engine.go#L47

electron0zero:
**What this PR does**:

Fixes #2086 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/encoding/vparquet/wal_block.go
```diff
@@ -589,6 +594,10 @@ func (b *walBlock) Fetch(ctx context.Context, req traceql.FetchSpansRequest, opt
 		Results: &mergeSpansetIterator{
 			iters: iters,
 		},
+		Bytes: func() uint64 {
+			// size of walBlock, we assume flushed block size was read during query
+			return b.FlushedSize()
```
joe-elliott, line 599: metrics includes an entry called "total block bytes" which is different then "inspected bytes". i think we should report data retrieved from the backend for "inspected" and use this flushed size for "total block bytes".

electron0zero: I need InspectedBytes to calculate query throughput down the line where TraceQL is executed. 

I am using an existing `Bytes` callback to pass one metric up the chain to caller of the Fetch, and added a TODO to do it in a better way.

since I can only pass one metric back using this callback, I am passing flushed size as an approximation for InspectedBytes.

This approximation is only for walBlock and backend Block is passing it's `TotalBytesRead` metric on `BackendReaderAt`, which is also kinda not ideal but works for now 🤞🏼 

I am going to drop few TODOs for me to come back and find a better solution.

joe-elliott: Why don't we just do a similar pattern as exists for blocks opened in object storage? They use the `BackendReaderAt` shim that records how many bytes are read. We can just make a `walReaderAt` that implements `io.ReaderAt` and counts the bytes read on the files?

I'm really struggling with returning such different values for the backend blocks and the wal. I'd also prefer not to add the `FlushedSize()` method to the `WALBlock` interface.

If we use a readerat shim like the backend does we can avoid both of these issues.

electron0zero: added a reader in wal page, and using that to compute accurate info about bytes read.

tempodb/encoding/vparquet/wal_block.go
```diff
@@ -642,13 +645,17 @@ func (b *walBlock) Fetch(ctx context.Context, req traceql.FetchSpansRequest, opt
...
+		totalBytesRead += file.r.TotalBytesRead.Load()
 	}
 
 	// combine iters?
 	return traceql.FetchSpansResponse{
 		Results: &mergeSpansetIterator{
 			iters: iters,
 		},
+		Bytes: func() uint64 {
+			return totalBytesRead
```
mdisibio, line 657: I'm not sure this will work.  Fetch just opens the iters, so the value `totalBytesRead` just counts the parquet headers more or less.  Since `mergeSpansetIterator` is the top-level iter returned to the engine, it would need to sum up these values when each iter is finished (or engine exits early).

electron0zero: updated, now I am collecting readers from pages and reading TotalBytesRead when Bytes callback is called.

this means, that we will read metadata + data size post iteration, and it will produce accurate result.

tempodb/encoding/vparquet2/wal_block.go
```diff
@@ -642,13 +645,22 @@ func (b *walBlock) Fetch(ctx context.Context, req traceql.FetchSpansRequest, opt
...
 		iters = append(iters, wrappedIterator)
+		readers = append(readers, file.r)
 	}
 
 	// combine iters?
 	return traceql.FetchSpansResponse{
 		Results: &mergeSpansetIterator{
 			iters: iters,
 		},
+		Bytes: func() uint64 {
```
electron0zero, line 656: cc @stoewer because I am making same changes in vparquet2 format as well.

tempodb/encoding/vparquet2/block_search.go
```diff
@@ -229,7 +225,7 @@ func makePipelineWithRowGroups(ctx context.Context, req *tempopb.SearchRequest,
 			max = (time.Millisecond * time.Duration(req.MaxDurationMs)).Nanoseconds()
 		}
 		durFilter := pq.NewIntBetweenPredicate(min, max)
-		traceIters = append(traceIters, makeIter("DurationNano", durFilter, "Duration"))
+		traceIters = append(traceIters, makeIter("DurationNanos", durFilter, "Duration"))
```
mdisibio, line 228: What was the rationale for this change?  It looks like the column is really named [DurationNano](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet2/schema.go#L183) and this change will not work.

mdisibio (commented):
The wal metrics lgtm, but I would like @stoewer to review the vparquet2 DurationNanos change.

tempodb/encoding/vparquet2/block_search.go
electron0zero: oopps, copy paste error. reverting back.

electron0zero: > The wal metrics lgtm, but I would like @stoewer to review the vparquet2 DurationNanos change.

cc @stoewer

mdisibio (approved):

stoewer (commented):
The vParquet2 changes look good to me

I still added a couple of minor suggestions

tempodb/encoding/vparquet/readers.go
```diff
@@ -128,3 +128,22 @@ func (r *cachedReaderAt) ReadAt(p []byte, off int64) (int, error) {
...
+var _ io.ReaderAt = (*WalReaderAt)(nil)
+
+func NewWalReaderAt(r io.ReaderAt) *WalReaderAt {
+	return &WalReaderAt{r, atomic.Uint64{}}
+}
+
+func (wr *WalReaderAt) ReadAt(p []byte, off int64) (int, error) {
+	// parquet-go will call ReadAt when reading data from a parquet file.
+	wr.TotalBytesRead.Add(uint64(len(p)))
+	return wr.r.ReadAt(p, off)
```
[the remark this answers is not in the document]
electron0zero: TIL, change sounds good to me.

tempodb/encoding/vparquet/readers.go
```diff
@@ -128,3 +128,22 @@ func (r *cachedReaderAt) ReadAt(p []byte, off int64) (int, error) {
 
 	return r.r.ReadAt(p, off)
 }
+
+// WalReaderAt is wrapper over io.ReaderAt, and is used to measure the total bytes read when searching walBlock.
+type WalReaderAt struct {
```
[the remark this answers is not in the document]
electron0zero: cool, will make it private.

tempodb/encoding/vparquet/wal_block.go
```diff
@@ -642,13 +645,22 @@ func (b *walBlock) Fetch(ctx context.Context, req traceql.FetchSpansRequest, opt
...
 	// combine iters?
 	return traceql.FetchSpansResponse{
 		Results: &mergeSpansetIterator{
 			iters: iters,
 		},
+		Bytes: func() uint64 {
+			// read value when callback is called
+			var totalBytesRead uint64
+			for _, r := range readers {
+				totalBytesRead += r.TotalBytesRead.Load()
```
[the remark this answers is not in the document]
electron0zero: Added `BytesRead()` method on `walReaderAt` and `BackendReaderAt`

stoewer (approved):
LGTM

PR #2090, opened 2023-02-08
Clarify the specific intrinsic fields that currently exist

09jvilla:
**What this PR does**:
Update docs to specify that the intrinsics listed in the table is an exhaustive list of the intrinsics available today.


knylander-grafana (approved):

PR #2105, opened 2023-02-14
Log when a trace is too large to compact

### LINKED ISSUE #1931: Log more information when a trace is too large to compact
When a trace exceeds `max_bytes_per_trace` the compactor will drop spans.  Currently this is tracked with a metric but it would be nice to log the trace ID and/or spans that were dropped.  Need to decide on log level, a case could be made for warning, info, or debug.

This looks straightforward to do and adding [here](https://github.com/grafana/tempo/blob/main/modules/compactor/compactor.go#L240) and [here](https://github.com/grafana/tempo/blob/main/modules/compactor/compactor.go#L244) would cover both v2 and parquet formats (and should be future-proof for other formats).

scalalang2:
**What this PR does**:
Log when a trace is too large to compact.
If the total bytes of traces exceeds maximum limit, the following messages will be logged.

```json
{
    "msg": "max size of trace exceeded",
    "tenant": "XXXXX",
    "traceId": "XXXXX",
    "discarded_span_count": 100
}
```

There will be some dicussions in this PR.
Please refer to following comments.

Fixes #1931

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/compactor/compactor.go
+++ b/modules/compactor/compactor.go
@@ -10,6 +10,7 @@ import (
 	"github.com/grafana/dskit/kv"
 	"github.com/grafana/dskit/ring"
 	"github.com/grafana/dskit/services"
+	tempoUtil "github.com/grafana/tempo/pkg/util"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 
@@ -236,12 +237,13 @@ func (c *Compactor) Combine(dataEncoding string, tenantID string, objs ...[]byte
 		return objs[0], wasCombined, nil
 	}
 
-	spansDiscarded := countSpans(dataEncoding, objs[1:]...)
-	overrides.RecordDiscardedSpans(spansDiscarded, reasonCompactorDiscardedSpans, tenantID)
+	totalDiscarded := countSpans(dataEncoding, objs[1:]...)
+	overrides.RecordDiscardedSpans(totalDiscarded, reasonCompactorDiscardedSpans, tenantID)
 	return objs[0], wasCombined, nil
 }
 
-func (c *Compactor) RecordDiscardedSpans(count int, tenantID string) {
+func (c *Compactor) RecordDiscardedSpans(count int, tenantID string, traceID string) {
+	level.Warn(log.Logger).Log("msg", "max size of trace exceeded", "tenant", tenantID, "traceId", traceID, "discarded_span_count", count)
 	overrides.RecordDiscardedSpans(count, reasonCompactorDiscardedSpans, tenantID)
 }
 
@@ -293,12 +295,12 @@ func (c *Compactor) OnRingInstanceStopping(lifecycler *ring.BasicLifecycler) {}
 func (c *Compactor) OnRingInstanceHeartbeat(lifecycler *ring.BasicLifecycler, ringDesc *ring.Desc, instanceDesc *ring.InstanceDesc) {
 }
 
-func countSpans(dataEncoding string, objs ...[]byte) int {
+func countSpans(dataEncoding string, objs ...[]byte) (total int) {
+	var traceID string
 	decoder, err := model.NewObjectDecoder(dataEncoding)
 	if err != nil {
 		return 0
 	}
-	spans := 0
 
 	for _, o := range objs {
 		t, err := decoder.PrepareForRead(o)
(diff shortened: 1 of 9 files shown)


modules/compactor/compactor.go
```diff
@@ -308,10 +318,17 @@ func countSpans(dataEncoding string, objs ...[]byte) int {
 
 		for _, b := range t.Batches {
 			for _, ilm := range b.ScopeSpans {
-				spans += len(ilm.Spans)
+				for _, span := range ilm.Spans {
+					discardedMap[tempoUtil.TraceIDToHexString(span.TraceId)]++
```
scalalang2, line 322: This code iterates all spans to extract `traceId`.
However, as @mdisibio mentioned in issue #1931. There may be millions spans discarded. therefore, this approach is highly inefficient

Any suggestions for obtaining taceId more efficient?

joe-elliott: Perhaps just logging the spans here at a debug level instead of returning them?

modules/compactor/compactor.go
```diff
@@ -236,12 +237,16 @@ func (c *Compactor) Combine(dataEncoding string, tenantID string, objs ...[]byte
 		return objs[0], wasCombined, nil
 	}
 
-	spansDiscarded := countSpans(dataEncoding, objs[1:]...)
-	overrides.RecordDiscardedSpans(spansDiscarded, reasonCompactorDiscardedSpans, tenantID)
+	totalDiscarded, discarded := countSpans(dataEncoding, objs[1:]...)
```
joe-elliott, line 240: in the combine function all of the input traces will have the same id. we don't need to change the return of "countSpans" we only need to log the dropped trace id if `totalDiscarded` is greater than 0.

joe-elliott (commented):
two comments. looking good!

modules/compactor/compactor.go
scalalang2: I changed to log number of spans and traceId if number of `ilm.Spans` greater than zero.

modules/compactor/compactor.go
scalalang2: I reverted that countSpans only returns `totalDiscarded` and also moved logging code inside of the function.

tempodb/encoding/vparquet/compactor.go
```diff
@@ -315,18 +316,27 @@ func estimateMarshalledSizeFromParquetRow(row parquet.Row) (size int) {
...
-		return 0
+		return "", 0
 	}
 
 	for _, v := range row {
 		if v.Column() == spanID.ColumnIndex {
 			spans++
 		}
+
+		if v.Column() == traceIDColumn.ColumnIndex {
```
scalalang2, line 337: Does this seems to be ok?
I've want to find `traceId` in a parquet row.

modules/compactor/compactor.go
```diff
@@ -308,10 +309,13 @@ func countSpans(dataEncoding string, objs ...[]byte) int {
 
 		for _, b := range t.Batches {
 			for _, ilm := range b.ScopeSpans {
-				spans += len(ilm.Spans)
+				if len(ilm.Spans) > 0 {
+					level.Debug(log.Logger).Log("msg", "max size of trace exceeded", "traceId", tempoUtil.TraceIDToHexString(ilm.Spans[0].TraceId), "discarded_span_count", len(ilm.Spans))
```
joe-elliott, line 313: this will hit repeatedly for every batch/scope spans in the same trace. let's just move it to line 319 after the loop and only log once.

tempodb/encoding/vparquet/compactor.go
joe-elliott: i think this works :). add a test to confirm?

joe-elliott (commented):
this is looking good. two small comments.

i forgot that `v2` and `vParquet` have different paths for counting dropped spans. thank you for tracking this down. i suppose ultimately we'd prefer both act like `vParquet` does, but that's too much change for this PR.

tempodb/encoding/vparquet/compactor.go
scalalang2: Okay I'll add a test.

scalalang2: done :)

modules/compactor/compactor.go
scalalang2: yeap :) it's applied.

joe-elliott (approved):
Nice! Thanks for this improvement. It looks like there's a conflict in the changelog. If you can clean that up we'll merge.

scalalang2:
I cleaned changelog, thanks joe :)

PR #2139, opened 2023-02-23
[TraceQL] Add support to directly compare floats and ints

### LINKED ISSUE #2010: [TraceQL] Number comparisons should work for both float and int attributes
**Describe the bug**
Currently traceql makes no attempt to convert any data types while doing comparisons. This is intended for types like string <-> int, but not for int <-> float.

The impact of this bug is that this query will not return spans if `floatAttribute` is of [type float](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/common/v1/common.proto#L35) and greater than 1. 

```{ span.floatAttribute > 1 }```

Currently this query can be used and will return as expected:

```{ span.floatAttribute > 1. }```

### PR DESCRIPTION
**Pull request by joe-elliott**
**What this PR does**:
Previously the following traceql query would always return nothing b/c floats and ints were not comparable via equality:
```
{ 1 = 1. }
```
This PR makes it so that both static and attribute comparisons will work for all number types.

Fixes #2010 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


mapno (approved):
LGTM

Issue #2131, opened 2023-02-23
Request to Add Configuration Verification Option


agrib-01:
**Issue Report: Adding a Config Verification Option to Tempo**

**Problem:**
I am trying to validate the Tempo configuration file before deploying it. However, I have noticed that there is currently no option available to verify the configuration file.

**Proposed Solution:**
I would like to request the addition of a configuration verification option in Tempo, similar to the "verify-config" option available in Loki, which allows for the validation of configuration syntax before applying the configuration. This option would be very useful to have when deploying and managing Tempo, as it would ensure that the configuration file is valid and reduce the risk of errors during deployment.

**Additional Context:**
Currently, there is no option available to check the syntax of the configuration file in Tempo, which is available in other related tools such as Loki and Mimir. I believe that adding this option to Tempo would make the tool more versatile and user-friendly.

agrib-01:
Thank you for pointing out the "CheckConfig" method as a potential solution. I appreciate your input and will look into implementing this flag in Tempo. Once I have made the necessary changes, I will create a pull request for review.

agrib-01:
Hi @joe-elliott, 
I've created a PR that allows configuration validation using the flag -config.verify and then terminates the execution. 
I would appreciate it if you could review it.

agrib-01:
Hi, 
since the pull request has been merged, I'm closing this issue here.
Thanks again for pointing me to the correct function.

PR #2141, opened 2023-02-24
Add configuration verification flag to CLI (#2131)

agrib-01:
**What this PR does**:
This pull request introduces a new feature to the application in the form of a config.verify flag. When enabled, this flag allows the user to verify the configuration and then exit the application. This can be useful in scenarios where the user wants to validate the configuration file before launching the application or to troubleshoot configuration-related issues.

Fixes #[2131](https://github.com/grafana/tempo/issues/2131)

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo/main.go
+++ b/cmd/tempo/main.go
@@ -92,17 +92,7 @@ func main() {
 	// Allocate a block of memory to alter GC behaviour. See https://github.com/golang/go/issues/23044
 	ballast := make([]byte, *ballastMBs*1024*1024)
 
-	// Warn the user for suspect configurations
-	if warnings := config.CheckConfig(); len(warnings) != 0 {
-		level.Warn(log.Logger).Log("-- CONFIGURATION WARNINGS --")
-		for _, w := range warnings {
-			output := []any{"msg", w.Message}
-			if w.Explain != "" {
-				output = append(output, "explain", w.Explain)
-			}
-			level.Warn(log.Logger).Log(output...)
-		}
-	}
+	configIsValid(config)
 
 	// Start Tempo
 	t, err := app.New(*config)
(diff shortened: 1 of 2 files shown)


agrib-01:
I'm uncertain about the reason for the failure of the TestInstanceFailsLargeTracesEvenAfterFlushing test. 
As far as I know, my changes should not have any impact on this test. 
Could you point me in the right direction? Locally, the test-with-cover test is passing fine. 
Your help would be greatly appreciated.

cmd/tempo/main.go
```diff
@@ -92,17 +92,7 @@ func main() {
 	// Allocate a block of memory to alter GC behaviour. See https://github.com/golang/go/issues/23044
 	ballast := make([]byte, *ballastMBs*1024*1024)
 
-	// Warn the user for suspect configurations
-	if warnings := config.CheckConfig(); len(warnings) != 0 {
```
joe-elliott, line 96: should we just move this logic into `loadConfig()` instead of calling it in main and loadConfig()? i'm not sure why it's here.

cmd/tempo/main.go
```diff
@@ -171,11 +179,20 @@ func loadConfig() (*app.Config, error) {
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse configFile %s: %w", configFile, err)
 		}
+
+		if configVerify {
```
joe-elliott, line 183: you need to add a line to the top of this method to correctly set the value of `configVerify`. See the lines that set configFileOption and configExpandOption

cmd/tempo/main.go
```diff
@@ -171,11 +179,20 @@ func loadConfig() (*app.Config, error) {
...
+			if !configIsValid(config) {
```
joe-elliott, line 184: as said above, let's move the config verification here instead of calling it from two spots, but I believe this needs to move below the `flag.Parse()` line to take into account both file config and cli config.

joe-elliott (commented):
thanks for taking this on! we have some flakey tests in CI. i will rerun if it's an unrelated test.

joe-elliott (approved):
nice improvement. thanks!

PR #2148, opened 2023-02-27
Stop tag values search when the limit is reached in V1 endpoint

mapno:
**What this PR does**:

It uses the new generic `util.NewDistinctValueCollector` and stops when the max bytes has been reached for `/api/search/tag/{tagName}/values` and `/api/search/tags`.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * [CHANGE] Removing leading zeroes in span id [#2062](https://github.com/grafana/tempo/pull/2062) (@ie-pham)
 * [ENHANCEMENT] Metrics generator to make use of counters earlier [#2068](https://github.com/grafana/tempo/pull/2068) (@zalegrala)
 * [ENHANCEMENT] Log when a trace is too large to compact [#2105](https://github.com/grafana/tempo/pull/2105) (@scalalang2)
+* [ENHANCEMENT] Stop tag values search when the limit is reached in V1 endpoint [#](https://github.com/grafana/tempo/pull/) (@mapno)
```
electron0zero, line 9: missing PR ID: 2148

pkg/util/distinct_value_collector.go
```diff
@@ -8,7 +8,7 @@ type DistinctValueCollector[T comparable] struct {
 	totalLen int
 }
 
-// NewDistinctStringCollector with the given maximum data size. This is calculated
+// NewDistinctValueCollector with the given maximum data size. This is calculated
```
joe-elliott, line 11: i like that you removed the DistinctStringCollector and consolidated here. personally i would have just made the DistinctStringCollector a wrapper on DistinctValueCollector so all of the calling code could stay the same.

joe-elliott (commented):
This change seems broader then the headline suggests. For search tags, tag values v1 and tag values v2 i think this will stop a search mid block when the limit is reached. Previously we only checked if the limit was reached between blocks.

this change is quite wide reaching and i'm concerned about backporting it to 2.0. i'd prefer merging and waiting for 2.1.

can we make the suggested simplification in the comments? it will make this change easier to review.

CHANGELOG.md
mapno: Done

pkg/util/distinct_value_collector.go
mapno: Having to clone strings to free the underlying file made consolidating this too ugly.

tempodb/encoding/vparquet/block_search_tags.go
```diff
@@ -153,7 +153,9 @@ func searchTags(_ context.Context, cb common.TagCallback, pf *parquet.File) erro
 
 					for i := 0; i < dict.Len(); i++ {
 						s := dict.Index(int32(i)).String()
-						cb(s)
+						if cb(s) {
```
joe-elliott, line 156: seeing one place above where we can add this. line 111:
```
cb(lbl)
```

pkg/tempofb/searchdatamap.go
```diff
@@ -21,7 +21,7 @@ const (
 	maxBufferLen = 900 << 20
 )
 
-type TagCallback func(t string)
+type TagCallback func(t string) (stop bool)
```
joe-elliott, line 24: is there a reason to change this one? i would prefer leaving it alone. we are removing all of this in #2159

modules/ingester/instance_search.go
```diff
@@ -572,18 +576,22 @@ func (i *instance) SearchTagValuesV2(ctx context.Context, req *tempopb.SearchTag
...
 
 	se := &tempofb.SearchEntry{}
+
+liveTraces:
 	for _, t := range i.traces {
 		for _, s := range t.searchData {
 			se.Reset(s)
-			visitFn(se)
+			if visitFn(se) {
+				break liveTraces
```
joe-elliott, line 593: nit: return nil over break to label

joe-elliott (commented):
Added some thoughts. Does this actually impact tag values search? or just tags search?

I think you'd have to make changes here:

https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet/block_search_tags.go#L205

to impact tag values search

tempodb/search/backend_search_block.go
```diff
@@ -178,7 +178,9 @@ func (s *BackendSearchBlock) TagValues(ctx context.Context, tagName string, cb t
 	if kv != nil {
 		for j, valueLength := 0, kv.ValueLength(); j < valueLength; j++ {
 			value := string(kv.Value(j))
-			cb(value)
+			if stop := cb(value); stop {
```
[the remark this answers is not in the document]
joe-elliott: oh, this doesn't even matter. will be removed in #2159

mapno:
This fell through the cracks. Closing as it no longer seems relevant.

Issue #2034, opened 2023-01-31
Tempo Ingester Flushes Failing 


inbarpatashnik:
Ever since updating tempo version to be in line with the main branch (commit for PR #1997) we have been getting the follow alert

TempoIngesterFlushesFailing

Any idea why this might be happening? 

(Before this we were running on a previous version of main from a few months ago. The only changes made to the configuration were the image tag and getting rid of the max_bytes feature flag from the increasing storage bug which was solved) 
![AB895088-9776-4EFF-B7A4-4C08FBCD70D8](https://user-images.githubusercontent.com/75210443/215742615-cf57ca80-18b1-4eaa-bbcd-f48ffa72fc75.jpeg)

inbarpatashnik:
I saw an error message that said 
failed to cut traces because storage is full.
I assumed that it refers to the PVC which is getting full because there are no flushes?
So I struggle to find the problem :(

joe-elliott:
If you can share the exact text of the log messages as well as a config, I may be able to help you with this issue. As is, it is difficult to make guesses based on this information.

inbarpatashnik:
![Uploading 2BBFB754-799B-4E99-92CA-B9FB6E12580E.jpeg…]()

inbarpatashnik:
level-error ts=2023-01-30T13:33:25.50724828Z caller-flush.go:156 org_id=single-tenant msg="failed to cut traces"

err="error closing writer: writing buffered pages of row group column 0: write

/var/tempo/wal/8d2d8482-6807-4b71-bb54-b78c0ebd1e68+single-tenant+vP: no space left on device"

joe-elliott:
It appears that you've run out of space on your PVC. I would recommend increasing it.

inbarpatashnik:
The thing is that the PVC is getting full because the flushes are failing.
Ever since we upgraded to branch main and removed the flag I mentioned above, the flushes didn't succeed and because of that the PVC's storage was getting full.
I saw that it used to get to about 20% of its space until the flushes occurred and now it gets to 100% and flushes always fail, even when the storage is still not full (I didn't see any error message in this case).
We also tried to use the renamed flag, parquet_max_bytes, but it still didn't help.

mdisibio:
Hi, can you check the logs for the below messages?  These should indicate the root cause error when flushes are failing:

* `error performing op in flushQueue`
* `error copying block from local to remote backend`

inbarpatashnik:
[...]

PR #2152, opened 2023-02-27
Close WAL block file before deleting everything

kostya9:
**What this PR does**: Close WAL block file before deleting everything

Also added some other files closing that I found during a look around the code.

PS: I not proficient in Go, so I could have misunderstood, so please check if it looks good

Attempt to fix https://github.com/grafana/tempo/issues/2034
Fixes https://github.com/grafana/tempo/issues/2153
**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@
 * [BUGFIX] Correctly coalesce trace level data when combining Parquet traces. [#2095](https://github.com/grafana/tempo/pull/2095) (@joe-elliott)
 * [BUGFIX] Unescape query parameters in AWS Lambda to allow TraceQL queries to work. [#2114](https://github.com/grafana/tempo/issues/2114) (@joe-elliott)
 * [CHANGE] Pad leading zeroes in span id to always be 16 chars [#2062](https://github.com/grafana/tempo/pull/2062) (@ie-pham)
+* [BUGFIX] Fix float/int comparisons in TraceQL. [#2139](https://github.com/grafana/tempo/issues/2139) (@joe-elliott)
+* [BUGFIX] Fix not closing WAL block file before attempting to delete the folder. [#2139](https://github.com/grafana/tempo/pull/2152) (@kostya9)
 
 ## v2.0.0 / 2023-01-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * [CHANGE] Update Go to 1.20 [#2079](https://github.com/grafana/tempo/pull/2079) (@scalalang2)
 * [BUGFIX] Retry copy operations during compaction in GCS backend [#2111](https://github.com/grafana/tempo/pull/2111) (@mapno)
 * [BUGFIX] Fix float/int comparisons in TraceQL. [#2139](https://github.com/grafana/tempo/issues/2139) (@joe-elliott)
+* [BUGFIX] Fix not closing WAL block file before attempting to delete the folder. [#2139](https://github.com/grafana/tempo/pull/2152) (@kostya9)
 
 ## v2.0.1 / 2023-03-03
 
@@ -20,8 +21,6 @@
 * [BUGFIX] Correctly coalesce trace level data when combining Parquet traces. [#2095](https://github.com/grafana/tempo/pull/2095) (@joe-elliott)
 * [BUGFIX] Unescape query parameters in AWS Lambda to allow TraceQL queries to work. [#2114](https://github.com/grafana/tempo/issues/2114) (@joe-elliott)
 * [CHANGE] Pad leading zeroes in span id to always be 16 chars [#2062](https://github.com/grafana/tempo/pull/2062) (@ie-pham)
-* [BUGFIX] Fix float/int comparisons in TraceQL. [#2139](https://github.com/grafana/tempo/issues/2139) (@joe-elliott)
-* [BUGFIX] Fix not closing WAL block file before attempting to delete the folder. [#2139](https://github.com/grafana/tempo/pull/2152) (@kostya9)
 
 ## v2.0.0 / 2023-01-31
 
--- a/tempodb/backend/local/local.go
+++ b/tempodb/backend/local/local.go
@@ -131,6 +131,7 @@ func (rw *Backend) Read(ctx context.Context, name string, keypath backend.KeyPat
 
 	stat, err := f.Stat()
 	if err != nil {
+		f.Close()
 		return nil, -1, err
 	}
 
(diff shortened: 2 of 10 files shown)


tempodb/encoding/vparquet/wal_block.go
```diff
@@ -446,6 +446,7 @@ func (b *walBlock) Iterator() (common.Iterator, error) {
 }
 
 func (b *walBlock) Clear() error {
+	b.file.Close()
```
kostya9, line 449: This is the thing that I think may fix the original issue. `b.writer` is not closed too, but as far as I see, that it is based on the `file`, so there is no reason to double close

tempodb/encoding/vparquet/wal_block.go
```diff
@@ -446,6 +446,9 @@ func (b *walBlock) Iterator() (common.Iterator, error) {
 }
 
 func (b *walBlock) Clear() error {
+	if b.file != nil {
+		b.file.Close()
```
kostya9, line 450: This is the thing that I think may fix the original issue. b.writer is not closed too, but as far as I see, that it is based on the file, so there is no reason to double close

kostya9: Here, I debugged test `TestWalDropsZeroLength`. Not sure whether the test is simulating real-world workload, but the file is still open when we call `CutCompleteTraces`

image

tempodb/backend/local/local.go
```diff
@@ -131,6 +131,7 @@ func (rw *Backend) Read(ctx context.Context, name string, keypath backend.KeyPat
 
 	stat, err := f.Stat()
 	if err != nil {
+		f.Close()
```
kostya9, line 134: I am not sure about this one - should we close if open succeeded and stat failed?

joe-elliott: how about just deferring close? i think that's ok for read only files.

tempodb/encoding/vparquet/wal_block.go
joe-elliott: i definitely agree with this change, but let's check and return the error from `b.file.Close()`.

should we run `os.RemoveAll()` even if `b.file.Close()` fails? it might be the safer thing to do

tempodb/encoding/vparquet/wal_block.go
```diff
@@ -460,6 +463,8 @@ func (b *walBlock) FindTraceByID(ctx context.Context, id common.ID, _ common.Sea
 			}
 
 			r := parquet.NewReader(file)
+			defer r.Close()
```
joe-elliott, line 466: should we also defer close the file above? and all of the files returned from `page.file()` calls?

joe-elliott (commented):
thanks for taking a look at this :+1:

tempodb/backend/local/local.go
kostya9: Cant defer close in the higher scope because f is returned from the function. I assume the caller expects an opened file

tempodb/encoding/vparquet/wal_block.go
kostya9: `parquet.File` unfortunately does not implement a Close function, so we can not close it. I noticed that when I was figuring out what is going on and logged an issue - https://github.com/grafana/tempo/issues/2153

tempodb/encoding/vparquet/wal_block.go
kostya9: Added error aggregation. I agree, I think it's safer to remove everything rather than not, so did that.

joe-elliott (commented):
apologies. i had a pending review that i did not submit from a couple days back.

tempodb/encoding/vparquet/wal_block.go
kostya9: Sounds interesting.

What do you think of wrapping the parquet.File with another type implements Close()? I feel like maybe that would be a more intuitive design. Also, it could later be used in other call sites theoretically (which would probably also need to close the file at some point).

joe-elliott: so like:
```
type pageFile struct {
  parquet.File
  os.File
}
```
then the caller can interact with the parquet.file and eventually close the os.file? i'm good on that

kostya9: Added closing the file opened in `file()`. It resulted in a little bit more changes that I anticipated, but we should close everything opened in the wal_block module now

kostya9: Ah, did all that work, but didnt call the finel `iterator.Close()`. Fixed that

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@
 * [BUGFIX] Correctly coalesce trace level data when combining Parquet traces. [#2095](https://github.com/grafana/tempo/pull/2095) (@joe-elliott)
 * [BUGFIX] Unescape query parameters in AWS Lambda to allow TraceQL queries to work. [#2114](https://github.com/grafana/tempo/issues/2114) (@joe-elliott)
 * [CHANGE] Pad leading zeroes in span id to always be 16 chars [#2062](https://github.com/grafana/tempo/pull/2062) (@ie-pham)
+* [BUGFIX] Fix float/int comparisons in TraceQL. [#2139](https://github.com/grafana/tempo/issues/2139) (@joe-elliott)
```
joe-elliott, line 22: we redid the changelog for 2.0.1. can you merge main or rebase and clean this up?

joe-elliott (approved):
@mapno I'm good on this, but before merging can you take a quick look? this does a lot of subtle changes with regard to the wal and searching. i'm concerned we may see some strange issues crop up b/c we relied on these files being opened, but i think explicitly closing these files is the right direction.

don't feel the need to do a line by line review (unless you want). mainly flagging you due to the changes along the SearchTags and TagValues paths.

@kostya9 thanks for the persistence and level of detail on this! after @mapno takes a look we will merge.

CHANGELOG.md
kostya9: Sure, rebased and fixed this

mapno (approved):
LGTM. This change touches a lot of code, so it's difficult to follow all the `close()` calls. Tests should cover most edge cases. Thanks!

kostya9:
I see failures in the generator storage tests, from my analysis, the generator doesn't use the code I changed here. Are these tests flaky, or is there a chance something is broken?

joe-elliott:
> I see failures in the generator storage tests, from my analysis, the generator doesn't use the code I changed here. Are these tests flaky, or is there a chance something is broken?

Yeah. It's a flakey test. Will rerun.

joe-elliott:
Thanks for the fixes @kostya9!

PR #2164, opened 2023-03-03
Tag search improvements

mapno:
**What this PR does**:

This PR contains a few fixes to tag search in vParquet.

* Pass correct context to tag search func `searchTagValues()`
* Search head block in `SearchTagValuesV2`
* Take `blocksMtx` lock only once in `SearchTagValuesV2`
* Don't cache parquet files for search

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


tempodb/encoding/vparquet/block_search.go
```diff
@@ -48,6 +48,7 @@ func (b *backendBlock) openForSearch(ctx context.Context, opts common.SearchOpti
 	if b.pf != nil && b.readerAt != nil {
 		// Reset metrics, is there a better way to do this?
 		b.readerAt.TotalBytesRead.Store(0)
+		// TODO: We should reset the context here, but it's not possible to do so without creating a new parquet.File.
```
joe-elliott, line 51: this is a nasty context leak.

when working on the wal i found that it was actually more memory efficient to _not_ cache parquet.files and the performance penalty of repeatedly opening was quite low. would recommend pushing a test image that does not do this file caching and see if it hurts perf.

joe-elliott (commented):
Some other things we can try:
- We pass `common.DefaultSearchOptions()` to all searches done on disk. this sets up reads of 1MB which is tuned for object storage. Parquet defaults to 4KB per read which is better tuned for disk access. Might be worth seeing what happens if leave the 4KB default. If you do this remember the potential issue with ReadRange() opening a file everytime!
- wal is read synchronously and backend blocks are read async. if one is outperforming the other we could use that as a hint about what works best for performance on local disk ingester search.

joe-elliott (approved):
lgtm. some good cleanup here.

does locking only once in SearchTagValuesV2 improve performance? or just simplifies code?

mapno:
> does locking only once in SearchTagValuesV2 improve performance? or just simplifies code?

Can't say for sure. Most of the improvements came after we stopped caching files. I think it's a nice improvement anyway.

PR #2119, opened 2023-02-21
TraceQL Perf: Reduce metadata retrieved by implementing two pass iteration.

joe-elliott:
**What this PR does**:
Dramatically reduces the amount of data pulled from the backend for "needle in the haystack" queries by implementing a two pass iteration over the data. The first pass retrieves all of the fields necessary to evaluate the query and then metadata is only retrieved for those spans that actually fulfill the query. Previously metadata was pulled along with the span data which resulted in slower queries when there were few matches.

Fixes #2138 

e.g. for the query `{ resource.service.name = "foo" }`

Previously iterators looked like the following. In this case span metadata is always retrieved.
```
JoinIterator: 0	
	LeftJoinIterator: 1
		required: 
			ColumnIterator: rs.Resource.ServiceName 
			InstrumentedPredicate{0, StringInPredicate{foo, }}
			LeftJoinIterator: 3
			required: 
				ColumnIterator: rs.ils.Spans.StartUnixNanos 
				InstrumentedPredicate{0, nil}
				ColumnIterator: rs.ils.Spans.EndUnixNanos 
				InstrumentedPredicate{0, nil}
				ColumnIterator: rs.ils.Spans.ID 
				InstrumentedPredicate{0, nil}
			optional: 
		optional: 
	ColumnIterator: TraceID 
		InstrumentedPredicate{0, nil}
	ColumnIterator: StartTimeUnixNano 
		InstrumentedPredicate{0, nil}
	ColumnIterator: DurationNanos 
		InstrumentedPredicate{0, nil}
	ColumnIterator: RootSpanName 
		InstrumentedPredicate{0, nil}
	ColumnIterator: RootServiceName 
		InstrumentedPredicate{0, nil}
traceCollector{})
```

Now they look like the following. In this case only the data necessary to evaluate the query is pulled, a "filter" callback is called and only those spans that survive the callback pass through the trace and span metadata iterators.
```
JoinIterator: 0	
	JoinIterator: 3	
		spansToMetaIterator: 
			JoinIterator: 0	
			LeftJoinIterator: 1
				required: 
					ColumnIterator: rs.Resource.ServiceName 
					InstrumentedPredicate{0, StringInPredicate{bar, }}
					LeftJoinIterator: 3
					required: 
						ColumnIterator: rs.ils.Spans.Kind 
						InstrumentedPredicate{0, nil}
					optional: 
					spanCollector(0, [])
				optional: 
				batchCollector{true, 1}
			ColumnIterator: StartTimeUnixNano 
[...]

PR #2165, opened 2023-03-03
Synchronous iterator

mdisibio:
**What this PR does**:
This PR creates a new synchronous version of the ColumnIterator.   Current tests are passing and performance is looking better, but needs a bit more polish and testing.  Need to run it against the enhanced set of tests in #2119.  

```
name                                 old time/op    new time/op    delta
BackendBlockTraceQL/noMatch-12         33.8ms ± 2%    31.6ms ± 2%   -6.37%  (p=0.000 n=9+10)
BackendBlockTraceQL/partialMatch-12    53.4ms ±14%    17.0ms ± 1%  -68.19%  (p=0.000 n=10+9)
BackendBlockTraceQL/service.name-12    2.48ms ± 1%    2.43ms ± 3%   -1.94%  (p=0.001 n=10+8)

name                                 old speed      new speed      delta
BackendBlockTraceQL/noMatch-12        505MB/s ± 2%   539MB/s ± 2%   +6.81%  (p=0.000 n=9+10)
BackendBlockTraceQL/partialMatch-12   462MB/s ±13%   667MB/s ± 1%  +44.39%  (p=0.000 n=10+9)
BackendBlockTraceQL/service.name-12   293MB/s ± 1%   299MB/s ± 3%   +1.99%  (p=0.001 n=10+8)

name                                 old MB_io/op   new MB_io/op   delta
BackendBlockTraceQL/noMatch-12           17.0 ± 0%      17.0 ± 0%     ~     (all equal)
BackendBlockTraceQL/partialMatch-12      24.5 ± 0%      11.3 ± 0%  -53.85%  (p=0.000 n=10+10)
BackendBlockTraceQL/service.name-12      0.73 ± 0%      0.73 ± 0%     ~     (all equal)

name                                 old alloc/op   new alloc/op   delta
BackendBlockTraceQL/noMatch-12         2.25MB ± 4%    2.23MB ± 6%     ~     (p=0.400 n=9+10)
BackendBlockTraceQL/partialMatch-12    46.4MB ± 0%    23.4MB ± 1%  -49.52%  (p=0.000 n=10+10)
BackendBlockTraceQL/service.name-12     234kB ± 1%     221kB ± 2%   -5.78%  (p=0.000 n=10+10)

name                                 old allocs/op  new allocs/op  delta
BackendBlockTraceQL/noMatch-12          31.3k ± 0%     31.2k ± 0%   -0.12%  (p=0.000 n=10+8)
BackendBlockTraceQL/partialMatch-12     72.0k ± 0%     49.7k ± 0%  -30.94%  (p=0.000 n=10+10)
BackendBlockTraceQL/service.name-12     3.36k ± 0%     3.32k ± 0%   -1.01%  (p=0.000 n=10+10)
```

n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added


knylander-grafana:
@mdisibio Will we need doc for this?

joe-elliott:
Can we resurrect this PR? I keep seeing traces like this: 

image

Issue here: https://github.com/grafana/tempo/issues/2336

This PR would eliminate the possibility that the large gaps seen here are due to synchronization. It would also hopefully allow for easier performance analysis by simplifying this bit of code.

mdisibio:
> Can we resurrect this PR?

Yes.  A couple updates:

1. What do you think about putting this behind an environment variable like `PARQUET_SYNC_ITERATOR=1`?  So we can safely merge and test, and also compare performance between pods.
2. There is still a bug or two that surfaced when using this for the new metrics stuff on real blocks, which isn't caught by the current test suite.  Hopefully a simple fix.

joe-elliott:
> What do you think about putting this behind an environment variable like PARQUET_SYNC_ITERATOR=1? So we can safely merge and test, and also compare performance between pods.

no concerns. do what makes sense to you. is the PR that risky? or are the gains not very clear?

> There is still a bug or two that surfaced when using this for the new metrics stuff on real blocks, which isn't caught by the current test suite. Hopefully a simple fix.

:+1:

mdisibio:
> @mdisibio Will we need doc for this?

I don't think so. This is a performance experiment and if it pans out we will make it permanent.  No user-facing changes.

pkg/parquetquery/iters.go
```diff
@@ -252,6 +296,370 @@ func columnIteratorResultPoolPut(r *IteratorResult) {
...
+	currPageMax     RowNumber
+	currValues      pq.ValueReader
+	currBuf         []pq.Value
+	currBufN        int
+}
+
+var _ Iterator = (*SyncIterator)(nil)
+
+func NewSyncIterator(ctx context.Context, rgs []pq.RowGroup, column int, columnName string, readSize int, filter Predicate, selectAs string) *SyncIterator {
+
```
joe-elliott, line 330: ctx not used? it looks like we're losing the `columnIterator.iterate` span with this change which i find somewhat useful.

pkg/parquetquery/iters.go
```diff
@@ -252,6 +296,370 @@ func columnIteratorResultPoolPut(r *IteratorResult) {
...
+				if skip > 0 {
+					if err := c.currPages.SeekToRow(skip); err != nil {
+						return true, err
+					}
+					c.curr.Skip(skip)
+				}
+			}*/
+
+		for c.currPage == nil {
+			pg, err := c.currPages.ReadPage()
```
joe-elliott, line 484: i think this is showing we have to read the page to count its rows and skip it correctly? unfortunate and could be a nice future improvement for needle in the haystack style queries

joe-elliott (approved):
lgtm. some small comments, but excited to get this one in.

pkg/parquetquery/iters.go
joe-elliott: oof, good point. our traces need attention generally and this PR is too obvious a win to block.

fine with the "NewSyncIterator" approach to see how it goes.

mdisibio: 👍 Tested it out and it works well enough for a first pass or if we have a better idea.  The side-effect is that the span starts as soon as the iter is created, instead of on the first "pull" like the async one.

mdisibio:
Pushed a small fix to `mergeSpanSetIterator` to `Close()` iters as soon as possible when they are exhausted, instead of waiting all the way til the end. Mostly for timing accuracy, performance not really affected.

PR #2171, opened 2023-03-03
include internal server

ie-pham:
**What this PR does**: This fixes a problem in GET where Tokengen is used. The problem was that Tokengen also spins up the server, which has internal server as a dependency. Since internal server was added as a service to wait for, the server service for Tokengen was never allowed to properly stop.  

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott (approved):
nice work, pham

PR #2172, opened 2023-03-04
Metrics Span Multiplier

### LINKED ISSUE #2166: Dynamic Sampling
In one of our services we want to sample every route, user, app version by another ratio, the problem is generated metrics won't reflect the reality.

using float64 attribute named `X-SampleRatio` we can multiply all the metrics by `1.0/ratio` 

**Additional context**
It's just like what HoneyComb does in https://github.com/honeycombio/honeycomb-opentelemetry-go/blob/main/deterministic_sampler.go and https://docs.honeycomb.io/manage-data-volume/sampling/

If it's okay i would like to implement it.

altanozlu:
**What this PR does**:
Adds span multiplier feature by using "X-SampleRatio" attribute.

Fixes #2166

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/generator/processor/servicegraphs/config.go
+++ b/modules/generator/processor/servicegraphs/config.go
@@ -27,6 +27,7 @@ type Config struct {
 	// If client and server spans have the same attribute, behaviour is undetermined
 	// (either value could get used)
 	Dimensions []string `yaml:"dimensions"`
+
 	// If enabled X-SampleRatio attribute will be used for metric calculation
 	EnableSpanMultiplier bool `yaml:"enable_span_multiplier"`
 }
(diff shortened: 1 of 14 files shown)


modules/generator/processor/servicegraphs/config.go
```diff
@@ -27,6 +27,9 @@ type Config struct {
 	// If client and server spans have the same attribute, behaviour is undetermined
 	// (either value could get used)
 	Dimensions []string `yaml:"dimensions"`
+
+	// If enabled X-SampleRatio attribute will be used for metric calculation
+	EnableSpanMultiplier bool `yaml:"enable_span_multiplier"`
```
mapno, line 32: If there isn't a semantic convention for this attribute, we could use a string, that if defined, will be used as key to get the information.

mapno (commented):
I like the changes. This is something we've wanted to do for a long time. Thanks for starting the work.

modules/generator/processor/servicegraphs/config.go
altanozlu: Yea we can have that

altanozlu:
Hi @mapno, do you have any idea why "Test Packages" test failed and can we rerun it ?

mapno (approved):
LGTM

PR #2183, opened 2023-03-08
Release spans when nothing matches

joe-elliott:
**What this PR does**:
- Adds an interface method that allows the engine to signal to the fetch layer it is dropping spans
- Use the new method to release spans back to the fetch pool when nothing matches

When doing a needle in the haystack query that is rejected by the engine, we drop millions of spans on the floor which puts a lot of pressure on the GC. This PR releases spans back to the pool when the engine rejects everything. This is the 99.99...% case when searching (b)millions of traces for just one or two.

### DIFF
--- a/pkg/traceql/ast_test.go
+++ b/pkg/traceql/ast_test.go
@@ -250,3 +250,4 @@ func (m *mockSpan) StartTimeUnixNanos() uint64 {
 func (m *mockSpan) EndtimeUnixNanos() uint64 {
 	return m.endTimeUnixNanos
 }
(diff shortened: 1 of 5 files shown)


mdisibio (approved):
:rocket: Nice catch.  I wonder if there is a meaningful difference between putting Release on the individual spans vs the fetcher.

joe-elliott:
> I wonder if there is a meaningful difference between putting Release on the individual spans vs the fetcher.

That's a good question. Performance wise I'd guess they are roughly the same, but we may prefer one pattern over another.

PR #2195, opened 2023-03-13
Add Azure Workload Identity authentication support

LambArchie:
**What this PR does**:
Adds support for [Azure Workload Identity](https://azure.github.io/azure-workload-identity/docs/) authentication.

Partially based off the following PR https://github.com/grafana/loki/pull/7540 which implemented Azure Workload Identity into Loki.

**Checklist**
- [x] Tests updated
- [x] Documentation added

---

This is my first time writing anything in Go, so I might have made a mistake with the syntax. If I have feel free to fix it or let me know and I'll fix it myself.  
I have tested this and it seems to work so I'm hoping I haven't made a mistake but you never know.

### DIFF
--- a/tempodb/backend/azure/azure_helpers.go
+++ b/tempodb/backend/azure/azure_helpers.go
@@ -166,7 +166,7 @@ func getStorageAccountKey(cfg *Config) string {
 }
 
 func getOAuthToken(cfg *Config) (*blob.TokenCredential, error) {
-	spt, err := getServicePrincipalToken(cfg, defaultAuthFunctions)
+	spt, err := getServicePrincipalToken(cfg)
 	if err != nil {
 		return nil, err
 	}
@@ -194,19 +194,19 @@ func getOAuthToken(cfg *Config) (*blob.TokenCredential, error) {
 	return &tc, nil
 }
 
-func getServicePrincipalToken(cfg *Config, authFunctions authFunctions) (*adal.ServicePrincipalToken, error) {
+func getServicePrincipalToken(cfg *Config) (*adal.ServicePrincipalToken, error) {
 	endpoint := cfg.Endpoint
 
 	resource := fmt.Sprintf("https://%s.%s", cfg.StorageAccountName, endpoint)
 
 	if cfg.UseFederatedToken {
-		token, err := servicePrincipalTokenFromFederatedToken(resource, authFunctions.NewOAuthConfigFunc, authFunctions.NewServicePrincipalTokenFromFederatedTokenFunc)
+		token, err := servicePrincipalTokenFromFederatedToken(resource, defaultAuthFunctions)
 		if err != nil {
 			return nil, err
 		}
 
 		var customRefreshFunc adal.TokenRefresh = func(context context.Context, resource string) (*adal.Token, error) {
-			newToken, err := servicePrincipalTokenFromFederatedToken(resource, authFunctions.NewOAuthConfigFunc, authFunctions.NewServicePrincipalTokenFromFederatedTokenFunc)
+			newToken, err := servicePrincipalTokenFromFederatedToken(resource, defaultAuthFunctions)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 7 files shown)


tempodb/backend/azure/azure_helpers.go
```diff
@@ -181,11 +201,37 @@ func getOAuthToken(cfg *Config) (*blob.TokenCredential, error) {
 	return &tc, nil
 }
 
-func getServicePrincipalToken(cfg *Config) (*adal.ServicePrincipalToken, error) {
+func getServicePrincipalToken(cfg *Config, authFunctions authFunctions) (*adal.ServicePrincipalToken, error) {
```
joe-elliott, line 204: it seems like for `getServicePrincipalToken()` we always pass `defaultAuthFunctions`. Can we remove the `authFunctions` parameter here? and only take it at `servicePrincipalTokenFromFederatedToken` level?

tempodb/backend/azure/azure_helpers.go
```diff
@@ -196,3 +242,32 @@ func getServicePrincipalToken(cfg *Config) (*adal.ServicePrincipalToken, error)
 
 	return msiConfig.ServicePrincipalToken()
 }
+
+func servicePrincipalTokenFromFederatedToken(environment string, resource string, newOAuthConfigFunc func(activeDirectoryEndpoint, tenantID string) (*adal.OAuthConfig, error), newServicePrincipalTokenFromFederatedTokenFunc func(oauthConfig adal.OAuthConfig, clientID string, jwt string, resource string, callbacks ...adal.TokenRefreshCallback) (*adal.ServicePrincipalToken, error)) (*adal.ServicePrincipalToken, error) {
```
joe-elliott, line 246: as suggested above, let's change this to take the `authFunctions` struct. Alternatively we can create function types to clean up this function signature. Something like:

```
type oAuthConfig func(activeDirectoryEndpoint, tenantID string) (*adal.OAuthConfig, error)
```

joe-elliott (commented):
Thanks for the PR! I added some thoughts on cleanup. I'm also concerned about changing this functionality as its difficult to test and I am far from an Azure auth expert. We will still be willing to merge, but any testing you can do on your end to feel confident in this change would be appreciated.

tempodb/backend/azure/azure_helpers.go
```diff
@@ -14,14 +14,34 @@ import (
...
 
 const (
 	maxRetries = 1
+
+	// Environment
+	azureGlobal       = "AzureGlobal"
+	azurePublicCloud  = "AzurePublicCloud"
+	azureChinaCloud   = "AzureChinaCloud"
+	azureGermanCloud  = "AzureGermanCloud"
+	azureUSGovernment = "AzureUSGovernment"
```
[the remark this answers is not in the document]
LambArchie: https://github.com/Azure/go-autorest/blob/main/autorest/azure/environments.go#L34

However, I've refactored the logic to not require an environment being set so this has been [removed](https://github.com/grafana/tempo/pull/2195/commits/d5f3332c85914414aaf26c9abfd3a16a3198da10) (instead it gets the AAD URL via an environment variable which is set by automatically and if for some reason that is missing default to AzurePublicCloud).

(If you are curious about azureGlobal, it was unnecessary, it was originally there as on Loki that was there to maintain backwards compatibility and I forgot to remove it when copying across)

tempodb/backend/azure/azure_helpers.go
LambArchie: Done in https://github.com/grafana/tempo/pull/2195/commits/83b5d32b2d091cd8646ef047d99c26d802e6d8ce

tempodb/backend/azure/azure_helpers.go
LambArchie: Done in https://github.com/grafana/tempo/pull/2195/commits/83b5d32b2d091cd8646ef047d99c26d802e6d8ce

LambArchie:
> Thanks for the PR! I added some thoughts on cleanup. I'm also concerned about changing this functionality as its difficult to test and I am far from an Azure auth expert. We will still be willing to merge, but any testing you can do on your end to feel confident in this change would be appreciated.

I tested my original changes and it seemed to work, including token renewal as I continued to see changes on the test azure storage account 32 hours after I initially started tempo (tokens are renewed every hour but if not are valid for 24h).

I will test again with these changes to ensure no regression has occurred

joe-elliott (approved):
LGTM. Nice improvement!

Ivalberto:
What is the rigth configuration to have tempo working with workload identity ? I receiving failed to created store : getting storage container: open: no such file or directory. My container exist, relation between sa and manage identity also exist

PR #2219, opened 2023-03-16
[TraceQL] Quit tag values search if virtual tags returns anything

### LINKED ISSUE #2200: Autocomplete returning both integer and string types for `status` field
For the `status` field, Tempo autocomplete returns both strings (`unset`, `error`, `ok`) and integers `0`, `1`, `2`. This causes a confusing result when users looks at the possible values for `status` (see screenshot) - they're left wondering if status should be treated as an integer or a string for the purposes of building a TraceQL query?

status-autocomplete

It appears that numerical comparison operators (`=`) aren't working with the integer values (`0`). 

autocomplete-status-fail

joe-elliott:
**What this PR does**:
Halts v2 tag values search if any virtual tags are found.

image

cc @mapno since you're working in this area
cc @mdisibio since you had recently changed the behavior of v1/v2 tag values search w/r to `status`

Fixes #2200 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


zalegrala (approved):

mapno (approved):
Approach looks good 👍

PR #2240, opened 2023-03-20
Split tests between unit and integration

joe-elliott:
**What this PR does**:
Attempts to split our tests into unit and integration tests. I did this for a few reasons:

- To help distinguish between true unit tests in our `_test.go` files and tests that have broader coverage.
- To reduce the size of the unit tests job in CI

I think it accomplishes the above, but choosing "integration" and "unit" tests felt arbitrary at times. We have plenty of examples of tests that don't fall neatly into either category. If @grafana/tempo as a team works to maintain a distinction between the two, then this could be a powerful way to structure future tests. Otherwise it's just going to rot and leave people scratching their heads in a few years time.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


Makefile
```diff
@@ -34,8 +34,7 @@ ifeq ($(BUILD_DEBUG), 1)
 	GO_OPT+= -gcflags="all=-N -l"
 endif
 
-GOTEST_OPT?= -race -timeout 30m -count=1
-GOTEST_OPT_WITH_COVERAGE = $(GOTEST_OPT) -cover
+GOTEST_OPT?= -race -timeout 30m -count=1 -cover
```
zalegrala, line 37: Do we want to include `-v` here?  I'm good to drop it and bring it back if we miss it.

tempodb/compactor_integration_test.go
```diff
@@ -0,0 +1,393 @@
+package tempodb
```
zalegrala, line 1: Do we need to add a go build flag here to limit the scope of the build?  This will require that the jobs run the job with the given tags to include.

zalegrala: Applies to the integration test files.

zalegrala (approved):
Good thinking.

tempodb/compactor_integration_test.go
joe-elliott: I considered build tags, but I think we'd need to add tags to every single test file to split the tests in CI. The trick here is we are using a prefix on the test name. The make file runs all tests that start with `TestIntegration` in a different target than just `Test`

Makefile
joe-elliott: I considered it, but left it out since we weren't doing it before.

.github/workflows/ci.yml
```diff
@@ -48,9 +48,8 @@ jobs:
         with:
           version: v1.51.0
 
-
   unit-tests:
-    name: Test packages
+    name: Unit test packages
```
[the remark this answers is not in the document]
joe-elliott: unsure what you're suggesting. change the job name to "make test-unit" ? i have no objection if that's what the team wants

zalegrala:
This looks okay to me.

electron0zero:
opened https://github.com/grafana/tempo/pull/2324 to spit tests job into multiple jobs

PR #2247, opened 2023-03-21
Update compactor to handle context

zalegrala:
**What this PR does**:

* Update tempodb.Compactor interface to include a context
* Update implementations of tempodb.Compactor to pass a context when available
* Handle context.Done() to break the compaction cycle
* Small format and docstring changes

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/compactor.go
+++ b/tempodb/compactor.go
@@ -61,16 +61,18 @@ var (
 	}, []string{"tenant"})
 )
 
-// todo: pass a context/chan in to cancel this cleanly
-func (rw *readerWriter) compactionLoop() {
+func (rw *readerWriter) compactionLoop(ctx context.Context) {
 	compactionCycle := DefaultCompactionCycle
 	if rw.compactorCfg.CompactionCycle > 0 {
 		compactionCycle = rw.compactorCfg.CompactionCycle
 	}
 
 	ticker := time.NewTicker(compactionCycle)
-	for range ticker.C {
+	select {
+	case <-ticker.C:
 		rw.doCompaction()
+	case <-ctx.Done():
+		return
 	}
 }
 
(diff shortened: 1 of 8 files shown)


zalegrala:
If we pass the context through to the do compaction loop, I think we can cancel the in flight compaction, though I do like the idea of tidying up the current compaction and having the side benefit of staggering the restart time.  A subservice to block the shutdown could possibly handle that.

zalegrala:
Makes sense.  I'll have to review the block ownership to understand the startup behavior you mentioned.

I've updated to include context cancellation in the retention and compaction blocks.  This feels like a good place to settle for now, but we could push cancellation further, perhaps later though.  In my testing, this terminates the compactor pretty darn quickly, but I'll be interested to see how it performs in some of our larger environments.  Hows this look after the update?  Should do something now :)

tempodb/compactor.go
```diff
@@ -111,44 +113,48 @@ func (rw *readerWriter) doCompaction() {
...
+				level.Error(rw.logger).Log("msg", "error during compaction cycle", "err", err)
+				metricCompactionErrors.Inc()
+			}
+
+			// after a maintenance cycle bail out
+			if start.Add(rw.compactorCfg.MaxTimePerTenant).Before(time.Now()) {
+				measureOutstandingBlocks(tenantID, blockSelector, rw.compactorSharder.Owns)
+
+				level.Info(rw.logger).Log("msg", "compacted blocks for a maintenance cycle, bailing out", "tenantID", tenantID)
+				return
```
zalegrala, line 148: No longer a break, but return.  Hard to see through the diff.  A couple other `break` is now `return`.

joe-elliott (commented):
this now passes context all the way down into the actual requests (i think). this will cause the actual http clients to cancel their requests.

i'm fine with this change. my guess it is basically the same thing that's happening now just more intentional, but we do need to closely monitor this in all 3 cloud envs to make sure this doesn't have unexpected consequences with the various object storage backends.

joe-elliott (approved):

Issue #1868, opened 2022-11-07
TraceQL autocomplete


mapno:
## Context

As we develop TraceQL and its main integration with Grafana, it is important for the language that we make it easy to use and provide a gentle learning curve. A big part of that experience is making writing queries easy for the users.

Many well-known query languages, such as PromQL and LogQL, provide the user with suggestions of possible conditions that will result in a successful query. We aim to add this concept to TraceQL as well.

As it is right now, suggestions are very limited by the API design and implementation. Grafana queries Tempo for all tags and their values, and Tempo will return everything that it has, without any filtering based on the current query. That makes it possible for Grafana to suggest tags and values that produce no matches in Tempo.

## Objectives

- Improve tag values search results, so only values that can produce a query that returns a non-empty results set are returned.
- Design new API that makes querying for tags and values more efficient that current `# of queries = 1 tag query + (tag values queries * # of tags)`.

#### Objective 1: Extend `/api/v2/search/tag/<tag.name>/values`

- [x] Support query param that contains TraceQL query
  - eg. `/api/v2/search/tag/resource.namespace/values?q="{cluster="foo"}"`
- [x] Parse TraceQL query and extract conditions to filter
- [x] Implement filtering in WAL blocks
- [ ] Integrate in Grafana - https://github.com/grafana/grafana/issues/66956
- [x] Implement filtering in backend blocks
  - [x] Add non-collecting column iterators to tag value search
- [ ] ~Implement filtering in live traces (?)~ (Implemented using the traceql engine, which is not supported in live traces)
- [x] Compare performance between filtering / no filtering
- [ ] (Stretch) Support more than one spanset selector `{}` per query—ie. `{} && {}`.
  - [ ] ~Define identifier for the active spanset selector (eg. cursor position)~ No longer relevant

#### Objective 2: Improvements over Objective 1

- [ ] Support complete TraceQL spec
  - [ ] Support multiple spanset filters `{ true } && { true }`
  - [ ] Support OR conditions `{ true || false }`, `{ true } || { false }`
- [ ] Use yacc parser instead of regex
  - [ ] Implement graceful failure mode for parser

#### Objective 3 (Optional): Design and implementation of new API

Extending the current API has its limitations. Performance could become a significant bottleneck as the query that's being typed grows too long or complicated, since Tempo would be basically doing a search for every tag, fetching its values.

That's why, there's already been some consideration on how an improved solution could look like.

- [ ] Design new API together with Grafana
[...]

PR #2253, opened 2023-03-23
[Tag value search] Support `q` query param in `/api/v2/search/<tag.name>/values`

mapno:
**What this PR does**:

Add support for passing a traceql query to tag values search, via query param `q`. The query can be incomplete—eg. `{ .cluster = "foo" && .namespace = }`. In such cases, Tempo will extract the value matchers, ie. ` foo = "bar"`, and build a query that can be parse by the storage layer.

This is done with a regex that captures the valid matchers. Below is a benchmark of this function for some queries. Overall, the cost of extracting the matchers is very low, and shouldn't affect latency.

```zsh
go test -bench=BenchmarkExtractMatchers -run=^$ ./pkg/traceql
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
BenchmarkExtractMatchers/{.service_name_=_"foo"}-8                                                                                                 	  786459	      1482 ns/op
BenchmarkExtractMatchers/{.service_name_=_"foo"_&&_.http.status_code_=_200}-8                                                                      	  405924	      2781 ns/op
BenchmarkExtractMatchers/{.service_name_=_"foo"_&&_.http.status_code_=_200_&&_.http.method_=_"GET"}-8                                              	  301669	      3908 ns/op
BenchmarkExtractMatchers/{.service_name_=_"foo"_&&_.http.status_code_=_200_&&_.http.method_=_"GET"_&&_.http.url_=_"/foo"}-8                        	  450922	      2730 ns/op
BenchmarkExtractMatchers/{.service_name_=_"foo"_&&_.cluster_=_}-8                                                                                  	  453796	      2488 ns/op
BenchmarkExtractMatchers/{.service_name_=_"foo"_&&_.http.status_code_=_200_&&_.cluster_=_}-8                                                       	  297260	      3895 ns/op
BenchmarkExtractMatchers/{.service_name_=_"foo"_&&_.http.status_code_=_200_&&_.http.method_=_"GET"_&&_.cluster_=_}-8                               	  232723	      5048 ns/op
BenchmarkExtractMatchers/{.service_name_=_"foo"_&&_.http.status_code_=_200_&&_.http.method_=_"GET"_&&_.http.url_=_"/foo"_&&_.cluster_=_}-8         	  454668	      2646 ns/op
PASS
ok  	github.com/grafana/tempo/pkg/traceql	9.972s
```

Contributes to https://github.com/grafana/tempo/issues/1868

**Checklist**
- [x] Tests updated
- [x] Documentation added


mapno:
> Do we need to add docs for this?

Yes. I will update the API docs shortly 👍

pkg/traceql/engine.go
```diff
@@ -232,3 +340,33 @@ func (s Static) asAnyValue() *common_v1.AnyValue {
...
 }
+
+// Regex to extract matchers from a query string
+// This regular expression matches a string that contains three groups separated by operators.
+// The first group is a string of alphabetical characters, dots, and underscores.
+// The second group is a comparison operator, which can be one of several possibilities, including =, >, <, and !=.
+// The third group is either a string of alphabetical characters or a number.
+// Example: "http.status_code = 200" from the query "{ .http.status_code = 200 && .http.method = }"
+// TODO: Support booleans (true|false) and durations
+var re = regexp.MustCompile(`([a-zA-Z._]+)\s*(=|=>|=<|=~|!=|>|<|!~)\s*("([a-zA-Z._]+)"|([0-9]+))`)
```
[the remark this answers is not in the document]
mapno: Fixed, nice!

pkg/traceql/engine.go
```diff
@@ -232,3 +343,34 @@ func (s Static) asAnyValue() *common_v1.AnyValue {
...
+// The first group is a string of alphabetical characters, dots, and underscores.
+// The second group is a comparison operator, which can be one of several possibilities, including =, >, <, and !=.
+// The third group is one of several possible values: a string enclosed in double quotes,
+// a number with an optional time unit (such as "ns", "ms", "s", "m", or "h"),
+// a plain number, or the boolean values "true" or "false".
+// Example: "http.status_code = 200" from the query "{ .http.status_code = 200 && .http.method = }"
+var re = regexp.MustCompile(`[a-zA-Z._]+\s*[=|<=|<=|=~|!=|>|<|!~]\s*(?:"[a-zA-Z._]+"|[0-9smh]+|true|false)`)
+
+// extractMatchers extracts matchers from a query string and returns a string that can be parsed by the storage layer.
+func extractMatchers(query string) string {
```
joe-elliott, line 358: it seems unlikely this is solvable with a regex. what about:

```
{ (<cond> || <cond>) && <cond> } // we can nest ()'s infinitely with booleans operators in all kinds of combinations
```

or 

```
{ <some conditions> } && { <different conditions> } 
```

I think some of these edge cases will always remain unsolved which is fine, but picking up all matchers out of a query string with a regex will provide very wrong results in a lot of cases.

mapno: We're only targeting support for queries of the type `{ a = b && c = d && .. }`, which matches what the UI editor supports.

Every other query will probably remain unsupported until a parser in the frontend (Grafana) exists. Or a better alternative in Tempo is found.

mapno: Those cases could be handled if OR `||` or more that one selector is found.

joe-elliott: Is there somewhere that we assert that the query matches this shape and if it doesn't fall back to the old behavior?
```
{ a = b && c = d && .. }
```

mapno: Not yet. But I was planning in adding it if we're ok with the approach.

knylander-grafana (commented):
Thank you for adding doc updates! Doc changes look good.

pkg/traceql/engine.go
mapno: Done in 3ca8ba6ee

pkg/traceql/engine.go
```diff
@@ -99,6 +102,114 @@ func (e *Engine) Execute(ctx context.Context, searchReq *tempopb.SearchRequest,
...
+		}
+
+		for _, ss := range evalSS {
+			for _, s := range ss.Spans {
+				switch attribute.Scope {
+				case AttributeScopeNone: // If tag is unscoped, we need to check all tags by name
+					for attr, v := range s.Attributes() {
+						if attr.Name == attribute.Name {
+							// TODO: Should we just quit if cb returns true?
+							cb(v) // Can't exit early because resource and span scoped tags may have the same name
```
joe-elliott, line 175: why can't we exit early? `cb()` returning false just indicates that the distinct value collector has crossed the limit right? it has nothing to do with span vs resource vs no scope

pkg/traceql/engine.go
```diff
@@ -99,6 +102,114 @@ func (e *Engine) Execute(ctx context.Context, searchReq *tempopb.SearchRequest,
...
+			return nil, err
+		}
+
+		if len(evalSS) == 0 {
+			return nil, nil
+		}
+
+		for _, ss := range evalSS {
+			for _, s := range ss.Spans {
+				switch attribute.Scope {
```
joe-elliott, line 170: instead of switching inside a double loop maybe set a func above that extracts the attribute value?

pkg/traceql/engine_test.go
```diff
@@ -386,3 +386,80 @@ func TestExamplesInEngine(t *testing.T) {
 		})
 	}
 }
+
+func TestExtractMatchers(t *testing.T) {
+	testCases := []struct {
```
joe-elliott, line 391: test multiple spansets and pipelined queries

pkg/traceql/engine.go
```diff
@@ -232,3 +343,48 @@ func (s Static) asAnyValue() *common_v1.AnyValue {
...
+
+// Regex to extract matchers from a query string
+// This regular expression matches a string that contains three groups separated by operators.
+// The first group is a string of alphabetical characters, dots, and underscores.
+// The second group is a comparison operator, which can be one of several possibilities, including =, >, <, and !=.
+// The third group is one of several possible values: a string enclosed in double quotes,
+// a number with an optional time unit (such as "ns", "ms", "s", "m", or "h"),
+// a plain number, or the boolean values "true" or "false".
+// Example: "http.status_code = 200" from the query "{ .http.status_code = 200 && .http.method = }"
+var matchersRegexp = regexp.MustCompile(`[a-zA-Z._]+\s*[=|<=|>=|=~|!=|>|<|!~]\s*(?:"[a-zA-Z._]+"|[0-9smh]+|true|false)`)
```
joe-elliott, line 355: i'm concerned about these regex's being difficult to maintain as we grow this functionality. frankly they may already be difficult to maintain.

i think the right choice is likely an error tolerant traceql parser, but i'm fine with this approach for now.

modules/ingester/instance_search.go
```diff
@@ -350,35 +345,33 @@ func (i *instance) SearchTagValuesV2(ctx context.Context, req *tempopb.SearchTag
 		return distinctValues.Collect(tv)
 	}
 
-	search := func(s common.Searcher, dv *util.DistinctValueCollector[tempopb.TagValue]) error {
-		if s == nil || dv.Exceeded() {
-			return nil
-		}
+	engine := traceql.NewEngine()
 
-		err = s.SearchTagValuesV2(ctx, tag, cb, common.DefaultSearchOptions())
```
joe-elliott, line 358: can we remove this from the interface?

docs/sources/tempo/api_docs/_index.md
```diff
@@ -319,8 +319,16 @@ $ curl -G -s http://localhost:3200/api/search/tag/service.name/values  | jq
...
+
+- `q = (TraceQL query)`: Url encoded [TraceQL query](https://grafana.com/docs/tempo/latest/traceql/).
+  - If a query is provided, results are filtered down to only those traces that match the query.
+  - Queries can be incomplete (eg. `{ .cluster = }`).
+    Tempo will extract only the valid matchers and build a valid query.
+  - Only queries with a single selector `{}` and AND `&&` operators are supported.
+
 ```
-GET /api/v2/search/tag/.service.name/values
+GET /api/v2/search/tag/.service.name/values?q="{.http.method='GET'}"
```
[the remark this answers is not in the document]
mapno: Good point, added.

docs/sources/tempo/api_docs/_index.md
```diff
@@ -319,8 +319,16 @@ $ curl -G -s http://localhost:3200/api/search/tag/service.name/values  | jq
 This endpoint retrieves all discovered values and their data types for the given TraceQL identifier.  The endpoint is available in the query frontend service in
 a microservices deployment, or the Tempo endpoint in a monolithic mode deployment. This endpoint is similar to `/api/search/tag/<tag>/values` but operates on TraceQL identifiers and types. See [TraceQL](../traceql/) documention for more information. The following request returns all discovered service names.
 
+The URL query parameters support the following values:
+
+- `q = (TraceQL query)`: Url encoded [TraceQL query](https://grafana.com/docs/tempo/latest/traceql/).
+  - If a query is provided, results are filtered down to only those traces that match the query.
+  - Queries can be incomplete (eg. `{ .cluster = }`).
+    Tempo will extract only the valid matchers and build a valid query.
+  - Only queries with a single selector `{}` and AND `&&` operators are supported.
```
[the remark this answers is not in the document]
mapno: Done!

modules/ingester/instance_search.go
```diff
@@ -350,35 +345,33 @@ func (i *instance) SearchTagValuesV2(ctx context.Context, req *tempopb.SearchTag
...
-		if err != nil && err != common.ErrUnsupported {
-			return fmt.Errorf("unexpected error searching tag values v2 (%s): %w", tag, err)
-		}
-		return nil
+	// TODO: fallback to old approach if query is not supported or valid
```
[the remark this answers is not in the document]
mapno: Right, it's not. Nice catch.

pkg/api/http.go
```diff
@@ -309,6 +310,37 @@ func ParseSearchBlockRequest(r *http.Request) (*tempopb.SearchBlockRequest, erro
...
 }
 
+func ParseSearchTagValuesRequest(r *http.Request) (*tempopb.SearchTagValuesRequest, error) {
+	vars := mux.Vars(r)
+	tag, ok := vars[tagName]
+	if !ok {
+		return nil, errors.New("please provide a tag")
+	}
+
+	// TODO: Support non-traceql tags
```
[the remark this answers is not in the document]
mapno: We don't. This was from when I was thinking on extending it to the v1 endpoint. No longer relevant, removed.

pkg/traceql/engine.go
```diff
@@ -99,6 +102,114 @@ func (e *Engine) Execute(ctx context.Context, searchReq *tempopb.SearchRequest,
...
+		End:   math.MaxUint32,
+	}
+
+	attribute, err := ParseIdentifier(req.TagName)
+	if err != nil {
+		return err
+	}
+
+	fetchSpansRequest := e.createFetchSpansRequest(searchReq, rootExpr.Pipeline)
+	// TODO: remove other conditions for the attribute we're searching for
```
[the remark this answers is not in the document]
mapno: I was thinking more on requests which search for tag values which are already used in the query we're filtering by.

pkg/traceql/engine.go
mapno: You're right. I misunderstood that return.

pkg/traceql/engine.go
mapno: I've changed the returns to exit the loop properly. Don't mind using a func either

pkg/traceql/engine_test.go
mapno: Added!

pkg/traceql/engine.go
mapno: I share your concerns. I dislike the solution, but it's functional and I think it's the best we've got for now :(.

pkg/traceql/engine.go
```diff
@@ -238,3 +355,48 @@ func (s Static) asAnyValue() *common_v1.AnyValue {
...
+// The third group is one of several possible values: a string enclosed in double quotes,
+// a number with an optional time unit (such as "ns", "ms", "s", "m", or "h"),
+// a plain number, or the boolean values "true" or "false".
+// Example: "http.status_code = 200" from the query "{ .http.status_code = 200 && .http.method = }"
+var matchersRegexp = regexp.MustCompile(`[a-zA-Z._]+\s*[=|<=|>=|=~|!=|>|<|!~]\s*(?:"[a-zA-Z._-]+"|[0-9smh]+|true|false)`)
+
+// TODO: Merge into a single regular expression
+
+// Regex to extract selectors from a query string
+// This regular expression matches a string that contains a single selector and no OR `||` conditions.
```
mdisibio, line 372: Let's change `selector` to `spanset filter` which matches the traceql definitions, and add a few examples of queries that match or don't match.

pkg/traceql/engine.go
```diff
@@ -105,6 +108,120 @@ func (e *Engine) Execute(ctx context.Context, searchReq *tempopb.SearchRequest,
...
+				}
+			}
+			return false
+		}
+
+	evalLoop:
+		for _, ss := range evalSS {
+			for _, s := range ss.Spans {
+				if collectAttributeValue(s) {
+					break evalLoop // Exit early if we have enough data
```
mdisibio, line 196: We can return `io.EOF` to completely stop the Fetch.

mdisibio (approved):
LGTM - Left a few comments, which we discussed offline.

PR #2254, opened 2023-03-23
[TraceQL] Multiple &&ed or ||ed should return unique spans

### LINKED ISSUE #2246: [TraceQL] &&ing spansets into count() will give erroneous results
**Describe the bug**
This issue was discovered while inspecting code for this [PR](https://github.com/grafana/tempo/pull/2245/files).

When performing a spanset operation like this:
```
{<cond>} && {<cond>} && {<cond>}
```
The results of each spanset operation are [simply appended](https://github.com/grafana/tempo/blob/70272b8221ae9456c79b8337984068ecd0a91b34/pkg/traceql/ast_execute.go#L37-L51) to the output spanset. The result is that the same spanset may be included multiple times in the output which will definitely give erroneous results when piping into count and is likely to break other assumptions in the fetch layer.

joe-elliott:
**What this PR does**:
Forces the spanset operations && and || to always return a unique spanset. Before they would return the cross product of all spansets. For instance this query `{ true } && { true }` would return a spanset that was double the input trace. Now it will return a spanset that is equal to the input trace.

Fixes #2246 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/ast_execute_test.go
+++ b/pkg/traceql/ast_execute_test.go
@@ -214,7 +214,6 @@ func TestSpansetOperationEvaluate(t *testing.T) {
 			"{ true } && { true } && { true }",
 			[]*Spanset{
 				{Spans: []Span{
-					// This spanset will be kept because it satisfies both conditions
 					&mockSpan{id: []byte{1}, attributes: map[Attribute]Static{NewAttribute("foo"): NewStaticString("a")}},
 				}},
 			},
@@ -228,7 +227,6 @@ func TestSpansetOperationEvaluate(t *testing.T) {
 			"{ true } || { true } || { true }",
 			[]*Spanset{
 				{Spans: []Span{
-					// This spanset will be kept because it satisfies both conditions
 					&mockSpan{id: []byte{1}, attributes: map[Attribute]Static{NewAttribute("foo"): NewStaticString("a")}},
 				}},
 			},
--- a/pkg/traceql/ast_execute_test.go
+++ b/pkg/traceql/ast_execute_test.go
@@ -1,6 +1,7 @@
 package traceql
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
@@ -181,8 +182,8 @@ func TestSpansetOperationEvaluate(t *testing.T) {
 			},
 			[]*Spanset{
 				{Spans: []Span{
-					&mockSpan{id: []byte{1}, attributes: map[Attribute]Static{NewAttribute("foo"): NewStaticString("a")}},
 					&mockSpan{id: []byte{2}, attributes: map[Attribute]Static{NewAttribute("foo"): NewStaticString("b")}},
+					&mockSpan{id: []byte{1}, attributes: map[Attribute]Static{NewAttribute("foo"): NewStaticString("a")}},
 				}},
 			},
 		},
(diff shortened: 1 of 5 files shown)


tempodb/encoding/vparquet/block_traceql.go
```diff
@@ -295,6 +296,12 @@ func (i *spansetIterator) Next() (*span, error) {
 			}
 		}
 
+		// spans returned from the filter are not guaranteed to be in file order
+		// we need them to be so that the meta iterators work correctly. sort here
+		sort.Slice(i.currentSpans, func(j, k int) bool {
+			return parquetquery.CompareRowNumbers(DefinitionLevelResourceSpans, i.currentSpans[k].rowNum, i.currentSpans[j].rowNum) == 1
```
mdisibio, line 302: Is this different than `compare(j,k) == -1` ?

pkg/traceql/ast_execute_test.go
```diff
@@ -676,3 +703,29 @@ func BenchmarkBinOp(b *testing.B) {
 		})
 	}
 }
+
+// BenchmarkUniquespans benchmarks the performance of the uniqueSpans function using
+// different numbers of spansets and spans.
+func BenchmarkUniqueSpans(b *testing.B) {
```
mdisibio, line 709: Neat, did the benchmarks show a benefit from the current approach?  It seems like it would be the same number of insertions and lookups at first glance.

tempodb/encoding/vparquet/block_traceql.go
joe-elliott: i'm not sure what you mean. is there a faster way to do this?

pkg/traceql/ast_execute_test.go
joe-elliott: I tried building the dictionary using the first, largest and smallest spanset. Smallest outperformed the others. Here is first -> smallest:

```
name                       old time/op    new time/op    delta
UniqueSpans/10000|1-8         152µs ± 1%     169µs ±15%  +11.23%  (p=0.032 n=5+5)
UniqueSpans/100|1-8          2.20µs ±23%    2.25µs ± 5%     ~     (p=0.968 n=5+5)
UniqueSpans/1|1-8             112ns ±12%     112ns ±18%     ~     (p=0.738 n=5+5)
UniqueSpans/10000|100-8       364µs ±15%     345µs ±10%     ~     (p=0.548 n=5+5)
UniqueSpans/100|100-8        12.5µs ±21%    14.5µs ±17%     ~     (p=0.095 n=5+5)
UniqueSpans/1|100-8          9.80µs ±11%    2.01µs ± 7%  -79.47%  (p=0.008 n=5+5)
UniqueSpans/10000|10000-8    1.62ms ±12%    1.60ms ±13%     ~     (p=1.000 n=5+5)
UniqueSpans/100|10000-8      1.45ms ±13%    0.39ms ± 6%  -73.18%  (p=0.008 n=5+5)
UniqueSpans/1|10000-8        1.21ms ± 7%    0.23ms ±25%  -81.21%  (p=0.008 n=5+5)

name                       old alloc/op   new alloc/op   delta
UniqueSpans/10000|1-8         164kB ± 0%     164kB ± 0%     ~     (all equal)
UniqueSpans/100|1-8          1.79kB ± 0%    1.79kB ± 0%     ~     (all equal)
UniqueSpans/1|1-8             32.0B ± 0%     32.0B ± 0%     ~     (all equal)
UniqueSpans/10000|100-8       169kB ± 0%     169kB ± 0%     ~     (p=0.992 n=5+5)
UniqueSpans/100|100-8        8.43kB ± 0%    8.42kB ± 0%   -0.02%  (p=0.000 n=4+5)
UniqueSpans/1|100-8          7.02kB ± 0%    1.79kB ± 0%     ~     (p=0.079 n=4+5)
UniqueSpans/10000|10000-8    1.00MB ± 0%    1.00MB ± 0%     ~     (p=0.548 n=5+5)
UniqueSpans/100|10000-8       840kB ± 0%     169kB ± 0%  -79.87%  (p=0.008 n=5+5)
UniqueSpans/1|10000-8         840kB ± 0%     164kB ± 0%  -80.49%  (p=0.008 n=5+5)

name                       old allocs/op  new allocs/op  delta
UniqueSpans/10000|1-8          1.00 ± 0%      1.00 ± 0%     ~     (all equal)
UniqueSpans/100|1-8            1.00 ± 0%      1.00 ± 0%     ~     (all equal)
UniqueSpans/1|1-8              1.00 ± 0%      1.00 ± 0%     ~     (all equal)
UniqueSpans/10000|100-8        9.00 ± 0%      9.00 ± 0%     ~     (all equal)
UniqueSpans/100|100-8          9.00 ± 0%      9.00 ± 0%     ~     (all equal)
UniqueSpans/1|100-8            9.00 ± 0%      1.00 ± 0%  -88.89%  (p=0.008 n=5+5)
UniqueSpans/10000|10000-8       214 ± 0%       214 ± 0%     ~     (all equal)
UniqueSpans/100|10000-8         214 ± 0%         9 ± 0%     ~     (p=0.079 n=4+5)
UniqueSpans/1|10000-8           214 ± 0%         1 ± 0%  -99.53%  (p=0.008 n=5+5)
```

tempodb/encoding/vparquet/block_traceql.go
joe-elliott: oh. understood. i agree that the suggested ordering is more clear.

mdisibio (approved):

PR #2261, opened 2023-03-27
Add job & instance labels to span metrics, a new target_info metrics, and the ability to customize dimension label mapping

ie-pham:
**What this PR does**: Add the `service` label to `job` in span metrics and add `instance` as additional intrinsic dimensions. Add feature to allow custom dimension label mapping. 

Dimension Mapping config:
```
span_metrics+: {
          dimensions: [],
          dimension_mappings: [
            { // label would be "vulture-zero" and the value would be "<vulture-0>"
              name: 'vulture-zero',
              source_labels: ['vulture-0'],
              join: '/'
            },
            { // label would be "vulture-index" and the value would be "<vulture-0>/<vulture-1>"
              name: 'vulture-index',
              source_labels: ['vulture-0', 'vulture-1'],
              join: '/'
            },
          ],
        },
```

Decisions for target_info/job/instance:
https://github.com/grafana/tempo/pull/2261#discussion_r1151153721

> 1. If service.name is present, then job is present. If not, the label is skipped. If ONLY service.namespace is present and not service.name, then also job is skipped.
> 2. If service.instance.id is present then instance is present. If not, the label is skipped.
> 3. If there are NO resource labels other than the job and instance labels, then the target_info is skipped entirely.

>* target_info should contain all resource attributes + job + instance

**Make sure this is a non-breaking change**

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -64,6 +69,7 @@ func New(cfg Config, registry registry.Registry) gen.Processor {
 		spanMetricsCallsTotal:      registry.NewCounter(metricCallsTotal, labels),
 		spanMetricsDurationSeconds: registry.NewHistogram(metricDurationSeconds, labels, cfg.HistogramBuckets),
 		spanMetricsSizeTotal:       registry.NewCounter(metricSizeTotal, labels),
+		spanMetricsTargetInfo:      registry.NewCounter(targetInfo, labels),
```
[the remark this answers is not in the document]
mapno: Also, can't have its labels statically defined, since it'll depend on what resource attributes are present.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -125,10 +134,12 @@ func (p *Processor) aggregateMetricsForSpan(svcName string, rs *v1.Resource, spa
 
 	p.spanMetricsCallsTotal.Inc(registryLabelValues, 1*spanMultiplier)
 	p.spanMetricsSizeTotal.Inc(registryLabelValues, float64(span.Size())*spanMultiplier)
+	p.spanMetricsTargetInfo.Inc(registryLabelValues, 0)
```
mapno, line 137: It shouldn't use the same labels as the rest. `target_info` should contain all resource attributes.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -84,24 +90,24 @@ func (p *Processor) Shutdown(_ context.Context) {
 
 func (p *Processor) aggregateMetrics(resourceSpans []*v1_trace.ResourceSpans) {
 	for _, rs := range resourceSpans {
-		// already extract service name, so we only have to do it once per batch of spans
-		svcName, _ := processor_util.FindServiceName(rs.Resource.Attributes)
-
+		// already extract job name & instance id, so we only have to do it once per batch of spans
+		jobName := processor_util.GetJobValue(rs.Resource.Attributes)
+		instanceID, _ := processor_util.FindInstanceID(rs.Resource.Attributes)
```
mapno, line 95: I think we should handle the case when there is no job and instance ID. In that case, `target_info` can't be matched to the rest of the metrics.

mapno (commented):
Left a few comments. Please correct me if I'm wrong, I'm no expert in the requirements for `target_info`.

jpkrohling:
Is this change drifting this component further away from the upstream component? Is there an intention of converging them into one?

modules/generator/processor/spanmetrics/spanmetrics.go
ie-pham: so right now if either of them are blank, it is set to "". What do you suggest should happen instead?

modules/generator/processor/util/util.go
```diff
@@ -36,3 +45,11 @@ func GetSpanMultiplier(ratioKey string, span *v1.Span) float64 {
 	}
 	return spanMultiplier
 }
+
+func GetJobValue(attributes []*v1_common.KeyValue) string {
+	svName, _ := FindServiceName(attributes)
```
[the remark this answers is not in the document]
ie-pham: So if there is namespace but no service, we will set job to just namespace?

modules/generator/processor/spanmetrics/spanmetrics.go
ie-pham: ok just for clarity, we want _all_ attributes including the ones that the users are not asking for? Isn't that very very high cardinality?

knylander-grafana:
@ie-pham  Do we need documentation for this? It looks like new features that need content.

modules/generator/processor/util/util.go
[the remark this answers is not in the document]
cyrille-leclerc: @kovrus what was the choice on the OTel Collector Span Metrics Connector? cc @jpkrohling

modules/generator/processor/spanmetrics/spanmetrics.go
cyrille-leclerc: @kovrus what was the choice on the OTel Collector Span Metrics Connector? cc @jpkrohling 
I guess there is no value in creating a `target_info` if `service.instance.id` is missing

modules/generator/processor/spanmetrics/spanmetrics.go
mapno: Sorry, I'm not sure. That was what I thought, but it's possible I'm very wrong, so I'll just step back and let the owners clarify this.

modules/generator/processor/spanmetrics/spanmetrics.go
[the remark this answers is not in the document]
mapno: cc/ @gouthamve

ie-pham:
> @ie-pham Do we need documentation for this? It looks like new features that need content.

yes we do, can you point me to where i need to update this?

modules/generator/processor/util/util.go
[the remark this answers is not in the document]
jpkrohling: The upstream processor assumes that the service name is always present, given it's a required value for resource attributes: https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/#service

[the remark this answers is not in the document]
jpkrohling: Looking at the upstream code, it looks like there's one case check for this, which will then just skip the data point if the service name is empty:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/5f5770535d75a51ca07c6ea4e19906b0fa7add9c/processor/spanmetricsprocessor/processor.go#L346-L349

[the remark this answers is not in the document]
ie-pham: empty service name has been handled in the code

modules/generator/processor/spanmetrics/spanmetrics.go
ie-pham: @gouthamve ^^

modules/generator/processor/spanmetrics/spanmetrics.go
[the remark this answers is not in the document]
gouthamve: Its _okay_ if its a counter with value 1. Best if its gauge, but because Prometheus is weakly typed, it'll be fine.

modules/generator/processor/spanmetrics/spanmetrics.go
gouthamve: > Isn't that very very high cardinality?

The cardinality is driven by `job` and `instance`. Adding additional resource labels won't add extra cardinality as each (`job`, `instance`) combination is unique. There won't be two sets of resource labels with the same job, instance but different resource labels.

modules/generator/processor/spanmetrics/spanmetrics.go
gouthamve: Please note that this drifting towards the connector and not the processor. 

In the connector, not having a `service.instance.id` will mean that `target_info` is indeed generated. However, I don't think there is any point to it. I am okay to sync with upstream behavior:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/5f5770535d75a51ca07c6ea4e19906b0fa7add9c/pkg/translator/prometheusremotewrite/helper.go#L188-L205

1. If `service.name` is present, then `job` is present. If not, the label is skipped. If ONLY `service.namespace` is present and not `service.name`, then also `job` is skipped.
2. If `service.instance.id` is present then `instance` is present. If not, the label is skipped.

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/translator/prometheusremotewrite/helper.go#L536-L550

If there are NO resource labels other than the `job` and `instance` labels, then the `target_info` is skipped entirely.


-------------

@kovrus What do you think about changing the upstream behavior? Skip `target_info` generation if `service.instance.id` is not present?

gouthamve:
Also this will cause an increase in cardinality because we are not including `service.instance.id` in the labels. It should be documented clearly.

jpkrohling (approved):
I'm happy with the changes, as I understand they will indeed bring this closer to the expected pipelines in upstream (otelcol with span to metrics connector and Prometheus exporter)

modules/generator/processor/spanmetrics/spanmetrics.go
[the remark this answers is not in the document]
ie-pham: @gouthamve does it have to be 1 or can it be 0?

modules/generator/processor/spanmetrics/config.go
```diff
@@ -10,49 +10,60 @@ import (
...
+
+type DimensionMappings struct {
+	Name        string   `yaml:"name"`
+	SourceLabel []string `yaml:"source_labels"`
+	Join        string   `yaml:"join"`
 }
 
 type IntrinsicDimensions struct {
-	Service       bool `yaml:"service"`
+	Job           bool `yaml:"job"`
```
joe-elliott, line 54: does this impact existing metrics? we should not be make breaking changes to existing metrics w/o figuring out how we would roll that out to cloud.

ie-pham:
@gouthamve @jpkrohling @kovrus can the user disable job and instance?

modules/generator/processor/util/util.go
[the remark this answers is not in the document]
kovrus: See the specification https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/#service . `service.name` is the **required** resource attribute. The OTel span metrics connector and processor drop spans resource scope completely if the service name is missing.

kovrus:
> @gouthamve @jpkrohling @kovrus can the user disable job and instance?

What is the motivation behind that?

Those are generated by Prometheus remote write, it is not possible to disable it manually, but if the `service.name` or `service.instance.id` is missing then they won't be created https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/translator/prometheusremotewrite/helper.go#L188-L205. When using span metrics connector with prom remote write there always be `service.name` generated (if not, those points will be dropped).

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -20,6 +20,7 @@ const (
 	metricCallsTotal      = "traces_spanmetrics_calls_total"
 	metricDurationSeconds = "traces_spanmetrics_latency"
 	metricSizeTotal       = "traces_spanmetrics_size_total"
+	targetInfo            = "traces_spanmetrics_target_info"
```
kovrus, line 23: Should that be just `target_info`? 

OTel Upstream: span metrics connector + prometheus remote write will generate just `target_info`.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -84,21 +109,28 @@ func (p *Processor) Shutdown(_ context.Context) {
...
 	for _, rs := range resourceSpans {
-		// already extract service name, so we only have to do it once per batch of spans
+		// already extract job name & instance id, so we only have to do it once per batch of spans
 		svcName, _ := processor_util.FindServiceName(rs.Resource.Attributes)
-
+		jobName := processor_util.GetJobValue(rs.Resource.Attributes)
+		instanceID, _ := processor_util.FindInstanceID(rs.Resource.Attributes)
+		if p.Cfg.EnableTargetInfo {
+			resourceLabels := processor_util.GetTargetInfoAttributes(rs.Resource.Attributes)
+			p.targetInfoLabels = append(p.targetInfoLabels, resourceLabels...)
```
kovrus, line 118: are we adding `service.name`, `service.namespace`, and `service.instance.id` as target_info labels?

kovrus:
@ie-pham is it possible to add some tests for the target info generation?

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -50,10 +50,14 @@ The following metrics are exported:
...
 
-By default, the metrics processor adds the following labels to each metric: `service`, `span_name`, `span_kind`, `status_code`, `status_message`.
+By default, the metrics processor adds the following labels to each metric: `service`, `span_name`, `span_kind`, `status_code`, `status_message`, `job`, `instance`.
 Additional user defined labels can be created using the [`dimensions` configuration option]({{< relref "../configuration/#metrics-generator" >}}).
 When a configured dimension collides with one of the default labels (e.g. `status_code`), the label for the respective dimension is prefixed with double underscore (i.e. `__status_code`).
 
+Custom labeling of dimensions is also supported using the [`dimension_mapping` configuration option]({{< relref "../configuration/#metrics-generator" >}}).
+
+An optional metric called `traces_spanmetrics_target_info` using all resource level attributes as dimensions can be enabled in the [`enable_target_info` configuration option]({{< relref "../configuration/#metrics-generator" >}}).
```
gouthamve, line 59: Why not `target_info`?

ie-pham:
> @ie-pham is it possible to add some tests for the target info generation?

@kovrus tests have already been added to "span_metrics_test"

modules/overrides/limits.go
```diff
@@ -57,18 +57,20 @@ type Limits struct {
...
+	MetricsGeneratorProcessorServiceGraphsDimensions        []string            `yaml:"metrics_generator_processor_service_graphs_dimensions" json:"metrics_generator_processor_service_graphs_dimensions"`
+	MetricsGeneratorProcessorSpanMetricsHistogramBuckets    []float64           `yaml:"metrics_generator_processor_span_metrics_histogram_buckets" json:"metrics_generator_processor_span_metrics_histogram_buckets"`
+	MetricsGeneratorProcessorSpanMetricsDimensions          []string            `yaml:"metrics_generator_processor_span_metrics_dimensions" json:"metrics_generator_processor_span_metrics_dimensions"`
+	MetricsGeneratorProcessorSpanMetricsIntrinsicDimensions map[string]bool     `yaml:"metrics_generator_processor_span_metrics_intrinsic_dimensions" json:"metrics_generator_processor_span_metrics_intrinsic_dimensions"`
+	MetricsGeneratorProcessorSpanMetricsDimensionMappings   []map[string]string `yaml:"metrics_generator_processor_span_metrics_dimension_mappings" json:"metrics_generator_processor_span_metrics_dimension_mapings"`
```
zalegrala, line 72: Good additions.  Do you want to add a test for these two?

zalegrala (commented):
This looks good to me.  Nice work.  A couple comments but otherwise I'm +1 here.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -115,17 +143,93 @@ func (p *Processor) aggregateMetricsForSpan(svcName string, rs *v1.Resource, spa
...
+		} else {
+			p.targetInfoLabels = removeLabel(dimInstance, p.targetInfoLabels)
+		}
+	}
+
+	//update labels to remove job and/or instance if they're blank
+	p.spanMetricsCallsTotal.UpdateLabels(p.labels)
+	p.spanMetricsDurationSeconds.UpdateLabels(p.labels)
+	p.spanMetricsSizeTotal.UpdateLabels(p.labels)
+	p.spanMetricsTargetInfo.UpdateLabels(p.targetInfoLabels)
```
mapno, line 182: Not sure if this is thread-safe. Batches can be concurrently pushed to a processor. If a set of traces has `instance_id`, and other doesn't, how does that affect all the labels that are being mutated in the metrics and the processor itself (eg. line 174)?

ie-pham: good catch! i made some changes and added tests for this scenario

modules/generator/registry/counter.go
```diff
@@ -23,6 +21,7 @@ type counter struct {
 }
 
 type counterSeries struct {
+	labels []string
```
mapno, line 24: Nit: in Prometheus, the key-value pair is as a whole called `label`. We could use that terminology and simplify the two slices into one.

modules/generator/registry/gauge.go
```diff
@@ -0,0 +1,188 @@
...
+
+func (g *gauge) Set(labelValueCombo *LabelValueCombo, value float64) {
+	g.change(labelValueCombo, value, set)
+}
+
+func (g *gauge) Inc(labelValueCombo *LabelValueCombo, value float64) {
+	g.change(labelValueCombo, value, add)
+}
+
+func (g *gauge) change(labelValueCombo *LabelValueCombo, value float64, operation string) {
```
mapno, line 78: `change` is a strange name for this func. What about `updateSeries` instead, and `series` can have its own `update` method?

mapno (approved):
I'm happy with the PR. The concurrency problem is now fixed and I think it's ready to be merged.

Left two code-style comments, but good to go otherwise.

ie-pham:
We will need to communicate with users that this may increase their cardinality due to job and instance being added

modules/generator/registry/counter.go
```diff
@@ -126,7 +119,7 @@ func (c *counter) collectMetrics(appender storage.Appender, timeMs int64, extern
 
 	activeSeries = len(c.series)
 
-	lbls := make(labels.Labels, 1+len(externalLabels)+len(c.labels))
+	lbls := make(labels.Labels, 1+len(externalLabels)+4)
```
zalegrala, line 122: Where does the 4 come from here?  Might be worth a comment on the code.

modules/generator/processor/util/util.go
```diff
@@ -11,6 +11,15 @@ func FindServiceName(attributes []*v1_common.KeyValue) (string, bool) {
 	return FindAttributeValue(semconv.AttributeServiceName, attributes)
 }
 
+func FindServiceNamespace(attributes []*v1_common.KeyValue) (string, bool) {
+	return FindAttributeValue(semconv.AttributeServiceNamespace, attributes)
+}
+
+func FindInstanceID(attributes []*v1_common.KeyValue) (string, bool) {
+	string, bool := FindAttributeValue(semconv.AttributeServiceInstanceID, attributes)
+	return string, bool
```
zalegrala, line 20: Could return `FindAttributeValue()` here.

zalegrala (approved):
This looks good to me, nice work here.  We'll be adding two additional labels to spanmetrics when the conditions are met within the span, and it will be interesting to see how this performs when we put some load on it, but improvements in that area can be made over time.

modules/generator/processor/util/util.go
ie-pham: fixed!

modules/generator/registry/counter.go
ie-pham: good eye! I removed the line but @kvrhdn maybe has more context on whether that line is needed.

kvrhdn: In the original code we are preallocating the slice used by `labels.Builder` to ensure `lb.Set` doesn't have to expand the underlying array.

https://github.com/grafana/tempo/blob/089772bba3c50293ef3ac3fa615e7e1ff8a9ad73/modules/generator/registry/counter.go#L129-L130

I'm guessing the 4 must be something you set during testing. The sum should be:
- `1`: `__name__` (the metric name, always present)
- `len(externalLabels)`: constant labels injected by the registry config
- `len(c.labels)`: labels present on the counter itself

Since every serie can now have a different number of labels, it will be difficult to figure out that last value. You could calculate which series has the most labels, but iterating through all series is not going to be worth it.

I see the current code doesn't preallocate a slice anymore (it does `labels.NewBuilder(nil)`). This will still work but it might we might reallocate the slice several times.

Instead I'd suggest allocating at least enough space for the metric name and `externalLabels`, this data is easy to get and it's an easy win. As we are iterating through the series the labels slice might have to reallocated when it stumbles on a serie with more labels but that should only happen a couple of time.

kvrhdn (commented):
Did a light review, besides following comments seems okay to me :+1:

docs/sources/tempo/configuration/_index.md
```diff
@@ -297,6 +297,14 @@ metrics_generator:
             # the metrics if present.
             [dimensions: <list of string>]
 
+            # Custom labeling of dimensions is possible via a list of maps consisting of 
+            # "name" <string>, "source_labels" <list of string>, "join" <string> 
+            # "name" will be what appears in the metrics, "source_labels" are the actual
```
knylander-grafana, line 302:
```suggestion
            # "name" appears in the metrics, "source_labels" are the actual
```

knylander-grafana (commented):
Thank you for adding docs. I made a minor suggestion, and otherwise looks great.

PR #2265, opened 2023-03-27
Update docker image to run as non-root

### LINKED ISSUE #334: Allow for images to run as non-root
In the baseline: security. It could pose as a serious security concern.
Do we need to run as root? No.

Especially in enterprise environment, we are simply not even allowed to run images as root. Therefore I would like to see a fix.

Add users to the docker images, perhaps change a few paths (i.e. placement of the executables).

The compactor can be used with the following workaround:

```
        volumeMounts:
        - mountPath: /conf
          name: tempo-conf
        - mountPath: /var/tempo
          name: tempo  
      volumes:
      - name: tempo
        emptyDir: {}

```
By adding an emptydir we override the /var/tempo folder as non-root. 

For the querier we can do the same "trick" for the /var/tempo folder BUT it also wants to write and use files in the `/tmp` folder. This is an issue because we cannot mount an empty dir on the `/tmp` folder. This is because the original `/tmp` folder contains the `tempo-query` executable.

The real dodgy workaround is to build an image with `tempo-query` in a different place. I.e.: 

```
FROM  tempo-query:latest
COPY tempo-query /tempo-query 
```

Then use an initcontainer:

```
      initContainers:
      - name: please-no
        image: busybox
        command: ['sh', '-c', 'ln -s /tempo-query /tmp/tempo-query']
        volumeMounts:
          - name: tmp
            mountPath: /tmp
```

while having the following emptydir mounts:

```
        volumeMounts:
        - mountPath: /tmp
          name: tmp
        - mountPath: /conf
          name: tempo-query-conf         
      volumes:
      - name: tempo
        emptyDir: {}
      - name: tmp
        emptyDir: {}
```

**Additional context**

The workaround is fine for a proof of concept, but not production usage :)

zalegrala:
**What this PR does**:

**BREAKING CHANGE**

Here we make the image adjustment necessary to run Tempo as non-root in the
Docker container, as well as include some util jsonnet to allow statefulsets to
chown their data directories to match the new permissions.

With this new chown init contianer, both the ingester and metrics-generator
statefulsets start and function.

```
❯ docker run -it --entrypoint sh zalegrala/tempo:tempoNonRootImage-eefae10f9
/ $ whoami
tempo
/ $ id
uid=10001(tempo) gid=10001(tempo) groups=10001(tempo)
/ $ ls -ld /var/tempo
drwxr-xr-x    2 tempo    tempo         4096 Apr  2 19:40 /var/tempo
```

Note that the `securityContext.fsGroup(10001)` may be required for environments
that mount additional volumes which do not have read/write permissions for the
`tempo` user.  Users may also wish to recursively `chown` the `/var/tempo`
directory for the new ownership.  This will need to be done only once.

Fixes #334

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 * [ENHANCEMENT] Add querier metrics for requests executed [#3524](https://github.com/grafana/tempo/pull/3524) (@electron0zero)
 * [FEATURE] Added gRPC streaming endpoints for all tag queries. [#3460](https://github.com/grafana/tempo/pull/3460) (@joe-elliott)
 * [CHANGE] Align metrics query time ranges to the step parameter [#3490](https://github.com/grafana/tempo/pull/3490) (@mdisibio)
+* [CHANGE] Change the UID and GID of the `tempo` user to avoid root [#2265](https://github.com/grafana/tempo/pull/2265) (@zalegrala)
+  **BREAKING CHANGE** Ownership of /var/tempo is changing.  Specifically the
+  ingester and metrics-generator statefulsets may need to be `chown`'d in order
+  to come up properly.  A jsonnet example of an init container is included with
+  the PR.
 * [ENHANCEMENT] Add string interning to TraceQL queries [#3411](https://github.com/grafana/tempo/pull/3411) (@mapno)
 * [ENHANCEMENT] Add new (unsafe) query hints for metrics queries [#3396](https://github.com/grafana/tempo/pull/3396) (@mdisibio)
 * [ENHANCEMENT] Add nestedSetLeft/Right/Parent instrinsics to TraceQL. [#3497](https://github.com/grafana/tempo/pull/3497) (@joe-elliott)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,12 @@
 * [FEATURE] Added gRPC streaming endpoints for all tag queries. [#3460](https://github.com/grafana/tempo/pull/3460) (@joe-elliott)
 * [CHANGE] Align metrics query time ranges to the step parameter [#3490](https://github.com/grafana/tempo/pull/3490) (@mdisibio)
 * [CHANGE] Change the UID and GID of the `tempo` user to avoid root [#2265](https://github.com/grafana/tempo/pull/2265) (@zalegrala)
-  **BREAKING CHANGE** Ownership of /var/tempo is changing.  Specifically the
-  ingester and metrics-generator statefulsets may need to be `chown`'d in order
-  to come up properly.  A jsonnet example of an init container is included with
-  the PR.
+  **BREAKING CHANGE** Ownership of /var/tempo is changing.  Historyically this
+  has been owned by root:root, and with this change it will now be owned by
+  tempo:tempo with the UID/GID of 10001.  The `ingester` and
+  `metrics-generator` statefulsets may need to be `chown`'d in order to start
+  properly.  A jsonnet example of an init container is included with the PR.
+  This impacts impacts all users of the `grafana/tempo` Docker image.
 * [ENHANCEMENT] Add string interning to TraceQL queries [#3411](https://github.com/grafana/tempo/pull/3411) (@mapno)
 * [ENHANCEMENT] Add new (unsafe) query hints for metrics queries [#3396](https://github.com/grafana/tempo/pull/3396) (@mdisibio)
 * [ENHANCEMENT] Add nestedSetLeft/Right/Parent instrinsics to TraceQL. [#3497](https://github.com/grafana/tempo/pull/3497) (@joe-elliott)
--- a/operations/jsonnet/microservices/common.libsonnet
+++ b/operations/jsonnet/microservices/common.libsonnet
@@ -7,16 +7,16 @@
   local volumeMount = k.core.v1.volumeMount,
   local statefulset = k.apps.v1.statefulSet,
 
-  tempo_chown_container(data_volume, uid)::
+  tempo_chown_container(data_volume, uid='10001', gid='10001', dir='/var/tempo')::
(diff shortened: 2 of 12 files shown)


operations/jsonnet/microservices/common.libsonnet
```diff
@@ -1,4 +1,27 @@
 {
+
+  local k = import 'k.libsonnet',
+  local kausal = import 'ksonnet-util/kausal.libsonnet',
+
+  local container = k.core.v1.container,
+  local volumeMount = k.core.v1.volumeMount,
+  local statefulset = k.apps.v1.statefulSet,
+
+  tempo_chown_container(data_volume, uid)::
```
zalegrala, line 10: Add argument to specify the directory.

operations/jsonnet/microservices/common.libsonnet
```diff
@@ -1,4 +1,27 @@
...
+    container.new('chown-' + data_volume, $._images.tempo) +
+    container.withCommand('chown') +
+    container.withArgs([
+      '-R',
+      uid,
```
zalegrala, line 15: Change the GID also.

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 * [ENHANCEMENT] Add querier metrics for requests executed [#3524](https://github.com/grafana/tempo/pull/3524) (@electron0zero)
 * [FEATURE] Added gRPC streaming endpoints for all tag queries. [#3460](https://github.com/grafana/tempo/pull/3460) (@joe-elliott)
 * [CHANGE] Align metrics query time ranges to the step parameter [#3490](https://github.com/grafana/tempo/pull/3490) (@mdisibio)
+* [CHANGE] Change the UID and GID of the `tempo` user to avoid root [#2265](https://github.com/grafana/tempo/pull/2265) (@zalegrala)
```
zalegrala, line 6: Mention this data used by owned by root.

kvrhdn (approved):
This looks good to me, nice work cleaning up some secuirty debt.

I'm fine with merging but I think @mdisibio was going to leave a comment too.

electron0zero (approved):

CHANGELOG.md
zalegrala: Have done.

zalegrala:
Thanks for the review folks.

@mdisibio I pushed a commit for your feedback on the jsonnet.  Feel free to call out anything else you'd like to see addressed here.

zalegrala:
I caught up with mdisibio earlier and we're good on the changes here.

gangstead:
For anyone else whose helm release broke when this was released (chart version 1.9.0).

The correct place to add the securityContext at the container level in the chart is here: https://github.com/grafana/helm-charts/blob/main/charts/tempo/values.yaml#L234

```yaml
    securityContext:
      fsGroup: 10001
```

zalegrala:
Thanks for the note @gangstead.  I checked the `tempo-distributed` chart before, but not the `tempo` chart, so apologies for the trouble.  I think you will want more than `fsGroup` on your pods.  The change would be something similar to what is in the `tempo-distributed` chart here: https://github.com/grafana/helm-charts/blob/main/charts/tempo-distributed/values.yaml#L68-L70

StefanLobbenmeierObjego:
In understood this as add this to the tempo values:

```yaml
securityContext:
  runAsNonRoot: true
  runAsUser: 1000
  runAsGroup: 1000
```

but actually I have to make sure to use the UID/GID from a previous release? Or use 10001?

StefanLobbenmeierObjego:
Combining both and using 10001 seems to work:

```
securityContext:
  fsGroup: 10001
  runAsNonRoot: true
  runAsUser: 10001
  runAsGroup: 10001
```

gangstead:
It worked for me with just `fsGroup: 10001`.  I think this PR already sets the user/group with the `USER 10001:10001` command in the dockerfile.  

I'm not really sure on the interaction between the  `runAs*` in the k8s `securityContext` and the `user` command in the dockerfile.  They seem to be two ways of setting the user / group but I don't know if there's any subtle distinction.

zalegrala:
If only `fsGroup` is used, I expect that the reads would be functional because of the `umask`, but writes and deletes would not.  A one-time `chown` of the data is also an approach to drop the security context handling entirely.  A successul read of the local data is enough to let the ingester start.

Is the ingester writing data locally successfully after using your `securityContext` updates? @StefanLobbenmeierObjego @gangstead

gangstead:
Yes.  Looking in Grafana there are new traces being saved.  Maybe I had some data loss and just didn't realize it?  We only store traces for 36 hours due to the huge volume of storage it takes.  It's too late for me to go back and verify. Losing some traces is fine in our case.

StefanLobbenmeierObjego:
> Is the ingester writing data locally successfully after using your securityContext updates?

yeah I also had to issues with my tempo configuration

StefanLobbenmeierObjego:
Actually, no, something is off. I can see the traces from today without issues, but no traces from past days. I am assuming the writing to the disk fails but it still shows some cached in memory?

StefanLobbenmeierObjego:
Nevermind, is seems that was just a hickup. I tried again now and I can see traces from today and traces from up to 14 days ago (which is the limit we configured). So everything is working, sorry for the confusion

pharaujo:
While testing Tempo v2.5.0 with `runAsNonRoot` in our incubator environment, I noticed `vulture` was missed: https://github.com/grafana/tempo/blob/v2.5.0/cmd/tempo-vulture/Dockerfile

zalegrala:
Thanks @pharaujo, would you like to PR a change?

zalegrala:
@StefanLobbenmeierObjego Would you like to PR the `tempo` chart for this update?

PR #2274, opened 2023-03-29
[metrics-generator] filter out spans based on policy

### LINKED ISSUE #1482: Metrics-generator should filter out spans based upon attributes
If the metrics-generator is enabled for a tenant, it will generate metrics for all traces that are ingested. In some cases it would be nice if we could only generate metrics for a subset of spans.

Example:
- don't collect metrics for traces that don't have the attribute `http.code`
- only generate metrics for spans from service XXX

It should be possible to configure filters that include/exclude spans based upon their attributes.

There is no alternative AFAIK. You could drop the metrics when remote writing them, but that is definitely not ideal.

**Additional context**

zalegrala:
**What this PR does**:

Here we implement an approach to filtering out spans based on a policy, loosely based around the OTEL collector filterspan config format. 

Fixes #1482 

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/cmd/tempo/app/config.go
+++ b/cmd/tempo/app/config.go
@@ -186,6 +186,13 @@ func (c *Config) CheckConfig() []ConfigWarning {
 			if a.Scope == traceql.AttributeScopeNone {
 				warnings = append(warnings, newMetricsGeneratorProcessorSpanMetricsFilterPoliciesWarning(fmt.Sprintf("invalid attribute scope: %s", attr.Key)))
 			}
+
+			switch a.Intrinsic.String() {
+			case "name", "status", "kind": // currently supported
+			default:
+				warnings = append(warnings, newMetricsGeneratorProcessorSpanMetricsFilterPoliciesWarning(fmt.Sprintf("currently unsupported intrinsic: %s", a.Intrinsic)))
+			}
+
 		}
 	}
 
@@ -256,6 +263,6 @@ func newV2Warning(setting string) ConfigWarning {
 func newMetricsGeneratorProcessorSpanMetricsFilterPoliciesWarning(setting string) ConfigWarning {
 	return ConfigWarning{
 		Message: "c.Generator.Processor.SpanMetrics.FilterPolicies attribute is invalid: " + setting,
-		Explain: "FilterPolicy attributes must be `resource` or `span` scopped.",
+		Explain: "FilterPolicy attributes must be `resource` or `span` scopped, or a supported intrinic attribute",
 	}
 }
--- a/cmd/tempo/app/config.go
+++ b/cmd/tempo/app/config.go
@@ -192,7 +192,6 @@ func (c *Config) CheckConfig() []ConfigWarning {
 			default:
 				warnings = append(warnings, newMetricsGeneratorProcessorSpanMetricsFilterPoliciesWarning(fmt.Sprintf("currently unsupported intrinsic: %s", a.Intrinsic)))
 			}
-
 		}
 	}
 
--- a/cmd/tempo/app/config.go
+++ b/cmd/tempo/app/config.go
@@ -239,10 +239,3 @@ func newV2Warning(setting string) ConfigWarning {
 		Explain: "This setting is only used in v2 blocks",
 	}
 }
-
-func newMetricsGeneratorProcessorSpanMetricsFilterPoliciesWarning(setting string) ConfigWarning {
-	return ConfigWarning{
-		Message: "c.Generator.Processor.SpanMetrics.FilterPolicies attribute is invalid: " + setting,
-		Explain: "FilterPolicy attributes must be `resource` or `span` scopped, or a supported intrinic attribute",
-	}
-}
--- a/cmd/tempo/app/config.go
+++ b/cmd/tempo/app/config.go
@@ -18,7 +18,6 @@ import (
 	"github.com/grafana/tempo/modules/querier"
 	"github.com/grafana/tempo/modules/storage"
 	internalserver "github.com/grafana/tempo/pkg/server"
-	"github.com/grafana/tempo/pkg/traceql"
 	"github.com/grafana/tempo/pkg/usagestats"
 	"github.com/grafana/tempo/pkg/util"
 	"github.com/grafana/tempo/tempodb"
@@ -175,24 +174,6 @@ func (c *Config) CheckConfig() []ConfigWarning {
 
 	if c.StorageConfig.Trace.Block.Version != "v2" && c.Compactor.Compactor.IteratorBufferSize != tempodb.DefaultIteratorBufferSize {
 		warnings = append(warnings, newV2Warning("v2_prefetch_traces_count"))
-	}
-
-	for _, x := range c.Generator.Processor.SpanMetrics.FilterPolicies {
-		for _, attr := range x.Include.Attributes {
-			a, w := traceql.ParseIdentifier(attr.Key)
-			if w != nil {
-				warnings = append(warnings, newMetricsGeneratorProcessorSpanMetricsFilterPoliciesWarning(w.Error()))
-			}
-			if a.Scope == traceql.AttributeScopeNone {
-				warnings = append(warnings, newMetricsGeneratorProcessorSpanMetricsFilterPoliciesWarning(fmt.Sprintf("invalid attribute scope: %s", attr.Key)))
-			}
-
-			switch a.Intrinsic.String() {
-			case "name", "status", "kind": // currently supported
-			default:
-				warnings = append(warnings, newMetricsGeneratorProcessorSpanMetricsFilterPoliciesWarning(fmt.Sprintf("currently unsupported intrinsic: %s", a.Intrinsic)))
-			}
-		}
 	}
 
 	return warnings
--- a/cmd/tempo/app/config.go
+++ b/cmd/tempo/app/config.go
@@ -18,6 +18,7 @@ import (
 	"github.com/grafana/tempo/modules/querier"
 	"github.com/grafana/tempo/modules/storage"
 	internalserver "github.com/grafana/tempo/pkg/server"
+	"github.com/grafana/tempo/pkg/traceql"
 	"github.com/grafana/tempo/pkg/usagestats"
 	"github.com/grafana/tempo/pkg/util"
 	"github.com/grafana/tempo/tempodb"
@@ -176,6 +177,18 @@ func (c *Config) CheckConfig() []ConfigWarning {
 		warnings = append(warnings, newV2Warning("v2_prefetch_traces_count"))
 	}
 
+	for _, x := range c.Generator.Processor.SpanMetrics.FilterPolicies {
+		for _, attr := range x.Include.Attributes {
+			a, w := traceql.ParseIdentifier(attr.Key)
+			if w != nil {
+				warnings = append(warnings, newMetricsGeneratorProcessorSpanMetricsFilterPoliciesWarning(w.Error()))
+			}
+			if a.Scope == traceql.AttributeScopeNone {
+				warnings = append(warnings, newMetricsGeneratorProcessorSpanMetricsFilterPoliciesWarning(fmt.Sprintf("invalid attribute scope: %s", attr.Key)))
+			}
+		}
+	}
+
 	return warnings
 }
 
@@ -237,5 +250,12 @@ func newV2Warning(setting string) ConfigWarning {
 	return ConfigWarning{
 		Message: "c.StorageConfig.Trace.Block.Version != \"v2\" but " + setting + " is set",
 		Explain: "This setting is only used in v2 blocks",
+	}
+}
+
+func newMetricsGeneratorProcessorSpanMetricsFilterPoliciesWarning(setting string) ConfigWarning {
+	return ConfigWarning{
+		Message: "c.Generator.Processor.SpanMetrics.FilterPolicies attribute is invalid: " + setting,
+		Explain: "FilterPolicy attributes must be `resource` or `span` scopped.",
 	}
 }
--- a/docs/sources/tempo/metrics-generator/span_metrics.md
+++ b/docs/sources/tempo/metrics-generator/span_metrics.md
@@ -1,7 +1,7 @@
 ---
 aliases:
-- /docs/tempo/latest/server_side_metrics/span_metrics/
-- /docs/tempo/latest/metrics-generator/span_metrics/
+  - /docs/tempo/latest/server_side_metrics/span_metrics/
+  - /docs/tempo/latest/metrics-generator/span_metrics/
 title: Generate metrics from spans
 weight: 400
 ---
@@ -11,8 +11,9 @@ weight: 400
 The span metrics processor generates metrics from ingested tracing data, including request, error, and duration (RED) metrics.
 
 Span metrics generate two metrics:
-* A counter that computes requests
-* A histogram that tracks the distribution of durations of all requests
+
+- A counter that computes requests
+- A histogram that tracks the distribution of durations of all requests
 
 Span metrics are of particular interest if your system is not monitored with metrics,
 but it has distributed tracing implemented.
@@ -43,7 +44,7 @@ This processor is designed with the goal to mirror the implementation from the O
 The following metrics are exported:
 
 | Metric                         | Type      | Labels     | Description                  |
-|--------------------------------|-----------|------------|------------------------------|
+| ------------------------------ | --------- | ---------- | ---------------------------- |
 | traces_spanmetrics_latency     | Histogram | Dimensions | Duration of the span         |
 | traces_spanmetrics_calls_total | Counter   | Dimensions | Total count of the span      |
 | traces_spanmetrics_size_total  | Counter   | Dimensions | Total size of spans ingested |
@@ -55,7 +56,6 @@ Additional user defined labels can be created using the [`dimensions` configurat
 When a configured dimension collides with one of the default labels (e.g. `status_code`), the label for the respective dimension is prefixed with double underscore (i.e. `__status_code`).
 
 If you use ratio based sampler you can use custom sampler below to not lose metric information, you also need to set `metrics_generator.processor.span_metrics.span_multiplier_key` to `"X-SampleRatio"`
-
 
 ```go
 package tracer
@@ -90,6 +90,67 @@ func (ds RatioBasedSampler) Description() string {
 	return "Ratio Based Sampler which gives information about sampling ratio"
 }
 ```
+
+### Filtering
+
+In some cases it may be desirable to reduce the number of metrics produced by the `spanmetrics` processor. This can be accomplished by configuring the processor to use a filter. An `include` filter is used to match criteria that must be present in the span in order to be included. Following the include filter, an `exclude` filter may be used to reject portions of what was previously included by the filter policy.
+
+Currently supported is filtering by resource and span attributes with the following value types.
+
+- string
+- int
+- double
+
+Additionally, the following intrinsic span attributes may be filtered upon.
+
+- name
+- status (code)
+- kind
+
+When implementing a filter policy, intrinsic keys to act on may be used directly, as in the following snippet.
+
+```yaml
+---
+filter_policies:
+  - include:
+      match_type: strict
+      attributes:
+        - key: kind
+          value: SPAN_KIND_SERVER
+```
+
+In this example, spans which are of `kind` "server" are included for metrics export.
+
+When selecting spans based on non-intrinsic attributes, it is required to specify the scope of the attribute, similar to how it is specified in TraceQL. For example, if the `resource` contains a `location` attribute which is to be used in a filter policy, then the reference needs to be specified as `resource.location`. This requires users to know and specify which scope an attribute is to be found and avoids the ambiguity of conflicting values at differing scopes. The following may help illustrate.
+
+```yaml
+---
+filter_policies:
+  - include:
+      match_type: strict
+      attributes:
+        - key: resource.location
+          value: earth
+```
+
+In the above examples, we are using `match_type` of `strict`, which is a direct comparison of values. An additional option for `match_type` is `regex`. This allows users to build a regular expression to match against.
+
+```yaml
+---
+filter_policies:
+  - include:
+      match_type: regex
+      attributes:
+        - key: resource.location
+          value: eu-.*
+  - exclude:
+      match_type: regex
+      attributes:
+        - key: resource.tier
+          value: dev-.*
+```
+
+In the above, we first include all spans which have a `resource.location` that begins with `eu-` with the `include` statement, and then exclude those with begin with `dev-`. In this way, a flexible approach to filtering can be achieved to ensure that only metrics which are important are generated.
 
 ## Example
 
--- a/docs/sources/tempo/metrics-generator/span_metrics.md
+++ b/docs/sources/tempo/metrics-generator/span_metrics.md
@@ -95,7 +95,7 @@ func (ds RatioBasedSampler) Description() string {
 
 In some cases it may be desirable to reduce the number of metrics produced by the `spanmetrics` processor. This can be accomplished by configuring the processor to use a filter. An `include` filter is used to match criteria that must be present in the span in order to be included. Following the include filter, an `exclude` filter may be used to reject portions of what was previously included by the filter policy.
 
-Currently supported is filtering by resource and span attributes with the following value types.
+Currently, only filtering by resource and span attributes with the following value types is supported.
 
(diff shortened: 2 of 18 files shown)


modules/generator/processor/spanmetrics/config.go
```diff
@@ -67,3 +70,25 @@ func (ic *IntrinsicDimensions) ApplyFromMap(dimensions map[string]bool) error {
 	}
 	return nil
 }
+
+type FilterPolicy struct {
```
ie-pham, line 74: I think we're missing one layer here. We want to do a slice of include/exclude policies right?

```
{
    include: [
        {
            match_type: 'strict'
            attributes: [
                {
                    key: 'resource.service.name',
                    value: 'test-service',
                },
                {
                    key: 'span.status',
                    value: 'OK',
                }
            ],
        },
        {
            match_type: 'strict'
            attributes: [
                {
                    key: 'span.name',
                    value: 'Read',
                },
            ],
        },
    ],
    exclude: [],
}
```

where the attributes are "or"s and the policies are "and"s:
So the above would read: I want only Read spans that are either from "test-service" or with a status of "OK"

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -1,7 +1,7 @@
 ---
 aliases:
-- /docs/tempo/latest/server_side_metrics/span_metrics/
-- /docs/tempo/latest/metrics-generator/span_metrics/
+  - /docs/tempo/latest/server_side_metrics/span_metrics/
+  - /docs/tempo/latest/metrics-generator/span_metrics/
```
knylander-grafana, line 4: Good catch on the indent. I had to go look it up and you're right (which means I need to update my other PR).

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -91,6 +91,67 @@ func (ds RatioBasedSampler) Description() string {
 }
 ```
 
+### Filtering
+
+In some cases it may be desirable to reduce the number of metrics produced by the `spanmetrics` processor. This can be accomplished by configuring the processor to use a filter. An `include` filter is used to match criteria that must be present in the span in order to be included. Following the include filter, an `exclude` filter may be used to reject portions of what was previously included by the filter policy.
```
knylander-grafana, line 96:
```suggestion
In some cases, you may want to reduce the number of metrics produced by the `spanmetrics` processor. You can configure the processor to use an `include` filter to match criteria that must be present in the span in order to be included. Following the include filter, an `exclude` filter may be used to reject portions of what was previously included by the filter policy.
```

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -91,6 +91,67 @@ func (ds RatioBasedSampler) Description() string {
...
+
+Currently supported is filtering by resource and span attributes with the following value types.
```
knylander-grafana, line 98:
```suggestion
Currently, only filtering by resource and span attributes with the following value types is supported.
```

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -91,6 +91,67 @@ func (ds RatioBasedSampler) Description() string {
...
+
+- string
+- int
+- double
+
+Additionally, the following intrinsic span attributes may be filtered upon.
```
knylander-grafana, line 104:
```suggestion
Additionally, these intrinsic span attributes may be filtered upon:
```

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -91,6 +91,67 @@ func (ds RatioBasedSampler) Description() string {
...
+
+- name
+- status (code)
+- kind
+
+When implementing a filter policy, intrinsic keys to act on may be used directly, as in the following snippet.
```
knylander-grafana, line 110: The wording in this sentence is a little awkward. Maybe we can make it clearer? I have a suggestion below, but I"m not sure I'm getting the correct meaning. 
```suggestion
Intrinsic keys can be acted on directly when implementing a filter policy. For example: 
```

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -91,6 +91,67 @@ func (ds RatioBasedSampler) Description() string {
...
+
+- string
+- int
+- double
+
+Additionally, the following intrinsic span attributes may be filtered upon.
+
+- name
+- status (code)
+- kind
```
knylander-grafana, line 108: Are these values that would appear in the code snippet? Should they be formatted with `? In other words: 
```suggestion
- `name`
- `status` (code)
- `kind`
```

modules/generator/processor/spanmetrics/config.go
zalegrala: That was my original thinking too, but then I read closer in the original issue that pointed upstream to what OTEL is doing.

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/filter/filterconfig/config.go#L32

Double check me there, but I think we're doing what is described in that godoc.

For the example above, the same can be achieved by including multiple policies with each their own `include`.

I'd initially been thinking of a different approach to polices that was a little more flexible, but I couldn't come up with a scenario where the same results couldn't have been achieved with the otel config example.  Keeping some alignment seems okay to me, but the goofy part is that we're keeping alignment with an `internal` spec.  What do you think?

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -128,6 +133,130 @@ func (p *Processor) aggregateMetricsForSpan(svcName string, rs *v1.Resource, spa
...
+				return false
+			}
+		}
+	}
+
+	return true
+}
+
+// policyMatch returns true when the resource attribtues and span attributes match the policy.
+func policyMatch(policy *PolicyMatch, rs *v1.Resource, span *v1_trace.Span) bool {
```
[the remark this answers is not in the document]
zalegrala: I tried it that way also, but I found it less readable.  Since we're passing a pointer here it should be cheap, but I think if we want to construct the three policies to match outside of this method then we'll pass all thee in, which might clean up this method, but muddy the water on the other.  Are you thinking it might be more readable if we move it up a layer?

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -91,6 +91,67 @@ func (ds RatioBasedSampler) Description() string {
...
 
+### Filtering
+
+In some cases it may be desirable to reduce the number of metrics produced by the `spanmetrics` processor. This can be accomplished by configuring the processor to use a filter. An `include` filter is used to match criteria that must be present in the span in order to be included. Following the include filter, an `exclude` filter may be used to reject portions of what was previously included by the filter policy.
+
+Currently supported is filtering by resource and span attributes with the following value types.
+
+- string
+- int
+- double
```
knylander-grafana, line 102: Are these items things taht should be formatted in ` ? 
```suggestion
- `string`
- `int`
- `double`
```

knylander-grafana (commented):
Docs look good! Thank you for adding them.

modules/generator/processor/spanmetrics/spanmetrics.go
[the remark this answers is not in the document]
ie-pham: Sorry yup that's what I meant, just the three policies since they are static

modules/generator/processor/spanmetrics/config.go
ie-pham: Yeah ok, maybe the extra flexibility isn't needed at this time if we want to keep parity with Otel

modules/generator/processor/spanmetrics/spanmetrics.go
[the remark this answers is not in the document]
zalegrala: I'll commit an update here shortly with your suggestion.  Here are the benchmark results.
```
❯ benchstat old.txt new.txt
goos: linux
goarch: amd64
pkg: github.com/grafana/tempo/modules/generator/processor/spanmetrics
cpu: Intel(R) Core(TM) i7-10610U CPU @ 1.80GHz
                                      │   old.txt    │                new.txt                │
                                      │    sec/op    │    sec/op     vs base                 │
SpanMetrics_applyFilterPolicyNone-8      32.31 ± ∞ ¹    33.50 ± ∞ ¹        ~ (p=1.000 n=1) ²
SpanMetrics_applyFilterPolicySmall-8    220.3m ± ∞ ¹   128.1m ± ∞ ¹        ~ (p=1.000 n=1) ²
SpanMetrics_applyFilterPolicyMedium-8   871.6m ± ∞ ¹   216.6m ± ∞ ¹        ~ (p=1.000 n=1) ²
geomean                                  1.837         976.0m        -46.88%
² need >= 4 samples to detect a difference at alpha level 0.05

                                      │      old.txt       │                new.txt                 │
                                      │        B/op        │     B/op       vs base                 │
SpanMetrics_applyFilterPolicyNone-8          15.67Gi ± ∞ ¹   15.67Gi ± ∞ ¹        ~ (p=1.000 n=1) ²
SpanMetrics_applyFilterPolicySmall-8     32001685.00 ± ∞ ¹     80.00 ± ∞ ¹        ~ (p=1.000 n=1) ²
SpanMetrics_applyFilterPolicyMedium-8   224008272.00 ± ∞ ¹     80.00 ± ∞ ¹        ~ (p=1.000 n=1) ²
geomean                                      471.2Mi         46.46Ki        -99.99%
² need >= 4 samples to detect a difference at alpha level 0.05

                                      │      old.txt      │                new.txt                │
                                      │     allocs/op     │  allocs/op    vs base                 │
SpanMetrics_applyFilterPolicyNone-8          170.9M ± ∞ ¹   170.9M ± ∞ ¹        ~ (p=1.000 n=1) ²
SpanMetrics_applyFilterPolicySmall-8    1000004.000 ± ∞ ¹    3.000 ± ∞ ¹        ~ (p=1.000 n=1) ²
SpanMetrics_applyFilterPolicyMedium-8   3000004.000 ± ∞ ¹    3.000 ± ∞ ¹        ~ (p=1.000 n=1) ²
geomean                                      8.004M         1.154k        -99.99%
² need >= 4 samples to detect a difference at alpha level 0.05
```

joe-elliott (commented):
one broader question that just occurred to me: should all of this logic apply to both span metrics and service graph metrics? should we move this up a level?

modules/generator/overrides.go
```diff
@@ -3,6 +3,7 @@ package generator
 import (
 	"github.com/grafana/tempo/modules/generator/registry"
 	"github.com/grafana/tempo/modules/overrides"
+	"github.com/grafana/tempo/pkg/sharedconfig"
```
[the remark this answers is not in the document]
zalegrala: Oh!  Is that why my tests are failing? :P  My goodness.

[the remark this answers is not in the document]
zalegrala: I was following what was going on in #2261, but I like the idea of `modules/overrides`, or alternatively a new `spanmetrics/config` package.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -10,31 +12,50 @@ import (
...
 
 	registry registry.Registry
 
-	spanMetricsCallsTotal      registry.Counter
-	spanMetricsDurationSeconds registry.Histogram
-	spanMetricsSizeTotal       registry.Counter
+	spanMetricsCallsTotal       registry.Counter
+	spanMetricsDurationSeconds  registry.Histogram
+	spanMetricsSizeTotal        registry.Counter
+	spanMetricsFilterDropsTotal registry.Counter
```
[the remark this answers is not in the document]
zalegrala: Nope, good call out.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -89,7 +125,11 @@ func (p *Processor) aggregateMetrics(resourceSpans []*v1_trace.ResourceSpans) {
 
 		for _, ils := range rs.ScopeSpans {
 			for _, span := range ils.Spans {
-				p.aggregateMetricsForSpan(svcName, rs.Resource, span)
+				if p.applyFilterPolicy(rs.Resource, span) {
```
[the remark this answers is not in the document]
zalegrala: I also prefer that.  I'll adjust.

zalegrala:
Thanks for the review @joe-elliott.  If we move this up a level, do you suppose the processors should have an independent filter config, or share one?  Sharing one would be simpler and likely more performant, but I wonder if folks would want to tune them independently.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -58,13 +79,28 @@ func New(cfg Config, registry registry.Registry) gen.Processor {
...
 
+	var filterPolicies []*filterPolicy
+
+	for _, policy := range cfg.FilterPolicies {
+		p := &filterPolicy{
+			Include: getSplitPolicy(policy.Include),
+			Exclude: getSplitPolicy(policy.Exclude),
+		}
+
+		if p.Include != nil || p.Exclude != nil {
```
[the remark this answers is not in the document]
zalegrala: Probably yes.  I'll adjust.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -128,6 +168,118 @@ func (p *Processor) aggregateMetricsForSpan(svcName string, rs *v1.Resource, spa
...
 
+// applyFilterPolicy returns true if the span should be included in the metrics.
+func (p *Processor) applyFilterPolicy(rs *v1.Resource, span *v1_trace.Span) bool {
+	// With no filter policies specified, all spans are included.
+	if len(p.Cfg.FilterPolicies) == 0 {
+		return true
+	}
+
+	for _, policy := range p.filterPolicies {
+		if policy.Include != nil {
```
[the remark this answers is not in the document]
zalegrala: I'm not sure what `policyMatch()` should do in the case of a nil policy.  I agree it would be a little cleaner, but include is the inverse of exclude here, so for example returning a `true` value from `policyMatch()` isn't as clean.  Perhaps I can make exclude and include behave the same here, and then make the suggested adjustment.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -128,6 +168,118 @@ func (p *Processor) aggregateMetricsForSpan(svcName string, rs *v1.Resource, spa
...
+		}
+	}
+
+	return len(policy.Attributes) == matches
+}
+
+// policyMatchAttrs returns true if all attributes in the policy match the attributes in the span.  String, bool, int, and floats are supported.  Regex MatchType may be applied to string span attributes.
+func policyMatchAttrs(policy *sharedconfig.PolicyMatch, attrs []*v1_common.KeyValue) bool {
+
+	matches := 0
```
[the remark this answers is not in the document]
zalegrala: I'll give some thought to how to restructure this.  I took the fastest(to code) approach for me, but I see what you're getting at.  The double nested loops means that we will often compare attributes that don't match, and so returning false on the first non-match won't be the correct behavior.

modules/generator/overrides.go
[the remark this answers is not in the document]
zalegrala: I'm going to restructure this into a `pkg/spanfilter/config` that can be used in common.

zalegrala:
In lieu of an specific idea about how widely to apply the filtering, I'm going to restructure the span filtering into `pkg/spanfilter` and implement it in the `spanmetrics` processor.  I think this will give is the ability to re-use the code elsewhere if we want to pick it up later and still give us the short term benefit of filtering on spans in the `spanmetrics` currently.

modules/generator/overrides.go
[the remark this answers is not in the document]
zalegrala: This has been refactored a bit.

cmd/tempo/app/config.go
```diff
@@ -176,6 +177,25 @@ func (c *Config) CheckConfig() []ConfigWarning {
...
+			a, w := traceql.ParseIdentifier(attr.Key)
+			if w != nil {
+				warnings = append(warnings, newMetricsGeneratorProcessorSpanMetricsFilterPoliciesWarning(w.Error()))
+			}
+			if a.Scope == traceql.AttributeScopeNone {
+				warnings = append(warnings, newMetricsGeneratorProcessorSpanMetricsFilterPoliciesWarning(fmt.Sprintf("invalid attribute scope: %s", attr.Key)))
+			}
+
+			switch a.Intrinsic.String() {
+			case "name", "status", "kind": // currently supported
```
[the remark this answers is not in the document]
zalegrala: I've refactored this a bit into the `pkg/spanfilter/config` package.

modules/generator/processor/spanmetrics/spanmetrics.go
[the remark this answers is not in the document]
zalegrala: I spoke too soon.  For a generator instance, we don't have the ID available currently, so to create a metric where we include the tenant label isn't feasible without a bit of refactor.  This currently pushes the metric to the remote endpoint and would increase the series count, so perhaps this isn't something we want to do.  Though, if we host the metric on the generator itself, then we'd have access to know which tenants would be filtering which spans, but likely this wouldn't have the desired value for folks who don't have access to those metrics.  I.e. if a cloud user has a filter, they wouldn't be able to see how many spans are being rejected by their filter, which seems like the primary utility.  I'm a little torn on even including this, but it does seem like we want some indication that the spans are being filtered out.

modules/generator/processor/spanmetrics/spanmetrics.go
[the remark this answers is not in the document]
zalegrala: I've added a commit to exit early when we're not matched.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -68,10 +73,21 @@ func New(cfg Config, registry registry.Registry) gen.Processor {
 	if cfg.Subprocessors[Size] {
 		p.spanMetricsSizeTotal = registry.NewCounter(metricSizeTotal, labels)
 	}
+
+	filter, err := spanfilter.NewSpanFilter(cfg.FilterPolicies)
+	if err != nil {
+		return nil, err
+	}
+
+	if filter != nil {
```
joe-elliott, line 82: based on this it look like you expect that filter can sometimes be nil, but below 

```
  if p.filter.ApplyFilterPolicy(rs.Resource, span) {
```

you don't test for nil. is there a mismatch here?

pkg/spanfilter/spanfilter.go
```diff
@@ -0,0 +1,238 @@
...
+				return false
+			}
+			matches++
+		case traceql.IntrinsicStatus:
+			if !stringMatch(policy.MatchType, span.GetStatus().GetCode().String(), pa.Value.(string)) {
+				return false
+			}
+			matches++
+		case traceql.IntrinsicKind:
+			if !stringMatch(policy.MatchType, span.GetKind().String(), pa.Value.(string)) {
```
joe-elliott, line 121: anyway we can do int matches here and on status? would be much faster

joe-elliott (commented):
Really liking the refactor to pull out the filters. Some thoughts but this is close

modules/generator/processor/spanmetrics/spanmetrics.go
zalegrala: This has been refactored.

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -91,6 +91,77 @@ func (ds RatioBasedSampler) Description() string {
...
+- `bool`
+- `double`
+- `int`
+- `string`
+
+Additionally, these intrinsic span attributes may be filtered upon:
+
+- `name`
+- `status` (code)
+- `kind`
```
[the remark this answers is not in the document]
zalegrala: Just added a commit.

pkg/spanfilter/spanfilter.go
zalegrala: Probably yes, but it might be a little clumsy, since we'd need to take this as an int in the config I think.  There is a special yaml struct tag we could use iirc to parse a string as an int, or perhaps some custom unmarshaling.  For now I'm inclined to leave it and come back to it later.

ie-pham:
> Thanks for the review @joe-elliott. If we move this up a level, do you suppose the processors should have an independent filter config, or share one? Sharing one would be simpler and likely more performant, but I wonder if folks would want to tune them independently.

Yeah this occurred to me just now too. When I added the ingestion_slack filter last time, it was applied before pushing to both processors and so the code lived on level above (instance.go).

modules/generator/instance.go
```diff
@@ -52,7 +52,10 @@ var (
 	}, []string{"tenant", "reason"})
 )
 
-const reasonOutsideTimeRangeSlack = "outside_metrics_ingestion_slack"
+const (
+	reasonOutsideTimeRangeSlack = "outside_metrics_ingestion_slack"
+	reasonSpanMetricsFiltered   = "span_metrics_filtered"
```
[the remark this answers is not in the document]
zalegrala: Yeah, this doesn't read well.  Its used to indicate the number of spans rejected by the filter, so what about `spans_rejected`  or `filter_policy_rejections` maybe?  I was thinking that if we had another processor also filtering, differentiating between them would be nice.  We could include a label to include a label perhaps for the name of the processor doing the rejecting.  Not sure though.

pkg/spanfilter/spanfilter.go
zalegrala: I think I can work this.  I'll take a closer look next week.

zalegrala: I've added a commit for this.  It made some parts less readable, but seems okay to me.  How's that look?

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -31,11 +33,15 @@ type Processor struct {
 	spanMetricsDurationSeconds registry.Histogram
 	spanMetricsSizeTotal       registry.Counter
 
+	filter *spanfilter.SpanFilter
+
+	discardCounter prometheus.Counter
+
```
joe-elliott, line 39: nit: name it something to make it clear it belongs with span filters? and remove the space between to visually pair them?

`filteredSpansCounter` ? dunno :)

modules/overrides/limits.go
```diff
@@ -58,18 +59,19 @@ type Limits struct {
...
+	MetricsGeneratorProcessorServiceGraphsDimensions        []string                    `yaml:"metrics_generator_processor_service_graphs_dimensions" json:"metrics_generator_processor_service_graphs_dimensions"`
+	MetricsGeneratorProcessorSpanMetricsHistogramBuckets    []float64                   `yaml:"metrics_generator_processor_span_metrics_histogram_buckets" json:"metrics_generator_processor_span_metrics_histogram_buckets"`
+	MetricsGeneratorProcessorSpanMetricsDimensions          []string                    `yaml:"metrics_generator_processor_span_metrics_dimensions" json:"metrics_generator_processor_span_metrics_dimensions"`
+	MetricsGeneratorProcessorSpanMetricsIntrinsicDimensions map[string]bool             `yaml:"metrics_generator_processor_span_metrics_intrinsic_dimensions" json:"metrics_generator_processor_span_metrics_intrinsic_dimensions"`
+	MetricsGeneratorProcessorSpanMetricsFilterPolicies      []filterconfig.FilterPolicy `yaml:"metrics_generator_processor_span_metrics_filter_policies" json:"metrics_generator_processor_span_metrics_filter_policies"`
```
joe-elliott, line 74: one of these days we're going to break configmap sizes in k8s and be very very sad.

pkg/spanfilter/config/config.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if policy.Include != nil {
+		if err := ValidatePolicyMatch(policy.Include); err != nil {
+			return err
+		}
+	}
+
+	if policy.Exclude != nil {
+		if err := ValidatePolicyMatch(policy.Exclude); err != nil {
+			return err
```
joe-elliott, line 52: similar nit here. may want to wrap with the policy that failed to match for clarity.

modules/overrides/limits.go
zalegrala: Hmm, that won't be much fun.  We might be able to make more configmaps and load them from a mounted directory or something.  Might want to approach that before that days comes :)

PR #2335, opened 2023-04-13
Prefix service graph extra dimensions

### LINKED ISSUE #1860: [Service graph] Add prefix to differentiate client and server attributes
In the service graph processor, when [adding custom dimensions to metrics](https://github.com/grafana/tempo/blob/58637f39cb91d707a43a1ba246d9b1ff97c1548b/modules/generator/processor/servicegraphs/servicegraphs.go#L224-L230), values can be overridden if parent and children span have the same attribute but with different values. The attributes of the last span to be processed are the ones selected for the metrics. This results in inconsistent behaviour.

We should a prefix to the label metrics to differentiate between values coming from the parent span and the children span.

domasx2:
**What this PR does**:
In the service graph processor, when [adding custom dimensions to metrics](https://github.com/grafana/tempo/blob/58637f39cb91d707a43a1ba246d9b1ff97c1548b/modules/generator/processor/servicegraphs/servicegraphs.go#L224-L230), values can be overridden if parent and children span have the same attribute but with different values. The attributes of the last span to be processed are the ones selected for the metrics. This results in inconsistent behaviour, and does not match what otel collector service graph connector does.

One use case is to add `service.namespace` dimension, in case you want services to be identified by both name and namespace. You have to have a label for both client and server.

This PR prefixes values coming from client span with `client_` and values coming from server span with `server_`. 

Fixes #1860

**Checklist**
- [x] Tests updated
- [x] Documentation added


Jxlam-pdx:
@cyrille-leclerc @gouthamve can you drive understanding and communicating the impact with teams involved? It's a breaking change and potentially have cost implication with Cardinality increase.  Thanks!

domasx2:
@joe-elliott  thanks, that makes sense. Any thoughts on introducing this in a non breaking way? For example, adding syntax to be explicit about desired behaviour, like `client:service.namespace` to generate `client_service_namespace` taken only from client span. Or a config opt-in flag.

domasx2:
OTEL service graph processor does prefix: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/servicegraphprocessor/processor.go#L293

cyrille-leclerc:
@joe-elliott @Jxlam-pdx point taken. Sorry for the delay, @gouthamve is busy at KubeCon this week.

zalegrala:
Have we started the review on current customers?  Is there any update to that end?

ie-pham:
If we are intending on adding `enable_client_server_prefix` as a per tenant config, this needs to also be in the overrides section. You can follow what I did for `metrics_generator_processor_span_metrics_enable_target_info` in my [PR](https://github.com/grafana/tempo/pull/2261/files#diff-a7c59378c5a4634425b60403c61a59b4cf746ffba074f0289a8ddfe52708a0ea) 😅

domasx2:
@ie-pham thanks! I did so, but now am completely stumped as to why CI is failing :( everything works locally and the error+ location in CI doesn't make any sense... Maybe you have a clue? :pray:

domasx2:
Thanks @ie-pham , resolved :)

ie-pham:
This LGTM

domasx2:
@joe-elliott  can you take a look again? Feature is now opt-in, no longer a breaking change

joe-elliott (approved):

domasx2:
hold on with merging this for a bit, still some ongoing discussions on app o11y side

domasx2:
good to merge :+1:

PR #2338, opened 2023-04-14
Doc - Tempo meta-monitoring configuration

camsellem:
**What this PR does**:

Adding mention of the fact that tempo supports only Jaeger Thrift protocol for meta-monitoring. Added the env var with a sample value.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/operations/monitoring.md
+++ b/docs/sources/tempo/operations/monitoring.md
@@ -34,9 +34,11 @@ The tracer can be configured [using environment variables](https://github.com/ja
 To enable tracing, set one of the following: `JAEGER_AGENT_HOST` and `JAEGER_AGENT_PORT`, or `JAEGER_ENDPOINT`.
 
 The Jaeger client uses remote sampling by default, if the management server is not available no traces will be sent.
+Only Jaeger Thrift HTTP protocol is supported for traces format (default port: 14268).
 To always send traces (no sampling), set the following environment variables:
 
 ```
+JAEGER_ENDPOINT=http://<tempo_endpoint>:14268/api/traces
 JAEGER_SAMPLER_TYPE=const
 JAEGER_SAMPLER_PARAM=1
 ```


docs/sources/tempo/operations/monitoring.md
```diff
@@ -34,9 +34,11 @@ The tracer can be configured [using environment variables](https://github.com/ja
 To enable tracing, set one of the following: `JAEGER_AGENT_HOST` and `JAEGER_AGENT_PORT`, or `JAEGER_ENDPOINT`.
 
 The Jaeger client uses remote sampling by default, if the management server is not available no traces will be sent.
+Only Jaeger Thrift HTTP protocol is supported for traces format (default port: 14268).
```
joe-elliott, line 37: i don't think this is true. tempo can send all the formats that the jaeger client supports:

https://github.com/jaegertracing/jaeger-client-go

thrift compact/binary over udp or thrift over http. by default it uses thrift compact over udp on port 6831.

camsellem: Hey Joe - we tried Thirft over gRPC (14250) but it didnt work, over HTTP worked tho (14268) - I agree I made a mistake by using the word "only".

Should I just list what is supported and the ports users can use instead?

camsellem: The customer was also a bit confused because our doc redirected them to a page that is referring to a deprecated library.

joe-elliott: > Should I just list what is supported and the ports users can use instead?

If you want. It might make more sense to just refer to Jaeger client documentation. Such as this:

https://www.jaegertracing.io/docs/1.44/client-features/

or just the main readme for the go client:

https://github.com/jaegertracing/jaeger-client-go

> The customer was also a bit confused because our doc redirected them to a page that is referring to a deprecated library.

Yup, we would like to switch as well. However, the Jaeger clients works and there are more pressing matters so we haven't pushed on it. There was an attempt at one point to do this, but I believe context propagation is broken.

People can experiment with it by setting this config option:
https://github.com/grafana/tempo/blob/aa4531fef5c1271d3f79db86605c2b82007ee55a/cmd/tempo/app/config.go#L68

PR #2348, opened 2023-04-19
Add example of finding a trace by its ID

charleskorn:
**What this PR does**:

It wasn't immediately clear to me by looking at [the TraceQL docs](https://grafana.com/docs/tempo/latest/traceql/) how I can search for a trace by its ID, so this PR adds an example of how to do that.

(none)

**Checklist**
- [n/a] Tests updated
- [x] Documentation added
- [n/a] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/docs/sources/tempo/traceql/_index.md
+++ b/docs/sources/tempo/traceql/_index.md
@@ -283,6 +283,14 @@ It's a convenient request to identify misconfigurations and leaks across product
 { resource.deployment.environment = "production" } && { resource.deployment.environment = "staging" }
 ```
 
+### Find traces by trace ID
+
+Find any trace with trace ID `1a2b3c4e5f678901`:
+
+```
+1a2b3c4e5f678901
+```
+
 ### Other examples
 
 Find any trace with a `deployment.environment` attribute set to `production` and `http.status_code` attribute set to `200`:


joe-elliott (commented):
Thanks for the PR! I wonder if this belongs here though. Grafana combines the traceql and trace by id lookup into one spot, but just typing a trace id isn't really traceql.

Having said that I understand that someone who is reading the docs may find this information useful. Does this make more sense in the Grafana section? wdyt?

charleskorn:
Ah, I didn't realise the Grafana UI combined both TraceQL and trace ID search into one field. 

Given the field is labelled as just being for TraceQL, I went looking in the TraceQL docs, so perhaps a note in the TraceQL docs that the Grafana UI also supports searching for a particular trace ID would be helpful? Or perhaps we can change the Grafana UI to make this clearer?

joe-elliott:
Yeah, I agree. This is more of a UX concern on Grafana's side. I have filed an issue here:

https://github.com/grafana/grafana/issues/67055

Closing this PR. Thanks for looking to improve Tempo docs!

PR #2356, opened 2023-04-20
add a regex example

soeirosantos:
**What this PR does**:
Adds a simple example of regex for TraceQL.

### DIFF
--- a/docs/sources/tempo/traceql/_index.md
+++ b/docs/sources/tempo/traceql/_index.md
@@ -306,5 +306,5 @@ Find any trace where any span has an `http.method` attribute set to `GET` as wel
 Find any trace with a `deployment.environment` attribute that matches the regex `prod-.*` and `http.status_code` attribute set to `200`:
 
 ```
-{ .deployment.environment = "prod-.*" && .http.status_code = 200 }
+{ resource.deployment.environment =~ "prod-.*" && span.http.status_code = 200 }
 ```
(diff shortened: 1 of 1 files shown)


docs/sources/tempo/traceql/_index.md
```diff
@@ -302,3 +302,9 @@ Find any trace where any span has an `http.method` attribute set to `GET` as wel
 ```
 { span.http.method = "GET" && status = ok } && { span.http.method = "DELETE" && status != ok }
 ```
+
+Find any trace with a `deployment.environment` attribute that matches the regex `prod-.*` and `http.status_code` attribute set to `200`:
+
+```
+{ .deployment.environment = "prod-.*" && .http.status_code = 200 }
```
joe-elliott, line 309: The regex operator needs to be used to do the comparison correctly. I'd also prefer to keep as many example as possible to use the `resource` and `span` scopes where appropriate. it's so much more efficient!

```suggestion
{ resource.deployment.environment =~ "prod-.*" && span.http.status_code = 200 }
```

knylander-grafana (commented):
Thank you for adding documentation.

docs/sources/tempo/traceql/_index.md
soeirosantos: Oh! The example of the regex operator was the only intention of this PR. I think I copied things around and missed updating the example :palmface: sorry about that, and thanks for the fix

joe-elliott (approved):

joe-elliott:
Thanks for the contribution!

PR #2365, opened 2023-04-25
Add ability to detect virtual nodes in the servicegraph processor

mapno:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -233,6 +244,15 @@ func (p *Processor) upsertDimensions(m map[string]string, resourceAttr []*v1_com
 	}
 }
 
+func (p *Processor) upsertPeerNode(e *store.Edge, spanAttr []*v1_common.KeyValue) {
+	for _, peerKey := range p.Cfg.PeerAttributes {
+		if v, ok := processor_util.FindAttributeValue(peerKey, spanAttr); ok {
+			e.PeerNode = v
```
mdisibio, line 250: Looking at this logic - is it correct to say that the order of `PeerAttributes` is their priority?  If `net.sock.peer.addr` exists it will always be used first.  If so let's add a note to the documentation snippet in `service_graphs.md`

mdisibio (approved):
LGTM - A few small q's and comments.

docs/sources/tempo/metrics-generator/service_graphs.md
```diff
@@ -43,6 +43,17 @@ Each emitted metrics series have the `client` and `server` label corresponding w
   tempo_service_graph_request_total{client="app", server="db", connection_type="database"} 20
 ```
 
+#### Virtual nodes
+
+Virtual nodes are nodes that form part of the lifecycle of a trace,
+but spans for them are not being collected because they're outside the user's reach (eg. an external service for payment processing) or are not instrumented (eg. a frontend application).
```
knylander-grafana, line 49:
```suggestion
but spans for them are not being collected because they're outside the user's reach (for example, an external service for payment processing) or are not instrumented (for example, a frontend application).
```

docs/sources/tempo/metrics-generator/service_graphs.md
```diff
@@ -43,6 +43,17 @@ Each emitted metrics series have the `client` and `server` label corresponding w
...
+
+Virtual nodes can be detected in two different ways:
+
+- The root span has `span.kind` set to `server`. This indicates that the request has initiated by an external system that's not instrumented, like a frontend application or an engineer via curl
+- A `client` span does not have its matching `server` span, but has a peer attribute present. In this case we make the assumption that a call was made to an external service, for which Tempo won't receive spans.
+  - Default peer attributes are `net.sock.peer.addr`, `net.sock.peer.name`, `rpc.service.key`, `net.peer.name`, `http.url`, `http.target`.
```
knylander-grafana, line 55:
```suggestion
Virtual nodes can be detected in two different ways:

- The root span has `span.kind` set to `server`. This indicates that the request has initiated by an external system that's not instrumented, like a frontend application or an engineer via `curl`.
- A `client` span does not have its matching `server` span, but has a peer attribute present. In this case, we make the assumption that a call was made to an external service, for which Tempo won't receive spans.
   - The default peer attributes are `net.sock.peer.addr`, `net.sock.peer.name`, `rpc.service.key`, `net.peer.name`, `http.url`, `http.target`.
```

knylander-grafana (commented):
Doc changes look good. Thank you for updating the docs.

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -258,6 +278,17 @@ func (p *Processor) onComplete(e *store.Edge) {
 
 func (p *Processor) onExpire(e *store.Edge) {
 	p.metricExpiredEdges.Inc()
+
+	e.ConnectionType = store.VirtualNode
```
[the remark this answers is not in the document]
mapno: Added comments to clarify what this is doing.

>Should e.ConnectionType only be set if one of the conditions is true?

Doesn't really matter, as if it doesn't match one of the two conditions to be considered a virtual node, it won't be collected.

modules/generator/processor/servicegraphs/servicegraphs.go
mapno: Nice suggestions. Applied both.

mohammadjizi:
Hi. Can i know please when this feature will be released?

joe-elliott:
This will be in Tempo 2.2. There is no set release date, but due to our normal cadence it will likely be cut late July/early August

PR #2366, opened 2023-04-25
GRPC streaming search

joe-elliott:
**What this PR does**:
Adds a GRPC streaming endpoint to the query frontend that allows a client to stream search results from Tempo. Additionally this PR adds two new fields to SearchMetrics (TotalJobs and JobsCompleted) that allow a client to estimate the amount of work to be done and the rate at which it is being completed.

Other Changes:
- Small refactor around the old SearchResponse object to make it work better for this use case
- Removed SkippedBlocks and SkippedTraces from the SearchMetrics object. These metrics were only relevant with the now removed flatbuffer search
- Updated CLI to use the new streaming endpoint
- Adjusted SearchSharder to take a factory function to create the SearchProgress object. This allows the streaming search code to capture this object and possibly return a different one. 

Todo:
- ~Currently this PR sends the entire search response every time. For larger search results this could create a fair amount of traffic. I'm considering changing it to only stream diffs. wdyt?~ Went ahead and implemented diffs.
- ~Considering renaming "InspectedBlocks" in SearchMetrics to "TotalBlocks". It is set once at the beginning of the search call and never changes.~ Performed the rename.

**Checklist**
- [x] Tests updated
- [x] Documentation added


docs/sources/tempo/api_docs/_index.md
```diff
@@ -546,3 +548,63 @@ GET /status/usage-stats
 ```
 
 Displays anonymous usage stats data that is reported back to Grafana Labs.
+
+# Tempo GRPC API
```
knylander-grafana, line 552: Making this a subheading since most pages only have one level 1 heading, which is also the page title. 
```suggestion
## Tempo GRPC API
```

docs/sources/tempo/api_docs/_index.md
```diff
@@ -546,3 +548,63 @@ GET /status/usage-stats
...
+
+Tempo uses GRPC to internally communicate with itself, but only has one externally supported client. The query-frontend component implements
+the streaming querier interface defined below. [See here](https://github.com/grafana/tempo/blob/main/pkg/tempopb/) for the complete proto definition and generated code.
+
+Note that the below rpc call returns only traces that are new or have updated each time SearchResponse is returned except for the last response. The
```
knylander-grafana, line 557:
```suggestion
The below `rpc` call returns only traces that are new or have updated each time `SearchResponse` is returned except for the last response. The
```

knylander-grafana (commented):
Docs look good. Minor wording suggestions. Thank you for adding docs.

modules/frontend/search_streaming.go
```diff
@@ -0,0 +1,195 @@
...
+	// clear seen traces
+	p.seenTraces = map[string]struct{}{}
+
+	return res
+}
+
+// finalResult gives the user the ability to pull all results w/o filtering
+// to ensure that all results are sent to the caller
+func (p *diffSearchProgress) finalResult() *shardedSearchResults {
+	return p.progress.result()
```
mdisibio, line 91: Thinking through how a downstream system will consume the results (Grafana) - we might need to tweak something to resolve some ambiguity. Ex: if we stream results X, then Y, they should be added to the UI immediately. But then the final results X,Y,Z come in, it either has to replace the whole results or do its own diff to detect Z. Also search results for XY might change if new partials are found, and the new values would be streamed again.    I am thinking we could add a flag to denote the "final" results, or maybe there is some other mechanic that I'm missing?

modules/frontend/searchsharding.go
```diff
@@ -71,7 +74,7 @@ type SearchSharderConfig struct {
 }
 
 // newSearchSharder creates a sharding middleware for search
-func newSearchSharder(reader tempodb.Reader, o *overrides.Overrides, cfg SearchSharderConfig, sloCfg SLOConfig, logger log.Logger) Middleware {
+func newSearchSharder(reader tempodb.Reader, o *overrides.Overrides, cfg SearchSharderConfig, sloCfg SLOConfig, progress searchProgressFactory, logger log.Logger) Middleware {
```
mdisibio, line 77: If I'm reading this right it looks like the progress factory is only called once.   I think it would help readability if we passed in the actual `shardedSearchProgress` and let `Roundtrip` set the static fields like total blocks/jobs directly.

modules/frontend/searchsharding.go
```diff
@@ -182,7 +185,12 @@ func (s searchSharder) RoundTrip(r *http.Request) (*http.Response, error) {
 		startedReqs++
 
 		go func(innerR *http.Request) {
-			defer wg.Done()
+			defer func() {
+				if progress.shouldQuit() {
+					subCancel()
```
mdisibio, line 190: What does this change do?  It looks like `subCancel` is already deferred up above, and the main for loop on line 174 will exit on the same criteria (`progress.shouldQuit() == true`)

mdisibio (commented):
I love this PR, and the ability to show a progress bar for search results is amazing!  Have a few small questions, but overall this looks great.

modules/frontend/search_streaming.go
joe-elliott: The contract that I'm seeing with the frontend is this:

If we return a traceid in the resultset then Tempo guarantees it is the complete set of results for that trace. The frontend can then replace any results it has that match that trace id. This will come into play more with grouping.

pkg/tempopb/tempo.proto
```diff
@@ -26,6 +26,10 @@ service Querier {
   rpc SearchTagValuesV2(SearchTagValuesRequest) returns (SearchTagValuesV2Response) {};
 }
 
+service StreamingQuerier {
+  rpc Search(SearchRequest) returns (stream SearchResponse);
```
zalegrala, line 30: Should we also stream the requests?

integration/util.go
```diff
@@ -217,6 +220,15 @@ func NewJaegerGRPCClient(endpoint string) (*jaeger_grpc.Reporter, error) {
 	return jaeger_grpc.NewReporter(conn, nil, logger), err
 }
 
+func NewSearchGRPCClient(endpoint string) (tempopb.StreamingQuerierClient, error) {
+	clientConn, err := grpc.DialContext(context.Background(), endpoint, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
zalegrala, line 224: It might be nice to allow for configuring the transport for things like TLS.

cmd/tempo/app/modules.go
```diff
@@ -287,11 +287,12 @@ func (t *App) initQueryFrontend() (services.Service, error) {
...
 	)
 
-	traceByIDHandler := middleware.Wrap(queryFrontend.TraceByID)
-	searchHandler := middleware.Wrap(queryFrontend.Search)
+	traceByIDHandler := middleware.Wrap(queryFrontend.TraceByIDHandler)
+	searchHandler := middleware.Wrap(queryFrontend.SearchHandler)
 
 	// register grpc server for queriers to connect to
 	frontend_v1pb.RegisterFrontendServer(t.Server.GRPC, t.frontend)
+	tempopb.RegisterStreamingQuerierServer(t.Server.GRPC, queryFrontend)
```
zalegrala, line 295: Will this new server be traced for us?

modules/frontend/search_streaming.go
zalegrala: Just wondering, do we need a finished state?  If the streaming were bidirectional the caller could decide what what was long enough to wait, or cancel the reuqest for whatever reason.

modules/frontend/searchsharding.go
joe-elliott: yup, seems obvious now that you point it out. good call

modules/frontend/searchsharding.go
joe-elliott: The searchResponse object/searchSharder used to cancel the context directly in `shouldQuit()`. I wanted to keep that responsibility in the main search sharding code so I moved it here.

Everytime a search sharding job completes this code checks to see if we should cancel all other jobs and does so if necessary. Reasons to cancel include things like a job failed or the results limit was reached.

pkg/tempopb/tempo.proto
joe-elliott: I considered that, but it's more complex and given the nature of Tempo queries I think that a one way stream is sufficient.

integration/util.go
joe-elliott: that will remain on the long list :). since this is just a simple integration test function it's fine for now.

cmd/tempo/app/modules.go
joe-elliott: it should. tracing is initiated in weaveworks common and propagated correctly these changes (i think). if there are any issues I will patch them up.

modules/frontend/search_streaming.go
mdisibio: > do we need a finished state?

Yeah now I'm thinking we don't need to send everything at the end.  The final diffs if any would be sufficient.

modules/frontend/searchsharding.go
mdisibio: I'm still a little unclear.   I think it would have the same effect if not called here.   The outer loop on requests would break because `shouldQuit` is true, and then the method would exit and `subCancel()` is already deferred.

modules/frontend/search_streaming.go
mdisibio: After talking offline, we decided there isn't a straightforward path here and it's ok as-is.  Some things that need to be considered, not exhaustive:  bounded wait group .Add/.Done block because they don't take contexts, the http roundtripping can have other non-context related errors like json marshalling.

mdisibio (approved):

PR #2367, opened 2023-04-25
Update link to docs instead of outdated blog post

Condla:
**What this PR does**:
There's a link in the docs that refers to an outdated blog post. Linking the corresponding config page instead.

Fixes #<issue number>


joe-elliott (approved):
Thanks for the update! Good catch

cc @knylander-grafana

PR #2368, opened 2023-04-25
New metrics generator processor to store a local copy of traces

mdisibio:
**What this PR does**:
This PR is part of a wider initiative to perform on-demand metrics calculations.  It adds a new processor `local-blocks` that stores a copy of all received traces on disk.  It works almost identically to an ingester with a live traces map, wal, local blocks, retention period, and with similar configuration.  Currently this serves **no purpose** but more functionality will be added in later PRs.

The metrics generator config looks like:
```yaml
metrics-generator:
  processor:
    local_blocks:
       block: (tempodb block config, optional)
       flush_check_period: (duration, optional)
       trace_idle_period: (duration, optional)
       max_block_duration: (duration, optional)
       max_block_bytes: (optional)
       complete_block_timeout: (optional)

  traces_storage:
    path: /tmp/tempo/generator/traces (required) # the only required field
    version: (optional)

overrides:
  metrics_generator_processors: [local-blocks]  # Enable the processor
```

Notes:

1. This forks a lot of code from the ingester for live traces, flushing, deleting, etc.  I originally hoped to call the ingester module but it's not prepared for that and would need a combination of updates (exporting symbols, making parts like flushing to object storage disable-able). So forking was not only more straightforward but also a chance to simplify the code.
2. Flushing/completing/deleting is done via goroutines per instance. The real ingester has a shared set of workers and work queue across all instances.  That approach should be adopted here but unsure yet how to make it work with the processors, since they can be dynamically configured, enabled, and disabled.
3. This tries to reuse config structs and defaults where possible, but counter-intuitively it doesn't reuse actual same settings as the rest of Tempo. There are a few reasons:  they are separate modules and probably should have separate config.  This processor's storage has to be distinct from the main storage area since the data is a little different.  Real storage will be RF3 (if configured), but the metrics-generator traffic is best-effort and only RF1.  

Additional changes:
1. I think this should fix the "duplication registration" bug that sometimes happens in metrics generator. 
2. Relocates some config defaults to be callable by the generator module.

Fixes n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/processor/localblocks/config.go
+++ b/modules/generator/processor/localblocks/config.go
@@ -23,7 +23,7 @@ type Config struct {
 func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet) {
 	cfg.FlushCheckPeriod = 5 * time.Second
 	cfg.MaxTraceIdle = 5 * time.Second
-	cfg.MaxBlockDuration = 15 * time.Minute
+	cfg.MaxBlockDuration = 1 * time.Minute
 	cfg.MaxBlockBytes = 500_000_000
-	cfg.CompleteBlockTimeout = time.Hour
+	cfg.CompleteBlockTimeout = 5 * time.Minute
 }
--- a/modules/generator/processor/localblocks/config.go
+++ b/modules/generator/processor/localblocks/config.go
@@ -17,7 +17,7 @@ type Config struct {
 	Block                *common.BlockConfig   `yaml:"block"`
 	Search               *tempodb.SearchConfig `yaml:"search"`
 	FlushCheckPeriod     time.Duration         `yaml:"flush_check_period"`
-	MaxTraceIdle         time.Duration         `yaml:"trace_idle_period"`
+	TraceIdlePeriod      time.Duration         `yaml:"trace_idle_period"`
 	MaxBlockDuration     time.Duration         `yaml:"max_block_duration"`
 	MaxBlockBytes        uint64                `yaml:"max_block_bytes"`
 	CompleteBlockTimeout time.Duration         `yaml:"complete_block_timeout"`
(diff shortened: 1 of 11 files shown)


knylander-grafana:
Will we need to update the metrics-generator docs for this?

modules/generator/processor/localblocks/config.go
```diff
@@ -0,0 +1,39 @@
...
+
+const (
+	Name = "local-blocks"
+)
+
+type Config struct {
+	Block                *common.BlockConfig   `yaml:"block"`
+	Search               *tempodb.SearchConfig `yaml:"search"`
+	FlushCheckPeriod     time.Duration         `yaml:"flush_check_period"`
+	MaxTraceIdle         time.Duration         `yaml:"trace_idle_period"`
```
zalegrala, line 20: Nit: do we want to align the field tag with the name?  This is the only one that doesn't match.

modules/generator/processor/localblocks/processor.go
```diff
@@ -0,0 +1,402 @@
...
+	}
+
+	p := &Processor{
+		Cfg:            cfg,
+		tenant:         tenant,
+		wal:            wal,
+		walBlocks:      map[uuid.UUID]common.WALBlock{},
+		completeBlocks: map[uuid.UUID]common.BackendBlock{},
+		liveTraces:     newLiveTraces(),
+		closeCh:        make(chan struct{}),
```
zalegrala, line 57: Just thinking outloud, not blocking; do we prefer a close chan over a context cancellation?

zalegrala (approved):
Approved with a couple comments.  Nice work.

modules/generator/processor/localblocks/config.go
mdisibio: Good catch, I copied from https://github.com/grafana/tempo/blob/main/modules/ingester/config.go#L23.

modules/generator/processor/localblocks/processor.go
mdisibio: I like the chose channel for simplicity, but you make a good point that they have different behaviors.  Right now the processor would finish any in-progress flush or block conversion before shutting down, whereas context cancellation would immediately break out because it would be passed along to the i/o operations.

mdisibio:
> Will we need to update the metrics-generator docs for this?

@knylander-grafana Yes we will need new docs for this processor and the API coming in later PRs.

PR #2385, opened 2023-04-27
Update deployment.md

MattiasSegerdahl:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/sources/tempo/setup/deployment.md
+++ b/docs/sources/tempo/setup/deployment.md
@@ -1,6 +1,6 @@
 ---
 title: Plan your Tempo deployment
-menuTitle: Plan your deployment
+menuTitle: Plan your Tempo deployment
 aliases:
   - /docs/tempo/deployment
   - /docs/tempo/deployment/deployment
@@ -28,7 +28,7 @@ descriptions of the components.
 
 To enable this mode, `-target=all` is used, which is the default.
 
-Find docker-compose deployment examples at:
+Find docker-compose deployment examples at the following links:
 
 - [https://github.com/grafana/tempo/tree/main/example/docker-compose/local](https://github.com/grafana/tempo/tree/main/example/docker-compose/local)
 - [https://github.com/grafana/tempo/tree/main/example/docker-compose/s3](https://github.com/grafana/tempo/tree/main/example/docker-compose/s3)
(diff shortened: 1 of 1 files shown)


docs/sources/tempo/setup/deployment.md
```diff
@@ -28,7 +28,7 @@ descriptions of the components.
 
 To enable this mode, `-target=all` is used, which is the default.
 
-Find docker-compose deployment examples at:
+Find docker-compose deployment examples at the following links:
```
knylander-grafana, line 31:
```suggestion
Here are two docker-compose deployment examples:
```

docs/sources/tempo/setup/deployment.md
```diff
@@ -39,12 +39,11 @@ Monolithic mode can be horizontally scaled out.
...
 Horizontal scale out is achieved by instantiating more than one process, with each having `-target` set to `scalable-single-binary`.
 
-This mode offers some flexibility of scaling without the configuration complexity of the full
-microservices deployment.
+This mode offers some scaling flexibility without the configuration complexity of the full microservices deployment.
 
 Each of the `queriers` perform a DNS lookup for the `frontend_address` and connect to the addresses found within the DNS record.
 
-Find a docker-compose deployment example at:
+Find a docker-compose deployment example at the following link:
```
knylander-grafana, line 46:
```suggestion
Here is a docker-compose deployment example:
```

docs/sources/tempo/setup/deployment.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Plan your Tempo deployment
-menuTitle: Plan your deployment
+menuTitle: Plan your Tempo deployment
```
knylander-grafana, line 3: In this case, we're not using Tempo in the menu title since we're in Tempo documentation. It should be obvious to the user that this is for Tempo deployment. The title: element should actually be updated to remove Tempo, as well as the page title.  
```suggestion
menuTitle: Plan your deployment
```

docs/sources/tempo/setup/deployment.md
knylander-grafana: If you have time, changing the sentence to include a brief description of the docker-compose example with the link would be better than my suggested revision above.

knylander-grafana (changes requested):
Thank you for submitting updates to the documentation. I have some minor suggestions to the PR.

docs/sources/tempo/setup/deployment.md
MattiasSegerdahl: I think including the link is better as the target and method might change

PR #2399, opened 2023-04-28
Metrics-generator customizable ring port

mdisibio:
**What this PR does**:
Adds the ability to override the port that metrics-generator advertises for itself in the ring.  This functionality is needed for certain deployments that need to route distributor->generator traffic through a proxy.  This functionality already exists for the ingester and compactor but was missing from the metrics-generator.

Didn't see any of these advanced options in the documentation, but they are in the config manifest and regenerated it.  Seems our yaml indentation changed so diff has a lot of noise, but should be readable ignoring whitespace.

Fixes n/a

**Checklist**
- [ ] Tests updated
- [x] Documentation added


joe-elliott (approved):
nice!

PR #2402, opened 2023-04-28
Upgrade parquet to fix boolean storage (#2400)

joe-elliott:
Backports #2400 to v2.1


mdisibio (approved):
👍 lgtm but I think the changelog isn't right

PR #2407, opened 2023-04-28
Use TLS config from `dskit`

### LINKED ISSUE #1914: Panic in S3 Setup
Trying to setup Tempo to use a local object store setup from Nutanix Objects which is S3 API compatible (Velero and others all work with it as S3) - using `docker.io/grafana/tempo:1.5.0` as the version

I get panics when starting Tempo components. i.e.
```
k logs pod/tempo-ingester-1 -n grafana                                                                                                                                                                                                                                                                                                                                                                                            ✔ │ dre-admin-beta-onprem/vault ⎈ │ 14:37:24 
level=info ts=2022-11-28T14:37:38.499004916Z caller=main.go:191 msg="initialising OpenTracing tracer"
level=info ts=2022-11-28T14:37:38.531184092Z caller=main.go:106 msg="Starting Tempo" version="(version=, branch=HEAD, revision=fd5743d5d)"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xa0 pc=0x1790afe]

goroutine 1 [running]:
github.com/grafana/tempo/tempodb/backend/s3.createCore(0xc000150380, 0x0)
	/drone/src/tempodb/backend/s3/s3.go:360 +0x61e
github.com/grafana/tempo/tempodb/backend/s3.internalNew(0xc000150380, 0x1)
	/drone/src/tempodb/backend/s3/s3.go:79 +0x5d
github.com/grafana/tempo/tempodb/backend/s3.New(...)
	/drone/src/tempodb/backend/s3/s3.go:73
github.com/grafana/tempo/tempodb.New(0xc0000e6600, {0x250dd40?, 0xc0004bfae0})
	/drone/src/tempodb/tempodb.go:148 +0xe5
github.com/grafana/tempo/modules/storage.NewStore({{0xc000056bd0, 0xc0004ec780, 0xc0004bfa40, 0xc00050b5f0, 0x45d964b800, 0x32, 0x1, 0x2, 0x0, 0x0, ...}}, ...)
	/drone/src/modules/storage/store.go:51 +0x565
github.com/grafana/tempo/cmd/tempo/app.(*App).initStore(0xc00062e000)
	/drone/src/cmd/tempo/app/modules.go:279 +0xbb
github.com/grafana/dskit/modules.(*Manager).initModule(0xc0004a8570, {0x7ffcd8a4de2e, 0x8}, 0xc0001ba370?, 0x0?)
	/drone/src/vendor/github.com/grafana/dskit/modules/modules.go:120 +0x224
github.com/grafana/dskit/modules.(*Manager).InitModuleServices(0x1ccc0c0?, {0xc0005e5590, 0x1, 0x8?})
	/drone/src/vendor/github.com/grafana/dskit/modules/modules.go:92 +0x10c
github.com/grafana/tempo/cmd/tempo/app.(*App).Run(0xc00062e000)
	/drone/src/cmd/tempo/app/app.go:172 +0x1c5
main.main()
	/drone/src/cmd/tempo/main.go:108 +0x890
```

My configMap has this for `storage`

```
    storage:
      trace:
        backend: s3
        blocklist_poll: 5m
        cache: memcached
        local:
          path: /var/tempo/traces
        memcached:
          consistent_hash: true
          host: tempo-memcached
          service: memcached-client
          timeout: 500ms
        s3:
          access_key: <key>
          bucket: traces
          endpoint: objects.example.com
          insecure: true
          insecure_skip_verify: true
          secret_key: <secret_key>
        wal:
          path: /var/tempo/wal
```

Is this an issue with Objects, or have I misconfigured something ?

Thanks.

### LINKED ISSUE #2262: http_tls_config under S3 to support CA cert setting
is there any plan to add http_tls_config under storage.s3 like Loki?  This setting can help us fulfill zero trust stragegy.

### LINKED ISSUE #2373: Support specifying self-signed certificate to trust for object storage
I'd like to use Tempo with a self-hosted S3 compatible object storage solution that's configured with TLS, without having to resort to insecureSkipVerify.

I need to be able to configure tempo to use a specific CA bundle file with the list of certificates it should use when verifying the certificate for the S3 endpoint.

insecureSkipVerify

**Additional context**

Relevant code: https://github.com/grafana/tempo/blob/e92189167552b547d2b9feede5c59d34d35518e4/tempodb/backend/s3/s3.go#L365-L399

zalegrala:
Here we implement the initial TLS configuration options from dskit on the s3 configuration.  The change here is pretty minimal, but will set us up to make use of the other TLS common components, such as CA cert file loading and server name.  Currently only the s3 config is modified, but I can see this impacting the others as well later.

**What this PR does**:

* add tls.ClientConfig inline config
* add documentation for new supported TLS client configurations
* replace the minio TLS configuration with the one based on tls.ClientConfig
* **breaking change** rename `storage.trace.s3.insecure_skip_verify` config option

Fixes #2262
Fixes #1914
Fixes #2373

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/storage/config.go
+++ b/modules/storage/config.go
@@ -5,7 +5,6 @@ import (
 	"time"
 
 	"github.com/grafana/tempo/pkg/cache"
-
 	"github.com/grafana/tempo/pkg/util"
 	"github.com/grafana/tempo/tempodb"
 	"github.com/grafana/tempo/tempodb/backend"
@@ -78,6 +77,7 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
 	f.StringVar(&cfg.Trace.S3.Prefix, util.PrefixConfig(prefix, "trace.s3.prefix"), "", "s3 root directory to store blocks in.")
 	f.StringVar(&cfg.Trace.S3.Endpoint, util.PrefixConfig(prefix, "trace.s3.endpoint"), "", "s3 endpoint to push blocks to.")
 	f.StringVar(&cfg.Trace.S3.AccessKey, util.PrefixConfig(prefix, "trace.s3.access_key"), "", "s3 access key.")
+	f.StringVar(&cfg.Trace.S3.MinVersion, util.PrefixConfig(prefix, "trace.s3.min_version"), "VersionTLS12", "minimum version of TLS to use when connecting to s3.")
 	f.Var(&cfg.Trace.S3.SecretKey, util.PrefixConfig(prefix, "trace.s3.secret_key"), "s3 secret key.")
 	f.Var(&cfg.Trace.S3.SessionToken, util.PrefixConfig(prefix, "trace.s3.session_token"), "s3 session token.")
 	cfg.Trace.S3.HedgeRequestsUpTo = 2
--- a/modules/storage/config.go
+++ b/modules/storage/config.go
@@ -77,7 +77,7 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
 	f.StringVar(&cfg.Trace.S3.Prefix, util.PrefixConfig(prefix, "trace.s3.prefix"), "", "s3 root directory to store blocks in.")
 	f.StringVar(&cfg.Trace.S3.Endpoint, util.PrefixConfig(prefix, "trace.s3.endpoint"), "", "s3 endpoint to push blocks to.")
 	f.StringVar(&cfg.Trace.S3.AccessKey, util.PrefixConfig(prefix, "trace.s3.access_key"), "", "s3 access key.")
-	f.StringVar(&cfg.Trace.S3.MinVersion, util.PrefixConfig(prefix, "trace.s3.min_version"), "VersionTLS12", "minimum version of TLS to use when connecting to s3.")
+	f.StringVar(&cfg.Trace.S3.MinVersion, util.PrefixConfig(prefix, "trace.s3.tls_min_version"), "VersionTLS12", "minimum version of TLS to use when connecting to s3.")
 	f.Var(&cfg.Trace.S3.SecretKey, util.PrefixConfig(prefix, "trace.s3.secret_key"), "s3 secret key.")
 	f.Var(&cfg.Trace.S3.SessionToken, util.PrefixConfig(prefix, "trace.s3.session_token"), "s3 session token.")
 	cfg.Trace.S3.HedgeRequestsUpTo = 2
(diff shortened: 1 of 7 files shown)


knylander-grafana (commented):
Thank you for adding doc.

modules/storage/config.go
```diff
@@ -78,6 +77,7 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
 	f.StringVar(&cfg.Trace.S3.Prefix, util.PrefixConfig(prefix, "trace.s3.prefix"), "", "s3 root directory to store blocks in.")
 	f.StringVar(&cfg.Trace.S3.Endpoint, util.PrefixConfig(prefix, "trace.s3.endpoint"), "", "s3 endpoint to push blocks to.")
 	f.StringVar(&cfg.Trace.S3.AccessKey, util.PrefixConfig(prefix, "trace.s3.access_key"), "", "s3 access key.")
+	f.StringVar(&cfg.Trace.S3.MinVersion, util.PrefixConfig(prefix, "trace.s3.min_version"), "VersionTLS12", "minimum version of TLS to use when connecting to s3.")
```
kousikmitra, line 80: `tls_min_version` to keep it consistent

tempodb/backend/s3/s3.go
```diff
@@ -372,8 +376,13 @@ func createCore(cfg *Config, hedge bool) (*minio.Core, error) {
 		return nil, errors.Wrap(err, "create minio.DefaultTransport")
 	}
 
-	if cfg.InsecureSkipVerify {
-		customTransport.TLSClientConfig.InsecureSkipVerify = true
+	tlsConfig, err := cfg.GetTLSConfig()
```
joe-elliott, line 379: how does this work? i'm reading the code here:

https://github.com/grafana/dskit/blob/main/crypto/tls/tls.go#L87

and it looks like you have to set a reader somewhere?

Also, it looks like `GetTLSConfig()` always returns a value and so it will always be set on line 385 below. Is this ok? Should there be a way to not set this and just use defaults?

zalegrala: Good find.  That commit isn't in the branch currently, but it looks like was introduced here: https://github.com/grafana/dskit/pull/274

I assume the reader interface was just to provide a little more flexibility on where to read the config data.

The nil check is just a habit when I see a pointer that could be nil, just in case the implementation changes we're still being defensive.  I can remove.

I'll double check, but it looked to me like if we don't set the config items, then we get a mostly default tls config, which looks okay to me.  Also note, that what we got back from the minio tls code is non-default, and so we set the min version in our config `RegisterFlagsAndApplyDefaults` to account for this.

joe-elliott: ok, rereading this makes way more sense. I missed that `GetTLSConfig()` is being called on the composed struct above. What is returned here will be the product of the settings. Wuf.

joe-elliott (approved):
Nice one, Zach.

PR #2410, opened 2023-04-29
Support to filter traces using `!~`

### LINKED ISSUE #1990: [TraceQL] Support Not Regex (!~)
Add support for the NotRegex operator (!~).  It is part of the language grammar and parses, but returns an `operation unsupported` error.

kousikmitra:
**What this PR does**:

* Enables parsing `!~` in tracql
* Support for filtering values using negated regex pattern

Fixes #1990

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/parquetquery/predicate_test.go
+++ b/pkg/parquetquery/predicate_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/segmentio/parquet-go"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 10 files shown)


pkg/parquetquery/predicate_test.go
```diff
@@ -32,41 +33,153 @@ func (p *mockPredicate) KeepValue(parquet.Value) bool             { p.valCalled
...
+			},
+		},
+		{
+			testName:   "dictionary in the page header allows for skipping a page",
+			predicate:  NewSubstringPredicate("x"), // Not present in any values
+			keptChunks: 1,
+			keptPages:  0,
+			keptValues: 0,
+			writeData: func(w *parquet.Writer) { //nolint:all
+				type dictString struct {
```
mdisibio, line 69: Since this struct is now repeated in several cases, let's move it out and maybe rename it `testDictString`.  The key is that it is dictionary encoded which allows the predicate to skip pages.  Not sure what happened because now both `String` and `dictString` are dictionary encoded, but it's ok and let's just use `testDictString` for all 6 cases.

pkg/parquetquery/predicate_test.go
```diff
@@ -32,41 +33,153 @@ func (p *mockPredicate) KeepValue(parquet.Value) bool             { p.valCalled
...
+	}
+}
+
+func TestNewRegexInPredicate(t *testing.T) {
+	testCases := []predicateTestCase{
+		{
+			testName: "all chunks/pages/values inspected",
+			predicate: func() Predicate {
+				pred, err := NewRegexInPredicate([]string{"a.*"})
+				assert.NoError(t, err)
```
mdisibio, line 94: Can you switch these to `require`?  We favor that over `assert` for fail fast and early.

pkg/parquetquery/predicates.go
```diff
@@ -129,7 +140,7 @@ func (p *RegexInPredicate) keep(v *pq.Value) bool {
 
 	matched := false
 	for _, r := range p.regs {
-		if r.MatchString(s) {
+		if r.MatchString(s) == p.shouldMatch {
```
mdisibio, line 143: I've been wondering if there is a performance benefit in using `r.Match()` instead.  Mostly because it would allow us to use `v.ByteArray()` instead of `v.String()`.   What do you think about creating a `BenchmarkRegexPredicate` and trying it out?

mdisibio (commented):
Thanks a lot for this.  Overall looks great, just have a few small comments.

pkg/parquetquery/predicates.go
kousikmitra: With r.Match()
```
goos: darwin
goarch: arm64
pkg: github.com/grafana/tempo/pkg/parquetquery
BenchmarkRegexInPredicate-8   	   19226	     62875 ns/op
BenchmarkRegexInPredicate-8   	   18411	     64000 ns/op
BenchmarkRegexInPredicate-8   	   18625	     63660 ns/op
BenchmarkRegexInPredicate-8   	   19041	     66662 ns/op
BenchmarkRegexInPredicate-8   	   18672	     67335 ns/op
PASS
ok  	github.com/grafana/tempo/pkg/parquetquery	11.338s
```

with r.MatchString()
```
goos: darwin
goarch: arm64
pkg: github.com/grafana/tempo/pkg/parquetquery
BenchmarkRegexInPredicate-8   	   14379	     82846 ns/op
BenchmarkRegexInPredicate-8   	   14172	     82937 ns/op
BenchmarkRegexInPredicate-8   	   14335	     82151 ns/op
BenchmarkRegexInPredicate-8   	   14313	     85904 ns/op
BenchmarkRegexInPredicate-8   	   13190	     83861 ns/op
PASS
ok  	github.com/grafana/tempo/pkg/parquetquery	12.070s
```

kousikmitra: this is after turning off memoization

kousikmitra: * updated above with benchmem

mdisibio: Ah my bad I forgot that the reason we use MatchString is because the memoization map key.  Shame because those benchmarks are nice.  Thanks for investigating.

pkg/traceql/test_examples.yaml
```diff
@@ -1,258 +1,257 @@
...
+  - "{ true || false }"
+  - "{ 1 = 2 }"
+  - "{ 1 != 2 }"
+  - "{ 1 > 2 }"
+  - "{ 1 >= 2 }"
+  - "{ 1 < 2 }"
+  - "{ 1 <= 2 }"
+  - "{ -1 = 2 }"
   - '{ "test" =~ "test" }'
+  - '{ "test" !~ "test" }'
```
mdisibio, line 16: What was the rationale for the quoting change in this file?   Fine with it if there is a reason, but now there is now a mixture of quoting styles.  TraceQL also supports embedded backticks for strings, and this works for me locally and would let us update the remainders:  
```
"{ `test` =~ `test` }"
```

kousikmitra: sorry for this, it's cause of my editor config for yaml. I will revert this

kousikmitra: reverted

pkg/parquetquery/predicates.go
kousikmitra: @mdisibio check this out.

mdisibio: Thanks, I was just reading about the `map[string([]byte)]` lookup optimization as well.  My concern is that the byte slices from `v.ByteArray()` can't be held onto long-term because they point to the underlying page buffers and are reused ( unless you call `v.Clone()`).  So we need to dig into the go internals a bit and see if the string really does persist in the map or is just hashed.  Even if tests are passing, this might be tricky to catch.

kousikmitra: agree on that, I am also not much aware about it. I will look into `if the string really does persist in the map or is just hashed`

* also `v.String()` internally uses
```
	case ByteArray, FixedLenByteArray:
		return string(v.byteArray())
```

* and another impl, if we instead use `v.String()` for memoization and `v.ByteArray()` for matching. it still ~2x better than current implementation but ya the mem allocation will still be same as current one.
```
goos: darwin
goarch: arm64
pkg: github.com/grafana/tempo/pkg/parquetquery
BenchmarkRegexInPredicate-8   	   29047	     41418 ns/op	   48005 B/op	    1000 allocs/op
BenchmarkRegexInPredicate-8   	   28704	     41063 ns/op	   48005 B/op	    1000 allocs/op
BenchmarkRegexInPredicate-8   	   28605	     42268 ns/op	   48005 B/op	    1000 allocs/op
BenchmarkRegexInPredicate-8   	   29046	     41118 ns/op	   48005 B/op	    1000 allocs/op
BenchmarkRegexInPredicate-8   	   28389	     41852 ns/op	   48005 B/op	    1000 allocs/op
PASS
ok  	github.com/grafana/tempo/pkg/parquetquery	10.042s
```

kousikmitra:
e2e passed on my local, might be some flaky test

knylander-grafana (commented):
Thank you for adding doc.

kousikmitra:
> I'm good with the map[string] implementation as-is. There is no regression on current performance, so let's not hold up this awesome feature. We can follow up with performance improvements as needed.

Sure, then we are good to go 🚀

Issue #2307, opened 2023-04-06
Span sets returned by TraceQL should return the number of all matching spans


stoewer:
**Is your feature request related to a problem? Please describe.**

When the Search API finds traces via TraceQL queries, each returned `spanSet` contains the matching spans within each trace. The maximum number spans in the `spanSet` is currently limited by [spansPerSpanSet](https://github.com/grafana/tempo/blob/97daaacebb58e6e0005f1c380894082dd1081753/pkg/traceql/engine.go#L22). Therefore, the number of returned spans in the `spanSet` can be lower than the actual number of matches. The `spanSet` contains a field [`matched`](https://grafana.com/docs/tempo/latest/api_docs/#example-of-traceql-search) but this field is always set to the [number of returned spans](https://github.com/grafana/tempo/blob/97daaacebb58e6e0005f1c380894082dd1081753/pkg/traceql/engine.go#L134).

As a consequence it is not possible to know whether the returned spans are all spans that matched the query.

**Describe the solution you'd like**

The field `matched` in `spanSet` should contain the number of all spans that matched by the query.

**Additional context**

This feature also allows the UI to implement visual aides that indicate whether a span set contains all matched spans or not.

mghildiy:
I would like to take this one.

joe-elliott:
Sure. The spansets are passed into the query engine here:

https://github.com/grafana/tempo/blob/4cdf859fc1583cac4ebf6ba70afc668764813cce/pkg/traceql/engine.go#L47

Any spansets that have "passing spans" are trimmed down here. Really this should be configurable but is currently hardcoded to 3:

https://github.com/grafana/tempo/blob/4cdf859fc1583cac4ebf6ba70afc668764813cce/pkg/traceql/engine.go#L60

After the lines above the spansets go back into the fetch layer (with a max of 3 spans each) and are decorated with metadata such as span id, trace id, etc. Then they come out here and are translated into the object that is eventually unmarshalled to json and returned by the api:

https://github.com/grafana/tempo/blob/4cdf859fc1583cac4ebf6ba70afc668764813cce/pkg/traceql/engine.go#L88

[...]

PR #2411, opened 2023-04-30
fix2308: Add query parameter to Search API to configure spans-per-span-set in response

### LINKED ISSUE #2308: Add query parameter to Search API to configure the spans per span set

When the Search API returns traces from TraceQL queries, each returned `spanSet` contains the matching spans within each trace. The maximum number spans in the `spanSet` is limited by [spansPerSpanSet](https://github.com/grafana/tempo/blob/97daaacebb58e6e0005f1c380894082dd1081753/pkg/traceql/engine.go#L22), which is currently set to 3.

The `spansPerSpanSet` should be configurable with a query parameter to adjust the maximum number of matched spans returned in a `spanSet`.

**Additional context**

Together with #2307 this enables the UI to implement a feature to show more matched spans on demand.

mghildiy:
**Summary of changes**:

- Add query parameter spss to API /api/search in http.go
- set number of spans during filtering
- Update http_test.go for new query parameter
- Update engine_test.go
- Update instance_search_test.go

**Adds a query parameter to Search API to make spans-per-span-set in response configurable**:
Fixes #2308 

**Checklist**

- [http_test.go, engine_test.go,instance_search_test.go ] Tests updated
- [docs/sources/tempo/api_docs/_index.md] Documentation added
- [CHANGELOG.md] updated


cmd/tempo-serverless/handler.go
```diff
@@ -110,7 +110,7 @@ func Handler(r *http.Request) (*tempopb.SearchResponse, *HTTPError) {
 	var resp *tempopb.SearchResponse
 
 	if api.IsTraceQLQuery(searchReq.SearchReq) {
-		engine := traceql.NewEngine()
+		engine := traceql.NewEngineWithSpansPerSpanSet(int(searchReq.SearchReq.SpansPerSpanSet))
```
joe-elliott, line 113: Let's have `NewEngine()` take the spans per spanset parameter instead of making a new method. There are 2 places in `instance_search` that will also need this parameter for it to work.

pkg/api/http.go
```diff
@@ -23,13 +23,14 @@ import (
...
-	urlParamStart       = "start"
-	urlParamEnd         = "end"
+	urlParamQuery           = "q"
+	urlParamTags            = "tags"
+	urlParamMinDuration     = "minDuration"
+	urlParamMaxDuration     = "maxDuration"
+	urlParamLimit           = "limit"
+	urlParamStart           = "start"
+	urlParamEnd             = "end"
+	urlParamSpansPerSpanSet = "spansPerSpanSet"
```
joe-elliott, line 33: Tempo can make 10s or even 100s of thousands of jobs for a given query. Is there a shorter parameter name we can use that will make sense to reduce bandwidth between components for distributing jobs?

pkg/tempopb/tempo.proto
```diff
@@ -54,6 +54,7 @@ message SearchRequest {
   uint32 end = 6;
   // TraceQL query
   string Query = 8;
+  uint32 SpansPerSpanSet = 9;
```
joe-elliott, line 57: maybe a uint16? i doubt anyone will ever want more than 2^16 spans per spanset. 2^8 is tempting :)

joe-elliott (commented):
This is looking good but we need a few tweaks before we can get it in. A few other things

- needs a changelog
- vendor-check is failing. try running `make gen-proto` locally. you can also run `make vendor-check` locally to make sure things are in order.
- please add the new query param here: https://github.com/grafana/tempo/blob/main/docs/sources/tempo/api_docs/_index.md?plain=1#L158

If you have noticed CI is a mess at the moment. We are working on it, but for now if things are passing locally feel free to submit and i will rerun CI.

Thanks for giving this a shot!

pkg/api/http.go
mghildiy: spss?

cmd/tempo-serverless/handler.go
mghildiy: I see three invocations of NewEngine in instance_search, 2 of which has access to request. Third one desn't have, so I guess I can pass default value of 3 there.

mghildiy:
I have made the changes as suggested above. But engine_test:TestEngine_Execute still fails because what filter function does is not reflected in the final response, for eactly the same reason as discussed [here](https://github.com/grafana/tempo/issues/2307).

pkg/traceql/engine.go
```diff
@@ -20,12 +20,18 @@ type Engine struct {
...
 
-func NewEngine() *Engine {
+func NewEngine(spss int) *Engine {
 	return &Engine{
-		spansPerSpanSet: 3, // TODO make configurable
+		spansPerSpanSet: spss,
 	}
 }
 
+/*func NewEngineWithSpansPerSpanSet(spansPerSpanSet int) *Engine {
```
joe-elliott, line 29: let's just remove this

cmd/tempo-serverless/handler.go
joe-elliott: I looked a little deeper and I'm wondering why we don't just use the value on the `searchReq`? We pass it in when we call `ExecuteSearch`. Can't we just the value on that struct?

pkg/api/http.go
joe-elliott: works for me

joe-elliott (commented):
> I have made the changes as suggested above. But engine_test:TestEngine_Execute still fails because what filter function does is not reflected in the final response, for eactly the same reason as discussed https://github.com/grafana/tempo/issues/2307.

This should work though b/c every span returned by the filter function is decorated with metadata by the second pass of the fetch layer. I don't think we'll have the same issue here.

pkg/traceql/engine.go
mghildiy: Very strange. I infact removed it and committed.

cmd/tempo-serverless/handler.go
mghildiy: You mean to add spanperspanset filed in SearchBlockRequest instead of SearchRequest?

mghildiy:
Now I understand the cause of the test failure. Its down to the way we are using those filtered spans in mocked iterator defined in engine_test. There we seem to be not correctly assigning them to the response.

mghildiy:
Test failures here made me to try to run tests locally, and I see many test failures in e2e, wal, OTHERS etc. Need to look in more details why those failures happening.

joe-elliott:
> So Matched is actually not exactly the number of spans matched in storage.

Right, this is what you were looking at fixing previously, but it is going to require some additional work that I was going to do anyway when implementing grouping and select. It's on the list to fix!

> Test failures here made me to try to run tests locally, and I see many test failures in e2e, wal, OTHERS etc. Need to look in more details why those failures happening.

Yeah, our CI is not great at the moment. @mapno is looking into cleaning it up some.

mghildiy:
> Yeah, our CI is not great at the moment. @mapno is looking into cleaning it up some.

It would be really helpful if makefile can be made window compatible.

stoewer:
> It would be really helpful if makefile can be made window compatible

Have you considered using [WSL](https://learn.microsoft.com/en-us/windows/wsl/install) for development?

There also is a [vscode plugin](https://code.visualstudio.com/docs/remote/wsl) to seamlessly develop using WSL. I have not tried it myself as I don't have a windows environment - but it looks worth trying.

mghildiy:
> > It would be really helpful if makefile can be made window compatible
> 
> Have you considered using [WSL](https://learn.microsoft.com/en-us/windows/wsl/install) for development?
> 
> There also is a [vscode plugin](https://code.visualstudio.com/docs/remote/wsl) to seamlessly develop using WSL. I have not tried it myself as I don't have a windows environment - but it looks worth trying.

Yeah, I tried with WSL, but things become pretty slow with it.
Anyways, I have got hold of a mac machine, so not an issue anymore.

mghildiy:
I ran docker-compose for local, referring to a tempo image build with local changes.
And on UI I see:

image

I don't see the new query parameter I added for spans-pers-span-set. It seems I need to make UI changes too.

mghildiy:
> > I don't see the new query parameter I added for spans-pers-span-set. It seems I need to make UI changes too.
> 
> The UI will not automatically update for added query parameters. Once this PR is in we will submit an issue in the [grafana](https://github.com/grafana/grafana) repo to support the new param. If you are interested in working on it, you'd be more than welcome to. If not, the team will get to it when they can.

Ok, got it. Yup, I am interested on Grafana UI work.
But I am not sure if any more changes are needed on this ticket now.

joe-elliott:
There is one change I'd like you to make. Let's remove the parameter from `NewEngine()` and just use the value passed into `ExecuteSearch`:

image

There's no need to construct the engine with this value since we're already passing it into the only place that uses it.

mghildiy:
So I was able to run vendor-check locally courtesy input i got on slcak channel.
Issue is that proto doesnt seem to allow uint8, or even uint16. I would need to use uint32 instead.

mghildiy:
I am not able to reproduce failure of test-with-cover-others locally.

mghildiy:
test-with-cover-tempodb seems ot be flaky. Sometimes it fails, sometimes not.

joe-elliott:
> test-with-cover-tempodb seems to be flaky. Sometimes it fails, sometimes not.

Yup. We are working on it, but it's really struggling right now. If anyone is interested in trying to improve some of those tests please do.

The bigger problem here is:
image
for some reason the CLA job is not kicking off? Maybe due to GH issues the past few days?

fwiw, i think this is ready to merge, i'm just waiting on CI to work itself out.

mghildiy:
I am not sure why it shows so, but I have signed CLA few times.

joe-elliott:
> I am not sure why it shows so, but I have signed CLA few times.

The job is not even kicking off. It has nothing to do with whether or not you've signed the CLA. One option is to push an empty commit and see if that kick starts it. Otherwise we can wait for it to give up and force restart it. As long as its pending I can't do anything.

mghildiy:
> If anyone is interested in trying to improve some of those tests please do.

I can debug it. 
Any specific tests that can be directly looked into, as there are many tests in 'test-with-cover-tempodb'.
Or any approach I can follow to zero-in on faulty ones? Currently I see tests failing randomly.

PR #2414, opened 2023-05-01
Remove tenant_header_key from config

### LINKED ISSUE #2381: [tempo-query] Remove `tenant_header_key` config header option
It's highly misleading. While you can configure the plugin to use different headers to pass the tenant ID, Jaeger Query will only [propagate the bearer token](https://github.com/jaegertracing/jaeger/blob/main/plugin/storage/grpc/shared/grpc_client.go#L44) to the storage plugin. So, the only real option is using the default `bearer.token`. We should remove that config option.

kousikmitra:
**What this PR does**:

Removes `tenant_header_key` option from config

Fixes #2381

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## main / unreleased
 
+* [ENHANCEMENT] Remove `tenant_header_key` option from `tempo-query` config [#2414](https://github.com/grafana/tempo/pull/2414) (@kousikmitra)
```
mdisibio, line 3: I think this should be `[CHANGE]`

mdisibio (changes requested):
Hi thanks for the cleanup.  Just one small nit in the changelog and LGTM

CHANGELOG.md
kousikmitra: Updated

mapno (approved):
LGTM. Can you fix the conflict?

mdisibio (approved):

PR #2416, opened 2023-05-01
Remove tolerate_failed_blocks from trace by id path

joe-elliott:
**What this PR does**:
The original idea was to allow installations with very large numbers of blocks to still return success even if a few blocks failed. In practice we found that the negative interaction with the query frontend retry and the fact that this sometimes just hid corrupt blocks made this an operational anti-pattern.

**Checklist**
- [x] Tests updated
- [x] Documentation added


knylander-grafana (commented):
Thank you for updating docs.

PR #2417, opened 2023-05-01
Increase query shards

### LINKED ISSUE #1905: Allow more than 256 max shards on trace by id lookup
This function: 

https://github.com/grafana/tempo/blob/cf6a9337e7ce58b75bc9e98b50db50c5953eaf4f/modules/frontend/tracebyidsharding.go#L228

Is responsible for breaking up the the guid space into a smaller number of evenly sized shards. Currently, due to the way it is written, the maximum number of possible shards is 256. Let's improve this function to allow for an arbitrary number of shards. 

This constant will also have to change:
https://github.com/grafana/tempo/blob/cf6a9337e7ce58b75bc9e98b50db50c5953eaf4f/modules/frontend/tracebyidsharding.go#L27

and it should be confirmed that this method:
https://github.com/grafana/tempo/blob/cf6a9337e7ce58b75bc9e98b50db50c5953eaf4f/modules/frontend/tracebyidsharding.go#L196

can handle the new number of shards correctly.

joe-elliott:
**What this PR does**:
Improves the `createBlockBoundaries()` function to accept more than 255 shards. Also fixes a bug where, if the query_shards configured was not evenly divisible by the total range, it would place all of the "extra space" in the last bucket.

Fixes #1905 

**Checklist**
- [x] Tests updated
- [x] Documentation added


modules/frontend/tracebyidsharding_test.go
```diff
@@ -38,22 +40,53 @@ func TestCreateBlockBoundaries(t *testing.T) {
 			name:        "multiple shards",
 			queryShards: 4,
 			expected: [][]byte{
-				{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0},  // 0
-				{0x3f, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}, // 0x3f = 255/4 * 1
-				{0x7e, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}, // 0x7e = 255/4 * 2
-				{0xbd, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}, // 0xbd = 255/4 * 3
+				{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0},
+				{0x40, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0},
```
mdisibio, line 44: Good catch, after staring at `0x3f` vs `0x40` I agree that was a bug and `0x40` is the expected evenly divided value.  The entries define the start of each range, but the last `0xFFFF...` is different and defines the end of the second-to-last entry.

modules/frontend/tracebyidsharding_test.go
```diff
@@ -38,22 +40,53 @@ func TestCreateBlockBoundaries(t *testing.T) {
...
+				{0x80, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0},
+				{0xc0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0},
 				{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff},
 			},
 		},
+		{
+			name:        "large number of evenly divisible shards",
+			queryShards: 255,
```
mdisibio, line 52: Is this right?  I think 256 should be evenly divisible?

modules/frontend/tracebyidsharding_test.go
```diff
@@ -38,22 +40,53 @@ func TestCreateBlockBoundaries(t *testing.T) {
...
+		},
+		{
+			name:        "large number of not evenly divisible shards",
+			queryShards: 1111,
```
mdisibio, line 56: Could this be tested with hardcoded expected values and a smaller non even value like 3 shards?  Instead of the code down below.

modules/frontend/tracebyidsharding.go
```diff
@@ -259,14 +260,25 @@ func createBlockBoundaries(queryShards int) [][]byte {
...
 	for i := 0; i < queryShards; i++ {
-		binary.LittleEndian.PutUint64(blockBoundaries[i][:8], (MaxUint/uint64(queryShards))*uint64(i))
-		binary.LittleEndian.PutUint64(blockBoundaries[i][8:], 0)
+		binary.BigEndian.PutUint64(blockBoundaries[i][:8], boundary)
+		binary.BigEndian.PutUint64(blockBoundaries[i][8:], 0)
+
+		boundary += bucketSz
+		if numLarger != 0 {
+			numLarger--
+			boundary++
```
mdisibio, line 276: 👍 I like this way of distributing the extras.

mdisibio (approved):
LGTM and I look forward to configuring 6 digit query shard values. 🚀  Left a few non-blocking questions/comments.

modules/frontend/tracebyidsharding.go
joe-elliott: i stole this from stack overflow :)

modules/frontend/tracebyidsharding_test.go
mghildiy: Just curious, how is all this tested apart from the unit tests?

joe-elliott: The various shard sizes are not directly tested anywhere else that I'm aware of, but we do test the entire e2e stack here:

https://github.com/grafana/tempo/tree/main/integration/e2e

mghildiy: So now that number of shards has changed, would it require changes in e2e tests too?

joe-elliott: The number of shards hasn't changed, but the maximum configurable number of shards has. We choose not to test every combination of parameters in the e2e tests b/c it would make for an exceptionally long test. The e2e tests are "smoke tests" to make sure that the basic functionality of Tempo is working.

PR #2429, opened 2023-05-04
Replace "go test" with "gotestsum" in Makefile

zalegrala:
**What this PR does**:

Here is a proposal to use `gotestsum` in place of `go test` for the purpose of improving the output of the test results.  Color adds a lot of context, and the particular format chosen here is one of many available.

- Create `tools/` package and add initial go.mod
- Imports new `make` targets for installing tools from sub directory
- Include the `tools` target as a `make` dependency in several CI job tasks
- Propose structure to support additional CI tools in the future, managed as go dependencies in `tools/`

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


Makefile
```diff
@@ -94,37 +94,37 @@ benchmark:
 
 # Not used in CI, tests are split in pkg, tempodb, tempodb-wal and others in CI jobs
 .PHONY: test-with-cover
-test-with-cover: test-serverless
+test-with-cover: tools test-serverless
```
zalegrala, line 97: Alternative to the make target dependency here, we could add `make tools` in the CI as a step.

Makefile
```diff
@@ -338,3 +338,7 @@ ifndef DRONE_TOKEN
 	$(error DRONE_TOKEN is not set, visit https://drone.grafana.net/account)
 endif
 	DRONE_SERVER=https://drone.grafana.net drone sign --save grafana/tempo .drone/drone.yml
+
+
+# Import fragments
+include build/tools.mk
```
zalegrala, line 344: I thought this was a good time to start breaking up the make config a bit, but if we like it all in one, I can move these targets into the `Makefile`.

zalegrala:
Feel free to review a test run here, as the output will be noticeable.

tools/tools.go
```diff
@@ -0,0 +1,8 @@
+//go:build tools
+
+package tools
+
+import (
+	_ "github.com/psampaz/go-mod-outdated"
```
joe-elliott, line 6: what is go mod outdated?

joe-elliott (approved):
i like this. one question below. also can we add golangci-lint and jsonnetfmt to the tools list?

does gofmt come with go? or should we add that to?

tools/tools.go
zalegrala: Shows you out dated modules in your go.mod file.  Pretty handy if you like to keep that sort of thing up to date.  Pretty annoying if you don't.  In this case, its only used as part of the tools module.

zalegrala:
I agree other tools can be slotted in here, perhaps in a follow up PR.  I like `gofumpt` from https://github.com/mvdan/gofumpt but we should chat before introducing it unless we make it automatic.  See my slack comment from earlier today :)

tools/go.mod
```diff
@@ -0,0 +1,25 @@
+module github.com/grafana/tempo/tools
+
+go 1.20
+
+require (
+	github.com/psampaz/go-mod-outdated v0.9.0
+	gotest.tools/gotestsum v1.10.0
```
electron0zero, line 7: q: why gotest.tools, and why not github? imo, I would trust github over gotest.tools to not serve me a bad or compromised verison of gotestsum :)

zalegrala: Not really sure, but its in their recommendation for `go install`, so that's what I'd put here.

zalegrala: And indeed their module is `gotest.tools` also.

zalegrala: Merged, since this wasn't a blocking comment, but if we want to discuss and revisit I'm open to it.

PR #2430, opened 2023-05-04
Query-frontend: allow unknown fields

joe-elliott:
**What this PR does**:
Currently if there is a mismatch between the json sent by the querier and the one expected by the query-frontend it will fail to parse. This was recently revealed with the rename of `inspectedBlocks` -> `totalBlocks`.

This PR allows the frontend to leniently parse the response from the querier which prevent this issue and simplifies rollouts.


ie-pham (approved):

zalegrala (approved):

zalegrala:
Interesting.

PR #2436, opened 2023-05-06
[ENHANCEMENT] log client ip to help identify which client is no org id

Komey:
**What this PR does**:
add log for client ip to help identify which client is no org id
Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


electron0zero (approved):

modules/distributor/receiver/middleware.go
```diff
@@ -61,12 +61,14 @@ func (m *multiTenancyMiddleware) Wrap(next consumer.Traces) consumer.Traces {
 			info := client.FromContext(ctx)
 			orgIDs := info.Metadata.Get(user.OrgIDHeaderName)
 			if len(orgIDs) == 0 {
-				log.Logger.Log("msg", "failed to extract org id from both grpc and HTTP", "err", err)
+				log.Logger.Log("msg", "failed to extract org id from both grpc and HTTP",
+					"err", err, "client", info.Addr.String())
```
mapno, line 65: This is causing a panic when `info.Addr == nil`. You need to add a check for that.

mapno (commented):
LGTM, but the panic needs to be fixed.

modules/distributor/receiver/middleware.go
Komey: HI @mapno, 
Thanks for the reviews, for version [v0.74.0](https://github.com/grafana/tempo/blob/main/go.mod#LL73C13-L73C13) of the opentelemetry-collector SDK, `info.Addr` is not a pointer. we should pay attention to future updates.

https://github.com/open-telemetry/opentelemetry-collector/blob/v0.74.0/client/client.go#L104

mapno: `net.Addr` is an interface, so it panics when undefined. See example https://go.dev/play/p/dx6a1BYB16J.

Komey: @mapno Thanks for the correction, the panics has been fixed

mapno (approved):
LGTM, thank you for the contribution.

PR #2438, opened 2023-05-07
Add e2e tests for jaeger query plugin

rubenvp8510:
**What this PR does**:
This PR adds an e2e test fot the existing jaeger query plugin.  

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


integration/util.go
```diff
@@ -143,7 +144,7 @@ func NewTempoQuerier() *e2e.HTTPService {
 
 	s := e2e.NewHTTPService(
 		"querier",
-		image,
+		queryImage,
```
mapno, line 147:
```suggestion
		image,
```


`querier != tempo-query`. Naming is confusing, but they're different components. The querier is a component in the read path, while tempo-query is an adapter to make Tempo work with Jaeger

integration/e2e/config-plugin-test.yaml
mapno, line 1: This config looks practically identical to `config-all-in-one-local.yaml`. Is it possible to use that one instead?

mapno (commented):
Thanks for the contribution. Leaving two comments, but LGTM overall.

integration/e2e/config-plugin-test.yaml
rubenvp8510: Sorry for the delay on this, and thanks for the review, I wanted to reuse the configuration, but I need to have a long flush period here due the services test, the original one had 1s.

integration/util.go
rubenvp8510: Fixed! thanks!

mapno (approved):
LGMT, thanks!

PR #2448, opened 2023-05-10
Keep blocks for longer in e2e test

mapno:
**What this PR does**:

Increases `ingester.complete_block_timeout` in e2e test config to reduce probability of race condition.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


joe-elliott (approved):
why 20s? why not 10m? or 1h?

hopefully this doesn't negatively impact any tests

mapno:
>why 20s? why not 10m? or 1h?

No reason, it seemed like a good value.

> hopefully this doesn't negatively impact any tests

The other test using this config should benefit from this change as well.

PR #2453, opened 2023-05-10
Copy tempo-query binary to /tempo-query

BertelBB:
- previous implementation forces tempo to run with root privileges in Kubernetes because it is also writing to files in /tmp directory
- this solution is inline to how other tempo components are built and ensures /tmp dir exists with mkdir -p

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo-query/Dockerfile
+++ b/cmd/tempo-query/Dockerfile
@@ -3,7 +3,8 @@ FROM jaegertracing/jaeger-query:1.43.0
 ENV SPAN_STORAGE_TYPE=grpc-plugin \
     GRPC_STORAGE_PLUGIN_BINARY=/tmp/tempo-query
 
-# This is silly, but it's important that tempo-query gets copied into /tmp
-#  b/c it forces a /tmp dir to exist which hashicorp plugins depend on
+# This is silly, but it's important b/c it forces a /tmp dir to exist which hashicorp plugins depend on
+RUN mkdir -p /tmp
+
 ARG TARGETARCH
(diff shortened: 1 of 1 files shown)


electron0zero (approved):
lgtm 👍🏼 

btw, tempo-query is an optional component and we sent a [PR](https://github.com/grafana/helm-charts/pull/2387#issue-1698226860) to disable it by default in helm chars

cmd/tempo-query/Dockerfile
```diff
@@ -1,9 +1,10 @@
 FROM jaegertracing/jaeger-query:1.43.0
 
 ENV SPAN_STORAGE_TYPE=grpc-plugin \
-    GRPC_STORAGE_PLUGIN_BINARY=/tmp/tempo-query
+    GRPC_STORAGE_PLUGIN_BINARY=/tempo-query
+
+# Ensure /tmp dir exists, hashicorp plugins need a /tmp dir to exist.
+RUN mkdir -p /tmp
 
-# This is silly, but it's important that tempo-query gets copied into /tmp
```
joe-elliott, line 6: you didn't like my cheeky comment ;) ?

joe-elliott: oh, @electron0zero suggested that!

BertelBB: Haha I personally liked it and that’s why I initially kept that part in there 😄

electron0zero: I was okay with either but wanted to sound like we know what we are doing 😆

PR #2462, opened 2023-05-11
Disable tempo-query by default in Jsonnet libs

electron0zero:
**What this PR does**:
tempo-query was required for Grafana version <7.5 for compatibility with jaeger-ui. grafana version <7.5 didn't have Tempo datasource, and we used jaeger datasource to query tempo via tempo-query.

Grafana 7.5 was released on Mar 25, 2021, which was 2+ year ago.

It is already disabled by default in helm-chart: https://github.com/grafana/helm-charts/pull/2387

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/operations/jsonnet-compiled/util/example/main.jsonnet
+++ b/operations/jsonnet-compiled/util/example/main.jsonnet
@@ -5,10 +5,10 @@ local tempo = import 'microservices/tempo.libsonnet';
 tempo {
   _images+:: {
     tempo: 'grafana/tempo:latest',
-    tempo_query: 'grafana/tempo-query:latest',
     tempo_vulture: 'grafana/tempo-vulture:latest',
   },
 
+  // generate with `tempo_query: true` to include tempo-query manifests
   _config+:: {
     namespace: 'tracing',
     compactor+: {
--- a/operations/jsonnet-compiled/util/example/main.jsonnet
+++ b/operations/jsonnet-compiled/util/example/main.jsonnet
@@ -6,9 +6,10 @@ tempo {
   _images+:: {
     tempo: 'grafana/tempo:latest',
     tempo_vulture: 'grafana/tempo-vulture:latest',
+    tempo_query: 'grafana/tempo-query:latest',
   },
 
-  // generate with `tempo_query: true` to include tempo-query manifests
+  // generate with `tempo_query.enabled: true` to include tempo-query manifests
   _config+:: {
     namespace: 'tracing',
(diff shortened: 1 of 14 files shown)


operations/jsonnet/microservices/config.libsonnet
```diff
@@ -18,6 +18,9 @@
     node_selector: null,
     ingester_allow_multiple_replicas_on_same_node: false,
 
+    // disable tempo-query by default
+    tempo_query: false,
```
mapno, line 22:
```suggestion
    tempo_query: {
        enabled: false,
    },
```

tempo-query has some TLS config options that should be configurable. This way, it opens the possibility of adding them.

mapno (commented):
Left a comment, but overall good change.

operations/jsonnet/microservices/config.libsonnet
electron0zero: makes sense, turned it into an object.

operations/jsonnet-compiled/util/example/main.jsonnet
```diff
@@ -5,10 +5,10 @@ local tempo = import 'microservices/tempo.libsonnet';
 tempo {
   _images+:: {
     tempo: 'grafana/tempo:latest',
-    tempo_query: 'grafana/tempo-query:latest',
     tempo_vulture: 'grafana/tempo-vulture:latest',
   },
 
+  // generate with `tempo_query: true` to include tempo-query manifests
```
mapno, line 11:
```suggestion
  // generate with `tempo_query.enabled: true` to include tempo-query manifests
```

operations/jsonnet-compiled/util/example/main.jsonnet
```diff
@@ -5,10 +5,10 @@ local tempo = import 'microservices/tempo.libsonnet';
 tempo {
   _images+:: {
     tempo: 'grafana/tempo:latest',
-    tempo_query: 'grafana/tempo-query:latest',
```
mapno, line 8: Maybe comment instead of delete?

mapno (approved):

operations/jsonnet-compiled/util/example/main.jsonnet
electron0zero: will keep it as is, this var is no-op when `tempo_query.enabled: false`

PR #2463, opened 2023-05-11
metrics-generator: use fanout storage to correctly update Prometheus remote write

kvrhdn:
**What this PR does**:

Fixes a bug in our usage of the Prometheus `agent.DB` and `remote.Storage` structures. When we append data onto `agent.DB` we also need to append it onto `remote.Storage` to ensure `prometheus_remote_storage_highest_timestamp_in_seconds` is updated ([append implementation](https://github.com/prometheus/prometheus/blob/5c5fa5c319fca713506fa144ec6768fddf00d466/storage/remote/write.go#L268
)).

This timestamp is used to determine how many shards are needed, since this metric wasn't updated desired shards is always set to 1, even if remote write is running behind and needs to catch up.

~~Fixes #<issue number>~~

**Checklist**
- [x] Tests updated
- [ ] ~~Documentation added~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ To make use of filtering, configure `autocomplete_filtering_enabled`.
 * [ENHANCEMENT] Add support for IPv6 [#1555](https://github.com/grafana/tempo/pull/1555) (@zalegrala)
 * [ENHANCEMENT] Add span filtering to spanmetrics processor [#2274](https://github.com/grafana/tempo/pull/2274) (@zalegrala)
 * [BUGFIX] tempodb integer divide by zero error [#2167](https://github.com/grafana/tempo/issues/2167) (@kroksys)
+* [BUGFIX] metrics-generator: use fanout storage to correctly update Prometheus remote write [#2463](https://github.com/grafana/tempo/issues/2463) (@kvrhdn)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ To make use of filtering, configure `autocomplete_filtering_enabled`.
 * [ENHANCEMENT] Add support for IPv6 [#1555](https://github.com/grafana/tempo/pull/1555) (@zalegrala)
 * [ENHANCEMENT] Add span filtering to spanmetrics processor [#2274](https://github.com/grafana/tempo/pull/2274) (@zalegrala)
 * [BUGFIX] tempodb integer divide by zero error [#2167](https://github.com/grafana/tempo/issues/2167) (@kroksys)
+* [BUGFIX] metrics-generator: use fanout storage to correctly update Prometheus remote write [#2463](https://github.com/grafana/tempo/issues/2463) (@kvrhdn)
```
bboreham, line 34: Will an end-user understand this?
Suggest to focus more on the effect, which I think is Tempo will now use multiple connections to send samples as the rate gets higher.

09jvilla:
fixes https://github.com/grafana/tempo-squad/issues/218 no?

CHANGELOG.md
kvrhdn: Good point thanks! I tried to improve it somewhat

joe-elliott (approved):
lgtm

PR #2468, opened 2023-05-12
Split Search and SearchTags Handler

electron0zero:
**What this PR does**:
Split Search and SearchTags Handler to fix `tempo_request_duration_seconds_count` metric.

Till now we were reusing SearchHandler for Search Tags requests, and because of that `tempo_request_duration_seconds_count{route=tempo_api_search}` metric for `/search` path includes SearchTags requests, and makes it harder to alert on search and search tags separately. 

Before this change, we were relaying on our code that directly searches ingesters when `start` and `end` is not supplied. Search Tags relied on that behaviour.

Now, it splits both handlers into two, and adds `start` and `end` to search only in ingesters if `start` and `end` was not supplied in Search requests. This keeps the same behavours from users, point of view but splits the code make it easy.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


cmd/tempo/app/modules.go
```diff
@@ -300,12 +301,12 @@ func (t *App) initQueryFrontend() (services.Service, error) {
...
 	// http search endpoints
 	t.Server.HTTP.Handle(addHTTPAPIPrefix(&t.cfg, api.PathSearch), searchHandler)
-	t.Server.HTTP.Handle(addHTTPAPIPrefix(&t.cfg, api.PathSearchTags), searchHandler)
-	t.Server.HTTP.Handle(addHTTPAPIPrefix(&t.cfg, api.PathSearchTagsV2), searchHandler)
-	t.Server.HTTP.Handle(addHTTPAPIPrefix(&t.cfg, api.PathSearchTagValues), searchHandler)
-	t.Server.HTTP.Handle(addHTTPAPIPrefix(&t.cfg, api.PathSearchTagValuesV2), searchHandler)
+	t.Server.HTTP.Handle(addHTTPAPIPrefix(&t.cfg, api.PathSearchTags), searchTagsHandler)
+	t.Server.HTTP.Handle(addHTTPAPIPrefix(&t.cfg, api.PathSearchTagsV2), searchTagsHandler)
+	t.Server.HTTP.Handle(addHTTPAPIPrefix(&t.cfg, api.PathSearchTagValues), searchTagsHandler)
+	t.Server.HTTP.Handle(addHTTPAPIPrefix(&t.cfg, api.PathSearchTagValuesV2), searchTagsHandler)
```
mapno, line 307: By the same logic, shouldn't we split tag value search as well?

electron0zero: Before this change, recent search was going directly to querier, instead of going via search sharder (where we calculate throughout and slo metrics), and was not accounted for in slo and throughout metrics.

before this split `/search` for recent data queries (search request without start and end), and tag searchs was sent directly to a querier without going via search sharder.

for the case of tags and tag values, we want to direcly go to querier, and we do not compute throughout and slo metrics on these endpoints. And because of that I think makes sense to just have single handler for both tags and tag values endpoint.

electron0zero: PR description was bit misleading, updated now.

joe-elliott (commented):
I'm good with this as is, but Mario had some comments. Will leave it unapproved until we resolve all concerns.

cmd/tempo/app/modules.go
mapno: It took me a couple of reads, but now I understand it. Thank you for the explanation.

pkg/api/search.go
```diff
@@ -12,6 +13,17 @@ func IsBackendSearch(r *http.Request) bool {
 	return q.Get(urlParamStart) != "" && q.Get(urlParamEnd) != ""
 }
 
+func IsMissingStartOrEnd(r *http.Request) bool {
```
mapno, line 16: Isn't this identical to `!IsBackendSearch()`? Can we use one or the other?

mapno (approved):
LGTM, just have a small nit.

modules/frontend/searchsharding.go
```diff
@@ -367,11 +372,16 @@ func (s *searchSharder) backendRequests(ctx context.Context, tenantID string, pa
...
 }
 
-// queryIngesterWithin returns a new start and end time range for the backend as well as an http request
+// ingesterRequest returns a new start and end time range for the backend as well as an http request
 // that covers the ingesters. If nil is returned for the http.Request then there is no ingesters query.
 // since this function modifies searchReq.Start and End we are taking a value instead of a pointer to prevent it from
 // unexpectedly changing the passed searchReq.
 func (s *searchSharder) ingesterRequest(ctx context.Context, tenantID string, parent *http.Request, searchReq tempopb.SearchRequest) (*http.Request, error) {
+	// request without start or end, search only in ingester
+	if searchReq.Start == 0 || searchReq.End == 0 {
```
joe-elliott, line 381: i like this fix. i think it nicely combines the old idea of "SearchRecent" with existing search.

pkg/api/search.go
```diff
@@ -6,7 +6,7 @@ import (
 	"github.com/grafana/tempo/pkg/tempopb"
 )
 
-// IsBackendSearch returns true if the request has a start, end and tags parameter and is the /api/search path
+// IsBackendSearch returns true if the request has a start, and end parameter and is the /api/search path
 func IsBackendSearch(r *http.Request) bool {
```
joe-elliott, line 10: is this method still necessary? maybe we can just remove it?

modules/frontend/searchsharding.go
```diff
@@ -242,6 +246,7 @@ func (s searchSharder) RoundTrip(r *http.Request) (*http.Response, error) {
 	throughput := float64(overallResponse.response.Metrics.InspectedBytes) / reqTime.Seconds()
 	searchThroughput.WithLabelValues(tenantID).Observe(throughput)
 
+	// also this is being called?? we log here??
```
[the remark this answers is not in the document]
electron0zero: missed in cleanup

electron0zero:
This is ready for final review, I added some tests around WAL, and other places where I was worried about bugs :)

joe-elliott (approved):

joe-elliott:
lgtm, i think we landed in a good spot with this one.

PR #2475, opened 2023-05-15
Add range in traceql search for vulture

ie-pham:
**What this PR does**: Vulture was doing test traceql queries for but since time range was not added in the query, it was getting false errors

Fixes https://github.com/grafana/tempo-squad/issues/240

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


ie-pham:
tested in ops overnight and things look good

cmd/tempo-vulture/main.go
```diff
@@ -262,6 +263,7 @@ func pushMetrics(metrics traceMetrics) {
 	metricTracesErrors.WithLabelValues("incorrectresult").Add(float64(metrics.incorrectResult))
 	metricTracesErrors.WithLabelValues("missingspans").Add(float64(metrics.missingSpans))
 	metricTracesErrors.WithLabelValues("notfound_search").Add(float64(metrics.notFoundSearch))
+	metricTracesErrors.WithLabelValues("notfound_search").Add(float64(metrics.notFoundTraceQL))
```
joe-elliott, line 266: we should change the reason label from `notfound_search` to `notfound_traceql` or something

pkg/util/client.go
```diff
@@ -171,3 +171,13 @@ func (c *Client) SearchTraceQL(query string) (*tempopb.SearchResponse, error) {
 
 	return m, nil
 }
+
+func (c *Client) SearchTraceQLWithRange(query string, start int64, end int64) (*tempopb.SearchResponse, error) {
+	m := &tempopb.SearchResponse{}
+	_, err := c.getFor(c.BaseURL+"/api/search?q="+url.QueryEscape(query)+"&start="+strconv.FormatInt(start, 10)+"&end="+strconv.FormatInt(end, 10), m)
```
joe-elliott, line 177: i think there are more correct ways to build a query string like this. likely using the url object and then asking it to encode all at once, but this is fine given what we're doing here

joe-elliott (approved):
approving to unblock but please change the metric reason.

pkg/util/client.go
```diff
@@ -164,10 +164,33 @@ func (c *Client) QueryTrace(id string) (*tempopb.Trace, error) {
...
+	m := &tempopb.SearchResponse{}
+	_, err := c.getFor(c.buildQueryURL("q", query, start, end), m)
+	if err != nil {
+		return nil, err
+	}
+
+	return m, nil
+}
+
+func (c *Client) buildQueryURL(queryType string, query string, start int64, end int64) string {
```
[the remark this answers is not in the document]
joe-elliott: up to you. if it works it works. i do think this is "more correct"

PR #2478, opened 2023-05-15
Add spanset attributes

joe-elliott:
**What this PR does**:
This PR does some cleanup in preparation for passing arbitrary attributes through the Fetch method.
- Adds an arbitrary attributes map to the spansets
- Adds "matched" into this attributes map
- Adds aggregate information to the spansets

In order to keep these PRs more easily reviewable I'm going to cut this off here. Next I will work on actually preserving these through the fetch layer.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/traceql/ast.go
```diff
@@ -443,6 +443,89 @@ func (s Static) Equals(other Static) bool {
...
+			return 1
+		} else if s.Status < other.Status {
+			return -1
+		}
+	case TypeKind:
+		if s.Kind > other.Kind {
+			return 1
+		} else if s.Kind < other.Kind {
+			return -1
+		}
```
mapno, line 499: Does this comparison make sense for `status` and `kind`? What does a status bigger than another mean?

mapno (approved):

pkg/traceql/ast.go
joe-elliott: I agree that it's strange and one status isn't really comparable to another status in this way. The only reason I included them was so that if we were to sort Statics like Kind/Status using this function the results would be deterministic.

PR #2480, opened 2023-05-16
Local blocks processor query all blocks

mdisibio:
**What this PR does**:
This PR is part of a wider initiative to perform on-demand metrics calculations.  Updates the `/generator/api/metrics` API to query all blocks.  Includes some basic caching with simplelru that definitely needs to be replaced with something more robust.  Finally added a test but it's a good one 😬 

Fixes n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/processor/localblocks/processor.go
+++ b/modules/generator/processor/localblocks/processor.go
@@ -4,12 +4,12 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"math"
 	"sort"
 	"sync"
 	"time"
 
 	"github.com/go-kit/log/level"
+	"github.com/golang/groupcache/lru"
 	"github.com/google/uuid"
 	gen "github.com/grafana/tempo/modules/generator/processor"
 	"github.com/grafana/tempo/pkg/model"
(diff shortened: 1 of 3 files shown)


modules/generator/processor/localblocks/processor.go
```diff
@@ -200,62 +204,105 @@ func (p *Processor) metricLoop() {
...
 
 func (p *Processor) GetMetrics(ctx context.Context, req *tempopb.SpanMetricsRequest) (*tempopb.SpanMetricsResponse, error) {
+	p.blocksMtx.RLock()
+	defer p.blocksMtx.RUnlock()
 
-	fetcher := traceql.NewSpansetFetcherWrapper(func(ctx context.Context, req traceql.FetchSpansRequest) (traceql.FetchSpansResponse, error) {
-		if p.headBlock == nil {
-			return traceql.FetchSpansResponse{}, fmt.Errorf("head block is nil")
+	// Blocks to check
+	blocks := make([]common.BackendBlock, 0, 1+len(p.walBlocks)+len(p.completeBlocks))
```
mapno, line 268: This is going to create a lot of memory pressure with each read by copying memory around. We could loop through the blocks directly instead.

modules/generator/processor/localblocks/processor.go
```diff
@@ -25,13 +26,15 @@ import (
...
-	closeCh chan struct{}
-	wg      sync.WaitGroup
-
-	blocksMtx      sync.Mutex
+	tenant   string
+	Cfg      Config
+	wal      *wal.WAL
+	closeCh  chan struct{}
+	wg       sync.WaitGroup
+	cacheMtx sync.Mutex
```
mapno, line 34: It can be `sync.RWMutex`

mapno (approved):
Overall LGTM

modules/generator/processor/localblocks/processor.go
zalegrala: Good point.  I wonder if we could keep this `blocks` object on the struct, since we're already locking, we could keep it updated on the iteration but not need to re-allocate new memory.

zalegrala (approved):
Looks good to me.

modules/generator/processor/localblocks/processor.go
mdisibio: I took this approach because it greatly simplifies the below logic and error handling, one loop instead of two or three (head block, wal blocks, complete blocks).  The overhead of this slice should be minimal - expect a few dozen blocks and single digit requests per second to this api.

PR #2482, opened 2023-05-17
Introduce `overrides.Interface` to decouple implementation from usage

kvrhdn:
**What this PR does**:

A lot of modules in Tempo rely on the overrides package for runtime config, they directly use the `overrides.Overrides` struct. This makes the coupling very tight and limits modularity.

Changes:
- extract `overrides.Interface` from the methods of `overrides.Overrides`
- use this interface in as much places as possible
- export the Overrides module from `App`

~~Fixes #<issue number>~~

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/forwarder/manager.go
+++ b/modules/distributor/forwarder/manager.go
@@ -13,6 +13,7 @@ import (
 	"go.uber.org/multierr"
 
 	"github.com/grafana/tempo/modules/distributor/queue"
+	"github.com/grafana/tempo/modules/overrides"
 )
 
 const (
@@ -21,9 +22,10 @@ const (
 )
 
 type Overrides interface {
-	TenantIDs() []string
 	Forwarders(tenantID string) []string
 }
+
+var _ Overrides = (overrides.Interface)(nil)
 
 type Manager struct {
 	services.Service
--- a/modules/overrides/interface.go
+++ b/modules/overrides/interface.go
@@ -1,25 +1,49 @@
-package generator
+package overrides
 
 import (
+	"io"
+	"net/http"
 	"time"
 
-	"github.com/grafana/tempo/modules/generator/registry"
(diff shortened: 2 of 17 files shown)


modules/distributor/forwarder/manager.go
```diff
@@ -21,10 +22,11 @@ const (
 )
 
 type Overrides interface {
-	TenantIDs() []string
```
kvrhdn, line 24: This method wasn't used, I've removed it and added a type assertion to avoid drift.

modules/overrides/interface.go
```diff
@@ -0,0 +1,57 @@
...
+
+type Service interface {
+	services.Service
+	prometheus.Collector
+	Interface
+
+	WriteStatusRuntimeConfig(w io.Writer, r *http.Request) error
+}
+
+type Interface interface {
```
joe-elliott, line 23: this is a really intense interface which will make implementation/mocking difficult. is there anything we can do to simplify?

maybe:
```
type Interface interface {
  Limit(type, userID) 
}
```
That doesn't seem to work b/c each limit returns a different type and returning empty interface is kind of gross.

Is there a better point of abstraction? maybe something inside the overrides module?

kvrhdn: Interesting, I think extending/reusing code of the runtimeconfig.Manager to work with other sources would be neat. The polling logic will be needed for every source I think.

Besides fetching the limits we also need logic to merge the overrides of multiple sources and I think that will still end up somewhere in Tempo and not in dskit since dskit isn't aware of our limits. Adding a first layer of abstraction (the `overrides.Interface`) will make it easier to put a component in between that can deal with the merging logic.

joe-elliott: understood :+1:

PR #2487, opened 2023-05-18
Fill parent ID column and nested set columns

stoewer:
**What this PR does**:

Fill the new columns that were introduced with `vParquet2` with the correct values.

Fixes grafana/tempo-squad#207

**Checklist**
- [x] Tests updated
- [ ] ~Documentation added~


stoewer:
Benchmark results without and with the calculation for the nested set model:
```
goos: linux
goarch: amd64
pkg: github.com/grafana/tempo/tempodb/encoding/vparquet2
cpu: AMD Ryzen 7 PRO 6850U with Radeon Graphics     
                                  │ bench-nested-set-01.txt │       bench-nested-set-02.txt        │
                                  │         sec/op          │    sec/op      vs base               │
ProtoToParquet/SpanCount:10_k-16                8.746m ± 4%   10.763m ± 12%  +23.06% (p=0.002 n=6)
ProtoToParquet/SpanCount:100_k-16               58.16m ± 3%    69.99m ±  1%  +20.33% (p=0.002 n=6)
ProtoToParquet/SpanCount:1_M-16                 607.7m ± 2%    726.3m ±  2%  +19.52% (p=0.002 n=6)
geomean                                         67.61m         81.79m        +20.96%

                                  │ bench-nested-set-01.txt │      bench-nested-set-02.txt       │
                                  │          B/op           │     B/op      vs base              │
ProtoToParquet/SpanCount:10_k-16               8.338Mi ± 1%   8.641Mi ± 1%  +3.64% (p=0.002 n=6)
ProtoToParquet/SpanCount:100_k-16              82.80Mi ± 0%   86.69Mi ± 0%  +4.70% (p=0.002 n=6)
ProtoToParquet/SpanCount:1_M-16                827.4Mi ± 0%   867.1Mi ± 0%  +4.80% (p=0.002 n=6)
geomean                                        82.97Mi        86.60Mi       +4.38%

                                  │ bench-nested-set-01.txt │      bench-nested-set-02.txt      │
                                  │        allocs/op        │  allocs/op   vs base              │
ProtoToParquet/SpanCount:10_k-16                25.71k ± 1%   25.65k ± 1%       ~ (p=0.974 n=6)
ProtoToParquet/SpanCount:100_k-16               253.6k ± 0%   253.8k ± 0%       ~ (p=0.937 n=6)
ProtoToParquet/SpanCount:1_M-16                 2.535M ± 0%   2.533M ± 0%  -0.09% (p=0.041 n=6)
geomean                                         254.7k        254.5k       -0.08%
```

tempodb/encoding/vparquet2/combiner.go
```diff
@@ -160,6 +160,7 @@ func (c *Combiner) Result() (*Trace, int) {
 	if c.result != nil && c.combined {
 		// Only if anything combined
 		SortTrace(c.result)
```
joe-elliott, line 162: idle thought: since we are already sorting the trace can we dedupe batches here to reduce parquet size and improve performance on resource level attributes? wdyt?

tempodb/encoding/vparquet2/nested_set_model.go
```diff
@@ -0,0 +1,105 @@
...
+		assignmentNeeded bool
+		rootSpans        []*wrappedSpan
+		spanChildren     = map[[8]byte][]*Span{}
+	)
+
+	// find root spans and map span IDs to child spans
+	for _, rs := range trace.ResourceSpans {
+		for _, ss := range rs.ScopeSpans {
+			for i, s := range ss.Spans {
+				if s.NestedSetLeft == 0 || s.NestedSetRight == 0 {
```
joe-elliott, line 16: this seems to be checking for a case where we have combined two traces and some spans already have NestedSetLeft/Right set. can we trust the values that exist tho? this seems to trust existing NestedSetLeft and NestedSetRight values but if we combine two existing traces don't we have to completely recalculate all left/right/parent ids?

tempodb/encoding/vparquet2/nested_set_model.go
```diff
@@ -0,0 +1,105 @@
...
+					assignmentNeeded = true
+				}
+
+				if len(s.ParentSpanID) == 0 {
```
joe-elliott, line 20: nice work handling multiple roots

tempodb/encoding/vparquet2/nested_set_model.go
```diff
@@ -0,0 +1,105 @@
...
+					rootSpans = append(rootSpans, &wrappedSpan{span: &ss.Spans[i], id: util.SpanIDToArray(s.SpanID)})
+				} else {
+					parentID := util.SpanIDToArray(s.ParentSpanID)
+					spanChildren[parentID] = append(spanChildren[parentID], &ss.Spans[i])
```
joe-elliott, line 24: i wonder if we can do any preallocating or slice reuse here. that may be pre-optimizing and not worth it.

tempodb/encoding/vparquet2/nested_set_model.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+}
+
+type wrappedSpan struct {
+	span      *Span
+	id        [8]byte
+	nextChild int
+}
+
+type stack[T any] []*T
```
joe-elliott, line 76: wondering if this should be in `/pkg/stack` ? i'm also a little surprised there's no generic go stack in the runtime.

joe-elliott (commented):
this is some excellent work. after we merge this i will need to restrain myself from immediately writing `{} >> {}` :)

i have a fair number of questions and i'd like to give @mdisibio a chance to review (if he wants).

overall this looks great. i don't see any major issues.

tempodb/encoding/vparquet2/nested_set_model.go
stoewer: You are right. Combined traces can have incorrect bounds while all spans have non-zero left and right values. I added a parameter `forceAssignment`. When the function is called in the combiner, the assignment is now enforced.

tempodb/encoding/vparquet2/nested_set_model.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	// traverse the tree
+	var (
+		ancestors      stack[wrappedSpan]
+		nestedSetBound int32 = 1
+	)
+
+	for _, root := range rootSpans {
+		root.span.NestedSetLeft = nestedSetBound
```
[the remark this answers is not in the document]
stoewer: I added a comment and a method `IsRoot()` that uses the correct parent ID

tempodb/encoding/vparquet2/combiner.go
stoewer: I think that sounds like something that's worth to look into. But I guess that's probably better handled in a separate issue. Should I create one?

tempodb/encoding/vparquet2/nested_set_model.go
stoewer: I think this would be possible. I'm not so sure about the performance benefits though: It would require another iteration through all spans with a map lookup for every span. In the benchmark above number of allocations did not increase much when calling `assignNestedSetModelBounds`

tempodb/encoding/vparquet2/nested_set_model.go
stoewer: > i'm also a little surprised there's no generic go stack in the runtime

There is nothing in the standard library. There are generic collection packages emerging, but AFAIK there isn't really a standard package yet.

> wondering if this should be in /pkg/stack ?

I think it's a good idea to make it available outside of  the `vparquet2` package. An own package is a bit much though. I moved it to `pkg/util`.

tempodb/encoding/vparquet2/nested_set_model.go
joe-elliott: Ignore for now. I'd say this is not worth it unless it starts showing up on profiles.

tempodb/encoding/vparquet2/nested_set_model.go
joe-elliott: i've been trying to kill the util package in favor of a lot of smaller packages :), but if your instincts want it there then i'm good.

joe-elliott (approved):

tempodb/encoding/vparquet2/nested_set_model.go
joe-elliott: we're forcing reassignment in the combiner and in traceToParquet we're guaranteed that all NestedSet's will be 0. do we even need this check anymore?

joe-elliott (commented):
two questions, but this is looking good. maybe a quick sync discussion about the changes will be helpful to get this merged.

i am sad about the additional complexity to handle zipkin and (if the work weren't already complete) would be tempted to just say that hierarchical operators are undefined for zipkin traces.

great work. in particular i'm impressed with the clever optimizations in `(span id, kind) -> hash`

tempodb/encoding/vparquet2/nested_set_model.go
```diff
@@ -0,0 +1,154 @@
...
+					}
+				} else {
+					nodesByID[id] = []*spanNode{node}
+				}
+			}
+		}
+	}
+
+	// check preconditions before assignment
+	if undoAssignment {
```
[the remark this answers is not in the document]
stoewer: This is for the case where more than two spans have the same ID and it is no longer possible to make a reasonable assignment. In case there are nested set bounds assigned, it is necessary to undo them.

tempodb/encoding/vparquet2/nested_set_model.go
stoewer: You are right. I removed the check and the `forceAssignment` parameter

PR #2494, opened 2023-05-22
TraceQL: Selection

### LINKED ISSUE #2175: [TraceQL] Add Support for Selecting Fields in TraceQL

The current Search API endpoint returns 5 useful fields for each trace: `traceID`, `rootServiceName`, `rootTraceName`, `startTimeUnixNano`, and `durationMs`. When doing analysis on large amounts of traces, it can be very useful to get an additional field or two per trace, instead of parsing it out of the full trace via the Query Trace API. For example, one of my current use cases involves logging both the `rootServiceName` and the `span.http.target` fields. 

To modify the Search API endpoint to allow extra fields to be returned alongside the 5 fields it returns right now. This could be configured via a `select` statement in TraceQL or via a tag in tag-based search.

I can get the Tempo UI to show this information by including a TraceQL clause like `span.http.target != ""` to see the field value for `span.http.target` when I drill into a trace, but this does not work via API. 
**Additional context**

joe-elliott:
**What this PR does**:
Adds the ability to select fields to the language:

image

Fixes #2175 

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/traceql/lexer.go
```diff
@@ -245,7 +247,7 @@ func isAttributeRune(r rune) bool {
 	}
 
 	switch r {
-	case scanner.EOF, '{', '}', '(', ')', '=', '~', '!', '<', '>', '&', '|', '^':
+	case scanner.EOF, '{', '}', '(', ')', '=', '~', '!', '<', '>', '&', '|', '^', ',':
```
mdisibio, line 250: Highlighting that this means we can't use attribute names containing a comma after this change.  I think that is ok, we will figure out escaping at some point.

mdisibio (approved):
I love the fact this PR only touches the engine-layer.  Means the storage-layer flexibility is paying off.   LGTM 👍

PR #2497, opened 2023-05-23
Update "use-environment-variables-in-the-configuration" section

tavsec:
Update instructions on enabling env variable parsing in config file

**What this PR does**:
Updated the `-config.expand-env` option instructions, which now include the full option with value (`-config.expand-env=true`), instead of only specifying the option name. This will also unify the instructions on enabling env. variables in config with [Loki configuration](https://grafana.com/docs/loki/latest/configuration/), where the full option with value is specified.

- 

**Checklist**
- [ ] Tests updated (nothing to update)
- [x] Documentation added


knylander-grafana (approved):
Changes look good. Thank you for updating the doc.

joe-elliott (approved):
Thanks for the improvement!

PR #2500, opened 2023-05-23
Possibly reduce cost of tempodb tests

joe-elliott:
Previously we created 4 vParquet and 4 vParquet2 block. one for each of the different search tests and then ran the tests.

This PR flips things around and creates 1 vParquet and vParquet2 block and runs all 4 tests on each.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,7 @@ test-with-cover-pkg: tools
 # tests in tempodb (excluding tempodb/wal)
 .PHONY: test-with-cover-tempodb
 test-with-cover-tempodb: tools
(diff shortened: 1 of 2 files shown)


joe-elliott:
oof. still failing with 143. i do think this should obviously reduce mem/cpu costs of these tests, but maybe they're not as impactful as i thought?

joe-elliott:
gomemlimit seems to have helped.

electron0zero (approved):
lgtm.

btw, #2502 got merged. can rebease to see if that helps with mem usage issue.

PR #2502, opened 2023-05-24
Run TestCountSpans with predictable batch and span size

stoewer:
**What this PR does**:

`TestCountSpans` randomly selects a batch and span size to run the tests with. When high numbers are chosen for both values, the tests will allocate > 10 GB memory. This PR makes the 
batch and span size more predictable in order to avoid excess memory allocation.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/encoding/vparquet/compactor_test.go
+++ b/tempodb/encoding/vparquet/compactor_test.go
@@ -147,8 +147,8 @@ func TestValueAlloc(t *testing.T) {
 }
 
 func TestCountSpans(t *testing.T) {
-	batchSize := rand.Intn(1000) + 1
-	spansEach := rand.Intn(1000) + 1
+	batchSize := 300 + rand.Intn(25)
+	spansEach := 250 + rand.Intn(25)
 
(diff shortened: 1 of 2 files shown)


tempodb/encoding/vparquet/compactor_test.go
```diff
@@ -147,8 +147,8 @@ func TestValueAlloc(t *testing.T) {
 }
 
 func TestCountSpans(t *testing.T) {
-	batchSize := rand.Intn(1000) + 1
-	spansEach := rand.Intn(1000) + 1
+	batchSize := 300 + rand.Intn(25)
```
electron0zero, line 150: nit: let's leave a comment about high mem usage when these values are set to big numbers.

mapno (approved):

tempodb/encoding/vparquet/compactor_test.go
electron0zero: same for `tempodb/encoding/vparquet2/compactor_test.go`

electron0zero (approved):

PR #2510, opened 2023-05-26
add config to drop certain labels from target info

ie-pham:
**What this PR does**: Allows user to specify labels they want to drop from `traces_target_info`

Example:
```
traces_target_info{label_a="abc", label_b="xyz"}
```

If we set 
```
target_info_excluded_dimensions: ['label_b']
```

We will get
```
traces_target_info{label_a="abc"}
```

There is no filtering involved. The label is simply dropped. 

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [x] Documentation added


zalegrala:
Interesting.  If we drop labels, do we know what happens when there are duplicate metrics?  Presumably if two metrics were unique, but after dropping the labels, we would have two metrics merged into one I suspect.

zalegrala (approved):
This looks good to me.  It might be worth testing this out against some use cases we've heard about, but I expect this to work as intended.

knylander-grafana (commented):
Thank you for adding docs.

PR #2515, opened 2023-05-30
FIX-699: Flush all the traces during stopping stage

### LINKED ISSUE #699: [Ingester] Add ability to completely flush on shutdown
When an ingester shuts down it is expected it will come back up, replay its wal and continue using the same disk. However, in some ephemeral deployments this is not true and it is desired for an ingester to always flush its entire contents to the backend on shutdown.

Add the ability for an ingester to always flush its entire disk to the backend.

Note: With this setting there may be gaps in trace availability. After the ingester shuts down but before the queriers get a chance to poll the new blocks those traces would not be available. This should be noted on the configuration option, but does not need to be solved.

mghildiy:
[Ingester] Add ability to completely flush on shutdown(WIP)

Fixes #699

**Checklist**
- [ ] Tests updated
- [x] Documentation added


zalegrala:
This looks like a good start to me.  Thanks for putting this up.  Our CI has been struggling a little, so I kicked the failed jobs again.

joe-elliott:
So the concern with this PR is that there is a chance that flushqueues will be empty even if there is still work to do as detailed here:

https://github.com/grafana/tempo/issues/699#issuecomment-1564495737

I see there are more comments in that issue about a potential approach. I will review when I get a chance.

modules/ingester/ingester.go
```diff
@@ -148,6 +148,14 @@ func (i *Ingester) loop(ctx context.Context) error {
 func (i *Ingester) stopping(_ error) error {
 	i.markUnavailable()
 
+	if i.cfg.FlushAllOnShutdown {
+		i.lifecycler.SetUnregisterOnShutdown(true)
```
joe-elliott, line 152: i don't think we should set this. allow this value to be configured independently of `flush_all_on_shutdown`.

joe-elliott (commented):
I finally had time to read your comment here: https://github.com/grafana/tempo/issues/699#issuecomment-1566065223

After digging through the code I agree with your analysis. Nice work deep diving this. Let's do some cleanup and then we're good to merge this.

- Since the added lines are the same as what's in the ShutdownHandler let's move these out to a new function and use it in both places. Please comment why it works b/c it's a little unintuitive.
  - Adding a test for this new method would be nice. Not sure how feasible that is.
- Add a changelog entry
- Document the new field in the ingester docs here: https://github.com/grafana/tempo/blob/main/docs/sources/tempo/configuration/_index.md#ingester

modules/ingester/ingester.go
mghildiy: But i see it being done in flush.ShutDownHandler. Shouldn't we have same behaviour for stopping too?

mghildiy: So now ShutDownHandler too would flush only if this new flag is true?

joe-elliott: No. Shutdown should continue to do exactly what it does. It should not require a flag to set set.

joe-elliott:
I'm unsure what caused the large changes. They should definitely not be present.

> make: gotestsum: No such file or directory

We swapped to gotestsum recently for better test output in CI. It should be automatically installed if you run `make tools`

knylander-grafana (commented):
Thank you for adding doc.

mghildiy:
> Thank you for adding doc.

[Better PR](https://github.com/grafana/tempo/pull/2538) is here.

joe-elliott:
Closing in favor of: https://github.com/grafana/tempo/pull/2538

PR #2517, opened 2023-05-31
vParquet3: add dedicated columns to parquet schema and block meta

### LINKED ISSUE #2528: vParquet3: Add dedicated columns to schema and block meta
Add spare columns to the parquet schema and extend the block metadata with column assignments. Change the encoding such that attributes are written to and read from dedicated spare columns if a column assignment is specified in the block metadata.

Is part of #2527

stoewer:
**What this PR does**:

This PR is a first step towards dynamically assignable dedicated columns for resource and span attributes. Attributes written to and read from spare dedicated columns depending on a column assignment in the block metadata.

Fixes #2528

**Checklist**
- [x] Tests updated
- [ ] ~Documentation added~
- [ ] ~`CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`~


tempodb/encoding/vparquet2/dedicated_columns.go
```diff
@@ -0,0 +1,193 @@
...
+	dedicatedColumnTypeString    = "string"
+	dedicatedColumnScopeResource = "resource"
+	dedicatedColumnScopeSpan     = "span"
+)
+
+var (
+	// Column paths for spare dedicated attribute columns
+	dedicatedResourceColumnsByType = map[string][]string{
+		dedicatedColumnTypeString: {
+			"rs.list.element.Resource.DedicatedAttributes.String01",
```
stoewer, line 18: The column paths are actually not needed yet, but will be needed later for the querying part. For now only the number of entries is relevant.

tempodb/backend/block_meta_test.go
```diff
@@ -95,9 +95,41 @@ func TestBlockMetaObjectAdded(t *testing.T) {
...
+    "encoding": "zstd",
+    "indexPageSize": 250000,
+    "totalRecords": 124356,
+    "dataEncoding": "",
+    "bloomShards": 244,
+	"dedicatedColumns": [
+		{"scope": "resource", "name": "namespace", "type": "string"},
+		{"scope": "span", "name": "http.method", "type": "string"},
+		{"scope": "span", "name": "namespace", "type": "string"}
+	]
```
mapno, line 125:
```suggestion
    "dedicatedColumns": [
    	{"scope": "resource", "name": "namespace", "type": "string"},
    	{"scope": "span", "name": "http.method", "type": "string"},
    	{"scope": "span", "name": "namespace", "type": "string"}
    ]
```

mapno (approved):
LGTM

tempodb/encoding/vparquet2/dedicated_columns.go
```diff
@@ -0,0 +1,193 @@
...
+
+import (
+	v1 "github.com/grafana/tempo/pkg/tempopb/common/v1"
+	"github.com/grafana/tempo/tempodb/backend"
+)
+
+const (
+	dedicatedColumnTypeString    = "string"
+	dedicatedColumnScopeResource = "resource"
+	dedicatedColumnScopeSpan     = "span"
```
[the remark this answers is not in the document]
stoewer: I agree that can be beneficial to make types for the column scope and type. However, similar fields / types will exist in different places: config/overrides, block meta, and encoding. Adding types makes more sense if the definitions are shared across those parts. I’m not quite sure if we want to do this in the end. I’d rather wait until we have the other parts and then see if sharing types makes sense.

PR #2535, opened 2023-06-02
Make synchronous parquet iterator default

mdisibio:
**What this PR does**:
The new synchronous iterator added in #2165 has been running internally for awhile and going well, so this makes it the default.  Environment variables were swapped around so that we can revert back if needed.

I was also very lax about updating the changelog lately, and adds some missing entries. 

Fixes n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott (approved):

joe-elliott:
let's go synchronously!

PR #2536, opened 2023-06-02
Attempt to fix `"1" is not greater than or equal to "2"`

joe-elliott:
**What this PR does**:
Attempts to fix the infamous `"1" is not greater than or equal to "2"` error in CI. I was unable to track down exactly what the issue (or reproduce it locally), but I believe it's b/c the `appender.Commit()` is non atomic. In this case if the final flush happened to initiate in between writing the series and writing the exemplar then it would only receive 1 instead of the expected 2.

### DIFF
--- a/modules/generator/storage/instance_test.go
+++ b/modules/generator/storage/instance_test.go
@@ -26,9 +26,8 @@ import (
 	"go.uber.org/atomic"
 )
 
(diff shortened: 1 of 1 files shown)


joe-elliott:
alas, this has proven to be beyond me

PR #2540, opened 2023-06-05
Continue polling remaining tenants on error

mdisibio:
**What this PR does**:
Currently, if flakey errors occur during polling, the entire loop exits early and doesn't restart until the next timer. This can effectively starve some tenants and cause their indexes to fall out of date both due to failure to be written and failure to be read.  This PR changes the high-level error handling to be a circuit-breaker style where it will keep going until a configurable amount of consecutive errors happen.  The idea is that flakey errors will not halt polling, but consistent errors due to misconfiguration, auth, or networking will, so that we don't keep hammering the backend.   By default it's set to tolerate 1 error only, meaning it will fail if there are 2 consecutive errors.  It can also be set to 0 to revert to the prior behavior where every error is honored.

Marked as draft because I want to get feedback on this approach or see if we want to do a different approach.  Because polling iterates tenants in order, tenants that are alphanumerically last are impacted the most.  An alternative fix would be to iterate tenants randomly or keep order across loops like compactors do. But I think the approach in this PR is more robust.

Fixes n/a

**Checklist**
- [x] Tests updated
- [x] Documentation added


mapno (approved):
Good change 👍

zalegrala:
:ship:

PR #2542, opened 2023-06-06
[tempo-mixin] add new alert to identify block list growth

zalegrala:
**What this PR does**:

Here we include a new mixin alert to notify when the block list growth is over 40% over the last 7 days.  This may indicates an under-scaled compactor deployment.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


ie-pham (approved):
🙏

PR #2543, opened 2023-06-06
Add user-configurable overrides module

kvrhdn:
**What this PR does**:
Introduces a new optional feature: user-configurable overrides. Goal is to allow tenants to change a subset of overrides using an API. Changes will be limited to `forwarders` and later several `metrics_generator_...` overrides.

#### Architecture

For every tenant we will store an `overrides.json` file in a bucket. These files will be stored in a separate bucket from the backend.

```
Overrides bucket
├── 1
│   └── overrides.json
└── 2
    └── overrides.json
```

This bucket is regularly polled and a copy of the limits is kept in-memory. When requesting the overrides for a tenant, the overrides module will now:
1. check this override is set in the user-configurable overrides, if so return that value 
2. check if this override is set in the runtime config (configmap), if so return that value
3. return the default values

#### API

```
GET /api/overrides
{
  forwarders: ["k6-cloud-insights-staging"],
}
```

```
PUT /api/overrides
{
  forwarders: ["k6-cloud-insights-staging"],
}
```

```
DELETE /api/overrides
```

Later we might consider adding a PATCH endpoint to perform partial updates/deletes.

~~Fixes #<issue number>~~ New feature!

**Checklist**
- [x] Tests updated
- [ ] Documentation added https://github.com/grafana/tempo/issues/2627


modules/overrides/user_config_overrides.go
```diff
@@ -281,23 +294,24 @@ func (o *userConfigOverridesManager) DeleteLimits(ctx context.Context, userID st
...
-func (o *userConfigOverridesManager) getTenantLimitsAll() map[string]*UserConfigurableLimits {
+func (o *userConfigOverridesManager) getAllCachedTenantLimits() map[string]*UserConfigurableLimits {
 	o.mtx.RLock()
 	defer o.mtx.RUnlock()
 
 	return o.tenantLimits
 }
 
 func (o *userConfigOverridesManager) Forwarders(userID string) []string {
-	tenantLimits, ok := o.getTenantLimits(userID)
```
electron0zero: this will only return configured overrides when configured overrides are set, and runtime overrides when configured overrides are not set.

I am leaning towards keeping it as is. do we want this behaviour and want to merge both Forwarders?

knylander-grafana:
@kvrhdn Will we need doc for this? Should I create a doc issue for it or is it something you'd add to this PR?

kvrhdn:
> @kvrhdn Will we need doc for this? Should I create a doc issue for it or is it something you'd add to this PR?

Yes! This is adding a new feature with an API endpoint and it's own set of configurations. Feel free to create a doc issue and assign it to me, I can get a PR started and we can iterate from there 🙂

modules/overrides/http.go
```diff
@@ -0,0 +1,115 @@
...
+
+	"github.com/grafana/tempo/pkg/api"
+	"github.com/grafana/tempo/pkg/util/log"
+)
+
+// OverridesHandler is a http.HandlerFunc to return user configured overrides
+func (o *userConfigOverridesManager) OverridesHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	span, ctx := opentracing.StartSpanFromContext(ctx, "userConfigOverridesManager.OverridesHandler")
+	defer span.Finish()
```
zalegrala, line 21: Do you want to set the span error if there is one?

modules/overrides/user_config_overrides.go
```diff
@@ -0,0 +1,384 @@
...
+
+func (o *userConfigOverridesManager) starting(ctx context.Context) error {
+	if err := services.StartManagerAndAwaitHealthy(ctx, o.subservices); err != nil {
+		return errors.Wrap(err, "unable to start overrides subservices")
+	}
+
+	return o.reloadAllTenantLimits(ctx)
+}
+
+func (o *userConfigOverridesManager) loop(ctx context.Context) error {
```
zalegrala, line 159: Consider renaming to `running` for convention.  Not a blocker.

zalegrala (commented):
This looks pretty nice and clean to me.  :+1: Happy to approve, but will give others a chance to review also.

integration/e2e/overrides_client.go
```diff
@@ -0,0 +1,105 @@
...
+	tempoUtil "github.com/grafana/tempo/pkg/util"
+)
+
+// Client wraps tempoUtil.Client and adds functions to work with the user-configurable overrides API.
+// We can't add this functions to tempoUtil directly because this would result in an import cycle.
+type Client struct {
+	*tempoUtil.Client
+}
+
+func (c *Client) GetOverrides() (*overrides.UserConfigurableLimits, error) {
```
joe-elliott, line 21: current API client is in `/pkg/util/client.go`. i'm thinking we should move both of these to a new package. Something like `/pkg/util/httpclient`.

modules/overrides/http.go
```diff
@@ -0,0 +1,115 @@
...
+
+	userID, err := user.ExtractOrgID(ctx)
+	if err != nil {
+		handleError(r, w, userID, errors.Wrap(err, "failed to find org id in request"))
+		return
+	}
+	level.Info(log.Logger).Log("tenant", userID, "method", r.Method, "url", r.URL.RequestURI())
+
+	switch r.Method {
```
joe-elliott, line 30: we should use the mux to handle this routing

modules/overrides/user_config_overrides.go
```diff
@@ -0,0 +1,384 @@
...
+
+	return nil
+}
+
+func (o *userConfigOverridesManager) listTenantsUnderLock(ctx context.Context) ([]string, error) {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "overrides.listTenantsUnderLock")
+	defer span.Finish()
+
+	// TODO to avoid polling the entire bucket, use a tenant list and keep it in a shared cache
+	tenants, err := o.r.List(ctx, []string{overridesKeyPath})
```
joe-elliott, line 227: this is threadsafe and doe not require locking

modules/overrides/user_config_overrides.go
```diff
@@ -0,0 +1,384 @@
...
+	return tenants, nil
+}
+
+// getTenantLimits will look up the UserConfigurableLimits for a tenant, it performs a backend request
+// and will update the entry in the local cache.
+func (o *userConfigOverridesManager) getTenantLimits(ctx context.Context, userID string) (tenantLimits *UserConfigurableLimits, err error) {
+	o.mtx.Lock()
+	defer o.mtx.Unlock()
+
+	return o.getTenantLimitsUnderLock(ctx, userID)
```
joe-elliott, line 240: i find it a little strange (but not wrong) that calling GET to get the limits also updates the local, in memory copy.

i feel like the API should be separate from the mechanism that maintains limits internally. so an api that's just a passthrough to object storage and a separate module that polls tenant limits and provides them as part of the overrides chain.

wdyt? is this separation of concerns worth it?

modules/overrides/util.go
```diff
@@ -0,0 +1,10 @@
+package overrides
+
+func doesNotContain(list []string, item string) bool {
```
joe-elliott, line 3: just not the output of this? https://pkg.go.dev/golang.org/x/exp/slices#Contains

tempodb/backend/cache/cache.go
```diff
@@ -110,6 +110,11 @@ func (r *readerWriter) CloseAppend(ctx context.Context, tracker backend.AppendTr
 	return r.nextWriter.CloseAppend(ctx, tracker)
 }
 
+func (r *readerWriter) Delete(ctx context.Context, name string, keypath backend.KeyPath) error {
+	// TODO also clear from cache
```
joe-elliott, line 114: we should do this before merging. otherwise it's a difficult to find bug in the future

modules/overrides/user_config_limits.go
```diff
@@ -0,0 +1,15 @@
+package overrides
+
+type UserConfigurableLimits struct {
+	// TODO how do we ensure Version is always filled in?
+	Version string `json:"version" yaml:"version"`
```
joe-elliott, line 5: if this is an api version we should keep it in the path. that's what we do on other endpoints

knylander-grafana:
> > @kvrhdn Will we need doc for this? Should I create a doc issue for it or is it something you'd add to this PR?
> 
> Yes! This is adding a new feature with an API endpoint and it's own set of configurations. Feel free to create a doc issue and assign it to me, I can get a PR started and we can iterate from there 🙂

Thanks, @kvrhdn! The issue is here: https://github.com/grafana/tempo/pull/2543

modules/overrides/http.go
kvrhdn: I have addressed this in my last commit by setting the error in `handleError`

modules/overrides/http.go
kvrhdn: Good point, I've split up the handler and register them separately now in modules.go. It's a bit more verbose but not more code overall I think.

modules/overrides/util.go
kvrhdn: Woah, today I learned

tempodb/backend/cache/cache.go
kvrhdn: This comment is misleading, it's intended to say "once we implement caching, we will have to delete stuff from the cache here". We don't have caching yet so I'm just going to delete this comment.

modules/overrides/user_config_overrides.go
kvrhdn: I was thinking of a scenario in which someone sends a POST or DELETE request right after we called list. In this case the list we just fetched would be outdated. I think I should take another look at this and simplify it, we really don't need a constant lock for the whole thing...

modules/overrides/user_config_overrides.go
kvrhdn: I've linked them with the idea "we have spent work to fetch the data anyway, let's immediately update our local copy". Advantage is you have a quicker update in the instance handling the request. In the context of a distributed system that isn't really a big deal though since all the other components will have to fetch it too...

I'll give separating the two a try and see if it cleans up the code a bit 🙂

integration/e2e/overrides_client.go
joe-elliott: edited the suggested path to `/pkg/httpclient`. overall i'm trying to move things out of util

modules/overrides/user_config_overrides.go
joe-elliott: These kinds of state based race conditions are unavoidable. We should just push them to object storage as much as possible.

tempodb/backend/cache/cache.go
joe-elliott: > To allow merging this PR maybe I can panic instead?

I suppose there's no code path to this b/c you don't implement caching on the overrides api? i'm fine with a panic. heads up @zalegrala

zalegrala: Maybe log a message and return nil?  I suppose panic might be warranted if we really never intent to add support.

modules/overrides/user_config_overrides.go
kvrhdn: I've split up the manager, api and client structures. Agree each is simpler now and can focus on its own thing :+1:

So the manager and the API both use client to talk with the backend. All components will run the manager but only the query-frontend needs to run the API.

integration/e2e/overrides_client.go
kvrhdn: PR to move it: https://github.com/grafana/tempo/pull/2637
Once merged I can update this PR

tempodb/backend/cache/cache.go
kvrhdn: I have changed it to return a panic for now. This isn't as clean as returning nil/logging an error but it would ensure we don't accidentally use this in production. If someone isn't careful while testing they might never call delete through a cache or not check the logs of their app. Having the application crash should ensure it doesn't fly under the radar.

modules/overrides/user_config_limits.go
kvrhdn: I've opted to drop the version field for now. Usually when we do a breaking change we just introduce a `/v2/` endpoint. Dealing with versioning even before we know how our api could change might be a bit over eager...

modules/overrides/user_configurable_overrides.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+func (o *userConfigurableOverridesManager) deleteTenantLimit(userID string) {
+	o.mtx.Lock()
+	defer o.mtx.Unlock()
+
+	delete(o.tenantLimits, userID)
+}
+
+func (o *userConfigurableOverridesManager) Forwarders(userID string) []string {
```
joe-elliott, line 182: i can honestly say i've not yet seen this pattern in go. are you "overriding" the composed interface?

modules/overrides/user_configurable_overrides.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	mgr.subservices, err = services.NewManager(subOverrides)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create subservices: %w", err)
+	}
+	mgr.subservicesWatcher = services.NewFailureWatcher()
+	mgr.subservicesWatcher.WatchManager(mgr.subservices)
+
+	mgr.Service = services.NewBasicService(mgr.starting, mgr.running, mgr.stopping)
```
joe-elliott, line 82: there's a service.TimerService (i'm pretty sure)

modules/overrides/user_configurable_overrides.go
kvrhdn: It's called struct embedding. This page explains it way better than I ever could: https://eli.thegreenplace.net/2020/embedding-in-go-part-1-structs-in-structs/

By embedding `overrides.Interface` in `userConfigurableOverridesManager` our manager inherits the full implementation. And by overwriting specific methods we can inject our own overrides.
For now we only implemented `Forwarders`, the other methods are automatically handled by the embedded `Interface`.

It's definitely a complicated Go feature but we thought it would be worth it as it saves us a lot of boilerplate and potential of mistakes.

The alternative is to make `overrides.Interface` a regular field, but that would require us to implement each method of `Interface` over again (34 methods!). Every implementation would just look like this:
```go
func (o *userConfigurableOverridesManager) IngestionRateStrategy() string {
	return o.interface.IngestionRateStrategy()
}
```

modules/overrides/userconfigurableapi/client.go
```diff
@@ -0,0 +1,163 @@
...
+	defer span.Finish()
+
+	tenantDirectories, err := o.r.List(ctx, []string{OverridesKeyPath})
+	if err != nil {
+		return nil, err
+	}
+
+	var tenants []string
+	for _, tenant := range tenantDirectories {
+		_, _, err := o.r.Read(ctx, OverridesFileName, []string{OverridesKeyPath, tenant}, false)
```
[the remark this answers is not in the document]
kvrhdn: I've added this logic because the local backend would leave empty directories behind. So list returns the tenant even if the overrides file doesn't exist. I'll see if we can address this in the implementation of the local backend.

modules/overrides/userconfigurableapi/client.go
```diff
@@ -0,0 +1,163 @@
...
+		reader, writer, _, err = s3.New(cfg.S3)
+	case "azure":
+		reader, writer, _, err = azure.New(cfg.Azure)
+	default:
+		err = fmt.Errorf("unknown backend %s", cfg.Backend)
+	}
+	return
+}
+
+func (o *userConfigOverridesClient) List(ctx context.Context) ([]string, error) {
```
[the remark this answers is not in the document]
kvrhdn: Agree, I think we should log a warning if the exact same path is used. If overrides are stored next to the backend blocks the `overrides/` key will be interpreted as a tenant name and the tenants itself will be parsed as block uuids and fail.

In the documentation I will mention we recommend using 2 buckets but if you use 1 bucket you need to configure a prefix (https://github.com/grafana/tempo/issues/2627#issuecomment-1632729681). Prefixes are supported on all backends so it's a decent solution.

modules/overrides/userconfigurableapi/http.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// TODO validate the received data
+
+	err = a.client.Set(ctx, userID, limits)
+	if err != nil {
+		handleError(span, userID, r, w, http.StatusInternalServerError, errors.Wrap(err, "failed to store user-configurable limits"))
+	}
+
+	w.WriteHeader(http.StatusOK)
+	w.Header().Set(api.HeaderContentType, "text/plain; charset=utf-8")
```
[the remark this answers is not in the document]
kvrhdn: Looks like we usually only set status ok and return nothing on successful requests. I'll make this consistent with the other handlers.

modules/overrides/userconfigurableapi/http.go
```diff
@@ -0,0 +1,134 @@
...
+	defer span.Finish()
+
+	userID, err := user.ExtractOrgID(ctx)
+	if err != nil {
+		handleError(span, userID, r, w, http.StatusBadRequest, err)
+		return
+	}
+	logRequest(userID, r)
+
+	limits, err := a.client.Get(ctx, userID)
```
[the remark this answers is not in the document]
kvrhdn: Changed it so it returns 404 when no overrides are set

modules/overrides/userconfigurableapi/http.go
```diff
@@ -0,0 +1,134 @@
+package userconfigurableapi
```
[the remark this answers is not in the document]
kvrhdn: This file contains the http handlers for `/api/overrides`. `httpclient` is for making request to Tempo?
Do you mean `pkg/api`?

So far we have kept http handlers inside their respective module so it is consistent with the rest of Tempo (afaik). I think we can move out logic if we are doing more than just parsing the request and passing it on to the backend client.

zalegrala (approved):
This looks good to me.  A few TODOs which we discussed and can follow up in future PRs.

modules/overrides/userconfigurableapi/http.go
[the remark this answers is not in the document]
joe-elliott: so i consider the `/api/overrides` endpoints to be part of Tempo's external api surface which is why I think they should be part of `/pkg/httpclient`

joe-elliott (approved):
One comment on why I still think we should move the overrides client into `pkg/httpclient`, but I'll leave that up to you. Your call!

PR #2546, opened 2023-06-08
Fix target_info labels

ie-pham:
**What this PR does**: The `labels` variable in the gauge & counter `collectMetrics` function was created outside the for loop and so we were constantly adding labels to metrics after metrics. 

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/registry/gauge.go
+++ b/modules/generator/registry/gauge.go
@@ -2,6 +2,7 @@ package registry
 
 import (
 	"sync"
+	"fmt"
 	"time"
 
 	"github.com/prometheus/prometheus/model/labels"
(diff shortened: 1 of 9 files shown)


modules/generator/registry/gauge.go
```diff
@@ -118,31 +118,27 @@ func (g *gauge) collectMetrics(appender storage.Appender, timeMs int64, external
...
-	// set metric name
-	lb.Set(labels.MetricName, g.metricName)
-	// set external labels
-	for name, value := range externalLabels {
-		lb.Set(name, value)
-	}
-
 	for _, s := range g.series {
 		t := time.UnixMilli(timeMs)
+		lbls := make(labels.Labels, 1+len(externalLabels)+len(s.labels.names))
```
zalegrala, line 123: We might want to keep the allocations outside of the loop.  I think I saw a `Reset()` method on the lb object that might empty this out.

ie-pham: good call

calebamiles (approved):
🚀

mapno (commented):
I think it's missing a test to ensure the bug is fixed. Otherwise, LGTM

ie-pham:
added test and made sure they failed before the fix and passed after the fix

mapno (approved):
LGMT, nice fix

PR #2549, opened 2023-06-09
Refactor microservices jsonnet into reusable component

zalegrala:
**What this PR does**:

Here we introduce a new `component` jsonnet library that encapsulates much of the functionality for each tempo component with flexibility to be re-used between them.  Additionally, all the tempo components are adjusted to implement the component library, passing in necessary details from the existing global `$._config` while avoiding any global reference internally to the library.  To maintain backwards compatibility with existing field overrides, all existing fields are appended to the appropriate locations for the relevant component objects.  This will mean that existing implementations that wrap the microservices jsonnet will be able to continue to function and allow a slower transition to the new approach.

** BREAKING CHANGE **: tempo-query has been dropped from the jsonnet deployment of the query-frontend.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


operations/jsonnet/microservices/component.libsonnet
```diff
@@ -0,0 +1,87 @@
...
+    configVolumeName:: 'tempo-conf',
+    overridesVolumeName:: 'overrides',
+
+    containerArgs:: {
+      target: target,
+      'config.file': '/conf/tempo.yaml',
+      'mem-ballast-size-mbs': $._config.ballast_size_mbs,
+    },
+
+    workload: {},
```
zalegrala, line 26: Since we're not responsible for realizing the workoad, we don't need this key and can instead reference the `target.deployment` directly, or the `target.statefulset` directly.

operations/jsonnet/microservices/querier.libsonnet
```diff
@@ -7,49 +7,28 @@
...
-      config_hash: std.md5(std.toString($.tempo_querier_configmap.data['tempo.yaml'])),
-    }) +
-    deployment.mixin.spec.template.spec.withVolumes([
-      volume.fromConfigMap(tempo_config_volume, $.tempo_querier_configmap.metadata.name),
-      volume.fromConfigMap(tempo_overrides_config_volume, $._config.overrides_configmap_name),
-    ]),
-
-  tempo_querier_service:
-    k.util.serviceFor($.tempo_querier_deployment),
+  tempo_querier_container:: {},
```
joe-elliott, line 31: what are these for?

joe-elliott (commented):
i overall like these changes, but how does this impact the structure of our internal deployment jsonnet?

for instance is per env scaling.libsonnet still honored? does the way we've structured our overlays in main.jsonnet still apply correctly?

operations/jsonnet/microservices/querier.libsonnet
zalegrala: Its a way of creating the hidden object first, so that appends to the top level keys are capture and used above.  Previously, the `_deployment` and `_service` keys below were the public interface, but we move the public to be encapsulated under the key `querier` for this component.

zalegrala: Strictly to maintain backwards compatibility.

zalegrala:
That last commit covers the delta that I saw in the delta in one of our clusters.  Now what I see in that diff on that cluster is represented in the compiled jsonnet.

kvrhdn (commented):
This is really nice work! I like how you managed to simplify the creation of each component and removed so much jsonnet/k8s 'noise' from each file.
It's also clear you thought about how to migrate this.

There are a couple of changes in the generated yaml, but as far as I can tell these don't change any settings since the default is the same right?

Just a thought: could we warn people when they use the old variables? We still expose `tempo_compactor_container` for compatibility with how we used to override settings, but I think you are now supposed to tweak `compactor.container`.
Could we issue a warning if people still change `tempo_compactor_container` (or any other old variable)?

operations/jsonnet/microservices/frontend.libsonnet
```diff
@@ -1,93 +1,42 @@
...
-    ]),
-
-  tempo_query_frontend_deployment:
-    deployment.new(
-      target_name,
-      $._config.query_frontend.replicas,
-      std.prune([
-        $.tempo_query_frontend_container,
-        if $._config.tempo_query.enabled then $.tempo_query_container,
-      ]),
```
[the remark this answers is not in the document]
zalegrala: Good eye.  I added a commit to drop the tempo-query configuration and deployment, and updated the PR description.  Probably worth a changelog, which I'll also include after testing again.  The generator deployment->statefulset happened while this PR was opened and so want to check on things.

operations/jsonnet/microservices/configmap.libsonnet
```diff
@@ -121,42 +121,6 @@
...
-      'tempo.yaml': $.util.manifestYaml($.tempo_querier_config),
-    }),
-
-  tempo_query_frontend_configmap:
-    configMap.new('tempo-query-frontend') +
-    configMap.withData({
-      'tempo.yaml': $.util.manifestYaml($.tempo_query_frontend_config),
-    }),
-
   tempo_query_configmap: if $._config.tempo_query.enabled then
```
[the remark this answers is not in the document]
zalegrala: I've opted to drop this based on your other comment.

zalegrala:
Thanks for the review @kvrhdn.  The changes in the manifests are a result of moving away from the `kausal` utils, which have been deprecated for a long time.  All the changes seemed okay to me since as you mention, I think they are the default values.

Good question about warning users if they use an old variable.  I'm not entirely sure, but I'll do a little research and see if we can come up with something.

PR #2555, opened 2023-06-14
[vParquet3] Write path

### LINKED ISSUE #2552: Wiring the config with the schema changes and make write path work
Dynamic columns configured through overrides need to be propagated through block creation in the ingesters.

Part of https://github.com/grafana/tempo/issues/2527

mapno:
**What this PR does**:
Propagates dynamic columns configuration to the write path. Ingesters now are able to create and write to blocks with dedicated columns.

Fixes #2552

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/ingester/ingester_test.go
+++ b/modules/ingester/ingester_test.go
@@ -344,7 +344,7 @@ func TestFlush(t *testing.T) {
 func TestDedicatedColumns(t *testing.T) {
 	tmpDir, err := os.MkdirTemp("/tmp", "")
 	require.NoError(t, err, "unexpected error getting tempdir")
-	defer os.RemoveAll(tmpDir)
+	t.Cleanup(func() { _ = os.RemoveAll(tmpDir) })
 
 	limits := overrides.Limits{}
 	flagext.DefaultValues(&limits)
@@ -402,7 +402,7 @@ func TestDedicatedColumns(t *testing.T) {
 }
 
 func defaultIngesterModule(t testing.TB, tmpDir string) *Ingester {
-	return defaultIngesterWithOverrides(t, tmpDir, defaultLimitsTestConfig())
+	return defaultIngesterWithOverrides(t, tmpDir, defaultOverrides())
 }
 
 func defaultIngesterWithOverrides(t testing.TB, tmpDir string, o overrides.Limits) *Ingester {
@@ -496,7 +496,7 @@ func defaultIngesterTestConfig() Config {
 	return cfg
 }
 
-func defaultLimitsTestConfig() overrides.Limits {
+func defaultOverrides() overrides.Limits {
 	limits := overrides.Limits{}
 	flagext.DefaultValues(&limits)
 	return limits
--- a/modules/ingester/ingester_test.go
+++ b/modules/ingester/ingester_test.go
@@ -7,6 +7,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+
 	"github.com/go-kit/log"
 	"github.com/gogo/protobuf/proto"
 	"github.com/grafana/dskit/flagext"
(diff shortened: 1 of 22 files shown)


modules/ingester/ingester_test.go
```diff
@@ -339,9 +341,73 @@ func TestFlush(t *testing.T) {
...
+	// TODO: This check should be included as part of the read path
+	inst.blocksMtx.RLock()
+	for _, b := range inst.completeBlocks {
+		assert.Equal(t, limits.DedicatedColumns, b.BlockMeta().DedicatedColumns)
+	}
+	inst.blocksMtx.RUnlock()
+}
+
 func defaultIngesterModule(t testing.TB, tmpDir string) *Ingester {
+	return defaultIngesterWithOverrides(t, tmpDir, defaultLimitsTestConfig())
```
stoewer, line 405: `nitpick` This could be an opportunity to rename `defaultLimitsTestConfig` to `defaultOverridesTestConfig` or just `defaultOverrides`.

modules/ingester/instance.go
```diff
@@ -478,6 +486,13 @@ func (i *instance) resetHeadBlock() error {
 	return nil
 }
 
+func (i *instance) getDedicatedColumns() []backend.DedicatedColumn {
+	if cols := i.overrides.DedicatedColumns(i.instanceID); len(cols) > 0 {
```
stoewer, line 490: If it should be possible to unset the dedicated columns via overrides, it would be necessary to make a distinction here between `nil` and the empty slice `[]backend.DedicatedColumn{}`. In this case the condition should be `cols != nil`

tempodb/backend/block_meta.go
```diff
@@ -63,15 +65,16 @@ type DedicatedColumn struct {
 	Type string `yaml:"type" json:"type"`
 }
 
-func NewBlockMeta(tenantID string, blockID uuid.UUID, version string, encoding Encoding, dataEncoding string) *BlockMeta {
+func NewBlockMeta(tenantID string, blockID uuid.UUID, version string, encoding Encoding, dataEncoding string, dedicatedColumns ...DedicatedColumn) *BlockMeta {
```
stoewer, line 68: I'd prefer to use `dedicatedColumns []DedicatedColumn` instead of a variadic parameter here
* It makes it easier to change or extend the function signature
* In this situation I find it actually positive too explicitly pass in `nil` or empty slice when no dedicated columns are used

tempodb/backend/block_meta.go
```diff
@@ -104,3 +107,24 @@ func (b *BlockMeta) ObjectAdded(id []byte, start uint32, end uint32) {
 
 	b.TotalObjects++
 }
+
+// separatorByte is a byte that cannot occur in valid UTF-8 sequences
+var separatorByte = []byte{255}
+
+// TODO: Find a better way of comparing dedicated columns config
```
stoewer, line 114: I think the method is absolutely valid

tempodb/encoding/v2/encoding.go
```diff
@@ -47,7 +47,7 @@ func (v Encoding) OpenWALBlock(filename string, path string, ingestionSlack time
 }
 
 // CreateWALBlock creates a new appendable block
-func (v Encoding) CreateWALBlock(id uuid.UUID, tenantID string, filepath string, e backend.Encoding, dataEncoding string, ingestionSlack time.Duration) (common.WALBlock, error) {
+func (v Encoding) CreateWALBlock(id uuid.UUID, tenantID string, filepath string, e backend.Encoding, dataEncoding string, ingestionSlack time.Duration, dedicatedColumns ...backend.DedicatedColumn) (common.WALBlock, error) {
```
stoewer, line 50: I'd prefer to use `dedicatedColumns []backend.DedicatedColumn` instead of a variadic parameter here too

stoewer (changes requested):

stoewer:
Currently we don't validate the values for `scope` and `type`. Since we have very strict requirements for both I think we should at least log an error when the config or overrides contain invalid values.

modules/ingester/ingester_test.go
```diff
@@ -339,9 +341,73 @@ func TestFlush(t *testing.T) {
 	}
 }
 
+func TestDedicatedColumns(t *testing.T) {
+	tmpDir, err := os.MkdirTemp("/tmp", "")
+	require.NoError(t, err, "unexpected error getting tempdir")
+	defer os.RemoveAll(tmpDir)
```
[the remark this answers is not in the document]
mapno: Applied

modules/ingester/instance.go
mapno: Changed the tests in `overrides_test.go` to verify that the behaviour is as you describe.

tempodb/backend/block_meta.go
mapno: Changed. Added a new method when necessary, and changed variadic params to slices.

modules/ingester/instance.go
stoewer: I just tested this and unfortunately this does not work: If `dedicated_columns` is set to `[]` in the overrides it takes the dedicated columns from the configuration. Therefore it is currently not possible to "unset" the dedicated columns.

In my opinion the following override configuration should remove all configured dedicated columns:
```yaml
overrides:
  dedicated_columns: []
```
But this isn't the case. If the line above is changed like this it works.
```go
    if cols := i.overrides.DedicatedColumns(i.instanceID); cols != nil {
```

mapno: You're right. Changed 👍

stoewer (approved):
LGTM
Awesome

PR #2564, opened 2023-06-16
Fix ingester locking

### LINKED ISSUE #2534: Ingester search occassionally fails due to race conditions
Ingester search will occasionally throw strange errors such as the below. This occasionally can be seen in the field and in tests.

```
error opening file /tmp/TestInstanceSearchDoesNotRace3179494870/001/753f46d7-ae0d-46c8-8cdc-1f93da4a54a6+fake+vParquet2/0000000001: error opening file: open /tmp/TestInstanceSearchDoesNotRace3179494870/001/753f46d7-ae0d-46c8-8cdc-1f93da4a54a6+fake+vParquet2/0000000001: no such file or directory
```

I was able to semi reliably reproduce this in tests by running the following:

```
go test -run TestInstanceSearchDoesNotRace -cpu 1
```

The issue is here:
https://github.com/grafana/tempo/blob/05aad36fee2dd710c98931565930e799faf3edfb/modules/ingester/instance_search.go#L47-L50

The problem is that `searchWal()` and `searchLocalBlocks()` kick off goroutines to perform the search. Then immediately drop the lock. It's possible for a wal block to be cleared or other changes to occur to the blocks before the search occurs. This can result in error messages like the above. The obvious fix is:

```
  i.blocksMtx.RLock()
  defer i.blocksMtx.RUnlock()
  i.searchWAL(ctx, req, sr)
  i.searchLocalBlocks(ctx, req, sr)
```

The problem with this is that a long running search will block the ingest pattern of live traces -> head block -> completing block -> completed block. If the search goes long enough then ingester memory will balloon b/c it can't dump live traces and it will also stall out the block flushing mechanics resulting in significantly larger than configured blocks.

mdisibio:
**What this PR does**:
Fixes #2534 by keeping the lock until the search is complete, but also splits out a separate mutex for the headblock ease contention on the core write path (live traces -> disk).   I added a new `BenchmarkInstanceContention` to help measure the effect, which is similar to the `xxxDoesNotRace` methods but prints out the op/sec of each thing.   Comparison between `main` and this branch are below. There's definitely a draw back in reads due to the increased lock time, but the write path actually looks better, since it only has to lock the headblock, and I took care to ensure that the read paths release the headblock as soon as possible.

```
name                   old time/op            new time/op            delta
InstanceContention-12             2.37s ±21%             2.69s ±29%      ~     (p=0.105 n=10+10)

name                   old blockflushes/sec   new blockflushes/sec   delta
InstanceContention-12              257 ±124%              886 ±101%  +244.84%  (p=0.000 n=10+10)

name                   old finds/sec          new finds/sec          delta
InstanceContention-12             6.84k ±20%             0.10k ±78%   -98.58%  (p=0.000 n=10+10)

name                   old pushes/sec         new pushes/sec         delta
InstanceContention-12             2.83k ±16%             2.39k ±28%   -15.51%  (p=0.009 n=10+10)

name                   old retentions/sec     new retentions/sec     delta
InstanceContention-12             1.91k ±22%             4.82k ±72%  +151.49%  (p=0.000 n=10+10)

name                   old searchTags/sec     new searchTags/sec     delta
InstanceContention-12               612 ±19%                52 ±36%   -91.51%  (p=0.000 n=10+10)

name                   old searchedBytes/sec  new searchedBytes/sec  delta
InstanceContention-12             25.2M ±13%             29.9M ±37%   +18.80%  (p=0.017 n=9+10)

name                   old searches/sec       new searches/sec       delta
InstanceContention-12               191 ±21%                34 ±20%   -82.00%  (p=0.000 n=10+9)

name                   old traceflushes/sec   new traceflushes/sec   delta
InstanceContention-12             5.43 ±126%              9.83 ±30%   +81.00%  (p=0.006 n=9+10)
```

**Note** Having the two mutexes makes it possible to introduce deadlocks if they are accessed in the wrong order. Left a couple huge comments explaining it, but the brittleness of future dev error is something to consider.

**Note2** I have been debating between this change and an alternate approach [here](https://github.com/grafana/tempo/compare/main...mdisibio:locklist).  The alternative approach is the `granular per-block locking nirvana` that I would like to see.  It reduces contention even more but it's not quite enough to balance out the increased complexity yet.  I'll keep working on it if there are no objections.

Fixes #2534 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/ingester/instance.go
+++ b/modules/ingester/instance.go
@@ -272,8 +272,8 @@ func (i *instance) CutBlockIfReady(maxBlockLifetime time.Duration, maxBlockBytes
 		// Now that we are adding a new block take the blocks mutex.
 		// A warning about deadlocks!!  This area does a hard-acquire of both mutexes.
 		// To avoid deadlocks this function and all others must acquire them in
-		// the ** same_order ** or else. i.e. another function can't acquire blocksMtx
-		// then headblockMtx. Even if the likelihood is low it is a statical certainly
+		// the ** same_order ** or else!!! i.e. another function can't acquire blocksMtx
+		// then headblockMtx. Even if the likelihood is low it is a statistical certainly
 		// that eventually a deadlock will occur.
 		i.blocksMtx.Lock()
 		defer i.blocksMtx.Unlock()
--- a/modules/ingester/instance.go
+++ b/modules/ingester/instance.go
@@ -98,8 +98,10 @@ type instance struct {
 	traceSizes map[uint32]uint32
 	traceCount atomic.Int32
 
+	headBlockMtx sync.RWMutex
+	headBlock    common.WALBlock
+
 	blocksMtx        sync.RWMutex
-	headBlock        common.WALBlock
 	completingBlocks []common.WALBlock
 	completeBlocks   []*localBlock
 
@@ -241,16 +243,16 @@ func (i *instance) CutCompleteTraces(cutoff time.Duration, immediate bool) error
 		tempopb.ReuseByteSlices(t.batches)
 	}
 
-	i.blocksMtx.Lock()
-	defer i.blocksMtx.Unlock()
+	i.headBlockMtx.Lock()
+	defer i.headBlockMtx.Unlock()
 	return i.headBlock.Flush()
 }
 
 // CutBlockIfReady cuts a completingBlock from the HeadBlock if ready.
 // Returns the ID of a block if one was cut or a nil ID if one was not cut, along with the error (if any).
 func (i *instance) CutBlockIfReady(maxBlockLifetime time.Duration, maxBlockBytes uint64, immediate bool) (uuid.UUID, error) {
-	i.blocksMtx.Lock()
-	defer i.blocksMtx.Unlock()
+	i.headBlockMtx.Lock()
+	defer i.headBlockMtx.Unlock()
 
 	if i.headBlock == nil || i.headBlock.DataLength() == 0 {
 		return uuid.Nil, nil
@@ -266,6 +268,15 @@ func (i *instance) CutBlockIfReady(maxBlockLifetime time.Duration, maxBlockBytes
 		}
 
 		completingBlock := i.headBlock
+
+		// Now that we are adding a new block take the blocks mutex.
+		// A warning about deadlocks!!  This area does a hard-acquire of both mutexes.
+		// To avoid deadlocks this function and all others must acquire them in
+		// the ** same_order ** or else. i.e. another function can't acquire blocksMtx
+		// then headblockMtx. Even if the likelihood is low it is a statical certainly
+		// that eventually a deadlock will occur.
+		i.blocksMtx.Lock()
+		defer i.blocksMtx.Unlock()
 
 		i.completingBlocks = append(i.completingBlocks, completingBlock)
 
@@ -390,18 +401,20 @@ func (i *instance) FindTraceByID(ctx context.Context, id []byte) (*tempopb.Trace
 	}
 	i.tracesMtx.Unlock()
 
-	i.blocksMtx.RLock()
-	defer i.blocksMtx.RUnlock()
-
 	combiner := trace.NewCombiner()
 	combiner.Consume(completeTrace)
 
 	// headBlock
+	i.headBlockMtx.RLock()
 	tr, err := i.headBlock.FindTraceByID(ctx, id, common.DefaultSearchOptions())
+	i.headBlockMtx.RUnlock()
 	if err != nil {
 		return nil, fmt.Errorf("headBlock.FindTraceByID failed: %w", err)
 	}
 	combiner.Consume(tr)
+
+	i.blocksMtx.RLock()
+	defer i.blocksMtx.RUnlock()
 
 	// completingBlock
 	for _, c := range i.completingBlocks {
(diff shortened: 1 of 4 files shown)


modules/ingester/instance_search.go
```diff
@@ -83,16 +100,21 @@ func (i *instance) Search(ctx context.Context, req *tempopb.SearchRequest) (*tem
...
 
-// searchWAL starts a search task for every WAL block. Must be called under lock.
-func (i *instance) searchWAL(ctx context.Context, req *tempopb.SearchRequest, sr *search.Results) {
-	searchWalBlock := func(b common.WALBlock) {
-		blockID := b.BlockMeta().BlockID.String()
-		span, ctx := opentracing.StartSpanFromContext(ctx, "instance.searchWALBlock", opentracing.Tags{
-			"blockID": blockID,
-		})
-		defer span.Finish()
+// searchLocalBlocks starts a search task for every local block. Must be called under lock.
```
electron0zero, line 103:
```suggestion
// searchBlock starts a search task for every wal/local block. Must be called under lock.
```

modules/ingester/instance_search.go
```diff
@@ -83,16 +100,21 @@ func (i *instance) Search(ctx context.Context, req *tempopb.SearchRequest) (*tem
...
+func (i *instance) searchBlock(ctx context.Context, req *tempopb.SearchRequest, sr *search.Results, blockID uuid.UUID, block common.Searcher, cleanup func()) {
+	sr.StartWorker()
+	go func(e common.Searcher, cleanup func()) {
+		if cleanup != nil {
+			defer cleanup()
+		}
 		defer sr.FinishWorker()
 
+		span, ctx := opentracing.StartSpanFromContext(ctx, "instance.searchLocalBlocks")
```
electron0zero, line 112:
```suggestion
		span, ctx := opentracing.StartSpanFromContext(ctx, "instance.searchBlocks")
```

modules/ingester/instance_search.go
```diff
@@ -83,16 +100,21 @@ func (i *instance) Search(ctx context.Context, req *tempopb.SearchRequest) (*tem
...
+		defer span.Finish()
+
+		span.LogFields(ot_log.Event("local block entry mtx acquired"))
```
electron0zero, line 115:
```suggestion
		span.LogFields(ot_log.Event("block entry mtx acquired"))
```

modules/ingester/instance_search.go
```diff
@@ -101,92 +123,30 @@ func (i *instance) searchWAL(ctx context.Context, req *tempopb.SearchRequest, sr
...
 		}
 
 		if err == common.ErrUnsupported {
-			level.Warn(log.Logger).Log("msg", "wal block does not support search", "blockID", b.BlockMeta().BlockID)
+			level.Warn(log.Logger).Log("msg", "block does not support search", "blockID", blockID)
 			return
 		}
 		if err != nil {
-			level.Error(log.Logger).Log("msg", "error searching local block", "blockID", blockID, "block_version", b.BlockMeta().Version, "err", err)
+			level.Error(log.Logger).Log("msg", "error searching local block", "blockID", blockID, "err", err)
```
electron0zero, line 137:
```suggestion
			level.Error(log.Logger).Log("msg", "error searching block", "blockID", blockID, "err", err)
```

electron0zero: this code will be working on wal and local blocks both? right?

electron0zero (approved):
minor comments, rest lgtm

modules/ingester/instance.go
```diff
@@ -267,6 +269,15 @@ func (i *instance) CutBlockIfReady(maxBlockLifetime time.Duration, maxBlockBytes
 
 		completingBlock := i.headBlock
 
+		// Now that we are adding a new block take the blocks mutex.
+		// A warning about deadlocks!!  This area does a hard-acquire of both mutexes.
+		// To avoid deadlocks this function and all others must acquire them in
+		// the ** same_order ** or else. i.e. another function can't acquire blocksMtx
```
joe-elliott, line 275: not enough !!! after "or else".

how confident are we that "TestInstanceDoesNotRace" will catch deadlocks?

modules/ingester/instance_search.go
```diff
@@ -41,13 +42,29 @@ func (i *instance) Search(ctx context.Context, req *tempopb.SearchRequest) (*tem
 	sr := search.NewResults()
 	defer sr.Close() // signal all running workers to quit
 
-	// Lock blocks mutex until all search tasks have been created. This avoids
+	// Lock headblock separately from other blocks and release it as soon as this
+	// subtask is finished.
+	// A warning about deadlocks!!  This area does a hard-acquire of both mutexes.
+	// To avoid deadlocks this function and all others must acquire them in
+	// the ** same_order ** or else. i.e. another function can't acquire blocksMtx
+	// then headblockMtx. Even if the likelihood is low it is a statical certainly
```
joe-elliott, line 50: statical -> statistical

joe-elliott (approved):

modules/ingester/instance.go
mdisibio: Maybe you were joking but more !!! added.    Actually the deadlocks were being caught by `TestInstanceSearchDoesNotRace` because it exercises more read path (search, tags, tag values). I tested with count=10 and it virtually always triggered until I fixed them.  I think just one time it didn't.  So probably nearly 100% within 20 runs?   Our CI runs often enough it should be good, and the failure mode will be a timeout in one of these DoesNotRace tests.

joe-elliott (approved):

Issue #1958, opened 2022-12-15
Fix distributor http status codes


joe-elliott:
**Describe the bug**
As reported by the community if the ingester returns trace too large then the returned http error is 500. This error suggests to the otel collector to retry however it should return 400 which suggests to not retry. Currently the GRPC status codes correctly suggest retrying or not.

Review the status codes for various errors and return 400s where appropriate.

joe-elliott:
@mghildiy is going to take a crack at this.

cc @ie-pham b/c you're also looking at return codes (but for proto/grpc)

mghildiy:
So currently error code returned is 'FailedPrecondition

https://github.com/grafana/tempo/blob/c0127c78c368319433c7c67ca8967adbfed2259e/modules/ingester/instance.go#L189

https://github.com/grafana/tempo/blob/c0127c78c368319433c7c67ca8967adbfed2259e/modules/ingester/instance.go#L198

...defined here:

https://github.com/grafana/tempo/blob/c0127c78c368319433c7c67ca8967adbfed2259e/vendor/google.golang.org/grpc/codes/codes.go#L121

A better option can be:
https://github.com/grafana/tempo/blob/c0127c78c368319433c7c67ca8967adbfed2259e/vendor/google.golang.org/grpc/codes/codes.go#L57

Surprisingly, [grpC codes to http codes mapping](https://chromium.googlesource.com/external/github.com/grpc/grpc/+/refs/tags/v1.21.4-pre1/doc/statuscodes.md) says that both map to 400.

ie-pham:
Hi @mghildiy ,

Thank you so much for your contribution to this issue. Unfortunately, this issue is no longer is no longer relevant due to our shift towards a "partial success" solution, as outlined in #2571. In addition, [Otel](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#partial-success-1) states that partial successes should return a 200 instead of an error. 

Your contribution to this repo is well observed by the team and is highly valued. Don't be deterred to explore other opportunities to contribute!  

Closing this issue.

PR #2571, opened 2023-06-19
Handle Partial Success in ConsumeTraces

### LINKED ISSUE #1957: Complete writing a batch even if some spans refused

ie-pham:
**What this PR does**: 

1. Right now if a single trace in the batch triggers an error, Tempo drops the entire batch. This PR allows every trace in the batch to get a turn at getting pushed before returning a response to the client. 

2. From internal discussion, we are going to match what is required [here](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#partial-success-1) for partial success and return a `200` instead of an error. 

The tricky part was tallying up the discarded span count.

If replication factor is 3, then the minimum success is 2 (rep factor / 2 +1) and the max allowable error is 1. So initially I thought a simple logic should account for this. Tally up all the responses and if a trace is marked as "error" twice then we can consider that trace discarded. However, this approach is only doable if we were working with cases where the number of responses = replication factor. That is not the case. We think the ring.DoBatch function returns as soon as 2 errors or 2 successes are recorded without waiting for the third response (a partial success is still considered a success). In a perfect world where two those responses are exact the same, we would not have a problem. However, if we have a scenario where the lists of discarded traces are different in the two responses, it is hard to accurately determine whether a trace was discarded or not without the third response. 

Decision:
Since the discarded span count currently does not play an important role in the functionality of Tempo, I am going with the "over counting" solution. 

Since the number of responses is always the minimum success or more
```
	// max number of error allowed is {number of responses} - {minimum success required}
```
So if the replication factor is 3 and we get only 2 responses (and the minimum success required is 2), then the max number of error allowed is 0. If a trace is marked "error" even once, it will be considered discarded. However, for example, for replication of 5 with a minimum success of 3, if we only get 4 responses then with the same rules, the max error allowed is 1 not 0. 

This solution has drawbacks because we may overcount the number of discarded spans.  For example if we have one partial success and one total success, without the third response, there is a chance that whatever trace failed in the partial success actually was successfully processed in the third response but we will mark it as discarded. 

Another note is I am adding a mutex lock to collect the responses from the ingester. Will need to test this to make sure it does not impact performance too much. 

Fixes #1957

**Checklist**
- [x] Tests updated
- [ ] Documentation added


ie-pham:
So the logic works but I'm just not sure if it's too convoluted to record the correct discarded span count by reason

modules/distributor/distributor.go
```diff
@@ -378,11 +393,20 @@ func (d *Distributor) sendToIngestersViaBytes(ctx context.Context, userID string
...
 			SearchData: nil, // support for flatbuffer/v2 search has been removed. todo: cleanup the proto
+			SpanCounts: make([]tempopb.PreallocBytes, len(indexes)),
 		}
 
 		for i, j := range indexes {
 			req.Traces[i].Slice = marshalledTraces[j][0:]
 			req.Ids[i].Slice = traces[j].id
+
+			// convert span count int to byte : int => string => byte
+			spanCount, err := tempo_util.HexStringToTraceID(strconv.Itoa(traces[j].spanCount))
```
joe-elliott, line 404: definitely shouldn't be going int -> string -> byte. that's a large performance penalty to go from int -> byte

pkg/tempopb/tempo.proto
```diff
@@ -172,6 +172,8 @@ message PushBytesRequest {
   repeated bytes ids = 3 [(gogoproto.nullable) = false, (gogoproto.customtype) = "PreallocBytes"];
   // search data, length must match traces
   repeated bytes searchData = 4 [(gogoproto.nullable) = false, (gogoproto.customtype) = "PreallocBytes"];
+    // span count per trace
+  repeated bytes spanCounts = 5 [(gogoproto.nullable) = false, (gogoproto.customtype) = "PreallocBytes"];
```
joe-elliott, line 176: if span count is 1 byte then the max spans it can track is 2^8 which is likely not enough in a lot of scenarios. i question whether passing the span count at all is worth it tho.

modules/ingester/instance.go
```diff
@@ -149,12 +166,66 @@ func newInstance(instanceID string, limiter *Limiter, writer tempodb.Writer, l *
...
+				internalErorCount += spanCount
+				if !strings.Contains(errors.Error(), err.Error()) {
+					errors.Add(err)
+				}
+			}
 		}
 	}
+
+	if len(errors) > 0 {
+		discardCountMessage := fmt.Sprintf(
```
joe-elliott, line 218: if we're going to pass back rich information like this we should just change the `PushResponse` method to have these fields.

joe-elliott (commented):
I question the efforts to perfectly record dropped spans for each reason. I'm wondering if a better approach would be to change `PushResponse` to return trace indexes and rejection reasons back to the distributor. The distributor could then update the appropriate metrics based on the response.

modules/ingester/instance.go
ie-pham: Yeah I originally had tried doing this but the response object never gets back to the distributor and I just assumed that was not possible. I will try to figure out what I did wrong there. Is there any additional things I need to do besides changing the protos for the response?

modules/distributor/distributor.go
```diff
@@ -353,12 +360,12 @@ func (d *Distributor) PushTraces(ctx context.Context, traces ptrace.Traces) (*te
...
 }
 
-func (d *Distributor) sendToIngestersViaBytes(ctx context.Context, userID string, traces []*rebatchedTrace, keys []uint32) error {
-	// Marshal to bytes once
+func (d *Distributor) sendToIngestersViaBytes(ctx context.Context, userID string, traces []*rebatchedTrace, keys []uint32, totalSpanCount int, tracesByID map[uint32]*rebatchedTrace) error {
 	marshalledTraces := make([][]byte, len(traces))
 	for i, t := range traces {
 		b, err := d.traceEncoder.PrepareForWrite(t.trace, t.start, t.end)
 		if err != nil {
+			overrides.RecordDiscardedSpans(totalSpanCount, reasonInternalError, userID)
```
joe-elliott, line 368: i think we should leave this outside `sendToIngesterViaBytes`. Basically either:

- `sendToIngesterViaBytes` returns an error meaning something unexpected happened and we can record drops due to internal errors.
- `sendToIngesterViaBytes` returns nil and internally it will record drops due to policy reasons.

i'm wondering if we want to completely stop recording internal error in metrics, and just rely on the error return. that would be for a future PR.

modules/distributor/distributor.go
```diff
@@ -390,14 +401,51 @@ func (d *Distributor) sendToIngestersViaBytes(ctx context.Context, userID string
...
 		}
 
-		_, err = c.(tempopb.PusherClient).PushBytesV2(localCtx, &req)
+		response, err := c.(tempopb.PusherClient).PushBytesV2(localCtx, &req)
+		// response = nil, err = nil ==> entire batch processed successfully
+		// response = nil, err != nill ==> entire batch failed not related to livetraces or tracetoolong error
+		// response != nil, err = nil  ==>  at least one trace was discarded
 		metricIngesterAppends.WithLabelValues(ingester.Addr).Inc()
+
+		r.mu.Lock()
```
joe-elliott, line 410: should we move this to the bottom? after checking response and err?

modules/distributor/distributor.go
```diff
@@ -390,14 +401,51 @@ func (d *Distributor) sendToIngestersViaBytes(ctx context.Context, userID string
...
+	}
+
+	// count discarded span count
+	// due to the way that doBatch works, there is a possibility of overcounting discarded span count
+	// because we are not waiting for all responses to come back before returning
+	// but we predict a low frequency for that scenario
+	// todo: figure out a more accurate way to count discarded spans
+	r.mu.Lock()
+	maxLiveDiscardedCount, traceTooLargeDiscardedCount, internalErrorDiscardedCount := countDiscaredSpans(r.responses, tracesByID, d.ingestersRing.ReplicationFactor())
+	r.mu.Unlock()
```
joe-elliott, line 439: deferring unlock is preferred when possible

modules/distributor/distributor.go
```diff
@@ -515,23 +567,53 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount int
...
+	numResponses := len(responses)
+	quorum := (repFactor / 2) + 1 // min success required
+
+	// max number of error allowed is {number of responses} - {minimum success required}
+	// + 1 for error count that should mark a trace as discarded
+	maxError := (numResponses - quorum) + 1
+
+	// if a trace triggers two different errors in two different responses, the error that triggers the maxError count will be used
+	for _, response := range responses {
+		if response == nil {
```
joe-elliott, line 587: i'm concerned about this error handling. if we get an error in DoBatch() above i'm fine with keeping it simple and just letting that error propagate all the way back just like it does now.

modules/ingester/ingester.go
```diff
@@ -239,15 +239,18 @@ func (i *Ingester) PushBytesV2(ctx context.Context, req *tempopb.PushBytesReques
...
 	if err != nil {
+		level.Warn(log.Logger).Log(err.Error())
 		return nil, err
 	}
 
-	err = instance.PushBytesRequest(ctx, req)
-	if err != nil {
-		return nil, err
+	response, _ := instance.PushBytesRequest(ctx, req)
+	// distributor will determine if an error should be returned to the client
```
joe-elliott, line 247: i think we should continue returning error. any unexpected error in the path should just make the whole thing 500. previously we treated drops due to limits as an error, but this PR changes that.

- everything works perfectly: 200/OK
- nothing fails, but we drop spans: 200/OK and tick up our metrics
- unexpected error: 500/INTERNAL

joe-elliott (commented):
I have reviewed mainly the distributor part and I think we need to make some changes so I'm not quite going to dig deep on the ingester part yet.

Overall, this is really heading the right direction. A lot of comments and questions b/c this is a very performance sensitive part of the code and a very nuanced change.

Also, I think this invalidates: https://github.com/grafana/tempo/issues/1958 since we are just going to start returning 200 on dropped spans. Can you confirm?

pkg/tempopb/tempo.proto
```diff
@@ -163,6 +163,12 @@ message Trace {
 message PushResponse {
 }
 
+message IngesterPushResponse {
+  repeated uint32 maxLiveErrorTraces = 1;
```
joe-elliott, line 167: we can just add these fields to the existing `PushResponse`. changing the response type will change the method signature which i think will break Tempo on rollout. 

speaking of, we will need to be thoughtful about what a rollout of this change will look like. we will need to nail down some more of the behavior as discussed in the comments first.

modules/distributor/distributor.go
```diff
@@ -330,15 +338,14 @@ func (d *Distributor) PushTraces(ctx context.Context, traces ptrace.Traces) (*te
 			size)
 	}
 
-	keys, rebatchedTraces, err := requestsByTraceID(batches, userID, spanCount)
+	keys, rebatchedTraces, tracesByID, err := requestsByTraceID(batches, userID, spanCount)
```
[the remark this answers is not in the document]
ie-pham: good call, fixed

modules/distributor/distributor.go
ie-pham: Have we ever used this for any debugging purposes?

joe-elliott: Not that I'm aware of, but that metric is exposed to our cloud users so we would want to consider the impact before changing it.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 * [FEATURE] New encoding vParquet3 with support for dedicated attribute columns (@mapno, @stoewer) [#2649](https://github.com/grafana/tempo/pull/2649)
 * [ENHANCEMENT] Assert ingestion rate limits as early as possible [#2640](https://github.com/grafana/tempo/pull/2703) (@mghildiy)
 
+* [CHANGE] **Breaking Change** Fix issue where tempo drops the entire batch if one trace triggers an error [#2571](https://github.com/grafana/tempo/pull/2571) (@ie-pham)
```
joe-elliott, line 6: Add details about why this is breaking. I believe it's b/c we are now returning 200/OK when we have partial success on write. A sentence about how this impacts the discarded spans metric would be good too.

modules/distributor/distributor.go
```diff
@@ -141,6 +142,11 @@ type Distributor struct {
 	logger log.Logger
 }
 
+type PushByteRes struct {
```
joe-elliott, line 145: except for bundling these two fields together there's not really a reason for this struct to exist. i would personally not create this for simplicity, but i'm not strongly opposed to it either

modules/distributor/distributor.go
```diff
@@ -398,14 +407,36 @@ func (d *Distributor) sendToIngestersViaBytes(ctx context.Context, userID string
...
-		_, err = c.(tempopb.PusherClient).PushBytesV2(localCtx, &req)
+		response, err := c.(tempopb.PusherClient).PushBytesV2(localCtx, &req)
+		// response = {empty}, err = nil ==> entire batch processed successfully
+		// response = nil, err != nill ==> entire batch failed not related to livetraces or tracetoolong error
+		// response != {empty}, err = nil  ==>  at least one trace was discarded
 		metricIngesterAppends.WithLabelValues(ingester.Addr).Inc()
+
+		r.mu.Lock()
+		defer r.mu.Unlock()
+		r.responses = append(r.responses, response)
```
joe-elliott, line 418: i think we should check for err != nil before the locking. if err != nil there's no reason to accumulate the responses.

modules/distributor/distributor.go
```diff
@@ -526,20 +561,52 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount int
...
-func recordDiscaredSpans(err error, userID string, spanCount int) {
-	s := status.Convert(err)
-	if s == nil {
-		return
-	}
-	desc := s.Message()
+func countDiscaredSpans(responses []*tempopb.PushResponse, traces []*rebatchedTrace, repFactor int) (int, int, int) {
+	maxLiveDiscardedCount := 0
+	traceTooLargeDiscardedCount := 0
+	internalErrorDiscardedCount := 0
```
joe-elliott, line 567: i don't think we need to count internal error here. `DoBatch()` will do the work of accounting for an error while sending to the ingesters.

i think we can drop the error handling/response == nil logic here and simplify.

modules/ingester/instance.go
```diff
@@ -154,14 +170,50 @@ func newInstance(instanceID string, limiter *Limiter, overrides ingesterOverride
...
 	for j := range req.Traces {
+		index := int32(j)
+
 		err := i.PushBytes(ctx, req.Ids[j].Slice, req.Traces[j].Slice)
 		if err != nil {
-			return err
+			errorMessage := err.Error()
+			totalTracesDiscarded++
+
+			if strings.Contains(errorMessage, overrides.ErrorPrefixLiveTracesExceeded) {
```
joe-elliott, line 188: we can compare the errors directly. We no longer need to do string comparisons.

modules/ingester/instance.go
```diff
@@ -154,13 +172,49 @@ func newInstance(instanceID string, limiter *Limiter, overrides ingesterOverride
...
 	for j := range req.Traces {
+		index := int32(j)
+
 		err := i.PushBytes(ctx, req.Ids[j].Slice, req.Traces[j].Slice)
 		if err != nil {
-			return err
+			errorMessage := err.Error()
+			totalTracesDiscarded++
+
+			if _, ok := err.(*maxLiveTracesError); ok {
```
joe-elliott, line 190: let's use `errors.Is()`. it will handle error wrapping and is the canonical method of detecting error equality in go

modules/ingester/instance.go
```diff
@@ -154,13 +172,49 @@ func newInstance(instanceID string, limiter *Limiter, overrides ingesterOverride
 	return i, nil
 }
 
-func (i *instance) PushBytesRequest(ctx context.Context, req *tempopb.PushBytesRequest) error {
+func (i *instance) PushBytesRequest(ctx context.Context, req *tempopb.PushBytesRequest) *tempopb.PushResponse {
+	var errors util.MultiError
+
+	maxLiveErrorTraces := make([]int32, 0)
```
joe-elliott, line 178: the compiler likely does this for us, but we should not call `make()` here so the slices are only allocated if an error occurs

modules/ingester/instance.go
```diff
@@ -154,13 +172,49 @@ func newInstance(instanceID string, limiter *Limiter, overrides ingesterOverride
...
 		if err != nil {
-			return err
+			errorMessage := err.Error()
+			totalTracesDiscarded++
+
+			if _, ok := err.(*maxLiveTracesError); ok {
+				maxLiveErrorTraces = append(maxLiveErrorTraces, index)
+				// only log one of each occurrence
+				if !strings.Contains(errorMessage, overrides.ErrorPrefixLiveTracesExceeded) {
+					errors.Add(err)
```
joe-elliott, line 194: how about we just log these at a warn level?

modules/ingester/instance.go
```diff
@@ -154,13 +172,49 @@ func newInstance(instanceID string, limiter *Limiter, overrides ingesterOverride
...
+			}
 		}
 	}
+
+	// at least one failure
+	if totalTracesDiscarded > 0 {
+		return &tempopb.PushResponse{
+			MaxLiveErrorTraces:       maxLiveErrorTraces,
+			TraceTooLargeErrorTraces: traceTooLargeErrorTraces,
+			Error:                    errors.Error(),
```
joe-elliott, line 213: is there any reason to return the error to the ingester?

modules/ingester/instance.go
```diff
@@ -154,13 +172,49 @@ func newInstance(instanceID string, limiter *Limiter, overrides ingesterOverride
...
+		}
+	}
+
+	// no failure
```
joe-elliott, line 217: to reduce complexity it might make more sense to always return a value `tempopb.PushResponse`. even if there are no errors. in this case the slices will just be empty.

joe-elliott (commented):
heading the right direction

modules/ingester/instance.go
ie-pham: I was thinking that we might want to eventually return this error message once we start implementing OTEL `export` call with the partial success error. Should we just leave it out until we implement it?

joe-elliott: yeah, leave it out for now. i think it tightens the design.

modules/distributor/distributor.go
```diff
@@ -398,15 +401,39 @@ func (d *Distributor) sendToIngestersViaBytes(ctx context.Context, userID string
 			return err
 		}
 
-		_, err = c.(tempopb.PusherClient).PushBytesV2(localCtx, &req)
+		response, err := c.(tempopb.PusherClient).PushBytesV2(localCtx, &req)
+		// response = {empty}, err = nil ==> entire batch processed successfully
+		// response = nil, err != nill ==> entire batch failed not related to livetraces or tracetoolong error
+		// response != {empty}, err = nil  ==>  at least one trace was discarded
```
kvrhdn, line 407: I feel like this would be easier to read as follows:

```suggestion
		response, err := c.(tempopb.PusherClient).PushBytesV2(localCtx, &req)
		metricIngesterAppends.WithLabelValues(ingester.Addr).Inc()

		if err != nil {
			// internal error, abandon the entire batch
			metricIngesterAppendFailures.WithLabelValues(ingester.Addr).Inc()
			return err
		}
		
		// batch was (partially) accepted, gather responses and we are done

		return nil
```

modules/distributor/distributor.go
```diff
@@ -398,15 +401,39 @@ func (d *Distributor) sendToIngestersViaBytes(ctx context.Context, userID string
...
 		metricIngesterAppends.WithLabelValues(ingester.Addr).Inc()
-		if err != nil {
+
+		if response != nil {
+			mu.Lock()
+			defer mu.Unlock()
+			responses = append(responses, response)
 			metricIngesterAppendFailures.WithLabelValues(ingester.Addr).Inc()
 		}
```
kvrhdn, line 415: The `metricIngesterAppendFailures.WithLabelValues(ingester.Addr).Inc()` in here is wrong, this is the branch for successful push requests right?

modules/distributor/distributor.go
```diff
@@ -526,20 +557,39 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount int
...
 
-func recordDiscaredSpans(err error, userID string, spanCount int) {
-	s := status.Convert(err)
-	if s == nil {
-		return
-	}
-	desc := s.Message()
+func countDiscaredSpans(responses []*tempopb.PushResponse, traces []*rebatchedTrace, repFactor int) (int, int) {
+	maxLiveDiscardedCount := 0
+	traceTooLargeDiscardedCount := 0
```
kvrhdn, line 562: Nit: you can used named return values, makes the method more explicit + you don't need to initialise these

```suggestion
func countDiscaredSpans(responses []*tempopb.PushResponse, traces []*rebatchedTrace, repFactor int) (maxLiveDiscardedCount int, traceTooLargeDiscardedCount int) {
```

The line gets a bit long though... 🤔

kvrhdn (commented):
Still looking through instance.go and the e2e test but already going to release some comments 🙂

modules/distributor/distributor.go
ie-pham: ohh i like this

modules/distributor/distributor.go
```diff
@@ -397,15 +400,37 @@ func (d *Distributor) sendToIngestersViaBytes(ctx context.Context, userID string
...
-		return err
+
+		mu.Lock()
+		defer mu.Unlock()
+		responses = append(responses, response)
+		return nil
 	}, func() {})
+	// if err != nil, we discarded everything because of an internal error
+	if err != nil {
+		return err
```
joe-elliott, line 418: nit: it might be clearer at this point to move:
```
overrides.RecordDiscardedSpans(spanCount, reasonInternalError, userID)
```
in here since all discarded span counts are recorded right here.

modules/distributor/distributor.go
```diff
@@ -376,6 +376,9 @@ func (d *Distributor) sendToIngestersViaBytes(ctx context.Context, userID string
 		op = ring.Write
 	}
 
+	responses := make([]*tempopb.PushResponse, 0, d.ingestersRing.ReplicationFactor())
```
joe-elliott, line 379: won't this be the count of all ingesters? one response for each ingester? not sure how easy it is to get that number.

modules/distributor/distributor.go
```diff
@@ -525,20 +554,40 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount int
...
+	maxError := (numResponses - quorum) + 1
+
+	// if a trace triggers two different errors in two different responses, the error that triggers the maxError count will be used
+	for _, response := range responses {
+		if response == nil { // no errors
+			continue
+		}
+		for _, traceIndex := range response.MaxLiveErrorTraces {
+			traceIndex := int(traceIndex)
+			discardedTraces[traceIndex]++
```
joe-elliott, line 573: need safeguard here. it would be too easy for a bug to panic the distributors

modules/distributor/distributor.go
```diff
@@ -525,20 +554,40 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount int
...
 
-	if strings.HasPrefix(desc, overrides.ErrorPrefixLiveTracesExceeded) {
-		overrides.RecordDiscardedSpans(spanCount, reasonLiveTracesExceeded, userID)
-	} else if strings.HasPrefix(desc, overrides.ErrorPrefixTraceTooLarge) {
-		overrides.RecordDiscardedSpans(spanCount, reasonTraceTooLarge, userID)
-	} else {
-		overrides.RecordDiscardedSpans(spanCount, reasonInternalError, userID)
+	// max number of error allowed is {number of responses} - {minimum success required}
+	// + 1 for error count that should mark a trace as discarded
+	maxError := (numResponses - quorum) + 1
```
joe-elliott, line 564: i may be misunderstanding but i don't think this is true. i think there's one response for each ingester, but each trace will actually only be written to 3 (in the case of RF3). I think max error is:
```
maxError = (repFactor / 2) // integer division
```
the reason max error is below true quorum is b/c we don't require 1 write per batch to complete before this logic hits and for safety we consider it a failure.

modules/distributor/distributor.go
```diff
@@ -525,20 +554,40 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount int
...
+			discardedTraces[traceIndex]++
+			if discardedTraces[traceIndex] == maxError {
+				spanCount := traces[traceIndex].spanCount
+				maxLiveDiscardedCount += spanCount
+			}
+		}
+
+		for _, traceIndex := range response.TraceTooLargeErrorTraces {
+			traceIndex := int(traceIndex)
+			discardedTraces[traceIndex]++
```
joe-elliott, line 582: same as above

modules/ingester/instance.go
```diff
@@ -15,7 +16,7 @@ import (
 	"github.com/gogo/status"
 	"github.com/google/uuid"
 	"github.com/opentracing/opentracing-go"
-	"github.com/pkg/errors"
+	pkgErrors "github.com/pkg/errors"
```
joe-elliott, line 19: any particular reason we're adding this? .New, .Wrap, and .Wrapf all appear in the standard `errors` package

joe-elliott (commented):
getting down to the nitty gritty. Have a few comments, let me know what you think.

This change is close!

modules/ingester/instance.go
ie-pham: That was already there. When i imported the standard `errors` package, I had to rename that particular `errors` package. But I could also just not import the standard package and just use what's already there, which is that `pkg/errors`.

joe-elliott: oh, i see. it would be my inclination to just move everything to `errors`

ie-pham:
So... I refactored

Before we would collect all the responses and then loop through the responses to count the number of times a specific trace has errored to determine whether it was truly discarded or not. 

Now:

We will have
```go
batchResults := make([][]int, len(keys))
```
where the outer lay represents the traces and the inner represents the responses from `ingester.PushBytesV2()` collected for each trace. The responses will be represented by 0, 1, 2 where 0=success, 1=maxlivetraces, 2=tracetoolarge

At the end we will iterate through batchResults and any trace that do not have the minimum required successes (the minimum required 0s) will be marked as discarded.

modules/distributor/distributor.go
```diff
@@ -397,15 +405,54 @@ func (d *Distributor) sendToIngestersViaBytes(ctx context.Context, userID string
...
+		if err != nil { // internal error, drop entire batch
 			metricIngesterAppendFailures.WithLabelValues(ingester.Addr).Inc()
+			return err
 		}
-		return err
+
+		// assign a number for each index that had an error
+		// maxLive = 1, traceTooLarge = 2
+		// indexes that do not have an error will have a default 0 value
+		ringBatchResults := make([]int, len(indexes))
```
joe-elliott, line 419: i'm fine with something simpler. we don't need to perfectly count all reasons failed, right?  for instance in RF3 if a trace is rejected once for max live and once for too large then there is no "right answer" for how to metric dropped spans. all we really need to know is total errors as well as one (the first?) of the rejection reasons.

modules/distributor/distributor.go
```diff
@@ -397,15 +405,54 @@ func (d *Distributor) sendToIngestersViaBytes(ctx context.Context, userID string
...
+		for _, value := range response.MaxLiveErrorTraces {
+			ringBatchResults[value] = maxLiveTracesErrInt
+		}
+		for _, value := range response.TraceTooLargeErrorTraces {
+			ringBatchResults[value] = traceTooLargeErrorInt
```
joe-elliott, line 424: probably should double check the value before indexing into the slice.

modules/distributor/distributor.go
joe-elliott: allocating this for every push might get costly. can't we translate the response data directly into the batchResults from the response using the indexes?

modules/distributor/distributor.go
```diff
@@ -376,6 +378,12 @@ func (d *Distributor) sendToIngestersViaBytes(ctx context.Context, userID string
 		op = ring.Write
 	}
 
+	batchResults := make([][]int, len(keys))
```
joe-elliott, line 381: this is tough. for each trace we need to know 2 things to accurately count refused spans:

- number of successes
- first (or last or any) reported error

it's possible there is an optimization to be made here to only track these fields instead of every returned value for each trace. we'd need to do some benchmarking to determine if there was a better path. possibly something to keep in mind while we finish up.

modules/ingester/instance.go
```diff
@@ -154,14 +152,28 @@ func newInstance(instanceID string, limiter *Limiter, overrides ingesterOverride
...
-func (i *instance) PushBytesRequest(ctx context.Context, req *tempopb.PushBytesRequest) error {
+func (i *instance) PushBytesRequest(ctx context.Context, req *tempopb.PushBytesRequest) *tempopb.PushResponse {
+	pr := &tempopb.PushResponse{}
+
 	for j := range req.Traces {
+
 		err := i.PushBytes(ctx, req.Ids[j].Slice, req.Traces[j].Slice)
 		if err != nil {
-			return err
+			if errors.Is(err, errMaxLiveTraces) {
```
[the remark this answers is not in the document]
ie-pham: it is using `errMaxLiveTraces/errMaxLiveTraces`

the functions newMaxLiveTracesError/newMaxLiveTracesError just add a step to print a warning log

```go
var (
	errTraceTooLarge = errors.New(overrides.ErrorPrefixTraceTooLarge)
	errMaxLiveTraces = errors.New(overrides.ErrorPrefixLiveTracesExceeded)
)

func newTraceTooLargeError(traceID common.ID, instanceID string, maxBytes, reqSize int) error {
	level.Warn(log.Logger).Log("msg", fmt.Sprintf("%s: max size of trace (%d) exceeded while adding %d bytes to trace %s for tenant %s",
		overrides.ErrorPrefixTraceTooLarge, maxBytes, reqSize, hex.EncodeToString(traceID), instanceID))
	return errTraceTooLarge
}
```

modules/ingester/instance.go
```diff
@@ -34,34 +35,31 @@ import (
...
 
 // Errors returned on Query.
 var (
-	ErrTraceMissing = errors.New("Trace missing")
+	ErrTraceMissing = pkgErrors.New("Trace missing")
 )
 
 const (
-	traceDataType = "trace"
+	traceDataType         = "trace"
```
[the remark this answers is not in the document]
ie-pham: apparently here
```
	// measure received bytes as sum of slice lengths
	// type byte is guaranteed to be 1 byte in size
	// ref: https://golang.org/ref/spec#Size_and_alignment_guarantees
	i.bytesReceivedTotal.WithLabelValues(i.instanceID, traceDataType).Add(float64(len(traceBytes)))
```

modules/ingester/instance.go
```diff
@@ -34,34 +35,31 @@ import (
...
-		overrides.ErrorPrefixTraceTooLarge, e.maxBytes, e.reqSize, hex.EncodeToString(e.traceID), e.instanceID)
+func newMaxLiveTracesError(instanceID string, limit string) error {
+	level.Warn(log.Logger).Log("msg", fmt.Sprintf("%s max live traces exceeded for tenant %s: %v", overrides.ErrorPrefixLiveTracesExceeded, instanceID, limit))
+	return errMaxLiveTraces
 }
 
 // Errors returned on Query.
 var (
-	ErrTraceMissing = errors.New("Trace missing")
+	ErrTraceMissing = pkgErrors.New("Trace missing")
```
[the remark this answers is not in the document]
ie-pham: apparently it is not used anywhere, I'll get rid of it

modules/distributor/distributor.go
ie-pham: the only optimization I can think of which I don't know if it will be an optimization is:

```go

// instead of batchResults
NumSuccessByTraceIndex   = make([]int, len(keys)
LastPushErrorByTraceIndex = make([]int, len(keys)
```

And every time we get a PushResponse, we update each trace accordingly. What do you think?

modules/frontend/v1/frontendv1pb/frontend.pb.go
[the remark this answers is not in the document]
ie-pham: i probably just need to rebase and gen-proto again 👍

jcmackie:
Hiya @joe-elliott et al.

We are pretty keen to see this merged into Tempo.
As you can probably imagine we have many traces that hit our size limit, and it's nice to be able to set that limit and protect our Tempo deployment from ridiculously sized traces.

But if we are also losing lots of spans for traces that are otherwise okay. We've tuned down our batch size to reduce the blast radius but it only helps so much.

If we feel confident that all the trace data that should get into Tempo can do so successfully, that will go a long way towards our confidence in the reliability of this tool.

Please let us know if there's any way we can help.

modules/ingester/instance.go
```diff
@@ -154,14 +144,39 @@ func newInstance(instanceID string, limiter *Limiter, overrides ingesterOverride
...
+	atLeastOneError := false
+	pr := &tempopb.PushResponse{}
+
 	for j := range req.Traces {
+
 		err := i.PushBytes(ctx, req.Ids[j].Slice, req.Traces[j].Slice)
 		if err != nil {
-			return err
+			// only make list if there is at least one error
+			if !atLeastOneError {
```
joe-elliott, line 156: let's make this block a function for clarity. we can also test it directly which is nice b/c it does some weird things.

something like:
```
  addTraceError([]tempopb.PushErrorReason, pos int, err error) []tempopb.PushErrorReason
```

modules/ingester/instance.go
```diff
@@ -154,14 +144,39 @@ func newInstance(instanceID string, limiter *Limiter, overrides ingesterOverride
...
+			}
+			if errors.Is(err, errMaxLiveTraces) {
+				pr.ErrorsByTrace = append(pr.ErrorsByTrace, tempopb.PushErrorReason_MAX_LIVE_TRACES)
+				continue
+			}
+
+			if errors.Is(err, errTraceTooLarge) {
+				pr.ErrorsByTrace = append(pr.ErrorsByTrace, tempopb.PushErrorReason_TRACE_TOO_LARGE)
+				continue
+			}
```
joe-elliott, line 173: if error is neither max live traces or trace too large do we just swallow it here? should we return other errors?

modules/distributor/distributor.go
```diff
@@ -530,19 +561,54 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount int
...
-	desc := s.Message()
 
-	if strings.HasPrefix(desc, overrides.ErrorPrefixLiveTracesExceeded) {
-		overrides.RecordDiscardedSpans(spanCount, reasonLiveTracesExceeded, userID)
-	} else if strings.HasPrefix(desc, overrides.ErrorPrefixTraceTooLarge) {
-		overrides.RecordDiscardedSpans(spanCount, reasonTraceTooLarge, userID)
+	return maxLiveDiscardedCount, traceTooLargeDiscardedCount
+}
+
+func (d *Distributor) processPushResponse(pushResponse *tempopb.PushResponse, numSuccessByTraceIndex []int, lastErrorReasonByTraceIndex []tempopb.PushErrorReason, numOfTraces int, indexes []int) {
```
joe-elliott, line 583: this function needs some cleanup and tests. going to make some stylistic suggestions.

modules/distributor/distributor.go
```diff
@@ -530,19 +561,54 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount int
...
+	// no errors
+	if len(pushResponse.ErrorsByTrace) == 0 {
+		for _, reqBatchIndex := range indexes {
+			if reqBatchIndex < numOfTraces {
```
joe-elliott, line 587: i prefer writing this like. it reduces nesting and makes it easier to read. 

```
<for loop> {
  if <thing is bad> {
    <log my error>
    continue
  }
  
  <do happy path things>
}
```

modules/distributor/distributor.go
```diff
@@ -530,19 +561,54 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount int
...
+				currentNumSuccess := numSuccessByTraceIndex[reqBatchIndex]
+				numSuccessByTraceIndex[reqBatchIndex] = currentNumSuccess + 1
```
joe-elliott, line 589: i think this can be simplified to:
```
numSuccessByTraceIndex[reqBatchIndex]++
```

modules/distributor/distributor.go
```diff
@@ -530,19 +561,54 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount int
...
+			} else {
+				level.Warn(d.logger).Log("msg", fmt.Sprintf("batch index %d out of bound for length %d", reqBatchIndex, numOfTraces))
+			}
+		}
 	} else {
-		overrides.RecordDiscardedSpans(spanCount, reasonInternalError, userID)
+		for ringIndex, pushError := range pushResponse.ErrorsByTrace {
```
joe-elliott, line 595: similar to above you can actually remove two levels of nesting here.

modules/distributor/distributor.go
```diff
@@ -530,19 +561,54 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount int
...
-func recordDiscaredSpans(err error, userID string, spanCount int) {
-	s := status.Convert(err)
-	if s == nil {
-		return
+func countDiscaredSpans(numSuccessByTraceIndex []int, lastErrorReasonByTraceIndex []tempopb.PushErrorReason, traces []*rebatchedTrace, repFactor int) (maxLiveDiscardedCount int, traceTooLargeDiscardedCount int) {
+	quorum := int(math.Floor(float64(repFactor)/2)) + 1 // min success required
+
+	for traceIndex, numSuccess := range numSuccessByTraceIndex {
+		// we will count anything that did not receive min success as discarded
+		if numSuccess < quorum {
```
joe-elliott, line 569: loop nesting can be reduced. see below.

joe-elliott (commented):
some stylistic comments. almost there!

modules/overrides/config.go
```diff
@@ -30,9 +30,9 @@ const (
 	GlobalIngestionRateStrategy = "global"
 
 	// ErrorPrefixLiveTracesExceeded is used to flag batches from the ingester that were rejected b/c they had too many traces
-	ErrorPrefixLiveTracesExceeded = "LIVE_TRACES_EXCEEDED:"
+	ErrorPrefixLiveTracesExceeded = "LIVE_TRACES_EXCEEDED"
```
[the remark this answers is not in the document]
ie-pham: the fmt.Print was different but will change to make everything match

joe-elliott (approved):
Nice addition!

adugar-conga:
In which release version of Grafana Tempo will this be addressed ?

adugar-conga:
@joe-elliott  @ie-pham In which release version of Grafana Tempo will this be addressed ?

joe-elliott:
This feature is merged into main and will be in the next release of Tempo. We attempt a 3 month cadence on OSS releases but make no guarantees.

Releases are listed here: https://github.com/grafana/tempo/releases/

PR #2579, opened 2023-06-21
Metrics summary time range

mdisibio:
**What this PR does**:
Adds optional start/end time range to the new /api/metrics/summary endpoint.  Also fixes a couple edge cases in the histogram calculations.

**Which issue(s) this PR fixes**:
Fixes n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #2582, opened 2023-06-22
Dynamic metrics group by multiple attributes

mdisibio:
**What this PR does**:
Updated /api/metrics/summary to support grouping by mulitple attributes.  The `groupBy` parameter can be a comma delimited list of attributes.  Response was changed to return a slice of key/value pairs, in the same order that they are specified in groupBy.  A nil is returned for any missing value. Originally it was just the values, but I thought it was better to include the key instead of putting the onus on the caller to know what to expect.

This is a draft because it should go after #2579 which also has changes in this area.

Fixes n.a

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/generator/processor/localblocks/processor.go
```diff
@@ -612,15 +612,25 @@ func filterBatch(batch *v1.ResourceSpans) *v1.ResourceSpans {
...
-	return &tempopb.TraceQLStatic{
-		Type:   int32(static.Type),
-		N:      int64(static.N),
-		F:      static.F,
-		S:      static.S,
-		B:      static.B,
-		D:      uint64(static.D),
-		Status: int32(static.Status),
-		Kind:   int32(static.Kind),
+func toStaticProto(series traceqlmetrics.MetricSeries) []*tempopb.KeyValue {
```
zalegrala, line 615: Consider renaming this function to match the change in behavior.

zalegrala (commented):
This looks like a good start to me.  Do you want to plumb through the frontend, or leave that for another PR?

mdisibio:
> This looks like a good start to me. Do you want to plumb through the frontend, or leave that for another PR?

Can you clarify what you mean?  It already works end-to-end by reusing the `groupBy` parameter to be a comma-delimited list.  The output proto/json is also updated to return the slice of group values.

zalegrala (approved):
Sorry, I meant to approve yesterday.  This looks good to me with one comment about whitespace on the query.

pkg/tempopb/tempo.proto
```diff
@@ -219,16 +219,21 @@ message RawHistogram {
...
 
+message KeyValue {
+  string key = 1;
+  TraceQLStatic value = 2;
+}
+
 message SpanMetrics {
   repeated RawHistogram latency_histogram = 1;
-  TraceQLStatic static = 2;
+  repeated KeyValue series = 2;
```
zalegrala, line 229: Maybe we also include a status code to indicate control messages like not-found?

zalegrala: Or possibly on the `KeyValue`.

mdisibio: When an attribute is not found it will return a traceql static with Nil type (empty value).  This works because real nils are not possible (against spec but also not persisted).

PR #2583, opened 2023-06-22
Implement tenant index deletion for tenants with zero blocks

### LINKED ISSUE #2520: Tempo leaves tenant indexes on tenants with no blocks
If a tenant's blocks are completely removed Tempo will still leave the `index.json.gz` in the root of the tenant's "folder". This tenant index correctly indicates there are no blocks for this tenant, but incurs a cost to Tempo.

During the polling phase Tempo components first list all tenants and then pull all tenant index's for all tenants. In a highly multitenant environment these "empty" tenant indexes can significantly impact Tempo.

Remove `index.json.gz` if there are no blocks.

zalegrala:
**What this PR does**:

Here we interrupt the tenant index write in the compactor when a tenant is found to have zero blocks, and zero compacted blocks.

Fixes #2520

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/backend/backend.go
```diff
@@ -64,4 +79,6 @@ type Compactor interface {
 	ClearBlock(blockID uuid.UUID, tenantID string) error
 	// CompactedBlockMeta returns the compacted blockmeta given a block and tenant id
 	CompactedBlockMeta(blockID uuid.UUID, tenantID string) (*CompactedBlockMeta, error)
+	// DeleteTenantIndex deletes the tenant index.
+	DeleteTenantIndex(ctx context.Context, tenantID string) error
```
mdisibio, line 83: Have an idea that could simplify things - what if this was done automatically inside `WriteTenantIndex`?  If there are no blocks then delete.

zalegrala:
I'll open a new PR for this.  If we don't extend the interface, I don't think we need anything here.

PR #2584, opened 2023-06-22
Run `gofumpt -w` on all go files and integrate with CI

zalegrala:
**What this PR does**:

In working in the `tempodb` package, I notice a few areas that could use a little formatting.  Here is a proposal to introduce a bit more opinionated formatting using [gofumpt](https://github.com/mvdan/gofumpt).  After review of the changes, this has been integrated into `make fmt` to apply to the entire codebase.

If we agree on these changes, we could introduce this in `tools/` and integrate as part of CI perhaps in a future PR.  All editors probably have an auto-formatter as well, but worth checking on your editor.  A make target could cover most of us though. 

Thoughts?

**Checklist**
- [x] Tests updated
- [x] Documentation added


mdisibio:
The changes look fairly unopinionated and ok to me, a little bit cleaner than the default.  Main concern is the tooling and preventing churn between contributors.  A make target would be great.

zalegrala:
Those last few commits will do the needful.  If this goes to `main`, `make fmt` will be required for all committers who don't have their editor configured to use `gofumpt` as their auto-format.

mdisibio:
Enabling it to format on save in VS Code was easy following the settings snippet in https://github.com/mvdan/gofumpt#visual-studio-code

However it looks like there isn't a way to include the `-extra` through gopls, it only has [gofumpt: bool](https://github.com/golang/tools/blob/master/gopls/doc/settings.md#gofumpt-bool). There is a discussion of the rationale in https://github.com/golang/go/issues/56403

zalegrala:
Thanks for testing @mdisibio.  I'm completely open to removing the `-extra` flag.

joe-elliott:
i'm good on this as long as everything can be automated on save in popular editors. 

we should include details about this tool and how to set it up in `./CONTRIBUTING.md`.

PR #2587, opened 2023-06-23
Ipv6 docs

### LINKED ISSUE #2586: Add note to IPv6 docs to use port 3100 for the HTTP server
**Describe the bug**
In the IPv6 setup instructions, the recommended configuration mentions that a port of 3200 should be used as the http_listen_port. When using the Tempo Helm chart to set-up Tempo, the ingesters and distributors are configured to use port 3100, and this is hard-coded. In the jsonnet, this port has been changed to 3200.

Setting the http_listen_port to 3200 in the config in the Helm chart leads to cryptic errors regarding unexpected node messages. It's only upon close reading and analysis of the Helm chart templates, ports fields, and Tempo configuration files that the port misalignment can be found and resolved. Mentioning this discrepancy in the docs may help future Helm chart users in setting up IPv6.

ajpauwels:
**What this PR does**:

Addresses a missing note in the IPv6 docs to use port 3100 as the http_listen_port when using the Helm chart to deploy Tempo.

Fixes #2586 

**Checklist**
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/configuration/ipv6.md
+++ b/docs/sources/tempo/configuration/ipv6.md
@@ -14,7 +14,7 @@ Tempo can be configured to communicate between the components using Internet Pro
 
 This sample listen configuration will allow the gRPC and HTTP servers to listen on IPv6, and configure the various memberlist components to enable IPv6.
 
-NOTE: If using the Helm chart, set `http_listen_port: 3100` as the pod ports are hard-coded to 3100.
+NOTE: If using the Helm chart, set `http_listen_port: 3100` as the pod ports are hard-coded to 3100. The Jsonnet files are configured to use 3200 by default.
 
 ```yaml
 memberlist:
--- a/docs/sources/tempo/configuration/ipv6.md
+++ b/docs/sources/tempo/configuration/ipv6.md
@@ -14,8 +14,6 @@ Tempo can be configured to communicate between the components using Internet Pro
 
 This sample listen configuration will allow the gRPC and HTTP servers to listen on IPv6, and configure the various memberlist components to enable IPv6.
 
-NOTE: If using the Helm chart, set `http_listen_port: 3100` as the pod ports are hard-coded to 3100. The Jsonnet files are configured to use 3200 by default.
-
 ```yaml
 memberlist:
   bind_addr:
@@ -34,14 +32,11 @@ metrics_generator:
 
 ingester:
   lifecycler:
-    address: '::'
     enable_inet6: true
 
 server:
   grpc_listen_address: '[::0]'
-  grpc_listen_port: 9095
   http_listen_address: '[::0]'
-  http_listen_port: 3200
 ```
 
 ## Kubernetes service configuration
--- a/docs/sources/tempo/configuration/ipv6.md
+++ b/docs/sources/tempo/configuration/ipv6.md
@@ -14,6 +14,8 @@ Tempo can be configured to communicate between the components using Internet Pro
 
 This sample listen configuration will allow the gRPC and HTTP servers to listen on IPv6, and configure the various memberlist components to enable IPv6.
 
+NOTE: If using the Helm chart, set `http_listen_port: 3100` as the pod ports are hard-coded to 3100.
+
 ```yaml
 memberlist:
   bind_addr:


docs/sources/tempo/configuration/ipv6.md
```diff
@@ -14,6 +14,8 @@ Tempo can be configured to communicate between the components using Internet Pro
 
 This sample listen configuration will allow the gRPC and HTTP servers to listen on IPv6, and configure the various memberlist components to enable IPv6.
 
+NOTE: If using the Helm chart, set `http_listen_port: 3100` as the pod ports are hard-coded to 3100. The Jsonnet files are configured to use 3200 by default.
```
zalegrala, line 17: Thanks for the PR.  I think we might even be able to drop `http_listen_port` and `grpc_listen_port` from the example below since they are defaults.  This would also cover the helm case I believe.  Would this have helped you in your discovery of this issue?  Being able to rely on the defaults would mean that helm would also use the configured default.  I'd be curious if you are able to test that.

ajpauwels: I just tested it and looks like it works fine, I'll update the docs!

I did find one other issue however: I needed to set the `memberlist.advertise_addr` setting to the pod's IP (this is in an EKS single-stack IPv6 cluster). I did this by assigning the pod's IP address to a POD_IP environment variable via the downward API, enabling the expand-env option, and then using `${POD_IP}` as the advertise_addr field value. Without doing this, the distributor and ingester both output logs that look like this:


```
ts=2023-06-24T17:43:46.556975173Z caller=memberlist_logger.go:74 level=info msg="Suspect grafana-tempo-ingester-1-ea9253ae has failed, no acks received"
ts=2023-06-24T17:43:46.557388484Z caller=memberlist_logger.go:74 level=warn msg="Got ping for unexpected node 'grafana-tempo-ingester-2-4832517a' from=[::6]:7946"
level=warn ts=2023-06-24T17:43:50.121095145Z caller=lifecycler.go:247 msg="found an existing instance(s) with a problem in the ring, this instance cannot become ready until this problem is resolved. The /ring ht
tp endpoint on the distributor (or single binary) provides visibility into the ring." ring=ingester err="instance :9095 in state PENDING"
level=warn ts=2023-06-24T17:44:00.120682325Z caller=lifecycler.go:247 msg="found an existing instance(s) with a problem in the ring, this instance cannot become ready until this problem is resolved. The /ring ht
tp endpoint on the distributor (or single binary) provides visibility into the ring." ring=ingester err="instance :9095 in state PENDING"
ts=2023-06-24T17:44:06.463116266Z caller=memberlist_logger.go:74 level=info msg="Marking grafana-tempo-querier-5bfc68d957-85sdr-05471e0f as failed, suspect timeout reached (0 peer confirmations)"
ts=2023-06-24T17:44:06.463176645Z caller=memberlist_logger.go:74 level=info msg="Marking grafana-tempo-compactor-cbd6f685b-rsf55-9de46201 as failed, suspect timeout reached (0 peer confirmations)"
ts=2023-06-24T17:44:06.551814042Z caller=memberlist_logger.go:74 level=info msg="Marking grafana-tempo-ingester-1-ea9253ae as failed, suspect timeout reached (0 peer confirmations)"
level=warn ts=2023-06-24T17:44:10.120604678Z caller=lifecycler.go:247 msg="found an existing instance(s) with a problem in the ring, this instance cannot become ready until this problem is resolved. The /ring ht
tp endpoint on the distributor (or single binary) provides visibility into the ring." ring=ingester err="instance :9095 in state PENDING"
ts=2023-06-24T17:44:16.56267006Z caller=memberlist_logger.go:74 level=info msg="Suspect grafana-tempo-ingester-2-4832517a has failed, no acks received"
level=warn ts=2023-06-24T17:44:20.143056966Z caller=lifecycler.go:247 msg="found an existing instance(s) with a problem in the ring, this instance cannot become ready until this problem is resolved. The /ring ht
tp endpoint on the distributor (or single binary) provides visibility into the ring." ring=ingester err="instance :9095 in state PENDING"
```

Should I add this to the docs as well?

Thank you for your review! I know this seems like a pretty small/insignificant PR but I believe it can save future IPv6 installers from a couple hours of debugging.

ajpauwels: I've also removed the `ingester.lifecycler.address` field as I found it would make querying fail

zalegrala: I'd say update the docs to include the $POD_IP usage.  Do we know why this is necessary?  Is it that the pod in these environment doesn't have the same address locally that is represented outward, or that `::` is not handled correctly?  Note that not all users are using kubernetes so just call that out as something specific where it makes sense.  Fixing docs for future users is always appreciated, so thanks for running through it and testing out the IPv6 support.  Let me know when its ready for another look.

knylander-grafana:
Just wanted to follow up and see if you needed any help with docs.

PR #2589, opened 2023-06-23
Make metrics-generator ingestion slack per tenant

ie-pham:
**What this PR does**: Make metrics-generator ingestion slack a per tenant config for when a tenant requires a much higher slack than the default '30s'

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [x] Documentation added


modules/generator/instance.go
```diff
@@ -108,6 +108,11 @@ func newInstance(cfg *Config, instanceID string, overrides metricsGeneratorOverr
 	}
 	go i.watchOverrides()
 
+	// add ingestion slack overrides
+	if slack := overrides.MetricsGeneratorIngestionSlack(instanceID); slack > 0 {
+		i.cfg.MetricsIngestionSlack = slack
```
joe-elliott, line 113: cfg is a pointer. is this modifying the ingestion slack for other tenants?

modules/generator/instance.go
```diff
@@ -108,6 +108,11 @@ func newInstance(cfg *Config, instanceID string, overrides metricsGeneratorOverr
 	}
 	go i.watchOverrides()
 
+	// add ingestion slack overrides
+	if slack := overrides.MetricsGeneratorIngestionSlack(instanceID); slack > 0 {
```
zalegrala, line 112: It might make sense to move this to the `updateProcessors` so that if it changes, we don't have to restart the instance for changes to take effect.

joe-elliott: good call. overrides should be dynamically loaded.

knylander-grafana (commented):
Doc changes look good! Thank you for adding doc.

modules/generator/instance.go
```diff
@@ -342,6 +351,11 @@ func (i *instance) preprocessSpans(req *tempopb.PushSpansRequest) {
 	size := 0
 	spanCount := 0
 	expiredSpanCount := 0
+	ingestionSlack := i.cfg.MetricsIngestionSlack
+	if overrideSlack, found := i.ingestionSlackOverrides[i.instanceID]; found {
```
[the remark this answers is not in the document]
kvrhdn: +1, it's fine to retrieve the override frequently. The overrides module is built with this assumption and caches overrides internally.

We would still need to check whether the per-tenant value is set and if not, use the default value `i.cfg.MetricsIngestionSlack`.

So something like:
```go
ingestionSlack := i.overrides.MetricsGeneratorIngestionSlack(i.instanceID) 
if ingestionSlack = 0 {
    ingestionSlack = i.cfg.MetricsIngestionSlack
}
```

modules/generator/instance.go
```diff
@@ -184,6 +185,11 @@ func (i *instance) updateProcessors() error {
 		return err
 	}
 
+	i.ingestionSlackOverride = i.overrides.MetricsGeneratorIngestionSlack(i.instanceID)
+	if i.ingestionSlackOverride == 0 {
```
joe-elliott, line 189: i hate that we do this everywhere. i wish the overrides system integrated the config and returned the correct result if it was unset.

nothing to fix in this PR tho :)

modules/generator/instance.go
```diff
@@ -184,6 +185,11 @@ func (i *instance) updateProcessors() error {
 		return err
 	}
 
+	i.ingestionSlackOverride = i.overrides.MetricsGeneratorIngestionSlack(i.instanceID)
```
joe-elliott, line 188: i think there's a race condition here with the main ingestion path

modules/generator/instance.go
```diff
@@ -189,6 +191,13 @@ func (i *instance) updateProcessors() error {
 		return err
 	}
 
+	ingestionSlackInt := i.overrides.MetricsGeneratorIngestionSlack(i.instanceID).Nanoseconds()
+	if ingestionSlackInt == 0 {
+		ingestionSlackInt = i.cfg.MetricsIngestionSlack.Nanoseconds()
+	}
+
+	atomic.StoreInt64(&i.ingestionSlackOverride, ingestionSlackInt)
```
joe-elliott, line 199: use an atomic package over the primitives. it helps prevent making mistakes later. we mostly use: https://github.com/uber-go/atomic

modules/generator/instance.go
```diff
@@ -347,16 +357,21 @@ func (i *instance) preprocessSpans(req *tempopb.PushSpansRequest) {
...
 	expiredSpanCount := 0
+
 	for _, b := range req.Batches {
 		size += b.Size()
 		for _, ss := range b.ScopeSpans {
 			spanCount += len(ss.Spans)
 			// filter spans that have end time > max_age and end time more than 5 days in the future
 			newSpansArr := make([]*v1.Span, len(ss.Spans))
 			timeNow := time.Now()
+			ingestionSlackNano := i.ingestionSlackOverride.Load()
```
joe-elliott, line 368: one last change, load this once at the top of this method instead of per span.

kvrhdn (approved):
LGTM! This will be handy, thanks!

Just a small nitpick, do with it what you will, I'm fine with merging it as-is.

PR #2592, opened 2023-06-26
[vParquet3] dedicated columns read path

### LINKED ISSUE #2576: TraceQL searches through dedicated columns
TraceQL queries return conditions containing resource and span attributes from dedicated columns.

stoewer:
**What this PR does**:
TraceQL searches through attributes stored in dedicated columns.

Fixes #2576

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/encoding/vparquet2/block_search.go
```diff
@@ -130,18 +131,31 @@ func (b *backendBlock) Search(ctx context.Context, req *tempopb.SearchRequest, o
...
-func makePipelineWithRowGroups(ctx context.Context, req *tempopb.SearchRequest, pf *parquet.File, rgs []parquet.RowGroup) pq.Iterator {
+func makePipelineWithRowGroups(ctx context.Context, req *tempopb.SearchRequest, pf *parquet.File, rgs []parquet.RowGroup, dc []backend.DedicatedColumn) pq.Iterator {
 	makeIter := makeIterFunc(ctx, rgs, pf)
 
 	// Wire up iterators
 	var resourceIters []pq.Iterator
 	var traceIters []pq.Iterator
 
+	// Dedicated column mappings
+	resourceColumnMapping := dedicatedColumnsToColumnMapping(dc, dedicatedColumnScopeResource)
```
mdisibio, line 142: It looks like these end up with the same logic so maybe we don't need to split them by scope.  Are you doing this so that one scope can be prioritized over the other?  If so maybe span-level should take precedence over resource-level.   Not really concerned as this is the old search functionality, but just mentioning it.

mdisibio (commented):
I'm very excited for this 🚀   Left a few questions and comments.

pkg/traceql/enum_statics.go
```diff
@@ -17,6 +17,31 @@ const (
 	TypeKind
 )
 
+func StaticTypeFromString(s string) StaticType {
+	switch s {
+	case "spanset":
```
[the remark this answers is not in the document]
stoewer: Good point. @mapno also [suggested earlier](https://github.com/grafana/tempo/pull/2517#discussion_r1212815182) to make separate types for the dedicated column scope and type. Therefore, I added the types and replaced `StaticTypeFromString()` with `DedicatedColumnType.ToStaticType()`.

tempodb/encoding/vparquet2/block_search.go
stoewer: >  Are you doing this so that one scope can be prioritized over the other?

No, this is just because in most other places require the mapping only for one scope. I changed `dedicatedColumnsToColumnMapping()` such that it can create mappings for multiple scopes.

mapno (approved):
Awesome, LGTM

Issue #2477, opened 2023-05-15
Support RBAC on remote calls from the querier to the serverless backend 


modulitos:
**Is your feature request related to a problem? Please describe.**

As an admin of a self-hosted Tempo installation, I want to use my cloud provider's HTTP api when sending requests from the querier to the serverless backend (aka `external_endpoint`), so that I can control access via my cloud provider's RBAC policies.

The current [`external_endpoints` configuration](https://grafana.com/docs/tempo/latest/configuration/#querier) (used for hosting a Tempo serverless backend) doesn't allow for sending auth info. We can see this in the code where the querier sends an http request to the backend, in the `Querier.searchExternalEndpoint` function:
https://github.com/grafana/tempo/blob/dcf8a2a5b0af920b47603220f7fa1d798d9b1089/modules/querier/querier.go#L632

**Describe alternatives you've considered**

The current workaround seems to rely on firewalls/VPCs/etc to allow the queriers to send requests to the serverless backend, while blocking all other traffic. This solution isn't ideal because it doesn't leverage the existing ecosystem of RBAC, service accounts, etc from a cloud provider. These firewalls can also be difficult to configure correctly, potentially exposing security vulnerabilities.

**Describe the solution you'd like**

I would like a solution that leverages the RBAC policies of the cloud providers when sending requests to the serverless backend. Queriers are already configured with a service account or role from the cloud provider - after all, this is how they are accessing the storage bucket.

Here's a quick summary for how we might implement this feature:
Instead of using `http.NewRequest(http.MethodGet, ...)` in  `Querier.searchExternalEndpoint` (linked above), I think we can delegate this call to the appropriate AWS/GCP/Azure client library.

We're already doing this when accessing the storage buckets:
https://github.com/grafana/tempo/blob/9ad992451a657a2a221dc8e526295bc5e347e62a/cmd/tempo-serverless/handler.go#L148-L159

I think we leverage a similar pattern when selecting the cloud provider when making an HTTP service request in `Querier.searchExternalEndpoint`.

As for the HTTP request itself, I believe AWS/GCP/Azure all have APIs for sending HTTP requests to their services. Here's an example where we are sending an HTTP request using Google's `NewTransport` api:
https://github.com/grafana/tempo/blob/dcf8a2a5b0af920b47603220f7fa1d798d9b1089/tempodb/backend/gcs/gcs.go#L266

[...]

PR #2593, opened 2023-06-26
Queriers: Implement RBAC on external queriers

modulitos:
**What this PR does**:
This PR adds RBAC support when querying the external endpoints. It adds an "external client" struct to encapsulate the concerns around sending remote requests to an external endpoint, which makes it easier to inject RBAC behavior into that struct.

It only supports RBAC on CloudRun, but adding Lambda support should be an easy followup. I also did a red/green smoke test using a CloudRun backend, and it's behaving as expected.

Resolves https://github.com/grafana/tempo/issues/2477

**Checklist**
- [x] Tests updated
- [x] Documentation added


CONTRIBUTING.md
```diff
@@ -60,9 +61,11 @@ vendor/
 ## Coding Standards
 
 ### go imports
+
```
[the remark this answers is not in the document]
joe-elliott: it's fine. nice update to this doc.

knylander-grafana (commented):
Thank you for adding some doc content. When we add more support, it would be great to add a configuration page with information about use cases, benefits, etc.

modules/querier/external/cloud_run.go
```diff
@@ -0,0 +1,19 @@
...
+	"fmt"
+
+	"golang.org/x/oauth2"
+	"google.golang.org/api/idtoken"
+)
+
+// Return Oauth2 token according to our google application credentials:
+// https://cloud.google.com/run/docs/authenticating/service-to-service#acquire-token
+func googleToken(ctx context.Context, endpoint string) (*oauth2.Token, error) {
+	ts, err := idtoken.NewTokenSource(ctx, endpoint)
```
joe-elliott, line 14: this will be called 100s or 1000s of times/second. what is the correct way to use `NewTokenSource`? should we cache it locally or is it meant to be invoked like this?

docs/sources/tempo/configuration/_index.md
```diff
@@ -489,6 +489,10 @@ querier:
         # The default value of "" disables this feature.
         [external_endpoints: <list of strings> | default = <empty list>]
 
+        # If external_backend is set, then authorization credentials will be provided when querying the external endpoints. "cloud-run" is the only value supported at this time.
+        # The default value of "" omits credentials when querying the external backend.
+        [external_backend: <string> | default = ""]
```
joe-elliott, line 494: reading the code below i feel like this should be "external_backend_auth"? 

or a slightly different direction:

i'm wondering if longterm we'd prefer a pattern like we do in the storage backend:

```
external_backend: <google-cloud-run|aws-lambda|http>
google-cloud-run:
  <gcr settings>
aws-lambda:
  <lambda settings>
...

File: CHANGELOG.md [hunk 4]
```diff
@@ -1,5 +1,6 @@
 ## main / unreleased
 
+* [ENHANCEMENT] Support RBAC for serverless backend queries [#2487](https://github.com/grafana/tempo/pull/2593) (@modulitos)
```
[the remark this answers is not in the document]
modulitos: Updated :+1:

modules/querier/external/cloud_run.go
modulitos: Great catch! Yes, we should cache it. I pushed up an implementation which leverages a cached `oauth2.TokenSource` per endpoint.

docs/sources/tempo/configuration/_index.md
modulitos: Sounds good! I pushed up a change that restructures the config with an eye toward these longer-term considerations. I'm hoping the new structure will avoid breaking changes in the future, while providing enough flexibility to define the cloud-run/lambda/http specific as we discover them. For now, I'm holding off on making any breaking changes in the existing config as well.

modules/querier/external/token_provider.go
```diff
@@ -0,0 +1,73 @@
...
+			mtx.Lock()
+			defer mtx.Unlock()
+
+			if err != nil {
+				tsErr = multierr.Combine(tsErr, err)
+			}
+			sources[ep] = ts
+		}(endpoint)
+	}
+	wg.Wait()
```
modulitos, line 45: Note that this code is only invoked when the app starts up initially, so performance isn't a huge concern, so I'm happy to remove the goroutine + WaitGroup if we want to simplify. But I figured concurrent execution was worthwhile since `getTokenSource` can be a blocking call.

For example, when the `idtoken.NewTokenSource` function (used to auth into Google's api) is passed as `getTokenSource` (as done in the `newGoogleProvider` function below), it'll [perform 2 legged auth to retrieve the token](https://github.com/googleapis/google-api-go-client/blob/ed3a635a8d60c849d27763cc619105088abb1e06/idtoken/idtoken.go#L137-L138) as well as [read from a credentials file on disk](https://github.com/googleapis/google-api-go-client/blob/ed3a635a8d60c849d27763cc619105088abb1e06/idtoken/idtoken.go#L102).

docs/sources/tempo/configuration/_index.md
modulitos: Just to clarify, the SearchConfig looks like this now:

```
// These are the existing configs:
external_endpoints: []string
HedgeRequestsAt: int
HedgeRequestsUpTo: int
// ^ moving these under their own "http" external_backend block is a followup,
// requiring a breaking change

external_backend: <google_cloud_run> // "aws-lambda" and "http" will be a followup
google_cloud_run:
  external_endpoints: []string // eg: ["https://endpoint1.run.app", "https://endpoint2.run.app" ]
  HedgeRequestsAt: int
  HedgeRequestsUpTo: int
```

There's some repetition in the `external_endpoints` and `HedgeRequests*` configs for each external backend type, but I figured it made more sense to make that declarative instead of trying to re-use it across multiple backends. This declarative approach is also consistent with the storage backend config [`/tempodb/config.go`](https://github.com/grafana/tempo/blob/d6544e80c076a3b96a0c991bc329252c926ffede/tempodb/config.go#L52-L57).

Also, I kept `external_endpoints` as a list under the `google_cloud_run` config. I'm not totally sure why it's a list to begin with - perhaps we want to load balance across multiple serverless deployments or regions? I figured that we'll want to retain the list, but I'm calling that out here just to be sure.

joe-elliott: I like the direction this is heading. See some more detailed comments below.

modules/querier/external/client.go
```diff
@@ -0,0 +1,197 @@
...
+			endpoints:         cfg.HTTPConfig.Endpoints,
+			hedgeRequestsAt:   cfg.HTTPConfig.HedgeRequestsAt,
+			hedgeRequestsUpTo: cfg.HTTPConfig.HedgeRequestsUpTo,
+		})
+	case "google_cloud_run":
+		provider, err := newGoogleProvider(ctx, cfg.CloudRunConfig.Endpoints)
+		if err != nil {
+			return nil, err
+		}
+		return newClientWithOpts(&commonConfig{
```
joe-elliott, line 98: i expect this to eventually drift into two different clients, but i think this is good for now.

modules/querier/external/token_provider.go
joe-elliott: no, i think this is a good call. this kind of structure is common and does not require much mental overhead to parse.

modules/querier/external/token_provider.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil, fmt.Errorf("failed to create one or more token sources: %w", tsErr)
+	}
+
+	return &cachedTokenProvider{
+		tokenSources: sources,
+	}, nil
+}
+
+func (t *cachedTokenProvider) getToken(ctx context.Context, endpoint string) (*oauth2.Token, error) {
+	if src, containsKey := t.tokenSources[endpoint]; containsKey {
```
joe-elliott, line 56: does this ever expire?

modules/querier/external/client.go
```diff
@@ -0,0 +1,197 @@
...
+
+type Client struct {
+	httpClient    *http.Client
+	endpoints     []string
+	tokenProvider tokenProvider
+}
+
+type CloudRunConfig struct {
+	Endpoints         []string      `yaml:"external_endpoints"`
+	HedgeRequestsAt   time.Duration `yaml:"external_hedge_requests_at"`
```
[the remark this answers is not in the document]
modulitos: Sounds good - I moved the request hedging settings up to the parent 👍

I think "endpoints" is still a reasonable name, since we're still invoking Cloud Run over HTTP. So the semantics of "endpoints" here lines up with HTTP. The term "endpoints" is also used in [the Cloud Run docs](https://cloud.google.com/run/):
> Each Cloud Run service gets an out-of-the-box stable HTTPS endpoint

Re adding more settings: there are no additional settings added here, since we're using the existing HTTP interface. But having a dedicated Cloud Run config will come in handy eventually. For example, it'll give us a place to configure features like [invoking Cloud Run over gRPC](https://cloud.google.com/run/docs/triggering/grpc). I think supporting gRPC will require some changes on the serverless endpoint, but taking some benchmarks and implementing it could be a nice followup feature!

modules/querier/external/token_provider.go
modulitos: Nope, I don't think it expires.

TokenSource handles expired tokens under the hood, as described here: https://pkg.go.dev/golang.org/x/oauth2#ReuseTokenSource
but the TokenSource itself shouldn't expire.

I suppose there's a case where the creds used to create the TokenSource expire (ie: the `GOOGLE_APPLICATION_CREDENTIALS` env var), but at that point, I think you'll need to update the creds on the server and restart it.

modules/querier/external/token_provider.go
```diff
@@ -0,0 +1,73 @@
...
+// Caches an oauth2.TokenSource to enable efficient auth on each of our external
+// endpoints.
+type cachedTokenProvider struct {
+	tokenSources map[string]oauth2.TokenSource
+}
+
+func newTokenProvider(
+	ctx context.Context,
+	endpoints []string,
+	getTokenSource func(ctx context.Context, endpoint string) (oauth2.TokenSource, error),
```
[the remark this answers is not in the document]
modulitos: Yeah, I can go either way, but I went ahead and added this approach to see. There's a standalone commit that implements this change, if that's easier to review.

On one hand, this approach normalizes our various token providers under the same interface. From an OO perspective, I can see how it's easier to reason about.

On the other hand, we now have an extra `getTokenSource` function on the interface, which is required when wrapping the provider with caching. So we have to define the function on the cachedTokenProvider to satisfy the interface, even though it's not meant to be called.

I think either approach is reasonable, and it's fun thinking through these tradeoffs. But lmk if I missed something in your suggestion :)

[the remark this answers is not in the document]
modulitos: I thought more about this, and decided to revert my approach described above because I think the original is more maintainable. You can still review the change under the commit titled `implement same interface for all token providers` - perhaps I missed something in your suggestion that made the interface feel bloated. If you still think we should use the same interface, I'm always happy to chat and/or address it as a follow up. I'll continue thinking about this as well :thinking:

joe-elliott (approved):
great work! appreciate this contribution. adds a nice authentication option for GCR and starts laying the groundwork for additional configuration per backend.

PR #2598, opened 2023-06-28
Enable Cross Cluster Querying

### LINKED ISSUE #2594: Support multiple ingester rings in the Querier
Tempo should support multiple ingester rings in the Querier. This will allow seamless migrations of clusters and other exotic architectures.

When configured in this way. One set of queriers would be able to query ingesters from two different clusters.

joe-elliott:
**What this PR does**:
Adds two configuration options that allows queriers in two independent clusters to query each other's ingesters. In particular this is useful for two deployment patterns:

- active/active clusters where two different clusters write to the same bucket and you need the queriers in each cluster to see each other's ingesters
- seamless cluster migration. a second cluster can be brought up which can read the recent traces from the old cluster

These options are very niche and an operator would have to understand exactly what they are doing to use these successfully. As such I have decided not to document them, but I'm open to being persuaded otherwise.

Fixes #2594 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


example/docker-compose/cross-cluster/prometheus.yaml
```diff
@@ -0,0 +1,19 @@
...
+    static_configs:
+      - targets:
+        - 'compactor:3200'
+        - 'distributor:3200'
+        - 'ingester-0:3200'
+        - 'ingester-1:3200'
+        - 'ingester-2:3200'
+        - 'querier:3200'
+        - 'query-frontend:3200'
+        - 'metrics-generator:3200'
```
mapno, line 19: This needs to be split into `a` and `b` clusters

mapno (approved):
LGTM

PR #2603, opened 2023-06-30
[vParquet3] pass dedicated columns to querier

### LINKED ISSUE #2576: TraceQL searches through dedicated columns

stoewer:
**What this PR does**:

Add dedicated columns to `SearchBlockRequest` in order to make them available in the Querier. There they are needed in order to search through dedicated attribute columns via TraceQL.

Fixes #2576

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/tempopb/tempo.proto
```diff
@@ -80,6 +80,14 @@ message SearchBlockRequest {
...
   uint64 size = 10; // total size of data file
   uint32 footerSize = 11; // size of file footer (parquet)
+  repeated DedicatedColumn dedicatedColumns = 12;
+}
+
+// Configuration for a single dedicated attribute column.
+message DedicatedColumn {
+  string type = 1;
+  string name = 2;
+  string scope = 3;
```
mapno, line 90: Will we want enums for `type` and `scope` eventually?

stoewer: So far we are using string values to represent `type` and `scope`. To me this makes sense as these are exposed to the user in the configuration and overrides.
AFAIK protocol buffers does not support enums with string values. This means we would have to convert back and forth. Not sure if this is worth the effort. What do you think?

stoewer: I added enums for type and scope to tempo.proto :+1:

mapno (approved):
LGTM

PR #2612, opened 2023-07-05
FIX1958: Error code 400 when input trace is too large(WIP)

### LINKED ISSUE #1958: Fix distributor http status codes

mghildiy:
**Return status code 400 when input trace is too large**:

**Currently, status code returned for traces larger than permissible size is 500, which makes client to retry. To avoid this, return status code 400 instead.**:
Fixes #1958

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/ingester/instance.go
+++ b/modules/ingester/instance.go
@@ -186,7 +186,7 @@ func (i *instance) push(ctx context.Context, id, traceBytes []byte) error {
 		prevSize := int(i.traceSizes[tkn])
 		reqSize := len(traceBytes)
 		if prevSize+reqSize > maxBytes {
-			return status.Errorf(codes.FailedPrecondition, (newTraceTooLargeError(id, i.instanceID, maxBytes, reqSize).Error()))
+			return status.Errorf(codes.InvalidArgument, (newTraceTooLargeError(id, i.instanceID, maxBytes, reqSize).Error()))
 		}
 	}
 
@@ -195,7 +195,7 @@ func (i *instance) push(ctx context.Context, id, traceBytes []byte) error {
 	err := trace.Push(ctx, i.instanceID, traceBytes)
 	if err != nil {
 		if e, ok := err.(*traceTooLargeError); ok {
-			return status.Errorf(codes.FailedPrecondition, e.Error())
+			return status.Errorf(codes.InvalidArgument, e.Error())
 		}
 		return err
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## main / unreleased
 
+* [BUGFIX] Return status code 400 when input trace is too large error [#2612](https://github.com/grafana/tempo/pull/2612)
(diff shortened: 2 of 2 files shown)


modules/ingester/instance.go
```diff
@@ -186,7 +186,7 @@ func (i *instance) push(ctx context.Context, id, traceBytes []byte) error {
 		prevSize := int(i.traceSizes[tkn])
 		reqSize := len(traceBytes)
 		if prevSize+reqSize > maxBytes {
-			return status.Errorf(codes.FailedPrecondition, (newTraceTooLargeError(id, i.instanceID, maxBytes, reqSize).Error()))
+			return status.Errorf(codes.InvalidArgument, (newTraceTooLargeError(id, i.instanceID, maxBytes, reqSize).Error()))
```
joe-elliott, line 189: have you tested this. does this really cause the distributor to return a 400?

mghildiy: This is WIP.

mghildiy: How do I test it? I have created a small spring boot app, and configured it with otel to send traces to tempo instance. 
I face 2 issues:

1. I am not able to configure max_bytes_per_trace for overrides. When I do it, tempo doesnt start, though its a valid config parameter as per official documentation.

```
overrides:
  per_tenant_override_config: /etc/overrides.yaml
```

In overrides.yaml, I have:
```
overrides:
  max_bytes_per_trace: 500
```

2. As a workaround, I hardcode  this config in instance.go. But this stops even other requests from being accepted(I see some requests with request bytes size around 10064 at repeated intervals. I think they are compaction related activity)

mghildiy: In collector logs I see:

> 023-07-12 19:05:03 2023-07-12T13:35:03.418Zerrorexporterhelper/queued_retry.go:391Exporting failed. The error is not retryable. Dropping data.{"kind": "exporter", "data_type": "traces", "name": "otlp", "error": "Permanent error: rpc error: code = InvalidArgument desc = TRACE_TOO_LARGE: max size of trace (500) exceeded while adding 5024 bytes to trace 64aeac0154073ef25bea20d464cc57d3 for tenant single-tenant", "dropped_items": 4}

Seems good to me.

mghildiy: Is there a way to turn off compaction?

joe-elliott: Perhaps try something like:

```
compactor:
  compaction:
    max_block_bytes: 1
```

mghildiy: In collector logs, I see error message containing 'The error is not retryable. Dropping data' no matter which code I use in ingester,  FailedPrecondition OR InvalidArgument.

PR #2625, opened 2023-07-07
[TraceQL] Implement structural operators

mdisibio:
**What this PR does**:
Implements descendant (>>), child (>), and sibling (~) structural operators in vParquet2.   The TraceQL engine now passes new intrinsics to storage so it can load the minimum columns (each operator uses a different set of columns).  The `Span` interface was updated with generic methods `DescendantOf(span)` that the engine calls, so it is agnostic to how this data is represented in the backend.  I.e. we are not exposing the raw left/right/parent integers to the engine.  There are pros and cons to this approach, it is more flexible and as we implement TraceQL on other formats (live traces?), I think it will be better.  The drawback is it means a very naive algorithm in the engine to detect the relationships O(n * m).  If the engine had the integers it could do fancy things like put them in a map or sorted slice.  I think for most cases n and m will be small.

Note:  My formatter is now gofumpt, so there is a little bit of noise from the fumpting.

Fixes ??  <-- If we have an issue to track structural operators I couldn't find it.

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/traceql/ast_execute.go
```diff
@@ -99,6 +98,72 @@ func (o SpansetOperation) evaluate(input []*Spanset) (output []*Spanset, err err
...
 			}
 
+		case OpSpansetDescendant:
+			if len(lhs) > 0 && len(rhs) > 0 {
+				// Clone here to capture previously computed aggregates, grouped attrs, etc.
+				matchingSpanset := input[i].clone()
+				matchingSpanset.Spans = nil
+				// TODO: In what situations do lhs and rhs have more than one spanset,
+				// and what should the result be?
+				for _, l := range lhs[0].Spans {
```
joe-elliott, line 108: should we fail in the event of multiple spansets? i can see the same for the other operators.

"spanset operators are not supported for multiple spansets per trace. consider using coalesce()"

?

pkg/traceql/enum_attributes.go
```diff
@@ -61,6 +61,10 @@ const (
 	// not yet implemented in traceql but will be
 	IntrinsicParent
 
+	IntrinsicStructuralDescendant
```
joe-elliott, line 64: perhaps a note about the unique nature of these?

tempodb/encoding/vparquet2/block_traceql.go
```diff
@@ -66,6 +94,9 @@ func (s *span) attributesMatched() int {
 	if len(s.id) > 0 {
 		count++
 	}
+	if s.nestedSetLeft > 0 || s.nestedSetRight > 0 || s.nestedSetParent > 0 {
+		count++
```
joe-elliott, line 98: should these be evaluated independently and each increment count separately? since they come in as different conditions?

tempodb/encoding/vparquet2/block_traceql.go
```diff
@@ -184,12 +221,13 @@ var intrinsicColumnLookups = map[traceql.Intrinsic]struct {
...
-	traceql.IntrinsicKind:          {intrinsicScopeSpan, traceql.TypeKind, columnPathSpanKind},
-	traceql.IntrinsicSpanID:        {intrinsicScopeSpan, traceql.TypeString, columnPathSpanID},
-	traceql.IntrinsicSpanStartTime: {intrinsicScopeSpan, traceql.TypeString, columnPathSpanStartTime},
+	traceql.IntrinsicName:                 {intrinsicScopeSpan, traceql.TypeString, columnPathSpanName},
+	traceql.IntrinsicStatus:               {intrinsicScopeSpan, traceql.TypeStatus, columnPathSpanStatusCode},
+	traceql.IntrinsicDuration:             {intrinsicScopeSpan, traceql.TypeDuration, columnPathDurationNanos},
+	traceql.IntrinsicKind:                 {intrinsicScopeSpan, traceql.TypeKind, columnPathSpanKind},
+	traceql.IntrinsicSpanID:               {intrinsicScopeSpan, traceql.TypeString, columnPathSpanID},
+	traceql.IntrinsicSpanStartTime:        {intrinsicScopeSpan, traceql.TypeString, columnPathSpanStartTime},
+	traceql.IntrinsicStructuralDescendant: {intrinsicScopeSpan, traceql.TypeNil, ""},
```
joe-elliott, line 230: why just one added here?

tempodb/encoding/vparquet2/block_traceql.go
joe-elliott: oh, reading below. it's kind of impossible to exactly count the number of attributes. just incrementing once is playing it safe?

pkg/traceql/storage.go
```diff
@@ -84,6 +84,10 @@ type Span interface {
 	ID() []byte
 	StartTimeUnixNanos() uint64
 	DurationNanos() uint64
+
+	SiblingOf(Span) bool
```
joe-elliott, line 88: is there a way to work this into being returned through the existing `Attributes()` func? maybe 
```
span.Attributes()[NewIntrinsic(IntrinsicStructuralDescendant)]
```
returns a func that can be executed to get the desired results? i find these a little awkward but i can think of no good alternatives. i'm afraid this pattern will grow and i'm trying to think of a way to jam it into the existing attributes. wdyt?

- returning left/right/parent binds the engine to the current implementation
- returning slices of valid siblings/descendants/children is far more costly then what you've done

joe-elliott (commented):
looks great! obvious asks: changelog and docs for the new traceql features

pkg/traceql/ast_execute.go
mdisibio: Agree.  When we see a query and use case that requires this, then hopefully we can also make a determination on what the output should be algorithmically.

tempodb/encoding/vparquet2/block_traceql.go
mdisibio: Yeah each operation uses a different subset. We could know the case for a single operator, but it's not possible to distinguish between child (which loads left + parent), and parent (which loads only parent) for example.  However it didn't seem necessary to increment exactly, just by 1 is enough to ensure the span passes through the filter logic in spanCollector.

mdisibio: Want to add more detail on why it works to only increment by 1.  It's because of this line:  https://github.com/grafana/tempo/blob/main/pkg/traceql/ast.go#L238   Any spanset operation unchecks the AllConditions optimization, which means a span only has to meet 1 condition to be passed up to the engine.  The exact count is only needed when the optimization is present (only spans meeting all are passed up to the engine).

tempodb/encoding/vparquet2/block_traceql.go
mdisibio: Thanks, good catch. It actually still worked because these are just hints for default scope, but still want them. Fixed.

pkg/traceql/storage.go
mdisibio: What do you see are the pros of using Attributes()?  I agree that this interface is feeling unfinished and the split between dedicated fields (like ID()) and dynamic Attributes() isn't well-defined.  Maybe one idea is that anything addressable by a native TraceQL enum/attr (i.e. IntrinsicName) could go in Attributes(), and everything else has a dedicated field.  Like Span.ID() could be eliminated and referenced by `span.Attributes()[NewIntrinsic(IntrinsicSpanID)]` now.

To bring it back to the original question:  if we exposed the parent/left/right values to the engine and bound the implementation, we'd switch to real IntrinsicLeft/Right/Parent and put them in Attributes.  However I really think the current interface is a better abstraction but we won't know it until we implement TraceQL on another format (I really want to do things like TraceQL tail-sampling, compaction rules, etc).

joe-elliott: > What do you see are the pros of using Attributes()?

Mainly the consistency of all data requested from the Fetch layer coming through the same path. Also simplicity of the interface is nice. It's a good point that intrinsics like span id and duration are coming through predefined functions. That did not occur to me while making the comment.

> To bring it back to the original question: if we exposed the parent/left/right values to the engine and bound the implementation, we'd switch to real IntrinsicLeft/Right/Parent and put them in Attributes. However I really think the current interface is a better abstraction but we won't know it until we implement TraceQL on another format (I really want to do things like TraceQL tail-sampling, compaction rules, etc).

Agree the current implementation is superior to adding left/right/parent or all valid relationship data in slices. This is definitely not a blocker. I just wanted to muse on it some since it is expanding the interface.

joe-elliott (approved):

PR #2631, opened 2023-07-10
Frontend Queues: Reduces Lock contention

joe-elliott:
**What this PR does**:
Reduces lock contention, simplifies code, and adds tests/benchmarks for the frontend request queue. 

Offering this PR up for discussion. I like a lot of things about it but it comes with two cons:

- Removes "max queriers per tenant" functionality. This is unused in Tempo but perhaps something we want eventually.
- Despite impressive local benchmarks it was difficult to show definitive performance improvements in a real environment.

**Benchmarks against main**:
```
name                     old time/op    new time/op    delta
GetNextForQuerier100-8      182ms ± 5%      55ms ± 4%  -69.64%  (p=0.008 n=5+5)
GetNextForQuerier1000-8     796ms ±27%      58ms ± 3%  -92.67%  (p=0.008 n=5+5)
GetNextForQuerier5000-8     10.8s ±35%      0.1s ± 7%  -99.37%  (p=0.008 n=5+5)

name                     old alloc/op   new alloc/op   delta
GetNextForQuerier100-8      702MB ±12%      63MB ±11%  -90.97%  (p=0.008 n=5+5)
GetNextForQuerier1000-8    1.68GB ±22%    0.05GB ±20%  -96.87%  (p=0.008 n=5+5)
GetNextForQuerier5000-8    4.00GB ±34%    0.03GB ±36%  -99.26%  (p=0.008 n=5+5)

name                     old allocs/op  new allocs/op  delta
GetNextForQuerier100-8       102k ± 0%      100k ± 0%   -1.67%  (p=0.008 n=5+5)
GetNextForQuerier1000-8      104k ± 1%      100k ± 0%   -4.16%  (p=0.008 n=5+5)
GetNextForQuerier5000-8      114k ± 4%      100k ± 0%  -12.17%  (p=0.008 n=5+5)
```

**Which issue(s) this PR fixes**:
Addresses the first part of #2464 

**Checklist**
- [x] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #2632, opened 2023-07-10
Queue cleanup and refactoring

joe-elliott:
**What this PR does**:
This is a much tamer version of https://github.com/grafana/tempo/pull/2631. It is not nearly aggressive in terms of refactoring this code, but it does include some small cleanup items as well as a deadlock fix: https://github.com/grafana/mimir/pull/667. (#2631 fixes this as well)

The main benefit of this PR over #2631 is that it retains the per tenant max querier functionality.


modules/frontend/queue/queue.go
```diff
@@ -140,19 +140,17 @@ FindQueue:
 		}
 
 		// Pick next request from the queue.
-		for {
```
mapno, line 143: The `for` in line 135 could also be removed, and the block here encapsulated with `if queue != nil{}`. That'll get rid of the linter warning.

mapno (commented):
I can't say I fully understand `contextCond`, but I trust it's been used in mimir for a long time.

Only left a lint comment.

modules/frontend/queue/queue.go
```diff
@@ -179,7 +173,7 @@ func (q *RequestQueue) stopping(_ error) error {
 	defer q.mtx.Unlock()
 
 	for q.queues.len() > 0 && q.connectedQuerierWorkers.Load() > 0 {
-		q.cond.Wait()
+		q.cond.Wait(context.Background())
```
zalegrala, line 176: We have a context to pass.  Is there a reason not to pass it?

PR #2634, opened 2023-07-11
Update `golangci-lint` and fix `revive` errors

mapno:
**What this PR does**:

This PR updates `golangci-lint` to v1.53.3 and fixes all the errors from the `revive` linter. The failing rules were `redefines-builtin`, `superfluous-else`, and `unused-parameter`.

### DIFF
--- a/modules/generator/processor/localblocks/processor.go
+++ b/modules/generator/processor/localblocks/processor.go
@@ -82,7 +82,7 @@ func (*Processor) Name() string {
 	return "LocalBlocksProcessor"
 }
 
-func (p *Processor) PushSpans(ctx context.Context, req *tempopb.PushSpansRequest) {
+func (p *Processor) PushSpans(_ context.Context, req *tempopb.PushSpansRequest) {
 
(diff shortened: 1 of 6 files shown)


joe-elliott (approved):
we apparently love naming arguments we don't use. thx for the cleanup!

tempodb/encoding/vparquet/block_iterator.go
```diff
@@ -28,7 +28,7 @@ func (b *backendBlock) open(ctx context.Context) (*parquet.File, *parquet.Reader
 	return pf, r, nil
 }
 
-func (b *backendBlock) RawIterator(ctx context.Context, pool *rowPool) (*rawIterator, error) {
+func (b *backendBlock) rawIter(ctx context.Context, pool *rowPool) (*rawIterator, error) {
```
joe-elliott, line 31: why did we change this?

mapno: It was complaining that an exported function was returning and unexported type. And we don't need the function to be exported.

PR #2647, opened 2023-07-12
[Jsonnet] Make metrics-generator deployment volume mounts customizable

mdisibio:
**What this PR does**:
The microservices jsonnet was using `container.withVolumeMounts` on the generator deployment, and this prevents downstream users of the jsonnet from customizing the mounts through the `$.tempo_metrics_generator_container`.  Except they still worked for the statefulset because it doesn't do that, which is unexpected and hard to troubleshoot.

The root cause was that we were attempting to keep the volume mounts on the deployment set to `/var/tempo/wal` but let the statefulset mount at `/var/tempo` (because of the new metrics feature that uses `/var/tempo/blocks`).  So the fix here is both removing `withVolumeMounts` and making the mounts the same between both deployment modes. This addresses the downstream customize issue, but also simplifies.

Fixes n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


zalegrala (approved):
Thanks for putting this together.  Looks good.

PR #2652, opened 2023-07-13
Implement polling improvements

### LINKED ISSUE #2521: Improve Polling Performance at high block counts
As Tempo approaches 1M+ blocks the TCO and performance of polling becomes negatively impactful. Polling operates using the following steps.

Queriers and query-frontends:
1) List all tenants
2) Get `<tenant>/index.json.gz` for each tenant

Compactors:
1) List all tenants
2) If I am not the tenant index builder for this tenant
    - Get `<tenant>/index.json.gz` for the tenant
3) If I am the tenant index builder for this tenant
    - List all blocks beneath the tenant: `<tenant>/*`
    - For each block get `<tenant>/<guid>/meta.json`. If this 404s get `<tenant>/<guid>/meta.compacted.json`
    - Use the above information to build a new tenant index
    - Put `<tenant>/index.json.gz`

The cost in time and api calls of getting all of the meta.json's and meta.compacted.json's adds up considerably. Given that meta.json and meta.compacted.json are immutable we don't need to GET them individually and we need to find a way to reduce these calls.

zalegrala:
**What this PR does**:

Currently, each polling cycle, the block list is read from the backend, followed by a read of each meta.json file.  If the meta.json returns not-found, then an additional call is made to the backend to fetch the meta.compacted.json.  During compaction, the meta.json is replaced with meta.compacted.json, but the details about the block remain unchanged.  This means that tempo performs lots of fetching for information that is already known from the previous poll. This can be a significant source of backend calls, and in turn, latency of the polling cycle.  In environments with a high number of blocks, this problem is amplified.

Here we extend the backend interfaces to create a ListBlocks() method that can be used to target block information direclty from each backend.

* list the backend, splitting on an empty delimter to get all objects in the backend
* parse the response, sorting the results into compacted and not-compacted
* for each block, use the previous blocklist to match data we have in memory
* for each new block, we perform the fetch as normal to read the metadata

This results in fewer calls to the backend, and a may decrease in total time spent for each poll cycle.

Additional changes to the structure of the backend are made to improve the duration while keeping the total calls in check.

* the s3 and gcs ListBlocks() call has implemented a parallel list over the search space

Fixes #2521

**Checklist**
- [x] Tests updated
- [x] Documentation added


zalegrala:
Blocklist poll duration decreases with an image built from this PR in GCS.
image

zalegrala:
Blocklist poll duration increases in s3.
image

The list call itslef is now taking longer in s3, but likely still saves cost on requests.  Perhaps there is a different approach to the s3 implementation we can take.  I'll do a little research and see if I can come up with something, but feel free to chime in if you have ideas.

joe-elliott:
As long as the polling takes less time than the polling cycle I think we're ok. Double the polling time might not matter if it's going from 20s -> 40s on 200k blocks or something like that.

zalegrala:
I've made a small change to use `n` as a delimiter for both the `s3` and the `gcs` case.  This works because `.json` files end with `n`.  Use against the actual GCS API was working fine, but against the fake-gcs-server that we use in our CI was not accepting multi-character delimiters.  I applied the same change to s3 thinking that perhaps the multi-character delimiter was slowing down the responses, but this didn't seem to pan out when I deployed.  I think its worth a review at this point.

tempodb/blocklist/poller.go
```diff
@@ -360,3 +485,19 @@ func sumTotalBackendMetaMetrics(blockMeta []*backend.BlockMeta, compactedBlockMe
 		compactedBlockMetaTotalBytes:   sumTotalBytesCBM,
 	}
 }
+
+func uuidInBlocklist(tenantID string, id uuid.UUID, list backend.Blocklist) bool {
+	for _, i := range list.Metas(tenantID) {
```
joe-elliott, line 490: this is very inefficient. we should request the list.Metas and list.CompactedMetas once and reuse them when checking for what's already in the blocklist

tempodb/backend/azure/azure.go
```diff
@@ -175,6 +176,12 @@ func (rw *readerWriter) List(ctx context.Context, keypath backend.KeyPath) ([]st
 	return objects, nil
 }
 
+// ListBlocks implements backend.Reader
+func (rw *readerWriter) ListBlocks(_ context.Context, _ backend.KeyPath) ([]uuid.UUID, []uuid.UUID, error) {
+	// TODO: implement
+	return nil, nil, fmt.Errorf("not implemented")
```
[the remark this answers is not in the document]
zalegrala: Ah yes, I didn't search for an exiting error and can use one of them.  I agree about `pkg/errors`, though do you mean  as part of this PR or a follow up?

tempodb/backend/raw.go
```diff
@@ -40,6 +40,8 @@ type RawWriter interface {
 type RawReader interface {
 	// List returns all objects one level beneath the provided keypath
 	List(ctx context.Context, keypath KeyPath) ([]string, error)
+	// ListBlocks returns the IDs for blocks and compacted blocks at the keypath
```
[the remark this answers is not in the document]
zalegrala: I took this approach because I wanted to avoid touching the existing implementations.  `List()` is a pretty generic way to get the contents of a path, but what we're aiming at here is a special method of getting the blocks we know about, which feels like it deserves its own method.  We could extend `List()` perhaps to include an additional argument, but then this would enable the specific features, and so it feels like we're back to the `RawReader` having special handling for special cases.

Since we're dealing with the specifics around implementation, I'm not sure what that would look like, but I'm all ears if you have thoughts on how to unify them.

tempodb/backend/raw.go
```diff
@@ -175,6 +194,12 @@ func (r *reader) Blocks(ctx context.Context, tenantID string) ([]uuid.UUID, erro
 	return blockIDs, nil
 }
 
+// Blocks implements backend.Reader
+func (r *reader) QuickBlocks(ctx context.Context, tenantID string) ([]uuid.UUID, []uuid.UUID, error) {
```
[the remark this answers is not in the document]
zalegrala: What we get with `QuickBlocks` is both the compacted uuids and the non-compacted uuidss in one call, which felt like a win.  The existing `Blocks()` doesn't differentiate between compacted and not, and so we have to do extra work here to determine which is which.  Also, I left the existing `Blocks()` intact so we can fall back to the old code if we get an error on the Quick try.

tempodb/blocklist/poller.go
```diff
@@ -239,19 +254,114 @@ func (p *Poller) pollTenantBlocks(ctx context.Context, tenantID string) ([]*back
...
+		return nil, nil, err
+	}
+
+	return p.flushMetaChannels(chMeta, chCompactedMeta)
+}
+
+func (p *Poller) quickPollTenantBlocks(
+	ctx context.Context,
+	tenantID string,
+	previous backend.Blocklist,
```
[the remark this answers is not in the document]
zalegrala: It felt like a clean way not to modify the old code and be able to fall back to existing behavior.  We can merge them into one if we want, I just didn't want to modify the existing in case we want to make it optional or some other approach.

tempodb/blocklist/poller.go
zalegrala: Good call, I'll make the adjustments to pass in the variables we've already captured above.

zalegrala:
Thanks for the review @joe-elliott.  I'm happy to chat through this whenever.

tempodb/backend/azure/azure.go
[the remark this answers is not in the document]
joe-elliott: we should consolidate this error (and all error logic) in a follow up PR

tempodb/backend/raw.go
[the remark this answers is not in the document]
joe-elliott: imo we should just change the way "Blocks" works and only have one codepath here to reduce complexity.

tempodb/blocklist/poller.go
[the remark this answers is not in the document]
joe-elliott: same thought here. we should have just one way of doing this (the new way) to reduce code complexity

tempodb/backend/raw.go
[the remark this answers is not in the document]
joe-elliott: this is harder. let's sync offline. i'm against adding any "tempo" specific information to this layer as it's mostly tempo agnostic now, but this may be an exception if there really is substantial gains by splitting this off.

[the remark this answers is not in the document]
zalegrala: I've refactored to include a `Find()` method that allows options to be passed in.  Future changes can be made to the opts struct to allow more complex searching, but for now I think this works and keeps the code agnostic as to what is being searched from the backend.

tempodb/blocklist/poller.go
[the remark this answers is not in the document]
zalegrala: I've refactored this out so that we only have one code path now.  Need to stare at some tests for a while until they pass again.

modules/frontend/v1/frontendv1pb/frontend.pb.go
```diff
@@ -308,38 +308,38 @@ func init() {
 }
 
 var fileDescriptor_8e6c94795ed772cd = []byte{
-	// 486 bytes of a gzipped FileDescriptorProto
```
zalegrala, line 311: I'm not sure why this changes.

tempodb/backend/s3/s3.go
```diff
@@ -276,6 +278,87 @@ func (rw *readerWriter) List(_ context.Context, keypath backend.KeyPath) ([]stri
...
+
+							matched, e := f(opts)
+							if e == backend.ErrDone {
+								return
+							}
+							if !matched {
+								continue
+							}
+
+							keys = append(keys, c.Key)
```
zalegrala, line 338: I don't believe this is thread safe.

zalegrala:
I'll drop the Find references next.  This now implements the ListBlocks on the RawReader after consideration of our options and some discussion.  S3 is currently the only concurrent implementation, and GCS should be as simple.

tempodb/backend/raw.go
```diff
@@ -40,6 +49,8 @@ type RawWriter interface {
 type RawReader interface {
 	// List returns all objects one level beneath the provided keypath
 	List(ctx context.Context, keypath KeyPath) ([]string, error)
+	// ListBlocks returns all blockIDs and compactedBlockIDs for a tenant.
+	ListBlocks(ctx context.Context, keypath KeyPath) (blockIDs []uuid.UUID, compactedBlockIDs []uuid.UUID, err error)
```
mapno, line 53: `ListBlocks` is not really _raw_, it's leaking backend implementation details. It could be made more generic with a callback if necessary to do the dir depth check,

```go
		parts := strings.Split(tenantFilePath, "/")
		// i.e: <tenantID/<blockID>/meta
		if len(parts) != 3 {
			return nil
		}
```

and merge the `switch` statement to `backend.Reader`

```go
		switch parts[2] {
		case backend.MetaName:
			metas = append(metas, id)
		case backend.CompactedMetaName:
			compactedMetas = append(compactedMetas, id)
		}
```

mapno: Not a super strong opinion, just think it'll be a cleaner interface.

mapno (commented):
Overall, I like the changes. Left a comment.

tempodb/backend/raw.go
zalegrala: I thought so too since this was the last approach I'd taken, but it came up in review that folks didn't like the concurrency living one layer above the raw.  See this commit for what was removed.  https://github.com/grafana/tempo/pull/2652/commits/c1da3a6416a36eb14f6225164fbfcfd9b0315d05

Not all backeneds support starting a list call in the middle, so it was necessary to know what backend we were working with.  This was part of the objection to this approach also.  This discovery was removed in this commit.
https://github.com/grafana/tempo/pull/2652/commits/d9145775d28bc9a7978980aecbeb043a13a432af

Do you have ideas about how we might approach this differntly?  I struggled to find a generic solution that could:
1. parallelize the calls
2. avoid mentioning anything about UUID or Blocks in the RawReaders
3. make something that was generic enough to be used in the UUID case (0-f searching) and in a general find case (0-9a-zA-Z) where the number of workers would take an even amount of the load depending on the use case.

The idea of providing "hits" to the backend about what to search within came up, but this felt like we were just dancing around the idea of using the word blocks in the backend.

What do you think?

tempodb/tempodb.go
```diff
@@ -426,10 +426,15 @@ func (rw *readerWriter) EnablePolling(sharder blocklist.JobSharder) {
 		rw.cfg.BlocklistPollTenantIndexBuilders = DefaultTenantIndexBuilders
 	}
 
-	level.Info(rw.logger).Log("msg", "polling enabled", "interval", rw.cfg.BlocklistPoll, "concurrency", rw.cfg.BlocklistPollConcurrency)
+	if rw.cfg.TenantPollConcurrency <= 0 {
```
joe-elliott, line 429: i really don't like that these are all uints. you're just following suit so it's fine, but they should really be ints

tempodb/blocklist/poller.go
```diff
@@ -225,33 +267,84 @@ func (p *Poller) pollTenantAndCreateIndex(ctx context.Context, tenantID string)
...
+		return nil, nil, err
 	}
 
+	previousMetas := previous.Metas(tenantID)
+	previousCompactedMetas := previous.CompactedMetas(tenantID)
+
 	bg := boundedwaitgroup.New(p.cfg.PollConcurrency)
-	chMeta := make(chan *backend.BlockMeta, len(blockIDs))
-	chCompactedMeta := make(chan *backend.CompactedBlockMeta, len(blockIDs))
+	chMeta := make(chan *backend.BlockMeta, len(currentBlockIDs))
```
joe-elliott, line 284: looking back, i really don't like that i used a channel as a buffer like this. would prefer just appending the block ids to a normal slice. it worked better when everything was async, now it's a bit weird.

not going to ask you to change it, but if you want to do a simpler pattern but felt bound to this, go for it.

tempodb/blocklist/poller.go
```diff
@@ -360,3 +468,19 @@ func sumTotalBackendMetaMetrics(blockMeta []*backend.BlockMeta, compactedBlockMe
 		compactedBlockMetaTotalBytes:   sumTotalBytesCBM,
 	}
 }
+
+func uuidInBlocklist(id uuid.UUID, metas []*backend.BlockMeta, compactedMetas []*backend.CompactedBlockMeta) bool {
+	for _, i := range metas {
```
joe-elliott, line 473: the thing that's being passed in to the `Do()` function is a `*List` right? the list object has maps to meta and compactedMetas which would be much faster lookup than this.

why don't we just take a *List instead of an interface?

tempodb/backend/azure/azure.go
```diff
@@ -182,6 +183,55 @@ func (rw *readerWriter) List(ctx context.Context, keypath backend.KeyPath) ([]st
...
+			if len(parts) != 2 {
+				continue
+			}
+
+			id, err = uuid.Parse(parts[0])
+			if err != nil {
+				return nil, nil, err
+			}
+
+			switch parts[1] {
```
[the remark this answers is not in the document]
zalegrala: Good call, this should be moved after, so we only parse the UUID at the last moment.

kvrhdn (commented):
Still going through the code but already some preliminary comments

cmd/tempo-cli/cmd-migrate-tenant.go
```diff
@@ -38,7 +38,7 @@ func (cmd *migrateTenantCmd) Run(opts *globalOptions) error {
 
 	// TODO create dest directory if it doesn't exist yet?
 
-	blocksDest, err := readerDest.Blocks(ctx, cmd.DestTenantID)
+	blocksDest, _, err := readerDest.Blocks(ctx, cmd.DestTenantID)
```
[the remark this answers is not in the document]
zalegrala: I think even without this change, compacted blocks are not migrated.  Is the intention here to move the blocks to a new tenant?  If we don't migrate the compacted blocks, they are still not query-able and will get picked up by retention and then the tenant is deleted which seems okay to me.

cmd/tempo-cli/shared.go
```diff
@@ -54,7 +54,7 @@ type blockStats struct {
 }
 
 func loadBucket(r backend.Reader, c backend.Compactor, tenantID string, windowRange time.Duration, includeCompacted bool) ([]blockStats, error) {
-	blockIDs, err := r.Blocks(context.Background(), tenantID)
+	blockIDs, _, err := r.Blocks(context.Background(), tenantID)
```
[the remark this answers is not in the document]
zalegrala: I've made an update to include the compacted results in a few places.  The places not updated aren't making use of the compacted blocks that I saw.

docs/sources/tempo/configuration/_index.md
```diff
@@ -731,6 +731,11 @@ storage:
             # Example: "endpoint: s3.dualstack.us-east-2.amazonaws.com"
             [endpoint: <string>]
 
+            # The number of list calls to make in parallel to the backend per instance.
+            # Adjustments here will impact the polling time, as well as the number of Go routines.
+            # Default is 3
```
zalegrala, line 736: I'm rethinking this and perhaps a default of `1` is better here.  Thoughts?

tempodb/backend/gcs/gcs.go
```diff
@@ -186,6 +190,112 @@ func (rw *readerWriter) List(ctx context.Context, keypath backend.KeyPath) ([]st
...
+
+		wg.Add(1)
+		go func(min, max uuid.UUID) {
+			defer wg.Done()
+
+			iter := rw.bucket.Objects(ctx, &storage.Query{
+				Prefix:      prefix,
+				Delimiter:   "",
+				Versions:    false,
+				StartOffset: prefix + min.String(),
```
mdisibio, line 225: Could we make use of EndOffset?  Without it it seems like there is redundant work.  For 4 shards the first request would list 100% of the bucket, the second request would list 75% of the bucket, etc.  I think you could use EndOffset=max.String(), and then that block=max gets picked up in the next shard where it equals min.  It gets a little tricky for max UUID 0xFFFF... because EndOffset is exclusive so it must not be present on the last request.  I haven't looked into the block boundaries not sure if that is actually a concern.

mdisibio:
Is there a way to enable the old polling behavior?  Thinking more about alternative backends like S3-compatible products.  Since this exercises new behavior (recursive prefix listing) it's possible that an issue could be found only after upgrade.  They could revert Tempo versions, but just switching back the polling until we could identify a fix would be nicer.   Not sure how feasible this is in the current PR, and I'm not necessarily pushing for it, but wanted to mention it.

tempodb/backend/gcs/gcs.go
```diff
@@ -186,6 +190,112 @@ func (rw *readerWriter) List(ctx context.Context, keypath backend.KeyPath) ([]st
...
+					}
+
+					x := bytes.Compare(id[:], min[:])
+					if x < 0 {
+						return
+					}
+
+					y := bytes.Compare(id[:], max[:])
+					if y > 0 {
+						return
```
joe-elliott, line 267: if we've gone past the max block boundary should we exit the loop?

this comment applies to the s3 implementation as well.

tempodb/backend/gcs/gcs.go
```diff
@@ -186,6 +190,112 @@ func (rw *readerWriter) List(ctx context.Context, keypath backend.KeyPath) ([]st
...
+					}
+					if err != nil {
+						errChan <- fmt.Errorf("iterating blocks: %w", err)
+						return
+					}
+
+					parts = strings.Split(attrs.Name, "/")
+					// ie: <tenant>/<blockID>/meta.json
+					if len(parts) != 3 {
```
joe-elliott, line 247: the prefix should contain the tenant id. i would expect the parts to contain 2, but i suppose this is just an a difference between s3 and the other backends? which only return the parts not in the prefix?

Also, wdyt think about these methods failing if not provided with a keypath of length 1? or maybe we should just change the "keypath []string" param to "tenant string"?

tempodb/backend/gcs/gcs.go
```diff
@@ -186,6 +190,112 @@ func (rw *readerWriter) List(ctx context.Context, keypath backend.KeyPath) ([]st
...
+					}
+
+					y := bytes.Compare(id[:], max[:])
+					if y > 0 {
+						return
+					}
+
+					switch parts[2] {
+					case backend.MetaName:
+						mtx.Lock()
```
joe-elliott, line 272: generally very detailed locking like this should be avoided unless absolutely necessary. ideal is:
```
Lock()
defer Unlock()
```
but sometimes that's unwieldy in its own way. I think a nice compromise here is to lock before the switch and unlock after.

this comment applies to the s3 implementation as well.

tempodb/blocklist/poller.go
```diff
@@ -95,6 +98,15 @@ func (ownsNothingSharder) Owns(_ string) bool {
 	return false
 }
 
+// OwnsEverythingSharder owns everything. You do not want this developer on your team.
+var OwnsEverythingSharder = ownsEverythingSharder{}
```
joe-elliott, line 102: unused? remove?

tempodb/blocklist/poller.go
```diff
@@ -73,6 +75,7 @@ var (
 // Config is used to configure the poller
 type PollerConfig struct {
 	PollConcurrency           uint
+	PollTenantConcurrency     uint
```
joe-elliott, line 78: unused? remove?

tempodb/blocklist/poller.go
```diff
@@ -226,69 +256,141 @@ func (p *Poller) pollTenantAndCreateIndex(ctx context.Context, tenantID string)
...
 	if err != nil {
-		metricBlocklistErrors.WithLabelValues(tenantID).Inc()
-		return []*backend.BlockMeta{}, []*backend.CompactedBlockMeta{}, err
+		return nil, nil, err
 	}
 
-	bg := boundedwaitgroup.New(p.cfg.PollConcurrency)
-	chMeta := make(chan *backend.BlockMeta, len(blockIDs))
-	chCompactedMeta := make(chan *backend.CompactedBlockMeta, len(blockIDs))
+	readPreviousSpan, _ := opentracing.StartSpanFromContext(derivedCtx, "readPreviousResults")
```
joe-elliott, line 272: an event may be more appropriate here. same for the other spans below (assemble/learn/etc)

tempodb/blocklist/poller.go
```diff
@@ -226,69 +256,141 @@ func (p *Poller) pollTenantAndCreateIndex(ctx context.Context, tenantID string)
...
 			}
 
-			m, cm, err := p.pollBlock(ctx, tenantID, uuid)
+			m, cm, pollBlockErr := p.pollBlock(pollSpanCtx, tenantID, id)
 			if m != nil {
-				chMeta <- m
-			} else if cm != nil {
-				chCompactedMeta <- cm
-			} else if err != nil {
+				mtx.Lock()
```
joe-elliott, line 343: lock and then defer an unlock over these detailed locks/unlocks

tempodb/blocklist/poller.go
```diff
@@ -226,69 +256,141 @@ func (p *Poller) pollTenantAndCreateIndex(ctx context.Context, tenantID string)
...
+
+	mtx := sync.Mutex{}
+	newBlockList := make([]*backend.BlockMeta, 0, len(currentBlockIDs))
+	newCompactedBlocklist := make([]*backend.CompactedBlockMeta, 0, len(currentCompactedBlockIDs))
 	anyError := atomic.Error{}
 
-	for _, blockID := range blockIDs {
+	learnSpan, _ := opentracing.StartSpanFromContext(derivedCtx, "learnFromPrevious")
+	newBlockIDs := []uuid.UUID{}
+	for _, blockID := range currentBlockIDs {
```
joe-elliott, line 299: we know which uuids are compacted and which are not. we lose that by putting everything in a newBlockIDs slice and then calling pollBlock. ideally we would use this information to reduce backend calls and have more direct? purposeful? terse? code. 

however, this is simple and easy to follow so i'm fine with it. it's still a very large improvement over current.

joe-elliott (commented):
Lot of details that need to be addressed before merging but only one big concern. We need to be 100% confident that block guids on the block boundaries in the GCS and S3 backends are correctly owned by exactly one shard.

I think an integration test in which you wrote a bad block on guids like a min boundary/max boundary/0000/FFFF/etc and confirm that tempo had polling errors would be sufficient.

tempodb/blocklist/poller.go
zalegrala: This is useful for building binaries outside of tempo main.  I'd like to keep it for this purpose.

docs/sources/tempo/configuration/_index.md
zalegrala: I had tested with 20 on gcs and got the best performance, but this was not running with the large number of instances we see in some of our environments.  The good thing is that we have a configuration option to tune.  I'm fine with coming out the gate with 3 and then finding a nice balance as we roll this out, then come back and adjust our defaults if it turns out we were way off.

tempodb/backend/gcs/gcs.go
zalegrala: There is currently an early return of we reach a block ID that is out of the scope for the particular routine.  In paged results we would still throw away the last part of a page, but not quite as bad as you mention.

tempodb/backend/s3/s3.go
```diff
@@ -277,6 +281,117 @@ func (rw *readerWriter) List(_ context.Context, keypath backend.KeyPath) ([]stri
...
+						if parts[2] != backend.MetaName && parts[2] != backend.CompactedMetaName {
+							continue
+						}
+
+						id, err := uuid.Parse(parts[1])
+						if err != nil {
+							continue
+						}
+
+						x := bytes.Compare(id[:], min[:])
```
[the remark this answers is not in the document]
zalegrala: I think no, but since we're comparing the end of the range, I thought it was easy enough to check the start of the range as well.

cmd/tempo-cli/cmd-query-trace-summary.go
```diff
@@ -87,11 +87,13 @@ func sortServiceNames(nameFrequencies map[string]int) PairList {
 }
 
 func queryBucketForSummary(ctx context.Context, percentage float32, r backend.Reader, c backend.Compactor, tenantID string, traceID common.ID) (*TraceSummary, error) {
-	blockIDs, err := r.Blocks(context.Background(), tenantID)
+	blockIDs, compactedBlockIDs, err := r.Blocks(context.Background(), tenantID)
```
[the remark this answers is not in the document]
zalegrala: Any reason not to include them?

tempodb/blocklist/poller.go
zalegrala: Yep, I started with events, but since we don't have support for them in traceql I opted for using full spans.

tempodb/blocklist/poller.go
zalegrala: This seems okay to me.  What's the reason for your suggestion?  Just readability?

tempodb/blocklist/poller.go
zalegrala: The `newBlockIDs` slice is used to build a list of UUIDs that we have not previously fetched, and then poll them.  What information are we losing here by building the slice?  We still need to fetch the meta in order to build up the blocklist.

zalegrala:
Thanks for the review.

@mdisibio This came up in https://github.com/grafana/tempo/pull/2652#discussion_r1288488417 in this PR.  We opted to chose one path that worked to avoid the complexity.

@joe-elliott I'll work up a test that illustrates what you mention about blocks being on either side and right on the boundary.

tempodb/blocklist/poller.go
joe-elliott: I don't think I understand? If fa dependent codebase wants an `OwnsEverythingSharder` it can make it itself. We shouldn't keep this code unless we're using it.

tempodb/blocklist/poller.go
joe-elliott: Sounds like a good incentive to get it in :)

tempodb/blocklist/poller.go
joe-elliott: readability as well as reducing the likelihood of writing a deadlock

tempodb/blocklist/poller.go
joe-elliott: At this point we are aware which block ids are compacted and which are not, but we put those all in the same slice and lose that information.

The "pollBlock" function pulls the meta and if that fails it bulls the compacted meta. We still need this 2 tiered approach for the metas, but for the compacted metas we can skip the first GET and go straight to trying to grab the compacted meta.

tempodb/blocklist/poller.go
zalegrala: Agreed, but still trying to get this one in :)

tempodb/blocklist/poller.go
zalegrala: Sorry, is there risk deadlock on the append?  What am I missing?

The only reason I'm pushing back here is that I'd rather not lock on error.  I can adjust to check error first and such, but locking only when appending.

tempodb/blocklist/poller.go
zalegrala: I may use this in the test that I mentioned I'd write.  Let's see.

cmd/tempo-cli/cmd-query-trace-summary.go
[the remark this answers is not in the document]
zalegrala: Okay, I'll adjust.

tempodb/blocklist/poller.go
zalegrala: I saw that also, but didn't chase it down.  I've added a TODO for this.

tempodb/backend/gcs/gcs.go
mdisibio: Right, I am thinking about overhead on the object storage backend.  The implementation for S3 seems good, the work is split across multiple requests with the continuation token, and it will break after reaching the shard max (up to only 1 page of results thrown away).    But for GCS it looks like a single call, and even though we stop processing the results for the shard, the work has already been done by the object storage to provide them.

zalegrala: Ah interesting.  I'd thought the iteration token worked similarly, but you could be right.  I've made the adjustment to include the `EndOffset`.

tempodb/backend/s3/s3.go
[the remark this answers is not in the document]
zalegrala: We have test coverage for this now.

tempodb/backend/gcs/gcs.go
zalegrala: This return will stop the goroutine for the range.  We want the other routines to finish as well.

tempodb/backend/gcs/gcs.go
zalegrala: If we lock before the switch, then we lock just to check if there is match, which is on every object even if we won't end up making a modification that requires the lock.  I suppose we could check if either names are matched and then lock and then switch, but that seems less readable.  I'm find with making the change here, but I don't think its unreadable as is, since we lock only for one line, and then unlock.  I agree about the general pattern, but in this case I think it warranted the specific locks.  Given this, would you still prefer the lock + defer?

zalegrala: Also defer in a for loop inside a go routine seems a little more dangerous to me.  So explicit lock + unlock seems like the right approach.

tempodb/blocklist/poller.go
```diff
@@ -138,52 +149,81 @@ func (p *Poller) Do() (PerTenant, PerTenantCompacted, error) {
 		return nil, nil, err
 	}
 
+	m := &sync.Mutex{}
 	blocklist := PerTenant{}
 	compactedBlocklist := PerTenantCompacted{}
-
+	errs := []error{}
 	consecutiveErrors := 0
```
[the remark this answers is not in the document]
zalegrala: I'll adjust to ensure that we do the lock/defer where applicable.  Then we don't have to worry as much, even if we do end up locking more than we need.

tempodb/blocklist/poller.go
```diff
@@ -226,69 +254,164 @@ func (p *Poller) pollTenantAndCreateIndex(ctx context.Context, tenantID string)
...
+
+		go func(id uuid.UUID) {
 			defer bg.Done()
 
 			if p.cfg.PollJitterMs > 0 {
 				time.Sleep(time.Duration(rand.Intn(p.cfg.PollJitterMs)) * time.Millisecond)
 			}
 
-			m, cm, err := p.pollBlock(ctx, tenantID, uuid)
+			m, cm, pollBlockErr := p.pollBlock(pollSpanCtx, tenantID, id)
```
mdisibio, line 364: This call will make 2 reads for every compactedBlock, since it first tries meta.json.  However we already knew the block is compacted from the ListBlocks call.  I think this could be addressed by keeping the `newBlockIDs` slice split between compacted and regular, and then pass that info to pollBlock (maybe an option like `compacted bool`.  So on a compacted block it doesn't need to try meta.json.   I think for a regular block it still has to try both because there is a non-zero chance the block got compacted in between ListBlocks and pollBlock.

integration/poller/poller_test.go
```diff
@@ -0,0 +1,221 @@
...
+			incrementUUIDBytes(base)
+			expected = append(expected, uuid.UUID(base))
+
+			incrementUUIDBytes(base)
+			expected = append(expected, uuid.UUID(base))
+
+			writeTenantBlocks(t, w, tenant, expected)
+
+			mmResults, cmResults, err := rr.ListBlocks(context.Background(), backend.KeyPath([]string{tenant}))
+			require.NoError(t, err)
```
joe-elliott, line 144: let's assert here that mmResults and cmResults are expected lengths/contents

tempodb/backend/gcs/gcs.go
joe-elliott: > This return will stop the goroutine for the range. We want the other routines to finish as well.

if the guid is > max then this goroutine is done right? it should exit instead of churning through larger guids

tempodb/backend/gcs/gcs.go
joe-elliott: > If we lock before the switch, then we lock just to check if there is match, which is on every object even if we won't end up making a modification that requires the lock.

You're checking a few lines up if this a meta or a compacted meta. So we should already know it's a match.

> Also defer in a for loop inside a go routine seems a little more dangerous to me. So explicit lock + unlock seems like the right approach.

That's why I said: "I think a nice compromise here is to lock before the switch and unlock after."

tempodb/blocklist/poller.go
joe-elliott: Can you move it to the test then?

tempodb/blocklist/poller.go
```diff
@@ -226,69 +254,164 @@ func (p *Poller) pollTenantAndCreateIndex(ctx context.Context, tenantID string)
...
 
-	for _, blockID := range blockIDs {
+	for _, blockID := range currentCompactedBlockIDs {
+		// if we already have this block id in our previous list, use the existing data.
+		if v, ok := cm[blockID]; ok {
+			newCompactedBlocklist = append(newCompactedBlocklist, v)
+			continue
+		}
+
+		// TODO: Review this hack to avoid polling for compacted blocks that we
```
joe-elliott, line 312: there's a lot of commented out code below, can we remove? it's just cluttering

tempodb/backend/gcs/gcs.go
joe-elliott: Yup. I see. You are correct :+1:

tempodb/backend/gcs/gcs.go
zalegrala: I think it depends if you include a delimiter or not on s3 as to the number of parts.  The comment there is an example result.

I'm okay with adjusting the interface to be a string instead of a path.

joe-elliott: not a big deal either way. if you find that changing keypath -> tenantid makes things very difficult you can skip it. it's just closer to the true meaning of what this method is doing and will make it harder to mess up and pass an empty keypath

tempodb/backend/azure/v2/v2.go
```diff
@@ -164,6 +165,62 @@ func (rw *V2) List(ctx context.Context, keypath backend.KeyPath) ([]string, erro
...
+
+	if len(prefix) > 0 {
+		prefix += dir
+	}
+
+	var parts []string
+	var id uuid.UUID
+
+	pager := rw.containerClient.NewListBlobsFlatPager(&container.ListBlobsFlatOptions{
+		Include: container.ListBlobsInclude{},
```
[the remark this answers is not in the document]
zalegrala: Sure, I'll make a note to file an issue after merge.

tempodb/backend/gcs/gcs.go
zalegrala: Sorry you're right, you did say that.

zalegrala: I've pushed a change for this.

tempodb/backend/gcs/gcs.go
zalegrala: I've adjusted the interface and implementations to use tenant string instead of keypath.

joe-elliott (approved):
Couple of smaller comments, but I'm g2g. Great improvement!

tempodb/blocklist/poller.go
```diff
@@ -226,71 +245,172 @@ func (p *Poller) pollTenantAndCreateIndex(ctx context.Context, tenantID string)
...
-	close(chMeta)
-	close(chCompactedMeta)
+	for _, blockID := range currentCompactedBlockIDs {
+		// if we already have this block id in our previous list, use the existing data.
+		if v, ok := cm[blockID]; ok {
+			newCompactedBlocklist = append(newCompactedBlocklist, v)
+			continue
+		}
+
+		// TODO: Review the ability  to avoid polling for compacted blocks that we
```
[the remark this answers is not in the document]
zalegrala: Not quite.  What I had in mind might not be possible, but the note here is about avoiding pulling the data entirely if the block transitions from meta -> compactedMeta.  We have all the detail except the time it was compacted.  More generally, I wonder if we already had this data in memory, then we know that it is only one polling cycle out of date, and perhaps we could be fuzzy enough on this that we don't have to know the precise time.  More thinking needed probably.

PR #2656, opened 2023-07-14
Remove mention to `$__tags`

fabrizio-grafana:
[This documentation page](https://grafana.com/docs/tempo/latest/getting-started/tempo-in-grafana/#linking-traces-and-metrics) contains a mention to the use of `$__tags`:

> For example, you can use span attributes to metric labels by using the $__tags keyword to convert span attributes to metrics labels.

[As shown in this issue](https://github.com/grafana/grafana/issues/64871), its use might be misleading when dealing with YAML files, which require to escape the literal `$` by writing `$$`. Since the documentation later suggests to [refer to this page](https://grafana.com/docs/grafana/latest/datasources/tempo/#trace-to-metrics) for more information and that page also contains a YAML example which uses `$$__tags`, I propose to simply remove the potentially misleading mention to `$__tags` and instead redirect the user to the dedicated page.

If we like the changes, I'll take care of updating also [the corresponding page for Grafana Cloud](https://grafana.com/docs/grafana-cloud/data-configuration/traces/#linking-traces-and-metrics).

Outcome after the changes:
image 

**Checklist**
- [x] Tests updated
- [x] Documentation added


joe-elliott:
Thanks for the PR! I will let @knylander-grafana make the call on how to clarify this.

PR #2657, opened 2023-07-14
Add ability to disable histogram in servicegraphs

mapno:
**What this PR does**:

Add config param `use_histograms` to control cardinality in servicegraph metrics. If set to false, it'll use a counter to measure latency instead of histograms. This saves a lot of active series

**Checklist**
- [x] Tests updated
- [ ] Documentation added


knylander-grafana:
@mapno Will we need docs for this?

mapno:
If it ever moves forward, yes. For now, it's just a draft. Likely, this other approach will be accepted instead https://github.com/grafana/tempo/pull/2667.

PR #2660, opened 2023-07-14
Fix a few bugs in vParquet2 structural operators

### LINKED ISSUE #2658: Child operator sometimes returns incorrect results
**Describe the bug**
Occassionally the child structural operator returns incorrect results. This is easily reproduced using the https://github.com/grafana/intro-to-mlt/ demo.

**To Reproduce**
1. Start up the mentioned demo
2. Write a query like 
```
{resource.service.name="mythical-requester"} > { resource.service.name="mythical-server" && kind != server} 
```
3. See that it returns spans that are not children of a span with `resource.service.name="mythical-requester"`

image

mdisibio:
**What this PR does**:
A few fixes in the structural operators:
1. Need to skip over spans with broken nesting. This is detected as any value of either span == 0.  Without this it was falsely identifying matches on partial traces where the spans had broken nesting
2. Sibling operator would falsely identify a span as its own sibling if it was present in multiple spansets.  I.e. `{ .foo} ~ {.foo}` because each input technically had the same parent.  There are two ways to fix this, and I chose the more efficient one.  Theoretically we could loading both Parent and Left and ensure Lefts are different (not same spans).  However instead I am just checking struct pointers to ensure not the same `*span`.  This works until the TraceQL engine starts making multiple passes over a file (which is not the plan and may be never). 
3. There was also a small typo in DescendantOf that could have led to more false positives.  I never saw this in the wild.

Fixes #2658 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/encoding/vparquet2/block_traceql.go
+++ b/tempodb/encoding/vparquet2/block_traceql.go
@@ -74,7 +74,12 @@ func (s *span) SiblingOf(x traceql.Span) bool {
 			ss.nestedSetParent == 0 {
 			return false
 		}
-		return ss.nestedSetParent == s.nestedSetParent
+		// Same parent but not ourself
+		// Checking pointers here means we don't have to load
+		// an additional column of nestedSetLeft but assumes the span
+		// object. This is true because all TraceQL executions are
+		// currently single-pass.
+		return ss.nestedSetParent == s.nestedSetParent && s != ss
 	}
 	return false
 }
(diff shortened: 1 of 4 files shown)


tempodb/encoding/vparquet2/block_traceql.go
```diff
@@ -55,21 +55,41 @@ func (s *span) DurationNanos() uint64 {
...
 	if ss, ok := x.(*span); ok {
-		return s.nestedSetLeft > ss.nestedSetLeft && s.nestedSetLeft < ss.nestedSetRight
+		if s.nestedSetLeft == 0 ||
+			s.nestedSetRight == 0 ||
+			ss.nestedSetLeft == 0 ||
+			ss.nestedSetRight == 0 {
+			// Spans with missing data, never a match.
+			return false
+		}
+		return s.nestedSetLeft > ss.nestedSetLeft && s.nestedSetRight < ss.nestedSetRight
```
mdisibio, line 65: The typo was the third var in this statement `s.nestedSetLeft` -> `s.nestedSetRight`

zalegrala (approved):

PR #2661, opened 2023-07-14
Refactor poller into new module and adjust tempodb.New everywhere

zalegrala:
**What this PR does**:

The poller has write operations which, if not canceled, have unknown behavior.  In order to pass the service context into the poller, this change proposes moving the poller itself into a module that can be used as a dependency and allow for more control over the termination of the write operations to the backend.

Also included here is a rename of the instance variables as well as the instance type of tempoDB.  I believe this is more clear as to the intention.

Structurally, the `tempodb.New` now returns the `tempoDB` instance. This allows the caller to use the returned instance as whatever interface is desired. Adjustments for this contract have been updated everywhere.

Additional formatting changes for `gofumpt` taking over my editor.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo/app/app.go
+++ b/cmd/tempo/app/app.go
@@ -38,10 +38,13 @@ import (
 	"github.com/grafana/tempo/pkg/usagestats"
 	"github.com/grafana/tempo/pkg/util"
 	"github.com/grafana/tempo/pkg/util/log"
+	"github.com/grafana/tempo/tempodb"
 )
 
-const metricsNamespace = "tempo"
-const apiDocs = "https://grafana.com/docs/tempo/latest/api_docs/"
+const (
+	metricsNamespace = "tempo"
+	apiDocs          = "https://grafana.com/docs/tempo/latest/api_docs/"
+)
 
 var (
 	metricConfigFeatDesc = prometheus.NewDesc(
@@ -71,6 +74,7 @@ type App struct {
 	ingester       *ingester.Ingester
 	generator      *generator.Generator
 	store          storage.Store
+	poller         tempodb.Poller
 	usageReport    *usagestats.Reporter
 	MemberlistKV   *memberlist.KVInitService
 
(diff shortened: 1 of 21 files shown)


zalegrala:
The linter is failing due to `tempoDB` being private.  Is there a reason not to make it public?

joe-elliott:
This PR is difficult to parse b/c of all the formatting changes. I think we should leave gofumpt off until we merge it into main to help with reviews.

I think the real win here is treating polling as a service instead of something you just "Enable" on the store module. I think this PR should double down on that idea and remove everything else.

**Simpler Option:**

Create a new `/pkg/poller`. Have a simple call that returns a service:

`poller.New()`

pass in the store and any other necessary components and have the simple wrapper call `.EnablePolling(ctx)`. This will accomplish the goal of making the polling cycle context aware with a very small amount of code change.

**More Complete Option:**

Remove the `EnablePolling()` function from tempodb and initiate polling directly through the `tempodb/blocklist` package. `pkg/poller` looks the same externally, but now instantiates a blocklist poller and manages the lifecycle completely outside of tempodb. This will require a deeper analysis of the interaction between polling, tempodb, and the blocklist package.

zalegrala:
For the first option, I think we probably just need to pass a config that switches the job sharder, probably based on config and defaults.  This is probably a good direction for this PR and I don't think backs us into a corner with respect to the second idea.

For the second option, we need a way to update the `blocklist` on the tempodb instance.  Currently the poller runs periodically and updates the blocklist.  The other components of tempodb just call into the blocklist to get the in memory results.  We could create a `db.SetPoller()` or but the current approach where its created at New and everything is somewhat self contained.  I think the interfaces could shift a little more, but still self contained within tempodb.  I don't immediately see a clear option to update the blocklist and keep access to it private without tempodb creating that on New.  Maybe there are some approaches I'm not thinking of, so if you have ideas I'm all ears.  It think the mention of dependency injection made me wonder a little about Set() calls for updating private struct keys, but then there is more complexity around protecting updates, etc.

joe-elliott:
Agree the second options will be a bit more sweeping. I think the first option is a good path to get the desired effect and perhaps, longer term, setup the something like the second.

zalegrala:
I pushed a few commits that move the poller into a module/service.  The `running()` function doesn't call the EnablePolling due to the way that the Poller is started with the sharder.   It gets a little squirrelly.

PR #2667, opened 2023-07-18
Allow to disable client metrics in service graphs processor

mapno:
**What this PR does**:

In the service graph processor, we generate latency metrics from the POVs of the client and server span (`traces_service_graph_request_client_seconds` and `traces_service_graph_request_server_seconds` respectively), but only the server histogram is used in the Grafana visualization. The client histogram remains mostly ignored, adding active series which impacts the performance of the processor.

This PR adds a new config param `enable_client_metrics` and override `metrics_generator_processor_service_graphs_enable_client_metrics` to allow disable creating that metric.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


zalegrala (approved):
Nice, this looks pretty good to me.  What's the reason for the `QueryExact` change in the tests?

cyrille-leclerc:
FYI Application Observability workflows rely on service graph client metrics to generate the RED metrics of the downstream services & databases.
On the other hand, App O11y is interested in disabling the span metrics for span types client and producer as we don't use them.

mapno:
> FYI Application Observability workflows rely on service graph client metrics to generate the RED metrics of the downstream services & databases.

That's ok, since this config option will be disabled by default.

PR #2678, opened 2023-07-19
Delete the index for empty tenants

### LINKED ISSUE #2520: Tempo leaves tenant indexes on tenants with no blocks

zalegrala:
**What this PR does**:

When a tenant stops writing to tempo, the blocks eventually get cleaned up by the retention, but the tenant index is still left and so the tenant is still left in the list when the poller determines which indexes to pull.  This can create additional load for no benefit long term if the tenant doesn't return with traffic.

Here we call the recently implemented `Delete` on the `RawWriter` to remove the tenant indexes from the backend to complete the deletion of all tenant objects when no blocks or compacted blocks have been found to exist.

Fixes #2520

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/querier/external/client_test.go
+++ b/modules/querier/external/client_test.go
@@ -93,6 +93,7 @@ func (t *stubbedProvider) getToken(_ context.Context, _ string) (*oauth2.Token, 
 		AccessToken: t.dummyToken,
 	}, nil
 }
+
 func getStubbedTokenProvider(dummyToken string) tokenProvider {
 	return &stubbedProvider{dummyToken: dummyToken}
 }
--- a/tempodb/backend/mocks.go
+++ b/tempodb/backend/mocks.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"io"
+	"strings"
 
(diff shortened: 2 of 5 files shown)


mdisibio (approved):
Nice cleanup.  I think all of the real object stores will "delete" the tenant folder when the last item is removed (i.e. index.json.gz).  Will this be a problem for the `local` backend?  It might be used in some cases for non-production setups.    The `local-blocks` processor and the ingester also use a `local` backend but not with polling so they should be ok.

zalegrala:
The `local` backend would leave the directories around, good eyes.  I suspect this isn't an issue due to the speed of local disk in comparison to network calls.  Since the use cases you mention don't keep an index, I'm okay with this as is.  We can revisit if it comes up again.

PR #2679, opened 2023-07-20
Change default search query timeout on querier

joe-elliott:
**What this PR does**:
Adjusts the default search query timeout from 30s to 10s. This aligns both trace by id query timeout and search. Both of these values still feel quite high, but additional experimentation would be necessary to push them lower.

**Checklist**
- [ ] Tests updated
- [x] Documentation added


zalegrala (approved):

knylander-grafana (approved):
Thank you for adding doc!

Issue #2627, opened 2023-07-07
[DOC] Document user-configurable overrides module


knylander-grafana:
Document the user-configurable overrides module. Refer to https://github.com/grafana/tempo/pull/2543 for details of what the feature does and it's architecture.

kvrhdn:
Quick list with tasks:

- [ ] Document new configuration options
- [ ] Describe how this feature works
- [ ] Describe how to use it from a client perspective (API, behaviour)
- [ ] Describe how to deploy it (prefer 2 bucket setup, mention 1 bucket with prefix)

PR #2681, opened 2023-07-20
user-configurable overrides: handle versioning, add PATCH endpoint

kvrhdn:
**What this PR does**:
Add version handling and a new PATCH endpoint to the user-configurable overrides. Existing endpoints gain support for [conditional headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Conditional_requests) adding protection against data races.

New PATCH endpoint: instead of providing the full limits struct, clients can send a partial json with the other fields set to null or omitted. End result: Tempo will read the existing user-configurable overrides and only updates the fields that are set in the patch.

Changes to GET, POST and DELETE:

GET and POST will return the current version of the overrides using the `Etag` header.
```
GET http://localhost:3200/api/overrides

HTTP/1.1 200 OK
Content-Type: application/json
Etag: 123
```
POST and DELETE must have the `If-Match` hedaer containing the version of a previous request. This ensures the overrides haven't been modified since the last GET.

`backend.VersionedReaderWriter` is a new interface to add version handling to backends. While GCS and Azure support conditional headers, S3 does not. Instead we adopt a best effort mechanism in which we fetch the current version and check it before writing.

Implementation:
- GCS: All good ✅
- S3: as good as we can get it without API support or external dependencies
- Azure: TODO ❌ reading versions and setting if-match headers requires us to update to the new SDK, I will do this in a separate PR

~~Fixes #<issue number>~~

**Checklist**
- [x] Tests updated
- [ ] Documentation added TODO https://github.com/grafana/tempo/issues/2627


modules/overrides/userconfigurableapi/client.go
```diff
@@ -42,74 +47,89 @@ var (
 type UserConfigurableOverridesClientConfig struct {
 	Backend string `yaml:"backend"`
 
+	// ConfirmVersioning is enabled when creating the backend client. If versioning is disabled no
+	// checks against concurrent writes will be performed.
+	ConfirmVersioning bool `yaml:"confirm_versioning"`
```
kvrhdn, line 52: I considered placing this on `gcs.Config` instead but I'm worried this would conflict with our existing bucket configuration. For storing traces you don't need versioning, so setting this to default true would cause a lot of issues.

modules/overrides/userconfigurableapi/client.go
```diff
@@ -118,12 +138,9 @@ func (o *userConfigOverridesClient) Get(ctx context.Context, userID string) (ten
...
 
-	reader, _, err := o.r.Read(ctx, OverridesFileName, []string{OverridesKeyPath, userID}, false)
-	if err == backend.ErrDoesNotExist {
-		return nil, nil
-	}
+	reader, version, err := o.rw.ReadVersioned(ctx, OverridesFileName, []string{OverridesKeyPath, userID})
 	if err != nil {
-		return
+		return nil, "", err
 	}
```
kvrhdn, line 144: This is a change in behaviour: instead of returning `nil` on not found just pass through the `backend.ErrDoesNotExist` error. I thought this would be easier and make the client more of a simple passthrough component.

tempodb/backend/azure/azure.go
```diff
@@ -198,6 +211,44 @@ func (rw *readerWriter) ReadRange(ctx context.Context, name string, keypath back
 func (rw *readerWriter) Shutdown() {
 }
 
+func (rw *readerWriter) WriteVersioned(ctx context.Context, name string, keypath backend.KeyPath, data io.Reader, version backend.Version) (backend.Version, error) {
+	// TODO use conditional if-match API
```
kvrhdn, line 215: To properly implement this we need to upgrade our Azure SDK from [azure-storage-blob-go](https://github.com/Azure/azure-storage-blob-go) to the newer [azure-sdk-for-go](https://github.com/Azure/azure-sdk-for-go). Since this impacts the entire package I'm going to do this in a separate PR...

tempodb/backend/versioned.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+type fakeVersionedReaderWriter struct {
+	RawReader
+	RawWriter
+}
+
+var _ VersionedReaderWriter = (*fakeVersionedReaderWriter)(nil)
+
+func NewFakeVersionedReaderWriter(r RawReader, w RawWriter) VersionedReaderWriter {
```
zalegrala, line 45: Consider returning the struct instead of the interface.

zalegrala (commented):
This looks pretty good to me.  One comment about the signatures in a couple spots.  I was reminded the other day while digging around in tempodb, that there are some good reasons to accept interfaces as arguments, and return structs.  I think it can make working with those objects a little easier in some cases, like testing.  "accept interfaces, return structs" goes back a ways in google, but I don't know the source.  Anyway, maybe worth considering or some conversation.  What do you think?

tempodb/backend/gcs/gcs.go
```diff
@@ -30,33 +31,62 @@ type readerWriter struct {
...
 }
 
 // New gets the GCS backend
 func New(cfg *Config) (backend.RawReader, backend.RawWriter, backend.Compactor, error) {
-	return internalNew(cfg, true)
+	rw, err := internalNew(cfg, true)
+	return rw, rw, rw, err
+}
+
+func NewVersionedReaderWriter(cfg *Config, confirmVersioning bool) (backend.VersionedReaderWriter, error) {
```
mapno, line 53: I find the signature a bit redundant. We're instantiating a `backend.VersionedReaderWriter`, but if `confirmVersioning` is not `true`, then it really isn't versioned.

IMO we could keep `confirmVersioning` to the caller.

modules/overrides/userconfigurableapi/client.go
mapno: Agree with the reasoning here 👍

modules/overrides/userconfigurableapi/client.go
```diff
@@ -42,79 +47,95 @@ var (
...
+		if err != nil {
+			return nil, err
+		}
+		rw = backend.NewFakeVersionedReaderWriter(r, w)
 	case backend.GCS:
-		reader, writer, _, err = gcs.New(cfg.GCS)
+		rw, err = gcs.NewVersionedReaderWriter(cfg.GCS, cfg.ConfirmVersioning)
 	case backend.S3:
-		reader, writer, _, err = s3.New(cfg.S3)
+		rw, err = s3.NewVersionedReaderWriter(cfg.S3)
```
mapno, line 118: I think we could log that the S3 implementation is best-effort.

mapno (approved):
LGTM. Just left a few minor comments.

tempodb/backend/gcs/gcs.go
kvrhdn: The reason I pass it down to the gcs layer is because we need the client to lookup the bucket settings and verify versioning is enabled. So the caller can't do this (unless we expose more but I'd like to keep that limited).

I agree it's weird to create a VersionedReaderWriter but disable versioning. This boolean is more of a "check versioning is enabled, otherwise crash" setting. Disabling means "I don't care if versioning is actually enabled, let's just roll with it".
If versioning is not enabled the precondition headers just won't be checked but the operations will still work.

I added it because our e2e tests can't use versioning on gcs (unfortunately).

Alternatively we could also fall back to a regular RawReader and RawWriter but that would require a lot of annoying code because the method interface is different.

tempodb/backend/versioned.go
kvrhdn: That's a good tip that I almost always forget about. I have the tendency to hide abstractions from my Java days, but I agree returning the struct gives more flexibility to the caller since they can create their own interface around it.

For this specific example: `fakeVersionedReaderWriter` is private and returning it gives me this warning in GoLand. Not sure if making it public or just returning the interface would be best here? 🤔😅 

Screenshot 2023-07-26 at 15 40 06

tempodb/backend/azure/azure.go
kvrhdn: https://github.com/grafana/tempo/issues/2705

tempodb/backend/gcs/gcs.go
mapno: I'm fine if the reason is keeping the rest of the code clean.

tempodb/backend/versioned.go
zalegrala: I think making it public is okay.

PR #2682, opened 2023-07-20
Validate user-configurable overrides

kvrhdn:
**What this PR does**:
Validate the user-configurable overrides before storing them in the backend.

I've also split off overrides-api into a separate module to make matters more explicit. This also opens the way to running the override-api as a separate component.

Note this will conflict slightly with https://github.com/grafana/tempo/pull/2681. I can address merge conflicts when either PR gets merged.

~~Fixes #<issue number>~~

**Checklist**
- [x] Tests updated
- [ ] Documentation added TODO https://github.com/grafana/tempo/issues/2627


cmd/tempo/app/overrides_validation.go
```diff
@@ -0,0 +1,30 @@
...
+
+func (v *overridesValidator) Validate(limits *userconfigurableapi.UserConfigurableLimits) error {
+	var validForwarders []string
+	for _, f := range v.cfg.Distributor.Forwarders {
+		validForwarders = append(validForwarders, f.Name)
+	}
+
+	if limits.Forwarders != nil {
+		for _, f := range *limits.Forwarders {
+			if !slices.Contains(validForwarders, f) {
```
mapno, line 23: Nit: a map would be more efficient

mapno (approved):
LGTM

cmd/tempo/app/overrides_validation.go
kvrhdn: Good point, I've updated the implementation :)

PR #2691, opened 2023-07-24
Avoid panic in cachedWriter

zalegrala:
Here we pass the cache.Delete() call on to the nextWriter instead of panic in cases where we do not want to cache.  Cache deletion is still not supported and will panic.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/backend/cache/cache.go
+++ b/tempodb/backend/cache/cache.go
@@ -110,8 +110,11 @@ func (r *readerWriter) CloseAppend(ctx context.Context, tracker backend.AppendTr
 	return r.nextWriter.CloseAppend(ctx, tracker)
 }
 
-func (r *readerWriter) Delete(ctx context.Context, name string, keypath backend.KeyPath) error {
-	return r.nextWriter.Delete(ctx, name, keypath)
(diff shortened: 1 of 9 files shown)


tempodb/backend/cache/cache.go
```diff
@@ -110,8 +110,8 @@ func (r *readerWriter) CloseAppend(ctx context.Context, tracker backend.AppendTr
 	return r.nextWriter.CloseAppend(ctx, tracker)
 }
 
-func (r *readerWriter) Delete(context.Context, string, backend.KeyPath) error {
-	panic("delete is not supported for cache.Cache backend")
+func (r *readerWriter) Delete(ctx context.Context, name string, keypath backend.KeyPath) error {
+	return r.nextWriter.Delete(ctx, name, keypath)
```
joe-elliott, line 114: we can't really do this. the point of the panic was to highlight that the cache does not support deleting. this is making it silently fail which means we will create some difficult to detect bugs in the future.

zalegrala: This has to do with the reads for these objects will still be cached?

joe-elliott (approved):

PR #2694, opened 2023-07-25
Update traces.md

shanvichid:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/traces.md
+++ b/docs/sources/tempo/traces.md
@@ -96,4 +96,4 @@ Span
 : A unit of work done within a trace.
 
 Trace
-: An observed execution path of a request through a distributed system.
+: An observed execution path of a request through a distributed system. Traces capture details of requests as they move through a system. They're composed of spans, which are data structures that represent individual operations in the flow of execution. 


zalegrala:
This looks reasonable to me, thank you.  Would you like to sign the CLA so we can take your contribution?

docs/sources/tempo/traces.md
```diff
@@ -96,4 +96,4 @@ Span
 : A unit of work done within a trace.
 
 Trace
-: An observed execution path of a request through a distributed system.
+: An observed execution path of a request through a distributed system. Traces capture details of requests as they move through a system. They're composed of spans, which are data structures that represent individual operations in the flow of execution.
```
knylander-grafana, line 99: Thank you for updating the definition. We already have a definition from the opening paragraph of the page, which your suggestion is similar to. What do you think about changing the definition to match that content? 

We're currently building a glossary solution for the docs website, so these updates will be included and consistent across all of the Grafana docs (not just Tempo docs). 

I'm not including the definition for spans here, because we want to keep the traces definition focused on traces.  Also, the user can look immediately above the trace definition to see the span definition. 

```suggestion
: A trace represents the whole journey of a request or an action as it moves through all the nodes of a distributed system, especially containerized applications or microservices architectures. Traces are composed of spans.  
```

knylander-grafana:
I've incorporated the changes from this PR here: https://github.com/grafana/tempo/pull/3060

PR #2698, opened 2023-07-25
Add traceql support for `{.a != nil}`

ie-pham:
**What this PR does**: Accepts {.a != nil} pattern by changing it to use OpNone instead

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


tempodb/encoding/vparquet2/block_traceql.go
```diff
@@ -726,6 +729,16 @@ func createAllIterator(ctx context.Context, primaryIter parquetquery.Iterator, c
 		traceConditions    []traceql.Condition
 	)
 	for _, cond := range conds {
+		// check for existence filter {.a != nil}
+		// and change it to opnone
```
mdisibio, line 733: I was thinking we could do this in the engine around [here](https://github.com/grafana/tempo/blob/main/pkg/traceql/ast_conditions.go#L17) and then maybe it would work without storage-layer changes.  What do you think?

knylander-grafana:
@ie-pham Should there be a doc update to cover this?

ie-pham:
> @ie-pham Should there be a doc update to cover this?

yes! i ran into some git problem and it got lost but I brought it back

docs/sources/tempo/traceql/_index.md
```diff
@@ -151,6 +151,11 @@ Find all traces where the `http.method` attribute is either `GET` or `DELETE`:
 { span.http.method =~ “DELETE|GET” }
 ```
 
+Find all traces where `any_attribute` is not nil or where `any_attribute` exists in a span
```
knylander-grafana, line 154:
```suggestion
Find all traces where `any_attribute` is not `nil` or where `any_attribute` exists in a span:
```

mdisibio (commented):
This is looking good, it's really close.  I found a few edge cases while testing locally that I think would be good to handle before merging.

* We lost the validation error on `.foo = nil` .  Now the query appears to succeed but never returns anything.  Let's see if we can reject this query still.
* `status != nil` or other intrinsics also appear to succeed but never return anything.  I think this kind of query should work, probably if it just passes `InstrinsicStatus/OpNone` to fetch like other attributes.  Maybe just additional cases in the BinaryOperation logic.

pkg/traceql/test_examples.yaml
```diff
@@ -249,8 +249,6 @@ unsupported:
   - '{ 1 >= parent }'
   - '{ -parent = nil }'
   - '{ !parent = nil }'
-  # nil - will be valid when supported
-  - '{ .foo = nil }'
```
mdisibio, line 253: I think this test case should be re-added.

mdisibio (approved):

mdisibio:
LGTM.  Left one comment about reenabling the test for unsupported on = nil, but going ahead and approving.

Issue #2699, opened 2023-07-25
Correct dns entry for memberlist config


joe-elliott:
**Describe the bug**
An issue was discovered in the way a new pod attempted to join a cluster. Details and fix are here:

https://github.com/grafana/loki/pull/9723

We should get this fix in our jsonnet and helm chart.

PR #2700, opened 2023-07-25
microservices: use DNS lookup for memberlist.join_members by default

zalegrala:
**What this PR does**:

Related: https://github.com/grafana/loki/pull/9723

The current default configuration will result in a connection to each member returned by a DNS entry and result in more connections than is necessary.  In large clusters, this can become a problem.

Here we include the fix that Loki has for using a `dns+` lookup feature of memberlist to ensure that only a few members of the DNS entry are added.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/operations/jsonnet-compiled/ConfigMap-tempo-compactor.yaml
+++ b/operations/jsonnet-compiled/ConfigMap-tempo-compactor.yaml
@@ -18,7 +18,7 @@ data:
         abort_if_cluster_join_fails: false
         bind_port: 7946
         join_members:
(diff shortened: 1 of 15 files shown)


joe-elliott (approved):
- Fixes #2699 
- We should also submit this fix to the helm chart
- You need to re-gen the yaml. `make jsonnet` i think?

zalegrala:
Oh nice, thanks for filing an issue.  I just saw the note from Ed, but making an issue was a good call.

PR #2711, opened 2023-07-26
Add several metrics-generator fields to user-configurable overrides

kvrhdn:
**What this PR does**:

Extends the user-configurable overrides added in https://github.com/grafana/tempo/pull/2543 with several metrics_generator_... overrides.

I adopted the nested structure proposed in https://github.com/grafana/tempo/pull/2688

New JSON payload:
```json
{
  "forwarders": [],
  "metrics_generator": {
    "disable_collection": true,
    "processors": ["service-graphs"],
    "processor": {
      "span_metrics": {
        "dimensions": ["env"],
        "enable_target_info": true
      },
      "service_graphs": {
        "dimensions": ["env"],
        "enable_client_server_prefix": true,
        "peer_attributes": ["att"]
      }
    }
  }
}
```

~~Fixes #<issue number>~~

**Checklist**
- [x] Tests updated
- [ ] Documentation added --> TODO still https://github.com/grafana/tempo/issues/2627


modules/overrides/userconfigurableapi/limits.go
```diff
@@ -1,5 +1,110 @@
...
+func (l *UserConfigurableOverridesMetricsGeneratorProcessor) GetSpanMetrics() *UserConfigurableOverridesMetricsGeneratorProcessorSpanMetrics {
+	if l != nil {
+		return l.SpanMetrics
+	}
+	return nil
+}
+
+type UserConfigurableOverridesMetricsGeneratorProcessorServiceGraphs struct {
+	Dimensions     *[]string
+	PeerAttributes *[]string `json:"peer_attributes,omitempty"`
```
domasx2, line 76: we also require `enable_client_server_prefix`

domasx2 (commented):
Also need `service_graphs.enable_client_server_prefix`, otherwise lgtm!

modules/overrides/user_configurable_overrides.go
```diff
@@ -184,13 +184,61 @@ func (o *userConfigurableOverridesManager) setTenantLimit(userID string, limits
...
+
+func (o *userConfigurableOverridesManager) MetricsGeneratorProcessorSpanMetricsDimensions(userID string) []string {
+	if dimensions, ok := o.getTenantLimits(userID).GetMetricsGenerator().GetProcessor().GetSpanMetrics().GetDimensions(); ok {
+		return dimensions
+	}
+	return o.Interface.MetricsGeneratorProcessorSpanMetricsDimensions(userID)
+}
+
+func (o *userConfigurableOverridesManager) MetricsGeneratorProcessorSpanMetricsEnableTargetInfo(userID string) bool {
+	if enableTargetInfo, ok := o.getTenantLimits(userID).GetMetricsGenerator().GetProcessor().GetSpanMetrics().GetEnableTargetInfo(); ok {
```
mdisibio, line 236: It seems things in this call-chain can be nil, like `GetMetricsGenerator()`.  Is that possible?

kvrhdn: Yes, since the overrides can be not set I've made all of the fields pointers. So if you are fetching a nested field every step until that field could be potentially nil.

To make it easier to use I added these helper Get-functions that can deal with a nil receiver (similar to what protobuf generates for you). It's a lot of annoying boilerplate but the nested if-cases would be even worse imo.

mdisibio (approved):

PR #2722, opened 2023-07-28
Queriers: Fix external search config when using RBAC with Google Cloud Run

modulitos:
**What this PR does**:

I implemented a feature enabling RBAC on Google Cloud Run endpoints, which was merged last week: https://github.com/grafana/tempo/pull/2593

But there is a bug with the configuration, as described below. This PR fixes that bug.

**Repro:**

Run Tempo with the following config:
```
querier:
  search:
    external_backend: google_cloud_run
    google_cloud_run:
      external_endpoints: ["https://my-cloud-run-endpoint.run.app"]
```

Open the Grafana Tempo explore panel, and search for traces.
We'll see traces in our results, but they should be coming from our external Cloud Run endpoint, not an internal block search.

**Expected:**

Block queries are sent to the external Cloud Run endpoint.

**Actual:**

An internal search is performed instead of an external search. This is because the `querier.search.external_endpoints` list is unset, so it defaults to an empty list, and is caught by this conditional:
https://github.com/grafana/tempo/blob/41dfbc49e640b45c6e98c92e8889692bf27c53c0/modules/querier/querier.go#L716-L718

We can work around it by defining a non-empty `querier.search.external_endpoints` list:
```
querier:
  search:
    external_endpoints: ["dummy-value"]
    external_backend: google_cloud_run
    google_cloud_run:
      external_endpoints: ["https://my-cloud-run-endpoint.run.app"]
```
but clearly that's not desirable. This PR fixes the conditional and adds a unit test to cover this code path.

**Checklist**
- [x] Tests updated
- [x] Documentation added
  - No documentation changes required
  - Since this is a bug on a pre-release feature, I figured it's not worth updating the changelog. Happy to update it though if requested.

### DIFF
--- a/modules/querier/external/client.go
+++ b/modules/querier/external/client.go
@@ -17,7 +17,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/weaveworks/common/user"
-	"golang.org/x/oauth2"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


modules/querier/external/client.go
```diff
@@ -54,20 +53,13 @@ type Client struct {
 
 type CloudRunConfig struct {
 	Endpoints []string `yaml:"external_endpoints"`
+	NoAuth    bool     // For testing
```
mapno, line 56: Nit: Unexport if only for testing

mapno (approved):
LGTM. 

cc/ @joe-elliott, bugfix to include in v2.2

PR #2755, opened 2023-08-02
Refactor user-configurable overrides API and client; add detailed logging

kvrhdn:
**What this PR does**:

This PR contains two changes:

#### 1. Split up the `userconfigurable` package into `userconfigurable/api` and `userconfigurable/client`

The `userconfigurable` package contained the client to interact with the object store and the API module that serves client requests. The client is used by both the API and by userconfigurableoverrides manager (the process that runs in every instance of Tempo).
Splitting them into pacakges makes it clearer they are both self-contained and allows the API module to grow on its own.

Dependencies:
- `modules/overrides` depends on `modules/overrides/userconfigurable/client`
- `modules/overrides/userconfigurable/api` depends on `modules/overrides/userconfigurable/client` (api can now depend on `modules/overrides` if we want)

#### 2. Clean up the handler of `UserConfigOverridesAPI` and add audit logging

The overrides API will be low volume but it's important to know who changed what, when. I've added more detailed request logging including full request bodies. Traces will also contain the payload so it should be easier to debug issues when they arise.
Examples:

GET
```
level=info ts=2023-08-02T18:01:20.216992Z caller=http.go:163 component=overrides-api traceID=624f8ede50374cc5 method=GET url=/api/overrides
```
Good POST (logs the full request)
```
level=info ts=2023-08-02T18:01:42.336188Z caller=http.go:163 component=overrides-api traceID=61b37e59319a4464 method=POST url=/api/overrides
level=info ts=2023-08-02T18:01:42.336418Z caller=api.go:89 component=overrides-api traceID=61b37e59319a4464 msg="storing user-configurable overrides" userID=single-tenant limits="{\"metrics_generator\":{\"disable_collection\":true}}" version=1
```
Bad POST (logs the error the cient received)
```
level=info ts=2023-08-02T18:01:36.947411Z caller=http.go:163 component=overrides-api traceID=763fb545f8c1499b method=POST url=/api/overrides
level=error ts=2023-08-02T18:01:36.947505Z caller=http.go:171 component=overrides-api traceID=763fb545f8c1499b method=POST url=/api/overrides err="client.Limits.MetricsGenerator: client.LimitsMetricsGenerator.DisableCollection: ReadBool: expect t or f, but found 5, error found in #10 byte of ...|ection\": 5\n  }\n}|..., bigger context ...|\"metrics_generator\": {\n    \"disable_collection\": 5\n  }\n}|..."
```
DELETE
```
level=info ts=2023-08-02T18:01:26.887397Z caller=http.go:163 component=overrides-api traceID=3ad40b007e18012f method=DELETE url=/api/overrides
level=info ts=2023-08-02T18:01:26.887474Z caller=api.go:149 component=overrides-api traceID=3ad40b007e18012f msg="deleting user-configurable overrides" userID=single-tenant version=1
```

To reduce complexity and length of the methods I've moved all HTTP stuff into `http.go` while `api.go` only handles with actual logic now.

#### 3. Fixed bugs related to versioning edge cases

I've updated the `overrides_test.go` in e2e to test each edge case and check Tempo replies with the correct HTTP status code.

~~Fixes #<issue number>~~

**Checklist**
- [x] Tests updated (No changes!)
- [ ] Documentation added


modules/overrides/userconfigurable/client/limits.go
```diff
@@ -0,0 +1,118 @@
+package client
+
+import (
+	"github.com/grafana/tempo/pkg/util/listtomap"
+)
+
+type Limits struct {
+	Forwarders *[]string `json:"forwarders,omitempty"`
```
zalegrala, line 8: Is the pointer to the slice required to achieve the desired result of `omitempty`?  I thought that a slice could be treated as nil.

kvrhdn: There is a slight nuance: `*[]string` can express 3 states, namely not set (`null`), empty (`[]`) and full (`["a"]`).
Example: https://go.dev/play/p/v-bQU9BHIqg

While this usually doesn't matter, for the overrides we want to know when you _don't_ set the overrides because in that case we will use the override from the configmap.

We use this property here:
https://github.com/grafana/tempo/blob/21fe98e8d2e9b3c52b5fa7fab2d3d2c72fe3977f/modules/overrides/user_configurable_overrides.go#L188-L193

So if `forwarders` is not set (= nil) we pass on the call to the runtimemanager (configmap). If it is set to empty slice `[]` we just return that.

zalegrala (approved):
This looks reasonable to me.  I like the separation in here.

PR #2768, opened 2023-08-04
Replace weavework with dskit for recent merge

zalegrala:
**What this PR does**:

Unfortunate merge timing meant that we re-introduced some weaveworks packages which we intended to drop.  Here we make the weaveworks -> dskit transition that was intended.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/frontend/frontend.go
+++ b/modules/frontend/frontend.go
@@ -236,7 +236,7 @@ func newSpanMetricsMiddleware() Middleware {
 
 // buildUpstreamRequestURI returns a uri based on the passed parameters
 // we do this because dskit/common uses the RequestURI field to translate from http.Request to httpgrpc.Request
-// https://github.com/dskit/common/blob/47e357f4e1badb7da17ad74bae63e228bdd76e8f/httpgrpc/server/server.go#L48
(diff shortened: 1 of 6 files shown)


modules/frontend/frontend.go
```diff
@@ -235,8 +235,8 @@ func newSpanMetricsMiddleware() Middleware {
 }
 
 // buildUpstreamRequestURI returns a uri based on the passed parameters
-// we do this because weaveworks/common uses the RequestURI field to translate from http.Request to httpgrpc.Request
-// https://github.com/weaveworks/common/blob/47e357f4e1badb7da17ad74bae63e228bdd76e8f/httpgrpc/server/server.go#L48
+// we do this because dskit/common uses the RequestURI field to translate from http.Request to httpgrpc.Request
+// https://github.com/dskit/common/blob/47e357f4e1badb7da17ad74bae63e228bdd76e8f/httpgrpc/server/server.go#L48
```
kvrhdn, line 239: I think this should be the correct line
```suggestion
// https://github.com/grafana/dskit/blob/740f56bd293423c5147773ce97264519f9fddc58/httpgrpc/server/server.go#L59
```

kvrhdn (approved):
Looks okay to me. If CI is happy I'm happy 🙂

PR #2775, opened 2023-08-09
Cliquery

ie-pham:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


cmd/tempo-cli/cmd-query-blocks-summary.go
```diff
@@ -0,0 +1,135 @@
...
+	for i, result := range results {
+		combiner.ConsumeWithFinal(result.trace, i == len(results)-1)
+	}
+
+	combinedTrace, _ := combiner.Result()
+	size := 0
+	spanCount := 0
+	firstStartTime := combinedTrace.Batches[0].ScopeSpans[0].Spans[0].StartTimeUnixNano
+	lastEndTime := combinedTrace.Batches[0].ScopeSpans[0].Spans[0].EndTimeUnixNano
+	rootSpan := combinedTrace.Batches[0].ScopeSpans[0].Spans[0]
```
joe-elliott, line 60: the root span is not always the first. as you're iterating through all the spans below look for one with no parent span id

cmd/tempo-cli/cmd-query-blocks-summary.go
```diff
@@ -0,0 +1,135 @@
...
+		size += b.Size()
+		for _, attr := range b.Resource.Attributes {
+			if "service.name" == attr.Key {
+				serviceNameMap[attr.Value.GetStringValue()] ++
+			}
+		}
+		for _, scope := range b.ScopeSpans {
+			spanCount += len(scope.Spans)
+			for _, span := range scope.Spans {
+				if span.EndTimeUnixNano > lastEndTime {
```
joe-elliott, line 80: need to do the math for `firstStartTime` as well

cmd/tempo-cli/cmd-query-blocks-summary.go
```diff
@@ -0,0 +1,135 @@
...
+					lastEndTime = span.EndTimeUnixNano
+				}
+			}
+		}
+	}
+
+
+	// get top 5 most frequent service names
+	lowest := 0
+	topFiveName := make([]string, 5)
```
joe-elliott, line 90: i would probably just put all of the key/value pairs in a slice and sort it. it won't be as efficient, but it will be clearer and this CLI command doesn't need to be super efficient.

cmd/tempo-cli/cmd-query-blocks-summary.go
```diff
@@ -0,0 +1,143 @@
...
+		for _, scope := range b.ScopeSpans {
+			spanCount += len(scope.Spans)
+			for _, span := range scope.Spans {
+				if span.StartTimeUnixNano < firstStartTime {
+					firstStartTime = span.StartTimeUnixNano
+				}
+				if span.EndTimeUnixNano > lastEndTime {
+					lastEndTime = span.EndTimeUnixNano
+				}
+				if span.ParentSpanId == nil {
```
joe-elliott, line 75: might want to check `len() == 0`

cmd/tempo-cli/cmd-query-blocks-summary.go
joe-elliott: instead of grabbing this first span like this we should init firstStartTime/lastEndTime/rootSpan/rootResource to better first values. something like this diff:

```
@@ -4,11 +4,14 @@ import (
        "bytes"
        "context"
        "fmt"
+       "math"
        "sort"
 
        "github.com/gogo/protobuf/jsonpb"
 
+       v1resource "github.com/grafana/tempo/pkg/tempopb/resource/v1"
+       v1 "github.com/grafana/tempo/pkg/tempopb/trace/v1"
        "github.com/grafana/tempo/pkg/util"
 )
 
@@ -47,15 +50,16 @@ func (cmd *queryBlocksSummaryCmd) Run(ctx *globalOptions) error {
        }
 
        combinedTrace, _ := combiner.Result()
+
+       var rootSpan *v1.Span
+       var rootSpanResource *v1resource.Resource
        size := 0
        spanCount := 0
-       firstStartTime := combinedTrace.Batches[0].ScopeSpans[0].Spans[0].StartTimeUnixNano
-       lastEndTime := combinedTrace.Batches[0].ScopeSpans[0].Spans[0].EndTimeUnixNano
-       rootSpan := combinedTrace.Batches[0].ScopeSpans[0].Spans[0]
-       rootSpanResource := combinedTrace.Batches[0].Resource
+       firstStartTime := uint64(math.MaxUint64)
+       lastEndTime := uint64(0)
+
        rootServiceName := ""
        serviceNameMap := make(map[string]int)
-
        for _, b := range combinedTrace.Batches {
                size += b.Size()
                for _, attr := range b.Resource.Attributes {
```

joe-elliott (commented):
heading the right way. we need:

- changelog
- docs

i also don't like the command name, but it does nicely match the query blocks command.

knylander-grafana:
@ie-pham This looks like a useful tool. Should we add some doc for this?

cmd/tempo-cli/cmd-query-trace-summary.go
```diff
@@ -0,0 +1,150 @@
...
+					rootSpan = span
+					rootSpanResource = b.Resource
+				}
+			}
+		}
+	}
+
+	for _, attr := range rootSpanResource.Attributes {
+		if "service.name" == attr.Key {
+			rootServiceName = attr.Value.GetStringValue()
```
joe-elliott, line 92: nit: can add a break statement

cmd/tempo-cli/cmd-query-trace-summary.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+	}
+
+	// get top 5 most frequent service names
+	topFiveSortedPL := sortServiceNames(serviceNameMap)
+	topFiveServiceName := make([]string, 5)
+	length := len(topFiveSortedPL)
```
joe-elliott, line 99: what if length is less than 5? i think this panics

ie-pham: since we are iterating through the actual list of service names using its length (or if it's longer than 5 use 5), even if it's less than 3, we wouldn't get to `out of range`.  It just prints out a weird looking array with extra spaces

`[a b c      ]`

joe-elliott (approved):

PR #2781, opened 2023-08-10
Avoid index errors for deleted tenants

zalegrala:
**What this PR does**:

Here we ensure that when Tempo tries to delete a tenant index that has already been deleted, that this is not an error condition, and doesn't not log or count against the metrics.  This can happen when an index was deleted because no blocks have been found, but for block paths that still contain spurious objects.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * [BUGFIX] Fix panic in metrics summary api [#2738](https://github.com/grafana/tempo/pull/2738) (@mdisibio)
 * [BUGFIX] Fix node role auth IDMSv1 [#2760](https://github.com/grafana/tempo/pull/2760) (@coufalja)
 * [BUGFIX] Only search ingester blocks that fall within the request time range. [#2783](https://github.com/grafana/tempo/pull/2783) (@joe-elliott)
+* [BUGFIX] Fix incorrect metrics for index failures [#2781](https://github.com/grafana/tempo/pull/2781) (@zalegrala)
 
 ## v2.2.0 / 2023-07-31
 
--- a/tempodb/backend/raw.go
+++ b/tempodb/backend/raw.go
@@ -91,6 +91,7 @@ func (w *writer) CloseAppend(ctx context.Context, tracker AppendTracker) error {
 func (w *writer) WriteTenantIndex(ctx context.Context, tenantID string, meta []*BlockMeta, compactedMeta []*CompactedBlockMeta) error {
 	// If meta and compactedMeta are empty, call delete the tenant index.
 	if len(meta) == 0 && len(compactedMeta) == 0 {
+		// Skip returning an error when the object is already deleted.
 		if err := w.w.Delete(ctx, TenantIndexName, KeyPath([]string{tenantID}), false); err != nil && err != ErrDoesNotExist {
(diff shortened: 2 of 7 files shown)


zalegrala:
It does not appear to be an error in s3 when you delete an object that is not present, so the s3 implementation has not been updated.

tempodb/backend/raw.go
```diff
@@ -91,7 +91,10 @@ func (w *writer) CloseAppend(ctx context.Context, tracker AppendTracker) error {
 func (w *writer) WriteTenantIndex(ctx context.Context, tenantID string, meta []*BlockMeta, compactedMeta []*CompactedBlockMeta) error {
 	// If meta and compactedMeta are empty, call delete the tenant index.
 	if len(meta) == 0 && len(compactedMeta) == 0 {
-		return w.w.Delete(ctx, TenantIndexName, KeyPath([]string{tenantID}), false)
+		if err := w.w.Delete(ctx, TenantIndexName, KeyPath([]string{tenantID}), false); err != nil && err != ErrDoesNotExist {
```
joe-elliott, line 94: maybe more simple?

```
if err := delete(); err == ErrDoesNotExist {
  // eat DNE error on delete
  return nil
}
return err
```

tempodb/blocklist/poller.go
```diff
@@ -216,7 +216,7 @@ func (p *Poller) pollTenantAndCreateIndex(ctx context.Context, tenantID string)
 	// everything is happy, write this tenant index
 	level.Info(p.logger).Log("msg", "writing tenant index", "tenant", tenantID, "metas", len(blocklist), "compactedMetas", len(compactedBlocklist))
 	err = p.writer.WriteTenantIndex(derivedCtx, tenantID, blocklist, compactedBlocklist)
-	if err != nil {
+	if err != nil && err != backend.ErrDoesNotExist {
```
joe-elliott, line 219: this is handled one layer above. we shouldn't handle it twice.

tempodb/backend/raw.go
zalegrala: Simpler, but doesn't compile.

zalegrala: Added a comment about eating the error.

joe-elliott (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * [BUGFIX] Fix panic in metrics summary api [#2738](https://github.com/grafana/tempo/pull/2738) (@mdisibio)
 * [BUGFIX] Fix node role auth IDMSv1 [#2760](https://github.com/grafana/tempo/pull/2760) (@coufalja)
 * [BUGFIX] Only search ingester blocks that fall within the request time range. [#2783](https://github.com/grafana/tempo/pull/2783) (@joe-elliott)
+* [BUGFIX] Fix incorrect metrics for index failures [#2781](https://github.com/grafana/tempo/pull/2781) (@zalegrala)
```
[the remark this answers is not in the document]
zalegrala: Not that I'm thinking of.  We still leave the spurious objects in the backend, so perhaps we can have a conversation about cleaning that up also.

Issue #2785, opened 2023-08-11
Metrics-generator in 2.2 panics with: `fatal error: concurrent map iteration and map write`


kvrhdn:
**Describe the bug**
Reported in community Slack: https://grafana.slack.com/archives/C01D981PEE5/p1691583814497569

> Hi, we upgraded our production cluster to tempo v2.2.0 today and since then we have been having sporadic Metric generator restarts with the error fatal error: concurrent map iteration and map write. Any idea what might be causing this? Our staging cluster is running no problem and when we downgraded back to 2.1.1 it ran fine again. We are using a helm chart to run

Stack trace
```
github.com/grafana/tempo/modules/generator.(*instance).diffProcessors(_, _, {{0x2540be400, 0x2710, 0xa, {0xc0001f1600, 0x8, 0x8}, {0x0, 0x0, ...}, ...}, ...})
 Aug 09 12:33:54 54.246.15.219 tempo[tempo-tempo-distributed-metrics-generator- 	/drone/src/modules/generator/instance.go:229 +0x6c
 Aug 09 12:33:54 54.246.15.219 tempo[tempo-tempo-distributed-metrics-generator- github.com/grafana/tempo/modules/generator.(*instance).updateProcessors(0xc0e3ae1950)
 Aug 09 12:33:54 54.246.15.219 tempo[tempo-tempo-distributed-metrics-generator- 	/drone/src/modules/generator/instance.go:190 +0x21e
 Aug 09 12:33:54 54.246.15.219 tempo[tempo-tempo-distributed-metrics-generator- github.com/grafana/tempo/modules/generator.(*instance).watchOverrides(0xc0e3ae1950)
 Aug 09 12:33:54 54.246.15.219 tempo[tempo-tempo-distributed-metrics-generator- 	/drone/src/modules/generator/instance.go:123 +0xf6
 Aug 09 12:33:54 54.246.15.219 tempo[tempo-tempo-distributed-metrics-generator- created by github.com/grafana/tempo/modules/generator.newInstance
 Aug 09 12:33:54 54.246.15.219 tempo[tempo-tempo-distributed-metrics-generator- 	/drone/src/modules/generator/instance.go:109 +0x405
 Aug 09 12:33:54 54.246.15.219 tempo[tempo-tempo-distributed-metrics-generator- goroutine 1 [select, 48 minutes]:
[...]

PR #2786, opened 2023-08-11
Fix panic in the metrics-generator when using multiple tenants with default overrides

kvrhdn:
**What this PR does**:
We have a concurrent read/write panic while reading from `desiredProcessors`, this is the only place we modify it really. Let's copy it before modifying it.

Fixes https://github.com/grafana/tempo/issues/2785

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/instance.go
+++ b/modules/generator/instance.go
@@ -12,6 +12,7 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"golang.org/x/exp/maps"
 
 	"github.com/grafana/tempo/modules/generator/processor"
 	"github.com/grafana/tempo/modules/generator/processor/localblocks"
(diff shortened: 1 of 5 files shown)


modules/generator/instance.go
```diff
@@ -151,8 +152,11 @@ func (i *instance) updateSubprocessors(desiredProcessors map[string]struct{}, de
 	_, latencyOk := desiredProcessors[spanmetrics.Latency.String()]
 	_, sizeOk := desiredProcessors[spanmetrics.Size.String()]
 
+	newDesiredProcessors := map[string]struct{}{}
+	maps.Copy(newDesiredProcessors, desiredProcessors)
```
joe-elliott, line 156: if desiredProcessors was capable of racing in the previous code, I don't see how this fixes it. it could still race here while copying to newDesiredProcessors

kvrhdn: This only reads the map, afaik concurrent map reads are fine. Later in this function we would modify `desiredProcessors` but with this change we only modify the copy `newDesiredProcessors`.

kvrhdn:
The new test fails on a data race ☝🏻

joe-elliott (approved):

PR #2790, opened 2023-08-15
Add trace quality metrics

### LINKED ISSUE #2749: Add trace quality metrics
We need metrics that helps operators determine the "quality" of their trace ingestion. We will start with metric'ing broken traces at 3 points:

- Live Traces -> WAL
- WAL -> Complete Block (when combining traces)
- Compaction (when combining traces)

The latter two require some thought. We want to make sure that the metric is clear in what it communicates. Counting broken traces only of traces that require combining could be misleading.

joe-elliott:
**What this PR does**:
Counts disconnected traces in two "phases" of a trace's lifetime. The first is when the trace is cut from live traces to the head block. The second is when the trace is combined in the compactor. These counters should be combined with other metrics to get a percentage of traces disconnected:

```
rate(tempo_warnings_total{reason="disconnected_trace_flushed_to_wal"}[1m]) /
rate(tempo_ingester_traces_created_total{}[1m])
```
```
rate(tempo_warnings_total{reason="disconnected_trace_compactor_combine"}[1m]) /
rate(tempodb_compaction_objects_combined_total{}[1m])
```

*Thoughts*
- The first metric percentage above isn't exact. "Traces created" is counted when a live trace is created. The warning is counted when a live trace is flushed to the wal.
- There is an opportunity to also record this metric when we flush from wal -> complete block. See comments in code for why it is quite difficult to count in this case.
- This is implemented only on vParquet3 for simplicity. This should soon be our default block version.
- Found and fixed a small issue in `assignNestedSetModelBounds`. Added tests.

cc @stoewer This touches the `assignNestedSetModelBounds` function. If you have a chance to review, it would be appreciated.

Fixes #2749 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/encoding/vparquet3/nested_set_model.go
+++ b/tempodb/encoding/vparquet3/nested_set_model.go
@@ -15,7 +15,8 @@ type spanNode struct {
 
 // assignNestedSetModelBounds calculates and assigns the values Span.NestedSetLeft, Span.NestedSetRight,
 // and Span.ParentID for all spans in a trace.
-func assignNestedSetModelBounds(trace *Trace) {
+// Returns true if the trace tree is a connected graph which is useful for calculating data quality
+func assignNestedSetModelBounds(trace *Trace) bool {
 	// count spans in order be able to pre-allocate tree nodes
 	var spanCount int
 	for _, rs := range trace.ResourceSpans {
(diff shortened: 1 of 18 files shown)


tempodb/encoding/vparquet3/nested_set_model.go
```diff
@@ -45,7 +46,8 @@ func assignNestedSetModelBounds(trace *Trace) {
 				id := util.SpanIDToUint64(s.SpanID)
 				if nodes, ok := nodesByID[id]; ok {
 					// zipkin traces may contain client/server spans with the same IDs
-					nodesByID[id] = append(nodes, node)
+					nodes = append(nodes, node)
+					nodesByID[id] = nodes
```
stoewer, line 50: Good catch!

stoewer (approved):
LGTM
Just a small comment about documenting the new parameter

PR #2791, opened 2023-08-15
Add RegisterFlagsAndApplyDefault function to every backend config

kvrhdn:
**What this PR does**:
We configure a lot of default settings for each tempodb backend in `modules/storage/config.go`. The user-configurable overrides now also uses the backends and doesn't use these defaults.

Move the defaults from `modules/storage` into each backend so both storage and user-configurable overrides can use them.

~~Fixes #<issue number>~~

Fixes deployments on Azure, it sets `endpoint_suffix` to `blob.core.windows.net` which is needed to get Azure working.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


zalegrala (approved):
Nice cleanup.

PR #2799, opened 2023-08-17
Upgrade dskit and improve log.InitLogger function

duricanikolic:
**What this PR does**:
This PR upgrades `dskit` and uses `dskit`'s `log` package for the creation of the global logger.
For the main in the `dskit` see https://github.com/grafana/dskit/pull/359.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/util/log/log.go
```diff
@@ -6,51 +6,24 @@ import (
...
-	}
+// InitLogger initialises the global gokit logger and returns that logger.
+func InitLogger(logFormat string, logLevel dslog.Level) kitlog.Logger {
+	writer := kitlog.NewSyncWriter(os.Stderr)
+	logger := dslog.NewGoKitWithWriter(logFormat, writer)
 
-	// add support for level based logging
-	logger = level.NewFilter(logger, LevelFilter(cfg.LogLevel.String()))
+	// use UTC timestamps and skip 5 stack frames.
+	logger = kitlog.With(logger, "ts", kitlog.DefaultTimestampUTC, "caller", kitlog.Caller(5))
```
zalegrala, line 22: What does it mean to change the call stack depth here?  Do we now report a different line is doing the logging?

zalegrala (commented):
This looks good to me.  Just one question about the stack depth change.

pkg/util/log/log.go
duricanikolic: Hi @zalegrala,
The reason for that is the implementation of `dslog.NewGoKitWithWriter()` that is now used to create the actual logger, and that replaces the “manual” creation that was used before.
I did a manual test to check the outcome and it prints tve same source and line it used to print before my change.
I will prepare a Unit test that ensures the right thing is printed once I am back from PTO. I will also replace the usage of `RateLimitedLogger` with the one from `dskit`.

PR #2803, opened 2023-08-17
Configurable route http to grpc

joe-elliott:
**What this PR does**:
Disables the grpc/streaming endpoint routing over http by default. This was breaking TLS. Adds a new setting allowing the operator to turn on streaming over HTTP along with some docs.

This new setting:
```
stream_over_http_enabled: false
```
can be used as long as TLS is not needed on the query-frontend. This is common in scenarios were TLS is terminated in front of Tempo.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/cmd/tempo/app/config.go
+++ b/cmd/tempo/app/config.go
@@ -33,6 +33,7 @@ type Config struct {
 	Target                       string `yaml:"target,omitempty"`
 	AuthEnabled                  bool   `yaml:"auth_enabled,omitempty"`
 	MultitenancyEnabled          bool   `yaml:"multitenancy_enabled,omitempty"`
+	StreamOverHTTPEnabled        bool   `yaml:"stream_over_http_enabled,omitempty"`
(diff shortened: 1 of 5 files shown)


mdisibio:
I like the new setting name, it's nice and clear.   Fwiw `DoNotRouteHTTPToGRPC` was never exposed to the config yaml or CLI options so we should be able to remove it and then fix the upstream application references.   Then we would be left with just this new setting and it's a lot clearer.

joe-elliott:
> Fwiw DoNotRouteHTTPToGRPC was never exposed to the config yaml or CLI options so we should be able to remove it and then fix the upstream application references. Then we would be left with just this new setting and it's a lot clearer.

I did it this way on purpose so we wouldn't have to make any upstream changes. If you're willing to just drop `DoNotRouteHTTPToGRPC` I'd be glad to.

mdisibio:
> drop `DoNotRouteHTTPToGRPC`

👍 Yes let's drop it.

mdisibio (approved):

PR #2819, opened 2023-08-21
Fix tools package listing on Go 1.21

zalegrala:
**What this PR does**:

Without this change, the `go list` used to track with tools are required by the project is broken in Go 1.21, resulting in the following error.

```
tools.go:6:2: import "github.com/psampaz/go-mod-outdated" is a program, not an importable package
```

Here we add `-e` to the `go list` to allow the list to complete and handle the tool management as before.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott (approved):
man, go 1.21 is just going to give us hell, isn't it?

zalegrala:
Yeah, probably.

PR #2826, opened 2023-08-22
Fix incorrectly reported vulture errors

### LINKED ISSUE #2792: Vulture incorrectly reports failures since r107
**Describe the bug**
Since installing r107 the vulture is reporting errors for trace not found and incorrect results. After rolling back to the r106 vulture the errors went away. This strongly suggests that the data is fine and some change in vulture/the http client caused the issue:

image

Fix the vulture!

stoewer:
Fix vulture problems:
* Since #2649 the function `RandomAttrFromTrace()` may return attributes with resource scope. This caused several problems in vulture:
  * When searching for `service.name` the matching traces can exceed the limit and the relevant trace could be missing from the result
    *Solution: skip the `service.name` attribute when selecting random attributes*
  * Vulture's `searchTraceql()` function was only searching through span attributes
    *Solution: change query template from `span.%s="%s"` to `.%s="%s"`*
  
* The order of attributes in generated data can differ from the order of attributes in traces retrieved from the API.
  *Solution: sort attributes before testing traces for equality*

Fixes #2792

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/util/trace_info.go
+++ b/pkg/util/trace_info.go
@@ -265,7 +265,7 @@ func RandomAttrFromTrace(t *tempopb.Trace) *v1common.KeyValue {
 
 	// maybe choose resource attribute
 	res := batch.Resource
-	if len(res.Attributes) > 0 && rand.Int()%2 == 1 {
+	if len(res.Attributes) > 0 && r.Int()%2 == 1 {
 		return randFrom(r, res.Attributes)
 	}
 
--- a/pkg/util/trace_info.go
+++ b/pkg/util/trace_info.go
@@ -266,7 +266,12 @@ func RandomAttrFromTrace(t *tempopb.Trace) *v1common.KeyValue {
 	// maybe choose resource attribute
 	res := batch.Resource
 	if len(res.Attributes) > 0 && r.Int()%2 == 1 {
-		return randFrom(r, res.Attributes)
+		attr := randFrom(r, res.Attributes)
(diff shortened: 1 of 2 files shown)


pkg/util/trace_info.go
```diff
@@ -265,8 +265,13 @@ func RandomAttrFromTrace(t *tempopb.Trace) *v1common.KeyValue {
 
 	// maybe choose resource attribute
 	res := batch.Resource
-	if len(res.Attributes) > 0 && rand.Int()%2 == 1 {
-		return randFrom(r, res.Attributes)
+	if len(res.Attributes) > 0 && r.Int()%2 == 1 {
+		attr := randFrom(r, res.Attributes)
+		// skip service.name because service names have low cardinality and produce queries with
+		// too many results in tempo-vulture
+		if attr.Key != "service.name" {
```
mapno, line 272: Should we also skip other attributes such as `namespace`, `cluster`?

joe-elliott (approved):
I'm good on this, but do what change caused vulture to start failing?

stoewer:
> but do what change caused vulture to start failing?

The vulture problems were introduced by #2649. Before vulture was only testing span attributes. Since dedicated attribute columns are supported on resource and span scope, the test data was changed such that it also includes a resource attributes. This caused all the problems mentioned in the PR description

PR #2840, opened 2023-08-24
Rework SLO metrics

joe-elliott:
**What this PR does**:
This PR reworks the way our SLO metrics are recorded with these goals in mind;

- Push the calculations out to be as close the http.Server as possible
- Clearly metric both total queries and SLO queries in one spot
- Cleanup and remove unused metrics

This PR accomplishes the above with the following sharp corners:

- In order to pass the throughput up to the code that calculates the SLOs a *float was added to the context. This works, but is brittle and difficult to detect if it breaks.
- Specialized code needed to be added to the streaming endpoint

TODO
- [x] correct changelog
- [x] add tests

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/slos.go
+++ b/modules/frontend/slos.go
@@ -69,20 +69,25 @@ func sloHook(allByTenantCounter, withinSLOByTenantCounter *prometheus.CounterVec
 			return
 		}
 
-		passedThroughput := true
+		passedThroughput := false
 		// final check is throughput
 		if cfg.ThroughputBytesSLO > 0 {
 			throughput, ok := thoughputFromContext(ctx)
 
 			// if we didn't find the key, but expected it, we consider throughput a failure
-			passedThroughput = !ok && throughput >= cfg.ThroughputBytesSLO
+			passedThroughput = ok && throughput >= cfg.ThroughputBytesSLO
 		}
 
-		passedDuration := cfg.DurationSLO == 0 || latency < cfg.DurationSLO
+		passedDuration := false
+		if cfg.DurationSLO > 0 {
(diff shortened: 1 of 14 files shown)


modules/frontend/slos.go
```diff
@@ -0,0 +1,119 @@
...
+	"net/http"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+const (
+	traceByIDOp = "traces"
+	searchOp    = "search"
```
zalegrala, line 14: Do we want to track SLOs on the other frontend operation?  Looks like we lost a couple.

zalegrala (commented):
This looks reasonable to me.  Just a question about tracking the other operations.  Have you had a chance to test this out?

modules/frontend/slos.go
joe-elliott: We currently don't support "policy based" SLOs for those other ops and just use `tempo_request_duration_seconds` histograms. Only these two operations were using these metrics for SLOs so I removed the other two to align `tempo_query_frontend_queries_total` and `tempo_query_frontend_queries_within_slo_total`. Previously, `tempo_query_frontend_queries_total` had more values for the `op` label then `tempo_query_frontend_queries_within_slo_total` and I wanted this pair of metrics to be aligned.

mdisibio (approved):
LGTM, accuracy of the metrics for SLO calculation is important and I think this is a good change.  I think this is Tempo's first case of context smuggling so it gets a pass. 😄

zalegrala (approved):
Nice work Joe.  Thanks for giving this area some :heart: .

PR #2854, opened 2023-08-26
WIP: Fix1858:  Ability to search all tenants in multi-tenant set up

### LINKED ISSUE #1858: Ability to search all tenants in multi-tenant setups
I'd like the ability to search all tenants at once in a multi-tenant setup. At least when pulling a trace by trace ID. We have multiple departments who communicate with each other, and each department is set up as its own tenant. If a trace goes from tenant A->B->A, it will appear as a broken trace when querying tenant A and as a root span missing trace when querying tenant B.

mghildiy:
**Adds ability to search all tenants in multi-tenant setup**:

Fixes #1858

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/config.go
+++ b/modules/frontend/config.go
@@ -16,10 +16,11 @@ import (
 var statVersion = usagestats.NewString("frontend_version")
 
 type Config struct {
-	Config     v1.Config       `yaml:",inline"`
-	MaxRetries int             `yaml:"max_retries,omitempty"`
-	Search     SearchConfig    `yaml:"search"`
-	TraceByID  TraceByIDConfig `yaml:"trace_by_id"`
+	Config                    v1.Config       `yaml:",inline"`
+	MaxRetries                int             `yaml:"max_retries,omitempty"`
+	Search                    SearchConfig    `yaml:"search"`
+	TraceByID                 TraceByIDConfig `yaml:"trace_by_id"`
+	MultitenantQueriesEnabled bool            `yaml:"multitenant_queries_enabled"`
 }
 
 type SearchConfig struct {
@@ -53,6 +54,7 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
 	cfg.Config.MaxOutstandingPerTenant = 2000
 	cfg.Config.MaxBatchSize = 5
 	cfg.MaxRetries = 2
+	cfg.MultitenantQueriesEnabled = false
 	cfg.Search = SearchConfig{
 		Sharder: SearchSharderConfig{
 			QueryBackendAfter:     15 * time.Minute,
--- a/modules/frontend/crosstenant.go
+++ b/modules/frontend/crosstenant.go
@@ -1,0 +1,114 @@
+package frontend
+
+import (
+	"io"
+	"net/http"
+	"strings"
+	"sync"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/grafana/dskit/tenant"
+	"github.com/grafana/dskit/user"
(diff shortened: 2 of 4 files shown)


mghildiy:
Its WIP. Test part is pending

modules/frontend/tracebyidtenantparsing.go
```diff
@@ -0,0 +1,169 @@
...
+	})
+}
+
+type parseQuery struct {
+	next   http.RoundTripper
+	cfg    *TraceByIDConfig
+	logger log.Logger
+}
+
+func (p parseQuery) RoundTrip(r *http.Request) (*http.Response, error) {
```
joe-elliott, line 38: we will need this same functionality on all paths: tags, metrics, trace by id and search. i think we should try to implement this as a generic piece of middleware.

the middleware would define a "combiner" interface that it used to recombine the incoming http requests. wdyt?

mghildiy: Ok. So every path then would provide implementation for this new combiner interface as per path's requirements.

joe-elliott: yes, but tbh i don't think this is possible atm on the metrics endpoints. maybe just return an error there for now since it's experimental?

mghildiy: I check that search and trace by id have diff ways of processing requests and combining the responses. So another approach may be to pass this behaviour as a function to the middleware.

mghildiy:
```
type Combiner interface {
	Consume(input any) (count int)
	Result(statusCode int, statusMsg string) (*http.Response, error)
}
```
This is the Combiner interface I plan to have.

Consume is for merging responses of individual requests.
Result is for putting together final response based on merge result(and final statuscode, status message after all requests are processed).

joe-elliott: - `Consume()` should take an `*http.Response`. the middleware should just give it all responses it receives and rely on the interface's implementation to work out the details
- Does result need the status code/message? These values will be in each consumed http.response?

I'm thinking like:

```
type Combiner interface {
	Consume(*http.Response)
	Result() (*http.Response, error)
}
```

as discussed out of band, let's:

1) make this configurable (off by defafult)
2) only worry about the trace by id path right now

if we establish a good generic set of code we can add the others later.

mghildiy: As consuming a response may result into an error too(like when reading response body or during unmarshalling the body), so Consume must return error, and then maybe a we call it Process.

`Process(*http.Response) error`

joe-elliott: I'm fine with either Process or Result. I think they are clear enough

mghildiy: Added a cross tenant middleware which takes combiner as input, and added one such combiner for traces

electron0zero:
Thank you, @mghildiy for putting in time into this feature. This feature requires significant efforts and we are thankful for your time and efforts you have put into this. 

As mentioned [here](https://github.com/grafana/tempo/issues/1858#issuecomment-1631457869), We have decided to take this feature from our enterprise version(GET) and move it into OSS Tempo. 

We believe that it's better if someone from the team takes our enterprise code and moves it into OSS tempo. It will avoid the duplicate efforts. 

I am working on the PR to move it from Enterprise to OSS and will have a PR out soon.

Thank you for the contribution ❤️

electron0zero:
update: multi-tenant search support was added via https://github.com/grafana/tempo/pull/3087 & https://github.com/grafana/tempo/pull/3262

PR #2860, opened 2023-08-28
Remove Grafana Cloud from metrics-gen arch diagram

knylander-grafana:
**What this PR does**:

Removes Grafana Cloud from the metrics-generator architecture diagram. This diagram is used here: https://grafana.com/docs/tempo/latest/metrics-generator/

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [X] Documentation added


09jvilla (approved):
LGTM!

Any promtheus compatible TSDB would work (so not just Mimir, but also Thanos, Prometheus, Victoria Metrics, etc...), but I think its fine to just leave the diagram as is. I believe our actual content already says 'any prometheus compatible tsdb'.

joe-elliott (approved):

Issue #2671, opened 2023-07-18
Query frontend should substitute "<root span not yet received>"


joe-elliott:
Currently the querier substitutes "<root span not yet received>" only on results that come from the ingester. We should move this substitution up to be done in query frontend so it applies on backend blocks as well.

Additionally, a bit of research to see if we can distinguish between a root span with an empty name and no root span should be conducted.

PR #2866, opened 2023-08-29
substitute empty root span in query frontend

galalen:
**What this PR does**:
This PR moved the substitution of empty root span to the query frontend, to be applied on results coming from backend blocks.
Fixes [#2671](https://github.com/grafana/tempo/issues/2671)

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ defaults:
   metrics_generator:
     processors: [service-graphs, span-metrics]
 ```  
+* [BUGFIX] Moved empty root span substitution from `querier` to `query-frontend`. [#2671](https://github.com/grafana/tempo/issues/2671) (@galalen)
 
 # v2.2.2 / 2023-08-30
 
--- a/modules/frontend/search_progress.go
+++ b/modules/frontend/search_progress.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"sync"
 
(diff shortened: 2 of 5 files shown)


joe-elliott:
This PR looks good, but seeing some failures in on the e2e suite. You can run it locally doing `make test-e2e`. It can also be easier to run just one test sometimes. To do that run `make docker-compose` and then cd into `./integration/e2e` and run `go test -run <test name i care about>`

Thanks for looking into this, if you'd like any help debugging the e2e tests, let me know.

galalen:
> This PR looks good, but seeing some failures in on the e2e suite. You can run it locally doing `make test-e2e`. It can also be easier to run just one test sometimes. To do that run `make docker-compose` and then cd into `./integration/e2e` and run `go test -run <test name i care about>`
> 
> Thanks for looking into this, if you'd like any help debugging the e2e tests, let me know.

Awesome, I'll be looking into it.

joe-elliott (commented):
This PR is looking good. I think you found the reason integration tests are failing and fixed it, but it looks like changes from 2 other PRs somehow snuck in here? Can we remove those changes? I noted the first one, but then realized there were quite a few.

After we clean up the unnecessary diffs, this should be good to merge.

joe-elliott (approved):
Nice fix!

Issue #1903, opened 2022-11-21
Min/Max ID occasionally wrong on a block


joe-elliott:
Blocks have been observed with a recorded min and max id that does not correctly reflect the range of ids in the block. This causes trace by id search to occasionally 404 when the trace is actually present in the system.

- This has only been observed on blocks with level > 0, but it it is likely this also occurs on level 0 blocks.
- Best guess is that this occurs when an ingester rolls, replays the WAL and rebuilds the meta.

mdisibio:
We no longer write min/max IDs to the block, because it isn't helpful because each block effectively has IDs across the entire 000...FFF range.

PR #2867, opened 2023-08-29
Fix correctness of block min/max IDs

mdisibio:
**What this PR does**:
Tempo has had a long-standing bug where the min/max trace IDs in meta.json were wrong.  These fields were no longer read in https://github.com/grafana/tempo/pull/1904 and then this issue surfaced again in https://github.com/grafana/tempo/pull/2697.  I think this PR fixes it and the root issue is that the byte slice returned by the raw iterator is still attached to the underlying parquet row which gets reused through pooling.   Try as I might, I wasn't able to reproduce this with a test or running locally, only by trying it out in a medium-sized staging cluster.    Marked as draft for now because I would really like to cover it with a test.

Fixes n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott (approved):
Fixes https://github.com/grafana/tempo/issues/1903

PR #2869, opened 2023-08-30
Move `jsonnetfmt`, `jsonnet`, `goimports` and others install into `tools/`

zalegrala:
**What this PR does**:

Here we create a new docker image to use within CI and locally to ensure that
we are using the same version of `jsonnetfmt`, `jsonnet` and `goimports` that
we are using in CI.  This will ensure that the version we are using to compile
locally is the same version that we are using in CI.  The `goimports` was
already included in the tools, and so the CI was redundant.  Also not, this
moves us from the C jsonnet to the Go jsonnet, which looks to introduce slight
formatting changes.

This is somewhat forward looking in that we may wish to push yet further CI
tools into this image and begin the process of backing them out of github
actions.

**Checklist**
- [x] CI updated
- [ ] Tests updated
- [ ] Documentation added


build/tools.mk
```diff
@@ -11,8 +11,26 @@ GO_MOD_OUTDATED  ?= go-mod-outdated
 TOOL_DIR     ?= tools
 TOOL_CONFIG  ?= $(TOOL_DIR)/tools.go
 
+TOOLS_IMAGE ?= zalegrala/tempo-ci-tools
```
zalegrala, line 14: Once we have an image published, we can switch to using it.  This PR creates the image publication as well as the implementation.  We can break this into two PRs if desired, but I thought leaving it as one gave a more holistic picture of what is happening.

electron0zero: yeah, let's clean it up in follow up...

electron0zero (approved):
:shipit:

electron0zero (approved):

PR #2877, opened 2023-08-31
Add tracql support for ancestor and parent operator

### LINKED ISSUE #2670: TraceQL: Support Parent and Ancestor structural operators
Currently TraceQL supports descendant `>>` and child `>` operators. These operators are evaluated on two spansets and return the spans from the right hand side. i.e. `{ span.http.target = "/customers" } >> { status = error }` returns spans with the error status below the `span.http.target` endpoint.

However, it might be useful to find all `/customer` spans that have errors somewhere beneath them. This is not possible with the current operators but could be acheived using something like:

`{ status = error } << { span.http.target = "/customers" }`

By reversing the operator we keep the behavior that structural operators always return spans from the right hand side. We also retain the meaning of the direction of the arrow and count of arrows.

Ancestor:
```
{} << {}
```

Parent
```
{} < {}
```

kousikmitra:
**What this PR does**:

Fixes #2670

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/traceql/enum_attributes.go
```diff
@@ -64,11 +64,13 @@ const (
 
 	// These intrinsics do not map to specific data points, but are used to
 	// indicate that Spans must be able to answer the structural methods
-	// DescdendantOf, SiblingOf, and ChildOf.  The details of those methods
+	// DescdendantOf, AncestorOf, SiblingOf, ChildOf and ParentOf.  The details of those methods
 	// and how these intrinsics are handled is left to the implementation.
 	IntrinsicStructuralDescendant
+	IntrinsicStructuralAncestor
```
mdisibio, line 70: I don't think we need these new intrinsics.  For example, for ancestor check we are calling the existing method `span.IsDescendantOf(other)`, but with reversed arguments, we should be able to reuse `IntrinsicStructuralDescendant`.   This way we don't have to make any changes at the storage layer (vParquet2/3 encodings).

docs/sources/tempo/traceql/_index.md
```diff
@@ -211,7 +211,9 @@ The second expression returns no traces because it's impossible for a single spa
 These spanset operators look at the structure of a trace and the relationship between the spans.
 
 - `{condA} >> {condB}` - The descendant operator (`>>`) looks for spans matching `{condB}` that are descendants of a span matching `{condA}`
+- `{condA} << {condB}` - The ancestor operator (`<<`) looks for spans matching `{condB}` that are ancestor of a span matching `{condA}`
 - `{condA} > {condB}` - The child operator (`>`) looks for spans matching `{condB}` that are direct child spans of a parent matching `{condA}`
+- `{condA} > {condB}` - The parent operator (`<`) looks for spans matching `{condB}` that are direct parent spans of a child matching `{condA}`
```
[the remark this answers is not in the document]
kousikmitra: Fixed

pkg/traceql/enum_attributes.go
```diff
@@ -64,7 +64,7 @@ const (
 
 	// These intrinsics do not map to specific data points, but are used to
 	// indicate that Spans must be able to answer the structural methods
-	// DescdendantOf, SiblingOf, and ChildOf.  The details of those methods
+	// DescdendantOf, AncestorOf, SiblingOf, ChildOf and ParentOf.  The details of those methods
```
mdisibio, line 67:
```suggestion
	// DescdendantOf, SiblingOf, and ChildOf.  The details of those methods
```

mdisibio (approved):
This looks great 👍   Please fix the couple docs tweaks and we will merge.  Thanks.

pkg/traceql/enum_attributes.go
kousikmitra: Done

PR #2882, opened 2023-09-01
Upgrade Azure SDK from `azure-storage-blob-go` to `azure-sdk-for-go`

### LINKED ISSUE #2129: [Ingester] Intermittent KubePersistentVolumeFillingUp alerts
**Describe the bug**
Our ingesters' `PersistentVolumeClaim`s intermittently fill up, seemingly at random.

**To Reproduce**
We have not been able to consistently reproduce the issue.

**Expected behavior**
We would expect that the used space on the `PersistentVolumeClaim` remains fairly consistent, and does not spike.

**Environment:**
 - Infrastructure: We run Tempo in microservices mode on Azure Kubernetes Service.
 - Deployment tool: We configure Tempo via jsonnet. The configuration files generated by jsonnet are applied using [Fluxcd](https://fluxcd.io/).

**Additional Context**
We run a total of nine ingesters; three in each zone across three zones.

For the past couple of months, the used space of an ingester's PVC will start gradually increasing. This gradual increase continues until it approaches 100%, at which point we have to manually intervene.

Here are a couple of screenshots of the PVC space usage when it happens:
1. Between `2023-02-21T00:25:00Z` and `2023-02-21T04:00:00Z`, the PVC of one of our ingesters in zone `A` went from ~3GB of usage to ~19GB: Screenshot 2023-02-23 at 09 23 59
2. Between `2023-02-21T23:50:00Z` and `2023-02-22T02:50:00Z`, the PVC of one of our ingesters in zone `C` went from ~
3GB of usage to ~18.4GB: Screenshot 2023-02-23 at 09 26 26
3. Between `2023-02-12T20:20:00Z` and `2023-02-13T15:28:00Z`, the PVC of one of our ingesters in zone `C` went from ~2GB of usage to ~18.9GB:  Screenshot 2023-02-23 at 09 30 43

We have observed this issue across different Tempo versions, namely `v1.5.0` and `v2.0.0` (which we recently upgraded to).

Initially, we figured that it was an issue with the flushing mechanism of the ingester, but preliminary log analysis has been unable to confirm this theory.

Using our first example from above (the ingester from zone `A`):
Error logs (excluding `TRACE_TOO_LARGE`) from the period before and after the increase in space usage:
Screenshot 2023-02-23 at 09 37 55

As we can see, no `error` logs were emitted in the period leading up to the disk being full. The first `error` log emitted is at `2023-02-21T04:02:34`, which is when there was no space left:
`"error completing wal block with local backend: error creating block: write /var/tempo/wal/blocks/single-tenant/f291dded-9dea-4dfa-985b-5984c5421f49/data.parquet: no space left on device"`

For the record, I am also attaching the graph of all logs (excluding `TRACE_TOO_LARGE`):
Screenshot 2023-02-23 at 09 40 22

This graph shows two spikes: one at ~`2023-02-21T04:00:00Z`, and a second one at ~`2023-02-21T06:30:00Z`. The first spike here are the `no space left on device` error messages; the second spike are logs from the manual intervention.

We would be much appreciative of any help with this issue, as we are somewhat stuck at the moment, and are unsure of how to proceed.

Please let me know of any additional information that may be of use in debugging the issue. I will make sure it is swiftly provided.

As a final note: I have created this issue as a bug, but I appreciate the fact that the problem can just as well be a misconfiguration on our side as it can be a bug in Tempo itself. Our team is very happy to use Tempo in general, and this is a nut that we cannot seem to crack on our own.

Thanks.

### LINKED ISSUE #2835: Upgrade from the `azure-storage-blob-go` SDK to the `azure-sdk-for-go ` SDK
[`azure-storage-blob-go`](https://github.com/Azure/azure-storage-blob-go) has been superseded by the [`azure-sdk-for-go`](https://github.com/Azure/azure-sdk-for-go), and is no longer being actively maintained.

Upgrading to [`azure-sdk-for-go`](https://github.com/Azure/azure-sdk-for-go) is a prerequisite for at least one task (#2705), and will fix at least one bug (#2129):
* https://github.com/grafana/tempo/issues/2705
* https://github.com/grafana/tempo/issues/2129

LasseHels:
**What this PR does**:
This pull request updates Tempo's `azure` package to use the [azure-sdk-for-go](https://github.com/Azure/azure-sdk-for-go) SDK instead of the deprecated [azure-storage-blob-go](https://github.com/Azure/azure-storage-blob-go) SDK.

This pull request aims to be fully backwards compatible, and is mainly a refactor. At least one functional change is included:
* Requests to Azure Blob Storage will now be retried once by default instead of zero time.

See https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/migrationguide.md for a non-exhaustive list of the differences between the two SDKs.

Fixes #2129
Fixes #2835

**Checklist**
- [x] Tests updated - Note that no new tests have been added, as this pull request should be functionally equivalent to `main`.
- [ ] ~Documentation added~ - I'm not sure documentation updates are necessary for this change.


LasseHels:
This pull request is not ready for review, but anyone interested is welcome to take a look at the work in progress. I've left a couple of comments in the code where upgrade quirks have been discovered. These comments will be removed before we merge.

CHANGELOG.md
```diff
@@ -14,7 +14,9 @@
 * [ENHANCEMENT] Add `TempoUserConfigurableOverridesReloadFailing` alert [#2784](https://github.com/grafana/tempo/pull/2784) (@kvrhdn)
 * [ENHANCEMENT] Add RootSpanName and RootServiceName to log about discarded spans [#2816](https://github.com/grafana/tempo/pull/2816) (@marcinginszt)
 * [ENHANCEMENT] Add `UserID` to log message about rate limiting [#2850](https://github.com/grafana/tempo/pull/2850) (@lshippy)
+* [ENHANCEMENT] Upgrade from deprecated [azure-storage-blob-go](https://github.com/Azure/azure-storage-blob-go) SDK to [azure-sdk-for-go](https://github.com/Azure/azure-sdk-for-go) [#2835](https://github.com/grafana/tempo/issues/2835) (@LasseHels)
```
LasseHels, line 17:
```
* [ENHANCEMENT] Upgrade from deprecated [azure-storage-blob-go](https://github.com/Azure/azure-storage-blob-go) SDK to [azure-sdk-for-go](https://github.com/Azure/azure-sdk-for-go) [#2835](https://github.com/grafana/tempo/issues/2835) (@LasseHels)
```

Not sure if an SDK upgrade counts as an enhancement? Wasn't sure where to put this (if anywhere).

tempodb/backend/azure/azure.go
```diff
@@ -132,12 +138,13 @@ func (rw *readerWriter) CloseAppend(context.Context, backend.AppendTracker) erro
 }
 
 func (rw *readerWriter) Delete(ctx context.Context, name string, keypath backend.KeyPath, _ bool) error {
-	blobURL, err := GetBlobURL(ctx, rw.cfg, backend.ObjectFileName(keypath, name))
+	blobClient, err := getBlobClient(ctx, rw.cfg, backend.ObjectFileName(keypath, name))
 	if err != nil {
-		return errors.Wrapf(err, "cannot get Azure blob URL, name: %s", backend.ObjectFileName(keypath, name))
+		return errors.Wrapf(err, "cannot get Azure blob client, name: %s", backend.ObjectFileName(keypath, name))
```
LasseHels, line 143: Are we OK to rename errors? I guess this could _technically_ be considered a breaking change? As a mostly academic argument, one might imagine that a Tempo consumer has an alert rule set up to listen for a log statement like `cannot get Azure blob URL`. In this case, this change would break that rule. However, for all intents and purposes, I'd consider this change OK. This also applies to other errors that have been renamed.

Thoughts?

tempodb/backend/azure/azure_helpers.go
```diff
@@ -9,40 +9,31 @@ import (
...
-func GetContainerURL(ctx context.Context, cfg *Config, hedge bool) (blob.ContainerURL, error) {
+func getContainerClient(ctx context.Context, cfg *Config, hedge bool) (container.Client, error) {
 	var err error
-	var p pipeline.Pipeline
 
-	retryOptions := blob.RetryOptions{
-		MaxTries: int32(maxRetries),
-		Policy:   blob.RetryPolicyExponential,
+	retry := policy.RetryOptions{
+		MaxRetries:    int32(maxRetries) - 1,
```
LasseHels, line 30: This configuration option was called "MaxTries" in the old SDK. It stated:
> "A value of 1 means 1 try and no retries."

Since the new option is "MaxRetries" (retries instead of tries), we need to set it to (previous value - 1) to keep the same behaviour.

This seems rather clumsy; a well-meaning engineer might update `maxRetries` to `0`, in which case we would be setting `MaxRetries` here to `-1` which I imagine would cause all sorts of errors.

We could hardcode the value to `0` here.

In fact, I wonder if this was an oversight of the implementation of the old SDK. With the old SDK, we were using `maxRetries` to set both `retryOptions.MaxTries` _and_ `DownloadFromBlobOptions.MaxRetryRequests`:
```go
	retryOptions := blob.RetryOptions{
		MaxTries: int32(maxRetries),
		Policy:   blob.RetryPolicyExponential,
	}
```

```go
		destBuffer, blob.DownloadFromBlobOptions{
			BlockSize:   blob.BlobDefaultDownloadBlockSize,
			Parallelism: maxParallelism,
			Progress:    nil,
			RetryReaderOptionsPerBlock: blob.RetryReaderOptions{
				MaxRetryRequests: maxRetries,
			},
		},
```

Is the desired behaviour that we use the same amount of retries across the board? If it is, then that would be a functional change, which I'm trying to stay away from in this pull request.

tempodb/backend/azure/azure_helpers.go
```diff
@@ -9,40 +9,31 @@ import (
...
+		TryTimeout:    1 * time.Minute,
```
LasseHels, line 31: The below three values were set implicitly via `blob.RetryPolicyExponential` in the old SDK. `blob.RetryPolicyExponential` was a shorthand for setting these defaults:
```go
IfDefault(&o.TryTimeout, 1*time.Minute)
IfDefault(&o.RetryDelay, 4*time.Second)
IfDefault(&o.MaxRetryDelay, 120*time.Second)
```

The new SDK contains no shorthand, so we manually set the same defaults. We might consider making these configuration options in Tempo, but I've deliberately not done that in the interest of diff leanness.

tempodb/backend/azure/azure_helpers.go
```diff
@@ -59,88 +50,112 @@ func GetContainerURL(ctx context.Context, cfg *Config, hedge bool) (blob.Contain
...
+	if err != nil {
+		return container.Client{}, err
 	}
 
-	if !cfg.UseFederatedToken && !cfg.UseManagedIdentity && cfg.UserAssignedID == "" {
-		credential, err := blob.NewSharedKeyCredential(getStorageAccountName(cfg), getStorageAccountKey(cfg))
+	var client *azblob.Client
+	accountKey := getStorageAccountKey(cfg)
+
+	switch {
```
LasseHels, line 82: In my opinion, the most interesting part of the upgrade is the "new" authentication switch.

I'm not sure how to verify that the switch results in the same authentication behaviour as the old implementation. If I'm not mistaken, our end-to-end tests only cover shared key credential, and doesn't touch the other three authentication mechanisms at all.

Extra careful review might be in order here; any input is appreciated.

LasseHels:
Tests are passing locally, and this should be ready for review. I can't run the pull request workflows myself, and am thus unable to verify that the CI tests will also pass.

tempodb/backend/azure/azure.go
joe-elliott: I'm fine with changing the text of a log line and not list it as a breaking change.

tempodb/backend/azure/azure.go
```diff
@@ -147,31 +154,26 @@ func (rw *readerWriter) Delete(ctx context.Context, name string, keypath backend
...
-
-		for _, blob := range list.Segment.BlobPrefixes {
-			objects = append(objects, strings.TrimPrefix(strings.TrimSuffix(blob.Name, dir), prefix))
-		}
 
-		// Continue iterating if we are not done.
-		if !marker.NotDone() {
-			break
+		for _, b := range page.Segment.BlobPrefixes {
+			objects = append(objects, strings.TrimPrefix(strings.TrimSuffix(*b.Name, dir), prefix))
```
joe-elliott, line 176: i'm concerned about blindly derefencing b.Name. do we have any guarantees that this will always have a value? i'd rather return an error here than panic

tempodb/backend/azure/azure.go
```diff
@@ -292,76 +294,74 @@ func (rw *readerWriter) append(ctx context.Context, src []byte, name string) err
...
+	props, err := blobClient.GetProperties(ctx, &blob.GetPropertiesOptions{})
 	if err != nil {
 		return err
 	}
 
 	length := int64(len(destBuffer))
 	var size int64
 
-	if length > 0 && length <= props.ContentLength()-offset {
+	if length > 0 && length <= *props.ContentLength-offset {
```
joe-elliott, line 348: similar pointer concerns here

tempodb/backend/azure/azure.go
```diff
@@ -292,76 +294,74 @@ func (rw *readerWriter) append(ctx context.Context, src []byte, name string) err
...
 
-	if err := blob.DownloadBlobToBuffer(context.Background(), blobURL.BlobURL, offset, size,
-		destBuffer, blob.DownloadFromBlobOptions{
-			BlockSize:   blob.BlobDefaultDownloadBlockSize,
-			Parallelism: maxParallelism,
-			Progress:    nil,
-			RetryReaderOptionsPerBlock: blob.RetryReaderOptions{
-				MaxRetryRequests: maxRetries,
-			},
+	if _, err := blobClient.DownloadBuffer(context.Background(), destBuffer, &blob.DownloadBufferOptions{
```
joe-elliott, line 354: interesting that we break context here. likely a bug. can we use the passed context instead?

tempodb/backend/azure/azure.go
```diff
@@ -373,42 +373,42 @@ func (rw *readerWriter) readRange(ctx context.Context, name string, offset int64
...
 
-	var props *blob.BlobGetPropertiesResponse
-	props, err := blobURL.GetProperties(ctx, blob.BlobAccessConditions{}, blob.ClientProvidedKeyOptions{})
+	props, err := blobClient.GetProperties(ctx, &blob.GetPropertiesOptions{})
 	if err != nil {
 		return nil, "", err
 	}
 
-	destBuffer := make([]byte, props.ContentLength())
+	destBuffer := make([]byte, *props.ContentLength)
```
joe-elliott, line 384: more pointer derefs

tempodb/backend/azure/azure.go
```diff
@@ -373,42 +373,42 @@ func (rw *readerWriter) readRange(ctx context.Context, name string, offset int64
...
 
 func readError(err error) error {
-	var storageError blob.StorageError
-	errors.As(err, &storageError)
-
-	if storageError == nil {
-		return errors.Wrap(err, "reading storage container")
-	}
-	if storageError.ServiceCode() == blob.ServiceCodeBlobNotFound {
+	if bloberror.HasCode(err, bloberror.BlobNotFound) {
```
joe-elliott, line 409: what is our confidence in this change? we have a fair amount of functionality dependent on correctly returning `backend.ErrDoesNotExist`

tempodb/backend/azure/compactor.go
```diff
@@ -127,33 +124,33 @@ func (rw *readerWriter) readAllWithModTime(ctx context.Context, name string) ([]
...
 	return BlobAttributes{
-		Size:         props.ContentLength(),
-		LastModified: props.LastModified(),
+		Size:         *props.ContentLength,
+		LastModified: *props.LastModified,
 	}, nil
 }
 
 // Delete removes the blob with the given name.
 func (rw *readerWriter) delete(ctx context.Context, name string) error {
```
joe-elliott, line 146: is this different than the other `Delete()` method? should we consolidate?

tempodb/backend/azure/azure_helpers.go
joe-elliott: so when constructing the "url" we were using max retries incorrectly and actually setting it to 0 retries. however, when calling `DownloadBlobToBuffer` we were presumably using it correctly and signalling we wanted 1 retry. naturally we have no idea how these two values interact in the real world.

since the intent seemed to be that we always had 1 retry, i think we should just set:
```
MaxRetries: int32(maxRetries),
```
and forget the -1. Likely worth noting in the changelog since it may cause behavioral changes in Tempo.

tempodb/backend/azure/azure_helpers.go
joe-elliott: agreed with these changes. please add a comment noting where these values came from and why they were chosen.

tempodb/backend/azure/azure_helpers.go
joe-elliott: I have basically no experience with Azure auth mechanisms. Unfortunately, we may just have to note this change loudly in the changelog and ask for quick feedback if operators see issues in azure.

If anyone knows how to write azure auth e2e tests it would be an excellent addition, but it's hard to throw resources down that rabbit hole.

joe-elliott (commented):
So far so good. I'm going to build this image and run in our azure dev environment. Will report back if there are issues.

LasseHels:
@joe-elliott Thanks for your review! I've got some internal tasks I need to prioritise, after which I will take a look at this again. ETA a couple of days.

joe-elliott:
> Thanks for your review! I've got some internal tasks I need to prioritise, after which I will take a look at this again. ETA a couple of days.

No worries. Thank you for taking this on. I have run this branch in our azure dev cluster for the past 12 or so hours without incident. Everything seems fine.

CHANGELOG.md
LasseHels: Roger that.

Action taken: this is now listed in the changelog as a `CHANGE` instead of `ENHANCEMENT`.

tempodb/backend/azure/azure_helpers.go
LasseHels: > since the intent seemed to be that we always had 1 retry, i think we should just set:
> ```
> MaxRetries: int32(maxRetries),
> ```

Roger that.

> Likely worth noting in the changelog since it may cause behavioral changes in Tempo.

Good idea.

Actions taken:
* We now set `MaxRetries` to `1` instead of `0`.
* Added changelog `ENHANCEMENT` entry for the change

tempodb/backend/azure/azure_helpers.go
LasseHels: Action taken: added this comment:
>  // The values for TryTimeout, RetryDelay and MaxRetryDelay are inherited from the old Azure SDK
> // (azure-storage-blob-go).
> //
> // See https://github.com/Azure/azure-storage-blob-go/blob/905b628ceb292e8d769ae62fb7cc5c5e949360db/azblob/zc_policy_retry.go#L89.

tempodb/backend/azure/azure_helpers.go
LasseHels: I took a brief look at whether we could set up some Azurite tests with different authentication mechanisms, but came up mostly empty-handed. It looks like Azurite [_might_](https://github.com/Azure/Azurite#default-storage-account) support some other authentication mechanisms, but I couldn't find documentation on how to set it up.

As a side note, the authentication switch currently assumes workload identity in the `default` case. After thinking about it some more, I think I prefer failing fast in case of invalid authentication configuration.

We have the `cfg.UseFederatedToken` option to explicitly specify workload identity authentication. A user who has accidentally misconfigured their authentication configuration might land in the `default` case and see an obscure error related to workload identity, when in reality their problem lies elsewhere.

I think I am in favour of returning an error in the `default` case. Something like this:
```go
	default:
		return container.Client{}, errors.New("please configure at least one Azure Blob Storage authentication mechanism")
```

Thoughts?

tempodb/backend/azure/azure.go
LasseHels: According to the SDK documentation, this value is required:
```go
type BlobPrefix struct {
	// REQUIRED
	Name *string `xml:"Name"`
}
```

The [API documentation](https://learn.microsoft.com/en-us/rest/api/storageservices/list-blobs) itself is unclear on whether this value is required.

I am fine with a defensive check.

Action taken: we now return an error if `b.Name == nil`.

tempodb/backend/azure/azure.go
LasseHels: Action taken: we now return an error if `props.ContentLength == nil`.

tempodb/backend/azure/azure.go
LasseHels: Sure. I was wondering why we used `context.Background()` here.

Action taken: we now pass `ctx` to `blobClient.DownloadBuffer` instead of `context.Background()`.

tempodb/backend/azure/azure.go
LasseHels: Action taken: we now return an error if `props.ContentLength == nil`.

tempodb/backend/azure/compactor.go
LasseHels: Looks like the two methods are mostly identical. The exported `Delete()` method returns `readError(err)` if `blobClient.Delete()` errors, but they are otherwise the same.

Action taken: removed `delete()` method in favour of `Delete()`.

tempodb/backend/azure/azure.go
LasseHels: The "new" functionality should be identical to the old. `bloberror.HasCode` is a shorthand for checking:
1. Is the response `azcore.ResponseError`.
2. Does the response contain error code `BlobNotFound`.

Additionally, the `TestReadError` test is still passing even with only changing `blob.ServiceCodeBlobNotFound` to `bloberror.BlobNotFound`.

LasseHels:
@joe-elliott Comments have been addressed. Are you able to take another look?

tempodb/backend/azure/azure_helpers.go
joe-elliott: I'm looking at our own azure configs and we don't set any of these flags. with the old way did it defafult to workload identity? if so we should probably keep it.

if we do keep default as workload identity can we use a fallthrough from the specific case to the default?

tempodb/backend/azure/azure_helpers.go
```diff
@@ -59,88 +54,112 @@ func GetContainerURL(ctx context.Context, cfg *Config, hedge bool) (blob.Contain
...
-		return blob.ContainerURL{}, err
+		client, err = azblob.NewClient(u.String(), credential, &opts)
+
+		if err != nil {
+			return container.Client{}, err
+		}
 	}
-	return c, nil
+
+	return *client.ServiceClient().NewContainerClient(cfg.ContainerName), nil
```
joe-elliott, line 145: it seems the sdk prefers to return a pointer to container.Client? any reason not to follow suit in `getContainerClient`. seems like it would be simpler

tempodb/backend/azure/azure_helpers.go
LasseHels: > I'm looking at our own azure configs and we don't set any of these flags

I'm guessing you set `storage_account_name` and `storage_account_key`? I'd hazard that this is how most users authenticate.

There are five configuration values of interest when it comes to authentication:
* `use_managed_identity`
* `use_federated_token`
* `user_assigned_id`
* `storage_account_name`
* `storage_account_key`

If _none_ of these values are set, then:
* The **old** authentication logic would default to `NewSharedKeyCredential`, and fails immediately, _unless_ the `AZURE_STORAGE_ACCOUNT` and `AZURE_STORAGE_KEY` environment variables are set.
* The **new** authentication logic defaults to `NewWorkloadIdentityCredential`, and fails immediately _unless_ the `AZURE_CLIENT_ID`, `AZURE_TENANT_ID` and `AZURE_FEDERATED_TOKEN_FILE` environment variables are set.

There might be some users thet rely on the old default behaviour, so I'll update the new switch to honor that by defaulting to shared key credential.

LasseHels: Action taken: the authentication switch now has three cases:
1. `case cfg.UseFederatedToken` covers workload identity.
2. `case cfg.UseManagedIdentity` covers both system-assigned and user-assigned managed identity.
3. `default` covers shared key credential.

This way, the default behaviour of the new authentication logic is the same as the old.

tempodb/backend/azure/azure_helpers.go
LasseHels: > any reason not to follow suit in `getContainerClient`

The main reason is that the old SDK implementation did not use pointers, but it seems that was because the old SDK didn't return pointers.

> seems like it would be simpler

Agreed.

Action taken: we now return pointers for SDK clients.

LasseHels:
@joe-elliott Thank you for the review. All comments have been addressed. Can you take another look? :eyes:

tempodb/backend/azure/azure_helpers.go
giskou: Side note: Looks like use_federated_token is not mentioned in the current [documentation](https://grafana.com/docs/tempo/latest/configuration/azure/#azure-blob-storage-permissions-and-management).
Took me a while to figure out how to enable workload identity. This discussion actually made that possible for me.

joe-elliott: @LasseHels Thanks for keeping it consistent with the old behavior
@giskou Thank you for the heads up. I'm going to make a quick docs issue:

https://github.com/grafana/tempo/issues/2927

joe-elliott (approved):
Amazing contribution. Thank you :pray: 

Looks like there's a changelog conflict. Clean that up and we'll merge!

LasseHels:
Conflicts have been fixed and this should be ready for a merge!

LasseHels:
Looks like a test is failing, I'm taking a look.

LasseHels:
Hmm, looks like the `TestHedge/hedge_disabled` test panicked. I can't reproduce the issue locally, neither when running the test directly nor when running `make test-with-cover-tempodb`. @joe-elliott any insight? Should we re-run the pipeline?

zalegrala:
@LasseHels If there is a failing test that you don't think is a result of your change, always feel free to re-run the failed job if you have the permission.  Thanks for the PR.

zalegrala:
@LasseHels Have you been able to run this code change in your environment?  We continue to see this test failure pop up intermittently and wondering about next steps.  It would help to know if we believe this was only a test failure, or if the change is also experiencing issues in an Azure environment.

zalegrala:
`go test -race -cover ./tempodb/backend/azure/ -count=1000 -failfast -run='(TestHedge)' -timeout=30m` is reproducing this pretty reliably for me.

zalegrala:
The failure looks specific to the Write test and a large buffer to force the multiple calls.  https://github.com/grafana/tempo/pull/2948 to work around the test.  Would still like to know how this is performing in a real environment.

LasseHels:
@zalegrala We are not running this change in our own environments. As such, I can't say whether the test failure is a false or true positive.

I might be able to spend some investigation time over the next couple of days to try to figure out what's going on. If you deem this to be an urgent issue, then you may be better off not relying on me, as I can't guarantee my availability.

LasseHels:
@zalegrala Spent a couple of hours looking into the issue and got nowhere. I can reproduce it locally, but I don't know the root cause yet. I can't tell if the issue is in our implementation of the Azure SDK, our test, the SDK itself or somewhere else.

I also see we created an issue in the SDK repository, will link that here for visibility: https://github.com/Azure/azure-sdk-for-go/issues/21555

zalegrala:
Thanks for the info @LasseHels.  I also spent some time researching and wasn't able to nail it down.  I've got https://github.com/grafana/tempo/pull/2952 up to bring back the old code as v1, and use the new code as v2 with a config option to switch.  This will allow us to gain some confidence as we move forward and give time for the Azure folks to dig into that issue.  Keep an eye on the changelog and docs for the final name of the config option in case it changes before merge.

PR #2884, opened 2023-09-01
Move ingester_traces_created_total

joe-elliott:
**What this PR does**:
Moves the location we are calculating `tempo_ingester_traces_created_total` to the point at which a trace is appended to the wal block. This will allow for better alignment with `tempodb_warnings_total{reason="disconnected_trace_flushed_to_wal"}` which improves the accuracy of the formulas recommended here:

https://github.com/grafana/tempo/pull/2790

Preferring this to adding a new multitenant metric. This change doesn't really lower the value of `tempo_ingester_traces_created_total` it just subtly changes it's meaning. Given that it still broadly metrics traces created total this feels fine to me, but willing to be disagreed with.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


mapno (approved):
LGTM

PR #2899, opened 2023-09-06
Add collection-interval to metrics-generator user-configurable overrides

rlankfo:
**What this PR does**:

Extends the user-configurable overrides added in https://github.com/grafana/tempo/pull/2543 with collection_interval override.

New JSON payload:

```json
{
    "forwarders": [],
    "metrics_generator": {
        "processors": [
            "service-graphs",
            "span-metrics"
        ],
        "disable_collection": false,
        "collection_interval": "60s",
        "processor": {
            "service_graphs": {
                "dimensions": [
                    "service.namespace"
                ],
                "enable_client_server_prefix": true,
                "peer_attributes": [
                    "peer.service",
                    "db.name",
                    "db.system",
                    "db.url"
                ]
            },
            "span_metrics": {
                "dimensions": [
                    "service.namespace"
                ],
                "enable_target_info": true
            }
        }
    }
}
```

~~Fixes #~~

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/overrides/userconfigurable/client/limits.go
```diff
@@ -52,6 +55,13 @@ func (l *LimitsMetricsGenerator) GetProcessor() *LimitsMetricsGeneratorProcessor
 	return nil
 }
 
+func (l *LimitsMetricsGenerator) GetCollectionInterval() (time.Duration, bool) {
+	if l != nil && l.CollectionInterval != nil {
+		return *l.CollectionInterval, true
+	}
+	return time.Minute, false
```
rlankfo, line 62: nit: as this value gets thrown out, it probably makes more sense to use `time.Duration(0)` here. Thoughts?
```suggestion
	return time.Duration(0), false
```

modules/overrides/userconfigurable/client/limits.go
```diff
@@ -25,8 +27,9 @@ func (l *Limits) GetMetricsGenerator() *LimitsMetricsGenerator {
 }
 
 type LimitsMetricsGenerator struct {
-	Processors        listtomap.ListToMap `json:"processors,omitempty"`
-	DisableCollection *bool               `json:"disable_collection,omitempty"`
+	Processors         listtomap.ListToMap `json:"processors,omitempty"`
+	DisableCollection  *bool               `json:"disable_collection,omitempty"`
+	CollectionInterval *time.Duration      `json:"collection_interval,omitempty"`
```
rlankfo, line 32: The json decoder expects `time.Duration` fields to be passed in nanoseconds. This means 60s is represented as 60000000000. This seemed like an acceptable trade-off and cleaner than making this a string and calling `time.ParseDuration`.

kvrhdn: This should be fine, we use `time.Duration` in a lot of configuration structs and it will automatically parse durations formatted like `15s` or `1m`.

modules/overrides/userconfigurable/client/limits.go
kvrhdn: I think we can also just return
```suggestion
	return 0, false
```
Doesn't really matter tbh, people shouldn't be reading this value anyway.

kvrhdn (commented):
The changes look good, thanks!

Can you add some validation around collection interval? While users are free to set their preferred interval, we have to make sure they don't break anything or stress Tempo too much.
We have a validator for the overrides in this file: https://github.com/grafana/tempo/blob/main/cmd/tempo/app/overrides_validation.go

Right now people could post negative values which might lead to a panic. I think it's reasonable to say collection interval must be a value between 15s and 5m?

modules/overrides/userconfigurable/client/limits.go
rlankfo: sounds good!

rlankfo:
> Right now people could post negative values which might lead to a panic. I think it's reasonable to say collection interval must be a value between 15s and 5m?

Thanks for the review @kvrhdn.  I've added validation for this, please take a look.

modules/overrides/userconfigurable/client/limits.go
rlankfo: I believe that will work when unmarshaling yaml because it's handled here:
https://github.com/go-yaml/yaml/blob/v3.0.1/decode.go#L647-L653

The json decoder won't accept a non numeric value, such as "60s".  So I think here we'll either have to use nanoseconds as I pointed out, or this would need to be a string and we'd have to figure out the best place to run `time.ParseDuration`.

rlankfo: Another option could be to change the JSON field to "collection_interval_seconds" and return `*l.CollectionInterval * 1000000` in `GetCollectionInterval`

kvrhdn: Oh I see, it's an oversight of the json decoder to not use `time.ParseDuration`. I'd prefer to implement our own unmarshaler in that case so we can use `15s` and `1m` etc. That is easier to use and consistent with configuration in the configmap.

There are articles explaining how to implement the unmarshaler interface: https://biscuit.ninja/posts/go-unmarshalling-json-into-time-duration/

rlankfo: thanks @kvrhdn. I implemented the custom marshal/unmarshal and added some tests. Let me know what you think!

modules/overrides/userconfigurable/client/limits.go
```diff
@@ -25,9 +29,48 @@ func (l *Limits) GetMetricsGenerator() *LimitsMetricsGenerator {
 	return nil
 }
 
+type Duration struct {
+	time.Duration
+}
```
kvrhdn, line 34: Mostly a nitpick: I think it would be cleaner to move the `Duration` struct into a separate file:
- this file (unfortunately) already has a lot of boilerplate making it difficult to read
- keeping it in a separate file will make it easier to reuse later, we can just move the file to shared package when we need it

What do you think of moving this struct and the corresponding test functions into `duration.go`?

kvrhdn (commented):
LGTM, works as expected 👍🏻 Just a slight comment about file structure.

modules/overrides/userconfigurable/client/limits.go
rlankfo: Makes sense to me. I moved the code. Thanks!

PR #2907, opened 2023-09-08
Enable errorlint and check for plain error comparisons

stoewer:
**What this PR does**:

Enable the linter [`errorlint`](https://github.com/polyfloyd/go-errorlint) to and configure it to check for plain error comparisons and type assertions. Enabling this linter can help to catch errors like the one fixed by #2890.

Fix all plain comparisons and type assertions found by `errorlint`.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,7 +154,7 @@ check-jsonnetfmt: jsonnetfmt
 
 .PHONY: lint
 lint:
-	$(LINT) run
+	$(LINT) run --config .golangci.yml
 
 ### Docker Images
 
--- a/cmd/tempo/app/app.go
+++ b/cmd/tempo/app/app.go
@@ -207,14 +207,11 @@ func (t *App) Run() error {
 		// let's find out which module failed
 		for m, s := range serviceMap {
 			if s == service {
-				switch service.FailureCase() {
-				case modules.ErrStopProcess:
+				if errors.Is(service.FailureCase(), modules.ErrStopProcess) {
 					level.Info(log.Logger).Log("msg", "received stop signal via return error", "module", m, "err", service.FailureCase())
-				case context.Canceled:
-				default:
+				} else {
 					level.Error(log.Logger).Log("msg", "module failed", "module", m, "err", service.FailureCase())
 				}
-
 				return
 			}
 		}
--- a/cmd/tempo/app/app.go
+++ b/cmd/tempo/app/app.go
@@ -207,10 +207,11 @@ func (t *App) Run() error {
 		// let's find out which module failed
 		for m, s := range serviceMap {
 			if s == service {
-				if errors.Is(service.FailureCase(), modules.ErrStopProcess) {
-					level.Info(log.Logger).Log("msg", "received stop signal via return error", "module", m, "err", service.FailureCase())
-				} else {
-					level.Error(log.Logger).Log("msg", "module failed", "module", m, "err", service.FailureCase())
+				err = service.FailureCase()
+				if errors.Is(err, modules.ErrStopProcess) {
+					level.Info(log.Logger).Log("msg", "received stop signal via return error", "module", m, "err", err)
+				} else if err != nil && !errors.Is(err, context.Canceled) {
+					level.Error(log.Logger).Log("msg", "module failed", "module", m, "err", err)
 				}
 				return
 			}
--- a/cmd/tempo/app/app.go
+++ b/cmd/tempo/app/app.go
@@ -210,7 +210,9 @@ func (t *App) Run() error {
 				err = service.FailureCase()
 				if errors.Is(err, modules.ErrStopProcess) {
 					level.Info(log.Logger).Log("msg", "received stop signal via return error", "module", m, "err", err)
-				} else if err != nil && !errors.Is(err, context.Canceled) {
+				} else if errors.Is(err, context.Canceled) {
+					return
+				} else if err != nil {
 					level.Error(log.Logger).Log("msg", "module failed", "module", m, "err", err)
 				}
 				return
--- a/modules/ingester/instance_search.go
+++ b/modules/ingester/instance_search.go
@@ -7,6 +7,10 @@ import (
 	"strings"
 	"sync"
 
+	ot_log "github.com/opentracing/opentracing-go/log"
+	"github.com/pkg/errors"
+	"go.uber.org/atomic"
+
 	"github.com/go-kit/log/level"
 	"github.com/grafana/dskit/user"
 	"github.com/grafana/tempo/pkg/api"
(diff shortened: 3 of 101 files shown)


zalegrala (commented):
This change looks okay to me.  I notice that there is now an `errors` package in the std lib.  https://pkg.go.dev/errors
Is now a good time to make that switch?  I think it might be a drop in replacement, at least for the changes I see in this PR.

stoewer:
> I notice that there is now an errors package in the std lib. https://pkg.go.dev/errors
Is now a good time to make that switch?

The package `github.com/pkg/errors` is a drop in replacement for `errors` from the standard library but not the other way around. The `errors` package does not have a function `errors.Wrap()` and also does not have errors with stack traces.

If we don't need the stack traces we could switch to `errors` and wrap errors with `fmt.Errorf("%w", err)` instead. But I'd rather not do this in this PR.

cmd/tempo/app/app.go
```diff
@@ -207,14 +207,12 @@ func (t *App) Run() error {
...
-				switch service.FailureCase() {
-				case modules.ErrStopProcess:
-					level.Info(log.Logger).Log("msg", "received stop signal via return error", "module", m, "err", service.FailureCase())
-				case context.Canceled:
-				default:
-					level.Error(log.Logger).Log("msg", "module failed", "module", m, "err", service.FailureCase())
+				err = service.FailureCase()
+				if errors.Is(err, modules.ErrStopProcess) {
+					level.Info(log.Logger).Log("msg", "received stop signal via return error", "module", m, "err", err)
+				} else if err != nil && !errors.Is(err, context.Canceled) {
```
joe-elliott, line 213: this is tough to read. maybe do a:

```
if errors.Is(context.Canceled) {
  continue
}
```

at the top of the loop. any idea why we are ignoring context canceled? also, it looks like the old code logs an error even it's nil?

modules/ingester/instance_search.go
```diff
@@ -7,6 +7,10 @@ import (
 	"strings"
 	"sync"
 
+	ot_log "github.com/opentracing/opentracing-go/log"
+	"github.com/pkg/errors"
```
joe-elliott, line 11: aren't Is and As in the standard errors package? i wonder why we import this

modules/overrides/userconfigurable/api/http.go
```diff
@@ -128,15 +129,15 @@ func (a *UserConfigOverridesAPI) DeleteHandler(w http.ResponseWriter, r *http.Re
...
 		w.WriteHeader(http.StatusNotFound)
 		return
 	}
-	if err == backend.ErrVersionDoesNotMatch || err == backend.ErrVersionInvalid {
+	if errors.Is(err, backend.ErrVersionDoesNotMatch) || errors.Is(err, backend.ErrVersionInvalid) {
 		http.Error(w, err.Error(), http.StatusPreconditionFailed)
 		return
 	}
-	if err, ok := err.(validationError); ok {
+	if ok := errors.As(err, &validationError{}); ok {
```
joe-elliott, line 140: is validationError really an error? i don't see an `Error()` method. should we make it one and do the pointer treatment suggested in this PR?

pkg/traceql/parse.go
```diff
@@ -22,7 +22,7 @@ func Parse(s string) (expr *RootExpr, err error) {
 		if r := recover(); r != nil {
 			var ok bool
 			if err, ok = r.(error); ok {
-				if errors.Is(err, ParseError{}) {
+				if errors.Is(err, &ParseError{}) {
```
joe-elliott, line 25: does this work? it seems more like an 'As' situation then testing against a sentinel error with `Is`. If it works it's fine, just looks strange to my eye

pkg/util/errors.go
```diff
@@ -46,7 +46,8 @@ func (es *MultiError) Add(err error) {
 	if err == nil {
 		return
 	}
-	if merr, ok := err.(MultiError); ok {
+	var merr MultiError
```
joe-elliott, line 49: should we swap the mutlierror `Error()` method to the pointer receiver?

Makefile
```diff
@@ -154,7 +154,7 @@ check-jsonnetfmt: jsonnetfmt
 
 .PHONY: lint
 lint:
-	$(LINT) run
+	$(LINT) run --config .golangci.yml
```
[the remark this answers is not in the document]
stoewer: This just makes it explicit

modules/ingester/instance_search.go
stoewer: We are using both packages `errors` and `github.com/pkg/errors` in the code base. And we are also using both wrapping techniques `fmt.Errorf("%w")` from the standard library and `errors.Wrap()` from `github.com/pkg/errors`. It isn't clear to me which one we prefer. For this PR I made the decision to use `github.com/pkg/errors` where both packages can be used.

modules/overrides/userconfigurable/api/http.go
stoewer: It used an embedded error and "inherited" it's `Error()` method. I changed it to an explicit implementation and with pointer receiver

pkg/util/errors.go
stoewer: I just noticed that MultiError is not used anywhere. Since there is also `github.com/grafana/dskit/multierror` and Go 1.20 added `errors.Join()` which serves a similar purpose, I think it's safe to remove MultiError

pkg/traceql/parse.go
stoewer: I replaced it with `errors.As`

modules/ingester/instance_search.go
joe-elliott: i'm surprised `Wrap` isn't in the standard library. I think this PR is big enough as is, but we should consider standardizing in the future.

modules/overrides/userconfigurable/api/api.go
```diff
@@ -168,11 +168,15 @@ func (a *UserConfigOverridesAPI) parseLimits(body io.Reader) (*client.Limits, er
...
 type validationError struct {
-	error
+	err error
 }
 
-func newValidationError(err error) validationError {
-	return validationError{err}
+func (e *validationError) Error() string {
+	return e.err.Error()
+}
```
joe-elliott, line 176: add Unwrap func? it feels like this is necessary for Is and As to correctly descend the wrapped error tree structure

modules/overrides/userconfigurable/api/http.go
joe-elliott: good catch. we use composition so rarely i don't think about it

pkg/util/errors.go
joe-elliott: +1

modules/ingester/instance_search.go
stoewer: > I think this PR is big enough as is, but we should consider standardizing in the future

I created this issue to replace github.com/pkg/errors with errors #2946

joe-elliott (approved):
lgtm!

PR #2909, opened 2023-09-09
Removing N for example generating timestamp

alileza:
**What this PR does**:

Generating the date with the command instructed in the docs resulted in `invalid character 'N' after object key:value pair` on MacOS arm64. Removing it works

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/api_docs/pushing-spans-with-http.md
+++ b/docs/sources/tempo/api_docs/pushing-spans-with-http.md
@@ -57,7 +57,7 @@ curl -X POST http://localhost:9411 -H 'Content-Type: application/json' -d '[{
 }]'
 ```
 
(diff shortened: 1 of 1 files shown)


mapno:
`date +%s%6` doesn't really adhere to any date formatting.

What the original `date +%s%6N` does is: 
* `%s`: seconds since the Epoch (1970-01-01 00:00 UTC)
* `%N`: Fractional seconds digits, default is 9 digits (nanosecond)
  *  `%6N`: microsecond (6 digits)

To get the example working, try using `gdate` from [coreutils](https://formulae.brew.sh/formula/coreutils) instead of mac's `date`.

PR #2916, opened 2023-09-11
Return first unmarshal error when parsing overrides config

kvrhdn:
**What this PR does**:
When using the newer overrides config format, if your config has an error it will always returns something like "field defaults does not exist".
This isn't helpful, instead hold on to the original error and return that instead when unmarshalling as legacy config fails as well.

~~Fixes #<issue number>~~

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/overrides/config.go
+++ b/modules/overrides/config.go
@@ -4,15 +4,15 @@ import (
 	"flag"
 	"time"
 
-	"github.com/grafana/tempo/pkg/util/listtomap"
-
-	"github.com/grafana/tempo/tempodb/backend"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/model"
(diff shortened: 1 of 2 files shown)


modules/overrides/config.go
```diff
@@ -200,7 +206,8 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	legacyCfg.UserConfigurableOverridesConfig = c.UserConfigurableOverridesConfig
 
 	if err := unmarshal(&legacyCfg); err != nil {
-		return err
+		// unmarshaling failed twice, return the first error
+		return originalError
```
kvrhdn, line 210: I'm not sure if we should combine both errors instead here? I'm worried that might be too confusing.

PR #2933, opened 2023-09-14
Get tags and tag values from backend if start and end are provided

rubenvp8510:
**What this PR does**:
This PR adds the ability to fetch tags and tags values from the backend, eliminating the limitation of having on the ingesters. This is still a preliminar PR that requires a lot of work.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


rubenvp8510:
Hello @mapno thank you very much, Can you start to review it? I'm trying to figure out why azure tests failed :(

zalegrala:
Azure test failure is related to https://github.com/grafana/tempo/pull/2952

modules/frontend/tagsharding.go
```diff
@@ -0,0 +1,489 @@
...
+	m := &jsonpb.Marshaler{}
+	bodyString, err := m.MarshalToString(&tempopb.SearchTagValuesResponse{
+		TagValues: valuetags,
+	})
+	if err != nil {
+		return "", err
+	}
+
+	return bodyString, nil
+}
```
mapno, line 73: This only covers the V1 endpoint. `SearchTagValuesV2` should be supported as well.

modules/frontend/tagsharding.go
```diff
@@ -0,0 +1,489 @@
...
+	return &TagNameHandler{}
+}
+
+type searchTagProgress struct {
+	err              error
+	statusCode       int
+	statusMsg        string
+	ctx              context.Context
+	finishedRequests int
+	resultsCombiner  *util.DistinctStringCollector
```
mapno, line 109: Tag values in V2 also have a type we need to return. Check out `ingester_search.go` to see how that's handled in there.

mapno (commented):
Nice start. Left a few comments mainly in the frontend.

rubenvp8510:
Sorry I was on PTO, I'll take this ticker again and update with comments.

modules/frontend/tagsharding.go
```diff
@@ -0,0 +1,489 @@
...
+
+type SearchTagSharderConfig struct {
+	ConcurrentRequests    int           `yaml:"concurrent_jobs,omitempty"`
+	TargetBytesPerRequest int           `yaml:"target_bytes_per_job,omitempty"`
+	DefaultLimit          uint32        `yaml:"default_result_limit"`
+	MaxLimit              uint32        `yaml:"max_result_limit"`
+	MaxDuration           time.Duration `yaml:"max_duration"`
+	QueryBackendAfter     time.Duration `yaml:"query_backend_after,omitempty"`
+	QueryIngestersUntil   time.Duration `yaml:"query_ingesters_until,omitempty"`
+}
```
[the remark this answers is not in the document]
rubenvp8510: What would be the right things to preserve here?

I think the limits and the QueryBackendAfter and QueryIngestersUntil ?

modules/frontend/tagsharding.go
rubenvp8510: I'll add V2 endpoint also, just wanted to make sure this approach for V1 works first.

modules/frontend/tagsharding.go
```diff
@@ -0,0 +1,489 @@
...
+	return bodyString, nil
+}
+
+func NewTagValueHandler() *TagValueHandler {
+	return &TagValueHandler{}
+}
+
+func NewTagNameHandler() *TagNameHandler {
+	return &TagNameHandler{}
+}
```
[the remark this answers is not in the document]
rubenvp8510: Done!

modules/frontend/tagsharding.go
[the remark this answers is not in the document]
rubenvp8510: I realized that I'm currently using the same search configuration for this, Should we include may be an additional configuration for search tags? 

At least we need to know when to search on the ingester vs on backend, I'm not sure if it's a good idea to have more configurations like this, or use same values for search here

WDYT? @mapno

rubenvp8510:
This is ready for another review

modules/frontend/tagsharding.go
```diff
@@ -0,0 +1,354 @@
...
+			subR := parent.Clone(ctx)
+			subR.Header.Set(user.OrgIDHeaderName, tenantID)
+
+			dc, err := m.DedicatedColumns.ToTempopb()
+			if err != nil {
+				reqCh <- &backendReqMsg{err: err}
+				return
+			}
+
+			subR, err = api.BuildTagSearchBlockRequest(subR, &tempopb.SearchTagsBlockRequest{
```
mapno, line 238: Shouldn't this handle tag values requests as well?

mapno (commented):
Nice work since last review. I've left comments mostly around the frontend code.

Also, I think this PR would benefit greatly from some tests: unit tests and/or `/integration` tests.

rubenvp8510:
@mapno I did some simplifications over the code and remove some unnecessary things. Also address your comments!  I know the PR is a little bit large but thanks for all the reviews!

rubenvp8510:
> The code is looking much better. I'd like to see tests for all the added functionality 🙏

Thank you! I'm adding testing

rubenvp8510:
Some tests added. Also a basic E2E test

rubenvp8510:
This is ready for another review. all comments are already address and tests added.

integration/e2e/e2e_test.go
```diff
@@ -126,6 +126,8 @@ func TestAllInOne(t *testing.T) {
 			now := time.Now()
 			util.SearchAndAssertTraceBackend(t, apiClient, info, now.Add(-20*time.Minute).Unix(), now.Unix())
 
+			util.SearchAndAsserTagsBackend(t, apiClient, now.Add(-20*time.Minute).Unix(), now.Unix())
```
mapno, line 129: It'd be awesome if we could integrate tests for this change in `api_test.go`. There are tests in that file that cover the tags and tag values APIs.

modules/querier/querier.go
```diff
@@ -798,6 +828,166 @@ func (q *Querier) internalSearchBlock(ctx context.Context, req *tempopb.SearchBl
...
+	wg := sync.WaitGroup{}
+	for idx := range scopes {
+		resps[idx] = &tempopb.SearchTagsResponse{}
+
+		wg.Add(1)
+		go func(scope string, ret **tempopb.SearchTagsResponse) {
+			defer wg.Done()
+			newReq := *req
+			newReq.SearchReq.Scope = scope
+			resp, err := q.SearchTagsBlocks(ctx, &newReq)
```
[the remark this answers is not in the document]
rubenvp8510: Is it? I think this is for v2, so it is not calling it self but calling SearchTagsBlocks v1 to retrive all tags for certain scope.

rubenvp8510:
@mapno I added a couple of more e2e tests, I think the linter is crazy :/

modules/querier/querier.go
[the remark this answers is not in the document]
mapno: You're right. I got confused, sorry.

modules/querier/querier.go
```diff
@@ -798,6 +828,166 @@ func (q *Querier) internalSearchBlock(ctx context.Context, req *tempopb.SearchBl
...
+		BlockID:          blockID,
+		DataEncoding:     req.DataEncoding,
+		FooterSize:       req.FooterSize,
+		DedicatedColumns: dc,
+	}
+
+	opts := common.DefaultSearchOptions()
+	opts.StartPage = int(req.StartPage)
+	opts.TotalPages = int(req.PagesToSearch)
+	opts.MaxBytes = q.limits.MaxBytesPerTrace(tenantID)
```
mapno, line 969: This is not needed, it's used only in the trace combiner for ID lookups.

tempodb/tempodb.go
```diff
@@ -353,6 +356,93 @@ func (rw *readerWriter) Search(ctx context.Context, meta *backend.BlockMeta, req
...
+	engine := traceql.NewEngine()
+
+	tag, err := traceql.ParseIdentifier(req.TagName)
+	if err != nil {
+		return nil, err
+	}
+
+	query := traceql.ExtractMatchers(req.Query)
+
+	if !traceql.IsEmptyQuery(query) {
```
mapno, line 421: This is not checking is autocomplete_filtering is enabled or not. I'd add at least a comment mentioning this behaviour.

mapno (commented):
Awesome work. Left some comments, but I've tested this locally and it's working fine 👏

integration/e2e/api_test.go
```diff
@@ -56,111 +57,238 @@ func TestSearchTagValuesV2(t *testing.T) {
...
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			callSearchTagValuesAndAssert(t, tempo, tc.tagName, tc.query, tc.expected)
+			callSearchTagValuesV2AndAssert(t, tempo, tc.tagName, tc.query, tc.expected, 0, 0)
 		})
 	}
+
+	// Wait to block flushed to backend, 20 seconds is the complete_block_timeout configuration on all in one, we add
+	// 1s for security.
+	time.Sleep(time.Second * 30)
```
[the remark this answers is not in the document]
rubenvp8510: What I was trying to test here is more than just flush to backend, but the retention on the ingester, waiting for `complete_block_timeout`, so then in theory if I call flush, the ingester will retain the blocks

What I wanted to test here is that after complete_block_timeout, should not be more tags on the ingester, but if I call the API with a time range, it should return tags.

integration/e2e/api_test.go
```diff
@@ -56,111 +57,238 @@ func TestSearchTagValuesV2(t *testing.T) {
...
+	}
+
+	// Assert tags on storage backend
+	now := time.Now()
+	start := now.Add(-2 * time.Hour)
+	end := now.Add(2 * time.Hour)
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			callSearchTagValuesV2AndAssert(t, tempo, tc.tagName, tc.query, searchTagValuesV2Response{}, start.Unix(), end.Unix())
```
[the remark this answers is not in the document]
rubenvp8510: Changed, uncomment that line. nice catch thanks!

integration/e2e/api_test.go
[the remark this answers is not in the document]
mapno: You're still relying on async processes for removing blocks in the ingester. Calling `/flush` also takes care of clearing blocks flushed to the backend that are past the `complete_block_timeout`.

What I mean here is that it'd be nice to check metrics to verify what we expect has happened—ie. the block has been flushed to storage and is no longer in the ingester.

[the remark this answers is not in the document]
rubenvp8510: Make sense, I did the changes we talked about!

rubenvp8510:
Hello Is there anything pending required for this PR to be merged?

integration/e2e/api_test.go
```diff
@@ -56,111 +57,220 @@ func TestSearchTagValuesV2(t *testing.T) {
...
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			callSearchTagValuesAndAssert(t, tempo, tc.tagName, tc.query, tc.expected)
+			callSearchTagValuesV2AndAssert(t, tempo, tc.tagName, tc.query, tc.expected, 0, 0)
 		})
 	}
+
+	// Wait to block flushed to backend, 20 seconds is the complete_block_timeout configuration on all in one, we add
+	// 1s for security.
```
mapno, line 158:
```suggestion
	// Wait to block flushed to backend, 20 seconds is the complete_block_timeout configuration on all in one, we add
	// 10s for security.
```

10s might be too much. I'd go for 1-2s and assert that the block has been flushed to the backend and cleared from the ingester with metrics (`tempo_ingester_blocks_flushed_total`, `tempodb_blocklist_length`, `tempo_ingester_blocks_cleared_total`)

mapno: See a similar example in https://github.com/grafana/tempo/blob/main/integration/e2e/e2e_test.go#L118-L121

integration/e2e/api_test.go
```diff
@@ -56,111 +57,220 @@ func TestSearchTagValuesV2(t *testing.T) {
...
+
+	// Assert no more on the ingester
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			callSearchTagValuesV2AndAssert(t, tempo, tc.tagName, tc.query, searchTagValuesV2Response{}, 0, 0)
+		})
+	}
+
+	// Wait to blocklist_poll to be completed
+	time.Sleep(time.Second * 2)
```
mapno, line 171: You can assert this has happened with `tempodb_blocklist_length`.

mapno (commented):
Left a couple of comments. We also need to update the [API docs](https://github.com/grafana/tempo/blob/main/docs/sources/tempo/api_docs/_index.md?plain=1#L253-L439)

modules/frontend/tagsharding.go
```diff
@@ -0,0 +1,434 @@
...
+
+func newTagResultCollector(ctx context.Context, factory tagResultHandlerFactory, limit int) *tagResultCollector {
+	return &tagResultCollector{
+		statusCode: http.StatusOK,
+		ctx:        ctx,
+		delegate:   factory(limit),
+	}
+}
+
+func (s searchTagSharder) httpErrorResponse(err error) *http.Response {
```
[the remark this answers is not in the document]
rubenvp8510: not really, but I prefer to use a method instead of populating the package with more functions.

modules/frontend/tagsharding.go
```diff
@@ -0,0 +1,434 @@
...
+				return
+			}
+		}
+	}
+}
+
+// ingesterRequest returns a new start and end time range for the backend as well as a http request
+// that covers the ingesters. If nil is returned for the http.Request then there is no ingesters query.
+// since this function modifies searchReq.Start and End we are taking a value instead of a pointer to prevent it from
+// unexpectedly changing the passed searchReq.
```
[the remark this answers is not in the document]
rubenvp8510: Not sure what you mean, this is actually true for the case I'm using. but I agree because `tagSearchReq` is an interface, this might not be true always.

rubenvp8510:
I think there are some flaky tests unrelated to this PR :(

rubenvp8510:
Hello, any other change that needs to be done to get this merged? THank you!

joe-elliott (approved):

modules/frontend/tagsharding.go
```diff
@@ -0,0 +1,431 @@
...
+// until limit results are found
+func (s searchTagSharder) RoundTrip(r *http.Request) (*http.Response, error) {
+	requestCtx := r.Context()
+
+	tenantID, err := user.ExtractOrgID(requestCtx)
+	if err != nil {
+		return s.httpErrorResponse(err), nil
+	}
+
+	handler := newTagResultCollector(requestCtx, s.tagShardHandlerFactory, s.overrides.MaxBytesPerTagValuesQuery(tenantID))
```
mapno, line 148: `MaxBytesPerTagValuesQuery` is defined only for `/tag/<tag>/values`, and not for `/tags`. This is changing the behaviour.

I'm not opposed to the change, but it should at least get documented.

mapno: `MaxBytesPerTagValuesQuery` is also implemented in the ingester—this was fine since only a single ingester request was created. It'd be good to coalesce this limit in a single place.

mapno (commented):
I have some final thoughts on the max bytes limits. Otherwise it's looking pretty good. I've been testing it and works well.

modules/frontend/tagsharding.go
rubenvp8510: I put a TODO here, may be we can address it in a follow up PR?

mapno (approved):
Awesome job! Thanks for keeping it up for so long.

rubenvp8510:
Something else required to be merged? more approvals?

mapno (commented):
@knylander-grafana, could you review the docs? Thanks!

docs/sources/tempo/api_docs/_index.md
```diff
@@ -289,9 +289,20 @@ $ curl -G -s http://localhost:3200/api/search/tags?scope=span  | jq
...
 ```
 
+Parameters:
+- `scope = (resource|span|intrinsic)`
+  Specifies the scope of the tags, this is an optional parameter, if not specified it means all scopes.
+  Default = `all`
+- `start = (unix epoch seconds)`
+  Optional.  Along with `end` define a time range from which tags should be returned.
+- `end = (unix epoch seconds)`
+  Optional.  Along with `start` define a time range from which tags should be returned. Providing both `start` and `end` will include blocks for the specified time range only.
```
knylander-grafana, line 299:
```suggestion
- `scope = (resource|span|intrinsic)`
  Specifies the scope of the tags. This is an optional parameter. If not specified, it means all scopes.
  Default = `all`
- `start = (unix epoch seconds)`
  Optional.  Along with `end`, defines a time range from which tags should be returned.
- `end = (unix epoch seconds)`
  Optional.  Along with `start`, defines a time range from which tags should be returned. Providing both `start` and `end` will include blocks for the specified time range only.
```

docs/sources/tempo/api_docs/_index.md
```diff
@@ -368,6 +389,13 @@ $ curl -G -s http://localhost:3200/api/search/tag/service.name/values  | jq
 }
 ```
 
+Parameters:
+- `start = (unix epoch seconds)`
+  Optional.  Along with `end` define a time range from which tags should be returned.
+- `end = (unix epoch seconds)`
+  Optional.  Along with `start` define a time range from which tags should be returned. Providing both `start` and `end` will include blocks for the specified time range only.
```
knylander-grafana, line 396:
```suggestion
- `start = (unix epoch seconds)`
  Optional. Along with `end`, defines a time range from which tags should be returned.
- `end = (unix epoch seconds)`
  Optional. Along with `start`, defines a time range from which tags should be returned. Providing both `start` and `end` will include blocks for the specified time range only.
```

docs/sources/tempo/api_docs/_index.md
```diff
@@ -340,7 +360,8 @@ $ curl -G -s http://localhost:3200/api/v2/search/tags  | jq
 
 ### Search tag values
 
-Ingester configuration `complete_block_timeout` affects how long tags are available for search.
+Ingester configuration `complete_block_timeout` affects how long tags are available for search, if start or end are not specified it will only
+fetch blocks that wasn't flushed to backend.
```
knylander-grafana, line 364:
```suggestion
Ingester configuration `complete_block_timeout` affects how long tags are available for search. if start or end are not specified, it will only
fetch blocks that wasn't flushed to backend.
```

docs/sources/tempo/api_docs/_index.md
```diff
@@ -405,6 +433,7 @@ $ curl http://localhost:3200/api/v2/search/tag/.service.name/values | jq .
   ]
 }
 ```
+This endpoint can also receive `start` and `end` optional parameters, these parameters will define the time range from which the tags will be fetched
```
knylander-grafana, line 436:
```suggestion
This endpoint can also receive `start` and `end` optional parameters. These parameters will define the time range from which the tags will be fetched
```

knylander-grafana (commented):
Thank you for updating docs! I only had a few minor things.

rubenvp8510:
I added all the suggestions! ready for another review.

integration/e2e/api_test.go
```diff
@@ -56,111 +57,230 @@ func TestSearchTagValuesV2(t *testing.T) {
...
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			callSearchTagValuesAndAssert(t, tempo, tc.tagName, tc.query, tc.expected)
+			callSearchTagValuesV2AndAssert(t, tempo, tc.tagName, tc.query, tc.expected, 0, 0)
 		})
 	}
+
+	// Wait to block flushed to backend, 20 seconds is the complete_block_timeout configuration on all in one, we add
+	// 2s for security.
+	callFlush(t, tempo)
```
electron0zero, line 159: can we wait on the prom metric here? it feels more reliable way to wait for flush.

I see that we are waiting using metric below on line 163.

```	// test metrics
	require.NoError(t, tempo.WaitSumMetrics(e2e.Equals(1), "tempo_ingester_blocks_flushed_total"))
```

also, maybe it makes sense to copy the all-in-one config, modify it use it here?

integration/e2e/api_test.go
```diff
@@ -56,111 +57,230 @@ func TestSearchTagValuesV2(t *testing.T) {
...
+
+	batch := makeThriftBatch()
+	require.NoError(t, jaegerClient.EmitBatch(context.Background(), batch))
+
+	// Wait for the traces to be written to the WAL
+	time.Sleep(time.Second * 3)
+	callSearchTagValuesAndAssert(t, tempo, "service.name", searchTagValuesResponse{TagValues: []string{"my-service"}}, 0, 0)
+
+	callFlush(t, tempo)
+	time.Sleep(time.Second * 22)
```
electron0zero, line 247: same comment for sleep here? I 22 seconds is a lot of time and it adds up to e2e test runtime.

pkg/api/search_tags.go
```diff
@@ -17,6 +22,343 @@ const (
 	ParamScopeIntrinsic = "intrinsic"
 )
 
+// ParseSearchBlockRequest parses all http parameters necessary to perform a block search.
+func ParseSearchBlockRequest(r *http.Request) (*tempopb.SearchBlockRequest, error) {
```
electron0zero, line 26: can we add tests for these new functions? we have few tests in `pkg/api/search_tags_test.go`. would be nice to have test for these as well.

electron0zero (approved):
did a quick review and I have some minor comments. 

It's already approved my @mapno so my comments are non-blocking.

thank you for contribution 🙏🏼

integration/e2e/api_test.go
rubenvp8510: I need not only the blocks to be flushed from ingester to backend but also removed from ingester, that is the motivation to wait for  `complete_block_timeout`. For what I understand, is not enough to wait for blocks to be flushed.

rubenvp8510:
Is it posible to do enhancement for the test in a follow up PR? If that is the case. Is something else requieres to merge this? Thank you!

electron0zero:
> Is it posible to do enhancement for the test in a follow up PR? If that is the case. Is something else requieres to merge this? Thank you!

sure thing 👍🏼

rubenvp8510:
Do I need to do something else? any other comments on this?

rubenvp8510:
@mapno any update with this? Do I require to do more changes?

PR #2935, opened 2023-09-14
Enforce max trace size on trace by id path

### LINKED ISSUE #2794: Enforce limitations on parquet trace by id queries
The v2 format enforces `max_bytes_per_trace` on query to prevent OOMs in the query path. The vParquet* formats has no such equivalents and we have seen OOMs when very large traces (100s of MBs) are requested.

Enforce `max_bytes_per_trace` in the vParquet* formats when doing a trace by id lookup OR build equivalent limitations for vParquet. The challenge is that the size of a trace in bytes used to be enforced on the proto size which was trivial to calculate in v2, but impossible to calculate in vParquet. We have [various](https://github.com/grafana/tempo/blob/12bdeff1d7ee726f703914cd43b9fd3c2d55a3b1/tempodb/encoding/vparquet3/compactor.go#L293) [hacky](https://github.com/grafana/tempo/blob/12bdeff1d7ee726f703914cd43b9fd3c2d55a3b1/tempodb/encoding/vparquet3/compactor.go#L314C6-L314C42) [attempts](https://github.com/grafana/tempo/blob/12bdeff1d7ee726f703914cd43b9fd3c2d55a3b1/tempodb/encoding/vparquet3/create.go#L239) to calculate sizes, but none of them are great.

It's possible that simply applying some combination of existing estimates against the per tenant max_bytes_per_trace will be sufficient and should be attempted. It's also possible that new limitations specific to vParquet will be required.

Also consider profiling trace by id lookup to determine if there's any obvious inefficiencies to remove.

joe-elliott:
Fixes #2794. Enforces max trace size on the query path by checking it in 4 places:

- As it's read from the parquet file at the block level (querier and ingester)
- As it's combined in the querier
- As it's combined in the ingester
- As it's combined in the query frontend

Additional changes

- Fixes an undocumented bug where errors were being dropped querying ingesters
- Fixes e2e test timestamps on spans
- Changelog


tempodb/encoding/vparquet3/block_findtracebyid.go
```diff
@@ -125,10 +125,10 @@ func (b *backendBlock) FindTraceByID(ctx context.Context, traceID common.ID, opt
 		span.SetTag("inspectedBytes", rr.BytesRead())
 	}()
 
-	return findTraceByID(derivedCtx, traceID, b.meta, pf, rowGroup)
+	return findTraceByID(derivedCtx, traceID, opts.MaxBytes, b.meta, pf, rowGroup)
 }
 
-func findTraceByID(ctx context.Context, traceID common.ID, meta *backend.BlockMeta, pf *parquet.File, rowGroup int) (*tempopb.Trace, error) {
+func findTraceByID(ctx context.Context, traceID common.ID, maxTraceSizeBytes int, meta *backend.BlockMeta, pf *parquet.File, rowGroup int) (*tempopb.Trace, error) {
```
mapno, line 131: Default encoding still is `vParquet2`. We should add it there as well.

mapno (approved):
LGTM

tempodb/encoding/vparquet3/block_findtracebyid.go
joe-elliott: I did not realize that. We should switch default to 3

joe-elliott:
@mapno I have done a bit of restructuring:

- pushed the error generation into the combiner
- return 400 from the querier instead of 500 where possible
- extend tests to confirm the above

500s from the queriers will be retried by the frontend. 400s will be piped up to the trace sharding layer and fail quickly. this is better, but still not perfect. 

making these changes makes me realize our query path needs an operational facelift. i intend to work on that next.

tempodb/encoding/vparquet3/block_findtracebyid.go
```diff
@@ -251,6 +251,13 @@ func findTraceByID(ctx context.Context, traceID common.ID, meta *backend.BlockMe
 		return nil, errors.Wrap(err, "error reading row from backend")
 	}
 
+	if maxTraceSizeBytes > 0 {
+		estimatedSize := estimateMarshalledSizeFromTrace(tr)
+		if estimatedSize > maxTraceSizeBytes {
+			return nil, errors.Errorf("trace exceeds max size in the block. size: %d, max: %d", estimatedSize, maxTraceSizeBytes)
+		}
+	}
+
```
mapno, line 260: Size check in `backendBlock` is missing in `vParquet` and `vParquet2`

mapno (commented):
Coalescing the check into the combiner is nice. Left some comments.

Makefile
```diff
@@ -160,7 +160,7 @@ lint:
 
 .PHONY: docker-component # Not intended to be used directly
 docker-component: check-component exe
-	docker build -t grafana/$(COMPONENT) --build-arg=TARGETARCH=$(GOARCH) -f ./cmd/$(COMPONENT)/Dockerfile .
+	docker build -t grafana/$(COMPONENT) --load --build-arg=TARGETARCH=$(GOARCH) -f ./cmd/$(COMPONENT)/Dockerfile .
```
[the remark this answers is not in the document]
mapno: I think that nothing in the GH CI triggers docker builds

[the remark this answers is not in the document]
joe-elliott: i think e2e tests do. don't they build a local `grafana/tempo:latest` and then use that for the tests?

tempodb/encoding/vparquet3/block_findtracebyid.go
joe-elliott: i'm ok with this. we are purposefully not backporting all changes to all previous parquet versions for sanity.

PR #2941, opened 2023-09-18
Store per-trace service statistics in vparquet4 blocks

### LINKED ISSUE #2940: Return additional trace metadata in TraceQL search response
It would be great if  trace visualization tools (e.g. Grafana, Jaeger UI) can show additional metadata in the search results for each matched trace. Specifically, the service names of traversed services, span count and error count.

Example in the Jaeger UI:
grafik

The `/api/search?q=...` API should return the above metadata, for example:
```
{
  "traceID": "20ae1e9d40ab5759699035e3c09092b3",
  "rootServiceName": "frontend",
  "rootTraceName": "HTTP GET /dispatch",
  "startTimeUnixNano": "1695040911822070267",
  "durationMs": 738,
  "spanSets": [
    {
      "spans": [
        {
          "spanID": "f953b82c348121d2",
          "startTimeUnixNano": "1695040911822559586",
          "durationNanos": "321561466"
        },
        {
          "spanID": "299e15e2db12f15f",
          "startTimeUnixNano": "1695040911822502499",
          "durationNanos": "321668520"
        },
        {
          "spanID": "e6cfbd41f7438b69",
          "startTimeUnixNano": "1695040911822213022",
          "durationNanos": "322204809"
        }
      ],
      "matched": 51
    }
  ],
  "servicesMeta": [
    {
      "serviceName": "mysql",
      "spanCount": "1"
    },
    {
      "serviceName": "customer",
      "spanCount": "1"
    },
    {
      "serviceName": "frontend",
      "spanCount": "24"
    },
    {
      "serviceName": "route",
      "spanCount": "10"
    },
    {
      "serviceName": "driver",
      "spanCount": "1"
    },
    {
      "serviceName": "redis",
      "spanCount": "14",
      "errorCount": 2
    }
  ]
},
```

Alternatively, the visualization tool can query the entire trace by ID for each search result. This will result in slow performance however.

andreasgerstmayr:
**What this PR does**:
Based on our discussion in the latest Tempo Community call about adding additional trace meta information to the TraceQL search response (#2940), I started experimenting with adding per-trace service statistics to vparquet3 blocks.

This is an early work-in-progress prototype which modifies the vparquet3 block schema, therefore it won't work on existing data and should be planned for the next version of the parquet block schema. Main point of this early prototype is to determine if this approach makes sense, and to find any issues early on.

This PR
* adds a `ServicesMeta` field to the Trace meta data containing a map of `ServiceMeta` structs containing the number of spans and errors of each service
* computes this field on ingestion
* returns this field in the `/api/search?q=` response, for example
```
{
  "traceID": "20ae1e9d40ab5759699035e3c09092b3",
  "rootServiceName": "frontend",
  "rootTraceName": "HTTP GET /dispatch",
  "startTimeUnixNano": "1695040911822070267",
  "durationMs": 738,
  "spanSets": [...],
  "serviceStats": {
    "customer": {
      "spanCount": 1
    },
    "driver": {
      "spanCount": 1
    },
    "frontend": {
      "spanCount": 24
    },
    "mysql": {
      "spanCount": 1
    },
    "redis": {
      "spanCount": 14,
      "errorCount": 3
    },
    "route": {
      "spanCount": 10
    }
  }
},
```

Fixes #2940

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/encoding/vparquet3/schema.go
```diff
@@ -213,6 +213,12 @@ type ResourceSpans struct {
 	ScopeSpans []ScopeSpans `parquet:"ss,list"`
 }
 
+type ServiceMeta struct {
+	ServiceName    string `parquet:",snappy,dict"`
+	SpanCount      uint64 `parquet:",delta"`
+	ContainsErrors bool   `parquet:""`
```
mdisibio: We could just as easily record the number of errors instead of a single bool.  Any reason not to?

tempodb/encoding/vparquet3/schema.go
```diff
@@ -473,6 +480,28 @@ func traceToParquet(meta *backend.BlockMeta, id common.ID, tr *tempopb.Trace, ot
 		}
 	}
 
+	// Calculate service meta information/statistics per trace
+	ot.ServicesMeta = []ServiceMeta{}
```
mdisibio, line 484: This creates an entry for every batch.  I think we need to dedupe and sum.

tempodb/encoding/vparquet3/block_traceql.go
```diff
@@ -1830,6 +1846,17 @@ func (c *traceCollector) KeepGroup(res *parquetquery.IteratorResult) bool {
 		case columnPathRootServiceName:
 			finalSpanset.RootServiceName = e.Value.String()
 			c.traceAttrs[traceql.NewIntrinsic(traceql.IntrinsicTraceRootService)] = traceql.NewStaticString(finalSpanset.RootServiceName)
+		case columnPathServicesMetaServiceName:
+			serviceName := e.Value.String()
+			finalSpanset.ServicesMeta = append(finalSpanset.ServicesMeta, traceql.ServiceMeta{ServiceName: serviceName})
+		case columnPathServicesMetaSpanCount:
+			spanCount := e.Value.Uint64()
+			finalSpanset.ServicesMeta[spanCountIdx].SpanCount = spanCount // TODO: there must be a better way
```
mdisibio, line 1854: `// TODO: there must be a better **way**`
Ah clever way to rebuild the entries.  The reason for the friction here is that the service metas are child rows of the trace, but they were added to the trace-level iterator (a nesting level above), so they got "flattened".   A more official way would be create a separate iterator for them, and have a new `serviceMetaCollector` recreate `traceql.ServiceMeta` structs. 

But this approach is fine imo.  The order of res.Entries should be reliable. So if you wanted to experiment: instead of incrementing the idxs you could create a new `traceql.ServiceMeta` every time columnPathServicesMetaServiceName is encountered.  It should be first of the three and represent a new row.

mdisibio (commented):
Hi, I wasn't able to attend the community call but this sounds like a great addition.  Left a few thoughts.   It would be very interesting to use this data to also speed up searches for things like `{resource.service.name="foo" && status=error}`.

tempodb/encoding/vparquet3/schema.go
andreasgerstmayr: The same `res.Resource.ServiceName` can appear multiple times in this loop? I guess then I can create a map indexed by the service name and convert it to a list later?

Or keep the map in the block schema. I'm not sure if a list or map is better there - I assumed lists are easier to handle with parquet. Purely from the data perspective, a map would make more sense, as this is a grouping of per-service meta information.

tempodb/encoding/vparquet3/combiner.go
```diff
@@ -86,6 +86,9 @@ func (c *Combiner) ConsumeWithFinal(tr *Trace, final bool) (spanCount int) {
 		c.result.RootSpanName = tr.RootSpanName
 	}
 	c.result.DurationNano = c.result.EndTimeUnixNano - c.result.StartTimeUnixNano
+	if len(c.result.ServicesMeta) == 0 {
```
joe-elliott, line 89: shouldn't this make an effort to combine? instead of just choosing one side or another. i think the only correct way to do it would be to just do the calculation after the combine is complete in the `Result()` function.

tempodb/encoding/vparquet3/schema.go
```diff
@@ -221,11 +227,12 @@ type Trace struct {
...
-	EndTimeUnixNano   uint64 `parquet:",delta"`
-	DurationNano      uint64 `parquet:",delta"`
-	RootServiceName   string `parquet:",dict"`
-	RootSpanName      string `parquet:",dict"`
+	StartTimeUnixNano uint64        `parquet:",delta"`
+	EndTimeUnixNano   uint64        `parquet:",delta"`
+	DurationNano      uint64        `parquet:",delta"`
+	RootServiceName   string        `parquet:",dict"`
+	RootSpanName      string        `parquet:",dict"`
+	ServicesMeta      []ServiceMeta `parquet:",list"`
```
joe-elliott, line 235: the parquet-go library will technically attempt schema changes automatically but we have found it to be costly and, at times, unstable.

if we want to add this field i think we should cut a vparquet4. wdyt @mdisibio ?

tempodb/encoding/vparquet3/schema.go
```diff
@@ -473,6 +480,29 @@ func traceToParquet(meta *backend.BlockMeta, id common.ID, tr *tempopb.Trace, ot
 		}
 	}
 
+	// Calculate service meta information/statistics per trace
+	ot.ServicesMeta = []ServiceMeta{}
+	// TODO: set upper limit on maximum services per trace, to reduce potential overhead when a trace traverses 1000s of services
```
joe-elliott, line 485: +1

tempodb/encoding/vparquet3/combiner.go
andreasgerstmayr: Yep, probably using the same algorithm as in `pkg/traceql/combine.go` (currently empty). I think I could just combine two `[]ServicesMeta` by the service name and sum up their spanCount and errorCount?

In this case it'd also beneficial to store it as a `map[string]ServicesMeta` (keyed by service name) instead of `[]ServicesMeta`, but I'm not sure how good that'll work with parquet, or if keeping it as a slice is better for interoperability.

tempodb/encoding/vparquet3/schema.go
mdisibio: Yes this was mentioned in the PR description above - we're on the same page and this is just a proof-of-concept.

tempodb/encoding/vparquet3/schema.go
mdisibio: Yes, it's typical to receive many batches for the same resource because of the practical upper-limit on the batch size. The agent/collector can only buffer a limited amount of spans or length of time before it has to forward the batch onward.  

Map vs list:  they both end up in the parquet file nearly identical, key column + data columns, the only difference is in the parquet metadata (afaik, we haven't used MAP columns before).  And the low-level logic in Fetch() to return the data will be the same.  So I don't think I have strong opinions either way.    A map may be simpler because it will be allocated anyway.

andreasgerstmayr: I've updated the PR to sum up the service statistics and use a map instead. As you mentioned it's actually quite straightforward to convert the list to a map, as the data is stored similarly :)

tempodb/encoding/vparquet3/block_traceql.go
andreasgerstmayr: In my tests I got all service names first, then all spanCounts and then the errorCounts, so while I'm let's say receiving the second span count, I need to know where to set it. After moving to a map data structure that was even more trickier (I could create a temporary slice to get a idx -> servicename mapping, but that's imho not very elegant).

I went with a serviceStatsCollector and iterator now.

tempodb/encoding/vparquet3/combiner.go
andreasgerstmayr: I've updated the PR to accumulate the spanCount and errorCount now.

tempodb/encoding/vparquet3/schema.go
andreasgerstmayr: done

tempodb/encoding/vparquet3/schema.go
andreasgerstmayr: Is there a branch for the next version of the parquet schema?

pkg/tempopb/tempo.proto
```diff
@@ -110,6 +110,12 @@ message TraceSearchMetadata {
   uint32 durationMs = 5;
   SpanSet spanSet = 6; // deprecated. use SpanSets field below
   repeated SpanSet spanSets = 7;
+  map<string, ServiceStats> serviceStats = 8;
```
andreasgerstmayr, line 113: Making this field non-nullable `[(gogoproto.nullable) = false]` to get `map[string]ServiceStats` unfortunately doesn't work, as then `jsonpb.Marshaler` always marshalls the `ServiceStats` struct to an empty JSON object `{}`, even when `ServiceStats` is not empty.

andreasgerstmayr:
I've rebased the PR on the `main-vparquet4` branch, updated the existing unit tests and added a few new test cases.

stoewer (changes requested):
Thanks for this great PR. I left a couple of minor change requests

> I've rebased the PR on the main-vparquet4 branch, updated the existing unit tests and added a few new test cases.

This looks like a good addition to vParquet4. 

Just a heads up: we rebase the `main-vparquet4` to `main` every couple of weeks in order to make sure we have the latest improvements from `main`. This means you may need to rebase you branch to `main-vparquet4` and resolve potential conflicts

pkg/traceql/combine.go
```diff
@@ -69,6 +69,13 @@ func combineSearchResults(existing *tempopb.TraceSearchMetadata, incoming *tempo
 		existing.DurationMs = incoming.DurationMs
 	}
 
+	// Merge service stats
+	for service, incomingStats := range incoming.ServiceStats {
+		existingStats := existing.ServiceStats[service]
```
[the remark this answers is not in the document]
andreasgerstmayr: Thanks!
I forgot to update this code when I changed the map from `map[string]ServiceStats` to `map[string]*ServiceStats`.

tempodb/encoding/vparquet3/block_traceql.go
```diff
@@ -1961,13 +1980,52 @@ func (c *traceCollector) KeepGroup(res *parquetquery.IteratorResult) bool {
 		}
 	}
 
+	finalSpanset.ServiceStats = make(map[string]traceql.ServiceStats)
```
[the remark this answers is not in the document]
andreasgerstmayr: Good idea!
Done.

tempodb/encoding/vparquet3/schema.go
andreasgerstmayr: I've set `MaxServiceStats = 100`. Any thoughts of increasing or decreasing that number?

I guess 100 catches pretty much all regular traces, and exotic traces with 1000s of services won't be processed, avoiding slowing down the processing of huge traces due to this new feature.

andreasgerstmayr:
> Thanks for this great PR. I left a couple of minor change requests

Thank you for your review!
I've applied all your suggestions.

> Just a heads up: we rebase the `main-vparquet4` to `main` every couple of weeks in order to make sure we have the latest improvements from `main`. This means you may need to rebase you branch again `main-vparquet4` and resolve potential conflicts

Sure :+1:

tempodb/encoding/vparquet3/schema.go
stoewer: I talked with @joe-elliott about this limit today. We concluded, that even for large traces with 1000s of services, the size for the service stats would still be acceptable. It would also be valuable if the service stats were guaranteed to be correct. We are therefore leaning towards removing the `MaxServiceStats` entirely. What do you think?

andreasgerstmayr: I agree, having incorrect (empty) ServiceStats for some traces is also odd behavior. If you both conclude that the size of the service stats field is not an issue, I'm +1 on removing this limit altogether. I've updated the PR.

stoewer (approved):

PR #2942, opened 2023-09-19
Tag value search improvements

### LINKED ISSUE #2957: Autocomplete filtering optimisations
## Context

Autocomplete filtering is the ability of Tempo to suggest tag values to Grafana, and thus the user, based on conditions already present in the search. Using an example, when typing the query `{ resource.environment = "prod" && resource.service = | }`, Tempo returns possible service names that are only present in production environments.

### Previous work

This was originally discussed in #1868, which resulted in a number of PRs (#2253, #2433) that add a new query param `q` to tag value search, that accepts a TraceQL for filtering down values.

The implementation is based on heavily reusing the TraceQL code for building iterators, fetching results and the collecting only the values for the desired attributes. While this approach works well, it doesn't scale as good as expected in big production clusters.

## Design optimisations

As a result of reusing TraceQL code, autocomplete filtering does a lot more work that's necessary for the feature. The TraceQL engine is built around the concept of spansets, which require fetching data to build spans for returning as results to the user. That same logic is used in autocomplete filtering, which then throws away most of the retrieved data.

The autocomplete filtering code path can be optimised to do only the work that's necessary, improving performance.

* Create new [collectors](https://github.com/grafana/tempo/blob/cebafe4a9e587dbd1eeabbab99c5e6883f536464/tempodb/encoding/vparquet3/block_traceql.go#L1567-L1912) that gather values for the wanted attribute, instead of building high-level objects (ie. spans, spansets).
  * Profiling indicates that a significant amount of time (up to ~40% of the autocomplete subroutine) is spent in functions like `runtime.mapassign` in [`batchCollector.KeepGroup()`](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet2/block_traceql.go#L1662-L1748) and `sync.(*Pool).Get` that's used for [reusing spans](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet2/block_traceql.go#L138-L159) in the [`spanCollector`](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet2/block_traceql.go#L1546)
* Simplify iterator builder logic and create only the required iterators
  * Because of the need for building spansets, Tempo retrieves data (eg. span start/end time, ) that is of no use for autocomplete. Such as [`span.kind`](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet2/block_traceql.go#L1024-L1034) when there are no span conditions (because we still need to build spans for the engine).

## Other paths

* Reduce the number of inspected blocks/data: a config parameter controls how many blocks should be inspected for tag value queries, reducing the amount of data read.
* Stop search faster by calculating cardinality of collected values.
  * Algorithms like [HyperLogLog](https://en.wikipedia.org/wiki/HyperLogLog) could be used to quit early if we're not getting new values after a while.
* Return after a configured amount of time (eg. 2s) with whatever data it's been fetched. In most use-cases, searching for longer is useless as the user won't wait for the results.

# Tasklist

```[tasklist]
### Tasks
- [ ] Refactor autocomplete filtering internals to read less data and allocate less memory https://github.com/grafana/tempo/pull/2942
- [ ] https://github.com/grafana/grafana/issues/77380
```

mapno:
**What this PR does**:

Adapts code for building iterators for traceql for autocomplete. Changes are focused in two areas:

* Create new [collectors](https://github.com/grafana/tempo/blob/cebafe4a9e587dbd1eeabbab99c5e6883f536464/tempodb/encoding/vparquet3/block_traceql.go#L1567-L1912) that gather only attribute values, instead of building high-level objects (ie. spans, spansets).
  * Profiling indicates that a significant amount of time (up to ~40% of the autocomplete subroutine) is spent in functions like `runtime.mapassign` in [`batchCollector.KeepGroup()`](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet2/block_traceql.go#L1662-L1748) and `sync.(*Pool).Get` that's used for [reusing spans](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet2/block_traceql.go#L138-L159) in the [`spanCollector`](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet2/block_traceql.go#L1546)
* Trim down iterator builder logic and create only the required iterators
  * Because of the need for building spansets, Tempo retrieves data (eg. span start/end time, ) that is of no use for autocomplete. Such as [`span.kind`](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet2/block_traceql.go#L1024-L1034) when there are no span conditions (because we still need to build spans for the engine).

Fixes #2957 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/encoding/vparquet3/block_autocomplete.go
```diff
@@ -0,0 +1,869 @@
...
+				if err != nil {
+					return nil, errors.Wrap(err, "creating predicate")
+				}
+				addPredicate(entry.columnPath, pred)
+				columnSelectAs[entry.columnPath] = cond.Attribute.Name
+				continue
+			}
+		}
+
+		// Attributes stored in dedicated columns
```
mdisibio, line 297: There is a lot of overlap between this and the iterators for search.  Do you have any thoughts on if this can/should be addressed?   Could we expand the existing `createThingIterator` funcs to take the group predicates, which I believe are the main differences?

tempodb/encoding/vparquet3/block_autocomplete.go
```diff
@@ -0,0 +1,869 @@
...
+}
+
+func mapSpanAttr(e entry) (traceql.Attribute, traceql.Static) {
+	switch e.Key {
+	case columnPathSpanID,
+		columnPathSpanParentID,
+		columnPathSpanNestedSetLeft,
+		columnPathSpanNestedSetRight,
+		columnPathSpanStartTime:
+		return traceql.Attribute{}, traceql.Static{}
```
mdisibio, line 731: I'm wondering if there is a way to avoid this, ideally we totally ignore these columns (no allocs).  I think the root cause is how `keep` is a callback function.  The lower layers like `distinctSpanCollector.KeepGroup` don't know the name of the target attribute so they have to parse every one.  If `keep` were the `Attribute` from the original AutocompleteRequest we could use it here to only look at the results of interest.


Edit:  See the other comment about `selectAs=""`.

tempodb/encoding/vparquet3/block_autocomplete.go
```diff
@@ -0,0 +1,869 @@
...
+			traceIters = append(traceIters, makeIter(columnPathTraceID, nil, columnPathTraceID))
+		case traceql.IntrinsicTraceDuration:
+			var pred parquetquery.Predicate
+			if allConditions {
+				pred, err = createIntPredicate(cond.Op, cond.Operands)
+				if err != nil {
+					return nil, err
+				}
+			}
+			traceIters = append(traceIters, makeIter(columnPathDurationNanos, pred, columnPathDurationNanos))
```
mdisibio, line 599: Using this line to start a discussion but it applies throughout the entire file.  I think there's a further improvement here by using `""` for the last `makeIter` parameter `selectAs`.  SelectAs means the parquet iterators pull back the value, but it's possible to use `selectAs=""` and only assert the predicate.  This will eliminate a lot of allocations and work in the KeepGroup callbacks.  The only one that needs a `selectAs` is the target attribute.   In TraceQL we want to pull back all values, pass them to the engine and the UI but here we don't.

tempodb/encoding/vparquet3/block_autocomplete.go
```diff
@@ -0,0 +1,869 @@
...
+				continue
+			}
+		}
+
+		// Else: generic attribute lookup
+		genericConditions = append(genericConditions, cond)
+	}
+
+	for columnPath, predicates := range columnPredicates {
+		iters = append(iters, makeIter(columnPath, parquetquery.NewOrPredicate(predicates...), columnSelectAs[columnPath]))
```
mdisibio, line 323: Can use `orIfNeeded` here.

tempodb/encoding/vparquet3/block_autocomplete.go
```diff
@@ -0,0 +1,869 @@
...
```
mdisibio, line 544: Can use `orIfNeeded` here

modules/ingester/instance_search.go
```diff
@@ -399,7 +367,9 @@ func (i *instance) SearchTagValuesV2(ctx context.Context, req *tempopb.SearchTag
 	query := extractMatchers(req.Query)
```
mdisibio, line 367: Having trouble with the query `{ resource.service.name="shop-backend" && span.http.method="PUT" && kind=server && span.http.url= }`.   It fails with a user-facing UI error `unexpected error searching complete block (ac7f17fd-033f-4ff3-816a-9aeaf90edc8c): parse error at line 1, col 73: syntax error: unexpected IDENTIFIER`

extractMatchers returns `{resource.service.name="shop-backend" && span.http.method="PUT" && kind=s}` so I think the issue is with the unquoted `=server`.

mapno: Fixed the regex to support those queries, nice catch.

tempodb/encoding/vparquet3/block_autocomplete.go
```diff
@@ -0,0 +1,869 @@
...
+	if err != nil {
+		return err
+	}
+
+	iter, err := autocompleteIter(ctx, req, pf, opts, b.meta.DedicatedColumns)
+	if err != nil {
+		return errors.Wrap(err, "creating fetch iter")
+	}
+
+	// TODO: The iter shouldn't be exhausted here, it should be returned to the caller
```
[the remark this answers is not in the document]
mapno: I'm leaning towards the callback approach, since it's how the other tag values functions (`SearchTagValues, `SearchTagValuesV2`) do it. It keeps the code simpler I think.

joe-elliott:
Should we include a test runner here: https://github.com/grafana/tempo/blob/main/tempodb/tempodb_search_test.go#L45 that exercises the tag/tag value search functionality?

tempodb/encoding/vparquet3/block_autocomplete.go
mapno: I've gone with your suggestion of `selectAs=""`, so empty structs shouldn't be returned anymore. I've kept `keep` as a safeguard nonetheless. I didn't have the courage to delete it 😆

tempodb/encoding/vparquet3/block_autocomplete.go
mapno: Very nice suggestion. Applied! :)

mapno:
> Should we include a test runner here: https://github.com/grafana/tempo/blob/main/tempodb/tempodb_search_test.go#L45 that exercises the tag/tag value search functionality?

I think it'd be difficult to adapt to the autocomplete interfaces. I'd rather extend `block_autocomplete_test.go`. Is there something clearly missing from those tests?

tempodb/encoding/vparquet3/block_autocomplete.go
mdisibio: The consensus is this isn't worth trying to deduplicate the code path for Autocomplete and Search.  It seems possible eventually but very complex and not worth tackling in this PR.

mdisibio (approved):
Amazing addition.  We've been running this internally it's looking great.  Let's merge and get this update in.

PR #2955, opened 2023-09-21
Use the standard `errors` package to create, wrap, and handle errors

### LINKED ISSUE #2946: Use the standard 'errors' package to create, wrap, and handle errors

The Tempo code base uses two packages to create, wrap, and handle errors:
* `errors` from the standard library
* `github.com/pkg/errors` a drop-in replacement for `errors` that was created before [Go 1.13 added error wrapping](https://go.dev/blog/go1.13-errors)

Using both packages creates the following problems:
*  Where both packages can be used it can be confusing which one to import
* `github.com/pkg/errors` implements a function `errors.Wrap(err, "...")` that does the same as `fmt.Errorf("... : %w", err)` and it can be confusing which function should be used
* `github.com/pkg/errors` is archived and does not provide more recent additions to the `errors` package such as `errors.Join()`: It is no longer a real drop-in replacement

* Replace all imports of `github.com/pkg/errors` with `errors`
* Use `fmt.Errorf("%w")` instead of `errors.Wrap()` and `errors.Errorf("%w")`
* Remove `github.com/pkg/errors` from direct dependencies (it is probably still an indirect dependency that is used by other packages we use)
* Look for errors that are not properly wrapped with i.e. `fmt.Errorf("%v", err)` and wrap them properly

**Additional context**
Here is a [recent discussion about this problem](https://github.com/grafana/tempo/pull/2907#discussion_r1328699378)

sonisr:
**What this PR does**:
Standardises the error formatting across the Tempo code base, using the `errors` package from the standard library, replacing the `github.com/pkg/errors` package.

Fixes #2946 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo/app/app.go
+++ b/cmd/tempo/app/app.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -20,7 +21,6 @@ import (
 	"github.com/grafana/dskit/services"
 	"github.com/grafana/dskit/signals"
 	"github.com/jedib0t/go-pretty/v6/table"
-	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/version"
 	"google.golang.org/grpc"
@@ -276,7 +276,7 @@ func (t *App) writeStatusConfig(w io.Writer, r *http.Request) error {
 	case "":
 		output = t.cfg
 	default:
-		return errors.Errorf("unknown value for mode query parameter: %v", mode)
+		return fmt.Errorf("unknown value for mode query parameter: %v", mode)
 	}
 
 	out, err := yaml.Marshal(output)
@@ -410,7 +410,7 @@ func (t *App) statusHandler() http.HandlerFunc {
 					if err == nil {
 						err = e
 					} else {
-						err = errors.Wrap(err, e.Error())
+						err = fmt.Errorf("%w", err)
 					}
 				}
 			}
@@ -488,7 +488,7 @@ func (t *App) writeStatusEndpoints(w io.Writer) error {
 		return nil
 	})
 	if err != nil {
-		return errors.Wrap(err, "error walking routes")
+		return fmt.Errorf("error walking routes %w", err)
 	}
 
 	sort.Slice(endpoints[:], func(i, j int) bool {
@@ -510,7 +510,7 @@ func (t *App) writeStatusEndpoints(w io.Writer) error {
 
 	_, err = w.Write([]byte(fmt.Sprintf("\nAPI documentation: %s\n\n", apiDocs)))
 	if err != nil {
-		return errors.Wrap(err, "error writing status endpoints")
+		return fmt.Errorf("error writing status endpoints %w", err)
 	}
 
 	return nil
--- a/cmd/tempo/app/app.go
+++ b/cmd/tempo/app/app.go
@@ -114,7 +114,7 @@ func New(cfg Config) (*App, error) {
 	app.setupAuthMiddleware()
 
 	if err := app.setupModuleManager(); err != nil {
-		return nil, fmt.Errorf("failed to setup module manager %w", err)
+		return nil, fmt.Errorf("failed to setup module manager: %w", err)
 	}
 
 	return app, nil
@@ -171,7 +171,7 @@ func (t *App) Run() error {
 
 	serviceMap, err := t.ModuleManager.InitModuleServices(t.cfg.Target)
 	if err != nil {
-		return fmt.Errorf("failed to init module services %w", err)
+		return fmt.Errorf("failed to init module services: %w", err)
 	}
 	t.serviceMap = serviceMap
 
@@ -182,7 +182,7 @@ func (t *App) Run() error {
 
 	sm, err := services.NewManager(servs...)
 	if err != nil {
-		return fmt.Errorf("failed to start service manager %w", err)
+		return fmt.Errorf("failed to start service manager: %w", err)
 	}
 
 	// before starting servers, register /ready handler and gRPC health check service.
@@ -234,7 +234,7 @@ func (t *App) Run() error {
 	// in other state than New, which should not be the case.
 	err = sm.StartAsync(context.Background())
 	if err != nil {
-		return fmt.Errorf("failed to start service manager %w", err)
+		return fmt.Errorf("failed to start service manager: %w", err)
 	}
 
 	return sm.AwaitStopped(context.Background())
@@ -488,7 +488,7 @@ func (t *App) writeStatusEndpoints(w io.Writer) error {
 		return nil
 	})
 	if err != nil {
-		return fmt.Errorf("error walking routes %w", err)
+		return fmt.Errorf("error walking routes: %w", err)
 	}
 
 	sort.Slice(endpoints[:], func(i, j int) bool {
@@ -510,7 +510,7 @@ func (t *App) writeStatusEndpoints(w io.Writer) error {
 
 	_, err = w.Write([]byte(fmt.Sprintf("\nAPI documentation: %s\n\n", apiDocs)))
 	if err != nil {
-		return fmt.Errorf("error writing status endpoints %w", err)
+		return fmt.Errorf("error writing status endpoints: %w", err)
 	}
 
 	return nil
--- a/cmd/tempo/app/app.go
+++ b/cmd/tempo/app/app.go
@@ -410,7 +410,7 @@ func (t *App) statusHandler() http.HandlerFunc {
 					if err == nil {
 						err = e
 					} else {
-						err = fmt.Errorf(e.Error(), ": %w", err)
+						err = fmt.Errorf("%s: %w", e.Error(), err)
 					}
 				}
 			}
--- a/cmd/tempo/app/app.go
+++ b/cmd/tempo/app/app.go
@@ -410,7 +410,7 @@ func (t *App) statusHandler() http.HandlerFunc {
 					if err == nil {
 						err = e
 					} else {
-						err = fmt.Errorf("%w", err)
+						err = fmt.Errorf(e.Error(), ": %w", err)
 					}
 				}
 			}
--- a/modules/ingester/local_block.go
+++ b/modules/ingester/local_block.go
@@ -2,6 +2,7 @@ package ingester
 
 import (
 	"context"
+	"fmt"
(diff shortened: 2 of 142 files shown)


sonisr:
I'd consider this a WIP. I have yet to address errors that are not properly wrapped.

> Look for errors that are not properly wrapped with i.e. fmt.Errorf("%v", err) and wrap them properly

cmd/tempo/app/app.go
```diff
@@ -114,7 +114,7 @@ func New(cfg Config) (*App, error) {
 	app.setupAuthMiddleware()
 
 	if err := app.setupModuleManager(); err != nil {
-		return nil, fmt.Errorf("failed to setup module manager %w", err)
+		return nil, fmt.Errorf("failed to setup module manager: %w", err)
```
stoewer, line 117: Good catch spotting the missing `:`

cmd/tempo/app/app.go
```diff
@@ -410,7 +410,7 @@ func (t *App) statusHandler() http.HandlerFunc {
 					if err == nil {
 						err = e
 					} else {
-						err = errors.Wrap(err, e.Error())
+						err = fmt.Errorf(e.Error(), ": %w", err)
```
stoewer, line 413: I think should be like this:
```suggestion
						err = fmt.Errorf("%s: %w", e.Error(), err)
```
Maybe we could also use `errors.Join()` or dskit multierror instead of `joinErrors`?

modules/ingester/local_block.go
```diff
@@ -85,7 +85,7 @@ func (c *localBlock) SetFlushed(ctx context.Context) error {
 func (c *localBlock) Write(ctx context.Context, w backend.Writer) error {
 	err := encoding.CopyBlock(ctx, c.BlockMeta(), c.reader, w)
 	if err != nil {
-		return errors.Wrap(err, "error copying block from local to remote backend")
+		return fmt.Errorf("error copying block from local to remote backen: %w", err)
```
stoewer, line 88:
```suggestion
		return fmt.Errorf("error copying block from local to remote backend: %w", err)
```

modules/querier/querier.go
```diff
@@ -376,12 +376,12 @@ func forOneIngesterRing(ctx context.Context, replicationSet ring.ReplicationSet,
 
 		client, err := pool.GetClientFor(ingester.Addr)
 		if err != nil {
-			return nil, errors.Wrap(err, fmt.Sprintf("failed to get client for %s", ingester.Addr))
+			return nil, fmt.Errorf(fmt.Sprintf("failed to get client for %s: ", ingester.Addr), "%w", err)
```
stoewer, line 379: I think the inner `fmt.Sprintf` isn't strictly necessary
```suggestion
			return nil, fmt.Errorf("failed to get client for %s:  %w", ingester.Addr, err)
```

modules/querier/querier.go
```diff
@@ -376,12 +376,12 @@ func forOneIngesterRing(ctx context.Context, replicationSet ring.ReplicationSet,
...
 		}
 
 		resp, err := f(funcCtx, client.(tempopb.QuerierClient))
 		if err != nil {
-			return nil, errors.Wrap(err, fmt.Sprintf("failed to execute f() for %s", ingester.Addr))
+			return nil, fmt.Errorf(fmt.Sprintf("failed to execute f() for %s: ", ingester.Addr), "%w", err)
```
stoewer, line 384:
```suggestion
			return nil, fmt.Errorf("failed to execute f() for %s: %w", ingester.Addr, err)
```

modules/querier/querier.go
```diff
@@ -412,20 +412,20 @@ func (q *Querier) forGivenGenerators(
 
 		client, err := q.generatorPool.GetClientFor(generator.Addr)
 		if err != nil {
-			return nil, errors.Wrap(err, fmt.Sprintf("failed to get client for %s", generator.Addr))
+			return nil, fmt.Errorf(fmt.Sprintf("failed to get client for %s: ", generator.Addr), "%w", err)
```
stoewer, line 415:
```suggestion
			return nil, fmt.Errorf("failed to get client for %s: %w", generator.Addr, err)
```

modules/querier/querier.go
```diff
@@ -412,20 +412,20 @@ func (q *Querier) forGivenGenerators(
...
 		}
 
 		resp, err := f(funcCtx, client.(tempopb.MetricsGeneratorClient))
 		if err != nil {
-			return nil, errors.Wrap(err, fmt.Sprintf("failed to execute f() for %s", generator.Addr))
+			return nil, fmt.Errorf(fmt.Sprintf("failed to execute f() for %s: ", generator.Addr), "%w", err)
```
stoewer, line 420:
```suggestion
			return nil, fmt.Errorf("failed to execute f() for %s: %w", generator.Addr, err)
```

tempodb/encoding/vparquet3/compactor.go
```diff
@@ -198,7 +198,7 @@ func (c *Compactor) Compact(ctx context.Context, l log.Logger, r backend.Reader,
 			currentBlockPtrCopy.meta.EndTime = maxBlockEnd
 			err := c.finishBlock(ctx, currentBlockPtrCopy, l)
 			if err != nil {
-				return nil, errors.Wrap(err, fmt.Sprintf("error shipping block to backend, blockID %s", currentBlockPtrCopy.meta.BlockID.String()))
+				return nil, fmt.Errorf(fmt.Sprintf("error shipping block to backend, blockID %s: ", currentBlockPtrCopy.meta.BlockID.String()), "%w", err)
```
stoewer, line 201:
```suggestion
				return nil, fmt.Errorf("error shipping block to backend, blockID %s: %w", currentBlockPtrCopy.meta.BlockID.String(), err)
```

tempodb/encoding/vparquet3/compactor.go
```diff
@@ -210,7 +210,7 @@ func (c *Compactor) Compact(ctx context.Context, l log.Logger, r backend.Reader,
 		currentBlock.meta.EndTime = maxBlockEnd
 		err := c.finishBlock(ctx, currentBlock, l)
 		if err != nil {
-			return nil, errors.Wrap(err, fmt.Sprintf("error shipping block to backend, blockID %s", currentBlock.meta.BlockID.String()))
+			return nil, fmt.Errorf(fmt.Sprintf("error shipping block to backend, blockID %s: ", currentBlock.meta.BlockID.String()), "%w", err)
```
stoewer, line 213:
```suggestion
			return nil, fmt.Errorf("error shipping block to backend, blockID %s: %s", currentBlock.meta.BlockID.String(), err)
```

tempodb/tempodb.go
```diff
@@ -321,12 +320,12 @@ func (rw *readerWriter) Find(ctx context.Context, tenantID string, id common.ID,
 		r := rw.getReaderForBlock(meta, curTime)
 		block, err := encoding.OpenBlock(meta, r)
 		if err != nil {
-			return nil, errors.Wrap(err, fmt.Sprintf("error opening block for reading, blockID: %s", meta.BlockID.String()))
+			return nil, fmt.Errorf(fmt.Sprintf("error opening block for reading, blockID: %s: ", meta.BlockID.String()), "%w", err)
```
stoewer, line 323:
```suggestion
			return nil, fmt.Errorf("error opening block for reading, blockID: %s: %w", meta.BlockID.String(), err)
```

tempodb/tempodb.go
```diff
@@ -321,12 +320,12 @@ func (rw *readerWriter) Find(ctx context.Context, tenantID string, id common.ID,
...
 		}
 
 		foundObject, err := block.FindTraceByID(ctx, id, opts)
 		if err != nil {
-			return nil, errors.Wrap(err, fmt.Sprintf("error finding trace by id, blockID: %s", meta.BlockID.String()))
+			return nil, fmt.Errorf(fmt.Sprintf("error finding trace by id, blockID: %s: ", meta.BlockID.String()), "%w", err)
```
stoewer, line 328:
```suggestion
			return nil, fmt.Errorf("error finding trace by id, blockID: %s: %w", meta.BlockID.String(), err)
```

stoewer (commented):
Thank you for contributing. The PR looks quite good already. I left a couple of comments.

cmd/tempo/app/app.go
sonisr: AFAIK about how `errors.Join()` would parse the error message, it uses a newline separator. I'm under the assumption we're looking to have single-line rather than multi-line errors.

E.g.
```
error writing status endpoints: <err>

vs.

<err_one>
<err_two>
```

Curious if `dskit` would format it similar to the latter?

stoewer: Your are right, using dskit multi error or errors.Join will both alter the format or the error string.
Regarding dskit: for 5 errors the multierror creates a string that looks like this:
```
5 errors: err1; err2; err3; err4; err5
```
I think it should be fine to use dskit multierror here. But it is also completely OK to keep the `joinErrors`.

sonisr: Maybe we leave it as-is for now, for consistency's sake? I'd be happy to revisit it again in the future, if need be.

tempodb/encoding/vparquet2/block_findtracebyid.go
```diff
@@ -242,13 +242,13 @@ func findTraceByID(ctx context.Context, traceID common.ID, meta *backend.BlockMe
 	r := parquet.NewReader(pf)
 	err = r.SeekToRow(rowMatch)
 	if err != nil {
-		return nil, errors.Wrap(err, "seek to row")
+		return nil, fmt.Errorf("seek to row:: %w", err)
```
stoewer, line 245:
```suggestion
		return nil, fmt.Errorf("seek to row: %w", err)
```

tempodb/encoding/vparquet2/block_findtracebyid.go
```diff
@@ -242,13 +242,13 @@ func findTraceByID(ctx context.Context, traceID common.ID, meta *backend.BlockMe
...
 	}
 
 	tr := new(Trace)
 	err = r.Read(tr)
 	if err != nil {
-		return nil, errors.Wrap(err, "error reading row from backend")
+		return nil, fmt.Errorf("error reading row from backend:: %w", err)
```
stoewer, line 251:
```suggestion
		return nil, fmt.Errorf("error reading row from backend: %w", err)
```

stoewer (approved):
LGTM
PRs like this are always super tedious with all those little changes. Thank you for contributing it.

Issue #2973, opened 2023-10-01
metrics-generator: Option to disable adding X-Scope-OrgID in remote write requests


vineetjp:
**Describe the bug**

When using metrics-generator to remote write metrics to Amazon Managed Prometheus (Cortex), the authentication to the server fails. Support case has been logged with AWS

**To Reproduce**
1. Start Tempo 2.21 in multi tenant mode with metrics-generator enabled and storage config with remote-write to amazon managed prometheus server.
2. Start metric ingestion . Following error will appear in metrics generator logs
`ts=2023-09-30T11:50:02.769098012Z caller=dedupe.go:112 tenant=Amway component=remote level=error remote_name=e3f5ba url=https://aps-workspaces.ap-southeast-1.amazonaws.com/workspaces/xxx/api/v1/remote_write msg="non-recoverable error" count=500 exemplarCount=0 err="server returned HTTP status 403 Forbidden: {\"message\":\"The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.\\n\\nThe Canonical String for this request should have been\\n'POST\\n/workspaces/x/api/v1/remote_write\\n\\ncontent-encoding:snappy\\ncontent-type:application/x-protobuf\\nhost:aps-workspaces.ap-southeast-1.amazonaws.com\\nx-amz-date:20230930T115002Z\\nx-amz-security-token:xxx//////////xxx//////////xxx/xxx"`

**Expected behavior**
It will be better if X-Scope-OrgID header in remote write request can be turned off.

**Environment:**
 - Infrastructure: Kubernetes
 - Deployment tool: helm

**Additional Context**
In order to verify that X-Scope-OrgID header is causing the issue, a prometheus server's remote write configuration was modified to add X-Scope-OrgID header. After removing the X-Scope-OrgID header configuration in the remote write configuration, remote writing was completed successfully with the same sigv4 arn.

Prometheus Server remote write config
`metricsGenerator:
  enabled: true
  config:
    storage:
      remote_write:
      - url: https://aps-workspaces.ap-southeast-1.amazonaws.com/workspaces/x/api/v1/remote_write
        sigv4:
          region: ap-southeast-1
          role_arn: arn:aws:iam::xxxx:role/xxx`

Tempo metrics generator helm config
`metricsGenerator:
  enabled: true
  config:
    storage:
      remote_write:
[...]

PR #2974, opened 2023-10-01
metrics-generator disable x-scope-orgid header append

vineetjp:
**What this PR does**:
Adds option to disable addition of X-Scope-OrgID header in metrics generator with remote writing
Fixes #1554

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/configuration/_index.md
+++ b/docs/sources/tempo/configuration/_index.md
@@ -376,6 +376,9 @@ metrics_generator:
         # How long to wait when flushing samples on shutdown
         [remote_write_flush_deadline: <duration> | default = 1m]
 
+        # Whether to remove X-Scope-OrgID header in remote write requests
+        [remote_write_remove_org_id_header: <bool> | default = false]
+
         # A list of remote write endpoints.
         # https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write
         remote_write:
--- a/docs/sources/tempo/configuration/_index.md
+++ b/docs/sources/tempo/configuration/_index.md
@@ -376,8 +376,8 @@ metrics_generator:
         # How long to wait when flushing samples on shutdown
         [remote_write_flush_deadline: <duration> | default = 1m]
 
-        # Whether to remove X-Scope-OrgID header in remote write requests
-        [remote_write_remove_org_id_header: <bool> | default = false]
+        # Whether to add X-Scope-OrgID header in remote write requests
+        [remote_write_add_org_id_header: <bool> | default = true]
 
         # A list of remote write endpoints.
         # https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write
--- a/modules/generator/storage/config.go
+++ b/modules/generator/storage/config.go
@@ -17,6 +17,9 @@ type Config struct {
 	// How long to wait when flushing sample on shutdown
 	RemoteWriteFlushDeadline time.Duration `yaml:"remote_write_flush_deadline"`
 
+	// Add X-Scope-OrgID header in remote write requests
+	RemoteWriteRemoveOrgIDHeader bool `yaml:"remote_write_remove_org_id_header,omitempty"`
+
 	// Prometheus remote write config
 	// https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write
 	RemoteWrite []prometheus_config.RemoteWriteConfig `yaml:"remote_write,omitempty"`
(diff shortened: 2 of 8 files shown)


joe-elliott:
Fixes #2973

modules/generator/storage/config_util.go
```diff
@@ -36,6 +36,15 @@ func generateTenantRemoteWriteConfigs(originalCfgs []prometheus_config.RemoteWri
 			cloneCfg.Headers[user.OrgIDHeaderName] = tenant
 		}
 
+		if removeOrgIDHeader {
```
joe-elliott, line 39: can we just not add the header above instead of adding it and removing it here?

docs/sources/tempo/configuration/_index.md
```diff
@@ -376,6 +376,9 @@ metrics_generator:
         # How long to wait when flushing samples on shutdown
         [remote_write_flush_deadline: <duration> | default = 1m]
 
+        # Whether to remove X-Scope-OrgID header in remote write requests
+        [remote_write_remove_org_id_header: <bool> | default = false]
```
joe-elliott, line 380: would prefer an "add org id" setting that defaults to true

joe-elliott (commented):
thanks for this improvement! 

A couple of small comments, but overall it's looking good.

docs/sources/tempo/configuration/_index.md
vineetjp: @joe-elliott I have changed the config name to remote_write_add_org_id_header. https://github.com/grafana/tempo/pull/2974/commits/2ac2622c01979c901674f3e7610e3ed5af5819db

modules/generator/storage/config_util.go
vineetjp: @joe-elliott  Thanks for the review.
 Changes has been committed to skip addition of OrgIDHeader if config to add id header is set to false, please review the same https://github.com/grafana/tempo/pull/2974/commits/2ac2622c01979c901674f3e7610e3ed5af5819db

modules/generator/storage/config.go
```diff
@@ -17,6 +17,9 @@ type Config struct {
 	// How long to wait when flushing sample on shutdown
 	RemoteWriteFlushDeadline time.Duration `yaml:"remote_write_flush_deadline"`
 
+	// Add X-Scope-OrgID header in remote write requests
+	RemoteWriteAddOrgIDHeader *bool `yaml:"remote_write_add_org_id_header,omitempty"`
```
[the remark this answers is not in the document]
vineetjp: @joe-elliott RemoteWriteAddOrgIDHeader is changed to normal bool, please review ab2e9fd9aa4f2892213164736402b8e1d5d29f0f

modules/generator/storage/config_util.go
```diff
@@ -33,7 +33,9 @@ func generateTenantRemoteWriteConfigs(originalCfgs []prometheus_config.RemoteWri
 				}
 			}
 
-			cloneCfg.Headers[user.OrgIDHeaderName] = tenant
+			if addOrgIDHeader {
```
kvrhdn, line 36: I'm wondering if we should move this if-case higher and merge it with `if tenant != util.FakeTenantID` at line 24.
Make it:

```go
if tenant != util.FakeTenantID && addOrgIDHeader {
```

Reasoning: we only need to remove any pre-existing X-Scope-OrgID headers if we want to inject our own. If we are not injecting any, there won't be a conflict and we don't need to remove them either.

We have the following possible scenarios:

- single-tenant: we don't remove X-Scope-OrgID
- multi-tenant with `addOrgIDHeader = true`: we remove X-Scope-OrgID headers and inject our own
- multi-tenant with `addOrgIDHeader = false`: we do not inject our own X-Scope-OrgID header so I don't think we need to remove them as well

kvrhdn (commented):
Changes look good overall, just left a comment about removing pre-existing X-Scope-OrgID headers

modules/generator/storage/config_util.go
vineetjp: @kvrhdn I agree with your point of view. addOrgIDHeader = false make metrics generator more lenient on remote write authentication then it should also allow X-Scope-OrgID that a user wants to pass using remote write config. The suggested changes has been committed 364bdc39351837042c9c37b0430dfd091f4fc493

kvrhdn (approved):
LGTM! I'll give Joe a chance to check in again as well 🙂

joe-elliott (approved):
Nice addition. Thanks!

PR #2986, opened 2023-10-05
Fix error returns that didn't check `err != nil`

mapno:
**What this PR does**:

In https://github.com/grafana/tempo/pull/2955 tempo replaced usage of github.com/pkg/errors
with errors package from the standard library.

In that process errors.Wrap calls were replaced with fmt.Errorf,
and there a difference in the way both work with `nil` error.

e.g: `errors.Wrap(nil, "test")` is `<nil>`, but `fmt.Errorf("test", nil)` is `test%!(EXTRA <nil>)`

so we need to guard fmt.Errorf calls to avoid returning errors when `err` is `nil`.


electron0zero (approved):

mapno (commented):
Thanks @electron0zero

PR #2993, opened 2023-10-06
TraceQL: Experimental "not structural" operators

### LINKED ISSUE #1811: Negated structural query operators for TraceQL
Envisage a solution that expects a authentication span `{ .action = "authenticate" }` to be active whilst allowing descendant privileged actions `{ .action = "privileged" }`. Given such a configuration, being able to query for traces where privileged actions had occurred **without** an ancestor authentication span would allow for the potential identification of security vulnerabilities.

Currently the [structural operators](https://github.com/grafana/tempo/blob/main/docs/design-proposals/2022-04%20TraceQL%20Concepts.md#structural-operators) only allow querying for positive aspects, i.e., `{ .action = "authenticate" } >> { .action = "privileged" }` would bring back traces where the desired behaviour **had** occurred - but the above failure case would not be easy to catch. Including negated operators (i.e., !>>, !> and !~) in TraceQL would allow for such queries:

    { .action = "authenticate" } !>> { .action = "privileged" }

In some cases, similar/equivalent functionality could be achieved by searching for traces that contain the privileged action but not the authentication action (i.e., `{ action = "privileged" } && { action != "authenticate" }`, or if direct child relationship is expected, `{ action != "authenticate" } > { action = "privileged" }` would likely work. But both of these don't meet the more general case, and the former would fail to catch privileged actions before or after the authentication span had ended.

joe-elliott:
**What this PR does**:
- Adds five experimental "not structural" operators: `!>`, `!>>`, `!<`, `!>>`, and `!~`. 
- Consolidates structural operator code
- Extends the lexer to be able to handle 2+ combination tokens.

Fixes #1811

**Checklist**
- [x] Tests updated
- [x] Documentation added


docs/sources/tempo/traceql/_index.md
```diff
@@ -208,20 +208,44 @@ The second expression returns no traces because it's impossible for a single spa
...
-These spanset operators look at the structure of a trace and the relationship between the spans.
+These spanset operators look at the structure of a trace and the relationship between the spans. Structural operators ALWAYS return
+matches from the right hand side of the operator.
 
 - `{condA} >> {condB}` - The descendant operator (`>>`) looks for spans matching `{condB}` that are descendants of a span matching `{condA}`
 - `{condA} << {condB}` - The ancestor operator (`<<`) looks for spans matching `{condB}` that are ancestor of a span matching `{condA}`
 - `{condA} > {condB}` - The child operator (`>`) looks for spans matching `{condB}` that are direct child spans of a parent matching `{condA}`
 - `{condA} < {condB}` - The parent operator (`<`) looks for spans matching `{condB}` that are direct parent spans of a child matching `{condA}`
-- `{condA} ~ {condB}` - The sibling operator (`~`) checks that spans matching `{condA}` and `{condB}` are siblings of the same parent span.
+- `{condA} ~ {condB}` - The sibling operator (`~`) looks that spans matching `{condB}` that have at least one sibling matching `{condA}`.
```
knylander-grafana, line 218: Should this be looks at spans instead of looks that? 
```suggestion
- `{condA} ~ {condB}` - The sibling operator (`~`) looks at spans matching `{condB}` that have at least one sibling matching `{condA}`.
```

docs/sources/tempo/traceql/_index.md
```diff
@@ -208,20 +208,44 @@ The second expression returns no traces because it's impossible for a single spa
...
 
 For example, to find a trace where a specific HTTP API interacted with a specific database:
 
 ```
 { span.http.url = "/path/of/api" } >> { span.db.name = "db-shard-001" }
 ```
 
+### Experimental Structural
```
knylander-grafana, line 226: I added the standard wording that we use for an experimental feature or capability. What do you think? 
```suggestion
### Experimental structural

{{% admonition type="warning" %}}
These experimental spanset operators is an [experimental feature](/docs/release-life-cycle/). Engineering and on-call support is not available. Documentation is either limited or not provided outside of code comments. No SLA is provided.  
{{% /admonition %}}
```

docs/sources/tempo/traceql/_index.md
```diff
@@ -208,20 +208,44 @@ The second expression returns no traces because it's impossible for a single spa
...
+
+These spanset operators look at the structure of a trace and the relationship between the spans. They are marked experimental b/c they
+are known to sometimes return false positives. However, they can be very useful and we encourage users to try them and give feedback.
```
knylander-grafana, line 229: We say that these operators can be useful. Do we need to elaborate more on that or are the provided examples enough? 
```suggestion
These spanset operators look at the structure of a trace and the relationship between the spans. They are marked experimental because they
are known to sometimes return false positives. However, they can be very useful (see examples below). We encourage users to try them and give feedback.
```

docs/sources/tempo/traceql/_index.md
```diff
@@ -208,20 +208,44 @@ The second expression returns no traces because it's impossible for a single spa
...
+
+- `{condA} !>> {condB}` - The not descendant operator (`!>>`) looks for spans matching `{condB}` that are not descendant spans of a parent matching `{condA}`
+- `{condA} !<< {condB}` - The not ancestor operator (`!<<`) looks for spans matching `{condB}` that are not ancestor spans of a child matching `{condA}`
+- `{condA} !> {condB}` - The not child operator (`!>`) looks for spans matching `{condB}` that are not direct child spans of a parent matching `{condA}`
+- `{condA} !< {condB}` - The not parent operator (`!<`) looks for spans matching `{condB}` that are not direct parent spans of a child matching `{condA}`
+- `{condA} !~ {condB}` - The not sibling operator (`!~`) looks that spans matching `{condB}` that do not have at least one sibling matching `{condA}`.
```
knylander-grafana, line 235: Suggesting hyphens for the compound adjectives. The words like "not-descendant" and "not-ancestor" function as descriptors for operator, so I've made them compound adjectives by adding the hyphen. 
```suggestion
- `{condA} !>> {condB}` - The not-descendant operator (`!>>`) looks for spans matching `{condB}` that are not descendant spans of a parent matching `{condA}`
- `{condA} !<< {condB}` - The not-ancestor operator (`!<<`) looks for spans matching `{condB}` that are not ancestor spans of a child matching `{condA}`
- `{condA} !> {condB}` - The not-child operator (`!>`) looks for spans matching `{condB}` that are not direct child spans of a parent matching `{condA}`
- `{condA} !< {condB}` - The not-parent operator (`!<`) looks for spans matching `{condB}` that are not direct parent spans of a child matching `{condA}`
- `{condA} !~ {condB}` - The not-sibling operator (`!~`) looks that spans matching `{condB}` that do not have at least one sibling matching `{condA}`.
```

knylander-grafana (commented):
Thank you for adding documentation with examples! Awesome.

docs/sources/tempo/traceql/_index.md
joe-elliott: I don't mind adding a warning, but these are OSS docs. I don't think it makes sense to talk about SLAs or support.

pkg/traceql/ast_execute.go
```diff
@@ -100,76 +103,54 @@ func (o SpansetOperation) evaluate(input []*Spanset) (output []*Spanset, err err
...
-			}
-
-			if len(spans) > 0 {
-				// Clone here to capture previously computed aggregates, grouped attrs, etc.
-				// Copy spans to new slice because of internal buffering.
-				matchingSpanset := input[i].clone()
-				matchingSpanset.Spans = append([]Span(nil), spans...)
-				output = append(output, matchingSpanset)
 			}
+			falseForAll = o.Op == OpSpansetNotParent
```
mdisibio, line 138: Could the falseForAlls be set in the prior cases before the fallthrough?

pkg/traceql/lexer.go
```diff
@@ -167,18 +171,34 @@ func (l *lexer) Lex(lval *yySymType) int {
 		return FLOAT
 	}
 
-	tokStrNext := l.TokenText() + string(l.Peek())
-	if tok, ok := tokens[tokStrNext]; ok {
-		l.Next()
-		l.parsingAttribute = startsAttribute(tok)
-		return tok
+	// look for combination tokens starting with 2 and working up til there is no match
```
mdisibio, line 174: Interesting.  I think this works because `!<` is a valid prefix of `!<<` (for example)?  If we had a three character token that didn't overlap a two character token, not sure it would work.  Not an issue, just thinking out loud and maybe a comment to explain for next time.

docs/sources/tempo/traceql/_index.md
```diff
@@ -208,20 +208,44 @@ The second expression returns no traces because it's impossible for a single spa
...
+For example, to find a trace with a leaf span in the service "foo":
+
+```
+{ } !< { resource.service.name = "foo" }
+```
+
+To find a span that is the last error in a series of cascading errors:
+
+```
+{ status != error } !< { status = error }
```
mdisibio, line 246: Is this example 100% right?  I think the both clauses should be `status = error`, and then the query would be "find errors that aren't parents of other errors"

mdisibio (approved):
Awesome addition, good cleanup and tests.  Left a few small thoughts, but going ahead and approving.

pkg/traceql/lexer.go
joe-elliott: yes, this is to disambiguate common prefixes for 2 and 3 char tokens. it's a good point that it would not find 3 char tokens that did not have a 2 char prefix.

pkg/traceql/ast_execute.go
joe-elliott: if we set it before the fallthrough won't it just get overwritten in the next case? i might be confused about what you're asking for. Something like?
```
		case OpSpansetDescendant:
			falseForAll = false
			fallthrough
		case OpSpansetNotDescendant:
                        falseForAll = true
			relFn = <some func>
```
we could init it to `true` and then set to false in all of the cases like `OpSpansetDescendant` ?

docs/sources/tempo/traceql/_index.md
joe-elliott: good catch. fixing.

pkg/traceql/ast_execute.go
mdisibio: Was thinking something like:
```go
		case OpSpansetNotDescendant:
			falseForAll = true
			fallthrough
		case OpSpansetDescendant:
			relFn = <some func>
```

I didn't check the order of each case pair

joe-elliott: oh. yeah. that does read better.

PR #3011, opened 2023-10-11
Metrics summary API root spans

mdisibio:
**What this PR does**:
Also keeps root spans in the local-blocks processor so that things like `traceDuration`, `rootServiceName` can work in the metrics summary API.  It's best effort and has the same limitations as other cross-span traceql features:  when the root span and matching span are not in the same segment (wal flush or complete block) then it doesn't work.

Also adds a metric to track spans after filtering.

Fixes n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/processor/localblocks/metrics.go
+++ b/modules/generator/processor/localblocks/metrics.go
@@ -19,6 +19,12 @@ var (
 		Name:      "traces_total",
 		Help:      "Total number of traces created",
 	}, []string{"tenant"})
+	metricTotalSpans = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: namespace,
+		Subsystem: subsystem,
+		Name:      "spans_total",
+		Help:      "Total number of spans after filtering",
+	}, []string{"tenant"})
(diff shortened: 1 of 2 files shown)


joe-elliott (approved):

09jvilla:
> so that things like traceDuration, rootServiceName can work in the metrics summary API. 

When you say "things" does this mean "all intrinsics"? What is specific about `traceDuration` and `rootServiceName` that made them problematic? The fact that they're trace level intrinsics?

09jvilla:
I see. 

Before, pretend we had traceA and traceB...

traceA
- root span id=123, kind=client, env=prod, servicename=foo
      - child span id=567, kind=server, env=prod, dc=dc1

traceB
- root span id=333, kind=server, env=prod, servicename=foo
      - child span id=777, kind=server, env=prod, dc=dc2

and then we did

{ rootServiceName=foo } | aggregateGy (dc)

Previously we would have not counted TraceA because the root span was kind=client, which meant that our aggregate statistics only considered relevant spans from TraceB and not traceA?

PR #3012, opened 2023-10-11
Fix: pass-through to runtime overrides for FilterPolicies and TargetInfoExcludedDimensions

electron0zero:
**What this PR does**:

When userconfigurableoverrides are enabled, we were returning metrics_generator `filter_policies` as empty even when it's set in per tenant runtime overrides.

This is a bug, we should return per tenant runtime overides when filter_policies are not set at userconfigurableoverrides.

this PR added a len > 0 check to fix this for multiple attribuets, and adds a test for all overrides when userconfigurableoverrides is enabled.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## main / unreleased
+* [BUGFIX] Fix check for zero length attributes correctly in user configurable overrides [#3012](https://github.com/grafana/tempo/pull/3012) (@electron0zero)
 * [ENHANCEMENT] Unescape tag names [#2894](https://github.com/grafana/tempo/pull/2894) (@fabrizio-grafana)
 * [FEATURE] New TraceQL structural operators ancestor (<<), parent (<) [#2877](https://github.com/grafana/tempo/pull/2877) (@kousikmitra)
 * [ENHANCEMENT] Add support for searching by span status message using  `statusMessage` keyword [#2848](https://github.com/grafana/tempo/pull/2848) (@kousikmitra)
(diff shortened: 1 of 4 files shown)


electron0zero:
> > this PR added a len > 0 check to fix this for multiple attributes, and adds a test for all overrides when userconfigurableoverrides is enabled.
> 
> This is actually intentional in the API: if you set an override to `[]` you are explicitly setting it to an empty list and it should take priority over overrides set in the configmap.
> 
> So if you pass this to the API:
> 
> ```
> {
>   "metrics_generator": {
>     "processor": {
>       "service_graphs": {
>         "peer_attributes": []
>       }
>     }
>   }
> }
> ```
> 
> And the configmap has:
> 
> ```
> {
>   "metrics_generator": {
>     "processor": {
>       "service_graphs": {
>         "peer_attributes": ["db.name", "db.system", "db.url"]
>       }
>     }
>   }
> }
> ```
> 
> The merged result should be:
> 
> ```
> {
>   "metrics_generator": {
>     "processor": {
>       "service_graphs": {
>         "peer_attributes": []
>       }
>     }
>   }
> }
> ```
> 
> `PeerAttributes` is a `*[]string` to allows us to distinguish between "empty list" and "not set".
> 
> https://github.com/grafana/tempo/blob/67fd1710b6781c4f39ffceefcd9ccb2a0f11cdbc/modules/overrides/userconfigurable/client/limits.go#L88

makes sense, I was not aware of this. looks like this is not the bug, and but is in the pass through code. updated the fix now.

kvrhdn (approved):

PR #3013, opened 2023-10-11
Add generate jsonnet option to analyse blocks

ie-pham:
**What this PR does**: Add a new tag to allow generation of overrides for dedicated columns in the form of jsonnet

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


cmd/tempo-cli/main.go
```diff
@@ -22,8 +22,11 @@ const (
 	bloomFilePrefix = "bloom-"
 )
 
+var generateJsonnet = false
```
mdisibio, line 25: Instead of a global var this could be an option in `analyseBlockCmd`.  Or do you see it being useful for other commands?

cmd/tempo-cli/cmd-analyse-block.go
```diff
@@ -334,6 +343,34 @@ func printSummary(scope string, max int, summary genericAttrSummary) error {
 	return w.Flush()
 }
 
+func printDedicatedColumnOverridesJsonnet(spanSummary genericAttrSummary, resourceSummary genericAttrSummary) {
+	fmt.Println("")
+	fmt.Printf("parquet_dedicated_columns: [\n")
+	numColumnsSpan := 10
+	if numColumnsSpan > len(spanSummary.attributes) {
```
mdisibio, line 350: Can we make topN handle the case where there aren't 10 attributes?

mdisibio (approved):

PR #2928, opened 2023-09-13
Add spanmetrics, servicegraph histogram buckets to user-configurable overrides api

mar4uk:
**What this PR does**:
Extends the user-configurable overrides added in https://github.com/grafana/tempo/pull/2543 with histogram buckets override.

New JSON payload:
```
{
    "forwarders": [],
    "metrics_generator": {
        "processors": [
            "service-graphs",
            "span-metrics"
        ],
        "disable_collection": false,
        "collection_interval": "60s",
        "processor": {
            "service_graphs": {
                "dimensions": [
                    "service.namespace"
                ],
                "enable_client_server_prefix": true,
                "peer_attributes": [
                    "peer.service",
                    "db.name",
                    "db.system",
                    "db.url"
                ],
                "histogram_buckets": [0, 0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10]
            },
            "span_metrics": {
                "dimensions": [
                    "service.namespace"
                ],
                "enable_target_info": true,
                "histogram_buckets": [0, 0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10],
                "filter_policies": [
                    {
                        "include": {
                            "match_type": "strict",
                            "attributes": [
                                {
                                    "key": "span.kind",
                                    "value": "SPAN_KIND_SERVER"
                                }
                            ]
                        }
                    },
                    {
                        "include": {
                            "match_type": "strict",
                            "attributes": [
                                {
                                    "key": "span.kind",
                                    "value": "SPAN_KIND_CLIENT"
                                }
                            ]
                        }
                    }
                ]
            }
        }
    }
}
```

**Which issue(s) this PR fixes**:
~Fixes #<issue number>~

**Checklist**
- [x] Tests updated
- [ ] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #3018, opened 2023-10-12
Add missing fields in overrides API merged response

kvrhdn:
**What this PR does**:

While reviewing https://github.com/grafana/tempo/pull/3012 I noticed some fields weren't being populated correctly. This was a massive oversight on my part as the PR adding the merge logic (#2915) was merged during the same time window as PRs adding extending the limits struct (#2899, #2928, #2945, ).
When merged the code got merged without issues but the merge function wasn't updated.

Fields that weren't being populated:
- `metrics_generator.collection_interval`
- `metrics_generator.service_graphs.histogram_buckets`
- `metrics_generator.span_metrics.histogram_buckets`
- `metrics_generator.span_metrics.target_info_excluded_dimensions`

Before this change they would always be returned from the result even though a value would be set.

~~Fixes #<issue number>~~

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/overrides/userconfigurable/api/limits_test.go
+++ b/modules/overrides/userconfigurable/api/limits_test.go
@@ -1,0 +1,119 @@
+package api
+
+import (
+	"encoding/json"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/grafana/tempo/modules/overrides"
+	filterconfig "github.com/grafana/tempo/pkg/spanfilter/config"
+)
+
+func Test_limitsFromOverrides(t *testing.T) {
+	userID := "foo"
+
+	cfg := overrides.Config{
+		Defaults: overrides.Overrides{
+			Forwarders: []string{"my-forwarder"},
+			MetricsGenerator: overrides.MetricsGeneratorOverrides{
+				Processors:         map[string]struct{}{"service-graphs": {}},
+				CollectionInterval: 15 * time.Second,
+				DisableCollection:  true,
+				Processor: overrides.ProcessorOverrides{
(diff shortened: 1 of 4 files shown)


modules/overrides/userconfigurable/api/limits_test.go
```diff
@@ -0,0 +1,119 @@
...
+		},
+	}
+	overridesInt, err := overrides.NewOverrides(cfg)
+	assert.NoError(t, err)
+
+	limits := limitsFromOverrides(overridesInt, userID)
+	limitsJson, err := json.MarshalIndent(limits, "", "  ")
+	assert.NoError(t, err)
+
+	expectedJson := `{
```
mdisibio, line 61: Does this test pick up if we forget to add a new field to `limitsFromOverrides` ?  I think it does since the output json will get a null but just checking.

mdisibio (approved):
Left one question and I think the answer is yes and lgtm.

modules/overrides/userconfigurable/api/limits_test.go
kvrhdn: It does because `overrides.MetricsGeneratorOverrides` will set the new value to the zero value which will be omitted in the json and you get a diff.

This what happens if I remove collection_interval from both the input overrides and the expected result:

```
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -9,2 +9,3 @@
        	            	     "disable_collection": true,
        	            	+    "collection_interval": "0s",
        	            	     "processor": {
        	Test:       	Test_limitsFromOverrides
--- FAIL: Test_limitsFromOverrides (0.00s)
```

PR #3019, opened 2023-10-12
Correctly return RetryInfo on resource exhausted

joe-elliott:
**What this PR does**:
The otel collector requires a "RetryInfo" to be sent back with an error to retry on "ResourceExhausted". This is the primary error code returned when Tempo refuses spans due to rate limit. 

https://github.com/open-telemetry/opentelemetry-collector/blob/d7b49df5d9e922df6ce56ad4b64ee1c79f9dbdbe/exporter/otlpexporter/otlp.go#L172

**Checklist**
- [x] Tests updated
- [x] Documentation added


docs/sources/tempo/configuration/_index.md
```diff
@@ -204,6 +204,11 @@ distributor:
     # Disables write extension with inactive ingesters. Use this along with ingester.lifecycler.unregister_on_shutdown = true
     #  note that setting these two config values reduces tolerance to failures on rollout b/c there is always one guaranteed to be failing replica
     [extend_writes: <bool>]
+
+    # Optional.
+    # Configures the time sent back in a gRPC RetryInfo struct when Tempo returns a GRPC ResourceExhausted. If this
+    # parameter is set to 0 the client may not retry ResourceExhausted errors.
```
knylander-grafana, line 210:
```suggestion
    # parameter is set to 0, the client may not retry ResourceExhausted errors.
```

knylander-grafana (commented):
Thank you for updating the doc! :)

modules/distributor/config.go
```diff
@@ -64,6 +68,7 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
 	cfg.DistributorRing.KVStore.Store = "memberlist"
 	cfg.DistributorRing.HeartbeatTimeout = 5 * time.Minute
 
+	cfg.RetryAfterOnResourceExhausted = time.Second
```
mdisibio, line 71: Do you think we should default to 0?  I guess I just have concerns/questions about how agents/collectors will behave and things going awry when the new Tempo is installed.  Since tempo always returns "retryable" on rate limiting, do they have a max number of retries or will it turn into an infinite loop?

joe-elliott: I was hoping to default to non-zero b/c it seemed to me that the default behavior should be retrying. I do think there's some sense in defaulting to 0 to keep current behavior though which people have likely become accustomed to.

> Since tempo always returns "retryable" on rate limiting, do they have a max number of retries or will it turn into an infinite loop?

This change will make the agent/collector honor it's own [retry settings](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md) which includes a max cutoff. In practice you will see some log messages like "retrying due to resource exhausted" and then "max retries exceeded. data dropped".

joe-elliott: As per discussions. Changing this to default to 0 which means retries are disabled by default.

mdisibio (approved):

PR #3025, opened 2023-10-14
Created strongly typed PolicyMatch struct from policy config to match values

### LINKED ISSUE #3015: Metrics generator regex policy filter compiles on each attribute value match
Metrics generator has a filter policy with match type `regex`. Generator is iterating over attributes and performs regex compilation every time it tries to match value. This results in high memory consumption.  
https://github.com/grafana/tempo/blob/main/pkg/spanfilter/spanfilter.go#L85

Compile regex once and perform match with already compiled regex. 

Used `strict` match type on the policy, which works well and memory consumption is stable. Although I have a use-case where I would like to use `regex`, but can't afford the memory consumption.

**Additional context**
Using `strict` type used to consume just over 1.2GB of memory. Changing to config presented below resulted in ~60GB of memory being used.
```
filter_policies:
  - include:
      match_type: regex
      attributes:
        - key: resource.telemetry.sdk.version
          value: 1\.1[6-9]\.[0-9]
        - key: service.namespace
          value: payments|access
```

andriusluk:
**What this PR does**:
Introduced `AttributePolicyMatch` & `IntrinsicPolicyMatch` structures to match span attributes based on strongly typed values & precompiled regex which are used to match span attribute values.

Fixes #3015 

**Checklist**
- [x] Tests updated
- [x] Documentation added


CHANGELOG.md
```diff
@@ -74,6 +74,7 @@ defaults:
   metrics_generator:
     processors: [service-graphs, span-metrics]
 ```  
+* [BUGFIX] Introduced `PolicyMatch` struct with attributes having strongly typed values and precompiled regexp which are used to match span attribute values. [#3025](https://github.com/grafana/tempo/pull/3025) (@andriusluk)
```
knylander-grafana, line 77: struct = structure? Spelling it out for when your note is included. 
```suggestion
* [BUGFIX] Introduced `PolicyMatch` structure with attributes having strongly typed values and precompiled regexp which are used to match span attribute values. [#3025](https://github.com/grafana/tempo/pull/3025) (@andriusluk)
```

pkg/spanfilter/policymatch/policymatch.go
```diff
@@ -0,0 +1,282 @@
...
+			matches++
+		}
+	}
+
+	return len(p.attributes) == matches
+}
+
+// MatchPolicyAttribute is an attribute that must match a span for the span to match the policy.
+type MatchPolicyAttribute interface {
+	Key() string
```
joe-elliott, line 108: i'm trying to think of a way to simplify this interface. i think it will benefit the code using the MatchPolicyAttributes.

Could it be as simple as:
```
type matchPolicyAttribute interface {
  matchesSpan(*span)    // used by MatchIntrinsicAttrs
  matchesAttribute(*attr)  // used by MatchAttrs
}
```

wdyt?

joe-elliott (commented):
thx for the improvements! 

as far as i can tell there should be no change to a user's config correct? the old config will parse and work just fine?

this seems like it would result in some nice perf improvements. dropping the reflection and only doing the regex once. if you had time for benchmarks that would be cool, but no worries if you're busy.

i have one suggestion regarding the interface. i think it would help me see the structure of code better. do you think it makes sense?

zalegrala (commented):
This looks like a nice change to me, good work.  Thanks for the PR.

pkg/spanfilter/policymatch/policymatch.go
andriusluk: Very good suggestion. I will do an update to this PR.

CHANGELOG.md
```diff
@@ -74,6 +74,7 @@ defaults:
   metrics_generator:
     processors: [service-graphs, span-metrics]
 ```  
+* [BUGFIX] Introduced `PolicyMatch` structure with attributes having strongly typed values and precompiled regexp which are used to match span attribute values. [#3025](https://github.com/grafana/tempo/pull/3025) (@andriusluk)
```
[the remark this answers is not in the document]
andriusluk: You are correct, this makes makes more sense

pkg/spanfilter/policymatch/policymatch.go
andriusluk: I took  a bit different direction, but with the same idea in mind. I have created a separate structure IntrinsicPolicyMatch to perform matches span's intrinsic attributes.

andriusluk: Attaching benchmarks here.
```
goos: darwin
goarch: arm64
pkg: github.com/grafana/tempo/pkg/spanfilter
                                        │    o.txt     │                n.txt                 │
                                        │    sec/op    │    sec/op     vs base                │
SpanFilter_applyFilterPolicyNone-8        4.447m ±  9%   4.444m ±  2%        ~ (p=0.436 n=10)
SpanFilter_applyFilterPolicySmall-8       57.37m ± 10%   54.05m ±  4%   -5.79% (p=0.000 n=10)
SpanFilter_applyFilterPolicyMedium-8      92.36m ±  7%   54.05m ± 14%  -41.48% (p=0.000 n=10)
SpanFilter_applyFilterPolicyRegex-8       67.96m ±  4%   53.47m ±  7%  -21.33% (p=0.000 n=10)
SpanFilter_applyFilterPolicyIntrinsic-8   28.64m ±  2%   22.68m ±  2%  -20.84% (p=0.000 n=10)
geomean                                   34.06m         27.50m        -19.26%

```

joe-elliott: nice!

pkg/spanfilter/policymatch/attribute.go
```diff
@@ -0,0 +1,163 @@
...
+	attr := AttributeFilter{
+		key: key,
+	}
+
+	switch v := value.(type) {
+	case string:
+		attr.stringValue = v
+		if code, exists := tracev1.Status_StatusCode_value[v]; exists {
+			attr.typ = StatusCodeAttributeFilter
+			attr.statusCodeValue = tracev1.Status_StatusCode(code)
```
joe-elliott, line 109: i don't think if the value happens to be a status code or kind we should change to a status code/kind filter. shouldn't this always be a string filter? aren't status code/kind intrinsic filters?

pkg/spanfilter/splitpolicy.go
```diff
@@ -0,0 +1,91 @@
...
+		attr, err := traceql.ParseIdentifier(pa.Key)
+		if err != nil {
+			return nil, err
+		}
+
+		if attr.Intrinsic > 0 {
+			var attribute policymatch.IntrinsicFilter
+			if policy.MatchType == config.Strict {
+				value := pa.Value.(string)
+				if code, ok := tracev1.Status_StatusCode_value[value]; ok {
```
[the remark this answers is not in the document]
andriusluk: I have reverted it to handle filter creation in a similar fashion as it was before

pkg/spanfilter/spanfilter.go
```diff
@@ -1,46 +1,42 @@
...
-	v1_common "github.com/grafana/tempo/pkg/tempopb/common/v1"
 	v1 "github.com/grafana/tempo/pkg/tempopb/resource/v1"
-	v1_trace "github.com/grafana/tempo/pkg/tempopb/trace/v1"
-	"github.com/grafana/tempo/pkg/traceql"
+	tracev1 "github.com/grafana/tempo/pkg/tempopb/trace/v1"
 )
 
 type SpanFilter struct {
-	filterPolicies []*filterPolicy
+	filterPolicies []filterPolicy
```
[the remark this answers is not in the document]
andriusluk: There was no particular reason, I forgot to undo ti after experimenting with implementations. Reverted the change

andriusluk:
@joe-elliott  `tempodb` tests seem to be timing out for some reason. I see the same timeout on the other PRs as well.

```
=== Failed
=== FAIL: tempodb  (0.00s)
panic: test timed out after 20m0s
...
```

pkg/spanfilter/policymatch/attribute.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+	}
+
+	return true
+}
+
+func matchesAnyFilter(pa AttributeFilter, attrs []*commonv1.KeyValue) bool {
+	for _, attr := range attrs {
+		// If the attribute key does not match, then it cannot match the policy.
+		if pa.key != attr.Key {
```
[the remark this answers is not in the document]
joe-elliott: you're right. for each filter you are checking all attributes to see if any attribute matches it. the broader loop is something like:
```
for each filter
  for each attr
    check if matches attr matches filter
  if no attrs match return false // every filter has to find one matching attribute or we fail

return true // all filters found one matching attribute so we match
```

You could invert the loop, use a map on filters and do some shenanigans where you see if all filters were found at least one match by the end of the loop. I don't know if that would be faster or not with a real workload. I'm fine with it as written

pkg/spanfilter/policymatch/attribute.go
```diff
@@ -0,0 +1,140 @@
...
+	case int64:
+		attr.typ = Int64AttributeFilter
+		attr.int64Value = v
+	case float64:
+		attr.typ = Float64AttributeFilter
+		attr.float64Value = v
+	case bool:
+		attr.typ = BoolAttributeFilter
+		attr.boolValue = v
+	}
```
joe-elliott, line 115: i think we should default to returning an error here. that way if someone tries to create a filter for a type we don't support it will be apparent

pkg/spanfilter/policymatch/intrinsic.go
```diff
@@ -0,0 +1,87 @@
...
+	return IntrinsicFilter{intrinsic: traceql.IntrinsicName, name: value}
+}
+
+// NewRegexpIntrinsicFilter returns a new IntrinsicFilter that matches spans based on the given regex and intrinsic.
+func NewRegexpIntrinsicFilter(intrinsic traceql.Intrinsic, regex string) (IntrinsicFilter, error) {
+	r, err := regexp.Compile(regex)
+	if err != nil {
+		return IntrinsicFilter{}, err
+	}
+	return IntrinsicFilter{intrinsic: intrinsic, regex: r}, nil
```
joe-elliott, line 63: since we only support name/status/kind below, i think this should return an error if `intrinsic` is an unsupported (i.e. not name/status/kind)

pkg/spanfilter/splitpolicy.go
```diff
@@ -0,0 +1,115 @@
...
+type splitPolicy struct {
+	// ResourceMatch is a set of resource attributes that must match a span for the span to match the policy.
+	ResourceMatch *policymatch.AttributePolicyMatch
+	// SpanMatch is a set of span attributes that must match a span for the span to match the policy.
+	SpanMatch *policymatch.AttributePolicyMatch
+	// IntrinsicMatch is a set of intrinsic attributes that must match a span for the span to match the policy.
+	IntrinsicMatch *policymatch.IntrinsicPolicyMatch
+}
+
+func newSplitPolicy(policy *config.PolicyMatch) (*splitPolicy, error) {
```
joe-elliott, line 25: i find this a bit tough to read. some suggestions on cleanup below. some basic tests to confirm that a config becomes the expected split policy would probably be good too.

pkg/spanfilter/splitpolicy.go
```diff
@@ -0,0 +1,115 @@
...
+	for _, pa := range policy.Attributes {
+		attr, err := traceql.ParseIdentifier(pa.Key)
+		if err != nil {
+			return nil, err
+		}
+
+		if attr.Intrinsic > 0 {
+			var filter policymatch.IntrinsicFilter
+			if policy.MatchType == config.Strict {
+				switch attr.Intrinsic {
```
joe-elliott, line 39: we could create one method here `NewIntrinsicFilter()` and move the switch down into it to make this code a bit clearer.

pkg/spanfilter/splitpolicy.go
```diff
@@ -0,0 +1,115 @@
...
+				case traceql.IntrinsicKind:
+					switch v := pa.Value.(type) {
+					case tracev1.Span_SpanKind:
```
joe-elliott, line 42: is it possible for `pa.Value` to be a SpanKind? I feel like the only valid values will be strings

pkg/spanfilter/splitpolicy.go
```diff
@@ -0,0 +1,115 @@
...
+							return nil, fmt.Errorf("currently unsupported status intrinsic string value: %s; supported values: %v", v, maps.Keys(tracev1.Status_StatusCode_value))
+						}
+					default:
+						return nil, fmt.Errorf("currently unsupported intrinsic: %v", v)
+					}
+				case traceql.IntrinsicName:
+					filter = policymatch.NewNameIntrinsicFilter(pa.Value.(string))
+				}
+			} else {
+				filter, err = policymatch.NewRegexpIntrinsicFilter(attr.Intrinsic, pa.Value.(string))
```
joe-elliott, line 70: we should avoid the unsafe cast `pa.Value.(string)`. I believe it would be possible to type an integer into the config and panic the application. let's do the ol' safe cast: `s, ok := pa.Value.(string)`

joe-elliott (commented):
It's looking good. I think the code in attribute.go/intrinsic.go is much more straightforward with your changes.

Looking for some cleanup in splitpolicy.go. Other than that I think this is good to go. Given how we much we rely on this feature I will run it in some on some of our internal clusters after the next round of changes. If everything is looking good then I think this will be ready.

joe-elliott:
> @joe-elliott tempodb tests seem to be timing out for some reason. I see the same timeout on the other PRs as well.

Yup, this times out ~occasionally~ regularly. Don't worry about it. I will re-run in CI that is unrelated to your PR.

pkg/spanfilter/splitpolicy.go
andriusluk: I guess not, will remove this case

PR #3029, opened 2023-10-17
Improve Structural Operators Perf

joe-elliott:
**What this PR does**:
The PR improves performance by using a standard sort/search procedure but that requires having both the LHS and RHS of the operator at the fetch layer (where it is aware of nested set implementation details). The cost is a very ungainly change to the span interface:

```
	SiblingOf([]Span, []Span, bool, bool, []Span) []Span
	DescendantOf([]Span, []Span, bool, bool, []Span) []Span
	ChildOf([]Span, []Span, bool, bool, []Span) []Span
```

These methods no longer operate on the span itself which is strange. However, there is no other clean places to place it that can be easily accessed deep in the engine where the evaluations occur.

Other fixes:
- Fixed an issue where `{} !~ {}` would not find the root span. The way the fix is implemented multiple root spans will be considered siblings using a query like `{} ~ {}`
- Fixed the bridge iterator sort to use the correct definition level

Benchmarks to follow.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


joe-elliott:
Benchmarks show _significant_ improvements on the worst case and even performance in the normal case. 

**Queries**
```
{"sibling", "{ resource.service.name = `loki-querier` } ~ { resource.service.name = `loki-querier` }"},
{"desc", "{ resource.service.name = `loki-querier` } >> { resource.service.name = `loki-querier` }"},
{"child", "{ resource.service.name = `loki-querier` } > { resource.service.name = `loki-querier` }"},
```
```
name                           old time/op    new time/op    delta
BackendBlockTraceQL/sibling-8     956ms ± 6%     841ms ± 2%  -12.05%  (p=0.008 n=5+5)
BackendBlockTraceQL/desc-8        1.04s ± 7%     1.08s ±12%     ~     (p=0.548 n=5+5)
BackendBlockTraceQL/child-8       1.06s ± 8%     1.12s ±10%     ~     (p=0.222 n=5+5)

name                           old speed      new speed      delta
BackendBlockTraceQL/sibling-8  14.9MB/s ± 6%  16.9MB/s ± 2%  +13.56%  (p=0.008 n=5+5)
BackendBlockTraceQL/desc-8     15.1MB/s ± 6%  14.6MB/s ±11%     ~     (p=0.548 n=5+5)
BackendBlockTraceQL/child-8    14.9MB/s ± 7%  14.1MB/s ±10%     ~     (p=0.222 n=5+5)

name                           old MB_io/op   new MB_io/op   delta
BackendBlockTraceQL/sibling-8      14.2 ± 0%      14.2 ± 0%     ~     (all equal)
BackendBlockTraceQL/desc-8         15.7 ± 0%      15.7 ± 0%     ~     (all equal)
BackendBlockTraceQL/child-8        15.7 ± 0%      15.7 ± 0%     ~     (all equal)

name                           old alloc/op   new alloc/op   delta
BackendBlockTraceQL/sibling-8    50.4MB ± 3%    48.0MB ±14%     ~     (p=0.286 n=4+5)
BackendBlockTraceQL/desc-8       44.5MB ± 5%    50.8MB ±18%  +14.18%  (p=0.016 n=5+5)
BackendBlockTraceQL/child-8      46.3MB ± 8%    43.8MB ± 0%     ~     (p=0.690 n=5+5)

name                           old allocs/op  new allocs/op  delta
BackendBlockTraceQL/sibling-8      109k ± 3%      109k ± 4%     ~     (p=0.905 n=4+5)
BackendBlockTraceQL/desc-8         107k ± 0%      111k ± 6%   +3.95%  (p=0.008 n=5+5)
BackendBlockTraceQL/child-8        106k ± 0%      107k ± 0%   +0.83%  (p=0.008 n=5+5)
```
**Queries**
```
{"sibling", "{ resource.service.name = `loki-querier` } ~ { name != `` }"},
{"desc", "{ resource.service.name = `loki-querier` } >> { name != `` }"},
{"child", "{ resource.service.name = `loki-querier` } > { name != `` }"},
```
```
name                           old time/op    new time/op     delta
BackendBlockTraceQL/sibling-8     1.82s ± 7%      1.48s ±16%  -18.61%  (p=0.016 n=5+5)
BackendBlockTraceQL/desc-8        1.84s ± 3%      1.68s ± 6%   -9.08%  (p=0.008 n=5+5)
BackendBlockTraceQL/child-8       1.88s ± 4%      1.58s ± 5%  -15.88%  (p=0.008 n=5+5)

name                           old speed      new speed       delta
BackendBlockTraceQL/sibling-8  8.47MB/s ± 7%  10.45MB/s ±15%  +23.35%  (p=0.016 n=5+5)
BackendBlockTraceQL/desc-8     12.2MB/s ± 3%   13.4MB/s ± 6%  +10.06%  (p=0.008 n=5+5)
BackendBlockTraceQL/child-8    11.9MB/s ± 4%   14.2MB/s ± 4%  +18.86%  (p=0.008 n=5+5)

name                           old MB_io/op   new MB_io/op    delta
BackendBlockTraceQL/sibling-8      15.4 ± 0%       15.4 ± 0%     ~     (all equal)
BackendBlockTraceQL/desc-8         22.4 ± 0%       22.4 ± 0%     ~     (all equal)
BackendBlockTraceQL/child-8        22.5 ± 0%       22.5 ± 0%     ~     (all equal)

name                           old alloc/op   new alloc/op    delta
BackendBlockTraceQL/sibling-8     267MB ±43%      218MB ±73%     ~     (p=0.381 n=5+5)
BackendBlockTraceQL/desc-8        153MB ± 1%      153MB ± 9%     ~     (p=0.730 n=4+5)
BackendBlockTraceQL/child-8       152MB ± 3%      150MB ± 3%     ~     (p=0.968 n=5+5)

name                           old allocs/op  new allocs/op   delta
BackendBlockTraceQL/sibling-8     3.41M ± 8%      3.30M ±10%     ~     (p=0.952 n=5+5)
BackendBlockTraceQL/desc-8        3.15M ± 0%      3.15M ± 0%   +0.03%  (p=0.008 n=5+5)
BackendBlockTraceQL/child-8       3.15M ± 0%      3.15M ± 0%   +0.03%  (p=0.008 n=5+5)
```

**Queries**
These queries don't even return on main. A single pass takes ~5 minutes. Results are for this branch only.
```
{"sibling", "{ name != `` } ~ { name != `` }"},
{"desc", "{ name != `` } >> { name != `` }"},
{"child", "{ name != `` } > { name != `` }"},
```
```
name                           time/op
BackendBlockTraceQL/sibling-8     1.86s ± 6%
BackendBlockTraceQL/desc-8        3.35s ± 4%
BackendBlockTraceQL/child-8       2.06s ± 3%

name                           speed
BackendBlockTraceQL/sibling-8  10.2MB/s ± 6%
BackendBlockTraceQL/desc-8     6.11MB/s ± 4%
BackendBlockTraceQL/child-8    10.0MB/s ± 3%

name                           MB_io/op
BackendBlockTraceQL/sibling-8      19.0 ± 0%
BackendBlockTraceQL/desc-8         20.5 ± 0%
BackendBlockTraceQL/child-8        20.5 ± 0%

name                           alloc/op
BackendBlockTraceQL/sibling-8    1.36GB ± 1%
BackendBlockTraceQL/desc-8       1.42GB ± 0%
BackendBlockTraceQL/child-8      1.43GB ± 1%

name                           allocs/op
BackendBlockTraceQL/sibling-8     6.13M ± 0%
BackendBlockTraceQL/desc-8        6.22M ± 0%
BackendBlockTraceQL/child-8       6.22M ± 0%
```

pkg/traceql/storage.go
```diff
@@ -85,9 +85,9 @@ type Span interface {
 	StartTimeUnixNanos() uint64
 	DurationNanos() uint64
 
-	SiblingOf(Span) bool
-	DescendantOf(Span) bool
-	ChildOf(Span) bool
+	SiblingOf([]Span, []Span, bool, bool, []Span) []Span
```
mdisibio, line 88: Do we really invert or could the engine swap lhs and rhs? I think falseForAll is a required though.

Thinking out loud on other ways to organize this:

1) Put them on the Fetcher?
2) Put them on the Spanset: `lhs.SiblingsOf(rhs, bool, bool, buffer)`. But looking at that, doesn't really seem much better.   And we'd need to make Spanset an interface.
3) Expose left/right/etc to the engine move the algorithm there.

I think only your current interface and Fetcher are practical for now.  Is there any drawback to putting it on the Fetcher?

joe-elliott: > Do we really invert or could the engine swap lhs and rhs? I think falseForAll is a required though.
 
 Unfortunately, I think we do b/c we always return the right hand side. invert just changes the direction that the relationship is evaluated. Child becomes parent. We could make 3 more functions (the inversions) and that would allow us to drop it

> Put them on the Fetcher?

My only pushback on this suggestion is that we'd have to change the pipelineElement interface to pass the fetcher down through all of the elements. It would be a rather sweeping (although straighforward) change.

```
type pipelineElement interface {
 	evaluate([]*Spanset, *SpansetFetcher) ([]*Spanset, error)
}
```

I like this suggestion about the same as I like keeping it where it is. Which is to say I don't think either option is great, but both work and are practical. Both options muddy the waters and make writing new fetch layers for the engine a bit more unwieldy.

pkg/traceql/storage.go
```diff
@@ -85,9 +85,9 @@ type Span interface {
 	StartTimeUnixNanos() uint64
 	DurationNanos() uint64
 
-	SiblingOf(Span) bool
-	DescendantOf(Span) bool
-	ChildOf(Span) bool
+	SiblingOf([]Span, []Span, bool, []Span) []Span
+	DescendantOf([]Span, []Span, bool, bool, []Span) []Span
+	ChildOf([]Span, []Span, bool, bool, []Span) []Span
```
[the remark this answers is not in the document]
joe-elliott: agreed. adding some comments :+1:

tempodb/encoding/vparquet3/nested_set_model.go
```diff
@@ -97,6 +99,7 @@ func assignNestedSetModelBounds(trace *Trace) bool {
 	for _, root := range rootNodes {
 		node := root
 		node.span.NestedSetLeft = nestedSetBound
+		node.span.ParentID = nestedSetRootParent
```
mdisibio, line 102: 1) This won't be present on existing data, does that matter?

2) Can you explain a bit which use case needs it?  Maybe NotSibling and NotAncestor? Sibling should be ok because either 0 legit or 0 unset, root nodes don't have siblings.

joe-elliott: > This won't be present on existing data, does that matter?

I think it's fine. We've marked these as experimental and it only impacts `{} !~ {}` on root spans.

NotSibling is the only one that matters. This allows it to distinguish between a root span and a span that is disconnected in the trace. Ancestor/Descendant only checks left/right so this doesn't impact that algorithm.

tempodb/encoding/vparquet3/block_traceql.go
```diff
@@ -444,7 +547,7 @@ func (i *bridgeIterator) Next() (*parquetquery.IteratorResult, error) {
 		parquetquery.ReleaseResult(res)
 
 		sort.Slice(i.nextSpans, func(j, k int) bool {
-			return parquetquery.CompareRowNumbers(DefinitionLevelResourceSpans, i.nextSpans[j].rowNum, i.nextSpans[k].rowNum) == -1
+			return parquetquery.CompareRowNumbers(DefinitionLevelResourceSpansILSSpanAttrs, i.nextSpans[j].rowNum, i.nextSpans[k].rowNum) == -1
```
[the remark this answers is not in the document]
joe-elliott: Yeah. Agreed. There is perhaps something strange going on here. I will investigate.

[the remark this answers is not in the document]
joe-elliott: After doing some digging this has been broken forever. My change is not correct either. It should be:

`DefinitionLevelResourceSpansILSSpan`

Pushing the fix!

tempodb/encoding/vparquet3/nested_set_model.go
mdisibio: Oh sorry I meant NotParent (child, invert, falseforall).

mdisibio: Agree if it's a graceful fallback that sounds good.  To clarify it means that `{ } !~ { }` will skip existing roots (parent=0), and going forward it will pick them up (parent=-1).   Can you add a comment?

joe-elliott: So this?
```
{} !< {}
```
I think this worked before b/c child/parent pulls both NestedSetLeft and NestedSetParent and NestedSetLeft on Root was nonzero. Therefore it passed filters and made it to the engine.

The reason
```
{} !~ {}
```
did not work is b/c this only pulls nested set parent and [these lines](https://github.com/grafana/tempo/blob/419288d49dae887d8887aae24e83f3407665f8f1/tempodb/encoding/vparquet3/block_traceql.go#L1666-L1672) dropped the span. The code was basically seeing the span as having no valid attributes, so it didn't match anything and so just let 'er go.

joe-elliott: > Agree if it's a graceful fallback that sounds good. To clarify it means that { } !~ { } will skip existing roots (parent=0), and going forward it will pick them up (parent=-1). Can you add a comment?

Sorry, this comment wasn't visible when I replied. The real fix was more like:

If you requested NestedSetParent ONLY from the fetch layer it would fail to return root spans due to the filtering linked above. It drops root spans even though we have actual information about the parent of a root span. So by using -1 we are giving a clear signal to the fetch layer that we have valid information about the parent of this span b/c normally 0 means "we don't know anything".

Will add a comment.

PR #3031, opened 2023-10-17
Ingester Search honor ctx errors

mdisibio:
**What this PR does**:
Big queries sent to Ingester.SearchRecent may keep running after a context timeout.  The root cause is the Local backend wraps local file i/o which is not context aware.  This fixes the Local backend to check errors, but also addresses some handling in instance.Search to properly wait for all goroutines to complete before releasing the lock.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/backend/local/local.go
+++ b/tempodb/backend/local/local.go
@@ -41,7 +41,11 @@ func New(cfg *Config) (backend.RawReader, backend.RawWriter, backend.Compactor, 
 }
 
 // Write implements backend.Writer
-func (rw *Backend) Write(_ context.Context, name string, keypath backend.KeyPath, data io.Reader, _ int64, _ bool) error {
+func (rw *Backend) Write(ctx context.Context, name string, keypath backend.KeyPath, data io.Reader, _ int64, _ bool) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+
 	blockFolder := rw.rootPath(keypath)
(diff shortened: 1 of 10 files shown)


tempodb/backend/local/local.go
```diff
@@ -104,13 +116,21 @@ func (rw *Backend) CloseAppend(_ context.Context, tracker backend.AppendTracker)
 	return dst.Close()
 }
 
-func (rw *Backend) Delete(_ context.Context, name string, keypath backend.KeyPath, _ bool) error {
+func (rw *Backend) Delete(ctx context.Context, name string, keypath backend.KeyPath, _ bool) error {
+	if err := ctx.Err(); err != nil {
```
joe-elliott, line 120: cc @zalegrala heads up you will want this context logic in the Blocks() method you added to the local backend

joe-elliott (commented):
looking good! do we need similar protections in our search tags/values endpoints? or in the metrics generator endpoints that take traceql?

mdisibio:
> do we need similar protections in our search tags/values endpoints? or in the metrics generator endpoints that take traceql?

The local backend and wal block fixes should address them by catching the context error on the next i/o. But would appreciate another set of eyes.

joe-elliott:
> The local backend and wal block fixes should address them by catching the context error on the next i/o. But would appreciate another set of eyes.

Good call. Agreed!

joe-elliott (approved):

PR #3074, opened 2023-10-25
make the traceID label name configurable

zeitlinger:
**What this PR does**: make the traceID label name configurable, especially to make it possible to be otel compliant: https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#exemplars

**Checklist**
- [x] Tests updated
- [x] Documentation added


zeitlinger:
should I add a new config to legacy config - or adjust this [test](https://github.com/grafana/tempo/blob/cac2954e39751be6f0e044ac2c1de44659bde62d/modules/overrides/config_test.go#L227)?

modules/generator/registry/histogram.go
```diff
@@ -149,7 +149,7 @@ func (h *histogram) name() string {
 	return h.metricName
 }
 
-func (h *histogram) collectMetrics(appender storage.Appender, timeMs int64, externalLabels map[string]string) (activeSeries int, err error) {
+func (h *histogram) collectMetrics(appender storage.Appender, timeMs int64, externalLabels map[string]string, traceIDLabelName string) (activeSeries int, err error) {
```
joe-elliott, line 152: does it make sense to add this as a member of the struct instead of passing it in every time? or do we pass it in every time to handle dynamically changing values through overrides?

joe-elliott (commented):
This generally looks good to me! What is the goal here? Do we need this to be per tenant?

I'm not going to approve b/c I'd like @kvrhdn to take a look regarding how this interacts with overrides/user configurable overrides.

modules/generator/registry/histogram_test.go
```diff
@@ -295,7 +295,7 @@ func Test_histogram_concurrencyDataRace(t *testing.T) {
 	})
 
 	go accessor(func() {
-		_, err := h.collectMetrics(&noopAppender{}, 0, nil)
+		_, err := h.collectMetrics(&noopAppender{}, 0, nil, "")
```
joe-elliott, line 298: should we write a test that uses the new functionality?

modules/generator/registry/counter.go
```diff
@@ -115,7 +115,7 @@ func (c *counter) name() string {
 	return c.metricName
 }
 
-func (c *counter) collectMetrics(appender storage.Appender, timeMs int64, externalLabels map[string]string) (activeSeries int, err error) {
+func (c *counter) collectMetrics(appender storage.Appender, timeMs int64, externalLabels map[string]string, _ string) (activeSeries int, err error) {
```
joe-elliott, line 118: adding the param to satisfy an interface?

zeitlinger:
> What is the goal here? Do we need this to be per tenant?

I think ultimately we want to migrate everyone over, so I see this as a feature flag.

modules/generator/registry/histogram.go
zeitlinger: good idea - changed

modules/generator/registry/histogram_test.go
zeitlinger: done

knylander-grafana (commented):
Thank you for adding documentation. The update looks good.

zeitlinger:
@knylander-grafana fixed build

modules/overrides/user_configurable_overrides.go
```diff
@@ -215,6 +215,10 @@ func (o *userConfigurableOverridesManager) MetricsGeneratorDisableCollection(use
 	return o.Interface.MetricsGeneratorDisableCollection(userID)
 }
 
+func (o *userConfigurableOverridesManager) MetricsGenerationTraceIDLabelName(userID string) string {
+	return o.Interface.MetricsGenerationTraceIDLabelName(userID)
+}
+
```
kvrhdn, line 221: If we don't want to expose this setting through the API we can just omit this function completely.
`Interface` is a nested struct of `userConfigurableOverridesManager` so if we don't implement this function Go will automatically call the implementation on the nest struct (`Interface`).

kvrhdn (commented):
Looks good to me, no comments besides that we don't need to do anything for user-configurable overrides.

modules/overrides/user_configurable_overrides.go
zeitlinger: thanks - removed

zeitlinger:
@kvrhdn can this be merged now?

zeitlinger:
> Looks good to me, no comments besides that we don't need to do anything for user-configurable overrides.

addressed

kvrhdn (approved):
Yep, this looks good to me. Thanks!

PR #3084, opened 2023-10-30
Include statusMessage intrinsic attribute in tag search

### LINKED ISSUE #3081: `statusMessage` not included in the list of intrinsic attributes returned by `api/v2/search/tags`
**Describe the bug**
The `api/v2/search/tags` endpoint (and probably also the associated v1 version) does not include the `statusMessage` intrinsic attribute, added with [this PR](https://github.com/grafana/tempo/pull/2848) more than one month ago, in the list of tags. Partial example of response:
```
{"name":"intrinsic","tags":["duration","kind","name","rootName","rootServiceName","status","traceDuration"]}
```

**To Reproduce**
1. Start Tempo locally, latest version
2. Call the `api/v2/search/tags` endpoint (e.g., `http://localhost:3200/api/v2/search/tags`), for example from browser

**Expected behavior**
Partial output:
```
{"name":"intrinsic","tags":[...,"status","statusMessage", ...]}
```
In other words, similarly for instance to `status`, I would expect `statusMessage` to be included too in this list.

**Additional Context**
I tried to have a look by myself at Tempo's code, but I didn't manage to understand whether this is actually a real bug and in such case where is the error.

rcrowe:
**What this PR does**:

Includes `statusMessage` from the `api/v2/search/tags` endpoint.

Fixes #3081 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott:
Thanks for the PR!

It looks like you have a [test to patch up](https://github.com/grafana/tempo/actions/runs/6691657599/job/18181025671?pr=3084) and you'll need to sign the CLA in order for us to merge, but these should be easy enough.

Feel free to ignore the tempodb test failure. It is just timing out. We need to split it into multiple tests.

rcrowe:
Thanks @joe-elliott 
Hopefully, both those things have now been fixed correctly

joe-elliott:
One more think I should have mentioned! Can you add a bugfix changelog entry? Thanks :pray:

joe-elliott (approved):
Thanks for the quick bugfix!

PR #3091, opened 2023-10-31
Set GRAFANA_VERSION used throughout shared lookups

jdbaldry:
It was working in "next" because the fallback behavior is to use the page's version for any "SOMETHING_VERSION" lookup.
Therefore, the `docs/shared` lookup coincidentally targeted "next" which exists in the Grafana documentation.

Now that we have cut a new release with these docs, the released version is being used which is causing errors because `v2.3.x` is not a valid Grafana version.

### DIFF
--- a/docs/sources/tempo/_index.md
+++ b/docs/sources/tempo/_index.md
@@ -3,6 +3,8 @@ title: Tempo documentation
 description: Grafana Tempo is an open source distributed tracing backend.
 aliases:
   - /docs/tempo/
+cascade:
+  GRAFANA_VERSION: next
 ---
 
 # Tempo documentation
--- a/docs/sources/tempo/traceql/query-editor/traceql-editor.md
+++ b/docs/sources/tempo/traceql/query-editor/traceql-editor.md
@@ -11,7 +11,7 @@ title: Write TraceQL queries with the editor
 
 # Write TraceQL queries with the editor
 
-[//]: # 'Shared content for the TraceQL query editor'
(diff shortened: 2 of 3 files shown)


jdbaldry:
@knylander-grafana I'm not sure if "next" is the preferred place to look up this shared page. What do you think?

jdbaldry:
> > @knylander-grafana I'm not sure if "next" is the preferred place to look up this shared page. What do you think?
> 
> Let me make sure I understand what's going on. Are you looking at where the content for the shared content is pulled from? For example, the query editor content is shared from the grafana repository (docs/shared). If I'm reading this PR correctly, you're setting the Tempo repository to pulled shared content from the tip of main (next) from Grafana? Is that correct?
> 
> If so, that should be fine. However, if you're looking at sharing content from Tempo into Grafana, then that might be a different conversation.

Yeah that's correct. It was previously using "next" (`main` branch) when the documentation was unreleased because the Tempo docs version also happened to be "next". For the released Tempo docs, the version is "latest" or "v2.3.x" and the latter is not a valid Grafana version so is failing on pages like https://grafana.com/docs/tempo/v2.3.x/traceql/query-editor/traceql-search/ (https://github.com/grafana/tempo/blob/release-v2.3/docs/sources/tempo/traceql/query-editor/traceql-search.md?plain=1#L32).

This PR and the backport will fix that but it will also set the targeted Grafana docs version to "next" for all of the "next", "latest", and "v2.3.x" versions of Tempo docs.

knylander-grafana (approved):
Approving changes. Thank you for the explanation, Jack.

PR #3096, opened 2023-10-31
Update dskit and grpc

zalegrala:
**What this PR does**:

Here we update dskit and grpc.

* Remove go mod replace for grpc: 1.59.0 patched in https://github.com/grpc/grpc-go/pull/6704
* Update typo for https://github.com/grafana/dskit/commit/30221e73f47ec5a6d17dfbd3f4e2707ea5ec6ca8
* Update runtime config implementation for "overrides" usage: https://github.com/grafana/dskit/pull/405

**Checklist**
- [x] Tests updated
- [ ] Documentation added


joe-elliott (approved):

modules/overrides/runtime_config_overrides.go
```diff
@@ -109,7 +109,7 @@ func newRuntimeConfigOverrides(cfg Config) (Service, error) {
 			ReloadPeriod: time.Duration(cfg.PerTenantOverridePeriod),
 			Loader:       loadPerTenantOverrides(cfg.ConfigType),
 		}
-		runtimeCfgMgr, err := runtimeconfig.New(runtimeCfg, prometheus.WrapRegistererWithPrefix("tempo_", prometheus.DefaultRegisterer), log.Logger)
+		runtimeCfgMgr, err := runtimeconfig.New(runtimeCfg, "overrides", prometheus.WrapRegistererWithPrefix("tempo_", prometheus.DefaultRegisterer), log.Logger)
```
joe-elliott, line 112: what's the new string do?

zalegrala: I think it makes it so there can be more than one runtime manager.  I'm still checking on the tests to confirm we don't have a change in behavior.  I think if we just name it anything that matches the intent, then everything will be internally consistent.

PR #3115, opened 2023-11-06
[tempo-mixin] Increase notification delay for TempoTenantIndexTooOld

zalegrala:
**What this PR does**:

Be a little more forgiving about the sensitivity of  `TempoTenantIndexTooOld`.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/operations/tempo-mixin/alerts.libsonnet
+++ b/operations/tempo-mixin/alerts.libsonnet
@@ -143,7 +143,7 @@
             expr: |||
               max by (%s) (tempodb_blocklist_tenant_index_age_seconds{}) > %s
             ||| % [$._config.group_by_tenant, $._config.alerts.max_tenant_index_age_seconds],
-            'for': '5m',
+            'for': '20m',
             labels: {
               severity: 'critical',
             },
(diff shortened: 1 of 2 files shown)


operations/tempo-mixin/alerts.libsonnet
```diff
@@ -143,7 +143,7 @@
             expr: |||
               max by (%s) (tempodb_blocklist_tenant_index_age_seconds{}) > %s
             ||| % [$._config.group_by_tenant, $._config.alerts.max_tenant_index_age_seconds],
-            'for': '5m',
+            'for': '20m',
```
zalegrala, line 146: Maybe 700s is more reasonable.

electron0zero: Q: why 700s? also why 20m?

I was assuming that tempo operators cam use `max_tenant_index_age_seconds` to tune this alert? and don't have to bump `for`?.

I am not aginst the chage, just want to know the reason for the change.

zalegrala: Good questions.  The PR is because I've been noticing that this flaps quite a bit, causing alert noise when no action is required to being it back into desired state.  Perhaps there is another approach we can take here, or tweak to be slightly more targeted.  What do you think?

zalegrala: Oh 20m because I was just rusing up a PR after getting paged.  The 700s is 2x the polling cycle (600s) plus 100s for good measure in case we're on a boundary.

PR #3129, opened 2023-11-09
Add per-tenant compaction window

zalegrala:
**What this PR does**:

Add a `compaction_window` override to be used per-tenant.  This will allow large and small tenants to scale the window independently and be a little more flexible.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/overrides/config.go
+++ b/modules/overrides/config.go
@@ -230,8 +230,6 @@ func (c *Config) RegisterFlagsAndApplyDefaults(f *flag.FlagSet) {
 	f.IntVar(&c.Defaults.Read.MaxBytesPerTagValuesQuery, "querier.max-bytes-per-tag-values-query", 50e5, "Maximum size of response for a tag-values query. Used mainly to limit large the number of values associated with a particular tag")
 	f.IntVar(&c.Defaults.Read.MaxBlocksPerTagValuesQuery, "querier.max-blocks-per-tag-values-query", 0, "Maximum number of blocks to query for a tag-values query. 0 to disable.")
 
-	// f.DurationVar(&c.Defaults.Compaction.CompactionWindow.(*time.Duration), "compactor.max-compaction-window", 1*time.Hour, "Maximum duration of a search query. 0 to disable.")
-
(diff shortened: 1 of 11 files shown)


joe-elliott (commented):
is this referenceable using the old overrides config style? we might need it to be to used in our internal clusters since we haven't migrated forward.

zalegrala:
The config is bi-directional, but you can chose only new or legacy at a time.  I updated the round-trip test for this change.

modules/overrides/config.go
```diff
@@ -247,5 +250,6 @@ func (c *Config) Collect(ch chan<- prometheus.Metric) {
 	ch <- prometheus.MustNewConstMetric(metricLimitsDesc, prometheus.GaugeValue, float64(c.Defaults.Read.MaxBlocksPerTagValuesQuery), MetricMaxBlocksPerTagValuesQuery)
 	ch <- prometheus.MustNewConstMetric(metricLimitsDesc, prometheus.GaugeValue, float64(c.Defaults.Global.MaxBytesPerTrace), MetricMaxBytesPerTrace)
 	ch <- prometheus.MustNewConstMetric(metricLimitsDesc, prometheus.GaugeValue, float64(c.Defaults.Compaction.BlockRetention), MetricBlockRetention)
+	ch <- prometheus.MustNewConstMetric(metricLimitsDesc, prometheus.GaugeValue, float64(c.Defaults.Compaction.CompactionWindow), MetricCompactionWindow)
```
joe-elliott, line 253: we seem to only expose some overrides here. perhaps only the ones a user may be interested in?

my guess is that we should not include this one, but i'm not 100% sure what this method is doing. perhaps determine its intent and add a comment?

joe-elliott:
> The config is bi-directional, but you can chose only new or legacy at a time. I updated the round-trip test for this change.

Good call. I see the code now :+1:

modules/overrides/config.go
zalegrala: I was on the fence about this one.  I think the intent is to expose the config, but I somewhat wonder why metric a default value.  I think we can leave it out.

joe-elliott (approved):
lgtm. nice!

PR #3138, opened 2023-11-13
TraceQL Metrics design proposal

mdisibio:
**What this PR does**:
Adds design proposal for TraceQL language additions for metrics. The rendered version can be viewed at: https://github.com/mdisibio/tempo/blob/traceql-metrics-design-proposal/docs/design-proposals/2023-11%20TraceQL%20Metrics.md

Fixes n/a

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/design-proposals/2023-11 TraceQL Metrics.md
+++ b/docs/design-proposals/2023-11 TraceQL Metrics.md
@@ -34,7 +34,7 @@ A normal TraceQL query finds matching spans and returns data points like time, d
 `<span query...> | <metrics query...>`
 
 An important understanding is that _any_ valid TraceQL query can be turned into a metrics query. This finds the call rates to service B but only when initiated upstream by service A:  
-` {resource.service.name="A"} >> {resouce.service.name="B"} | rate()`
+`{resource.service.name="A"} >> {resouce.service.name="B"} | rate()`
 
 
 ## First Stage
--- a/docs/design-proposals/2023-11 TraceQL Metrics.md
+++ b/docs/design-proposals/2023-11 TraceQL Metrics.md
@@ -1,0 +1,112 @@
+---
+Authors: Martin Disibio (@mdisibio)
+Created: 2023 November
+Last updated: 2023-11-13
+---
+
+# TraceQL Metrics
+
+## Summary
+
+This design document describes TraceQL language additions for extracting time-series metrics like error rates and latencies from traces.   
+
+This document is **not** meant to be a complete language specification. Rather it is meant to invite the community to comment on the direction and capabilities.
+
+### Table of Contents
+
+- [Intro](#intro)
+- [First Stage](#first-stage)
+  - [Interval](#interval)
+  - [Grouping](#grouping)
+  - [Filtering](#filtering)
+- [Additional Stages](#additional-stages)
+- [Arithmetic](#arithmetic)
+- [Notes](#notes)
+  - [Step Interval](#step-interval)
+
+## Intro
+
+Let's look at an example straight away.  This query plots the error rate by service over time:  
+`{status = error } | rate() by (resource.service.name)`
+
+A normal TraceQL query finds matching spans and returns data points like time, duration, service name, status code, etc.  A metrics query at the core does the same thing: _find matching spans_, but instead of seeing individual results, they are aggregated over time to create time-series.  A metrics query is built by extending a regular query with metrics pipeline stages.  The structure looks like this:
+
+`<span query...> | <metrics query...>`
+
+An important understanding is that _any_ valid TraceQL query can be turned into a metrics query. This finds the call rates to service B but only when initiated upstream by service A:  
+` {resource.service.name="A"} >> {resouce.service.name="B"} | rate()`
+
+
+## First Stage
+
+The first stage in a metrics query is responsible for turning spans into time series. This is a unique intersection between the two data types and has specialized aggregation functions.  We use the `..._over_time()` suffix to distinguish them from their counterparts.
+
+| function                                   | description                |
+---------------------------------------------| ---------------------------- 
+|`rate()`                                    | The rate of spans per second
+|`count_over_time()`                         | The count of spans
+|`avg_over_time(<field>)`                    | The average of a numeric value like `duration` or [http.request.body.size](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/)
+|`max_over_time(<field>)`                    | The maximum of a numeric value.
+|`min_over_time(<field>)`                    | The minimum of a numeric value.
+|`quantile_over_time(<field>, q1, q2, ...)`  | The quantile of a numeric value (e.g. p95). Multiple quantiles can be requested and a time-series is created for each. |
+
+### Interval
+All aggregations take an optional interval like `30s`, `5m`, `1h`. If not specified then the interval will automatically match the step interval of the query. The interval is the last parameter of each function.
+
+`rate(5m)`  
+`quantile_over_time(duration, 0.95, 5m)`
+
+### Grouping
+All aggregations can group by one or more attributes and generate a time-series for each combination of values.  
+
+Plot the request rate of `/myapi` by user ID and HTTP status code:  
+`{ span.http.path = "/myapi" } | rate() by (span.user_id, span.http.status_code)`
+
+Plot the p95 duration of requests to service by HTTP endpoint:  
+`{ resource.service.name = "myservice" } | quantile_over_time(duration, 0.95) by (span.http.path)`
+
+### Filtering
+Aggregations can be combined with comparison operations to filter the output.
+
+Find the request rate per service but only if more than 1000 req/s:  
+`{ } | rate() by (resource.service.name) > 1000`
+
+
+## Additional Stages
+Additional metrics stages can be added to further refine and transform the time-series. These stages receive time-series and generate new time-series so they are separate from the first stage.  
+
+Most functions support the same ability to regroup using `by()` and comparison operators to refilter.  These can be thought of as instant functions that work across inputs at each point in time. Therefore they do not have an interval and do not alter the frequency of data points. If the original aggregation creates a data point every 60s, these will also output a data point every 60s.
+
+| function                        | description                |
+----------------------------------| ---------------------------- 
+| `... \| max() [by(...)]`        | Maximum value for each point in time
+| `... \| min() [by(...)]`        | Minimum value for each point in time
+| `... \| avg() [by(...)]`        | Average across values for each point in time
+| `... \| stddev() [by(...)]`     | Standard deviation across values for each point in time
+| `... \| quantile(q) [by(...)]`  | Quantile across values for each point in time
+| `... \| topk(N)`                | Return the top series for each point in time (grouping TBD)
+
+Find the highest per-pod failure rate in each cluster:  
+`{ status = error } | rate() by (cluster, pod) | max() by (cluster)`
+
+
+## Arithmetic
+Arithmetic operations between time-series can be used to calculate things like error rates.
+
+Find the 5xx error rate for an endpoint:
+```
+({ span.http.path = "/myapi" && span.http.status_code >= 500 } | rate()) 
+   /
+({ span.http.path = "/myapi" | rate())
+```
+
+Operations:
+* `*`
+* `/`
+* `+`
+* `-`
+
+## Notes
+
+### Step Interval
+The step interval is the explicit resolution of a query.  It is identical to the [Prometheus step interval](https://prometheus.io/docs/prometheus/latest/querying/api/#range-queries).  A step interval of `1m` means that a data point will be returned once every 60 seconds.  It is separate from the `interval` of an aggregation like `rate(1h)`.  The query `rate(1h)` with `step=1m` will still return a data point every 60 seconds but each data point is the rate smoothed over the previous hour.
--- a/docs/design-proposals/2023-11 TraceQL Metrics.md
+++ b/docs/design-proposals/2023-11 TraceQL Metrics.md
@@ -1,7 +1,7 @@
 ---
 Authors: Martin Disibio (@mdisibio)
 Created: 2023 November
-Last updated: 2023-11-13
+Last updated: 2023-11-27
 ---
 
 # TraceQL Metrics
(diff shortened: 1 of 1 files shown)


docs/design-proposals/2023-11 TraceQL Metrics.md
```diff
@@ -0,0 +1,112 @@
...
+`{resource.service.name="A"} >> {resouce.service.name="B"} | rate()`
+
+
+## First Stage
+
+The first stage in a metrics query is responsible for turning spans into time series. This is a unique intersection between the two data types and has specialized aggregation functions.  We use the `..._over_time()` suffix to distinguish them from their counterparts.
+
+| function                                   | description                |
+---------------------------------------------| ---------------------------- 
+|`rate()`                                    | The rate of spans per second
```
cyriltovena, line 46: The field could be optional for those when specifying one you could rate/sum that field instead of the span count.

`{ resource.service.name = "foo"} | rate(span.attribute.order_count) by (resource.pod)`

docs/design-proposals/2023-11 TraceQL Metrics.md
```diff
@@ -0,0 +1,112 @@
...
+`{ span.http.path = "/myapi" } | rate() by (span.user_id, span.http.status_code)`
+
+Plot the p95 duration of requests to service by HTTP endpoint:  
+`{ resource.service.name = "myservice" } | quantile_over_time(duration, 0.95) by (span.http.path)`
+
+### Filtering
+Aggregations can be combined with comparison operations to filter the output.
+
+Find the request rate per service but only if more than 1000 req/s:  
+`{ } | rate() by (resource.service.name) > 1000`
```
cyriltovena, line 72: Syntax wise this is a bit confusing for me. Not sure how to make it better but that is less confusing for me to read

```
({ } | rate() by (resource.service.name)) > 1000
```

Becomes even weirder when you have 

```
{ } | rate() by (resource.service.name) > {} |  rate() by (resource.bar)
```

docs/design-proposals/2023-11 TraceQL Metrics.md
```diff
@@ -0,0 +1,112 @@
...
+---------------------------------------------| ---------------------------- 
+|`rate()`                                    | The rate of spans per second
+|`count_over_time()`                         | The count of spans
+|`avg_over_time(<field>)`                    | The average of a numeric value like `duration` or [http.request.body.size](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/)
+|`max_over_time(<field>)`                    | The maximum of a numeric value.
+|`min_over_time(<field>)`                    | The minimum of a numeric value.
+|`quantile_over_time(<field>, q1, q2, ...)`  | The quantile of a numeric value (e.g. p95). Multiple quantiles can be requested and a time-series is created for each. |
+
+### Interval
+All aggregations take an optional interval like `30s`, `5m`, `1h`. If not specified then the interval will automatically match the step interval of the query. The interval is the last parameter of each function.
```
cyriltovena, line 54: 👍

docs/design-proposals/2023-11 TraceQL Metrics.md
mdisibio: In the first case the comparison operator is part of the pipeline stage: `{ } | (rate() by (resource.service.name) > 1000)` so you can drop data but keep going.  `{} | rate() > 1000 | max() > 2000`.  I think this is equivalent to the functional style `max(rate({}) > 1000) > 2000)`.

Cross-series query:  agree that syntax could be clearer.  Although in this case I'm not sure that query works since they have different labels.  We could add `on(x,y,z)` to specify the join conditions like in PromQL.  But in the first release we simplify and require same labels.

docs/design-proposals/2023-11 TraceQL Metrics.md
mdisibio: Interesting idea.  Do you mean it would inner sum the attribute and calculate rate?   

Example: `{ } | rate(span.read_bytes)` calculates read throughput?

docs/design-proposals/2023-11 TraceQL Metrics.md
```diff
@@ -0,0 +1,112 @@
...
+| `... \| stddev() [by(...)]`     | Standard deviation across values for each point in time
+| `... \| quantile(q) [by(...)]`  | Quantile across values for each point in time
+| `... \| topk(N)`                | Return the top series for each point in time (grouping TBD)
+
+Find the highest per-pod failure rate in each cluster:  
+`{ status = error } | rate() by (cluster, pod) | max() by (cluster)`
+
+
+## Arithmetic
+Arithmetic operations between time-series can be used to calculate things like error rates.
```
joe-elliott, line 94: if we're defining operators between time series, should we also discuss join semantics?

docs/design-proposals/2023-11 TraceQL Metrics.md
joe-elliott: I feel like the second example only returns one series with `resource.bar` having no value. The attribute `resource.bar` has already been aggregated away in the stage before it.

Actually, is the second example even allowed? `rate()` should require an input of a spanset pipeline. I don't think it can take metrics.

mdisibio: @joe-elliott  The second example is a comparison between series:  `>`.  Easy to miss? Which is further confirmation the syntax would be better.   You are correct if it was a normal multi-stage pipeline.   We don't plan to "backfill" the `by` clauses so `by(a) | by(b)` doesn't work you need `by(a,b) | by(b)`

joe-elliott: ah, I see! it is easy to miss. i'll point out that spanset pipelines currently require parentheses to be parsed:

https://github.com/grafana/tempo/blob/5e7023521e080153e7af83017c7cc9a0acd963c8/pkg/traceql/expr.y#L120-L124

This is not valid:
```
{ span.http.status_code = 200 } | count() > { span.http.status_code = 500 } | count()
```
It must be this:
```
({ span.http.status_code = 200 } | count()) > ({ span.http.status_code = 500 } | count())
```

I assume metrics would follow suit? iirc the `()` requirement was added to prevent ambiguity in the parser.

docs/design-proposals/2023-11 TraceQL Metrics.md
mdisibio: Agree, @cyriltovena 's [comment](https://github.com/grafana/tempo/pull/3138#discussion_r1391738304) about comparison also highlighted this gap.   Maybe we don't need the full featureset of PromQL [vector matching](https://prometheus.io/docs/prometheus/latest/querying/operators/#vector-matching).

docs/design-proposals/2023-11 TraceQL Metrics.md
```diff
@@ -0,0 +1,112 @@
...
+|`count_over_time()`                         | The count of spans
+|`avg_over_time(<field>)`                    | The average of a numeric value like `duration` or [http.request.body.size](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/)
+|`max_over_time(<field>)`                    | The maximum of a numeric value.
+|`min_over_time(<field>)`                    | The minimum of a numeric value.
+|`quantile_over_time(<field>, q1, q2, ...)`  | The quantile of a numeric value (e.g. p95). Multiple quantiles can be requested and a time-series is created for each. |
+
+### Interval
+All aggregations take an optional interval like `30s`, `5m`, `1h`. If not specified then the interval will automatically match the step interval of the query. The interval is the last parameter of each function.
+
+`rate(5m)`
```
Davenury, line 56: Wouldn't it be more prometheus-like to use `[]` to indicate interval? E.g. `rate[5m]()` or `quantile_over_time[5m](duration, 0.95)`. Then the arguments of the function wouldn't be mixed with interval

docs/design-proposals/2023-11 TraceQL Metrics.md
```diff
@@ -0,0 +1,112 @@
...
+  - [Filtering](#filtering)
+- [Additional Stages](#additional-stages)
+- [Arithmetic](#arithmetic)
+- [Notes](#notes)
+  - [Step Interval](#step-interval)
+
+## Intro
+
+Let's look at an example straight away.  This query plots the error rate by service over time:  
+`{status = error } | rate() by (resource.service.name)`
```
faustodavid, line 30: I believe adopting the logQL/promQL syntax for aggregations in traceQL would make it more familiar to the users and straightforward to adopt.

Current proposal:
`{status = error } | rate() by (resource.service.name)`

Suggested logQL/promQL Style:
`sum(rate({status = error })) by (resource.service.name)`

docs/design-proposals/2023-11 TraceQL Metrics.md
mdisibio: Hi indeed we debated that internally too.  Our rationale is:  in prometheus `[5m]` is meaningful on its own without a specific function.  Both `metric{}` and `metric{}[5m]` are valid constructs (with different types).  That doesn't apply to Tempo because there are no underlying `metric` time-series.  They must be created by these aggregation functions and in that case there is no need to make the interval special, it could be simplified to an argument.  To clarify even further:  `{.http.status_code>=500}` is valid and finds spans. But `{.http.status_code>=500}[5m]` doesn't mean anything without the aggregation.

To tweak your example it would probably look like this:  `rate([5m])` and `quantile_over_time(duration, 0.95 [5m])`.

docs/design-proposals/2023-11 TraceQL Metrics.md
mdisibio: Yep this is a good called and we invested a lot of thought here but I didn't cover it in the proposal.  Ultimately decided that Tempo can't execute these basic queries and it would require breaking changes.  If if breaking changes are required, then even better to use the pipeline structure rather than a close-but-different promql syntax.

Here is some explanation:

`sum(rate({status = error }))`

Inside this query is:

`rate({status =error})` which returns the raw series in promql/logql. These series have implied labels which are a side effect of your application instrumentation and your data pipeline.  For instance your app might include the label `userID` and your pipeline might include the label `region`.  The labels are a fundamental design aspect of the data and your infrastructure.   Tempo has no implied labels and doesn't require decisions about them or limits to cardinality.  So it's not clear what to return from the query `rate({status=error})`.  

If it returns the total rate, there is no need for `sum()`. 

If it returns "raw series" it would have to pull something from the span/resource/trace (pulling "everything" isn't feasible or useful).  The `by()` clause contains this information but it must now be attached to `rate` instead of `sum`.  

This is how we arrived at `... | rate() by ()`.

docs/design-proposals/2023-11 TraceQL Metrics.md
mdisibio: Yes this is likely but won't be certain until we try to implement the parsing changes.  Do you think the proposal should be updated to reflect this now?

docs/design-proposals/2023-11 TraceQL Metrics.md
faustodavid: Thanks for the detailed explanation. I understand the team concerns about the breaking changes and the unique aspects of Tempo.

To clarify my suggestion: My expectation was that `rate` would return all intrinsics, along with span/resource attributes, similar to Loki/logQL's behavior. This approach might not be useful in isolation, but it becomes powerful when combined with `sum`.

For instance, in the **intro-to-mlt** project, using the feature `log_received_spans` this raw data is overwhelming in Loki but when we apply the `sum` function to group this data, it transforms into actionable insights.

image

image

I understand that implementing this would require significant changes. However, I believe aligning traceQL more closely with familiar syntaxes could lower the learning curve.

mdisibio: @faustodavid To be clear the proposed traceql `rate()` func is already the summed total.  

Traceql query for the second screenshot would be:

`{status=error && name="pg.query:INSERT postgres"} | rate() by (span.db.statement)`

jcarres-mdsol:
With this current proposal would it be possible to retrieve multiple timeseries from the same set of spans?

```
{resource.service.namespace = "shopping" } | rate(), avg_over_time(), quantile_over_time(...) by (resource.service.name)
```
I think it would be very common to want to know rate and latency on the same conditions and ensure the data comes from the same original set (instead of running this as multiple queries). Also would be more efficeint to retrieve the spans once and convert them into series.

jcarres-mdsol:
It would be fantastic if there is some way to do string manipulations at some point. 
For instance maybe for reasons some of my `service.namespace` is `UserSystem` and other is `usersystem`
And I'd like to lowercase both prior to creating metrics so when I create metrics `by (resource.service.namespace)` both are counted together.

Same could be thought of splitting strings etc.

jcarres-mdsol:
And best of best would be if even more complex scenarios could be coded

```
{resource.service.namespace = "shopping" } | add(type="database") | add(type="cache") if (span.db.name="redis" or span.db.name="memcache") | rate() by (type)
```
This probably is a further extension outside this proposal, probably a "processing" stage could be developed in some future extension.

mdisibio:
@jcarres-mdsol 

> retrieve multiple timeseries from the same set of spans

I had the same thoughts and quantile_over_time taking multiple quantiles is an example where TraceQL could do it trivially.   Extending that to any arbitrary calculation was also on our radar.  Added a new section in the proposal for Future enhancements like this.

Commas are decent (because they aren't used in TraceQL yet) but the ideal solution supports embedding additional metrics pipelines and not just the one-off functions.  Maybe something akin to a pipeline-splitting syntax.  

> string manipulations

Agree this would be a great feature.  I think the `select()` stage is a good place for it like `select(service.namespace=lower(service.namespace))`.  Then it would be useable in any downstream pipeline not just metrics.  

> if (span.db.name="redis" or span.db.name="memcache")

Ok this might be too much for the poor `select()` function.  Definitely interesting. 

I think these last 2 are good thoughts but outside the scope of this proposal.  We could track them in a discussion, it will be a few iterations before we get to them.

docs/design-proposals/2023-11 TraceQL Metrics.md
mdisibio: Added this to the Future section.  It's TBD, we can define it once we add the features that need it like time-series arithmetic.

borchero:
This is exciting! From a user's POV: what can be expected from this design proposal? Does it lay the groundwork for implementation s.t. there will be an experimental implementation available some time this year? :smile:

joe-elliott:
@borchero 

> This is exciting! From a user's POV: what can be expected from this design proposal? Does it lay the groundwork for implementation s.t. there will be an experimental implementation available some time this year? 😄

An experimental implementation that supports the `rate()` function only is available now in Tempo 2.4.1. 2.5.x should have support for quantiles or other methods of aggregating durations. We are primarily focused on performance at scale right now.

PR #3163, opened 2023-11-22
links and events schema changes [wip]

ie-pham:
**What this PR does**: Add new schema for links and updated both links and events to use the generic Attribute struct. 

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/encoding/vparquet3/block_traceql_test.go
+++ b/tempodb/encoding/vparquet3/block_traceql_test.go
@@ -498,13 +498,13 @@ func fullyPopulatedTestTrace(id common.ID) *Trace {
 									{Key: "unsupported-kv-list", ValueDropped: kvListValue, ValueType: attrTypeNotSupported},
 								},
 								Events: []Event{
-									{TimeUnixNano: 1, Name: "e1", Attrs: []EventAttribute{
-										{Key: "foo", Value: []byte("fake proto encoded data. i hope this never matters")},
-										{Key: "bar", Value: []byte("fake proto encoded data. i hope this never matters")},
+									{TimeSinceStartUnixNano: 1, Name: "e1", Attrs: []Attribute{
+										{Key: "foo", Value: strPtr("fake proto encoded data. i hope this never matters")},
+										{Key: "bar", Value: strPtr("fake proto encoded data. i hope this never matters")},
 									}},
-									{TimeUnixNano: 2, Name: "e2", Attrs: []EventAttribute{}},
+									{TimeSinceStartUnixNano: 2, Name: "e2", Attrs: []Attribute{}},
 								},
-								Links: linkBytes,
+								Links: []Link{},
 								DedicatedAttributes: DedicatedAttributes{
 									String01: ptr("dedicated-span-attr-value-1"),
 									String02: ptr("dedicated-span-attr-value-2"),
--- a/tempodb/encoding/vparquet3/block_traceql_test.go
+++ b/tempodb/encoding/vparquet3/block_traceql_test.go
@@ -13,7 +13,6 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/grafana/tempo/pkg/tempopb"
-	v1_common "github.com/grafana/tempo/pkg/tempopb/common/v1"
 	v1 "github.com/grafana/tempo/pkg/tempopb/trace/v1"
 	"github.com/grafana/tempo/pkg/traceql"
 	"github.com/grafana/tempo/pkg/traceqlmetrics"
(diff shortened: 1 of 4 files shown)


tempodb/encoding/vparquet3/block_traceql_test.go
```diff
@@ -498,13 +498,17 @@ func fullyPopulatedTestTrace(id common.ID) *Trace {
...
+										Name:                   "e1",
+										Attrs: []Attribute{
+											attr("foo", "fake proto encoded data. i hope this never matters"),
+											attr("bar", "fake proto encoded data. i hope this never matters"),
+										},
+									},
+									{TimeSinceStartUnixNano: 2, Name: "e2", Attrs: []Attribute{}},
 								},
-								Links: linkBytes,
+								Links: []Link{},
```
stoewer, line 511: * `linkBytes` contained an encoded link, it would be good to define similar link data here
* `linkBytes` and `links` defined in line 393-414 is no longer needed and can be removed

tempodb/encoding/vparquet3/schema.go
```diff
@@ -150,16 +150,27 @@ type DedicatedAttributes struct {
 	String10 *string `parquet:",snappy,optional,dict"`
 }
 
-type EventAttribute struct {
-	Key   string `parquet:",snappy,dict"`
-	Value []byte `parquet:",snappy"` // Was json-encoded data, is now proto encoded data
+type Event struct {
+	TimeSinceStartUnixNano uint64      `parquet:",delta"`
```
stoewer, line 154: The suffix `UnixNano` indicates nanoseconds since unix epoch. However, this field contains nanoseconds since the span start, therefore `TimeSinceStartNano` could a better name (not 100% sure about this)

tempodb/encoding/vparquet3/schema.go
```diff
@@ -150,16 +150,27 @@ type DedicatedAttributes struct {
...
+	Name                   string      `parquet:",snappy"`
```
stoewer, line 155: Event names have probably a cardinality similar to span names. Therefore it can make sense to use dictionary encoding. 
```suggestion
	Name                   string      `parquet:",snappy,dict"`
```

tempodb/encoding/vparquet3/schema.go
```diff
@@ -150,16 +150,27 @@ type DedicatedAttributes struct {
...
+	Attrs                  []Attribute `parquet:",list"`
+	DroppedAttributesCount int32       `parquet:",snappy"`
```
stoewer, line 157: `DroppedAttributesCount` is zero for most spans, links, resources, and events. It would be interesting to run benchmarks with `parquet:",snappy,delta"` for all dropped attribute count columns.

tempodb/encoding/vparquet3/schema.go
```diff
@@ -183,7 +194,7 @@ type Span struct {
 	DroppedAttributesCount int32       `parquet:",snappy"`
 	Events                 []Event     `parquet:",list"`
 	DroppedEventsCount     int32       `parquet:",snappy"`
-	Links                  []byte      `parquet:",snappy"` // proto encoded []*v1_trace.Span_Link
+	Links                  []Link      `parquet:",list"` // proto encoded []*v1_trace.Span_Link
```
stoewer, line 197:
```suggestion
	Links                  []Link      `parquet:",list"`
```

tempodb/encoding/vparquet3/schema_test.go
```diff
@@ -99,7 +99,15 @@ func TestFieldsAreCleared(t *testing.T) {
 									{
 										// An attribute of every type
```
stoewer, line 100: No longer true
```suggestion
```

tempodb/encoding/vparquet3/schema_test.go
```diff
@@ -99,7 +99,15 @@ func TestFieldsAreCleared(t *testing.T) {
...
 										Attributes: []*v1.KeyValue{
-											{Key: "i", Value: &v1.AnyValue{Value: &v1.AnyValue_IntValue{IntValue: 123}}},
+											{Key: "event-attr", Value: &v1.AnyValue{Value: &v1.AnyValue_IntValue{IntValue: 123}}},
+										},
+									},
+								},
+								Links: []*v1_trace.Span_Link{
+									{
+										Attributes: []*v1.KeyValue{
+											// String
```
stoewer, line 109: Actually int, but I don't think the comment is needed here
```suggestion
```

stoewer (commented):
The PR looks already pretty good. I left a couple of (mostly minor) comments.
It would be great to have grafana/xk6-client-tracing#17 in order to test it properly

tempodb/encoding/vparquet3/schema.go
```diff
@@ -472,7 +483,7 @@ func traceToParquet(meta *backend.BlockMeta, id common.ID, tr *tempopb.Trace, ot
 
 				ss.Events = extendReuseSlice(len(s.Events), ss.Events)
 				for ie, e := range s.Events {
-					eventToParquet(e, &ss.Events[ie])
+					eventToParquet(e, &ss.Events[ie], s.StartTimeUnixNano)
```
stoewer, line 486: `nitpick` The function `eventToParquet()` receives the event start time in order to calculate `TimeSiceStartNano`. The reverse function ` parquetToProtoEvents()` receives the whole span. It would be more consistent to pass the span or only the start time to both (I personally favor only passing in the start time)

tempodb/encoding/vparquet3/schema.go
```diff
@@ -708,7 +726,32 @@ func parquetToProtoAttrs(parquetAttrs []Attribute, counter droppedAttrCounter, i
...
+			}
+
+			protoLinks = append(protoLinks, protoLink)
+		}
+	}
+
+	return protoLinks
+}
+
+func parquetToProtoEvents(parquetEvents []Event, span Span) []*v1_trace.Span_Event {
```
stoewer, line 754: Span is a very large struct, it might be a good idea to pass it as a pointer `*Span` (or pass the span start time instead)

stoewer (changes requested):

stoewer (approved):

PR #3166, opened 2023-11-22
Refactor caching to support multiple independent caches

joe-elliott:
**What this PR does**:
This PR refactors caches to allow for multiple independent caches that can be assigned to various roles. It lays the ground work for adding additional caches for things like parquet pages or frontend search jobs. This PR is completely backwards compatible, but does deprecate all the existing cache configuration fields in the `store:` config block.

- Replaces "shouldCache" on the backend calls with a "cacheInfo" struct that allows the cache layer to make its own decisions.
- Create a new top level "cacheProvider" module and config block (see below for an example). Different components can now request caches that fulfill different roles from the provider.
- Came with some nice cleanup that helped validate the refactor.
  - removing `writerCallback func(*backend.BlockMeta, time.Time)` from compaction
  - removing CacheControl from SearchOptions
  - deprecating CacheControl and replacing it with the idea of roles
  
**Open Questions**
- `cache_min_compaction_level` and `cache_max_block_age` were the only caching settings that were left alone. These are policy settings and I decided to leave them under `storage.trace` in the config block. Should we leave "policy" style settings in the normal config blocks? and only put caching settings in the cache provider? This is the approach I preferred but I thought it was worth discussing.

- Roles are currently configured as a yaml list, but I'm tempted to make them a pipe delimited list of roles. I think it would read better, but seems strange when yaml supports lists directly.

**New example config**
```yaml
cache:
  background_cache:
    writeback_goroutines: 5
  caches:
  - roles:
    - parquet-footer
    - parquet-column-idx
    - parquet-offset-idx
    memcached:
      host: memcached-instance
  - roles:
    - bloom
    redis:
      endpoint: redis-instance
```

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/cmd/tempo/app/config.go
+++ b/cmd/tempo/app/config.go
@@ -10,6 +10,7 @@ import (
 	"github.com/grafana/dskit/server"
 	"github.com/prometheus/client_golang/prometheus"
 
+	"github.com/grafana/tempo/modules/cache"
 	"github.com/grafana/tempo/modules/compactor"
 	"github.com/grafana/tempo/modules/distributor"
 	"github.com/grafana/tempo/modules/frontend"
@@ -53,6 +54,7 @@ type Config struct {
 	Overrides       overrides.Config        `yaml:"overrides,omitempty"`
 	MemberlistKV    memberlist.KVConfig     `yaml:"memberlist,omitempty"`
 	UsageReport     usagestats.Config       `yaml:"usage_report,omitempty"`
+	CacheProvider   cache.Config            `yaml:"cache,omitempty"`
 }
 
 func newDefaultConfig() *Config {
--- a/docs/sources/tempo/configuration/_index.md
+++ b/docs/sources/tempo/configuration/_index.md
@@ -29,6 +29,7 @@ This document explains the configuration options for Tempo as well as the detail
       - [Standard overrides](#standard-overrides)
       - [Tenant-specific overrides](#tenant-specific-overrides)
       - [Override strategies](#override-strategies)
+  - [Cache](#cache)
   - [Usage-report](#usage-report)
 
 Additionally, you can review [TLS]({{< relref "./tls" >}}) to configure the cluster components to communicate over TLS, or receive traces over TLS.
@@ -921,6 +922,7 @@ storage:
 
         # Cache type to use. Should be one of "redis", "memcached"
         # Example: "cache: memcached"
+        # Deprecated. See [cache](#cache) section below.
         [cache: <string>]
 
         # Minimum compaction level of block to qualify for bloom filter caching. Default is 0 (disabled), meaning
@@ -958,6 +960,7 @@ storage:
             [read_buffer_count: <int>]
 
             # Granular cache control settings for parquet metadata objects
+            # Deprecated. See [cache](#cache) section below.
             cache_control:
 
                 # Specifies if footer should be cached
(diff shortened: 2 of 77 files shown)


docs/sources/tempo/configuration/_index.md
```diff
@@ -1572,3 +1475,159 @@ This value is available in the the [tempo-distributed](https://github.com/grafan
...
+            # (default: 100ms)
+            [timeout: <duration>]
+
+            # Optional
+            # Maximum number of idle connections in pool.
+            # (default: 16)
+            [max_idle_conns: <int>]
+
+            # Optional
+            # period with which to poll DNS for memcache servers.
```
knylander-grafana, line 1539:
```suggestion
            # Period with which to poll DNS for memcache servers.
```

joe-elliott: Paging down through the file it seems inconsistent. I think I'd prefer starting with a capital. WDYT?

I'd be glad to update all of the lines in the caching block I just added, but I don't think I should cleanup the whole file in this PR (would add a lot of noise).

knylander-grafana:
@joe-elliott We'll probably want to update the Upgrade page with information about this breaking change.

docs/sources/tempo/configuration/_index.md
knylander-grafana: I think using the sentence case would be correct, and would be great if you want to make your section consistent. I think we have a doc issue open to make the doc structure for the configuration page the same. We can tackle the consistency across the entire document in that issue.

tempodb/tempodb.go
```diff
@@ -531,3 +497,50 @@ func includeCompactedBlock(c *backend.CompactedBlockMeta, id common.ID, blockSta
 	}
 	return includeBlock(&c.BlockMeta, id, blockStart, blockEnd, timeStart, timeEnd)
 }
+
+// createLegacyCache uses the config to return a cache and a list of roles.
+func createLegacyCache(cfg *Config, logger gkLog.Logger) (cache.Cache, []cache.Role, error) {
```
mapno, line 502: Instead of this, we could translate the legacy config to the new cache config in `modules.go` when the module is instantiated. And we wouldn't need `AddCache`.

mapno (commented):
I like the changes. Only have a couple of comments.

>cache_min_compaction_level and cache_max_block_age were the only caching settings that were left alone. These are policy settings and I decided to leave them under storage.trace in the config block. Should we leave "policy" style settings in the normal config blocks? and only put caching settings in the cache provider? This is the approach I preferred but I thought it was worth discussing.

I like it the way it is right now.

>Roles are currently configured as a yaml list, but I'm tempted to make them a pipe delimited list of roles. I think it would read better, but seems strange when yaml supports lists directly.

I actually it reads better as a list hah. I wouldn't change it.

>Trace ID index had support for caching so I replicated it in this PR. Since this feature is known to be broken, I'm wondering if I should drop support until we invest the time to improve it.

Yeah, I think we should do that. Or at least log something when that cache is enabled. It's not very well documented that that cache is not working.

tempodb/backend/azure/azure_test.go
```diff
@@ -111,10 +111,10 @@ func TestHedge(t *testing.T) {
 					assert.Equal(t, int32(1), atomic.LoadInt32(&count))
 					atomic.StoreInt32(&count, 0)
 
-					_ = w.Write(ctx, "object", backend.KeyPathForBlock(uuid.New(), "tenant"), bytes.NewReader(make([]byte, 10)), 10, false)
+					_ = w.Write(ctx, "object", backend.KeyPathForBlock(uuid.New(), "tenant"), bytes.NewReader(make([]byte, 10)), 10, nil)
 					// Write consists of two operations:
 					// - Put Block operation
-					//   https://docs.microsoft.com/en-us/rest/api/storageservices/put-block
+					//   https://docs.microso).com/en-us/rest/api/storageservices/put-block
```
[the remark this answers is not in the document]
joe-elliott: wups. nice catch

modules/cache/config.go
```diff
@@ -0,0 +1,84 @@
...
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/grafana/tempo/modules/cache/memcached"
+	"github.com/grafana/tempo/modules/cache/redis"
+	"github.com/grafana/tempo/pkg/cache"
+)
+
+// role must be pipe based and valid
```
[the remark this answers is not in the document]
joe-elliott: early note that i accidentally left in. cleaning up.

modules/storage/store.go
```diff
@@ -39,15 +40,15 @@ type store struct {
 }
 
 // NewStore creates a new Tempo Store using configuration supplied.
-func NewStore(cfg Config, logger log.Logger) (Store, error) {
+func NewStore(cfg Config, cacheProvider cache.Provider, logger log.Logger) (Store, error) {
 	statCache.Set(cfg.Trace.Cache)
```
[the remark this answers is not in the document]
joe-elliott: I've added stats for the new config in modules/cache.go. I figured we'd leave this until we removed the old cache config.

tempodb/tempodb.go
joe-elliott: Point taken, but I'm going to leave this where it is. It's legacy code that we will be removing in a version or 2 and it will feel like a wart no matter where we add it.

cmd/tempo/app/config.go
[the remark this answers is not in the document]
joe-elliott: done. code/tests added

PR #3171, opened 2023-11-23
[spanmetrics] Sanitize name in mapped dimensions

mapno:
**What this PR does**:

If not sanitized, Prometheus incompatible labels can be sent.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/processor/spanmetrics/spanmetrics_test.go
+++ b/modules/generator/processor/spanmetrics/spanmetrics_test.go
@@ -1011,7 +1011,7 @@ func TestSpanMetricsDimensionMappingMissingLabels(t *testing.T) {
 		"status_message": "OK",
 		"first_only":     "first-value",
 		"world_only":     "world-value",
-		"first/last":     "first-value->last-value",
+		"first_last":     "first-value->last-value",
 	})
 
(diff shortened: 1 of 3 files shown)


modules/generator/processor/spanmetrics/spanmetrics_test.go
```diff
@@ -1011,7 +1011,7 @@ func TestSpanMetricsDimensionMappingMissingLabels(t *testing.T) {
 		"status_message": "OK",
 		"first_only":     "first-value",
 		"world_only":     "world-value",
-		"first/last":     "first-value->last-value",
+		"first_glast":    "first-value->last-value",
```
electron0zero, line 1014: do we replace `/` with `_g`?

mapno: No, it was a typo. Fixed now.

kvrhdn (approved):
LGTM!

Since we are calling the sanitize function at multiple points, I wonder if we should move this into the registry? The registry could do it more centralized and we don't risk forgetting about it in the processors.

PR #3175, opened 2023-11-23
Add configurable remote_write headers to metrics-generator

### LINKED ISSUE #2975: metrics_generator: different header option for remote write
**Describe the bug**
Remote write setting from metrics_generator is not writing to the correct mimir instance. instead of using the header I pass in the configuration, it takes the header configured for the tempo setup.

**To Reproduce**
```
metrics_generator:
  storage:
    remote_write:
    - headers:
       X-Scope-OrgID: tracing-metrics
     remote_timeout: 30s
     url: https://xxx/api/v1/push
```
**Expected behavior**
Metrics generated by the metrics_generator should be sent to the url including the header specified in the config above instead of the header we use for the tempo stack.

mapno:
**What this PR does**:

Adds per-tenant configurable headers for remote_write requests in the metrics-generator.

It also adds envvar expansion to the overrides configuration.

Fixes #2975

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/generator/storage/instance.go
```diff
@@ -98,6 +119,44 @@ func (s *storageImpl) Close() error {
 	).Err()
 }
 
+func (s *storageImpl) watchOverrides() {
```
joe-elliott, line 122: we already watch overrides at the generator/instance.go level. does it make sense to reload these overrides there as well? or does this need to be done here?

modules/generator/storage/instance.go
```diff
@@ -98,6 +119,44 @@ func (s *storageImpl) Close() error {
...
+		case <-t.C:
+			newHeaders := s.overrides.MetricsGeneratorRemoteWriteHeaders(s.tenantID)
+			if !headersEqual(s.currentHeaders, newHeaders) {
+				level.Info(s.logger).Log("msg", "updating remote write headers")
+				s.currentHeaders = newHeaders
+				err := s.remote.ApplyConfig(&prometheus_config.Config{
+					RemoteWriteConfigs: generateTenantRemoteWriteConfigs(s.cfg.RemoteWrite, s.tenantID, newHeaders, s.cfg.RemoteWriteAddOrgIDHeader, s.logger),
+				})
+				if err != nil {
+					level.Error(s.logger).Log("msg", "Failed to update remote write headers. Remote write will continue with old headers", "err", err)
```
joe-elliott, line 137: do we know this is true? if it the prom remote storage fails to apply new configs do we know it continues with the old?

updateProcessors at the level above increments a counter metric on failure which i like. `metrics_generator_active_processors_update_failed_total`. i would love to just have all of this reloaded and counted under the same metric, but that name is pretty opinionated.

modules/generator/storage/instance.go
mapno: Either approach works.

IMO, separating overrides updates per pkg results for cleaner code. If we moved this logic to `generator/instance.go` we would leak storage specific details outside `/generator/storage` and still would have dedicated code in this file in case of a required update.

modules/generator/storage/instance.go
mapno: >do we know this is true? if it the prom remote storage fails to apply new configs do we know it continues with the old?

Yes. If the update fails, the function returns and the original queue is left untouched https://github.com/prometheus/prometheus/blob/main/storage/remote/write.go#L148-L177.

>updateProcessors at the level above increments a counter metric on failure which i like. metrics_generator_active_processors_update_failed_total. i would love to just have all of this reloaded and counted under the same metric, but that name is pretty opinionated.

Good point. Added!

joe-elliott (approved):

knylander-grafana:
@mapno It looks like we'll need docs for authentication and configuration for this feature.

ToonTijtgat2:
Dear

I'm not sure if I'm doing something wrong, but it does not seem to work for me.
I tried with the legacy config:
```max_bytes_per_trace: 1500000
      metrics_generator_processors:
      - span-metrics
      - service-graphs
      metrics_generator_remote_write_headers:
        tenant-id: prometheus-apps-app-dev```
        
This shows me in the /status/runtime-config the correct settings.
![image](https://github.com/grafana/tempo/assets/109516706/9739ada2-dda1-4d6e-9147-d2bdb14f3dda)

So I started to try with the new config
I placed it in the conf/tempo.yaml
Placed it on runtime-config/overrides.yaml
placed it under conf/overrides.yaml
It always keeps saying defaults is not found on the legacyoverrides.
The config I try to accomplice looks like this.
```overrides.yaml: |
    defaults:
      global:
        max_bytes_per_trace: 1500000
      metrics_generator:
        processors: [service-graphs, span-metrics]
        remote_write_headers:
          tenant-id: prometheus-apps-app-dev```
          
          
Can someone please help me to set this up correctly, or check what went wrong?

Thanks Toon Tijtgat

PR #3176, opened 2023-11-23
Fix handling for returned error added in f9100fb

gebn:
f9100fb added a second return value to `NewLeftJoinIterator()`, which was not handled in all places.


mapno (approved):
How did tests pass? 🤨 I totally missed this one. Thanks for fixing it!

PR #3180, opened 2023-11-27
User-configurable overrides: add optional check to block API requests that conflict with runtime overrides

kvrhdn:
**What this PR does**:

Adds an optional check that refuses setting user-configurable overrides if the tenant already has conflicting settings in runtime overrides. This is a crude check, ti doesn't check field-by-field but instead checks if any of the possible overrides are set in runtime overrides.

This check can be enabled with:

```yaml
overrides:
  user_configurable_overrides:
    enabled: true
    client:
      # ...
    api:
      check_for_conflicting_runtime_overrides: true
```
Callers can skip/bypass it by adding the query paramter `?skip-conflicting-runtime-check=true`.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/overrides/runtime_config_overrides.go
```diff
@@ -241,6 +241,14 @@ func (o *runtimeConfigOverridesManager) WriteStatusRuntimeConfig(w io.Writer, r
 	return nil
 }
 
+func (o *runtimeConfigOverridesManager) GetLevel(_ Level) Interface {
+	return o
```
mdisibio, line 245: Should it be an error to request a different level of overrides from this manager?  I.e. requesting `LevelUserConfigurable` from this will silently return the `LevelRuntime` overrides.   It looks like the [UserConfigurableOverridesManager does check](https://github.com/grafana/tempo/pull/3180/files#diff-d1960db0e556609f29af683322ba95e6cdd4da202f089cc08d14b3ecda5f4c1bR208).

modules/overrides/userconfigurable/client/duration.go
```diff
@@ -39,7 +41,13 @@ func (d *Duration) UnmarshalJSON(input []byte) error {
 	return nil
 }
 
+func (d *Duration) Fake(faker *gofakeit.Faker) (any, error) {
```
mdisibio, line 44: Are fakes used for production functionality or only tests?  If only tests, then would prefer if there is a way to keep this dependency isolated to only the tests.

modules/overrides/runtime_config_overrides.go
kvrhdn: Yeah, that's a good point. It's kind of weird because the `runtimeConfigOverridesManager` will always be the lowest level, it doesn't have a child to call like the `UserConfigurableOverridesManager` does.

Maybe I should just keep it simple and call the method `GetRuntimeOverrides()` Interface. It's less generic but I'm not sure it's really worth introducing the concept of an overrides level if it isn't used anywhere else yet...

modules/overrides/userconfigurable/api/http.go
```diff
@@ -82,13 +85,18 @@ func (a *UserConfigOverridesAPI) PostHandler(w http.ResponseWriter, r *http.Requ
 		return
 	}
 
+	skipConflictingOverridesCheck := false
+	if value, ok := r.URL.Query()[queryParamSkipConflictingOverridesCheck]; ok && len(value) > 0 {
+		skipConflictingOverridesCheck, _ = strconv.ParseBool(value[0])
```
zalegrala, line 90: Do we want to handle the error?

modules/overrides/userconfigurable/api/http.go
```diff
@@ -113,7 +121,12 @@ func (a *UserConfigOverridesAPI) PatchHandler(w http.ResponseWriter, r *http.Req
 		return
 	}
 
-	patchedLimits, version, err := a.update(ctx, userID, patch)
+	skipConflictingOverridesCheck := false
+	if value, ok := r.URL.Query()[queryParamSkipConflictingOverridesCheck]; ok && len(value) > 0 {
+		skipConflictingOverridesCheck, _ = strconv.ParseBool(value[0])
```
zalegrala, line 126: Same as above, do we want to handle the error?

modules/overrides/userconfigurable/api/http.go
```diff
@@ -82,13 +85,18 @@ func (a *UserConfigOverridesAPI) PostHandler(w http.ResponseWriter, r *http.Requ
 		return
 	}
 
+	skipConflictingOverridesCheck := false
```
zalegrala, line 88: Do we want to pull this from config instead of a default value here?

modules/overrides/user_configurable_overrides.go
```diff
@@ -39,7 +39,16 @@ type UserConfigurableOverridesConfig struct {
 	// PollInterval controls how often the overrides will be refreshed by polling the backend
 	PollInterval time.Duration `yaml:"poll_interval"`
 
-	Client userconfigurableoverrides.Config `yaml:"client"`
+	Client userconfigurableoverrides.Config   `yaml:"client"`
+	API    UserConfigurableOverridesAPIConfig `yaml:"api"`
```
zalegrala, line 43: Do we want to call for API defaults in `RegisterFlagsAndApplyDefaults` below?

modules/overrides/userconfigurable/client/limits.go
```diff
@@ -8,9 +8,9 @@ import (
 )
 
 type Limits struct {
-	Forwarders *[]string `json:"forwarders,omitempty"`
+	Forwarders *[]string `yaml:"forwarders,omitempty" json:"forwarders,omitempty"`
 
-	MetricsGenerator *LimitsMetricsGenerator `json:"metrics_generator,omitempty"`
+	MetricsGenerator LimitsMetricsGenerator `yaml:"metrics_generator,omitempty" json:"metrics_generator,omitempty"`
```
zalegrala, line 13: What's the change here by becoming a non-pointer on the struct?  I'm not sure I follow.

zalegrala (commented):
Looks like a nice start.  Just a few notes.

modules/overrides/userconfigurable/api/http.go
kvrhdn: We read the value from config in `api.go`, the http layer is just parsing the header and passing the value to `UserConfigOverridesAPI.set`.

modules/overrides/userconfigurable/api/http.go
kvrhdn: I opted to silently 'fail' but just returning an error isn't a lot of work + feels nicer

modules/overrides/userconfigurable/client/limits.go
kvrhdn: I did this change to simplify comparing structs. If each child-struct is a pointer it becomes nearly impossible to compare two empty structs as the pointer values will never match.

For example, the following two structs are both empty/unset:

```go
Limits {
  MetricsGenerator: &LimitsMetricsGenerator{
    DisableCollection: nil,
  },
}
```

```go
Limits {
  MetricsGenerator: nil,
}
```

But they are not equal. Having every child-struct behind a pointer creates so many variations of 'empty'.

Making them all values instead of pointers avoids this situation + it doesn't change parsing (I added a test to check).

modules/overrides/interface.go
```diff
@@ -18,6 +18,8 @@ type Service interface {
 	Interface
 }
 
+type Level string
```
zalegrala, line 21: I didn't see this used, did I miss it?

modules/overrides/userconfigurable/api/http.go
```diff
@@ -159,6 +181,10 @@ func writeError(w http.ResponseWriter, err error) {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
+	if ok := errors.Is(err, errConflictingRuntimeOverrides); ok {
```
zalegrala, line 184: Does this allocate `ok` or is the compiler smart enough to skip it?

modules/overrides/config_test.go
```diff
@@ -250,3 +254,33 @@ func rvCountFields(rv reflect.Value) int {
 	}
 	return n
 }
+
+func TestOverrides_AssertUserConfigurableOverridesAreASubsetOfRuntimeOverrides(t *testing.T) {
+	userConfigurableOverrides := client.Limits{}
+
+	err := gofakeit.Struct(&userConfigurableOverrides)
```
zalegrala, line 261: :+1:

zalegrala (approved):
This looks good to me.  Looking forward to it.

modules/overrides/interface.go
kvrhdn: It was used in a previous iteration of the API, nice catch!

modules/overrides/userconfigurable/api/http.go
kvrhdn: Huh, I'm not entirely sure why I created a temporary variable here 😅 I probably used a different function before that returned 2 values instead of juts a boolean 🤷🏻 

Again nice catch! Surprised the linter didn't pick this up

PR #3185, opened 2023-11-28
Design Doc: TraceQL Extensions

joe-elliott:
This PR adds a design doc aimed at the next set of functionality for the TraceQL language. We are looking for community feedback, consensus and comments before merging and beginning this work.

Please jump in and comment!

### DIFF
--- a/docs/design-proposals/2023-11 TraceQL Extensions.md
+++ b/docs/design-proposals/2023-11 TraceQL Extensions.md
@@ -1,0 +1,152 @@
+---
+Authors: Martin Disibio (@mdisibio), Joe Elliott (@joe-elliott), Jennie Pham (@ie-pham), Adrian Stoewer (@stoewer)
+Created: 2023 July - 2023 November
+Last updated: 2023-11-28
+---
+
+# TraceQL Extensions
+
+## Summary
+
+This design document describes extensions to the existing TraceQL language that allow for searching new fields and data types.
+
+This document is **not** meant to be a complete language specification. Rather it is meant to invite the community to comment on the basic structure and concepts of the language extensions.
+
+### Table of Contents
+
+- [Escaping Attribute Names](#escaping-attribute-names)
+- [Additional Scopes](#additional-scopes)
+  - [Trace](#trace)
+  - [Instrumentation Scope](#instrumentation-scope)
+  - [Event](#event)
+  - [Link](#link)
+- [Scoped Intrinsics](#scoped-intrinsics)
+- [Additional Data Types](#additional-data-types)
+  - [Arrays](#arrays)
+  - [IDs](#ids)
+- [Future Considerations](#future-considerations)
+
+## Escaping Attribute Names
+
+OpenTelemetry supports [any valid Unicode sequence](https://opentelemetry.io/docs/specs/otel/common/attribute-naming/) for attribute names. This includes characters problematic for a query language such as whitespace, mathematical symbols, various braces, etc. Previously TraceQL did not support [these problematic characters](https://github.com/grafana/tempo/blob/c00e7ef1103a8e7804b6f71fe8e90d0a2d9a9b02/pkg/traceql/lexer.go#L344) in attributes names.
+
+TraceQL will be extended to support these characters by allowing attribute names wrapped in `"` to be unparsed.
+
+`{ span."attribute with spaces" = "foo"}`
+
+TraceQL will also support the escape sequences `\\` and `\"` in an attribute name.
+
+`{ span."this is \"bad\"" = "foo"}`
+
+## Additional Scopes
+
+We have added four additional scopes to the TraceQL language: [trace](#trace), [scope](#instrumentation-scope), [event](#event), and [link](#link). See below for details.
+
+### Trace
+
+The trace scope allows a user to access intrinsic fields that exist at the trace level. There are no attributes at the trace scope. See [intrinsics](#scoped-intrinsics) for a complete list of allowed trace level intrinsics.
+
+```
+{ trace:duration > 100ms }
+```
+
+### Instrumentation Scope
+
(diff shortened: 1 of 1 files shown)


docs/design-proposals/2023-11 TraceQL Extensions.md
```diff
@@ -0,0 +1,152 @@
...
+
+The trace scope allows a user to access intrinsic fields that exist at the trace level. There are no attributes at the trace scope. See [intrinsics](#scoped-intrinsics) for a complete list of allowed trace level intrinsics.
+
+```
+{ trace:duration > 100ms }
+```
+
+### Instrumentation Scope
+
+The [Instrumentation Scope](https://github.com/open-telemetry/opentelemetry-proto/blob/ea449ae0e9b282f96ec12a09e796dbb3d390ed4f/opentelemetry/proto/common/v1/common.proto#L71) object contains information about the instrumentation used to generate telemetry data. This scope contains new intrinsics as well as attributes. It will be referenced using the `scope` keyword.
```
mapno, line 55: I'm concerned with overusing the word `scope`. We already use it as a concept in TraceQL (ie. `{span,resource}.foo`, and now it's a type of scope. Scope `scope`.

We could use `instrumentation` instead.

joe-elliott: I considered `instrumentation`, but it's so long! Since the name/version/attributes reflect the instrumentation we should probably name it something along those lines. `instrumentation` is clearer but unwieldy, but this will be rarely typed w/o auto complete so maybe it's ok?

mapno: >but this will be rarely typed w/o auto complete so maybe it's ok?

Right, rarely a user would have to type the whole thing. 

Scopes in general are a concept that many users already struggle with. Introducing the scope `scope` will only add to the confusion.

mapno: Also, besides the name collision, `instrumentation` is clearer in what it refers to. `scope` could be anything!

docs/design-proposals/2023-11 TraceQL Extensions.md
```diff
@@ -0,0 +1,152 @@
...
+Span [links](https://github.com/open-telemetry/opentelemetry-proto/blob/ea449ae0e9b282f96ec12a09e796dbb3d390ed4f/opentelemetry/proto/trace/v1/trace.proto#L242) contain attributes as well as intrinsic properties that can create connections to other spans. Note that both [events](#event) and links are different than existing scopes because there can be many events to one span. Due to this we are being purposefully conservative when defining event scopes and intend to extend functionality as we better understand use cases. See [future considerations](#future-considerations) for additional discussion.
+
+```
+{ link.foo = "bar" }
+{ link:traceID = "<hex string>" }
+```
+
+In the above cases the operators will return true if there are any links on the span that meets the specified condition.
+
+## Scoped Intrinsics
```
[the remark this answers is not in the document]
joe-elliott: Nope. No unscoped intrinsics.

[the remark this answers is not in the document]
mapno: Ok. It makes the most sense. Just wanted to make sure. Should it be documented?

docs/design-proposals/2023-11 TraceQL Extensions.md
```diff
@@ -0,0 +1,152 @@
...
+
+OpenTelemetry supports [any valid Unicode sequence](https://opentelemetry.io/docs/specs/otel/common/attribute-naming/) for attribute names. This includes characters problematic for a query language such as whitespace, mathematical symbols, various braces, etc. Previously TraceQL did not support [these problematic characters](https://github.com/grafana/tempo/blob/c00e7ef1103a8e7804b6f71fe8e90d0a2d9a9b02/pkg/traceql/lexer.go#L344) in attributes names.
+
+TraceQL will be extended to support these characters by allowing attribute names wrapped in `"` to be unparsed.
+
+`{ span."attribute with spaces" = "foo"}`
+
+TraceQL will also support the escape sequences `\\` and `\"` in an attribute name.
+
+`{ span."this is \"bad\"" = "foo"}`
```
jmichalek132, line 39: How about adding support for single quotes too, so users don't have to deal with escaping. For reference there is a plan to do that in prometheus to address full utf-8 support. Proposal [link](https://github.com/prometheus/proposals/blob/main/proposals/2023-08-21-utf8.md#syntax-examples).

joe-elliott: thanks for the heads up. i like the way they've used both `"` and `'` for their quoted attributes. it's something we should definitely keep in mind for a future improvement.

docs/design-proposals/2023-11 TraceQL Extensions.md
lmickh: The OTEL docs and concepts will still refer to it as the instrumentation scope.  I have to use `instrumentation_scope.name` in [ottl filters](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/README.md) in the otel-collector for some of our endpoints for example.  New users come across references to the instrumentation scope before they start digging around in the protos to find the `Scope[Spans|Metrics|Logs]` messages in my experience since the SDK hides the proto details from them.  For better or worse, the [docs](https://opentelemetry.io/docs/concepts/instrumentation-scope/) inform how the users think about it more than the actual proto.

joe-elliott: @lmickh what would you say you are in favor of? we're discussing this internally with our otel folks and they have pointed out that "instrumentation scope" is more closely aligned with a scope ("logical unit of code") than with instrumentation. i'm a bit torn on the choice myself.

mdisibio (approved):

Issue #2489, opened 2023-05-18
[DOC] Missing "ListObjects" permission in the S3 configuration doc


chrisduong:
**Describe the bug**

When I deployed the Tempo with this S3 configuration doc https://grafana.com/docs/tempo/latest/configuration/s3/, I faced this error

```json
{"caller": "main.go:105", "err": "failed to init module services error initialising module: usage-report: failed to initialize usage report: unexpected error from ListObjects on tempo-traces-foo: Access Denied","level":"error","msg":"error running Tempo","ts":"2023-05-18T10:09:09.908963824Z"}
```

**Expected behavior**

The S3 configuration doc should add **s3:ListObjects** action into the policy. 

**Environment:**
 - Infrastructure: K8s
 - Deployment tool: Helm
 - Tempo version `2.1.1`

joe-elliott:
cc @knylander-grafana

absolutemikex:
I am running into this now - however, I added the policy and I am using IRSA. Is there anything else I can check?

`level=info ts=2023-05-19T18:22:40.23471841Z caller=main.go:215 msg="initialising OpenTracing tracer"
level=info ts=2023-05-19T18:22:40.243715669Z caller=main.go:102 msg="Starting Tempo" version="(version=2.1.1, branch=HEAD, revision=4157d7620)"
level=error ts=2023-05-19T18:22:40.396802855Z caller=main.go:105 msg="error running Tempo" err="failed to init module services error initialising module: store: failed to create store unexpected error from ListObjects on BUCKET: Access Denied"`

absolutemikex:
I am dumb - disregard. I remove my SA config accidentally from values. Appreciate all your hard work.

Eve832:
Hi @knylander-grafana , can you please update the status of this ticket in your view in https://github.com/orgs/grafana/projects/69/

z0rc:
@datsabk if you updated to 2.2.2, most likely it's #2888.

datsabk:
@z0rc Saved my day man! I had started to doubt myself ! 

Looks like a regression due to AWS SDK version upgrade. There's nothing much happening in the Tempo Code really.

ekristen:
This is still an issue, #2888 doesn't seem to fix anything.

knylander-grafana:
I'm getting ready to update the S3 config docs to address some of this content. Here are some related S3 issues we'll take into consideration for updating the doc: 

* https://github.com/grafana/tempo/issues/2895
[...]

PR #3198, opened 2023-11-30
[Docs] s3:ListObjects does not exist

joe-elliott:
According to this thread the `s3:ListObjects` permission does not actually exist. Removing!

https://github.com/grafana/tempo/issues/2489

cc @knylander-grafana

### DIFF
--- a/docs/sources/tempo/configuration/s3.md
+++ b/docs/sources/tempo/configuration/s3.md
@@ -29,7 +29,6 @@ The following IAM policy shows minimal permissions required by Tempo, where the 
                 "s3:PutObject",
                 "s3:GetObject",
                 "s3:ListBucket",
-                "s3:ListObjects",
                 "s3:DeleteObject",
                 "s3:GetObjectTagging",
                 "s3:PutObjectTagging"


mdisibio (approved):

electron0zero (approved):

shayal01 (commented):
this doesn't work.even if we give full access to an iam user, with iam policy for full access of s3 ,s3accesspoint policy with all the actions and and bucket policy delegating the access to the s3accesspoint .the following error throws up when i am using in the rosa classic cluster`msg="error running Tempo" err="failed to init module services: error initialising module: store: failed to create store: unexpected error from ListObjects on tempo-openshift: The specified method is not allowed against this `.and i think there is no more listobject action

vsomwanshi:
I've Tempo operator `0.14.1-2 provided by Red Hat` installed in my environment, i've given all the necessary permission i.e `s3:*` and created a separate policy and attached to my IAM user still getting similar error `msg="error running Tempo" err="failed to init module services: error initialising module: store: failed to create store: unexpected error from ListObjects on <bucket-name>: Get \"http:/<bucket-name>.s3.us-east-2.amazonaws.com/?delimiter=%2F&encoding-type=url&prefix=\": dial tcp 3.5.130.179:80: i/o timeout"`.

PR #3199, opened 2023-12-02
New example for integration with external docker-compose application

avpps:


joe-elliott:
We really appreciate this PR, but does this add anything besides a customer docker network? We'd prefer the examples show off different elements of configuring Tempo.

PR #3202, opened 2023-12-06
Don't increment overrides reload failed counter when request was cancelled

yvrhdn:
**What this PR does**:

If reloading failed due to a context cancelled error, don't increment the error counter. If the request is being cancelled it probably is because the pod is being shutdown.

~~Fixes #<issue number>~~

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott:
Does this happen when the client cancelled it's request?

kvrhdn:
> Does this happen when the client cancelled it's request?

I think this can only happen when the pod is being shut down. This code is part of the service's `running` function which is being executed on a ticker.
I think it's just really unfortunate timing this triggered an alert. Each pod shutting down should at most report a single failed reload.

PR #3208, opened 2023-12-08
Allow grpc dial config for querier worker

### LINKED ISSUE #1596: Configuration for querier.frontend_worker.grpc_client_config.max_recv_msg_size doesn't seem to work?
# Describe the bug

When trying to query very large traces in grafana it gives the follow error.

## Error message in grafana
```
failed to get trace with id: 19d5517f5f92a4c58286b227773b81cc 
Status: 500 Internal Server Error Body: error querying ingesters in Querier.FindTraceByID: 
rpc error: 
code = ResourceExhausted 
desc = grpc: received message after decompression larger than max (104857601 vs. 104857600)
```

This is in spite of the following being in my configuration file for tempo

## tempo config yaml
```yaml

server:
  http_listen_port: 3200
  grpc_server_max_recv_msg_size: 200_000_000
  grpc_server_max_send_msg_size: 200_000_000

querier:
  frontend_worker:
    grpc_client_config:
      max_recv_msg_size: 200_000_000
      max_send_msg_size: 200_000_000
```

Is there another component that also needs to updated with this new value?

# To Reproduce
1. Run tempo docker container with the configuration above and generate a very large trace.
```
docker.io/grafana/tempo                 latest        b11ca097701b  5 hours ago   81.6 MB
```
2. Query the trace in grafana

# Expected behavior

**Environment:**
 - Infrastructure: bare-metal
 - Deployment tool: docker-compose

If the query on grafana fails with this size limit I would expect it to use the limit configured.

zalegrala:
**What this PR does**:

Without this change, the querier worker does not honor the send/recv max message size.

Fixes #1596

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/querier/worker/worker.go
+++ b/modules/querier/worker/worker.go
@@ -251,6 +251,11 @@ func (w *querierWorker) connect(ctx context.Context, address string) (*grpc.Clie
 		return nil, err
 	}
 
+	opts = append(opts, grpc.WithDefaultCallOptions(
+		grpc.MaxCallRecvMsgSize(w.cfg.GRPCClientConfig.MaxRecvMsgSize),
+		grpc.MaxCallSendMsgSize(w.cfg.GRPCClientConfig.MaxSendMsgSize),
+	))
+
 	conn, err := grpc.DialContext(ctx, address, opts...)
 	if err != nil {
 		return nil, err


joe-elliott:
We're setting these on the query frontend server. Does GRPC negotiate the settings between the 2? Did we see something that makes us think we should set this?

zalegrala:
I was reading in this area recently and also noticed the #1596 and thought to put up the PR.  I think client and server settings are separate, and so without this, we are using the default values on the client.  We could certainly do more testing to confirm, like clamping down the max bytes on the client and seeing if the server responds in a way that represents the negotiation.

joe-elliott:
If we want to do this change, we should repro the issue and then confirm this fixes it.

zalegrala:
I mixed on this.  Here we're only using the configuration that we already expose, so to test we'd be testing the deeper levels of gRPC.  I haven't heard back on #1596.

zalegrala:
Withdrawing the PR.

Issue #2754, opened 2023-08-02
Tempo leaves behind data files for blocks with no metadata


zalegrala:
**Describe the bug**

In some circumstances, a `data.parquet` file is the only object in a block path, which means this block shows up in the list, but the metadata is available.  As of https://github.com/grafana/tempo/pull/2678 tempo now deletes the tenant index when the tenant is found to have zero blocks, but because these paths still show up, the tenant is not completely deleted which causes index failures to occur repeatedly for these tenants and results in additional calls to the backend that aren't helpful.

The index deletion I believe only uncovered this issue, which was dormant due to the index being left in place prior to #2678.

**To Reproduce**

Running `r106` we see this in environments where a tenant index was deleted due to no blocks being found.  My hunch is that this has something to do with unclean shutdown, but have no data to back this up.

**Expected behavior**
Block meta is either reconstructed from the data, or the data is removed.

**Environment:**
 - Infrastructure: Kubernetes
 - Deployment tool: jsonnet

mdisibio:
Adding that this situation will cause repeated errors like the following until all files are cleaned up:  `msg="failed to write tenant index" tenant=<id> err="storage: object doesn't exist"`. This is because it sees no more blocks and tries to delete the tenant index, but it's already been deleted.  The delete call at https://github.com/grafana/tempo/blob/main/tempodb/backend/raw.go#L94 isn't handling the `does not exist` error gracefully.  In this situation it should not be considered a failure and propagate up.  I think this should be an easy fix by ignoring `backend.ErrDoesNotExist`.

joe-elliott:
Options to improve this situation include:

- Clean up partial blocks over a certain age.
- Deterministically generate UUIDs for output blocks on compaction like the block builder

Also note that recommended bucket policies will eventually clean this up.

PR #3223, opened 2023-12-11
Improve poller missing index error condition

zalegrala:
**What this PR does**:

Here we handle an error condition for a missing tenant index on non-index-builder pollers.  This avoids unnecessary work when a tenant has been deleted, but remains in the tenant list due to the behavior described in #2754.

For non-index-builders:
* a missing index will skip the tenant
* a corrupted index will poll the tenant if fallback is enabled
* a corrupted index will not poll the tenant if fallback is disabled

This change resulted in the compactor being required as part of some tests.  Previously, if the index was not available, all pollers would attempt to build and write the index.

Related #2754

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/integration/e2e/e2e_test.go
+++ b/integration/e2e/e2e_test.go
@@ -211,10 +211,12 @@ func TestMicroservicesWithKVStores(t *testing.T) {
 			tempoIngester2 := util.NewTempoIngester(2)
 			tempoIngester3 := util.NewTempoIngester(3)
 
+			tempoCompactor := util.NewTempoCompactor()
+
 			tempoDistributor := util.NewTempoDistributor()
 			tempoQueryFrontend := util.NewTempoQueryFrontend()
 			tempoQuerier := util.NewTempoQuerier()
-			require.NoError(t, s.StartAndWaitReady(tempoIngester1, tempoIngester2, tempoIngester3, tempoDistributor, tempoQueryFrontend, tempoQuerier))
+			require.NoError(t, s.StartAndWaitReady(tempoIngester1, tempoIngester2, tempoIngester3, tempoDistributor, tempoQueryFrontend, tempoQuerier, tempoCompactor))
 
 			// wait for active ingesters
 			time.Sleep(1 * time.Second)
--- a/tempodb/blocklist/poller.go
+++ b/tempodb/blocklist/poller.go
@@ -145,14 +145,19 @@ func (p *Poller) Do(previous *List) (PerTenant, PerTenantCompacted, error) {
 		return nil, nil, err
 	}
 
-	blocklist := PerTenant{}
-	compactedBlocklist := PerTenantCompacted{}
-
-	consecutiveErrors := 0
+	var (
+		blocklist          = PerTenant{}
+		compactedBlocklist = PerTenantCompacted{}
+		consecutiveErrors  = 0
+	)
 
 	for _, tenantID := range tenants {
 		newBlockList, newCompactedBlockList, err := p.pollTenantAndCreateIndex(ctx, tenantID, previous)
 		if err != nil {
+			if errors.Is(err, backend.ErrDoesNotExist) {
+				continue
+			}
+
 			level.Error(p.logger).Log("msg", "failed to poll or create index for tenant", "tenant", tenantID, "err", err)
 			consecutiveErrors++
 			if consecutiveErrors > p.cfg.TolerateConsecutiveErrors {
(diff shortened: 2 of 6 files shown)


tempodb/blocklist/poller.go
```diff
@@ -145,14 +145,19 @@ func (p *Poller) Do(previous *List) (PerTenant, PerTenantCompacted, error) {
...
+	var (
+		blocklist          = PerTenant{}
+		compactedBlocklist = PerTenantCompacted{}
+		consecutiveErrors  = 0
+	)
 
 	for _, tenantID := range tenants {
 		newBlockList, newCompactedBlockList, err := p.pollTenantAndCreateIndex(ctx, tenantID, previous)
 		if err != nil {
+			if errors.Is(err, backend.ErrDoesNotExist) {
```
joe-elliott, line 157: the goal here is to detect that this tenant had no valid blocks despite being in the tenant list. should we detect that in the child functions and pass a custom error up? instead of keying on DNE?

or maybe a nil err and an empty index which we then ignore?

tempodb/blocklist/poller.go
```diff
@@ -223,12 +235,13 @@ func (p *Poller) pollTenantAndCreateIndex(
 
 		// polling fallback is true, log the error and continue in this method to completely poll the backend
 		level.Error(p.logger).Log("msg", "failed to pull bucket index for tenant. falling back to polling", "tenant", tenantID, "err", err)
+	} else {
+		metricTenantIndexBuilder.WithLabelValues(tenantID).Set(1)
```
joe-elliott, line 239: any particular reason to move this? is this to distinguish between a "true" tenant index builder and a compactor that's just falling through?

tempodb/blocklist/poller.go
```diff
@@ -196,11 +201,18 @@ func (p *Poller) pollTenantAndCreateIndex(
...
 	builder := p.tenantIndexBuilder(tenantID)
 	span.SetTag("tenant_index_builder", builder)
+
 	if !builder {
 		metricTenantIndexBuilder.WithLabelValues(tenantID).Set(0)
 
 		i, err := p.reader.TenantIndex(derivedCtx, tenantID)
 		err = p.tenantIndexPollError(i, err)
+
+		if errors.Is(err, backend.ErrDoesNotExist) {
```
joe-elliott, line 211: so the negative here is if we manually delete a tenant index to force recreation then only the assigned tenant index builders will attempt to recreate. which means that certain queriers will be unaware of any blocks in the backend for a tenant until it gets recreated (multiple minutes).

instead we could
1) log info/debug "no tenant index found. falling through to polling" 
2) falling through to the blocklist poll below
3) if the full blocklist list call returns an empty list log a warning. "no blocks found for tenant that should exist"
4) raise the metric `tempod_blocklist_poll_errors_total` to build alerts off of/warn the operator to clean up partial blocks

wdyt?

tempodb/blocklist/poller.go
zalegrala: This is to avoid setting the builder to true when the poller doesn't own the writing.  I think this is what we intend.  Or are you thinking that even if not configured to be the writer, if we fall though and end up writing, that we should metric as such?

integration/e2e/e2e_test.go
```diff
@@ -210,10 +210,12 @@ func TestMicroservicesWithKVStores(t *testing.T) {
 			tempoIngester2 := util.NewTempoIngester(2)
 			tempoIngester3 := util.NewTempoIngester(3)
 
+			tempoCompactor := util.NewTempoCompactor()
+
 			tempoDistributor := util.NewTempoDistributor()
 			tempoQueryFrontend := util.NewTempoQueryFrontend()
 			tempoQuerier := util.NewTempoQuerier()
-			require.NoError(t, s.StartAndWaitReady(tempoIngester1, tempoIngester2, tempoIngester3, tempoDistributor, tempoQueryFrontend, tempoQuerier))
+			require.NoError(t, s.StartAndWaitReady(tempoIngester1, tempoIngester2, tempoIngester3, tempoDistributor, tempoQueryFrontend, tempoQuerier, tempoCompactor))
```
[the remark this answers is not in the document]
zalegrala: The compactor needs to start in the tests now, since without it, the index won't get written.

tempodb/blocklist/poller.go
zalegrala: I think we want to avoid polling in the case of a missing index.  The only signal that we have currently that we wish to delete the tenant, is to delete the index.  If you wanted to force a recreation of the index, you could corrupt the index, which would cause the poller to poll and recreate it.

zalegrala:
Closing for non-response.

PR #3241, opened 2023-12-17
Add specific configuration for Tempo Distributed Mode

observability-guy:
**What this PR does**:

I am proposing this update to the Tempo configuration for Azure Storage in the light of issues I encountered when trying to run Tempo in Distributed Mode.

I found that, when using the existing documentation the trace specification did not get successfully applied to the storage object - it was still pointing to 'local' as its backend.

I also found that the extraArgs and etraEnv configuration did not seem to cascade down to any of the relevant services - so they were unable to retrieve the Storage Account key. This meant that the services threw errors during initialization and were unable to start up.

It seems to me that Tempo Monolithic and Tempo Distributed have different architectures so they each need their own distinct configuration for Azure Storage.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott (approved):
Thanks for the improvement!

PR #3252, opened 2023-12-19
[Traceql Metrics] PR 2 - API

mdisibio:
**What this PR does**:

This is the result of a collaboration effort and requires https://github.com/grafana/tempo/pull/3227 for the underlying work of providing the data.  Here we build out the API for metrics in Tempo.

* Update API and proto definitions
* Build out the generator API
* Wire up the generator API to queriers
* Handle the query and backend sharding on the frontend

**Notes**
1. #3251
2. #3252
3. #3258 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/traceql/ast.go
```diff
@@ -305,7 +304,7 @@ func newSpansetFilter(e FieldExpression) *SpansetFilter {
 func (*SpansetFilter) __spansetExpression() {}
 
 func (f *SpansetFilter) evaluate(input []*Spanset) ([]*Spanset, error) {
-	f.outputBuffer = f.outputBuffer[:0]
+	var outputBuffer []*Spanset
```
mdisibio, line 307: This change is required to make the traceql eval callback safe for concurrent use (so we can execute the query against multiple blocks in parallel).  This was ok previously because: regular searches were single threaded or the query was compiled for each block and they got their own AST (and SpansetFilters). But now for metrics the AST keeps more state and it should be more efficient to reuse the same AST across all blocks.  Otherwise it would have to create and squash repeatedly the same timeseries from each block to get the final results.

modules/generator/processor/localblocks/processor.go
```diff
@@ -387,6 +391,78 @@ func (p *Processor) GetMetrics(ctx context.Context, req *tempopb.SpanMetricsRequ
...
+		blocks = append(blocks, b)
+	}
+
+	eval, err := traceql.NewEngine().CompileMetricsQueryRange(req, false)
+	if err != nil {
+		return nil, err
+	}
+
+	var (
+		wg     = boundedwaitgroup.New(10)
```
zalegrala, line 422: Do we want to push the value here into config?

modules/generator/processor/localblocks/processor.go
```diff
@@ -387,6 +391,78 @@ func (p *Processor) GetMetrics(ctx context.Context, req *tempopb.SpanMetricsRequ
...
+			defer span.Finish()
+
+			// TODO - caching
+			f := traceql.NewSpansetFetcherWrapper(func(ctx context.Context, req traceql.FetchSpansRequest) (traceql.FetchSpansResponse, error) {
+				return b.Fetch(ctx, req, common.DefaultSearchOptions())
+			})
+
+			err := eval.Do(ctx2, f)
+			if err != nil {
+				jobErr.Store(err)
```
zalegrala, line 452: Alternatively we could append to a slice to get all of the errors.  I think this just stores the last error.

mdisibio: I think returning just the last error is ok.  Now looking at it more closely, `QueryRange` ideally exits immediately after the first error and cancels all remaining work.

modules/querier/http.go
```diff
@@ -296,6 +297,78 @@ func (q *Querier) SpanMetricsSummaryHandler(w http.ResponseWriter, r *http.Reque
...
+		if err != nil {
+			span.SetTag("error", err.Error())
+		}
+	}
+
+	defer func() {
+		errHandler(ctx, span, err)
+		m := jsonpb.Marshaler{}
+
+		jsBytes, funcErr := m.MarshalToString(resp)
```
mdisibio, line 335: Testing out the error handling in local blocks processor and it leads to nil resp here triggers the error `Marshal called with nil`.

zalegrala (approved):
Looks good to me.

modules/querier/http.go
zalegrala: Updated the error handling here.  Good call out.

PR #3257, opened 2023-12-20
Implement a priority queue for tenant polling

zalegrala:
**What this PR does**:

Here is an implementation of using a priority queue to poll the tenants who have been polled least recently, in combination with a modest concurrent polling.   In a small environment with 30k blocks and 10k tenants, this results in a 70% decrease in overall polling time.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/blocklist/poller.go
+++ b/tempodb/blocklist/poller.go
@@ -290,6 +290,10 @@ func (p *Poller) tenantPollLoop(ctx context.Context, j int) {
 	}()
 
 	for {
+		if ctx.Err() != nil { // context cancelled
+			return
+		}
+
 		o := p.tenantQueues.Dequeue(j)
 		if o == nil {
 			return
--- a/tempodb/blocklist/poller.go
+++ b/tempodb/blocklist/poller.go
@@ -72,7 +72,7 @@ var (
 		Help:      "Age in seconds of the last pulled tenant index.",
 	}, []string{"tenant"})
 	metricTenantQueueLength = promauto.NewGauge(prometheus.GaugeOpts{
-		Namespace: "tempo",
+		Namespace: "tempodb",
 		Name:      "poller_tenant_queue_length",
 		Help:      "The total number of tenants pending in the queue.",
 	})
@@ -117,14 +117,6 @@ type Poller struct {
 	sharder JobSharder
 	logger  log.Logger
 
-	tenantQueues     *flushqueues.ExclusiveQueues
-	tenantQueuesDone sync.WaitGroup
-	tenantsPolled    map[string]time.Time
-
-	perTenantChan          chan *PerTenant
-	perTenantCompactedChan chan *PerTenantCompacted
-	perTenantErrChan       chan error
-
 	blocklist *List
 }
 
@@ -135,15 +127,9 @@ func NewPoller(cfg *PollerConfig, sharder JobSharder, reader backend.Reader, com
 		compactor: compactor,
 		writer:    writer,
 
-		cfg:           cfg,
-		sharder:       sharder,
-		logger:        logger,
-		tenantQueues:  flushqueues.New(int(cfg.TenantPollConcurrency), metricTenantQueueLength),
-		tenantsPolled: make(map[string]time.Time, 1000),
-
-		perTenantChan:          make(chan *PerTenant),
-		perTenantCompactedChan: make(chan *PerTenantCompacted),
-		perTenantErrChan:       make(chan error, 1000),
+		cfg:     cfg,
+		sharder: sharder,
+		logger:  logger,
 
 		blocklist: blocklist,
 	}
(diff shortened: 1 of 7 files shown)


tempodb/blocklist/poller.go
```diff
@@ -145,50 +173,181 @@ func (p *Poller) Do(previous *List) (PerTenant, PerTenantCompacted, error) {
...
+		/* 	retry, err = p.handleComplete(op) */
+		/* } else { */
+		/* 	retry, err = p.handleFlush(context.Background(), op.userID, op.blockID) */
+		/* } */
+
+		/* if err != nil { */
+		/* 	handleFailedOp(op, err) */
+		/* } */
+
+		if retry {
```
zalegrala, line 340: TODO: Retry is always false.

tempodb/blocklist/poller.go
```diff
@@ -109,23 +117,41 @@ type Poller struct {
...
-		cfg:     cfg,
-		sharder: sharder,
-		logger:  logger,
+		cfg:           cfg,
+		sharder:       sharder,
+		logger:        logger,
+		tenantQueues:  flushqueues.New(int(cfg.TenantPollConcurrency), metricTenantQueueLength),
+		tenantsPolled: make(map[string]time.Time, 1000),
+
+		perTenantChan:          make(chan *PerTenant, 1000),
```
zalegrala, line 145: The static `1000` here is the part of this PR I'm least comfortable with.  I'm not sure how to size this.

zalegrala: Turns out if this is 0 the code still seems to work.  Continuing to test.

tempodb/blocklist/poller.go
```diff
@@ -109,23 +116,41 @@ type Poller struct {
...
+		perTenantCompactedChan: make(chan *PerTenantCompacted),
+		perTenantErrChan:       make(chan error, 1000),
+
+		blocklist: blocklist,
 	}
 }
 
 // Do does the doing of getting a blocklist
-func (p *Poller) Do(previous *List) (PerTenant, PerTenantCompacted, error) {
+func (p *Poller) Do() (PerTenant, PerTenantCompacted, error) {
```
joe-elliott, line 153: Let's combine `Do()` and `Start()`. Right now `Start()` kicks of N goroutines which watch the tenant queue polling.

If we combine these 2 methods than our channels, go routines, mutex, queues, etc all have a very clean lifecycle. We can remove all of those fields from the struct and instead create and destroy them all in `Do()`.

joe-elliott (approved):

tempodb/tempodb.go
```diff
@@ -535,9 +542,15 @@ func (rw *readerWriter) pollingLoop(ctx context.Context) {
 }
 
 func (rw *readerWriter) pollBlocklist() {
-	blocklist, compactedBlocklist, err := rw.blocklistPoller.Do(rw.blocklist)
+	blocklist, compactedBlocklist, err := rw.blocklistPoller.Do()
 	if err != nil {
-		level.Error(rw.logger).Log("msg", "failed to poll blocklist", "err", err)
+		switch {
+		case errors.Is(err, context.Canceled):
+		case errors.Is(err, context.DeadlineExceeded):
```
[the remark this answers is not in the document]
zalegrala: I think just avoiding the noise in the log is good for the poller.  Do we care if the context was canceled or timed out?

tempodb/blocklist/poller.go
zalegrala: I've pushed a commit for this request.  Let's see what CI says.  What do you think of the change?

tempodb/blocklist/poller.go
```diff
@@ -145,50 +170,168 @@ func (p *Poller) Do(previous *List) (PerTenant, PerTenantCompacted, error) {
...
+		blocklist          = PerTenant{}
+		compactedBlocklist = PerTenantCompacted{}
+		consecutiveErrors  = 0
+		lastTenantPoll     time.Time
+		errs               []error
+	)
 
 	for _, tenantID := range tenants {
-		newBlockList, newCompactedBlockList, err := p.pollTenantAndCreateIndex(ctx, tenantID, previous)
+		lastTenantPoll = time.Now()
```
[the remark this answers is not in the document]
zalegrala: Hmm, I think that could make more sense.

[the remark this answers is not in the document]
zalegrala: I've pushed a commit for this.

tempodb/tempodb.go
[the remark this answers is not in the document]
joe-elliott: if we are debugging issues where tenant indexes are not being updated seeing these errors in the log would be valuable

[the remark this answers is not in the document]
zalegrala: I've pushed a commit for this.

tempodb/blocklist/poller.go
joe-elliott: I don't see a need for the channels. Let's do the straightforwad thing:

```
// for all tenants insert jobs

// for N workers start a goroutine

     // pull job from queue

     // pollBlocklist
     
     // add job to blocklist
```

the blocklist and the queue are concurrency safe so need to worry about that. if you get an error just store it in an atomic err or use a mutex to protect a multierror or whatever makes sense.

let's simplify!

zalegrala: The current approach works based on the tests as well as a deployment.  Is there an issue with the channels you are seeing?  I quite like the channelized approach here.

joe-elliott: It's overcomplicated. This can be done in a much more straightforward way.

zalegrala: I don't like the changes required to avoid the channels.  Additionally, I think we should start the workers separately as before, and so the workers remain with the poller through the `Do()` iteration.  The startup + alloc on each iteration doesn't sit well with me

Stylistically, the lock -> append -> unlock in a few places I find noisy and less clean that the original implementation.  The channelized version avoided the locking because collection of the data was done outside of a routine, which I appreciated.

Pragmatically, I'm looking at `Dequeue()` which blocks waiting for an op and I don't see a clean way to avoid attempting to read from the queue without knowing the size of the tenants which feels fragile and prone to race.

Happy to chat.

joe-elliott: I would like the change to look something like this: 
https://github.com/grafana/tempo/compare/main...joe-elliott:code

It reads cleanly top to bottom, the flow of data is simpler, the lifecycles of the objects are constrained, and it is more in line with the existing structure making it easier to review. It is easier to debug and maintain long term.

> The startup + alloc on each iteration doesn't sit well with me

This is one of the best features of keeping it one function. Limiting the lifecycle of these objects reduces potential for bugs.

> Pragmatically, I'm looking at Dequeue() which blocks waiting for an op

Just make a "DequeueWIthoutBlocking()" call. There's no race b/c you've already inserted all tenants. This priority queue has some features unneeded here, it might make sense to just make a simple local one.

zalegrala: Thanks for the review.  I'll close this out and reconsider the approach.  In the mean time, if you'd like to PR your changes for review, feel free.

PR #3258, opened 2023-12-21
[Traceql Metrics] PR 3 - Trace ID sharding

mdisibio:
**What this PR does**:
This PR implements the trace-ID sharding used by metrics in the block Fetch() method.  For context:  the query-frontend divides a metrics query into jobs along 2 dimensions:   horizontally across time (i.e. a job every 5 minutes) and vertically across trace IDs (i.e. 10 shards).  This PR is the vertical scaling. For example:  shard "10 of 16" processes only traces A00.... through AFF....  This is done with a predicate against the TraceID column.  

There are some other details to be aware:
* Some tricks to shard both 8-byte and 16-byte trace IDs.
* It uses the previously-experimental trace ID index json in each block.  This eliminates quite a few reads because we can go straight to the row groups that contain the matching trace IDs instead of testing every one.  I think we fixed the issues with index, it's been working well for me.

**Notes**
1. https://github.com/grafana/tempo/pull/3251
2. https://github.com/grafana/tempo/pull/3252
3. <this PR>

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/encoding/vparquet3/block_traceql.go
+++ b/tempodb/encoding/vparquet3/block_traceql.go
@@ -1,6 +1,7 @@
 package vparquet3
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


zalegrala (approved):
This looks good to me.  Nice work.

PR #3261, opened 2023-12-24
Docs: indent code block to allow proper list numbers

fabrizio-grafana:
A code block breaks the numbering of list items:
image

Indenting it, as done in the preceding list, should fix the issue:
image


jdbaldry (commented):
I haven't checked this locally but I would guess that this does fix the issue given the screenshots.

I would consider using [`prettier`](https://prettier.io/) for consistent Markdown formatting. It is already used in the grafana repository and also Writers' Toolkit.

The consistent formatting makes the Markdown easier to read and also exposes ambiguities in Markdown parsing which can affect the rendered HTML.

knylander-grafana (approved):
Thank you for fixing the code block indent.

knylander-grafana:
> BTW and JTLYK, I tried to rerun the CI steps a few times, but I still have one of them failing. Also, I am not allowed to merge the PR (and maybe that's why the CI step is failing?).

I think because this is just docs, we should be okay with merging as-is. I'll go ahead and merge it and backport to the 2.3 docs.

PR #3087, opened 2023-10-30
Add support for multi-tenant queries

electron0zero:
**What this PR does**:

Add multi-tenant query support in selected routes:
- TraceByID
- Search
- SearchTags
- SearchTagsV2
- SearchTagsValues
- SearchTagsValuesV2

> Note: SpanMetricsSummary and Streaming Search endpoints does not support multi-tenant queries.

**Which issue(s) this PR fixes**:
Fixes #1858

**Checklist**
- [x] Tests updated
- [x] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #3262, opened 2023-12-26
Add support for multi-tenant queries in streaming search

### LINKED ISSUE #1858: Ability to search all tenants in multi-tenant setups

electron0zero:
**What this PR does**:

Add multi-tenant query support in streaming search endpoints.

follow up on https://github.com/grafana/tempo/pull/3087

Fixes #1858

```[tasklist]
### Tasks
- [x] multi-tenant streaming Search works
- [x] test for multiProgress
- [x] fix metrics in streamed results
- [x] test that streaming actually streams
```

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/example/docker-compose/multi-tenant/docker-compose.yaml
+++ b/example/docker-compose/multi-tenant/docker-compose.yaml
@@ -1,0 +1,48 @@
+version: "3"
+services:
+  tempo:
+    image: grafana/tempo:latest
+    environment:
+      - GRPC_TRACE=api
+    command: [ "-config.file=/etc/tempo.yaml" ]
+    volumes:
+      - ./tempo.yaml:/etc/tempo.yaml
+      - ./tempo-data:/tmp/tempo
+    ports:
+      - "14268:14268"  # jaeger ingest
+      - "3200:3200"   # tempo
+      - "9095:9095" # tempo grpc
+      - "4317:4317"  # otlp grpc
+      - "4318:4318"  # otlp http
+      - "9411:9411"   # zipkin
+
+  k6-tracing:
+    image: ghcr.io/grafana/xk6-client-tracing:latest
+    environment:
+      - ENDPOINT=tempo:4317
+      - TEMPO_X_SCOPE_ORGID=test
+    restart: always
+    depends_on:
+      - tempo
+
+  k6-tracing-2:
+    image: ghcr.io/grafana/xk6-client-tracing:latest
+    environment:
+      - ENDPOINT=tempo:4317
+      - TEMPO_X_SCOPE_ORGID=test2
+    restart: always
+    depends_on:
+      - tempo
+
+  grafana:
+    image: grafana/grafana:main
+    volumes:
+      - ./grafana-datasources.yaml:/etc/grafana/provisioning/datasources/datasources.yaml
+    environment:
+      - GF_DEFAULT_APP_MODE=development
+      - GF_AUTH_ANONYMOUS_ENABLED=true
+      - GF_AUTH_ANONYMOUS_ORG_ROLE=Admin
+      - GF_AUTH_DISABLE_LOGIN_FORM=true
+      - GF_FEATURE_TOGGLES_ENABLE=traceqlEditor traceQLStreaming metricsSummary
+    ports:
+      - "3000:3000"
--- a/integration/e2e/multi_tenant_test.go
+++ b/integration/e2e/multi_tenant_test.go
@@ -220,24 +220,9 @@ func TestMultiTenantSearch(t *testing.T) {
 				Query: "{}", Start: uint32(now.Add(-20 * time.Minute).Unix()), End: uint32(now.Unix()),
 			})
 			require.NoError(t, err)
-			// actual error comes in resp, need to call Recv to get it.
+			// actual error comes in response, need to call Recv to get it.
 			_, grpcErr := grpcResp.Recv()
-			// fmt.Printf("==== grpcResp.Recv() grpcErr: %s\n", grpcErr)
-			// fmt.Printf("==== grpcResp.Recv() grpcSearchResp: %s\n", grpcSearchResp)
-
-			// drain the stream until everything is returned
-			// for {
-			// 	_, grpcErr = grpcResp.Recv()
-			// 	if errors.Is(err, io.EOF) {
-			// 		break
-			// 	}
-			// 	require.NoError(t, err)
-			// }
-
-			// find the trace with streaming. using the http server b/c that's what Grafana will do
-			// FIXME: info is for the last tenant so it's gonna fail, assert streaming in it's own test?? and only check for no error here
-			// rpc error: code = Unknown desc = no org id
-			// util.SearchStreamAndAssertTrace(t, grpcClient, info, now.Add(-20*time.Minute).Unix(), now.Unix())
+			assertRequestCountMetric(t, tempo, "/tempopb.StreamingQuerier/Search", 1)
 
 			if tc.tenantSize > 1 {
 				// error for multi-tenant request for unsupported endpoints
--- a/integration/e2e/multi_tenant_test.go
+++ b/integration/e2e/multi_tenant_test.go
@@ -222,11 +222,28 @@ func TestMultiTenantSearch(t *testing.T) {
 			require.NoError(t, err)
 			// actual error comes in resp, need to call Recv to get it.
 			_, grpcErr := grpcResp.Recv()
+			// fmt.Printf("==== grpcResp.Recv() grpcErr: %s\n", grpcErr)
+			// fmt.Printf("==== grpcResp.Recv() grpcSearchResp: %s\n", grpcSearchResp)
+
+			// drain the stream until everything is returned
+			// for {
+			// 	_, grpcErr = grpcResp.Recv()
+			// 	if errors.Is(err, io.EOF) {
+			// 		break
+			// 	}
+			// 	require.NoError(t, err)
+			// }
+
+			// find the trace with streaming. using the http server b/c that's what Grafana will do
+			// FIXME: info is for the last tenant so it's gonna fail, assert streaming in it's own test?? and only check for no error here
+			// rpc error: code = Unknown desc = no org id
+			// util.SearchStreamAndAssertTrace(t, grpcClient, info, now.Add(-20*time.Minute).Unix(), now.Unix())
 
 			if tc.tenantSize > 1 {
-				// we expect error in case of multi-tenant request for unsupported endpoints
+				// error for multi-tenant request for unsupported endpoints
 				require.Error(t, msErr)
-				require.Error(t, grpcErr)
+				// no error for multi-tenant streaming search
+				require.NoError(t, grpcErr)
 			} else {
 				require.NoError(t, msErr)
 				require.NoError(t, grpcErr)
--- a/integration/e2e/multi_tenant_test.go
+++ b/integration/e2e/multi_tenant_test.go
@@ -205,10 +205,6 @@ func TestMultiTenantSearch(t *testing.T) {
 				assertRequestCountMetric(t, tempo, rt.route, rt.reqCount)
 			}
 
-			// test all the unsupported endpoints
-			now := time.Now()
-			_, msErr := apiClient.MetricsSummary("{}", "name", 0, 0)
-
 			// test streaming search over grpc
 			grpcCtx := user.InjectOrgID(context.Background(), tc.tenant)
 			grpcCtx, err = user.InjectIntoGRPCRequest(grpcCtx)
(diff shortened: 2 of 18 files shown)


example/docker-compose/multi-tenant/docker-compose.yaml
```diff
@@ -0,0 +1,48 @@
+version: "3"
```
joe-elliott, line 1: i'm always wary of adding new examples b/c they require manual testing/maintenance occassionally.

wdyt about coming this with the otel-collector multitenant example?

integration/e2e/multi_tenant_test.go
```diff
@@ -215,6 +215,19 @@ func TestMultiTenantSearch(t *testing.T) {
 				Query: "{}", Start: uint32(now.Add(-20 * time.Minute).Unix()), End: uint32(now.Unix()),
 			}, func(sr *tempopb.SearchResponse) {})
 
+			// fmt.Printf("==== WS Search wsErr: %s\n", wsErr)
```
joe-elliott, line 218: remove commented code?

integration/e2e/multi_tenant_test.go
```diff
@@ -228,12 +241,29 @@ func TestMultiTenantSearch(t *testing.T) {
 			require.NoError(t, err)
 			// actual error comes in resp, need to call Recv to get it.
 			_, grpcErr := grpcResp.Recv()
+			// fmt.Printf("==== grpcResp.Recv() grpcErr: %s\n", grpcErr)
```
joe-elliott, line 244: same as above

integration/e2e/multi_tenant_test.go
```diff
@@ -228,12 +241,29 @@ func TestMultiTenantSearch(t *testing.T) {
...
+			// rpc error: code = Unknown desc = no org id
+			// util.SearchStreamAndAssertTrace(t, grpcClient, info, now.Add(-20*time.Minute).Unix(), now.Unix())
 
 			if tc.tenantSize > 1 {
-				// we expect error in case of multi-tenant request for unsupported endpoints
+				// error for multi-tenant request for unsupported endpoints
 				require.Error(t, msErr)
-				require.Error(t, wsErr)
-				require.Error(t, grpcErr)
+				// no error for multi-tenant streaming search
```
joe-elliott, line 264: does this test assert the expected content? or just that there's no error?

modules/frontend/search_streaming.go
```diff
@@ -253,14 +316,19 @@ func (s *streamingSearcher) handle(r *http.Request, forwardResults func(*tempopb
...
-	progress := atomic.NewPointer[*diffSearchProgress](nil)
-	fn := func(ctx context.Context, limit, totalJobs, totalBlocks int, totalBlockBytes uint64) shardedSearchProgress {
-		p := newDiffSearchProgress(ctx, limit, totalJobs, totalBlocks, totalBlockBytes)
-		progress.Store(&p)
-		return p
+	mProgress := newMultiProgress()
+	// create diffSearchProgress for each tenant, and keep a reference to
+	// stream results back to the client.
+	progressFactoryFn := func(ctx context.Context, limit, totalJobs, totalBlocks int, totalBlockBytes uint64) shardedSearchProgress {
+		ap := atomic.NewPointer[*diffSearchProgress](nil)
```
joe-elliott, line 323: the atomic pointer was used to protect race conditions on the pointer access itself between the factory function and the below code. since these pointers are now protected by the multi progress object we can drop the atomic access.

modules/frontend/search_streaming.go
```diff
@@ -287,22 +356,33 @@ func (s *streamingSearcher) handle(r *http.Request, forwardResults func(*tempopb
...
+				result := (*p).result()
+				if result.err != nil || result.statusCode != http.StatusOK { // ignore errors here, we'll get them in the resultChan
+					continue
+				}
 
-			err := forwardResults(result.response)
-			if err != nil {
-				level.Error(s.logger).Log("msg", "search streaming: send failed", "err", err)
-				return fmt.Errorf("search streaming send failed: %w", err)
+				// send response to client
```
joe-elliott, line 371: clean up comments

modules/frontend/search_streaming.go
```diff
@@ -287,22 +356,33 @@ func (s *streamingSearcher) handle(r *http.Request, forwardResults func(*tempopb
...
+				// FIXME: this will ping pong between each tenant's metrics and client will not get accurate metrics
+				// -- metrics when results are streaming back in??
+				// {"metrics":{"inspectedBytes":452084,"totalBlocks":1,"completedJobs":2,"totalJobs":2,"totalBlockBytes":535418}}
+				// {"metrics":{"inspectedBytes":452945,"totalBlocks":1,"completedJobs":2,"totalJobs":2,"totalBlockBytes":506656}}
+				// {"metrics":{"inspectedBytes":452084,"totalBlocks":1,"completedJobs":2,"totalJobs":2,"totalBlockBytes":535418}}
+				// {"metrics":{"inspectedBytes":452945,"totalBlocks":1,"completedJobs":2,"totalJobs":2,"totalBlockBytes":506656}}
+				// FIXME: need to merge metrics from all tenants and modify response with merged metrics.
+				err := forwardResults(result.response)
```
joe-elliott, line 379: this seems like it will forward each tenant's response individually. we should combine results before returning.

i think you can just call `mProgress.finalResults()` and return that

modules/frontend/search_streaming.go
```diff
@@ -95,17 +96,74 @@ func (p *diffSearchProgress) finalResult() *shardedSearchResults {
...
+
+	if len(mp.progress) == 1 {
+		// single tenant query, only one progress object.
+		p := *mp.progress[0].Load()
+		return p.finalResult()
+	}
+
+	// multi-tenant query, merge result from all progress objects.
+	// note: newSearchProgress is used to combine results from multiple progress objects, don't use it for anything else.
+	comb := newSearchProgress(ctx, 0, 0, 0, 0)
```
joe-elliott, line 140: should we be using `combiner.NewSearch` here? i'm honestly not sure. wdyt?

modules/frontend/search_streaming.go
```diff
@@ -95,17 +96,74 @@ func (p *diffSearchProgress) finalResult() *shardedSearchResults {
...
+		logger:      logger,
+		downstream:  downstream,
+		reader:      reader,
+		postSLOHook: searchSLOPostHook(cfg.Search.SLO),
+		o:           o,
+		searchCache: searchCache,
+		cfg:         &cfg,
+		// pass NoOp combiner because we combine results ourselves, and don't use
+		// combined results from middleware
+		preMiddleware: newMultiTenantMiddleware(cfg, combiner.NewNoOp, logger),
```
joe-elliott, line 166: multitenant middleware waits for all tenants to have responses which it then combines and returns. 

how does this work with streaming?

integration/e2e/multi_tenant_test.go
electron0zero: just no error, and not content.

I am adding more unit tests for multi-tenant streaming in `search_streaming_test.go`

modules/frontend/search_streaming.go
electron0zero: finalResults will not return finalResults, because it calls finalResult on diff.

we want to call result in each diff and return that, I am adding a method on multiProcess called result, will call that here. 

will merge metrics as well, and that should fix the ping ping issue with metrics.

modules/frontend/search_streaming.go
electron0zero: we keep reference to progress objects and stream results via that, and we don't care about the results that it combines. that's why I am using a NoOp combiner that just throws away the results.

blocking in waits for all tenants to finish, and when it returns, we signal our select case to send the final result.

we again use progress objects to build final result.

example/docker-compose/multi-tenant/docker-compose.yaml
electron0zero: agree, it adds the maintaince overhead but I would prefer to keep it out because otel-collector multitenant is an example setup of using otel collector to send data to a multi-tenant tempo setup, and this is an example of multi-tenant search.

they are same in the spirit I am worrited that if we merge, It might confused folks who are using these examples as reference config for standing up tempo...

so I am leaning on the side of keeping otel-collector multitenant as simple as possible and keep this out of otel-collector multitenant.

modules/frontend/search_streaming.go
electron0zero: thanks, updated. that makes it simple as well :)

modules/frontend/search_streaming.go
electron0zero: this is cleaned up and fixed.

knylander-grafana:
@electron0zero What would you think about adding a query example to the TraceQL page to demonstrate this new capability? Are there any configuration options that need to be updated?

integration/e2e/multi_tenant_test.go
joe-elliott: any particular reason not to? we use this method in other tests to do it:

https://github.com/grafana/tempo/blob/main/integration/util.go#L368

modules/frontend/search_streaming.go
joe-elliott: ah, i see now. i am a bit concerned about orphaning goroutines in the multi tenant middleware, but it makes sense.

it's about as gorpy as the rest of this code. :)

modules/frontend/search_progress.go
```diff
@@ -185,6 +193,13 @@ func (r *searchProgress) result() *shardedSearchResults {
 	return res
 }
 
+func (r *searchProgress) metrics() *tempopb.SearchMetrics {
+	r.mtx.Lock()
+	defer r.mtx.Unlock()
+
+	return r.resultsMetrics
```
joe-elliott, line 200: this will create race conditions on the metric fields b/c you are returning a pointer to a memory location that other goroutines have access to and will attempt to update. 

you should likely clone this, but that can be costly. if you call it occassionally it should be ok.

modules/frontend/search_streaming.go
```diff
@@ -87,23 +87,104 @@ func (p *diffSearchProgress) result() *shardedSearchResults {
...
+}
+
+func (mp *multiProgress) Add(p *diffSearchProgress) {
+	mp.mu.Lock()
+	defer mp.mu.Unlock()
+
+	mp.progress = append(mp.progress, p)
+
+	// combine metrics
+	m := *p.metrics()
```
joe-elliott, line 120: why are you dereferencing the pointer?

modules/frontend/search_streaming.go
```diff
@@ -87,23 +87,104 @@ func (p *diffSearchProgress) result() *shardedSearchResults {
...
+	defer mp.mu.Unlock()
+
+	if len(mp.progress) == 1 {
+		// single tenant query, only one progress object.
+		return mp.progress[0].result()
+	}
+
+	comb := newSearchProgress(ctx, 0, mp.combinedMetrics.TotalJobs, mp.combinedMetrics.TotalBlocks, mp.combinedMetrics.TotalBlockBytes)
+	for _, progress := range mp.progress {
+		r := *progress.result() // only get new results from each progress object
```
joe-elliott, line 139: i don't think there's a reason to dereference

modules/frontend/search_streaming.go
```diff
@@ -87,23 +87,104 @@ func (p *diffSearchProgress) result() *shardedSearchResults {
...
+
+	if len(mp.progress) == 1 {
+		// single tenant query, only one progress object.
+		return mp.progress[0].finalResult()
+	}
+
+	// init the progress object with combinedMetrics to report accurate metrics
+	comb := newSearchProgress(ctx, 0, mp.combinedMetrics.TotalJobs, mp.combinedMetrics.TotalBlocks, mp.combinedMetrics.TotalBlockBytes)
+	for _, progress := range mp.progress {
+		r := *progress.finalResult()
```
joe-elliott, line 164: same thought as above

electron0zero:
> @electron0zero What would you think about adding a query example to the TraceQL page to demonstrate this new capability? 

@knylander-grafana It will be transperent to user querying data and should work for all queries, and doesn't add anything new in TraceQL.

> Are there any configuration options that need to be updated?

this is a follow up on https://github.com/grafana/tempo/pull/3087, and we documented config option in that PR, no updates now.

integration/e2e/multi_tenant_test.go
electron0zero: updated the test to assert the content.

modules/frontend/search_progress.go
electron0zero: updated this to return a copy resultsMetrics.

this is only called once each tenant (when we create a progress project), so should be fine.

modules/frontend/search_streaming.go
electron0zero: left out in refactoring, cleaned up...

modules/frontend/search_streaming.go
electron0zero: yup, cleaned up.

modules/frontend/search_streaming.go
joe-elliott: ah, i see now, the multi tenant middleware is used to split the requests but they're recombined in the streaming multi progress.

joe-elliott (approved):
let's go!

PR #3265, opened 2024-01-02
Remove multi-tenancy config in docker-compose examples

### LINKED ISSUE #3260: docker-compose examples: Non-multitenant otel-collector example and the local example (maybe others?) currently broken
I'm going to use Tempo to help me profile an application and improve its performance, so I want to be running it in as minimalistic of an environment as possible. I took a look at the docker-compose examples and for both the `local` one and the `otel-collector` one, when I went to go view traces in the Explore -> Tempo section, I got a lot of errors like `rpc error: code = Unknown desc = no org id` and the logs didn't really tell me much more. I spent a long time trying to debug and learn until it started to dawn on me to try the otel-collector-multitenant example, which worked perfectly fine.

I then realized that `example/docker-compose/shared/tempo.yaml` had `multitenancy_enabled: true` at the top! So Tempo was expecting the otel-collector to act like a multitenant reporter, but otel-collector wasn't configured to. It looks like this was a recent change, made in commit 0fc6bf1283, so maybe not all the examples' working-ness got double-checked for that change.

Luckily, once I finally understood what was going on, I worked around it pretty easily, without modifying the shared tempo.yaml config, by going to the docker-compose.yaml for the otel-collector example and changed the tempo Docker command from `command: [ "-config.file=/etc/tempo.yaml" ]` to `command: [ "-config.file=/etc/tempo.yaml", "-multitenancy.enabled=false" ]`. Then everything was fine. I guess putting that field of the tempo.yaml config back to `multitenancy_enbabled: false` would work too. But if you don't want to, note that the flag to enable multitenancy in the `otel-collector-multitenant` example is also now redundant.

I could open a PR to fix it but I don't know what the preferred the stylistic decision would be on where to use flags vs where to change the config file, so I kind of want to leave it up to you all :)

mapno:
**What this PR does**:

Removes `multitenancy_enabled: true` in non-multi-tenant` docker-compose examples

Fixes #3260 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/example/docker-compose/shared/tempo.yaml
+++ b/example/docker-compose/shared/tempo.yaml
@@ -1,4 +1,3 @@
-multitenancy_enabled: true
 stream_over_http_enabled: true
 server:
   http_listen_port: 3200
--- a/example/docker-compose/otel-collector-multitenant/docker-compose.yaml
+++ b/example/docker-compose/otel-collector-multitenant/docker-compose.yaml
@@ -50,6 +50,6 @@ services:
       - GF_AUTH_ANONYMOUS_ENABLED=true
       - GF_AUTH_ANONYMOUS_ORG_ROLE=Admin
       - GF_AUTH_DISABLE_LOGIN_FORM=true
-      - GF_FEATURE_TOGGLES_ENABLE=traceqlEditor
+      - GF_FEATURE_TOGGLES_ENABLE=traceqlEditor traceQLStreaming metricsSummary
     ports:
       - "3000:3000"


example/docker-compose/shared/tempo.yaml
```diff
@@ -1,4 +1,3 @@
-multitenancy_enabled: true
```
joe-elliott, line 1: did you check to see if all examples worked with this change ? particularly concerned about the otelcol multitenant one.

mapno: I checked a couple, including that one. It configures multi-tenancy via a flag https://github.com/grafana/tempo/blob/main/example/docker-compose/otel-collector-multitenant/docker-compose.yaml#L23

joe-elliott: +1

joe-elliott (approved):

PR #3267, opened 2024-01-04
Ensure 429 internal status is passed back to the caller

### LINKED ISSUE #2843: Tempo eating 429s on trace by ID path
If the query frontend receives a non 2XX it converts it to a 500 on the trace by id path here:

https://github.com/grafana/tempo/blob/2c05a8782a9b253f3e90d675b32595872c6eb12c/modules/frontend/tracebyidsharding.go#L162

This is currently eating 429s and returning 500 to the client. We need to propagate this status code back to the client correctly.

zalegrala:
**What this PR does**:

Here we ensure that 429 on the search and tracebyid paths are returned to the caller instead of a 500.

Fixes #2843

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/searchsharding.go
+++ b/modules/frontend/searchsharding.go
@@ -272,10 +272,20 @@ func (s searchSharder) RoundTrip(r *http.Request) (*http.Response, error) {
 		return nil, overallResponse.err
 	}
 
-	if overallResponse.statusCode != http.StatusOK {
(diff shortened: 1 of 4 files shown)


mapno (approved):
LGTM

electron0zero (approved):

PR #3277, opened 2024-01-10
Fix autocomplete panic when condition has wrong type

mapno:
**What this PR does**:

It prevents a panic when a filtering query is passed to `/search/tag/<tag>/values` with a attribute condition (`.a=b`) where `b` is a special type (`server/client`, `ok/error/unset`) but `.a` is of a different type (`string`, `int`). For example, `span.http.status_code=server`. This would create iterators with nil predicates and panics when it called `iter.Next()`.

Now, such conditions are ignored and iterators are not built for them.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@
 * [BUGFIX] Prevent building parquet iterators that would loop forever. [#3159](https://github.com/grafana/tempo/pull/3159) (@mapno)
 * [BUGFIX] Sanitize name in mapped dimensions in span-metrics processor [#3171](https://github.com/grafana/tempo/pull/3171) (@mapno)
 * [BUGFIX] Fixed an issue where cached footers were requested then ignored. [#3196](https://github.com/grafana/tempo/pull/3196) (@joe-elliott)
-* [BUGFIX] Fix panic in autocomplete when query condition had wrong type [#](https://github.com/grafana/tempo/pull/) (@mapno)
+* [BUGFIX] Fix panic in autocomplete when query condition had wrong type [#3277](https://github.com/grafana/tempo/pull/3277) (@mapno)
 
 ## v2.3.1 / 2023-11-28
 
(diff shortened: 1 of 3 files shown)


tempodb/encoding/vparquet3/block_autocomplete.go
```diff
@@ -358,6 +358,8 @@ func createDistinctAttributeIterator(
 			}
 			keyIter = makeIter(keyPath, parquetquery.NewStringInPredicate([]string{cond.Attribute.Name}), selectAs("key", cond.Attribute))
 			valIter = makeIter(boolPath, pred, selectAs("bool", cond.Attribute))
+		default:
+			continue
```
mapno, line 362: Another option is returning an error, but opted for ignoring the condition.

The reasoning is that autocomplete should be a smooth experience to the user. If the filtering query has an error, they'll get that feedback via normal search—ie. when they execute the query.

ie-pham (approved):

glamcoder (approved):

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@
 * [BUGFIX] Prevent building parquet iterators that would loop forever. [#3159](https://github.com/grafana/tempo/pull/3159) (@mapno)
 * [BUGFIX] Sanitize name in mapped dimensions in span-metrics processor [#3171](https://github.com/grafana/tempo/pull/3171) (@mapno)
 * [BUGFIX] Fixed an issue where cached footers were requested then ignored. [#3196](https://github.com/grafana/tempo/pull/3196) (@joe-elliott)
+* [BUGFIX] Fix panic in autocomplete when query condition had wrong type [#](https://github.com/grafana/tempo/pull/) (@mapno)
```
joe-elliott, line 39:
```suggestion
* [BUGFIX] Fix panic in autocomplete when query condition had wrong type [#3277](https://github.com/grafana/tempo/pull/3277) (@mapno)
```

tempodb/encoding/vparquet3/block_autocomplete.go
joe-elliott: are we missing case statements for things like `kind = server` and `status = ok`?

it seems like we should support auto complete on things like:
```
{ name = "foo" && kind = server && status = error && span.foo = ...
```

joe-elliott: thx for the clarification.

agree with your initial thought we should just ignore the bad condition over returning an error.

joe-elliott (approved):

mapno:
The autocomplete changes weren't released in v2.3. Nothing to backport! 😅

PR #3279, opened 2024-01-10
docs: Update docker-example.md

jonasem:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/sources/tempo/getting-started/docker-example.md
+++ b/docs/sources/tempo/getting-started/docker-example.md
@@ -67,6 +67,6 @@ To follow this guide, you need:
 
 1. To stop the services:
     ```
-    docker-compose down -v
+    docker compose down -v
     ```
     
--- a/docs/sources/tempo/getting-started/docker-example.md
+++ b/docs/sources/tempo/getting-started/docker-example.md
@@ -44,7 +44,7 @@ To follow this guide, you need:
 
 1. Start the services defined in the docker-compose file:
    ```bash
-   docker compose up -d
+   docker-compose up -d
    ```
 
 1. Verify that the services are running using `docker compose ps`. You should see something like:
--- a/docs/sources/tempo/getting-started/docker-example.md
+++ b/docs/sources/tempo/getting-started/docker-example.md
@@ -44,7 +44,7 @@ To follow this guide, you need:
 
 1. Start the services defined in the docker-compose file:
    ```bash
-   docker-compose up -d
+   docker compose up -d
    ```
 
 1. Verify that the services are running using `docker compose ps`. You should see something like:


docs/sources/tempo/getting-started/docker-example.md
```diff
@@ -69,4 +69,4 @@ To follow this guide, you need:
     ```
     docker-compose down -v
```
knylander-grafana, line 70:
```suggestion
    docker compose down -v
```

docs/sources/tempo/getting-started/docker-example.md
```diff
@@ -44,7 +44,7 @@ To follow this guide, you need:
 
 1. Start the services defined in the docker-compose file:
    ```bash
-   docker compose up -d
+   docker-compose up -d
```
knylander-grafana, line 47: This fix had me digging into the docker compose commands. Both `docker compose` and `docker-compose` versions work. However, docker-compose is a v1 implementation that is being deprecated in favor of docker compose (see [this thread](https://stackoverflow.com/questions/66514436/difference-between-docker-compose-and-docker-compose)). 

I did not know this. I'll update the docs. Thank you!

```suggestion
   docker compose up -d
```

jonasem: Okay, wow, for me `docker compose` doesn't work at all, i installed docker and docker-compose through brew in mac. 

I get an error saying `docker: 'compose' is not a docker command.`, docker version 24.0.7.

Seems like the space version is only included in the docker for desktop system for now, and that linux and mac users installing through cli have to do extra steps in order for it to work.

I learned someething new today as well, thank you :)

jonasem:
Applied your suggestions now, so the net effect is to use the space version.

knylander-grafana:
> Applied your suggestions now, so the net effect is to use the space version.

Yes. This is a very good catch -- we can review the other docs and make sure we're consistent with our commands, too. Thank you for opening this PR. I would have no idea otherwise. 

We're good to merge your PR. Would you be willing to sing the CLA?

jonasem:
> > Applied your suggestions now, so the net effect is to use the space version.
> 
> Yes. This is a very good catch -- we can review the other docs and make sure we're consistent with our commands, too. Thank you for opening this PR. I would have no idea otherwise.
> 
> We're good to merge your PR. Would you be willing to sing the CLA?

Yes, I tried to sign on friday, but CLA was having issues, seems to be working today :)

zalegrala (approved):

zalegrala:
Thanks for the contribution.

PR #3280, opened 2024-01-10
Implement prometheus datasource health check

zalegrala:
**What this PR does**:
Here we implement the two endpoints necessary to pass the "test" check when Tempo is used as a Prometheus data source.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


zalegrala:
We don't like that we implement a false `/query` endpoint, since we have the intention of supporting an instant query in the future at this endpoint.

PR #3292, opened 2024-01-14
Profiling: import godeltaprof/http/pprof

korniltsev:
**What this PR does**:

This adds /debug/pprof/delta_{heap,mutex,block} endpoints to DefaultServeMux

**Why do we need this feature?**

godeltaprof is much more efficient for the scraper and slightly more efficient for the target process (tempo)

Using godeltaprof saves much inuse memory and cpu on the scraper.

https://github.com/grafana/grafana/pull/80509
https://github.com/grafana/pyroscope-go/blob/main/godeltaprof/README.md
https://github.com/grafana/dskit/commit/bdc225d8adc027b2e9884d9d84b64ffb71cd94d7

Fixes https://github.com/grafana/pyroscope-squad/issues/88 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


mapno:
If this change has been merged into dskit's server, should we update that dependency instead?

korniltsev:
Yes, dskit is merged and updating dskit should do the trick.

I did not try to update dskit as I thought it would be too invasive for this change. I also heard my teammate @ cyriltovena tried to update it for tracing-profiling integration and it was not trivial, but maybe I am missing something.

I can try updating dskit.

Or maybe someone from tempo team could update.

Let me know what is the best.

korniltsev:
dskit was updated

PR #3300, opened 2024-01-16
Fix GRPC/HTTP/TLS

### LINKED ISSUE #3278: Querying from grafana doesn't work with TLS
**Describe the bug**
Setting up tls on both http and grpc causes grafana to receive "Query error" when searching for spans.

I have followed the guide at https://grafana.com/docs/tempo/latest/configuration/tls/#client-configuration
It says to add querier.query-frontend.grpc_client_config however, that doesn't seem to exist but I found that there is a querier.frontend_worker.grpc_client_config which doesn't exist in the docs. Are they the same?

Everything else seems to work fine and I get messages these, so the ingestion seems to work fine.
`level=info ts=2024-01-10T12:33:28.488848715Z caller=distributor.go:616 msg=received spanid=af253cd234c97243 traceid=b0a34725613194d4a3868db54c33492b`

I do not get any errors at all in the logs in tempo but in grafana I see:
`logger=data-proxy-log userId=1 orgId=1 uname=admin path=/api/datasources/proxy/uid/f64be1ec-e7c5-4824-addc-51f922c2477a/api/search remote_addr=::ffff:172.16.1.0 referer="https://grafana.app.local/explore?left=%7B%22datasource%22:%22f64be1ec-e7c5-4824-addc-51f922c2477a%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22tempo%22,%22uid%22:%22f64be1ec-e7c5-4824-addc-51f922c2477a%22%7D,%22queryType%22:%22nativeSearch%22,%22limit%22:20%7D%5D,%22range%22:%7B%22from%22:%22now-12h%22,%22to%22:%22now%22%7D%7D&orgId=1" traceID=58b728656a8c1bc4a2d1d79b3928bb5d t=2024-01-10T12:53:30.495083622Z level=error msg="Proxy request failed" err="read tcp 172.16.1.72:56580->172.17.228.229:3200: read: connection reset by peer"`

**To Reproduce**
Run grafana/tempo:2.3.1 container with the following config:

```
    multitenancy_enabled: false
    stream_over_http_enabled: true
    server:
      http_listen_port: 3200
      grpc_listen_port: 9095

      log_level: info

      grpc_tls_config:
        cert_file: /certs/tls.crt
        key_file: /certs/tls.key
      http_tls_config:
        cert_file: /certs/tls.crt
        key_file: /certs/tls.key
    query_frontend:
      search:
        duration_slo: 5s
        throughput_bytes_slo: 1.073741824e+09
      trace_by_id:
        duration_slo: 5s

    distributor:
      receivers:
        otlp:
          protocols:
            http:
              tls:
                cert_file: /certs/tls.crt
                key_file: /certs/tls.key
            grpc:
              tls:
                cert_file: /certs/tls.crt
                key_file: /certs/tls.key
      log_received_spans:
        enabled: true
    
    memberlist:
      tls_enabled: true
      tls_cert_path: /certs/tls.crt
      tls_key_path: /certs/tls.key

    ingester:
      max_block_duration: 5m               # cut the headblock when this much time passes. this is being set for demo purposes and should probably be left alone normally
    ingester_client:
      grpc_client_config:
        tls_enabled: true
        tls_cert_path: /certs/tls.crt
        tls_key_path: /certs/tls.key
        tls_insecure_skip_verify: true

    compactor:
      compaction:
        block_retention: 720h                # overall Tempo trace retention

    metrics_generator:
      storage:
        path: /tmp/tempo/generator/wal
        remote_write:
          - url: https://prometheus-operated:9090/api/v1/write
            send_exemplars: true
            tls_config:
              insecure_skip_verify: true
    metrics_generator_client:
      grpc_client_config:
        tls_enabled: true
        tls_cert_path: /certs/tls.crt
        tls_key_path: /certs/tls.key
        tls_insecure_skip_verify: true

    storage:
      trace:
        backend: local                     # backend configuration to use
        wal:
          path: /tmp/tempo/wal             # where to store the the wal locally
        local:
          path: /tmp/tempo/blocks

    querier:
      frontend_worker:
        grpc_client_config:
          tls_enabled: true
          tls_cert_path: /certs/tls.crt
          tls_key_path: /certs/tls.key
          tls_insecure_skip_verify: true

    overrides:
      defaults:
        metrics_generator:
          processors: [service-graphs, span-metrics] # enables metrics generator
```
**Expected behavior**
Tempo should return the traces available and when not working, it should log the error

**Environment:**
 - Infrastructure: Kubernetes

joe-elliott:
**What this PR does**:
Tempo currently has the ability to support both HTTP and GRPC requests on its primary HTTP port. It accomplishes it by using cmux through a feature in dskit. Unfortunately cmux breaks TLS on this port. This PR reworks our implementation of the shared GRPC/HTTP port to allow TLS to work.

Also:
- Add a TLS integration test
- Adds a "TempoServer" shim that controls the way HTTP routes are registered. 
- Default internalserver to listen on all IPs instead of just localhost

Fixes #3278 

**Checklist**
- [x] Tests updated
- [x] Documentation added


cmd/tempo/app/server_service.go
```diff
@@ -3,14 +3,101 @@ package app
...
 	"github.com/grafana/dskit/server"
 	"github.com/grafana/dskit/services"
+	"golang.org/x/net/http2"
+	"golang.org/x/net/http2/h2c"
+	"google.golang.org/grpc"
 
 	util_log "github.com/grafana/tempo/pkg/util/log"
 )
 
+type TempoServer interface {
```
zalegrala, line 22: Do we need the interface?  It looks like we would be able to use the struct directly.  Are you anticipating adding another implementation?

zalegrala (commented):
This looks pretty good.  Just a couple comments.

cmd/tempo/app/server_service.go
joe-elliott: the primary reason i used an interface was to limit the ways in which the rest of the code could interact with the `t.server` var. Another option would have been to move this into its own package but i suppose that didn't feel necessary.

integration/e2e/https_test.go
```diff
@@ -0,0 +1,86 @@
...
+	resp, err := apiClient.Do(echoReq)
+	require.NoError(t, err)
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+
+	// query an in-memory trace
+	queryAndAssertTrace(t, apiClient, info)
+	util.SearchAndAssertTrace(t, apiClient, info)
+	util.SearchTraceQLAndAssertTrace(t, apiClient, info)
+
+	creds := credentials.NewTLS(&tls.Config{InsecureSkipVerify: true})
```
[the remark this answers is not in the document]
joe-elliott: i think this is fine. TLS is still in use and functional. we're just not validating the cert against the CA's

integration/e2e/ca/ca.go
```diff
@@ -0,0 +1,212 @@
...
+		ClientCABothCertFile:      clientCABothCertFile,
+		Client1CertFile:           client1CertFile,
+		Client1KeyFile:            client1KeyFile,
+		Client2CertFile:           client2CertFile,
+		Client2KeyFile:            client2KeyFile,
+	}
+}
+
+type ca struct {
+	key    *ecdsa.PrivateKey
```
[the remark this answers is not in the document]
joe-elliott: we are restricted to whatever go supports i suppose. we're using vanilla HTTP everything.

integration/e2e/ca/ca.go
```diff
@@ -0,0 +1,212 @@
...
+	keyBytes, err := x509.MarshalECPrivateKey(key)
+	if err != nil {
+		return err
+	}
+
+	if err := writeExclusivePEMFile(keyPath, "PRIVATE KEY", 0o600, keyBytes); err != nil {
+		return err
+	}
+
+	template.IsCA = false
```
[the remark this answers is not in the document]
joe-elliott: the ca cert is handled in `writeCACertificate()` which uses the ca.cert struct which sets IsCA to true.

to be transparent, i lifted this file directly from dskit and am not familiar with all choices made.

cmd/tempo/app/server_service.go
zalegrala: I suppose my feeling is that interfaces should have more than one implementation, but I see your perspective to limit the interactions.

pkg/server/internal_server.go
```diff
@@ -15,7 +15,7 @@ type Config struct {
 
 // RegisterFlags add internal server flags to flagset
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
-	f.StringVar(&cfg.Config.HTTPListenAddress, "internal-server.http-listen-address", "localhost", "HTTP internal server listen address.")
```
zalegrala, line 18: With an empty string here, does the underlying implementation listen on everything or local or fail?

zalegrala (approved):
Looks good to me.  One comment about the behavior of an empty listen address for the internal server.

pkg/server/internal_server.go
joe-elliott: it listens on everything which is the default behavior for our main server as well. it took me awhile to figure out why the internal server wasn't responding and this was the reason.

imo, if you turn on/configure internal server you know you are exposing a new port and at that time you will make configuration choices to secure it in whatever way makes sense for your org.

right now if you do the minimal config to start the internal server it refuses requests and it's difficult to determine why.

zalegrala: Nice, I agree listen on everything unless you know better.

zalegrala: `netstat -luntp` or `apk add iproute2 && ss -ltn` are friends of mine.

joe-elliott: i am not the networking wizard you are :)

joe-elliott:
thanks for the review!

PR #3304, opened 2024-01-17
Improve frontend search context handling

### LINKED ISSUE #2680: Review/Fix Query Frontend cancellation behavior
While reviewing frontend cancellation I noticed some possibly inconsistent behavior. This issue will document what was found as well as what needs to be researched and improved.

**Observed Behavior**

1. When executing a query that exceeded the Tempo timeout (`2m` in this case) Grafana frontend receives a 504 and shows the following error:

image

Tempo logs a partially complete job with a 200:

image

2. If you start a query and instantly click the "Cancel" button in Grafana the query will still complete in Tempo.

image

**To Do:**

1. Review behavior of the query frontend in the face of different cancellation events
1. Establish appropriate behavior and make any adjustments as necessary
1. Consider this behavior both in terms of the HTTP and streaming GRPC query endpoints

**Cancellation Events:**

1. Execute a request that exceeds the timeout
1. Execute and then cancel a request before it has timed out

It would be interesting to do these in the context of Grafana specifically since it is the most common Tempo client.

zalegrala:
**What this PR does**:

Here we aim to improve the context handling on the search path in the case of cancellation and timeouts.

Related #2680

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/searchsharding.go
+++ b/modules/frontend/searchsharding.go
@@ -279,29 +279,20 @@ func (s searchSharder) RoundTrip(r *http.Request) (*http.Response, error) {
 		return nil, overallResponse.err
 	}
 
-	if ctxErr := ctx.Err(); ctxErr != nil { // context cancelled or deadline exceeded
-		overallResponse.statusMsg = ctxErr.Error()
-		if errors.Is(ctxErr, context.DeadlineExceeded) {
-			overallResponse.statusCode = http.StatusRequestTimeout
-		} else if errors.Is(ctxErr, context.DeadlineExceeded) {
-			overallResponse.statusCode = 499
-		}
-	}
-
 	switch overallResponse.statusCode {
 	case http.StatusOK:
-	case http.StatusTooManyRequests, http.StatusRequestTimeout, 499:
-		// Ensure we pass back the 408, 429, and 499 status.
+	case http.StatusTooManyRequests:
+		// Ensure we pass back the 429 status.
 		return &http.Response{
 			StatusCode: overallResponse.statusCode,
 			Header:     http.Header{},
 			Body:       io.NopCloser(strings.NewReader(overallResponse.statusMsg)),
 		}, nil
 	default:
-		// translate all non-200s and unmatched 4xx above into 500s. if, for
-		// instance, we get a 400 back from an internal component it means that we
-		// created a bad request. 400 should not be propagated back to the user b/c
-		// the bad request was due to a bug on our side, so return 500 instead.
+		// translate all non-200s and non-429s. if, for instance, we get a 400 back
+		// from an internal component it means that we created a bad request. 400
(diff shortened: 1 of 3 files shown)


modules/querier/worker/frontend_processor.go
```diff
@@ -66,8 +68,10 @@ func (fp *frontendProcessor) processQueriesOnSingleStream(ctx context.Context, c
 		}
 
 		if err := fp.process(c); err != nil {
-			level.Error(fp.log).Log("msg", "error processing requests", "address", address, "err", err)
-			backoff.Wait()
+			if status.Code(err) != codes.Canceled {
```
joe-elliott, line 71: So this is really interesting and worth a comment for sure. I didn't notice this behavior until I dug for the log message below.

It seems that when you cancel a query in Grafana the cancel is somehow propagated to the querier? B/c this line is hitting? i like dropping the backoff as well in this case b/c it will cause our queriers to reconnect quicker in the event of quick cancellations. 

further evidence this area is in need of overall improvement, but i think this is a great standalone patch.

modules/frontend/searchsharding.go
```diff
@@ -167,6 +167,7 @@ func (s searchSharder) RoundTrip(r *http.Request) (*http.Response, error) {
 
 		// if shouldQuit is true, terminate and abandon requests
 		if progress.shouldQuit() {
+			subCancel()
```
joe-elliott, line 170: nice catch. as we find ourselves pushing trace shards higher and higher in larger clusters this is becoming more important.

joe-elliott (approved):

modules/frontend/searchsharding.go
```diff
@@ -188,43 +189,43 @@ func (s searchSharder) RoundTrip(r *http.Request) (*http.Response, error) {
...
 					return
 				}
 
 				// if the status code is anything but happy, save the error and pass it down the line
 				if resp.StatusCode != http.StatusOK {
 					statusCode := resp.StatusCode
-					bytesMsg, err := io.ReadAll(resp.Body)
-					if err != nil {
-						_ = level.Error(s.logger).Log("msg", "error reading response body status != ok", "url", innerR.RequestURI, "err", err)
+					bytesMsg, ioErr := io.ReadAll(resp.Body)
```
joe-elliott, line 209: any particular reason you split these errors?

zalegrala: This routine err was shadowing the err outside the routine, so I think the change make it more clear.

modules/querier/worker/frontend_processor.go
zalegrala: I've added a comment.

zalegrala:
I had a chance to test this out last week and still looks good.

PR #3305, opened 2024-01-17
Local blocks processor honor per-tenant max trace size

mdisibio:
**What this PR does**:
Updates the local-blocks processor to honor the per-tenant override `max_trace_bytes`.  The generators already make use of the per-tenant overrides but this field.  This goes one step further than the trace size limits in the ingesters.  The ingesters keep track of the historical trace size only for the lifetime of the headblock, and that means long-running traces can eventually append more data.  This goes one step further and maintains the historical trace size for as long as the trace is active (continues to send spans).

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/processor/localblocks/livetraces.go
+++ b/modules/generator/processor/localblocks/livetraces.go
@@ -1,15 +1,12 @@
 package localblocks
 
 import (
-	"errors"
 	"hash"
 	"hash/fnv"
 	"time"
 
 	v1 "github.com/grafana/tempo/pkg/tempopb/trace/v1"
 )
-
-var errMaxExceeded = errors.New("asdf")
 
 type liveTrace struct {
 	id        []byte
--- a/modules/generator/processor/localblocks/livetraces.go
+++ b/modules/generator/processor/localblocks/livetraces.go
@@ -39,12 +39,7 @@ func (l *liveTraces) Len() uint64 {
 	return uint64(len(l.traces))
 }
 
-func (l *liveTraces) Push(batch *v1.ResourceSpans, max uint64) error {
-	if len(batch.ScopeSpans) == 0 || len(batch.ScopeSpans[0].Spans) == 0 {
-		return nil
-	}
-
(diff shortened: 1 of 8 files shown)


modules/generator/processor/localblocks/traceSizes.go
```diff
@@ -0,0 +1,65 @@
...
+
+type traceSize struct {
+	size      int
+	timestamp time.Time
+}
+
+func newTraceSizes() *traceSizes {
+	return &traceSizes{
+		hash:  fnv.New64(),
+		sizes: map[uint64]*traceSize{},
```
zalegrala, line 24: We may wish to `make()` here.

modules/generator/processor/localblocks/livetraces.go
```diff
@@ -39,21 +36,16 @@ func (l *liveTraces) Len() uint64 {
 	return uint64(len(l.traces))
 }
 
-func (l *liveTraces) Push(batch *v1.ResourceSpans, max uint64) error {
-	if len(batch.ScopeSpans) == 0 || len(batch.ScopeSpans[0].Spans) == 0 {
-		return nil
-	}
-
-	traceID := batch.ScopeSpans[0].Spans[0].TraceId
+func (l *liveTraces) Push(traceID []byte, batches []*v1.ResourceSpans, maxTraces uint64) bool {
```
zalegrala, line 39: Can we assert that the `batches` here contains only spans with the same traceID?

zalegrala (commented):
A couple comments, but overall looks reasonable to me.

modules/generator/processor/localblocks/livetraces.go
mdisibio: Good catch. It actually doesn't because I missed that the distributor [regroups the batches across traces]
(https://github.com/grafana/tempo/blob/main/modules/distributor/distributor.go#L460).  I'll fix.

zalegrala (approved):
Looks good to me.

PR #3308, opened 2024-01-17
Change spanmetrics filtering with multiple policies

zalegrala:
**What this PR does**:

Here is an issue that was identified when trying to use multiple include policies against the same intrinsic attribute.  Without this change, multiple policies operating on the same attributes will never match since we find the first opportunity to reject a span from being included in the metrics.  Here we make a change to continue evaluating additional filter policies even if the current one is unmatched.  This means that a filter policy uses `AND` logic internally, but multiple polices use `OR` logic to determine if the span should be included.  Any filter policy that is matched will include the metric.

  **BREAKING CHANGE** Exclude policies are no longer global

  The documentation currently suggests the following policy to include and
  exclude using two filter policies.

  ```
    metrics_generator:
      processor:
        span_metrics:
          filter_policies:
            - include:
                match_type: regex
                attributes:
                  - key: resource.location
                    value: eu-.*
            - exclude:
                match_type: regex
                attributes:
                  - key: resource.tier
                    value: dev-.*
  ```

  Currently, at the first non-matching filter, the span is dropped.  The
  breaking change here is that when an include is not matched, we continue
  evaluating additional policies for possible matches.  In the above scenario,
  each exclude must now be paired with the include that it wishes to further
  filter.  The above can be re-written with the include and exclude in the same
  policy filter.

  ```
    metrics_generator:
      processor:
        span_metrics:
          filter_policies:
            - include:
                match_type: regex
                attributes:
                  - key: resource.location
                    value: eu-.*
              exclude:
                match_type: regex
                attributes:
                  - key: resource.tier
                    value: dev-.*
  ```

  Users are now able to write an additional filter policy that matches on the
  same attribute (`resource.location`), where previously the first non-match
  was used as a reason not to include the span.

  ```
    metrics_generator:
      processor:
        span_metrics:
          filter_policies:
            - include:
                match_type: regex
                attributes:
                  - key: resource.location
                    value: eu-.*
            - include:
                match_type: static
                attributes:
                  - key: resource.location
                    value: moon
  ```

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@
 * [BUGFIX] Change exit code if config is successfully verified [#3174](https://github.com/grafana/tempo/pull/3174) (@am3o @agrib-01)
 * [BUGFIX] The tempo-cli analyse blocks command no longer fails on compacted blocks [#3183](https://github.com/grafana/tempo/pull/3183) (@stoewer)
 * [BUGFIX] Move waitgroup handling for poller error condition [#3224](https://github.com/grafana/tempo/pull/3224) (@zalegrala)
-* [BUGFIX] Continue evaluating spanmetrics filter policies even if the first one doesn't match [#3308](https://github.com/grafana/tempo/pull/3308) (@zalegrala)
 * [ENHANCEMENT] Introduced `AttributePolicyMatch` & `IntrinsicPolicyMatch` structures to match span attributes based on strongly typed values & precompiled regexp [#3025](https://github.com/grafana/tempo/pull/3025) (@andriusluk)
 * [CHANGE] TraceQL/Structural operators performance improvement. [#3088](https://github.com/grafana/tempo/pull/3088) (@joe-elliott)
 * [CHANGE] Merge the processors overrides set through runtime overrides and user-configurable overrides [#3125](https://github.com/grafana/tempo/pull/3125) (@kvrhdn)
@@ -15,6 +14,74 @@
 * [CHANGE] Update Alpine image version to 3.19 [#3289](https://github.com/grafana/tempo/pull/3289) (@zalegrala)
 * [CHANGE] Introduce localblocks process config option to select only server spans 3303https://github.com/grafana/tempo/pull/3303 (@zalegrala)
 * [CHANGE] Localblocks processor honor tenant max trace size limit [3305](https://github.com/grafana/tempo/pull/3305) (@mdisibio)
+* [CHANGE] Continue evaluating spanmetrics filter policies even if not matched in one policy [#3308](https://github.com/grafana/tempo/pull/3308) (@zalegrala)
+  **BREAKING CHANGE** Exclude policies are no longer global
+
+  The documentation currently suggests the following policy to include and
+  exclude using two filter policies.
+
+  ```
+    metrics_generator:
+      processor:
+        span_metrics:
+          filter_policies:
+            - include:
+                match_type: regex
+                attributes:
+                  - key: resource.location
+                    value: eu-.*
+            - exclude:
+                match_type: regex
+                attributes:
+                  - key: resource.tier
+                    value: dev-.*
+  ```
+
+  Currently, at the first non-matching filter, the span is dropped.  The
+  breaking change here is that when an include is not matched, we continue
+  evaluating additional policies for possible matches.  In the above scenario,
+  each exclude must now be paired with the include that it wishes to further
+  filter.  The above can be re-written with the include and exclude in the same
+  policy filter.
+
+  ```
+    metrics_generator:
+      processor:
+        span_metrics:
+          filter_policies:
+            - include:
+                match_type: regex
+                attributes:
+                  - key: resource.location
+                    value: eu-.*
+              exclude:
+                match_type: regex
+                attributes:
+                  - key: resource.tier
+                    value: dev-.*
+  ```
+
+  Users are now able to write an additional filter policy that matches on the
+  same attribute (`resource.location`), where previously the first non-match
+  was used as a reason not to include the span.
+
+  ```
+    metrics_generator:
+      processor:
+        span_metrics:
+          filter_policies:
+            - include:
+                match_type: regex
+                attributes:
+                  - key: resource.location
+                    value: eu-.*
+            - include:
+                match_type: static
+                attributes:
+                  - key: resource.location
+                    value: moon
+  ```
+
 * [CHANGE] Major cache refactor to allow multiple role based caches to be configured [#3166](https://github.com/grafana/tempo/pull/3166).
   **BREAKING CHANGE** Deprecate the following fields. These have all been migrated to a top level "cache:" field.
   ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * [BUGFIX] Change exit code if config is successfully verified [#3174](https://github.com/grafana/tempo/pull/3174) (@am3o @agrib-01)
 * [BUGFIX] The tempo-cli analyse blocks command no longer fails on compacted blocks [#3183](https://github.com/grafana/tempo/pull/3183) (@stoewer)
 * [BUGFIX] Move waitgroup handling for poller error condition [#3224](https://github.com/grafana/tempo/pull/3224) (@zalegrala)
+* [BUGFIX] Continue evaluating spanmetrics filter policies even if the first one doesn't match [#3308](https://github.com/grafana/tempo/pull/3308) (@zalegrala)
 * [ENHANCEMENT] Introduced `AttributePolicyMatch` & `IntrinsicPolicyMatch` structures to match span attributes based on strongly typed values & precompiled regexp [#3025](https://github.com/grafana/tempo/pull/3025) (@andriusluk)
 * [CHANGE] TraceQL/Structural operators performance improvement. [#3088](https://github.com/grafana/tempo/pull/3088) (@joe-elliott)
 * [CHANGE] Merge the processors overrides set through runtime overrides and user-configurable overrides [#3125](https://github.com/grafana/tempo/pull/3125) (@kvrhdn)
--- a/docs/sources/tempo/metrics-generator/span_metrics.md
+++ b/docs/sources/tempo/metrics-generator/span_metrics.md
@@ -125,8 +125,8 @@ func (ds RatioBasedSampler) Description() string {
 
 ### Filtering
 
-In some cases, you may want to reduce the number of metrics that are produced
-from the `spanmetrics` processor by using a filter. With a filter, you can
+In some cases, you may want to use a filter to reduce the number of metrics that are produced
+from the `spanmetrics` processor. With a filter, you can
 configure the processor to use an `include` statement to match criteria that
 must be present in the span in order to be included in the metrics. Following
 the include filter, you can use an `exclude` filter to reject portions of what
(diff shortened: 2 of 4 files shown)


pkg/spanfilter/spanfilter.go
```diff
@@ -56,17 +56,21 @@ func (f *SpanFilter) ApplyFilterPolicy(rs *v1.Resource, span *tracev1.Span) bool
...
-			return false
+			continue
 		}
 
 		if policy.Exclude != nil && policy.Exclude.Match(rs, span) {
-			return false
+			continue
 		}
+
+		policyMatch++
```
joe-elliott, line 70: can't we just return true here instead of incrementing policyMatch and evaluating other policies?

joe-elliott (commented):
lgtm. one suggested improvement

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -125,9 +125,12 @@ func (ds RatioBasedSampler) Description() string {
 
 ### Filtering
 
-In some cases, you may want to reduce the number of metrics produced by the `spanmetrics` processor.
-You can configure the processor to use an `include` filter to match criteria that must be present in the span in order to be included.
-Following the include filter, you can use an `exclude` filter to reject portions of what was previously included by the filter policy.
+In some cases, you may want to reduce the number of metrics that are produced
+from the `spanmetrics` processor by using a filter. With a filter you can
```
knylander-grafana, line 129:
```suggestion
from the `spanmetrics` processor by using a filter. With a filter, you can
```

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -208,6 +211,49 @@ metrics_generator:
 In the above, we first include all spans which have a `resource.location` that begins with `eu-` with the `include` statement, and then exclude those with begin with `dev-`.
 In this way, a flexible approach to filtering can be achieved to ensure that only metrics which are important are generated.
 
+Additionally, when multiple filter policies are are used, any policy that is
+matched will result in the span being included in the metric export.
```
knylander-grafana, line 215:
```suggestion
matched results in the span being included in the metric export.
```

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -208,6 +211,49 @@ metrics_generator:
...
+              - key: resource.tier
+                value: dev-.*
+```
+
+In this example, spans that match a location beginning with `eu-` or `sa-` will
+be included so long as they are not a `resource.tier` beginning with `dev-`.
+This example could also be written with a single regex expression, except that
+in the second policy we have an additional constraint to match against.  If
+either the `resource.location` or the `resource.platform` is not matched in the
+second policy, then the span is not included.
```
knylander-grafana, line 252:
```suggestion
In this example, spans that match a location beginning with `eu-` or `sa-` are
included so long as they are not a `resource.tier` beginning with `dev-`.
This example could also be written with a single regex expression, except that
in the second policy, we have an additional constraint to match against. If
either the `resource.location` or the `resource.platform` is not matched in the
second policy, then the span is not included.
```

knylander-grafana (commented):
Thank you for adding docs. I've made some minor suggestions.

pkg/spanfilter/spanfilter.go
zalegrala: When testing this it showed me that we needed to also handle the case where only have exclude policies are in place.  I've adjusted the logic to be a bit more verbose, and added additional tests and docs.

docs/sources/tempo/metrics-generator/span_metrics.md
zalegrala: I reworked this a bit.

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -125,9 +125,12 @@ func (ds RatioBasedSampler) Description() string {
 
 ### Filtering
 
-In some cases, you may want to reduce the number of metrics produced by the `spanmetrics` processor.
-You can configure the processor to use an `include` filter to match criteria that must be present in the span in order to be included.
-Following the include filter, you can use an `exclude` filter to reject portions of what was previously included by the filter policy.
+In some cases, you may want to reduce the number of metrics that are produced
+from the `spanmetrics` processor by using a filter. With a filter, you can
```
knylander-grafana, line 129: Moving "use a filter" to the front of the sentence because it's the method people can use to reduce the number of metrics.
```suggestion
In some cases, you may want to use a filter to reduce the number of metrics that are produced
from the `spanmetrics` processor. With a filter, you can
```

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -150,6 +153,39 @@ The following intrinsic kinds are available for filtering.
 - `SPAN_KIND_PRODUCER`
 - `SPAN_KIND_CONSUMER`
 
+
+#### Filter policies
+
+Each filter policy is evaluated in order for every span to determine if the
+span should be included for metrics.  Both `include` and `exclude` policies are
+used to *reject* spans from the metrics.  With no filter polices applied, all
+spans are included.
```
knylander-grafana, line 162: I'm having some trouble parsing the first sentence. Do you mean 1) that every filter policy is evaluated in a particular order (i.e., policy A first, policy B, policy C, etc.) or 2) that each policy is used to evaluated every span to determine whether a span should be included for metrics? 
```suggestion
Each filter policy is evaluated in order for every span to determine if the
span should be included for metrics.  Both `include` and `exclude` policies are
used to *reject* spans from the metrics.  With no filter polices applied, all
spans are included.
```

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -150,6 +153,39 @@ The following intrinsic kinds are available for filtering.
...
+{{% admonition type="note" %}}
+An exclude match will take precedence over an include match within the *same*
+policy, but not later polices.
+{{% /admonition %}}
+
+{{% admonition type="note" %}}
+A matching exclude of one policy will be ignored if a later policy `include`
+matches, and that same policy does not also have a matching `exclude` for the
+span.
+{{% /admonition %}}
```
knylander-grafana, line 173: Changing future tense to present tense to follow our [style guidelines](https://grafana.com/docs/writers-toolkit/write/style-guide/style-conventions/#write-in-present-tense). 

```suggestion
An exclude match takes precedence over an include match within the *same*
policy, but not later polices.
{{% /admonition %}}

{{% admonition type="note" %}}
A matching exclude of one policy is ignored if a later policy `include`
matches, and that same policy does not also have a matching `exclude` for the
span.
{{% /admonition %}}
```

docs/sources/tempo/metrics-generator/span_metrics.md
knylander-grafana: Okay, reading later down I think you're referring to the order of the filter policies.

kvrhdn (approved):
LGTM :+1: Dealing with multiple policies gets complicated but I feel like the examples in the docs will help a lot. The logic makes sense to me.

docs/sources/tempo/metrics-generator/span_metrics.md
zalegrala: Okay I've updated.  The policies are evaluated in order yes (0,1,2), but only until there is a match to include without a matching exclude on the same policy.  Hows that read?

CHANGELOG.md
```diff
@@ -14,6 +14,74 @@
 * [CHANGE] Update Alpine image version to 3.19 [#3289](https://github.com/grafana/tempo/pull/3289) (@zalegrala)
 * [CHANGE] Introduce localblocks process config option to select only server spans 3303https://github.com/grafana/tempo/pull/3303 (@zalegrala)
 * [CHANGE] Localblocks processor honor tenant max trace size limit [3305](https://github.com/grafana/tempo/pull/3305) (@mdisibio)
+* [CHANGE] Continue evaluating spanmetrics filter policies even if not matched in one policy [#3308](https://github.com/grafana/tempo/pull/3308) (@zalegrala)
+  **BREAKING CHANGE** Exclude policies are no longer global
```
knylander-grafana, line 18: This is an excellent write-up of the breaking change. Thank you. We can add this to the Upgrade page when we do the next release.

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -150,6 +153,39 @@ The following intrinsic kinds are available for filtering.
 - `SPAN_KIND_PRODUCER`
 - `SPAN_KIND_CONSUMER`
 
+
+#### Filter policies
+
+Each filter policy is evaluated for every span until an `include` policy is
+matched without a corresponding `exclude` policy match within the same filter.
```
knylander-grafana, line 160: Much better! Thank you.

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -198,15 +240,59 @@ metrics_generator:
...
                 value: dev-.*
 ```
 
-In the above, we first include all spans which have a `resource.location` that begins with `eu-` with the `include` statement, and then exclude those with begin with `dev-`.
-In this way, a flexible approach to filtering can be achieved to ensure that only metrics which are important are generated.
+In the above, we first include all spans which have a `resource.location` that
+begins with `eu-` with the `include` statement, and then `exclude` those with
+begin with `dev-`. In this way, a flexible approach to filtering can be
+achieved to store only those metrics which are important.
+
```
knylander-grafana, line 254:
```suggestion
The above example first includes all spans which have a `resource.location` that
begins with `eu-` with the `include` statement, and then `exclude` those with
begin with `dev-`. In this way, you can create a flexible approach to filtering to store 
only those metrics which are important.

```

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -198,15 +240,59 @@ metrics_generator:
...
+                value: k8s-.*
+          exclude:
+            match_type: regex
+            attributes:
+              - key: resource.tier
+                value: research
+```
+
+In this example, spans that match a location beginning with `eu-` or `sa-` will
+be included, but in both cases we also exclude spans which match
```
knylander-grafana, line 289:
```suggestion
In this example, spans that match a location beginning with `eu-` or `sa-` are 
included, but both cases also exclude spans which match
```

docs/sources/tempo/metrics-generator/span_metrics.md
```diff
@@ -198,15 +240,59 @@ metrics_generator:
...
+`{resource.tier="research"}`.
+
+This example could also be written with a single regex expression, except that
+in the second policy we have an additional constraint to match against for
+`k8s-.*`.  If either the `resource.location` or the `resource.platform` is not
+matched in the second policy, then the span is not included.
```
knylander-grafana, line 295: Slight tweak to exclude second person "we". 
```suggestion
This example could also be written with a single regex expression, except that
in the second policy has an additional constraint to match against for
`k8s-.*`.  If either the `resource.location` or the `resource.platform` is not
matched in the second policy, then the span is not included.
```

knylander-grafana (approved):
Doc changes are excellent. Thank you for updating the content and adding so many examples.

PR #3324, opened 2024-01-24
Fix typo in service graph metric name

09jvilla:
Seems like we updated the metric names from tempo_ to traces_ in the section below, but not in this specific example.


joe-elliott (approved):

jamesrwhite:
This typo is still live on the site, does something need to be done for the fix to be released?

https://grafana.com/docs/tempo/latest/metrics-generator/service_graphs/

09jvilla:
I was confused by this too @jamesrwhite , but the fix should be live when we publish the next release of Tempo. Our documentation is versioned so fixes don't go live immediately, but rather are reflected in the next version.

09jvilla:
You can see its fixed here:
https://grafana.com/docs/tempo/next/metrics-generator/service_graphs/

PR #3326, opened 2024-01-24
Improve query cancellation handling in tracebyidsharding

zalegrala:
**What this PR does**:

Here we aim to improve the context handling on the TraceByID path in the case of cancellation and timeouts.

Related https://github.com/grafana/tempo/issues/2680

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/tracebyidsharding.go
+++ b/modules/frontend/tracebyidsharding.go
@@ -171,12 +171,8 @@ func (s shardQuery) RoundTrip(r *http.Request) (*http.Response, error) {
 		switch statusCode {
 		case http.StatusNotFound:
 			// Pass through 404s
-		case http.StatusRequestTimeout:
-			// Pass through 408s
 		case http.StatusTooManyRequests:
 			// Pass through 429s
-		case 499:
-			// Pass through 499s, client closed connection
 		default:
 			statusCode = http.StatusInternalServerError
 		}
@@ -248,12 +244,7 @@ func shouldQuit(ctx context.Context, statusCode int, err error) bool {
 		return true
 	}
 
-	switch statusCode {
-	case http.StatusTooManyRequests:
-		return true
-	case http.StatusRequestTimeout:
-		return true
-	case 499:
+	if statusCode == http.StatusTooManyRequests {
 		return true
 	}
 
--- a/modules/frontend/tracebyidsharding.go
+++ b/modules/frontend/tracebyidsharding.go
@@ -65,7 +65,10 @@ func (s shardQuery) RoundTrip(r *http.Request) (*http.Response, error) {
 
 	// context propagation
 	r = r.WithContext(ctx)
-	reqs, err := s.buildShardedRequests(r)
+	subCtx, subCancel := context.WithCancel(ctx)
+	defer subCancel()
+
+	reqs, err := s.buildShardedRequests(subCtx, r)
 	if err != nil {
 		return nil, err
 	}
@@ -75,19 +78,28 @@ func (s shardQuery) RoundTrip(r *http.Request) (*http.Response, error) {
 	if s.cfg.ConcurrentShards > 0 {
 		concurrentShards = uint(s.cfg.ConcurrentShards)
 	}
-	wg := boundedwaitgroup.New(concurrentShards)
-	mtx := sync.Mutex{}
 
-	var overallError error
+	var (
+		overallError error
+
+		mtx        = sync.Mutex{}
+		statusCode = http.StatusNotFound
+		statusMsg  = "trace not found"
+		wg         = boundedwaitgroup.New(concurrentShards)
+	)
+
 	combiner := trace.NewCombiner(s.o.MaxBytesPerTrace(userID))
 	_, _ = combiner.Consume(&tempopb.Trace{}) // The query path returns a non-nil result even if no inputs (which is different than other paths which return nil for no inputs)
-	statusCode := http.StatusNotFound
-	statusMsg := "trace not found"
 
 	for _, req := range reqs {
 		wg.Add(1)
 		go func(innerR *http.Request) {
-			defer wg.Done()
+			defer func() {
+				if shouldQuit(innerR.Context(), statusCode, overallError) {
+					subCancel()
+				}
+				wg.Done()
+			}()
 
 			resp, rtErr := s.next.RoundTrip(innerR)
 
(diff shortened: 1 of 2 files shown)


modules/frontend/tracebyidsharding.go
```diff
@@ -65,7 +65,10 @@ func (s shardQuery) RoundTrip(r *http.Request) (*http.Response, error) {
 
 	// context propagation
 	r = r.WithContext(ctx)
-	reqs, err := s.buildShardedRequests(r)
+	subCtx, subCancel := context.WithCancel(ctx)
+	defer subCancel()
+
```
electron0zero, line 70: nit: if it's not too hard to test, can we have test to test the cancellation behavour. test would ensure that we don't break it again :)

electron0zero (approved):
one nit, rest lgtm 👍🏼

modules/frontend/tracebyidsharding.go
```diff
@@ -168,8 +171,12 @@ func (s shardQuery) RoundTrip(r *http.Request) (*http.Response, error) {
 		switch statusCode {
 		case http.StatusNotFound:
 			// Pass through 404s
+		case http.StatusRequestTimeout:
```
joe-elliott, line 174: are we seeing this error anywhere? does it make sense to bubble this up?

if the querier returns this code to the frontend and then the frontend propagates it to the client that doesn't seem correct b/c the frontend isn't actually closing an unused connection as suggested by the code.

modules/frontend/tracebyidsharding.go
```diff
@@ -97,20 +104,16 @@ func (s shardQuery) RoundTrip(r *http.Request) (*http.Response, error) {
...
-			if rtErr != nil {
-				_ = level.Error(s.logger).Log("msg", "error querying proxy target", "url", innerR.RequestURI, "err", rtErr)
-				overallError = rtErr
+			// Check the context of the worker request
+			if shouldQuit(innerR.Context(), statusCode, overallError) {
 				return
 			}
 
-			// if the status code is anything but happy, save the error and pass it down the line
+			// if the status code is anything but happy, save the error and pass it
```
joe-elliott, line 112: why did we remove the `rtErr` check that was here?

modules/frontend/tracebyidsharding.go
```diff
@@ -168,8 +171,12 @@ func (s shardQuery) RoundTrip(r *http.Request) (*http.Response, error) {
 		switch statusCode {
 		case http.StatusNotFound:
 			// Pass through 404s
+		case http.StatusRequestTimeout:
+			// Pass through 408s
 		case http.StatusTooManyRequests:
 			// Pass through 429s
+		case 499:
```
joe-elliott, line 178: similar thought here? have we seen 499s? do we know if the go http client actually returns this status code on timeout? i would expect it just to return errors.

joe-elliott (approved):

modules/frontend/tracebyidsharding.go
zalegrala: It looks like its unreachable to me.  The `shouldQuit` above would have caught it because `overallError` is checked and if non-nil above, then then we've returned.

joe-elliott: yup, good call!

modules/frontend/tracebyidsharding.go
zalegrala: Good pointing out.  Currently with this PR, if the timeout config is very low, the response to grafana is the following.

```json
{
    "results": {
        "A": {
            "error": "failed to get trace with id: febc8dcb5c086ba3c18506f49042c86b Status: 499 status code 499 Body: context canceled",
            "errorSource": "",
            "status": 500
        }
    }
}
```

Additionally, the error code show up in the logs.
```
2024-01-26T19:32:17.608743968Z stderr F level=info ts=2024-01-26T19:32:17.607466177Z caller=handler.go:87 tenant=ops method=GET traceID=d8c8ed5e419f91542b89ce6c6daf503a url=/api/traces/febc8dcb5c086ba3c18506f49042c86b duration=74.136385ms response_size=0 status=500 err="rpc error: code = Code(499) desc = context canceled"
```

Here is a case of a "context cancelled" error with a `499` but because the config timeout is so low, I would have expected a "deadline exceeded" message.  I added the case here because I think both reasons for a context error should be covered, deadline exceeded and cancelled and probably should be returned to the frontend in the same way.  What do you think?

Here is a change we could make right above this block you are pointing at, but I'm not sure its needed.

```go
	if ctxErr := ctx.Err(); ctxErr != nil { // context cancelled or deadline exceeded
		statusMsg = ctxErr.Error()
		if errors.Is(ctxErr, context.DeadlineExceeded) {
			statusCode = http.StatusRequestTimeout
		} else if errors.Is(ctxErr, context.DeadlineExceeded) {
			statusCode = 499
		}
	}
```

This would synthetically set the status based on the context error.  The 499 here is a non-standard, but is in use in a few places, including `frontend.StatusClientClosedRequest`, so it seems like a reasonable to continue using.

modules/frontend/tracebyidsharding.go
zalegrala: See the response above.

modules/frontend/tracebyidsharding.go
zalegrala: I added a test case, but its only against eh mock.  This is challenging to test e2e.  How does the test look to you?  Do you see a simpler path to testing this change?

modules/frontend/tracebyidsharding.go
joe-elliott: > if the timeout config is very low

which timeout config?

> Additionally, the error code show up in the logs.

This looks like the frontend gives up on the querier and the 499 is propagated up the handlers and we 500 at the frontend? I honestly don't think we should ever return a 499. It's a "fake" status code that a client will generate when it gives up. It doesn't make sense to me to return this in a real http response.

Which raise the question of what should we return? If an internal request between Tempo components times out then 500 feels like a pretty good catch all. Basically some kind of unknown communication issue between components.

zalegrala: I'll drop the status updates here, which will continue to be a 500 back to the frontend.

modules/frontend/tracebyidsharding.go
zalegrala: I backed out the test case since it was requested we don't change the status handling.  I don't think there is good way to test this.

PR #3328, opened 2024-01-24
Ingester lock tweaks to minimize headblock locking during search

mdisibio:
**What this PR does**:
We have seen occasional high read latencies on ingesters where tag lookups and searches take 20+ seconds instead of the typical few.  Investigation showed that it is contention on the block mutexes, and specifically due to the design of `instance.Search` and `SearchResults`.  Instead of searching and releasing the headblock asap, the lock was held until after also acquiring a lock on the rest of the blocks and beginning to read from `SearchResults`.  High latency occurs when read/write access to the mutexes stack up, such that it might have to wait for a current search to finish (releasing read lock), then a block flush (write lock), then acquiring read lock again.

This moves away from `SearchResults` and the channel-based flow to a simpler setup like the other endpoints.  Since this was the only usage of `SearchResults` this seems better than trying to patch it (again).

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


mdisibio:
Putting back to draft because allocating the channel size based on the limit doesn't work for cases where the limit is large (1,000,000+)

tempodb/encoding/vparquet3/wal_block.go
```diff
@@ -699,6 +699,7 @@ func (b *walBlock) FetchTagValues(ctx context.Context, req traceql.AutocompleteR
 		if err != nil {
 			return fmt.Errorf("error opening file %s: %w", page.path, err)
 		}
+		defer file.Close()
```
zalegrala, line 702: Is the defer inside the loop enough, or do we want to close it out each iteration through `blockFlushes`?

mdisibio: That's correct, it's inside the loop and all of them get closed at the end, which is not optimal.   This func and several others are all following the same pattern, and I was aiming to have minimum changes necessary.  But happy to make larger changes and defer as expected if we want.

modules/ingester/instance_search.go
```diff
@@ -149,20 +75,93 @@ func (i *instance) searchBlock(ctx context.Context, req *tempopb.SearchRequest,
...
 			}
 		}
-		sr.AddBlockInspected()
+	}
 
-		sr.AddBytesInspected(resp.Metrics.InspectedBytes)
-		sr.AddTraceInspected(resp.Metrics.InspectedTraces)
-	}(block, cleanup)
+	// Search headblock (synchronously)
+	// Lock headblock separately from other blocks and release it as quickly as possible.
```
joe-elliott, line 103: i only see one mutex locked (unless this comment is also referring to the resultsMtx). is this still relevant?

modules/ingester/instance_search.go
```diff
@@ -149,20 +75,93 @@ func (i *instance) searchBlock(ctx context.Context, req *tempopb.SearchRequest,
...
+	}
+	if combiner.Count() >= maxResults {
+		return &tempopb.SearchResponse{
+			Traces:  combiner.Metadata(),
+			Metrics: metrics,
+		}, nil
+	}
+
+	// Search all other blocks (concurrently)
+	// Lock blocks mutex until all search tasks are finished and this function exists. This avoids
```
joe-elliott, line 125: nit:

exists -> exits

modules/ingester/instance_search.go
```diff
@@ -149,20 +75,93 @@ func (i *instance) searchBlock(ctx context.Context, req *tempopb.SearchRequest,
...
+
+		if resp.Metrics != nil {
+			metrics.InspectedTraces += resp.Metrics.InspectedTraces
+			metrics.InspectedBytes += resp.Metrics.InspectedBytes
+		}
+
+		for _, tr := range resp.Traces {
+			combiner.AddMetadata(tr)
+			if combiner.Count() >= maxResults {
+				return
```
joe-elliott, line 97: can we cancel context or anything here? to force the local block searches to bail out?

modules/ingester/instance_search.go
mdisibio: The rest of the blocks are locked [here](https://github.com/grafana/tempo/pull/3328/files#diff-8c9aac72ec16ad39eb093753942cde7f893be0b5950578ffe10ef003d2ff5237R128)

There are 3 mutexes:
* i.headBlockMtx - for head block, unlocked asap
* i.blocksMtx - for all completing and complete blocks, unlocked when the last goroutine exits
* resultsMtx - for the concurrent results

joe-elliott: since they aren't nested, does the order matter? wouldn't the headblock be released and then it would sit waiting for the blocks mutex? if another goroutine did it in the opposite order they would unblock for each other?

ultimately it's not a big deal. i just read this comment and it seemed that the order was unimportant.

joe-elliott (approved):
lgtm 👍

PR #3320, opened 2024-01-22
[DOC] Add documentation for prometheus metrics compatibility

zalegrala:
**What this PR does**:

Here we add some documentation about how to use Tempo as a Prometheus datasource in Grafana.  We don't expect this current approach to be long-lived, but without a timeline in sight for the frontend, we want folks to be able to try it out.

**Checklist**
- [ ] Tests updated
- [x] Documentation added
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #3329, opened 2024-01-25
Add TEMPO_VERSION for links

knylander-grafana:
**What this PR does**:
Sets the version for Tempo for release. 

@jdbaldry Would you check this PR to make sure that it's correct? 

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [X] Documentation added

### DIFF
--- a/docs/sources/tempo/_index.md
+++ b/docs/sources/tempo/_index.md
@@ -5,6 +5,7 @@ aliases:
   - /docs/tempo/
 cascade:
   GRAFANA_VERSION: next
+  TEMPO_VERSION: latest
 ---
 
 # Tempo documentation


knylander-grafana:
@jdbaldry  What's the best way of making sure we have support for the cascade: TEMPO_VERSION in previous versions? I'm assuming checkout those branches, update the index file, and then merge it instead of backporting this one and updating those PRs.

jdbaldry (changes requested):
I think this PR should not be merged because version substitution of `<TEMPO_VERSION>` in these docs already does the right thing.

Right now:

> The special syntax <SOMETHING_VERSION> is substituted by the version that is inferred from the page’s URL. If the page’s URL has the prefix /docs/grafana/latest/, the syntax <SOMETHING_VERSION> is replaced by latest in the final URL.

(https://grafana.com/docs/writers-toolkit/write/shortcodes/#about-version-substitution)

If you set this front matter, `<TEMPO_VERSION>` will always be replaced by "latest" which won't make sense for "next" documentation or older versions.

jdbaldry:
> > I think this PR should not be merged because version substitution of `<TEMPO_VERSION>` in these docs already does the right thing.
> > Right now:
> > > The special syntax <SOMETHING_VERSION> is substituted by the version that is inferred from the page’s URL. If the page’s URL has the prefix /docs/grafana/latest/, the syntax <SOMETHING_VERSION> is replaced by latest in the final URL.
> > 
> > 
> > ([grafana.com/docs/writers-toolkit/write/shortcodes/#about-version-substitution](https://grafana.com/docs/writers-toolkit/write/shortcodes/#about-version-substitution))
> > If you set this front matter, `<TEMPO_VERSION>` will always be replaced by "latest" which won't make sense for "next" documentation or older versions.
> 
> Thank you. I wasn't sure how this process worked and was still unclear after reading Writers' Toolkit. If the version is set automatically by tooling, then I'm good to close this PR. The linked PR (#3320) uses TEMPO_VERSION in a link. I opened this PR to make sure we had TEMPO_VERSION set some where.

Could we improve the explanation in Writers' Toolkit to make this easier to understand?

PR #3332, opened 2024-01-25
Add HTML pages to view tenants with overrides

kvrhdn:
**What this PR does**:
Changes `/status/overrides` and `/status/overrides/{tenant}` to HTML page which are nicer and easier to use.

Examples:

`/status/overrides` with user-configurable overrides disabled

> Screenshot 2024-01-25 at 20 13 38

`/status/overrides/{tenant}` with user-configurable overrides disabled

> Screenshot 2024-01-25 at 20 13 54

`/status/overrides` with user-configurable overrides enabled

> Screenshot 2024-01-25 at 20 12 08

`/status/overrides/{tenant}` with user-configurable overrides enabled

> Screenshot 2024-01-25 at 20 12 27

> Screenshot 2024-01-25 at 20 12 43

~~Fixes #<issue number>~~

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/overrides/interface.go
+++ b/modules/overrides/interface.go
@@ -71,7 +71,4 @@ type Interface interface {
 
 	// Management API
 	WriteStatusRuntimeConfig(w io.Writer, r *http.Request) error
-
-	TenantsHandler(w http.ResponseWriter, req *http.Request)
-	TenantStatusHandler(w http.ResponseWriter, req *http.Request)
 }
--- a/modules/overrides/interface.go
+++ b/modules/overrides/interface.go
@@ -21,6 +21,9 @@ type Service interface {
 type Interface interface {
 	prometheus.Collector
 
+	// GetTenantIDs returns all tenants that have non-default overrides.
+	GetTenantIDs() []string
+
 	// GetRuntimeOverridesFor returns the runtime overrides set for the given user excluding
 	// overrides from the user-configurable overrides, if enabled.
 	GetRuntimeOverridesFor(userID string) *Overrides
@@ -68,5 +71,7 @@ type Interface interface {
 
 	// Management API
 	WriteStatusRuntimeConfig(w io.Writer, r *http.Request) error
-	WriteTenantOverrides(w io.Writer, r *http.Request, tenant string) error
+
+	TenantsHandler(w http.ResponseWriter, req *http.Request)
+	TenantStatusHandler(w http.ResponseWriter, req *http.Request)
 }
--- a/modules/overrides/overrides_tenants_http.go
+++ b/modules/overrides/overrides_tenants_http.go
@@ -26,58 +26,54 @@ type tenantsPageTenant struct {
 	HasUserConfigurableOverrides bool   `json:"has_user_configurable_overrides"`
 }
 
-func (o *runtimeConfigOverridesManager) TenantsHandler(w http.ResponseWriter, req *http.Request) {
-	var tenants []tenantsPageTenant
-	for _, tenant := range o.GetTenantIDs() {
-		tenants = append(tenants, tenantsPageTenant{
-			Name:                         tenant,
-			HasRuntimeOverrides:          true,
-			HasUserConfigurableOverrides: false,
-		})
-	}
+func TenantsHandler(o Interface) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		tenants := make(map[string]tenantsPageTenant)
 
-	sortTenantsPageTenant(tenants)
-
-	util.RenderHTTPResponse(w, tenantsPageContents{
-		Now:     time.Now(),
-		Tenants: tenants,
-	}, tenantsTemplate, req)
-}
-
-func (o *userConfigurableOverridesManager) TenantsHandler(w http.ResponseWriter, req *http.Request) {
-	tenants := make(map[string]tenantsPageTenant)
-
-	// runtime overrides
-	for _, tenant := range o.Interface.GetTenantIDs() {
-		tenants[tenant] = tenantsPageTenant{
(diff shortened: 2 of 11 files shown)


modules/overrides/interface.go
```diff
@@ -68,5 +71,7 @@ type Interface interface {
 
 	// Management API
 	WriteStatusRuntimeConfig(w io.Writer, r *http.Request) error
-	WriteTenantOverrides(w io.Writer, r *http.Request, tenant string) error
+
+	TenantsHandler(w http.ResponseWriter, req *http.Request)
```
mdisibio, line 75: I'm wondering if these should not be in `Interface`.  It seems like they are an aspect of the hosting in `/cmd/tempo/` and not part of the overrides modules directly.  Let's chat offline to help me get up to speed on the interaction between overrides.

kvrhdn: My reply earlier

> Yeah it’s an unfortunate consequence from hiding the runtime and user-config overrides behind an interface. The only public API exposed in the overrides package is `Interface`. When you setup the overrides, we return either a `runtimeConfigOverridesManager` or a `userConfigurableOverridesManager` based upon your configuration.
> 
> The reason there are two handlers for each page is we have a version with runtime overrides only and one with runtime+user-config overrides.
> We could also try to extend `Interface` to pass the config but that would be difficult to do generically I think.
> 
> By moving the status page into the overrides module we can access internal data more easily and it kind of makes sense that the module and its management page are tied closer together.

I actually just thought of an implementation that doesn't use two implementations, let me try it out.

knylander-grafana (commented):
Doc update looks good.

modules/overrides/overrides_tenants_http.go
```diff
@@ -0,0 +1,83 @@
...
+			}
+		}
+
+		// user-configurable overrides
+		userConfigurableOverridesManager, ok := o.(*userConfigurableOverridesManager)
+		if ok {
+			for _, tenant := range userConfigurableOverridesManager.GetTenantIDs() {
+				_, hasRuntimeOverrides := tenants[tenant]
+
+				tenants[tenant] = tenantsPageTenant{
```
mdisibio, line 57: nit: Can this be simplified?  Either by using `*tenantsPageTenant` or making use of how `tenants[tenant]` will return an uninitialized struct (with HasRunTimeOverrides=false)

modules/overrides/overrides_tenants_http.go
```diff
@@ -0,0 +1,83 @@
...
+	"html/template"
+	"net/http"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/grafana/tempo/pkg/util"
+)
+
+//go:embed tenants.gohtml
```
mdisibio, line 14: This is awesome 👍

modules/overrides/tenant_status.gohtml
```diff
@@ -0,0 +1,17 @@
+{{- /*gotype: github.com/grafana/mimir/pkg/storegateway.tenantsPageContents*/ -}}
```
mdisibio, line 1: Do we need to update this from mimir to tempo? Same q in both files.

modules/overrides/user_configurable_overrides.go
```diff
@@ -195,6 +198,14 @@ func (o *userConfigurableOverridesManager) setTenantLimit(userID string, limits
 	}
 }
 
+func (o *userConfigurableOverridesManager) GetTenantIDs() []string {
+	var ids []string
+	for tenant := range o.getAllTenantLimits() {
```
mdisibio, line 203: nit: Could use `maps.Keys(o.getAllTenantLimits)` (same for both implementations).

mdisibio (approved):
Tested most of this locally and LGTM.  One thing I noticed is that when browsing to a non-existent tenant like `/status/overrides/doesnotexist` it prints the inherited settings from tenant `*` as expected - but is there a way to indicate that it's using the defaults and not configured specifically?   Not a blocker, just curious.   Also left a few nits for things that I think would improve readability.  Non-blocking.

modules/overrides/overrides_tenants_http.go
```diff
@@ -0,0 +1,83 @@
...
+
+				tenants[tenant] = tenantsPageTenant{
+					Name:                         tenant,
+					HasRuntimeOverrides:          hasRuntimeOverrides,
+					HasUserConfigurableOverrides: true,
+				}
+			}
+		}
+
+		var tenantsList []tenantsPageTenant
```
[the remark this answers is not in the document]
kvrhdn: These are neat, didn't know they existed!

kvrhdn:
> Tested most of this locally and LGTM. One thing I noticed is that when browsing to a non-existent tenant like `/status/overrides/doesnotexist` it prints the inherited settings from tenant `*` as expected - but is there a way to indicate that it's using the defaults and not configured specifically? Not a blocker, just curious. Also left a few nits for things that I think would improve readability. Non-blocking.

I gave this a shot and it gets a bit complicated because a tenant can be using its own runtime overrides, the wildcard (`*`) or the default overrides.

https://github.com/grafana/tempo/blob/21fde18eec19df2edb41a4df5bd400acb81aa2d3/modules/overrides/runtime_config_overrides.go#L484-L498

I added a text line that tells you where it found the overrides, I hope this is clear enough.

If a tenant has its own runtime overrides:
> Screenshot 2024-02-01 at 16 06 29

If a tenant does not have its own runtime overrides, but we have `*` overrides:
> Screenshot 2024-02-01 at 16 06 57

If neither is set and we fall back to the `overrides.defaults` configuration:
> Screenshot 2024-02-01 at 16 08 17

PR #3339, opened 2024-01-26
Fix autocomplete returning erroneous values

joe-elliott:
**What this PR does**:
In certain circumstances filtered autocomplete would return incorrect results due to an incorrect iterator scope. The actual fix is [this line](https://github.com/grafana/tempo/pull/3339/files#diff-75a709e18d228f16dc762d46a5b3d80e86ac0c8c632d12d6618f72052f26c58aR503)

- removed an unused param in `createDistinctAttributeOperator`
- added integration tests for filtered autocomplete

**Checklist**
- [x] Tests updated
- [ ] Documentation added


mapno (approved):
LGTM. Nice catch

PR #3340, opened 2024-01-26
Return Bad Request when exceeding limit

joe-elliott:
**What this PR does**:
Currently if the configured max limit is exceeded Tempo will just cut off results at the max limit. Instead we should return a 400 to clearly notify the user their request is invalid.

**Checklist**
- [x] Tests updated
- [x] Documentation added


zalegrala (approved):
This looks reasonable to me.

PR #3350, opened 2024-01-30
Add configuration for fetch services older than complete_block_timeout

rubenvp8510:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/cmd/tempo-query/tempo/plugin.go
+++ b/cmd/tempo-query/tempo/plugin.go
@@ -53,11 +53,12 @@ var tlsVersions = map[string]uint16{
 }
 
 type Backend struct {
-	tempoBackend    string
-	tlsEnabled      bool
-	tls             tlsCfg.ClientConfig
-	httpClient      *http.Client
-	tenantHeaderKey string
+	tempoBackend          string
+	tlsEnabled            bool
+	tls                   tlsCfg.ClientConfig
+	httpClient            *http.Client
+	tenantHeaderKey       string
+	QueryServicesDuration *time.Duration
 }
 
 func New(cfg *Config) (*Backend, error) {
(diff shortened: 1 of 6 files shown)


rubenvp8510:
This is ready for review.

rubenvp8510:
Hello @joe-elliott Do you know who can review this? Thank you very much

cmd/tempo-query/tempo/plugin.go
```diff
@@ -53,24 +53,38 @@ var tlsVersions = map[string]uint16{
...
-	tlsEnabled      bool
-	tls             tlsCfg.ClientConfig
-	httpClient      *http.Client
-	tenantHeaderKey string
+	tempoBackend          string
+	tlsEnabled            bool
+	tls                   tlsCfg.ClientConfig
+	httpClient            *http.Client
+	tenantHeaderKey       string
+	QueryServicesDuration *time.Duration
```
joe-elliott, line 61: since this is all tag values and not just services i'd adjust the name to reflect that.

cmd/tempo-query/tempo/plugin.go
```diff
@@ -379,7 +399,14 @@ func createTagsQueryParam(service string, operation string, tags map[string]stri
...
 
 func (b *Backend) lookupTagValues(ctx context.Context, span opentracing.Span, tagName string) ([]string, error) {
-	url := fmt.Sprintf("%s://%s/api/search/tag/%s/values", b.apiSchema(), b.tempoBackend, tagName)
+	var url string
+
+	if b.QueryServicesDuration == nil {
+		url = fmt.Sprintf("%s://%s/api/search/tag/%s/values", b.apiSchema(), b.tempoBackend, tagName)
+	} else {
+		startTime, endTime := b.calculateTimeRange()
+		url = fmt.Sprintf("%s://%s/api/search/tag/%s/values?start=%d&end=%d", b.apiSchema(), b.tempoBackend, tagName, startTime, endTime)
```
joe-elliott, line 408: should this be controlled with a config value? does jaeger not pass similar params when it searches its own backend?

rubenvp8510: It pass the parameters only when you search for traces, but for do an initial search we need the service name.

cmd/tempo-query/tempo/plugin.go
rubenvp8510: well, not really, this is only used for fetch service list.

Usually that list only contains services that are in between the 15 minutes (or the defined complete_block_timeout) but now that we can get it from the backend we can extend that time.

joe-elliott: i see 👍

integration/e2e/query_plugin_test.go
```diff
@@ -73,6 +73,46 @@ func TestSearchUsingJaegerPlugin(t *testing.T) {
 	callJaegerQuerySearchTraceAssert(t, tempoQuery, "execute", "backend")
 }
 
+func TestSearchUsingBackendTagsService(t *testing.T) {
```
joe-elliott, line 76: nice addition!

joe-elliott (approved):

PR #3353, opened 2024-01-31
Update traceql metrics to use the trace-level timestamp columns

mdisibio:
**What this PR does**:
This PR circles back to a metrics TODO and uses the trace-level timestamp columns to more quickly filter the relevant traces in a block.   The columns are not free and hinder performance unless they are able to filter out enough values.   So it is decided at runtime based on the request and the block in question.

For now, did some local benchmarking, and using the simple heuristic that if we are looking at less than 20% of the block, then it's worth pulling in the columns.  Between 20 and 100%, we avoid the extra I/O but spend more cycles throwing away traces (this is the current behavior).   Open to suggestions on a better heuristic.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


mapno (approved):
Very cool optimisation. LGTM

PR #3358, opened 2024-02-02
Ingester readonly on startup until replay and rediscover is done to prevent broken head blocks

### LINKED ISSUE #3346: Ingesters can start in broken state and cannot write traces to disk for one or more tenants
**Describe the bug**
After a crash or unclean restart (OOM or readinessprobe failure) an ingester can enter a bad state where it is unable to write traces to the headblock.  It continually logs the following error and is unable to recover on its own.  The error means that all incoming data for the tenant is lost because it cannot be persisted to disk.

```
msg="failed to cut traces" err="error writing meta json: open /var/tempo/wal/<guid>+<tenant>+vParquet3/meta.json: no such file or directory"
```

RF=3 helps mitigate this issue because it prevents data loss from 1 failing ingester.  However not if there are 2 or more.

**To Reproduce**
1. Start Tempo `main`
2. Ingesters get killed for various reasons: OOMs or readiness probe failures.  
3. Large numbers of tenants, and pressure on the read path such as search and autocomplete lookups increase the likelihood of this.
4. Eventually an ingester will start in this state for 1 or more tenants.

**Expected behavior**
Ingesters are able to recover from all crashes and unclean restarts.  Existing data in memory or partially written blocks might be lost, but the ingester should always be able to receive new data.

**Environment:**
 - Infrastructure: Kubernetes
 - Deployment tool: jsonnet

**Additional Context**

mdisibio:
**What this PR does**:
This is one of those bugs that makes you wonder "how did this ever work?"   This PR changes the ingester to enter a read-only state on startup (similar to shutdown) until wal replay and local blocks rediscover are complete.  This prevents pushes from creating headblocks that get tangled up in the replay process (and broken).  I really like having the startup error different than the shutting down error.  It makes it easy to check logs for `Ingester is starting` and verify the fix is working.

**Description of the Bug**
Here is a full sequence of the bug:
1. An Ingester unexpectedly terminates (OOM, panic, readinessprobe failure etc)
2. The ingester restarts and begins WAL replay
3. Because it didn't leave the ring gracefully and propagating the new state takes some time, it still appears HEALTHY to (some or all) the distributors.
4. The distributors push traffic to it while it is replaying
5. PushBytes works (unexpectedly) and creates a headblock     <--- This is where we fix it
6. The headblock gets picked up by the replay process and deleted
7. All pushes to that headblock fail and the ingester never recovers

**Steps to Reproduce**
1. Requires a distributed setup with separate distributors, ingesters, and replication factor >= 2
2. While continuously pushing traffic, quickly kill and restart a single ingester via `docker kill`/etc
3. Only kill 1 ingester, so that the distributor sees the minimum replicas and keeps pushing traffic
4. Eventually it will trigger the bug

Fixes #3346 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * [BUGFIX] The tempo-cli analyse blocks command no longer fails on compacted blocks [#3183](https://github.com/grafana/tempo/pull/3183) (@stoewer)
 * [BUGFIX] Move waitgroup handling for poller error condition [#3224](https://github.com/grafana/tempo/pull/3224) (@zalegrala)
 * [BUGFIX] Fix head block excessive locking in ingester search [#3328](https://github.com/grafana/tempo/pull/3328) (@mdisibio)
+* [BUGFIX] Fix issue with ingester failed to cut traces no such file or directory [#3346](https://github.com/grafana/tempo/issues/3346) (@mdisibio)
 * [ENHANCEMENT] Introduced `AttributePolicyMatch` & `IntrinsicPolicyMatch` structures to match span attributes based on strongly typed values & precompiled regexp [#3025](https://github.com/grafana/tempo/pull/3025) (@andriusluk)
 * [CHANGE] TraceQL/Structural operators performance improvement. [#3088](https://github.com/grafana/tempo/pull/3088) (@joe-elliott)
 * [CHANGE] Merge the processors overrides set through runtime overrides and user-configurable overrides [#3125](https://github.com/grafana/tempo/pull/3125) (@kvrhdn)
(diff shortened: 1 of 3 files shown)


mdisibio:
> is there anyway to just not raise the healthy flag in the ring until we've done wal replay?

It already doesn't raise the healthy flag, but the issue is the non-zero propagation time during which the distributors continue sending traffic.  We should expect to see a short burst of "Ingester is starting" errors until the distributors catch up.  But I can dig more in this area and double check things.

joe-elliott:
> It already doesn't raise the healthy flag, but the issue is the non-zero propagation time during which the distributors continue sending traffic. 

Oh, I see. It's explained in your original steps. In a normal rollout this shouldn't even occur.

joe-elliott (approved):

PR #3368, opened 2024-02-06
Add vparquet4 encoding

stoewer:
**What this PR does**:
This adds a new encoding vParquet4 to Tempo (the default encoding remains unchanged). The new encoding improves support for links, events, and attributes with array values. This is a prerequisite better support of links, events, and arrays in TraceQL as proposed by the [TraceQL Extensions design proposal](https://github.com/grafana/tempo/blob/main/docs/design-proposals/2023-11%20TraceQL%20Extensions.md).

The vParquet4 encoding is considered experimental for now.

There are still some things missing from this PR, those will be subject to separate PRs:
* A CLI to analyze blocks in order to find good candidates for dedicated attributes
* Documentation about how to enable vParquet4

*Hints for reviewers*: This PR was created from the commits to the `vparquet3` directory in the branch `main-vparquet4` ([diff](https://github.com/grafana/tempo/compare/main...main-vparquet4)). Since this PR adds a whole new directory `vparquet4`, it can be helpful to review the individual PRs in order to see the differences between vParquet3 and vParquet4.

Fixes grafana/tempo-squad#384

**Checklist**
- [x] Tests updated
- [ ] Documentation added


ie-pham:
Screenshot 2024-02-07 at 2 24 35 PM
Screenshot 2024-02-07 at 2 24 58 PM
Ingester and querier memory usage is up

.golangci.yml
```diff
@@ -63,6 +63,8 @@ linters-settings:
     exclude: ./.errcheck-exclude.txt
   errorlint:
     errorf: false # don't check whether fmt.Errorf uses %w to wrap errors
+  goconst:
+    ignore-tests: true
```
joe-elliott, line 67: i think this is fine? it just turns of `goconst` for tests? i've had the linter find test issues before like failing to check an err return

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## main / unreleased
 
 * [ENHANCEMENT] Add tracing integration to profiling endpoints [#3276](https://github.com/grafana/tempo/pull/3276) (@cyriltovena)
+* [FEATURE] New block encoding vParquet4 with support for links, events, and arrays [#3368](https://github.com/grafana/tempo/pull/3368) (@stoewer @ie-pham @andreasgerstmayr)
```
joe-elliott, line 4: 🎉

modules/frontend/search_progress.go
```diff
@@ -159,6 +159,17 @@ func (r *searchProgress) result() *shardedSearchResults {
 			SpanSet:           copySpanset(m.SpanSet),
 		}
 
+		// copy service stats
+		if len(m.ServiceStats) > 0 {
```
joe-elliott, line 163: i do, admittedly, wish we had separated the parquet changes from the search pipeline changes, but i don't think it's worth reverting at this point.

cc @grafana/observability-traces-and-profiling search results will be enriched with services, span and error counts when we move to vparquet4. nothing actionable now, but something we can add in the future.

pkg/traceql/combine.go
```diff
@@ -70,6 +70,17 @@ func combineSearchResults(existing *tempopb.TraceSearchMetadata, incoming *tempo
 		existing.DurationMs = incoming.DurationMs
 	}
 
+	// Merge service stats
+	for service, incomingStats := range incoming.ServiceStats {
+		existingStats, ok := existing.ServiceStats[service]
```
joe-elliott, line 75: i don't think this merge works. if we find multiple trace fragments that satisfy a traceql result we will be repeatedly re-adding the service span/error counts.

should it be max?

pkg/traceql/engine.go
```diff
@@ -231,9 +231,17 @@ func (e *Engine) asTraceSearchMetadata(spanset *Spanset) *tempopb.TraceSearchMet
 		RootTraceName:     spanset.RootSpanName,
 		StartTimeUnixNano: spanset.StartTimeUnixNanos,
 		DurationMs:        uint32(spanset.DurationNanos / 1_000_000),
+		ServiceStats:      make(map[string]*tempopb.ServiceStats, len(spanset.ServiceStats)),
 		SpanSet:           &tempopb.SpanSet{},
 	}
 
+	for service, stats := range spanset.ServiceStats {
```
joe-elliott, line 238: would like to see some of the existing tests extended to cover this and other service stats functionality. this area is pretty well tested and it would be nice to catch this logic as well.

tempodb/encoding/versioned.go
```diff
@@ -78,7 +81,7 @@ func DefaultEncoding() VersionedEncoding {
 
 // LatestEncoding returns the most recent encoding.
 func LatestEncoding() VersionedEncoding {
-	return vparquet3.Encoding{}
+	return vparquet4.Encoding{}
```
joe-elliott, line 84: i'm concerned about this. if we merge vp4 primarily to prevent it from rotting and then don't roll it out for awhile this change will prevent our ingester tests from running vp3. wdyt?

tempodb/encoding/vparquet4/schema.go
```diff
@@ -0,0 +1,963 @@
...
+			}
+
+			protoEvents = append(protoEvents, protoEvent)
+		}
+	}
+
+	return protoEvents
+}
+
+func parquetTraceToTempopbTrace(meta *backend.BlockMeta, parquetTrace *Trace, includeDroppedAttr bool) *tempopb.Trace {
```
joe-elliott, line 804: it's not clear to me why `includeDroppedAttr` is a boolean. would we not want to always return the count of dropped attrs?

tempodb/encoding/vparquet4/block_traceql.go
```diff
@@ -0,0 +1,2311 @@
...
+			traceIters = append(traceIters, makeIter(columnPathRootServiceName, pred, columnPathRootServiceName))
+		}
+	}
+
+	// order is interesting here. would it be more efficient to grab the span/resource conditions first
+	// or the time range filtering first?
+	traceIters = append(traceIters, resourceIter)
+
+	// collect service stats of the trace
+	traceIters = append(traceIters, createServiceStatsIterator(makeIter))
```
joe-elliott, line 1505: we should not add this iterator unless the fetch layer requests it. otherwise we will unnecessarily load this data for every trace before conditions are asserted in the engine

tempodb/encoding/vparquet4/combiner.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	if c.result.RootSpanName == "" {
+		c.result.RootSpanName = tr.RootSpanName
+	}
+	c.result.DurationNano = c.result.EndTimeUnixNano - c.result.StartTimeUnixNano
+
+	// Merge service stats
+	for service, incomingStats := range tr.ServiceStats {
+		combinedStats := c.result.ServiceStats[service]
+		combinedStats.SpanCount += incomingStats.SpanCount
```
joe-elliott, line 93: i don't think we can just sum here. i think we have to rescan the trace and recalculate similar to nested set.

tempodb/encoding/vparquet4/schema.go
```diff
@@ -0,0 +1,963 @@
...
+
+type Attribute struct {
+	Key string `parquet:",snappy,dict"`
+
+	ValueType    AttrType  `parquet:",snappy,delta"`
+	Value        []string  `parquet:",snappy,dict,list"`
+	ValueInt     []int64   `parquet:",snappy,list"`
+	ValueDouble  []float64 `parquet:",snappy,list"`
+	ValueBool    []bool    `parquet:",snappy,list"`
+	ValueDropped string    `parquet:",snappy,optional"`
```
joe-elliott, line 139: quick comment on how this is encoded? json?

tempodb/encoding/vparquet4/schema.go
```diff
@@ -0,0 +1,963 @@
...
+func (e *Event) addDroppedAttr(n int32) {
+	e.DroppedAttributesCount += n
+}
+
+type Link struct {
+	TraceID                []byte      `parquet:","`
+	SpanID                 []byte      `parquet:","`
+	TraceState             string      `parquet:",snappy"`
+	Attrs                  []Attribute `parquet:",list"`
+	DroppedAttributesCount int32       `parquet:",snappy,delta"`
```
joe-elliott, line 172: i'm confused by this dropped attributes tracking. if we store it and return it faithfully on request by trace id then i don't think we need to track dropped anything.

these attrs would be invisible to traceql but we can just clearly document what we support

joe-elliott (commented):
Adding comments, but I'm not 100% done. I'm quite close and only have the second half of schema.go left

.golangci.yml
stoewer: Yes, this only disables `goconst` in test files ([docs](https://golangci-lint.run/usage/linters/#goconst)). Without this setting `goconst` finds a lot of test parameters  that appear often in tests like "resource." or "span."

tempodb/encoding/vparquet4/schema.go
stoewer: This is not quite what `includeDroppedAttr` does
* If `includeDroppedAttr` is `true` the converted `tempopb.Trace` will contain all attributes that are dropped by Tempo as if they weren't dropped and will also not show the incremented dropped attribute count
* If it is `false` dropped attributes will not appear in the converted trace and the dropped attribute count will be incremented by one for every dropped attribute. On most places this function is called with `false` i.e. from the outside dropped attributes will appear as dropped: They are not searchable with TraceQL, can't be selected, and do not appear in traces found via trace by ID

tempodb/encoding/vparquet4/schema.go
stoewer: Yes this this JSON encoded. The same encoding was used for the columns `ValueKVList` and `ValueArray` in vParquet3

tempodb/encoding/vparquet4/schema.go
stoewer: * Unsupported attributes values that are stored in the `ValueDropped` columns are not returned by the API and appear as dropped. Therefore the incrementing the dropped attribute count for those values is appropriate
* `DroppedAttributesCount` may also be > 0 when attributes are removed by other parts of the tracing pipeline

pkg/traceql/combine.go
stoewer: Good point, I think max is more correct

tempodb/encoding/versioned.go
joe-elliott: after some thought, i'm good on being more aggressive here. let's keep this change to update `LatestEncoding()`

tempodb/encoding/vparquet4/schema.go
joe-elliott: going to mark this resolve and consolidate dropped attributes discussion above.

tempodb/encoding/vparquet4/schema.go
joe-elliott: let's back up and discuss dropped attributes handling. in my opinion we should simplify this logic by:

- removing the `includeDroppedAttr` boolean and always rendering dropped attributes. we can just return the json string.
- removing the logic that internally increments/decrements dropped attributes and just faithfully storing whatever the trace pipeline sends us.

wdyt?

joe-elliott (commented):
I've made a full pass of the PR now and all concerns are currently in unresolved conversations.

tempodb/encoding/vparquet4/schema.go
stoewer: Sounds good. I added a commit to change this behavior. Attributes with unsupported values are no longer treated as dropped values:
* Changed the column name from `ValueDropped` to `ValueUnsupported`
* Removed the logic to increment `DroppedAttributesCount`
* Removed the parameter `includeDroppedAttr` and the associated logic.

pkg/traceql/combine.go
andreasgerstmayr: If you both agree on using `max()`, I've created a PR targeting the branch of this PR: https://github.com/stoewer/tempo/pull/1

pkg/traceql/engine.go
andreasgerstmayr: Added ServiceStats in the existing TraceQL engine tests: https://github.com/stoewer/tempo/pull/2
Which other areas should be tested? We've coverage of the main logic in `vparquet4/schema.go` and the serviceStatsIters are tested in `block_traceql_meta_test.go`.

tempodb/encoding/vparquet4/block_traceql.go
andreasgerstmayr: I'm happy to work on this, could you give me some tips where to move this iterator?
I noticed it should probably be moved after the time range filtering (not sure why I added it before that), but I guess you've more optimization in mind.

tempodb/encoding/vparquet4/combiner.go
andreasgerstmayr: Should we apply the same logic as in https://github.com/stoewer/tempo/blob/add-vparquet4-encoding/tempodb/encoding/vparquet4/schema.go#L560-L575 with the Trace in `c.result` (and moving this to the end of the function, after the loop where the spans are copied)?

tempodb/encoding/vparquet4/block_traceql.go
joe-elliott: the engine builds conditions that it passes into the fetch layer to tell communicate its needs. we currently perform two passes on the data. the first pass asserts any conditions in the query and the second pass requests relevant metadata for the matching rows.

This is the metadata it requests on the second pass.
https://github.com/grafana/tempo/blob/4f53eb9eda1672699898f8ed1d76aac51b530de6/pkg/traceql/storage.go#L15

Currently each condition is requesting a specific column that are span intrinsics. It's likely we don't want to make a new intrinsic here b/c we don't intend for these fields to be refernceable in TraceQL. So we should probably just have a new condition that specifically requests "service metadata".

Then we would only build/add this iterator if that condition were requested.

tempodb/encoding/vparquet4/combiner.go
joe-elliott: yes. notice the function `assignNestedSetModelBounds` it's called in `traceToParquet` (when we first create a trace) and in the combiner.

stoewer: It could also be a good idea to rename the function to `assignNestedSetModelBoundsAndServiceStats` and move the logic there. The downside is, that it adds complexity to an already complex function. But it also has two advantages:
* It makes sure that service stats are always (re-) calculated when needed
* `assignNestedSetModelBounds` already iterates through the whole trace. Moving the logic there, eliminates a second traversal of the trace

@joe-elliott what do you think?

stoewer:
@andreasgerstmayr thanks for submitting the PRs

tempodb/encoding/vparquet4/combiner.go
joe-elliott: > @joe-elliott what do you think?

+1 do this. i could see more "recalculate after combining" summary columns being added. co-locating them in one spot for performance feels good to me.

it is a relatively easy to function to test. some clear tests should help with the complexity concerns.

andreasgerstmayr: :+1:

I moved the logic to `assignNestedSetModelBoundsAndServiceStats()` and added a few new testcases: https://github.com/stoewer/tempo/pull/3

tempodb/encoding/vparquet4/block_traceql.go
andreasgerstmayr: Thanks for the explanation!
I moved the iterator to the second pass: https://github.com/stoewer/tempo/pull/4

tempodb/encoding/vparquet4/schema.go
```diff
@@ -0,0 +1,920 @@
...
+func instrumentationScopeToParquet(s *v1.InstrumentationScope, ss *InstrumentationScope) {
+	if s == nil {
+		ss.Name = ""
+		ss.Version = ""
+	} else {
+		ss.Name = s.Name
+		ss.Version = s.Version
+	}
+
+	// TODO: handle attributes correctly once they are added to the proto
```
joe-elliott, line 572: these have been added to the proto. i don't think we should handle that in this PR, but perhaps make an issue?

https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/common/v1/common.proto#L79-L80

pkg/traceql/enum_attributes.go
```diff
@@ -9,6 +9,7 @@ const (
 	AttributeScopeResource
 	AttributeScopeSpan
 	AttributeScopeUnknown
+	AttributeScopeTrace
```
joe-elliott, line 12: addressed elsewhere but trace attributes don't make sense to me. i feel like all trace level fields are intrinsic

joe-elliott (commented):
This was well reviewed a month or so back. A few questions and thoughts but this is quite close.

pkg/traceql/storage.go
```diff
@@ -22,6 +26,7 @@ func SearchMetaConditions() []Condition {
 		{NewIntrinsic(IntrinsicSpanID), OpNone, nil},
 		{NewIntrinsic(IntrinsicSpanStartTime), OpNone, nil},
 		{NewIntrinsic(IntrinsicDuration), OpNone, nil},
+		{Attribute{Scope: AttributeScopeTrace, Name: ServiceStatsAttributeName}, OpNone, nil},
```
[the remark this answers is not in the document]
andreasgerstmayr: Yes, makes sense. I think I misunderstood this comment from above
> Currently each condition is requesting a specific column that are span intrinsics. It's likely we don't want to make a new intrinsic here b/c we don't intend for these fields to be refernceable in TraceQL. So we should probably just have a new condition that specifically requests "service metadata".

tempodb/encoding/vparquet4/block_traceql.go
```diff
@@ -0,0 +1,2403 @@
...
+		}
+	}
+
+	// loop over all spans and add the trace-level attributes
+	for _, s := range finalSpanset.Spans {
+		s := s.(*span)
+		s.setTraceAttrs(c.traceAttrs)
+	}
+
+	finalSpanset.ServiceStats = make(map[string]traceql.ServiceStats, numServiceStats)
```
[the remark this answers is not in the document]
joe-elliott: i disagree with my thoughts here about making servicestats alloc/clear related to the pooling code. only a very small %age of spansets will need this field.

it does seem we are alloc'ing in two places though. perhaps we can get down to one.

pkg/traceql/storage.go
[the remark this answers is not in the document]
joe-elliott: Oof, that was several months ago. I apologize. I did not realize I was running you in circles. I'm fine with intrinsic. It's definitely better than an Attribute scope here. We also have other unqueryable intrinsics such as span id.

Sry friend! 🙏

tempodb/encoding/vparquet4/block_traceql.go
```diff
@@ -0,0 +1,2403 @@
...
+				sp.cbSpanset.RootSpanName = ss.RootSpanName
+			}
+			if len(sp.cbSpanset.RootServiceName) == 0 {
+				sp.cbSpanset.RootServiceName = ss.RootServiceName
+			}
+			if sp.cbSpanset.StartTimeUnixNanos == 0 {
+				sp.cbSpanset.StartTimeUnixNanos = ss.StartTimeUnixNanos
+			}
+			if len(sp.cbSpanset.ServiceStats) == 0 {
+				sp.cbSpanset.ServiceStats = map[string]traceql.ServiceStats{}
```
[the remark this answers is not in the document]
andreasgerstmayr: If the spanset is allocated in
https://github.com/grafana/tempo/blob/148ba6dbb980f27fff57073a534a44a698f3c02d/pkg/traceql/ast_execute.go#L36
 we need to allocate the `ServiceStats` map here, otherwise it'll get allocated in the keepgroup afaics.

Addressed in https://github.com/stoewer/tempo/pull/5

pkg/traceql/storage.go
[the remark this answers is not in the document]
andreasgerstmayr: np! the code is complex, so I appreciate that you're taking a thorough look at my changes :)

I made it an intrinsic in https://github.com/stoewer/tempo/pull/5.

pkg/traceql/enum_attributes.go
andreasgerstmayr: :+1: done

pkg/traceql/storage.go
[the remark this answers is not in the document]
joe-elliott: thank you!

tempodb/encoding/vparquet4/schema_test.go
```diff
@@ -0,0 +1,894 @@
...
+}
+
+func TestFieldsAreCleared(t *testing.T) {
+	meta := backend.BlockMeta{
+		DedicatedColumns: test.MakeDedicatedColumns(),
+	}
+
+	traceID := []byte{0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0E, 0x0F}
+	complexTrace := parquetTraceToTempopbTrace(&meta, fullyPopulatedTestTrace(traceID))
+	simpleTrace := &tempopb.Trace{
```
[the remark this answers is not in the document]
joe-elliott: i was wanting to confirm that the service stats was correctly wiped out when reusing the trace in `traceToParquet`. I can see that service stats were added to the `fullyPopulatedTestTrace()` which covers my concern.

thx!

joe-elliott:
@stoewer @andreasgerstmayr this is very close. thank you for your work. there are 3 threads left to resolve. two of which are covered by andreas's PR: https://github.com/stoewer/tempo/pull/5

if we push on this i think we can get it quite soon. let me know when you're ready for a final review 🙏

tempodb/encoding/vparquet4/schema.go
stoewer: I'm aware of this and have it on my post merge TODO list :+1:

tempodb/encoding/vparquet4/schema.go
```diff
@@ -0,0 +1,920 @@
...
+			attrToParquetTypeUnsupported(a, p)
+		}
+	default:
+		attrToParquetTypeUnsupported(a, p)
+	}
+}
+
+func attrToParquetTypeUnsupported(a *v1.KeyValue, p *Attribute) {
+	jsonBytes := &bytes.Buffer{}
+	_ = jsonMarshaler.Marshal(jsonBytes, a.Value) // deliberately marshalling a.Value because of AnyValue logic
```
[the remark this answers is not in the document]
joe-elliott: looks like we swapped span links and event attributes to proto but left others as json?

https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet3/schema.go#L558
https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet3/schema.go#L671

i'm fine with leaving these as json. likely they will be rare. we should keep an eye on profiles as we roll this out to see if there is significant impact on resource usage.

tempodb/encoding/vparquet4/block_traceql.go
```diff
@@ -0,0 +1,2406 @@
...
+			finalSpanset.Spans = append(finalSpanset.Spans, span)
+		}
+	}
+
+	// loop over all spans and add the trace-level attributes
+	for _, s := range finalSpanset.Spans {
+		s := s.(*span)
+		s.setTraceAttrs(c.traceAttrs)
+	}
+
```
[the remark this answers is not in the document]
stoewer: Makes sense: I added the change :+1:

joe-elliott (approved):

PR #3377, opened 2024-02-08
deprecating vparquet v1

ie-pham:
**What this PR does**: Deprecating vParquet v1. vParquet v1 blocks can still be read but write is no longer allowed. 

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/configuration/parquet.md
+++ b/docs/sources/tempo/configuration/parquet.md
@@ -32,14 +32,14 @@ You can still use the previous format `vParquet2`.
 To enable it, set the block version option to `vParquet2` in the Storage section of the configuration file.
 
 ```yaml
-# block format version. options: v2, vParquet, vParquet2, vParquet3
+# block format version. options: v2, vParquet2, vParquet3
 [version: vParquet2]
 ```
 
 In some cases, you may choose to disable Parquet and use the old `v2` block format. Using the `v2` block format disables all forms of search, but also reduces resource consumption, and may be desired for a high-throughput cluster that does not need these capabilities. To make this change, set the block version option to `v2` in the Storage section of the configuration file.
 
 ```yaml
-# block format version. options: v2, vParquet, vParquet2, vParquet3
+# block format version. options: v2, vParquet2, vParquet3
 [version: v2]
 ```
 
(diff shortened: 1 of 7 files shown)


docs/sources/tempo/configuration/parquet.md
```diff
@@ -32,14 +32,14 @@ You can still use the previous format `vParquet2`.
 To enable it, set the block version option to `vParquet2` in the Storage section of the configuration file.
```
joe-elliott, line 32: this is weird advice. we should likely not be advising anyone to set their version to vParquet2 at this point.

tempodb/encoding/common/config.go
```diff
@@ -65,5 +65,9 @@ func ValidateConfig(b *BlockConfig) error {
 		return fmt.Errorf("positive value required for bloom-filter shard size")
 	}
 
+	if b.Version == "vParquet" {
```
joe-elliott, line 68: can be replaced with this constant:

https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet/encoding.go#L14

tempodb/encoding/versioned.go
```diff
@@ -85,7 +85,7 @@ func LatestEncoding() VersionedEncoding {
 func AllEncodings() []VersionedEncoding {
 	return []VersionedEncoding{
 		v2.Encoding{},
-		vparquet.Encoding{},
+		//	vparquet.Encoding{},
```
joe-elliott, line 88: this will help with tests. good call.

joe-elliott (approved):

tempodb/encoding/common/config.go
joe-elliott: ah, i'm good with this then 👍

docs/sources/tempo/configuration/parquet.md
ie-pham: Well, the full context was. Default = vParquet3. But "You can still use the previous format `vParquet2`."

tempodb/encoding/versioned.go
electron0zero: can we remove it instead of commenting it out?

electron0zero (approved):
one nit, rest lgtm. 

let's goooo 🚀

PR #3387, opened 2024-02-13
Fix trace context propagation in the read path

mapno:
**What this PR does**:

Fixes trace context propagation between query-frontend and querier. The context was not being correctly passed and unpacked between both.

It also adds tracing to the distributed docker-compose example.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


mdisibio (approved):

joe-elliott (commented):
nice :)

PR #3395, opened 2024-02-14
Add shutdown delay and e2e test

joe-elliott:
**What this PR does**:
Adds a shutdown delay configurable parameter modeled after mimir:

https://github.com/grafana/mimir/pull/3298

After receiving SIGTERM Tempo will drop its readiness handler and wait the configured seconds. In a k8s environment this will allow Tempo to cleanly roll its frontend without dropping queries.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo/app/app.go
+++ b/cmd/tempo/app/app.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net/http"
 	"sort"
+	"time"
 
 	"github.com/go-kit/log/level"
 	"github.com/gorilla/mux"
@@ -23,6 +24,7 @@ import (
 	"github.com/jedib0t/go-pretty/v6/table"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/version"
+	"go.uber.org/atomic"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/health/grpc_health_v1"
 	"gopkg.in/yaml.v3"
@@ -42,6 +44,7 @@ import (
 	"github.com/grafana/tempo/pkg/usagestats"
 	"github.com/grafana/tempo/pkg/util"
 	"github.com/grafana/tempo/pkg/util/log"
+	util_log "github.com/grafana/tempo/pkg/util/log"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


cmd/tempo/app/config.go
```diff
@@ -76,6 +77,7 @@ func (c *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet) {
 	f.BoolVar(&c.UseOTelTracer, "use-otel-tracer", false, "Set to true to replace the OpenTracing tracer with the OpenTelemetry tracer")
 	f.BoolVar(&c.EnableGoRuntimeMetrics, "enable-go-runtime-metrics", false, "Set to true to enable all Go runtime metrics")
 	f.BoolVar(&c.AutocompleteFilteringEnabled, "autocomplete-filtering.enabled", true, "Set to false to disable autocomplete filtering")
+	f.DurationVar(&c.ShutdownDelay, "shutdown-delay", 0, "How long to wait between SIGTERM and shutdown. After receiving SIGTERM, Tempo will report not-ready status via /ready endpoint.")
```
electron0zero, line 80: Q: do we want to set a non zero default here?? so folks have a clean shutdown by default?

electron0zero (approved):
one small Q, rest lgtm 👍🏼

integration/e2e/e2e_test.go
```diff
@@ -312,6 +312,39 @@ func TestMicroservicesWithKVStores(t *testing.T) {
...
+	go func() {
+		_ = tempo.Stop()
+	}()
+
+	time.Sleep(500 * time.Millisecond)
+
+	// confirm the readiness flag is down
+	res, err := e2e.DoGet("http://" + tempo.Endpoint(3200) + "/ready")
+	require.NoError(t, err)
+	require.Equal(t, http.StatusServiceUnavailable, res.StatusCode)
```
zalegrala, line 344: Perhaps perform a query before and after to confirm the change in behavior.

cmd/tempo/app/app.go
```diff
@@ -231,6 +240,14 @@ func (t *App) Run() error {
 	handler := signals.NewHandler(t.Server.Log())
 	go func() {
 		handler.Loop()
+
+		shutdownRequested.Store(true)
+		t.Server.SetKeepAlivesEnabled(false)
+
+		if t.cfg.ShutdownDelay > 0 {
+			time.Sleep(t.cfg.ShutdownDelay)
```
zalegrala, line 248: Do we also want to sleep when there are no requests in flight?

cmd/tempo/app/config.go
joe-elliott: i prefer to 0 for a few reasons:
- consistency: mimir defaults to 0
- just setting this param alone doesn't guarantee a clean shutdown.  it requires coordination with your orchestration platform
- a default of 0 means there's no change in behavior

integration/e2e/e2e_test.go
joe-elliott: it's not apparent, but this line waits until `/ready` returns 200:

https://github.com/grafana/tempo/pull/3395/files#diff-b12b20a2065695807b91723666ebb99faca53ac562a4156ad198e724013661adR331

cmd/tempo/app/app.go
joe-elliott: i'm concerned about the complexity about that change. this is a very simple one that will do nearly all of what we want with some config choices in k8s

integration/e2e/e2e_test.go
joe-elliott: added a comment to help clarify that the /ready point is confirmed to be up before we stop tempo

cmd/tempo/app/config.go
electron0zero: makes sense.

zalegrala (approved):
Its a step in the right direction.

Issue #3366, opened 2024-02-05
Add streaming support to all read path endpoints


mapno:
Since its introduction, [streaming](https://grafana.com/docs/tempo/latest/traceql/query-editor/traceql-editor/#streaming-results) in [`/search`](https://grafana.com/docs/tempo/latest/api_docs/#search) requests has been a major driver of usability of Tempo and Grafana. Streaming is accomplished by a combination of periodically computing the partial results returned by queriers and gRPC streams.

All other read path endpoints could benefit from streaming as well:
* [ID lookups](https://grafana.com/docs/tempo/latest/api_docs/#query)
* [Tags](https://grafana.com/docs/tempo/latest/api_docs/#search-tags) ([v2](https://grafana.com/docs/tempo/latest/api_docs/#search-tags-v2)) and [tag values](https://grafana.com/docs/tempo/latest/api_docs/#search-tag-values) ([v2](https://grafana.com/docs/tempo/latest/api_docs/#search-tag-values-v2)). Tag values v2 in particular, which is the (at the momento) only endpoint that supports autocomplete filtering, which can be slow with certain queries.

The query-frontend streaming code is comprised of two parts: [a gRPC streaming handler](https://github.com/grafana/tempo/blob/main/modules/frontend/search_streaming.go) and [a combiner](https://github.com/grafana/tempo/blob/main/modules/frontend/search_progress.go) that can compute diffs from incoming results. This lays a good foundation for building generic code to handle all read path endpoints.

PR #3400, opened 2024-02-15
Frontend Refactor

### LINKED ISSUE #1392: Query limits don't work or work unexpectedly

joe-elliott:
**What this PR does**:
Refactors the frontend pipeline to accomplish a number of goals detailed below. Currently only the search pipelines have been moved over to the new structure. There will be follow up PRs for other endpoints.

Goals:
- Provide streaming on all query endpoints
- Consolidate recombination logic in the frontend/combiner package
- Consolidate all pipeline building logic in the frontend/pipeline package

See the [readme in the pipeline package](https://github.com/grafana/tempo/pull/3400/files#diff-3c5207b6a5e79bfb1b84334ee056b7cd262e6017abc2a316692f2778c84e6662) for details.

Fixes #1392

This PR is progress toward https://github.com/grafana/tempo/issues/3366 with an eventual goal to close it.

**Changes**
- removes `tempo_query_frontend_multitenant`... metrics b/c they don't quite fit into the current pattern. this can be improved later. 

**todo**:
- [x] Flesh out readme.md in /pipeline to describe how to build new pipeline elements

**Checklist**
- [x] Tests updated
- [ ] Documentation added


joe-elliott:
> @joe-elliott Was there doc added? I didn't see any in the file list to review.

There was not. This change is invisible to the user. Unchecking the docs item!

modules/frontend/pipeline/pipeline.go
```diff
@@ -0,0 +1,98 @@
...
+// Wrap implements Middleware.
+func (f MiddlewareFunc) Wrap(w http.RoundTripper) http.RoundTripper {
+	return f(w)
+}
+
+//
+// Builder and Bridge
+//
+
+// MergeMiddlewares takes a set of ordered middlewares and merges them into a pipeline
```
mdisibio, line 62: nit: comment

modules/frontend/pipeline/async_sharding.go
```diff
@@ -0,0 +1,104 @@
...
+	go func() {
+		defer asyncResp.done()
+
+		reqIdx := 0
+		for {
+			req := reqFn(reqIdx)
+			reqIdx++
+
+			// else check for a request to pass down the pipeline
+			if req == nil {
```
mdisibio, line 37: What do you think about the caller passing the expected number of requests?  It looks like the only reference is for multi-tenant where the number is known.

modules/frontend/pipeline/async_sharding.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		wg.Wait()
+	}()
+
+	return asyncResp
+}
+
+// NewAsyncSharderLimitedGoroutines creates a new AsyncResponse that shards requests to the next AsyncRoundTripper[*http.Response] using a limited number of goroutines.
+func NewAsyncSharderLimitedGoroutines(concurrent int, reqs <-chan *http.Request, resps Responses[*http.Response], next AsyncRoundTripper[*http.Response]) Responses[*http.Response] {
```
mdisibio, line 62: These two methods would be clearer to me if they were renamed to `NewAsyncSharderFunc` and `NewAsyncSharderChan`.  That looks like the main difference (in this current version at least). Since they both have concurrently limits.

modules/frontend/combiner/common.go
```diff
@@ -1,74 +1,104 @@
...
-	code          int
-	statusMessage string
-	err           error
+	//
+	httpStatusCode int
+	httpRespBody   string
 }
 
-func (c *genericCombiner[R]) AddRequest(res *http.Response, _ string) error {
+// AddRequest actually adds a response. Rename?
```
mdisibio, line 37: Yes please 👍

modules/frontend/frontend.go
```diff
@@ -55,19 +56,19 @@ func New(cfg Config, next http.RoundTripper, o overrides.Interface, reader tempo
...
 
 	// inject multi-tenant middleware in multi-tenant routes
 	traceByIDMiddleware := MergeMiddlewares(
 		newMultiTenantMiddleware(cfg, combiner.NewTraceByID, logger),
 		newTraceByIDMiddleware(cfg, o, logger), retryWare)
 
-	searchMiddleware := MergeMiddlewares(
-		newMultiTenantMiddleware(cfg, combiner.NewSearch, logger),
-		newSearchMiddleware(cfg, o, reader, searchCache, logger), retryWare)
+	searchPipeline := pipeline.Build(
```
mdisibio, line 68: Double checking my understanding:  Search is the only API moved to the new pipeline?

modules/frontend/frontend.go
```diff
@@ -55,19 +56,19 @@ func New(cfg Config, next http.RoundTripper, o overrides.Interface, reader tempo
...
+		asyncPipeline(cfg, newAsyncSearchSharder(reader, o, cfg.Search.Sharder, logger), logger),
```
mdisibio, line 69: This kind of hides the multi-tenant handling.  Is there a way to "flatten" it up to this level like the other apis are built?

modules/frontend/frontend.go
```diff
@@ -265,12 +253,26 @@ func buildUpstreamRequestURI(originalURI string, params url.Values) string {
...
 
-		return RoundTripperFunc(func(r *http.Request) (*http.Response, error) {
+		return pipeline.RoundTripperFunc(func(r *http.Request) (*http.Response, error) {
 			return searchRT.RoundTrip(r)
 		})
 	})
 }
+
+func asyncPipeline(cfg Config, sharder pipeline.AsyncMiddleware[*http.Response], logger log.Logger) []pipeline.AsyncMiddleware[*http.Response] {
+	if cfg.MultiTenantQueriesEnabled {
```
mdisibio, line 267: I think this might be changing multi-tenant behavior unintentionally.  Previously when `cfg.MultiTenantQueriesEnabled==false` it would pass-through to next().  But now it looks like it will fail whenever multiple tenant IDs are given.   

For example if `cfg.MultiTenantQueriesEnabled==false` and `X-Scope-OrgID: foo|bar` it will fail instead of using tenantID `foo|bar` literally?

modules/frontend/frontend.go
joe-elliott: good call. it's nice to see all the pipeline components in one spot. changed to:
```
	searchPipeline := pipeline.Build(
		[]pipeline.AsyncMiddleware[*http.Response]{
			newAsyncSearchSharder(reader, o, cfg.Search.Sharder, logger),
		},
		[]pipeline.Middleware{cacheWare, statusCodeWare, retryWare},
		next)
```

modules/frontend/frontend.go
joe-elliott: yup, you are correct. fixed to return a "no-op" middleware when `MultiTenantQueriesEnabled==false`. this will cause the pipeline to treat `foo|bar` literally.

modules/frontend/pipeline/collector_grpc.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return grpcError(err)
+	}
+
+	// stores any error that occurs during the streaming
+	//  the wg protects the store from concurrent writes/reads
+	var overallErr error
+
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
```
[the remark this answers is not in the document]
joe-elliott: good catch! removed.

modules/frontend/pipeline/async_sharding.go
joe-elliott: The intention was for `NewAsyncSharder` to create one goroutine for every concurrrent downstream request. This is the way the current frontend works.

`NewAsyncSharderLimitedGoroutines` creates a limited number of goroutine "workers" that can then fire up to concurrency requests downstream.

I don't think this is working the way I intended it to. This was the last area of code I was looking at closely. I'm going to do some more experimentation but I think that we will still need one goroutine per request. Perhaps we eventually migrate to an async request pipeline as well as response?

joe-elliott: Ok, I've cleaned these up and used your suggested names. I like where they are now, but think there's room for improvement.

modules/frontend/pipeline/async_sharding.go
```diff
@@ -0,0 +1,104 @@
...
+
+				asyncResp.Send(resp)
+				concurrencyLimiter.Done()
+			}
+		}()
+	}
+
+	go func() {
+		if resps != nil {
+			asyncResp.Send(resps)
```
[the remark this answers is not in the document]
mdisibio: Got it, thanks.   Caching is another instance, where the cacheKey is passed through the context.

modules/frontend/pipeline/sync_handler_cache.go
```diff
@@ -0,0 +1,165 @@
...
+
+	if len(buffer) == 0 {
+		return
+	}
+
+	c.c.Store(ctx, []string{key}, [][]byte{buffer})
+}
+
+// fetch fetches the response body from the cache. the caller assumes the responsibility of closing the response body.
+func (c *frontendCache) fetch(key string, pb proto.Message) bool {
```
mdisibio, line 131: This method looks unused except for the test, and wasn't sure about json+proto.Message.  Since we know it's a `proto.Message` we could store the binary representation in cache instead.

mdisibio (approved):
I really like this, the async parts seem clearer and streaming everywhere will be awesome.  Definitely think we can make the sync parts (cache, etc) also async in the future.

modules/frontend/pipeline/sync_handler_cache.go
joe-elliott: yup. i do think there's an opportunity to make the sync part of the pipeline type aware with generics which would allow this, but we're not quite there yet

PR #3403, opened 2024-02-16
Restore tempo_request_duration_seconds metrics for querier_api_* requests

kvrhdn:
**What this PR does**:

Recent changes (I suspect https://github.com/grafana/tempo/pull/3300) broke middleware on our querier requests. We have HTTP middleware that provides tracing and metrics. Requests sent to the querier are routed slightly differently and due to this change we didn't get metrics anymore for the HTTP routes.

__Issue__
So `tempo_request_duration_seconds` does not show up for routes like `querier_api_search_tags` and `querier_api_traces_traceid`. GRPC routes were still instrumented correctly (`/tempopb.Querier/FindTraceByID` and `/tempopb.Querier/SearchTags`).

__How it works__
Normal middleware flow:

```
http.Handler.Serve --> middleware (tracing, metrics) --> mux.Router (calls the correct http.Handler for each path)
```

When we set `stream_over_http_enabled: false` this will be

```
http.Handler.Serve --> dskit.Server (with instrumentation middleware) --> Tempo's mux.Router
```

When we set `stream_over_http_enabled: true` we switch up the order a bit:

```
http.Handler.Server --> GRPC interceptor --> instrumentation middleware --> Tempo's Router
                        (1)                                                 (2)
```

The bug: requests from query-frontend to querier are handled differently, they are _pulled_ by the frontend worker process.
The frontend worker will pull a request from the frontend, unpack the GRPC requests and then push the HTTP request to the regular HTTP handler. This handler was the router (marked `(2)`), since this handler is after instrumentation we didn't get any metrics for it. By instead passing the start of the middleware chain (`(1)`) we fix this.

Adding `HTTPRouter` and `HTTPHandler` to `TempoServer` should make this more explicit.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo/app/server_service.go
+++ b/cmd/tempo/app/server_service.go
@@ -21,7 +21,8 @@ import (
 )
 
 type TempoServer interface {
-	HTTP() *mux.Router
+	HTTPRouter() *mux.Router
+	HTTPHandler() http.Handler
 	GRPC() *grpc.Server
 	Log() log.Logger
 	EnableHTTP2()
@@ -32,7 +33,8 @@ type TempoServer interface {
 
 // todo: evaluate whether the internal server should be included as part of this
 type tempoServer struct {
-	mux *mux.Router // all tempo http routes are added here
+	mux     *mux.Router  // all tempo http routes are added here
+	handler http.Handler // the final handler which includes the router and any middleware
 
 	externalServer  *server.Server // the standard server that all HTTP/GRPC requests are served on
 	enableHTTP2Once sync.Once
@@ -45,8 +47,12 @@ func newTempoServer() *tempoServer {
 	}
 }
 
-func (s *tempoServer) HTTP() *mux.Router {
+func (s *tempoServer) HTTPRouter() *mux.Router {
 	return s.mux
+}
+
+func (s *tempoServer) HTTPHandler() http.Handler {
+	return s.handler
 }
 
 func (s *tempoServer) GRPC() *grpc.Server {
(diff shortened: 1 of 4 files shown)


cmd/tempo/app/server_service.go
```diff
@@ -71,22 +77,28 @@ func (s *tempoServer) StartAndReturnService(cfg server.Config, supportGRPCOnHTTP
...
+
+	if !supportGRPCOnHTTP {
+		// We don't do any GRPC handling, let the library handle all routing for us
+		cfg.Router = s.mux
+
+		s.externalServer, err = server.NewWithMetrics(cfg, metrics)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create server: %w", err)
+		}
+		s.handler = s.externalServer.HTTPServer.Handler
```
kvrhdn, line 90: @joe-elliott we were really close while pairing. We used `s.externalServer.HTTP` which is the router, `HTTPServer.Handler` is the handler which contains all middleware as well.

joe-elliott: so close!

cmd/tempo/app/server_service.go
```diff
@@ -21,7 +21,8 @@ import (
 )
 
 type TempoServer interface {
-	HTTP() *mux.Router
+	HTTPRouter() *mux.Router
```
joe-elliott, line 24: i like this rename for clarity

joe-elliott (approved):

Issue #3338, opened 2024-01-26
spanmetrics generator creates negative values for traces_spanmetrics_latency_bucket


mar4uk:
**Describe the bug**
When spanmetrics metrics-generator is enabled sometimes it generates negative values for traces_spanmetrics_latency_bucket
The corresponding issues in app-o11y https://github.com/grafana/app-o11y/issues/497, https://github.com/grafana/app-o11y/issues/178

**To Reproduce**
It reproduces from time to time. There is no reliable way to reproduce this

**Expected behavior**
There should be no buckets with negative values because latency is always non-negative

kvrhdn:
Doing a quick scan through the code, this is where we calculate the latency of each span:

https://github.com/grafana/tempo/blob/7d9defd2c802af7672412c309efbff11cd089733/modules/generator/processor/spanmetrics/spanmetrics.go#L141

We observe it here:

https://github.com/grafana/tempo/blob/7d9defd2c802af7672412c309efbff11cd089733/modules/generator/processor/spanmetrics/spanmetrics.go#L206-L208

So likely culprit seem to be:
- the end timestamp is earlier than the start, I'm not sure if this can happen, I'd imagine instrumentation should prevent this from happening
- span multiplier could be negative, but we calculate it here and there is a check to make sure this value isn't negative

https://github.com/grafana/tempo/blob/7d9defd2c802af7672412c309efbff11cd089733/modules/generator/processor/util/util.go#L33-L46

kvrhdn:
Is there anything special about the spans this is happening to? Does this happen to a lot of spans or just a few, are they especially short?

cedricziel:
hi @kvrhdn - we have a live example with accessible traces here https://otlpcortexdev01.grafana-dev.net/a/grafana-app-observability-app/services/python3.11?var-prometheus=grafanacloud-prom&var-loki=grafanacloud-logs&var-tempo=grafanacloud-traces&var-pyroscope=grafanacloud-profiles&var-environmentValue=$__all&var-groupBy=__empty

mar4uk:
> Is there anything special about the spans this is happening to? Does this happen to a lot of spans or just a few, are they especially short?

In the case Cedric mentioned above, there are spans `GET /alive` that take ~1-2ms
<img width="1242" alt="Screenshot 2024-02-06 at 12 41 50" src="https://github.com/grafana/tempo/assets/2266206/2b4896e8-3850-4cbf-9e75-64780b91a629">

PR #3412, opened 2024-02-20
Fix spanmetrics latency underflow, prevent partial histogram scrape

mdisibio:
**What this PR does**:
Fixes 2 small bugs in the span metrics processor.

1. Latency underflow

As mentioned in #3338 the latency calculation will be incorrect if span End is < Start.  However it doesn't lead to a negative float64.  Because start and end are uint64 the calculation underflows to a very large uint64 before float64 conversion, for example float64(uint64(-1)) is `1.8446744073709552e+19`.  I'm not sure of the side effects from this all the way to the metrics backend, but in addition to the bad value the counter also stops increasing.  It no longer has enough precision to capture increments of +1.0.

Rather than ignore these spans, they are treated as zero latency. This ensures that _count and le buckets are still incremented.

2. Partial histogram scrape

The histograms counters are updated outside of lock. Therefore it's possible for a histogram to be scraped with inconsistent values.  For example if the histogram is scraped after incrementing `_count` but before incrementing `le="+Inf"`.  In all cases these metrics should be identical.   I addressed this by doing `updateSeries` under lock and reverting to `sync.Mutex` instead of `RWMutex`.  This has performance implications, but not sure of the degree in practice.  If we really need to avoid the lock, then we could make it more "failsafe" by changing the order of increments in `updateSeries`.  I.e. update the buckets in reverse order so `+Inf` is always highest even in a partial scrape.  Happy to discuss more.

Relates to #3338   I don't think it fixes it

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/processor/spanmetrics/spanmetrics.go
+++ b/modules/generator/processor/spanmetrics/spanmetrics.go
@@ -138,7 +138,11 @@ func (p *Processor) aggregateMetrics(resourceSpans []*v1_trace.ResourceSpans) {
 }
 
 func (p *Processor) aggregateMetricsForSpan(svcName string, jobName string, instanceID string, rs *v1.Resource, span *v1_trace.Span, resourceLabels []string, resourceValues []string) {
-	latencySeconds := float64(span.GetEndTimeUnixNano()-span.GetStartTimeUnixNano()) / float64(time.Second.Nanoseconds())
+	// Spans with negative latency are treated as zero.
+	latencySeconds := 0.0
+	if start, end := span.GetStartTimeUnixNano(), span.GetEndTimeUnixNano(); start < end {
+		latencySeconds = float64(end-start) / float64(time.Second.Nanoseconds())
+	}
 
 	labelValues := make([]string, 0, 4+len(p.Cfg.Dimensions))
 	targetInfoLabelValues := make([]string, len(resourceLabels))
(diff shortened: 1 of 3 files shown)


joe-elliott (approved):

mdisibio:
Thinking about it some more - the lack of mutex lock might also explain situations where `_count` is < `le="+Inf"`.   If the histogram is incremented in between these two lines https://github.com/grafana/tempo/blob/main/modules/generator/registry/histogram.go#L192-L202  then it's possible for the Inf bucket to be greater than _count.  If `appender.Append` is performing i/o and pausing the collection routine, then the likelihood is increased.

```go
_, err = appender.Append(0, lb.Labels(), timeMs, s.count.Load())

...
		
for i, bucketLabel := range h.bucketLabels {
	ref, err := appender.Append(0, lb.Labels(), timeMs, s.buckets[i].Load())
```

kvrhdn (approved):
Nice find!

The performance impact from the lock shouldn't be _too bad_. We collect samples only once every 15s to 60s. While we are collecting a metric it would lock out ingest for that metric. Since we have a low amount of metrics (each processor has a handful) this can lead to contention, but collecting data itself is pretty fast so...

Some thoughts (maybe for a next refactor):
- I think we can drop the atomic types in `histogramSeries` now, using regular data types might be better for memory
- Instead of locking the entire `series` map, we can keep the `RWMutex` on `series` and keep a lock on each `histogramSeries`. Whenever you modify or read the series you would lock that series only. This would be a more granular approach to avoiding the partial reads/writes but requires more lock management.

kvrhdn (approved):
I've run this change in our largest cluster and didn't see significant changes in resource usage or latency. I'm not 100% yet this fixes our correctness issues, that will require some further investigation. I'm good with merging this though :+1:

PR #3426, opened 2024-02-26
Return unfiltered results when a bad TraceQL query is provided

mapno:
**What this PR does**:

Return unfiltered results when a bad TraceQL query is provided in autocomplete. Before it would error with a 500 without providing any useful feedback to the user.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/traceql/engine.go
```diff
@@ -152,7 +152,13 @@ func (e *Engine) ExecuteTagValues(
 
 	rootExpr, err := Parse(query)
 	if err != nil {
-		return err
+		// If the query has bad TraceQL, don't error out, return unfiltered results
+		var parseErr *ParseError
+		if errors.As(err, &parseErr) {
+			rootExpr, _ = Parse("{ true }")
```
mdisibio, line 158: How does this performance compare versus falling back to the non-traceql value lookup?

mapno: There is a last check in the fetcher itself https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet3/block_autocomplete.go#L28-L31. It'll go through the non-traceql lookup.

ie-pham (approved):
LGTM

PR #3440, opened 2024-02-28
[TraceQL Metrics] Fix handling of empties and nils + other fixes

### LINKED ISSUE #3429: [TraceQL Metrics] Unexpected error when rating by rootName/rootServiceName
image

Reproduced with the same images as here:
https://github.com/grafana/tempo/issues/3428

If I include a filter in the query it works:

image

mdisibio:
**What this PR does**:
Basically this overhauls the handling of timeseries labels for metrics queries.  There are several fixes in here, but it also lines up behavior with: https://github.com/grafana/grafana/pull/83619

* Empty strings were previously dropped but now they are preserved. For example: `{rootName=""} | rate() by (rootName)`.  This would look like a completely empty labelset in the response and cause issues downstream.  Now it's preserved as the single label `rootName` with an empty string value.
* When all labels are nil, we were backfilling a single label with the text "<nil>" to make Grafana happy like.  This was actually premature specialization for the prometheus-compatible API, and now that we have the new dedicated Tempo-UI we can fix this.  Now nils are dropped entirely and handled in the frontend. They will become `{}` in the Tempo UI and the query name in the Prometheus UI (a fallback method).  Note:  the `EmitDefaults:true` is part of this so that Tempo always outputs an empty array in the json (and not nil).
* Fix sorting.  Previously we only sorted during the conversion to Prometheus format.  Also fix sorting to handle non-strings and multiple label values correctly.
* Fix default step calculation and error handling.

Fixes #3429 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * [ENHANCEMENT] Add string interning to TraceQL queries [#3411](https://github.com/grafana/tempo/pull/3411) (@mapno)
 * [BUGFIX] Fix metrics query results when filtering and rating on the same attribute [#3428](https://github.com/grafana/tempo/issues/3428) (@mdisibio)
+* [BUGFIX] Fix metrics query results when series contain empty strings or nil values [#3429](https://github.com/grafana/tempo/issues/3429) (@mdisibio)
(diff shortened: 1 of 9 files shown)


mdisibio:
Requires: https://github.com/grafana/grafana/pull/83619

mapno (approved):
LGTM

PR #3442, opened 2024-02-29
Docs: Fix typo

fabrizio-grafana:
Remove unnecessary comma from Traces doc page.

### DIFF
--- a/docs/sources/tempo/traces.md
+++ b/docs/sources/tempo/traces.md
@@ -21,7 +21,7 @@ A span is a unit of work within a trace that has a start time relative to the be
 It may optionally include key/value attributes that are relevant to the span itself, for example the HTTP method being used by the span, as well as other metadata such as sub-span events or links to other spans.
 
 By definition, traces are never complete. You can always push a new batch of spans, even if days have passed since the last one.
-When receiving a query requesting a stored trace, tracing backends (for example Tempo), find all the spans for that specific trace and collate them into a returned result. For that reason, issues can start to arise predominantly on retrieval of the trace data if you are creating traces that are extremely large in size.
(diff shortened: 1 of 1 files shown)


docs/sources/tempo/traces.md
```diff
@@ -21,7 +21,7 @@ A span is a unit of work within a trace that has a start time relative to the be
 It may optionally include key/value attributes that are relevant to the span itself, for example the HTTP method being used by the span, as well as other metadata such as sub-span events or links to other spans.
 
 By definition, traces are never complete. You can always push a new batch of spans, even if days have passed since the last one.
-When receiving a query requesting a stored trace, tracing backends (for example Tempo), find all the spans for that specific trace and collate them into a returned result. For that reason, issues can start to arise predominantly on retrieval of the trace data if you are creating traces that are extremely large in size.
+When receiving a query requesting a stored trace, tracing backends (for example Tempo) find all the spans for that specific trace and collate them into a returned result. For that reason, issues can start to arise predominantly on retrieval of the trace data if you are creating traces that are extremely large in size.
```
joe-elliott, line 24: That sentence is kind of awkward. Since these are Tempo docs, I also think we should remove mention of other backends. Maybe?

"When receiving a query requesting a stored trace, Tempo will find all ..."

fabrizio-grafana: I guess the idea of this page was to be a generic introduction to traces, unrelated to Tempo. But GTM in rephrasing it!

knylander-grafana: This is a generic introduction, but we can be more specific by chanting it to tracing backends like Tempo -- which shifts the focus to Tempo and helps people understand that Tempo is the thing that does all this cool stuff. 
```suggestion
When receiving a query requesting a stored trace, tracing backends like Tempo find all the spans for that specific trace and collate them into a returned result.
For that reason, issues can start to arise predominantly on retrieval of the trace data if you are creating traces that are extremely large in size.
```

PR #3443, opened 2024-02-29
Correctly cancel GRPC context beneath the HTTP server

joe-elliott:
**What this PR does**:
Currently if a GRPC streaming query or HTTP query exceeds the HTTP write timeout when passing through the HTTP server it continues until completion. i.e. context is not correctly cancelled. In addition this unhelpful error message is presented in Grafana:

```
rpc error: code = Internal desc = stream terminated by RST_STREAM with error code: INTERNAL_ERROR
```

This PR correctly cancels context for both the gRPC and HTTP frontend endpoints. It updates our dskit dependency to access a new http timeout middleware and uses gRPC interceptor to cancel calls to the streaming querier.

Concerns:
- The http and gRPC timeout are enforced in very different places due to the nature of the way these options are configured.
- I don't like the config name `api_timeout`. Any suggestions?

**Checklist**
- [x] Tests updated
- [x] Documentation added


electron0zero (approved):

zalegrala (approved):
This looks reasonable to me.  Nice use of the interceptor.

PR #3453, opened 2024-03-03
Add new histogram to generator - messaging system latency

### LINKED ISSUE #3232: Service-graph metrics for messaging_system connection type: Taking into account network+middleware latency
### Discussed in https://github.com/grafana/tempo/discussions/3210

<div type='discussions-op-text'>

<sup>Originally posted by **adirmatzkin** December 10, 2023</sup>
I was wondering about the correctness of the way the service_graph metrics generated by the Metrics Generator are calculated for server side metrics.

As for the "classic" client-server interaction I have no doubt, but for the producer-consumer ones (`connection_type=messaging_system`), I believe it shouldn't be as how it is today (or at least expand to give a solution to another use case).

The case I'm describing is when the **total** latency (from the start time of the producer until the end time of the consumer) is the latency needed to be measured.

Imagine a situation where a message was published to some queue A, (publish span took 10ms), then the message sits in the queue for 20s (because no consumer was available to process another message for example), and then finally pulled and processed in 15ms by the consumer.

The service-graph metrics will only "count" the 10ms for the producer side, and the 15ms for the consumer side ([consumer calculates it's latency by the consumer span duration](https://github.com/grafana/tempo/blob/main/modules%2Fgenerator%2Fprocessor%2Fservicegraphs%2Fservicegraphs.go#L205)),
Leaving the 20s delay not mentioned on any side and in any metric...

I believe this 20s latency should be considered _somehow_, what do you guys think? 🙃

</div>

---

The need this issue comes to address is giving a more comprehensive metric to the calculated latency metric in the service-graph processor - specifically for the `messaging_system` connection_type.
As for today, the latency metric generated in the service-graph processor handles [producer-consumer] spans the same way [client-server] spans are. This situation leads us to essentially disregard the "actual" latency **between** the services.
To create a more "complete" picture, we should take into account the network overhead and middleware latency (for example the time a message waits in a queue).
 
My proposal is as follows:
1. Add a **new** latency histogram metric.
2. The calculation will involve subtracting the **`producer end_time` from the `consumer start_time`.**

That way, if considering the following trace:
```
Span A: |--------------------|
Span B:         (t1)|--------|(t2)
Span C:                                             (t3)|-----------|(t4)
Span D:                                                             |-----|  
```
we will be able to get all possible latencies aggregations:
- `t2-t1` ("span B") - using the client-side metric we already have today.
- `t4-t3` ("span C") - using the server-side metric we already have today.
- `t3-t2` (the missing part this issue comes to solve!)
- Any other combination of summing the above options (`t4-t2` might be an interesting one - the time alapsed from publishing a message to finishing processing it...).

adirmatzkin:
**What this PR does**:
This PR adds a new metric calculated by the metrics-generator service-graph processor.
The metric is a histogram for the latency **between** 2 services communicating through a messaging system.

Fixes #3232 

Screenshot 2024-03-03 18 14 48

**Checklist**
- [x] Tests updated
- [x] Documentation added


joe-elliott:
Thanks for putting some time into this. It's a fairly small change should be a pretty easy review, but we do need this to be configurable.

We are currently trying to keep metrics generator series down and histograms tend to generate a lot of series.

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -315,8 +329,14 @@ func (p *Processor) spanFailed(span *v1_trace.Span) bool {
 	return span.GetStatus().GetCode() == v1_trace.Status_STATUS_CODE_ERROR
 }
 
+func unixNanosDiffSec(unixNanoStart uint64, unixNanoEnd uint64) float64 {
+	// handling potential underflow of unit64s substraction
+	diff := int64(unixNanoEnd) - int64(unixNanoStart)
```
joe-elliott, line 334: instead of casting we could check to see if diff > end for overflow. is that better?

joe-elliott: also, we recently made a change to protect span metrics from negative latencies:

https://github.com/grafana/tempo/pull/3412/files

@mdisibio should we make sure the same protections are here?

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -100,10 +102,11 @@ func New(cfg Config, tenant string, registry registry.Registry, logger log.Logge
...
 
-		serviceGraphRequestTotal:                  registry.NewCounter(metricRequestTotal),
-		serviceGraphRequestFailedTotal:            registry.NewCounter(metricRequestFailedTotal),
-		serviceGraphRequestServerSecondsHistogram: registry.NewHistogram(metricRequestServerSeconds, cfg.HistogramBuckets),
-		serviceGraphRequestClientSecondsHistogram: registry.NewHistogram(metricRequestClientSeconds, cfg.HistogramBuckets),
+		serviceGraphRequestTotal:                           registry.NewCounter(metricRequestTotal),
+		serviceGraphRequestFailedTotal:                     registry.NewCounter(metricRequestFailedTotal),
+		serviceGraphRequestServerSecondsHistogram:          registry.NewHistogram(metricRequestServerSeconds, cfg.HistogramBuckets),
+		serviceGraphRequestClientSecondsHistogram:          registry.NewHistogram(metricRequestClientSeconds, cfg.HistogramBuckets),
+		serviceGraphRequestMessagingSystemSecondsHistogram: registry.NewHistogram(metricRequestMessagingSystemSeconds, cfg.HistogramBuckets),
```
joe-elliott, line 109: will this create series even if we never call `.Observe...`?

joe-elliott (commented):
a few questions, but looking good!

also, we will need a changelog

modules/generator/processor/servicegraphs/servicegraphs.go
adirmatzkin: Tested it locally with the intro-to-mltp setup - when the config wasn't set to true (defaults to false), there were no series in Mimir... so I guess that's fine 📈

modules/generator/overrides.go
```diff
@@ -33,6 +33,7 @@ type metricsGeneratorOverrides interface {
 	MetricsGeneratorProcessorSpanMetricsDimensionMappings(userID string) []sharedconfig.DimensionMappings
 	MetricsGeneratorProcessorSpanMetricsEnableTargetInfo(userID string) bool
 	MetricsGeneratorProcessorServiceGraphsEnableClientServerPrefix(userID string) bool
+	MetricsGeneratorProcessorServiceGraphsEnableMessagingSystemLatencyHistogram(userID string) bool
```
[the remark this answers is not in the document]
adirmatzkin: yeah.. tried to stick with the convention and it got quite lol
So... should we leave it like that? thinking of changing anything here?

modules/generator/processor/servicegraphs/servicegraphs.go
adirmatzkin: > instead of casting we could check to see if diff > end for overflow. is that better?

I agree we can make it cleaner,
Really liked your idea, very clever 🙃 My only concern about it is how clear it will be. After making the check if diff > end, if underflow did happen, meaning the subtraction yields a negative number - don't we want to return it?

If not, maybe we could make this function a bit more specific and move the negative values "protection" into it.. and then.. return 0?
It could also be that simple:
```
func unixNanosDiffSec(unixNanoStart uint64, unixNanoEnd uint64) float64 {
	if unixNanoStart > unixNanoEnd {
		// To prevent underflow, return 0.
		return 0
	}
	// Safe subtraction.
	return float64(unixNanoEnd-unixNanoStart) / 1e9
}
```

adirmatzkin (commented):
Also added to changelog 👍🏼 
Will update docs after making a decision about the "negative diffs"

modules/generator/overrides.go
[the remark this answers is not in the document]
joe-elliott: it's fine. i would have named it the same thing. i just did a double take a the length!

modules/generator/processor/servicegraphs/servicegraphs.go
joe-elliott: let's follow suit with the linked PR for consistency and return 0.

i think your suggested `unixNanosDiffSec` in this PR is perfect.

modules/generator/processor/servicegraphs/servicegraphs.go
joe-elliott: works for me!

joe-elliott:
Thanks for the work. This is looking good.

heads up you will need this commit in your branch (merge or rebase) to get CI to pass:

https://github.com/grafana/tempo/commit/7540d573c8b41f5f10ea6e4e5f6b791bb97cb911

adirmatzkin:
> heads up you will need this commit in your branch (merge or rebase) to get CI to pass:

Synced from upstream + added docs. 
@joe-elliott anything else needed? 🙃

docs/sources/tempo/configuration/_index.md
```diff
@@ -319,6 +319,9 @@ metrics_generator:
             # per additional dimension instead of one.
             [enable_client_server_prefix: <bool> | default = false]
 
+            # If enabled another histogram will be produced for interactions over messaging systems middlewares
+            [enable_messaging_system_latency_histogram: <bool> | default = false]
```
joe-elliott, line 323: this latency may be seconds, minutes, hours, days, etc. if someone is interested in this feature over longer time periods they should consider increasing the time that the processor waits to match an edge. default is 10s:

```
metrics_generator:
    processor:
        service_graphs:
            wait: 10s
```

can we note this in the docs?

docs/sources/tempo/metrics-generator/service_graphs/_index.md
```diff
@@ -60,14 +60,15 @@ Virtual nodes can be detected in two different ways:
...
+| traces_service_graph_request_server_seconds           | Histogram | client, server, connection_type | Time for a request between two nodes as seen from the server                                               |
+| traces_service_graph_request_client_seconds           | Histogram | client, server, connection_type | Time for a request between two nodes as seen from the client                                               |
+| traces_service_graph_request_messaging_system_seconds | Histogram | client, server, connection_type | (Off by default) Time between publisher and consumer for services communicating through a messaging system |
+| traces_service_graph_unpaired_spans_total             | Counter   | client, server, connection_type | Total count of unpaired spans                                                                              |
+| traces_service_graph_dropped_spans_total              | Counter   | client, server, connection_type | Total count of dropped spans                                                                               |
```
joe-elliott, line 71: do these metrics exist? I'm not finding them.

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -178,6 +181,7 @@ func (p *Processor) consume(resourceSpans []*v1_trace.ResourceSpans) (err error)
 						e.ConnectionType = connectionType
 						e.ClientService = svcName
 						e.ClientLatencySec = spanDurationSec(span)
+						e.ClientEndTimeUnixNano = span.EndTimeUnixNano
```
joe-elliott, line 184: should we put this line in the `Span_SPAN_KIND_PRODUCER` section of the switch to show it belongs with that functionality?

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -203,6 +207,7 @@ func (p *Processor) consume(resourceSpans []*v1_trace.ResourceSpans) (err error)
 						e.ConnectionType = connectionType
 						e.ServerService = svcName
 						e.ServerLatencySec = spanDurationSec(span)
+						e.ServerStartTimeUnixNano = span.StartTimeUnixNano
```
joe-elliott, line 210: same thoughts as above.

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -315,8 +329,17 @@ func (p *Processor) spanFailed(span *v1_trace.Span) bool {
 	return span.GetStatus().GetCode() == v1_trace.Status_STATUS_CODE_ERROR
 }
 
+func unixNanosDiffSec(unixNanoStart uint64, unixNanoEnd uint64) float64 {
+	if unixNanoStart > unixNanoEnd {
+		// To prevent underflow, return 0.
+		return 0
+	}
+	// Safe subtraction.
+	return float64(unixNanoEnd-unixNanoStart) / 1e9
```
joe-elliott, line 338: should we divide by the constant: `/ time.Second`  ?

joe-elliott (commented):
Looking good, just have a few small thoughts.

Also, what do you think about adding the ConnectionType to the expired edge metric? If someone is wanting to do service graph metrics through queues I'm concerned it will create a lot of expired edges. It might be nice for an operator to know if their expired edges are largely queues or synchronous calls.

docs/sources/tempo/metrics-generator/service_graphs/estimate-cardinality.md
```diff
@@ -19,30 +19,46 @@ The amount of edges depends on the number of nodes in the system and the directi
...
 
 Finally, we get the following cardinality estimation:
 
 ```
-  Sum: 8 * #hops + 2 * #services
+  Sum: [([2 * #hb] + 2) * #hops] + [2 * #services]
+```
+
+{{% admonition type="note" %}}
+If `enable_messaging_system_latency_histogram` configuration is set to `true`, another histogram will be produced:
```
knylander-grafana, line 48: We try to use present, when possible.  More information in the [Writers' Toolkit](https://grafana.com/docs/writers-toolkit/write/style-guide/style-conventions/#write-in-present-tense).
```suggestion
If `enable_messaging_system_latency_histogram` configuration is set to `true`, another histogram is produced:
```

docs/sources/tempo/metrics-generator/service_graphs/estimate-cardinality.md
```diff
@@ -19,30 +19,46 @@ The amount of edges depends on the number of nodes in the system and the directi
...
+
 - A system with 3 nodes `(A, B, C)` of which A only calls B and B only calls C will have 2 hops `(A → B, B → C)`
 - A system with 3 nodes `(A, B, C)` that call each other (i.e., all bidirectional link) will have 6 hops `(A → B, B → A, B → C, C → B, A → C, C → A)`
 
 We can’t calculate the amount of hops automatically based upon the nodes,
 but it should be a value between `#services - 1` and `#services!`.
 
 If we know the amount of hops in a system, we can calculate the cardinality of the generated
-[service graphs]({{< relref "../service_graphs" >}}):
+[service graphs]({{< relref "../service_graphs" >}}) (assuming `#hb` is the number of histogram buckets):
```
knylander-grafana, line 30: Thank you for adding #hb and explaining it.

docs/sources/tempo/metrics-generator/service_graphs/_index.md
knylander-grafana: Are these the metrics? https://github.com/grafana/tempo/blob/3d2850d457725aa61afd00965c9d8ca7092b84d3/modules/generator/processor/servicegraphs/servicegraphs.go#L49

modules/generator/processor/servicegraphs/servicegraphs.go
adirmatzkin: so this one is resolved? 🤔

docs/sources/tempo/configuration/_index.md
adirmatzkin: Added a sentence about it.
Although it's relevant for the entire processor... the whole service-graph concept is sensitive to significant time diffs between the spans pairs 🤷🏼‍♂️.
I personally believe that anyone configuring these knobs should be aware of this and understand it at that level... but I suppose another notice won't hurt anybody...

knylander-grafana: Thank you for adding a sentence. It's always good to have an a sentence or two explanation in case someone doesn't know.

modules/generator/processor/servicegraphs/servicegraphs.go
adirmatzkin: @joe-elliott 🙃

joe-elliott: apologies for the delay. after reviewing all of this logic more carefully I think my concern is that we assign values to the Edge struct in the `consume` function and then interpret them in the `onComplete` function. There are a lot of values and not a lot of clarity on how the values flow through to the various metrics. 

in my mind making a change in this function would help with that clarity but reviewing the logic as a whole i now think it's a mistake. we should just faithfully set the values here and let the `onComplete` and `onExpire` functions interpret them as they will. There's nothing in the future that would prevent something unrelated to messaging systems using these values.

I'm good as is. Thank you for your patience and thoughts here!

adirmatzkin: Sorry for the delay here...
So should we resolve this comment? (and the following similar one)

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -286,6 +291,15 @@ func (p *Processor) onComplete(e *store.Edge) {
 
 	p.serviceGraphRequestServerSecondsHistogram.ObserveWithExemplar(registryLabelValues, e.ServerLatencySec, e.TraceID, e.SpanMultiplier)
 	p.serviceGraphRequestClientSecondsHistogram.ObserveWithExemplar(registryLabelValues, e.ClientLatencySec, e.TraceID, e.SpanMultiplier)
+
+	if p.Cfg.EnableMessagingSystemLatencyHistogram && e.ConnectionType == store.MessagingSystem {
```
joe-elliott, line 295: this is looking really good. i think my final question is just related to this recent discussion item:

https://github.com/grafana/tempo/discussions/3616

Is there any reason in particular to limit this change to producer/consumer spans? Does making this apply to client/server as well help the request above?

cc @jSwanson99

knylander-grafana:
Thank you for updating the docs

modules/generator/processor/servicegraphs/servicegraphs.go
adirmatzkin: The calculation I did here won't help with what @jSwanson99 described in his [screenshot](https://github.com/grafana/tempo/discussions/3616#discussion-6565351).
I'm looking at the diff between the producer ("client") start time and the consumer ("server") start time.

So anyway, I don't see it relevant for this PR...
Anyway - commented there my opinion 🙃

adirmatzkin: @joe-elliott let me know if we're clear to rebase again and merge 🥳

joe-elliott: sorry friend. we have worked on this long enough. after you catch up to main I will approve and merge. it will be easy enough to extend if @jSwanson99 feels like it will solve his use case. 

apologies for the delay.

adirmatzkin: No worries,
I will sync to main tomorrow morning and let you know 🙂

adirmatzkin:
@joe-elliott I think we're ready 🥳

joe-elliott:
running CI!

joe-elliott (approved):

PR #3466, opened 2024-03-05
Handle prefixes when listing blocks from S3

### LINKED ISSUE #3465: Retention deletion does not work on S3 or GCS when a prefix is configured
**Describe the bug**
When object storage is configured to use AWS S3, and a prefix is configured, if retention is configured it does not appear to delete old data beyond the retention period.

**To Reproduce**
1. Start Tempo v2.4.0
2. Wait for `blocklist_poll` time to elapse.
3. Wait for `compacted_block_retention` time to elapse.
4. Wait a lot longer.
5. Witness no files deleted from S3.

**Expected behavior**
I expect data files in S3 older than the configured `block_retention` to be deleted from S3. 

**Environment:**
 - Infrastructure: Kubernetes
 - Deployment tool: helm

**Additional Context**
I see 0 blocks listed in the compactor logs:
```
level=debug ts=2024-03-05T17:03:31.370043415Z caller=s3.go:273 msg="listing blocks" keypath=foo/tempo/ found=1 IsTruncated=false NextMarker=
level=debug ts=2024-03-05T17:03:32.513352722Z caller=s3.go:389 msg="listing blocks complete" blockIDs=0 compactedBlockIDs=0
```

Notable configuration:
```
compactor:
  compaction:
    block_retention: 48h
    compacted_block_retention: 10m
storage:
  trace:
    backend: s3
    blocklist_poll: 1m
    s3:
      bucket: foobar-bucket
      prefix: foo/tempo
```

When I have no prefix configured, I see a lot of compaction/retention activity in the (debug) log.  With a prefix configured, I see virtually nothing.

bpfoster:
**What this PR does**:  Adjusts the logic to list blocks from S3 to consider object path prefix.
You can see that `tempodb/backend/s3/s3.go` would ignore any results whose key, upon splitting by `/`, did not have 3 elements.  When a prefix is configured, e.g. `a/b/c/`, the returned key will instead be something like `a/b/c/<tenantID>/<blockID>/meta`.  Stip the prefix off this path before pulling out those 3 useful bits from the path.

Fixes #3465

**Checklist**
- [X] Tests updated
- [ ] Documentation added


joe-elliott:
ok, this is undoubtedly related to the polling improvements we made in 2.4.0. thx for looking at this.

@zalegrala can you review?

bpfoster:
Yes it looks like you were right that it was introduced in 2.4.0.  Running 2.3.1 looks like the compactor is cleaning up as expected.

Though I don't have Azure or GCS accounts to do integration testing, I have added to this PR unit tests covering these cases for both that are similar to S3.  As you expected, the issue is present in the GCS implementation (and I have added a fix for it to this PR), but looks like the Azure implementations already are correctly stripping the prefix.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * [BUGFIX] Fix metrics query results when series contain empty strings or nil values [#3429](https://github.com/grafana/tempo/issues/3429) (@mdisibio)
 * [BUGFIX] Return unfiltered results when a bad TraceQL query is provided in autocomplete. [#3426](https://github.com/grafana/tempo/pull/3426) (@mapno)
 * [BUGFIX] Correctly handle 429s in GRPC search streaming. [#3469](https://github.com/grafana/tempo/pull/3469) (@joe-ellitot)
+* [BUGFIX] Fix compaction/retention in AWS S3 and GCS when a prefix is configured. [#3465](https://github.com/grafana/tempo/issues/3465) (@bpfoster)
```
zalegrala, line 9: Mention the azure change also please.

tempodb/backend/azure/azure_test.go
```diff
@@ -252,6 +252,150 @@ func TestObjectWithPrefix(t *testing.T) {
...
+				Prefix:             tc.prefix,
+				Endpoint:           server.URL[7:], // [7:] -> strip http://,
+			})
+			require.NoError(t, err)
+
+			ctx := context.Background()
+			blockIDs, compactedBlockIDs, err2 := r.ListBlocks(ctx, "single-tenant")
+			assert.NoError(t, err2)
+
+			assert.Equal(t, 1, len(blockIDs))
```
zalegrala, line 393: Perhaps instead of checking just the length, we can check that the block IDs are what we expect.  Perhaps extend the test case for a `blockIDs` and include the static IDs that are in the response body.  This would apply to the other backends as well.

zalegrala (commented):
Thank you for the PR, and for fixing all the backends.  It might be nice to have an e2e test for this, but it digging in there is a little more work.  If you're willing that would be great, but if its too much, I won't push.  This is already a good step.

Part of the testing for the work that you are fixing was integrated into a test here to specifically test the poller with the various backends.  It might be simple enough to update the config to include a prefix.

Alternatively we have the main e2e tests that could be adjusted to include prefix testing.

https://github.com/grafana/tempo/blob/main/integration/e2e/e2e_test.go#L42

Probably duplicating the above with a config change would work.

bpfoster:
Thanks @zalegrala .  I'll address your comments and then take a look at the integrations tests and see if I can make something happen there.

CHANGELOG.md
zalegrala: Okay, let's see if it shows in the e2e tests when you get there.

bpfoster:
@zalegrala - please see the new commits when you have a chance.
- Addressed your comment with the unit test regarding checking block IDs
- Updated both the `poller_test` and `e2e_test` to exercise configuration with and without prefixes.  Feel free to choose both/either/neither :)  There are also other ways we could handle the test permutations, so let me know if you're not a fan of how I incorporated those changes into the tests.

Both the integrations tests confirm that S3 and GCS backends fail when configured with a prefix in v2.4.0 [^1].
Both confirm that Azure is in fact unaffected.
Both confirm that S3 and GCS are fixed in this PR.

I did slightly refactor my changes to the S3 and GCS backends to be more in line with the Azure backends.  This has the added benefit over the prior iteration of working when the configured prefix does not end in a `/`. This is covered here in prefix test cases in both integration tests.

Given that Azure has been verified in the integration tests, I believe we can resolve the discussion on [`CHANGELOG.md`](https://github.com/grafana/tempo/pull/3466#discussion_r1521779737)?

[^1]: I don't actually have a hard failure in the e2e test against v2.4.0.  I gave up waiting on the test to complete after 30 minutes (stuck on the `tempo.WaitSumMetricsWithOptions()` line).  I believe that the failure to complete, along with logs always showing `metas=0`, is a symptom of the issue - no blocks ever returned.

zalegrala:
Thank you for the updates, I've approved the CI.  I think the only concern I have is the one you mention, which is the amount of time it takes for the e2e tests.  The change to the `poller_test` looks good and I appreciate all the permutations, but perhaps we reduce the permutations on the e2e test as a matter of pragmatism.  The e2e tests take much longer to run generally.

I appreciate you confirming the Azure code with this test also.  Yes, I believe you are correct about the CHANGELOG.  I checked the code of both azure versions and it looks like this case is handled, so just the two you are fixing were missed.

I see a couple failures, but I'll check out the code and run locally to take a look.  Let's see what the rest of the CI checks look like.

bpfoster:
I addressed the format error, fixed the failing e2e tests, and removed 2/4 of the prefix test case permutations from the e2e test.

It looks like all these are passing the CI checks now.  The lint job is failing in the last run, but this looks more like a CI job issue, I don't think any changes here would cause the errors being reported.

A local run of `golangci-lint run --new-from-rev=772ebd79f1d9cbd75d2f190e1cb9e9390e30531f` reports 1 error but it is from to existing code that just had indentation changed.
```
integration/e2e/e2e_test.go:113:50: string `/api/echo` has 5 occurrences, make it a constant (goconst)
				assertEcho(t, "http://"+tempo.Endpoint(3200)+"/api/echo")
```

zalegrala (approved):
This looks good to me.  Thank you for the contribution.

PR #3470, opened 2024-03-06
Fix jsonnet check 1.21 -> 1.29

joe-elliott:
The 1.21 version of the k8s libsonnet has been retired. Upgrading to the most recent version: 1.29

Removed here: https://github.com/jsonnet-libs/k8s-libsonnet/commit/90d0fae415cff9d0c089d60e9a478efee092c42e


ie-pham (approved):
it feels good to me?
the vibes seem good to me?

PR #3490, opened 2024-03-14
[TraceQL Metrics] Step align query_range time range

mdisibio:
**What this PR does**:
This PR aligns metrics time ranges to the step parameter.  Without this, the output can subtly change across refreshes of queries like "Last 1 Hour" because spans fall into different buckets a little differently.   We have to align in three places: overall, for the generators, and the backend.  There's some slightly tricky order of operations, i.e. do we check the max_duration limit before or after alignment?  I chose to align first, so that the limit check is against what will actually be executed, but can see an argument for the reverse.

Also a bunch of other mainly cosmetic stuff:  code cleanup, error handling, fewer pointers.

Fixes n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/query_range_sharding.go
+++ b/modules/frontend/query_range_sharding.go
@@ -85,12 +85,12 @@ func (s queryRangeSharder) RoundTrip(r *http.Request) (*http.Response, error) {
 		now = time.Unix(now.Unix(), 0)
 	}
 
-	queryRangeReq, err := api.ParseQueryRangeRequest(r)
+	req, err := api.ParseQueryRangeRequest(r)
 	if err != nil {
 		return s.respErrHandler(isProm, err)
 	}
 
-	expr, err := traceql.Parse(queryRangeReq.Query)
+	expr, err := traceql.Parse(req.Query)
 	if err != nil {
 		return s.respErrHandler(isProm, err)
 	}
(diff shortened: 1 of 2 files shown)


modules/frontend/query_range_sharding.go
```diff
@@ -387,8 +395,7 @@ func (s *queryRangeSharder) backendRange(now time.Time, start, end uint64, query
 	return start, end
 }
 
-func (s *queryRangeSharder) generatorRequest(searchReq tempopb.QueryRangeRequest, samplingRate float64) *queryRangeJob {
-	now := time.Now()
+func (s *queryRangeSharder) generatorRequest(searchReq tempopb.QueryRangeRequest, now time.Time, samplingRate float64) *queryRangeJob {
```
mapno, line 398: `now` seems to be unused

mapno (commented):
LGTM

modules/frontend/query_range_sharding.go
mdisibio: It's used [a bit further down](https://github.com/grafana/tempo/pull/3490/files/0432ccee3ed4f790c4b71415463194ff4bb39bfa#diff-1205a2cd25b76a20cc27dc681f5073614b170b1dfbf29069af221e6f8e7ffed6R399) to determine the starting time range of the generator request: now-querybackendafter.

mapno (approved):

PR #3495, opened 2024-03-15
Query Frontend Refactor: Fix goroutine leaks

joe-elliott:
- Add Send/NextComplete methods that must be called by the producer/consumer to guarantee cleanup of resources
- Add goroutine leak tests using built pipelines and the async response

### DIFF
--- a/modules/frontend/pipeline/responses.go
+++ b/modules/frontend/pipeline/responses.go
@@ -67,7 +67,7 @@ type asyncResponse struct {
 	respChan chan Responses[*http.Response]
 	errChan  chan error
 	err      *atomic.Error
-	done     *atomic.Bool
+	//done     *atomic.Bool
 
 	curResponses Responses[*http.Response]
 }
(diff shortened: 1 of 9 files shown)


modules/frontend/pipeline/responses.go
```diff
@@ -87,10 +94,27 @@ func (a *asyncResponse) SendError(err error) {
...
+// SendComplete indicates the sender is done. We close the channel to give a clear signal to the consumer
+func (a *asyncResponse) SendComplete() {
 	close(a.respChan)
 }
 
+// NextComplete indicates the receiver is done. We drain all channels and subchannels to goroutines are orphaned
+func (a *asyncResponse) NextComplete() {
+	go func() {
+		// drain the response channel?
+		for resps := range a.respChan {
```
mdisibio, line 106: Can you check my understanding: The goroutine leak was that if a receiver only partially read the response channel (because it hit max search results for example), then a sender would be orphaned waiting on the channel send.  Now when the receiver calls NextComplete it drains the channel until.

joe-elliott: correct, per an out of band conversation i have moved towards using the cancel context signal from the `Next()`er to cancel the Sends.

This is a bit more hidden and easy to break, but it's also cleaner and more in line with channel architectures.

mdisibio (approved):

modules/frontend/pipeline/responses.go
mdisibio: 👍 lgtm thanks

PR #3505, opened 2024-03-19
Context.canceled handling changes for slo and receiver shim

mdisibio:
**What this PR does**:

Addresses two places where a client disconnect could cause unexpected behavior and false alarms:
* SLO calculation - if a client disconnects while calling an API, now it's considered within SLO
* Pushing traces - if a client disconnects while pushing traces, now we ignore and don't propagate a 5xx upstream. The error is still logged to help troubleshooting   

Fixes n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/receiver/shim.go
+++ b/modules/distributor/receiver/shim.go
@@ -2,6 +2,7 @@ package receiver
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"time"
@@ -342,6 +343,12 @@ func (r *receiversShim) ConsumeTraces(ctx context.Context, td ptrace.Traces) err
 	metricPushDuration.Observe(time.Since(start).Seconds())
 	if err != nil {
 		r.logger.Log("msg", "pusher failed to consume trace data", "err", err)
+
+		// Client disconnects are logged but not propogated back.
+		if errors.Is(err, context.Canceled) {
+			return nil
+		}
+
 		err = wrapErrorIfRetryable(err, r.retryDelay)
 	}
 
--- a/modules/frontend/slos.go
+++ b/modules/frontend/slos.go
@@ -1,6 +1,8 @@
 package frontend
 
 import (
+	"context"
+	"errors"
 	"net/http"
 	"time"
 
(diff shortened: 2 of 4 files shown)


modules/distributor/receiver/shim.go
```diff
@@ -342,6 +343,12 @@ func (r *receiversShim) ConsumeTraces(ctx context.Context, td ptrace.Traces) err
 	metricPushDuration.Observe(time.Since(start).Seconds())
 	if err != nil {
 		r.logger.Log("msg", "pusher failed to consume trace data", "err", err)
+
+		// Client disconnects are logged but not propogated back.
+		if errors.Is(err, context.Canceled) {
```
joe-elliott, line 348: this brings to mind a difficulty I have on the read path. it's impossible to tell where this context cancelled came from. Is it further up in the otel receiver code due to a client disconnect? or deeper down in the distributor code.

For instance, if we fail to write to 2+ ingesters due to this timeout I think that would bubble up as a context canceled as well:
https://github.com/grafana/tempo/blob/c41b078b13aa36f0b25ed67909efad262cdeb959/modules/distributor/distributor.go#L392-L393

`withCancelCause` was added in 1.20:

https://pkg.go.dev/context#WithCancelCause

to allow for communication of the reason, but I don't know if this is set correctly in the GRPC server. it's definitely not in our own code. Maybe we set it in our code and assume if there is no cause it's due to client disconnect?

we unfortunately cancel context in a lot of places and don't have good patterns for when, why or what is communicated when we do. as is, i think this would mask timeouts to the ingesters.

modules/frontend/slos.go
```diff
@@ -66,8 +68,10 @@ func sloHook(allByTenantCounter, withinSLOByTenantCounter *prometheus.CounterVec
 
 		// most errors are SLO violations
 		if err != nil {
-			// however, if this is a grpc resource exhausted error (429) then we are within SLO
-			if status.Code(err) == codes.ResourceExhausted {
+			// However these errors are considered within SLO:
+			// * grpc resource exhausted error (429)
+			// * context canceled (client disconnected or canceled)
+			if status.Code(err) == codes.ResourceExhausted || errors.Is(err, context.Canceled) {
```
joe-elliott, line 74: same thoughts here. maybe we just log the cancel cause if one exists and see if that's populated?

PR #3506, opened 2024-03-20
Add ingester ring shuffle sharding

kvrhdn:
**What this PR does**:
This adds the possibility to use ingester ring shuffle sharding to Tempo. This feature already exists in Mimir and Loki.
fyi docs from Mimir: https://grafana.com/docs/mimir/latest/configure/configure-shuffle-sharding/

How it works: a tenant can be configured to use a smaller ring with `ingestion_tenant_shard_size`. When set, the distributors will write to a subset of all ingesters, queriers will only query this subset. Different tenants will be randomly shuffled over the ingesters.

Advantages:
- resilience: an ingester going down will impact less tenants
- in highly multi-tenant cells tenants will be less scattered, instead of all ingesters ingesting a very small fraction of a tenant a subset of ingesters each can ingest more

Disadvantages:
- it adds complexity and management
- it's not possible to downscale a tenant without interrupting queries

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


kvrhdn:
Running a test in a test cluster and it's very responsive

Screenshot 2024-03-27 at 19 46 28

joe-elliott:
> You can lower shard size without issue down to 2 instances, 

With RF3 we should never set this value below 3, correct?

docs/sources/tempo/configuration/_index.md
```diff
@@ -1248,6 +1256,9 @@ overrides:
       # A value of 0 disables the check.
       [max_global_traces_per_user: <int> | default = 0]
 
+      # Shuffle sharding shards used for this user. A value of 0 uses all ingesters in the ring.
```
electron0zero, line 1259: let's leave a note that it should not be less then RF.

electron0zero: maybe we can do a runtime validation...

modules/querier/config.go
```diff
@@ -84,6 +87,7 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
 		},
 		DNSLookupPeriod: 10 * time.Second,
 	}
+	cfg.ShuffleShardingLookbackPeriod = 1 * time.Hour
```
electron0zero, line 90: we should be set to `query_ingesters_until` and some buffer for safety...

modules/distributor/distributor.go
```diff
@@ -388,7 +388,9 @@ func (d *Distributor) sendToIngestersViaBytes(ctx context.Context, userID string
 
 	var mu sync.Mutex
 
-	err := ring.DoBatch(ctx, op, d.ingestersRing, keys, func(ingester ring.InstanceDesc, indexes []int) error {
+	writeRing := d.ingestersRing.ShuffleShard(userID, d.overrides.IngestionTenantShardSize(userID))
```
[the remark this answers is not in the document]
kvrhdn: Yup, if shard size is 0 it just returns the ring, effectively a no-op.

https://github.com/grafana/dskit/blob/7bd314b462b234ffdcad12275efa6bcdd63d5be5/ring/ring.go#L665-L668

So it's safe to roll out this code as is. Setting the shard size should only be done if we have set `querier_shuffle_sharding_ingesters_enabled: false` though since lowering shard size means queriers will miss ingesters with data.

modules/querier/config.go
kvrhdn: Yep, I looked at this parameter. `query_ingesters_until` defaults to 30m so I thought 1hr would be safe. We can also default to a slightly higher value, the performance impact is small.

kvrhdn:
> > You can lower shard size without issue down to 2 instances,
> 
> With RF3 we should never set this value below 3, correct?

I think under normal operations setting shard size to 2 should be an error if you have 1) RF=3 and 2) 3 or more ingesters available. But while experimenting Tempo accepted it and it kept working fine. I'm not sure if the ring is just able to deal with 2 instances (like during an outage) or if it's related to our [eventually consistent ring](https://github.com/grafana/tempo/blob/main/pkg/ring/ring.go#L35).

I'm looking into adding validation to loading the overrides to just prevent this from happening.

electron0zero (approved):
code lgtm 👍🏼

electron0zero (approved):
code lgtm 👍🏼

PR #3515, opened 2024-03-26
Tryout otlp collector fix for retryable http errors

jesusvazquez:
**What this PR does**:

Tries out a collector version that includes https://github.com/open-telemetry/opentelemetry-collector/pull/9357 and adds two e2e tests to make sure that the receiver replies with retryable errors.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


jesusvazquez:
Waiting to see if the original PR can be merged upstream and then we just have to vendor the main collector repository instead of the fork.

integration/e2e/limits_test.go
```diff
@@ -90,6 +98,104 @@ func TestLimits(t *testing.T) {
...
+	httpClient := otlptracehttp.NewClient(
+		otlptracehttp.WithEndpoint(tempo.Endpoint(4318)),
+		otlptracehttp.WithInsecure(),
+		otlptracehttp.WithRetry(otlptracehttp.RetryConfig{Enabled: false}),
+	)
+	require.NoError(t, httpClient.Start(context.Background()))
+
+	httpErr := httpClient.UploadTraces(context.Background(), protoSpans)
+	assert.Error(t, httpErr)
+	require.Contains(t, httpErr.Error(), "retry-able request failure")
```
zalegrala, line 134: Would this fail on `main` using the non-forked receiver?

integration/e2e/limits_test.go
```diff
@@ -90,6 +98,104 @@ func TestLimits(t *testing.T) {
...
+			}
+
+			resp, err := http.DefaultClient.Do(req)
+			require.NoError(t, err)
+			defer func() { _ = resp.Body.Close() }()
+			bodyBytes, err := io.ReadAll(resp.Body)
+			require.NoError(t, err)
+			fmt.Println(string(bodyBytes))
+
+			assert.Equal(t, http.StatusTooManyRequests, resp.StatusCode)
```
zalegrala, line 194: Check my understanding, but we are sending multiple traces in one reqeust, and the config specifies a limit of one trace per user.  Should this fail with "TooManyRequests" since we are only making one request?

go.mod
```diff
@@ -348,3 +354,5 @@ replace (
 replace github.com/hashicorp/memberlist => github.com/grafana/memberlist v0.3.1-0.20220708130638-bd88e10a3d91
 
 replace golang.org/x/net => golang.org/x/net v0.17.0
+
+replace go.opentelemetry.io/collector/receiver/otlpreceiver => github.com/jpkrohling/opentelemetry-collector/receiver/otlpreceiver v0.0.0-20240320225130-6e76aa23e8d3
```
zalegrala, line 358: I didn't see the change mentioned merged into this fork, but perhaps I missed it.

jesusvazquez: It is actually vendoring a branch from this fork https://github.com/jpkrohling/opentelemetry-collector/commits/otlpreciever-http-response-code/ that has the commit in it.

integration/e2e/limits_test.go
zalegrala: Okay I see now, thanks for pointing that out.

zalegrala: It looks like the PR upstream has been merged.  Should we make the adjustment to this PR also?

integration/e2e/limits_test.go
jesusvazquez: The config limits fixture file used in this tests specifies the following https://github.com/mapno/tempo/blob/70bf4e253df931f68e747b267c496adaa8e77f6a/integration/e2e/config-limits.yaml#L20

```
overrides:
  defaults:
    ingestion:
      max_traces_per_user: 1
      rate_limit_bytes: 500
      burst_size_bytes: 500
    global:
      max_bytes_per_trace: 130
```

The test is batching up 10 traces
https://github.com/mapno/tempo/blob/5faf0f091584b25e43f57fdcb27e5f2da7cc342a/integration/e2e/limits_test.go#L146

So the quota is being hit and the status code returned is 429 which is known as Resource Exhausted but in go its also TooManyRequests https://github.com/golang/go/blob/master/src/net/http/status.go#L62

go.mod
jesusvazquez: It got merged so I'm vendoring the new changes.

zalegrala (approved):
Nice, thanks for the contribution.

Issue #3413, opened 2024-02-20
Tempo Query Frontend Refactor


joe-elliott:
The query frontend code has been pushed to its absolute limits. We need a refactor for improved code clarity, performance, enable streaming on all endpoints and pave the way for more advanced features demanded by TraceQL metrics.

The first PR has been opened and will address the first two tasks when merged: https://github.com/grafana/tempo/pull/3400

**Tasks**

- [x] Build tools to construct asynchronous pipelines
- [x] Migrate search to the new pipelines
- [x] Migrate tag endpoints to the new pipelines
- [x] Migrate trace id endpoints to the new pipelines. (remain synchronous but use new patterns)
- [x] Migrate metrics summary api to the new pipelines. (remain synchronous but use new patterns)
- [ ] Build and use structures to support deterministic search
- [x] Build experimental metrics streaming API

**Stretch**

- [x] Measure/Improve job throughput. At least don't slow things down.
- [ ] Issue multiple ingester search queries instead of one.

**Existing/Relevant Issues**

- https://github.com/grafana/tempo/issues/1392
- https://github.com/grafana/tempo/issues/3366
- Deterministic search
  - https://github.com/grafana/tempo/issues/2659
  - https://github.com/grafana/tempo/issues/3109

joe-elliott:
We discovered some goroutine leaks in the existing implementation. Fix here:

https://github.com/grafana/tempo/pull/3495

joe-elliott:
Tags streaming:

https://github.com/grafana/tempo/pull/3460

joe-elliott:
Multiple ingester queries are being handled here:

https://github.com/grafana/tempo/pull/3574

PR #3522, opened 2024-03-26
Tempo: Refactor Trace By ID and Metrics Summary

joe-elliott:
**What this PR does**:
Refactors 2 more endpoints to use the new frontend pipeline items. The final one will be the new query range endpoint. Trace by ID was more difficult than expected. It has a few unique features that make it harder to work with:

- Passes proto instead of json
- Changes the encoded object. All other paths propagate the same object from querier to outside
- Has a special middleware that occurs after recombination (deduper)
- Does not consider 404s an error

Testing is looking good for no regressions.

Other changes:
- Removes trace by id hedging. We have found it to be an antipattern.
- Fixes 429s being returned as 499s and adds an integration test
- Removes the addition of a tenant in multitenant trace id lookup. We are not using it and it's difficult to fit into the new pipeline. Ignoring for now.
- Reduce memory consumption in the frontend on trace by id lookup by pushing the deduper and content-type marshalling logic into the combiner. This prevents an unnecessary 2x marshalling/unmarshalling of a trace.

Works toward #3413 

**Checklist**
- [x] Tests updated
- [x] Documentation added


knylander-grafana (commented):
Doc updates look good. Thank you for updating the docs.

docs/sources/tempo/configuration/manifest.md
```diff
@@ -301,8 +301,6 @@ query_frontend:
         query_ingesters_until: 30m0s
     trace_by_id:
         query_shards: 50
```
electron0zero, line 303: nice, let's get it out 🚀 👨🏼‍🎤

modules/frontend/traceid_handlers.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/grafana/tempo/modules/overrides"
+	"github.com/grafana/tempo/pkg/api"
+)
+
+// newTraceIDHandler creates a http.handler for trace by id requests
+//
+// todo: the trace by id path consumes a lot of resources for large traces b/c it
+// repeatedly unmarshals/marshals the trace data. Note that it occurs once here
+// to marshal into the proto format and again in the deduper middleware. we should
+// collapse this into the combiner where the data is already unmarshalled
```
kvrhdn, line 23: I don't think this comment is up-to-date anymore? This function doesn't marshal/unmarshal anymore and the deduper directly accepts a `*tempopb.Trace` now.

kvrhdn (commented):
LGTM overall :+1: Left some minor comments

modules/frontend/combiner/trace_by_id_deduper.go
```diff
@@ -22,75 +16,29 @@ const (
...
-			return nil, err
-		}
-
-		return &http.Response{
-			StatusCode:    http.StatusOK,
-			Body:          io.NopCloser(bytes.NewReader(responseBytes)),
-			Header:        http.Header{},
-			ContentLength: resp.ContentLength,
-		}, nil
+func (s *spanIDDeduper) dedupe(trace *tempopb.Trace) *tempopb.Trace {
```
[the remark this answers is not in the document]
kvrhdn: We could even put it in a separate package now to isolate it a bit more but not a big deal 🤷🏻

kvrhdn (approved):
I'm fine with merging this :+1: just got some minor questions/issues, nothing blocking

modules/frontend/tenant.go
```diff
@@ -1,173 +1,19 @@
...
-
-const (
-	statusCodeLabel = "status_code"
-	tenantLabel     = "tenant"
 )
 
 var ErrMultiTenantUnsupported = errors.New("multi-tenant query unsupported")
 
-var (
-	tenantSuccessTotal = promauto.NewCounterVec(
```
electron0zero, line 31: looks like we removing these metrics, can we mention it as a breaking change. this might break dashboards and alerts for end users.

modules/frontend/combiner/trace_by_id_deduper.go
```diff
@@ -1,17 +1,11 @@
...
-	"bytes"
 	"encoding/binary"
 	"fmt"
-	"io"
-	"net/http"
 
-	"github.com/go-kit/log"
-	"github.com/golang/protobuf/proto" //nolint:all //deprecated
-	"github.com/opentracing/opentracing-go"
+	//nolint:all //deprecated
```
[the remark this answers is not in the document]
joe-elliott: agreed. i think i needed it until i removed the http elements from this file.

modules/frontend/traceid_handlers.go
joe-elliott: good call. removed.

modules/frontend/tenant.go
joe-elliott: yup, i am going to create a big changelog PR when i'm done and will include this detail

PR #3524, opened 2024-03-27
Add querier metrics for requests executed

electron0zero:
**What this PR does**:

Adds two metrics to provide observability into read path and query processing

1. `tempo_querier_worker_request_executed_total` is a counter, measuring the requests executed by the querier
2. log the concurrency of a querier

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/querier/worker/frontend_processor.go
+++ b/modules/querier/worker/frontend_processor.go
@@ -15,6 +15,8 @@ import (
 	"github.com/grafana/dskit/httpgrpc"
 	"github.com/grafana/tempo/pkg/util/httpgrpcutil"
 	"github.com/opentracing/opentracing-go"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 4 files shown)


modules/querier/worker/worker.go
```diff
@@ -242,6 +249,9 @@ func (w *querierWorker) resetConcurrency() {
 	if totalConcurrency > w.cfg.MaxConcurrentRequests {
 		level.Warn(w.log).Log("msg", "total worker concurrency is greater than promql max concurrency. Queries may be queued in the querier which reduces QOS")
 	}
+
+	// capture the current concurrency metric
+	metricConcurrency.Set(float64(totalConcurrency))
```
joe-elliott, line 254: why don't we just log this if we're concerned it's a different value then expected.

modules/querier/worker/frontend_processor.go
```diff
@@ -27,6 +29,12 @@ var processorBackoffConfig = backoff.Config{
 	MaxBackoff: 1 * time.Second,
 }
 
+var metricWorkerRequests = promauto.NewCounter(prometheus.CounterOpts{
```
joe-elliott, line 32: i don't mind this but do you expect it to be significantly different than the querier request rates?

electron0zero: yup, I think request rate is for batches, and this is at job level (single request). also, this is counter so we can capture how much work each querier is doing...

modules/querier/worker/worker.go
electron0zero: updated.

joe-elliott (approved):

PR #3530, opened 2024-03-28
Increase buckets in queryThroughput histogram

electron0zero:
**What this PR does**:

we noticed our production cluster that current buckets are not sufficient. so add more buckets for better observability. 

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott (approved):
should we just drop this metric and rely on log lines? i feel like we could chase this forever

electron0zero:
> should we just drop this metric and rely on log lines? i feel like we could chase this forever

agree on adding log lines, but I think there is value in metric as well. hopefully we can migrate to native histograms soon 🤞🏼

PR #3536, opened 2024-04-03
Feistel cipher implementation for trace ID distribution correction

mapno:
**What this PR does**:
Fesitel cipher implementation for trace ID distribution correction

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


joe-elliott:
The first question I'm going to ask is how perf compares to fnv1a. Is Feistel being explored b/c it's reversible? What are the situations in which we need to reverse?

PR #3537, opened 2024-04-03
Sharded compaction

mdisibio:
**What this PR does**:
This is the largest change to compaction in years.  It is a new sharding compactor that splits blocks up by trace ID divisions.  For example:  instead of having 2 blocks with the entire range of trace IDs 00..FF, the sharding compactor can combine and split them into 2 blocks with the ranges 00..80 and 80..FF where each block contains half of the range it did before.  

There are several benefits from this approach:
* The read path is more efficient.  Trace lookup can filter out more blocks faster by checking the Min/Max ID in the meta without having to resort to the bloom filter.  Metrics queries inspect fewer blocks to find the traces in the job shard.
* More effective deduplication of data in the backend.  Because the sharding compactor clusters traces with similar values together, it is more effective at finding and deduplicating the copies of a trace due to replication factor.  A recent internal analysis measured that compactors were only bringing the replication factor from 3 down to ~2.5.  The sharding compactor should achieve ~1.5 or better.

The shard count is configurable globally and per-tenant.  Setting to 2 or higher activates the new sharding compactor. 0 or 1 will fallback to the existing compactor.  Expecting useful values to be 2 through 8.  Values higher than 8 could be interesting, but at the cost of greatly increasing the size of the blocklist.

**Description**
How the sharding compactor works can be considered as 3 phases. 
1. The first phase is taking fresh blocks from the ingesters and splitting them.  The ingesters haven't been affected in this PR, and continue to flush blocks with the full range 00..FF.  The sharding compactor identifies them by looking for blocks with `CompactionLevel=0` and will split them.  This is a combination of combine/split like described at the time, so that up to 4 (maxInputBlocks) ingester blocks will be rewritten as N sharded blocks.  This step feeds work to the next step:
2. The second phase is the typical compaction/reduction of sharded blocks with other blocks in the same shard.  This is same as the existing compaction, but shard-aware.  The compactor identifies them by `CompactionLevel>0` and the block min/max trace IDs fall within the same shard (i.e. the block is "well-sharded")
3. A third phase exists that will upgrade/resplit older blocks too, but it is the lowest priority and only when the first 2 phases have no outstanding work.

Fixes n/a

**Checklist**
- [x] Tests updated
- [x] Documentation added


tempodb/encoding/vparquet3/block_iterator.go
```diff
@@ -75,6 +75,7 @@ func (i *rawIterator) Next(context.Context) (common.ID, parquet.Row, error) {
 	}
 
 	if errors.Is(err, io.EOF) {
+		i.pool.Put(rows[0])
```
mdisibio, line 78: This fixes a "memory leak" where the buffer wasn't returned to the pool.  This actually has a significant effect on memory because this isn't happening once at the end, but on every loop of the multiblock iterator.  Once 1 block is exhausted it begins leaking a pooled buffer on every call. The better fix is to make the multiblock iterator stop calling Next() once an iter is exhausted, but it was going to be a lot more involved.

tempodb/tempodb.go
```diff
@@ -545,13 +546,13 @@ func (rw *readerWriter) pollBlocklist() {
 }
 
 // includeBlock indicates whether a given block should be included in a backend search
-func includeBlock(b *backend.BlockMeta, _ common.ID, blockStart []byte, blockEnd []byte, timeStart int64, timeEnd int64) bool {
+func includeBlock(b *backend.BlockMeta, id common.ID, blockStart []byte, blockEnd []byte, timeStart int64, timeEnd int64) bool {
 	// todo: restore this functionality once it works. min/max ids are currently not recorded
 	//    https://github.com/grafana/tempo/issues/1903
 	//  correctly in a block
```
mapno, line 552: Is this still true?

mapno (commented):
Nice work!

modules/overrides/config_legacy.go
```diff
@@ -151,6 +152,7 @@ func (l *LegacyOverrides) toNewLimits() Overrides {
 		Compaction: CompactionOverrides{
 			BlockRetention:   l.BlockRetention,
 			CompactionWindow: l.CompactionWindow,
+			Shards:           l.CompactionShards,
```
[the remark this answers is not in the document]
mdisibio: Good catch.  Fixed.

tempodb/compactor_test.go
```diff
@@ -704,7 +725,7 @@ func makeTraceID(i int, j int) []byte {
 }
 
 func BenchmarkCompaction(b *testing.B) {
-	for _, enc := range encoding.AllEncodings() {
+	for _, enc := range /*encoding.AllEncodings()*/ []encoding.VersionedEncoding{vparquet3.Encoding{}} {
```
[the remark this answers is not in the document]
mdisibio: This is a benchmark run as needed, but good call.

tempodb/tempodb.go
mdisibio: No, it works now (fixed in #2867) which is good because traceql metrics and this new compactor rely on it :)  Removed the obsolete comment.

knylander-grafana (commented):
Thank you for adding some information to the configuration docs. Will we need additional information in the docs that talks about when to use this ocnfiguration?

tempodb/sharding_block_selector.go
```diff
@@ -0,0 +1,264 @@
...
+				b.TenantID, // Work sharding by tenant, window
+				window,     //
+				shardMin,   // and shard
+			)
+
+		case b.CompactionLevel == 0:
+			// Unsharded new block that needs to be split. This step generates work
+			// for the above step. MinBlocks=1 allows stray single blocks to still be split.
+			entry.split = true
+			entry.minBlocks = 1
```
joe-elliott, line 137: wdyt about trying 2 (or more) input blocks at compaction 0 split into the shards?

tempodb/sharding_block_selector.go
```diff
@@ -0,0 +1,264 @@
...
+				b.TenantID, // Work sharding by tenant and window
+				window,     //
+			)
+		}
+
+		s.entries = append(s.entries, entry)
+	}
+
+	// sort by group then order
+	sort.SliceStable(s.entries, func(i, j int) bool {
```
joe-elliott, line 173: is stable important?

joe-elliott (commented):
long overdue work. looks great. added some Qs but fine with things as is.

if you need my approval please ask. the only reason i'm not approving is b/c you and @mapno have been handling these PRs

tempodb/sharding_block_selector.go
mdisibio: That was the original behavior but if your ingesters flush an odd number it will leave behind unsplit blocks.   It happens more than expected after testing internally.

tempodb/sharding_block_selector.go
mdisibio: This was inherited from the original block selector.  Don't believe it is, and not seeing any clues from the original [commit](https://github.com/grafana/tempo/commit/9264916552304513c4586b2828e3e736fc5fa384#diff-094e7acc4b5e77da6729bc4f464aac009e471a8e78a3c3db82eda52cb1c1be35R133).

tempodb/encoding/common/interfaces.go
```diff
@@ -61,8 +61,14 @@ func DefaultSearchOptionsWithMaxBytes(maxBytes int) SearchOptions {
 	}
 }
 
+type CompactionRound interface {
```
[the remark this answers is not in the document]
mdisibio: This interface is the "command" sent to the block encodings in the `Compact` function.  It replaces `blocks []*backend.BlockMeta`.  Creating `tempodb/compaction` would mean the dependency chain `tempodb` -> `tempodb/encoding/...` -> `tempodb/compaction`.   The current location seems better since it is single direction.

tempodb/encoding/vparquet3/compactor.go
```diff
@@ -146,6 +147,18 @@ func (c *Compactor) Compact(ctx context.Context, l log.Logger, r backend.Reader,
 			return nil, fmt.Errorf("error iterating input blocks: %w", err)
 		}
 
+		// ship block to backend if done
+		if currentBlock != nil && cmd.CutBlock(currentBlock.meta, lowestID) {
+			currentBlockPtrCopy := currentBlock
+			currentBlockPtrCopy.meta.StartTime = minBlockStart
```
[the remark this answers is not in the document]
mdisibio: Hmm this block was moved from the bottom of the loop to the top so that the `CutBlock` callback can inspect the next trace (`lowestID`).  You're right it does seem to be out of date and the copy is no longer needed.  I'll take a look.

modules/querier/querier_query_range.go
```diff
@@ -64,24 +65,33 @@ func (q *Querier) queryBackend(ctx context.Context, req *tempopb.QueryRangeReque
...
+	// Get blocks that overlap this time range and shard
 	metas := q.store.BlockMetas(tenantID)
-	withinTimeRange := metas[:0]
+	matchingBlocks := metas[:0]
 	for _, m := range metas {
-		if m.StartTime.UnixNano() <= int64(req.End) && m.EndTime.UnixNano() > int64(req.Start) {
-			withinTimeRange = append(withinTimeRange, m)
+		if m.StartTime.UnixNano() <= int64(req.End) &&
+			m.EndTime.UnixNano() > int64(req.Start) &&
+			checkShard(m.MinID, m.MaxID) {
```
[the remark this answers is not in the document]
mdisibio: Agree, the current split between 8-byte and 16-byte handling is quite complicated, and it will be more straight-forward after trace ID changes.

mdisibio:
Hi everyone:  Putting this back to draft because it is not yet ready for large scale.  Testing it out in a large cluster is showing it difficult to achieve the balance between splitting and combining blocks, even with tuning.  Would like to review the switch statement and priorities of each group of blocks.    When there is an imbalance the result is that compactors are less effective and the blocklist increases more than acceptable:  i.e. If the compactor spends too much time splitting blocks then the backend is composed of (too) many small blocks, and if the compactor spends too much combining, then ingester level-0 blocks are neglected.

PR #3543, opened 2024-04-05
Surface new labels for uninstrumented services and systems

### LINKED ISSUE #3461: Label improvements for Service identification

The current reliance on the `connection_type` label values for the purpose of identifying if a node is un-instrumented is problematic. For example, its `virtual_node` value alone doesn't identify which node is the `virtual` one. A similar issue arises with the `database` and `messaging_system` values, or even if there's a confluence of any of those in the same mode.

In the case of `database` and `messaging_system`, further identification relies on inspection of resource attributes (`db.system` and `messaging.system`, for example).

Ideally:

1. In the cases where `connection_type` had the value of `virtual_node`, we’d like to generate a new label `virtual_node` with a value of either `client` or `server`.

2.1. In the cases where `connection_type` is of value `database`, then we'd like to surface either `client_db_system` or `server_db_system` labels, with the same value as the `db.system` resource attribute.

2.2. In the cases where `connection_type` is of value `messaging_system`, then surface either `client_messaging_system` or `server_messaging_system` labels, with the same value as the  `messaging.system` resource attribute.

We've also considered adding more values to `connection_type` itself, with the drawback of it not being enough to cover all cases of client/server and intersection of the previous values.

**Additional context**

Resources:

- You can follow discussion, comments, and links to original issues in this [DesignDoc](https://docs.google.com/document/d/1t3d4P7a6Yo_0Ly5zLpovC1rLqCCKj3G_SsrBuiMqijc/edit).

I have bandwidth to propose PR(s), as necessary.

t00mas:
**What this PR does**:

For the servicegraphs processor:
- Gates this feature behind the EnableExtraUninstrumentedServicesLabels flag in the Config
- If enabled, adds new labels to `traces_service_graph_request_*`:
  - `virtual_node` with a value of `client` or `server`, so the un-instrumented service's node can be identified.
  - `client|server_db|messaging_system` with the corresponding resource attribute value, to surface and identify the service and the node.

Fixes #3461 

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/configuration/_index.md
+++ b/docs/sources/tempo/configuration/_index.md
@@ -324,6 +324,9 @@ metrics_generator:
             # Attribute Key to multiply span metrics
             [span_multiplier_key: <string> | default = ""]
 
+            # Enables additional labels for services and virtual nodes.
+            [enable_extra_uninstrumented_services_labels: <bool> | default = false]
+
         span_metrics:
 
             # Buckets for the latency histogram in seconds.
--- a/docs/sources/tempo/configuration/_index.md
+++ b/docs/sources/tempo/configuration/_index.md
@@ -325,7 +325,7 @@ metrics_generator:
             [span_multiplier_key: <string> | default = ""]
 
             # Enables additional labels for services and virtual nodes.
-            [enable_extra_uninstrumented_services_labels: <bool> | default = false]
+            [enable_virtual_node_label: <bool> | default = false]
 
         span_metrics:
 
--- a/docs/sources/tempo/metrics-generator/service_graphs/_index.md
+++ b/docs/sources/tempo/metrics-generator/service_graphs/_index.md
@@ -73,13 +73,13 @@ Duration is measured both from the client and the server sides.
 
 Possible values for `connection_type`: unset, `virtual_node`, `messaging_system`, or `database`.
 
-Additional labels can be included using the `dimensions` configuration option, or the `enable_extra_uninstrumented_services_label` option.
+Additional labels can be included using the `dimensions` configuration option, or the `enable_virtual_node_label` option.
 
 Since the service graph processor has to process both sides of an edge,
 it needs to process all spans of a trace to function properly.
 If spans of a trace are spread out over multiple instances, spans are not paired up reliably.
 
-#### Activate `enable_extra_uninstrumented_services_label`
+#### Activate `enable_virtual_node_label`
 
 Activating this feature adds the following label and corresponding values:
 
--- a/docs/sources/tempo/metrics-generator/service_graphs/_index.md
+++ b/docs/sources/tempo/metrics-generator/service_graphs/_index.md
@@ -73,20 +73,16 @@ Duration is measured both from the client and the server sides.
 
 Possible values for `connection_type`: unset, `virtual_node`, `messaging_system`, or `database`.
 
-Additional labels can be included using the `dimensions` configuration option, or the `enable_extra_uninstrumented_services_labels` option.
+Additional labels can be included using the `dimensions` configuration option, or the `enable_extra_uninstrumented_services_label` option.
 
 Since the service graph processor has to process both sides of an edge,
 it needs to process all spans of a trace to function properly.
 If spans of a trace are spread out over multiple instances, spans are not paired up reliably.
 
-#### Activating enable_extra_uninstrumented_services_labels
+#### Activating enable_extra_uninstrumented_services_label
 
-The following labels, with the corresponding values, will be added if this feature is enabled:
+The following label, with the corresponding values, will be added if this feature is enabled:
 
 | Label                   | Possible Values             | Description                                                              |
 |-------------------------|-----------------------------|--------------------------------------------------------------------------|
 | virtual_node            | `unset`, `client`, `server` | Explicitly indicates the side that is uninstrumented                     |
-| client_db_system        | `unset`,  `<string>`        | Surfaces the `db.system` resource attribute of the client, if any        |
-| server_db_system        | `unset`,  `<string>`        | Surfaces the `db.system` resource attribute of the server, if any        |
-| client_messaging_system | `unset`,  `<string>`        | Surfaces the `messaging.system` resource attribute of the client, if any |
-| server_messaging_system | `unset`,  `<string>`        | Surfaces the `messaging.system` resource attribute of the server, if any |
--- a/docs/sources/tempo/metrics-generator/service_graphs/_index.md
+++ b/docs/sources/tempo/metrics-generator/service_graphs/_index.md
@@ -79,9 +79,9 @@ Since the service graph processor has to process both sides of an edge,
 it needs to process all spans of a trace to function properly.
 If spans of a trace are spread out over multiple instances, spans are not paired up reliably.
 
-#### Activating enable_extra_uninstrumented_services_label
+#### Activate `enable_extra_uninstrumented_services_label`
 
-The following label, with the corresponding values, will be added if this feature is enabled:
+Activating this feature adds the following label and corresponding values:
 
 | Label                   | Possible Values             | Description                                                              |
 |-------------------------|-----------------------------|--------------------------------------------------------------------------|
--- a/docs/sources/tempo/metrics-generator/service_graphs/_index.md
+++ b/docs/sources/tempo/metrics-generator/service_graphs/_index.md
@@ -71,10 +71,22 @@ The following metrics are exported:
 
 Duration is measured both from the client and the server sides.
 
-Possible values for `connection_type`: unset, `messaging_system`, or `database`.
+Possible values for `connection_type`: unset, `virtual_node`, `messaging_system`, or `database`.
 
-Additional labels can be included using the `dimensions` configuration option.
+Additional labels can be included using the `dimensions` configuration option, or the `enable_extra_uninstrumented_services_labels` option.
 
 Since the service graph processor has to process both sides of an edge,
 it needs to process all spans of a trace to function properly.
 If spans of a trace are spread out over multiple instances, spans are not paired up reliably.
+
+#### Activating enable_extra_uninstrumented_services_labels
+
+The following labels, with the corresponding values, will be added if this feature is enabled:
+
+| Label                   | Possible Values             | Description                                                              |
+|-------------------------|-----------------------------|--------------------------------------------------------------------------|
+| virtual_node            | `unset`, `client`, `server` | Explicitly indicates the side that is uninstrumented                     |
+| client_db_system        | `unset`,  `<string>`        | Surfaces the `db.system` resource attribute of the client, if any        |
+| server_db_system        | `unset`,  `<string>`        | Surfaces the `db.system` resource attribute of the server, if any        |
+| client_messaging_system | `unset`,  `<string>`        | Surfaces the `messaging.system` resource attribute of the client, if any |
+| server_messaging_system | `unset`,  `<string>`        | Surfaces the `messaging.system` resource attribute of the server, if any |
--- a/modules/generator/processor/servicegraphs/config.go
+++ b/modules/generator/processor/servicegraphs/config.go
@@ -41,7 +41,7 @@ type Config struct {
 	SpanMultiplierKey string `yaml:"span_multiplier_key"`
 
 	// EnableExtraUninstrumentedServicesLabels enables additional labels for uninstrumented services
-	EnableExtraUninstrumentedServicesLabels bool `yaml:"enable_extra_uninstrumented_services_labels"`
+	EnableVirtualNodeLabel bool `yaml:"enable_virtual_node_label"`
 }
 
 func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
--- a/modules/generator/processor/servicegraphs/config.go
+++ b/modules/generator/processor/servicegraphs/config.go
@@ -40,7 +40,7 @@ type Config struct {
 	// If enabled attribute value will be used for metric calculation
 	SpanMultiplierKey string `yaml:"span_multiplier_key"`
 
-	// EnableExtraUninstrumentedServicesLabels enables additional labels for uninstrumented services
+	// EnableVirtualNodeLabel enables additional labels for uninstrumented services
 	EnableVirtualNodeLabel bool `yaml:"enable_virtual_node_label"`
 }
 
--- a/modules/generator/processor/servicegraphs/config.go
+++ b/modules/generator/processor/servicegraphs/config.go
@@ -39,6 +39,9 @@ type Config struct {
 
 	// If enabled attribute value will be used for metric calculation
 	SpanMultiplierKey string `yaml:"span_multiplier_key"`
+
+	// EnableExtraUninstrumentedServicesLabels enables additional labels for uninstrumented services
+	EnableExtraUninstrumentedServicesLabels bool `yaml:"enable_extra_uninstrumented_services_labels"`
 }
 
 func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
--- a/modules/generator/processor/servicegraphs/servicegraphs.go
+++ b/modules/generator/processor/servicegraphs/servicegraphs.go
@@ -190,6 +190,17 @@ func (p *Processor) consume(resourceSpans []*v1_trace.ResourceSpans) (err error)
 							e.ServerService = dbName
 							e.ServerLatencySec = spanDurationSec(span)
 						}
+
+						if p.Cfg.EnableExtraUninstrumentedServicesLabels {
+							switch e.ConnectionType {
+							case store.Database:
+								e.ServerDbSystem = e.ServerService
+							case store.MessagingSystem:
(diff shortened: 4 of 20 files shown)


modules/generator/processor/servicegraphs/servicegraphs_test.go
```diff
@@ -228,6 +228,87 @@ func TestServiceGraphs_virtualNodes(t *testing.T) {
...
+	})
+
+	serverMsgSystemLabels := labels.FromMap(map[string]string{
+		"client":                  "mythical-requester",
+		"connection_type":         "messaging_system",
+		"server":                  "mythical-recorder",
+		"server_messaging_system": "rabbitmq",
+	})
+
+	fmt.Println(testRegistry)
```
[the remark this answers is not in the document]
t00mas: definitely not needed

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -190,6 +190,17 @@ func (p *Processor) consume(resourceSpans []*v1_trace.ResourceSpans) (err error)
 							e.ServerService = dbName
 							e.ServerLatencySec = spanDurationSec(span)
 						}
+
+						if p.Cfg.EnableExtraUninstrumentedServicesLabels {
+							switch e.ConnectionType {
```
domasx2, line 195: this is mutually exclusive, but is it possible that  a service can be both database and messaging system? eg Redis

t00mas: it is - because at this point it has been decided what the service is acting as, depending on the presence/absence of the `db.name` resource attribute

modules/generator/processor/servicegraphs/store/edge.go
```diff
@@ -27,6 +27,11 @@ type Edge struct {
 	// Additional dimension to add to the metrics
 	Dimensions map[string]string
 
+	// Additional labels for uninstrumented services
+	VirtualNode                           string // Indicates which service is the virtual node (client or server)
```
joe-elliott, line 31: instead of making these individual strings can we add them to the dimensions map above? and just allow the normal code to add them to the metrics?

modules/generator/processor/servicegraphs/config.go
```diff
@@ -39,6 +39,9 @@ type Config struct {
 
 	// If enabled attribute value will be used for metric calculation
 	SpanMultiplierKey string `yaml:"span_multiplier_key"`
+
+	// EnableExtraUninstrumentedServicesLabels enables additional labels for uninstrumented services
+	EnableExtraUninstrumentedServicesLabels bool `yaml:"enable_extra_uninstrumented_services_labels"`
```
joe-elliott, line 44: similar thought to the below on the edge struct. i'm wondering if there's a way to safely encode this information in the `Dimensions` field above?

joe-elliott (commented):
a few initial questions. other thoughts:

- what occurs if there are conflicts between these new values and configured span dimensions?
- what do you think about pooling edges in [the store](https://github.com/grafana/tempo/blob/main/modules/generator/processor/servicegraphs/store/store.go)?

modules/generator/processor/servicegraphs/store/edge.go
t00mas: Definitely can be done. I've now added them as Dimensions if the feature is enabled.

modules/generator/processor/servicegraphs/config.go
t00mas: Encoding this fully into `Dimensions` for this struct is more tricky.

On one side, `virtual_node` is not a resource attribute anywhere, so encoding that into `Dimensions` could be counter-intuitive.

Secondly, while the other four are resource attributes, encoding them here would mean they also need to enable the client/server prefixes for them.

Let me know what you think.

(docs updated too)

t00mas:
> a few initial questions. other thoughts:
> 
> * what occurs if there are conflicts between these new values and configured span dimensions?
> * what do you think about pooling edges in [the store](https://github.com/grafana/tempo/blob/main/modules/generator/processor/servicegraphs/store/store.go)?

Thanks for reviewing Joe.

I've added some testing to cover "conflicts" with manually added `Dimensions`.

Could you be more specific about the edge pooling? Happy to address that if you think there's an issue there. (Sorry, I'm just new to Tempo in general)

mapno (commented):
I wonder if we should make this a per-tenant override instead. If this feature is behind a global config parameter, it makes me think that we should just add it to the default generator without a flag. What's the use case for enabling or disabling it globally?

t00mas:
> I wonder if we should make this a per-tenant override instead. If this feature is behind a global config parameter, it makes me think that we should just add it to the default generator without a flag. What's the use case for enabling or disabling it globally?

It actually makes more sense as a per-tenant override instead, yes.

I've also discussing with Joe and I will:
- Scope-down the PR just to add the new `virtual_node` label (the others can be added with a combination of Dimensions and Prefixes)
- Pool edges in the store for performance reasons

docs/sources/tempo/metrics-generator/service_graphs/_index.md
```diff
@@ -71,10 +71,18 @@ The following metrics are exported:
...
-Additional labels can be included using the `dimensions` configuration option.
+Additional labels can be included using the `dimensions` configuration option, or the `enable_extra_uninstrumented_services_label` option.
 
 Since the service graph processor has to process both sides of an edge,
 it needs to process all spans of a trace to function properly.
 If spans of a trace are spread out over multiple instances, spans are not paired up reliably.
+
+#### Activating enable_extra_uninstrumented_services_label
+
+The following label, with the corresponding values, will be added if this feature is enabled:
```
knylander-grafana, line 84: We try to use non-gerund (-ing verbs) in headings and avoid passive voice. What do you think about this suggestion?

```suggestion
#### Activate `enable_extra_uninstrumented_services_label`

Activating this feature adds the following label and corresponding values:
```

knylander-grafana (commented):
Doc changes look good. Thank you for adding docs!

docs/sources/tempo/metrics-generator/service_graphs/_index.md
t00mas: Sounds good, thank you!

t00mas:
These are some benchmarks about `Edge` reuse with pooling:

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/tempo/modules/generator/processor/servicegraphs
```

|                   | 20_runs_bench.txt | 20_runs_bench_with_pooling.txt      |
| ----------------- | ----------------- | ----------------------------------- |
|                   | sec/op            | sec/op     vs base                  |
| ServiceGraphs-11  | 21.33µ ± 0%       | 21.63µ ± 2%  +1.41% (p=0.006 n=20)  |
|                   | B/op              | B/op      vs base                   |
| ServiceGraphs-11  | 24.89Ki ± 0%      | 24.06Ki ± 0%  -3.35% (p=0.000 n=20) |
|                   | allocs/op         | allocs/op   vs base                 |
| ServiceGraphs-11  | 450.0 ± 0%        | 443.0 ± 0%  -1.56% (p=0.000 n=20)   |

There's probably better numbers to be had at scale, since these are limited to the testdata size.

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -86,8 +86,20 @@ type Processor struct {
 
 func New(cfg Config, tenant string, registry registry.Registry, logger log.Logger) gen.Processor {
 	labels := []string{"client", "server", "connection_type"}
+
+	if cfg.EnableVirtualNodeLabel {
```
joe-elliott, line 90: why don't we add this to the list of labels above? this will prevent us from having to add lines 96-102 which adds special handling for this value.

then we could add a new param to `onComplete` where we pass the string from `onExpire`

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -86,8 +86,20 @@ type Processor struct {
...
+		cfg.Dimensions = append(cfg.Dimensions, "virtual_node")
```
joe-elliott, line 91: should we make this slightly harder to conflict? double prefixed underscore (`__virtual_node`) seems like a common prometheus practice.

should we make this value a const?

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -301,12 +324,22 @@ func (p *Processor) onExpire(e *store.Edge) {
 		// We check if the span we have is the root span, and if so, we set the client service to "user".
 		if _, parentSpan := parseKey(e.Key()); len(parentSpan) == 0 {
 			e.ClientService = "user"
+
+			if p.Cfg.EnableVirtualNodeLabel {
+				e.Dimensions["virtual_node"] = "client"
+			}
```
joe-elliott, line 330: does this add any new info that just checking the client service label wouldn't have otherwise? i.e. if clientservice = user then we know what value virtual node would have had?

modules/generator/processor/servicegraphs/store/edge.go
```diff
@@ -37,12 +37,22 @@ type Edge struct {
...
+func zeroStateEdge(e *Edge) {
+	e.TraceID = ""
+	e.ConnectionType = Unknown
+	e.ServerService = ""
+	e.ClientService = ""
+	e.ServerLatencySec = 0
+	e.ClientLatencySec = 0
+	e.Failed = false
+	for k := range e.Dimensions {
+		// saves 30ns/op, 50 B/op, 1 allocs/op
```
joe-elliott, line 51: use the `clear(<map>)` func?

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -301,12 +324,22 @@ func (p *Processor) onExpire(e *store.Edge) {
...
 		}
 	} else if len(e.ServerService) == 0 && len(e.PeerNode) > 0 {
 		// If client span does not have its matching server span, but has a peer attribute present,
 		// we make the assumption that a call was made to an external service, for which Tempo won't receive spans.
 		e.ServerService = e.PeerNode
+
+		if p.Cfg.EnableVirtualNodeLabel {
+			e.Dimensions["virtual_node"] = "server"
+		}
+
```
joe-elliott, line 342: `onExpire` needs to call `ReturnEdge` when it doesn't call onComplete.

modules/generator/processor/servicegraphs/servicegraphs.go
t00mas: definitely doable - however what you save in this lines to avoid the prefixes, is added back to the onComplete logic for slice allocation and label value assignment.

also, it'll make this new label behave as half-dimension half-edge-property. do you think that'd be ok?

can you also explain further what's the modification of the onComplete signature for?

modules/generator/processor/servicegraphs/servicegraphs.go
t00mas: sounds good, I'm adding this

modules/generator/processor/servicegraphs/servicegraphs.go
t00mas: at the moment no, but I'd vouch for adding it anyway because:

1. label consistency for the new label: since adding `server` adds new info, by not adding `client` too we can end up in a strange situation where you need a different query for filtering (something like `{__virtual_node: server}` vs. `{__virtual_node: "", client: user}`)
2. it looks like the client service label could be improved in the future as well (the current value of `user` may not be accurate for all cases, but a best guess)

modules/generator/processor/servicegraphs/store/edge.go
t00mas: can do - similar benchmarks now, maybe also better at volume anyway

modules/generator/processor/servicegraphs/servicegraphs.go
t00mas: adding this - it doesn't seem to have coverage though, so I'll look into that too soon

modules/generator/processor/servicegraphs/store/edge.go
joe-elliott: probably no perf change, but i prefer it for clarity. 

also maybe some future perf improvements would make it into `clear()` that wouldn't be in the loop?

modules/generator/processor/servicegraphs/servicegraphs.go
joe-elliott: yup, i see. i was trying to remove some of the state wrangling in the `consume()` method but you've already completely removed it.

agree that the tradeoffs mentioned:

> what you save in this lines to avoid the prefixes, is added back to the onComplete logic for slice allocation and label value assignment.

are not worth it. let's keep it as implemented now.

> also, it'll make this new label behave as half-dimension half-edge-property. do you think that'd be ok?

i'm not seeing an issue with that. do you have some concerns in mind?

joe-elliott (approved):
Docs update, feature addition, a perf improvement. Love it. Thank you!

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -52,6 +52,8 @@ const (
 	metricRequestClientSeconds = "traces_service_graph_request_client_seconds"
 )
 
+const virtualNodeLabel = "__virtual_node"
```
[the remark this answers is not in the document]
joe-elliott: I asked for this. Should we drop it? Just trying to avoid conflicts with actual span attributes.

docs/sources/tempo/configuration/_index.md
```diff
@@ -324,6 +324,9 @@ metrics_generator:
             # Attribute Key to multiply span metrics
             [span_multiplier_key: <string> | default = ""]
 
+            # Enables additional labels for services and virtual nodes.
+            [enable_extra_uninstrumented_services_labels: <bool> | default = false]
```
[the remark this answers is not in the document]
t00mas: Thanks! Amending now

modules/generator/processor/servicegraphs/store/edge.go
```diff
@@ -37,12 +37,19 @@ type Edge struct {
...
 
-func newEdge(key string, ttl time.Duration) *Edge {
-	return &Edge{
-		key:        key,
-		Dimensions: make(map[string]string),
-		expiration: time.Now().Add(ttl).Unix(),
-	}
+// zeroStateEdge resets the Edge to its zero state.
+// Useful for reusing an Edge without allocating a new one.
+func zeroStateEdge(e *Edge) {
```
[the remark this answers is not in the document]
t00mas: `clear` sounds good to me

modules/generator/processor/servicegraphs/servicegraphs.go
t00mas: ok agree, leaving it behaving as before, as a dimension.

I think it's better to "treat it" as if it's fully one thing or the other. What I mean by that, from a future reader standpoint, it'll be easier to follow logic if it's treated as a dimension or as a struct property (first-class label?)

modules/generator/processor/servicegraphs/servicegraphs.go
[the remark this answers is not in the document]
t00mas: I'll be OK with and without the `__` prefix

modules/generator/processor/servicegraphs/store/interface.go
```diff
@@ -8,4 +8,8 @@ type Store interface {
 	UpsertEdge(key string, update Callback) (isNew bool, err error)
 	// Expire evicts expired edges from the store.
 	Expire()
+	// GrabEdge returns an edge from the store, creating a new one if it doesn't exist.
+	GrabEdge(key string) *Edge
+	// ReturnEdge returns an edge to the store.
+	ReturnEdge(e *Edge)
```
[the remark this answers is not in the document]
t00mas: I gave it a try, and it's definitely possible. With the new changes there's no interface mod needed as everything is now kept as the store's inner logic.
I think I've covered all the `returnEdge` paths and the benchmarks reach the same allocs/op gains as before, but if you see an edge case I haven't covered please let me know.

modules/generator/processor/servicegraphs/store/store.go
```diff
@@ -61,11 +61,13 @@ func (s *store) tryEvictHead() bool {
 
 	headEdge := head.Value.(*Edge)
 	if !headEdge.isExpired() {
+		s.returnEdge(headEdge)
```
mapno, line 64:
```suggestion
```

The edge is not expired here, so it shouldn't be returned to the pool.

mapno (changes requested):
I'm good with the PR, except that one comment. Nice work!

modules/generator/processor/servicegraphs/servicegraphs.go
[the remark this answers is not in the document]
t00mas: OK - then to not raise any questions to future readers, and because it seems a collision is unlikely, I'll leave it as **without** the `__` prefix.

It'd be good in the future to identify and sanitize in the same way as in spanmetrics, happy to work on that small improvement in a future PR if you need me.

mapno (approved):

PR #3548, opened 2024-04-08
clear metrics generator wal on start up

ie-pham:
**What this PR does**: Set metrics generator to clear its wal on start up.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/generator.go
+++ b/modules/generator/generator.go
@@ -66,13 +66,7 @@ func New(cfg *Config, overrides metricsGeneratorOverrides, reg prometheus.Regist
 		return nil, ErrUnconfigured
 	}
 
-	// clean the wal before everything
-	err := os.RemoveAll(cfg.Storage.Path)
-	if err != nil {
-		level.Warn(logger).Log("msg", "failed to remove wal on start up")
-	}
-
-	err = os.MkdirAll(cfg.Storage.Path, os.ModePerm)
+	err := os.MkdirAll(cfg.Storage.Path, os.ModePerm)
 	if err != nil {
 		return nil, fmt.Errorf("failed to mkdir on %s: %w", cfg.Storage.Path, err)
 	}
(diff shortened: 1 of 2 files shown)


modules/generator/generator.go
```diff
@@ -113,6 +113,9 @@ func New(cfg *Config, overrides metricsGeneratorOverrides, reg prometheus.Regist
 }
 
 func (g *Generator) starting(ctx context.Context) (err error) {
+	// clean the wal on start up
+	os.RemoveAll(g.cfg.Storage.Path)
```
joe-elliott, line 117: even if we don't fail on the error, we should log it if it's non-nil

joe-elliott (commented):
a simple enough change with hopefully a big impact. add a changelog please.

one other request:

i do not know the startup logic for the prom rw clients and if putting this code in `starting` could have a negative impact. can you verify that the prom clients are not initialized until after this line? if you have any concern you could move it to the new method.

modules/generator/storage/instance.go
```diff
@@ -57,12 +57,19 @@ func New(cfg *Config, o Overrides, tenant string, reg prometheus.Registerer, log
 
 	walDir := filepath.Join(cfg.Path, tenant)
 
+	// clean the wal before everything
+	level.Info(logger).Log("msg", "clearing old WAL on start up", "dir", walDir)
```
joe-elliott, line 61: any reason not to just do this once in generator.go instead of per tenant?

ie-pham: when I moved it to the very first line of code that touches the generator, the test kept failing because in the config file, we set the wal path to `/var/tempo` which is the same path for the other components -- and this would remove other folders created by other components. We normally set our walp ath to `/var/tempo/wal` but I was thinking... what if there are users out there that store their wal folders in the same path as their other components. Clearing the entire folder will nuke everything else they might have stored there so it will be safer to remove exactly only what we want to remove.

joe-elliott: +1 good call

joe-elliott (approved):

PR #3557, opened 2024-04-10
Add spss and limit to search request hash

joe-elliott:
**What this PR does**:
Currently if you repeat a search with different limits and spans per spanset values it will potentially grab incorrect results from cache. This PR adds limit and spss to the cache key to make the job goes through to the backend in this case.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


ie-pham (approved):
LGTM

PR #3560, opened 2024-04-10
Ensure poller only runs on the frontend for the frontend target

zalegrala:
**What this PR does**:

Without this change, SSB mode will end up running multiple pollers per instance; one for the compactor, and one for the frontend.  The frontend only needs the poller when we are running it with the QueryFrontend target.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott (approved):
nice fix!

PR #3574, opened 2024-04-15
Add support for ingester query shards

zalegrala:
**What this PR does**:

This change allows the search queries against the ingesters to be broken into smaller queries based on the time span.  

**Checklist**
- [x] Tests updated
- [x] Documentation added


modules/frontend/search_sharder.go
```diff
@@ -93,26 +94,26 @@ func (s asyncSearchSharder) RoundTrip(r *http.Request) (pipeline.Responses[*http
 		return pipeline.NewBadRequest(fmt.Errorf("range specified by start and end exceeds %s. received start=%d end=%d", maxDuration, searchReq.Start, searchReq.End)), nil
 	}
 
-	// build request to search ingester based on query_ingesters_until config and time range
+	// buffer of shards+1 allows us to insert ingestReq and metrics
+	reqCh := make(chan *http.Request, s.cfg.IngesterShards+1)
```
zalegrala, line 98: In the case the search range is out of bounds for the ingesters, we will have created the channel depth larger than was required.  This is probably fine.

modules/frontend/search_sharder.go
```diff
@@ -225,13 +225,42 @@ func (s *asyncSearchSharder) ingesterRequest(ctx context.Context, tenantID strin
...
 	searchReq.Start = ingesterStart
 	searchReq.End = ingesterEnd
 
-	return buildIngesterRequest(ctx, tenantID, parent, &searchReq)
+	// Split the start and end range into sub requests for each range.
+	shards := s.cfg.IngesterShards
+	duration := searchReq.End - searchReq.Start
+	interval := duration / uint32(s.cfg.IngesterShards)
+	// If the interval is less than a minute, dont' shard
+	if interval < 60 {
```
zalegrala, line 239: I'm not sure if this is right.  Perhaps we force-reduce the shard count until the interval is above a minute.  Using 1 minute here as a notion of "doing too much work" on overlapping blocks.  Should this be configurable?

modules/frontend/search_sharder.go
zalegrala: This does make we want to consume from the channel right away, but I think its outside the scope of this PR.

modules/frontend/config.go
```diff
@@ -69,6 +69,7 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
 			MaxDuration:           168 * time.Hour, // 1 week
 			ConcurrentRequests:    defaultConcurrentRequests,
 			TargetBytesPerRequest: defaultTargetBytesPerRequest,
+			IngesterShards:        1,
```
zalegrala, line 72: I thought about making the shards per-tenant, but I think this doesn't keep the query fairness.  This is probably fine for now, but we can discuss.

modules/frontend/search_sharder.go
```diff
@@ -201,18 +202,17 @@ func (s *asyncSearchSharder) backendRequests(ctx context.Context, tenantID strin
 // that covers the ingesters. If nil is returned for the http.Request then there is no ingesters query.
 // since this function modifies searchReq.Start and End we are taking a value instead of a pointer to prevent it from
 // unexpectedly changing the passed searchReq.
-func (s *asyncSearchSharder) ingesterRequest(ctx context.Context, tenantID string, parent *http.Request, searchReq tempopb.SearchRequest) (*http.Request, error) {
+func (s *asyncSearchSharder) ingesterRequests(ctx context.Context, tenantID string, parent *http.Request, searchReq tempopb.SearchRequest, reqCh chan *http.Request) error {
 	// request without start or end, search only in ingester
 	if searchReq.Start == 0 || searchReq.End == 0 {
-		return buildIngesterRequest(ctx, tenantID, parent, &searchReq)
+		return buildIngesterRequest(ctx, tenantID, parent, &searchReq, reqCh)
```
electron0zero, line 208: this will still build one request, should be okay imo but wanted to point it out.

electron0zero: this is from past, where we allowed requests without start and end.

modules/frontend/search_sharder.go
```diff
@@ -225,13 +225,54 @@ func (s *asyncSearchSharder) ingesterRequest(ctx context.Context, tenantID strin
...
 	if ingesterStart == ingesterEnd {
-		return nil, nil
+		return nil
 	}
 
 	searchReq.Start = ingesterStart
 	searchReq.End = ingesterEnd
 
-	return buildIngesterRequest(ctx, tenantID, parent, &searchReq)
+	// Split the start and end range into sub requests for each range.
```
electron0zero, line 234: I was thinking if we can make this code simpler by making the min interval a fixed size??

I was thinking, something like this??

image

modules/frontend/search_sharder_test.go
```diff
@@ -332,101 +333,198 @@ func TestBackendRequests(t *testing.T) {
...
+		// start/end when entirely within ingeste search window, but check that we don't shard too much.
+		{
+			request:             "/?tags=foo%3Dbar&minDuration=11ms&maxDuration=30ms&limit=50&start=" + strconv.Itoa(ago("15m")) + "&end=" + strconv.Itoa(ago("0s")),
+			queryIngestersUntil: 5*time.Minute + 10*time.Second,
+			expectedURI: []string{
+				"/querier?end=" + strconv.Itoa(ago("4m8s")) + "&limit=50&maxDuration=30ms&minDuration=11ms&spss=3&start=" + strconv.Itoa(ago("5m10s")) + "&tags=foo%3Dbar",
+				"/querier?end=" + strconv.Itoa(ago("3m6s")) + "&limit=50&maxDuration=30ms&minDuration=11ms&spss=3&start=" + strconv.Itoa(ago("4m8s")) + "&tags=foo%3Dbar",
+				"/querier?end=" + strconv.Itoa(ago("2m4s")) + "&limit=50&maxDuration=30ms&minDuration=11ms&spss=3&start=" + strconv.Itoa(ago("3m6s")) + "&tags=foo%3Dbar",
+				"/querier?end=" + strconv.Itoa(ago("1m2s")) + "&limit=50&maxDuration=30ms&minDuration=11ms&spss=3&start=" + strconv.Itoa(ago("2m4s")) + "&tags=foo%3Dbar",
+				"/querier?end=" + strconv.Itoa(ago("0s")) + "&limit=50&maxDuration=30ms&minDuration=11ms&spss=3&start=" + strconv.Itoa(ago("1m2s")) + "&tags=foo%3Dbar",
```
electron0zero, line 447: good tests 👍🏼

modules/frontend/search_sharder.go
zalegrala: Good call out.  I think let's follow up with a breaking change to validate that we pass a start/end and remove this chunk.

modules/frontend/search_sharder.go
zalegrala: https://github.com/grafana/tempo/pull/3574/commits/b8906ab6191d3a4d4310d267771d1b0e472762d4

I've pushed the above for this request.

.golangci.yml
```diff
@@ -63,6 +63,8 @@ linters-settings:
     exclude: ./.errcheck-exclude.txt
   errorlint:
     errorf: false # don't check whether fmt.Errorf uses %w to wrap errors
+  goconst:
```
joe-elliott, line 66: didn't know you could do this. i humorously included the same change in a different way here:

https://github.com/grafana/tempo/pull/3584/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9R77-R81

i'm going to change mine to match yours. it does mean one of us may have to resolve the conflict.

modules/frontend/search_sharder.go
```diff
@@ -225,13 +225,46 @@ func (s *asyncSearchSharder) ingesterRequest(ctx context.Context, tenantID strin
...
+	duration := searchReq.End - searchReq.Start
+	interval := duration / uint32(s.cfg.IngesterShards)
```
joe-elliott, line 236: i think this will underscan by a few seconds due to the flooring of integer division.

maybe make the break condition the for loop `shardStart > ingesterEnd` instead of ingester shards? a lot of fixes for this.

zalegrala: Thanks for calling this out.  The tests have been updated and verified against this case.

knylander-grafana (commented):
Thank you for adding updates to the doc.

joe-elliott (approved):

PR #3579, opened 2024-04-16
Compactor tweaks:  Fix pool leak when iter exhausted and async pages for better throughput

mdisibio:
**What this PR does**:
Two small improvements to compactors that I noticed in #3537 .  Bringing them forward since it is currently paused for a bit, and the changes are very small.

Async pages improves throughput by ~40% in our internal cluster. Below are screenshots showing total output traces and data rate.  Benchmarks don't show this since there is no latency like object storage.  (Aside, the benchmarks could use some work but skipping for now).

image

image

The pooling leak is mentioned [here](https://github.com/grafana/tempo/pull/3537#discussion_r1551637156)  It only happens when compacting blocks of different trace counts. The benchmarks are also lacking this (they compact blocks of identical counts).  Small but useful reduction in memory.

image
 

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott (approved):
changelog?

PR #3584, opened 2024-04-17
Query Frontend Refactor: Metrics Query Range

### LINKED ISSUE #3413: Tempo Query Frontend Refactor

joe-elliott:
**What this PR does**:
Moves the final frontend endpoint (metrics query range) over to the new streaming pipeline. Note that the streaming is done in a naive fashion. We simply return whatever is available every 500ms. After this is integrated into a frontend we will revisit and determine what a good streaming pattern will be.

**Open Questions**
- Should we add a new SLO config and SLO metric explicitly for metrics? Currently it piggybacks on search.
- Should we reorg the code layout to be around pipeline? i.e. should we bring the combiner/sharder/handlers all together in modules? They are starting to have dependencies that would be easier to see if they were co-located.

**Other changes:**
- Refactors the pipeline to return an interface instead of *http.Response. The interface allows additional data to accompany the returned result.
  - Adds the ability for a request to add a value into the context. This value is then returned with its associated response. This is used to return a per job sampling rate to the combiner.
- Drops prom compatibility support
- Adds support in the CLI for the new endpoints
- Updates the local docker-compose example to support metrics queries
- Adds guard code in the frontend to prevent invalid config options
- Removes a number of unused files such as the no-op combiner and the old multitenant sharder
- Added total jobs and completed jobs to the metrics query range response
- Excludes goconst linter from _test.go files.

Fixes #3413 

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/api/http.go
```diff
@@ -337,6 +335,13 @@ func ParseQueryRangeRequest(r *http.Request) (*tempopb.QueryRangeRequest, error)
 	req.Query = r.Form.Get("query")
 	req.QueryMode = r.Form.Get(QueryModeKey)
 
+	// if we don't have a query then let's attempt to parse them out of the url
+	//  Grafana posts these params as form data, but BuildQueryRangeRequest below puts them in the url
+	if len(req.Query) == 0 && len(req.QueryMode) == 0 {
+		req.Query = r.Form.Get(urlParamQuery)
```
mdisibio, line 341: It looks like we are checking for both `query` and `q`.  Since we are dropping Prom support I was hoping we could drop one, but unfortunately it looks like we are using `query` for query_range and `q` for all other search endpoints.    Probably ok to leave both then and we can consolidate starting with Grafana.

Either way, I think this area could be simplifed to make the handling of the query and mode separate, it looks like it calls `r.Form.Get(QueryModeKey)` twice?

modules/frontend/combiner/metrics_query_range.go
```diff
@@ -0,0 +1,85 @@
...
+				if partial.Metrics.TotalJobs == 0 {
+					partial.Metrics.CompletedJobs = 1
+				}
+			}
+
+			samplingRate := resp.AdditionalData()
+			if samplingRate != nil {
+				fRate := samplingRate.(float64)
+
+				if fRate != 0.0 {
```
mdisibio, line 35: nit:  can skip this when rate == 1.0.  Maybe switch to `fRate < 1.0` ?

modules/frontend/frontend.go
```diff
@@ -65,63 +67,81 @@ func New(cfg Config, next http.RoundTripper, o overrides.Interface, reader tempo
...
 		[]pipeline.Middleware{statusCodeWare, retryWare},
 		next)
 
-	queryRangeMiddleware := MergeMiddlewares(
-		newMultiTenantUnsupportedMiddleware(cfg, logger),
-		newQueryRangeMiddleware(cfg, o, reader, logger), retryWare)
+	// traceql metrics
+	queryRangePipeline := pipeline.Build(
+		[]pipeline.AsyncMiddleware[combiner.PipelineResponse]{
+			multiTenantMiddleware(cfg, logger),
```
mdisibio, line 130: Is this enabling multi-tenant support?  I think it would work in the current setup (combiner), but wondering if we should hold on this until we support the next round of features.

pkg/api/http.go
joe-elliott: > It looks like we are checking for both query and q. Since we are dropping Prom support I was hoping we could drop one, but unfortunately it looks like we are using query for query_range and q for all other search endpoints. Probably ok to leave both then and we can consolidate starting with Grafana.

I added this b/c I figured Grafana was using this form post and not url params. Since traceql metrics are already in explore I figured we had to keep it, but let me run a test.

> Either way, I think this area could be simplifed to make the handling of the query and mode separate, it looks like it calls r.Form.Get(QueryModeKey) twice?

Agree there's a bug here. Those internal calls should be attempting to pull from the url. It does have me wondering how it's working at all :). Will dig in.

modules/frontend/metrics_query_range_sharder.go
```diff
@@ -0,0 +1,404 @@
...
+type QueryRangeSharderConfig struct {
+	ConcurrentRequests    int           `yaml:"concurrent_jobs,omitempty"`
+	TargetBytesPerRequest int           `yaml:"target_bytes_per_job,omitempty"`
+	MaxDuration           time.Duration `yaml:"max_duration"`
+	QueryBackendAfter     time.Duration `yaml:"query_backend_after,omitempty"`
+	Interval              time.Duration `yaml:"interval,omitempty"`
+}
+
+// newAsyncQueryRangeSharder creates a sharding middleware for search
+func newAsyncQueryRangeSharder(reader tempodb.Reader, o overrides.Interface, cfg QueryRangeSharderConfig, logger log.Logger) pipeline.AsyncMiddleware[combiner.PipelineResponse] {
```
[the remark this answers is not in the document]
joe-elliott: Yup. I did add support to precalc and return total jobs which may be worth reviewing.

modules/frontend/frontend.go
joe-elliott: Yup, this is enabling multitenant. I'm fine with going back to unsupported. Does this impact the metrics sharding?

modules/frontend/combiner/metrics_query_range.go
joe-elliott: the 0.0 check is just a sanity check to prevent panics. `fRate < 1.0 && fRate > 0.0`. would there ever be a time we deflated metrics with the rate?

pkg/api/http.go
joe-elliott: Ok, I see. `ParseForm` pulls in the url parameters to the form which is why it works. I've updated the code and added some comments to clearly always pull from the url. This works with Grafana/CLI.

modules/frontend/combiner/metrics_query_range.go
mdisibio: No, the sampling rate will always be between 0 and 1 due to the check in the sharder: https://github.com/grafana/tempo/pull/3584/files#diff-85974350e62b77bee6b5d88d537e3946839ccb1bba5f6dc89450a1c5489b5d4fR364

modules/frontend/frontend.go
mdisibio: Sharding is ok, I had in mind the changes needed to the sharder/combiner for histograms. But after more consideration it is good.

modules/frontend/combiner/metrics_query_range.go
joe-elliott: good catch. updated to the suggested condition

mdisibio (approved):
LGTM and nice work.  I look forward to glorious streaming metrics panels!

mdisibio:
> Should we add a new SLO config and SLO metric explicitly for metrics? Currently it piggybacks on search.

I think in the future yes, but not required for this PR.  Even after we make the next round of changes that brings the metrics read path much closer to the search read path, I still expect the minimum throughout and latency to be different.

> Should we reorg the code layout to be around pipeline? i.e. should we bring the combiner/sharder/handlers all together in modules? They are starting to have dependencies that would be easier to see if they were co-located.

Yeah open to another reshuffle if it makes sense.

PR #3591, opened 2024-04-19
Stop counting client cancels as errors for the SLO

mapno:
**What this PR does**:

Makes `context.Cancel` errors count as successes for the SLO.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott:
I don't believe there's a way to know that a `context.Canceled` error is from the client. If we cancel context in the pipeline and let that error bubble up it would look the same here. I'm not opposed to this, but I think we should throw in some tests to feel confident we don't ever bubble context canceled up here?

in the near-ish future we should move SLO calculations out a layer in cloud to avoid issues like this.

electron0zero (approved):

PR #3594, opened 2024-04-19
Updated Grafana image in the docker-compose examples to latest

utr1903:
**What this PR does**:

The Grafana image used in the `docker-compose` examples are hardcoded to `10.1.1` which does not support the functionality of creating dashboard variables out of Tempo. I believe this is a great feature which everyone would like to be able test the moment when they run these examples.

More of such features will definitely come along with improvements in Tempo. Therefore, I would suggest to keep the Grafana images up-to-date just like the images for Prometheus and Tempo in these examples.

### DIFF
--- a/example/docker-compose/agent/docker-compose.yaml
+++ b/example/docker-compose/agent/docker-compose.yaml
@@ -45,7 +45,7 @@ services:
       - "9090:9090"
 
   grafana:
-    image: grafana/grafana:10.1.1
+    image: grafana/grafana:10.4.2
     volumes:
       - ../shared/grafana-datasources.yaml:/etc/grafana/provisioning/datasources/datasources.yaml
     environment:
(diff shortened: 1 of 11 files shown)


utr1903:
I have only changed the `otel-collector` example. I'm not sure if there is a special reason the keep the version in `10.1.1`. If not, I'd be happy to update all examples.

Plus, I see that the `otel-collector` images are also hard-coded (to `0.86.0` and `0.61.0`). If you'd see fit, I can update them as well.

joe-elliott:
I'd prefer not using the `:latest` tag b/c this will be different on everyone's machine. What works for me may not work for you. Instead of the latest tag can we update them to use the latest Grafana release?

`grafana/grafana:10.4.2` ?

joe-elliott (approved):

PR #3596, opened 2024-04-19
Update examples and docs for UID path ownership change

zalegrala:
**What this PR does**:

The recent UID change means that some of the paths in our examples and docs are not correctly owned by the `tempo` user.  Here we modify those examples and docs to use a path specified in the Dockerfile to be owned by the `tempo` user.

**Checklist**
- [ ] Tests updated
- [x] Documentation added


joe-elliott (approved):
i've not run through the examples but agree that `/var` is a more appropriate path for tempo data.

PR #3609, opened 2024-04-25
Align OTEL `semconv` package to `v1.25.0`

zalegrala:
**What this PR does**:

Here we update the use of `semconv` use for OTEL semantics.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/generator/processor/util/util.go
```diff
@@ -1,22 +1,23 @@
 package util
 
 import (
+	semconv "go.opentelemetry.io/otel/semconv/v1.25.0"
```
zalegrala, line 4: This is changing from the `/collector` package, but this update resolves to the same string value as before.

go.mod
```diff
@@ -298,7 +298,7 @@ require (
 	go.etcd.io/etcd/api/v3 v3.5.10 // indirect
 	go.etcd.io/etcd/client/pkg/v3 v3.5.10 // indirect
 	go.etcd.io/etcd/client/v3 v3.5.10 // indirect
-	go.mongodb.org/mongo-driver v1.13.0 // indirect
+	go.mongodb.org/mongo-driver v1.15.0 // indirect
```
zalegrala, line 301: I didn't have this in my local cache and so I updated so I could vendor locally.

zalegrala:
I'm not introducing the lint failures here, and I don't see where they are coming in from.  I'd like to get some review on this, and possibly merge without fixing the lint, since they won't show up on the next PRs.

kvrhdn (approved):
LGTM :+1:

I agree the linting errors are from unchanged files, the ones I checked are already there in main. While we should fix them eventually, maybe not in this PR 😅

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -203,7 +203,7 @@ func (p *Processor) consume(resourceSpans []*v1_trace.ResourceSpans) (err error)
 
 						// A database request will only have one span, we don't wait for the server
 						// span but just copy details from the client span
-						if dbName, ok := processor_util.FindAttributeValue("db.name", rs.Resource.Attributes, span.Attributes); ok {
+						if dbName, ok := processor_util.FindAttributeValue(string(semconv.DBNameKey), rs.Resource.Attributes, span.Attributes); ok {
```
joe-elliott, line 206: i'm concerned about letting this float. a new semconv package may change the behavior of Tempo unexpectedly.

maybe we should confirm through tests it doesn't change? so at least we can make a conscious choice instead of an accidental one.

zalegrala: Agree that floating could be dangerous.  Happy to add a test, but here are a few link that might tamp the concerns?

https://opentelemetry.io/docs/specs/otel/versioning-and-stability/
https://opentelemetry.io/blog/2023/http-conventions-declared-stable/
https://opentelemetry.io/blog/2023/http-conventions-declared-stable/#migrating-from--v1180
https://opentelemetry.io/blog/2023/http-conventions-declared-stable/#migrating-from--v1190

I suppose in my head we definitely want to align with otel here, since `db.name` here is targeting the otel semver, but not pinned to a version.  But I also see your point about this changing without us knowing.  Those links seem to indicate that the otel community won't be changing things without calling them out explicitly.

In the next couple of months I expect us to PR more reliance on the semver package and its use, so thinking more generally about how we want to explicitly test and ensure changes don't happen without us knowing would be good. 

What do you think after taking a look at those links?

joe-elliott: it's good that these conventions are stable and will likely not change. i also really like this change b/c it communicates something better than the string "db.name". a reader can see clearly the source of the string and why we chose it.

however, i am always concerned about changes in dependencies having inadvertent changes in Tempo behavior. I think a simple test that confirms "db.name" == semconv.DBNameKey with a comment about why it exists. Then, if it ever changes, we could choose to move forward and mark it a breaking change or keep it the same.

zalegrala: That sounds good to me.  I'll add that to the PR before merge.  I'll keep this in mind also as we may start to depend on those strings in the near future.

zalegrala: I dug on this a bit and see that we already have some test coverage for this.  The test would have failed if the attribute key had changed, so this is good news.

joe-elliott (approved):

PR #3610, opened 2024-04-25
Golangci deprecations and update using tools pattern

zalegrala:
**What this PR does**:

Here we move golangci to the tools pattern to execute in docker.  Users can now run `make lint` with the same version of golangci.

* add golangci to tools/
* update the golangci version
* handle some deprecations in the golangci config

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


zalegrala:
I have a follow up commit ready to update CI.  Once the new image has been published, then we can use `make lint` in github actions.

joe-elliott (approved):
Love how clean this addition is. Nice work!

PR #3611, opened 2024-04-26
Implement tenant deletion for empty tenants

### LINKED ISSUE #2878: Tenant Index Failures for deleted tenants
**Describe the bug**

A change in https://github.com/grafana/tempo/pull/2781 was an attempt to fix this, but is incomplete.  Also related to https://github.com/grafana/tempo/issues/2754

Currently when Tempo attempts to write a tenant index with 0 blocks and 0 compacted blocks, the tenant index is deleted, and errors to delete the already deleted index are now handled since #2781.  However, some deleted tenants may still show up in the tenant list if any objects are left within the tenant object path.  This results in the poller attempting to read the tenant index and failing (because the index does not exist), and then falling back to a full poll after incrementing the error counter and resulting in incorrect alerts.

**Expected behavior**

I viewed the change in #2781 as a stop gap, but since it didn't achieve the goal, I think we should discuss.  My preferred approach would be to delete all tenant objects left over when we delete the tenant index.

1. `WriteTenantIndex`)_ is called for a tenant with zero blocks and zero compacted blocks
2. `RawWriter.Delete()` is called for the tenant index
3. `List()` at the tenant path to find all remaining objects
4. Loop over remaining objects and call delete.

This will result in a full tenant deletion, instead of only deleting the index.  Note that even if we address the incorrect metric counting for a deleted tenant, we still make the calls to the backend, so my preference would be to delete all the objects from the backend.

**Environment:**
 - Infrastructure: kubernetes
 - Deployment tool: jsonnet

**Additional Context**
https://github.com/grafana/tempo/issues/2754

zalegrala:
**What this PR does**:

In previous attempts to delete the tenant index, it was possible that undiscoverable objects remained in the backend as a partial blocks, which lead to unintended behavior for the poller, whereby the poller would discover a tenant, determine there were no blocks, and delete the index.  However, the partial blocks kept the tenant in the list and so would repeat this forever, or until bucket retention policies removed the objects.

Here we take a more complete approach to tenant deletion.  Upon deletion of a tenant index, we search the backend for objects on a given tenant, and then delete them after a configurable amount of time has expired since the object was last modified.

* Implement Find() on backend readers
* Implement Delete() on backend writers
* Add configuration for `EmptyTenantDeletionAge`, default to 4x the polling cycle, which will delete objects from empty tenants after this time has expired.

**Background Summary:**

* A partial block is written to a tenant and then stops sending spans.
* Retention eventually removes all known blocks.
* The partial block remains for the tenant because it is not known.
    * i.e: `<tenant>/<blockID>/data.parquet` exists
    * No `meta.json` nor `meta.compacted.json` exist for this block.
* A poller owning this tenant calls [Tenants()](https://github.com/zalegrala/tempo/blob/9e77d6f6bba2f06b53acbd70097d9dc7df86282b/tempodb/blocklist/poller.go#L142) to know on which tenants to build an index.
    * This call is simply a [list](https://github.com/zalegrala/tempo/blob/9e77d6f6bba2f06b53acbd70097d9dc7df86282b/tempodb/backend/raw.go#L167) at `/` to determine all the tenant “directory” names.
* For the given tenant, no blocks are found, and so the tenant performs a [WriteTenantIndex()](https://github.com/zalegrala/tempo/blob/9e77d6f6bba2f06b53acbd70097d9dc7df86282b/tempodb/blocklist/poller.go#L239)
    * If we are writing an index with no blocks, the tenant index is instead [deleted](https://github.com/zalegrala/tempo/blob/9e77d6f6bba2f06b53acbd70097d9dc7df86282b/tempodb/backend/raw.go#L112).
* A poller that does not own the tenant performs the above, with a subtle but important difference.
    * First attempt to read the index with a [TenantIndex()](https://github.com/zalegrala/tempo/blob/9e77d6f6bba2f06b53acbd70097d9dc7df86282b/tempodb/blocklist/poller.go#L202) call
    * The call will fail because the owning poller has deleted the index.
    * The failing index metric used to alert is [incremented](https://github.com/zalegrala/tempo/blob/9e77d6f6bba2f06b53acbd70097d9dc7df86282b/tempodb/blocklist/poller.go#L214).
    * The poller will fallback to [full polling]() as described above.
    

Related #2678
Related #2754
Related #2781
Fixes #2878

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/configuration/_index.md
+++ b/docs/sources/tempo/configuration/_index.md
@@ -982,12 +982,18 @@ storage:
         # Default 1
         [blocklist_poll_tolerate_consecutive_errors: <int>]
 
-        # Polling will delete the index for a tenant if no blocks are found to
-        # exist.  The empty_tenant_deletion_age configuration is used to tune how
-        # quickly the poller will also delete any remaining backend objects found in the
-        # tenant path.
+        # Used to tune how quickly the poller will delete any remaining backend
+        # objects found in the tenant path.  This functionality requires enabling
+        # below.
         # Default: 12h
         [empty_tenant_deletion_age: <duration>]
+
+        # Polling will delete the index for a tenant if no blocks are found to
+        # exist.  If this setting is enabled, the poller will also delete any
+        # remaining backend objects found in the tenant path.  This is used to
+        # clean up partial blocks which may have not been cleaned up by the
+        # retention.
+        [empty_tenant_deletion_enabled: <bool> | default = false]
 
         # Cache type to use. Should be one of "redis", "memcached"
         # Example: "cache: memcached"
--- a/docs/sources/tempo/configuration/_index.md
+++ b/docs/sources/tempo/configuration/_index.md
@@ -982,6 +982,13 @@ storage:
         # Default 1
         [blocklist_poll_tolerate_consecutive_errors: <int>]
 
+        # Polling will delete the index for a tenant if no blocks are found to
+        exist.  The empty_tenant_deletion_age configuration is used to tune how
+        quickly the poller will also delete any remaining backend objects found in the
+        tenant path.
+        # Default: 20m; four times the blocklist_poll configuration above.
+        [empty_tenant_deletion_age: <duration>]
+
         # Cache type to use. Should be one of "redis", "memcached"
         # Example: "cache: memcached"
         # Deprecated. See [cache](#cache) section below.
--- a/docs/sources/tempo/configuration/_index.md
+++ b/docs/sources/tempo/configuration/_index.md
@@ -983,9 +983,9 @@ storage:
         [blocklist_poll_tolerate_consecutive_errors: <int>]
 
         # Polling will delete the index for a tenant if no blocks are found to
-        exist.  The empty_tenant_deletion_age configuration is used to tune how
-        quickly the poller will also delete any remaining backend objects found in the
-        tenant path.
+        # exist.  The empty_tenant_deletion_age configuration is used to tune how
+        # quickly the poller will also delete any remaining backend objects found in the
+        # tenant path.
         # Default: 20m; four times the blocklist_poll configuration above.
         [empty_tenant_deletion_age: <duration>]
 
--- a/docs/sources/tempo/configuration/_index.md
+++ b/docs/sources/tempo/configuration/_index.md
@@ -986,7 +986,7 @@ storage:
         # exist.  The empty_tenant_deletion_age configuration is used to tune how
         # quickly the poller will also delete any remaining backend objects found in the
         # tenant path.
-        # Default: 20m; four times the blocklist_poll configuration above.
+        # Default: 12h
         [empty_tenant_deletion_age: <duration>]
 
         # Cache type to use. Should be one of "redis", "memcached"
--- a/tempodb/backend/raw.go
+++ b/tempodb/backend/raw.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"io"
 	"path"
+	"time"
 
 	"github.com/google/uuid"
 
@@ -25,7 +26,15 @@ const (
 // from the backend
 type KeyPath []string
 
-type Feature int
+// FundFunc is executed for each object in the backend.  The provided FindOpts
+// are used to determine how to handle the object.  Any collection of these
+// objects is the callers responsibility.
+type FindFunc func(FindOpts)
+
+type FindOpts struct {
+	Modified time.Time
+	Key      string
+}
 
 // RawWriter is a collection of methods to write data to tempodb backends
 type RawWriter interface {
@@ -45,6 +54,8 @@ type RawReader interface {
 	List(ctx context.Context, keypath KeyPath) ([]string, error)
 	// ListBlocks returns all blockIDs and compactedBlockIDs for a tenant.
 	ListBlocks(ctx context.Context, tenant string) (blockIDs []uuid.UUID, compactedBlockIDs []uuid.UUID, err error)
+	// Find executes the FindFunc for each object in the backend starting at the specified keypath.  Collection of these objects is the callers responsibility.
+	Find(ctx context.Context, keypath KeyPath, f FindFunc) error
 	// Read is for streaming entire objects from the backend.  There will be an attempt to retrieve this from cache if shouldCache is true.
 	Read(ctx context.Context, name string, keyPath KeyPath, cacheInfo *CacheInfo) (io.ReadCloser, int64, error)
 	// ReadRange is for reading parts of large objects from the backend.
@@ -225,6 +236,11 @@ func (r *reader) TenantIndex(ctx context.Context, tenantID string) (*TenantIndex
 	}
 
 	return i, nil
+}
+
+// Find implements backend.Reader
+func (r *reader) Find(ctx context.Context, keypath KeyPath, f FindFunc) error {
+	return r.r.Find(ctx, keypath, f)
 }
 
 // Shutdown implements backend.Reader
--- a/tempodb/backend/raw.go
+++ b/tempodb/backend/raw.go
@@ -26,7 +26,7 @@ const (
 // from the backend
 type KeyPath []string
 
-// FundFunc is executed for each object in the backend.  The provided FindOpts
+// FundFunc is executed for each object in the backend.  The provided FindMatch
 // are used to determine how to handle the object.  Any collection of these
 // objects is the callers responsibility.
 type FindFunc func(FindMatch)
--- a/tempodb/backend/raw.go
+++ b/tempodb/backend/raw.go
@@ -29,9 +29,9 @@ type KeyPath []string
 // FundFunc is executed for each object in the backend.  The provided FindOpts
 // are used to determine how to handle the object.  Any collection of these
 // objects is the callers responsibility.
-type FindFunc func(FindOpts)
+type FindFunc func(FindMatch)
 
-type FindOpts struct {
+type FindMatch struct {
 	Modified time.Time
 	Key      string
 }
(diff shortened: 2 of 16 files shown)


tempodb/blocklist/poller.go
```diff
@@ -456,6 +466,46 @@ func (p *Poller) tenantIndexPollError(idx *backend.TenantIndex, err error) error
...
+	if p.cfg.EmptyTenantDeletionAge == 0 {
+		return fmt.Errorf("empty tenant deletion age must be greater than 0")
+	}
+
+	var foundObjects []string
+	err := p.reader.Find(ctx, backend.KeyPath{tenantID}, func(opts backend.FindOpts) {
+		level.Info(p.logger).Log("msg", "checking object for deletion", "object", opts.Key, "modified", opts.Modified)
+
+		if time.Since(opts.Modified) > p.cfg.EmptyTenantDeletionAge {
+			foundObjects = append(foundObjects, opts.Key)
```
zalegrala, line 482: We could also call delete directly here, but I thought checking once more for the index would be good, which we do below.  Additionally, not having an error return here makes the interface for FindFunc simpler.

docs/sources/tempo/configuration/_index.md
```diff
@@ -974,6 +974,13 @@ storage:
         # Default 1
         [blocklist_poll_tolerate_consecutive_errors: <int>]
 
+        # Polling will delete the index for a tenant if no blocks are found to
+        exist.  The empty_tenant_deletion_age configuration is used to tune how
+        quickly the poller will also delete any remaining backend objects found in the
+        tenant path.
```
knylander-grafana, line 980: I think each line needs to be commented, judging by the other examples in this file. 
```suggestion
        # Polling will delete the index for a tenant if no blocks are found to
        # exist.  The empty_tenant_deletion_age configuration is used to tune how
        # quickly the poller will also delete any remaining backend objects found in the
        # tenant path.
```

zalegrala: Good catch, thanks.

tempodb/backend/raw.go
```diff
@@ -25,7 +26,15 @@ const (
 // from the backend
 type KeyPath []string
 
-type Feature int
+// FundFunc is executed for each object in the backend.  The provided FindOpts
+// are used to determine how to handle the object.  Any collection of these
+// objects is the callers responsibility.
+type FindFunc func(FindOpts)
+
+type FindOpts struct {
```
mdisibio, line 34: I'd like to rename this to `FindResult` or something similar, to disambiguate it from the options pattern.

tempodb/blocklist/poller.go
mdisibio: Agree, separating the search and delete steps seems good, also to avoid any complications from changing the search results while going.  I think it would be good to log what's being deleted.    One more thought:  it looks like are deleting everything older than the cutoff, on each pass. Does it matter if anything newer is found?

docs/sources/tempo/configuration/_index.md
```diff
@@ -974,6 +974,13 @@ storage:
         # Default 1
         [blocklist_poll_tolerate_consecutive_errors: <int>]
 
+        # Polling will delete the index for a tenant if no blocks are found to
+        # exist.  The empty_tenant_deletion_age configuration is used to tune how
+        # quickly the poller will also delete any remaining backend objects found in the
+        # tenant path.
+        # Default: 20m; four times the blocklist_poll configuration above.
+        [empty_tenant_deletion_age: <duration>]
```
mdisibio, line 982: 20m seems quick, and in the range of normal block operations (maybe small network blips, and some compactions can take over an hour).  To start, what do you think about something more on the order of 2h to 24h?

tempodb/blocklist/poller.go
```diff
@@ -456,6 +466,46 @@ func (p *Poller) tenantIndexPollError(idx *backend.TenantIndex, err error) error
...
+	_, err = p.reader.TenantIndex(ctx, tenantID)
+	// If we have any error other than that which indicates that the tenant index
+	// call was made successfully, and that it does not exist, do nothing.  Only
+	// proceed if we know that the index does not exist.
+	if !errors.Is(err, backend.ErrDoesNotExist) {
+		return nil
+	}
+
+	for _, object := range foundObjects {
+		dir, name := path.Split(object)
```
mdisibio, line 499: There is some friction here between Find and Delete, where Find returns the path as a single string and Delete requires it split between name and keypath.  Does it make sense for them to match?  Obviously very minor, but something to consider.  Honestly maybe the better approach is to do a cleanup pass over the whole backend interfaces since they have grown closer towards generic filesystem operations after adding List/Find/Delete.

tempodb/backend/raw.go
zalegrala: How about `FindMatch` or something to indicate we're matching these sets of fields against the objects.

tempodb/blocklist/poller.go
zalegrala: I'm not sure that it matters if newer data is found, but I think I see what you are getting at.  The reason I think it probably doesn't matter is that the partial block may have been written ages ago.  Would it perhaps be better to only return the values if __all__ data is older than the `p.cfg.EmptyTenantDeletionAge`?  So any more recent data would be a reason to bail on the delete, not just the appearance of the index below.

docs/sources/tempo/configuration/_index.md
zalegrala: How about 12h default value?

tempodb/blocklist/poller.go
zalegrala: Hmm, I see this also.  Looking at `raw.go` , the `Delete` was modeled after `Write` and `Append`.  We could break from this and just receive the string path on `Delete`.

In the case of the find function implementation here, we could collect the objects in a different form than the string path to avoid the split and then rejoin later [here](https://github.com/zalegrala/tempo/blob/aa6308f7627564fcc91c4de5da521297c50e984f/tempodb/backend/s3/s3.go#L248), but I'm not sure.

zalegrala: Checking again, all the backends are doing the join on `backend.KeyPath` and `name`.  So how do we feel about the `Delete` call just taking a string path?

Alternatively, we could follow up in another PR to come up with a more holistic approach.

tempodb/backend/raw.go
zalegrala: I've pushed a commit for this change.  Let me know if that name seems better.

docs/sources/tempo/configuration/_index.md
zalegrala: I've pushed a commit for the 12h default.

tempodb/blocklist/poller.go
zalegrala: I pushed an additional safety check so that any object with a recent timestamp, we avoid deleting.  This way, all objects are subject to deletion at the same time, only after the most recently modified object crosses the threshold.  Hows that seem to you?

docs/sources/tempo/configuration/_index.md
mdisibio: Yep, 12h sounds good.  Just thinking about if this could interact with new tenants with partially written files, or other normal operations.

tempodb/backend/raw.go
```diff
@@ -25,7 +26,15 @@ const (
 // from the backend
 type KeyPath []string
 
-type Feature int
+// FundFunc is executed for each object in the backend.  The provided FindOpts
```
mdisibio, line 29: nit: update comment for renamed FindMatch

tempodb/blocklist/poller.go
mdisibio: Yeah this change seems good.  I was considering that indiscriminately deleting any old files might be dangerous, not under normal circumstances, but situations like a misconfigured or shared bucket, custom workflows, etc.    Not recommending it, but theoretically Tempo could also only delete files it owns like `data.parquet`, `bloom-*`.

tempodb/blocklist/poller.go
mdisibio: That's ok. Since there's not a clear preference for either way, let's leave it.

tempodb/blocklist/poller.go
```diff
@@ -456,6 +466,57 @@ func (p *Poller) tenantIndexPollError(idx *backend.TenantIndex, err error) error
...
+		foundObjects  []string
+		recentObjects int
+	)
+	err := p.reader.Find(ctx, backend.KeyPath{tenantID}, func(opts backend.FindMatch) {
+		level.Info(p.logger).Log("msg", "checking object for deletion", "object", opts.Key, "modified", opts.Modified)
+
+		if time.Since(opts.Modified) > p.cfg.EmptyTenantDeletionAge {
+			foundObjects = append(foundObjects, opts.Key)
+		} else {
+			recentObjects++
```
mdisibio, line 487: This could exit early as soon as it finds the first match, if we added a return value to FindFunc to signal.  Not required for now, but it's keep it in mind if we run into situations where continuing to iterate results is an issue.

mdisibio (approved):
Nice work. Left a few more idle thoughts and responses, but nothing blocking.

tempodb/blocklist/poller.go
zalegrala: That's a good thing to keep in mind.  Let's see how it goes without matching the file names.

tempodb/blocklist/poller.go
zalegrala: That's true.  I suspect that the number of objects are in the 10s at most, and the call is in batches of 1000, so likely won't result in additional calls to the backend.  Signaling a stop sounds reasonable if this does come up later.

PR #3612, opened 2024-04-26
Fix span-metrics' subprocessors bug

mapno:
**What this PR does**:
Fixes a bug in which a shared map in the metrics-generator override config was not properly copied to avoid modifying the config of other tenants.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott:
How difficult is it to construct a test that reproes and then confirm this fixes it?

mapno:
> How difficult is it to construct a test that reproes and then confirm this fixes it?

I think it's feasible. I'll build one 👍

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
 * [BUGFIX] Add spss and limit to the frontend cache key to prevent the return of incorrect results. [#3557](https://github.com/grafana/tempo/pull/3557) (@joe-elliott)
 * [BUGFIX] Use os path separator to split blocks path. [#3552](https://github.com/grafana/tempo/issues/3552) (@teyyubismayil)
 * [BUGFIX] Correctly parse traceql queries with > 1024 character attribute names or static values. [#3571](https://github.com/grafana/tempo/issues/3571) (@joe-elliott)
+* [BUGFIX] Fix span-metrics' subprocessors bug that applied wrong configs when running multiple tenants. [#](https://github.com/grafana/tempo/pull/) (@mapno)
```
joe-elliott, line 38:
```suggestion
* [BUGFIX] Fix span-metrics' subprocessors bug that applied wrong configs when running multiple tenants. [#3612](https://github.com/grafana/tempo/pull/3612) (@mapno)
```

joe-elliott (commented):
does it make more sense to break the shared memory here:

https://github.com/grafana/tempo/blob/main/modules/generator/config.go#L66

where we copy the config?

joe-elliott (approved):
looks like a simple test failure, but i'm good on this approach

PR #3618, opened 2024-04-26
max_global_traces_per_user: take into account ingestion.tenant_shard_size when converting to local limit

kvrhdn:
**What this PR does**:
Fixes an oversight when adding ingester shuffle sharding support (https://github.com/grafana/tempo/pull/3506). 

`max_global_traces_per_user` is a global limit enforced in the ingester. It is converted to a local limit by dividing the global limit by the amount of (healthy) ingesters in the ring. When `ingestion.tenant_shard_size` is set the tenant will have a smaller ring than the total count of ingesters. As a result the actual local limit would be higher in those cases.

Before:

```
actual local limit = max_global_traces_per_user / ingester instances
```

After:

```
actual local limit = max_global_traces_per_user / min(ingester instances, tenant_shard_size)
```

~~Fixes #<issue number>~~

**Checklist**
- [ ] ~~Tests updated~~
- [ ] ~~Documentation added~~


mapno (approved):
Nice fix

electron0zero (approved):
good catch :shipit:

PR #3624, opened 2024-04-29
Update set-up-monitoring.md

JStickler:
Remove backticks from Link (code formatting was overriding the link and displaying the link text as code).


knylander-grafana (approved):
Thank you for the fix!

PR #3628, opened 2024-04-30
Add replication factor to block meta [1 of 3]

mapno:
**What this PR does**:

Adds replication factor to the block meta and updates the compactors to use it as a grouping parameter. By default is set to 0, even when replication factor is enabled. It's only used in blocks created in the local-blocks processor in the metrics-generator.

It also contains a refactor of `CreateWALBlock()`. Instead of passing individual parameters, it accepts a `backend.BlockMeta`. It coalesces creating new blocks for all the different encodings (`v2`, `vParquet2`, `vParquet3`, `vParquet4`).

It also has a fix to the local-blocks processor when cutting traces _immediately_.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/wal/wal.go
```diff
@@ -158,28 +157,16 @@ func (w *WAL) RescanBlocks(additionalStartSlack time.Duration, log log.Logger) (
...
-//  1. Add a dedicatedColumns parameter to NewBlock, and have it default to nil.
-//  2. Have encoding-specific config be part of the encoding itself
-//  3. Pass the meta file path to the WAL constructor
-
-func (w *WAL) NewBlockWithDedicatedColumns(id uuid.UUID, tenantID, dataEncoding string, dedicatedColumns backend.DedicatedColumns) (common.WALBlock, error) {
-	return w.newBlock(id, tenantID, dataEncoding, w.c.Version, dedicatedColumns)
-}
-
-func (w *WAL) newBlock(id uuid.UUID, tenantID string, dataEncoding string, blockVersion string, dedicatedColumns backend.DedicatedColumns) (common.WALBlock, error) {
+func (w *WAL) newBlock(meta *backend.BlockMeta, blockVersion string) (common.WALBlock, error) {
```
mdisibio, line 164: Any reason not to consolidate these methods?

tempodb/encoding/vparquet/block_search_test.go
```diff
@@ -241,7 +242,7 @@ func makeBackendBlockWithTraces(t *testing.T, trs []*Trace) *backendBlock {
 		BloomShardSizeBytes: 100 * 1024,
 	}
 
-	meta := backend.NewBlockMeta("fake", uuid.New(), VersionString, backend.EncNone, "")
+	meta := backend.NewBlockMeta("fake", uuid.New(), VersionString, backend.EncNone, model.CurrentEncoding)
```
mdisibio, line 245: I think `model.CurrentEncoding` is not correct to combine with the vParquet formats. For them the dataEncoding is always blank.  The v2 block format contained different flavors of trace byte layout within the same overall block layout, so the block was a combination of version + dataEncoding.   Unless I missed a change to `model.CurrentEncoding`.

mdisibio:
Running locally, not seeing replicationFactor=1 in the complete blocks flushed by the generator.  It looks like [newStreamingBlock](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet3/create.go#L109) is dropping it and also needs to be updated (at least vp2 and vp3)

tempodb/encoding/vparquet/block_search_test.go
mapno: Changed this, nice catch

mdisibio (approved):

PR #3629, opened 2024-04-30
TraceQL - Add support for scoped intrinsics using `:`

ie-pham:
**What this PR does**: Add support for scoped intrinsics using `:`.  Users can now make queries like so: `{span:duration > 100ms}`. Which is different from `{span.duration > 100ms}` as the `:` signifies that we are looking at the duration of the span where as the `.` will signify that it is an attribute name `duration` of a span. 

This is the change will allow us to additionally features to Traceql like `trace:id` and `link:spanId` in the future. 

Screenshot 2024-04-30 at 7 20 51 PM
Screenshot 2024-04-30 at 7 19 33 PM

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * [ENHANCEMENT] Better compaction throughput and memory usage [#3579](https://github.com/grafana/tempo/pull/3579) (@mdisibio)
 * [ENHANCEMENT] Add support for sharded ingester queries  [#3574](https://github.com/grafana/tempo/pull/3574) (@zalegrala)
 * [ENHANCEMENT] TraceQL - Add support for scoped intrinsics using `:` [#3629](https://github.com/grafana/tempo/pull/3629) (@ie-pham)
+  available scoped intrinsics: trace:duration, trace:rootName, trace:rootService, span:duration, span:kind, span:name, span:status, span:statusMessage
 * [BUGFIX] Fix metrics query results when filtering and rating on the same attribute [#3428](https://github.com/grafana/tempo/issues/3428) (@mdisibio)
 * [BUGFIX] Fix metrics query results when series contain empty strings or nil values [#3429](https://github.com/grafana/tempo/issues/3429) (@mdisibio)
 * [BUGFIX] Fix metrics query duration check, add per-tenant override for max metrics query duration [#3479](https://github.com/grafana/tempo/issues/3479) (@mdisibio)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 * [ENHANCEMENT] Add tenant to frontend job cache key. [#3527](https://github.com/grafana/tempo/pull/3527) (@joe-elliott)
 * [ENHANCEMENT] Better compaction throughput and memory usage [#3579](https://github.com/grafana/tempo/pull/3579) (@mdisibio)
 * [ENHANCEMENT] Add support for sharded ingester queries  [#3574](https://github.com/grafana/tempo/pull/3574) (@zalegrala)
+* [ENHANCEMENT] TraceQL - Add support for scoped intrinsics using `:` [#3629](https://github.com/grafana/tempo/pull/3629) (@ie-pham)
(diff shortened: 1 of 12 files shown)


ie-pham:
```
func NewIntrinsic(n Intrinsic) Attribute {
	return Attribute{
		Scope:     AttributeScopeNone,
		Parent:    false,
		Name:      n.String(),
		Intrinsic: n,
	}
}

// NewScopedAttribute creates a new scopedattribute with the given identifier string.
// this handles parent, trace, span, and resource scopes.
func NewScopedIntrinsic(scope AttributeScope, parent bool, in string) Attribute {
	return Attribute{
		Scope:     scope,
		Parent:    parent,
		Name:      in,
		Intrinsic: IntrinsicScoped,
	}
}
```

Added a new scoped intrinsic struct and leaving every other `intrinsic` logic as is

docs/sources/tempo/traceql/_index.md
```diff
@@ -95,6 +95,22 @@ these intrinsics are significantly more performant because they have to inspect
 possible to span-level intrinsics.
 {{% /admonition %}}
 
+### Scoped Intrinsic Fields
```
knylander-grafana, line 98: Changing this to sentence case to match our style 
```suggestion
### Scoped intrinsic fields
```

pkg/traceql/test_examples.yaml
```diff
@@ -51,6 +51,14 @@ valid:
   - '{ duration > 1s }'
   - '{ 1 < 1h }'
   - '{ 1 <= 1.1 }'
+  # scoped intrinsics
```
joe-elliott, line 54: i'd add a few examples under "validate_fails" to make sure that we can't compare the new intrinsics in incorrect ways.

maybe the other sections?

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 * [ENHANCEMENT] Add tenant to frontend job cache key. [#3527](https://github.com/grafana/tempo/pull/3527) (@joe-elliott)
 * [ENHANCEMENT] Better compaction throughput and memory usage [#3579](https://github.com/grafana/tempo/pull/3579) (@mdisibio)
 * [ENHANCEMENT] Add support for sharded ingester queries  [#3574](https://github.com/grafana/tempo/pull/3574) (@zalegrala)
+* [ENHANCEMENT] TraceQL - Add support for scoped intrinsics using `:` [#3629](https://github.com/grafana/tempo/pull/3629) (@ie-pham)
```
joe-elliott, line 28: please add the list of new instrinics here.

joe-elliott (commented):
looking good! a few comments but this is close to mergeable.

after this is merged please also file an issue in github.com/grafana/grafana requesting support for the new intrinsics. let me know and i will help you tag it correctly.

pkg/traceql/ast.go
```diff
@@ -722,7 +722,7 @@ func (Attribute) referencesSpan() bool {
 func NewScopedAttribute(scope AttributeScope, parent bool, att string) Attribute {
 	intrinsic := IntrinsicNone
 	// if we are explicitly passed a resource or span scopes then we shouldn't parse for intrinsic
-	if scope != AttributeScopeResource && scope != AttributeScopeSpan {
+	if scope == AttributeScopeNone && !parent {
```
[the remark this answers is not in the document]
ie-pham: currently
```
if scope != AttributeScopeResource && scope != AttributeScopeSpan
```
this would allow {parent.duration} to parse to IntrinsicDuration and we discussed to make this to parse it as "duration" as an attribute

pkg/traceql/expr.y
```diff
@@ -167,8 +170,9 @@ selectOperation:
   ;
 
 attribute:
-  intrinsicField    { $$ = $1 }
-  | attributeField  { $$ = $1 }
+  intrinsicField          { $$ = $1 }
```
joe-elliott, line 173: can we add a comment above the `intrinsicField` definition that we are no longer adding these?

joe-elliott (approved):
asking for one small comment added, but otherwise lgtm.

great work!

PR #3641, opened 2024-05-01
fix(traceql): fix extract matcher regexes to work with regexp-type matchers

### LINKED ISSUE #3635: Autocomplete regex several issues
**Describe the bug**
(1) If we try to autocomplete a query like `{ .foo=~"(a|b)"}` it seems like it should work but it doesn't (any tag name, doesn't matter).  From debugging it is failing the check for single spanset operation, so probably the pipe `|` is incorrectly caught as a OR operation.  Likely just need a tweak to the `singleFilterRegexp` regex:  https://github.com/grafana/tempo/blob/main/pkg/traceql/extractmatcher.go#L36

(2) Simplifying the condition to `=~"a"` it still doesn't work.  Here it is failing the `matchersRegexp`. I believe it's this section:  `[=|<=|>=|=~|!=|>|<|!~]` which needs to be wrapped in parens instead of brackets to catch the multi-char operators: `(=|<=|>=|=~|!=|>|<|!~)`

Here's are some test cases for TextExtractMatchers:
```go
{
    name:     "test",
    query:    `{.foo=~"(a|b)"}`,
    expected: `{.foo=~"(a|b)"}`,
},
{
    name:     "test",
    query:    `{.foo=~"a"}`,
    expected: `{.foo=~"a"}`,
},
```

```
expected: "{.foo=~\"(a|b)\"}"
actual  : "{}"

expected: "{.foo=~\"a\"}"
actual  : "{}"
```

**To Reproduce**

**Expected behavior**

**Environment:**

**Additional Context**

sd2k:
This improves the regexes used by autocomplete (and I think the search tags v2 endpoint?) so that they work with queries like `{ resource.service.name = "app" && resource.k8s.cluster.name=~"prod-.+" }`.

Fixes #3635.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## main / unreleased
 
+* [BUGFIX] Fix handling of regex matchers in autocomplete endpoints [#3641](https://github.com/grafana/tempo/pull/3641) (@sd2k)
 * [ENHANCEMENT] Surface new labels for uninstrumented services and systems [#3543](https://github.com/grafana/tempo/pull/3543) (@t00mas)
 * [FEATURE] Add TLS support for Memcached Client [#3585](https://github.com/grafana/tempo/pull/3585) (@sonisr)
 * [FEATURE] TraceQL metrics queries: add quantile_over_time [#3605](https://github.com/grafana/tempo/pull/3605) [#3633](https://github.com/grafana/tempo/pull/3633) (@mdisibio) 
--- a/pkg/traceql/extractmatcher.go
+++ b/pkg/traceql/extractmatcher.go
@@ -18,7 +18,7 @@ import (
 //  3. The boolean values "true" or "false".
 //
 // Example: "http.status_code = 200" from the query "{ .http.status_code = 200 && .http.method = }"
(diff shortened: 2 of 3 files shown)


joe-elliott:
unsure why the drone CI failed. I have restarted it.

can you sign the CLA and add a changelog entry?

sd2k:
Hmm I've rebased on main just in case. Signed the CLA and added a changelog entry too :+1:

mdisibio:
Drone is failing on the step to test the debian package because tempo isn't starting up.  It doesn't seem like it captures the tempo logs.  Unlikely to be caused by the changes in this PR, but digging into it.

mdisibio:
I wasn't able to do exact reproduction of the drone steps because of systemd-in-docker woes, but was able to confirm that `make release-snapshot` builds the deb, it installs, and the included tempo binary starts ok with `/usr/bin/tempo -config.file /etc/tempo/config.yml`.  Leaning towards bypassing the drone step and merging.

```
level=info ts=2024-05-08T12:36:44.545535434Z caller=main.go:118 msg="Starting Tempo" version="(version=2.4.0-SNAPSHOT-1539758b9, branch=3635-autocomplete-regex-issues, revision=1539758b9)"
level=info ts=2024-05-08T12:36:44.548547946Z caller=server.go:240 msg="server listening on addresses" http=[::]:3200 grpc=[::]:9095
...
level=info ts=2024-05-08T12:36:44.623156247Z caller=app.go:214 msg="Tempo started"
```

mdisibio (approved):

mdisibio:
The drone issues were resolved in #3657 and all looks good. Thanks.

PR #3647, opened 2024-05-03
Implement polling tenants concurrently

zalegrala:
**What this PR does**:

Here we implement basic polling for tenants concurrently.  This is another step
towards making the polling more efficient.  The next step will be to implement
a weighted polling strategy to poll tenants by some priority.  For now, I think
this is a reasonable start and should stand on its own.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/integration/poller/poller_test.go
+++ b/integration/poller/poller_test.go
@@ -171,8 +171,6 @@ func TestPollerOwnership(t *testing.T) {
 					mmResults, cmResults, err := rr.ListBlocks(context.Background(), testTenant)
 					require.NoError(t, err)
 					sort.Slice(mmResults, func(i, j int) bool { return mmResults[i].String() < mmResults[j].String() })
-					// t.Logf("mmResults: %s", mmResults)
-					// t.Logf("cmResults: %s", cmResults)
 
 					require.Equal(t, tenantExpected[testTenant], mmResults)
 					require.Equal(t, len(tenantExpected[testTenant]), len(mmResults))
--- a/integration/poller/poller_test.go
+++ b/integration/poller/poller_test.go
@@ -4,8 +4,10 @@ import (
 	"bytes"
 	"context"
 	"crypto/rand"
+	mathrand "math/rand"
 	"os"
 	"sort"
+	"strconv"
 	"testing"
 	"time"
 
@@ -121,7 +123,8 @@ func TestPollerOwnership(t *testing.T) {
 				var ww backend.RawWriter
 				var cc backend.Compactor
 
-				concurrency := 3
+				concurrency := 10
+				tenantCount := 1000
 
 				e := hhh.Endpoint(hhh.HTTPPort())
 				switch tc.name {
@@ -150,47 +153,31 @@ func TestPollerOwnership(t *testing.T) {
 
 				blocklistPoller := blocklist.NewPoller(&blocklist.PollerConfig{
 					PollConcurrency:        3,
-					TenantPollConcurrency:  2,
+					TenantPollConcurrency:  3,
 					TenantIndexBuilders:    1,
 					EmptyTenantDeletionAge: 10 * time.Minute,
 				}, OwnsEverythingSharder, r, cc, w, logger)
 
 				// Use the block boundaries in the GCS and S3 implementation
 				bb := blockboundary.CreateBlockBoundaries(concurrency)
-				// Pick a boundary to use for this test
-				base := bb[1]
-				expected := []uuid.UUID{}
 
-				expected = append(expected, uuid.MustParse("00000000-0000-0000-0000-000000000000"))
-				expected = append(expected, uuid.MustParse("ffffffff-ffff-ffff-ffff-ffffffffffff"))
+				tenantExpected := map[string][]uuid.UUID{}
 
-				// Grab the one before the boundary
-				decrementUUIDBytes(base)
-				expected = append(expected, uuid.UUID(base))
+				// Push some data to a few tenants
+				for i := 0; i < tenantCount; i++ {
+					testTenant := tenant + strconv.Itoa(i)
+					tenantExpected[testTenant] = pushBlocksToTenant(t, testTenant, bb, w)
 
-				incrementUUIDBytes(base)
-				expected = append(expected, uuid.UUID(base))
+					mmResults, cmResults, err := rr.ListBlocks(context.Background(), testTenant)
+					require.NoError(t, err)
+					sort.Slice(mmResults, func(i, j int) bool { return mmResults[i].String() < mmResults[j].String() })
+					// t.Logf("mmResults: %s", mmResults)
+					// t.Logf("cmResults: %s", cmResults)
 
-				incrementUUIDBytes(base)
-				expected = append(expected, uuid.UUID(base))
-
-				incrementUUIDBytes(base)
-				expected = append(expected, uuid.UUID(base))
-
-				writeTenantBlocks(t, w, tenant, expected)
-
-				sort.Slice(expected, func(i, j int) bool { return expected[i].String() < expected[j].String() })
-				t.Logf("expected: %v", expected)
-
-				mmResults, cmResults, err := rr.ListBlocks(context.Background(), tenant)
-				require.NoError(t, err)
-				sort.Slice(mmResults, func(i, j int) bool { return mmResults[i].String() < mmResults[j].String() })
-				t.Logf("mmResults: %s", mmResults)
-				t.Logf("cmResults: %s", cmResults)
-
-				assert.Equal(t, expected, mmResults)
-				assert.Equal(t, len(expected), len(mmResults))
-				assert.Equal(t, 0, len(cmResults))
+					require.Equal(t, tenantExpected[testTenant], mmResults)
+					require.Equal(t, len(tenantExpected[testTenant]), len(mmResults))
+					require.Equal(t, 0, len(cmResults))
+				}
 
 				l := blocklist.New()
 				mm, cm, err := blocklistPoller.Do(l)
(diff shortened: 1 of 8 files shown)


zalegrala:
I'm aware of a race in the tests when we check the error message if we exceed the threshold during polling.  Since now we poll concurrently, we can't know which error will be the final error, and so an adjustment needs to be made to assert the error is one of n, rather than one specific.  I'll follow up.

In the meantime, feedback welcome.

My thought to follow this up this PR is to modify the tenant loop, so we loop over tenants which are sorted by some metric, like block size, or last poll duration.  This way we can start work on the ones that require the most effort first.

tempodb/blocklist/poller.go
```diff
@@ -148,44 +150,69 @@ func (p *Poller) Do(previous *List) (PerTenant, PerTenantCompacted, error) {
...
+		if err := finalErr.Load(); err != nil {
+			level.Error(p.logger).Log("msg", "exiting polling loop early because too many errors", "errCount", consecutiveErrors)
+			return nil, nil, err
+		}
+
+		wg.Add(1)
+		go func(tenantID string) {
+			defer wg.Done()
+
+			newBlockList, newCompactedBlockList, err := p.pollTenantAndCreateIndex(ctx, tenantID, previous)
```
javiermolinar, line 175: Maybe we could do something like:

```go
newBlockList, newCompactedBlockList, err := p.pollTenantAndCreateIndex(ctx, tenantID, previous)
resultCh <- result{
		tenantID:           tenantID,
		blockList:          newBlockList,
		compactedBlockList: newCompactedBlockList,
		err:                err,
}
```

And handle the errors outside the coroutine to avoid the mutex?

integration/poller/poller_test.go
```diff
@@ -4,8 +4,10 @@ import (
 	"bytes"
 	"context"
 	"crypto/rand"
+	mathrand "math/rand"
```
javiermolinar, line 7: Probably not very relevant at the moment but I think we can already use [v2](https://pkg.go.dev/math/rand/v2)

tempodb/blocklist/poller.go
joe-elliott: i personally like the simplicity of this approach. it's quite easy to read. i wouldn't be opposed to a struct through a channel if we really felt we needed it.

joe-elliott (commented):
this lgtm. not approving b/c it's still marked draft

integration/poller/poller_test.go
```diff
@@ -455,3 +444,50 @@ func writeBadBlockFiles(t *testing.T, ww backend.RawWriter, rr backend.RawReader
 	require.NoError(t, err)
 	t.Logf("items: %v", found)
 }
+
+func pushBlocksToTenant(t *testing.T, tenant string, bb [][]byte, w backend.Writer) []uuid.UUID {
+	// Randomly pick a block boundary
+	r := mathrand.Intn(len(bb))
```
zalegrala, line 450: Log something with regard to the random number here.

integration/poller/poller_test.go
zalegrala: Looks like it supports the use case, so I've pushed a commit for this.

zalegrala:
I've modified the error handling to be consistent, which required keeping of a `[]error` for the duration of the `Do()` call.  This allows us to know if the sequence of errors matches the sequence of the tenants.  This might be less important than the order in which the errors arrive, but at least its consistent.

tempodb/blocklist/poller.go
```diff
@@ -148,44 +150,73 @@ func (p *Poller) Do(previous *List) (PerTenant, PerTenantCompacted, error) {
...
-		newBlockList, newCompactedBlockList, err := p.pollTenantAndCreateIndex(ctx, tenantID, previous)
-		if err != nil {
-			level.Error(p.logger).Log("msg", "failed to poll or create index for tenant", "tenant", tenantID, "err", err)
-			consecutiveErrors++
-			if consecutiveErrors > p.cfg.TolerateConsecutiveErrors {
-				level.Error(p.logger).Log("msg", "exiting polling loop early because too many errors", "errCount", consecutiveErrors)
-				return nil, nil, err
-			}
+		finalErr atomic.Error
+		pollErrs = make([]error, len(tenants))
```
joe-elliott, line 161: instead of tracking errors with this slice i think we can do something like this in the go func():

```
go func(tenantID string) {
			defer wg.Done()
			var err error 
			errLefts := p.cfg.TolerateConsecutiveErrors

                         for errsLeft > 0 {
   			     newBlockList, newCompactedBlockList, err := p.pollTenantAndCreateIndex(ctx, tenantID, previous)
   			     if err == nil {
   			        break
   			    }
   			    
   			    // log error and errsLeft count
   			    errsLeft--
   			}
   			
   			if err != nil {
   			   // fatal err!
   			}
   			...
```
			
I like that this keeps the bookkeeping of errors local to the function and doesn't require tracking it with a slice across all tenants.

joe-elliott (commented):
looking good. one thought.

can we add the new config option to the docs?

tempodb/blocklist/poller.go
zalegrala: Hmm, interesting.  So this would move the error toleration from global to a per-tenant, which sounds pretty nice.  I think this means we could fail all but one tenant by default and still call the poll a success.  I'm not sure how I feel about that.

Since this used to be sequential, checking consecutive error checking here made sense, but I don't know what to make of it when running concurrently.  Does this just need to become a "total errors while polling" config option, rather than a "consequetive" toleration?  Then we could do like you say, but instead of `errLefts` being defined in the routine, we move it to the main tenant loop and do something like you've got there.

zalegrala: I added a commit for this, and an additional config for the new behavior.  I think we don't want a single failing tenant to bring down the entire poll by default.  This sort of merges old behavior with new behavior, and includes the encapsulation mentioned above.  I think it looks pretty tidy at this point, but let me know your thoughts.  Let's sync if we need to.

tempodb/blocklist/poller.go
```diff
@@ -151,44 +154,90 @@ func (p *Poller) Do(previous *List) (PerTenant, PerTenantCompacted, error) {
...
 	for _, tenantID := range tenants {
-		newBlockList, newCompactedBlockList, err := p.pollTenantAndCreateIndex(ctx, tenantID, previous)
-		if err != nil {
-			level.Error(p.logger).Log("msg", "failed to poll or create index for tenant", "tenant", tenantID, "err", err)
-			consecutiveErrors++
-			if consecutiveErrors > p.cfg.TolerateConsecutiveErrors {
-				level.Error(p.logger).Log("msg", "exiting polling loop early because too many errors", "errCount", consecutiveErrors)
-				return nil, nil, err
+		// Exit early if we have exceeded our tolerance for number of failing tenants.
+		if tenantFailures.Load() >= int32(tenantFailuresRemaining) {
```
joe-elliott, line 172: i don't think you need tenantFailuresRemaining. i believe the check on line 237 is correct. (comparing number of failed tenants vs the config'ed value).

there's also a race on tenantFailuresRemaining b/c you're checking it here not under lock and modifying it  on line 208.

final question:
previously we'd attempt to poll a tenant multiple times based on the consecutive errors config. if any tenant failed the configurable number of times we'd return an error from this func. this change will make it so a tenant can fail to poll and this function doesn't return an error. 

i'm fine with this behavior change if we feel like it is the right choice operationally.

tempodb/blocklist/poller.go
```diff
@@ -151,44 +154,90 @@ func (p *Poller) Do(previous *List) (PerTenant, PerTenantCompacted, error) {
...
+			defer mtx.Unlock()
+
+			if err != nil {
+				level.Error(p.logger).Log("msg", "failed to poll or create index for tenant", "tenant", tenantID, "err", err)
+				blocklist[tenantID] = previous.Metas(tenantID)
+				compactedBlocklist[tenantID] = previous.CompactedMetas(tenantID)
+
+				tenantFailures.Inc()
+				tenantFailuresRemaining--
+				finalErr.Store(err)
```
joe-elliott, line 209: instead of storing and returning only 1 error, let's just return a generic error on line 239.
```
	if tenantFailures.Load() > int32(p.cfg.TolerateTenantFailures) {
		return errors.New("too many tenant failures. abandoning polling cycle")
	}
```
just returning a single error could be misleading. if we return this one error we can review the logs and see all the errors that caused this method to fail.

tempodb/blocklist/poller.go
```diff
@@ -151,44 +154,90 @@ func (p *Poller) Do(previous *List) (PerTenant, PerTenantCompacted, error) {
...
-			level.Error(p.logger).Log("msg", "failed to poll or create index for tenant", "tenant", tenantID, "err", err)
-			consecutiveErrors++
-			if consecutiveErrors > p.cfg.TolerateConsecutiveErrors {
-				level.Error(p.logger).Log("msg", "exiting polling loop early because too many errors", "errCount", consecutiveErrors)
-				return nil, nil, err
+		// Exit early if we have exceeded our tolerance for number of failing tenants.
+		if tenantFailures.Load() >= int32(tenantFailuresRemaining) {
+			if lastErr := finalErr.Load(); lastErr != nil {
+				level.Error(p.logger).Log("msg", "exiting polling loop early because too many errors")
+				return nil, nil, lastErr
```
joe-elliott, line 175: i wouldn't return here. just `break` if we've had too many tenant failures and allow the method to exit naturally due to the check on 237-239.

zalegrala: Good call, cleaner with fewer places to reason about the return.

tempodb/blocklist/poller.go
zalegrala: The check at the top is to break early, so we don't continue to try and poll tenants after we exceeded our tolerance for failing tenants.  The check at the bottom is so that we catch exceeding the threshold if it happens on the last tenant.  I think we need both, but let's chat if you're seeing something else.  Maybe the one at the end is questionable since if we failed on the last tenant, perhaps better behavior is to return what we have worked for?

Nice, I see the race, I'll add another atomic for this.

One slight correction; previously there was no retry on a single tenant.  A tenant would fail, and we would store the previous results.  The previous early exit condition was based on the number of tenants failing in a row.  If more than one tenant failed (default tolerance of 1 failure) in a row then we would return an error.  Now we both retry failed tenants once by default, and we tolerate one failed tenant.  Operationally I like this pretty well, since if a tenant fails and then the retry succeeds, it doesn't stop the show.  Each tenant could fail once, but as long as they all succeed on the retry, we still have results.

What do you think about the change in behavior?  We can chat this afternoon.

tempodb/blocklist/poller.go
zalegrala: I pushed a commit to clean this up a bit and return a generic error.

joe-elliott (approved):
let's go!

PR #3649, opened 2024-05-07
vParquet4: update opentelemetry-proto and add scope attrs

stoewer:
**What this PR does**:

Follow up changes that were not included in the original vParquet4 PR

* Update the opentelemetry-proto to v1.3.0
  **Breaking Change** The update removes the deprecated `InstrumentationLibrary`  and `InstrumentationLibrarySpan` from the OTLP receivers
* Handle attributes for instrumentation scope correctly

Contribute to grafana/tempo-squad#299

**Checklist**
- [x] Tests updated
- [ ] Documentation added


mapno (commented):
This update removes the deprecated `instrumentation_library` fields. I believe this is a breaking change for the gRPC API. It's been deprecated for over 2 years now, but I think we should still note it somewhere.

mapno (approved):

Issue #3601, opened 2024-04-23
Tempo is unable to create directories when using local storage


AdrianColaianni:
**Describe the bug**
When starting Tempo for the first time, it is unable to create the directories in it's local storage.

**To Reproduce**
1. `git clone https://github.com/grafana/tempo`
2. `cd tempo/examples/docker-compose/local`
3. `docker-compose up -dV`
4. `docker logs local-tempo-1` and the following error will show
```
level=error ts=2024-04-23T14:06:56.550397594Z caller=main.go:121 msg="error running Tempo" err="failed to init module services: error initialising module: store: failed to create store: mkdir /var/tempo/blocks: permission denied"
```

**Expected behavior**
Tempo creates the directories and uses the storage

**Environment:**
 - Infrastructure: Linux 6.8.7-arch1-1 x86_64
 - Deployment tool: docker v26.0.2 & docker-compose v2.26.1

AdrianColaianni:
I am on main at `148ba6d` and my grafana/tempo image is `d523ba7846350e8e529c3f65effcfab8d97a57f3385ce5a6bc09f3b428f97fbe`

zalegrala:
Hmm, I don't have `148ba6d` locally, but try to pull an updated `main` and run the docker-compose example again.

zalegrala:
Oh I see now, you were ahead of me.  You may need to chown the path for the data to be owned by `10001:10001`.

AdrianColaianni:
Thank you zalegrala,

Changing ownership fixed the issue, but this should be documented or automated during startup.

PR #3657, opened 2024-05-08
Update deb and rpm packages to create and assign permissions for /var/tempo

mdisibio:
**What this PR does**:
The motivation for this PR is to fix the deb and rpm Drone test process, but we decided it's also a good improvement that should be shipped in the packages.

The deb and rpm packages install Tempo as a systemd service, under a newly-created user `tempo`.  However the services can't start out of the box because the `tempo` user lacks permissions to create and write to `/var/tempo` which is the default storage path.   This updates the post install process to create and assign perms to `/var/tempo`.  Unfortunately we couldn't pin down when this became a problem or what changed.

**Note**  A tricky bit is that Drone only runs for tags and PRs of branches in the Tempo repo itself.  If you PR is from a fork, it doesn't run. This made the failures go unnoticed for some time as most PRs are from forks.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/tools/packaging/tempo-postinstall.sh
+++ b/tools/packaging/tempo-postinstall.sh
@@ -17,7 +17,7 @@ cleanInstall() {
     fi
 
     # Create and assign permissions for default storage
-    mkdir /var/tempo
+    mkdir -m 0700 /var/tempo
     chown tempo /var/tempo
 
     # rhel/centos7 cannot use ExecStartPre=+ to specify the pre start should be run as root
--- a/tools/packaging/tempo-postinstall.sh
+++ b/tools/packaging/tempo-postinstall.sh
@@ -16,6 +16,10 @@ cleanInstall() {
         adduser --system --shell /bin/false "tempo"
     fi
 
+    # Create and assign permissions for default storage
+    mkdir /var/tempo
+    chown tempo /var/tempo
+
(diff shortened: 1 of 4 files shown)


tools/packaging/tempo-postinstall.sh
```diff
@@ -16,6 +16,10 @@ cleanInstall() {
         adduser --system --shell /bin/false "tempo"
     fi
 
+    # Create and assign permissions for default storage
+    mkdir /var/tempo
```
zalegrala, line 20: maybe `mkdir -m 0700 /var/tempo` so that non-privileged users can't read block data, etc.

zalegrala (approved):
One comment, but looks good to me.

Enrice:
this should REALLY be fixed in the docker distributions too!

zalegrala:
@Enrice I'm not sure what you mean.  Are you encountering an issue?

Enrice:
My issue is that I would need to apply the workaround depicted in #3601 which I cannot do at the moment.

zalegrala:
If you'd like to give us more information about your environment, we can help you work through the issue.  Consider joining the community slack.  I'm zleslie in there and would be happy to take a look.

PR #3670, opened 2024-05-13
[Traceql] Add support for searching by trace:id and span:id

ie-pham:
**What this PR does**: Add support for searching by trace:id and span:id. One thing to note is that when comparing bytes array, Tempo will left trim "x00". This means that `[]byte{0x02} == []byte{0x00, 0x02}`

Fixes #<issue number>
Screenshot 2024-05-13 at 5 47 16 PM

**Checklist**
- [x] Tests updated
- [x] Documentation added


tempodb/encoding/vparquet2/block_traceql.go
```diff
@@ -999,7 +1000,7 @@ func createSpanIterator(makeIter makeIterFn, primaryIter parquetquery.Iterator,
 		switch cond.Attribute.Intrinsic {
 		case traceql.IntrinsicSpanID:
 
-			pred, err := createStringPredicate(cond.Op, cond.Operands)
+			pred, err := createBytesPredicate(cond.Op, cond.Operands, true)
```
ie-pham, line 1003: placeholder until https://github.com/grafana/tempo/pull/3667 is merged with the new predicates

pkg/util/traceid.go
```diff
@@ -9,7 +9,7 @@ import (
 	"unsafe"
 )
 
-func HexStringToTraceID(id string) ([]byte, error) {
+func HexStringToID(id string, isSpan bool) ([]byte, error) {
```
stoewer, line 12: The function `HexStringToID()` is no longer used outside this package (and probably should no longer be used, as it has merely become a helper function). Consider making the function private and maybe move it below the functions `HexStringToTraceID()` and `HexStringToSpanID()`

stoewer:
When I use an ID string that is too long when querying or `span:id` or `trace:id` (e.g.  `span:id="ffffffffffffffff0"`) I'm seeing the following error: 

```
Query error
rpc error: code = Internal desc = error querying ingesters in Querier.Search: failed to execute f() for 127.0.0.1:9095: rpc error: code = Unknown desc = creating fetch iter: error creating iterator: creating span iterator: cannot convert hex to bytes array: ffffffffffffffff0
```

This is of course not wrong. But I'm wondering whether it would be better to just not return any matches

tempodb/encoding/vparquet2/block_traceql.go
joe-elliott: The new predicates will not have a bytes predicate. I would just write a new one in `pkg/parquetquery/predicates.go`.

ie-pham: Since the string predicate actually uses bytes array, reusing the string predicate works well. Any concerns about this?

joe-elliott: nope. not if it does what we want

stoewer (approved):
LGTM
Awesome PR!

PR #3674, opened 2024-05-14
Pin the goreleaser version for Go version support

zalegrala:
**What this PR does**:

Without this change, goreleaser@latest requires a newer version of Go than is currently supported by the project.  Here we pin the version.

**Checklist**
- [x] CI update
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/build/tools.mk
+++ b/build/tools.mk
@@ -12,10 +12,11 @@ TOOL_DIR     ?= tools
 TOOL_CONFIG  ?= $(TOOL_DIR)/tools.go
 
 TOOLS_IMAGE ?= grafana/tempo-ci-tools
+TOOLS_IMAGE_TAG ?= main-d8c69d4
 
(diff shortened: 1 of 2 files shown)


build/tools.mk
```diff
@@ -12,10 +12,11 @@ TOOL_DIR     ?= tools
 TOOL_CONFIG  ?= $(TOOL_DIR)/tools.go
 
 TOOLS_IMAGE ?= grafana/tempo-ci-tools
+TOOLS_IMAGE_TAG ?= main-d8c69d4
```
zalegrala, line 15: The `rXX` tags also tag `latest` which is strange, since the `rXX` are always behind `main`.  We can follow up with a better idea later, but this at least will use a tools image which has the expected tools.

joe-elliott (approved):
"latest" strikes again

electron0zero (approved):

PR #3684, opened 2024-05-16
Fixed multiple typos

xogoodnow:
Fixed multiple typos within the project

PR #3685, opened 2024-05-16
"ingestion_rate_limit_bytes" is set too short

xogoodnow:
**What this PR does**:
Fixing integration test for e2e suite

The "ingestion_rate_limit_bytes" within "integration/e2e/config-limits-partial-success.yaml"is set too short and I causes the integration test for e2e suite to fail

Increasing this value would prevent the job to fail in the future (When the job ran again it did not fail for the second time but in any case increasing the value would prevent future failures)

Here is the PR that lead to this:
https://github.com/grafana/tempo/pull/3684

Here is the link to the failed pipeline:
https://github.com/grafana/tempo/actions/runs/9114003097/job/25057452018?pr=3684#step:4:591

Screenshot 2024-05-16 at 6 53 12 PM


xogoodnow:
I sincerely apologize for the mess. :poop:
I mistakenly merged another branch into this.
I would appreciate it if you could squash all the commits together. 🙈

joe-elliott:
Thanks!

PR #3687, opened 2024-05-18
Remove hardcoded sleep in shutdown for distributor

### LINKED ISSUE #2353: Shutting down tempo always takes at least 30 seconds
**Describe the bug**

Shutting down the tempo service takes 30 seconds. This is annoying when rebooting because it holds up the reboot. I assume it's because [of this code in the distributor](https://github.com/grafana/tempo/blob/469c3337763d7ce75bdcdf74dce0730932753b3e/modules/distributor/receiver/shim.go#L253-L278) that just sleeps for 30 seconds

**To Reproduce**

1. Run tempo
2. SIGTERM tempo
3. It takes 30 seconds

**Expected behavior**

It should shut down when all requests have been properly processed

**Environment:**

I'm running tempo in a systemd service on NixOS, the configuration for it can be found here: https://github.com/NixOS/nixpkgs/blob/master/pkgs/servers/tracing/tempo/default.nix

```
tempo, version 2.0.0 (branch: <release>, revision: 2.0.0)
  build user:
  build date:
  go version:       go1.19.5
  platform:         linux/amd64
```

chodges15:
**What this PR does**:
This sleep is being removed in favor of using the configuration item `shutdown_delay` for generic server shutdown across all Tempo components introduced in #3395.

Fixes #2353 

**Checklist**
- [ ] Tests updated **[N/A, existing coverage]**
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## main / unreleased
 
+* [ENHANCEMENT] Remove hardcoded delay in distributor shutdown [#3687](https://github.com/grafana/tempo/pull/3687) (@chodges15)
 * [ENHANCEMENT] Update OTLP and add attributes to instrumentation scope in vParquet4 [#3649](https://github.com/grafana/tempo/pull/3649) (@stoewer)
   **Breaking Change** The update to OTLP 1.3.0 removes the deprecated `InstrumentationLibrary`
   and `InstrumentationLibrarySpan` from the OTLP receivers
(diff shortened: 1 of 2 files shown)


joe-elliott:
I had forgotten about this. We semi-recently added a `shutdown_delay` option that works for all components:

https://github.com/grafana/tempo/pull/3395

I am fine with just removing the lines that sleep for 30s. They were not well thought out when they were added, but have never bothered us enough to remove.

knylander-grafana (approved):
Approving doc update. Thank you for adding doc update withj your PR!

chodges15:
@joe-elliott that makes way more sense, I should have used git blame to look at the existing `shutdown_delay` and see it was added somewhat recently. Was really scratching my head as to why Tempo needed two.

joe-elliott:
I've been wanting to get rid of this forever and keep forgetting to just remove it. Thanks so much for getting this done!

PR #3691, opened 2024-05-20
Allow flushing blocks from the local-blocks processor [2 of 3]

mapno:
**What this PR does**:

This PR adds the ability to the local-blocks processor to flush blocks to storage, with the objective of flushing blocks with a replication factor of 1 (see previous PR: https://github.com/grafana/tempo/pull/3628).

It adds the `store` module as a dependency of the `metrics-generator`. The module is only instantiated when the storage is configured, working as before when it is not configured.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


cmd/tempo/app/modules.go
```diff
@@ -437,6 +437,12 @@ func (t *App) initCompactor() (services.Service, error) {
 }
 
 func (t *App) initStore() (services.Service, error) {
+	// Used by the local-blocs processor to flush RF1 blocks to storage.
+	// Only initialize if it's configured.
+	if t.cfg.Target == MetricsGenerator && t.cfg.StorageConfig.Trace.Backend == "" {
+		return services.NewIdleService(nil, nil), nil
+	}
```
mapno, line 444: Other alternatives considered:
* New `MaybeStore` module that contains this logic
* Configurable feature flag, eg. `metrics_generator_flush_to_storage`

mdisibio: Yes this is interesting. There is definitely a use case for generators without object store access.  I think the combination of both sounds the best?   The service has a dependency on the new module `MaybeStore/OptionalStore` which only starts if storage configured.  I like the approach of returning an IdleService.  Then the generator only flushes when `flush_to_storage=true`.  The generator can return a startup error if configured to flush but storage isn't.

modules/generator/processor/localblocks/processor.go
```diff
@@ -238,6 +255,38 @@ func (p *Processor) completeLoop() {
...
+			return
+		}
+
+		op := o.(*flushOp)
+		op.attempts++
+		err := p.flushBlock(op.blockID)
+		if err != nil {
+			_ = level.Error(p.logger).Log("msg", "failed to flush a block", "err", err)
+
+			if op.attempts < 3 {
```
mdisibio, line 278: Let's copy the behavior of the ingester, which doesn't limit retries on flushing to object storage. This will make it tolerate network or other transient errors as long as needed.

modules/generator/processor/localblocks/processor.go
```diff
@@ -310,6 +365,24 @@ func (p *Processor) completeBlock() error {
...
+		return nil
+	}
+
+	ctx := context.Background()
+	err := p.writer.WriteBlock(ctx, completeBlock)
+	if err != nil {
+		return err
+	}
+	metricFlushedBlocks.WithLabelValues(p.tenant).Inc()
+	return nil
```
mdisibio, line 383: As discussed offline, let's add a log message with the tenant and block ID.  level=info seems better than debug, and log volume isn't a concern.

mapno: Added!

modules/generator/processor/localblocks/processor.go
mapno: Done!

mapno: Isn't allowing infinite retries dangerous?

cmd/tempo/app/modules.go
mapno: Changed! Now it uses both approaches.

cmd/tempo/app/modules.go
```diff
@@ -436,6 +437,21 @@ func (t *App) initCompactor() (services.Service, error) {
 	return t.compactor, nil
 }
 
+func (t *App) initOptionalStore() (services.Service, error) {
+	// Used by the local-blocs processor to flush RF1 blocks to storage.
+	// Only initialize if it's configured.
+	if t.cfg.StorageConfig.Trace.Backend == "" {
+
+		if t.cfg.Target == MetricsGenerator && t.cfg.Generator.Processor.LocalBlocks.FlushToStorage {
```
mdisibio, line 445: nit: Can we move this check to initGenerator?

mdisibio (approved):
Nice work!  One more small nit but going ahead and approving.

cmd/tempo/app/modules.go
mapno: Good point, done

PR #3698, opened 2024-05-21
Fix TraceQL failures on boolean operations with statics and attributes

joe-elliott:
**What this PR does**:
Fixes an issue where TraceQL will pass a bad condition to the fetch layer when you use add a static and attribute such as:
```
{ span.http.status_code > span.http.status_code + 1 }
```

Currently this returns an error:

image

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/ast_conditions.go
+++ b/pkg/traceql/ast_conditions.go
@@ -44,7 +44,7 @@ func (o *BinaryOperation) extractConditions(request *FetchSpansRequest) {
 	case Attribute:
 		switch o.RHS.(type) {
 		case Static:
-			if o.RHS.(Static).Type == TypeNil && o.Op == OpNotEqual {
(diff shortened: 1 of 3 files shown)


pkg/traceql/ast_conditions.go
```diff
@@ -44,7 +44,7 @@ func (o *BinaryOperation) extractConditions(request *FetchSpansRequest) {
 	case Attribute:
 		switch o.RHS.(type) {
 		case Static:
-			if o.RHS.(Static).Type == TypeNil && o.Op == OpNotEqual {
+			if o.RHS.(Static).Type == TypeNil && o.Op == OpNotEqual || !o.Op.isBoolean() { // the fetch layer can't build predicates on operators that are not boolean
```
mdisibio, line 47: Let's add parens to make the `x && y || z` logic more clear

mdisibio (approved):

PR #3707, opened 2024-05-22
TraceQL: Collapse statics

joe-elliott:
**What this PR does**:
On creation of binary and unary operations this PR attempts to collapse values down to a single static where possible. This allows the fetch layer to directly apply predicates for conditions like `{ span.foo = 1 + 1 }` b/c they turn into `{ span.foo = 2 }`.

In order to make this PR work I needed to split out integer math from float. Currently the engine processes all values as floats. However, it's important not to change the type when collapsing or it breaks the fetch layer. `{ span.foo = 1 + 1 }` needs to collapse to `{ span.foo = 2 }` and not `{ span.foo = 2. }`

This has interesting consequences. Note the changed test. This used to be true `{ 1 / 10 = .1 }` but, due to integer division it is not. These two statements are now true: `{ 1 / 10. = .1 }` and `{ 1 / 10 = 0 }`. Overall I don't mind this. Generally TraceQL is quite strict about datatypes and this was an exception.

**Other options**

This whole exercise started when I noticed how bad performance was on `{ nestedSetParent = -1 }` which I realized was due to an `OpNone` condition being passed to the fetch layer instead of `OpEqual` and -1. In main -1 is parsed as an `UnaryOperation` with `OpSub` and 1.

We could just subtly change the parser by adding lines like these to fieldExpression:

```
  | fieldExpression AND fieldExpression      { $$ = newBinaryOperation(OpAnd, $1, $3) }
  | fieldExpression OR fieldExpression       { $$ = newBinaryOperation(OpOr, $1, $3) }
+ | SUB INTEGER                      { $$ = newStaticInt(-$2) }
+ | SUB FLOAT                      { $$ = newStaticFloat(-$2) }
  | SUB fieldExpression                      { $$ = newUnaryOperation(OpSub, $2) }
  | NOT fieldExpression                      { $$ = newUnaryOperation(OpNot, $2) }
  | static                                   { $$ = $1 }
```

This solves the -N case, but nothing else. It also introduces a lot of conflicts in the language, but testing suggests this is fine. Perhaps we prefer this b/c it imposes no other changes on TraceQL? Overall I prefer my proposed solution but am game to try the simpler thing.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/ast.go
+++ b/pkg/traceql/ast.go
@@ -402,12 +402,20 @@ type BinaryOperation struct {
 	compiledExpression *regexp.Regexp
 }
 
-func newBinaryOperation(op Operator, lhs, rhs FieldExpression) *BinaryOperation {
-	return &BinaryOperation{
+func newBinaryOperation(op Operator, lhs, rhs FieldExpression) FieldExpression {
+	binop := &BinaryOperation{
 		Op:  op,
 		LHS: lhs,
 		RHS: rhs,
 	}
+
+	if !binop.referencesSpan() && binop.validate() == nil {
+		if simplified, err := binop.execute(nil); err == nil {
+			return simplified
+		}
+	}
+
+	return binop
 }
 
 // nolint: revive
@@ -437,11 +445,19 @@ type UnaryOperation struct {
 	Expression FieldExpression
 }
 
-func newUnaryOperation(op Operator, e FieldExpression) UnaryOperation {
-	return UnaryOperation{
+func newUnaryOperation(op Operator, e FieldExpression) FieldExpression {
+	unop := UnaryOperation{
 		Op:         op,
 		Expression: e,
 	}
+
+	if !unop.referencesSpan() && unop.validate() == nil {
+		if simplified, err := unop.execute(nil); err == nil {
+			return simplified
+		}
+	}
+
+	return unop
 }
 
 // nolint: revive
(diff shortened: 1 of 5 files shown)


joe-elliott:
Benches of `{ nestedSetParent = -1 }`
```
goos: darwin
goarch: arm64
pkg: github.com/grafana/tempo/tempodb/encoding/vparquet3
BackendBlockTraceQL/=_-1-11   524.5m ± ∞ ¹   107.7m ± ∞ ¹  -79.46% (p=0.008 n=5)
geomean                       239.5m         107.1m        -55.27%

                            │  before.txt   │               after.txt                │
                            │      B/s      │      B/s        vs base                │
BackendBlockTraceQL/=_-1-11   62.08Mi ± ∞ ¹   302.31Mi ± ∞ ¹  +386.93% (p=0.008 n=5)
geomean                       136.0Mi          304.0Mi        +123.59%

                            │ before.txt  │              after.txt              │
                            │  MB_io/op   │  MB_io/op    vs base                │
BackendBlockTraceQL/=_-1-11   34.15 ± ∞ ¹   34.15 ± ∞ ¹       ~ (p=1.000 n=5) ²
geomean                       34.15         34.15        +0.00%
² all samples are equal

BackendBlockTraceQL/=_-1-11   591.27Mi ± ∞ ¹   40.82Mi ± ∞ ¹  -93.10% (p=0.008 n=5)
geomean                        169.1Mi         40.76Mi        -75.89%

                            │  before.txt   │              after.txt              │
                            │   allocs/op   │  allocs/op    vs base               │
BackendBlockTraceQL/=_-1-11   4430.2k ± ∞ ¹   482.5k ± ∞ ¹  -89.11% (p=0.008 n=5)
geomean                        1.574M         482.5k        -69.35%
```

PR #3708, opened 2024-05-22
TraceQL event intrinsic event:name

stoewer:
**What this PR does**:

Add `event` scope and the intrinsic `event:name`

Autocomplete is currently not working, but is also part of a separate issue.

Contributes to grafana/tempo-squad#424

**Checklist**
- [x] Tests updated
- [ ] Documentation added


stoewer:
image

stoewer:
There are two behaviors that I'm not quite sure about. It would be great to get feedback about them

1. The query `{ event:name != "exception" }` also returns spans without any events at all. Is this correct or should we only return spans with events that don't have the specified name?
2. When running a with `query:name` on vParquet2 or vParquet3 it shows the following error: *"rpc error: code = Internal desc = error querying ingesters in Querier.Search: failed to execute f() for 127.0.0.1:9095: rpc error: code = Unknown desc = creating fetch iter: error creating iterator: unsupported traceql scope: event:name"*. Do we want to show this error, or should we rather just show no results?

tempodb/encoding/vparquet3/block_traceql.go
```diff
@@ -833,6 +834,7 @@ var intrinsicColumnLookups = map[traceql.Intrinsic]struct {
 	traceql.IntrinsicTraceID:          {intrinsicScopeTrace, traceql.TypeString, columnPathTraceID},
 	traceql.IntrinsicTraceStartTime:   {intrinsicScopeTrace, traceql.TypeDuration, columnPathStartTimeUnixNano},
 
+	traceql.IntrinsicEventName:    {intrinsicScopeEvent, traceql.TypeNil, ""}, // Not used in vparquet3, this entry is only used to assign default scope.
```
mdisibio, line 837: Since this isn't supported in vParquet2/3, I think we need to ensure that a query using event intrinsics doesn't execute.  Currently, it looks like the condition will be silently dropped while fetching, which means the query results could be incorrect:   the `{name=x && event:name=y}` will be executed against a vp2/3 block as `{name=x}`.    checkConditions could return ErrUnsupported if any event scope condition is present.

tempodb/encoding/vparquet4/block_traceql.go
```diff
@@ -1445,6 +1492,31 @@ func createAllIterator(ctx context.Context, primaryIter parquetquery.Iterator, c
...
+			}
+			eventIters = append(eventIters, makeIter(columnPathEventName, pred, columnPathEventName))
+		}
+	}
+
+	if primaryIter != nil {
+		eventIters = append(eventIters, primaryIter)
+	}
+
+	return parquetquery.NewJoinIterator(4, eventIters, &eventCollector{}, parquetquery.WithPool(pqEventPool)), nil
```
mdisibio, line 1517: nit: let's create a const for the definitionlevel=4

tempodb/encoding/vparquet3/block_traceql.go
stoewer: Currently queries with `event:name` fail `categorizeConditions()`. But I agree, it's better to check this explicitly in `checkConditions()`. This also allows us to return a better error message

ie-pham:
LGTM! Awesome work! Will steal this for my PR 😄 

Nit: May I suggest a test in `block_traceql_meta_test.go` ?

```
{
			"Event name lookup",
			makeReq(
				parse(t, `{event:name = "e1"}`), // 
			),
			makeSpansets(
				makeSpanset(
					wantTr.TraceID,
					wantTr.RootSpanName,
					wantTr.RootServiceName,
					wantTr.StartTimeUnixNano,
					wantTr.DurationNano,
					&span{
						id:                 wantTr.ResourceSpans[0].ScopeSpans[0].Spans[0].SpanID,
						startTimeUnixNanos: wantTr.ResourceSpans[0].ScopeSpans[0].Spans[0].StartTimeUnixNano,
						durationNanos:      wantTr.ResourceSpans[0].ScopeSpans[0].Spans[0].DurationNano,
						spanAttrs: []attrVal{
							{traceql.NewIntrinsic(traceql.IntrinsicDuration), traceql.NewStaticDuration(100 * time.Second)},
							{traceql.NewIntrinsic(traceql.IntrinsicSpanID), traceql.NewStaticString(util.SpanIDToHexString(wantTr.ResourceSpans[0].ScopeSpans[0].Spans[0].SpanID))},
						},
						traceAttrs: []attrVal{
							{traceql.NewIntrinsic(traceql.IntrinsicTraceRootService), traceql.NewStaticString("RootService")},
							{traceql.NewIntrinsic(traceql.IntrinsicTraceRootSpan), traceql.NewStaticString("RootSpan")},
							{traceql.NewIntrinsic(traceql.IntrinsicTraceDuration), traceql.NewStaticDuration(100 * time.Millisecond)},
							{traceql.NewIntrinsic(traceql.IntrinsicTraceID), traceql.NewStaticString(util.TraceIDToHexString(wantTr.TraceID))},
						},
						eventAttrs: []attrVal{
							{traceql.NewIntrinsic(traceql.IntrinsicEventName), traceql.NewStaticString("e1")},
						},
					},
				),
			),
		},
```

stoewer:
> May I suggest a test in block_traceql_meta_test.go?

Good point, thank you. I've added the test case

ie-pham (approved):

PR #3711, opened 2024-05-23
Improve servicegraph handling of newer OTEL semantic conventions

### LINKED ISSUE #3010: [Metrics generator] Prioritise OTel DB semantic attributes over `db.name` and `db.system`
If a database node does not produce spans of its own, Tempo looks for `db.name` and `db.system` in the calling span's attributes to create an appropriate downstream edge and node to complete a service graph.

However, these two attributes actually pertain to the database name inside the downstream service and the database type, rather than the hostname of the database itself.

For example, this means that should there be several upstream services that call the same downstream database *service* on a specific host, but actually make requests to different database *names* within that service, you'll see several different service nodes generated, instead of just one that is called by multiple upstream services.

For `db.system`, this would actually have the reverse effect, where there may be multiple hosts serving databases of the same type, but which only produce a single service graph node that will appear to be called by all upstream services.

The [OpenTelemetry semantic conventions for connection level attributes](https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/database/#connection-level-attributes) is arguably what we probably want to prioritise instead, if they are present. I'd suggest maybe:
* `server.address`
* `server.socket.address`
* `db.connection_string` (bit handwavey about this, especially if it includes authentication details, although it's no worse than already seeing them in a span)

Should none of those connection attributes be present, then falling back to `db.name` and then `db.system` should occur, on the pretext that generating a node is better than not doing so.

You can, of course, use the attribute processor in a relevant collector to copy any relevant attribute that's already present to `peer.service`, `db.name` or `db.system` instead, but if these are already present then the potential for overwriting data that is actually of use is high. 

**Additional context**
N/A

zalegrala:
**What this PR does**:

Currently, the `db.name` attribute is used as the node identity on the service graph.  This leads to situations where the name of the database may be reused between services, leading to a single node on the graph where more than one actual service is in use.

This PR changes the node identity when newer OTEL semantic conventions are available as attributes on the span.  The following attributes are considered in order:
- `peer.service`
- `server.address`
- `network.peer.address` -- additionally `network.peer.port` will be used if available
- fall back to `db.name`

Fixes #3010

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * [FEATURE] TraceQL support for event scope and event:name intrinsic [#3708](https://github.com/grafana/tempo/pull/3708) (@stoewer)
 * [FEATURE] Flush blocks to storage from the metrics-generator [#3628](https://github.com/grafana/tempo/pull/3628) [#3691](https://github.com/grafana/tempo/pull/3691) (@mapno)
+* [ENHANCEMENT] Improve use of OTEL semantic conventions on the service graph [#3711](https://github.com/grafana/tempo/pull/3711) (@zalegrala)
 
 ## v2.5.0-rc.1
 
@@ -58,7 +59,6 @@
 * [ENHANCEMENT] TraceQL - Add support for scoped intrinsics using `:` [#3629](https://github.com/grafana/tempo/pull/3629) (@ie-pham)
   available scoped intrinsics: trace:duration, trace:rootName, trace:rootService, span:duration, span:kind, span:name, span:status, span:statusMessage
 * [ENHANCEMENT] Performance improvements on TraceQL and tag value search. [#3650](https://github.com/grafana/tempo/pull/3650),[#3667](https://github.com/grafana/tempo/pull/3667) (@joe-elliott)
-* [ENHANCEMENT] Improve use of OTEL semantic conventions on the service graph [#3711](https://github.com/grafana/tempo/pull/3711) (@zalegrala)
 * [BUGFIX] Fix handling of regex matchers in autocomplete endpoints [#3641](https://github.com/grafana/tempo/pull/3641) (@sd2k)
 * [BUGFIX] Update golang.org/x/net package to 0.24.0 to fix CVE-2023-45288 [#3613](https://github.com/grafana/tempo/pull/3613) (@pavolloffay)
 * [BUGFIX] Fix metrics query results when filtering and rating on the same attribute [#3428](https://github.com/grafana/tempo/issues/3428) (@mdisibio)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@
 * [ENHANCEMENT] TraceQL - Add support for scoped intrinsics using `:` [#3629](https://github.com/grafana/tempo/pull/3629) (@ie-pham)
   available scoped intrinsics: trace:duration, trace:rootName, trace:rootService, span:duration, span:kind, span:name, span:status, span:statusMessage
 * [ENHANCEMENT] Performance improvements on TraceQL and tag value search. [#3650](https://github.com/grafana/tempo/pull/3650),[#3667](https://github.com/grafana/tempo/pull/3667) (@joe-elliott)
+* [ENHANCEMENT] Improve use of OTEL semantic conventions on the service graph [#3711](https://github.com/grafana/tempo/pull/3711) (@zalegrala)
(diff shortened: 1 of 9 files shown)


modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -280,6 +273,72 @@ func (p *Processor) upsertPeerNode(e *store.Edge, spanAttr []*v1_common.KeyValue
...
+func (p *Processor) upsertDatabaseRequest(e *store.Edge, resourceAttr []*v1_common.KeyValue, span *v1_trace.Span) {
+	var (
+		isDatabase bool
+
+		// The fallback database name
+		dbName string
+	)
+
+	// Check for db.name first.  The dbName is set initially to maintain backwards compatbility.
+	if name, ok := processor_util.FindAttributeValue(string(semconv.DBNameKey), resourceAttr, span.Attributes); ok {
```
zalegrala, line 293: The pattern here loops over the span and resource attributes to find specific keys.  This means that if you use the preferred attributes in the spans, then the generator has to do less work.  I'd considered finding all available keys in one pass, but following the existing pattern elsewhere feels simpler.

hedss (approved):
From a GTM perspective, this is great, thank you Zach!

knylander-grafana (commented):

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -280,6 +273,72 @@ func (p *Processor) upsertPeerNode(e *store.Edge, spanAttr []*v1_common.KeyValue
...
+	if !isDatabase {
+		return
+	}
+	e.ConnectionType = store.Database
+	e.ServerLatencySec = spanDurationSec(span)
+
+	// Set the service name by order of precedence
+
+	// Check for peer.service
+	if name, ok := processor_util.FindAttributeValue(string(semconv.PeerServiceKey), resourceAttr, span.Attributes); ok {
```
joe-elliott, line 315: the `dbName` from `DBNameKey` above is being overwritten here. i think we need a condition above this that checks if name is already set

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -280,6 +273,72 @@ func (p *Processor) upsertPeerNode(e *store.Edge, spanAttr []*v1_common.KeyValue
...
+	// Check for server.address
+	if name, ok := processor_util.FindAttributeValue(string(semconv.ServerAddressKey), resourceAttr, span.Attributes); ok {
+		e.ServerService = name
+		return
+	}
+
+	// Check for network.peer.address and network.peer.port.  Use port if it is present.
+	if host, ok := processor_util.FindAttributeValue(string(semconv.NetworkPeerAddressKey), resourceAttr, span.Attributes); ok {
+		if port, ok := processor_util.FindAttributeValue(string(semconv.NetworkPeerPortKey), resourceAttr, span.Attributes); ok {
+			e.ServerService = fmt.Sprintf("%s:%s", host, port)
```
joe-elliott, line 329: little concerned about cardinality here. it should be fine b/c it should accurately reflect real databases in an org.

we had similar issues before and @mapno submitted this.

https://github.com/grafana/tempo/pull/2505

my guess is that it's fine, but we should keep an eye out for cardinality jumps in prod as we roll this out.

zalegrala: I thought about the carnality here too, but I think because this is server side information, we should be okay since I expect server name/port combos to be significantly less than the ephemeral ports in use by the client.  Good thing to watch out for when we roll for sure.

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@
 * [ENHANCEMENT] TraceQL - Add support for scoped intrinsics using `:` [#3629](https://github.com/grafana/tempo/pull/3629) (@ie-pham)
   available scoped intrinsics: trace:duration, trace:rootName, trace:rootService, span:duration, span:kind, span:name, span:status, span:statusMessage
 * [ENHANCEMENT] Performance improvements on TraceQL and tag value search. [#3650](https://github.com/grafana/tempo/pull/3650),[#3667](https://github.com/grafana/tempo/pull/3667) (@joe-elliott)
+* [ENHANCEMENT] Improve use of OTEL semantic conventions on the service graph [#3711](https://github.com/grafana/tempo/pull/3711) (@zalegrala)
```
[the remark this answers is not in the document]
zalegrala: Good catch, thank you.

PR #3713, opened 2024-05-24
Query-frontend: Parrallelize unmarshalling of responses

joe-elliott:
**What this PR does**:

Adds a configurable number of dedicated goroutines per request to consume the responses and add them to the combiner. The combiner was also adjusted to move the unmarshalling steps outside of the locking logic so we could do multiple unmarshals in parrallel.

- Added a parameter to configure the number of goroutines. Default: 10
- Added tests to confirm the combiner doesn't race with the new changes
- Found and fixed an issue where tag values v2 was not initializing its status code to 200 which caused it to fail if GRPCDiff was called before a response was received.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/frontend/pipeline/collector_http.go
+++ b/modules/frontend/pipeline/collector_http.go
@@ -3,8 +3,10 @@ package pipeline
 import (
 	"context"
 	"net/http"
+	"sync"
 
 	"github.com/grafana/tempo/modules/frontend/combiner"
+	"go.uber.org/atomic"
 )
 
 type httpCollector struct {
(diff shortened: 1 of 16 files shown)


modules/frontend/pipeline/collector_http.go
```diff
@@ -35,29 +39,79 @@ func (r httpCollector) RoundTrip(req *http.Request) (*http.Response, error) {
...
 
+	err = addNextAsync(ctx, r.consumers, resps, r.combiner, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return r.combiner.HTTPFinal()
+}
+
+func addNextAsync(ctx context.Context, consumers int, resps Responses[combiner.PipelineResponse], c combiner.Combiner, callback func() error) error {
```
mapno, line 50: Nit: I find the function name a bit generic for the pkg

modules/frontend/pipeline/collector_http.go
```diff
@@ -35,29 +39,79 @@ func (r httpCollector) RoundTrip(req *http.Request) (*http.Response, error) {
...
+	overallErr := atomic.Error{}
+	wg := sync.WaitGroup{}
+
+	setErr := func(err error) {
+		overallErr.CompareAndSwap(nil, err)
+	}
+
+	if consumers <= 0 {
+		consumers = 10
+	}
```
mapno, line 61: Unsure about this, I understand 0 as "sync".

mapno (approved):

knylander-grafana (commented):
Thank you for updating the config doc.

modules/frontend/pipeline/collector_http.go
joe-elliott: i consider `consumers == 1` sync and `consumers <= 0` as simply invalid which is why i chose a default. 

technically `consumers == 1` is synchronous just with silly channel shenanigans.

modules/frontend/pipeline/collector_http.go
joe-elliott: agreed. I'm going to change it to`consumeAndCombineResponses`. thoughts?

joe-elliott:
Fixed a flakey test and add http status code to the logged response on searches.

mapno (approved):

PR #3723, opened 2024-05-28
Read path for RF1 generator blocks [3 of 3]

mapno:
**What this PR does**:

Follow-up to https://github.com/grafana/tempo/pull/3691.

Adds the ability to query generator blocks (RF=1) for TraceQL metrics. If `frontend.metrics.query_generator_blocks` is set to true, the frontend will use the new query path and build requests only targeting blocks with `ReplicationFactor=1`.

This allows for more efficient query sharding—because there's no need to dedup data, querier jobs can target blocks or portions of them, instead of each job opening all blocks.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/querier/querier_query_range.go
```diff
@@ -24,6 +25,10 @@ func (q *Querier) QueryRange(ctx context.Context, req *tempopb.QueryRangeRequest
 		return q.queryRangeRecent(ctx, req)
 	}
 
+	if req.ShardCount == 0 { // RF1 search
```
mdisibio, line 28: I would swap this to look for BlockID

pkg/traceql/engine_metrics.go
```diff
@@ -628,6 +628,10 @@ func optimize(req *FetchSpansRequest) {
 			req.SecondPassConditions = nil
 		}
 	}
+
+	if len(req.SecondPassConditions) == 0 {
```
mdisibio, line 632: Note: This fixes a performance regression unrelated to this PR.

modules/querier/querier_query_range.go
```diff
@@ -59,6 +64,84 @@ func (q *Querier) queryRangeRecent(ctx context.Context, req *tempopb.QueryRangeR
...
+
+	eval, err := traceql.NewEngine().CompileMetricsQueryRange(req, false, timeOverlapCutoff, unsafe)
+	if err != nil {
+		return nil, err
+	}
+
+	f := traceql.NewSpansetFetcherWrapper(func(ctx context.Context, req traceql.FetchSpansRequest) (traceql.FetchSpansResponse, error) {
+		return q.store.Fetch(ctx, meta, req, opts)
+	})
+	err = eval.Do(ctx, f, 0, 0)
```
mdisibio, line 127: Since the request time range is trimmed to match the block, we can now fill in the start/end times here. This enables use of the trace-level columns to increase performance.  Long-term we should include the block start/end times in the query as new RF1 parameters, and it makes this clearer (and the step alignment can also be relocated here).

modules/frontend/metrics_query_range_sharder.go
```diff
@@ -319,12 +438,14 @@ func (s *queryRangeSharder) generatorRequest(searchReq tempopb.QueryRangeRequest
 
 	searchReq.QueryMode = querier.QueryModeRecent
 
-	// No sharding on the generators (unnecessary), but we do apply sampling
-	// rates.  In this case we execute a single arbitrary shard. Choosing
-	// the last shard works. The first shard should be avoided because it is
-	// weighted slightly off due to int63/128 sharding boundaries.
-	searchReq.ShardID = uint32(1.0 / samplingRate)
-	searchReq.ShardCount = uint32(1.0 / samplingRate)
+	if s.replicationFactor == backend.DefaultReplicationFactor {
```
mdisibio, line 441: I'm ok with deleting this entirely.  Otherwise the check could be clearer against s.cfg.QueryGeneratorBlocks?

modules/frontend/metrics_query_range_sharder.go
```diff
@@ -38,10 +39,15 @@ type QueryRangeSharderConfig struct {
 	MaxDuration           time.Duration `yaml:"max_duration"`
 	QueryBackendAfter     time.Duration `yaml:"query_backend_after,omitempty"`
 	Interval              time.Duration `yaml:"interval,omitempty"`
+	QueryGeneratorBlocks  bool          `yaml:"query_generator_blocks,omitempty"`
```
mdisibio, line 42: Would probably rename this to `rf1_read_path`, and then the logic related to setting things like `replicationFactor=1` is also clearer.

mdisibio (approved):

PR #3742, opened 2024-05-31
Fix block reload in local-blocks

mapno:
**What this PR does**:

Fix for bug introduced in https://github.com/grafana/tempo/pull/3691.

If `FlushToStorage` was set to `false` (default), complete blocks were not handled correctly:
* local blocks are no longer pushed to the queue to be flushed during reload (start-up)
* complete blocks can now be deleted if old (`endTime < now + completeBlockTimeout`)

### DIFF
--- a/modules/generator/processor/localblocks/processor.go
+++ b/modules/generator/processor/localblocks/processor.go
@@ -634,20 +634,7 @@ func (p *Processor) deleteOldBlocks() (err error) {
 	}
 
 	for id, b := range p.completeBlocks {
-		if p.Cfg.FlushToStorage {
-			flushedTime := b.FlushedTime()
-			if flushedTime.IsZero() {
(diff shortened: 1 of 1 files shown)


modules/generator/processor/localblocks/processor.go
```diff
@@ -634,6 +634,18 @@ func (p *Processor) deleteOldBlocks() (err error) {
 	}
 
 	for id, b := range p.completeBlocks {
+		if !p.Cfg.FlushToStorage {
+			if b.BlockMeta().EndTime.Before(before) {
+				level.Info(p.logger).Log("msg", "deleting complete block", "block", id.String())
```
mdisibio, line 639: Let's add this log line to the case below too

mdisibio (approved):

PR #3748, opened 2024-06-04
[wip]]TraceQL] add support for event attribute query

ie-pham:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


docs/sources/tempo/traceql/_index.md
```diff
@@ -128,6 +128,11 @@ Find any database connection string that goes to a Postgres or MySQL database:
 { span.db.system =~ "postgresql|mysql" }
 ```
 
+You can query for an exception in your span event
+```
+{ event.exception.message ~= "something went wrong" }
```
stoewer, line 133:
```suggestion
{ event.exception.message =~ "something went wrong" }
```

docs/sources/tempo/traceql/_index.md
```diff
@@ -101,7 +101,7 @@ possible to span-level intrinsics.
 
 ### Attribute fields
 
-There are two types of attributes: span attributes and resource attributes. By expanding a span in the Grafana UI, you can see both its span attributes (1 in the screenshot) and resource attributes (2 in the screenshot).
+There are three types of attributes: span attributes, resource attributes, and event attributes. By expanding a span in the Grafana UI, you can see both its span attributes (1 in the screenshot) and resource attributes (2 in the screenshot).
```
stoewer, line 104: I'm not sure about this one: What do you think about introducing the concept of scopes here instead of describing them as "attribute types"?
```suggestion
TraceQL has three different attribute scopes: span attributes, resource attributes, and event attributes. By expanding a span in the Grafana UI, you can see both its span attributes (1 in the screenshot) and resource attributes (2 in the screenshot).
```

tempodb/encoding/vparquet4/block_traceql.go
```diff
@@ -1364,27 +1369,38 @@ func (i *mergeSpansetIterator) Close() {
...
-//  Span column iterator N    ---------------------  |  |     |
-//    (ex: name, status)                          |  |  |     |
-//                                                V  V  V     V
-//                                            ------------------
-//                                            | span collector |
-//                                            ------------------
+//  Event attribute iterator: key    --------------------------------------------
+//                                                                            | |
+//  Event attribute iterator: valunN ---------------------------------------  | |
+//    											                            | | |
```
stoewer, line 1375: Nice doc update!!!

Minor formatting improvements:
```suggestion
//                            ...   ------------------------------------------- |
//  Event attribute iterator: valueN ---------------------------------------  | |
//                                                                          | | |
```

stoewer (commented):
Great PR! I tested the event attributes locally and everything worked as expected.

I left a couple of comments for improvements

pkg/parquetquery/iters.go
```diff
@@ -80,17 +80,41 @@ func EqualRowNumber(upToDefinitionLevel int, a, b RowNumber) bool {
...
+	case 4:
+		return RowNumber{t[0], t[1], t[2], t[3], t[4], -1, -1, -1}
+	case 5:
+		return RowNumber{t[0], t[1], t[2], t[3], t[4], t[5], -1, -1}
+	case 6:
+		return RowNumber{t[0], t[1], t[2], t[3], t[4], t[5], t[6], -1}
+	case 7:
+		return RowNumber{t[0], t[1], t[2], t[3], t[4], t[5], t[6], t[7]}
+	}
+	return TruncateRowNumber(definitionLevelToKeep, t)
```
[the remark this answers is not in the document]
ie-pham: Ah good catch! I meant to put the old truncateRowNumberSlow here but it might be good to log the error and return an empty row number instead.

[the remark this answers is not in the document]
stoewer: In my opinion logging and returning empty `RowNumber` is not a good way of error handling:
* It is not an idiomatic way to handle errors in Go
* Logging errors instead of handling them properly only delays discovery of bugs
* For the caller the function returns a potentially harmful value without returning an error. Callers might need to check for empty row number in order to detect "errors", which is not obvious for the caller

I think there are better ways to deal with invalid inputs here:
* Do the next best thing and treat `definitionLevelToKeep < 0` as 0 and `definitionLevelToKeep > MaxDefinitionLevel` as `MaxDefinitionLevel`. This is how the function behaved before.
* Simply panic (or log and panic). Code that calls `TruncateRowNumber()` is clearly a bug and should not happen at all. If someone adds definition levels to the schema in the future without increasing the size of `RowNumber` the bug is discovered quickly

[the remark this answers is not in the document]
ie-pham: That's a great point. It did take me a very long time to figure out why the code wasn't working at 6 levels before. 👍

pkg/parquetquery/iters.go
```diff
@@ -119,7 +115,11 @@ func TruncateRowNumber(definitionLevelToKeep int, t RowNumber) RowNumber {
 	case 7:
 		return RowNumber{t[0], t[1], t[2], t[3], t[4], t[5], t[6], t[7]}
 	}
-	level.Error(log.Logger).Log("msg", "definition level out of bound: should be [0:7] but got %d", definitionLevelToKeep)
+
+	if definitionLevelToKeep > MaxDefinitionLevel || definitionLevelToKeep < 0 {
+		level.Error(log.Logger).Log("msg", "definition level out of bound: should be [0:7] but got %d", definitionLevelToKeep)
+		panic("invalid definition level")
+	}
 	return EmptyRowNumber()
```
stoewer: This can just be the default case. This makes it also more clear that the `return EmptyRowNumber()` can never be reached (and thus be removed)

pkg/parquetquery/iters.go
```diff
@@ -543,6 +543,10 @@ func (t *RowNumber) Next(repetitionLevel, definitionLevel int) {
 			t[7] = -1
 		}
 	}
+	if definitionLevel > MaxDefinitionLevel || definitionLevel < 0 {
+		level.Error(log.Logger).Log("msg", "definition level out of bound: should be [0:7] but got %d", definitionLevel)
+		panic("invalid definition level")
+	}
```
stoewer: For all cases where the definition level is valid this code will make two additional assertions on `definitionLevel` that have already proven to be false in the switch statements. Normally this would not bother me, but since `Next()` is very performance critical, it could be worth improving this. For example:
* Make sure that this code is never reached by adding a return after each inner switch statement
* Panic instead in a default case of every inner switch statement

pkg/parquetquery/iters_test.go
```diff
@@ -43,21 +43,68 @@ func TestNext(t *testing.T) {
...
+		d := rand.Intn(MaxDefinitionLevel + 1)
+
+		newR := TruncateRowNumber(d, rn)
+		oldR := truncateRowNumberSlow(d, rn)
+
+		require.Equal(t, newR, oldR)
+	}
+}
+
+func TestInvalidDefinitionLevel(t *testing.T) {
```
stoewer, line 60: `nitpick` It is a bit unusual to test two different and only slightly related functions in the same test. It would be more idiomatic to split this in two tests `TestTruncateRowNumber_InvalidLevel` and `TestRowNumber_InvalidLevel (or similar names). Or as an alternative, add them as sub test to `TestTruncateRowNumber` and `TestRowNumber`

stoewer (approved):

docs/sources/tempo/traceql/_index.md
```diff
@@ -128,6 +128,11 @@ Find any database connection string that goes to a Postgres or MySQL database:
 { span.db.system =~ "postgresql|mysql" }
 ```
 
+You can query for an exception in your span event
+```
+{ event.exception.message =~ "something went wrong" }
```
joe-elliott, line 133: maybe 
```
{ event.exception.message =~ ".*something went wrong.*" }
```
or 
```
{ event.exception.message = "something went wrong" }
```

pkg/parquetquery/iters.go
```diff
@@ -30,14 +33,14 @@ import (
 //	    D      0,  1,  0
 //	  E        0,  2, -1
 //
-// Currently supports 6 levels of nesting which should be enough for anybody. :)
-type RowNumber [6]int32
+// Currently supports 8 levels of nesting which should be enough for anybody. :)
+type RowNumber [8]int32
```
joe-elliott, line 37: the perf hit for extending to 8 is not great. can we get two quick cpu profiles. one from main and one from this branch to identify and reduce the impact.

pkg/parquetquery/iters.go
```diff
@@ -80,17 +83,43 @@ func EqualRowNumber(upToDefinitionLevel int, a, b RowNumber) bool {
 		(a[2] == b[2] || upToDefinitionLevel < 2) &&
 		(a[3] == b[3] || upToDefinitionLevel < 3) &&
 		(a[4] == b[4] || upToDefinitionLevel < 4) &&
-		(a[5] == b[5] || upToDefinitionLevel < 5)
+		(a[5] == b[5] || upToDefinitionLevel < 5) &&
```
joe-elliott, line 86: i wonder if evaluating `upToDefinitionLevel < val` would be faster? i also wonder if this statement is too complex for the compiler to short circuit in the ways that make sense to us.

pkg/parquetquery/iters.go
```diff
@@ -80,17 +83,43 @@ func EqualRowNumber(upToDefinitionLevel int, a, b RowNumber) bool {
...
+		return RowNumber{t[0], t[1], t[2], t[3], t[4], -1, -1, -1}
+	case 5:
+		return RowNumber{t[0], t[1], t[2], t[3], t[4], t[5], -1, -1}
+	case 6:
+		return RowNumber{t[0], t[1], t[2], t[3], t[4], t[5], t[6], -1}
+	case 7:
+		return RowNumber{t[0], t[1], t[2], t[3], t[4], t[5], t[6], t[7]}
+	default:
+		level.Error(log.Logger).Log("msg", "definition level out of bound: should be [0:7] but got %d", definitionLevelToKeep)
+		panic("invalid definition level")
```
joe-elliott, line 119: fine with the panic, but would prefer not bringing a logging into this package. let's remove the logs and flesh out the panic message

use `panicWhenInvalidDefinitionLevel` below?

pkg/parquetquery/iters.go
```diff
@@ -123,36 +152,66 @@ func (t *RowNumber) Next(repetitionLevel, definitionLevel int) {
 			t[3] = -1
 			t[4] = -1
 			t[5] = -1
+			t[6] = -1
```
joe-elliott, line 155: i was noodling this and removing all the `t[N] = -1` from this function passes `TestSearchCompleteBlock`. This is b/c the row number is only `Next()'ed` in the iterators and it always receives the same definition level. So we are repeatedly (millions of times) setting those array positions to -1 for no reason. 

we could honestly change row number to be a slice and only have entries to the left of the required definition level.

pkg/parquetquery/iters.go
```diff
@@ -1940,6 +2203,11 @@ func (a *KeyValueGroupPredicate) KeepGroup(group *IteratorResult) bool {
 	return true
 }
 
+func panicWhenInvalidDefinitionLevel(definitionLevel int) {
+	level.Error(log.Logger).Log("msg", "definition level out of bound: should be [0:7] but got %d", definitionLevel)
```
joe-elliott, line 2207: same logging thoughts here.

tempodb/encoding/vparquet4/block_traceql.go
```diff
@@ -1364,27 +1369,38 @@ func (i *mergeSpansetIterator) Close() {
...
-//                                                            |
-//  Span column iterator 1    ---------------------------     |
-//                      ...   ------------------------  |     |
-//  Span column iterator N    ---------------------  |  |     |
-//    (ex: name, status)                          |  |  |     |
-//                                                V  V  V     V
-//                                            ------------------
-//                                            | span collector |
-//                                            ------------------
+//  Event attribute iterator: key    --------------------------------------------
```
joe-elliott, line 1372: lol

i love so much that you updated this diagram. there's no way i would have done it.

🙇

joe-elliott (commented):
fantastic work here. i think it's close, but i want to make an honest effort at this perf regression before merging.

I would like to see tests added to ./tempodb/tempodb_search_test.go:TestSearchCompleteBlock

And I would expect local manual testing using the local docker-compose to confirm things work as expected.

pkg/parquetquery/iters.go
ie-pham: will follow up in another discovery/PR

pkg/parquetquery/iters.go
ie-pham: this requires a return unless there's a better way to reuse the function?

pkg/parquetquery/iters.go
ie-pham: i also thought about this, will follow up

pkg/parquetquery/iters.go
joe-elliott: +1

joe-elliott (approved):
o7

PR #3751, opened 2024-06-06
search/utils: added missing new intrinsic names to search/tags endpoint

### LINKED ISSUE #3714: Tags Endpoint: Incorrect intrinsics returned
**Describe the bug**
When querying the `/api/v2/tags` endpoint Tempo still returns the old intrinsic names.  In https://github.com/grafana/tempo/pull/3629 we added support for the new intrinsic names but forgot to swap over this endpoint:

> curl -H 'X-Scope-OrgID: 1' 'http://localhost:3100/tempo/api/v2/search/tags?scope=intrinsic'
{"scopes":[{"name":"intrinsic","tags":["duration","kind","name","rootName","rootServiceName","status","statusMessage","traceDuration"]}]}

Should return names like:

`span:duration`, `span:kind`.

While we're at it, let's add these two new intrinsics as well:

https://github.com/grafana/tempo/pull/3670

We should consider returning both the legacy and new values since we intend to support them both for the foreseeable future.

**Code changes**

This is the code that returns intrinsic tags:
https://github.com/grafana/tempo/blob/main/pkg/search/util.go#L57

Also confirm that this method works for the new intrinsic names:
https://github.com/grafana/tempo/blob/main/pkg/search/util.go#L31

ishanjain28:
**What this PR does**:

This PR adds new intrinsic names to the response of `/v2/search/tags` endpoint.

The response now looks like, 
```
{"scopes":[{"name":"intrinsic","tags":["duration","kind","name","rootName","rootServiceName","span:duration","span:kind","span:name","span:status","span:statusMessage","status","statusMessage","trace:rootName","trace:rootServiceName","traceDuration"]}]}
```

It is missing `trace:traceDuration`. I wasn't sure what to call it in the code ?

Fixes #3714

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/enum_attributes.go
+++ b/pkg/traceql/enum_attributes.go
@@ -89,13 +89,13 @@ const (
 
 	IntrinsicTraceID
 	IntrinsicSpanID
-	IntrinsicSpanStatus
-	IntrinsicSpanStatusMessage
-	IntrinsicSpanDuration
-	IntrinsicSpanName
-	IntrinsicSpanKind
-	IntrinsicTraceRootName
-	IntrinsicTraceRootServiceName
+	ScopedIntrinsicSpanStatus
+	ScopedIntrinsicSpanStatusMessage
+	ScopedIntrinsicSpanDuration
+	ScopedIntrinsicSpanName
+	ScopedIntrinsicSpanKind
+	ScopedIntrinsicTraceRootName
+	ScopedIntrinsicTraceRootServiceName
 
 	// not yet implemented in traceql and may never be. these exist so that we can retrieve
 	// these fields from the fetch layer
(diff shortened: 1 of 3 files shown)


ishanjain28:
Hi  @ie-pham, Requesting a review

pkg/traceql/enum_attributes.go
```diff
@@ -89,6 +89,13 @@ const (
 
 	IntrinsicTraceID
 	IntrinsicSpanID
+	IntrinsicSpanStatus
```
ie-pham, line 92: To differentiate perhaps we can do "ScopedIntrinsic-" instead of continuing the "Intrinsic" pattern? What do you think?

ishanjain28: That sounds good to me, I'll make the change!

ishanjain28: Made the changes, Requesting re-review.

ie-pham:
Thank you for changing the prefix. Can you add `ScopedIntrinsicTraceDuration` now that it's not conflicting anymore?

ishanjain28:
Added `trace:traceDuration`.

ie-pham (approved):
Awesome work! Thank you for contributing

PR #3754, opened 2024-06-06
Search tags refactor

joe-elliott:
**What this PR does**:
Refactors tags search to better support tag name filtering and for performance.

- Moved all "distinct collectors" to a new package `./pkg/collector` and added `ScopedDistinctString` collector for use with tags v2.
- Fixed a bug where tag names limits were only applied per scope. Now they are applied across all scopes.
- Made SearchTags dependent on SearchTagsV2 (instead of the reverse). Previously SearchTagsV2 called SearchTags once for each scope which is much less efficient than calling it once for all scopes. Now SearchTagsV2 asks the block directly for the desired scopes.
- Added an explicit `TagsCallback` that returns both scope and tag name. Previously the block only returned tag name. Calling code was required to call the block once for each scope to know how to build the response. This is the required change that drove the refactor.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/integration/e2e/multi_tenant_test.go
+++ b/integration/e2e/multi_tenant_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/grafana/dskit/user"
 	"github.com/grafana/e2e"
+	"github.com/grafana/tempo/pkg/collector"
 	"github.com/grafana/tempo/pkg/httpclient"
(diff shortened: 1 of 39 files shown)


mapno (approved):
LGTM

PR #3760, opened 2024-06-07
Improve http connection reuse

### LINKED ISSUE #3750: S3 Not Reusing Connections
**Describe the bug**
S3 is not reusing connections when ranged requests are made.

**To Reproduce**
1. Start Tempo (version 2.5) pointing to minio running locally over http (I've included the docker-compose.yml and tempo.yml I used)
2. Insert a lot of traces
3. Run Wireshark on localhost
4. Try to retrieve a trace by id: ie http://localhost:3200/api/traces/<trace_id>
5. Watch as tempo makes a large number (159 for me) requests to minio and reuses the connection only once.

**Expected behavior**
When making requests to s3 (or Minio) the connections should be reused where possible.

**Environment:**
 - Infrastructure: docker
 
**Additional Context**

The bug is at https://github.com/grafana/tempo/blob/a6a44d3d9a0b7ac1fea2063edd8ad1f8f20f98f2/tempodb/backend/s3/s3.go#L552

```
err := options.SetRange(offset, offset+int64(len(buffer)))
...
byteCount, err := reader.Read(buffer[totalBytes:])
```

`reader.Read()` is called with a buffer which is EXACTLY the number of bytes requested in the range. this means an EOF will not be usually read. for example if I request 5 bytes, and read 5 bytes then an EOF should not been. because an EOF is not seen then the connection cannot be reused according to the the golang docs:

> It is the caller's responsibility to close Body. The default HTTP client's Transport may not reuse HTTP/1.x "keep-alive" TCP connections if the Body is not read to completion and closed.

as a bonus issue the code checks for `EOF` and silently discards the `EOF` error returning a `nil` error. the return value from this readRange function is just `error` so its impossible for a caller to know whether all of the bytes requested were actually returned. for example its possible to supply a buffer of size 10 to readRange and only 5 bytes are returned from the http request and the remaining 5 bytes will be whatever was originally in the buffer.

with the following change the number of connections drops from 159 to 78. The async pages code in parquet makes a lot of parallel requests for performance reasons so it's not possible or desirable to have the same connection reused for all requests.

```
diff --git a/tempodb/backend/s3/s3.go b/tempodb/backend/s3/s3.go
index cdc5369e8..b1f82f168 100644
--- a/tempodb/backend/s3/s3.go
+++ b/tempodb/backend/s3/s3.go
@@ -6,6 +6,7 @@ import (
        "errors"
        "fmt"
        "io"
+       "io/ioutil"
        "net/http"
        "os"
        "path"
@@ -561,19 +562,16 @@ func (rw *readerWriter) readRange(ctx context.Context, objName string, offset in
        }
        defer reader.Close()
 
-       totalBytes := 0
-       for {
-               byteCount, err := reader.Read(buffer[totalBytes:])
-               if errors.Is(err, io.EOF) {
-                       return nil
-               }
-               if err != nil {
-                       return fmt.Errorf("error in range read from s3 backend: %w", err)
-               }
-               if byteCount == 0 {
-                       return nil
-               }
-               totalBytes += byteCount
+       /* totalBytes == len(buffer) if and only if err == nil */
+       _, err = io.ReadFull(reader, buffer)
+
+       if err == nil {
+               /* read EOF so connection can be reused */
+               io.Copy(ioutil.Discard, reader)
+               return nil
+       } else {
+               /* ideally i think we would not want to use connections that are broken, so don't try and read EOF here */
+               return fmt.Errorf("error in range read from s3 backend: %w", err)
        }
 }
```

docker-compose.yml
```
services:

  tempo:
    image: grafana/tempo:latest
    command: "-target=scalable-single-binary -config.file=/etc/tempo.yaml"
    volumes:
      - ./tempo.yaml:/etc/tempo.yaml
    restart: always
    networks:
      default:
        aliases:
          - tempo
    depends_on:
      - minio
    ports:
      - 3200:3200


  minio:
    image: minio/minio:latest
    entrypoint: minio server /data --console-address ':9001'
    environment:
      - MINIO_ROOT_USER=tempo
      - MINIO_ROOT_PASSWORD=supersecret
    volumes:
      - ./data:/data
    ports:
      - 9000:9000
      - 9001:9001
 ```

tempo.yaml

```
server:
  http_listen_port: 3200

distributor:
  receivers:                           # this configuration will listen on all ports and protocols that tempo is capable of.
    otlp:
      protocols:
        http:
        grpc:

storage:
  trace:
    backend: s3                        # backend configuration to use
    wal:
      path: /var/tempo/wal             # where to store the wal locally
    s3:
      bucket: tempo                    # how to store data in s3
      endpoint: host.docker.internal:9000
      access_key: tempo
      secret_key: supersecret
      insecure: true

querier:
  frontend_worker:
    frontend_address: tempo:9095
```

### PR DESCRIPTION
**Pull request by bmteller**
**What this PR does**: 
Fixes connection reuse for S3 range requests and increases the number of MaxIdleConnsPerHost for the S3 backend to 100 to match the azure settings. Also, assumes GCP has the same connection reuse problem for range requests and performs the same fix. Note: I have tested the fix for S3 range requests and verified it is a problem and the change fixes the problem but I have not verified that it is a problem for GCP or that it fixes the problem for GCP. The GCP HTTP google storage code does look like it effectively returns the body reader so it does look like GCP would have a similar issue.

Fixes #3750

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * [ENHANCEMENT] Use multiple goroutines to unmarshal responses in parallel in the query frontend. [#3713](https://github.com/grafana/tempo/pull/3713) (@joe-elliott)
 * [BUGFIX] Fix metrics queries when grouping by attributes that may not exist [#3734](https://github.com/grafana/tempo/pull/3734) (@mdisibio)
 * [BUGFIX] max_global_traces_per_user: take into account ingestion.tenant_shard_size when converting to local limit [#3618](https://github.com/grafana/tempo/pull/3618) (@kvrhdn)
+* [BUGFIX] Fix http connection reuse on GCP and AWS by reading io.EOF through the http body. [#3760](https://github.com/grafana/tempo/pull/3760) (@bmteller)
 
 ## v2.5.0
 
--- a/tempodb/backend/gcs/gcs.go
+++ b/tempodb/backend/gcs/gcs.go
@@ -477,20 +477,17 @@ func (rw *readerWriter) readRange(ctx context.Context, name string, offset int64
 	}
 	defer r.Close()
 
-	totalBytes := 0
-	for {
-		byteCount, err := r.Read(buffer[totalBytes:])
-		if errors.Is(err, io.EOF) {
-			return nil
-		}
-		if err != nil {
-			return err
-		}
-		if byteCount == 0 {
-			return nil
-		}
-		totalBytes += byteCount
+	/* bytes read == len(buffer) if and only if err == nil */
+	_, err = io.ReadFull(r, buffer)
+
+	if err == nil {
+		/* read EOF so connection can be reused */
+		var dummy [1]byte
+		_, _ = r.Read(dummy[:])
+		return nil
 	}
+
+	return err
 }
 
 func createBucket(ctx context.Context, cfg *Config, hedge bool) (*storage.BucketHandle, error) {
--- a/tempodb/backend/s3/s3.go
+++ b/tempodb/backend/s3/s3.go
@@ -561,20 +561,17 @@ func (rw *readerWriter) readRange(ctx context.Context, objName string, offset in
 	}
 	defer reader.Close()
 
-	totalBytes := 0
-	for {
-		byteCount, err := reader.Read(buffer[totalBytes:])
-		if errors.Is(err, io.EOF) {
-			return nil
-		}
-		if err != nil {
-			return fmt.Errorf("error in range read from s3 backend: %w", err)
-		}
-		if byteCount == 0 {
-			return nil
-		}
-		totalBytes += byteCount
+	/* bytes read == len(buffer) if and only if err == nil */
+	_, err = io.ReadFull(reader, buffer)
+
+	if err == nil {
+		/* read EOF so connection can be reused */
+		var dummy [1]byte
+		_, _ = reader.Read(dummy[:])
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Review by joe-elliott — COMMENTED** [review]
Thanks for working on this.

- Do you think we should put something similar in Azure?
- Can you add a changelog entry detailing the fix

**Comment by bmteller** [comment]
I don't think azure has a similar problem unless it is in the azure library itself.

```
        if _, err := blobClient.DownloadBuffer(ctx, destBuffer, &blob.DownloadBufferOptions{
                Range: blob.HTTPRange{
                        Offset: offset,
                        Count:  size,
                },
                BlockSize:   blob.DefaultDownloadBlockSize,
                Concurrency: maxParallelism,
                RetryReaderOptionsPerBlock: blob.RetryReaderOptions{
                        MaxRetries: maxRetries,
                },

``` 

and this blobClient is from `github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob`.

bmteller:
what should it be classified in the changelog as? BUGFIX?

joe-elliott:
> what should it be classified in the changelog as? BUGFIX?

Yeah, I think that's fair.

joe-elliott (approved):
I went ahead and made the changelog entry b/c of how valuable this fix is.

Thanks!

bmteller:
thanks Joe!

PR #3780, opened 2024-06-17
Create a configurable list of TraceQL queries that are immediately 400'ed

### LINKED ISSUE #3769: Create a configurable list of TraceQL queries that are immediately 400'ed
Create a configurable list of regexes that are applied to every TraceQL query received by Tempo. If one of the regexes matches the query should be immediately 400ed instead of executed. This is to help in situations where a tenant is executing a query that is negatively impacting Tempo (e.g. OOMing or panicing queriers). 

The config can be added here:

https://github.com/grafana/tempo/blob/2870e82df7045b41c166da4a624ae8d68c0e6394/modules/frontend/config.go#L34

Ideally we would create a new [middleware](https://github.com/grafana/tempo/blob/main/modules/frontend/pipeline/pipeline.go#L29) and add it here:

https://github.com/grafana/tempo/blob/2870e82df7045b41c166da4a624ae8d68c0e6394/modules/frontend/frontend.go#L113

to reject the queries.

maliciousbucket:
**What this PR does**:
Adds a new filter middleware to the search and metrics endpoints that matches the search query  against a set of regex expressions. If it matches, a 400 result and generic message are returned.

Regex patterns can be added tot he config under FilterRegexes.

- I did try to write unit tests for this based off existing tests, but I am unsure how to set them up and test this feature properly.
  - It was manually tested by hardcoding values into the filter and running the tests for the search handlers
- The implementation of the matchng will probably need to be reworked

I would appreciate some guidance with regards to refactoring it and writign proper unit tests.

Fixes #3769

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/frontend/pipeline/trace_query_filter_test.go
```diff
@@ -0,0 +1 @@
+package pipeline
```
joe-elliott, line 1: i would look at this test:

https://github.com/grafana/tempo/blob/c300244823ca1b068535ce3207aeb7edff1b52c7/modules/frontend/pipeline/async_handler_multitenant_test.go#L112

for an example on how to create and exercise the middleware

maliciousbucket:
Thanks for the feedback! I've done some refactoring and added some basic unit tests. Fairly sure I'm building the URL's wrong for the tests ( hence the use of the raw url in the ParseSearchRequetQuery function). What's the best way to make the traceql queries?

modules/frontend/pipeline/trace_query_filter.go
```diff
@@ -20,47 +22,35 @@ func NewTraceQueryFilterWare(next http.RoundTripper) http.RoundTripper {
...
-	}
-	//Better way to do this?
-	if len(c.filters) == 0 {
-		return resp, nil
-	}
-	//need wait group
-	u, err := api.ParseSearchRequest(req)
-	if err != nil {
-		return resp, err
+	if c.filters == nil || len(c.filters) == 0 {
```
joe-elliott, line 44: in go `len(<nil slice>) == 0`

```suggestion
	if len(c.filters) == 0 {
```

modules/frontend/pipeline/trace_query_filter.go
```diff
@@ -92,5 +82,24 @@ func (c traceQueryFilterWare) RoundTrip(req *http.Request) (*http.Response, erro
...
 	}
-	return resp, nil
+	return c.next.RoundTrip(req)
+}
+
+func ParseSearchRequestQuery(req *http.Request) string {
+
+	//query, _ := api.ParseSearchTagValuesRequestV2(req)
+
+	//query, _ := api.ParseSearchRequest(req)
```
joe-elliott, line 92: it should be this one.

modules/frontend/pipeline/trace_query_filter.go
```diff
@@ -0,0 +1,105 @@
...
+
+	query := c.parseFn(req)
+
+	if len(query) == 0 || query == "" {
+		return c.next.RoundTrip(req)
+	}
+
+	match := make(chan bool, len(c.filters))
+	wg := sync.WaitGroup{}
+	for range c.filters {
```
joe-elliott, line 56: no need for concurrency. let's keep this simple and just check them one at a time inline

modules/frontend/frontend.go
```diff
@@ -97,7 +99,7 @@ func New(cfg Config, next http.RoundTripper, o overrides.Interface, reader tempo
 			multiTenantMiddleware(cfg, logger),
 			newAsyncSearchSharder(reader, o, cfg.Search.Sharder, logger),
 		},
-		[]pipeline.Middleware{cacheWare, statusCodeWare, retryWare},
+		[]pipeline.Middleware{cacheWare, statusCodeWare, retryWare, searchQueryFilterWare},
```
joe-elliott, line 102: searchQueryFilterWare should be the first in the list of the AsyncMiddleware. this will allow it to refuse a request before we do any work. it does mean it needs to be shaped like:

https://github.com/grafana/tempo/blob/main/modules/frontend/search_sharder.go#L51

joe-elliott:
We really appreciate your efforts here, but we have a sudden need for this internally and need to get this feature in. Thank you so much for giving this a shot!

Closing in favor of https://github.com/grafana/tempo/pull/3963

PR #3789, opened 2024-06-18
Begin implementation for native histograms

zalegrala:
**What this PR does**:

Here we add what is necessary to generate and remote-write native histograms from the metrics-generator.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/generator/instance_test.go
```diff
@@ -397,3 +397,7 @@ func (n noopAppender) Rollback() error { return nil }
 func (n noopAppender) UpdateMetadata(prometheus_storage.SeriesRef, labels.Labels, metadata.Metadata) (prometheus_storage.SeriesRef, error) {
 	return 0, nil
 }
+
+func (n noopAppender) AppendCTZeroSample(_ prometheus_storage.SeriesRef, _ labels.Labels, _, _ int64) (prometheus_storage.SeriesRef, error) {
```
joe-elliott, line 401: is this a leftover from a previous change to the appender interface?

modules/generator/registry/interface.go
```diff
@@ -17,6 +17,8 @@ type Counter interface {
 // Histogram
 // https://prometheus.io/docs/concepts/metric_types/#histogram
 type Histogram interface {
+	metric
```
joe-elliott, line 20: it seems like counter and gauge already support this interface as well. should we add to all 3 for clarity?

modules/generator/registry/native_histogram.go
```diff
@@ -0,0 +1,336 @@
...
+	buckets []float64
+
+	traceIDLabelName string
+}
+
+type nativeHistogramSeries struct {
+	// labels should not be modified after creation
+	labels        LabelPair
+	promHistogram prometheus.Histogram
+	lastUpdated   *atomic.Int64
```
joe-elliott, line 43: does this need to be atomic? it seems like its always protected by the seriesMtx above

modules/generator/registry/native_histogram.go
```diff
@@ -0,0 +1,336 @@
...
+		if s.lastUpdated.Load() < staleTimeMs {
+			delete(h.series, hash)
+			h.onRemoveSerie(h.activeSeriesPerHistogramSerie())
+		}
+	}
+}
+
+func (h *nativeHistogram) activeSeriesPerHistogramSerie() uint32 {
+	// TODO can we estimate this?
+	return 1
```
joe-elliott, line 309: this is interesting. i suppose this plays into how many series prom/mimir consider a native histogram?

modules/generator/registry/native_histogram.go
```diff
@@ -0,0 +1,336 @@
...
+func (h *nativeHistogram) name() string {
+	return h.metricName
+}
+
+func (h *nativeHistogram) collectMetrics(appender storage.Appender, timeMs int64, externalLabels map[string]string) (activeSeries int, err error) {
+	h.seriesMtx.Lock()
+	defer h.seriesMtx.Unlock()
+
+	labelsCount := 0
+	if h.series[0] != nil {
```
joe-elliott, line 136: `h.series` is a map of hashes to histograms. won't this basically always return nil?

modules/generator/registry/native_histogram.go
```diff
@@ -0,0 +1,336 @@
...
+
+		// Set series-specific labels
+		for i, name := range s.labels.names {
+			lb.Set(name, s.labels.values[i])
+		}
+
+		// Extract histogram
+		encodedMetric := &dto.Metric{}
+
+		// Encode to protobuf representation
```
joe-elliott, line 160: i don't think this encodes to protobuf (which is good. that would be costly). it just does a fairly complicated transition from one representation to another

modules/generator/registry/native_histogram.go
```diff
@@ -0,0 +1,336 @@
...
+		err = s.promHistogram.Write(encodedMetric)
+		if err != nil {
+			return activeSeries, err
+		}
+		s.histogram = encodedMetric.GetHistogram()
```
joe-elliott, line 165: is there a reason to store this as a field on the struct? it seems like we only use it in this method. if that's true we should remove it from the struct

modules/generator/registry/native_histogram.go
```diff
@@ -0,0 +1,336 @@
...
+
+		// *** Classic histogram
```
joe-elliott, line 167: it looks like we are adding both a classic and native histogram to the appender? any particular reason?

modules/generator/registry/registry.go
```diff
@@ -147,8 +147,17 @@ func (r *ManagedRegistry) NewCounter(name string) Counter {
...
-func (r *ManagedRegistry) NewHistogram(name string, buckets []float64) Histogram {
-	h := newHistogram(name, buckets, r.onAddMetricSeries, r.onRemoveMetricSeries, r.overrides.MetricsGenerationTraceIDLabelName(r.tenant))
+func (r *ManagedRegistry) NewHistogram(name string, buckets []float64) (h Histogram) {
+	traceIDLabelName := r.overrides.MetricsGenerationTraceIDLabelName(r.tenant)
+
+	// Temporary switch: use the old implementation when native histograms are disabled, eventually the new implementation can handle all cases
+	if !r.overrides.MetricsGeneratorGenerateNativeHistograms(r.tenant) {
+		h = newHistogram(name, buckets, r.onAddMetricSeries, r.onRemoveMetricSeries, traceIDLabelName)
+	} else {
+		level.Warn(r.logger).Log("msg", "creating native histogram!", "metric", name)
```
joe-elliott, line 157: should we warn on this? i'd think a debug log if anything

this will create quite a lot of log spam in generators with a lot of tenants creating native histos

joe-elliott (commented):
looks good. some comments to work through

modules/generator/registry/interface.go
zalegrala: Good catch.  Probably yes, which I think prometheus does this already in a similar way.

modules/generator/registry/native_histogram.go
zalegrala: Its doing something protobuf.  Perhaps not a `Marshall` call, but   dto.Metric is a protobuf struct.  [Here](https://github.com/zalegrala/tempo/blob/3ff3eb85abb298739b44b42ab444632ab815b996/vendor/github.com/prometheus/client_golang/prometheus/histogram.go#L751) is the `Write` method.

modules/generator/registry/native_histogram.go
zalegrala: This was added so that we could remember which histograms were sent.  The `Reset` call below captures this, and storing the state on the struct was a way for us to get around this issue.  If we don't know which buckets have had their exemplars reported, they get updated with the most recent timestamp and sent again, since the series map contains keeps the exemplars.

modules/generator/registry/native_histogram.go
zalegrala: Correct.  If we don't then there isn't an opportunity to slow transition.  Without the classic histogram here, we would not be generating them at all, which would break a bunch of things.  The idea here is that we keep writing both, but likely delete the older implementation of classic histograms at a future date.

modules/generator/instance_test.go
zalegrala: This was a required change to satisfy the interface updates for the go mod update.

modules/generator/registry/native_histogram.go
zalegrala: It appears that we are always under lock as you mention.  Do you suggest that `int64` might be a better type here?

joe-elliott: yup, if it's already protected by the global lock, would prefer an `int64`

modules/generator/registry/native_histogram.go
joe-elliott: yeah, i'm not seeing this. it seems only referenced in this method which makes me think we should just use a local var. perhaps i'm missing something. we can sync on this.

modules/generator/registry/native_histogram.go
joe-elliott: we should talk about how a transition would work. i don't like writing a classic histo here b/c it's a bit misleading in code and the costs of doubling all histograms would be quite a lot, but i see we might need some pattern to help people migrate from one to another. 

Maybe this override is tristate instead of bool? classic/native/both?
```
MetricsGeneratorGenerateNativeHistograms
```

modules/generator/registry/native_histogram.go
zalegrala: Hmm, I'm not sure when this would be non-nil, but it doesn't seem that we test for this condition.

modules/generator/registry/native_histogram.go
zalegrala: Right, I think we have something to learn here, since there doesn't seem to be a clean mapping between native histograms and active series.

modules/generator/registry/native_histogram.go
zalegrala: Yeah, let's sync.

modules/generator/registry/native_histogram.go
zalegrala: I like the tristate idea.  I agree it makes it a little messy in code to have them somewhat lumped together.  At least if we had the tristate the classic histograms would be in its own block/function and clearly separated.

modules/generator/registry/native_histogram.go
kvrhdn: A historical note since this code is copied from the original histogram implementation: we used to protect the series map with a `RWMutex`, this way we could grab the serie and update the atomic fields individually without holding a lock. After finding some concurrency issues updating the counters while reading them we opted to lock the entire map with a normal mutex.

Since `prometheus.Histogram` is concurrency-safe we could opt to switch back to a `RWMutex`. But I think this area might still change a lot has we integrate the prometheus code better, so it's fine to keep as-is for now imo.

modules/generator/registry/native_histogram.go
kvrhdn: Oof, yeah that's true. This is also a bug that exists in the original histogram code then 😬 The intention is "grab _any_ series in our map and check how many labels they have". But using `[0]` obviously isn't correct.

That this doesn't work isn't a huge deal, it just means the slice we allocate will be too small and gets reallocated as we add more labels (at most once per collection).

modules/generator/registry/native_histogram.go
kvrhdn: I also like the tristate idea.  Currently whether you create classic histograms or not is controlled by setting the buckets on each processor, if you set them to `nil` we won't create the classic histograms. Having a setting to enable/disable it for the entire tenant in the registry would be nice :+1:

modules/generator/registry/native_histogram.go
zalegrala: I checked the hash function, and there is a 0 value returned, but it doesn't seem useful here to check since the label names would also be empty and zero length.  I'll drop this shortly.

modules/generator/registry/native_histogram.go
zalegrala: I've pushed a commit for this.

example/docker-compose/local/docker-compose.yaml
```diff
@@ -42,6 +42,7 @@ services:
       - --config.file=/etc/prometheus.yaml
       - --web.enable-remote-write-receiver
       - --enable-feature=exemplar-storage
+      - --enable-feature=native-histograms
```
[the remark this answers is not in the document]
kvrhdn: For consistency, I've set this on all docker-compose deployments now. Even if the example doesn't use it, it does no harm enabling the feature.

modules/generator/registry/native_histogram.go
```diff
@@ -0,0 +1,379 @@
...
+
+		// NOTE: A subtle point here is that the new histogram implementation is
+		// chosen in the registry.  This means that if the user overrides change
+		// from "both" to "classic", we will still be using the new histograms
+		// implementation. This is because the registry is not recreated when the
+		// overrides change. This is a tradeoff to avoid recreating the registry
+		// for every change in the overrides, but it does mean that even if user
+		// changes to "classic" histograms, we will still be using the new
+		// histograms implementation and want to avoid generating native
+		// histograms.
```
kvrhdn, line 187: This will work when we switch from `native` or `both` to `classic`, but it will not work when we go from `classic` to anything else. If we initiate their tenant with `classic` we use the old histogram implementation, when we change the override nothing will be updated and we keep using this implementation.
To workaround this we need to recreate the processors I think so they all re-register their histograms.

But maybe something quick to fix in a follow-up PR?

joe-elliott (approved):
lgtm

modules/generator/registry/native_histogram.go
zalegrala: Right, I think let's follow up.

PR #3793, opened 2024-06-19
Update to Go 1.22.4

mapno:
**What this PR does**:

Updates Go's version to 1.22.4.

Fixes two high-severity CVEs: https://github.com/advisories/GHSA-vvjp-q62m-2vph, https://github.com/advisories/GHSA-4v7x-pqxf-cx7m.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott (approved):
would prefer to find ways to float the patch version so we are always building with the latest, but that may not be possible for all the various images/GHAs/etc.

PR #3796, opened 2024-06-20
Add caching to QueryRange requests

mapno:
**What this PR does**:

Adds caching to query range requests

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


mdisibio (approved):

modules/frontend/metrics_query_range_sharder.go
```diff
@@ -393,6 +396,10 @@ func (s *queryRangeSharder) buildBackendRequests(ctx context.Context, tenantID s
 
 			prepareRequestForQueriers(subR, tenantID, subR.URL.Path, subR.URL.Query())
 			// TODO: Handle sampling rate
+			key := queryRangeCacheKey(tenantID, queryHash, int64(queryRangeReq.Start), int64(queryRangeReq.End), m, int(queryRangeReq.StartPage), int(queryRangeReq.PagesToSearch))
```
joe-elliott, line 399: seeing a lot of other fields in the `QueryRangeRequest` struct. Should `Step` be in the cache key? `ShardID/Count`?

mapno: `Step` is part of the hash. `ShardID` and `ShardCount` are used in non-RF1 queries, not sure if we want to implement it for those.

PR #3797, opened 2024-06-20
Drop unnecessary concurrency check in metrics-generator histogram

kvrhdn:
**What this PR does**:
Simplify the `histogram.ObserveWithExemplar` a tad. Since we lock a `Mutex` for the entire method, we don't have to check whether the map was updated in the meantime. This code was added when we still used a `RWMutex` which was removed by #3412.

~~Fixes #<issue number>~~

**Checklist**
- [ ] ~~Tests updated~~
- [ ] ~~Documentation added~~


kvrhdn:
Simplification spotted by @zalegrala 🙏🏻

zalegrala (approved):

PR #3799, opened 2024-06-20
[TraceQL Metrics] Add local disk caching for generator completed blocks

mdisibio:
**What this PR does**:
Generators service metrics queries against recent data, which is done frequently and sensitive to user-facing performance. This adds a cache of query responses for local completed blocks which is typically the last 5-20 minutes.

The generators already have a similar cache for the metrics summary api, but this takes a different approach. Whereas the summary api cache is in-memory, this is disk-based and writes new caching files to the block folder. These files are unique to each request (query + params), and exist for the lifetime of the block.  They aren't flushed to object storage, and they get automatically cleaned up when the block is deleted.  (Note - this has precedent in the `flushed` file that is written to block folders to track their flushed status) Reasons to avoid the in-memory cache is that generators are already usually memory-intensive, and the cache is often inadequate as-is. Example block contents:

```
pwd = /tempo/generator/traces/single-tenant/blocks/single-tenant/
  65beeff3-d9c5-49c1-b69e-8c7b022024f3/bloom-0
  65beeff3-d9c5-49c1-b69e-8c7b022024f3/cache_query_range_14197044629227963207.buf
  65beeff3-d9c5-49c1-b69e-8c7b022024f3/cache_query_range_18089096181261108941.buf
  65beeff3-d9c5-49c1-b69e-8c7b022024f3/cache_query_range_433499706678913880.buf
  65beeff3-d9c5-49c1-b69e-8c7b022024f3/cache_query_range_4989462974766826143.buf
  65beeff3-d9c5-49c1-b69e-8c7b022024f3/data.parquet
  65beeff3-d9c5-49c1-b69e-8c7b022024f3/flushed
  65beeff3-d9c5-49c1-b69e-8c7b022024f3/index
  65beeff3-d9c5-49c1-b69e-8c7b022024f3/meta.json
```

There are additional changes in this PR which proved necessary:

* Generator RF1 block meta times weren't right. They always flushed traces with the timestamp of "now" so the metas were off by roughly trace_idle_period + trace_flush_period. This fixes it to flush proper times.
* As part of that, it also requires that the generator traces WAL has a real ingestion_slack time. This fixes the default to match ingesters of 2 minutes. NOTE - This is different than the slack time for metrics.

Fixes #<issue number>

**Checklist**
- [x ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/processor/localblocks/query_range.go
+++ b/modules/generator/processor/localblocks/query_range.go
@@ -170,23 +170,15 @@ func (p *Processor) queryRangeCompleteBlock(ctx context.Context, b *ingester.Loc
 	})
 	defer span.Finish()
 
-	fmt.Println("Before:", m.BlockID, time.Unix(0, int64(req.Start)), time.Unix(0, int64(req.End)))
-	fmt.Println("Block:", m.BlockID, m.StartTime, m.EndTime)
-
 	// Trim and align the request for this block. I.e. if the request is "Last Hour" we don't want to
 	// cache the response for that, we want only the few minutes time range for this block. This has
 	// size savings but the main thing is that the response is reuseable for any overlapping query.
-	// TODO - This buffer time is needed to close gaps between block meta time and actual
-	// span contents
-	blockStart := m.StartTime.Add(-time.Minute)
-	blockEnd := m.EndTime.Add(time.Minute)
+	req.Start, req.End = traceql.TrimToOverlap(req.Start, req.End, req.Step, uint64(m.StartTime.UnixNano()), uint64(m.EndTime.UnixNano()))
 
-	req.Start = max(uint64(blockStart.UnixNano()), req.Start)
-	req.End = min(uint64(blockEnd.UnixNano()), req.End)
-	req.Start = (req.Start / req.Step) * req.Step
-	req.End = (req.End/req.Step)*req.Step + req.Step
-
-	fmt.Println("After:", m.BlockID, time.Unix(0, int64(req.Start)), time.Unix(0, int64(req.End)))
+	if req.Start >= req.End {
+		// After alignment there is no overlap or something else isn't right
+		return nil, nil
+	}
 
 	cached, name, err := p.queryRangeCacheGet(ctx, m, req)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


modules/generator/processor/localblocks/query_range.go
```diff
@@ -0,0 +1,280 @@
...
+	return p.walW.Write(ctx, name, m.BlockID, m.TenantID, data, nil)
+}
+
+func queryRangeHashForBlock(req tempopb.QueryRangeRequest) uint64 {
+	h := fnv.New64a()
+	buf := make([]byte, 8)
+
+	h.Write([]byte(req.Query))
+
+	binary.BigEndian.PutUint64(buf, req.Step)
```
mapno, line 264: Why use this and not the native `fnv1a.AddUint64`? `binary.BigEndian.PutUint64` is only putting the int into the hash. Unsure if there's a big difference.

mdisibio: Good question, was just following practice from other code like trace combiner. Will take a look to see if there is any meaningful difference.

modules/generator/processor/localblocks/query_range.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	return results, nil
+}
+
+func (p *Processor) queryRangeCacheGet(ctx context.Context, m *backend.BlockMeta, req tempopb.QueryRangeRequest) (*tempopb.QueryRangeResponse, string, error) {
+	cacheable := req.Start <= uint64(m.StartTime.UnixNano()) && req.End >= uint64(m.EndTime.UnixNano())
+	if !cacheable {
+		return nil, "", nil
+	}
```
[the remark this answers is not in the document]
mdisibio: Agree, I think the requirement for caching changed during implementation.  Originally it was trying to avoid partial time range caching, but after solidifying the time handling with TrimToOverlap, I missed this. Will take a look.

modules/generator/processor/localblocks/query_range.go
mdisibio: They are identical and agree this new package is more efficient. Updated.

modules/generator/processor/localblocks/query_range.go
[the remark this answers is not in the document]
mdisibio: Cleaned up

mapno (approved):

PR #3805, opened 2024-06-24
Bugfix: Improved handling of complete blocks in localblocks processor after enabling flushing

mapno:
**What this PR does**:

This fix addresses an issue where complete blocks were not being properly processed in the `localblocks` processor. Before, complete blocks were kept for longer than necessary. This was a problem specially when enabling flushing to storage for the first time.

The condition for deleting complete blocks has been changed to be based on block end time. If the block has been flushed and it's past it the cutoff time, it's deleted.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * [BUGFIX] Fix frontend parsing error on cached responses [#3759](https://github.com/grafana/tempo/pull/3759) (@mdisibio)
 * [BUGFIX] max_global_traces_per_user: take into account ingestion.tenant_shard_size when converting to local limit [#3618](https://github.com/grafana/tempo/pull/3618) (@kvrhdn)
 * [BUGFIX] Fix http connection reuse on GCP and AWS by reading io.EOF through the http body. [#3760](https://github.com/grafana/tempo/pull/3760) (@bmteller)
+* [BUGFIX] Improved handling of old blocks in localblocks processor after enabling flusing [#3805](https://github.com/grafana/tempo/pull/3805) (@mapno)
 
 ## v2.5.0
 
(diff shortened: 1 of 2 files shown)


modules/generator/processor/localblocks/processor.go
```diff
@@ -547,10 +547,20 @@ func (p *Processor) deleteOldBlocks() (err error) {
 
 		flushedTime := b.FlushedTime()
 		if flushedTime.IsZero() {
+
+			if b.BlockMeta().EndTime.Before(cuttoff) { // Not flushed and old
```
mdisibio, line 551: I'm not sure we want to add this case.  The original design was to try to flush forever like the ingesters, so that it tolerates network outages and doesn't lose data.   

But mainly this fix isn't addressing the problem we saw.   After enabling "flush_to_storage" this reset the timer on all blocks, because line 553 below is checking based on `flushedTime`. Therefore it keeps blocks for up to 2*complete_block_time.

mapno: Changed 👍

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 * [BUGFIX] Fix frontend parsing error on cached responses [#3759](https://github.com/grafana/tempo/pull/3759) (@mdisibio)
 * [BUGFIX] max_global_traces_per_user: take into account ingestion.tenant_shard_size when converting to local limit [#3618](https://github.com/grafana/tempo/pull/3618) (@kvrhdn)
 * [BUGFIX] Fix http connection reuse on GCP and AWS by reading io.EOF through the http body. [#3760](https://github.com/grafana/tempo/pull/3760) (@bmteller)
+* [BUGFIX] Improved handling of old blocks in localblocks processor after enabling flusing [#3805](https://github.com/grafana/tempo/pull/3805) (@mapno)
```
mapno, line 24:
```suggestion
* [BUGFIX] Improved handling of complete blocks in localblocks processor after enabling flusing [#3805](https://github.com/grafana/tempo/pull/3805) (@mapno)
```

mdisibio (approved):

PR #3808, opened 2024-06-24
TraceQL Metrics: Diff streaming

### LINKED ISSUE #3765: Streaming gRPC does not avoid 'message larger than max' error
When querying Tempo with (streaming) gRPC using `tempo-cli --use-grpc` to do a large query, `tempo-cli` fails:

`tempo-cli: error: main.metricsQueryRangeCmd.Run(): rpc error: code = ResourceExhausted desc = grpc: received message larger than max (4560241 vs. 4194304)`

.. and the same when using the same max message size on `tempo-cli` as we have previously configured on the server:

`tempo-cli: error: main.metricsQueryRangeCmd.Run(): rpc error: code = Internal desc = response larger than the max (16842812 vs 16777216)`

**To Reproduce**
Perform a query with tempo-cli that returns a lot of data, specifically I tested with
`tempo-cli query api metrics localhost:3100 '{ selector } | quantile_over_time(duration, .99) by (span.http.url)' starttime endtime --use-grpc`

**Expected behavior**
The streaming gRPC should avoid the max message size by chunking appropriately, allowing to run large queries.

**Environment:**
Tempo 2.5.0 on Kubernetes 1.28

joe-elliott:
**What this PR does**:
Improves the query frontend to only return those streams and segments that have been updated instead of the entirety of the metric. This is my first foray into some of the metrics internals so I made some best guesses about implementation. Please steer me in the right direction if you disagree!

Also fixes a bug in the tempo cli where we were passing unix epoch seconds, but it only takes epoch nanoseconds.

Fixes #3765 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


zalegrala (approved):
Nice.

PR #3812, opened 2024-06-26
Add data quality metric to measure traces without a root

mapno:
**What this PR does**:

Adds new reason `rootless_trace` to data quality metric warnings.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/dataquality/warnings.go
+++ b/pkg/dataquality/warnings.go
@@ -8,6 +8,7 @@ import (
 const (
 	reasonOutsideIngestionSlack = "outside_ingestion_time_slack"
 	reasonDisconnectedTrace     = "disconnected_trace"
+	reasonRootlessTrace         = "rootless_trace"
 
 	PhaseTraceFlushedToWal     = "_flushed_to_wal"
 	PhaseTraceWalToComplete    = "_wal_to_complete"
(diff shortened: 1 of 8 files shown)


pkg/dataquality/warnings.go
```diff
@@ -27,3 +28,7 @@ func WarnOutsideIngestionSlack(tenant string) {
 func WarnDisconnectedTrace(tenant string, phase string) {
 	metric.WithLabelValues(tenant, reasonDisconnectedTrace+phase).Inc()
 }
+
+func WarnRootlessTrace(tenant string, phase string) {
```
joe-elliott, line 32: doesn't `disconnected_trace_compactor_combine` already cover this?

does the generator publish `disconnected_trace_flushed_to_wal` and `_wal_to_complete`. that would tell us a lot about the quality of traces produced by the generator

mapno: Yes, rootless traces would be a subset of disconnected traces. Maybe it makes more sense as an extension of that warning?

joe-elliott: I'm good on this PR and will approve, but I am struggling to think what it would mean to a Tempo operator to have the majority of their disconnected traces rootless or rooted. Would they investigate data quality issues differently?

That doesn't mean it doesn't have a use and we might be quite happy we added this at some point in the future.

joe-elliott (approved):

pkg/dataquality/warnings.go
mapno: I see it as a particularly bad type of disconnected trace—usually `higher in the trace == more important`. I think it just adds more information to the operator to investigate.

PR #3829, opened 2024-07-02
Added tempo vulture example

javiermolinar:
**What this PR does**:
It adds a new docker-compose example to run Vulture alongside a local Tempo instance
The hard coded port has been removed from the grpc client instantiation

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


cmd/tempo-vulture/main.go
```diff
@@ -311,7 +311,7 @@ func newJaegerGRPCClient(endpoint string) (*jaeger_grpc.Reporter, error) {
 	}
 
 	// new jaeger grpc exporter
-	conn, err := grpc.Dial(u.Host+":14250", dialOpts...)
```
zalegrala, line 314: We still require the port for the client call, correct?

example/docker-compose/local/readme.md
```diff
@@ -10,7 +10,6 @@ permissions set in order to start correctly.
 
 ```console
 mkdir tempo-data/
-sudo chown 10001:10001 tempo-data/
```
mapno, line 13: Why is this removed? Is it covered by `init` in the compose?

cmd/tempo-vulture/main.go
mapno: I'd try to keep it backwards compatible. This is a change that will break vultures everywhere. I'd add a port config param that defaults to `14250`

javiermolinar: We can do that of course, right now if you pass something like this:
```
--tempo-push-url localhost:14250
```

you will end with this:

```
localhost:14250:14250
```
Even more this line is the source of all that confusion:

https://github.com/grafana/tempo/blob/15ba2e14375b1776d9a78c715fa423d5522c2592/cmd/tempo-vulture/main.go#L66

example/docker-compose/local/readme.md
javiermolinar: If you do that the container won't start due to permissions errors. Just creating the folder and allowing the init to chown the folder is enough

cmd/tempo-vulture/main.go
```diff
@@ -285,15 +294,22 @@ func selectPastTimestamp(start, stop time.Time, interval, retention time.Duratio
...
+func getGRPCEndpoint(endpoint string) (string, error) {
 	u, err := url.Parse(endpoint)
 	if err != nil {
-		return nil, err
+		return "", err
 	}
+	dialAddress := u.Host
 
+	if u.Port() == "" {
+		dialAddress = dialAddress + fmt.Sprintf(":%d", defaultGRPCEndpoint)
```
mapno, line 305:
```suggestion
		dialAddress = fmt.Sprintf("%s:%d", dialAddress, defaultGRPCEndpoint)
```

cmd/tempo-vulture/main.go
```diff
@@ -58,6 +58,10 @@ type traceMetrics struct {
 	notFoundSearchAttribute int
 }
 
+const (
+	defaultGRPCEndpoint = 14250
```
mapno, line 62: Nit: More explicit

```suggestion
	defaultJaegerGRPCEndpoint = 14250
```

mapno (approved):
Just two nits, LGTM

mapno (approved):

PR #3836, opened 2024-07-03
feat: added new helper function to allow debugging e2e tests

javiermolinar:
**What this PR does**:

It can be handy to debug Tempo code when running an e2e test. This can be done in different ways. This PR aims to ease this task by providing a new function to spin a Tempo docker container based on the tempo-debug image, without modifying any aspect of the test so it can be run in the same way either locally or in the CI.

**Limitations**:
I haven't found a way to expose the docker host port locally ie: -p 2345:2345. 
This could be enhanced to allow that:

https://github.com/grafana/e2e/blob/db90b84177fccb4f023ea9f71ec0e8dcd2d3bdad/service.go#L328

**How to use it**:
The debug process is a bit cumbersome right now:
1. In our e2e test use the new function ``NewTempoAllInOnDebug`` to spin the debug container 
2. Set a breakpoint before the action we want to test
3. Start debugging the tests and wait till the breakpoint 
4. Get the docker container endpoint, this can be done by running this:
```
docker ps --format '{{.Ports}}'  
0.0.0.0:53467->2345
```
5. From a different instance of the IDE or directly from the console
```
 dlv connect localhost:53467
```

**Checklist**
- [X ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,12 +21,13 @@
 * [ENHANCEMENT] Added a boolean flag to enable or disable dualstack mode on Storage block config for S3 [#3721](https://github.com/grafana/tempo/pull/3721) (@sid-jar, @mapno)
 * [ENHANCEMENT] Add caching to query range queries [#3796](https://github.com/grafana/tempo/pull/3796) (@mapno)
 * [ENHANCEMENT] Add data quality metric to measure traces without a root [#3812](https://github.com/grafana/tempo/pull/3812) (@mapno)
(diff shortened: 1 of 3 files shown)


mapno (commented):
Would it be possible to include the user guide in the `/integration` readme?

mapno (approved):
LGTM

PR #3840, opened 2024-07-03
[CHANGE] tempo-query: switch from grpcPlugin to standalone

frzifus:
**What this PR does**:
This PR replaces the tempo-query gRPC storage plugin due to the [deprecation in Jaeger 1.58.0](https://github.com/jaegertracing/jaeger/blob/main/CHANGELOG.md#-breaking-changes) with a gRPC standalone service.

Instead of `jaeger+tempo storage plugin`, only the `jaeger remote storage proxy` is now delivered. This can be deployed as a sidecar to jaeger to achieve the same behavior again. 

Tempo-Query will serve on `0.0.0.0:7777` by default.

**Connect Jaeger-Query to Tempo-Query**
```bash
SPAN_STORAGE_TYPE=grpc  jaeger-query --grpc-storage.server=localhost:7777
```

**Checklist**
- [x] Tests updated
- [x] Documentation added

---

## How to run the proxy locally

1. Create a tempo configuration (`tempo.yaml`)
```yaml
server:
  http_listen_port: 3200
  grpc_listen_port: 9096

distributor:
  receivers:
    otlp:
      protocols:
        grpc:
          endpoint: "localhost:4417"
        http:
          endpoint: "localhost:4418"

storage:
  trace:
    backend: local
    wal:
      path: /tmp/tempo/wal
    local:
      path: /tmp/tempo/blocks
```
2. Run the tempo instance
```bash
docker run -v ${PWD}/tempo.yaml:/tempo.yaml:z --network=host --rm -it docker.io/grafana/tempo:latest -config.file=/tempo.yaml
```
3. Create a tempo-query config (`tempo-query.yaml`)
```
address: "0.0.0.0:7777"
backend: "localhost:3200"

tls_enabled: false

tenant_header_key: "X-Tenant-ID"
services_query_duration: "10s"
```
4. Run tempo-query.
```bash
docker run -v ${PWD}/query.yaml:/query.yaml:z  --rm -it --network=host localhost/tempo-query -config /query.yaml
```
5. Run Jaeger and point to tempo-query.
```bash
docker run --rm -it --network=host  jaegertracing/jaeger-query:1.60 --grpc-storage.server="localhost:7777" --span-storage.type=grpc
```


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 * [CHANGE] **BREAKING CHANGE** Remove `autocomplete_filtering_enabled` feature flag [#3729](https://github.com/grafana/tempo/pull/3729) (@mapno)
 * [CHANGE] Bump opentelemetry-collector to 0.102.1 [#3784](https://github.com/grafana/tempo/pull/3784) (@debasishbsws)
+* [CHANGE] **BREAKING CHANGE** Bump tempo-query dependency to Jaeger 1.58.1 and switch from grpcPlugin to standalone server [#3840](https://github.com/grafana/tempo/issues/3840) (@frzifus)
```
pavolloffay, line 5: How is this breaking change surfaced to users? What do they need to do? 

Does the configuration of tempo-query change?

cmd/tempo-query/Dockerfile
```diff
@@ -1,10 +1,16 @@
-FROM jaegertracing/jaeger-query:1.57.0
```
pavolloffay, line 1: Could this be kept and build the image with the jaeger-query?

CHANGELOG.md
frzifus: The configuration should be the same. The main change is that Jaeger is no longer part of the resulting image.

Instead of configuring the plugin with Jaeger a user needs to configure Jaeger itself to use the tempo-query proxy as storage backend.

cmd/tempo-query/Dockerfile
frzifus: And then a users has the option to run jaeger-query or tempo-query?
Y, we can do that. 👍

CHANGELOG.md
knylander-grafana: Are there any caveats that should be added to the documentation?

frzifus: Mainly that the image no longer serves the tempo connection as a grpc-plugin loaded by jaeger. Instead  its a standalone service which by default listens on `0.0.0.0:7777`. 

The configuration should remain the same, aside from the new address field e.g.:
```diff
+ address: "0.0.0.0:7777"
backend: "my-backend-service"

tls_enabled: true
tls:
  cert_path: "/path/to/server.crt"
  key_path: "/path/to/server.key"
  ca_path: "/path/to/ca.crt"
  server_name: "my.server.name"
  insecure_skip_verify: false
  cipher_suites: "TTLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"
  min_version: "TLS1.2"

tenant_header_key: "X-Tenant-ID"
services_query_duration: "10s"
```

frzifus:
I think we need to explicitly forward the tenent header and that commit is missing. I will double check and update this later today or tomorrow morning.

frzifus:
Changed the PR back to draft mode. Ive to adapt `integration/e2e/query_plugin_test.go`.

frzifus:
Seems only an unrelated test failed:
https://github.com/grafana/tempo/actions/runs/10612462691/job/29414164456?pr=3840#step:4:479

frzifus:
> I apologize for letting this sit, but I don't know much about this. Is this removing the Jaeger UI? I believe RedHat/OpenShift rely on this, but I honestly don't know.
> 
> @pavolloffay @andreasgerstmayr if you are good on these changes I will merge. RH is the primary user of tempo-query.

Maybe I should also have provided more context. ^^

Up to and including Jaeger 1.58, tempo-query was a Jaeger instance with an installed custom plugin for connecting to tempo. The plugin option was removed in Jaeger 1.59.

But the remote grpc storage is still supported.
By changing the plugin into a standalone proxy, we can deploy it as a sidecar and bump the jaeger version independently. 

I am currently working on a patch for the tempo operator.

pavolloffay (approved):

joe-elliott (approved):

PR #3846, opened 2024-07-04
[TraceQL] Autocomplete for events and links

ie-pham:
**What this PR does**: 
Add autocomplete support for events and links. Return empty responses for IDs in autocomplete. 
Also returns event:name, link:spanID, and link:traceID to search results to be displayed on the UI

Grafana UI is not updated yet but API works as expected
Screenshot 2024-07-04 at 5 05 02 PM
Screenshot 2024-07-04 at 5 05 22 PM
Screenshot 2024-07-04 at 5 28 04 PM

event:name, link:spanID, and link:traceID
Screenshot 2024-07-01 at 9 37 26 PM
Screenshot 2024-07-01 at 9 38 01 PM

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


ie-pham:
One caveat is because links and events are on the same definition level, they cannot be used to filter each other out effectively. I prefer to get the main functionality out first and create a new ticket to add this ability.

integration/e2e/multi_tenant_test.go
```diff
@@ -148,7 +148,8 @@ func TestMultiTenantSearch(t *testing.T) {
 
 			tagsV2Resp, err := apiClient.SearchTagsV2()
 			require.NoError(t, err)
-			require.Equal(t, 3, len(tagsV2Resp.GetScopes()))
+			fmt.Println(tagsV2Resp.Scopes)
```
stoewer, line 151: Was this print statement for debugging or is it meant to stay?

modules/ingester/instance_search_test.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/grafana/tempo/pkg/model/trace"
 	"github.com/grafana/tempo/pkg/tempopb"
 	v1 "github.com/grafana/tempo/pkg/tempopb/common/v1"
+	v11 "github.com/grafana/tempo/pkg/tempopb/trace/v1"
```
stoewer, line 26: `nitpick` we often use 'trace_v1', 'resource_v1` as import aliases
```suggestion
	trace_v1 "github.com/grafana/tempo/pkg/tempopb/trace/v1"
```

tempodb/encoding/vparquet4/block_autocomplete.go
```diff
@@ -114,6 +128,128 @@ func createDistinctIterator(
...
+	attrIter, err := createDistinctAttributeIterator(makeIter, tag, genericConditions, DefinitionLevelResourceSpansILSSpanEventAttrs,
+		columnPathEventAttrKey, columnPathEventAttrString, columnPathEventAttrInt, columnPathEventAttrDouble, columnPathEventAttrBool)
+	if err != nil {
+		return nil, errors.Wrap(err, "creating event attribute iterator")
+	}
+
+	// if no intrinsics and no primary then we can just return the attribute iterator
+	// if len(iters) == 0 && primaryIter == nil {
+	// 	return attrIter, nil
+	// }
```
stoewer, line 178: Is this meant to be commented out?  Just asking because the equivalent condition in `createDistinctLinkIterator` is not commented out (see L236)

stoewer:
> One caveat is because links and events are on the same definition level, they cannot be used to filter each other out effectively. I prefer to get the main functionality out first and create a new ticket to add this ability

I think it's a good idea to the the main functionality out and handle edge cases in another issue

stoewer:
Tested it locally and it works as expected. Just left some minor comments

modules/ingester/instance_search_test.go
ie-pham: ahhh I see, it's v11 in the pb.go files

tempodb/encoding/vparquet4/block_autocomplete.go
```diff
@@ -114,6 +128,128 @@ func createDistinctIterator(
...
+	for _, cond := range conditions {
+		// Intrinsic?
+		switch cond.Attribute.Intrinsic {
+		case traceql.IntrinsicEventName:
+			pred, err := createStringPredicate(cond.Op, cond.Operands)
+			if err != nil {
+				return nil, err
+			}
+			iters = append(iters, makeIter(columnPathEventName, pred, columnSelectAs[columnPathEventName]))
+			addSelectAs(cond.Attribute, columnPathEventName, columnPathEventName)
```
mdisibio, line 162: Should these 2 lines be flipped?  It looks like it reads `columnSelectAs` before setting it.    

The next thought is that maybe we don't need the `columnSelectAs` map at all.   Since this creates the iter directly in the case statement, we don't need to track it in the map.  `CreateDistinctSpan/Resource/Iterator` funcs do that because they create the iters later after assessing all conditions.  But `createDistinctTraceIterator` is doing it like here, and doesn't need the map.

tempodb/encoding/vparquet4/block_autocomplete.go
```diff
@@ -114,6 +128,128 @@ func createDistinctIterator(
...
+	for _, cond := range conditions {
+		// Intrinsic?
+		switch cond.Attribute.Intrinsic {
+		case traceql.IntrinsicLinkTraceID:
+			pred, err := createBytesPredicate(cond.Op, cond.Operands, false)
+			if err != nil {
+				return nil, err
+			}
+			iters = append(iters, makeIter(columnPathLinkTraceID, pred, columnPathLinkTraceID))
+			columnSelectAs[columnPathLinkTraceID] = "" // Don't select, just filter
```
mdisibio, line 214: Same thought here:  I don't think we need this map, and doesn't look like it's read?    To check the column but not select it, change the prev line to:  `iters = append(iters, makeIter(columnPathLinkTraceID, pred, ""))`

mdisibio (approved):
Events tested locally and working well.  I don't have a good testbed for links, but lgtm.

PR #3850, opened 2024-07-08
feat: refactor Tempo Vulture to reduce code complexity

javiermolinar:
**What this PR does**:
This PR aims to reduce the complexity of Vulture by refactoring the code into different functions. New tests have been added to provide a ground base for future changes.
Since this is a refactor, **the behavior is unaltered**. Some of the changes included:

- No references to global variables
- External dependencies are now interfaces
- Deduplicate function traceInTraces 
- Consistent use of the logger

Although the code coverage is not a goal itself it has been increased from 12% to 80%. Only the main method is not covered.

**Checklist**
- [ X] Tests updated
- [ ] Documentation added
- [ X] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`


pkg/util/trace_info.go
```diff
@@ -37,6 +37,12 @@ type TraceInfo struct {
 	tempoOrgID          string
 }
 
+// JaegerClient is an interface used to mock the underlying client in tests.
+type JaegerClient interface {
```
zalegrala, line 41: I slightly wonder if we should make this a generic grpc client interface.  For example, I think it would be good for the project to eventually have an OTEL vulture, which requires a bunch of shuffling around.  The interfaces you've started here are a step in that direction, and so thinking about what that might mean for this interface here would be good too.

javiermolinar: It's a good point to use Vulture to test other ways to ingest data into Tempo but it may be a premature optimization. The idea of this PR is to increase confidence in the code by adding some test coverage. We can explore further changes later. What do you think @zalegrala?

zalegrala: That sounds good to me.  Just something to think about for the future.

joe-elliott (approved):

PR #3851, opened 2024-07-09
chore: update mixin deps and add recording_rules_range_interval

### LINKED ISSUE #3815: Tempo mixin: alert and recording rules make range interval configurable

In our case we have scrape interval of 1m, however one alert and recording rules in the mixin seem to have hardcoded range interval, examples:

https://github.com/grafana/tempo/blob/main/operations/tempo-mixin/alerts.libsonnet#L202
https://github.com/grafana/tempo/blob/main/operations/tempo-mixin-compiled/rules.yaml#L4

Have a config option for the range interval window in alerts, similar to what is already present for recording rules in mimir mixin. https://github.com/grafana/mimir/blob/main/operations/mimir-mixin/config.libsonnet#L666.

Manually editing the alerts after they are rendered, which kind of defeats the point of mixins imho.

**Additional context**

https://www.robustperception.io/what-range-should-i-use-with-rate/
I can raise a PR for the changes.

jmichalek132:
**What this PR does**:
* Updating the mixin deps (needed to have newer version of `histogramRules` which allows specifying range_interval)
* Adding a mixin config option `recording_rules_range_interval` to allow overriding the range_interval in recording rules.

Fixes #3815

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/operations/tempo-mixin/alerts.libsonnet
+++ b/operations/tempo-mixin/alerts.libsonnet
@@ -199,7 +199,7 @@
             alert: 'TempoProvisioningTooManyWrites',
             // 30MB/s written to the WAL per ingester max
             expr: |||
-              avg by (%s) (rate(tempo_ingester_bytes_received_total{job=~".+/ingester"}[1m])) / 1024 / 1024 > 30
+              avg by (%s) (rate(tempo_ingester_bytes_received_total{job=~".+/ingester"}[5m])) / 1024 / 1024 > 30
             ||| % $._config.group_by_cluster,
             'for': '15m',
             labels: {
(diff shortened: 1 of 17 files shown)


operations/tempo-mixin/alerts.libsonnet
```diff
@@ -199,7 +199,7 @@
             alert: 'TempoProvisioningTooManyWrites',
             // 30MB/s written to the WAL per ingester max
             expr: |||
-              avg by (%s) (rate(tempo_ingester_bytes_received_total{job=~".+/ingester"}[1m])) / 1024 / 1024 > 30
```
jmichalek132, line 202: Is it okay to just bump this to 5m, or  should it be configurable trough the config?

mapno: The other alerts also use `5m`. I think it's fine this way.

mapno (commented):
LGTM. Could you add a changelog entry?

mapno (approved):
Nice, thanks Juraj :)

PR #3856, opened 2024-07-10
[tempo-cli] Add option to delete traces by id

joe-elliott:
**What this PR does**:
Adds an option to delete traces by id to the tempo cli.

Example:
```
go run ./cmd/tempo-cli rewrite-blocks delete-trace <id> <tenant>
```

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo-cli/cmd-rewrite-blocks.go.go
+++ b/cmd/tempo-cli/cmd-rewrite-blocks.go.go
@@ -1,0 +1,129 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strconv"
+
+	"github.com/google/uuid"
+	"github.com/grafana/tempo/pkg/boundedwaitgroup"
+	"github.com/grafana/tempo/pkg/util"
+	"github.com/grafana/tempo/tempodb"
+	"github.com/grafana/tempo/tempodb/backend"
+	"github.com/grafana/tempo/tempodb/encoding"
+	"github.com/grafana/tempo/tempodb/encoding/common"
+)
+
+type deleteTraceCmd struct {
+	backendOptions
+
+	TraceID  string `arg:"" help:"trace ID to retrieve"`
+	TenantID string `arg:"" help:"tenant ID to search"`
+}
+
+func (cmd *deleteTraceCmd) Run(ctx *globalOptions) error {
+	r, _, _, err := loadBackend(&cmd.backendOptions, ctx)
+	if err != nil {
+		return err
+	}
+
+	id, err := util.HexStringToTraceID(cmd.TraceID)
+	if err != nil {
+		return err
+	}
+
+	blocks, err := blocksWithTraceID(context.Background(), r, cmd.TenantID, id)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/tempo-cli/cmd-rewrite-blocks.go
```diff
@@ -0,0 +1,244 @@
...
+
+	if newMeta.TotalObjects != meta.TotalObjects-1 {
+		return nil, fmt.Errorf("expected output to have one less object then in. out: %d in: %d", newMeta.TotalObjects, meta.TotalObjects)
+	}
+
+	return newMeta, nil
+}
+
+func blocksWithTraceID(ctx context.Context, r backend.Reader, tenantID string, traceID common.ID) ([]*backend.BlockMeta, error) {
+	blockIDs, compactedBlockIDs, err := r.Blocks(context.Background(), tenantID)
```
mapno, line 166: Is it worth increasing the amount of work by picking also compacted blocks? They'll be deleted eventually anyway.

cmd/tempo-cli/cmd-rewrite-blocks.go
```diff
@@ -0,0 +1,244 @@
...
+			if meta != nil {
+				resultsCh <- meta
+			}
+		}(blockNum, id)
+	}
+
+	wg.Wait()
+	close(resultsCh)
+
+	results := make([]*backend.BlockMeta, 0)
```
mapno, line 199:
```suggestion
	results := make([]*backend.BlockMeta, 0, len(resultsCh))
```

tempodb/encoding/common/interfaces.go
```diff
@@ -75,6 +75,10 @@ type CompactionOptions struct {
 	BlockConfig        BlockConfig
 	Combiner           model.ObjectCombiner
 
+	// DropObject can be used to drop a trace from the compaction process. Currently it only receives the ID
+	// of the trace to be compacted. If it returns false, the trace will be dropped.
```
javiermolinar, line 79: " If it returns true, the trace will be dropped" ?

javiermolinar: Ah I see, nevermind

javiermolinar, line 79: It should not be: is it returns true, the trace will be dropped? The "DropObject" name of the method suggests the opposite.

cmd/tempo-cli/cmd-rewrite-blocks.go
joe-elliott: good call. removed compactedBlockIDs

tempodb/encoding/common/interfaces.go
joe-elliott: i updated the comment to say 
```
If the function returns true, the trace will be dropped.
```
it makes sense to me that `DropObject` returning true would cause an object to be dropped.

zalegrala (approved):
This is neat.  Looks pretty tidy to me.

PR #3859, opened 2024-07-11
Traceql instant query

mdisibio:
**What this PR does**:
Adds a new api `/api/metrics/query` which performs an instant query instead of a range query.  Includes both http and grpc streaming versions. The parameters are (1) query, (2) start, and (3) end.  Wanted to match the [Prometheus instant api](https://prometheus.io/docs/prometheus/latest/querying/api/#instant-queries), but for now we can't.  TraceQL is missing the custom interval tidbits like `[1h]`, so it needs the full time range (start and end).  I think this is fine.   

This new api is actually the same, conceptually and implementation-wise, as doing a range query with 1 step.   It reuses the existing query_range sharder, pipeline, combiner, etc, and does a transformation of the response at the end. 

Extends the tempo-cli command `tempo-cli query metrics` to take an `--instant` parameter to call the new apis. Works with both http and --use-grpc streaming version.

Also tried to cleanup/solidify the time range handling.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


zalegrala (approved):
Looks good to me.

modules/frontend/metrics_query_handler.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/grafana/dskit/user"
+	"github.com/grafana/tempo/modules/frontend/combiner"
+	"github.com/grafana/tempo/modules/frontend/pipeline"
+	"github.com/grafana/tempo/pkg/api"
+	"github.com/grafana/tempo/pkg/tempopb"
+)
+
+// newMetricsQueryInstantHTTPHandler handles instant queries.  Internally these are rewritten as query_range with single step
+// to make use of the existing pipeline.
+func newMetricsQueryInstantHTTPHandler(cfg Config, next pipeline.AsyncRoundTripper[combiner.PipelineResponse], logger log.Logger) http.RoundTripper {
```
joe-elliott, line 22: can we add a streaming version? is there something that makes it difficult?

mdisibio: I left it out because there was still ongoing work in that area: https://github.com/grafana/tempo/pull/3808   If we are good, yep I can add it.

> is there something that makes it difficult

I should be asking you that 😄  No I don't think so, it should be nearly identical to query_range streaming.

joe-elliott: > I should be asking you that 😄 No I don't think so, it should be nearly identical to query_range streaming.

I didn't think so either, but thought i might be missing something :). Agree it would basically identical to query_range streaming.

mdisibio: Added.  Nice, it's quite fun to stream `{ } | count_over_time()` and watch the value count up.

joe-elliott (commented):
two questions

pkg/traceql/engine_metrics.go
```diff
@@ -65,19 +65,86 @@ func IntervalOf(ts, start, end, step uint64) int {
...
+	if wasInstant {
+		// Maintain instant nature of the request
+		req.Step = req.End - req.Start
+	} else {
+		// Realign after trimming
+		AlignRequest(req)
+	}
+}
+
+func IsInstant(req tempopb.QueryRangeRequest) bool {
```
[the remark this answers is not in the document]
mdisibio: Do you mean if this happens to occur on a "true" query range request?  For example if you queried last 5m with step 5m?  Yes it would make that query process like an instant one.  The difference is that the time range would be honored exactly, instead of being aligned to 5m intervals from the epoch.

pkg/tempopb/tempo.proto
```diff
@@ -360,6 +361,28 @@ message SpanMetricsResultPoint {
   uint64 exemplarDuration = 4;
 }
 
+message QueryInstantRequest {
+  string query = 1;
+  uint64 start = 2;
+  uint64 end = 3;
+}
+
+message QueryInstantResponse {
```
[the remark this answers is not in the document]
mdisibio: Lean towards yes, but mainly because this is the route chosen by Prom/Mimir and also [Loki](https://grafana.com/docs/loki/latest/reference/loki-http-api/#query-logs-at-a-single-point-in-time).   Happy to tweak or change directions if we can think of other deciding factors - like this format impairs Grafana transformations or something?   One difference I see is that Prom/Loki still include the timestamp in the 1 sample response for instant queries. So it really boils down to instant returns `value: 1 sample`, range returns `values: [array of samples]`.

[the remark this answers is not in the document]
joe-elliott: no, i'm good with this. spoke with @mapno as well and agree for a different endpoint i am comfortable with a different return type.

i didn't want to impose the burden of more parsing/mapping/translating to grafana, but if we do this for all other data sources i'm good.

joe-elliott (approved):

mapno (approved):
LGTM

PR #3860, opened 2024-07-11
Fix poller to maintain previous blocklist results on tenant error

zalegrala:
**What this PR does**:

Here we fix a bug where a single tenant experiences an error during polling,
which is not enough to trip the max error threshold.  In this case, the tenant
which experienced the error would be removed from the blocklist, and would not
be re-added until the next time the poller runs.  Inconsistent results in the
blocklist can cause issues on the query path for this tenant.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


mdisibio (approved):
Yep this fixes my issue locally 👍

PR #3870, opened 2024-07-16
Add native histograms for internal metrics

zalegrala:
**What this PR does**:

Here we include the necessary options for the Prometheus client to expose
native histograms for internal metrics.  Users will need to scrape the native
histograms explicitly, but this will allow a migration to begin.  There will be
some follow up to add the necessary changes to dashboards etc, but here we can
begin the process.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott (commented):
looking good.

i don't know enough about the provided options to comment. does it make sense to keep them all independent like this? will we often want to make adjustments per histogram or should we just make one place to set them like:

```
func nativeHistoOpts(o prometheus.HistogramOpts) prometheus.HistogramOpts {
  // set native histo opts

  return o
}
```

zalegrala:
My impression is that we won't make changes to these often, and if we do, they might not align in the same way.  I don't mind the change you suggest, but not sure its worth it right now.

joe-elliott (approved):

PR #3888, opened 2024-07-19
Query-frontend: Performance Improvements and Refactor

joe-elliott:
**What this PR does**:
Refactors the frontend to pass a `Request` interface instead of an `*http.Request`. This is progress toward supporting deterministic search. Also added a search pipeline benchmark and found a perf improvement in `prepareRequestForQueriers`.

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/tempo/modules/frontend
                  │  before.txt  │             after.txt              │
                  │    sec/op    │    sec/op     vs base              │
SearchPipeline-11   609.4m ± ∞ ¹   604.0m ± ∞ ¹  -0.87% (p=0.008 n=5)

SearchPipeline-11   10.642Mi ± ∞ ¹   8.659Mi ± ∞ ¹  -18.63% (p=0.008 n=5)

                  │  before.txt  │             after.txt              │
                  │  allocs/op   │  allocs/op    vs base              │
SearchPipeline-11   163.0k ± ∞ ¹   148.3k ± ∞ ¹  -9.02% (p=0.008 n=5)
```

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/frontend.go
+++ b/modules/frontend/frontend.go
@@ -22,6 +22,13 @@ import (
 	"github.com/grafana/tempo/tempodb"
 )
 
+type RoundTripperFunc func(*http.Request) (*http.Response, error)
+
+// RoundTrip implememnts http.RoundTripper
+func (fn RoundTripperFunc) RoundTrip(req *http.Request) (*http.Response, error) {
+	return fn(req)
+}
+
 // these handler funcs could likely be removed and the code written directly into the respective
 // gRPC functions
 type (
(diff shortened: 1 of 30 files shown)


mapno (commented):
Refactor looks OK. Any idea why this is faster or uses less memory?

modules/frontend/frontend.go
```diff
@@ -239,20 +245,19 @@ func newMetricsSummaryHandler(next pipeline.AsyncRoundTripper[combiner.PipelineR
 // prepareRequestForQueriers modifies the request so they will be farmed correctly to the queriers
 //   - adds the tenant header
 //   - sets the requesturi (see below for details)
-func prepareRequestForQueriers(req *http.Request, tenant string, originalURI string, params url.Values) {
+func prepareRequestForQueriers(req *http.Request, tenant string) {
```
javiermolinar, line 248: I find this function pretty obscure. Why don't we return a new http.Request based on the original one?

```go
func getQuerierHTTPRequest(origin_req *http.Request, tenant string) *http.Request{
}
```

That way we will reduce the cognitive overload reading the code

joe-elliott:
> Refactor looks OK. Any idea why this is faster or uses less memory?

I removed this call to `q.Encode()` which allocs a map:

https://github.com/grafana/tempo/pull/3888/files#diff-9247a6648bf9e5a966f795337289b01d9240bc29a30b7cbcbc2831a6b9378a13L253

I have a follow up PR coming that will remove millions of these calls from the search path.

modules/frontend/frontend.go
joe-elliott: I would prefer modifying in place to reduce allocs

mapno (approved):

PR #3895, opened 2024-07-22
feat: rename batches property of Trace to ResourceSpans to be OTEL compatible

### LINKED ISSUE #3802: Trace by ID: JSON response does not align with OTLP JSON format
**Describe the bug**
According to the Tempo [documentation for Query API](https://grafana.com/docs/tempo/latest/api_docs/#query), if the `Content-Type: application/protobuf` header is absent Tempo "returns [OpenTelemetry](https://github.com/open-telemetry/opentelemetry-proto/tree/main/opentelemetry/proto/trace/v1) JSON".

However, the returned JSON output has `batches` as the top-level key instead of `resourceSpans`. This can also be seen in the Tempo docs: https://grafana.com/docs/tempo/latest/api_docs/pushing-spans-with-http/#retrieve-traces

Because of this, it is not possible to unmarshal the returned JSON into a `TracesData` message from the official OTLP spec.
When requesting protobuf results from tempo, it can be unmarshaled into the official `TracesData` message because the protobufs are compatible.

**To Reproduce**
2. Query a trace by ID as described here: https://grafana.com/docs/tempo/latest/api_docs/#query

**Expected behavior**
Returned JSON output is compatible with OTLP spec.

**Environment:**
 - Infrastructure: Kubernetes
 - Deployment tool: helm

**Additional Context**
Is it expected to be able to unmarshal the query responses into `TracesData` messages of the official OTLP spec? The documentation appears to hint that it is.
Or should applications querying Tempo unmarshal the response into the structs provided by Tempo (eg. https://github.com/grafana/tempo/blob/main/pkg/tempopb/tempo.proto)?

javiermolinar:
**What this PR does**:
Currently, the response from ``/api/traces/`` when the the header ``application/json`` is set is not OTEL compatible.
The goal of this PR is to prepare the ground for a V2 traces API by making the response OTEL compatible without breaking the current behavior.

I have explored two approaches. The first one was creating a new TraceV2 proto message type. But that introduced a cascade of changes ending with cluttered code, adding even more complexity to the code base. With this approach, we keep the code as it is while being more descriptive by replacing the generic ``Batches`` with a more specific ``ResourceSpans``. 

To make it retro-compatible, the response when the content-type header is ``application/json`` remains the same:
https://github.com/grafana/tempo/blob/a66720a018a2fa9631d286f6cb9d9a39ead9344c/modules/frontend/combiner/trace_by_id.go#L125

Fixes #3802
**Checklist**
- [X ] Tests updated
- [ ] Documentation added
- [ X] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`


pkg/tempopb/trace_utils.go
```diff
@@ -0,0 +1,28 @@
+package tempopb
+
+import (
+	"strings"
+
+	"github.com/gogo/protobuf/jsonpb"
+)
+
+// Marshal a Trace to an OTEL compatible JSON replacing resourceSpans by batches
+func MarshalToJSONV1(t *Trace) ([]byte, error) {
```
joe-elliott, line 10: it would be nice to get a short comment about why this method exists. how we renamed the batches field to resourceSpans and these utility methods are for (un)marshaling json from the v1 traces endpoint to the current proto object

.gitignore
```diff
@@ -18,3 +18,4 @@
 /tempodb/encoding/benchmark_block
 private-key.key
 integration/e2e/e2e_integration_test[0-9]*
+.tempo.yaml
```
joe-elliott, line 21: what's this?

joe-elliott (approved):

.gitignore
javiermolinar: It is my local config to debug Tempo. I can leave it out

pkg/tempopb/trace_utils.go
javiermolinar: Sure, I can add more info

.gitignore
joe-elliott: it's fine. just interested

PR #3903, opened 2024-07-24
[bugfix] Remove extra primary iter

ie-pham:
**What this PR does**: Primary iter is appended twice before. This fixes the issue in query range group by event attributes. 

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


mapno (commented):
LGTM. Do we need a changelog entry? Is the changes in #3847 needed now?

joe-elliott (approved):

PR #3911, opened 2024-07-25
fix: remove unused properties from the WAL config file

### LINKED ISSUE #3902: Bug in wal.go causes c.BlocksFilepath config to be overwritten in tempo.yaml
**Describe the bug**
Configuration file section for `cBlocksFilepath` is overwritten in tempo.yaml. 

@WesselAtWork reported [this issue in PR 3643](https://github.com/grafana/tempo/pull/3643/files#r1624087743): 

[tempo/tempodb/wal/wal.go](https://github.com/grafana/tempo/blob/3a4f309b0d9413fc8762daaa0bd9769a08cc8bf8/tempodb/wal/wal.go#L70-L75)

Lines 70 to 75 in [3a4f309](https://github.com/grafana/tempo/commit/3a4f309b0d9413fc8762daaa0bd9769a08cc8bf8)
```
 p := filepath.Join(c.Filepath, blocksDir) 
 err = os.MkdirAll(p, os.ModePerm) 
 if err != nil { 
 	return nil, err 
 } 
 c.BlocksFilepath = p
```

I think it's supposed to look like this (but with c.BlocksFilepath == ""):
[tempo/tempodb/wal/wal.go](https://github.com/grafana/tempo/blob/3a4f309b0d9413fc8762daaa0bd9769a08cc8bf8/tempodb/wal/wal.go#L56-L67)

Lines 56 to 67 in [3a4f309](https://github.com/grafana/tempo/commit/3a4f309b0d9413fc8762daaa0bd9769a08cc8bf8)
```
 // The /completed/ folder is now obsolete and no new data is written, 
 // but it needs to be cleared out one last time for any files left 
 // from a previous version. 
 if c.CompletedFilepath == "" { 
 	completedFilepath := filepath.Join(c.Filepath, completedDir) 
 	err = os.RemoveAll(completedFilepath) 
 	if err != nil { 
 		return nil, err 
 	} 
  
 	c.CompletedFilepath = completedFilepath 
 }
```

**To Reproduce**

**Expected behavior**

**Environment:**

**Additional Context**

javiermolinar:
**What this PR does**:

It removes the properties ``completedfilepath`` and ``blocksfilepath`` from the configuration of the WAL. The former one is not needed anymore and setting the latter one is a bit pointless since it can be just hanging from ``file path``

Fixes #3902

**Checklist**
- [X] Tests updated
- [X] Documentation added


CHANGELOG.md
```diff
@@ -50,7 +50,7 @@
 * [BUGFIX] Improved handling of complete blocks in localblocks processor after enabling flusing [#3805](https://github.com/grafana/tempo/pull/3805) (@mapno)
 * [BUGFIX] Handle out of boundaries spans kinds [#3861](https://github.com/grafana/tempo/pull/3861) (@javiermolinar)
 * [BUGFIX] Maintain previous tenant blocklist on tenant errors [#3860](https://github.com/grafana/tempo/pull/3860) (@zalegrala)
-
+* [BUGFIX] Remove unused properties from the WAL configuration [#3911](https://github.com/grafana/tempo/pull/3911) (@javiermolianr)
```
joe-elliott, line 53: we need to list this as a breaking change. we do that by typing `**Breaking Change**` on the changelog line like this:

https://github.com/grafana/tempo/blob/d0f0a5fddb7b3234e70f4bd39bb592a8e0b5df50/CHANGELOG.md?plain=1#L3

that will help us find and note the removal of these config optoins for our users when we cut 2.6

docs/sources/tempo/configuration/manifest.md
```diff
@@ -597,8 +595,6 @@ storage:
             queue_depth: 20000
         wal:
             path: /var/tempo/wal
-            completedfilepath: /var/tempo/wal/completed
```
joe-elliott, line 600: nice. i'm impressed you found and updated the manifest :)

CHANGELOG.md
javiermolinar: Done, I didn't think this was a breaking change because the configuration was not honored ever :P

joe-elliott: That is a good point, but we are removing config options that previously were parsed. It might break someone's config I suppose.

joe-elliott (approved):

PR #3912, opened 2024-07-25
feat: expose a Traces api V2

javiermolinar:
**What this PR does**:

This PR adds a new API endpoint for retrieving traces based on the trace id.
It includes the following changes:
- **Querier**:  
        - No longer returns a 404 when the trace is not found. Now it returns a 200 OK and an empty trace response
        - The V2 always returns a proto message. **This is pretty opinionated**
- **Frontend**: 
        - The ``TraceByIDResponse`` now wraps the trace. This leaves room for adding extra metadata in the response.
        - The response is now OTEL compatible

**Implementation details**
- The new ``NewTraceByIDV2`` combiner now benefits from the common combiner
- The common combiner now can serialize to protobuff for an HTTP request based on the request content-type
- The handlers on the querier and frontend are quite similar to the V1 ones. The code can be refactored but will add more complexity and no benefits.

**Post PR**
- We should look into a code refactoring / structuring it in a better way.

**Checklist**
- [X] Tests updated
- [] Documentation added


modules/frontend/traceid_handlers.go
```diff
@@ -83,3 +83,72 @@ func newTraceIDHandler(cfg Config, o overrides.Interface, next pipeline.AsyncRou
 		return resp, err
 	})
 }
+
+func newTraceIDHandlerV2(cfg Config, o overrides.Interface, next pipeline.AsyncRoundTripper[combiner.PipelineResponse], logger log.Logger) http.RoundTripper {
```
joe-elliott, line 87: i'm not opposed to forking everything if you think it's the correct choice, but what do you think about using the same pipeline as is? except now it returns the full TraceByIDResponse instead of just the trace?

modules/querier/http.go
```diff
@@ -100,6 +100,71 @@ func (q *Querier) TraceByIDHandler(w http.ResponseWriter, r *http.Request) {
...
+		BlockStart: blockStart,
+		BlockEnd:   blockEnd,
+		QueryMode:  queryMode,
+	}, timeStart, timeEnd)
+	if err != nil {
+		handleError(w, err)
+		return
+	}
+
+	if r.Header.Get(api.HeaderAccept) == api.HeaderAcceptProtobuf {
```
joe-elliott, line 142: i think we have a method that handles some of this:

`writeFormattedContentForRequest`

joe-elliott (commented):
one Q about forking vs code reuse. to be clear i'm on the fence myself. i will trust whatever direction you want to take this.

should we go ahead and add a "message" field to the tracebyidresponse? since we know we want it?

> The V2 always returns a proto message. This is pretty opinionated

is this true? it looks like you added code to the common combiner to respond to marshal as either proto or json.

> The new NewTraceByIDV2 combiner now benefits from the common combiner

nice, i've been wanting to do this, but haven't had a chance.

modules/frontend/traceid_handlers.go
javiermolinar: I didn't want to add any conditional expression (V1, V2) to reuse the code but you are right. We can change the handler function to accept the combiner and that way we don't need to fork anything. The only thing that bothers me is that the combiner has a lot of things on its plate.

modules/querier/http.go
javiermolinar: Oh, how I didn't notice? 🙀

javiermolinar:
>>should we go ahead and add a "message" field to the tracebyidresponse? since we know we want it?

I rather close this one and add it on a different PR

> > The V2 always returns a proto message. This is pretty opinionated
> 
> is this true? it looks like you added code to the common combiner to respond to marshal as either proto or json.

Yeah, this was true at some point in time, I guess I changed my mind 🫤

modules/frontend/combiner/common.go
```diff
@@ -129,15 +129,25 @@ func (c *genericCombiner[T]) HTTPFinal() (*http.Response, error) {
...
-		return nil, fmt.Errorf("error marshalling response body: %w", err)
+	var bodyString string
+	if c.marshallingFormat == api.HeaderAcceptProtobuf {
+		buff, err := proto.Marshal(final)
+		if err != nil {
+			return nil, fmt.Errorf("error marshalling response body: %w", err)
+		}
+		bodyString = string(buff)
+	} else {
+		c.marshallingFormat = api.HeaderAcceptJSON
```
joe-elliott, line 140: i find it a little strange to set this here. this feels like something that should be set on creation of the struct to control the response behavior and never changed.

i don't think it currently is causing a bug, but since its unexpected i could see how it would trip someone up.

modules/frontend/combiner/common.go
```diff
@@ -129,15 +129,25 @@ func (c *genericCombiner[T]) HTTPFinal() (*http.Response, error) {
...
+		bodyString, err = new(jsonpb.Marshaler).MarshalToString(final)
+		if err != nil {
+			return nil, fmt.Errorf("error marshalling response body: %w", err)
+		}
 	}
 
 	return &http.Response{
 		StatusCode: c.httpStatusCode,
```
joe-elliott, line 148: it's silly we set this to 200 in all the combiners and then return it here. we should hardcode this to return 200 and only set `c.httpStatusCode` if there's an error

the combiner was brought together from ~3 different sources when i refactored the frontend. it created some warts like this b/c i wanted to change as little code as possible

this (and the above comment) suggest we really need a `newCombiner(foo &genericCombiner)` method that properly initializes these fields

knylander-grafana:
Should we add docs for this?

joe-elliott:
> Should we add docs for this?

Great call out. Yes! Javi, api docs are here:

https://github.com/grafana/tempo/blob/main/docs/sources/tempo/api_docs/_index.md

modules/frontend/combiner/common.go
javiermolinar: Yes, I know what it looks like. The reason for that is to avoid a scenario where the implementation of the combiner function does not set it and therefore we will return an empty header. Another reason is how difficult is to know what kind of combiner are we returning since we always use the ``Combiner`` interface to later on do a type assertion:

```
NewSearchTagValues(limitBytes).(GRPCCombiner[*tempopb.SearchTagValuesResponse])
```

I wonder if we could return directly 

```
func NewSearchTags(limitBytes int) GRPCCombiner[*tempopb.SearchTagsResponse]
```

Nevertheless I have renamed it to httpMarshalingFormat to be more explicit

modules/frontend/combiner/common.go
javiermolinar: I'm really on the fence on this. Especially because you don't need to initiate the status code to make it work. In the first iteration, it won't stop and it will be replaced by the response's one. But it's "magic". I've added the method to make it more explicit

javiermolinar:
> Should we add docs for this?

Hi @knylander-grafana. I've added the documentation. Some questions:
- The current API is not OTEL compatible, that's why I've added a **mostly compatible** in the existing section
- The response from the V2 is yet not completed, not sure if we should state something or just keep adding new fields as we go along.

joe-elliott (approved):
love it

> The response from the V2 is yet not completed, not sure if we should state something or just keep adding new fields as we go along.

let's add the fields to the docs as we add them in code 👍

modules/frontend/combiner/common.go
```diff
@@ -129,15 +137,24 @@ func (c *genericCombiner[T]) HTTPFinal() (*http.Response, error) {
...
+	} else {
+		bodyString, err = new(jsonpb.Marshaler).MarshalToString(final)
+		if err != nil {
+			return nil, fmt.Errorf("error marshalling response body: %w", err)
+		}
 	}
 
 	return &http.Response{
-		StatusCode: c.httpStatusCode,
+		StatusCode: 200,
```
mapno, line 155: Doesn't this change the behaviour for all endpoints?

CHANGELOG.md
```diff
@@ -42,7 +42,9 @@
 * [ENHANCEMENT] Rename batches property of Trace to ResourceSpans to be OTEL compatible [#3895](https://github.com/grafana/tempo/pull/3895)
 * [ENHANCEMENT] Reduce memory consumption of query-frontend[#3888](https://github.com/grafana/tempo/pull/3888) (@joe-elliott)
 * [ENHANCEMENT] Reduce log level verbosity for e2e tests[#3900](https://github.com/grafana/tempo/pull/3900) (@javiermolinar)
-* [ENHANCEMENT] Support reloading client certificates [#537](https://github.com/grafana/dskit/pull/537) (@rubenvp8510)
```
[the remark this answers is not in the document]
javiermolinar: Actually, I'm correcting that. Andreas pasted the wrong PR, one from dskit instead of him:

https://github.com/grafana/tempo/pull/3915/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR45

modules/frontend/combiner/common.go
javiermolinar: No, only the TraceId handler uses that combiner, and it was returning 200 always. Ultimately the status code in the response can be modified in the handler once the whole pipeline has finished

mapno: Isn't `genericCombiner` used in other combiners? See https://github.com/grafana/tempo/blob/main/modules/frontend/combiner/search_tag_values.go#L17

javiermolinar: Yes, but in the form of GRPC combiner:

https://github.com/grafana/tempo/blob/6e9d93aefe7f103dc394c068e8d558ff3e4d40db/modules/frontend/combiner/search_tag_values.go#L41

Which doesn't use the HTTPFinal

mapno: `HTTPFinal()` is called by `httpCollector` in [`RoundTrip()`](https://github.com/grafana/tempo/blob/main/modules/frontend/pipeline/collector_http.go#L47), which is instantiated in all pipelines ([example](https://github.com/grafana/tempo/blob/main/modules/frontend/metrics_query_range_handler.go#L95))

javiermolinar: But even more, it doesn't matter because of this line:

https://github.com/grafana/tempo/blob/6e9d93aefe7f103dc394c068e8d558ff3e4d40db/modules/frontend/combiner/common.go#L122

Which is going to return an errored HTTP response if the status code != 200. This is kinda wrong since the queriers may return another code in the 200 range like 204.  

So, it doesn't matter at all because it will always return 200 unless there is an error :P

javiermolinar: > `HTTPFinal()` is called by `httpCollector` in [`RoundTrip()`](https://github.com/grafana/tempo/blob/main/modules/frontend/pipeline/collector_http.go#L47), which is instantiated in all pipelines ([example](https://github.com/grafana/tempo/blob/main/modules/frontend/metrics_query_range_handler.go#L95))

Super confusing this stuff since there is also a GRPCCollector.
I don't mind to change it again if you want

mapno: Oh okay, I see now. It's definitely confusing.

mapno (approved):

PR #3923, opened 2024-07-30
Update processor replacement logic for native histograms

zalegrala:
**What this PR does**:

Here we make an adjustment the processor replacement logic to include the previous and current values of the desired histogram implementation.  When moving from the new implementation to the old implementation, the servicegraph and span-metrics processors need to be replaced so that when NewHistogram is called on the registry, it can make the determination about which implementation to use.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/instance.go
+++ b/modules/generator/instance.go
@@ -71,6 +71,7 @@ type instance struct {
 	instanceID             string
 	overrides              metricsGeneratorOverrides
 	ingestionSlackOverride atomic.Int64
+	histogramsOverride     atomic.String
 
 	registry *registry.ManagedRegistry
 	wal      storage.Storage
@@ -110,6 +111,16 @@ func newInstance(cfg *Config, instanceID string, overrides metricsGeneratorOverr
 		reg:    reg,
 		logger: logger,
 	}
+
+	// NOTE: When a new histogram is created, the overrides are checked to determine
+	// which histogram implementaion to use.  This happens when the servicegraphs
+	// or spanmetrics processors are created.  The new implementation can
+	// generate native histograms or both classic and native.  However, if the
+	// new implementation is chosen, and then the overrieds are updated to
+	// indicated that the old implementation should be use by setting "classic",
+	// we need to know if the processors should be re-created.  Store the current
+	// value for later comparison.
+	i.histogramsOverride.Store(i.overrides.MetricsGeneratorGenerateNativeHistograms(i.instanceID))
 
 	err := i.updateProcessors()
 	if err != nil {
@@ -238,6 +249,7 @@ func (i *instance) updateProcessors() error {
 		}
 	}
 
+	i.histogramsOverride.Store(i.overrides.MetricsGeneratorGenerateNativeHistograms(i.instanceID))
 	i.updateProcessorMetrics()
 
 	return nil
@@ -246,6 +258,14 @@ func (i *instance) updateProcessors() error {
 // diffProcessors compares the existing processors with the desired processors and config.
 // Must be called under a read lock.
 func (i *instance) diffProcessors(desiredProcessors map[string]struct{}, desiredCfg ProcessorConfig) (toAdd, toRemove, toReplace []string, err error) {
+	var reloadForHistograms bool
+
+	currentHistogramsOverrides := i.overrides.MetricsGeneratorGenerateNativeHistograms(i.instanceID)
+	previousHisograms := i.histogramsOverride.Load()
+	if currentHistogramsOverrides == "classic" && currentHistogramsOverrides != previousHisograms {
+		reloadForHistograms = true
+	}
+
 	for processorName := range desiredProcessors {
 		if _, ok := i.processors[processorName]; !ok {
 			toAdd = append(toAdd, processorName)
@@ -259,11 +279,11 @@ func (i *instance) diffProcessors(desiredProcessors map[string]struct{}, desired
 
 		switch p := processor.(type) {
 		case *spanmetrics.Processor:
-			if !reflect.DeepEqual(p.Cfg, desiredCfg.SpanMetrics) {
+			if !reflect.DeepEqual(p.Cfg, desiredCfg.SpanMetrics) || reloadForHistograms {
 				toReplace = append(toReplace, processorName)
 			}
 		case *servicegraphs.Processor:
-			if !reflect.DeepEqual(p.Cfg, desiredCfg.ServiceGraphs) {
+			if !reflect.DeepEqual(p.Cfg, desiredCfg.ServiceGraphs) || reloadForHistograms {
 				toReplace = append(toReplace, processorName)
 			}
 		case *localblocks.Processor:
--- a/modules/generator/instance.go
+++ b/modules/generator/instance.go
@@ -113,10 +113,10 @@ func newInstance(cfg *Config, instanceID string, overrides metricsGeneratorOverr
 	}
 
 	// NOTE: When a new histogram is created, the overrides are checked to determine
-	// which histogram implementaion to use.  This happens when the servicegraphs
+	// which histogram implementation to use.  This happens when the servicegraphs
 	// or spanmetrics processors are created.  The new implementation can
 	// generate native histograms or both classic and native.  However, if the
-	// new implementation is chosen, and then the overrieds are updated to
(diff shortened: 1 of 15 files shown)


modules/generator/instance.go
```diff
@@ -111,6 +112,16 @@ func newInstance(cfg *Config, instanceID string, overrides metricsGeneratorOverr
...
 
+	// NOTE: When a new histogram is created, the overrides are checked to determine
+	// which histogram implementation to use.  This happens when the servicegraphs
+	// or spanmetrics processors are created.  The new implementation can
+	// generate native histograms or both classic and native.  However, if the
+	// new implementation is chosen, and then the overrides are updated to
+	// indicated that the old implementation should be use by setting "classic",
+	// we need to know if the processors should be re-created.  Store the current
+	// value for later comparison.
+	i.histogramsOverride.Store(i.overrides.MetricsGeneratorGenerateNativeHistograms(i.instanceID))
```
kvrhdn, line 123: Do we need this `Store()`? Since we call `updateProcessors` immediately after this we store it there anyways.

We can maybe drop this line and move the NOTE to `updateProcessors`?

modules/generator/instance.go
```diff
@@ -246,6 +258,14 @@ func (i *instance) updateProcessors() error {
 // diffProcessors compares the existing processors with the desired processors and config.
 // Must be called under a read lock.
 func (i *instance) diffProcessors(desiredProcessors map[string]struct{}, desiredCfg ProcessorConfig) (toAdd, toRemove, toReplace []string, err error) {
+	var reloadForHistograms bool
+
+	currentHistogramsOverrides := i.overrides.MetricsGeneratorGenerateNativeHistograms(i.instanceID)
+	previousHisograms := i.histogramsOverride.Load()
+	if currentHistogramsOverrides == "classic" && currentHistogramsOverrides != previousHisograms {
```
kvrhdn, line 265: Maybe add a small comment why we only reload if `currentHistogramsOverrides == "classic"`. Might not be obvious for everyone.

modules/generator/instance.go
```diff
@@ -238,6 +249,7 @@ func (i *instance) updateProcessors() error {
 		}
 	}
 
+	i.histogramsOverride.Store(i.overrides.MetricsGeneratorGenerateNativeHistograms(i.instanceID))
```
kvrhdn, line 252: I like this implementation, it's straightforward and easy to understand.

Only concern is what if the override changes _while_ we are creating the processors. So at the start of `updateProcessors` the override might be `native` and at the end `classic`. This would mean we create the processors with native histograms but then store classic in our flag.

This race condition seems fairly unlikely but I don't think it's impossible.

kvrhdn (commented):
Nice work, I appreciate the test too 🙂

modules/generator/instance.go
kvrhdn: I guess the workaround would be to store the override at the start of `updateProcessors` and then check again at the end, and if it changed restart the whole process? I'm also worried this would make it too convoluted.

zalegrala: If such a race happened, this would get corrected on the next reload correct, or are you seeing something more erroneous?

modules/generator/instance.go
zalegrala: Yep, your right.  Since the initial processors will be empty, we store the value before we need it on the next iteration.

modules/generator/instance.go
kvrhdn: Okay thinking about this some more I think we are usually fine. It's only when we go `native` -> `classic` that we need to reload. In the other direction we don't need to reload.

Timeline I was thinking about:

---

Initial situation:
- generate histograms: `native`

\*Some\* overrides change:
- `diffProcessors` detects some of the settings changed -> we start running `updateProcessors`
- `updateProcessors` is running, it's creating processors with `native` histograms

We change the overrides _again_, setting histograms `native` -> `classic`:
- `updateProcessors` is still replacing processors
- the registry picks up the new override and starts creating `classic` histograms now
- `updateProcessors` (finally) finishes and stores `classic` in `histogramsOverride`
  - because the override is currently `classic` we made the assumption that all histograms were created as `classic`, but we don't actually know that 100%

End result:
- some of the histograms will still be created as `native` while the others will already be `classic`
- next time we run `diffProcessors` we don't detect a difference and keep it as-is

---

This is pretty unlikely because it assumes:
- someone changes the overrides twice and both changes need to trigger `updateProcessors` (so `classic`>`native` and then `native`>`classic` would not work)
- the 2nd update doing `native` > `classic` has to happen while `updateProcessors` is busy
  - I think this function is fairly fast, so it's very unlikely

So it's possible to create histograms that are mixed `classic` and `native`, but the window is very small. This would persist until we reload the processors again or restart the instance.

kvrhdn: To avoid this we would have to pass down whether we want to create classic or native histograms from `instance`, through the processors all the way to the registry. That way we know exactly which type is being created and we don't risk that registry picks up another setting mid-update.
So this requires a lot of changes honestly for a very unlikely scenario.

kvrhdn (approved):
I'm fine with merging this, I think the race conditions is very theoretical and shouldn't occur in practice. If it does the impact would be that we generate the wrong type of histograms for a while which is not a big deal and can be resolved with a restart.

zalegrala:
I've pushed a significant commit to reflect the changes that we discussed.  I didn't like adding a known race even if it was probably never going to surface.

Now the each processor config will hold the values, which ends up working nicely with the configuration diff, since we compare the configuration of the processors to determine if we need to reload them.  Each new processor config option is not exposed to the user, but only used internally to determine if the processor needs to be replaced.  The tests have been updated to ensure we only reload the processors when desired.

A slight change here is that any change to the processor config will cause the reload.  This means that to go from "native" -> "both" will reload the processor, even though we will end up selecting the same implementation.  I think this is probably a fine trade off.  I'd like to deploy this change somewhere and observe the behavior probably next week, but I think its ready for another look @kvrhdn.

modules/generator/registry/interface.go
```diff
@@ -4,7 +4,7 @@ package registry
 type Registry interface {
 	NewLabelValueCombo(labels []string, values []string) *LabelValueCombo
 	NewCounter(name string) Counter
-	NewHistogram(name string, buckets []float64) Histogram
+	NewHistogram(name string, buckets []float64, histogramOverride string) Histogram
```
kvrhdn, line 7: Minor: can we make this an enum? I know enums are kind of weak in Go, but it helps document the interface what values are allowed.

modules/generator/registry/native_histogram.go
```diff
@@ -176,18 +174,8 @@ func (h *nativeHistogram) collectMetrics(appender storage.Appender, timeMs int64
...
-		// implementation. This is because the registry is not recreated when the
-		// overrides change. This is a tradeoff to avoid recreating the registry
-		// for every change in the overrides, but it does mean that even if user
-		// changes to "classic" histograms, we will still be using the new
-		// histograms implementation and want to avoid generating native
-		// histograms.
-
 		// If we are in "both" or "classic" mode, also emit classic histograms.
-		if overrides.HasClassicHistograms(h.modeFunc()) {
+		if overrides.HasClassicHistograms(h.histogramOverride) {
```
kvrhdn, line 178: Interesting side-effect from making the override a configuration setting. Just replacing the histogram all together instead of dynamically switching is a bit more overhead but I think it will be more reliable/sturdy.

kvrhdn (approved):
Thanks for putting in the extra effort, this really cleaned up the implementation and makes it more future-ready. Nice work!

I left a minor comment about changing `histogramOverride` into an enum for documentation purposes. Enums aren't enforced in Go but it shows what the values are we expect.

modules/generator/registry/native_histogram.go
zalegrala: Right, I noticed this also.  Sturdy sounds right to me.

modules/generator/registry/interface.go
zalegrala: The diff is somewhat large.  Let's take this and I'll put up another for this request.

zalegrala: https://github.com/grafana/tempo/pull/3938 to follow up.

PR #3926, opened 2024-07-31
chore: bump go module path

### LINKED ISSUE #3817: Module path needs to be updated with '/v2'
Currently this repo doesn't work correctly with "go get" because the module name is not compatible with the current major version. Please help to update the module path, like you did in the Loki project: https://github.com/grafana/loki/blob/main/go.mod#L1

davidgiga1993:
**What this PR does**:
Changes the module path to match the major version, this resolves `go get` not working.

This change was based upon the changes required for the [v3 loki release](https://github.com/grafana/loki/commit/4c88be0ef2310406206c02d7747bf73b51c95bda)

Fixes #3817

**Checklist**
- [x] Tests updated
- [ ] Documentation added


davidgiga1993:
@zalegrala yes we're trying to use Tempo as package for a custom reverse proxy to parse the AST of the query. We're doing the same thing with Prometheus and Loki (which can be consumed as a dependency)

zalegrala:
This change seems reasonable to me.  We hesitate because this will cause some additional work for us internally.  However, it might be nice to get this in before 2.6.  If this can get updated I can help review early next week.

davidgiga1993:
> This change seems reasonable to me. We hesitate because this will cause some additional work for us internally. However, it might be nice to get this in before 2.6. If this can get updated I can help review early next week.

I've resolved the conflicts

davidgiga1993:
And again

PR #3927, opened 2024-07-31
Reorganise e2e tests

mapno:
**What this PR does**:

Moves some of the heaviests e2e tests to their own folder, so they're more organised and can be called independently—e2e tests have been timing out lately.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


.github/workflows/ci.yml
```diff
@@ -119,6 +119,9 @@ jobs:
       - name: Poller
         run: make test-integration-poller
 
+      - name: Deployment modes
```
javiermolinar, line 122: I'm wondering if we could take the poller and deployment tests outside as a job to see if they can run at the same time

javiermolinar (commented):
Cool!

.github/workflows/ci.yml
javiermolinar: Trying that here: https://github.com/grafana/tempo/pull/3928

zalegrala:
The change looks reasonable to me, but I see that it still took 29 minutes to run.  Is that an improvement in the duration?

mapno:
> The change looks reasonable to me, but I see that it still took 29 minutes to run. Is that an improvement in the duration?

Not really, the main goal was to circumvent the test timeout.

I've updated the PR to split the different targets into different jobs

mapno:
Required CI checks need to be reviewed after this PR gets merged.

javiermolinar (approved):
Great!

electron0zero (approved):

PR #3928, opened 2024-07-31
try parallel

javiermolinar:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


zalegrala:
The test failure here seems unrelated, but did the jobs run in parallel?

javiermolinar:
> The test failure here seems unrelated, but did the jobs run in parallel?

I think so but the e2e suite is still doing all the heavy lift. I'll try a second round

PR #3937, opened 2024-08-05
Render template for /memberlist endpoint

zalegrala:
**What this PR does**:

Here we copy from Mimir to render a nice looking HTML page from template for
the `/memberlist` endpoint.  Additionally, enable a config flag to buffer
messages for debugging purposes which are viewable on the endpoint.

image

A nice follow-up would be to enable this for the main `/status` endpoint as well.

Locating this in `cmd/tempo/app` wasn't my first choice, but after seeing we
currently add all other endpoints in `modules.go`, I think this is fine for
now, and we can follow up if we want to improve the structure in this area,
perhaps by createing an `API` struct in `pkg/api` similar to Mimir.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


cmd/tempo/app/modules.go
```diff
@@ -422,10 +423,15 @@ func (t *App) initQueryFrontend() (services.Service, error) {
 	// http endpoint to see usage stats data
 	t.Server.HTTPRouter().Handle(addHTTPAPIPrefix(&t.cfg, api.PathUsageStats), usageStatsHandler(t.cfg.UsageReport))
 
+	t.Server.HTTPRouter().PathPrefix("/static/").HandlerFunc(http.FileServer(http.FS(staticFiles)).ServeHTTP).Methods("GET")
```
joe-elliott, line 426: i am struggling to parse this line and I think it may be the most critical in this PR. adding a static file server to Tempo is a security concern, but it looks like this may only be allowed to access preloaded files in memory?

cmd/tempo/app/modules.go
```diff
@@ -422,10 +423,15 @@ func (t *App) initQueryFrontend() (services.Service, error) {
...
+
 	// todo: queryFrontend should implement service.Service and take the cortex frontend a submodule
 	return t.frontend, nil
 }
 
+//go:embed static
+var staticFiles embed.FS
```
joe-elliott, line 433: this is blowing my mind. how are the files  in ./static mapped here?

zalegrala: I know, me too.  With this PR http requests at `/static` will return files in the `./static` directory, which are stored in this variable at compile time.

cmd/tempo/app/modules.go
zalegrala: Right, the determination is at compile time, and read only.

cmd/tempo/app/modules.go
joe-elliott: oh! 

```
//go:embed static
```

"static" means the ./static folder. my goodness

joe-elliott (approved):
looks good! agree with comments that long term we should organize this better instead of leaving it in cmd/tempo/app

knylander-grafana:
Should we update the API doc for [Memberlist](https://grafana.com/docs/tempo/latest/api_docs/)? Currently, it's very sparse and only has an entry in the table at the top of the page.

zalegrala:
Good question.  I think its okay for now @knylander-grafana, and we can follow up.  The UI implements most of the functionality that users will need, so I'm not expecting them to interact directly with the API.

PR #3938, opened 2024-08-05
Native histograms processor replacement enum

zalegrala:
**What this PR does**:

Plumb an enum into the native histograms and update overrides.

As a follow up to https://github.com/grafana/tempo/pull/3923

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo/app/overrides_validation.go
+++ b/cmd/tempo/app/overrides_validation.go
@@ -7,6 +7,7 @@ import (
 	"golang.org/x/exp/slices"
 
 	"github.com/grafana/tempo/modules/generator"
+	"github.com/grafana/tempo/modules/generator/registry"
 	"github.com/grafana/tempo/modules/overrides"
 	"github.com/grafana/tempo/modules/overrides/userconfigurable/api"
 	"github.com/grafana/tempo/modules/overrides/userconfigurable/client"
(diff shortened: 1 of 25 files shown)


zalegrala:
Taking back to draft while I work out the CI.

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -88,7 +88,7 @@ type Processor struct {
 	logger             log.Logger
 }
 
-func New(cfg Config, tenant string, registry registry.Registry, logger log.Logger) gen.Processor {
+func New(cfg Config, tenant string, reg registry.Registry, logger log.Logger) (gen.Processor, error) {
```
kvrhdn, line 91: Did something push you to add an `error` to the return signature? I think it's fine for consistency but as far as I can tell it's never set to anything else but `nil`?

cmd/tempo/app/overrides_validation.go
```diff
@@ -33,9 +33,9 @@ func (r *runtimeConfigValidator) Validate(config *overrides.Overrides) error {
 		}
 	}
 
-	if config.MetricsGenerator.GenerateNativeHistograms != "classic" &&
-		config.MetricsGenerator.GenerateNativeHistograms != "native" &&
-		config.MetricsGenerator.GenerateNativeHistograms != "both" &&
+	if config.MetricsGenerator.GenerateNativeHistograms != overrides.HistogramMethodClassic &&
```
kvrhdn, line 36: Not sure if this is better but... we could also use the `overrides.HistogramModeToValue` to check if the provided value is a valid key. By checking whether the map contains the given value as a key.
That would mean if we update the map and add a key (which is fairly unlikely tbh) this validation will automatically be updated.

kvrhdn (approved):
This change rippled out a lot more than I expected but nice cleanup!

modules/generator/processor/servicegraphs/servicegraphs.go
zalegrala: Ah yes, this might have been a detour that I didn't end up committing.  I'll back out the signature change.

cmd/tempo/app/overrides_validation.go
zalegrala: This should work.
```golang
	if v, ok := registry.HistogramModeToValue[string(config.MetricsGenerator.GenerateNativeHistograms)]; !ok {
		if string(v) != "" {
			return fmt.Errorf("metrics_generator.generate_native_histograms \"%s\" is not a valid value, valid values: classic, native, both", config.MetricsGenerator.GenerateNativeHistograms)
		}
	}
```
Do you like that better?

kvrhdn: Yep that should work :+1: Readability suffers a bit but that's more of a Go issue due to its limited enum support.

PR #3941, opened 2024-08-06
feat: Allow partial traces to be returned

javiermolinar:
**What this PR does**:
Allow return partial traces  for traces API V2

This is a prequel to a more intelligent span-dropping algorithm. For version 1 of Traces API, an error is returned when the trace size exceeds the global max bytes limit. With this PR, V2 will return a partial trace, once the limit has been reached no more spans will be added.

The change has been propagated from the front-end way down the read path. 

How the response looks like:

```json
{
  "trace": {
    "resourceSpans": [
    ]
  },
  "traceType": "PARTIAL",
  "message": "Trace exceeds maximun size of 10 bytes, a partial trace is returned"
}
```
**Checklist**
- [X] Tests updated
- [ ] Documentation added


pkg/model/trace/combine.go
```diff
@@ -90,9 +93,17 @@ func (c *Combiner) ConsumeWithFinal(tr *tempopb.Trace, final bool) (int, error)
 				}
 			}
 		}
-		return spanCount, c.sizeError()
+		maxSizeErr := c.sizeError()
+		if maxSizeErr != nil && c.allowPartialTrace {
```
mdisibio, line 97: nit: The logic here seems a little repetitive.  I might move it into `sizeError` and change the signature to `check() (done bool, error)`.  It could check allowPartialTrace, and the cached maxTraceSizeReached bool, etc.

```
if done, err := c.check(); done {
   return spanCount, err
}
```

mdisibio (approved):
Overall this looks good.   Debating on if it we should continue the FindTraceByIDV2 path all the way to the `Finder` interface and fork a new Combiner, instead of extending v1 with the bool.   I think as we add smarter logic to the combiner, we will ultimately want that.  But this looks great for now and going ahead and approving.

Left one nit, non-blocking, but something to consider.

joe-elliott (commented):
a good start, but it needs a way to signal to the caller that a partial trace is returned. this is going to be tough b/c of how many places we truncate the trace based on size.

perhaps we need a sentinel error that signals the trace is truncated but otherwise valid? i'm open to other solutions.

javiermolinar:
> a good start, but it needs a way to signal to the caller that a partial trace is returned. this is going to be tough b/c of how many places we truncate the trace based on size.
> 
> perhaps we need a sentinel error that signals the trace is truncated but otherwise valid? i'm open to other solutions.

Do you have an example where this could be problematic? My rationale was that instead of returning an error, stop concating more traces. For the client, we will add a message in the TraceByIDResponse

pkg/model/trace/combine.go
javiermolinar: Makes sense to me

joe-elliott:
> My rationale was that instead of returning an error, stop concating more traces. For the client, we will add a message in the TraceByIDResponse

This is what I mean. I was thinking about an error as a way to internally signal that the trace had been partially dropped, but returning a 200 to the client with a message.

For instance how do we bubble up that a trace was truncated here:

https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet4/block_findtracebyid.go#L266-L271

Its possible the right choice is to reduce where we are checking the max trace size? 

And then how are communicating from the querier -> query-frontend that a trace is partial?

pkg/tempopb/tempo.proto
```diff
@@ -52,11 +52,14 @@ message TraceByIDRequest {
...
   string blockEnd = 3;
   string queryMode = 5;
+  bool allowPartialTrace = 6;
 }
 
 message TraceByIDResponse {
   Trace trace = 1;
   TraceByIDMetrics metrics = 2;
+  bool partialTrace = 3;
+  string warning = 4;
```
joe-elliott, line 62:
```
  bool partialTrace = 3;
  string warning = 4;
```

Perhaps an enum instead of a bool in case we want to add more functionality here in the future?

nit: `string message` instead of `string warning` ?

pkg/tempopb/tempo.proto
```diff
@@ -52,6 +52,7 @@ message TraceByIDRequest {
   string blockStart = 2;
   string blockEnd = 3;
   string queryMode = 5;
+  bool allowPartialTrace = 6;
```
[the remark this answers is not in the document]
joe-elliott: This is to communicate between the frontend and the queriers that a partial trace is acceptable. i.e. is this the v1 or v2 traces endpoint.

pkg/tempopb/tempo.proto
javiermolinar: Seems reasonable to me

pkg/tempopb/tempo.proto
```diff
@@ -52,11 +52,18 @@ message TraceByIDRequest {
...
 }
 
 message TraceByIDResponse {
   Trace trace = 1;
   TraceByIDMetrics metrics = 2;
+  enum TraceType {
+    COMPLETE = 0;
+    PARTIAL = 1;
+  }
+  TraceType traceType = 3;
```
joe-elliott, line 65: i'd like to not constrain this to only "type" b/c we may want to communicate more in the future through it. how about something very generic like "Status" and the enum:

```
 enum Status {
    None = 0;
    PartialTrace = 1;
  }
```

pkg/model/v1/segment_decoder.go
```diff
@@ -23,9 +23,9 @@ func (d *SegmentDecoder) PrepareForWrite(trace *tempopb.Trace, _, _ uint32) ([]b
 	return proto.Marshal(trace)
 }
 
-func (d *SegmentDecoder) PrepareForRead(segments [][]byte) (*tempopb.Trace, error) {
+func (d *SegmentDecoder) PrepareForRead(segments [][]byte, allowPartialTraces bool) (*tempopb.Trace, error) {
 	// each slice is a marshalled tempopb.Trace, unmarshal and combine
-	combiner := trace.NewCombiner(0)
+	combiner := trace.NewCombiner(0, allowPartialTraces)
```
joe-elliott, line 28: let's revert the changes to `PrepareForRead`. We shouldn't need any changes here. just hardcode `false` or `true`. It shouldn't matter since `0` means no limit.

modules/querier/querier.go
```diff
@@ -221,7 +221,7 @@ func (q *Querier) stopping(_ error) error {
 }
 
 // FindTraceByID implements tempopb.Querier.
-func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDRequest, timeStart int64, timeEnd int64) (*tempopb.TraceByIDResponse, error) {
+func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDRequest, timeStart int64, timeEnd int64, allowPartialTrace bool) (*tempopb.TraceByIDResponse, error) {
```
joe-elliott, line 224: the querier checks both backend blocks and queries the ingesters. as you've found there are multiple layers where we fail the query if the trace exceeds max size. we need to find a way to bubble up from those places that this is occurred so the frontend knows to communicate back to the caller that this is a partial trace.

technically it should be impossible for a trace to exceed the max size in one block. this should only occur when recombining and higher levels. perhaps we drop that code that tests this?

pkg/model/v1/segment_decoder.go
javiermolinar: Ahh that's why my comment here: https://github.com/grafana/tempo/blob/26721832e2b1320de39183338d0d58d9da4083a6/pkg/model/trace/combine.go#L152

From the tests, I wasn't sure of the purpose of that, maybe we should document it or adding something more explicit like checkSize bool

modules/querier/querier.go
javiermolinar: > the querier checks both backend blocks and queries the ingesters. as you've found there are multiple layers where we fail the query if the trace exceeds max size. we need to find a way to bubble up from those places that this is occurred so the frontend knows to communicate back to the caller that this is a partial trace.

We could use the TraceById Response for that. So we can wrap the trace there.

> technically it should be impossible for a trace to exceed the max size in one block. this should only occur when recombining and higher levels. perhaps we drop that code that tests this?

I would love to know why is that 🙇. I always favor deleting code :)

pkg/tempopb/tempo.proto
javiermolinar: The only problem I see with this approach is when we get a combination of status, PartialTrace and smth else. We can end doing bitmasking :P

modules/querier/querier.go
javiermolinar: I know kung fu now :P

tempodb/encoding/vparquet4/wal_block.go
```diff
@@ -546,7 +546,7 @@ func (b *walBlock) FindTraceByID(ctx context.Context, id common.ID, opts common.
 		}
 	}
 
-	combiner := trace.NewCombiner(opts.MaxBytes)
+	combiner := trace.NewCombiner(opts.MaxBytes, false)
```
javiermolinar, line 549: I have left this to false. Even if we know it shouldn't ever happen I think it is a good sentinel

modules/querier/querier.go
```diff
@@ -309,11 +310,17 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
-	return &tempopb.TraceByIDResponse{
+	resp := &tempopb.TraceByIDResponse{
 		Trace:   completeTrace,
 		Metrics: &tempopb.TraceByIDMetrics{},
-	}, nil
+	}
+
+	if combiner.IsPartialTrace() {
+		resp.Status = tempopb.TraceByIDResponse_PARTIAL
+		resp.Message = fmt.Sprintf("Trace exceeds maximun size of %d bytes, a partial trace is returned", maxBytes)
```
joe-elliott, line 320: maximun -> maximum

pkg/model/trace/combine.go
```diff
@@ -129,19 +142,26 @@ func (c *Combiner) ConsumeWithFinal(tr *tempopb.Trace, final bool) (int, error)
...
+	moveForward, maxSizeErr := c.canContinue()
+	if moveForward {
+		return spanCount, nil
+	}
+
+	return spanCount, maxSizeErr
 }
 
-func (c *Combiner) sizeError() error {
+func (c *Combiner) canContinue() (bool, error) {
```
joe-elliott, line 153: i don't understand this "move forward" bool. it seems like this function signature should stay the same.

pkg/model/trace/combine.go
```diff
@@ -90,9 +95,17 @@ func (c *Combiner) ConsumeWithFinal(tr *tempopb.Trace, final bool) (int, error)
...
 		}
-		return spanCount, c.sizeError()
+		moveForward, maxSizeErr := c.canContinue()
+		if moveForward {
+			return spanCount, nil
+		}
+		return spanCount, maxSizeErr
 	}
 
+	// Do not combine more spans for now
```
joe-elliott, line 105: i think this short circuit logic should go right after the if tr == nil check. it makes it clearer if these kinds of checks are all at the top of the func

modules/querier/querier.go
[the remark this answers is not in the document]
javiermolinar: Good point

pkg/model/trace/combine.go
javiermolinar: This is my answer to this: https://github.com/grafana/tempo/pull/3941#pullrequestreview-2224443581 
To be honest, both approaches are fine to me. Looks more like a stylistic thing

joe-elliott: Making `canContinue()` return a bool but keeping it in the combiner is not worth it. We need to either:

- Return the bool out of the combiner so the calling code can stop looping over traces to combine
- Restore it to just return an error for simplicity

modules/frontend/combiner/trace_by_id_v2.go
```diff
@@ -21,8 +23,13 @@ func NewTraceByIDV2(maxBytes int, marshalingFormat string) Combiner {
 			// dedupe duplicate span ids
 			deduper := newDeduper()
 			traceResult = deduper.dedupe(traceResult)
-
 			resp.Trace = traceResult
+
+			if combiner.IsPartialTrace() {
```
joe-elliott, line 28: [Below](https://github.com/grafana/tempo/pull/3941/files#diff-a1fec687cd87c7167aee355f831ebae6beb368066e3bfbf092ef0aaf5e216aa2R317-R320) we are communicating from the querier to the frontend that there is a partial trace, but we are not handling it here.

The rest of this PR is looking good. My final concern is that we should handle `partial.Status == tempopb.TraceByIDResponse_PARTIAL` at the top of this `combine` func.

javiermolinar: My rationale was that a partial trace will always have a size bigger than the maxBytes so it won't be needed more changes.

If we want to be more performance-wise we can inspect the status and don't do any kind of processing

pkg/model/trace/combine.go
```diff
@@ -58,12 +63,19 @@ func (c *Combiner) Consume(tr *tempopb.Trace) (int, error) {
 	return c.ConsumeWithFinal(tr, false)
 }
 
+// Consume the given partial trace, no further processing is needed
+func (c *Combiner) ConsumePartialTrace(tr *tempopb.Trace) (int, error) {
+	c.result = tr
```
javiermolinar, line 68: I guess we don't need to clone the struct here

modules/frontend/combiner/trace_by_id_v2.go
joe-elliott: It does seem odd to me to pass back "partial trace" from the queriers and not honor it here. We should still combine the results here exactly like we're doing. It's possible (although unlikely) to calculate a partial trace in a querier and not come to the same conclusion in the frontend. For completeness if a querier tells sends back "partial trace" we should propagate it back to the caller.

pkg/model/trace/combine.go
joe-elliott: this is just copying a pointer and so won't clone the struct. regardless, i don't think this func should exist. we are very close. let's sync offline to push this through. we're not seeing eye to eye on some details we need to be aligned on.

joe-elliott (approved):

PR #3944, opened 2024-08-06
Swap querier /api/search and remaining endpoints to proto

mdisibio:
**What this PR does**:
Follow up to #3745 and #3731 .  Swaps /api/search and almost all remaining endpoints (except metrics summary api), to use proto between the frontend and queriers.  Improves latency, cpu, and memory.   Now that this is universal for all communication between frontend and queriers, did the next step of cleanup and centralized in `prepareRequestForQueriers`. 

For a bit more context - this started out as troubleshooting performance and resource usage for some searches. When limit and spans per spanset are high, it reached a point where the json encoding/decoding was the bottleneck.  For example limit=1000 and spss=100, means up to 100K spans per job.  This could easily lead to 100MB+ json response sizes.

Fixes n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
 * [ENHANCEMENT] Implement arrays for traceql.Static with reused fields [#3827](https://github.com/grafana/tempo/pull/3827) (@stoewer)
 * [ENHANCEMENT] Tag value lookup use protobuf internally for improved latency [#3731](https://github.com/grafana/tempo/pull/3731) (@mdisibio)
 * [ENHANCEMENT] TraceQL metrics queries use protobuf internally for improved latency [#3745](https://github.com/grafana/tempo/pull/3745) (@mdisibio)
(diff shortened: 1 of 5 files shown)


javiermolinar (commented):
Looks good, should we keep this?

https://github.com/grafana/tempo/blob/8cac9f5fb8747176faaa10c5479e108a9a5b69fa/modules/querier/http.go#L453

If none of this endpoints are going to return a json response maybe we can delete it

joe-elliott:
> If none of these endpoints are going to return a json response maybe we can delete it

i'd like to keep it so that curling the queriers directly will return json. there may be other good/bad reasons for it to be around

mdisibio:
> i'd like to keep it so that curling the queriers directly will return json. there may be other good/bad reasons for it to be around

+1 Agree good for troubleshooting.  Occasionally I use it to replay a single job shard against a querier.

PR #3961, opened 2024-08-13
Clone the diff in GRPC streaming to prevent panics

joe-elliott:
**What this PR does**:
Clones the diff in GRPC streaming to prevent panics

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


javiermolinar (commented):
Should not this mutex prevent that? the lock is adquired on diff and grpc_final: 

https://github.com/grafana/tempo/blob/f0b76c413ac2f66a149d6c3d47015a971bdb33dd/modules/frontend/combiner/common.go#L107

joe-elliott:
> Should not this mutex prevent that? the lock is adquired on diff and grpc_final:

The issue is that the pointers passed out of these functions are eventually passed to the proto marshalling code. These same pointers are also held inside the combiner and modified. The issue occurred when one of the objects was modified as it was being marshalled (we think).

ie-pham (approved):

PR #3963, opened 2024-08-14
feat: frontend middleware to block matching urls

### LINKED ISSUE #3769: Create a configurable list of TraceQL queries that are immediately 400'ed

javiermolinar:
**What this PR does**:
It provides a mechanism to refuse requests based on a regex blacklist. A new middleware was created and executed the first in the query pipeline. Any request that matches will return a 400 status code response.

Due to the way we build the pipeline I decided to make the new middleware as an AsyncMiddleware, that way it can be executed first in the chain. We should explore the possibility of having just one type of middleware so that they can run in any order regardless of their work. 

Fixes #3769

**Checklist**
- [X] Tests updated
- [X] Documentation added


modules/frontend/pipeline/black_list_middleware.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func NewURLBlackListWare(blackList []string, logger log.Logger) AsyncMiddleware[combiner.PipelineResponse] {
+	compiledBlacklist := make([]*regexp.Regexp, 0)
+	for _, v := range blackList {
+		r, err := regexp.Compile(v)
+		if err == nil {
+			compiledBlacklist = append(compiledBlacklist, r)
+		} else {
+			level.Warn(logger).Log("msg", "error compiling query frontend blacklist regex", "error", err)
```
joe-elliott, line 24: this should prevent the pod from starting. panic or return an error?

docs/sources/tempo/configuration/_index.md
```diff
@@ -561,6 +561,9 @@ query_frontend:
     # (default: 0)
     [api_timeout: <duration>]
 
+    # A list of regular expressions for refusing matching requests, these will apply for every request regardless of the endpoint.
+    [url_black_list: <list of strings> | default = <empty list>]]
```
joe-elliott, line 565: let's change all the wording to "deny list"

javiermolinar: ah true :P

modules/frontend/pipeline/deny_list_middleware.go
```diff
@@ -0,0 +1,52 @@
...
+		return &urlDenylistWare{
+			next:     next,
+			denyList: compiledDenylist,
+		}
+	})
+}
+
+func (c urlDenylistWare) RoundTrip(req Request) (Responses[combiner.PipelineResponse], error) {
+	if len(c.denyList) != 0 {
+		err := c.validateRequest(req.HTTPRequest().URL.Path)
```
joe-elliott, line 36: i don't believe path includes the url parameters. we need to test against something that includes these params so we can block specific traceql queries

javiermolinar: right, I thought this was the whole URL

modules/frontend/pipeline/deny_list_middleware.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+	}
+
+	return c.next.RoundTrip(req)
+}
+
+func (c urlDenylistWare) validateRequest(url string) error {
+	for _, v := range c.denyList {
+		if v.MatchString(url) {
+			return fmt.Errorf("Invalid request %s, URL is blacklisted", url)
```
joe-elliott, line 48: This will likely show up in Grafana, let's try:

"Invalid request %s. This query has been identified as one that destabilizes our system. Contact your system administrator for more information"

joe-elliott (approved):

Issue #3957, opened 2024-08-13
pusher failed to consume trace data due to "context canceled"


dastrobu:
**Is your feature request related to a problem? Please describe.**

We encountered the log message `msg="pusher failed to consume trace data" err="context canceled"` and had significant difficulty identifying the cause. The root cause was not related to the reasons documented in the [troubleshooting guide](https://github.com/grafana/tempo/blob/53db2730a9e6d1176641650c9412249d2bebf543/docs/sources/tempo/troubleshooting/max-trace-limit-reached.md).

The metric `tempo_discarded_spans_total` indicated that spans were being discarded, but the logs did not clarify the reason.

We ended up patching Grafana Tempo and even the Go SDK to gather more information about the discarded spans. I would like to propose improvements to observability in this area.

Here is what we did:

1. **Additional Logging in Distributor**

   We added extra logging to [distributor.go](https://github.com/grafana/tempo/blob/v2.5.0/modules/distributor/distributor.go#L362) to identify which spans were causing issues:

   ```go
   if err != nil {
       logSpans(batches, &LogReceivedSpansConfig{ // TODO should be LogDiscardedSpansConfig
           Enabled:              true,
           IncludeAllAttributes: true,
           FilterByStatusError:  false,
       }, d.logger)
       return nil, err
   }
   ```

   I propose adding a `LogDiscardedSpansConfig` in addition to `LogReceivedSpansConfig` to log discarded spans that are causing issues. I can open a PR for this if you think it is a good idea.

2. **Context Cancellation Debugging**

   It was extremely challenging to determine why the `Context` was canceled. This is a weak spot in Go, in my opinion. We had to patch the Go SDK to get more information about context cancellation.

   We added stack traces to [context.go](https://github.com/golang/go/blob/master/src/context/context.go#L536):

   ```go
   func (c *cancelCtx) cancel(removeFromParent bool, err, cause error) {
   	if err == nil {
   		panic("context: internal error: missing cancel error")
   	}
   	if err == Canceled {
   		stack := debug.Stack()
   		s := string(stack)
   		err = fmt.Errorf("%w stack: %v", err, s)
   	}
   ```

   This led to the following logs:

   ```txt
   level=error ts=2024-06-20T06:05:40.085835663Z caller=rate_limited_logger.go:27 msg="pusher failed to consume trace data" err="context canceled stack:
   goroutine 869 [running]:
   runtime/debug.Stack()
[...]

PR #3964, opened 2024-08-14
add a LogDiscardedSpansConfig to the distributor to log discarded spa…

dastrobu:
…ns causing issues.

```yaml
distributor:
  config:
    log_discarded_spans:
      enabled: true
    log_received_spans:
      # ...

```

**What this PR does**:

Relates to #3957.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/cmd/tempo/app/config.go
+++ b/cmd/tempo/app/config.go
@@ -166,6 +166,10 @@ func (c *Config) CheckConfig() []ConfigWarning {
 		warnings = append(warnings, warnLogReceivedTraces)
 	}
 
+	if c.Distributor.LogDiscardedSpans.Enabled {
+		warnings = append(warnings, warnLogDiscardedTraces)
+	}
+
 	if c.StorageConfig.Trace.Backend == backend.Local && c.Target != SingleBinary {
 		warnings = append(warnings, warnStorageTraceBackendLocal)
 	}
@@ -266,7 +270,10 @@ var (
 		Explain: fmt.Sprintf("default=%d", tempodb.DefaultBlocklistPollConcurrency),
 	}
 	warnLogReceivedTraces = ConfigWarning{
-		Message: "Span logging is enabled. This is for debuging only and not recommended for production deployments.",
+		Message: "Span logging is enabled. This is for debugging only and not recommended for production deployments.",
+	}
+	warnLogDiscardedTraces = ConfigWarning{
+		Message: "Span logging for discarded traces is enabled. This is for debugging only and not recommended for production deployments.",
 	}
 	warnStorageTraceBackendLocal = ConfigWarning{
 		Message: "Local backend will not correctly retrieve traces with a distributed deployment unless all components have access to the same disk. You should probably be using object storage as a backend.",
--- a/docs/sources/tempo/configuration/_index.md
+++ b/docs/sources/tempo/configuration/_index.md
@@ -196,9 +196,16 @@ distributor:
 
 
     # Optional.
-    # Enable to log every received span to help debug ingestion or calculate span error distributions using the logs
+    # Enable to log every received span to help debug ingestion or calculate span error distributions using the logs.
     # This is not recommended for production environments
     log_received_spans:
+        [enabled: <boolean> | default = false]
+        [include_all_attributes: <boolean> | default = false]
+        [filter_by_status_error: <boolean> | default = false]
+      
+    # Optional.
+    # Enable to log every discarded span to help debug ingestion or calculate span error distributions using the logs.
+    log_discarded_spans:
         [enabled: <boolean> | default = false]
         [include_all_attributes: <boolean> | default = false]
         [filter_by_status_error: <boolean> | default = false]
--- a/docs/sources/tempo/troubleshooting/max-trace-limit-reached.md
+++ b/docs/sources/tempo/troubleshooting/max-trace-limit-reached.md
@@ -10,6 +10,23 @@ aliases:
 
 The two most likely causes of refused spans are unhealthy ingesters or trace limits being exceeded.
 
+To log spans that are discarded, add the `--distributor.log_discarded_spans.enabled` flag to the distributor or 
+adjust the [distributor config]({{< relref "../configuration#distributor" >}}):
+
+```yaml
+distributor:
+  log_discarded_spans:
+    enabled: true
+    include_all_attributes: false # set to true for more verbose logs
+```
+
+Adding the flag logs all discarded spans, as shown below:
+
+```
+level=info ts=2024-08-19T16:06:25.880684385Z caller=distributor.go:767 msg=discarded spanid=c2ebe710d2e2ce7a traceid=bd63605778e3dbe935b05e6afd291006
+level=info ts=2024-08-19T16:06:25.881169385Z caller=distributor.go:767 msg=discarded spanid=5352b0cb176679c8 traceid=ba41cae5089c9284e18bca08fbf10ca2
+```
+
 ## Unhealthy ingesters
 
 Unhealthy ingesters can be caused by failing OOMs or scale down events.
@@ -41,3 +58,17 @@ tempo_discarded_spans_total
 ```
 
 In this case, use available configuration options to [increase limits]({{< relref "../configuration#ingestion-limits" >}}).
+
+## Client resets connection
+
+When the client resets the connection before the distributor can consume the trace data, you see logs like this:
+
+```
+msg="pusher failed to consume trace data" err="context canceled"
+```
+
+This issue needs to be fixed on the client side. To inspect which clients are causing the issue, logging discarded spans
+with `include_all_attributes: true` may help. 
+
+Note that there may be other reasons for a closed context as well. Identifying the reason for a closed context is 
+not straightforward and may require additional debugging.
--- a/modules/distributor/distributor.go
+++ b/modules/distributor/distributor.go
@@ -348,9 +348,7 @@ func (d *Distributor) PushTraces(ctx context.Context, traces ptrace.Traces) (*te
 
 	batches := trace.ResourceSpans
 
-	if d.cfg.LogReceivedSpans.Enabled {
-		logSpans(batches, &d.cfg.LogReceivedSpans, d.logger)
-	}
+	logReceivedSpans(batches, &d.cfg.LogReceivedSpans, d.logger)
 	if d.cfg.MetricReceivedSpans.Enabled {
 		metricSpans(batches, userID, &d.cfg.MetricReceivedSpans)
 	}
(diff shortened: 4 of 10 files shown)


dastrobu:
From my point of view, this PR is ready for review. I did some local testing in a kind cluster by setting `global_overrides.defaults.global.max_bytes_per_trace=10` (10 bytes to reject every span). Find a log sample below: 

```
level=info ts=2024-08-19T16:06:25.873820385Z caller=distributor.go:767 msg=received spanid=c2ebe710d2e2ce7a traceid=bd63605778e3dbe935b05e6afd291006
level=info ts=2024-08-19T16:06:25.874267385Z caller=distributor.go:767 msg=received spanid=5352b0cb176679c8 traceid=ba41cae5089c9284e18bca08fbf10ca2
level=info ts=2024-08-19T16:06:25.880684385Z caller=distributor.go:767 msg=discarded spanid=c2ebe710d2e2ce7a traceid=bd63605778e3dbe935b05e6afd291006
level=info ts=2024-08-19T16:06:25.881169385Z caller=distributor.go:767 msg=discarded spanid=5352b0cb176679c8 traceid=ba41cae5089c9284e18bca08fbf10ca2
level=info ts=2024-08-19T16:06:40.894952385Z caller=distributor.go:767 msg=received spanid=1323c400fc0ce680 traceid=ec000a65e060b907fa7ab2456a962e48
level=info ts=2024-08-19T16:06:40.895498385Z caller=distributor.go:767 msg=received spanid=0e9e7cd70829db54 traceid=2740bbe5cf599094d8f11d89aea5654b
level=info ts=2024-08-19T16:06:40.895587385Z caller=distributor.go:767 msg=received spanid=6065f5079da643ac traceid=c4ac3ede4c3b27483800caddaecdc4b4
level=info ts=2024-08-19T16:06:40.895644385Z caller=distributor.go:767 msg=received spanid=5d63bced0e52e8a4 traceid=30aa77762683b85b395cacc8f9c23711
level=info ts=2024-08-19T16:06:40.895665385Z caller=distributor.go:767 msg=received spanid=2fc41741cc426d9c traceid=8ddd7b725bcc1fe67010b0afe997507a
level=info ts=2024-08-19T16:06:40.902021385Z caller=distributor.go:767 msg=discarded spanid=5d63bced0e52e8a4 traceid=30aa77762683b85b395cacc8f9c23711
level=info ts=2024-08-19T16:06:40.902141385Z caller=distributor.go:767 msg=discarded spanid=1323c400fc0ce680 traceid=ec000a65e060b907fa7ab2456a962e48
level=info ts=2024-08-19T16:06:40.902202385Z caller=distributor.go:767 msg=discarded spanid=6065f5079da643ac traceid=c4ac3ede4c3b27483800caddaecdc4b4
level=info ts=2024-08-19T16:06:40.902232385Z caller=distributor.go:767 msg=discarded spanid=0e9e7cd70829db54 traceid=2740bbe5cf599094d8f11d89aea5654b
level=info ts=2024-08-19T16:06:40.902250385Z caller=distributor.go:767 msg=discarded spanid=2fc41741cc426d9c traceid=8ddd7b725bcc1fe67010b0afe997507a
```

I am not sure if you want a warning to be logged on startup for `log_received_spans.enabled`. If so, I can update the PR accordingly. Basically adding:
```go
	warnLogDiscardedTraces = ConfigWarning{
		Message: "Span logging for discarded traces is enabled. This is for debugging only and not recommended for production deployments.",
	}
```
Given that `log_discarded_spans` should probably not log anything for a perfect setup, I am not sure if we should warn about enabling this in production. 

Furthermore, to make the adjustments in `microservices-tempo-values.yaml` work, a follow-up PR on the helm chart is required. 

In order to keep the code dry, I had to refactor the logging logic a bit to share logic between received and discarded span logging.

knylander-grafana:
Greetings! You have the documentation updated checked, but I don't see any docs in the PR. Are there things that have user-impact that we should document? For example, config options that need to be updated, CLI commands, API calls, etc. If we do need docs, we can do those updates with your PR or open a new doc issue to track the required updates.

dastrobu:
> Are there things that have user-impact that we should document?

Thanks for the hint. I updated the config docs and the troubleshooting guide. Let me know if there is any other place in the docs I might have missed.  

Regarding the failing TestSearchUsingBackendTagsService E2E test, it appears to me that this is unrelated to my PR. I did run this test locally from the main branch and it also failed. I also looked for a recent [CI run from the main branch](https://github.com/grafana/tempo/actions/workflows/ci.yml?query=branch%3Amain) but did not find any.  
Do you have a hint on what to do about the failing test?

modules/distributor/distributor.go
```diff
@@ -575,16 +584,32 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount int
 	return keys, traces, nil
 }
 
-func countDiscaredSpans(numSuccessByTraceIndex []int, lastErrorReasonByTraceIndex []tempopb.PushErrorReason, traces []*rebatchedTrace, repFactor int) (maxLiveDiscardedCount, traceTooLargeDiscardedCount, unknownErrorCount int) {
+func discardedTraces(numSuccessByTraceIndex []int, lastErrorReasonByTraceIndex []tempopb.PushErrorReason, traces []*rebatchedTrace, repFactor int) []*discardedTrace {
```
joe-elliott, line 587: so i like how you tried to organize this, but we can not incur the allocations of these discardedTraces structs on our write path.

even tho it may be ugly, we're just going to have to work the logging of the dropped spans into here directly:

https://github.com/grafana/tempo/blob/main/modules/distributor/distributor.go#L584

docs/sources/tempo/troubleshooting/max-trace-limit-reached.md
```diff
@@ -41,3 +58,17 @@ tempo_discarded_spans_total
 ```
 
 In this case, use available configuration options to [increase limits]({{< relref "../configuration#ingestion-limits" >}}).
+
+## Client Resets Connection
+
+When the client resets the connection before the distributor can consume the trace data, you will see logs like this:
```
knylander-grafana, line 64:
```suggestion
When the client resets the connection before the distributor can consume the trace data, you see logs like this:
```

docs/sources/tempo/troubleshooting/max-trace-limit-reached.md
```diff
@@ -10,6 +10,23 @@ aliases:
...
+adjust the [distributor config]({{< relref "../configuration#distributor" >}}):
+
+```yaml
+distributor:
+  log_discarded_spans:
+    enabled: true
+    include_all_attributes: false # set to true for more verbose logs
+```
+
+This will log all discarded spans, as shown below:
```
knylander-grafana, line 23: Clarifying the wording. Is this correct? 
```suggestion
Adding the flag logs all discarded spans, as shown below:
```

docs/sources/tempo/configuration/_index.md
```diff
@@ -202,6 +202,13 @@ distributor:
         [enabled: <boolean> | default = false]
         [include_all_attributes: <boolean> | default = false]
         [filter_by_status_error: <boolean> | default = false]
+      
+    # Optional.
+    # Enable to log every discarded span to help debug ingestion or calculate span error distributions using the logs
```
knylander-grafana, line 207:
```suggestion
    # Enable to log every discarded span to help debug ingestion or calculate span error distributions using the logs.
```

knylander-grafana (commented):
Thank you for adding docs! I've reviewed the docs section and they look good.

cmd/tempo/app/config.go
```diff
@@ -266,7 +266,7 @@ var (
 		Explain: fmt.Sprintf("default=%d", tempodb.DefaultBlocklistPollConcurrency),
 	}
 	warnLogReceivedTraces = ConfigWarning{
-		Message: "Span logging is enabled. This is for debuging only and not recommended for production deployments.",
+		Message: "Span logging is enabled. This is for debugging only and not recommended for production deployments.",
```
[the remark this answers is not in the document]
dastrobu: I was thinking about this as well, see https://github.com/grafana/tempo/pull/3964#issuecomment-2296953852, I had no clear opinion, though. I can add the warning if desired. 

> Given that log_discarded_spans should probably not log anything for a perfect setup, I am not sure if we should warn about enabling this in production.

modules/distributor/distributor.go
dastrobu: Could you explain your concern a bit more? Is it about the memory footprint? `unsafe.Sizeof(discardedTrace{})` is just 16 bytes (one pointer and one enum). So, I guess it would be barely measurable, as these are just allocated for discarded spans. On the normal write path, there is no additional allocation when no spans are discarded. Even if all spans are discarded, the footprint remains low from my perspective. Or is it a different concern?

joe-elliott: sure, it's not so much about the number of bytes allocated. i'm more concerned with the allocations of the slice and the individual struct pointers. creating this many objects on the heap would put undue pressure on the GC.

in a situation where the ingesters were struggling and started dropping more spans this would create a potential cascading failure where the distributors started creating a large number of objects, falling behind on heap cleanup and OOMing.

since this is a "debug" option i expect additional work/allocations when its turned on, but would prefer it have no impact when turned off.

dastrobu: I have to admit that I am not fully convinced that allocating or resizing a slice (sub millisecond time) is something that would hav any measurable impact compared to any network request (at least milliseconds). 
However, I went thought the code again and I think I found a good solution now, which shouldn't have any runtime impact and still is not too ugly to contribute it. 

I also adjusted the trace logging a bit (added `push_error_reason` and `tenant`), which might be helpful. Adding the tenant could also be something to consider for `log_received_spans`. 

Please do a fresh review on these changes.

cmd/tempo/app/config.go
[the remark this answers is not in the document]
dastrobu: based on the discussions about performance below, I added the warning for discarded spans as well.

modules/distributor/distributor.go
```diff
@@ -581,12 +578,21 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount int
 	return keys, traces, nil
 }
 
-func countDiscaredSpans(numSuccessByTraceIndex []int, lastErrorReasonByTraceIndex []tempopb.PushErrorReason, traces []*rebatchedTrace, repFactor int) (maxLiveDiscardedCount, traceTooLargeDiscardedCount, unknownErrorCount int) {
+// discardedPredicate determines if a trace is discarded based on the number of successful replications.
+type discardedPredicate func(int) bool
+
+func newDiscardedPredicate(repFactor int) discardedPredicate {
```
joe-elliott, line 584: i'm curious why you chose this pattern. it's not a blocker, but i suppose i would have just written

```
func wasDiscarded(numSuccess int, repFactor int) bool {
  // the stuff
}
```

dastrobu: It felt most natural to me not to pull the computation of `quorum` into the loop. Plus, having functions with the signature `(int, int)` is always a bit harder to read, as you need to inspect the parameter names more carefully. But it is certainly a matter of taste. Your solution is less code on the plus side. I can change it, if you prefer less code.

joe-elliott: i almost always prefer less code :)

but i also like different opinions, thoughts and patterns. leave it. it reminds me of function currying from my old js days.

joe-elliott (approved):
thank you for being flexible on your approach 🙏 

an excellent original analysis and solid contribution to our codebase that is appreciated.

PR #3967, opened 2024-08-14
[TraceQL] Add support for querying by instrumentation scope

ie-pham:
**What this PR does**: Add TraceQL to search by instrumentation scope name, version, and attributes. 
Screenshot 2024-08-20 at 4 14 52 PM

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


joe-elliott:
cc @knylander-grafana possibly merged before 2.6 and would be a big feature

docs/sources/tempo/traceql/_index.md
```diff
@@ -94,6 +94,8 @@ The following table shows the current available scoped intrinsic fields:
 | `event:timeSinceStart`  | duration    | time of event in relation to the span start time                | `{ event:timeSinceStart > 2ms}`        |
 | `link:spanID`           | string      | link span id using hex string                                   | `{ link:spanID = "0000000000000001" }` |
 | `link:traceID`          | string      | link trace id using hex string                                  | `{ link:traceID = "1234567890abcde" }` |
+| `scope:name`            | string      | instrumentation scope name                                      | `{ scope:name = "grpc" }`              |
```
jcarres-mdsol, line 97: I thought the thing at the left of `:` was a scope , i.e. `link` is a possible value for scope. 
So is very confusing for me that this is called `scope` , shouldn't it be called `instrumentation` ?
`instrumentation:name` `instrumentation:version` more clearly define what this points at IMHO

joe-elliott: this is a good call out. we have discussed this with our internal team and they believe that otel does not see this as an instrumentation only field. it just happens to be used that way and they intend for it to be used as a more generic "scope". 

agree that scope scope is strange. let me summon @jpkrohling. he may have more insight.

jpkrohling: An instrumentation scope can have a name and a version, in addition to arbitrary key/value pairs, so I think this is valid as is: `{ scope:name = "grpc" }`.

Reference: https://github.com/open-telemetry/opentelemetry-proto/blob/40b3c1b746767cbc13c2e39da3eaf1a23e54ffdd/opentelemetry/proto/common/v1/common.proto#L69-L81

jpkrohling: Are end-users exposed to "scope scope"? As an OTel user, I immediately thought of [InstrumentationScope](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.3.2/opentelemetry/proto/common/v1/common.proto#L69-L81) when I saw "scope:name" here. Unless there's an explicit source of confusion, I'd keep "scope:name".

joe-elliott: > Are end-users exposed to "scope scope"?

Not directly although there may be some awkward wording in the docs.

jcarres-mdsol: Personally `scope:name` is a generic term that says nothing to me. 
I can imagine for people better versed in opentelemetry parlance it is very obvious. 
I am very sure that around here I will need to tell people that "scope:name means the name of the instrumentation/area that created the span". or such explanation as I do not think it is clear to more casual users of observability tools.

knylander-grafana: Would it help to have these differences called out in the doc to avoid confusion? Or would that just create more confusion?

jcarres-mdsol: Docs will definitely need to be there.
When someone reads a query or an autocomplete with `span:duration` , that person needs to know almost nothing about tracing to understand what that would mean, without resorting to reading documentation.

I think when someone reads `scope:name` that person will definitely need to go to documentation to know what would that mean.

docs/sources/tempo/traceql/_index.md
```diff
@@ -94,6 +94,8 @@ The following table shows the current available scoped intrinsic fields:
...
+| `scope:version`         | string      | instrumentation scope version                                   | `{ scope:version = "1.0.0.0" }`        |
```
atsu85, line 98: Was one `.0` accidentally added?

```suggestion
| `scope:version`         | string      | instrumentation scope version                                   | `{ scope:version = "1.0.0" }`        |   
```

docs/sources/tempo/traceql/_index.md
ie-pham: @jcarres-mdsol thank you for pointing this out. I've changed the syntax to `instrumentation:name` and `instrumentation:version` 🚀

docs/sources/tempo/traceql/_index.md
ie-pham: Nice catch!

docs/sources/tempo/traceql/_index.md
```diff
@@ -111,7 +113,7 @@ This example searches all Kubernetes clusters called `service-name` that have a
 
 ### Attribute fields
 
-TraceQL has four different attribute scopes: span attributes, resource attributes, event attributes, and link attributes. By expanding a span in the Grafana UI, you can see both its span attributes (1 in the screenshot) and resource attributes (2 in the screenshot).
+TraceQL has five different attribute scopes: span attributes, resource attributes, event attributes, link attributes, instrumentation scope attributes. By expanding a span in the Grafana UI, you can see both its span attributes (1 in the screenshot) and resource attributes (2 in the screenshot).
```
knylander-grafana, line 116: Missing and in the attribute list. 
```suggestion
TraceQL has five different attribute scopes: span attributes, resource attributes, event attributes, link attributes, and instrumentation scope attributes.
By expanding a span in the Grafana UI, you can see both its span attributes (1 in the screenshot) and resource attributes (2 in the screenshot).
```

knylander-grafana (commented):
Thank you for updating the docs! Approving the docs only.

tempodb/encoding/vparquet4/block_traceql.go
```diff
@@ -898,12 +941,14 @@ const (
...
 	otherEntryEventKey   = "event"
 	otherEntryLinkKey    = "link"
+	otherEntryScopeKey   = "scope"
 
 	// a fake intrinsic scope at the trace lvl
 	intrinsicScopeTrace = -1
 	intrinsicScopeSpan  = -2
 	intrinsicScopeEvent = -3
 	intrinsicScopeLink  = -4
+	intrinsicScopeScope = -5
```
mdisibio, line 951: PR is looking good. Do we want to clean up the remaining references to scope scope?  Looks like there are a handful.

mdisibio (approved):
Running locally things are looking good, searching, autocomplete, and selectall via compare() (nice!).  Approving ✅

But did want to call out some unexpected behavior, that this PR is not causing, but unearthing.   If a span is sent with no instrumentation, we save name/value to parquet as empty strings, and they are maintained for reads.  Empty string is returned from autocomplete `/api/v2/search/tag/instrumentation:name/values` and `/api/v2/search/tag/instrumentation:version/values`, and the query `{instrumentation:name="" && instrumentation:version=""}` finds these spans. 

But I think we got lucky.  Although this behavior is unexpected, it's technically correct.  The otel proto [definition](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L71) is commented `Semantically when InstrumentationScope isn't set, it is equivalent with an empty instrumentation scope name (unknown)`.   Now it's sounding familiar and we likely considered this when designing vParquet1 schema to be non-nullable (which this inherited).  

Either way, I think we are good and just wanted to mention it for future reference.

PR #3971, opened 2024-08-16
fix: pass search options to the backend

javiermolinar:
**What this PR does**:
The search options from here: https://github.com/grafana/tempo/blob/cb35a075268deaf501ac4d25bbd58d687c1f9f35/modules/querier/querier.go#L892

was not passed downstream. 

**Checklist**


joe-elliott (approved):

joe-elliott:
nice find. this results in a lot of block overscan

PR #3975, opened 2024-08-19
feat: added min_over_time TraceQL metric

javiermolinar:
**What this PR does**:
It adds a new TraceQL metric `min_over_time` used to compute the minimum value of a span series over a specified date range.

**Checklist**
- [X] Tests updated
- [x] Documentation added


pkg/traceql/ast.go
```diff
@@ -1073,6 +1081,11 @@ func (a *MetricsAggregate) init(q *tempopb.QueryRangeRequest, mode AggregateMode
 		exemplarFn = func(s Span) (float64, uint64) {
 			return math.NaN(), a.spanStartTimeMs(s)
 		}
+	case metricsAggregateMinOverTime:
+		innerAgg = func() VectorAggregator { return NewMinOverTimeAggregator(a.attr) }
+		exemplarFn = func(s Span) (float64, uint64) {
```
javiermolinar, line 1086: I am not quite sure if the exemplar function is correct in this case

pkg/traceql/engine_metrics.go
```diff
@@ -338,6 +338,69 @@ func (c *CountOverTimeAggregator) Sample() float64 {
...
+}
+
+var _ VectorAggregator = (*MinOverTimeAggregator)(nil)
+
+func NewMinOverTimeAggregator(attr Attribute) *MinOverTimeAggregator {
+	var fn func(s Span) float64
+	switch attr {
+	case IntrinsicDurationAttribute:
+		fn = func(s Span) float64 {
+			return float64(s.DurationNanos()) / float64(time.Millisecond)
```
javiermolinar, line 357: Should we work in nanoseconds, ms, seconds? Doing something like min_over_time(duration, s), a min-max normalization?

pkg/traceql/engine_metrics.go
```diff
@@ -338,6 +338,69 @@ func (c *CountOverTimeAggregator) Sample() float64 {
...
+	if !c.firstTime {
+		c.min = val
+		c.firstTime = true
+	} else if val < c.min {
+		c.min = val
+	}
+}
+
+func (c *MinOverTimeAggregator) Sample() float64 {
+	return c.min * c.rateMult
```
mdisibio, line 381: I think we need to tweak the empty/uninitialized behavior.  In this case if we query min_over_time() and there are no matching spans, or none have the attribute, it will return zeroes.  I think not returning a sample for that time range is the most correct.  Let's chat about how to accomplish this.  This also relates to the implementation of floatizeAttribute below which doesn't indicate if the attribute was present.

pkg/traceql/ast.go
```diff
@@ -1032,7 +1040,7 @@ func (a *MetricsAggregate) extractConditions(request *FetchSpansRequest) {
 	switch a.op {
 	case metricsAggregateRate, metricsAggregateCountOverTime:
 		// No extra conditions, start time is already enough
-	case metricsAggregateQuantileOverTime, metricsAggregateHistogramOverTime:
+	case metricsAggregateQuantileOverTime, metricsAggregateHistogramOverTime, metricsAggregateMinOverTime:
 		if !request.HasAttribute(a.attr) {
```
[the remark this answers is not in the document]
javiermolinar: Yeah, makes sense

pkg/traceql/engine_metrics.go
```diff
@@ -338,6 +338,69 @@ func (c *CountOverTimeAggregator) Sample() float64 {
...
+	min             float64
+	rateMult        float64
+}
+
+var _ VectorAggregator = (*MinOverTimeAggregator)(nil)
+
+func NewMinOverTimeAggregator(attr Attribute) *MinOverTimeAggregator {
+	var fn func(s Span) float64
+	switch attr {
+	case IntrinsicDurationAttribute:
```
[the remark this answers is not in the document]
javiermolinar: I've moved it and simplified the floatize metric to a single place. We could move it into the very own Span interface but I guess is too much hassle

pkg/traceql/ast.go
mdisibio: It should use the real attribute value from the floatize function, so that the exemplar is plotted in the final graph like expected.  math.NaN is being used for rate and count_over_time because they don't come from an attribute, it's a placeholder and the value gets assigned later.

mdisibio (commented):
As discussed offline - we need to add a new aggregator for mode Sum/Final, which computes the min across time-series (i.e. the min of all mins), and use it in `MetricsAggregate.initSum` and `initFinal`.  Right now it is still falling back to SimpleAdditionAggregator.

pkg/traceql/ast.go
javiermolinar: Yes, it is. But the attr.Float() may return a NaN so we need to prevent the error

pkg/traceql/engine_metrics.go
```diff
@@ -1189,10 +1243,32 @@ func (b *SimpleAdditionAggregator) Combine(in []*tempopb.TimeSeries) {
...
+		if j >= 0 && j < len(existing.Values) {
+			existing.Values[j] += sample.Value
+		}
+	}
+}
+
+// Simple min aggregator. It calculates the minumun between existing values and a new sample, New indicates if the existing values were just initialized, ugly
+func simpleMinAggregator(new bool, start, end, step uint64, existing *TimeSeries, samples []tempopb.Sample) {
+	for _, sample := range samples {
+		j := IntervalOfMs(sample.TimestampMs, start, end, step)
```
javiermolinar, line 1263: I know this code is repeated. But I'm thinking on average where we will need to know the number of elements in the array

javiermolinar: $x =(m * x_a + n*x_b)/(m+n)$

javiermolinar:
Back to draft, we need to come out with a solution for missing/uninitialized time-series values

knylander-grafana:
Is this something we'll want to document?

javiermolinar:
> Is this something we'll want to document?

Yes, once we agreed on a solution we should document this new function

pkg/traceql/engine_metrics.go
```diff
@@ -20,6 +20,8 @@ const (
 	internalLabelBucket   = "__bucket"
 	maxExemplars          = 100
 	maxExemplarsPerBucket = 2
+	// NormalNaN is a quiet NaN. This is also math.NaN().
```
javiermolinar, line 23: I stole this  from prometheus: https://github.com/prometheus/prometheus/blob/849215d90cf48119bcc772603857fdc0b986ac4b/model/value/value.go#L22

knylander-grafana:
> > Is this something we'll want to document?
> 
> Yes, once we agreed on a solution we should document this new function

This may be a good spot for your content. The compare function content went on the traceql/metrics-queries.md file in the Functions section.

pkg/traceql/engine_metrics.go
```diff
@@ -239,7 +241,7 @@ func (set SeriesSet) ToProtoDiff(req *tempopb.QueryRangeRequest, rangeForLabels
 
 			// todo: this loop should be able to be restructured to directly pass over
 			// the desired intervals
-			if ts < start || ts > end {
+			if ts < start || ts > end || math.IsNaN(value) {
```
mdisibio, line 244: We need one more change which is to not output series which end up with no samples after filtering.  After this loop:

```go
		// Skip series with no samples after filtering
		if len(samples) == 0 {
			continue
		}
```

This handles cases like the following:

`{ } | min_over_time(span.foo)`  Currently returns a single series with no data points.

`{ } | min_over_time(span.http.status_code) by (span.http.status_code)` This is a contrived example, but it includes an empty `"nil"` series for the spans without http.status_code.

pkg/traceql/engine_metrics.go
```diff
@@ -1060,51 +1129,76 @@ func (b *SimpleAdditionAggregator) Combine(in []*tempopb.TimeSeries) {
...
+func (b *SimpleAggregator) Results() SeriesSet {
 	return b.ss
 }
 
+// Simple addition aggregator. It adds existing values with the new sample. New indicates if the existing values were just initialized, ugly
+func simpleAdditionAggregator(start, end, step uint64, existing *TimeSeries, samples []tempopb.Sample) {
+	for _, sample := range samples {
+		j := IntervalOfMs(sample.TimestampMs, start, end, step)
+		if j >= 0 && j < len(existing.Values) {
+			existing.Values[j] += sample.Value
```
mdisibio, line 1185: I kind of want to thin these out to just the 1 or 2 lines in question, so that the interval check/calculation isn't duplicated, and put it back in Combine.  Something like:

```go
func (existing, new float64) float64

add := func(...) { return existing + new }
min := func(...) { if math.IsNaN(existing) || ... { return new } else { return existing } }

// in Combine
for _, sample := range samples {
	j := IntervalOfMs(sample.TimestampMs, start, end, step)
	if j >= 0 && j < len(existing.Values) {
		existing.Values[j] = fn(existing.Values[j], sample.Value)
```

mdisibio (commented):
This is looking great.  I'm really happy with this PR because we are straightening up some ast and value handling quirks that needed it.  There is 1 more required change to drop series with no values, and then 1 small nit that I would prefer but not blocking.

pkg/traceql/engine_metrics.go
javiermolinar: Cool!

pkg/traceql/engine_metrics.go
javiermolinar: That's fine although I'm not sure it will work for the avg_over_time since we need to know the number of non-nan elements for the given interval to compute it

javiermolinar: but we can fine-tune it later

javiermolinar: It would be nice if Go had the  "=>" operator

add := (a float64, b float64) => a + b

like in c#

```cs
Func<string> greet = () => "Hello, World!";
Console.WriteLine(greet());
```

javiermolinar: infering the return type ...

javiermolinar:
> > > Is this something we'll want to document?
> > 
> > 
> > Yes, once we agreed on a solution we should document this new function
> 
> This may be a good spot for your content. The compare function content went on the traceql/metrics-queries.md file in the Functions section.

@knylander-grafana I've added some documentation. We should have a section for all the aggregation functions over_time. Like prometheus [does](https://prometheus.io/docs/prometheus/latest/querying/functions/#aggregation_over_time)

pkg/traceql/engine_metrics.go
mdisibio: Yes I was thinking about average as well, since it requires state.  That is true and a single func pointer isn't enough.

docs/sources/tempo/traceql/metrics-queries.md
```diff
@@ -91,6 +94,25 @@ down by HTTP route.
 This might let you determine that `/api/sad` had a higher rate of erroring
 spans than `/api/happy`, for example.
 
+### The `count_over_time` and `min_over_time` functions
+
+The `counts_over_time()` let you counts the number of matching spans per time interval.
```
mdisibio, line 99:
```suggestion
The `count_over_time()` let you counts the number of matching spans per time interval.
```

docs/sources/tempo/traceql/metrics-queries.md
```diff
@@ -91,6 +94,25 @@ down by HTTP route.
...
+
+```
+{ name = "GET /:endpoint" } | count_over_time() by (span.http.status_code)
+
+```
+The `min_over_time()` let you aggregate numerical values by computing the minimun value of them, such as the all important span duration.
```
mdisibio, line 105:
```suggestion
The `min_over_time()` let you aggregate numerical values by computing the minimum value of them, such as the all important span duration.
```

knylander-grafana:
If you have a PR with doc updates, add the type/docs label to it please.

knylander-grafana:
> @knylander-grafana I've added some documentation. Maybe is a good idea to have a section for all the aggregation functions over_time. Like prometheus [does](https://prometheus.io/docs/prometheus/latest/querying/functions/#aggregation_over_time)

Thank you for adding the doc! We have a [doc issue](https://github.com/grafana/tempo-squad/issues/413) to update the metrics query docs (they need some love). 

Before we restructure sections in the page, we should look at the content on the metrics query pages to consider who our target audience is, what are the topics that audience needs to know, what are they interested in doing (determine any tasks -- which helps us create meaningful examples), etc. 

The Prometheus example is a good way of just listing an item and defining what it does. It could be a good way to do an introduction to a section, and then provide either an example for every function (in which case you use the subheadings like we have) or your subheadings are focused on examples and use cases for the functions. Does that difference make sense? I'd love to see the latter (function list like Prom docs and then real-world examples and explanations), but it takes more time than providing the function heading, with a brief explanation, and then a simple example.

PR #3976, opened 2024-08-19
Cache: Read to a sized buffer

joe-elliott:
**What this PR does**:
Improves the memory performance of the caching middleware by dropping `io.ReadAll` in favor of reading to a sized buffer. 

Also
- moves the response read functionality into a shared area for future use
- removes one of our 2 mock caches

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/frontend/pipeline/sync_handler_cache.go
```diff
@@ -75,21 +76,27 @@ func (c cachingWare) RoundTrip(req Request) (*http.Response, error) {
 	}
 
 	if len(key) > 0 {
+		// don't bother caching if the response is too large
+		maxItemSize := c.cache.c.MaxItemSize()
+		if maxItemSize != 0 && resp.ContentLength > int64(maxItemSize) {
+			return resp, nil
+		}
+
+		buffer, err := api.ReadBodyToBuffer(resp)
```
electron0zero, line 85: if maxItemSize is set to 0, then we will always cache, no matter the size.

is my understanding correct?

electron0zero (approved):
minor comments, rest lgtm :shipit:

PR #3977, opened 2024-08-19
Add doc about long-running traces

zalegrala:
**What this PR does**:

Here we add a bit of doc to point folks at when they are using the long-running trace pattern.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/troubleshooting/long-running-traces.md
+++ b/docs/sources/tempo/troubleshooting/long-running-traces.md
@@ -36,7 +36,7 @@ can aid in understanding when this situation arises. In particular, the followin
 1 - sum(rate(tempo_warnings_total{reason="disconnected_trace_flushed_to_wal"}[5m])) / sum(rate(tempo_ingester_traces_created_total{}[5m]))
 ```
 
-Users of this pattern may also be interested in the
+If you have long-running traces, you may also be interested in the
 `rootless_trace_flushed_to_wal` reason to know when a trace is flushed to the
 wall without a root trace.
 
--- a/docs/sources/tempo/troubleshooting/long-running-traces.md
+++ b/docs/sources/tempo/troubleshooting/long-running-traces.md
@@ -9,7 +9,7 @@ aliases:
 # Long-running traces
 
 An issue arises in Tempo when a user exercises a usage pattern referred to as
-long-running traces. This happens is when Tempo spans receives spans for a
+long-running traces. This happens when Tempo receives spans for a
 trace, and then there is a delay, and then Tempo receives additional spans for
 the same trace. If the delay between spans is great enough, the spans end up in
 different blocks, which can lead to inconsistency in a few ways.
--- a/docs/sources/tempo/troubleshooting/long-running-traces.md
+++ b/docs/sources/tempo/troubleshooting/long-running-traces.md
@@ -19,8 +19,11 @@ different blocks, which can lead to inconsistency in a few ways:
    performing a TraceID lookup, Tempo searches for all parts of a trace in all
    matching blocks, which yields greater accuracy when combined.
 
-1. When using structural operators, the conditions may match on different
-   blocks, and so results can be confusing.
+1. When using [`spanset`
+   operators](https://grafana.com/docs/tempo/latest/traceql/#combining-spansets),
+   Tempo only evaluates the contiguous trace of the current block. This means
+   that for a single block the conditions may evaluate to false, but to
+   consider all parts of the trace from all blocks would evaluate true.
 
 You can tune the `ingester.trace_idle_period` configuration to allow for
 greater control about when traces are written to a block. Extending this beyond
@@ -40,7 +43,7 @@ If you have long-running traces, you may also be interested in the
 `rootless_trace_flushed_to_wal` reason to know when a trace is flushed to the
 wall without a root trace.
 
-You can use `reason` fields for discovery with this query: 
+You can use `reason` fields for discovery with this query:
 
 ```
 sum(rate(tempo_warnings_total{}[5m])) by (reason)
--- a/docs/sources/tempo/troubleshooting/long-running-traces.md
+++ b/docs/sources/tempo/troubleshooting/long-running-traces.md
@@ -56,5 +56,5 @@ be worth modifying the usage pattern slightly. For example, you may want to use
 [span
 links](https://opentelemetry.io/docs/concepts/signals/traces/#span-links), so
 that traces are split up, allowing one trace to complete, while pointing to the
-next trace in the causal chain . This would allow both traces to finish in a
+next trace in the causal chain . This allows both traces to finish in a
 shorter duration, and increase the chances of ending up in the same block.
--- a/docs/sources/tempo/troubleshooting/long-running-traces.md
+++ b/docs/sources/tempo/troubleshooting/long-running-traces.md
@@ -8,11 +8,10 @@ aliases:
 
 # Long-running traces
 
-An issue arises in Tempo when a user exercises a usage pattern referred to as
-long-running traces. This happens when Tempo receives spans for a trace,
+Long-running traces are created when Tempo receives spans for a trace,
 followed by a delay, and then Tempo receives additional spans for the same
 trace. If the delay between spans is great enough, the spans end up in
-different blocks, which can lead to inconsistency in a few ways.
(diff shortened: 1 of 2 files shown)


docs/sources/tempo/troubleshooting/long-running-traces.md
```diff
@@ -0,0 +1,24 @@
...
+the same trace. If the delay between spans is great enough, the spans end up in
+different blocks, which can lead to inconsistency in a few ways.
+
+1. When using TraceQL search, the duration information only pertains to the
+   information contained in a single block. This happens because Tempo consults
+   only the first matching block to determine this information. When searching a full
+   trace by ID, all blocks are searched for parts of this trace,
+   and so the information is more accurate.
+
+1. When using structural operators, the conditions may match on different
```
zalegrala, line 23: I'm not sure what else we want to say here about the structural operator case.

knylander-grafana:
@zalegrala Let me know when you want me to review. Thank you for adding doc!

zalegrala:
I'll take review any time @knylander-grafana.  Do you have thoughts?

docs/sources/tempo/troubleshooting/long-running-traces.md
```diff
@@ -0,0 +1,24 @@
...
+description: Troubleshoot search results when using long-running traces
+weight: 479
+aliases:
+  - ../operations/troubleshooting/long-running-traces/
+---
+
+# Long-running traces
+
+An issue arises in Tempo when a user exercises a usage pattern referred to as
+long-running traces. This happens is when Tempo spans receives spans for a
```
joe-elliott, line 12:
```suggestion
long-running traces. This happens when Tempo receives spans for a
```

docs/sources/tempo/troubleshooting/long-running-traces.md
```diff
@@ -0,0 +1,24 @@
...
+trace, and then there is a delay, and then Tempo receives additional spans for
+the same trace. If the delay between spans is great enough, the spans end up in
+different blocks, which can lead to inconsistency in a few ways.
+
+1. When using TraceQL search, the duration information only pertains to the
```
joe-elliott, line 17: here is our effort to combine metadata for the same trace. it is far from perfect, but it does attempt to consolidate across multiple traces.

https://github.com/grafana/tempo/blob/main/pkg/traceql/combine.go#L54

docs/sources/tempo/troubleshooting/long-running-traces.md
```diff
@@ -0,0 +1,24 @@
+---
```
joe-elliott, line 1: great start, i would mention that:

- the `trace_idle_period` plays an important role here and can allow operators to adjust for longer running trace pipelines.
- we provide some data accuracy metrics under `tempo_warnings_total`. one indicates if a disconnected trace is flushed to the wal, the other indicates if a rootless trace is flushed to the wal.

https://github.com/grafana/tempo/blob/main/pkg/dataquality/warnings.go

this is the formula we use in cloud to generate %age of connected traces flushed to wal:

```
1 - sum(rate(tempo_warnings_total{reason="disconnected_trace_flushed_to_wal"}[5m])) / sum(rate(tempo_ingester_traces_created_total{}[5m]))
```

zalegrala: This is handy to know, thanks for the pointers.

docs/sources/tempo/troubleshooting/long-running-traces.md
zalegrala: I think I've captured this in the update.  Instead of mentioning "a single block", I just mention that not all blocks are used to determine the resulting duration on the search, which I think is accurate.

docs/sources/tempo/troubleshooting/long-running-traces.md
joe-elliott: it's not really about structural operators. Its when you use any spanset operators:
```
{} && {}
{} || {}
{} >> {}
...
```
since we are only able to look at the contiguous trace to evaluate these its possible the query might be false when evaluating the trace shard in this block here, but the query is actually true if you were to look at every span for this trace across all blocks

docs/sources/tempo/troubleshooting/long-running-traces.md
```diff
@@ -0,0 +1,61 @@
...
+  - ../operations/troubleshooting/long-running-traces/
+---
+
+# Long-running traces
+
+An issue arises in Tempo when a user exercises a usage pattern referred to as
+long-running traces. This happens when Tempo receives spans for a trace,
+followed by a delay, and then Tempo receives additional spans for the same
+trace. If the delay between spans is great enough, the spans end up in
+different blocks, which can lead to inconsistency in a few ways.
```
knylander-grafana, line 15: Trying to rewrite to reduce passive voice. What do you think about this as a start? 
```suggestion
Long-running traces are created when Tempo receives spans for a trace,
followed by a delay, and then Tempo receives additional spans for the same
trace. If the delay between spans is great enough, the spans end up in
different blocks, which can lead to inconsistency in a few ways:
```

docs/sources/tempo/troubleshooting/long-running-traces.md
```diff
@@ -0,0 +1,61 @@
...
+1. When using TraceQL search, the duration information only pertains to a
+   subset of the blocks that contain a trace. This happens because Tempo
+   consults only enough blocks to know the TraceID of the matching spans. When
+   performing a TraceID lookup, Tempo searches for all parts of a trace in all
+   matching blocks, which yields greater accuracy when combined.
+
+1. When using structural operators, the conditions may match on different
+   blocks, and so results can be confusing.
+
+Users can tune the `ingester.trace_idle_period` configuration to allow for
```
knylander-grafana, line 26: Do we really mean "you" instead of "user"? 
```suggestion
You can tune the `ingester.trace_idle_period` configuration to allow for
```

docs/sources/tempo/troubleshooting/long-running-traces.md
```diff
@@ -0,0 +1,61 @@
...
+Tempo publishes a `tempo_warnings_total` metric from several components, which
+can aid in understanding when this situation arises. In particular, the following query can be used to know what percentage of traces which are flushed to the wall are connected.
+
+```
+1 - sum(rate(tempo_warnings_total{reason="disconnected_trace_flushed_to_wal"}[5m])) / sum(rate(tempo_ingester_traces_created_total{}[5m]))
+```
+
+Users of this pattern may also be interested in the
+`rootless_trace_flushed_to_wal` reason to know when a trace is flushed to the
+wall without a root trace.
```
knylander-grafana, line 42: Is this correct? Is "this pattern" referring to long-running traces or the preceding query example? 
```suggestion
If you have long-running traces, you may also be interested in the
`rootless_trace_flushed_to_wal` reason to know when a trace is flushed to the
wall without a root trace.
```

docs/sources/tempo/troubleshooting/long-running-traces.md
```diff
@@ -0,0 +1,61 @@
...
+
+Additional `reason` fields are available for discovery with the following
+query.
```
knylander-grafana, line 45:
```suggestion
You can use `reason` fields for discovery with this query: 
```

docs/sources/tempo/troubleshooting/long-running-traces.md
```diff
@@ -0,0 +1,61 @@
...
+In general, Tempo functions at its peak when all parts of a trace are stored
+within as few blocks as possible. There is a wide variety of tracing patterns
+in the wild, which makes it impossible to optimize for all of them.
+
+While the preceding information can help determine what Tempo is doing, it may
+be worth modifying the usage pattern slightly. For example, you may want to use
+[span
+links](https://opentelemetry.io/docs/concepts/signals/traces/#span-links), so
+that traces are split up, allowing one trace to complete, while pointing to the
+next trace in the causal chain . This would allow both traces to finish in a
```
knylander-grafana, line 60:
```suggestion
next trace in the causal chain . This allows both traces to finish in a
```

knylander-grafana (approved):
Thank you for adding the doc! A few minor suggestions, but otherwise looks good.

joe-elliott (approved):
i'm good! nice addition

PR #3979, opened 2024-08-19
tools: update go and add goyacc and drone-cli

electron0zero:
**What this PR does**:

update tools image to include goyaac and drone-cli, 
also update go to 1.23 in the tools image, required to update go version in tempo.

once we have new image with these tools, we can stop using loki-build-image and move these commands to use our own build image.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -273,8 +273,8 @@ gen-proto:  ## Generate proto files
 ##@ Gen Traceql
 
 .PHONY: gen-traceql 
-gen-traceql: ## Generate traceql 
-	docker run --rm -v${PWD}:/src/loki ${LOKI_BUILD_IMAGE} gen-traceql-local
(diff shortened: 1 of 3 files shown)


build/tools.mk
```diff
@@ -12,7 +12,7 @@ TOOL_DIR     ?= tools
 TOOL_CONFIG  ?= $(TOOL_DIR)/tools.go
 
 TOOLS_IMAGE ?= grafana/tempo-ci-tools
-TOOLS_IMAGE_TAG ?= main-5e4eafd
+TOOLS_IMAGE_TAG ?= latest
```
electron0zero, line 15:
```suggestion
TOOLS_IMAGE_TAG ?= main-5e4eafd
```

PR #3988, opened 2024-08-21
fix: use the builtin clear to clean the map of trace sizes

javiermolinar:
**What this PR does**:
Use the builtin clear to clean the map of trace sizes. Currently, we are allocating the current length of the map. This length therefore would continue growing.

The context was not being used so I've removed it as well.

**Checklist**
- [X] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## main / unreleased
 
 * [ENHANCEMENT] Add bytes and spans received to usage stats [#3983](https://github.com/grafana/tempo/pull/3983) (@joe-elliott)
-
+* [BUGFIX] Use the builtin clear to clean the map of trace sizes [#3988](https://github.com/grafana/tempo/pull/3988) (@javiermolinar)
 # v2.6.0-rc.0
 
 * [CHANGE] **BREAKING CHANGE** Remove `autocomplete_filtering_enabled` feature flag [#3729](https://github.com/grafana/tempo/pull/3729) (@mapno)
--- a/modules/ingester/ingester.go
+++ b/modules/ingester/ingester.go
@@ -257,7 +257,7 @@ func (i *Ingester) PushBytesV2(ctx context.Context, req *tempopb.PushBytesReques
 		return nil, err
 	}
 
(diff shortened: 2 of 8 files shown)


joe-elliott (approved):

joe-elliott:
There is a behavior change here. Please monitor the memory perf impact on ingesters after rolling out.

Currently we use the previous map length to reallocate. This is less efficient, but it allows the map size to expand and contract over time.

I do not know for sure, but I think `clear()` is more efficient, but it will never reduce the amount of memory used.

Related issue: https://github.com/golang/go/issues/20135

And another map implementation that claims to be faster: https://github.com/dolthub/swiss

javiermolinar:
> There is a behavior change here. Please monitor the memory perf impact on ingesters after rolling out.
> 
> Currently we use the previous map length to reallocate. This is less efficient, but it allows the map size to expand and contract over time.
> 
> I do not know for sure, but I think `clear()` is more efficient, but it will never reduce the amount of memory used.
> 
> Related issue: [golang/go#20135](https://github.com/golang/go/issues/20135)
> 
> And another map implementation that claims to be faster: https://github.com/dolthub/swiss

Not sure we should add another dependency for that.

joe-elliott:
> Not sure if we should add another dependency for that. 

Agreed

> Maybe we should drop the PR if its not add a perf gain

Any ingester memory profile will be dominated by two things. This will likely not show up.

- Allocating go structs from the stream of incoming trace proto
- Converting those structs into parquet

Here's an alloc objects from ops:

image

wdyt?

javiermolinar:
> wdyt?

I think it won't make any difference. The `CutBlockIfReady` doesn't even show up as you mentioned. The allocations are marginal compared with CutCompletedTraces. Maybe the pro of this pr is that the code seems cleaner.

joe-elliott:
If we think it won't make a difference then let's close this. We are doing a lot as a team and risk should be avoided unless we feel there is gain.

To be clear I think this is quite low risk, but it is a behavior change which would introduce different sharp edges to ingester behavior. Let's focus on our energy on changes we feel we be impactful.

Thank you for finding this and driving an investigation!

Issue #3806, opened 2024-06-24
TraceQL Metrics: { false } | rate() returns data


joe-elliott:
**Describe the bug**
The query

```
{ false } | rate()
```

returns data even though `{ false }` should match nothing:

![image](https://github.com/grafana/tempo/assets/2272392/8ab049bb-90a1-4ca6-bee7-323c26aa22eb)

mdisibio:
This is likely the optimization to remove the second pass in https://github.com/grafana/tempo/blob/main/pkg/traceql/engine_metrics.go#L598  It's not detecting the reason to keep the second pass, which is the evaluation of `false` and would reject all spans.

At first glance I think this is limited to falsey static statements that don't inspect any data.  If there was a column condition in there it should work as expected, because that would be pushed down to storage.   Probably worth capturing this case further up in the query-frontend.

ndk:
Some observations:

1. `{resource.service.name="A"} | rate()` - no metrics (all is filtered out, as expected because "A" does not exist).
2. `{resource.service.name!="A"} | rate()` - all metrics (no filtering, as expected because "A" does not exist).
3. `{.service.name="A"} | rate()` - all metrics (unexpected, should be filtered out like in case 1).
4. `{.service.name!="A"} | rate()` - all metrics (unexpected, should filter out all metrics with `.service.name="A",` even though "A" does not exist).
5. `{resource.service.name="auth-service"} | rate()` - only `auth-service` metrics (expected behavior).
6. `{resource.service.name!="auth-service"} | rate()` - all metrics except `auth-service` ones (expected behavior).
7. `{.service.name1="auth-service"} | rate()` - all metrics (unexpected, should behave like case 8).
8. `{resource.service.name1="auth-service"} | rate()` - no metrics (expected, because `service.name1` does not exist).

I'm not very strong in TraceQL, so these are my assumptions. As I understand, cases 3, 4, and 7 return incorrect results. Am I right? :)

PR #3991, opened 2024-08-21
TraceQL: make nil type symmetric

### LINKED ISSUE #3989: [TraceQL] `!= nil` is not symmetric
**Describe the bug**

`!= nil` is only evaluated correctly one way.

This returns results:
```
{ span.accept != nil }
```

This does not:
```
{ nil != span.accept }
```

The best fix would be to make it work both ways. If that is difficult for some reason we should at least throw an error.

image

image

electron0zero:
**What this PR does**:
makes the nil symmetric.

Fixes #3989 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/ast_execute_test.go
+++ b/pkg/traceql/ast_execute_test.go
@@ -1951,6 +1951,34 @@ func TestSpansetExistence(t *testing.T) {
 			},
 			matches: true,
 		},
+		// .foo exists but flip the query - test symmetric
+		{
+			query: `{ nil != .duration }`,
+			span: &mockSpan{
+				attributes: map[Attribute]Static{
+					NewAttribute("duration"): NewStaticDuration(time.Minute),
+				},
+			},
+			matches: true,
+		},
+		{
+			query: `{ nil != .bar }`,
+			span: &mockSpan{
+				attributes: map[Attribute]Static{
+					NewAttribute("bar"): NewStaticString("bzz"),
+				},
+			},
+			matches: true,
+		},
+		{
+			query: `{ nil != .float }`,
+			span: &mockSpan{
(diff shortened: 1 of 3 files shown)


electron0zero:
After some digging into the failing test, I have arrived at a question that boils down to, “should a query like `nil != 0` return data or not?”

I’m unclear about the expected behavior for queries `nil != 0`, or `nil != 5`, should they evaluate to `true` or `false`?. This behavior isn’t currently defined, and we need to establish a clear definition.

Even if we prevent queries `TypeNil != TypeInt` or TypeNil != TypeFloat` at the parse layers, users can write queries that end up evaluating upto `TypeNil != TypeInt`.

For instance, `avg(.foo)` (foo doesn't exist) is not evaluated to `nil`. so a query like `{  } | avg(.dne) != 0 ` will finally be evaluated to `nil != 0`

We introduced the nil type to allow easy checks for non-existent attributes using queries like `{ .foo != nil }`. If we want to maintain symmetry and allow queries like { nil != .foo }, we need to clearly define the behavior of nil.


Also a follow up question: should functions be evaluating to `nil`, should we be evaluating `avg(.foo)` to nil, if `foo` doesn't exist?

electron0zero:
the issue we see with avg here, seems like we see same issue with rate() in https://github.com/grafana/tempo/issues/3806

stoewer:
I continued working on this bug [here](https://github.com/grafana/tempo/pull/4869). I'm closing this for now just to avoid confusion. @electron0zero I hope that's OK for you

electron0zero:
thanks for taking care of this, I started it but never got to finally pushing it over finish line.

PR #3995, opened 2024-08-22
[TraceQL Metrics] Remove all obsolete settings and code for RF3 metrics

mdisibio:
**What this PR does**:
Deletes all obsolete settings and code for metrics queries against RF3 blocks.  This path didn't scale, so we pivoted in the last release to a read path using RF1 blocks flushed from the generators. These don't need deduping or complex job handling, and it scales really well.  This is the long-term duration, and some of the RF3 path is already broken accidentally by other feature additions, so it's good to delete it now.

This introduces some **breaking changes** against `main` builds in the past few months. (There is no breaking change with v2.5, these are unreleased features changing pre-2.6)
1. the rf1_read_path frontend setting is deleted. The read path is always rf1 now. This will cause a configuration parse error on startup if you have set this value. It must be removed.
2. The local-blocks processor must be enabled to start using metrics queries like `{ } | rate()`. If not enabled metrics queries fail with the error `localblocks processor not found`.  Enabling the local-blocks processor can be done two ways:

(a) .  Per-tenant in the per-tenant overrides:
```yaml
overrides:
  'tenantID':
    metrics_generator_processors:
      - local-blocks
```

(b) or by default for all tenants in the main config:
```yaml
overrides:
  defaults:
    metrics_generator:
      processors: [local-blocks] 
```

3. The get metrics queries on historical data, you must configure the local-blocks processor to flush rf1 blocks to object storage:

```yaml
metrics_generator:
  processor:
    local_blocks:
      flush_to_storage: true
```

Other changes:
1. The query hints `sample` and `job_interval` are obsolete and were only parts of the RF3 sharding mechanism. These hints no longer do anything.

Fixes n/a

**Checklist**
- [x] Tests updated
- [x] Documentation added


joe-elliott (approved):
I believe we can remove this as well:

https://github.com/grafana/tempo/blob/main/modules/frontend/combiner/metrics_query_range.go#L34-L48

javiermolinar:
@mdisibio I have found another "error" when the metrics-generator is disabled:

`error finding generators in Querier.SpanMetricsSummary: empty ring`

This comes from here: https://github.com/grafana/tempo/blob/1e5ea804162fdd54711ad13b643776d444368d63/modules/querier/querier_query_range.go#L32

When the `metrics_generator:` entry in the Tempo config yaml is not present.

Since this will be there for the whole release until RF1 is done, we should consider stopping any traceql metric query if the generator is not enabled, with a more descriptive error.

mdisibio:
> @mdisibio I have found another 500 error response when the metrics-generator is disabled:
> 
> `error finding generators in Querier.SpanMetricsSummary: empty ring`

Thanks.  That should stop the query already, when the querier job returns error to the frontend. By earlier do you mean do the check in the query-frontend?  The new error could be a little cleaner but the query-frontend doesn't access the generator ring currently, and not sure I want to introduce that as a last-minute change.

However the error message is incorrect, it was copy-and-pasted from the other API. Fixed that and a few others.

PR #3996, opened 2024-08-22
Frontend cleanup and perf improvement

joe-elliott:
**What this PR does**:
This PR cleans up unused code that Tempo inherited that allows the configuration of the max number of queriers a tenant has access to. This code is complicating attempts to improve this element of the frontend. It is currently unused and not even usable.

Additionally, I've moved the clean up of user queues to a timer. This prevents "flapping" of user queue creation/deletion. Currently, we delete the queue if we [notice it has 0 jobs](https://github.com/grafana/tempo/blob/8a0b0e72a0db8374152333b1a7401d184face9e0/modules/frontend/queue/queue.go#L182-L185). This requires us to over provision go routines to shove jobs into the queue to prevent it from being constantly recreated. This PR should allow us to reduce the number of goroutines used to put pressure on the [user queues lock](https://github.com/grafana/tempo/blob/8a0b0e72a0db8374152333b1a7401d184face9e0/modules/frontend/queue/queue.go#L148-L157) and achieve the same query performance. 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * [ENHANCEMENT] TraceQL: Attribute iterators collect matched array values [#3867](https://github.com/grafana/tempo/pull/3867) (@electron0zero, @stoewer)
 * [ENHANCEMENT] Add bytes and spans received to usage stats [#3983](https://github.com/grafana/tempo/pull/3983) (@joe-elliott)
 * [ENHANCEMENT] Prevent massive allocations in the frontend if there is not sufficient pressure from the query pipeline. [#3996](https://github.com/grafana/tempo/pull/3996) (@joe-elliott)
+  **BREAKING CHANGE** Removed `querier_forget_delay` setting from the frontend. This configuration option did nothing.
 
 # v2.6.0-rc.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * [ENHANCEMENT] TraceQL: Attribute iterators collect matched array values [#3867](https://github.com/grafana/tempo/pull/3867) (@electron0zero, @stoewer)
 * [ENHANCEMENT] Add bytes and spans received to usage stats [#3983](https://github.com/grafana/tempo/pull/3983) (@joe-elliott)
+* [ENHANCEMENT] Prevent massive allocations in the frontend if there is not sufficient pressure from the query pipeline. [#3996](https://github.com/grafana/tempo/pull/3996) (@joe-elliott)
 
 # v2.6.0-rc.0
 
--- a/modules/frontend/queue/queue.go
+++ b/modules/frontend/queue/queue.go
@@ -11,8 +11,7 @@ import (
 )
 
 const (
-	// How frequently to check for disconnected queriers that should be forgotten.
-	forgetCheckPeriod = 30 * time.Second // every 30 seconds b/c the the stopping code requires there to be no queues. i would like to make this 5-10 minutes but then shutdowns would be blocked
+	queueCleanupPeriod = 30 * time.Second // every 30 seconds b/c the the stopping code requires there to be no queues. i would like to make this 5-10 minutes but then shutdowns would be blocked
 )
 
 var (
@@ -42,9 +41,7 @@ func FirstUser() UserIndex {
 // Request stored into the queue.
 type Request interface{}
 
-// RequestQueue holds incoming requests in per-user queues. It also assigns each user specified number of queriers,
-// and when querier asks for next request to handle (using GetNextRequestForQuerier), it returns requests
-// in a fair fashion.
+// RequestQueue holds incoming requests in per-user queues.
 type RequestQueue struct {
 	services.Service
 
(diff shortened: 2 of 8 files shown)


modules/frontend/queue/queue.go
```diff
@@ -197,13 +190,22 @@ FindQueue:
...
 	defer q.mtx.Unlock()
 
-	if q.queues.forgetDisconnectedQueriers(time.Now()) > 0 {
-		// We need to notify goroutines cause having removed some queriers
-		// may have caused a resharding.
+	removedQueue := false
+
+	// look for 0 len queues and remove them
+	for userID, uq := range q.queues.userQueues {
+		if uq.ch != nil && len(uq.ch) == 0 {
```
javiermolinar, line 201: I think  we can simplify this:
```go
if  len(uq.ch) == 0
```
It's not possible to be nil because we always create the channel with make. Furthermore, len also handle nil.

modules/frontend/queue/queue.go
```diff
@@ -197,13 +190,22 @@ FindQueue:
...
 	q.mtx.Lock()
 	defer q.mtx.Unlock()
 
-	if q.queues.forgetDisconnectedQueriers(time.Now()) > 0 {
-		// We need to notify goroutines cause having removed some queriers
-		// may have caused a resharding.
+	removedQueue := false
+
+	// look for 0 len queues and remove them
+	for userID, uq := range q.queues.userQueues {
```
javiermolinar, line 200: We should move this logic to user_queues as before, I think it's the responsibility of user_queues to know how to cleanup queues:

```go
if q.queues.deleteEmptyQueues(){
   q.cond.Broadcast()
}
```

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 * [ENHANCEMENT] TraceQL: Attribute iterators collect matched array values [#3867](https://github.com/grafana/tempo/pull/3867) (@electron0zero, @stoewer)
 * [ENHANCEMENT] Add bytes and spans received to usage stats [#3983](https://github.com/grafana/tempo/pull/3983) (@joe-elliott)
+* [ENHANCEMENT] Prevent massive allocations in the frontend if there is not sufficient pressure from the query pipeline. [#3996](https://github.com/grafana/tempo/pull/3996) (@joe-elliott)
```
javiermolinar, line 5: Although querier_forget_delay is not documented and I hardly believe someone is using it we should consider making this a breaking change in the changelog

javiermolinar (commented):
Awesome! Love the code clean-up

This is currently the memory allocation in ops:
image

javiermolinar:
I wonder if we should rename "users" to "tenants". It's pretty confusing already

modules/frontend/v1/frontend.go
```diff
@@ -22,41 +23,34 @@ import (
...
 	MaxOutstandingPerTenant int                    `yaml:"max_outstanding_per_tenant"`
-	QuerierForgetDelay      time.Duration          `yaml:"querier_forget_delay"`
 	MaxBatchSize            int                    `yaml:"max_batch_size"`
 	LogQueryRequestHeaders  flagext.StringSliceCSV `yaml:"log_query_request_headers"`
 }
 
 // RegisterFlags adds the flags required to config this to the given FlagSet.
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&cfg.MaxOutstandingPerTenant, "querier.max-outstanding-requests-per-tenant", 2000, "Maximum number of outstanding requests per tenant per frontend; requests beyond this error with HTTP 429.")
-	f.DurationVar(&cfg.QuerierForgetDelay, "query-frontend.querier-forget-delay", 0, "If a querier disconnects without sending notification about graceful shutdown, the query-frontend will keep the querier in the tenant's shard until the forget delay has passed. This feature is useful to reduce the blast radius when shuffle-sharding is enabled.")
```
electron0zero, line 39: remove it from docs as well? and mark it as breaking change in case folks are using it?

electron0zero (approved):

modules/frontend/v1/frontend.go
```diff
@@ -371,20 +366,17 @@ func (f *Frontend) queueRequest(ctx context.Context, req *request) error {
...
-	return f.requestQueue.EnqueueRequest(joinedTenantID, req, maxQueriers)
+	return f.requestQueue.EnqueueRequest(joinedTenantID, req)
 }
 
 // CheckReady determines if the query frontend is ready.  Function parameters/return
 // chosen to match the same method in the ingester
 func (f *Frontend) CheckReady(_ context.Context) error {
 	// if we have more than one querier connected we will consider ourselves ready
-	connectedClients := f.requestQueue.GetConnectedQuerierWorkersMetric()
+	connectedClients := f.connectedQuerierWorkers.Load()
```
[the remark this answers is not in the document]
joe-elliott: I would like to leave it. There is a chance that a bug decrements the value of `connectedQuerierWorkers` to negative values. If so, the log message as is would help us see this.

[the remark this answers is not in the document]
javiermolinar: clever!

modules/frontend/queue/queue.go
joe-elliott: queue and user_queues are pretty grossly intertwined, but you're not wrong. fixing!

joe-elliott:
> I wonder if we should rename "users" to "tenants". It's pretty confusing already
We should drift towards "tenant", but I'm not willing to take that on in this PR.

modules/frontend/queue/queue.go
```diff
@@ -8,12 +8,11 @@ import (
 
 	"github.com/grafana/dskit/services"
 	"github.com/prometheus/client_golang/prometheus"
-	"go.uber.org/atomic"
 )
 
 const (
 	// How frequently to check for disconnected queriers that should be forgotten.
-	forgetCheckPeriod = 5 * time.Second
+	forgetCheckPeriod = 30 * time.Second // every 30 seconds b/c the the stopping code requires there to be no queues. i would like to make this 5-10 minutes but then shutdowns would be blocked
```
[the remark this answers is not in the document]
joe-elliott: not sure where i would do that

javiermolinar (approved):

PR #4007, opened 2024-08-26
Query Frontend: Reduce allocs

joe-elliott:
**What this PR does**:
When performing large queries the frontend can create over 100k queries which causes a lot of subsequent work. There are a lot of opportunities for improvement here. This PR addresses allocations due to repeatedly converting the same dedicated columns objects to proto and then to json by memoizing the results and storing them by hash.

Also, noticed an old note to add tests and a small perf improvement for the query builder.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/dedicated_columns_to_json.go
+++ b/modules/frontend/dedicated_columns_to_json.go
@@ -7,11 +7,12 @@ import (
 	"github.com/grafana/tempo/tempodb/backend"
 )
 
+// jpe - move to api?
 type DedicatedColumnsToJSON struct {
 	columnsToJSON map[uint64]string
 }
 
-func NewDedicatedColumnsToJSON() *DedicatedColumnsToJSON {
+func newDedicatedColumnsToJSON() *DedicatedColumnsToJSON {
 	return &DedicatedColumnsToJSON{
 		columnsToJSON: make(map[uint64]string),
 	}
@@ -19,7 +20,7 @@ func NewDedicatedColumnsToJSON() *DedicatedColumnsToJSON {
 
 func (d *DedicatedColumnsToJSON) JSONForDedicatedColumns(cols backend.DedicatedColumns) (string, error) {
 	if len(cols) == 0 {
-		return "", nil // jpe - api.Build needs to handle empty string
+		return "", nil
 	}
 
 	hash := cols.Hash()
--- a/modules/frontend/dedicated_columns_to_json.go
+++ b/modules/frontend/dedicated_columns_to_json.go
@@ -1,0 +1,44 @@
+package frontend
+
+import (
+	"encoding/json"
+	"unsafe"
+
+	"github.com/grafana/tempo/tempodb/backend"
+)
+
+type DedicatedColumnsToJSON struct {
+	columnsToJSON map[uint64]string
+}
+
+func NewDedicatedColumnsToJSON() *DedicatedColumnsToJSON {
+	return &DedicatedColumnsToJSON{
+		columnsToJSON: make(map[uint64]string),
+	}
+}
+
+func (d *DedicatedColumnsToJSON) JSONForDedicatedColumns(cols backend.DedicatedColumns) (string, error) {
+	if len(cols) == 0 {
+		return "", nil // jpe - api.Build needs to handle empty string
+	}
+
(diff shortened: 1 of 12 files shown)


modules/frontend/dedicated_columns_to_json.go
```diff
@@ -0,0 +1,44 @@
...
+		return "", nil
+	}
+
+	hash := cols.Hash()
+	if jsonString, ok := d.columnsToJSON[hash]; ok {
+		return jsonString, nil
+	}
+
+	proto, err := cols.ToTempopb()
+	if err != nil {
```
javiermolinar, line 31: We can move this code into backend.DedicatedColumns:

```go
func (dcs *DedicatedColumns) MarshallJSON() ([]byte, error) {

}
```

modules/frontend/dedicated_columns_to_json.go
```diff
@@ -0,0 +1,44 @@
...
+		return "", err
+	}
+
+	jsonBytes, err := json.Marshal(proto)
+	if err != nil {
+		return "", err
+	}
+
+	jsonString := unsafe.String(unsafe.SliceData(jsonBytes), len(jsonBytes))
```
javiermolinar, line 40:
```go
jsonString := string(jsonBytes)
```

joe-elliott: this will incur another allocation/clone of the byte slice. given that we have full control over the lifecycle of the byte slice it feels safe.

modules/frontend/dedicated_columns_to_json.go
joe-elliott: `MarshalJSON() ([]byte, error)` is picked up by the json encoder. I'm not sure what that would impact and I don't think we want to do that. 

https://pkg.go.dev/encoding/json#example-package-CustomMarshalJSON

The logic to go from DedicatedColumns -> Proto Objects -> Json String is weird enough I'm not sure we should put it with dedicated columns. If we moved it anywhere I think the `api` package would be appropriate. `MarshalToProtoJSON()` ?

modules/frontend/metrics_query_range_sharder.go
```diff
@@ -253,18 +254,18 @@ func (s *queryRangeSharder) buildBackendRequests(ctx context.Context, tenantID s
...
-				DedicatedColumns: dc,
-				Exemplars:        exemplars,
+				BlockID:       m.BlockID.String(),
+				StartPage:     uint32(startPage),
+				PagesToSearch: uint32(pages),
+				Version:       m.Version,
+				Encoding:      m.Encoding.String(),
+				Size_:         m.Size,
+				FooterSize:    m.FooterSize,
+				// DedicatedColumns: dc, for perf reason we pass dedicated columns json in directly to not have to realloc object -> proto -> json
```
javiermolinar, line 264: what about adding a new property DedicatedColumnsJSON to the QueryRangeRequest and deprecating the existing one?

modules/frontend/dedicated_columns_to_json.go
javiermolinar: Makes sense to me, get everything close together

modules/frontend/metrics_query_range_sharder.go
javiermolinar: I know it's used everywhere :P

joe-elliott: i like this change, but the most difficult part would be the rollout. the query frontends would roll first and start sending the new format which the queriers would not know how to interpret.

that doesn't mean we can't do it, but it does mean we should be thoughtful if the cost of manually rolling tempo cells is worth it.

imo we are asking questions about serverless now and how much/what data needs to be passed from frontend -> querier. if we make a big shift that requires a manual rollout we should do it along with deciding what the relationship between querier and frontend is.

modules/frontend/dedicated_columns_to_json.go
joe-elliott: moved dedicated columns to json thinger into `/pkg/api`

electron0zero (approved):
lgtm :shipit:

PR #4021, opened 2024-08-28
Generator remote write update handling of X-Scope-OrgID

mdisibio:
**What this PR does**:
This PR changes the header precedence in Generator remote write, so that statically configured X-Scope-OrgID is honored. This change is necessary to unlock a use case that is currently not possible.  Will do my best to explain.  The desired behavior is to have a setup like this:
* 2 remote write targets with different headers and auth:
  * remote write target A:
    * The usual setup:
    * dynamically set X-Scope-OrgID header == tenant ID
  * remote write target B:
    * statically configured X-Scope-OrgID (we want to target a fixed metrics tenant)
    * Dynamically injected metrics label with the tenant ID like `{__id__="<tenantid>"}`

Target B doesn't work currently because the logic gives highest precedence to dynamically set X-Scope-OrgID.  The statically configured header is deleted and a warning logged.

This logic was inherited and designed to protect against broken authentication, but after review it doesn't seem a compelling use case.  Remote write headers can be set in two locations:

1. In the global config:
```yaml
metrics_generator:
  storage:
    remote_write:
      - url: ...
        headers:
          X-Scope-OrgID: ...
```

2. In runtime overrides per-tenant:
```yaml
overrides:
  '*':
    metrics_generator_remote_write_headers:
      X-Scope-OrgID: ...
```

Both locations are statically configured by a Tempo operator, and it is more likely they are intentional than not.

Fixes n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/storage/config_util.go
+++ b/modules/generator/storage/config_util.go
@@ -32,7 +32,7 @@ func generateTenantRemoteWriteConfigs(inputs []prometheus_config.RemoteWriteConf
 		if tenant != util.FakeTenantID && addOrgIDHeader {
 			existing := ""
 			for k, v := range output.Headers {
-				if strings.EqualFold(user.OrgIDHeaderName, k) {
+				if strings.EqualFold(user.OrgIDHeaderName, strings.TrimSpace(k)) {
 					existing = v
 					break
 				}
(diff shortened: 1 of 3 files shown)


modules/generator/storage/config_util.go
```diff
@@ -13,43 +13,51 @@ import (
...
+
+		// Copy headers so we can modify them
+		output.Headers = copyMap(output.Headers)
+
+		// Inject/overwrite custom headers from runtime overrides
+		for k, v := range headers {
+			output.Headers[k] = v
+		}
 
 		// Inject/overwrite X-Scope-OrgID header in multi-tenant setups
```
kvrhdn, line 31: Maybe we should do a slight update since the logic changed
```suggestion
		// Inject X-Scope-OrgID header in multi-tenant setups if not set already
```

kvrhdn (approved):
LGTM, PR is clear :+1:

PR #4023, opened 2024-08-28
Convert tenant index to proto in the backend

zalegrala:
**What this PR does**:

In large clusters, a significant portion of the poller memory is used
unmarshaling json for the tenant index.  This PR converts the tenant index to a
proto to reduce memory usage.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/backend/tenantindex.go
```diff
@@ -64,3 +65,79 @@ func (b *TenantIndex) unmarshal(buffer []byte) error {
...
+		tenantIndex.CompactedMeta = append(tenantIndex.CompactedMeta, cPb)
+	}
+
+	return tenantIndex, nil
+}
+
+func (b *TenantIndex) fromProto(pb *backend_v1.TenantIndex) error {
+	b.CreatedAt = pb.CreatedAt
+	var (
+		meta          = make([]*BlockMeta, len(pb.Meta))
```
zalegrala, line 106: Be consistent with above `proto()`.

tempodb/backend/raw.go
```diff
@@ -5,19 +5,27 @@ import (
...
 )
 
 const (
+	// JSON
 	MetaName          = "meta.json"
 	CompactedMetaName = "meta.compacted.json"
 	TenantIndexName   = "index.json.gz"
+
+	// Proto
+	TenantIndexNameProto = "index"
```
zalegrala, line 27: The index is no longer gzipped.  I'm happy to add this back in if we'd like to see the file compressed.  Without it, in my small sample, the file itself was 2x.

tempodb/backend/v1/v1.proto
```diff
@@ -0,0 +1,43 @@
...
+
+import "google/protobuf/timestamp.proto";
+import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+
+
+option go_package = "github.com/grafana/tempo/tempodb/backend/v1";
+
+message BlockMeta {
+    string version = 1;
+    string block_id = 2;
```
javiermolinar, line 15: I wonder if instead of string we could use "bytes"

A GUID is a 16 byte array, as a string we are using 36 characters, which is 36 bytes (36 * uint8)

tempodb/backend/v1/v1.proto
```diff
@@ -0,0 +1,43 @@
...
+    bytes min_id = 3;
+    bytes max_id = 4;
+    string tenant_id = 5;
+    google.protobuf.Timestamp start_time = 6[(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
+    google.protobuf.Timestamp end_time = 7[(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
+    int32 total_objects = 8;
+    uint64 size = 9;
+    uint32 compaction_level = 10;
+    int32 encoding = 11;
```
javiermolinar, line 24: Could this be resized to uint32?

javiermolinar: it can be negative?

tempodb/backend/v1/v1.proto
zalegrala: I went back and forth on this, but I removed the code that make me choose string here.  I'll test with bytes.

tempodb/backend/v1/v1.proto
zalegrala: I think we can resize yes.  I was matching what was in BlockMeta currently, or as close as I could.

tempodb/backend/v1/v1.proto
zalegrala: I've pushed a commit for this.  Good call out.

tempodb/backend/v1/v1.proto
zalegrala: I've pushed a commit for this.

mapno (approved):

tempodb/backend/raw.go
```diff
@@ -222,25 +241,18 @@ func (r *reader) BlockMeta(ctx context.Context, blockID uuid.UUID, tenantID stri
...
-	reader, size, err := r.r.Read(ctx, TenantIndexName, KeyPath([]string{tenantID}), nil)
+	tenantIndexProto, err := r.tenantIndexProto(ctx, tenantID)
 	if err != nil {
 		return nil, err
 	}
 
-	defer reader.Close()
-
-	bytes, err := tempo_io.ReadAllWithEstimate(reader, size)
+	tenantIndex := &TenantIndex{}
```
joe-elliott, line 249: if the proto read fails should we fall back to reading the old json index? this might help with a smooth transition when all components first come up and there is no proto indexes

zalegrala: I thought about this and decided not, though let me know if you disagree.  My thought was that if we are not updating the old file, then it could potentially hang out for years and become completely wrong, so falling back doesn't help.  If we wanted to write both for a period, we could fallback, but I'm not sure its worth the extra complexity.  Presumably we would do this as a hardcode which would be on a release cycle, or as a config flag, which requires extra operational handling.

tempodb/backend/raw.go
joe-elliott: not gzipping would save a fair amount of memory/cpu. i think we'd need to see how large this file is uncompressed to decide.

note also that gzip'ing comes with the benefit of checksums added into the compressed file.

a potential option here is to use lz4, snappy, or zstd which all have better perf characteristics than gzip.

tempodb/backend/raw.go
joe-elliott: > then it could potentially hang out for years and become completely wrong, so falling back doesn't help.

yeah, good call out. i prefer your approach. we'll just need a bit of extra care on initial rollout, but i know that's on your list regardless

tempodb/backend/raw.go
zalegrala: What do you think about taking this change without the compression and then observing?  I think the approach of leaving the old indexes in place means we can do this again if we want to make the backend file `index.lz4` or whatever we want to do later.

joe-elliott: > What do you think about taking this change without the compression and then observing? 

i'm good if you're good :)

PR #4032, opened 2024-08-29
feat: implement simple Fetch by key for cache items

javiermolinar:
**What this PR does**:
It implements a new way to fetch by key. That way we can skip more expensive operations using batches of key lookups when we only want one key.

It replaces the existing Fetch in the sync_handler_cache middleware.

Tempo local example now can use memcache, I think this doesn't add too much overhead and it's good for testing

[X] Test it locally against both Memcached and Redis

Fixes #<issue number>

**Checklist**
- [X] Tests updated
- [ ] Documentation added


javiermolinar:
I would like to test this against a real memcache/redis

pkg/cache/memcached.go
```diff
@@ -147,10 +146,32 @@ func (c *Memcached) Fetch(ctx context.Context, keys []string) (found []string, b
 	return
 }
 
+// FetchKey gets a single key from the cache
+func (c *Memcached) FetchKey(ctx context.Context, key string) (buf []byte, found bool) {
+	const method = "Memcache.Get"
```
joe-elliott, line 151: i have noticed that we spend a _lot_ of resources making all these tiny Memcache.Get spans. wdyt about either adding them as an event to the current span? or maybe putting them behind a debug flag?

we don't currently have a debug flag but maybe we add the concept with an env var? `DISTRIBUTED_TRACE_LEVEL=debug`? `DEBUG_DISTRIBUTED_TRACE=true` ?

joe-elliott:
Oh, can we replace this as well:

https://github.com/grafana/tempo/blob/main/tempodb/backend/cache/cache.go#L118

knylander-grafana:
Is there any user impact for this change like a config file option?

modules/frontend/pipeline/sync_handler_cache.go
```diff
@@ -173,10 +152,10 @@ func (c *frontendCache) fetchBytes(key string) []byte {
 		return nil
 	}
 
-	_, bufs, _ := c.c.Fetch(context.Background(), []string{key})
-	if len(bufs) != 1 {
+	buf, found := c.c.FetchKey(context.Background(), key)
```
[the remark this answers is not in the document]
javiermolinar: No, it makes sense if we want to do a graceful shutdown

javiermolinar:
> Is there any user impact for this change like a config file option?

No, this is internal

pkg/cache/memcached.go
javiermolinar: I like the span event idea, we could track the missing hits as well

javiermolinar: Where did you find the spans created per memcache.Get @joe-elliott ? I haven't been able to see them. What I have found is a duplicate span creation on the redis implementation. We were using CollectedRequest from dskit which internally create an span:

https://github.com/grafana/dskit/blob/931a021fb06b39732425870848e12b5a61333cb9/instrument/instrument.go#L161

I don't even know what is happening after the migration from OpenTracing to OpenTelemetry. 

So I've tried to standardized the way we measure the elapsed time and use the events to track hit and missed keys

javiermolinar:
Now we have 0 references from code to the Fetch method, we even have an internal fetchKeysBatched method in Memcached code

https://github.com/grafana/tempo/blob/5e3e03ac319d4facdf251ab65cfae277408cbfac/pkg/cache/memcached.go#L206

Do think this is something we are going to need?

javiermolinar:
> This LGTM. Two things:
> 
> * Let's remove the batching code as you suggested.
> * I'd prefer to back out the event code, but I'm good if you want to leave it. I thought we were replacing a span creation, but this will just incur more cost. Let me know which direction you decide to go.

right, I think we should handle this with two new metric counters one for hits and another one for missed ones

joe-elliott (approved):

PR #4069, opened 2024-09-11
Add disk caching in ingester SearchTagValuesV2 for completed blocks

electron0zero:
**What this PR does**:

cache the response for TagValuesV2 requests alongside the local blocks on the ingesters, this should speedup the autocomplete queries hitting ingesters.

In our dev env, it shows sizeable difference in p90, p95 of latency and throughout for tags that has values with high cardinality.

also updated instrumentation to make the ingester search spans more closer to the code.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/ingester/instance_search.go
+++ b/modules/ingester/instance_search.go
@@ -438,25 +438,39 @@ func (i *instance) SearchTagValuesV2(ctx context.Context, req *tempopb.SearchTag
 		return engine.ExecuteTagValues(ctx, tag, query, traceql.MakeCollectTagValueFunc(collector.Collect), fetcher)
 	}
 
-	searchBlock := func(ctx context.Context, s common.Searcher, bc common.BlockCacher, spanName string) error {
-		ctx, span := tracer.Start(ctx, "instance.SearchTagValuesV2."+spanName)
-		defer span.End()
-
+	exitEarly := func() bool {
 		if anyErr.Load() != nil {
-			return nil // Early exit if any error has occurred
+			return true // Early exit if any error has occurred
 		}
 
 		if maxBlocks > 0 && inspectedBlocks.Inc() > maxBlocks {
+			return true
+		}
+
+		return false // Continue searching
+	}
+
+	searchBlock := func(ctx context.Context, s common.Searcher, spanName string) error {
+		ctx, span := tracer.Start(ctx, "instance.SearchTagValuesV2."+spanName)
+		defer span.End()
+
+		if exitEarly() {
 			return nil
 		}
 
-		// cacher not provided, just search the block
-		if bc == nil {
-			return performSearch(ctx, s, valueCollector)
+		return performSearch(ctx, s, valueCollector)
+	}
+
+	searchBlockWithCache := func(ctx context.Context, b *LocalBlock, spanName string) error {
+		ctx, span := tracer.Start(ctx, "instance.SearchTagValuesV2."+spanName)
+		defer span.End()
+
+		if exitEarly() {
+			return nil
 		}
 
-		// we have a cacher, so we check the cache first
-		cacheData, err := bc.GetDiskCache(ctx, cacheKey)
+		// check the cache first
+		cacheData, err := b.GetDiskCache(ctx, cacheKey)
 		if err != nil {
 			// just log the error and move on...we will search the block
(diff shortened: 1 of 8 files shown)


modules/ingester/instance_search.go
```diff
@@ -427,17 +450,59 @@ func (i *instance) SearchTagValuesV2(ctx context.Context, req *tempopb.SearchTag
 			return nil
 		}
 
-		// if the query is empty, use the old search
-		if traceql.IsEmptyQuery(query) {
-			return s.SearchTagValuesV2(ctx, tag, traceql.MakeCollectTagValueFunc(valueCollector.Collect), common.DefaultSearchOptions())
+		// cacher not provided, just search the block
+		if bc == nil {
```
mdisibio, line 454: Since only a few lines of code of this func are used when not caching - what do you think about splitting it up into caching and non-caching versions?

modules/ingester/instance_search.go
```diff
@@ -417,8 +421,27 @@ func (i *instance) SearchTagValuesV2(ctx context.Context, req *tempopb.SearchTag
 	}
 
 	query := traceql.ExtractMatchers(req.Query)
+	// cacheKey will be same for all blocks in a request so only compute it once
+	cacheKey := searchTagValuesV2CacheKey(req, limit, "cache_search_tagvaluesv2")
```
mdisibio, line 425: Good catch.  Do you know if we honor the time range here in the ingesters?  For example is there a difference when you autocomplete for `last 5 minutes` versus `last 30 minutes` ?  If so we might need account for that here.

The ideal case is:   we don't honor the time range once it hits the ingesters, and we zero out `req.Start` `req.End` before hashing, so that all incoming requests will match the existing cache, even if the time range was specified.

tempodb/encoding/common/interfaces.go
```diff
@@ -32,6 +32,11 @@ type Searcher interface {
 	FetchTagNames(context.Context, traceql.FetchTagsRequest, traceql.FetchTagsCallback, SearchOptions) error
 }
 
+type BlockCacher interface {
```
mdisibio, line 35: I'm ok with leaving this out until we cache items in non-LocalBlocks.   And make these methods part of `LocalBlock`. If we split `searchBlock` like I mentioned above, the caching variant can take the `*LocalBlock` directly.

modules/ingester/instance_search.go
```diff
@@ -427,17 +450,59 @@ func (i *instance) SearchTagValuesV2(ctx context.Context, req *tempopb.SearchTag
...
+		// results not in cache, so search the block
+		// using a local collector to collect values from the block and set cache
+		localCol := collector.NewDistinctValue[tempopb.TagValue](limit, func(v tempopb.TagValue) int { return len(v.Type) + len(v.Value) })
+		localErr := performSearch(ctx, s, localCol)
+		if localErr != nil {
+			return localErr
+		}
+
+		// marshal the local collector and set the cache
+		valuesProto, err := valuesToTagValuesV2RespProto(localCol.Values())
```
mdisibio, line 491: We can call `localCol.Values()` once to avoid building the slice twice.

modules/ingester/instance_search.go
electron0zero: we are passing start and end, and we adjust them to respect the `QueryIngestersUntil`

looks like we are not using start and end right now when handling the SearchTagValuesV2 but we have an TODO to make use of start and end so in future we might add support for start and end in ExecuteTagValues. https://github.com/grafana/tempo/blob/9fdc0324d8572366a0e6e40df728cb4416863f61/pkg/traceql/storage.go#L230

tempodb/encoding/common/interfaces.go
electron0zero: cool, can LocalBlock and keep it simple

modules/ingester/instance_search.go
electron0zero: done

modules/ingester/instance_search.go
electron0zero: done, dropped start and end from the cachekey and added a note.

mdisibio (approved):

PR #4072, opened 2024-09-11
Replace TenantIndex, BlockMeta and CompactedBlockMeta with proto near-equivalents

zalegrala:
**What this PR does**:

Here we replace the `TenantIndex`, `BlockMeta` and `CompactedBlockMeta` with structs which are generated from proto.  This allows us to convert the `TenantIndex` to proto-only without the translation between a backend struct and an internal struct, since the internal structs here are being replaced.  The `BlockMeta` and `CompactedBlockMeta` remain `json`.  We may consider replacing these with proto in the future.

The replacement structs are not identical.  Of note, the `BlockMeta` struct has several fields which have no proto equivalent.  In order to overcome this, several types have been defined in the proto as `bytes`, which allows the use of the `customtype` gogo proto option to point to a specific type.  Those types must conform to a specific interface, and so methods have been added where necessary.

- `backend.DedicatedColumns` updated to include proto marshal and unmarshal methods
- `backend.Encoding` updated to include prto marshal and unmarshal methods
- `backend.BlockMeta` has a few type changes
    - `TotalObjects` changes `int` -> `int64`
    - `Size` becomes a funcion and the old size is moved to `Size_`
    - `CompactionLevel` changes `uint8` -> `uint32`
    - `BloomShardCount` changes `uint16` -> `uint32`
    - `ReplicationFactor` changes `uint8` -> `uint32`

A new UUID type has been added which is backed by the google UUID type and provide
the necessary interface for use with proto. 

The `TenantIndex` is written both as proto and json, and read as protofirst with a fallback to json on error.

The `BlockMeta` and `CompactedBlockMeta` are written/read only as json.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


Makefile
```diff
@@ -222,7 +222,8 @@ PROTOC = docker run --rm -u ${shell id -u} -v${PWD}:${PWD} -w${PWD} ${DOCKER_PRO
 PROTO_INTERMEDIATE_DIR = pkg/.patched-proto
 PROTO_INCLUDES = -I$(PROTO_INTERMEDIATE_DIR)
 PROTO_GEN = $(PROTOC) $(PROTO_INCLUDES) --gogofaster_out=plugins=grpc,paths=source_relative:$(2) $(1)
-PROTO_GEN_WITH_VENDOR = $(PROTOC) $(PROTO_INCLUDES) -Ivendor -Ivendor/github.com/gogo/protobuf --gogofaster_out=plugins=grpc,paths=source_relative:$(2) $(1)
+PROTO_GEN_WITH_VENDOR = $(PROTOC) $(PROTO_INCLUDES) -Ivendor -Ivendor/github.com/gogo/protobuf -Ipkg/tempopb --gogofaster_out=plugins=grpc,paths=source_relative:$(2) $(1)
```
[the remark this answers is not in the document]
zalegrala: Thank you, it does not.  This was left over from a previous trial.

tempodb/backend/raw.go
```diff
@@ -220,29 +263,81 @@ func (r *reader) BlockMeta(ctx context.Context, blockID uuid.UUID, tenantID stri
...
+		return nil, err
+	}
+
+	return outPb, nil
+}
+
 // TenantIndex implements backend.Reader
 func (r *reader) TenantIndex(ctx context.Context, tenantID string) (*TenantIndex, error) {
-	reader, size, err := r.r.Read(ctx, TenantIndexName, KeyPath([]string{tenantID}), nil)
+	ctx, span := tracer.Start(ctx, "reader.TenantIndex")
```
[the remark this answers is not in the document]
zalegrala: I've dropped tenantindex json writing and also block proto writing/reading and will update the PR description to match.

tempodb/backend/raw.go
```diff
@@ -220,29 +263,81 @@ func (r *reader) BlockMeta(ctx context.Context, blockID uuid.UUID, tenantID stri
...
+	}
+	defer readerPb.Close()
+
+	bytesPb, err := tempo_io.ReadAllWithEstimate(readerPb, size)
+	if err != nil {
+		return nil, err
+	}
+
+	out := &TenantIndex{}
+	err = out.Unmarshal(bytesPb)
```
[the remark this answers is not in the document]
zalegrala: This has been implemented.  Good suggestion.

pkg/uuid/uuid.go
```diff
@@ -0,0 +1,87 @@
+// Package uuid provides a UUID type that can be used in protocol buffer
```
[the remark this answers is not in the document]
zalegrala: This has been moved.

tempodb/backend/mocks.go
```diff
@@ -72,16 +72,21 @@ func (m *MockRawReader) Shutdown() {}
...
 
 func (m *MockRawWriter) Write(_ context.Context, _ string, _ KeyPath, data io.Reader, size int64, _ *CacheInfo) error {
-	var err error
-	m.writeBuffer, err = tempo_io.ReadAllWithEstimate(data, size)
+	if m.writeBuffer == nil {
+		m.writeBuffer = make([][]byte, 0)
+	}
+
+	writeBuffer, err := tempo_io.ReadAllWithEstimate(data, size)
+	m.writeBuffer = append(m.writeBuffer, writeBuffer)
```
mdisibio, line 88: An alternative here would be to use the KeyPath and put the byte slices into a map. Probably about the same complexity as the `m.writeBuffer[len(m.writeBuffer)-2]` checks in the tests below, but thought I'd mention it as something to consider.

tempodb/backend/tenantindex.go
```diff
@@ -3,22 +3,19 @@ package backend
...
 )
 
-// TenantIndex holds a list of all metas and compacted metas for a given tenant
-// it is probably stored in /<tenantid>/blockindex.json.gz as a gzipped json file
-type TenantIndex struct {
-	CreatedAt     time.Time             `json:"created_at"`
-	Meta          []*BlockMeta          `json:"meta"`
-	CompactedMeta []*CompactedBlockMeta `json:"compacted"`
-}
+var _ proto.Message = &TenantIndex{}
```
mdisibio, line 18: +1 for checking the interface.  Usually it's done with a nil pointer to avoid allocating the object.
```suggestion
var _ proto.Message = (*TenantIndex)(nil)
```

tempodb/backend/block_meta.go
```diff
@@ -337,3 +294,64 @@ func (dcs DedicatedColumns) Hash() uint64 {
...
+	return len(bb), nil
+}
+
+func (dcs *DedicatedColumns) Unmarshal(data []byte) error {
+	if len(data) == 0 {
+		return nil
+	}
+
+	// NOTE: See UnmarshalJSON for a performance optimization to record the json
+	// bytes for reuse.
```
mdisibio, line 333: Let's specifically call out "interning" the dedicated column config in this coment.

tempodb/backend/tenantindex.go
```diff
@@ -64,3 +61,43 @@ func (b *TenantIndex) unmarshal(buffer []byte) error {
 	d := json.NewDecoder(gzipReader)
 	return d.Decode(b)
 }
+
+func (b *TenantIndex) marshalPb() ([]byte, error) {
+	buffer := &bytes.Buffer{}
+
+	z, err := zstd.NewWriter(buffer)
```
mdisibio, line 68: Did we do any benchmarks on zstd vs other?  We resulted with snappy in parquet because it had the best balance of speed vs resources.  Not blocking, because we could change it later easily by switching on the file name, but would be interested to see benchmarks if we have them.

tempodb/backend/v1/v1.proto
```diff
@@ -0,0 +1,39 @@
...
+}
+
+message CompactedBlockMeta {
+    BlockMeta block_meta = 1[(gogoproto.embed) = true, (gogoproto.nullable) = false, (gogoproto.jsontag) = ""];
+    google.protobuf.Timestamp compacted_time = 2[(gogoproto.stdtime) = true, (gogoproto.nullable) = false, (gogoproto.jsontag) = "compactedTime"];
+}
+
+message TenantIndex {
+    google.protobuf.Timestamp created_at = 1[(gogoproto.stdtime) = true, (gogoproto.nullable) = false, (gogoproto.jsontag) = "createdAt"];
+    repeated BlockMeta meta = 2;
```
mdisibio, line 37: These fields should probably be pluralized `metas` and `compacted_metas` ?

tempodb/backend/raw.go
```diff
@@ -222,27 +243,57 @@ func (r *reader) BlockMeta(ctx context.Context, blockID uuid.UUID, tenantID stri
 
 // TenantIndex implements backend.Reader
 func (r *reader) TenantIndex(ctx context.Context, tenantID string) (*TenantIndex, error) {
-	reader, size, err := r.r.Read(ctx, TenantIndexName, KeyPath([]string{tenantID}), nil)
+	ctx, span := tracer.Start(ctx, "reader.TenantIndex")
+	defer span.End()
+
+	outPb, err := r.tenantIndexProto(ctx, tenantID)
+	if err == nil {
```
mdisibio, line 250: Wonder if we want to be more strict in the error checks here, by only falling back to json if the proto error is `backend.ErrDoesNotExist` ?   I don't think it makes sense to try the json if there was a proto index but it was corrupted. In that case it would be better to surface the error and fail.

mdisibio (commented):
I like where we ended up with this PR.   The existing block meta.json and block state are unchanged, but we will get the index proto efficiency on the downstream components (frontend, querier).  This PR looks like a lot, but 90% is fallout from the uuid and Size changes. The actual change is much smaller.   99% LGTM left a few q's and comments.

tempodb/backend/v1/v1.proto
zalegrala: I noticed this also, but kept the name to be consistent with the original object.  I agree though, so pushed a commit for this.

tempodb/backend/raw.go
zalegrala: I've pushed a commit for this.

tempodb/backend/tenantindex.go
zalegrala: Hmm, I didn't benchmark, but I can.  We could also push this into an override so we can swap out the implementation without a rollout, then add more implementations later if we want.

mdisibio (approved):
Great work! Let's get the changelog entry and then merge.

PR #4073, opened 2024-09-12
feat: TraceQL metrics: average over time

javiermolinar:
**What this PR does**:
It implements the avg_over_time aggregation function

**Checklist**
- [X] Tests updated
- [X] Documentation added


pkg/traceql/engine_metrics.go
```diff
@@ -1175,12 +1250,21 @@ func (b *SimpleAggregator) Combine(in []*tempopb.TimeSeries) {
 			}
 
 			b.ss[ts.PromLabels] = existing
+			if b.initAvg {
```
javiermolinar, line 1253: Struggling a bit with how to include the average in the Combiner. I think this way we only allocate the extra bytes only for this case

pkg/traceql/engine_metrics.go
```diff
@@ -1096,43 +1146,67 @@ const (
...
+	initAvg        bool
+	ssCounter      map[string]map[int]float64 // Counter of processed elements to calculate median
+	ssCompensation map[string]map[int]float64 // Avg compensation
 }
 
 func NewSimpleCombiner(req *tempopb.QueryRangeRequest, op SimpleAggregationOp) *SimpleAggregator {
 	l := IntervalCount(req.Start, req.End, req.Step)
 	var initWithNaN bool
-	var f func(existingValue float64, newValue float64) float64
+	var f func(b *SimpleAggregator, promLabel string, pos int, newValue float64)
```
javiermolinar, line 1169: Since we need to modify more than one value on each iteration (counter and compensation) this does the trick

pkg/traceql/engine_metrics_average.go
```diff
@@ -0,0 +1,540 @@
...
+	avg             []float64
+	count           []float64
+	compensation    []float64
+	exemplars       []Exemplar
+	exemplarBuckets *bucketSet
+	init            bool
+	vals            S
+}
+
+// In charge of calculating the average over time for a set of spans
```
javiermolinar, line 279: This is the equivalent of the RangeAggregator and the VectorAggregator

pkg/traceql/engine_metrics_average.go
```diff
@@ -0,0 +1,540 @@
...
+	interval := IntervalOf(span.StartTimeUnixNanos(), g.start, g.end, g.step)
+	if interval == -1 {
+		return
+	}
+	inc := g.getSpanAttValue(span)
+	if math.IsNaN(inc) {
+		return
+	}
+	s.count[interval]++
+	mean, c := averageInc(s.avg[interval], inc, s.count[interval], s.compensation[interval])
```
javiermolinar, line 374: This calculates the mean at span level and bubbles up the count. The other possibility is just sum all the values and calculate the average in the final step

pkg/traceql/engine_metrics_average.go
```diff
@@ -0,0 +1,540 @@
...
+			newAvg := sample.Value
+			currentCount := b.ss[counterLabel].Values[pos]
+			newCount := newCountersTS[ts.PromLabels][pos]
+
+			if math.IsNaN(currentAvg) && !math.IsNaN(newAvg) {
+				b.ss[ts.PromLabels].Values[pos] = newAvg
+				b.ss[counterLabel].Values[pos] = newCount
+			} else if !math.IsNaN(newAvg) {
+				// Weighted mean
+				avg := (currentAvg*currentCount + newAvg*newCount) / (currentCount + newCount)
```
javiermolinar, line 181: It calculates the weighted mean. There is another to reduce the intermediate operations (and a possible loss of precision) by adding all the values and counts and calculating the average on the Results()

pkg/traceql/engine_metrics_average.go
```diff
@@ -0,0 +1,540 @@
...
+}
+
+func (b *SpanSetsAverageOverTimeAggregator) initSeriesAggregator(in []*tempopb.TimeSeries, newCountersTS map[string][]float64) {
+	for _, ts := range in {
+		counterPromLabel := ""
+		if strings.Contains(ts.PromLabels, "_type") {
+			counterPromLabel = getLabels(ts.Labels, "_type").String()
+			newCountersTS[counterPromLabel] = make([]float64, b.len)
+			for i, sample := range ts.Samples {
+				newCountersTS[counterPromLabel][i] = sample.Value
```
mdisibio, line 199: Getting a panic here. We need to account for TimestampMs like in Combine().

pkg/traceql/engine_metrics_average.go
```diff
@@ -0,0 +1,540 @@
...
+			if i < len(a.by)-1 {
+				s.WriteString(",")
+			}
+		}
+		s.WriteString(")")
+	}
+	return s.String()
+}
+
+type SpanSetsAverageOverTimeAggregator struct {
```
mdisibio, line 141: nit:  I would rename this to `averageOverTimeSeriesAggregator`.  Spansets is something else in traceql, a collection of individual spans. This is combining time series, so think this name fits better. 

Related:   I don't think these structs are used outside of this package, so let's unexport them by lower casing.

pkg/traceql/engine_metrics_average.go
```diff
@@ -0,0 +1,540 @@
...
+	}
+}
+
+func (g *AvgOverTimeSpanAggregator[F, S]) Observe(span Span) {
+	if !g.getGroupingValues(span) {
+		return
+	}
+
+	s := g.getSeries()
+	interval := IntervalOf(span.StartTimeUnixNanos(), g.start, g.end, g.step)
```
mdisibio, line 365: We could move this check up, as it doesn't depend on s.

pkg/traceql/engine_metrics_average.go
```diff
@@ -0,0 +1,540 @@
...
+			// At this stage, the current.val is an infinite. If the added
+			// value is neither an Inf or a Nan, we can keep that mean
+			// value.
+			// This is required because our calculation below removes
+			// the mean value, which would look like Inf += x - Inf and
+			// end up as a NaN.
+			return mean, compensation
+		}
+	}
+	mean, c := kahanSumInc(inc/count-mean/count, mean, compensation)
```
mdisibio, line 401: This is a neat algorithm.  Just checking my understanding - both inc and mean are intended to be divided by the new count, correct? It seems right.

pkg/traceql/engine_metrics_average.go
```diff
@@ -0,0 +1,540 @@
...
+	return t, c
+}
+
+func (g *AvgOverTimeSpanAggregator[F, S]) ObserveExemplar(span Span, value float64, ts uint64) {
+	if !g.getGroupingValues(span) {
+		return
+	}
+
+	// Observe exemplar
+	all := span.AllAttributes()
```
mdisibio, line 426: I think the allocation of all and lbls can be moved below after the tests.

pkg/traceql/engine_metrics_average.go
```diff
@@ -0,0 +1,540 @@
...
+		labels = append(labels, Label{g.by[i].String(), vals[i]})
+	}
+
+	if len(labels) == 0 {
+		// When all nil then force one
+		labels = append(labels, Label{g.by[0].String(), NewStaticNil()})
+	}
+
+	if t != "" {
+		labels = append(labels, Label{"_type", NewStaticString(t)})
```
mdisibio, line 472: Let's move these strings to consts.   `Compare()` is using `internalLabelMetaType`, can we use that here too?  It's currently in engine_metrics_compare.go but we can relocate it.  For `_count` what about `internalMetaTypeCount`?

pkg/traceql/engine_metrics_average.go
mdisibio: I'm leaning towards this approach as well.  Simple summation would be faster but overflow and precision could be worse?  I think we can start with this approach, and switch later, if/when we have objective numbers that the error rate of the other way is better, or this proves to be a bottleneck.

pkg/traceql/engine_metrics_average.go
mdisibio: Now that I've reviewed further - could we reuse the kahan incremental average here too?  Across a few thousands of jobs, it seems like it could be worthwhile.

mdisibio: Went ahead and did some basic testing of error rates, and kahan approach is definitely better.    Calculated the average of pi and 1e12 a million times each.  Kahan had zero error across either.  Simple sum and divide accumlated a small error on pi.

```
KahanSum: Input: 3.141592653589793 Avg: 3.141592653589793 Error: 0
KahanSum: Input: 1e+12 Avg: 1e+12 Error: 0

Simple: Input: 3.141592653589793 Avg: 3.1415926535849428 Error: 1.5438761380437427e-12
Simple: Input: 1e+12 Avg: 1e+12 Error: 0
```


Here is the code:
<details>

```go
func TestErrorKahanSum(t *testing.T) {
	fs := []float64{math.Pi, 1e12}

	for _, f := range fs {
		t.Run(strconv.FormatFloat(f, 'g', -1, 64), func(t *testing.T) {
			avg := 0.0
			count := 0.0
			c := 0.0

			for i := 0; i < 1_000_000; i++ {
				count++
				avg, c = averageInc(avg, f, count, c)
			}

			fmt.Println("KahanSum: Input:", f, "Avg:", avg, "Error:", 1.0-avg/f)
		})
	}
}

func TestErrorSimpleAvg(t *testing.T) {
	fs := []float64{math.Pi, 1e12}

	for _, f := range fs {
		t.Run(strconv.FormatFloat(f, 'g', -1, 64), func(t *testing.T) {
			sum := 0.0
			count := 0.0

			for i := 0; i < 1_000_000; i++ {
				count++
				sum += f
			}

			avg := sum / count

			fmt.Println("Simple: Input:", f, "Avg:", avg, "Error:", 1.0-avg/f)
		})
	}
}
```

</details>

pkg/traceql/engine_metrics_average.go
javiermolinar: Correct!

docs/sources/tempo/traceql/metrics-queries.md
```diff
@@ -97,7 +100,7 @@ down by HTTP route.
 This might let you determine that `/api/sad` had a higher rate of erroring
 spans than `/api/happy`, for example.
 
-### The `count_over_time`, `min_over_time` and `max_over_time` functions
+### The `count_over_time`, `min_over_time`, `max_over_time` and `avg_over_time` functions
 
 The `count_over_time()` let you counts the number of matching spans per time interval.
```
knylander-grafana, line 105: Should this be duration over time functions heading? 
```suggestion
### The `over_time` functions

Metrics queries supports `count_over_time`, `min_over_time`, `max_over_time`, and `avg_over_time` functions. 

The `count_over_time()` let you counts the number of matching spans per time interval.
```

docs/sources/tempo/traceql/metrics-queries.md
```diff
@@ -128,6 +131,16 @@ The `max_over_time()` let you aggregate numerical values by computing the maximu
 { name = "GET /:endpoint" } | max_over_time(span.http.status_code)
 ```
 
+The `avg_over_time()` let you aggregate numerical values by computing the average value of them, such as the all important span duration.
```
knylander-grafana, line 134:
```suggestion
The `avg_over_time()` function lets you aggregate numerical values by computing the average value of them, such as the all important span duration.
```

knylander-grafana (commented):
Thank you for adding doc! Approving the doc, with two minor suggestions. Please be aware that this PR updates the metrics queries doc: https://github.com/grafana/tempo/pull/4128

docs/sources/tempo/traceql/metrics-queries.md
javiermolinar: It's not only for duration, any numerical value is ok.

mdisibio (approved):
Thanks for the last round of changes.  This is looking great.

docs/sources/tempo/traceql/metrics-queries/_index.md
```diff
@@ -60,7 +60,7 @@ Refer to [Solve problems using metrics queries](./solve-problems-metrics-queries
 
 ### Functions
 
-TraceQL metrics queries currently include the following functions for aggregating over groups of spans: `rate`, `count_over_time`, `quantile_over_time`, `histogram_over_time`, and `compare`.
+TraceQL metrics queries currently include the following functions for aggregating over groups of spans: `rate`, `count_over_time`,`max_over_time`,`min_over_time`,`avg_over_time`, `quantile_over_time`, `histogram_over_time`, and `compare`.
```
knylander-grafana, line 63: Missing spaces after commas.
```suggestion
TraceQL metrics queries currently include the following functions for aggregating over groups of spans: `rate`, `count_over_time`, `max_over_time`, `min_over_time`, `avg_over_time`, `quantile_over_time`, `histogram_over_time`, and `compare`.
```

PR #4074, opened 2024-09-12
TraceByID: don't allow concurrent_shards greater than query_shards

electron0zero:
**What this PR does**:
concurrent shards grater then query shards should not be allowed because it would create more goroutines then the jobs to send these jobs to queriers.

if that happens, set concurrent_shards=query_shards and log a warning 

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/traceid_sharder.go
+++ b/modules/frontend/traceid_sharder.go
@@ -59,8 +59,8 @@ func (s asyncTraceSharder) RoundTrip(pipelineRequest pipeline.Request) (pipeline
 		concurrentShards = uint(s.cfg.ConcurrentShards)
 	}
 
-	// having more concurrent shards than query shards should not happen because then
-	// we would be creating more goroutines then the jobs to send these jobs to queriers
+	// concurrent_shards grater then query_shards should not be allowed because it would create
+	// more goroutines then the jobs to send these jobs to queriers.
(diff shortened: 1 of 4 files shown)


modules/frontend/traceid_sharder.go
```diff
@@ -53,10 +54,19 @@ func (s asyncTraceSharder) RoundTrip(pipelineRequest pipeline.Request) (pipeline
...
 	if s.cfg.ConcurrentShards > 0 {
 		concurrentShards = uint(s.cfg.ConcurrentShards)
 	}
 
+	// concurrent_shards grater then query_shards should not be allowed because it would create
+	// more goroutines then the jobs to send these jobs to queriers.
+	if concurrentShards > uint(s.cfg.QueryShards) {
+		// set the concurrent shards to the total shards
+		concurrentShards = uint(s.cfg.QueryShards)
+		_ = level.Warn(s.logger).Log("concurrent_shards greater than query_shards is invalid, setting concurrent shards equal to query shards")
```
joe-elliott, line 67: warning may make more sense here than on every query:

https://github.com/grafana/tempo/blob/main/cmd/tempo/app/config.go#L141

electron0zero: moved to config.go and added test.

joe-elliott (approved):

PR #4076, opened 2024-09-12
Query Frontend: Job weights

joe-elliott:
**What this PR does**:
The query frontend treats all jobs as the same size when it farms them out to the queriers. This can cause querier instability b/c some jobs actually require quite a bit more resources to execute. By assigning weights to jobs we can reduce the amount each querier is asked to do will hopefully:

1) reduce querier OOMs/timeouts/retries
1) reduce querier latency
1) increase total throughput

*Other changes*

- Removed the roundtripper httpgrpc bridge and pushed the concept of pipeline.Request all the way down into the cortex frontend code. This can be a nice perf improvement b/c translating http -> httpgrpc is costly and we are pushing it to the last moment. Currently for some queries we are translating thousands of jobs and then throwing them away.
- Removed redundant `parseQuery` and `createFetchSpansRequest` to consolidate on the `Compile` function in `pkg/traceql`
- Check for context error before going through retry logic in retryWare. This causes retry metrics to be more accurate in the event of many cancelled jobs.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/frontend/queue/queue.go
```diff
@@ -159,16 +161,30 @@ FindQueue:
...
+			actuallyInBatch++
+
+			// PRTODO: add tests that take weight into account
+			weight := batchBuffer[i].Weight()
+			if weight <= 0 {
+				weight = 1
+			}
+			totalWeight += weight
+
+			if totalWeight >= requestedCount {
```
zalegrala, line 183: I think this makes sense.  I suppose what we're saying here is that we request of this batch a certain high water mark of work that we're willing to take, and the weight increases the notion of complexity for a single item above this threshold.  Implicitly here I suppose is that weight and requestedCount are of the same unit of measure.

modules/frontend/weights/weights.go
```diff
@@ -0,0 +1,50 @@
...
+	for _, c := range req.Conditions {
+		if c.Op != traceql.OpNone {
+			conditions++
+		}
+		if c.Op == traceql.OpRegex || c.Op == traceql.OpNotRegex {
+			regexConditions++
+		}
+	}
+
+	if conditions > 4 { // yay, magic!
```
zalegrala, line 38: A fine starting point.  I was wonder if each condition is weight++, and maybe each regex is weight+2 or some such.  It means for the queue logic that if any condition is present, we'll never consume the entire requested batch.  :thinking:

modules/frontend/weights/weights.go
```diff
@@ -0,0 +1,50 @@
+package weights
+
+import "github.com/grafana/tempo/pkg/traceql"
+
+// PRTODO: test weight calculation
```
zalegrala, line 5: I wrote a quick test for this.

```
package weights

import (
	"testing"

	"github.com/grafana/tempo/pkg/traceql"
)

func TestFetchSpans(t *testing.T) {
	cases := []struct {
		req      *traceql.FetchSpansRequest
		expected int
	}{
		{
			req:      nil,
			expected: 1,
		},
		{
			req:      &traceql.FetchSpansRequest{},
			expected: 2,
		},
		{
			req: &traceql.FetchSpansRequest{
				AllConditions: true,
				Conditions: []traceql.Condition{
					{Attribute: traceql.NewIntrinsic(traceql.IntrinsicTraceRootService), Op: traceql.OpEqual, Operands: []traceql.Static{traceql.NewStaticString("foo")}},
					{Attribute: traceql.NewIntrinsic(traceql.IntrinsicTraceRootService), Op: traceql.OpEqual, Operands: []traceql.Static{traceql.NewStaticString("foo")}},
				},
			},
			expected: 1,
		},
		{
			req: &traceql.FetchSpansRequest{
				AllConditions: true,
				Conditions: []traceql.Condition{
					{Attribute: traceql.NewIntrinsic(traceql.IntrinsicTraceRootService), Op: traceql.OpEqual, Operands: []traceql.Static{traceql.NewStaticString("foo")}},
					{Attribute: traceql.NewIntrinsic(traceql.IntrinsicTraceRootService), Op: traceql.OpEqual, Operands: []traceql.Static{traceql.NewStaticString("foo")}},
					{Attribute: traceql.NewIntrinsic(traceql.IntrinsicTraceRootService), Op: traceql.OpEqual, Operands: []traceql.Static{traceql.NewStaticString("foo")}},
					{Attribute: traceql.NewIntrinsic(traceql.IntrinsicTraceRootService), Op: traceql.OpEqual, Operands: []traceql.Static{traceql.NewStaticString("foo")}},
					{Attribute: traceql.NewIntrinsic(traceql.IntrinsicTraceRootService), Op: traceql.OpEqual, Operands: []traceql.Static{traceql.NewStaticString("foo")}},
				},
			},
			expected: 2,
		},
		{
			req: &traceql.FetchSpansRequest{
				AllConditions: true,
				Conditions: []traceql.Condition{
					{Attribute: traceql.NewIntrinsic(traceql.IntrinsicTraceRootService), Op: traceql.OpEqual, Operands: []traceql.Static{traceql.NewStaticString("foo")}},
					{Attribute: traceql.NewIntrinsic(traceql.IntrinsicTraceRootService), Op: traceql.OpEqual, Operands: []traceql.Static{traceql.NewStaticString("foo")}},
					{Attribute: traceql.NewIntrinsic(traceql.IntrinsicTraceRootService), Op: traceql.OpEqual, Operands: []traceql.Static{traceql.NewStaticString("foo")}},
					{Attribute: traceql.NewIntrinsic(traceql.IntrinsicTraceRootService), Op: traceql.OpEqual, Operands: []traceql.Static{traceql.NewStaticString("foo")}},
					{Attribute: traceql.NewIntrinsic(traceql.IntrinsicTraceRootService), Op: traceql.OpRegex, Operands: []traceql.Static{traceql.NewStaticString("foo")}},
				},
			},
			expected: 3,
		},
	}

	for _, c := range cases {
		actual := FetchSpans(c.req)
		if actual != c.expected {
			t.Errorf("expected %d, got %d", c.expected, actual)
		}
	}
}
```

modules/frontend/transport/roundtripper.go
```diff
@@ -1,56 +0,0 @@
-package transport
```
zalegrala, line 1: Nice cleanup.

modules/frontend/weights/weights.go
joe-elliott: you should be able to pull my branch locally and push changes to it using https://cli.github.com/manual/gh_pr_checkout

modules/frontend/queue/queue.go
joe-elliott: > Implicitly here I suppose is that weight and requestedCount are of the same unit of measure.

yes! currently all jobs fill a single "slot" in the batch. the "weight" is basically just making it fill more slots.

modules/frontend/pipeline/async_weight_middleware.go
```diff
@@ -0,0 +1,103 @@
...
+func setTraceQLWeight(req Request) {
+	var traceQLQuery string
+	query := req.HTTPRequest().URL.Query()
+	if query.Has("q") {
+		traceQLQuery = query.Get("q")
+	}
+	if query.Has("query") {
+		traceQLQuery = query.Get("query")
+	}
+	if traceQLQuery != "" {
```
joe-elliott, line 76: nit: for ease of reading i prefer.

```
if traceQLQuery == "" {
   req.SetWeight(TraceQLSearchWeight)
   return
}

...
```

this reduces nesting of the code below and very clearly communicates the logic taken when the query is not found

modules/frontend/pipeline/pipeline.go
```diff
@@ -65,6 +70,18 @@ func (r *HTTPRequest) ResponseData() any {
...
 
+func (r *HTTPRequest) Weight() int {
+	return r.weight
+}
+
+func (r *HTTPRequest) SetWeight(w int) {
+	r.weight = w
+}
+
+func (r *HTTPRequest) FromHTTPRequest(request *http.Request) *HTTPRequest {
```
joe-elliott, line 81: i think we can remove this method and use the Clone method (once we add the weight).

see my comment in the metrics_query_range_sharder.go file

joe-elliott: finishing the review i see it's used more below. if we do need this i think we should call it:
```
CloneWithHTTPRequest(request *http.Request)
```
and have it copy all internal fields except for the http request which it gets from the arg.

javiermolinar: Is there a clone method for Pipeline HTTPRequest? The one I see is for the  stdlib http.Request

modules/frontend/config.go
```diff
@@ -32,6 +31,9 @@ type Config struct {
 
 	// A list of regexes for black listing requests, these will apply for every request regardless the endpoint
 	URLDenyList []string `yaml:"url_deny_list,omitempty"`
+
+	RequestWithWeights bool `yaml:"request_with_weights,omitempty"`
```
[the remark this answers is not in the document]
javiermolinar: Yeah

modules/frontend/pipeline/async_weight_middleware.go
```diff
@@ -0,0 +1,109 @@
...
+	TraceQLMetrics
+)
+
+const (
+	DefaultWeight       int = 1
+	TraceQLSearchWeight int = 1
+	TraceByIDWeight     int = 2
+
+	maxTraceQLConditions int = 4
+	maxRegexConditions   int = 1
```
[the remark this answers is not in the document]
javiermolinar: I wanted to keep it simple, especially for the config. But since we agree to have its config struct we can make it configurable.

modules/frontend/pipeline/pipeline.go
```diff
@@ -15,18 +15,23 @@ type Request interface {
...
 	ResponseData() any
+	FromHTTPRequest(request *http.Request) *HTTPRequest
 }
 
 type HTTPRequest struct {
 	req *http.Request
 
 	cacheKey     string
 	responseData any
+	weight       int
```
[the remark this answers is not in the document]
joe-elliott: apparently i'm mistaken. i could have sworn there was a clone func. i think you basically just added it with CloneFromHTTPRquest.

modules/frontend/pipeline/pipeline.go
joe-elliott: apologies, see above. i could have sworn there was, but i'm wrong.

javiermolinar (approved):

Issue #4060, opened 2024-09-05
Replace hedged requests roundtrips total with a counter


joe-elliott:
We are currently using a gauge incorrectly to report hedged roundtrips on both external endpoints and backend requests. This has worked in the past but recently we have seen this report erroneous results when `rate()`ing in prometheus. It's possible it never worked correctly and we have only recently noticed. Regardless, it's bad practice and we should fix it.

[External Hedged Requests Gauge](https://github.com/grafana/tempo/blob/f4b2642b16793ed13daf20c43cf37758c4fd6211/modules/querier/external/client.go#L33-L39)

[Backend Hedged Requests Gauge](https://github.com/grafana/tempo/blob/f4b2642b16793ed13daf20c43cf37758c4fd6211/tempodb/backend/instrumentation/hedged_requests.go#L10-L16)

[Shared code in use for both Gauges](https://github.com/grafana/tempo/blob/main/pkg/hedgedmetrics/metrics.go)

Let's replace the two linked gauges above with counters.

galalen:
I would like to work on it

joe-elliott:
Thanks @galalen. Let me know if you have any questions.

PR #4063, opened 2024-09-08
replace hedged requests roundtrips total with counter

galalen:
**What this PR does**:
Replace hedged requests round trips total with a counter
**Which issue(s) this PR fixes**:
Fixes #4060

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #4078, opened 2024-09-12
fix: correctly publish the hedgedmetrics counter with delta

electron0zero:
**What this PR does**:
`hedgedhttp.Stats` is already a counter so we were incrementing the counter by the value of hedgedhttp.Stats counter.

correctly calculate delta and increment the counter for hedgedhttp.Stats

fixes https://github.com/grafana/tempo/issues/4060, we missed this detail of hedgedhttp.Stats in https://github.com/grafana/tempo/pull/4063

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## main / unreleased
 
-* [BUGFIX] Correctly publish the hedgedmetrics counter with delta [#4078](https://github.com/grafana/tempo/pull/4078) (@electron0zero)
-* [BUGFIX] Replace hedged requests roundtrips total with a counter. [#4063](https://github.com/grafana/tempo/pull/4063) (@galalen)
+* [BUGFIX] Replace hedged requests roundtrips total with a counter. [#4063](https://github.com/grafana/tempo/pull/4063) [#4078](https://github.com/grafana/tempo/pull/4078) (@galalen)
 * [CHANGE] TraceByID: don't allow concurrent_shards greater than query_shards. [#4074](https://github.com/grafana/tempo/pull/4074) (@electron0zero)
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## main / unreleased
 
+* [BUGFIX] Correctly publish the hedgedmetrics counter with delta [#4078](https://github.com/grafana/tempo/pull/4078) (@electron0zero)
+* [BUGFIX] Replace hedged requests roundtrips total with a counter. [#4063](https://github.com/grafana/tempo/pull/4063) (@galalen)
```
electron0zero, line 4: this entry was in `# v2.6.0` section, should be in `## main / unreleased` so fixed it.

pkg/hedgedmetrics/metrics.go
```diff
@@ -1,27 +1,63 @@
...
+	value   int64
+	mu      sync.Mutex
+}
+
+func NewCounterWithValue(counter prometheus.Counter) *CounterWithValue {
+	return &CounterWithValue{counter: counter}
+}
+
+func (c *CounterWithValue) Add(v int64) {
+	c.mu.Lock()
```
joe-elliott, line 27: this is only accessed from one goroutine. doesn't need locking.

electron0zero: removed.

electron0zero: I am seeing a DATARACE in the test `TestCounterWithValue` after removing the locking,

joe-elliott: That's b/c in the test you are checking the value of the counter in a different goroutine than is changing the value, but this doesn't happen in the actual application.

joe-elliott (approved):
lgtm, me!

PR #4088, opened 2024-09-16
Fix memcache host for local tempo setup

naitikmalaviya:
**What this PR does**:
Fixes mamcache host path issue for local tempo example which was introduced in https://github.com/grafana/tempo/pull/4032

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott (approved):
Thanks for the fix!

PR #4100, opened 2024-09-20
Speedup collection of results from ingesters in the querier

electron0zero:
**What this PR does**:
Earlier we used to make calls to ingesters, get results and itarate over these results to collect them and then return them to query-frontend.

Iterating and then collecting the results was sync and a bottleneck and showed up in traces and profiles. 

This PR pushed down the collection of results and that makes it a parallel and removes the iteration step.

now we query an ingester and collect the results where we query, and once all ingesters response, we read the collected results and return them to query-frontend for further combining.

see this trace for example:
image

along with these changes, It also does the following:
- removed the type assertions and use of `interface{}` return types from `forIngesterRings` and callback function `forEachFn`
- added locking in ScopedDistinctString and DistinctString collectors to make the safe to use.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/collector/generic_collector.go
```diff
@@ -0,0 +1,33 @@
+package collector
+
+import "sync"
+
+// GenericCollector is a safe collector to collect values
+// and retrieve them later when needed.
+type GenericCollector[T comparable] struct {
```
mdisibio, line 7: It looks like this is only used in one place (SearchRecent).  Don't think we need this abstraction for now, we can easily append to the slice in the callback.

modules/querier/querier.go
```diff
@@ -626,21 +638,25 @@ func (q *Querier) SearchTagValues(ctx context.Context, req *tempopb.SearchTagVal
...
+	forEach := func(ctx context.Context, client tempopb.QuerierClient) error {
+		resp, err := client.SearchTagValues(ctx, req)
+		if err != nil {
+			return err
+		}
+	outerLoop:
+		for _, res := range resp.TagValues {
 			distinctValues.Collect(res)
 			if distinctValues.Exceeded() {
 				break outerLoop // break out of all loops
```
mdisibio, line 650: I know this wasn't changed in this PR, but it could be a simplified to `return nil` to exit the callback, instead of labeling and breaking the outer loop.

modules/querier/querier.go
```diff
@@ -326,23 +324,23 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
 
 type (
-	forEachFn        func(ctx context.Context, client tempopb.QuerierClient) (interface{}, error)
+	forEachFn        func(ctx context.Context, client tempopb.QuerierClient) error
 	replicationSetFn func(r ring.ReadRing) (ring.ReplicationSet, error)
 )
 
 // forIngesterRings runs f, in parallel, for given ingesters
-func (q *Querier) forIngesterRings(ctx context.Context, userID string, getReplicationSet replicationSetFn, f forEachFn) ([]responseFromIngesters, error) {
+func (q *Querier) forIngesterRings(ctx context.Context, userID string, getReplicationSet replicationSetFn, f forEachFn) error {
```
mdisibio, line 332: I really like the new approach of just a simple callback.  It avoids both interface typecasting, even the original thought of making it use generics. 👍

modules/querier/querier.go
```diff
@@ -251,34 +246,37 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
-			return client.FindTraceByID(funcCtx, req)
-		})
-		if err != nil {
-			return nil, fmt.Errorf("error querying ingesters in Querier.FindTraceByID: %w", err)
-		}
-
-		found := false
-		for _, r := range responses {
-			t := r.response.(*tempopb.TraceByIDResponse).Trace
+		forEach := func(funcCtx context.Context, client tempopb.QuerierClient) error {
```
mdisibio, line 254: nit:  Not critical, but what do you think about making these inline funcs again like:

```go
err := q.forIngesterRings(ctx, userID, getRSFn, func(ctx, client) error {
    <callback logic here>
})
```

pkg/collector/generic_collector.go
electron0zero: done, inlined the collection.

modules/querier/querier.go
electron0zero: removed.

modules/querier/querier.go
electron0zero: we can make the similar change in `forGivenGenerators` https://github.com/grafana/tempo/blob/807435d38ac5b78c5626d05d74eb0715081f1066/modules/querier/querier.go#L432 but I left it out to keep this PR focused on ingester changes. 

can make a follow up PR, if you want to make that change :)

modules/querier/querier.go
electron0zero: I moved them out to make them easy to read because now these callback functions is few more lines. 

I personally find it easy to parse out when the callback is outside but I don't have a strong preference for either way, if you think that it's easy to read when inlined, I can inline them.

modules/querier/querier.go
electron0zero: pushed [`e1139f4` (#4119)](https://github.com/grafana/tempo/pull/4119/commits/e1139f4e2200d9353615b58ee33361bab0c64fd8)

PR is in draft because it's branched from this PR's branch.

mdisibio (approved):
👍 LGTM

mapno (approved):

PR #4110, opened 2024-09-22
fix_3811: send semver version in api/stattus/buildinfo for cloud deployments

### LINKED ISSUE #3811: Add semver version to api/status/buildinfo endpoint
[isFeatureAvailable](https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/tempo/datasource.ts#L252) is used in the Tempo front end to check if features should be made available in the UI based on which version of Tempo is running.

The function above (in the link) returns something like “r154-0b3d297” but for semver to work it needs to be something like “2.5.1”
This means that the function above always returns true so it is possible this would enable streaming in an environment where an old version of Tempo is being used.

To return a semver string from the `api/status/buildinfo` endpoint.

Having this semver string would also allow us to check if other features are available and cleans up the code/makes it more resilient on the frontend.

Removing the function above and simply running a streaming query for example and if it fails then falling back to a non-streaming query.

**Additional context**
Screenshot 2024-06-26 at 09 27 28
Screenshot 2024-06-26 at 09 27 32

Aki0x137:
**What this PR does**:
- After every merge to main, a github action will update the VESRION file in root of project with latest tag.
- Added a rule in Makefile for the same.
- `image-tag` tool is updated to validate and use the semver version, if present. Otherwise default to current behavior.

Fixes #3811 

**Checklist**
- [X] Tests updated
- [ ] Documentation added


.github/workflows/update-version.yml
```diff
@@ -0,0 +1,36 @@
...
+        fetch-depth: 0
+
+    - name: Update VERSION file
+      run: make update-version
+
+    - name: Commit and push changes
+      run: |
+        if [ -n "$(git status --porcelain VERSION)" ]; then
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
```
joe-elliott, line 30: @ie-pham do you think this will work with our latest security changes to GHA?

i'll admit i'm not sure where the creds are that allow this to work?

tools/image-tag
```diff
@@ -4,6 +4,27 @@ set -o errexit
...
+            return 0
+      fi
+}
+
+# Check if there is a VERSION file
+if [ -f "${REPO_ROOT}/VERSION" ]; then
+      VERSION=$(cat "${REPO_ROOT}/VERSION" )
+      if is_valid_semver "$VERSION"; then
+            echo "$VERSION"
+            exit 0
```
joe-elliott, line 24: So I believe that changing this script will update the version as expected b/c of this line:

https://github.com/grafana/tempo/blob/main/Makefile#L50

However I think it will also change the output of the script as used here:

https://github.com/grafana/tempo/blob/main/.drone/drone.yml#L12

which I'm pretty sure is how our docker images are tagged:

https://hub.docker.com/r/grafana/tempo/tags

@zalegrala do you know off the top of your head if this is how it works?

joe-elliott (commented):
Thanks for looking into this! 

I think it's close but I'm afraid it will break the way we've named our docker images. I've pinged a few team members who I think will have more a immediate understanding of how this works.

Aki0x137:
> Thanks for looking into this!
> 
> I think it's close but I'm afraid it will break the way we've named our docker images. I've pinged a few team members who I think will have more a immediate understanding of how this works.

will check this. Didn't know same hash is used for docker images. Thanks for the review!

tools/image-tag
Aki0x137: question: is it fine if there is a separate `tools/version-tag`  shell script which is used in Makefile to update tag, while the existing `tools/image-tag` can be left unmodified for creating docker image id?

Thanks for higjlighting this. I didn't check this flow tbh.

joe-elliott: yeah. splitting the script into two for simplicity makes sense to me

Aki0x137: done

.github/workflows/update-version.yml
ie-pham: In terms of the GHA, I think this looks fine as it doesn't require any vault secrets. However, I am unsure if it is possible to push directly to main without a pull request.

VERSION
```diff
@@ -0,0 +1 @@
+v2.6.0
```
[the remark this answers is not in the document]
Aki0x137: Yup, file is not needed. I referenced another Grafana repo mentioned on issue where they are using a VERSION file. Made the requested changes.

.github/workflows/update-version.yml
Aki0x137: New GHA is not needed after fixing the comment from @zalegrala 👍

.github/workflows/ci.yml
```diff
@@ -79,6 +79,8 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v4
+      - name: Fetch tags
```
Aki0x137, line 82: Callout: @joe-elliott Since checkout GHA runs by default with with detached head and doesn't pull tags, I had to modify the CI for the test case I wrote to pass.
The directive: `with: fetch-tags: true` doesn't work because of an [issue](https://github.com/actions/checkout/issues/1471) on GHA, open from few years.

joe-elliott: It looks like [integration tests are failing](https://github.com/grafana/tempo/actions/runs/11192788791/job/31179779452?pr=4110#step:5:955) right now. Is that what you're referring to?

Why does this not work? I would expect GHA to use the changes in the branch when running CI.

Aki0x137: yup, the test case that failed is the one I added. The shell script's file permission change wasn't added so it failed. Added the change and tested out the GHA on my fork. The test should pass now.

joe-elliott (approved):
very cool improvement. thank you 🙏

Aki0x137:
Thanks for the patience with this PR😅🙏

PR #4119, opened 2024-09-24
Pushdown collection of results from generators in the querier

electron0zero:
**What this PR does**:

Iterating and then collecting the results was sync and a bottleneck for the ingester queries, and we fixed that in https://github.com/grafana/tempo/pull/4100. we did it by pushing down the collection of results and that makes it a parallel and removes the iteration step.

we can do the same for the call to generators in the queries and make them fast.

this is a follow up on https://github.com/grafana/tempo/pull/4100, as discussed https://github.com/grafana/tempo/pull/4100#discussion_r1772783985

along with these changes, removed the type assertions and use of `interface{}` return types from `forGivenGenerators` and callback function `forEachGeneratorFn`

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


mapno (approved):
LGTM

mdisibio (approved):

PR #4140, opened 2024-09-28
Init histogram counters with 0

mapno:
**What this PR does**:

Pushes a 0 to classic histogram's counter when the series is new to allow Prometheus to start from a non-null value.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/registry/histogram.go
+++ b/modules/generator/registry/histogram.go
@@ -44,6 +44,19 @@ type histogramSeries struct {
 	exemplars      []*atomic.String
 	exemplarValues []*atomic.Float64
 	lastUpdated    *atomic.Int64
+	// firstSeries is used to track if this series is new to the counter.  This
+	// is used to ensure that new counters being with 0, and then are incremented
+	// to the desired value.  This avoids Prometheus throwing away the first
+	// value in the series, due to the transition from null -> x.
+	firstSeries *atomic.Bool
+}
+
+func (hs *histogramSeries) isNew() bool {
+	return hs.firstSeries.Load()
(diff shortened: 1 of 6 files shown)


modules/generator/registry/histogram.go
```diff
@@ -171,16 +187,27 @@ func (h *histogram) collectMetrics(appender storage.Appender, timeMs int64, exte
 			lb.Set(name, s.labels.values[i])
 		}
 
+		// If we are about to call Append for the first time on a series,
+		// we need to first insert a 0 value to allow Prometheus to start from a non-null value.
+		if s.isNew() {
+			lb.Set(labels.MetricName, h.nameCount)
```
javiermolinar, line 193: Question, maybe it makes sense to create a new ìnit` method in the interface:

```go
type metric interface {
	name() string
        initMetrics(appender storage.Appender,...)
	collectMetrics(appender storage.Appender, timeMs int64, externalLabels map[string]string) (activeSeries int, err error)
}
```
since I see we do the same logic in the counter

mapno: This logic is actually of the series, not the metric. It needs passing a couple of values, the function would actually encapsulate very little. I don't think it's worth the hassle.

```go
func (co *counterSeries) init(appender storage.Appender, lbls labels.Labels, ts int64) {
	_, _ = appender.Append(0, lbls, ts, 0)
	co.registerSeenSeries()
}
```

electron0zero (approved):

PR #4142, opened 2024-09-30
Fix typo

vrebol:
If insecure is set to "true", the authentication and certificate checks for gcs are disabled rather than enabled.


mapno (approved):

knylander-grafana (approved):
Thank you for updating the docs

PR #4144, opened 2024-09-30
Correctly handle 400 Bad Request and 404 Not Found in gRPC streaming

mapno:
**What this PR does**:

Correctly handle 400 Bad Request and 404 Not Found in gRPC streaming.

Status codes in gRPC are changed to return `InvalidArgument` for 400 and `NotFound` for 404. Other non-special HTTP status codes are `Unknown`.

This also allows fixing how 400 responses are counted towards SLO calculations.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/search_handlers.go
+++ b/modules/frontend/search_handlers.go
@@ -139,6 +139,8 @@ func logResult(logger log.Logger, tenantID string, durationSeconds float64, req 
 	statusCode := -1
 	if httpResp != nil {
 		statusCode = httpResp.StatusCode
+	} else if st, ok := status.FromError(err); ok {
+		statusCode = int(st.Code())
 	}
 
 	if resp == nil {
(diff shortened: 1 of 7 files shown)


modules/frontend/search_handlers.go
```diff
@@ -139,6 +139,8 @@ func logResult(logger log.Logger, tenantID string, durationSeconds float64, req
 	statusCode := -1
 	if httpResp != nil {
 		statusCode = httpResp.StatusCode
+	} else if st, ok := status.FromError(err); ok {
```
joe-elliott, line 142: is httpResp nil in this case?

we may also want to add this same logic to the other log methods

joe-elliott (approved):
one comment to just add the logging logic to all the streaming handlers, but other than that lgTm

PR #4152, opened 2024-10-02
use slices and map from the go stdlib instead of golang.org/x/exp

dastrobu:
**What this PR does**:

use slices and map from the go stdlib instead of golang.org/x/exp

`nil`

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/overrides/runtime_config_overrides.go
+++ b/modules/overrides/runtime_config_overrides.go
@@ -5,12 +5,13 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"maps"
 	"net/http"
+	"slices"
 	"time"
 
 	"github.com/drone/envsubst"
 	"github.com/go-kit/log/level"
-	"golang.org/x/exp/maps"
 
 	"github.com/grafana/dskit/runtimeconfig"
 	"github.com/grafana/dskit/services"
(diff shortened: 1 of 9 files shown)


dastrobu:
As we have go 1.23 now, I guess it would be nice to prefer the stdlib over `golang/x/exp` for maps and slices. 

I did not add this to the CHANGELOG yet, as I am not sure if you want this kind of cleanup mentioned there.

modules/overrides/runtime_config_overrides.go
```diff
@@ -277,7 +278,7 @@ func (o *runtimeConfigOverridesManager) GetTenantIDs() []string {
 		return nil
 	}
 
-	return maps.Keys(tenantOverrides.TenantLimits)
+	return slices.Collect(maps.Keys(tenantOverrides.TenantLimits))
```
joe-elliott, line 281: it seems like `slices.Collect` calls result in repeated appends to to an empty slice which is a bit unfortunate.

https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/slices/iter.go;l=49-54

wdyt about pre-allocing the slice to the appropriate capacity from the map and then using AppendSeq?

joe-elliott (commented):
Thanks for the cleanup. Looking for your thoughts on a small perf improvement.

modules/overrides/runtime_config_overrides.go
dastrobu: not sure if this is going to be measurable, but why not. At least that should be closer to https://cs.opensource.google/go/x/exp/+/701f63a6:maps/maps.go;l=21

joe-elliott (approved):
Thanks!

PR #4160, opened 2024-10-04
Avoid returning 500 for TRACE_TOO_LARGE

zalegrala:
**What this PR does**:

Here we make an adjustment to the return value at the querier and query-frontend for the case where a trace is larger than the max configured size.  Previously this was reported as a 500 status code and has now been changed to a 422.  This will allow these conditions to be excluded from SLO failures, which is more appropriate since this is based on a configured max trace size.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/combiner/trace_by_id.go
+++ b/modules/frontend/combiner/trace_by_id.go
@@ -87,7 +87,7 @@ func (c *traceByIDCombiner) AddResponse(r PipelineResponse) error {
 
 	if errors.Is(err, trace.ErrTraceTooLarge) {
 		c.code = http.StatusUnprocessableEntity
-		c.statusMessage = trace.ErrTraceTooLarge.Error()
+		c.statusMessage =  fmt.Sprint(err)
 		return nil
 	}
 
--- a/modules/frontend/combiner/trace_by_id.go
+++ b/modules/frontend/combiner/trace_by_id.go
@@ -86,7 +86,7 @@ func (c *traceByIDCombiner) AddResponse(r PipelineResponse) error {
 	_, err = c.c.Consume(resp.Trace)
 
 	if errors.Is(err, trace.ErrTraceTooLarge) {
-		c.code = http.StatusRequestEntityTooLarge
+		c.code = http.StatusUnprocessableEntity
 		c.statusMessage = trace.ErrTraceTooLarge.Error()
 		return nil
 	}
@@ -164,8 +164,8 @@ func (c *traceByIDCombiner) shouldQuit() bool {
 		return false
 	}
 
-	// test special case for 413
-	if c.code == http.StatusRequestEntityTooLarge {
+	// test special case for 422
+	if c.code == http.StatusUnprocessableEntity {
 		return true
 	}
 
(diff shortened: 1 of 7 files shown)


modules/frontend/handler.go
```diff
@@ -163,6 +164,8 @@ func writeError(w http.ResponseWriter, err error) error {
 		err = errDeadlineExceeded
 	} else if isRequestBodyTooLarge(err) {
 		err = errRequestEntityTooLarge
+	} else if isTraceTooLarge(err) {
```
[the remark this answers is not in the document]
zalegrala: Nice, that looks a bit cleaner to me.  Just pushed a commit for this.

modules/frontend/combiner/trace_by_id.go
```diff
@@ -156,6 +163,11 @@ func (c *traceByIDCombiner) shouldQuit() bool {
 		return false
 	}
 
+	// test special case for 413
+	if c.code == http.StatusRequestEntityTooLarge {
+		return false
```
joe-elliott, line 168: if the trace is too large, shouldn't we quit?

joe-elliott (commented):
oh! changelog

modules/frontend/combiner/trace_by_id.go
```diff
@@ -83,6 +84,12 @@ func (c *traceByIDCombiner) AddResponse(r PipelineResponse) error {
 
 	// Consume the trace
 	_, err = c.c.Consume(resp.Trace)
+
+	if errors.Is(err, trace.ErrTraceTooLarge) {
```
javiermolinar, line 88: Maybe I'm testing it wrongly but I get a 500 if the frontend combiner returns an ErrTraceTooLarge here.

The error is handled here: https://github.com/grafana/tempo/blob/280128f7b740919a7e1c3ecc0ef9fd6b9843ca41/modules/frontend/handler.go#L85

javiermolinar: ServeHttp -> RoundTrip -> TraceByIDHandler 

The response from the TraceByIdHandler is (nil, ErrTraceTooLarge)

joe-elliott: oh yeah, we need to return nil in this if block. the issue is that it's falling through and still returning the error on line 93

modules/frontend/combiner/trace_by_id.go
zalegrala: I'm aligning with the comment here, which was counter to my intuition. 
 https://github.com/zalegrala/tempo/blob/0b86b085a23a07c16ed10e4137e349175c100ab0/modules/frontend/combiner/trace_by_id_test.go#L51

joe-elliott: yeah, ignore that test, we should quit 👍 

at this point the combiner has accepted as much as it will, by returning quit we cause the pipeline to stop farming out jobs to queriers and to return its results

joe-elliott (approved):

joe-elliott:
nice one 👍

javiermolinar (commented):
LGTM!

PR #4162, opened 2024-10-07
Feature:  Distributor usage trackers

mdisibio:
**What this PR does**:
This a new feature that allows a tenant to accurately track the amount of ingested traffic by a set of custom labels. It's similar to the existing `traces_spanmetrics_size_total` metric created by the generators but improves on it in some key ways.  

** Need **
The core need is to export a set of highly accurate metrics on ingested traffic that tenants can use for _cost-attribution_.  This means that every ingested byte can be attributed to something, i.e. a team or department, so that tracing costs can be reconciled.  Any attribute in the tracing data can be used. 

** Reasons for a new feature**:

1. The existing size metric isn't accurate enough. It doesn't include non-span data (i.e. resources and scopes). This can be significant, typically 15+% of the total payload.  However it can also not be fixed because of the way input data is sharded across the generator rings.  Each time a batch is split by trace ID, the non-span data is duplicated (the resource-level information is duplicated for each generator target in order to ensure an internally-consistent payload).  Trying to account for it then errs on the other side and over-counts the non-span data (85% -> 115%). Therefore we needed a new approach which is 99+% accurate. The only component in Tempo which has the original payload is the distributor so it is the ideal location to add this functionality.
2. The labels for usage tracking need to be separately configurable from span metrics.  Span metrics typically includes labels such as http url or status code, span success/failure, database targets. This level of detail is fine-grained and geared towards operational needs, which is separate from cost-attribution and cost reconciliation.
3. Rename and combine - To bridge gaps in instrumentation and provide value without application changes, we support the ability to rename and combine different attributes into the same output bucket. The prime case for this is to bridge the gap between open-telemetry-style semantic conventions (`service.name`) and historical conventions (`app`). The per-tenant dimensions are a map of input->output.  If multiple inputs correspond to the same output then they are added together.

** Important concepts about this new feature **
1. This lays the foundation in the distributor for generic _trackers_ in the future. The only one now is cost-attribution, and is controlled by per-tenant overrides.   Examples of other trackers are helpful things like tracking the adoption of instrumentation libraries, databases, etc.  
2. The trackers are exposed on a new endpoint `/usage_metrics`.  This is so that they aren't mixed with the existing operational `/metrics`, because they are expected to have much higher cardinality and a different purpose.
4.  Significant work went into the algorithm for measuring non-span data correctly and fairly.  Tracing payloads are composed of a batch with resource attributes and many spans with their own attributes (simplifying greatly here).  A span is always matched to a single category, but the non-span data cannot (the ~15% of data).  Therefore we split it proportionally based on the assignment of spans.   **Example** If the batch contains 10 spans with "foo", and 5 spans with "bar", then the category "foo" will get 67% of the non-span bytes, and "bar" will get 33%.
5. This adds overhead to the write-path of the distributor.  But it should be minimal.  It is effectively a single additional call to `proto.Size()`, and the series tracking contains buffering that enables it to be zero-allocation for existing series.  If you run the benchmark you will see:  
```BenchmarkUsageTracker 	 1995282	      5904 ns/op	       0 B/op	       0 allocs/op```
 
**TODOs**
~There is some remaining questions:~
~1. Behavior for unconfigured tenants.  Currently it's opt-in:  if `overrides.cost_attribution.dimensions` isn't set then the distributor records nothing.  But maybe it makes sense to have default behavior here?  Service name may be a good default.~
~2. Behavior when hitting against max cardinality.  Currently if we already have the max number of series, it records the data into the _unlabeled_ series.  But the existing series keep working. This means that it is not possible to assess the data quality after max cardinality is reached.  Thinking about alternative behaviors here.~

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added

### CODE REVIEW

**Comment by knylander-grafana** [comment]
Should we create a doc issue for this when it's ready?

File: modules/distributor/usage/tracker.go [hunk 1]
```diff
@@ -0,0 +1,371 @@
...
+	labels     []string // Sanitized dimensions => final labels
+	values     []string // Used to capture values and guaranteed to be a matching length.
+}
+
+// GetBuffersForDimensions takes advantage of the fact that the configuration for a tracker
+// changes slowly.  Reuses buffers from the previous call when the dimensions are the same.
+func (t *tenantUsage) GetBuffersForDimensions(dimensions []string) ([]string, []string) {
+	reset := false
+
+	if len(dimensions) == len(t.dimensions) {
```
joe-elliott, line 57: use `slices.Equal` ?

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,371 @@
...
+func (t *tenantUsage) getSeries(labels, values []string, maxCardinality int) *bucket {
+	h := hash(labels, values)
+
+	b := t.series[h]
+	if b == nil {
+		// Before creating a new series, check for cardinality limit.
+		if len(t.series) >= maxCardinality {
+			// Overflow
+			// It goes into the unlabeled bucket
+			// TODO - Do we want to do something else?
```
joe-elliott, line 90: i really like this solution and can't come up with a better one off the top of my head. you can clearly see the severity of the unknown data and choose whether or not to pick a different label set.

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,371 @@
...
+				if i == 0 && j == 0 {
+					sz += firstSpanPortion
+				}
+
+				for _, a := range s.Attributes {
+					v := a.Value.GetStringValue()
+					if v == "" {
+						continue
+					}
+					for i, d := range dimensions {
```
joe-elliott, line 226: so it looks like span data "overrides" resource level data. should we require scopes on the configured dimensions to enforce level?

I'm imagining a scenario like:
```
batches: [
  atts: {
    "foo": "A",
  },
  spans: [
    { 
      atts: {
        "foo": "B",
      },
    },
    {
       // no value for foo
    },
  ]
]
```

will the first span overwrite the foo value to be "B" and that will carry forward to the second span even though it doesn't have a value for "foo"?

joe-elliott (commented):
core code looks great. have some Qs but mainly just needs changelog and docs

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,381 @@
...
+}
+
+func (u *Tracker) Observe(tenant string, batches []*v1.ResourceSpans) {
+	u.mtx.Lock()
+	defer u.mtx.Unlock()
+
+	dimensions := u.labelsFn(tenant)
+	if len(dimensions) == 0 {
+		// Not configured
+		// TODO - Should we put it all in the unattributed bucket instead?
```
electron0zero, line 166: I would lean on the side of defaulting to service name, that seems more useful then the unattributed bucket 🤔

modules/overrides/config.go
```diff
@@ -194,7 +199,8 @@ type Overrides struct {
 	// Global enforced overrides.
 	Global GlobalOverrides `yaml:"global,omitempty" json:"global,omitempty"`
 	// Storage enforced overrides.
-	Storage StorageOverrides `yaml:"storage,omitempty" json:"storage,omitempty"`
+	Storage         StorageOverrides         `yaml:"storage,omitempty" json:"storage,omitempty"`
+	CostAttribution CostAttributionOverrides `yaml:"cost_attribution,omitempty" json:"cost_attribution,omitempty"`
```
electron0zero, line 203: we have been calling this `usage tracker` or `usage metrics` at other places but Cost Attribution in the overrides. 

let's use `usage tracker` or `usage metrics` here to keep it consistent and avoid confusion.

modules/overrides/config_legacy.go
```diff
@@ -134,6 +134,8 @@ type LegacyOverrides struct {
 	//  is not used when doing a trace by id lookup.
 	MaxBytesPerTrace int `yaml:"max_bytes_per_trace" json:"max_bytes_per_trace"`
 
+	CostAttribution CostAttributionOverrides `yaml:"cost_attribution,omitempty" json:"cost_attribution,omitempty"`
```
electron0zero, line 137: consistent naming, same as above?

modules/overrides/interface.go
```diff
@@ -76,6 +76,8 @@ type Interface interface {
 	MaxMetricsDuration(userID string) time.Duration
 	DedicatedColumns(userID string) backend.DedicatedColumns
 	UnsafeQueryHints(userID string) bool
+	CostAttributionMaxCardinality(userID string) uint64
```
electron0zero, line 79: CostAttributionMaxCardinality -> UsageMetricsMaxCardinality or UsageTrackerMaxCardinality?


CostAttributionDimensions -> UsageMetricsDimensions or UsageTrackerDimensions?

modules/overrides/userconfigurable/client/limits.go
```diff
@@ -8,8 +8,8 @@ import (
 )
 
 type Limits struct {
-	Forwarders *[]string `yaml:"forwarders,omitempty" json:"forwarders,omitempty"`
-
+	Forwarders       *[]string              `yaml:"forwarders,omitempty" json:"forwarders,omitempty"`
+	CostAttribution  CostAttribution        `yaml:"cost_attribution,omitempty" json:"cost_attribution,omitempty"`
```
electron0zero, line 12: same as above

modules/overrides/user_configurable_overrides.go
```diff
@@ -222,6 +222,13 @@ func (o *userConfigurableOverridesManager) Forwarders(userID string) []string {
 	return o.Interface.Forwarders(userID)
 }
 
+func (o *userConfigurableOverridesManager) CostAttributionDimensions(userID string) []string {
```
electron0zero, line 225: iiuc, we are only exposing `Dimensions` in user configurable overrides and MaxCardinality can only be set via runtime overrides?

is my understanding correct?

cmd/tempo/app/modules.go
```diff
@@ -243,6 +243,8 @@ func (t *App) initDistributor() (services.Service, error) {
 		t.Server.HTTPRouter().Handle("/distributor/ring", distributor.DistributorRing)
 	}
 
+	t.Server.HTTPRouter().Handle("/usage_metrics", distributor.UsageTrackerHandler())
```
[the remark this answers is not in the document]
mdisibio: Yes it results in a 404. But I can wrap it in an if statement if think there is any brittleness there.

modules/distributor/usage/tracker.go
mdisibio: Definitely considered this, but for this use case we need to check both areas, as different applications can be instrumented differently (service A puts the attribute on the resource, and service B puts the attribute on the span).  Was thinking we could evolve into that in the future, parsing as a traceql identifier allows us to know if it's a plain old label (current), or scoped (span.foo) or even unscoped again (.foo).

Another consideration was to keep the configuration simple, and consistent across signals. For example here is the PR for Mimir: https://github.com/grafana/mimir/pull/9392

Ultimately I think at this time I wouldn't want to require scopes yet.

modules/overrides/user_configurable_overrides.go
mdisibio: Yes that should be the case.  Dimensions are user-configurable, whereas MaxCardinality is still per-tenant, but controlled by the operator via runtime overrides.

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,371 @@
...
+const (
+	tenantLabel  = "tenant"
+	trackerLabel = "tracker"
+)
+
+var emptyHash = hash(nil, nil)
+
+type tenantLabelsFunc func(string) []string
+
+type bucket struct {
```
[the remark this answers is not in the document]
mdisibio: Agree, it didn't seem worth pulling it out of the generator, since this only needs counters, but also I wanted to try to tweak the structure a little bit.  It is precomputing the metric description up-front instead of storing the key=value labels.

modules/overrides/config.go
electron0zero: ah, please ignore my comments then.

modules/distributor/usage/tracker.go
mdisibio: Ah sorry reading closer, yes I think you are right and that is a bug.   It looks like "B" would carry forward to the second span.

mdisibio:
> Should we create a doc issue for this when it's ready?

I'm happy to add docs in this PR if we want. At the minimum I will update the config and url sections.

modules/distributor/usage/tracker.go
colega: Whatever is decided here, it should be the same for all our DBs.

mdisibio:
Pushed some changes to configure via `map[string]string]` instead of `[]string`, to handle the case of many-to-one dimensions with relabel.  This allows for the flexibility to accommodate gaps across a tenant's data.  For example we could now scan for both labels `k8s.namespace.name` and `app_namespace`, and combine both into a final `namespace` label.  There are still some outstanding changes TODO.

modules/distributor/usage/tracker.go
mdisibio: This is fixed and added a test case.   Did take a small regression in performance because basically it becomes harder to detect/reuse the last hashed series across spans.

modules/distributor/usage/tracker.go
mdisibio: This behavior was changed slightly.  Instead of going to an unlabeled bucket, it substitutes `__overflow__` for every configured dimension. So if your dimensions are "foo" and "bar", the overflow series will be `{foo="__overflow__", bar="__overflow__"}`.  This makes it more explicit, and simplifies the downstream querying and reporting (and the same approach is being taken in the other databases).

mdisibio:
Pushed new behavior for overflow and missing labels:
* Missing: we now always output a label=value pair for each configured dimension. If both the span and batch are missing this value, then it is set to dimension="__missing__".  Each dimension is handled separately, so dimensionA might be populated, and dimensionB could be "__missing__".
* Overflow:  when max-cardinality is reached it now goes into the series with each dimension="__overflow__". Previously it went to the unlabeled series. 

These two changes ensure that the output metrics always have the expected labels, and are consistent across edge cases.

modules/distributor/usage/tracker.go
mdisibio: I think for now we won't have any default behavior. It is still possible to configure default behavior yourself, by setting the default-tenant overrides (tenantID=`*`).  The tracker is also disabled by default since it is only relevant to large multi-tenant scenarios.

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,423 @@
...
+	labels []string
+
+	// Runtime data
+	bytes       uint64
+	lastUpdated int64
+}
+
+func (b *bucket) Inc(bytes uint64, unix int64) {
+	b.bytes += bytes
+	b.lastUpdated = unix
```
javiermolinar, line 44: Maybe we could skip passing the Unix timestamp from the outside

b.lastUpdated = time.Now().Unix()

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,423 @@
...
+type Tracker struct {
+	mtx      sync.Mutex
+	name     string
+	tenants  map[string]*tenantUsage
+	labelsFn tenantLabelsFunc
+	maxFn    tenantMaxFunc
+	reg      *prometheus.Registry
+	cfg      PerTrackerConfig
+}
+
```
javiermolinar, line 173: We can check that it fulfills the interface

```go
var _ prometheus.Collector = (*Tracker)(nil)
```

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,423 @@
...
+			descr:  prometheus.NewDesc("tempo_usage_tracker_bytes_received_total", "bytes total received with these attributes", t.sortedKeys, t.constLabels),
+			labels: v,
+		}
+		t.series[h] = b
+	}
+	return b
+}
+
+type Tracker struct {
+	mtx      sync.Mutex
```
javiermolinar, line 165: Is the mutex required for another concurrent operation besides getting the tenantUsage? Maybe we could use a sync.Map instead? It should be faster b/c this is a read-heavy

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,423 @@
...
+	dimensions := u.labelsFn(tenant)
+	if len(dimensions) == 0 {
+		// Not configured
+		// TODO - Should we put it all in the unattributed bucket instead?
+		return
+	}
+
+	max := u.maxFn(tenant)
+	if max == 0 {
+		max = u.cfg.MaxCardinality
```
javiermolinar, line 223: what happens if  `max > u.cfg.MaxCardinality`?

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,423 @@
...
+	buffer1    []string          // Batch-level values
+	buffer2    []string          // Span-level values
+	buffer3    []string          // Last hashed values
+	overflow   uint64
+}
+
+// GetBuffersForDimensions takes advantage of the fact that the configuration for a tracker
+// changes slowly.  Reuses buffers from the previous call when the dimensions are the same.
+func (t *tenantUsage) GetBuffersForDimensions(dimensions map[string]string) ([]mapping, []string, []string, []string) {
+	if !maps.Equal(dimensions, t.dimensions) {
```
javiermolinar, line 69: I wonder if we could have like a lastUpdated timestamp in the runtimeConfigOverridesManager so we could know if the runtime has changed or not

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,423 @@
...
+		}
+		slices.Sort(t.sortedKeys)
+
+		// Step 3
+		// Prepare the mapping from raw attribute names to the final location of
+		// where it goes in the output buffers. This avoids another layer of indirection.
+		t.mapping = t.mapping[:0]
+		for k := range dimensions {
+			i := slices.Index(t.sortedKeys, sanitizedDimensions[k])
+			t.mapping = append(t.mapping, mapping{
```
javiermolinar, line 110: Is it not better to use a map for that? the search is O(1) vs O(n), and less code

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,423 @@
...
+		total += sz + protoLengthMath(sz)
+	}
+
+	l := len(batch.SchemaUrl)
+	if l > 0 {
+		total += l + protoLengthMath(l)
+	}
+
+	for _, ss := range batch.ScopeSpans {
+		// This is the data to store the prescence of this ss
```
javiermolinar, line 400: presence*

modules/distributor/usage/tracker.go
mdisibio: This is effectively called for every span.  `time.Now()` is fairly fast but it's not free, and this is more efficient.

modules/distributor/usage/tracker.go
mdisibio: Yes, that is more how [enterprise-logs does it](https://github.com/grafana/enterprise-logs/blob/main/pkg/enterprise/loki/usage/module.go#L108).  Where the config changes are "pushed" to the tracker, which lets us remove this check from the hot-path.   However I didn't go this route as it seemed like it would be a large change in how our overrides work.  I'm open to doing it this way in a separate PR, if this proves to be a bottleneck.

modules/distributor/usage/tracker.go
mdisibio: Iterating small arrays (5 or so) is actually faster than the map.  We use this technique in other places.  A couple links (not necessarily the best examples), searching the blocks [here](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet4/block_traceql.go#L137), and also [here](https://github.com/grafana/dskit/blob/main/ring/replication_set.go#L548).

modules/distributor/usage/tracker.go
mdisibio: Good question.  Both of these cardinality limits are set by the Tempo operator, either in the distributor config, or the runtime per-tenant config.   It does not come from the user-configurable overrides, so in that sense, both of these limits are "safe".   The question is which one takes precedence.   I was thinking of the distributor limit being the default, and not a server-side maximum.   This way it's still possible to set 1 tenant higher if needed.

modules/distributor/usage/tracker.go
mdisibio: Yes the mutex is needed to handle other concurrent pushes received by the distributor.  I actually expect it to be write-heavy.  All received traffic will be written to the metrics, and scraping will be less frequent (15s or 1m).   

Regarding sync.Map, we've tried it in a few areas and have found plain map + mutex to provide better performance. The ingester live-traces is good example.   I think it's the case where the usage doesn't fit the optimal cases described in the [docs](https://pkg.go.dev/sync#Map):  read-heavy, or writing disjoint sets of keys.  In this case I expect it to be write-heavy, and each tenant is a separate map, so is generally writing the same keys.

If this is shown to be a bottleneck, another option would be 2-layer mutexes, one to guard the tenant map, and then another per-tenant mutex within.  Generators are using this approach: [registry mutex](https://github.com/grafana/tempo/blob/main/modules/generator/registry/registry.go#L66) and [series mutex](https://github.com/grafana/tempo/blob/main/modules/generator/registry/histogram.go#L27)

modules/distributor/usage/tracker.go
javiermolinar: Cool, good to know!

modules/distributor/usage/tracker.go
javiermolinar: Thank you Marty, your explanation has been very enlightening

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,425 @@
...
+			delete(u.tenants, t)
+		}
+	}
+}
+
+func (u *Tracker) Handler() http.Handler {
+	return promhttp.HandlerFor(u.reg, promhttp.HandlerOpts{})
+}
+
+func (u *Tracker) Describe(chan<- *prometheus.Desc) {
```
zalegrala, line 349: Do we want to write the `desc` here?

zalegrala (approved):
Solid work.

modules/distributor/distributor.go
```diff
@@ -328,6 +340,7 @@ func (d *Distributor) PushTraces(ctx context.Context, traces ptrace.Traces) (*te
 		return &tempopb.PushResponse{}, nil
 	}
 	// check limits
+	// todo - usage tracker include discarded bytes?
```
joe-elliott, line 343: i believe the answer is no? we don't want it to included discarded?

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,425 @@
...
+				from: k,
+				to:   i,
+			})
+		}
+
+		// Step 4
+		// Prepopulate the buffers and precompute the overflow bucket
+		t.buffer1 = make([]string, len(t.sortedKeys))
+		t.buffer2 = make([]string, len(t.sortedKeys))
+		t.buffer3 = make([]string, len(t.sortedKeys))
```
joe-elliott, line 120: reuse existing buffer? i think i'm just getting mixed messages about allocs b/c we're reusing t.sortedKeys but none of the other fields.

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,425 @@
...
+	b := t.series[h]
+	if b == nil {
+		// Before creating a new series, check for cardinality limit.
+		if uint64(len(t.series)) >= maxCardinality {
+			// Overflow
+			// This tenant is at the maximum number of series.  In this case all data
+			// goes into the final overflow bucket. It has the same dimensions as the
+			// current configuration, except every label is overridden to the special overflow value.
+			for k := range buffer {
+				buffer[k] = overflowLabel
```
joe-elliott, line 142: calling code is ok with the passed buffer being overwritten?

joe-elliott (commented):
partway done. will review the Observe method next

docs/sources/tempo/configuration/_index.md
```diff
@@ -228,6 +228,19 @@ distributor:
...
+    # Configures usage trackers in the distributor which expose metrics of ingested traffic grouped by configurable
+    # attributes exposed on /usage_metrics.
+    usage:
+        cost_attribution:
+            # Enables the "cost-attribution" usage tracker. Per-tenant attributes are configured in overrides.
+            [enabled: <boolean> | default = false]
+            # Maximum number of series per tenant.
+            [max_cardinality: <int> | default = 10000]
+            # Interval after which a series is considered stale and will be deleted from the registry.
+            # Once a metrics series is deleted it won't be emitted anymore, keeping active series low.
```
knylander-grafana, line 242:
```suggestion
            # Once a metrics series is deleted, it won't be emitted anymore, keeping active series low.
```

knylander-grafana (commented):
Thank you for updating the doc. At a future time, we should convert all of the links in the API doc from relrefs to standard links. We've changed our recommended way to do linking a while back.

modules/distributor/usage/tracker.go
joe-elliott: nevermind. i see it used below. i thought it was local only to this method.

modules/distributor/usage/tracker.go
joe-elliott: I think the the default behavior of nothing is the right choice. In a large multi-tenant cell we may only want this turned on for a few tenants.

Should we move this outside lock since it's called on every push and the normal case will likely be to return here?

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,425 @@
...
+			}
+			sanitized = strutil.SanitizeFullLabelName(v)
+			sanitizedDimensions[k] = sanitized
+		}
+
+		// Step 2
+		// Build the final list of sorted/distinct outputs
+		t.sortedKeys = t.sortedKeys[:0]
+		for _, v := range sanitizedDimensions {
+			if !slices.Contains(t.sortedKeys, v) {
```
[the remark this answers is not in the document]
mdisibio: Yes combining/summing the outputs is intentional.  Tried to explain in: https://github.com/grafana/tempo/pull/4162#issuecomment-2405814389

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,425 @@
...
+		now                             = time.Now().Unix()
+		data                            = u.getTenant(tenant)
+		mapping, buffer1, buffer2, last = data.GetBuffersForDimensions(dimensions)
+	)
+
+	for _, batch := range batches {
+		unaccountedForBatchData, totalSpanCount := nonSpanDataLength(batch)
+
+		if totalSpanCount == 0 {
+			// Mainly to prevent a panic below, but is this even possible?
```
joe-elliott, line 236: i think technically yes, so this is good guard code. i'm imagining someone writing a custom client to push to tempo or perhaps a bug in one of the otel clients creating this situation.

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,425 @@
...
+			buffer1[k] = missingLabel
+		}
+
+		if batch.Resource != nil {
+			for _, a := range batch.Resource.Attributes {
+				v := a.Value.GetStringValue()
+				if v == "" {
+					continue
+				}
+				for _, m := range mapping {
```
joe-elliott, line 262: should the mapping be the driving loop? in general i think we'd have fewer mappings then attributes in a batch

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,425 @@
...
+				sz += protoLengthMath(sz)
+				sz += batchPortion // Incrementally add 1/Nth worth of the unaccounted for batch data
+				if i == 0 && j == 0 {
+					sz += firstSpanPortion
+				}
+
+				// Reset to batch values to for some spans having missing values.
+				copy(buffer2, buffer1)
+
+				for _, a := range s.Attributes {
```
joe-elliott, line 285: similar thoughts as above

modules/distributor/usage/tracker.go
mdisibio: It is, and actually requires it to ensure that the caller optimization to skip unnecessary bucket rehashing works.   Looking at it now, this area could be simplified a bit more.  For example by pre-creating the overflow bucket, and hiding the complexity of the rehashing by moving it around.  However I'm good with the code in its current state.

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,425 @@
...
+				// If the values buffer hasn't changed then we
+				// know it's the same bucket and avoid hashing again.
+				// This shows up in 2 common cases:
+				//  - Dimensions are only resource attributes
+				//  - Runs of spans with the same attributes
+				// NOTE - Not happy about the slices.Equal to detect when
+				// to rehash, but couldn't figure out a better way for now.
+				// The difficulty is tracking bucket dirty status while
+				// resetting to batch values and recording the span values.
+				if bucket == nil || !slices.Equal(buffer2, last) {
```
joe-elliott, line 308: i feel like the normal case here is that people will configure resource level attributes only so we will almost always use the previously grabbed bucket.

the best idea i have to dodge `slices.Equal` is if we configured each dimension with resource or span level. then we would only search those attributes that could match and we'd know if we needed to look in span level at all. 

that might not work with the broader goals of the project though if we intend people to specify attribute names only w/o the overhead of knowing its span or resource.

modules/distributor/usage/tracker.go
```diff
@@ -0,0 +1,425 @@
...
+		// This is 1/Nth of the unaccounted for batch data that gets added to each span.
+		// Adding this incrementally as we go through the spans is the fastest method, but
+		// loses some precision. The other (original) implementation is to record span counts
+		// per series into a map and reconcile at the end. That method has more accurate data because
+		// it performs the floating point math once on the total, instead of accumulating 1/N + 1/N ... errors.
+		batchPortion := int(math.RoundToEven(float64(unaccountedForBatchData) / float64(totalSpanCount)))
+
+		// To account for the accumulated error we dump the remaining delta onto the first span, which can be negative.
+		// The result ensures the total recorded bytes matches the input.
+		firstSpanPortion := unaccountedForBatchData - batchPortion*totalSpanCount
```
joe-elliott, line 249: trying to figure out if there's any potential weird negatives here, but I got nothing. the first span should be basically random and so this should effectively randomly distribute the overflow. additionally i think most people will have resource only labels and so they will all end up in the same series anyway.

joe-elliott (commented):
final thoughts. overall lgtm just had some Qs. i'll drop a approval whenever you're ready

modules/distributor/usage/tracker.go
mdisibio: Good question. The short answer is no.  `Describe()` is called during [registration](https://github.com/prometheus/client_golang/blob/main/prometheus/registry.go#L279) on startup, so the list of tenants and series will always be empty.  Within there it mentions that a collector with no descriptions is an "unchecked" collector.  Unsure if there are any notable considerations from that.  It does work as-is.

modules/distributor/usage/tracker.go
mdisibio: Good call.  It's close but benchmarks do seem to be a bit faster:
```
                    │  before.txt  │          after.txt           │
                    │    sec/op    │   sec/op     vs base         │
UsageTrackerObserve   6.843µ ± 17%   6.190µ ± 8%  ~ (p=0.485 n=6)
```

modules/distributor/usage/tracker.go
mdisibio: Moved

modules/distributor/usage/tracker.go
mdisibio: Yes we discussed some of this in: https://github.com/grafana/tempo/pull/4162#discussion_r1793856850  For now I'd like to leave the configuration as-is (scopeless) because of the broader goals of the project and also it's similar to how dimensions for the metrics-generator are configured.

PR #4179, opened 2024-10-12
fix(logs): change log level of log of Find method in tempodb

### LINKED ISSUE #4105: "searching for trace in block" log flood
We are logging Tempo with INFO level. Every time we do a simple query, logs get bombarded by hundreds of instances saying "searching for trace in block" and some other non relevant info.
As we keep querying this becomes quickly in a issue, getting thousands of these unwanted records.
It would be nice these can be dropped down to debug or something like that? Not sure what is the value on having all these under INFO level.

Here a screen shot for a single and simple query causing almost 1k entries:

Image

Aki0x137:
**What this PR does**:
Change the log level for log in the tempoDB after searching for a trace from INFO to DEBUG

Fixes #4105 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


Aki0x137:
@javiermolinar updated the changelog.👍

javiermolinar (approved):
Thank you for your contribution!

Issue #4014, opened 2024-08-27
Remove support for serverless functions


joe-elliott:
Currently our codebase supports [AWS Lambda and GCP Cloud Run](https://github.com/grafana/tempo/tree/main/cmd/tempo-serverless). However, we have recently found that our AWS runtime has been deprecated and the team has no bandwidth to continue maintaining this functionality. We are also generally unsure of the value of serverless and if it actually provides the presumed reduced latency at a lower cost.

Opening this issue as an opportunity for community members to express the desire to keep or maintain Tempo's serverless functionality. As of now, our position is to remove it.

palerthanale:
Curious if we (Fleek.xyz / Network) might be able to help support here via Fleek Functions. Still in Alpha, but initial tests have been promising. Would love to connect and learn more, if of interest.

edgarkz:
I haven't configured it since the Tempo 2.0 version, as I found that the standard Kubernetes HPA scaling works well for our needs.

electron0zero:
serverless functions will be deprecated and deleted in the tempo next release, I am sending out a PR to mark them deprecated.

xoan-grafana:
@electron0zero is this Done?

PR #4203, opened 2024-10-18
drone: remove serverless pipeline

javiermolinar:
**What this PR does**:
Remove the serverless pipeline. It also changes the release trigger from `tag` and `pull-request` to just tag. Now we are doing all the steps but the last final release one more than needed

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -152,7 +152,7 @@ local deploy_to_dev() = {
         destination_branch: 'master',
         pull_request_branch_prefix: 'auto-merge/cd-tempo-dev',
         pull_request_enabled: true,
-        pull_request_existing_strategy: "ignore",
+        pull_request_existing_strategy: 'ignore',
         repo_name: 'deployment_tools',
         update_jsonnet_attribute_configs: [
           {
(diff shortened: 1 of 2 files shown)


kvrhdn:
I think you might have an indentation issue? The diff is really confusing

Screenshot 2024-10-18 at 13 13 37

You can use `jsonnetfmt` to format your code

```
jsonnetfmt -i .drone/drone.jsonnet
```

javiermolinar:
> jsonnetfmt -i .drone/drone.jsonnet

It was wrongly indented before

kvrhdn (commented):
Left some comments for missed variables etc.

But mostly I'm wondering _if_ we want to remove this pipeline? Pushing the images is a pretty small cost (I assume) and setting this back up would be a lot of effort. So maybe we should decide first what we do with serverless support going forward.

electron0zero:
> I'm wondering _if_ we want to remove this pipeline?

I would say let's put this PR on hold till we make a decision on the future of serverless in Tempo (mostly the answer is no, but we need to sit down and make a proper decision).

I agree with Yuna that it's little gain and if we need to invest back into serverless, it will be a lot of effort to set it back.

javiermolinar:
> > I'm wondering _if_ we want to remove this pipeline?
> 
> I would say let's put this PR on hold till we make a decision on the future of serverless functions in Tempo (see #4014).
> 
> I agree with @kvrhdn that if we decide to keep serverless functions around, it will be a lot of effort to set it back up.

I can drop it if we think it's still valuable. Thank you both for the review

PR #4204, opened 2024-10-18
fix: skip exemplars for instant queries

javiermolinar:
**What this PR does**:
Currently, we are computing the exemplars for `query_range` and ` instant queries`. Since exemplars are not supported for instant queries (same as Prometheus) we can save that computing time.

It also fix a bug where the `exemplars` query param was not being honored 

**Checklist**
- [X] Tests updated
- [X] Documentation added

### DIFF
--- a/modules/frontend/metrics_query_range_sharder.go
+++ b/modules/frontend/metrics_query_range_sharder.go
@@ -40,7 +40,6 @@ type QueryRangeSharderConfig struct {
 	MaxDuration           time.Duration `yaml:"max_duration"`
 	QueryBackendAfter     time.Duration `yaml:"query_backend_after,omitempty"`
 	Interval              time.Duration `yaml:"interval,omitempty"`
-	Exemplars             bool          `yaml:"exemplars,omitempty"`
 	MaxExemplars          int           `yaml:"max_exemplars,omitempty"`
 }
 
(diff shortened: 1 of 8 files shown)


javiermolinar:
Can we live without this config?

https://github.com/grafana/tempo/blob/280128f7b740919a7e1c3ecc0ef9fd6b9843ca41/modules/frontend/config.go#L96

If `max_exemplars` is set to 0 then it means that the Exemplars are disabled

mdisibio:
> Can we live without this config?
> 
> https://github.com/grafana/tempo/blob/280128f7b740919a7e1c3ecc0ef9fd6b9843ca41/modules/frontend/config.go#L96
> 
> If `max_exemplars` is set to 0 then it means that the Exemplars are disabled

This sounds good to me. @mapno You originally implemented this, can you think of a scenario where max_exemplars=0 and exemplars=true is meaningful?

modules/frontend/metrics_query_range_sharder.go
```diff
@@ -26,11 +26,12 @@ import (
...
-	reader    tempodb.Reader
-	overrides overrides.Interface
-	cfg       QueryRangeSharderConfig
-	logger    log.Logger
+	next        pipeline.AsyncRoundTripper[combiner.PipelineResponse]
+	reader      tempodb.Reader
+	overrides   overrides.Interface
+	cfg         QueryRangeSharderConfig
+	logger      log.Logger
+	instantMode bool
```
mdisibio, line 34: What do you think about not splitting the pipelines and using [traceql.IsInstant(req)](https://github.com/grafana/tempo/blob/main/pkg/traceql/engine_metrics.go#L135) ?

javiermolinar: We then need to compile the query, I think it's not that bad to split the pipeline in case we want to change something in the future. I don't have a strong opinion about that, both options are ok

mapno:
> > Can we live without this config?
> > https://github.com/grafana/tempo/blob/280128f7b740919a7e1c3ecc0ef9fd6b9843ca41/modules/frontend/config.go#L96
> > 
> > If `max_exemplars` is set to 0 then it means that the Exemplars are disabled
> 
> This sounds good to me. @mapno You originally implemented this, can you think of a scenario where max_exemplars=0 and exemplars=true is meaningful?

Can't think of one. Possible was left from a previous implementation. I'm good to remove it.

knylander-grafana (approved):
Thank you for updating the docs. The updates look good.

mdisibio (approved):

PR #4210, opened 2024-10-21
[ENHANCEMENT] Use resource attributes for multiplying span metrics

### LINKED ISSUE #4202: Enable SpanMultiplier Calculation from Resource Attributes, not Just Span Attributes
We manage deployment configurations by assigning resource attributes through environment variables. For instance, we set `OTEL_TRACES_SAMPLER_ARG: 0.05` and intend to specify `OTEL_RESOURCE_ATTRIBUTES: sampler_rate=$(OTEL_TRACES_SAMPLER_ARG)` accordingly. Therefore, we would like to calculate Tempo's SpanMultiplier using resource attributes to adjust and correct the metrics accurately.

By modifying the GetSpanMultiplier function in the util package, it seems like this feature could be implemented fairly easily.

JunhoeKim:
**What this PR does**:
The property previously named span_multiplier_key is now renamed to multiplier_key, utilizing the resource attributes.
Fixes #4202 

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/configuration/_index.md
+++ b/docs/sources/tempo/configuration/_index.md
@@ -369,8 +369,10 @@ metrics_generator:
             # Example: ["peer.service", "db.name", "db.system", "host.name"]
             [peer_attributes: <list of string> | default = ["peer.service", "db.name", "db.system"] ]
 
-            # Attribute Key to multiply metrics
-            [multiplier_key: <string> | default = ""]
+            # Attribute Key to multiply span metrics
+            # Note that the attribute name is searched for in both
+            # resouce and span level attributes
+            [span_multiplier_key: <string> | default = ""]
 
             # Enables additional labels for services and virtual nodes.
             [enable_virtual_node_label: <bool> | default = false]
@@ -412,8 +414,10 @@ metrics_generator:
             # Enable traces_target_info metrics
             [enable_target_info: <bool> | default = false]
 
-            # Attribute Key to multiply metrics
-            [multiplier_key: <string> | default = ""]
+            # Attribute Key to multiply span metrics
+            # Note that the attribute name is searched for in both
+            # resouce and span level attributes
+            [span_multiplier_key: <string> | default = ""]
 
             # List of policies that will be applied to spans for inclusion or exclusion.
             [filter_policies: <list of filter policies config> | default = []]
--- a/docs/sources/tempo/configuration/_index.md
+++ b/docs/sources/tempo/configuration/_index.md
@@ -369,8 +369,8 @@ metrics_generator:
             # Example: ["peer.service", "db.name", "db.system", "host.name"]
             [peer_attributes: <list of string> | default = ["peer.service", "db.name", "db.system"] ]
 
-            # Attribute Key to multiply span metrics
-            [span_multiplier_key: <string> | default = ""]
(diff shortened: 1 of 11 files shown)


modules/generator/processor/util/util.go
```diff
@@ -31,19 +32,19 @@ func FindAttributeValue(key string, attributes ...[]*v1_common.KeyValue) (string
...
+func GetMultiplier(ratioKey string, span *v1.Span, rs *v1_resource.Resource) float64 {
+	multiplier := 1.0
 	if ratioKey != "" {
-		for _, kv := range span.Attributes {
+		for _, kv := range append(span.Attributes, rs.Attributes...) {
 			if kv.Key == ratioKey {
 				v := kv.Value.GetDoubleValue()
 				if v > 0 {
-					spanMultiplier = 1.0 / v
+					multiplier = 1.0 / v
```
joe-elliott, line 42: we should return after we find the appropriate value. no reason to keep iterating. i'm not sure why the original code didn't return.

modules/generator/processor/util/util.go
```diff
@@ -31,19 +32,19 @@ func FindAttributeValue(key string, attributes ...[]*v1_common.KeyValue) (string
 	return "", false
 }
 
-func GetSpanMultiplier(ratioKey string, span *v1.Span) float64 {
-	spanMultiplier := 1.0
+func GetMultiplier(ratioKey string, span *v1.Span, rs *v1_resource.Resource) float64 {
+	multiplier := 1.0
 	if ratioKey != "" {
-		for _, kv := range span.Attributes {
+		for _, kv := range append(span.Attributes, rs.Attributes...) {
```
joe-elliott, line 38: this append will unnecessarily alloc. it would be preferred to just iterate each attribute slice separately.

docs/sources/tempo/configuration/_index.md
```diff
@@ -412,8 +412,8 @@ metrics_generator:
             # Enable traces_target_info metrics
             [enable_target_info: <bool> | default = false]
 
-            # Attribute Key to multiply span metrics
-            [span_multiplier_key: <string> | default = ""]
+            # Attribute Key to multiply metrics
+            [multiplier_key: <string> | default = ""]
```
joe-elliott, line 416: we can change this config name, but we need to mark it a breaking change. i don't see a particularly strong reason to change it. the behavior has changed some, but it still is a "multiplier for span metrics".

docs/sources/tempo/configuration/_index.md
```diff
@@ -412,8 +412,8 @@ metrics_generator:
             # Enable traces_target_info metrics
             [enable_target_info: <bool> | default = false]
 
-            # Attribute Key to multiply span metrics
-            [span_multiplier_key: <string> | default = ""]
+            # Attribute Key to multiply metrics
```
joe-elliott, line 415: we should indicate that the attribute name is searched for in both resource and span level attributes

joe-elliott (commented):
thanks for giving this a shot!

JunhoeKim:
@knylander-grafana @joe-elliott 
Thank you for the review. I have updated it to ensure no breaking changes occur based on your feedback.

PR #4213, opened 2024-10-21
[EngHub] Add "Drone Signature Check" workflow

zzehring:
Add a GitHub workflow that checks the signature of drone config file when the config file changes. The workflow will fail if the current signature does not match the expected one.


zzehring:
Test successful. But, closing as workflow needs to only run on pushes to main branch.

PR #4231, opened 2024-10-23
[EngHub] Add "Drone Signature Check" workflow

electron0zero:
Add a GitHub workflow that checks the signature of drone config file when the config file changes. The workflow will fail if the current signature does not match the expected one.


mapno (approved):
Nice addition, pretty easy to miss

PR #4236, opened 2024-10-25
fix: counter samples being downsampled when series is new

javiermolinar:
**What this PR does**:
It fixes an issue with adaptive metrics: when a series is new the init sample ends in the same aggregation bucket as the span one, and the later one is downsampled. The fix consists of backdating the initial one to the end of the previous minute:

previously:
(t-1ms, 0)
(t, 1)

current:
(t-1m, 0)
(t, 1)

**Checklist**
- [X] Tests updated
- [X] Documentation added

### DIFF
--- a/modules/generator/registry/registry.go
+++ b/modules/generator/registry/registry.go
@@ -87,7 +87,7 @@ type metric interface {
 	removeStaleSeries(staleTimeMs int64)
 }
 
-const HighestAggregationInterval = 1 * time.Minute
+const highestAggregationInterval = 1 * time.Minute
 
 var _ Registry = (*ManagedRegistry)(nil)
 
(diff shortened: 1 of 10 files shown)


mapno (approved):

modules/generator/registry/registry.go
```diff
@@ -87,6 +87,8 @@ type metric interface {
 	removeStaleSeries(staleTimeMs int64)
 }
 
+const HighestAggregationInterval = 1 * time.Minute
```
kvrhdn, line 90: We don't need to make this value public
```suggestion
const highestAggregationInterval = 1 * time.Minute
```
(you'll need to update all references too)

kvrhdn (approved):
LGTM. Just left a comment about a public constant that can be private instead.

PR #4242, opened 2024-10-28
Query-Frontend: Perf improvements

joe-elliott:
**What this PR does**:

Some small perf improvements and cleanup noticed while working on the [ordered results PR](https://github.com/grafana/tempo/pull/4238).

- Use our `ReadAllWithEstimate()` instead of `io.ReadAll()`. 
- Standardized the way we clone child requests and dropped unnecessary headers.
- Standardized our response logging. We were using both "results" and "response" so i picked one.
- Pre-alloc 100 bytes on the query builder slice

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/frontend.go
+++ b/modules/frontend/frontend.go
@@ -278,9 +278,8 @@ func newMetricsSummaryHandler(next pipeline.AsyncRoundTripper[combiner.PipelineR
 	})
 }
 
-// jpe comment, rename to create child?
-// cloneRequest returns a cloned http.Request from the pipeline.Request.
-func cloneRequest(parent pipeline.Request, tenant string, modHTTP func(*http.Request) (*http.Request, error)) (pipeline.Request, error) {
+// cloneChildRequest returns a cloned http.Request from the pipeline.Request.
+func cloneChildRequest(parent pipeline.Request, tenant string, modHTTP func(*http.Request) (*http.Request, error)) (pipeline.Request, error) {
 	// first clone the http request with headers nil'ed out. this prevents the headers from being copied saving allocs
 	// here and especially downstream in the httpgrpc bridge. prepareRequestForQueriers will add the only headers that
 	// the queriers actually need.
@@ -290,7 +289,7 @@ func cloneRequest(parent pipeline.Request, tenant string, modHTTP func(*http.Req
 	clonedHTTPReq := req.Clone(req.Context())
 
 	req.Header = saveHeaders
-	clonedHTTPReq.Header = make(http.Header, 2) // jpe - 2 b/c we know what prepareRequest is about to do
+	clonedHTTPReq.Header = make(http.Header, 2) // cheating here. alloc 2 b/c we know that's how many headers prepareRequestForQueriers will add
 
 	// give the caller a chance to modify the internal http request
 	if modHTTP != nil {
--- a/modules/frontend/frontend.go
+++ b/modules/frontend/frontend.go
@@ -269,13 +269,41 @@ func newMetricsSummaryHandler(next pipeline.AsyncRoundTripper[combiner.PipelineR
 		resp, _, err := resps.Next(req.Context()) // metrics path will only ever have one response
 
 		level.Info(logger).Log(
-			"msg", "search tag response",
+			"msg", "metrics summary response",
 			"tenant", tenant,
 			"path", req.URL.Path,
 			"err", err)
 
 		return resp.HTTPResponse(), err
 	})
+}
+
+// jpe comment, rename to create child?
+// cloneRequest returns a cloned http.Request from the pipeline.Request.
+func cloneRequest(parent pipeline.Request, tenant string, modHTTP func(*http.Request) (*http.Request, error)) (pipeline.Request, error) {
+	// first clone the http request with headers nil'ed out. this prevents the headers from being copied saving allocs
+	// here and especially downstream in the httpgrpc bridge. prepareRequestForQueriers will add the only headers that
+	// the queriers actually need.
+	req := parent.HTTPRequest()
+	saveHeaders := req.Header
+	req.Header = nil
+	clonedHTTPReq := req.Clone(req.Context())
+
+	req.Header = saveHeaders
+	clonedHTTPReq.Header = make(http.Header, 2) // jpe - 2 b/c we know what prepareRequest is about to do
+
+	// give the caller a chance to modify the internal http request
+	if modHTTP != nil {
+		var err error
+		clonedHTTPReq, err = modHTTP(clonedHTTPReq)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	prepareRequestForQueriers(clonedHTTPReq, tenant)
+
+	return parent.CloneFromHTTPRequest(clonedHTTPReq), nil
 }
 
 // prepareRequestForQueriers modifies the request so they will be farmed correctly to the queriers
--- a/modules/frontend/frontend.go
+++ b/modules/frontend/frontend.go
@@ -278,8 +278,9 @@ func newMetricsSummaryHandler(next pipeline.AsyncRoundTripper[combiner.PipelineR
 	})
 }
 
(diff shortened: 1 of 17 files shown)


modules/frontend/frontend.go
```diff
@@ -278,6 +278,33 @@ func newMetricsSummaryHandler(next pipeline.AsyncRoundTripper[combiner.PipelineR
 	})
 }
 
+// cloneChildRequest returns a cloned http.Request from the pipeline.Request.
+func cloneChildRequest(parent pipeline.Request, tenant string, modHTTP func(*http.Request) (*http.Request, error)) (pipeline.Request, error) {
```
javiermolinar, line 282: Since this is a transformation from the original request to the one we will send to the queriers, maybe a more descriptive name like: `cloneRequestForQueriers` could be better

We can also document what modHTTP does

modules/frontend/metrics_query_range_sharder.go
```diff
@@ -243,15 +244,13 @@ func (s *queryRangeSharder) buildBackendRequests(ctx context.Context, tenantID s
...
-			subR := parent.HTTPRequest().Clone(ctx)
-
-			dedColsJSON, err := colsToJSON.JSONForDedicatedColumns(m.DedicatedColumns)
-			if err != nil {
-				// errFn(fmt.Errorf("failed to convert dedicated columns. block: %s tempopb: %w", blockID, err))
-				continue
-			}
+		dedColsJSON, err := colsToJSON.JSONForDedicatedColumns(m.DedicatedColumns)
+		if err != nil {
+			// errFn(fmt.Errorf("failed to convert dedicated columns. block: %s tempopb: %w", blockID, err))
```
javiermolinar, line 249: do we want to log something here? if not we can delete the comment

modules/frontend/metrics_query_range_sharder.go
```diff
@@ -261,31 +260,30 @@ func (s *queryRangeSharder) buildBackendRequests(ctx context.Context, tenantID s
...
-				FooterSize:    m.FooterSize,
-				// DedicatedColumns: dc, for perf reason we pass dedicated columns json in directly to not have to realloc object -> proto -> json
-				Exemplars: exemplars,
-			}
-
-			subR = api.BuildQueryRangeRequest(subR, queryRangeReq, dedColsJSON)
-
-			prepareRequestForQueriers(subR, tenantID)
-			pipelineR := parent.CloneFromHTTPRequest(subR)
+			pipelineR, _ := cloneChildRequest(parent, tenantID, func(r *http.Request) (*http.Request, error) {
```
javiermolinar, line 263: should we handle errors cloning the request?

modules/frontend/search_sharder.go
```diff
@@ -310,36 +311,37 @@ func buildBackendRequests(ctx context.Context, tenantID string, parent pipeline.
...
 		blockID := m.BlockID.String()
-		for startPage := 0; startPage < int(m.TotalRecords); startPage += pages {
-			subR := parent.HTTPRequest().Clone(ctx)
 
-			dedColsJSON, err := colsToJSON.JSONForDedicatedColumns(m.DedicatedColumns)
-			if err != nil {
-				errFn(fmt.Errorf("failed to convert dedicated columns. block: %s tempopb: %w", blockID, err))
-				continue
-			}
+		dedColsJSON, err := colsToJSON.JSONForDedicatedColumns(m.DedicatedColumns)
```
javiermolinar, line 315: Nice to take this out of the loop

modules/frontend/tag_sharder.go
```diff
@@ -213,7 +213,7 @@ func (s searchTagSharder) RoundTrip(pipelineRequest pipeline.Request) (pipeline.
 
 	// build request to search ingester based on query_ingesters_until config and time range
 	// pass subCtx in requests, so we can cancel and exit early
-	ingesterReq, err := s.ingesterRequest(ctx, tenantID, pipelineRequest, searchReq)
+	ingesterReq, err := s.ingesterRequest(tenantID, pipelineRequest, searchReq)
```
javiermolinar, line 216: this context that we are not passing anymore comes from the tracer. Maybe we should be passing it on to the   cloneChildRequest to not lose the tracing context

modules/frontend/tag_sharder.go
```diff
@@ -296,14 +296,13 @@ func (s searchTagSharder) buildBackendRequests(ctx context.Context, tenantID str
...
 		for startPage := 0; startPage < int(m.TotalRecords); startPage += pages {
-			subR := parent.Clone(ctx)
-			subR, err := searchReq.buildTagSearchBlockRequest(subR, blockID, startPage, pages, m)
+			pipelineR, err := cloneChildRequest(parent, tenantID, func(r *http.Request) (*http.Request, error) {
+				return searchReq.buildTagSearchBlockRequest(r, blockID, startPage, pages, m)
+			})
 			if err != nil {
 				errFn(err)
-				return
+				continue
```
javiermolinar, line 304: What's the implication of this? Before we failed fast

modules/frontend/traceid_sharder.go
```diff
@@ -66,35 +63,38 @@ func (s asyncTraceSharder) RoundTrip(pipelineRequest pipeline.Request) (pipeline
...
-			params[querier.BlockEndKey] = hex.EncodeToString(s.blockBoundaries[i])
-			params[querier.QueryModeKey] = querier.QueryModeBlocks
-		}
-		reqs[i] = api.BuildQueryRequest(reqs[i], params)
-		prepareRequestForQueriers(reqs[i], userID)
+		i := i // save the loop variable locally to make sure the closure grabs the correct var.
+		pipelineR, _ := cloneChildRequest(parent, userID, func(r *http.Request) (*http.Request, error) {
+			if i == 0 {
+				// ingester query
+				params[querier.QueryModeKey] = querier.QueryModeIngesters
```
javiermolinar, line 87: Maybe we could take this out and start the loop on 1

javiermolinar (commented):
Looking good!

modules/frontend/traceid_sharder.go
```diff
@@ -40,13 +39,11 @@ func newAsyncTraceIDSharder(cfg *TraceByIDConfig, logger log.Logger) pipeline.As
 
 // RoundTrip implements http.RoundTripper
 func (s asyncTraceSharder) RoundTrip(pipelineRequest pipeline.Request) (pipeline.Responses[combiner.PipelineResponse], error) {
-	r := pipelineRequest.HTTPRequest()
-
-	ctx, span := tracer.Start(r.Context(), "frontend.ShardQuery")
+	ctx, span := tracer.Start(pipelineRequest.Context(), "frontend.ShardQuery")
 	defer span.End()
-	r = r.WithContext(ctx)
+	pipelineRequest.WithContext(ctx)
```
javiermolinar, line 44: Not sure if this creates more allocation than just passing the context to the clone function

modules/frontend/metrics_query_range_sharder.go
joe-elliott: yeah, probably :). i was cheating since i knew that cloneChildRequest would only return err if modHTTP returned err, but that's a bad idea. fixing!

modules/frontend/tag_sharder.go
joe-elliott: the context was an unused param so i removed it. context should be preserved b/c we are using the context of the pipeline.Request when we clone

modules/frontend/tag_sharder.go
joe-elliott: before we stopped creating jobs if this err hit and only issued the first N jobs. now we just skip one job. this failure should basically never happen, but we log it anyway.

we should do a better job of propagating this error back to the user if it were to ever happen

modules/frontend/traceid_sharder.go
joe-elliott: sure. i was trying to change the original code as little as possible, but agree it's a nicer structure. change made!

javiermolinar (approved):
Great!

PR #4244, opened 2024-10-29
Native histogram performance

zalegrala:
**What this PR does**:

Improve performance in the native histograms implementations.  This is following in the footsteps of #4232.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/registry/native_histogram.go
+++ b/modules/generator/registry/native_histogram.go
@@ -1,6 +1,7 @@
 package registry
 
 import (
+	"fmt"
 	"math"
 	"sync"
 	"time"
@@ -41,6 +42,11 @@ type nativeHistogram struct {
 	histogramOverride HistogramMode
 
 	externalLabels map[string]string
+
+	// classic
+	nameCount  string
+	nameSum    string
+	nameBucket string
 }
 
 type nativeHistogramSeries struct {
@@ -56,6 +62,11 @@ type nativeHistogramSeries struct {
 	// This avoids Prometheus throwing away the first value in the series,
 	// due to the transition from null -> x.
 	firstSeries *atomic.Bool
+
+	// classic
+	countLabels labels.Labels
+	sumLabels   labels.Labels
+	// bucketLabels []labels.Labels
 }
 
 func (hs *nativeHistogramSeries) isNew() bool {
@@ -94,6 +105,11 @@ func newNativeHistogram(name string, buckets []float64, onAddSeries func(uint32)
 		buckets:           buckets,
 		histogramOverride: histogramOverride,
 		externalLabels:    externalLabels,
+
+		// classic
+		nameCount:  fmt.Sprintf("%s_count", name),
+		nameSum:    fmt.Sprintf("%s_sum", name),
+		nameBucket: fmt.Sprintf("%s_bucket", name),
 	}
 }
 
@@ -150,6 +166,14 @@ func (h *nativeHistogram) newSeries(labelValueCombo *LabelValueCombo, value floa
 
 	newSeries.labels = lb.Labels()
 	newSeries.lb = lb
+
+	// _count
+	lb.Set(labels.MetricName, h.nameCount)
+	newSeries.countLabels = lb.Labels()
+
+	// _sum
+	lb.Set(labels.MetricName, h.nameSum)
+	newSeries.sumLabels = lb.Labels()
 
 	return newSeries
 }
@@ -288,9 +312,8 @@ func (h *nativeHistogram) nativeHistograms(appender storage.Appender, lbls label
 
 func (h *nativeHistogram) classicHistograms(appender storage.Appender, lb *labels.Builder, timeMs int64, s *nativeHistogramSeries) (activeSeries int, err error) {
 	if s.isNew() {
-		lb.Set(labels.MetricName, h.metricName+"_count")
 		endOfLastMinuteMs := getEndOfLastMinuteMs(timeMs)
-		_, err = appender.Append(0, lb.Labels(), endOfLastMinuteMs, 0)
+		_, err = appender.Append(0, s.countLabels, endOfLastMinuteMs, 0)
 		if err != nil {
 			return activeSeries, err
 		}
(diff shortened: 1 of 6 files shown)


modules/generator/registry/native_histogram.go
```diff
@@ -90,6 +104,12 @@ func newNativeHistogram(name string, buckets []float64, onAddSeries func(uint32)
 		traceIDLabelName:  traceIDLabelName,
 		buckets:           buckets,
 		histogramOverride: histogramOverride,
+		externalLabels:    externalLabels,
+
+		// classic
+		nameCount:  fmt.Sprintf("%s_count", name),
```
joe-elliott, line 110: does `name` need to be sanitized for prom? does it ever come from an external config?

modules/generator/registry/native_histogram.go
```diff
@@ -109,14 +129,12 @@ func (h *nativeHistogram) ObserveWithExemplar(labelValueCombo *LabelValueCombo,
...
 	}
 
-	newSeries := h.newSeries(labelValueCombo, value, traceID, multiplier)
-	h.series[hash] = newSeries
+	h.series[hash] = h.newSeries(labelValueCombo, value, traceID, multiplier)
 }
 
 func (h *nativeHistogram) newSeries(labelValueCombo *LabelValueCombo, value float64, traceID string, multiplier float64) *nativeHistogramSeries {
 	newSeries := &nativeHistogramSeries{
 		// TODO move these labels in HistogramOpts.ConstLabels?
```
joe-elliott, line 137: remove todo?

modules/generator/registry/native_histogram.go
```diff
@@ -132,6 +150,31 @@ func (h *nativeHistogram) newSeries(labelValueCombo *LabelValueCombo, value floa
...
+	}
+	// set external labels
+	for name, value := range h.externalLabels {
+		lb.Set(name, value)
+	}
+
+	lb.Set(labels.MetricName, h.metricName)
+
+	newSeries.labels = lb.Labels()
+	newSeries.lb = lb
```
joe-elliott, line 168: ok, for clarity this becomes the label builder for all `le` series? it seems like that's basically the only place it's used. 

just concerned that labels from one series will "leak" into another if we share this label builder too much. it seems like the same label builder is on the main histogram and passed to all the series?

modules/generator/registry/native_histogram.go
```diff
@@ -180,14 +208,15 @@ func (h *nativeHistogram) collectMetrics(appender storage.Appender, timeMs int64
...
+		// NOTE: Store the encoded histogram here so we can keep track of the
+		// exemplars that have been sent.  The value is updated here, but the
+		// pointers remain the same, and so Reset() call below can be used to clear
+		// the exemplars.
 		s.histogram = encodedMetric.GetHistogram()
 
 		// If we are in "both" or "classic" mode, also emit classic histograms.
 		if hasClassicHistograms(h.histogramOverride) {
-			classicSeries, classicErr := h.classicHistograms(appender, lb, timeMs, s)
+			classicSeries, classicErr := h.classicHistograms(appender, s.lb, timeMs, s)
```
joe-elliott, line 219: since s.lb is a member var it could be removed from the function signature?

keeping it as a param might make it easier to test?

modules/generator/registry/registry.go
```diff
@@ -156,7 +156,7 @@ func (r *ManagedRegistry) NewHistogram(name string, buckets []float64, histogram
 	// are disabled, eventually the new implementation can handle all cases
 
 	if hasNativeHistograms(histogramOverride) {
-		h = newNativeHistogram(name, buckets, r.onAddMetricSeries, r.onRemoveMetricSeries, traceIDLabelName, histogramOverride)
+		h = newNativeHistogram(name, buckets, r.onAddMetricSeries, r.onRemoveMetricSeries, traceIDLabelName, histogramOverride, r.externalLabels)
```
joe-elliott, line 159: adding external labels here is a new feature? or fixing a bug where we weren't applying external labels to native histos?

joe-elliott (commented):
some Qs but i think it looks great

modules/generator/registry/native_histogram.go
zalegrala: I think we're good on the name.  Each metric name is a constant.
https://github.com/zalegrala/tempo/blob/f367ceadb5d9c679ae13cadef8085833fef5123a/modules/generator/processor/spanmetrics/spanmetrics.go#L85
https://github.com/zalegrala/tempo/blob/f367ceadb5d9c679ae13cadef8085833fef5123a/modules/generator/processor/servicegraphs/servicegraphs.go#L123
Good question though.

modules/generator/registry/native_histogram.go
zalegrala: Yes that's correct.  Since Labels() doesn't return a pointer, then we just get whatever it had at that time, which I didn't realize until I started poking on this PR.  Here is why I think its okay.
For [count](https://github.com/zalegrala/tempo/blob/1cb964399993c7f8283a1b2e830030215c5aee96/modules/generator/registry/native_histogram.go#L169) and [sum](https://github.com/zalegrala/tempo/blob/1cb964399993c7f8283a1b2e830030215c5aee96/modules/generator/registry/native_histogram.go#L173) metrics, we capture the labels at init.  Then the only place that we use the stored label builder is in the `classicHistograms`, which you noted.  Then we set the name before we use any of the labels [here](https://github.com/zalegrala/tempo/blob/1cb964399993c7f8283a1b2e830030215c5aee96/modules/generator/registry/native_histogram.go#L336), then loop over all the buckets and set their `le` label [here](https://github.com/zalegrala/tempo/blob/1cb964399993c7f8283a1b2e830030215c5aee96/modules/generator/registry/native_histogram.go#L344), same for the +Inf bucket.  Then we clean up at the bottom [here](https://github.com/zalegrala/tempo/blob/1cb964399993c7f8283a1b2e830030215c5aee96/modules/generator/registry/native_histogram.go#L381).

One improvement I can see is that perhaps we want to defer that delete, since there are opportunities to return early which wouldn't delete the `le` label, but this probably doesn't matter because we set it before we ever use it, so even the delete at the bottom is probably unncessary.  (I say this before running the tests).

modules/generator/registry/native_histogram.go
zalegrala: Good call.  Tests cover this without needing the param.

modules/generator/registry/registry.go
zalegrala: The labels used to get passed on the `collectMetrics` function, and was just moved up to avoid the extra calls there, since they don't change after the histogram is created.

modules/generator/registry/native_histogram.go
joe-elliott: Nice digging. Yup, the .Labels() method copies the internal slice instead of returning it so this is safe.

joe-elliott (approved):
Let's go!

PR #4254, opened 2024-10-31
Metrics Generator: Max limit on number of failed flushes

joe-elliott:
**What this PR does**:
Institutes a very generous limit of 100 on failed flush attempts as well as a metric to help track this. This prevents constant retries on blocks that cannot be flushed due to corruption.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/processor/localblocks/processor.go
+++ b/modules/generator/processor/localblocks/processor.go
@@ -297,7 +297,7 @@ func (p *Processor) flushLoop() {
 
 		err := p.flushBlock(op.blockID)
 		if err != nil {
-			_ = level.Info(p.logger).Log("msg", "re-queueing block for flushing", "block", op.blockID, "attempts", op.attempts)
+			_ = level.Info(p.logger).Log("msg", "re-queueing block for flushing", "block", op.blockID, "attempts", op.attempts, "err", err)
 			metricFailedFlushes.Inc()
 
 			delay := op.backoff()
--- a/modules/generator/processor/localblocks/processor.go
+++ b/modules/generator/processor/localblocks/processor.go
@@ -34,7 +34,10 @@ import (
 
 var tracer = otel.Tracer("modules/generator/processor/localblocks")
 
-const timeBuffer = 5 * time.Minute
+const (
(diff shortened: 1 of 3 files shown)


modules/generator/processor/localblocks/processor.go
```diff
@@ -276,16 +279,36 @@ func (p *Processor) flushLoop() {
...
-			if _, err := p.flushqueue.Enqueue(op); err != nil {
-				_ = level.Error(p.logger).Log("msg", "failed to requeue block for flushing", "err", err)
-			}
+			delay := op.backoff()
+			op.at = time.Now().Add(delay)
+
+			go func() {
+				time.Sleep(delay)
+				if _, err := p.flushqueue.Enqueue(op); err != nil {
+					_ = level.Error(p.logger).Log("msg", "failed to requeue block for flushing", "err", err)
```
javiermolinar, line 309: What happens if we fail to requeue? Should we delete the block as well?

modules/generator/processor/localblocks/processor.go
```diff
@@ -276,16 +279,36 @@ func (p *Processor) flushLoop() {
...
+			p.blocksMtx.Unlock()
+
+			continue
+		}
+
 		err := p.flushBlock(op.blockID)
 		if err != nil {
-			_ = level.Error(p.logger).Log("msg", "failed to flush a block", "err", err)
-
 			_ = level.Info(p.logger).Log("msg", "re-queueing block for flushing", "block", op.blockID, "attempts", op.attempts)
```
javiermolinar, line 300: Should we also log the reason why the block cannot be flushed?

javiermolinar (approved):
LGTM

modules/generator/processor/localblocks/processor.go
joe-elliott: i think no. the only time enqueue can fail is if the queue is stopped:

https://github.com/grafana/tempo/blob/main/pkg/flushqueues/priority_queue.go#L91-L93

in this case the process is shutting down and so we want to keep the block for when it starts back up. we could definitely be more explicit about this tho.

PR #4255, opened 2024-10-31
fix: reduce allocations in the Time Window Block Selector

javiermolinar:
**What this PR does**:

I noticed a huge amount of objects allocated during compactions from the Sprintf method

Screenshot 2024-10-31 at 10 09 33

This PR reuses a string buffer to reduce these allocations

```
❯ benchstat old.txt new.txt
goos: darwin
goarch: arm64
pkg: github.com/grafana/tempo/tempodb
cpu: Apple M3 Pro
              │   old.txt   │               new.txt               │
              │   sec/op    │   sec/op     vs base                │
BlockSelector   60.07m ± 1%   25.37m ± 2%  -57.77% (p=0.000 n=10)

              │   old.txt    │               new.txt                │
              │     B/op     │     B/op      vs base                │
BlockSelector   43.74Mi ± 0%   28.23Mi ± 0%  -35.45% (p=0.000 n=10)

              │   old.txt   │               new.txt               │
              │  allocs/op  │  allocs/op   vs base                │
BlockSelector   800.0k ± 0%   700.0k ± 0%  -12.50% (p=0.000 n=10)
```

**Checklist**
- [X] Tests updated

### DIFF
--- a/tempodb/compaction_block_selector.go
+++ b/tempodb/compaction_block_selector.go
@@ -1,8 +1,9 @@
 package tempodb
 
 import (
-	"fmt"
 	"sort"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/grafana/tempo/tempodb/backend"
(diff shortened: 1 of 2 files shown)


tempodb/compaction_block_selector.go
```diff
@@ -74,29 +78,82 @@ func newTimeWindowBlockSelector(blocklist []*backend.BlockMeta, maxCompactionRan
 		}
 
 		age := currWindow - w
+		builder.Reset()
+
 		if activeWindow <= w {
+			// Grow size: 2+20+1+16+1+20
+			builder.Grow(60)
```
joe-elliott, line 85: this is a very tiny %age of all compactor allocations and the previous implementation had a lot of clarity on what the group/order/hash strings looked like. the current implementation loses this a bit

i'm fine with moving forward with this, but can we need the new implementation to read as cleanly as the old given that the value gained is so small

PR #4259, opened 2024-11-01
Fix tempo cli arguments

faridtmammadov:
**What this PR does**:
This PR fixes the unused S3Pass and S3User parameters in the CLI options by implementing them in the code. Additionally, it introduces a skip verify option to bypass SSL certificate validation when connecting to the S3 backend, enhancing flexibility for self-signed or internal certificates.

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/cmd/tempo-cli/main.go
+++ b/cmd/tempo-cli/main.go
@@ -30,9 +30,10 @@ type backendOptions struct {
 	Backend string `help:"backend to connect to (s3/gcs/local/azure), optional, overrides backend in config file" enum:",s3,gcs,local,azure" default:""`
 	Bucket  string `help:"bucket (or path on local backend) to scan, optional, overrides bucket in config file"`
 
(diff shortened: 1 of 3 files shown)


cmd/tempo-cli/main.go
```diff
@@ -131,6 +133,16 @@ func loadBackend(b *backendOptions, g *globalOptions) (backend.Reader, backend.W
 		cfg.StorageConfig.Trace.Azure.ContainerName = b.Bucket
 	}
 
+	cfg.StorageConfig.Trace.S3.InsecureSkipVerify = b.InsecureSkipVerify
```
joe-elliott, line 136: let's set the same field on the GCS config so it applies to both backends.

`cfg.StorageConfig.Trace.GCS.Insecure`

It seems like we don't have a similar option on Azure. Can we mention that in the `help` text? something like "only applies to S3 and GCS"

joe-elliott (commented):
i have never used these fields and now i know they wouldn't have worked anyway.

thanks for spotting this and the fix! requesting a small change to the insecure flag. 🙏

mapno (approved):
Thanks!

PR #4260, opened 2024-11-01
Traces Cleanup

joe-elliott:
**What this PR does**:
I set out to clean up our traces and remove all orphaned spans, but there were fewer than expected. Our query path is entirely connected.

- Connected some local block operations in the generator.
- Similar fix in the ingester and also updated log lines to use "tenant" over "userID" which is preferred.
image
- Gave context to some orphanned s3.Reads in usage stats
image
- Connected some orphanned spans in our polling cycle

There are still two orphanned spans I can't figure out how to fix related to ring management. They create single span "CAS" and "GET" traces. I'll let the next person figure those out :).

https://github.com/grafana/dskit/blob/a4a4dca2b2e64cc70a1a719c5f3dbf7a5c81adba/kv/metrics.go#L97
https://github.com/grafana/dskit/blob/a4a4dca2b2e64cc70a1a719c5f3dbf7a5c81adba/kv/metrics.go#L81

There are also a fair number of "low value" traces. These are auto instrumented by our http server and may be best removed through sampling techniques. Things like getting metrics or health checks.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


zalegrala (approved):
Nice cleanup.

PR #4262, opened 2024-11-01
Improve the readability of redis_client util method

zalegrala:
Here we modify a utility method in the redis client which is used to retrieve the bytes of the array backing a string.  New `unsafe` methods exist to support this use case, which is considerably more readable.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott (approved):
nice

PR #4266, opened 2024-11-03
tempo-cli: support dropping multiple traces in a single operation

### LINKED ISSUE #4246: tempo-cli: add support for dropping multiple traces in single operation
We currently have a command in `tempo-cli` under `rewrite-blocks drop-trace` to delete a [single trace from a block](https://grafana.com/docs/tempo/latest/operations/tempo_cli/#drop-trace-by-id).

However, there are scenarios where operators may need to drop multiple traces from a single block. 
To achieve this now, they must loop through the list of trace IDs and execute the `drop-trace` command. This is inefficient, as it requires rewriting the same block multiple times (_pulling and uploading the same block to backend_) when dropping several traces.

I propose adding a new command (or updating the current one) to accept a list of trace IDs to delete multiple traces from a block in a single operation.

ndk:
**What this PR does**:
This PR introduces support for dropping multiple traces from a single block in tempo-cli. The new command syntax allows specifying multiple trace IDs separated by commas:

```console
tempo-cli rewrite-blocks drop-traces --backend=local --bucket=./cmd/tempo-cli/test-data --drop-trace single-tenant 04d5f549746c96e4f3daed6202571db2,0fb764d54d68a10734268771fb3e00f0
```

Fixes #4246

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/cmd/tempo-cli/cmd-rewrite-blocks.go
+++ b/cmd/tempo-cli/cmd-rewrite-blocks.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/go-kit/log"
 	"github.com/google/uuid"
@@ -19,16 +20,16 @@ import (
 	"github.com/grafana/tempo/tempodb/encoding/common"
 )
 
-type dropTraceCmd struct {
+type dropTracesCmd struct {
 	backendOptions
 
-	TraceID   string `arg:"" help:"trace ID to retrieve"`
 	TenantID  string `arg:"" help:"tenant ID to search"`
+	TraceIDs  string `arg:"" help:"Trace IDs to drop"`
 	DropTrace bool   `name:"drop-trace" help:"actually attempt to drop the trace" default:"false"`
 }
 
-func (cmd *dropTraceCmd) Run(ctx *globalOptions) error {
-	fmt.Printf("beginning process to drop trace %v from tenant %v\n", cmd.TraceID, cmd.TenantID)
+func (cmd *dropTracesCmd) Run(opts *globalOptions) error {
+	fmt.Printf("beginning process to drop traces %v from tenant %v\n", cmd.TraceIDs, cmd.TenantID)
 	fmt.Println("**warning**: compaction must be disabled or a compactor may duplicate a block as this process is rewriting it")
 	fmt.Println("")
 	if cmd.DropTrace {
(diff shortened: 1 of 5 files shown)


cmd/tempo-cli/cmd-rewrite-blocks.go
```diff
@@ -43,48 +44,55 @@ func (cmd *dropTraceCmd) Run(ctx *globalOptions) error {
...
+		}
+		ids = append(ids, id)
 	}
 
-	blocks, err := blocksWithTraceID(context.Background(), r, cmd.TenantID, id)
-	if err != nil {
-		return err
-	}
+	for _, id := range ids {
+		blocks, err := blocksWithTraceID(context.Background(), r, cmd.TenantID, id)
```
electron0zero, line 58: I am thinking the following approach:

build a list of unique blocks that contains these IDs, one trace ID can be in multiple blocks and one block can have multiple trace IDs that we want to delete.

once we have a unique list of blocks that contain these traceIDs, we can go over each block, and then delete all the trace IDs and push that block. this means that we are only rewriting a block only once for all the traceIDs that needs to be deleted from it.

knylander-grafana:
Do we also need to update the [tempo-cli doc](https://github.com/grafana/tempo/blob/main/docs/sources/tempo/operations/tempo_cli.md) with this change?

ndk:
> Looks like some good ideas and improvements. Will we also need to update the [tempo-cli doc](https://github.com/grafana/tempo/blob/main/docs/sources/tempo/operations/tempo_cli.md)? If so, we can create a new doc issue (especially if this PR os an exploration) to make sure the CLI changes are documented for Tempo users.

It depends on which order of arguments you want to have. There are written: `tempo-cli rewrite-blocks drop-trace ... single-tenant a188ea38aa3a83d74523774ad6728cc8`, but in the code it has the opposite order.
If you want to have <TenantID> <TraceIDs> it should be fixed in the code.
If you want to have <TraceIDs> <TenantID> it should be fixed in the docuemntation.

ndk:
This commit is outside the PR to demonstrate the idea of suppressing test output to stdout. It also allows us to cover the code with unit tests.

https://github.com/ndk/tempo/commit/db8d6548fbd4aaa4cd59de0e786b589d355ccd1c

cmd/tempo-cli/cmd-rewrite-blocks.go
```diff
@@ -131,7 +153,13 @@ func rewriteBlock(ctx context.Context, r backend.Reader, w backend.Writer, meta
 
 		// hook to drop the trace
 		DropObject: func(id common.ID) bool {
-			return bytes.Equal(id, traceID)
+			// Should it be replaced with binary search, trie or other ones?
+			for _, tid := range traceIDs {
+				if bytes.Equal(id, tid) {
```
electron0zero, line 158: I think this if can be in lineed in the return like `return bytes.Equal(id, tid)`

cmd/tempo-cli/cmd-rewrite-blocks.go
```diff
@@ -38,51 +39,72 @@ func (cmd *dropTraceCmd) Run(ctx *globalOptions) error {
...
+		}
 	}
 
-	fmt.Println("rewriting blocks:")
-	for _, block := range blocks {
-		fmt.Printf("  rewriting %v\n", block.BlockID)
-		newBlock, err := rewriteBlock(context.Background(), r, w, block, id)
+	// Remove traces from blocks
+	for _, p := range tracesByBlock {
+		// print out trace IDs to be removed in the block
```
electron0zero, line 84: I am not sure if I understand the use of `{ }` and `\n\n` in the log lines here? another small nit is that `traceIDs` is kinda similar to `p.traceIDs` so a name like `strTraceIds` would be better suited to indicate that these are string trace IDs

I was thinking something like this:

```golang
		// print out trace IDs to be removed in the block
		strTraceIds := make([]string, len(p.traceIDs))
		for i, tid := range p.traceIDs {
			strTraceIds[i] = util.TraceIDToHexString(tid)
		}
		fmt.Printf("Found %d traces: %v found in block: %v\n", len(strTraceIds), strTraceIds, p.blockMeta.BlockID)
		fmt.Printf("blockInfo: ID: %v, Size: %d Total Traces: %d\n", p.blockMeta.BlockID, p.blockMeta.Size_, p.blockMeta.TotalObjects)
```

cmd/tempo-cli/cmd-rewrite-blocks.go
```diff
@@ -38,51 +39,72 @@ func (cmd *dropTraceCmd) Run(ctx *globalOptions) error {
...
+		{
+			traceIDs := make([]string, len(p.traceIDs))
+			for i, tid := range p.traceIDs {
+				traceIDs[i] = util.SpanIDToHexString(tid)
```
electron0zero, line 88:
```suggestion
				traceIDs[i] = util.TraceIDToHexString(tid)
```

we are dealing with TraceIDs and not SpanIDs, so we want to use `TraceIDToHexString` here, we were using `HexStringToTraceID(cmd.TraceID)` here.

docs:
> - SpanIDToHexString converts a span ID to its string representation and WITHOUT removing any leading zeros.
> - TraceIDToHexString converts a trace ID to its string representation and removes any leading zeros.

electron0zero (commented):
direction looks good, left some comments.

cmd/tempo-cli/cmd-rewrite-blocks_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/grafana/tempo/tempodb/encoding/common"
+	"github.com/grafana/tempo/tempodb/encoding/vparquet4"
+)
+
+func TestDropTraceCmd(t *testing.T) {
+	for s := 2; s < 10; s++ {
+		s := s
+		t.Run(fmt.Sprintf("drop every %d trace", s), func(t *testing.T) {
+			tempDir := t.TempDir()
+			err := os.CopyFS(tempDir, os.DirFS("./test-data2"))
```
electron0zero, line 27: I see that only `test-data2` is used, can we cleanup test-data dirs, and only keep one dir, and name it `testdata`.

the name `testdata` is the convention for storing test data.

electron0zero: we prefer to not checkin blocks and generate it for the testing.

see `makeBackendBlockWithTraces‎` for reference https://github.com/grafana/tempo/blob/403fdcf14e595281377a23b34a6c61612a6d5216/tempodb/encoding/vparquet4/block_search_test.go#L252

cmd/tempo-cli/cmd-rewrite-blocks.go
ndk: The loop checks if `traceIDs` contains `id`, while `return bytes.Equal(id, tid)` would only compare `id` to a single value, missing the iteration over all elements.

cmd/tempo-cli/cmd-rewrite-blocks.go
ndk: Just to reduce the scope. I'll remove to avoid confusing people. :) `\n\n` was directly carried over from the original code. I wasn't sure if it was critical or not, so I kept the message format consistent with the original style.

docs/sources/tempo/operations/tempo_cli.md
```diff
@@ -506,20 +506,27 @@ Options:
...
+## Drop traces by id
 
-Rewrites all blocks for a tenant that contain a specific trace id. The trace is dropped from
+Rewrites all blocks for a tenant that contain a specific trace ids. The traces are dropped from
 the new blocks and the rewritten blocks are marked compacted so they will be cleaned up.
 
 Arguments:
 - `tenant-id` The tenant ID. Use `single-tenant` for single tenant setups.
-- `trace-id` The trace id to drop
+- `trace-ids` The comma separated trace ids to drop (also supports single trace id)
```
knylander-grafana, line 516: ID should be upper case. I'll correct it in this section. 
```suggestion
- `trace-ids` The comma-separated trace IDs to drop (also supports single trace ID)
```

docs/sources/tempo/operations/tempo_cli.md
```diff
@@ -506,20 +506,27 @@ Options:
 tempo-cli analyse blocks --backend=local --bucket=./cmd/tempo-cli/test-data/ single-tenant
 ```
 
-## Drop trace by id
+## Drop traces by id
```
knylander-grafana, line 509:
```suggestion
## Drop traces by ID
```

docs/sources/tempo/operations/tempo_cli.md
```diff
@@ -506,20 +506,27 @@ Options:
...
 
-Rewrites all blocks for a tenant that contain a specific trace id. The trace is dropped from
+Rewrites all blocks for a tenant that contain a specific trace ids. The traces are dropped from
```
knylander-grafana, line 511:
```suggestion
Rewrites all blocks for a tenant that contain a specific trace IDs. The traces are dropped from
```

docs/sources/tempo/operations/tempo_cli.md
```diff
@@ -506,20 +506,27 @@ Options:
...
+- `--drop-traces` By default this command runs in dry run mode. Supplying this argument causes it to actually rewrite blocks with the traces dropped.
 
 **Example:**
+
+Dropping one trace
+```bash
+tempo-cli rewrite-blocks drop-trace --backend=local --bucket=./cmd/tempo-cli/test-data/ single-tenant 04d5f549746c96e4f3daed6202571db2
+```
+
+Dropping multiple traces
```
knylander-grafana, line 529:
```suggestion
Drop multiple traces:
```

ndk:
Rebased due to some conflicts in `CHANGELOG.md`

electron0zero (approved):
tested locally, works :shipit: 

thank you for your contribution.

PR #4287, opened 2024-11-05
Respect passed headers in read path requests

mapno:
**What this PR does**:

https://github.com/grafana/tempo/pull/4242 broke compatibility with non-vanilla distributions of Tempo that rely on headers in the read path.

This PR respects headers in queries and forwards them to queriers as before.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


electron0zero (approved):

javiermolinar (commented):
I think the change was ok, we need to identify the missing header (X-Org-id or Authorization) and add it to the prepareRequestForQueriers

mdisibio (approved):
This change LGTM, lean towards reverting this behavior now to fix things, and we can revisit the alloc savings once we know more.

PR #4292, opened 2024-11-06
fix: exemplars for compare function

### LINKED ISSUE #4284: `compare()` returns fewer exemplars than expected
## Context

Exemplars were added to TraceQL metrics in https://github.com/grafana/tempo/pull/3824. Queries such as `{} | rate()`, `{} | quantile_over_time(duration, 0,99)`, `{} | compare({status=error})` return exemplars if enabled.

Exemplars in `compare()` are only returned for the series `__meta_type="selection_total"` and `__meta_type="baseline_total"`, since every other series is a subset of the totals, and exemplars carry all labels of the query.

## Problem

Queries that use `compare()` are inconsistent in returning exemplars—usually return very few exemplars and only in some series, instead of in all series.

There are a couple of potential issues at hand here:
* Sampling is very naive and the [`CallbackPredicate`](https://github.com/grafana/tempo/blob/main/pkg/parquetquery/predicates.go#L498-L500) might be stopping collection too aggressively. The [callback](https://github.com/grafana/tempo/blob/main/pkg/traceql/engine_metrics.go#L862) might be quitting before enough exemplars are collected.
* Exemplars might not be propagated correctly to all series. There are multiple places where series are merged, and exemplars could be lost there. The first place is in the [engine itself](https://github.com/grafana/tempo/blob/main/pkg/traceql/engine_metrics_compare.go#L264-L283). But there are followup calls in the queriers, generators and frontends.

javiermolinar:
**What this PR does**:

It fixes the exemplars selected for the compare function by adding the exemplars to the `baseline` and `selection` series instead of `baseline_total` and `selection_total`

*Notes*

Current implementation of the SimpleCombiner that we use for the second layer of aggregation of the compare function uses only one Exemplar bucket. This bucket does not discriminate between `baseline` and `selection`, it bases its decision of dropping or not the exemplar just on the interval. That means that we can end with intervals without exemplars for either `baseline` or `selection`. If we want to correct this we would need two buckets

Fixes #4284

**Checklist**
- [X] Tests updated

### DIFF
--- a/pkg/traceql/engine_metrics_compare.go
+++ b/pkg/traceql/engine_metrics_compare.go
@@ -262,24 +262,37 @@ func (m *MetricsCompare) result() SeriesSet {
 	addTotals(internalLabelTypeSelectionTotal, m.selectionTotals)
 
 	// Add exemplars
-	addExemplar := func(prefix Label, e Exemplar) {
+	addExemplar := func(prefix Label, e Exemplar, usedExemplars map[string]bool) {
 		for _, l := range e.Labels {
 			seriesLabels := Labels{
 				prefix,
-				{Name: l.Name},
+				{Name: l.Name, Value: l.Value},
 			}
-			if ts, ok := ss[seriesLabels.String()]; ok {
+			seriesLabelsProm := seriesLabels.String()
+			usedExemplarKey := fmt.Sprintf("%s-%d", seriesLabelsProm, e.TimestampMs)
+
+			if usedExemplars[usedExemplarKey] {
+				continue
+			}
+			usedExemplars[usedExemplarKey] = true
+
+			if ts, ok := ss[seriesLabelsProm]; ok {
 				ts.Exemplars = append(ts.Exemplars, e)
-				ss[seriesLabels.String()] = ts
+				ss[seriesLabelsProm] = ts
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


mapno (commented):
I'm not convinced this PR is addressing the irregular sampling behaviour of the `compare()` function.

Running the [query range e2e test](https://github.com/grafana/tempo/blob/main/integration/e2e/query_range_test.go#L24), I get very similar results to before, only with exemplars returned in a different series.

IMO, the first step should be extending that test or creating a new one to be able to verify the behaviour we're looking for: evenly distributed exemplars across all series.

javiermolinar:
> I'm not convinced this PR is addressing the irregular sampling behaviour of the `compare()` function.
> 
> Running the [query range e2e test](https://github.com/grafana/tempo/blob/main/integration/e2e/query_range_test.go#L24), I get very similar results to before, only with exemplars returned in a different series.
> 
> IMO, the first step should be extending that test or creating a new one to be able to verify the behaviour we're looking for: evenly distributed exemplars across all series.

Previously only the series labeled with `baseline_total` and  `selection_total` had exemplars. Now all can have exemplars. This doesn't mean that all the series are going to have at least one exemplar due to the way the bucket works. Sampling by interval only

PR #4296, opened 2024-11-07
[Compaction] Add signals for duplicate rf1 data

joe-elliott:
**What this PR does**:
This PR adds telemetry to help the team understand the impact of compaction of RF1 data. It is likely these are temporary changes to help inform an improvement to prevent duplicate RF1 data. Due to this there is no changelog. Two things were added:

1) A log message when a compaction job completes if the compactor no longer owns that job
2) A metric to count deduped spans per replication factor. Presumably this should show 0 at RF1

Note that I changed the meaning of the returned spans count from the combiner. Nothing was using it so this seemed acceptable.


mdisibio (approved):

zalegrala (approved):
It might be nice to trace this area, but logs are a good start.

PR #4298, opened 2024-11-07
Query-Frontend: Add middleware to drop headers

joe-elliott:
**What this PR does**:
Adds a new middleware to the front of all http pipelines that drops all headers except for a configurable list. This is done primarily as a performance improvement to prevent the unnecessary copying and recopying of headers to all sub jobs.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/config.go
+++ b/modules/frontend/config.go
@@ -32,8 +32,8 @@ type Config struct {
 	// A list of regexes for black listing requests, these will apply for every request regardless the endpoint
 	URLDenyList []string `yaml:"url_deny_list,omitempty"`
 
-	RequestWithWeights bool `yaml:"request_with_weights,omitempty"`
-	RetryWithWeights   bool `yaml:"retry_with_weights,omitempty"`
+	// A list of headers allowed through the HTTP pipeline. Everything else will be stripped.
+	AllowedHeaders []string `yaml:"-"`
 }
 
 type SearchConfig struct {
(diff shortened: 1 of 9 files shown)


javiermolinar (commented):
I think this is going to break Tempo since we are dropping the ` X-Scope-OrgID` header. Maybe we should initialize the AllowedHearders config and hardcode the ones that Tempo needs

joe-elliott:
> I think this is going to break Tempo since we are dropping the  X-Scope-OrgID header. Maybe we should initialize the AllowedHearders config and hardcode the ones that Tempo needs

`prepareRequestForQueriers` adds all required headers for the queriers including X-Scope-OrgID. This is called on all pipelines by the job sharder. the one exception to this is the metrics summary pipeline which doesn't do job sharding. i removed the header strip ware from that pipeline

javiermolinar:
> > I think this is going to break Tempo since we are dropping the  X-Scope-OrgID header. Maybe we should initialize the AllowedHearders config and hardcode the ones that Tempo needs
> 
> `prepareRequestForQueriers` adds all required headers for the queriers including X-Scope-OrgID. This is called on all pipelines by the job sharder. the one exception to this is the metrics summary pipeline which doesn't do job sharding. i removed the header strip ware from that pipeline

Right, I see that we inject the tenant in the auth middleware which happens before stripping the headers

javiermolinar (approved):

PR #4317, opened 2024-11-12
Metrics-Generators: Reduce Wal TruncateFrequency to reduce memory usage

joe-elliott:
**What this PR does**:
Reduces memory usage in the metrics-generators by reducing the amount of data kept in the prom wal. This has been reduced from `2h` to `15m` on recommendation from @mattdurham of the github.com/grafana/alloy team.

The impact is that Tempo can now only suffer a 15m network outage before there is data loss on remote write.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


mdisibio (approved):
This seems reasonable.  Memory usage from the WAL is a significant portion, and the network tolerance seems acceptable.  This pipeline already makes trade-offs vs durability.

mattdurham:
`15m` is what we use for Alloy internally everywhere and rarely have issues with that time scale so it should be reasonable.

PR #4320, opened 2024-11-13
feat: limit tags and tag values search

### LINKED ISSUE #4288: Add a limit parameter to search tags and tag values endpoints
There are currently a few [server side options](https://grafana.com/docs/tempo/latest/configuration/#standard-overrides) that can be set to limit the impact of tag and tag value queries: 

`max_bytes_per_tag_values_query`
`max_blocks_per_tag_values_query`

However, it's likely a user of Tempo may want to purposefully limit the number of tags returned on large queries. Let's add a query parameter allowing the caller of Tempo to indicate the maximum number of results they would like to see.

`/api/v2/searc/tags?limit=<int>`
`/api/v2/search/tag/<tag>/values?limit=<int>`

javiermolinar:
**What this PR does**:

This PR introduces two new optional parameters to enhance the control of Tags and Tag Values lookups:  

- **`limit`**:  
  Specifies the maximum number of items to retrieve. For the `api/v2/tags` endpoint, the search stops if any scope reaches this limit.  

- **`maxStaleValues`**:  
  Restricts the search for tag values. If the number of stale (already known) values meets or exceeds this limit, the search is halted.  This is useful when we have a small set of tag values.

Fixes #4288

**Checklist**
- [x] Tests updated
- [x] Documentation added


joe-elliott (commented):
i'd also like to get your thoughts on a different kind of parameter. perhaps it could be added here as well?

what about a parameter that stopped searching early if N strings were added to the combiner and none were new? so the combiner would keep a running count of the number of added strings since a new one was found and stop searching once we crossed threshold.

consider `/api/v2/search/tag/resource.cluster/values?q=<some very expensive query>`. there are likely very few values for `resource.cluster` but we still spend a lot of resources to exhaustively evaluate `<some very expensive query>`. a parameter that allowed the ingester to just stop searching after N found values and nothing new could be a powerful way to dynamically circuit break.

docs/sources/tempo/api_docs/_index.md
```diff
@@ -360,6 +360,8 @@ Parameters:
   Optional. Along with `end`, defines a time range from which tags should be returned.
 - `end = (unix epoch seconds)`
   Optional. Along with `start`, defines a time range from which tags should be returned. Providing both `start` and `end` includes blocks for the specified time range only.
+- `limit = (integer)`
+  Optional. Limits the maximum number of tags
```
knylander-grafana, line 364:
```suggestion
  Optional. Limits the maximum number of tags.
```

docs/sources/tempo/api_docs/_index.md
```diff
@@ -385,6 +387,8 @@ Parameters:
   Optional. Along with `end` define a time range from which tags should be returned.
 - `end = (unix epoch seconds)`
   Optional. Along with `start` define a time range from which tags should be returned. Providing both `start` and `end` includes blocks for the specified time range only.
+- `limit = (integer)`
+  Optional. Limits the maximum number of tags per scope
```
knylander-grafana, line 391:
```suggestion
  Optional. Limits the maximum number of tags per scope.
```

docs/sources/tempo/api_docs/_index.md
```diff
@@ -515,6 +519,8 @@ Parameters:
   Optional. Along with `end`, defines a time range from which tags should be returned.
 - `end = (unix epoch seconds)`
   Optional. Along with `start`, defines a time range from which tags should be returned. Providing both `start` and `end` includes blocks for the specified time range only.
+- `limit = (integer)`
+  Optional. Limits the maximum number of tags values
```
knylander-grafana, line 523:
```suggestion
  Optional. Limits the maximum number of tags values.
```

knylander-grafana (commented):
Thank you for updating the docs! I have a few minor suggestions to correct missing periods.

pkg/collector/scoped_distinct_string.go
```diff
@@ -4,31 +4,44 @@ import (
...
+		maxTagsPerScope: maxTagsPerScope,
 	}
 }
 
-func NewScopedDistinctStringWithDiff(maxDataSize int) *ScopedDistinctString {
+// NewScopedDistinctStringWithDiff collects the tags per scope with diff
+// MaxDataSize is calculated as the total length of the recorded strings.
+// MaxTagsPerScope controls how many tags can be added per scope. The intrinsic scope is unbounded.
+// For ease of use, maxDataSize=0 and maxTagsPerScope=0 are interpreted as unlimited.
+func NewScopedDistinctStringWithDiff(maxDataSize int, maxTagsPerScope uint32) *ScopedDistinctString {
```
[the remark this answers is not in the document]
javiermolinar: I thought about that as well, if we split the limit between tags (in the same way the maxSize is working right now) we could end with an incomplete list of span tags and none of the resource tags. The way this limit works now cannot be worse than the current implementation but in the worst-case scenario it will end with an exhaustive search.

I'm not opposed to keep the search deterministic by scope.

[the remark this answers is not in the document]
joe-elliott: The best pattern I can come up with for a tag names limit is to stop searching if ANY scope exceeds the limit instead of ALL scopes.

joe-elliott:
The more I think about this idea the more I want it:

> what about a parameter that stopped searching early if N strings were added to the combiner and none were new? so the combiner would keep a running count of the number of added strings since a new one was found and stop searching once we crossed threshold.

The change here and the current max_bytes_per_tag_values_query protect us when a tag value query finds A LOT of matches. We currently have no protection for an expensive tag values query that find a very low number of values.

pkg/collector/scoped_distinct_string.go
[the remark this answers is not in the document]
javiermolinar: Agree I think it's a good compromise

pkg/collector/distinct_value_collector.go
```diff
@@ -52,28 +67,39 @@ func (d *DistinctValue[T]) Collect(v T) (exceeded bool) {
 	if d.limExceeded {
 		return true
 	}
-
 	// Calculate length
 	valueLen := d.len(v)
+	if d.maxDataSize > 0 && d.currDataSize+valueLen >= d.maxDataSize {
+		level.Warn(d.logger).Log("msg", "Max data exceeded", "dataSize", d.currDataSize, "maxDataSize", d.maxDataSize)
```
joe-elliott, line 73: i'm not sure logging lines at such a low level is useful. it would be difficult to pair this up with the related request to understand what happened.

if we want to log the reason for stopping early we should just store the reason as a string in the collector and then log it somewhere like here:

https://github.com/grafana/tempo/pull/4320/files#diff-a1fec687cd87c7167aee355f831ebae6beb368066e3bfbf092ef0aaf5e216aa2R568-R570

joe-elliott: i'm also disliking the amount of boilerplate necessary to pipe the logger object all the way down here. let's store the reason for ending early as suggested and then we can log it at a higher level.

pkg/collector/distinct_value_collector_test.go
```diff
@@ -31,8 +33,52 @@ func TestDistinctValueCollector(t *testing.T) {
...
+	require.Nil(t, res)
+	require.Error(t, err, errDiffNotEnabled)
+}
+
+func TestDistinctValueCollectorWithCacheHitsLimits(t *testing.T) {
+	logger := log.NewNopLogger()
+	d := NewDistinctValue(0, 0, 2, func(s string) int { return len(s) }, logger)
+
+	var stop bool
+	d.Collect("123")
```
joe-elliott, line 63: we should require that these are returning false as expected

pkg/collector/distinct_string_collector_test.go
```diff
@@ -26,8 +28,50 @@ func TestDistinctStringCollector(t *testing.T) {
 	require.Error(t, err, errDiffNotEnabled)
 }
 
+func TestDistinctStringCollectorWithMaxItemsLimit(t *testing.T) {
+	logger := log.NewNopLogger()
+	d := NewDistinctString(0, 3, 0, logger)
+
+	d.Collect("123")
+	d.Collect("4567")
```
joe-elliott, line 36: require false or true as expected. maybe a helper func to keep the code clean?

pkg/tempopb/tempo.proto
```diff
@@ -218,6 +222,8 @@ message SearchTagValuesRequest {
   string query = 2; // TraceQL query
   uint32 start = 4;
   uint32 end = 5;
+  uint32 maxTagValues = 6;
+  uint32 staleValueThreshold = 14; // Limit of stale values
```
joe-elliott, line 226: use 7?

tempodb/encoding/common/interfaces.go
```diff
@@ -43,6 +43,8 @@ type SearchOptions struct {
 	ReadBufferCount        int
 	ReadBufferSize         int
 	BlockReplicationFactor int // Only blocks with this replication factor will be searched. Set to 1 to search generator blocks (RF=1).
+	Limit                  uint32
```
joe-elliott, line 46: i'd prefer not adding these here since these fields only impact tag values and not normal search. it looks like they were added b/c of this code:

https://github.com/grafana/tempo/pull/4320/files#diff-d68ed2a23a87b802e3d85870a36046be8d2f61dd2caeab632503164c9b0ab57dR384

Can we construct the distinct collector thing outside the call and pass it in? is there a clean way to avoid adding these fields?

pkg/collector/distinct_value_collector.go
javiermolinar: It is true that injecting the logger requires a massive amount of changes. I just wanted to propagate the context (the org_id).

Said so, I don't think this is in the hot path; these logs will only be written if we hit a stop condition. Therefore, it is irrelevant to the level they are. The main reason for keeping them there was to not leak the internal implementation of the collector. 

I believe this info can be handy to fine-tune and understand how the searches are working, also it give us useful data. I will store it in string.

tempodb/encoding/common/interfaces.go
javiermolinar: I have opted for a different solution, I'm passing the request in the same way we are doing for `SearchTagValuesV2` that way we standardize the interface contract

joe-elliott: very clean fix. love it ❤️

joe-elliott (commented):
needs a changelog, but otherwise good. great work here.

joe-elliott (approved):
🙏

PR #4325, opened 2024-11-14
TraceQL metrics time range fixes

### LINKED ISSUE #4257: Count over time Traceql query doubles the actual span count randomly
**Describe the bug**
When trying to count the number of spans using `count_over_time` the actual count gets doubled randomly.  Query used was `{} | count_over_time() by (name)`

**To Reproduce**
1. Use https://github.com/grafana/docker-otel-lgtm/tree/main and make sure it uses Tempo version 2.6.0
2. Change backend to Azure blob storage and fill in required details. Mine was as follows,
```
  storage:
    trace:
      backend: azure
      azure:
        container_name: 'containername'
        storage_account_name: 'storageaccount'
        storage_account_key: 'key'
```
3. Start generating traffic using  curl command that can be found [here](https://github.com/grafana/docker-otel-lgtm/blob/main/generate-traffic.sh). Make sure to send traffic without a delay _(wanted to replicate our production scenario hence chose to skip delay)_ and I triggered traffic from command line as follows

```
while true
do
curl -s http://localhost:8081/rolldice
done
```
4. After waiting for almost an hour, go to Grafana and choose Tempo data source and run `{name="roll"} | count_over_time() by (name)` . Please make sure step is set to `1m` and time range is last 1 hour - this is really important as issue happens only when range is set to last 1 hour, if you choose an hour window in the past query works just fine. It will show a time series with a huge spike at a random point `(if you don't see any spike at once just try hitting Run Query every minute )`. Looking the value you could see that its exactly twice the actual value and the next minute the doubling behaviour moves to next point in time.
5. The following snap shot show the buggy behaviour
At 16:26 you could see that the count is 4.87K which is the actual value
Image
After a moment later the count is changed to 9.74K which exactly the double of the actual value.
Image
And then it continuously slides through the window.
Actual Count at 16:27  is 4.73K
Image
Couple of minutes later it is 9.45K
Image

**Expected behavior**
The count should n't get doubled when looking at the last 1 hour span count
**Environment:**
 - Infrastructure: k8s, docker-compose
 - Deployment tool: helm, docker

mdisibio:
**What this PR does**:
Fixes the time range filtering in the jobs that overlapped the handoff between recent and backend data stores (query_backend_after).  Overall change is to disconnect the time range filtering (on start/end) from the step, so that we can accurately split the bucket that falls on that line into the two pieces.

Fixes #4257 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/engine_metrics.go
+++ b/pkg/traceql/engine_metrics.go
@@ -50,6 +50,9 @@ func DefaultQueryRangeStep(start, end uint64) uint64 {
 
 // IntervalCount is the number of intervals in the range with step.
 func IntervalCount(start, end, step uint64) int {
+	start = alignStart(start, step)
+	end = alignEnd(end, step)
+
 	intervals := (end - start) / step
 	intervals++
 	return int(intervals)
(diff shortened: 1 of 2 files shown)


joe-elliott (approved):
lgtm!

PR #4330, opened 2024-11-14
WIP TraceQL: Allow comparison duration to `float`

### LINKED ISSUE #4046: TraceQL: Durations can't be compared to floats
Unsure if this is a feature request or a bug but it makes the language feel inconsistent so I'm writing it up as a bug. There are a few TraceQL constructions with duration that return an error in the fetch layer.

This works b/c 10 / 3 resolves to an int which is compared to the duration:

```
{ span:duration > 10 / 3 }
```

This is functionally the same query but does not work b/c `10ns / 3` resolves to a float and an error bubbles up from the [createIntPredicate](https://github.com/grafana/tempo/blob/34644c5e7b2f47ac1bdc8fd4b3a7c809d63b7ca9/tempodb/encoding/vparquet4/block_traceql.go#L2486-L2501) logic.

```
{ span:duration > 10ns / 3 }
```

Similarly these do not work:
```
{ span:duration > 10  * .3 }
```
```
{ span:duration > 10ns  * .3 }
```

Duration is stored as an int but it makes sense to me that it should be comparable to floats in case someone wants to do math with durations.

ndk:
**What this PR does**:

This PR proposes a solution for comparing `duration` fields with `float` values.

Initially, I attempted to select between `createIntPredicate` and `createFloatPredicate` based on the operand type. However, this approach didn’t work, as the max value used in predicate evaluation is derived directly from column indexes, which are already integer-based for `span:duration`.

It's way simpler just to cast `float` to `int,` allowing us to proceed with the default integer path.

BTW, this expression also causes the issue: `{ span:duration > 10ns * 3 }`

Fixes #4046

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/encoding/vparquet4/block_traceql.go
+++ b/tempodb/encoding/vparquet4/block_traceql.go
@@ -1896,7 +1896,7 @@ func createSpanIterator(makeIter makeIterFn, innerIterators []parquetquery.Itera
 			continue
 
 		case traceql.IntrinsicDuration:
-			pred, err := createIntPredicate(cond.Op, cond.Operands)
+			pred, err := createNumericPredicate(cond.Op, cond.Operands)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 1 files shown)


joe-elliott:
oh, why'd you close this? it's on my list to review.

ndk:
You've linked another issue, and I've realized that my solution covers only particular cases. :)

ndk:
Let's consider two cases.
- `span:duration > 100ns * .3`
We already know the duration is int, so we can just cast a float to an integer to get a correct predicate.

- `span.floatAttr > 123`.
We don't know the type of ' floatAttr ' at the place where the operand predicate is constructed. Having a difference between attribute and operand types causes max values to be wrongly evaluated, which filters out traces. It looks like a universal numeric predicate is needed.

It's how it looks from my perspective. But I'm unfamiliar with `parquet-go` well enough, so I might be wrong.

PR #4331, opened 2024-11-15
Fix incorrect query error when querying at max metrics time range

mdisibio:
**What this PR does**:
The previous PR #4325 introduced a bug where queries at your max time range would incorrectly fail.  This is because the request alignment behavior changed, and it could align it to max+1 more step interval.   This isn't the responsibility of the caller to account for, so this swaps the order of validation and alignment in the frontend.  Reviewing this code again, I think is the correct approach all along.  Checking things after alignment only works if the time range is perfectly divisible by step, so that the time range doesn't increase after alignment.  Currently it doesn't have to be.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


javiermolinar (approved):
It would be nice to have some tests in this area. But we can do it in another moment

PR #4335, opened 2024-11-17
Enforce max span attribute size

### LINKED ISSUE #3985: Add a limit that caps the maximum allowed size of an attribute
When trying to fetch traces that have spans with very large attributes, the Tempo queriers run out of memory and crash. We've observed this when trying to fetch a single trace using the `tracebyID` [endpoint](https://grafana.com/docs/tempo/latest/api_docs/#query)). The trace itself didn't have a ton of spans (roughly 500), but it was very large in size (approximately 250KB). It was very large because some of the spans in that trace had attributes whose values were very large in size. 

One way to avoid these out-of-memory crashes to create a limit on the maximum allowable size of any individual attribute. On the ingest path, Tempo would then reject any spans that had attributes above this size. Or possibly it could store the span, but throw out the specific attribute that was above the size limit. 

If those spans were never ingested, they would not exist to be queried, and therefore could not crash the queriers when fetched. 

**Additional context**
This solution obviously comes with the tradeoff that the very large attribute does not get stored in the database, which may be problematic if the user actually wanted to store and later read back that attribute. 

We're essentially making that assumption that above a certain size, it really doesn't make sense for something to be an attribute value, and likely what's getting sent in is just garbage or possibly the result of a misconfiguration. What that "certain size" is is of course a bit subjective.

ie-pham:
**What this PR does**: Add config for `max_span_attr_size` to enforce the max size a span attribute can be. If a span contains an attribute with either key or value over the limit is sent to Tempo, the attribute key/value will be truncated and marked with "_truncated". 

Fixes #3985

<details><summary> benchmarks</summary>

4 traces, 300 batches and 5000 spans each run

with adding "_truncated"

```
                 │ before.out  │        after_add_truncated.out         │
                 │   sec/op    │   sec/op      vs base                  │
TestsByRequestID   402.3m ± 7%   2384.8m ± 10%  +492.83% (p=0.000 n=10)
```

without adding "_truncated" to truncated attribute values

```
                 │  before.out  │         after-notruncate.out         │
                 │    sec/op    │    sec/op     vs base                │
TestsByRequestID   357.1m ± 52%   594.0m ± 34%  +66.33% (p=0.001 n=10)

                 │  before.out  │      after-notruncate.out      │
                 │     B/op     │     B/op      vs base          │
TestsByRequestID   58.71Mi ± 0%   58.71Mi ± 0%  ~ (p=1.000 n=10)

                 │ before.out  │      after-notruncate.out       │
                 │  allocs/op  │  allocs/op   vs base            │
TestsByRequestID   5.100k ± 0%   5.100k ± 0%  ~ (p=1.000 n=10) ¹
```

</details>

**Checklist**
- [x] Tests updated
- [x] Documentation added


modules/distributor/distributor.go
```diff
@@ -603,7 +616,16 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount int
 		traces = append(traces, r)
 	}
 
-	return keys, traces, nil
+	return keys, traces, discardedSpans, nil
+}
+
+func spanContainsAttributeTooLarge(span *v1.Span, maxAttrSize int) bool {
+	for _, attr := range span.Attributes {
+		if len(attr.Key) > maxAttrSize || attr.Value.Size() > maxAttrSize {
```
mdisibio, line 624: Good call, checking the attribute name.  I'm wondering if this calculation should be against the total Key.Size() + Value.Size()?

modules/distributor/distributor.go
```diff
@@ -526,18 +533,24 @@ func (d *Distributor) UsageTrackerHandler() http.Handler {
...
 	tracesByID := make(map[uint32]*rebatchedTrace, tracesPerBatch)
+	discardedSpans := 0
 
 	for _, b := range batches {
 		spansByILS := make(map[uint32]*v1.ScopeSpans)
 
 		for _, ils := range b.ScopeSpans {
 			for _, span := range ils.Spans {
+				// drop spans with attributes that are too large
+				if spanContainsAttributeTooLarge(span, maxSpanAttrSize) {
```
mdisibio, line 547: If there is a way we could drop only the attribute that seems better than dropping the span (preserve as much information as possible).  It doesn't fit as easily within our current metrics/logs though.

docs/sources/tempo/configuration/_index.md
```diff
@@ -229,6 +229,10 @@ distributor:
     # instruct the client how to retry.
     [retry_after_on_resource_exhausted: <duration> | default = '0' ]
 
+    # Optional
+    # Configures the max size a span attribute can be. Any span with at least one attribute over this limit would be discarded with reason "attribute_too_large"
+    [max_span_attr_size: <int> | default = '10000']
```
mdisibio, line 234: Thinking if this should be `max_attr_bytes` to follow the pattern of other limits, and should it also apply to resource/event/link attributes?

knylander-grafana (commented):
Thank you for adding docs :)

modules/distributor/distributor.go
ie-pham: if key is max/2 and value is max/2, does it still cause a problem?

ie-pham:
benchmark updated but not looking too good :/

modules/distributor/distributor.go
```diff
@@ -603,7 +625,28 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount int
...
+		if len(attr.Key) > maxAttrSize {
+			attr.Key = attr.Key[:maxAttrSize] + "_truncated"
+			*count++
+		}
+
+		switch value := attr.GetValue().Value.(type) {
+		case *v1_common.AnyValue_StringValue:
+			if len(value.StringValue) > maxAttrSize {
+				fmt.Printf("size: %d", len(value.StringValue))
+				value.StringValue = value.StringValue[:maxAttrSize] + "_truncated"
```
joe-elliott, line 643: this is quite expensive as it's going to alloc a completely new string every time.

sane but not as nice option:
1) just produce a metric that indicates number of truncated attributes

crazy but fun option:
2) rewrite the byte array underlying the string to contain "_truncated" at the right position

ie-pham: the benchmark for the scenario where every attribute needs to be truncated, adding the "_truncated" had a ~400+% degradation. And only 66% if we just truncate without adding the "_truncated". I'm in favor of just reporting the metrics without adding "_truncated". Wdyt?

<details><summary> benchmarks</summary>

4 traces, 300 batches and 5000 spans each run

with adding "_truncated"

```
                 │ before.out  │        after_add_truncated.out         │
                 │   sec/op    │   sec/op      vs base                  │
TestsByRequestID   402.3m ± 7%   2384.8m ± 10%  +492.83% (p=0.000 n=10)
```

without adding "_truncated" to truncated attribute values

```
                 │  before.out  │         after-notruncate.out         │
                 │    sec/op    │    sec/op     vs base                │
TestsByRequestID   357.1m ± 52%   594.0m ± 34%  +66.33% (p=0.001 n=10)

                 │  before.out  │      after-notruncate.out      │
                 │     B/op     │     B/op      vs base          │
TestsByRequestID   58.71Mi ± 0%   58.71Mi ± 0%  ~ (p=1.000 n=10)

                 │ before.out  │      after-notruncate.out       │
                 │  allocs/op  │  allocs/op   vs base            │
TestsByRequestID   5.100k ± 0%   5.100k ± 0%  ~ (p=1.000 n=10) ¹
```

</details>

joe-elliott: > rewrite the byte array underlying the string to contain "_truncated" at the right position

this does not work. strings are stored in a read only memory and this will segfault.

docs/sources/tempo/configuration/_index.md
```diff
@@ -229,6 +229,10 @@ distributor:
     # instruct the client how to retry.
     [retry_after_on_resource_exhausted: <duration> | default = '0' ]
 
+    # Optional
+    # Configures the max size an attribute can be. Any key or value that exceeds this limit will be truncated before storing
```
joe-elliott, line 233: we should allow the operator to set this value to 0 to disable the feature.

joe-elliott (approved):
one concern about the logged message, but other than that i think it's good

ogxd:
This was marked as an [ENHANCEMENT] but it actually is a breaking change since it's truncating attributes to 2048 characters by default (which was not the case before)

joe-elliott:
Thanks for the heads up. It looks like we added this in 2.7. The linked PR corrects the changelog. After this merged we will make the same fix to the release.

Appreciate the attention to detail!

PR #4355, opened 2024-11-20
slo: include all request cancellations within SLO

electron0zero:
**What this PR does**:
We count all request cancellations outside SLO, but that's not right because request cancellations are done by users and should fall within SLO.

Request cancellations are now exposed under `result` label in `tempo_query_frontend_queries_total` and `tempo_query_frontend_queries_within_slo_total` with `completed` or `canceled` values to differentiate between completed and canceled requests.

operators can filter `tempo_query_frontend_queries_within_slo_total` with `result=completed` to get the old behaviour of the metric where cancellations where excluded from the  metric.

with the result label, we also expose the information about the cancelled requests, and you can filter on `result=canceled` to see cancelled requests.

I also added a span and span events in the http and gRPC collector to expose more details of the pipeline.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## main / unreleased
 * [FEATURE] tempo-cli: support dropping multiple traces in a single operation [#4266](https://github.com/grafana/tempo/pull/4266) (@ndk)
+* [CHANGE] slo: include all request cancellations within SLO [#4355] (https://github.com/grafana/tempo/pull/4355) (@electron0zero)
+  `tempo_query_frontend_queries_within_slo_total` is now more accurate by including all request cancellations within the SLO.
 * [CHANGE] update default config values to better align with production workloads [#4340](https://github.com/grafana/tempo/pull/4340) (@electron0zero)
 * [CHANGE] fix deprecation warning by switching to DoBatchWithOptions [#4343](https://github.com/grafana/tempo/pull/4343) (@dastrobu)
 * [CHANGE] **BREAKING CHANGE** The Tempo serverless is now deprecated and will be removed in an upcoming release [#4017](https://github.com/grafana/tempo/pull/4017/) @electron0zero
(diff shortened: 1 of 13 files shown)


mdisibio:
Tested this out locally and I think there are still edge cases not handled.  

(1) What works:  if you cancel quickly, within 1-2s, it works.  The query is counted both in _total and _within_slo_total, and confirmed the new logic catches it.

(2) What doesn't work:  if you cancel further into the query, ~4s, then the query continues running and eventually returns in the frontend. But `err` is nil so it is counted in _total, and not necessarily _within_slo.  It still has a chance to be considered outside of SLO.

**Thoughts**
(1) Here is a debugger screenshot from the (2) case, where even though `err == nil`, there is `req.Context().Err == context.Canceled`.   Maybe we could catch this case by checking it here, in addition to `err`.   Or maybe it should be somewhere else, within the roundtripper and bubbled up.

image

(2) In the case above the breakpoint is hit long after pressing the cancel button, so it seems to be the difference between canceling while frontend is still farming out jobs, or too late after all jobs are already issued.  This means maybe the querier is also missing a context check.

**Test setup**
To test this I am using metrics.duration_slo=100ms, and for a slow query:  `{} | rate() by (resource.service.name, name, span:id)` which takes about 9s, so there is time to test canceling at different points.

mdisibio:
@electron0zero Adding some more thoughts about the default behavior of counting canceled requests within SLO:  I think we could also support an operator that wants to exclude them from SLO by adding an additional label that indicates if the request was completed or canceled like `{result="completed|canceled"}`.  Then the operator can exclude canceled requests from SLO calculations, or leave them in.  The default behavior `sum(rate(_within_slo))/sum(rate(_total))` would leave them in, which fits our current direction.  

An additional benefit is that this would give us metrics on canceled requests in general, which I don't think we have today.  Could be useful.

I think this should be feasible and localized to the postSLOHook method where it is checking for cancelations already.  But let me know if there are considerations I missed.

mdisibio (approved):

PR #4359, opened 2024-11-20
[Rhythm] Metrics generator read from kafka first pass

mdisibio:
**What this PR does**:
If enabled, reads from kafka every 2 seconds.  The data is sent to span metrics and service graphs, but not local blocks, to ensure that it doesn't get flushed to storage, which is now the task of the block builder.   

TODO - Maintain a separate local blocks processor to serve queries over recent data (that doesn't flush).  

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/config.go
+++ b/modules/generator/config.go
@@ -38,7 +38,7 @@ type Config struct {
 	QueryTimeout          time.Duration `yaml:"query_timeout"`
 	OverrideRingKey       string        `yaml:"override_ring_key"`
 
-	// This config is dynamically injected because defined outside the ingester config.
+	// This config is dynamically injected because defined outside the generator config.
 	Ingest             ingest.Config `yaml:"-"`
 	AssignedPartitions []int32       `yaml:"assigned_partitions" doc:"List of partitions assigned to this block builder."`
 }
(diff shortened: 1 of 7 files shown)


modules/generator/config.go
```diff
@@ -34,6 +37,10 @@ type Config struct {
 	MetricsIngestionSlack time.Duration `yaml:"metrics_ingestion_time_range_slack"`
 	QueryTimeout          time.Duration `yaml:"query_timeout"`
 	OverrideRingKey       string        `yaml:"override_ring_key"`
+
+	// This config is dynamically injected because defined outside the ingester config.
```
mapno, line 41:
```suggestion
	// This config is dynamically injected because defined outside the generator config.
```

modules/generator/generator_kafka.go
```diff
@@ -0,0 +1,231 @@
...
+						}
+
+						i.pushSpansFromQueue(ctx, &tempopb.PushSpansRequest{
+							Batches: trace.ResourceSpans,
+						})
+					}
+				}
+			}
+		}
+	}
```
mapno, line 150: These loops can be simplified since we're direct consuming and all the fetches belong to the topic/partition defined above

```go
for recIter := fetches.RecordIter(); !recIter.Done(); {
  <code>
}
```

mapno (approved):
I'd prefer to try group consuming, it'd simplify a lot the kafka logic. I'm ok to merge this to get things moving 🚀

mapno (approved):
Nice! 🚀

PR #4365, opened 2024-11-21
Ingester/Generator Live trace cleanup

joe-elliott:
**What this PR does**:
Moves the livetraces struct in the generator into shared location for both the ingester and metrics generator to use. This will cause the ingester to more aggressively (and correctly) refuse live traces.

Other changes:
- Fixed a bug in the live traces struct that was double counting the first batch
- Added a metric to both generator and ingester to count total bytes consumed
- Removed unnecessary and confusing traceCount atomic

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/ingester/instance_test.go
+++ b/modules/ingester/instance_test.go
@@ -570,9 +570,26 @@ func TestInstanceFailsLargeTracesEvenAfterFlushing(t *testing.T) {
 	_, _, traceTooLargeCount = CheckPushBytesError(response)
 	assert.Equal(t, true, traceTooLargeCount > 0)
 
-	// Cut block and then pushing works again
+	// Cut block and then pushing still fails b/c too large traces persist til they stop being pushed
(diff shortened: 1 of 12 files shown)


modules/ingester/instance_test.go
```diff
@@ -241,20 +235,20 @@ func TestInstanceLimits(t *testing.T) {
...
+		// 			req: makeRequestWithByteLimit(300, []byte{}),
+		// 		},
+		// 		{
+		// 			req: makeRequestWithByteLimit(500, []byte{}),
+		// 		},
+		// 		{
+		// 			req: makeRequestWithByteLimit(100, []byte{}),
+		// 		},
+		// 	},
+		// },
```
mapno, line 251: This one should work, right?

mapno (commented):
LGTM, only a few comment

javiermolinar (approved):
Nice cleanup

PR #4383, opened 2024-11-26
Add maximum spans per span set

### LINKED ISSUE #4275: Limit maximun number of spans for span sets
Currently, there is no way to limit the number of spans per span set. This query is ok:

`http://localhost:3200/api/search?q={}&spss=10000`

This can destabilize Tempo. We should add a new config to control this. Tempo must return an error if the spss is greater than the aforementioned limit. This is similar as the `limit` behaviour 

https://grafana.com/docs/tempo/latest/api_docs/#search

carles-grafana:
**What this PR does**:

Add maximum spans per span set.

Fixes #4275

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## main / unreleased
 * [FEATURE] tempo-cli: support dropping multiple traces in a single operation [#4266](https://github.com/grafana/tempo/pull/4266) (@ndk)
+* [CHANGE] **BREAKING CHANGE** Add maximum spans per span set. Users can set `max_spans_per_span_set` to 0 to obtain the old behavior. [#4275](https://github.com/grafana/tempo/pull/4383) (@carles-grafana)
 * [CHANGE] slo: include request cancellations within SLO [#4355] (https://github.com/grafana/tempo/pull/4355) (@electron0zero)
   request cancellations are exposed under `result` label in `tempo_query_frontend_queries_total` and `tempo_query_frontend_queries_within_slo_total` with `completed` or `canceled` values to differentiate between completed and canceled requests.
 * [CHANGE] update default config values to better align with production workloads [#4340](https://github.com/grafana/tempo/pull/4340) (@electron0zero)
--- a/docs/sources/tempo/configuration/_index.md
+++ b/docs/sources/tempo/configuration/_index.md
@@ -640,7 +640,10 @@ query_frontend:
 
         # The number of shards to break ingester queries into.
         [ingester_shards]: <int> | default = 3]
-        
+
+        # The maximum allowed value of spans per span set. 0 disables this limit.
+        [max_spans_per_span_set]: <int> | default = 100]
+
         # SLO configuration for Metadata (tags and tag values) endpoints.
         metadata_slo:
             # If set to a non-zero value, it's value will be used to decide if metadata query is within SLO or not.
--- a/modules/frontend/config.go
+++ b/modules/frontend/config.go
@@ -79,6 +79,7 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
 			ConcurrentRequests:    defaultConcurrentRequests,
 			TargetBytesPerRequest: defaultTargetBytesPerRequest,
 			IngesterShards:        3,
+			MaxSpansPerSpanSet:    100,
 		},
 		SLO: slo,
 	}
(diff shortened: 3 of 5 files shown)


modules/frontend/config.go
```diff
@@ -79,6 +79,7 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
 			ConcurrentRequests:    defaultConcurrentRequests,
 			TargetBytesPerRequest: defaultTargetBytesPerRequest,
 			IngesterShards:        3,
+			MaxSpansPerSpanSet:    1_000,
```
carles-grafana, line 82: Arbitrary number; to decide

modules/frontend/search_sharder.go
```diff
@@ -91,6 +92,10 @@ func (s asyncSearchSharder) RoundTrip(pipelineRequest pipeline.Request) (pipelin
 		return pipeline.NewBadRequest(fmt.Errorf("range specified by start and end exceeds %s. received start=%d end=%d", maxDuration, searchReq.Start, searchReq.End)), nil
 	}
 
+	if searchReq.SpansPerSpanSet > s.cfg.MaxSpansPerSpanSet {
+		return pipeline.NewBadRequest(fmt.Errorf("spans per span set exceeds %d. received %d", s.cfg.MaxSpansPerSpanSet, searchReq.SpansPerSpanSet)), nil
```
carles-grafana, line 96: Not sure how to call it in the error message because in the Web UI the field is called `Span Limit` while in the API it is called `spss`.

modules/frontend/config.go
joe-elliott: 100 feels like a good default to me? that's still quite a few spans.

By putting it in this struct it is already configurable! Please add some docs here:

https://github.com/grafana/tempo/blob/main/docs/sources/tempo/configuration/_index.md?plain=1#L642-L643

modules/frontend/search_sharder.go
joe-elliott: that's a good call out. i think calling it "spans per spanset" is fine here. that's what i would do.

joe-elliott (commented):
Looking good! Needs some doc, changelog and a simple unit test.

modules/frontend/config.go
joe-elliott: > I suppose the manifest YAML is automatically generated?

Nope. If you want to add that to CI, please do.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## main / unreleased
 * [FEATURE] tempo-cli: support dropping multiple traces in a single operation [#4266](https://github.com/grafana/tempo/pull/4266) (@ndk)
+* [CHANGE] Add maximum spans per span set [#4275](https://github.com/grafana/tempo/pull/4383) (@carles-grafana)
```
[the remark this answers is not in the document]
joe-elliott: yeah, that's a good call. list it as a breaking change and note that you can just set the limit to 0 if you want the old behavior.

docs/sources/tempo/configuration/_index.md
```diff
@@ -640,7 +640,10 @@ query_frontend:
 
         # The number of shards to break ingester queries into.
         [ingester_shards]: <int> | default = 3]
-        
+
+        # The maximum allowed value of spans per span set. 0 disables this limit.
+        [max_spans_per_span_set]: <int> | default = 100]
```
javiermolinar, line 645: Should we set it to 0 by default? then it won't be a breaking change

javiermolinar: That's okay to me. It simplify setting up Tempo for sure

jordipolocarres-wk: I think this needs more context so people know what a limit of 100 spansets mean. Are we talking about results of a traceql query?

joe-elliott: this is a limit of 100 spans per spanset. it would be this field in Grafana:

image

naturally this can be adjusted to whatever works for your org (or turned off with 0)

modules/frontend/search_sharder_test.go
```diff
@@ -731,6 +732,12 @@ func TestSearchSharderRoundTripBadRequest(t *testing.T) {
 	resp, err = testRT.RoundTrip(pipeline.NewHTTPRequest(req))
 	testBadRequestFromResponses(t, resp, err, "range specified by start and end exceeds 5m0s. received start=1000 end=1500")
 
+	// spans per span set greater than maximum
+	req = httptest.NewRequest("GET", "/?spss=200", nil)
+	req = req.WithContext(user.InjectOrgID(req.Context(), "blerg"))
```
joe-elliott, line 737: "blerg". well, i guess you're officially on the team now :)

joe-elliott (approved):
please fix the conflict in the changelog and we'll merge!

PR #4391, opened 2024-11-27
TraceQL: support mixed-type attribute querying (int/float)

### LINKED ISSUE #4332: TraceQL: Ints can't be compared to floats
Currently ints can't be compared to floats. For instance this will never find a matching span:

```
{ span.floatAttr = 3 }
```

This may find a matching span:

```
{ span.floatAttr = 3. }
```

Tempo will never fully cast/coerce datatypes. This will never work:

```
{ span.floatAttr = "3" }
```

But I do think it would be useful for ints and floats to be comparable. Let's add support for searching float columns with int values and vice versa.

ndk:
**What this PR does**:
Below is my understanding of the current limitations. Please feel free to correct me if I’ve misunderstood or overlooked something.

Attributes of the same type are stored in the same column. For example, integers are stored in one column and floats in another.

Querying operates in two stages:
- Predicate Creation: Predicates are created based on the operand types.
- Chunk Scanning: Chunks are scanned, and spans are filtered using the predicates.

The issue arises because predicates are generated based on the operand type. If an attribute is stored as a float but the operand is an integer, the predicate evaluates against the integers column instead of the floats column. This results in incorrect behavior.

**Proposed Solution**
The idea is to generate predicates for both integers and floats, allowing both columns to be scanned for the queried attribute.

In this PR, I’ve created a proof-of-concept by copying the existing `createAttributeIterator` function to `createAttributeIterator2`. This duplication is intentional, as the original function is used in multiple places, and I want to avoid introducing unintended side effects until the approach is validated.

```go
case traceql.TypeInt:
	{
		pred, err := createIntPredicate(cond.Op, cond.Operands)
		if err != nil {
			return nil, fmt.Errorf("creating attribute predicate: %w", err)
		}
		attrIntPreds = append(attrIntPreds, pred)
	}

	{
		if i, ok := cond.Operands[0].Int(); ok {
			operands := traceql.Operands{traceql.NewStaticFloat(float64(i))}
			pred, err := createFloatPredicate(cond.Op, operands)
			if err != nil {
				return nil, fmt.Errorf("creating attribute predicate: %w", err)
			}
			attrFltPreds = append(attrFltPreds, pred)
		}
	}

case traceql.TypeFloat:
	{
		operands := traceql.Operands{traceql.NewStaticInt(int(cond.Operands[0].Float()))}
		pred, err := createIntPredicate(cond.Op, operands)
		if err != nil {
			return nil, fmt.Errorf("creating attribute predicate: %w", err)
		}
		attrIntPreds = append(attrIntPreds, pred)
	}

	{
		pred, err := createFloatPredicate(cond.Op, cond.Operands)
		if err != nil {
			return nil, fmt.Errorf("creating attribute predicate: %w", err)
		}
		attrFltPreds = append(attrFltPreds, pred)
	}
```

WDYT? :)

Fixes #4332

**Checklist**
- [x] Tests updated
- [ ] Documentation added


joe-elliott:
I apologize for taking so long to get to this. Your analysis is correct! We do generate predicates per column and, since we store integers and floats independently we only scan one of the columns. Given how small int and float columns tend to be (compared to string columns) I think the performance hit of doing this is likely acceptable in exchange for the nicer behavior.

What is the behavior in this case? I'm pretty sure this will work b/c the engine will request all values for the two attributes and do the work itself. I believe the [engine layer](https://github.com/grafana/tempo/blob/main/pkg/traceql/ast_execute.go#L328) will compare ints and floats correctly but I'm not 100% sure.
  ```
  { span.intAttr > span.floatAttr }
  ```

Tests should also be [added here](https://github.com/grafana/tempo/blob/main/tempodb/tempodb_search_test.go#L45) for the new behavior. These tests build a block and then search for a known trace using a large range of traceql queries. If you add tests here and they pass it means that your changes work from the parquet file all the way up through the engine.

This will also break the "allConditions" optimization if the user types any query with a number comparison:

https://github.com/grafana/tempo/pull/4391/files#diff-a201423ab0b50d4455a497bf1804b1a9f596394413c28b7702710f89237c49c1R2815-R2821

I would like preserve the allConditions behavior in this case b/c it's such a nice optimization and number queries are common. I'm not quite sure why the `len(valueIters) == 1` condition exists so we'd need to do some research into it.

ndk:
> I apologize for taking so long to get to this. Your analysis is correct! We do generate predicates per column and, since we store integers and floats independently we only scan one of the columns. Given how small int and float columns tend to be (compared to string columns) I think the performance hit of doing this is likely acceptable in exchange for the nicer behavior.

Thank you for confirming the approach and pointing out the `allConditions` optimization.  Right now, the fix scans both integer and float columns for attributes that might be either type. I’ve also adjusted how float comparisons work for integer fields, taking into account the fraction part and the comparison operator.

> What is the behavior in this case? I'm pretty sure this will work b/c the engine will request all values for the two attributes and do the work itself. I believe the [engine layer](https://github.com/grafana/tempo/blob/main/pkg/traceql/ast_execute.go#L328) will compare ints and floats correctly but I'm not 100% sure.
> 
> ```
> { span.intAttr > span.floatAttr }
> ```

I verified that `{ span.intAttr > span.floatAttr }` behaves as expected. Wanna me to add a test to cover this case?

> Tests should also be [added here](https://github.com/grafana/tempo/blob/main/tempodb/tempodb_search_test.go#L45) for the new behavior. These tests build a block and then search for a known trace using a large range of traceql queries. If you add tests here and they pass it means that your changes work from the parquet file all the way up through the engine.

Done. Let me know if I missed something.

> This will also break the "allConditions" optimization if the user types any query with a number comparison:
> 
> https://github.com/grafana/tempo/pull/4391/files#diff-a201423ab0b50d4455a497bf1804b1a9f596394413c28b7702710f89237c49c1R2815-R2821
> 
> I would like preserve the allConditions behavior in this case b/c it's such a nice optimization and number queries are common. I'm not quite sure why the `len(valueIters) == 1` condition exists so we'd need to do some research into it.

Regarding the `allConditions` block, the optimization is lost because we generate two predicates (one for int, one for float) under the same attribute name, triggering a `LeftJoinIterator` instead of a `JoinIterator.` Possible workarounds I’m considering:

- Creating a variant of `JoinIterator` that uses logical `OR` rather than `AND.`
- Exploring `parquet.multiRowGroup`, `parquetquery.UnionIterator`, or `parquetquery.KeyValueGroupPredicate` to see if they can unify the int/float search without losing the optimization.
- ~~Refactoring single working with `ColumnChunk` to the multi-one.~~

Given my limited exposure to Tempo’s internals, I’d appreciate any guidance on whether these routes are viable or if there’s a simpler approach to preserve `allConditions`.

ndk:
> I'm fine with what you have now, but feel free to experiment with union if you want.

I'm not sure if it's worth it. I'd rather rely on your opinion here.

> > P.S. Do we care about comparisons with negative values? Should it also be covered?
> Yes, are they not already? reviewing your code I think they would work fine.

Actually, it turned out they didn't work correctly with negative values. I've updated the shifting logic to fix this. Also, another edge case raises questions: what happens if a float hits MaxInt/MinInt? In some cases, it might cause jumps between MaxInt and MinInt.

> I think my primary ask...

Done! Let me know if this aligns with what you had in mind.

Plus, I've added some tests in a separate commit. Feel free to let me know if they look odd or need adjustments.

> I'm a bit impressed you're taking this on. I wouldn't have guessed someone outside of Grafana would have had the time and patience to find this.

Haha, thanks! Honestly, it's just curiosity. Tempo is a fascinating system, and I've been wanting to dive into something challenging like this. It's fun to learn from real-world systems and see how they tackle performance and scalability. :)

joe-elliott:
> Also, another edge case raises questions: what happens if a float hits MaxInt/MinInt? In some cases, it might cause jumps between MaxInt and MinInt.

We could try to get tricky here. Like if you do `{ span.IntCol > IntMaxAsFloat }` then we just don't do the int comparison. `{ span.IntCol < IntMaxAsFloat }` would just return all values from the fetch layer. But I'm also fine with the easy path of just not attempting the float/int comparison is the float is outside the bounds of the int column. It feels like an acceptable edge case as long as we document it.

> Done! Let me know if this aligns with what you had in mind.

Yup, I think this communicates better to a future reader what's going on. Thanks for the change.

Ok, I was running your branch on Friday to test and we do have one final thing to figure out. This query does not work:

```
{ span.http.status_code = 200.0 }
```

The reason is b/c we handle this special column here:

https://github.com/grafana/tempo/blob/14efba0afa06cdb3fae54d03d3138c585463641a/tempodb/encoding/vparquet4/block_traceql.go#L1969-L1986

All well known and dedicated columns are strings ... except this one unfortunately. To do this correctly we have to scan both the well known column as well as the general float attribute column if the static value being compared against http status code is a float. To do this performantly I think we will need to build a `UnionIterator` that joins two sub iterators. One that scans the well known column and one that scans the float attribute column with the appropriate predicate.

ndk:
> We could try to get tricky here. Like if you do `{ span.IntCol > IntMaxAsFloat }` then we just don't do the int comparison. `{ span.IntCol < IntMaxAsFloat }` would just return all values from the fetch layer.

Sounds like a plan. Will do it later. :)

> ```
> { span.http.status_code = 200.0 }
> ```
> ...
> All well known and dedicated columns are strings ... except this one unfortunately. To do this correctly we have to scan both the well known column as well as the general float attribute column if the static value being compared against http status code is a float. To do this performantly I think we will need to build a `UnionIterator` that joins two sub iterators. One that scans the well known column and one that scans the float attribute column with the appropriate predicate.

Oh, that's a nice catch! But before rushing into handling this case, I want to address one quick concern. If a user specifies `span.http.status_code = 200.0`, isn't that likely just a typo? Automatically converting floats to ints might hide the mistake instead of surfacing it. Even though status codes are technically numbers, they're more like categorical values. 199 isn't "slightly less successful" than 200. It's a completely different outcome.

Anyway, if you see real value in covering this edge case, I'm happy to implement it. Let me know what you think!

ndk:
I added a rough coverage of this case. Am I on the right track?

joe-elliott:
It's funny because I really want this PR in, but the only thing blocking it is handling http status code correctly. However, I'd really like to cut a vparquet5 that removes all well known columns (and other cleanup) which would unblock this PR.

ndk:
I believe that I finally learn on how to use UnionInterator properly, but need more time to cover with tests to ensure if I didn't screw something up. I'll move the PR from draft state when I push the updated version. :)

ndk:
What's `vparquet5`? Is something coming?

joe-elliott:
> What's vparquet5? Is something coming?

I wish we had time to work on this. It's an undefined cleanup pass on vParquet with a focus on reducing complexity, number of columns and footer size. One of the things I'd like accomplished is removing the well known columns and instead relying on dedicated columns.

ndk:
> > What's vparquet5? Is something coming?
> 
> I wish we had time to work on this. It's an undefined cleanup pass on vParquet with a focus on reducing complexity, number of columns and footer size. One of the things I'd like accomplished is removing the well known columns and instead relying on dedicated columns.

Lemme know if I can help. :) BTW, I think I’m done (it's already in the PR). I still need to clean up the code and do a bit of refactoring. Also, I'm not totally sure if my code and tests match the style of the existing codebase, so please let me know if anything should be reworked or rearranged.

joe-elliott:
Tested and works! but there's definitely some cleanup to do. 

`{ span.http.status_code = 200 }` and `{ span.http.status_code = 200. }` are doing more work than necessary. They are creating this iterator structure:

```
UnionIterator: 3: %!s(<nil>)
	SyncIterator: rs.list.element.ss.list.element.Spans.list.element.HttpStatusCode : IntEqualPredicate{200}
	LeftJoinIterator: 4: attributeCollector{}
	required:
		SyncIterator: rs.list.element.ss.list.element.Spans.list.element.Attrs.list.element.Key : StringInPredicate{http.status_code}
	optional:
		SyncIterator: rs.list.element.ss.list.element.Spans.list.element.Attrs.list.element.ValueDouble.list.element : FloatEqualPredicate{200.000000}
		SyncIterator: rs.list.element.ss.list.element.Spans.list.element.Attrs.list.element.ValueInt.list.element : IntEqualPredicate{200})
```

but we don't need to scan the generic attribute column for an int. Int values are guaranteed to be stored in the dedicated column for this attribute name so we only need to scan the generic column for a float. This should simplify the iterators to something like:

```
UnionIterator: 3: %!s(<nil>)
	SyncIterator: rs.list.element.ss.list.element.Spans.list.element.HttpStatusCode : IntEqualPredicate{200}
	JoinIterator: 4: attributeCollector{}
		SyncIterator: rs.list.element.ss.list.element.Spans.list.element.Attrs.list.element.Key : StringInPredicate{http.status_code}
		SyncIterator: rs.list.element.ss.list.element.Spans.list.element.Attrs.list.element.ValueDouble.list.element : FloatEqualPredicate{200.000000}
```

unsure why you're seeing nils. I can dig into that a bit. we shouldn't need the filter nil thing.

ndk:
Oh my gosh! This is what happens when a review lasting too long. I started forgetting what I've been doing. :D Fixed.

[`{ span.http.status_code = 200 && span.http.status_code = 200. }`](https://github.com/grafana/tempo/pull/4391/files#diff-6a26fb1b00b9e140d9411c072f2d34042463dba6950f866e3e172adb681221ecR143) -> [iterator structure](https://github.com/grafana/tempo/pull/4391/files#diff-256d15523f8e9a158562150da820069f9eada18501fcd7410a4b6777682c42e4R19)
```
LeftJoinIterator: 3: spanCollector(1)
required: 
	UnionIterator: 3: %!s(<nil>)	
		SyncIterator: rs.list.element.ss.list.element.Spans.list.element.HttpStatusCode : IntEqualPredicate{200}
		SyncIterator: rs.list.element.Resource.Attrs.list.element.ValueDouble.list.element : FloatEqualPredicate{200.000000})
	UnionIterator: 3: %!s(<nil>)	
		SyncIterator: rs.list.element.ss.list.element.Spans.list.element.HttpStatusCode : IntEqualPredicate{200}
		SyncIterator: rs.list.element.Resource.Attrs.list.element.ValueDouble.list.element : FloatEqualPredicate{200.000000})
```

tempodb/encoding/vparquet4/testqt/floatattr.txt
```diff
@@ -0,0 +1,54 @@
+rebatchIterator:
```
joe-elliott, line 1: should we remove these?

joe-elliott (commented):
Awesome! it looks like we were able to get rid of those nil filter shenanigans. I think this is very very close. All functionality is accounted for. I did run some benchmarks and find a regression we should spend some time to understand. I do expect a bit of overhead due to this change but one particular query is showing a 20% increase in cpu.

I can help dig into this.

These are the queries used in the benches. The regression occurred on `traceOrMatch` which you can see below. As you can tell they are crafted for internal data, but they can be rewritten for any block where they get some matches.

```
statuscode:  { span.http.status_code = 200 }
traceOrMatch: { rootServiceName = `tempo-gateway` && (status = error || span.http.status_code = 500)}
complex: {resource.cluster=~"prod.*" && resource.namespace = "tempo-prod" && resource.container="query-frontend" && name = "HTTP GET - tempo_api_v2_search_tags" && span.http.status_code = 200 && duration > 1s}
```

<details>
<summary>benches</summary>

```
goos: darwin
goarch: arm64
cpu: Apple M3 Pro
                                    │ before.txt  │              after.txt              │
                                    │   sec/op    │   sec/op     vs base                │
BackendBlockTraceQL/statuscode-11     64.28m ± 2%   64.81m ± 1%   +0.83% (p=0.043 n=10)
BackendBlockTraceQL/traceOrMatch-11   249.8m ± 8%   303.7m ± 9%  +21.59% (p=0.000 n=10)
BackendBlockTraceQL/complex-11        4.944m ± 5%   4.880m ± 1%        ~ (p=0.190 n=10)
geomean                               42.98m        45.80m        +6.56%

                                    │  before.txt  │              after.txt               │
                                    │     B/s      │     B/s       vs base                │
BackendBlockTraceQL/statuscode-11     341.4Mi ± 2%   338.8Mi ± 1%        ~ (p=0.052 n=10)
BackendBlockTraceQL/traceOrMatch-11   6.695Mi ± 8%   5.541Mi ± 9%  -17.24% (p=0.000 n=10)
BackendBlockTraceQL/complex-11        182.0Mi ± 5%   184.4Mi ± 1%        ~ (p=0.190 n=10)
geomean                               74.66Mi        70.22Mi        -5.94%

                                    │ before.txt  │              after.txt               │
                                    │  MB_io/op   │  MB_io/op    vs base                 │
BackendBlockTraceQL/statuscode-11      23.01 ± 0%    23.02 ± 0%  +0.04% (p=0.000 n=10)
BackendBlockTraceQL/traceOrMatch-11    1.753 ± 0%    1.766 ± 0%  +0.74% (p=0.000 n=10)
BackendBlockTraceQL/complex-11        943.7m ± 0%   943.7m ± 0%       ~ (p=1.000 n=10) ¹
geomean                                3.364         3.373       +0.26%
¹ all samples are equal

BackendBlockTraceQL/statuscode-11      31.19Mi ±  1%   31.29Mi ±  1%       ~ (p=0.436 n=10)
BackendBlockTraceQL/traceOrMatch-11   10.597Mi ± 17%   9.867Mi ± 35%       ~ (p=0.579 n=10)
BackendBlockTraceQL/complex-11         5.387Mi ±  4%   5.413Mi ±  2%       ~ (p=0.631 n=10)
geomean                                12.12Mi         11.87Mi        -2.09%

BackendBlockTraceQL/statuscode-11     378.4k ± 0%   378.6k ± 0%  +0.04% (p=0.000 n=10)
BackendBlockTraceQL/traceOrMatch-11   86.49k ± 1%   86.57k ± 1%       ~ (p=0.218 n=10)
BackendBlockTraceQL/complex-11        79.81k ± 0%   79.83k ± 0%  +0.02% (p=0.000 n=10)
geomean                               137.7k        137.8k       +0.05%
```

</details>

tempodb/encoding/vparquet4/testqt/floatattr.txt
ndk: It seems it's not so easy to run benchmark tests. Need time to figure out how to deal with this. Looks like testblock should be somehow generated. :)
```go
	r, _, _, err := local.New(&local.Config{
		// Path: path.Join("/Users/marty/src/tmp"),
		// Path: path.Join("/Users/mapno/workspace/testblock"),
		Path: path.Join("/Users/joe/testblock"),
	})
	require.NoError(b, err)
```

joe-elliott:
> I cannot reproduce it. Could you tell me how you generated traces?

we generally pull a block generated from internal tracing data which is why the benchmarks contain references to loki and tempo. these blocks generally cover a large range of organically created trace data.

nice work generating a large block. likely some pattern of data internally at Grafana is causing the regression. maybe you should write some float value http status codes and see what happens?

> UPD: I'm wondering how to check if a block has dedicated columns at all.

the meta.json will list all dedicated columns in a block.

i do think there's a bug with the current implementation. fixing it may also resolve the regression. not sure. the query `{ span.http.status_code = 500 }` is generating the following iterators:

```
LeftJoinIterator: 3: spanCollector(1)
required:
	UnionIterator: 3: %!s(<nil>)
		SyncIterator: rs.list.element.ss.list.element.Spans.list.element.HttpStatusCode : IntEqualPredicate{500}
		SyncIterator: rs.list.element.Resource.Attrs.list.element.ValueDouble.list.element : FloatEqualPredicate{500.000000})
optional:
```

I believe it should be this:

```
LeftJoinIterator: 3: spanCollector(1)
required:
	UnionIterator: 3: %!s(<nil>)
		SyncIterator: rs.list.element.ss.list.element.Spans.list.element.HttpStatusCode : IntEqualPredicate{500}
		JoinIterator: 4: attributeCollector{}
			SyncIterator: rs.list.element.ss.list.element.Spans.list.element.Attrs.list.element.Key : StringInPredicate{http.status_code}
			SyncIterator: rs.list.element.ss.list.element.Spans.list.element.Attrs.list.element.ValueDouble.list.element : FloatEqualPredicate{500.000000}
optional:
```

I'm looking into the regression now.

ndk:
> ```
> LeftJoinIterator: 3: spanCollector(1)
> required:
> 	UnionIterator: 3: %!s(<nil>)
> 		SyncIterator: rs.list.element.ss.list.element.Spans.list.element.HttpStatusCode : IntEqualPredicate{500}
> 		JoinIterator: 4: attributeCollector{}
> 			SyncIterator: rs.list.element.ss.list.element.Spans.list.element.Attrs.list.element.Key : StringInPredicate{http.status_code}
> 			SyncIterator: rs.list.element.ss.list.element.Spans.list.element.Attrs.list.element.ValueDouble.list.element : FloatEqualPredicate{500.000000}
> optional:
> ```
> 
> I'm looking into the regression now.

Yeah, I have the same hypothesis. I've just been looking into how to correctly attach filtering by key.

ndk:
> nice work generating a large block. likely some pattern of data internally at Grafana is causing the regression. maybe you should write some float value http status codes and see what happens?

Yep, I also think so. Roaming around the code base I got an impression that dedicated columns aren't something by default. I feel I'm missing something.

joe-elliott:
> Yep, I also think so. Roaming around the code base I got an impression that dedicated columns aren't something by default. I feel I'm missing something.

Dedicated columns need to be configured manually. They allow us to move data from the main attribute columns into their own "dedicated" column. The primary thing we use this for is to move very large attributes out of the main columns like sql queries or large json objects. The secondary use is to isolate important columns for querying. The docs have some details on [how we pick which columns](https://grafana.com/docs/tempo/latest/operations/dedicated_columns/) to configure.

So I reran benchmarks and the regression is fairly intense and likely not something we can accept. I am still working on determining what is causing the regression and seeing if we can improve performance, but my heads down time is limited. This is really cool work and it's so close to mergeable, but the `span.http.status_code` column is frequently queried and we can't sustain a hit like the benchmarks are suggesting. I have not given up hope yet. I'm chasing a few threads, but I do wonder if this work will be more easily mergeable once we drop the "well known" columns. I'm starting a discussion about that here: 

https://github.com/grafana/tempo/issues/4694

<details>
<summary>benches</summary>

```
goos: darwin
goarch: arm64
cpu: Apple M3 Pro
                                                    │ before.txt  │              after.txt              │
                                                    │   sec/op    │   sec/op     vs base                │
BackendBlockTraceQL/spanAttValMatch-11                82.76m ± 1%   81.73m ± 2%   -1.25% (p=0.011 n=10)
BackendBlockTraceQL/spanAttValNoMatch-11              5.468m ± 1%   5.503m ± 1%   +0.64% (p=0.029 n=10)
BackendBlockTraceQL/spanAttIntrinsicMatch-11          58.59m ± 0%   58.41m ± 1%   -0.30% (p=0.005 n=10)
BackendBlockTraceQL/spanAttIntrinsicNoMatch-11        5.383m ± 2%   5.383m ± 1%        ~ (p=0.579 n=10)
BackendBlockTraceQL/resourceAttValMatch-11            394.6m ± 1%   392.6m ± 1%        ~ (p=0.052 n=10)
BackendBlockTraceQL/resourceAttValNoMatch-11          5.095m ± 0%   5.119m ± 0%   +0.46% (p=0.009 n=10)
BackendBlockTraceQL/resourceAttIntrinsicMatch-11      24.86m ± 0%   24.71m ± 0%   -0.59% (p=0.000 n=10)
BackendBlockTraceQL/resourceAttIntrinsicMatch#01-11   5.170m ± 0%   5.167m ± 0%        ~ (p=0.579 n=10)
BackendBlockTraceQL/traceOrMatch-11                   240.7m ± 1%   362.6m ± 0%  +50.66% (p=0.000 n=10)
BackendBlockTraceQL/traceOrNoMatch-11                 241.4m ± 0%   364.0m ± 0%  +50.74% (p=0.000 n=10)
BackendBlockTraceQL/mixedValNoMatch-11                178.4m ± 0%   180.6m ± 0%   +1.18% (p=0.000 n=10)
BackendBlockTraceQL/mixedValMixedMatchAnd-11          5.009m ± 1%   5.021m ± 1%        ~ (p=0.436 n=10)
BackendBlockTraceQL/mixedValMixedMatchOr-11           148.5m ± 0%   149.5m ± 0%   +0.62% (p=0.000 n=10)
BackendBlockTraceQL/count-11                          328.5m ± 0%   327.4m ± 0%   -0.34% (p=0.000 n=10)
BackendBlockTraceQL/struct-11                         429.4m ± 2%   424.2m ± 1%   -1.20% (p=0.011 n=10)
BackendBlockTraceQL/||-11                             152.5m ± 0%   154.4m ± 0%   +1.24% (p=0.000 n=10)
BackendBlockTraceQL/mixed-11                          25.34m ± 0%   25.23m ± 0%   -0.44% (p=0.000 n=10)
BackendBlockTraceQL/complex-11                        4.975m ± 1%   5.002m ± 1%        ~ (p=0.105 n=10)
BackendBlockTraceQL/select-11                         5.009m ± 1%   5.023m ± 1%        ~ (p=0.912 n=10)
geomean                                               40.72m        42.53m        +4.44%

                                                    │     B/s      │     B/s       vs base               │
BackendBlockTraceQL/spanAttValMatch-11                270.6Mi ± 1%   274.0Mi ± 2%  +1.27% (p=0.011 n=10)
BackendBlockTraceQL/spanAttValNoMatch-11              309.0Mi ± 1%   307.0Mi ± 1%  -0.64% (p=0.028 n=10)
BackendBlockTraceQL/spanAttIntrinsicMatch-11          397.6Mi ± 0%   398.8Mi ± 1%  +0.30% (p=0.005 n=10)
BackendBlockTraceQL/spanAttIntrinsicNoMatch-11        460.2Mi ± 2%   460.2Mi ± 1%       ~ (p=0.579 n=10)
BackendBlockTraceQL/resourceAttValMatch-11            55.69Mi ± 1%   55.98Mi ± 1%       ~ (p=0.050 n=10)
BackendBlockTraceQL/resourceAttValNoMatch-11          176.5Mi ± 0%   175.7Mi ± 0%  -0.46% (p=0.007 n=10)
BackendBlockTraceQL/resourceAttIntrinsicMatch-11      885.8Mi ± 0%   891.1Mi ± 0%  +0.59% (p=0.000 n=10)
BackendBlockTraceQL/resourceAttIntrinsicMatch#01-11   180.6Mi ± 0%   180.7Mi ± 0%       ~ (p=0.579 n=10)
BackendBlockTraceQL/traceOrMatch-11                   6.948Mi ± 0%   6.876Mi ± 0%  -1.03% (p=0.000 n=10)
BackendBlockTraceQL/traceOrNoMatch-11                 6.924Mi ± 0%   6.852Mi ± 0%  -1.03% (p=0.000 n=10)
BackendBlockTraceQL/mixedValNoMatch-11                11.05Mi ± 0%   10.91Mi ± 0%  -1.21% (p=0.000 n=10)
BackendBlockTraceQL/mixedValMixedMatchAnd-11          178.2Mi ± 1%   177.8Mi ± 0%       ~ (p=0.436 n=10)
BackendBlockTraceQL/mixedValMixedMatchOr-11           18.57Mi ± 0%   18.46Mi ± 0%  -0.62% (p=0.000 n=10)
BackendBlockTraceQL/count-11                          66.86Mi ± 0%   67.09Mi ± 0%  +0.34% (p=0.000 n=10)
BackendBlockTraceQL/struct-11                         12.72Mi ± 2%   12.87Mi ± 1%  +1.24% (p=0.014 n=10)
BackendBlockTraceQL/||-11                             144.8Mi ± 0%   143.0Mi ± 0%  -1.22% (p=0.000 n=10)
BackendBlockTraceQL/mixed-11                          843.2Mi ± 0%   846.9Mi ± 0%  +0.45% (p=0.000 n=10)
BackendBlockTraceQL/complex-11                        180.9Mi ± 1%   179.9Mi ± 1%       ~ (p=0.093 n=10)
BackendBlockTraceQL/select-11                         179.7Mi ± 1%   179.2Mi ± 1%       ~ (p=0.869 n=10)
geomean                                               101.9Mi        101.8Mi       -0.14%

                                                    │ before.txt  │               after.txt               │
                                                    │  MB_io/op   │  MB_io/op    vs base                  │
BackendBlockTraceQL/spanAttValMatch-11                 23.48 ± 0%    23.48 ± 0%        ~ (p=1.000 n=10) ¹
BackendBlockTraceQL/spanAttValNoMatch-11               1.772 ± 0%    1.772 ± 0%        ~ (p=1.000 n=10) ¹
BackendBlockTraceQL/spanAttIntrinsicMatch-11           24.43 ± 0%    24.43 ± 0%        ~ (p=1.000 n=10) ¹
BackendBlockTraceQL/spanAttIntrinsicNoMatch-11         2.598 ± 0%    2.598 ± 0%        ~ (p=1.000 n=10) ¹
BackendBlockTraceQL/resourceAttValMatch-11             23.04 ± 0%    23.04 ± 0%        ~ (p=1.000 n=10) ¹
BackendBlockTraceQL/resourceAttValNoMatch-11          943.2m ± 0%   943.2m ± 0%        ~ (p=1.000 n=10) ¹
BackendBlockTraceQL/resourceAttIntrinsicMatch-11       23.09 ± 0%    23.09 ± 0%        ~ (p=1.000 n=10) ¹
BackendBlockTraceQL/resourceAttIntrinsicMatch#01-11   979.0m ± 0%   979.0m ± 0%        ~ (p=1.000 n=10) ¹
BackendBlockTraceQL/traceOrMatch-11                    1.753 ± 0%    2.615 ± 0%  +49.17% (p=0.000 n=10)
BackendBlockTraceQL/traceOrNoMatch-11                  1.753 ± 0%    2.615 ± 0%  +49.17% (p=0.000 n=10)
BackendBlockTraceQL/mixedValNoMatch-11                 2.067 ± 0%    2.067 ± 0%        ~ (p=1.000 n=10) ¹
BackendBlockTraceQL/mixedValMixedMatchAnd-11          936.1m ± 0%   936.1m ± 0%        ~ (p=1.000 n=10) ¹
BackendBlockTraceQL/mixedValMixedMatchOr-11            2.893 ± 0%    2.893 ± 0%        ~ (p=1.000 n=10) ¹
BackendBlockTraceQL/count-11                           23.03 ± 0%    23.03 ± 0%        ~ (p=1.000 n=10) ¹
BackendBlockTraceQL/struct-11                          5.726 ± 0%    5.726 ± 0%        ~ (p=1.000 n=10) ¹
BackendBlockTraceQL/||-11                              23.14 ± 0%    23.14 ± 0%        ~ (p=1.000 n=10) ¹
BackendBlockTraceQL/mixed-11                           22.40 ± 0%    22.40 ± 0%        ~ (p=1.000 n=10) ¹
BackendBlockTraceQL/complex-11                        943.7m ± 0%   943.7m ± 0%        ~ (p=1.000 n=10) ¹
BackendBlockTraceQL/select-11                         943.7m ± 0%   943.7m ± 0%        ~ (p=1.000 n=10) ¹
geomean                                                4.351         4.538        +4.30%
¹ all samples are equal

                                                    │  before.txt   │               after.txt                │
                                                    │     B/op      │      B/op       vs base                │
BackendBlockTraceQL/spanAttValMatch-11                45.28Mi ±  1%    45.01Mi ±  1%        ~ (p=0.353 n=10)
BackendBlockTraceQL/spanAttValNoMatch-11              6.693Mi ±  1%    6.716Mi ±  1%        ~ (p=0.796 n=10)
BackendBlockTraceQL/spanAttIntrinsicMatch-11          39.78Mi ±  1%    39.65Mi ±  1%        ~ (p=0.123 n=10)
BackendBlockTraceQL/spanAttIntrinsicNoMatch-11        7.021Mi ±  2%    7.032Mi ±  1%        ~ (p=0.912 n=10)
BackendBlockTraceQL/resourceAttValMatch-11            576.8Mi ±  0%    576.0Mi ±  1%        ~ (p=0.481 n=10)
BackendBlockTraceQL/resourceAttValNoMatch-11          5.308Mi ±  1%    5.362Mi ±  1%        ~ (p=0.063 n=10)
BackendBlockTraceQL/resourceAttIntrinsicMatch-11      10.52Mi ±  1%    10.48Mi ±  1%        ~ (p=0.912 n=10)
BackendBlockTraceQL/resourceAttIntrinsicMatch#01-11   6.768Mi ±  1%    6.763Mi ±  1%        ~ (p=0.912 n=10)
BackendBlockTraceQL/traceOrMatch-11                   9.428Mi ± 20%   12.960Mi ± 12%  +37.46% (p=0.001 n=10)
BackendBlockTraceQL/traceOrNoMatch-11                 10.02Mi ± 23%    13.47Mi ± 15%  +34.47% (p=0.000 n=10)
BackendBlockTraceQL/mixedValNoMatch-11                8.002Mi ± 11%    8.127Mi ±  8%        ~ (p=0.853 n=10)
BackendBlockTraceQL/mixedValMixedMatchAnd-11          5.743Mi ±  1%    5.738Mi ±  1%        ~ (p=0.684 n=10)
BackendBlockTraceQL/mixedValMixedMatchOr-11           7.971Mi ±  8%    8.119Mi ±  8%        ~ (p=0.436 n=10)
BackendBlockTraceQL/count-11                          418.7Mi ±  0%    417.9Mi ±  0%        ~ (p=0.165 n=10)
BackendBlockTraceQL/struct-11                         13.66Mi ± 17%    15.59Mi ± 10%        ~ (p=0.143 n=10)
BackendBlockTraceQL/||-11                             16.74Mi ±  5%    17.05Mi ±  4%        ~ (p=0.912 n=10)
BackendBlockTraceQL/mixed-11                          7.048Mi ±  3%    7.015Mi ±  2%        ~ (p=0.165 n=10)
BackendBlockTraceQL/complex-11                        5.613Mi ±  1%    5.499Mi ±  3%        ~ (p=0.075 n=10)
BackendBlockTraceQL/select-11                         5.554Mi ±  2%    5.531Mi ±  2%        ~ (p=0.631 n=10)
geomean                                               14.61Mi          15.21Mi         +4.11%

BackendBlockTraceQL/spanAttValMatch-11                503.6k ± 0%   498.5k ± 0%  -1.02% (p=0.000 n=10)
BackendBlockTraceQL/spanAttValNoMatch-11              79.47k ± 0%   79.47k ± 0%       ~ (p=0.179 n=10)
BackendBlockTraceQL/spanAttIntrinsicMatch-11          288.1k ± 0%   287.4k ± 0%  -0.25% (p=0.000 n=10)
BackendBlockTraceQL/spanAttIntrinsicNoMatch-11        79.42k ± 0%   79.42k ± 0%       ~ (p=0.155 n=10)
BackendBlockTraceQL/resourceAttValMatch-11            3.450M ± 0%   3.445M ± 0%  -0.16% (p=0.000 n=10)
BackendBlockTraceQL/resourceAttValNoMatch-11          79.47k ± 0%   79.47k ± 0%       ~ (p=0.300 n=10)
BackendBlockTraceQL/resourceAttIntrinsicMatch-11      126.1k ± 0%   125.7k ± 0%  -0.31% (p=0.000 n=10)
BackendBlockTraceQL/resourceAttIntrinsicMatch#01-11   79.46k ± 0%   79.46k ± 0%  -0.00% (p=0.002 n=10)
BackendBlockTraceQL/traceOrMatch-11                   86.53k ± 2%   87.34k ± 2%  +0.93% (p=0.019 n=10)
BackendBlockTraceQL/traceOrNoMatch-11                 86.36k ± 0%   87.04k ± 4%  +0.78% (p=0.000 n=10)
BackendBlockTraceQL/mixedValNoMatch-11                80.29k ± 0%   80.28k ± 0%       ~ (p=0.870 n=10)
BackendBlockTraceQL/mixedValMixedMatchAnd-11          79.46k ± 0%   79.46k ± 0%  -0.00% (p=0.010 n=10)
BackendBlockTraceQL/mixedValMixedMatchOr-11           80.29k ± 0%   80.28k ± 0%       ~ (p=0.516 n=10)
BackendBlockTraceQL/count-11                          1.991M ± 0%   1.985M ± 0%  -0.26% (p=0.000 n=10)
BackendBlockTraceQL/struct-11                         93.63k ± 7%   93.42k ± 2%       ~ (p=0.529 n=10)
BackendBlockTraceQL/||-11                             161.9k ± 0%   161.5k ± 0%  -0.27% (p=0.000 n=10)
BackendBlockTraceQL/mixed-11                          87.86k ± 0%   87.83k ± 0%  -0.02% (p=0.000 n=10)
BackendBlockTraceQL/complex-11                        79.82k ± 0%   79.86k ± 0%  +0.04% (p=0.000 n=10)
BackendBlockTraceQL/select-11                         79.70k ± 0%   79.70k ± 0%  -0.00% (p=0.016 n=10)
geomean                                               147.6k        147.5k       -0.04%
```

</details>

ndk:
Here's what I can suggest from my side:
- Put this PR on hold for now and revisit it later when the timing is better. :)
- Trim down the functionality by dropping support for well-known columns.
- I can try to reproduce the performance regression myself by generating more advanced traces. I have some time to focus on debugging, so I can dedicate effort to tracking down the root cause.

The real question is whether this feature is worth the effort right now. I'll trust your judgment on its importance, as you likely have a better understanding of its priority. :)

ndk:
> * Trim down the functionality by dropping support for well-known columns.

One more thought: Simply cutting the feature might not be the best solution, since the root cause of the regression is still unclear. If we don't fully understand why this slowdown happens, we can't be certain that it's strictly tied to the well-known columns. I think it would be worth identifying the exact bottleneck first before making a decision on whether to drop support for them.

ndk:
I've downloaded a large dataset from Uber: https://zenodo.org/records/13947828.
Unfortunately, there's currently no built-in support for Tempo Parquet in the OpenTelemetry Collector (see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/27284).
However, I don't think it'll be a big issue to parse over 1.3M JSON files and ingest them into Tempo via OTEL/gRPC, although it might take some time. :)
After that, I'll try to reproduce the performance issue using this dataset.

joe-elliott:
> The dictionary load/time is the bottleneck.

Yes, dictionaries can be expensive. For low/medium cardinality columns we have found that they can be a massive performance improvement. Especially when searching for rarely occurring data. They also provide very nice compression.
The biggest issues with dictionaries are that:

- They take the first page of each row group and are loaded as one indivisible unit. You can't partially load a dictionary even if its enormous.
- The byte buffers are not pooled in parquet-go. I have been wanting to PR pooled buffers to parquet-go for awhile, but have not got around to it.

In [vParquet5](https://github.com/grafana/tempo/issues/4694) I am proposing a set of non-dictionary encoded dedicated columns to put high cardinality data to combat the first issue.

> The culprit is that checking the dictionary becomes expensive, even if the key isn't actually there.

The key not being there is actually one of the fastest things that can occur. It allows us to skip the entire row group.

> Below is an example of my attribute-keys column (Key) metadata. It's mostly NULLs and has only two distinct strings (error and grpc.status). Still, the dictionary size is about 316123, and Distinct Values: 0

I believe these stats are for the entire column in that row group and not just the dictionary. Distinct values 0 is weird. I wonder if that's a parquet-go bug. I'm seeing quite similar values for our internal datasets. The reason for the large number of nulls is b/c of the way [parquet nests values](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/36632.pdf). Everytime the structure iterates at a lower level than the column you are currently iterating there is a "null" value for this column b/c it didn't have a value.

These values are not actually encoded into the column's pages. They are encoded into the repetition and definition levels and [reinserted by parquet-go](https://github.com/grafana/parquet-go/blob/main/page_values.go#L23-L30) when you read the values. I have been digging deeper into this b/c of the PR you submitted. I have tried a few approaches not evaluating these nulls at all, but none of them come back with the expected perf improvements. While thinking about this I stumbled on [this perf improvement](https://github.com/grafana/tempo/pull/4753) which also improves our benchmarks. Once this is merged I'm going to try again b/c I think I'm close to a nice improvement for any situation in which heavy null iteration occurs (like yours!).

> If we replace real predicates with false stubs, the problem vanishes:

Yes! but these are also not doing anything :). Your example is returning false from [all these functions](https://github.com/grafana/tempo/blob/main/pkg/parquetquery/predicates.go#L476-L480) which is basically telling the iterator to skip everything.

> Even though the dataset doesn't have span.http.status_code, the predicate still opens the dictionary:

The dictionary must be opened in order to read the column. Even if we didn't have code that dealt with it explicitly it would still happen behind the scenes in parquet-go

Great work here. You are pushing me into details of parquet-go I hadn't reviewed in awhile and finding some nice improvements.

ndk:
If only we could bloom-filter away unnecessary dictionary loading. :)

stoewer:
I just read through the code and this the above conversation. Thanks @ndk for the detailed analysis and all the effort you put into this PR. I really appreciate the level of scrutiny.

> I'm considering another approach: if it's possible to cast http.status_code to an integer without loss of precision, we could store it as an integer instead of a float. That way, we wouldn't need to scan column 63 at all.  

Right now, Tempo doesn’t modify incoming traces - it preserves them exactly as received, even if they contain unsupported attribute types like maps or non-homogeneous arrays. If we were to convert `http.status_code` to an integer, we’d lose that property. Also, while it would be unusual and against OTEL HTTP semantic conventions, OTEL does allow status codes that can’t be represented as integers. So even though such cases are rare (and arguably nonsensical), they’re still valid OTEL tracing data.

stoewer:
> My dataset might still be too synthetic because almost every page has a floating-point http.status_code. So it might not reflect real-world data distributions.

Yes this can be tricky. I had a similar problem with the benchmarks for #4864 (see comments [here](https://github.com/grafana/tempo/pull/4864#discussion_r2006941169) and [here](https://github.com/grafana/tempo/pull/4864#discussion_r2011490866)).

One way to mitigate this could be in the k6 script by modifying the part with `client.push(gen.traces())` to something like this:
```javascript
let traces = gen.traces()
// logic that modifies only one in 1000 traces
client.push(traces)
```

ndk:
Hi! To be honest, I've run out of ideas for fixing the regression without a major overhaul under the hood, which probably reflects my limited familiarity with Tempo's internals. However, I think I can still reuse some parts of this code for #4046.

If you don't see a better path forward either, it might make sense to close this PR, since it seems like a dead end. Anyway, I had a lot of fun working on it, and I believe it was quite educational for me. Let me know if there's anything else I could help with or explore next! :)

stoewer:
I agree - I also don’t see a way to avoid the regressions without more fundamental changes.

That said, I think you’re right that some parts of this code could be reused for #4046. The two issues share similarities, but #4046 doesn’t require setting up an additional iterator for another column, so it should be possible to solve it without the same performance overhead.

Would you be interested in working on #4046? If so, you’re more than welcome!

PR #4395, opened 2024-11-28
[rhythm] Add more metrics

mapno:
**What this PR does**:

Adds metrics to the distributor and block-builder to measure different parameters of the new ingest path.

### DIFF
--- a/modules/blockbuilder/blockbuilder.go
+++ b/modules/blockbuilder/blockbuilder.go
@@ -17,6 +17,8 @@ import (
 	"github.com/grafana/tempo/tempodb"
 	"github.com/grafana/tempo/tempodb/encoding"
 	"github.com/grafana/tempo/tempodb/wal"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/twmb/franz-go/pkg/kadm"
 	"github.com/twmb/franz-go/pkg/kerr"
 	"github.com/twmb/franz-go/pkg/kgo"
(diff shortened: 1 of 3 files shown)


modules/blockbuilder/blockbuilder.go
```diff
@@ -27,6 +29,35 @@ const (
...
+		Subsystem: "block_builder",
+		Name:      "partition_lag",
+		Help:      "Lag of a partition.",
+	}, []string{"partition"})
+	metricConsumeCycleDuration = promauto.NewHistogram(prometheus.HistogramOpts{
+		Namespace:                   "tempo",
+		Subsystem:                   "block_builder",
+		Name:                        "consume_cycle_duration_seconds",
+		Help:                        "Time spent consuming a full cycle.",
+		NativeHistogramBucketFactor: 1.1,
```
electron0zero, line 44: do we want to limit the max buckets by setting NativeHistogramMaxBucketNumber to something high so there is a bound.

will this work for folks who are not using Prometheus v2.40+?

modules/blockbuilder/blockbuilder.go
```diff
@@ -27,6 +29,35 @@ const (
...
+	})
+	metricProcessPartitionSectionDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace:                   "tempo",
+		Subsystem:                   "block_builder",
+		Name:                        "process_partition_section_duration_seconds",
+		Help:                        "Time spent processing one partition section.",
```
electron0zero, line 50: same for this histogram

modules/distributor/distributor.go
```diff
@@ -116,6 +116,30 @@ var (
...
+	metricKafkaRecordsPerRequest = promauto.NewHistogram(prometheus.HistogramOpts{
+		Namespace: "tempo",
+		Subsystem: "distributor",
+		Name:      "kafka_records_per_request",
+		Help:      "The number of records in each kafka request",
+	})
+	metricKafkaWriteLatency = promauto.NewHistogram(prometheus.HistogramOpts{
+		Namespace: "tempo",
+		Subsystem: "distributor",
+		Name:      "kafka_write_latency_seconds",
```
electron0zero, line 128: nit: do we want to use the word Kafka here? or should we use a more generic word like `queue`,  `distributed_queue` or `durable_queue`?

electron0zero (approved):

PR #4396, opened 2024-11-28
[rhythm] Extract block-builder into its own module

mapno:
**What this PR does**:

Moves the block-builder out of the metrics-generator into its own standalone module.


modules/blockbuilder/config.go
```diff
@@ -38,13 +39,26 @@ type Config struct {
...
-	blockConfig BlockConfig `yaml:"block" doc:"Configuration for the block builder."`
+	BlockConfig BlockConfig `yaml:"block" doc:"Configuration for the block builder."`
+	WAL         wal.Config  `yaml:"wal" doc:"Configuration for the write ahead log."`
 
 	// This config is dynamically injected because defined outside the ingester config.
 	IngestStorageConfig ingest.Config `yaml:"-"`
 }
 
 func (c *Config) Validate() error {
+	if c.BlockConfig.BlockCfg.Version != c.WAL.Version {
```
mdisibio, line 50: If the WAL and block versions are required to be the same, then maybe we should swap the validation logic to prevent the WAL version from being set, and always copy over BlockCfg.Version.

mdisibio (approved):
LGTM. One small comment on validation logic.

modules/blockbuilder/config.go
mapno: Good point, I'll change it to that.

PR #4397, opened 2024-11-28
Add query-frontend limit for max length of query expression

electron0zero:
**What this PR does**:

In order to protect tempo from massive queries, it's helpful to put a cap on the max query length.

Having a limit on the size of the query saves us from being DOSed by issuing huge queries against tempo.

Added `query_frontend.max_query_expression_size_bytes` config param for query frontends, defaults to 128 KiB.

this can be configured to a value lower of higher by setting it in query_frontend config section like:

```yaml
query_frontend:
  max_query_expression_size_bytes: 10000
```

users will see it as an error like this:
image

Loki also limits the size of query expression to 128kb and Mimir has a configurable limit with same name (see https://github.com/grafana/mimir/pull/4604), and both were added for same reasons. 

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/frontend/pipeline/async_query_validator_middleware.go
+++ b/modules/frontend/pipeline/async_query_validator_middleware.go
@@ -40,17 +40,18 @@ func (c queryValidatorWare) validateTraceQLQuery(queryParams url.Values) error {
 		traceQLQuery = queryParams.Get("query")
 	}
 	if traceQLQuery != "" {
+		// reject query if the query expression size exceeds the maximum allowed size.
+		// reject huge queries before we parse them, this avoids parsing huge queries.
+		if len(traceQLQuery) > c.maxQuerySizeBytes {
+			return fmt.Errorf("TraceQL expression exceeds the configured maximum size of %d bytes, reduce the query expression size or contact your system administrator", c.maxQuerySizeBytes)
+		}
+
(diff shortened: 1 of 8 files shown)


modules/frontend/pipeline/async_query_validator_middleware.go
```diff
@@ -45,6 +47,11 @@ func (c queryValidatorWare) validateTraceQLQuery(queryParams url.Values) error {
 		if err != nil {
 			return fmt.Errorf("invalid TraceQL query: %w", err)
 		}
+
+		// reject query if the query expression size exceeds the maximum allowed size
+		if len(traceQLQuery) > c.maxQuerySizeBytes {
```
javiermolinar, line 52: We could check the length of the string before parsing the query

javiermolinar (approved):

modules/frontend/pipeline/async_query_validator_middleware.go
electron0zero: nice, this would allow us to skip parsing huge queries.

mapno (approved):
I prefer to have 0 disable the limit, but I'm to merge this PR 👍

PR #4404, opened 2024-12-02
More exemplar fixes

mdisibio:
**What this PR does**:
(1) Fix placeholder exemplars (NaNs) to be attached at the last step in the query-frontend. The previous location was too early and was partial job results instead of the final.
(2) Fix histogram_over_time to use placeholder exemplars. It outputs counts like count_over_time(), this means splitting out its case statement from quantile_over_time.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


mapno (approved):
LGTM

PR #4409, opened 2024-12-02
Fix metrics queries with unscoped attributes

### LINKED ISSUE #3806: TraceQL Metrics: { false } | rate() returns data

mdisibio:
**What this PR does**:
Fixes issues with using unscoped attributes in metrics queries: https://github.com/grafana/tempo/issues/3806#issuecomment-2505900926

The summary is that metrics queries do a lot of work to optimize away the engine callback, relying solely on storage predicate pushdowns.  Unscoped queries like `{ .foo && .bar }` were optimizing it away when they shouldn't.  

~Unfortunately still thinking about the best fix for `{ false } | rate()`.  To the storage layer it looks identical to `{ true } | rate()`.~

Finally introduces some high-level metrics queries tests by writing test blocks with known frequencies of spans and querying them.

Fixes #3806 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


mdisibio:
Pushed a possible fix for `{ false } | rate()` in the query frontend. This is the ideal behavior in that it doesn't execute any jobs or read any blocks. But it's hard to confidently say it works for all variations on falsey queries.

mapno (approved):
LGTM

PR #4411, opened 2024-12-03
Rhythm: Separate non-flushing local blocks processor to store new queue data for reads

mdisibio:
**What this PR does**:
As part of the new architecture rollout, this introduces a secondary instance of the local blocks processor.  It is used to serve traceql metrics queries of recent data received from the queue, and never flushes to object storage.  This isn't the long-term setup, but a transition between the current and new architectures. 

Some notes:
* The secondary local blocks processor lifetime follows the main processor. It is added or removed with it.
* It requires a separate WAL and config to ensure there is never any overlap with the data received to the normal processor.
* Updated some of the WAL config validation.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/blockbuilder/config.go
+++ b/modules/blockbuilder/config.go
@@ -55,7 +55,7 @@ func (c *Config) Validate() error {
 		return fmt.Errorf("block config validation failed: %w", err)
 	}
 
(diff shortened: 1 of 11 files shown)


mapno (approved):
LGTM

PR #4419, opened 2024-12-05
Fix Grpc streaming handling of Authorization header

mdisibio:
**What this PR does**:
The gRPC streaming endpoints are unintentionally dropping auth-related headers when issuing jobs to the queriers, so streaming + auth fails.  We need to extract them from the gRPC metadata and include them in the job.  The fix is here, but it is for downstream [Grafana Enterprise Traces](https://grafana.com/docs/enterprise-traces/latest/)

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott:
Should we put this in `prepareRequestForQueriers`? it's what i generally think of as the place where all the headers are lined up for the queriers:

https://github.com/grafana/tempo/blob/main/modules/frontend/frontend.go#L311

mdisibio:
> Should we put this in `prepareRequestForQueriers`? 

It should be possible since the request parameter's context should still have a copy of the gRPC metadata.  However it is called for every job, so it would be less efficient, and I do like that the current location (grpc handlers) limits the scope.

joe-elliott (approved):
Yup, good call. I do wish there was a way we could centralize the code to make it harder to mess up the next time we add a streaming endpoint.

But I think that about a lot of the boilerplate code between the streaming and http handlers.

PR #4420, opened 2024-12-05
Halt compaction if job is lost

joe-elliott:
**What this PR does**:
In a [previous PR](https://github.com/grafana/tempo/pull/4296) we added a log message when compactors finished a job and recognized it no longer owned the job. We have since seen this message pop up sporadically in production suggesting that unstable compactors will rarely increase the amount of data in object storage.

This PR extends on that work by adding code to abandon compaction in the event that ownership of the job changes. This is not foolproof but should make this an extremely rare occurrence.

~I also considered an approach where we created a child context in `doCompaction` and cancel it with cause in the event that job ownership changes. We could use the cause to confirm that a "context cancelled" that bubbles up from `.compact()` was in fact due to ownership change. The reason I don't like it is b/c we would have to create a side goroutine that polls `.Owns()` watching for ownership change which felt messier. I am definitely game to go this route if others prefer it.~

I have since pivoted to the approach described in the strikethrough for the reason mentioned below. The PR now contains the following changes:

- Cleaned up the compaction loop to use a simple `time.Sleep` to accomplish the same intent.
- Made the start and end compaction cycle log message "Info" level like other status updates for consistency
- Removed <-ctx.Done() and just checked ctx.Err() for simplicity in doCompaction.
- Renamed `doCompaction` and `compact` for clarity and added a `compactWhileOwns` method that does the job of checking job ownership during compaction

I have also left the first attempt as the first commit in this branch and then reverted it if a reviewer would like to see both options.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/compactor.go
+++ b/tempodb/compactor.go
@@ -84,7 +84,7 @@ func (rw *readerWriter) compactionLoop(ctx context.Context) {
 	ticker := time.NewTicker(compactionCycle)
 	defer ticker.Stop()
 	for {
-		doForAtLeast(compactionCycle, func() {
+		doForAtLeast(ctx, compactionCycle, func() {
 			rw.compactOneTenant(ctx)
 		})
 	}
@@ -423,14 +423,20 @@ func (i instrumentedObjectCombiner) Combine(dataEncoding string, objs ...[]byte)
 	return b, wasCombined, err
 }
 
-// doForAtLeast executes the function f. It blocks for at least the passed duration but can go longer
-func doForAtLeast(dur time.Duration, f func()) {
+// doForAtLeast executes the function f. It blocks for at least the passed duration but can go longer. if context is cancelled after
+// the function is done we will bail immediately. in the current use case this means that the process is shutting down
+// we don't force f() to cancel, we assume it also responds to the cancelled context
+func doForAtLeast(ctx context.Context, dur time.Duration, f func()) {
 	startTime := time.Now()
-
 	f()
 	elapsed := time.Since(startTime)
 
 	if elapsed < dur {
-		time.Sleep(dur - elapsed)
+		ticker := time.NewTicker(dur - elapsed)
+
+		select {
+		case <-ticker.C:
+		case <-ctx.Done():
+		}
 	}
 }
--- a/tempodb/compactor.go
+++ b/tempodb/compactor.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"sort"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/go-kit/log/level"
@@ -70,6 +71,8 @@ var (
 		Name:      "compaction_spans_combined_total",
 		Help:      "Number of spans that are deduped per replication factor.",
 	}, []string{"replication_factor"})
+
+	errCompactionJobNoLongerOwned = fmt.Errorf("compaction job no longer owned")
 )
 
 func (rw *readerWriter) compactionLoop(ctx context.Context) {
@@ -81,23 +84,14 @@ func (rw *readerWriter) compactionLoop(ctx context.Context) {
 	ticker := time.NewTicker(compactionCycle)
 	defer ticker.Stop()
 	for {
-		select {
-		case <-ctx.Done():
-			return
-		default:
-		}
-
-		select {
-		case <-ticker.C:
-			rw.doCompaction(ctx)
-		case <-ctx.Done():
-			return
-		}
+		doForAtLeast(compactionCycle, func() {
+			rw.compactOneTenant(ctx)
+		})
 	}
 }
 
-// doCompaction runs a compaction cycle every 30s
-func (rw *readerWriter) doCompaction(ctx context.Context) {
+// compactOneTenant runs a compaction cycle every 30s
+func (rw *readerWriter) compactOneTenant(ctx context.Context) {
 	// List of all tenants in the block list
 	// The block list is updated by constant polling the storage for tenant indexes and/or tenant blocks (and building the index)
 	tenants := rw.blocklist.Tenants()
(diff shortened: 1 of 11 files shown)


tempodb/compactor.go
```diff
@@ -142,51 +136,99 @@ func (rw *readerWriter) doCompaction(ctx context.Context) {
...
-			if start.Add(rw.compactorCfg.MaxTimePerTenant).Before(time.Now()) {
-				measureOutstandingBlocks(tenantID, blockSelector, rw.compactorSharder.Owns)
+	done := make(chan struct{})
+	defer close(done)
 
-				level.Info(rw.logger).Log("msg", "compacted blocks for a maintenance cycle, bailing out", "tenantID", tenantID)
+	// every second test if we still own the job. if we don't then cancel the context with a cause
+	// that we can then test for
+	go func() {
+		ticker := time.NewTicker(1 * time.Second)
```
zalegrala, line 192: Defer the ticker stop?

tempodb/compactor.go
```diff
@@ -142,51 +136,99 @@ func (rw *readerWriter) doCompaction(ctx context.Context) {
...
+
+		for {
+			if !owns() {
+				cancel(errCompactionJobNoLongerOwned)
```
zalegrala, line 196: Return after the cancel here?

tempodb/compactor.go
```diff
@@ -380,3 +422,21 @@ func (i instrumentedObjectCombiner) Combine(dataEncoding string, objs ...[]byte)
 	}
 	return b, wasCombined, err
 }
+
+// doForAtLeast executes the function f. It blocks for at least the passed duration but can go longer. if context is cancelled after
+// the function is done we will bail immediately. in the current use case this means that the process is shutting down
+// we don't force f() to cancel, we assume it also responds to the cancelled context
+func doForAtLeast(ctx context.Context, dur time.Duration, f func()) {
```
zalegrala, line 429: Hmm, I'm trying to understand this.  It looks like this still only executes `f` once no matter the context status.  What am I overlooking?

tempodb/compactor.go
```diff
@@ -380,3 +422,21 @@ func (i instrumentedObjectCombiner) Combine(dataEncoding string, objs ...[]byte)
...
+	startTime := time.Now()
+	f()
+	elapsed := time.Since(startTime)
+
+	if elapsed < dur {
+		ticker := time.NewTicker(dur - elapsed)
```
zalegrala, line 435: Defer the ticker stop.

tempodb/compactor.go
zalegrala: Oh we're holding the blocked call to f for at least duration.

tempodb/compactor.go
```diff
@@ -380,3 +422,21 @@ func (i instrumentedObjectCombiner) Combine(dataEncoding string, objs ...[]byte)
...
+
+		select {
+		case <-ticker.C:
+		case <-ctx.Done():
```
zalegrala, line 439: Return here?

tempodb/compactor.go
joe-elliott: good catch. 🙏

tempodb/compactor.go
joe-elliott: i thought i'd just let it drop down to the select which would bail out, but i like your suggestion. it's more clear.

tempodb/compactor.go
joe-elliott: both cases will just fall through to the end of the function. this basically just blocks until one of the two things occurs.

tempodb/compactor.go
joe-elliott: it was meant to replicate the current logic in a more straight forward way. i'm not 100% sure, but the current logic:
```
	for {
		select {
		case <-ctx.Done():
			return
		default:
		}

		select {
		case <-ticker.C:
			rw.doCompaction(ctx)
		case <-ctx.Done():
			return
		}
	}
```

Basically just attempts to run the `doCompaction` function once every `CompactionCycle`. Also, if compaction takes a long time then the ticker would have elapsed potentially hundreds of times over. I wonder if this is related to our go 1.23 timer panic?

So I tried to write a simpler method that:
- ran the compaction function
- if it took more than the CompactionCycle then immediately return so we can do another
- if it took less time than the CompactionCycle then wait for the difference while also watching for a context cancel.

tempodb/compactor.go
zalegrala: Yeah, its interesting but no issues.

tempodb/compactor.go
```diff
@@ -142,51 +136,101 @@ func (rw *readerWriter) doCompaction(ctx context.Context) {
...
 	start := time.Now()
 
-	level.Debug(rw.logger).Log("msg", "starting compaction cycle", "tenantID", tenantID, "offset", rw.compactorTenantOffset)
+	level.Info(rw.logger).Log("msg", "starting compaction cycle", "tenantID", tenantID, "offset", rw.compactorTenantOffset)
 	for {
-		select {
-		case <-ctx.Done():
+		// this context is controlled by the service manager. it being cancelled means that the process is shutting down
+		if ctx.Err() != nil {
+			level.Info(rw.logger).Log("msg", "caught context cancelled at the top of the compaction loop. bailing.", "err", ctx.Err(), "cause", context.Cause(ctx))
```
zalegrala, line 143: Should we return here?

tempodb/compactor.go
zalegrala: This is clever and I like the approach.  It took a minute to see the reason, but holding a wait period for the cancellation makes sense.

tempodb/compactor.go
```diff
@@ -142,51 +136,101 @@ func (rw *readerWriter) doCompaction(ctx context.Context) {
...
+func (rw *readerWriter) compactWhileOwns(ctx context.Context, blockMetas []*backend.BlockMeta, tenantID string, owns func() bool) error {
+	ownsCtx, cancel := context.WithCancelCause(ctx)
+
+	done := make(chan struct{})
+	defer close(done)
+
+	// every second test if we still own the job. if we don't then cancel the context with a cause
+	// that we can then test for
+	go func() {
+		ticker := time.NewTicker(1 * time.Second)
```
zalegrala, line 192: I see now it appears this 1 second is in conflict with the `compactionCycle`.  If we have a `compactionCycle` less than 1 second, it means we could continue compaction for one extra than desired.  Is this a concern?

zalegrala (commented):
This is neat.  Just a couple comments.

tempodb/compactor.go
joe-elliott: wups, yes

tempodb/compactor.go
joe-elliott: this doesn't block the return of the function so it won't stop the compaction. if a compaction takes < 1s then this method will still return successfully and some short amount of time later this goroutine will exit.

zalegrala (approved):

Wessie:
Hello, I just hit a code path introduced in this patch that made tempo spin a CPU core.

This is because the for loop on [this line](https://github.com/grafana/tempo/blob/main/tempodb/compactor.go#L86) has no exit condition. This means when tempo is exiting (but slowly for some reason), the `ctx` is canceled and this makes `f()` instantly return with this kind of log message:

`caller=compactor.go:143 msg="caught context cancelled at the top of the compaction loop. bailing." err="context canceled" cause="context canceled"` 

and then [this line](https://github.com/grafana/tempo/blob/main/tempodb/compactor.go#L443) also returns instantly inside of `doForAtLeast` because `ctx` is canceled. Causing the problem.

joe-elliott:
@Wessie Great find. Thank you 🙏. Would you like to submit a PR to fix?

If not I will patch it up today, no worries.

Wessie:
Feel free to patch.

PR #4429, opened 2024-12-09
Disable gRPC compression

carles-grafana:
Our benchmark suggests that without compression, queriers and distributors use less CPU and memory.

This setting is configured at the client end.
That is, in the querier and in the ingester and metrics_generator clients of the distributor.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 ## main / unreleased
 
+* [CHANGE] Disable gRPC compression in the querier and distributor for performance reasons [#4429](https://github.com/grafana/tempo/pull/4429) (@carles-grafana)
+  If you would like to re-enable it, we recommend 'snappy'. Use the following settings:
+```
+ingester_client:
+    grpc_client_config:
+        grpc_compression: "snappy"
+metrics_generator_client:
+    grpc_client_config:
+        grpc_compression: "snappy"
+querier:
+    frontend_worker:
+        grpc_client_config:
+            grpc_compression: "snappy"
+```
 * [CHANGE] **BREAKING CHANGE** Add maximum spans per span set. Users can set `max_spans_per_span_set` to 0 to obtain the old behavior. [#4275](https://github.com/grafana/tempo/pull/4383) (@carles-grafana)
 * [CHANGE] slo: include request cancellations within SLO [#4355] (https://github.com/grafana/tempo/pull/4355) (@electron0zero)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## main / unreleased
 * [FEATURE] tempo-cli: support dropping multiple traces in a single operation [#4266](https://github.com/grafana/tempo/pull/4266) (@ndk)
+* [CHANGE] Disable gRPC compression in the querier and distributor. [#4429] (https://github.com/grafana/tempo/pull/4429) (@carles-grafana)
```
joe-elliott, line 3: Let's add some more details here.

> Disable gRPC compression in the querier and distributor for performance reasons.
If you would like to reenable we recommend snappy. Use the following settings:
```
querier:
  <whatever the path to the setting is>
distributor:
  <whatever the path to the setting is>
```

joe-elliott (approved):

joe-elliott:
Looks like there's a changelog conflict. Get that cleaned up and we'll merge!

cmd/tempo/app/config.go
```diff
@@ -116,9 +116,9 @@ func (c *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet) {
 
 	// Everything else
 	flagext.DefaultValues(&c.IngesterClient)
-	c.IngesterClient.GRPCClientConfig.GRPCCompression = "snappy"
+	c.IngesterClient.GRPCClientConfig.GRPCCompression = ""
```
electron0zero, line 119: nit: since this is changing the default values, let's update `docs/sources/tempo/configuration/manifest.md` as well.

electron0zero (approved):

cmd/tempo/app/config.go
carles-grafana: I'm writing a script to generate it automatically and to the check it in CI. I will add it in another PR.

vladst3f:
some before/after benchmarks would have been nice ! otherwise everyone is aware of the tradeoff.

carles-grafana:
> some before/after benchmarks would have been nice ! otherwise everyone is aware of the tradeoff.

I don't have exact numbers but in our testing environment we've seen less CPU usage in queries that return a high number of series. Likewise, in our distributors we saw a significant drop after disabling compression.
image

PR #4430, opened 2024-12-09
Add script to generate manifest with default config

carles-grafana:
The script:
- generates the manifest with default config YAML which is included in the docs as a reference.
- runs in CI to check if the manifest needs to be updated as a result of new changes in the default config.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


.github/workflows/ci.yml
```diff
@@ -103,6 +106,9 @@ jobs:
       - name: Build Tempo
         run: make tempo
 
+      - name: generate-manifest
+        run: go run -v pkg/docsgen/generate_manifest.go
```
mapno, line 110: Should this be a make target?

carles-grafana: personally I'm not a fan of putting this kind of commands in the makefile as it tends to grow a lot, but I'm fine with it if it's the team convention

javiermolinar: I rather use make as well. That's its main purpose

carles-grafana: Something to keep in mind is that every time we run make, some commands are run which add an overhead:
```
# Version number
VERSION=$(shell ./tools/version-tag.sh | cut -d, -f 1)

GIT_REVISION := $(shell git rev-parse --short HEAD)
GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD)
```

For now this is negligible, but as the Makefile grows we are more likely to introduce significant overhead. I've seen this in my previous job where we had a giant Makefile, that's why I'm averse to it and I would rather not use make as a generic task runner.

carles-grafana:
By managing to reuse the build cache, now the new command only takes 2 seconds, I just had the use the exact same build flags GO111MODULE=on CGO_ENABLED=0 in order to reuse the cache

mapno (approved):
LGTM!

PR #4431, opened 2024-12-09
Exit early on autocomplete noops

mdisibio:
**What this PR does**:
This fixes a gap in autocomplete where it's possible to autocomplete an attribute that's already being filtered in the query. This should be a noop but is actually executed.

Example:
`/api/v2/search/tag/name/values?query={name="foo"}`   

Since we already know `name` must be `foo`, there is no reason to scan for the autocomplete values.  

This PR adds the check to ExecuteTagValues in the engine. This is lower than preferred, ideally we would add this to the query-frontend, but the amount of pipeline reuse, and the location of the loose parsing of `query` make that non-trivial.  But want to get this fix in sooner.

Notes:
It's only valid to exit early when it is the Equal operation, and not any case where the attribute is present in the query.  For example these queries are filtering the same tag but can still return a range of values:

* `{ .foo =~ "bar.*"}`
* `{ .foo > 0 }`
* `{ .foo != nil }`

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/engine.go
+++ b/pkg/traceql/engine.go
@@ -238,12 +238,6 @@ func (e *Engine) createAutocompleteRequest(tag Attribute, pipeline Pipeline) Fet
 
 	pipeline.extractConditions(&req)
 
-	// TODO: remove other conditions for the wantAttr we're searching for
-	// for _, cond := range fetchSpansRequest.Conditions {
-	// 	if cond.Attribute == wantAttr {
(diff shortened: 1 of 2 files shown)


mapno (approved):
This [TODO](https://github.com/grafana/tempo/blob/main/pkg/traceql/engine.go#L233-L238) can be cleared up

PR #4436, opened 2024-12-11
feat: include unscoped queries and default the batch size to 7

javiermolinar:
**What this PR does**:
It checks for unscoped attributes to weigh the request. 
It sets the default batch size to 7. Our tests have proved that a batch size of less than 7 in combination with weighted requests introduces a latency penalty due to contention. 

**Checklist**
- [X] Tests updated


carles-grafana (changes requested):
Looks good but the new manifest check has been merged so I think you'll need to rebase on main and run `make generate-manifest``

modules/frontend/pipeline/async_weight_middleware.go
```diff
@@ -114,12 +114,17 @@ func (c weightRequestWare) setTraceQLWeight(req Request) {
 		if c.Op != traceql.OpNone {
 			conditions++
 		}
+
+		if c.Attribute.Intrinsic == traceql.IntrinsicNone && c.Attribute.Scope == traceql.AttributeScopeNone {
```
joe-elliott, line 118: what are we trying to catch here?

joe-elliott: oh, you are looking for unscoped attributes. it may be safer to test `spanRequest.AllConditions` at the bottom?

i'm fine with this, but perhaps do a bit of testing before we merge?

joe-elliott (commented):
good with the change. needs a changelog and a quick test to confirm that the unscoped attributes change is a winner

modules/frontend/pipeline/async_weight_middleware.go
javiermolinar: Do you mean something like?:

```go
if heavyQuery  && !spanRequest.AllConditions{
		req.SetWeight(c.weights.TraceQLSearchWeight + 1)
	}
```

joe-elliott: apologies, i meant to edit this comment and got sidetracked. `AllConditions` will only be true if its a single spanset selector will all conditions `&&`'ed. It will be true even with unscoped attributes, so it will be true on both of these:

```
{ span.foo = "bar" && resource.baz = "bat" }
{  .foo = "bar" && .baz = "bat" }
```

If AllConditions is false -OR- there are unscoped attributes the querier is a heavier query b/c it cannot be evaluated in the fetch layer and requires the engine.

It's possible that both unscoped attributes and AllConditions false should result in `TraceQLSerachWeight + 1` but would need some testing.

knylander-grafana:
Does this impact configuration or how a user would run queries? In other words, are there any changes in this PR that would impact how a user interacts or configures Tempo? If so, we should consider updating the docs.

PR #4446, opened 2024-12-13
Blocklist update fix

mdisibio:
**What this PR does**:
Fixes an issue where the blocklist could get doubled entries for the same block after quick compactions, leading to incorrect results for TraceQL metrics queries.  Searches are unaffected because they dedupe, but the metrics path requires RF1. 

This issue is only present for the SingleBinary, where the compactor and the read paths are in the same process.  In distributed mode this isn't a problem, the indexes shared through object storage are correct.

Example:
* block A  -> compacted to B
* block B -> compacted to C
* both B and C would be in the in-memory blocklist

image

Notes:
tempodb Retention Test had to be updated because it was relying on previous bug.  It expected the same block couldn't be compacted and retained within the same polling cycle, which was because the blocklist didn't work as expected. 

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/blocklist/list.go
+++ b/tempodb/blocklist/list.go
@@ -1,6 +1,7 @@
 package blocklist
 
 import (
+	"slices"
 	"sync"
 
 	"github.com/google/uuid"
@@ -147,11 +148,18 @@ func (l *List) updateInternal(tenantID string, add []*backend.BlockMeta, remove 
 			newblocklist = append(newblocklist, b)
 		}
 	}
-	// add new blocks (only if they don't already exist)
+	// add new blocks (only if they don't already exist) and weren't also removed
 	for _, b := range add {
-		if _, ok := existingMetas[(uuid.UUID)(b.BlockID)]; !ok {
-			newblocklist = append(newblocklist, b)
+		if _, ok := existingMetas[(uuid.UUID)(b.BlockID)]; ok {
+			continue
 		}
+
+		if slices.ContainsFunc(remove, func(b2 *backend.BlockMeta) bool {
+			return b.BlockID == b2.BlockID
+		}) {
+			continue
+		}
+		newblocklist = append(newblocklist, b)
 	}
 
 	l.metas[tenantID] = newblocklist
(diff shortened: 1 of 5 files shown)


tempodb/blocklist/list.go
```diff
@@ -125,64 +125,70 @@ func (l *List) Update(tenantID string, add []*backend.BlockMeta, remove []*backe
...
+		for _, b := range existing {
+			if slices.ContainsFunc(remove, hasID(b.BlockID)) {
+				continue
 			}
+			final = append(final, b)
 		}
-	}
+		// add new if they don't already exist and weren't also removed
+		for _, b := range add {
+			if slices.ContainsFunc(final, hasID(b.BlockID)) ||
```
joe-elliott, line 156: before we built a map for lookups and it looks like we're just doing linear searches here. wdyt think about sorting and doing a binary search?

mdisibio: Sorry I forgot to include the benchmarks but the this version is already significantly faster, mainly by eliminating the maps, which reduces allocs.  

```
pkg: github.com/grafana/tempo/tempodb/blocklist
cpu: Apple M4 Pro
       │   sec/op    │   sec/op     vs base               │
Update   5.485m ± 2%   1.212m ± 2%  -77.91% (p=0.002 n=6)

       │     B/op     │     B/op      vs base               │
Update   6.345Mi ± 0%   1.746Mi ± 0%  -72.48% (p=0.002 n=6)

       │  before.txt   │             after.txt             │
       │   allocs/op   │ allocs/op   vs base               │
Update   4079.000 ± 0%   4.000 ± 0%  -99.90% (p=0.002 n=6)
```

Since the number of metas to add/remove is low (max 4?), sorting/binsearch may not be much better linear, but I can check.

joe-elliott: oh, ok. This is called with just a few metas each compaction to pass the results of the cycle into the polled blocklist. And then each polling cycle it's called once with all accumulated modifications that cycle b/c it assumes that the tenant index is still a bit behind.

given everything the compactor does this may not be noticeable with either a linear or binary search. i'm good as is and will approve. up to you if you want to explore binary search.

joe-elliott (approved):

tempodb/blocklist/list.go
mdisibio: Tested a bit and sorting the current blocklist is outweighing the benefits of the quicker lookups (benchmark using 100K blocklist).  I checked and the even the current implementation isn't showing up in profiles, so I think it's not worth digging more in here.

PR #4454, opened 2024-12-16
Migrate docker-ci-tools workflow to Github Actions

carles-grafana:
**What this PR does**:

Migrate the docker-ci-tools pipeline from Drone to Github Actions.
This includes building the Docker images, pushing them and creating the final manifest.

I've tested this in a branch of the main repo: https://github.com/grafana/tempo/actions/runs/12352580265/job/34469919545

Once merged, I'll test it by updating tools/go.mod and verifying that the produced image works.
Then we can delete the pipeline from Drone.

One change I've made from the original, to make it more efficient, is to trigger the workflow only when the tools directory changes, instead of on every commit.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/.github/workflows/docker-ci-tools.yml
+++ b/.github/workflows/docker-ci-tools.yml
@@ -1,0 +1,76 @@
+name: docker-ci-tools
+on:
+  push:
+    branches:
+      - 'main'
+    paths:
+      - 'tools/**'
+
+env:
+  IMAGE_NAME: grafana/tempo-ci-tools
+
+# Needed to login to DockerHub
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+
+  get-image-tag:
+    runs-on: ubuntu-24.04
+    outputs:
+      tag: ${{ steps.get-tag.outputs.tag }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - id: get-tag
+        run: |
+          echo "tag=$(./tools/image-tag)" >> "$GITHUB_OUTPUT"
+
+  docker-ci-tools:
+    needs: get-image-tag
+    strategy:
+      matrix:
+        runner_arch: [ { runner: ubuntu-24.04, arch: amd64 }, { runner: github-hosted-ubuntu-arm64, arch: arm64 } ]
+    runs-on: ${{ matrix.runner_arch.runner }}
+    env:
+      TAG: ${{ needs.get-image-tag.outputs.tag }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Login to DockerHub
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-ci-tools.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Login to DockerHub
+        uses: grafana/shared-workflows/actions/dockerhub-login@dockerhub-login-v1.0.0
+
+      - name: docker-build-and-push
+        run: |
+          TAG_ARCH="$TAG-${{ matrix.runner_arch.arch }}"
+          docker build -f tools/Dockerfile -t $IMAGE_NAME:$TAG_ARCH .
+          docker push $IMAGE_NAME:$TAG_ARCH
+
+  manifest:
```
javiermolinar, line 52: I think this is not needed since we are not pushing a multi-architecture image yet

javiermolinar: And it can be achieved as well with  docker `build-push-action `

carles-grafana: In Drone we do something very similar to what I'm doing here, that is, build each platform natively and then create a multi-platform manifest -- only that in Drone we use some opaque plugins.

To clarify, this workflow is a combination of these Drone pipelines:
- https://github.com/grafana/tempo/blob/main/.drone/drone.yml#L217
- https://github.com/grafana/tempo/blob/main/.drone/drone.yml#L250
- https://github.com/grafana/tempo/blob/main/.drone/drone.yml#L285

.github/workflows/docker-ci-tools.yml
```diff
@@ -0,0 +1,76 @@
...
+    env:
+      TAG: ${{ needs.get-image-tag.outputs.tag }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Login to DockerHub
+        uses: grafana/shared-workflows/actions/dockerhub-login@dockerhub-login-v1.0.0
+
+      - name: docker-build-and-push
```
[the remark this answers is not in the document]
javiermolinar: I'm not opposed to it but other teams are using it:

https://github.com/grafana/loki/blob/e6d82b9253a46a08120dccf4317fd1d25c1d4ca3/.github/workflows/images.yml#L48

I think it simplifies the process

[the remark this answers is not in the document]
carles-grafana: I say we review this once we have migrated everything, I'm sure there's room for optimization in our CI but I don't want to change things too much during the migration to make sure I don't break any process.

javiermolinar (approved):

PR #4459, opened 2024-12-16
rythm: fix ingestion slack time range

javiermolinar:
**What this PR does**:
It fixes the ingestion slack by calculating it according to the partition's last commit.

Using a fixed time on WAL creation won't work as we want for several reasons. Every partition can have a different offset and the wall is created before reading Kafka. Also, we don't have a way to know this start time in case of wal replay.

The adjustment of the ingestion slack should be done outside the WAL, since we could have different strategies. Another possibility could be storing a function to calculate it but I have opted for this approach that extends the current interface without modifying it.


modules/blockbuilder/partition_writer.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 type partitionSectionWriter interface {
-	pushBytes(tenant string, req *tempopb.PushBytesRequest) error
+	pushBytes(tenant string, req *tempopb.PushBytesRequest, startTime time.Time) error
```
mapno, line 21: `startTime` can be a field of the `writer` since it doesn't change for each section.

modules/blockbuilder/tenant_store.go
```diff
@@ -104,13 +106,55 @@ func (s *tenantStore) resetHeadBlock() error {
...
+			}
+		}
+	}
+	startSeconds = startSeconds / uint64(time.Second)
+	endSeconds = endSeconds / uint64(time.Second)
+
+	return
+}
+
+func (s *tenantStore) adjustTimeRangeForSlack(startTime time.Time, start, end uint64) (uint64, uint64) {
```
mapno, line 139: Since we're adjusting start/end here, we could adjust it to the start and end of the cycle — `[cycleStart-slack , cycleEnd+slack]`. Cycles could be made much longer than the current 5min, and slack would have to increase accordingly.

joe-elliott:
I have not dug into the details of this PR but to share some history about ingestion slack. When we first rolled Tempo out internally it would mark a block's start and time using the min start time and max end time of all spans in the block. We quickly found that every block had a start time of 0 and an end time 100 years in the future due to the data we were consuming. So I created the ingestion slack to prevent this.

Ingestion slack is gross to calculate and I have since just wished I had added 5 minutes to the beginning and end of every block instead of trying to watch all spans and only doing it conditionally.

You all are welcome to tackle this problem however you want, and I'd be glad to discuss it sync if you're like to, but just wanted to give some history.

modules/blockbuilder/tenant_store.go
javiermolinar: Makes sense to me

mapno (commented):
I like the new behaviour, but I'm unsure of changing how `Append` and `AppendTrace` work like that. Specially only changing it for one parquet version. There is no warning to the callers of the functions.

tempodb/encoding/vparquet4/wal_block.go
```diff
@@ -325,7 +329,7 @@ func (b *walBlock) Append(id common.ID, buff []byte, start, end uint32) error {
 	if err != nil {
 		return fmt.Errorf("error preparing trace for read: %w", err)
 	}
-
+	start, end = b.adjustTimeRangeForSlack(start, end)
```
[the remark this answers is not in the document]
javiermolinar: Oh yes, I forgot tho propagate it to the other versions, you are right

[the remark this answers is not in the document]
javiermolinar: You are right, I forgot to do the same change in the other encondings

mapno (commented):
The current changes correctly address the problem, but I still find strange that `Append` and `AppendTrace` have different behaviors. We should either change the method's signature or be very clear with comments or similar.

javiermolinar:
Closing to favor https://github.com/grafana/tempo/pull/4611

PR #4469, opened 2024-12-18
Nomad job example

bcirh:
**What this PR does**:
- Adding HashiCorp Nomad job example (microservices mode)

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/example/nomad/tempo-distributed/config.yml
+++ b/example/nomad/tempo-distributed/config.yml
@@ -1,0 +1,81 @@
+server:
+  log_level: info
+  http_listen_port: {{ env "NOMAD_PORT_http" }}
+  grpc_listen_port: {{ env "NOMAD_PORT_grpc" }}
+
+distributor:
+  ring:
+    kvstore:
+      store: consul
+      prefix: tempo/
+      consul:
+        host: {{ env "attr.unique.network.ip-address" }}:8500
+
+  receivers: # this configuration will listen on all ports and protocols that tempo is capable of.
+    otlp:
+      protocols:
+        http:
+        grpc:
+
+ingester:
+  max_block_duration: 5m
+  lifecycler:
+    ring:
+      kvstore:
+        store: consul
+        prefix: tempo/
+        consul:
+          host: {{ env "attr.unique.network.ip-address" }}:8500
+      replication_factor: 3
+
+compactor:
+  ring:
+    kvstore:
+      store: consul
+      prefix: tempo/
+      consul:
+        host: {{ env "attr.unique.network.ip-address" }}:8500
+
+  compaction:
+    block_retention: 336h  #Duration to keep blocks.  Default is 14 days (336h).
+
+querier:
+  frontend_worker:
+    frontend_address: tempo-query-frontend-grpc.service.consul:9095
+
+metrics_generator:
+  processor:
+    service_graphs:
+      max_items: 10000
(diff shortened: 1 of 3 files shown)


example/nomad/tempo-distributed/config.yml
```diff
@@ -0,0 +1,81 @@
...
+storage:
+  trace:
+    backend: s3
+    wal:
+      path: {{ env "NOMAD_ALLOC_DIR" }}/tempo/wal
+    local:
+      path: {{ env "NOMAD_ALLOC_DIR" }}/tempo/blocks
+    s3:
+      bucket: tempo # how to store data in s3
+      endpoint: seaweedfs-s3.service.consul
```
joe-elliott, line 71: is this a specific endpoint for your internal setup? should we remove it?

example/nomad/tempo-distributed/tempo.hcl
```diff
@@ -0,0 +1,546 @@
...
+      }
+    }
+
+    service {
+      name = "tempo-metrics-generator-grpc"
+      port = "grpc"
+      tags = [
+        "grpc",
+        "traefik.enable=true",
+        "traefik.http.routers.tempo-metrics-generator-grpc.rule=Host(`tempo-metrics-generator-grpc.service.consul`)",
```
joe-elliott, line 53: what do these traefik tags do? is this setting up a service mesh? is that standard in nomad?

example/nomad/tempo-distributed/tempo.hcl
```diff
@@ -0,0 +1,546 @@
...
+        EOH
+
+        destination = "secrets/s3.env"
+        env         = true
+      }
+
+      resources {
+        cpu        = 200
+        memory     = 128
+        memory_max = 2048
```
joe-elliott, line 542: do we want to specify a max? I guess as an example that's fine, but resource usage of Tempo can cover a huge range depending on usage patterns.

joe-elliott (commented):
A few questions. I have no nomad experience, but I'm fine with this PR. The danger is that this becomes out of date and no one on the team knows how to maintain it.

I think it's fine. If the nomad loving community keeps this up to date then we will keep it. If in a year it's had no updates and it no longer works we will likely remove it.

example/nomad/tempo-distributed/config.yml
bcirh: Yes, this is endpoint of my environment. It's safe to leave it as is, it's a generic endpoint.

example/nomad/tempo-distributed/tempo.hcl
bcirh: `memory` value is how much RAM will nomad reserve for the task, `memory_max` value is how much task can expand if it goes above reserved memory (memory oversubscription). `memory_max` values can be removed, in that case value of `memory` will be max value. I would leave it as is, since this is an example, pepole can decide what works best for them. I will update README.md to address this.

example/nomad/tempo-distributed/tempo.hcl
bcirh: You are correct on this. There is no need for traefik tags. I have used traefik tags in my env. I will remove the tags.

joe-elliott (approved):
Thanks for the contribution!

PR #4477, opened 2024-12-19
Distributor shim: add test verifying receiver works (including metrics)

yvrhdn:
**What this PR does**:

Adds a test that runs the OTel receiver shim and tries to write traces in various formats. It also checks the spans_received/spans_refused metrics are set correctly.

I'm adding this because updating OTel dependencies is breaking the metrics and we only have slow integration tests to catch this 😞 

~~Fixes #<issue number>~~

**Checklist**
- [x] Tests updated
- [ ] ~~Documentation added~~ Not needed


modules/distributor/receiver/metrics_provider.go
```diff
@@ -22,184 +22,82 @@ import (
...
-	_ metric.Float64Gauge                   = Float64Gauge{}
-	_ metric.Int64ObservableCounter         = Int64ObservableCounter{}
-	_ metric.Float64ObservableCounter       = Float64ObservableCounter{}
-	_ metric.Int64ObservableGauge           = Int64ObservableGauge{}
-	_ metric.Float64ObservableGauge         = Float64ObservableGauge{}
-	_ metric.Int64ObservableUpDownCounter   = Int64ObservableUpDownCounter{}
-	_ metric.Float64ObservableUpDownCounter = Float64ObservableUpDownCounter{}
-	_ metric.Int64Observer                  = Int64Observer{}
-	_ metric.Float64Observer                = Float64Observer{}
+	_ metric.MeterProvider = &MeterProvider{}
```
javiermolinar, line 31: Do you know if we need the pointer? I think we can keep the value receivers, none of the methods modify the struct state

javiermolinar (approved):
Looks good, have you tested that the metrics are still exposed?

modules/distributor/receiver/metrics_provider.go
yvrhdn: Sure, I've pushed a new commit with just value receivers. Works just as well.

PR #4478, opened 2024-12-19
WIP:  [Rhythm] Full RF1 read path

mdisibio:
**What this PR does**:
I believe these are the changes needed to migrate the backend read path to RF1.  New config options `rf1_after` which is a timestamp, after that only blocks with rf==1 are included in searches and trace by ID lookups.  And an option to discontinue flushes to object storage from the ingesters.

How each api is handled:

Search: the frontend determines the blocks:
```yaml
query_frontend:
  search:
    rf1_after: "2024-12-18T00:00:00Z"
```

Trace lookup: the querier determines the blocks:
```yaml
querier:
  trace_by_id:
    rf1_after: "2024-12-18T00:00:00Z"
```

Tags: this is actually not filtering on replication factor, so no changes needed
Metrics:  already limited to rf1 blocks

The migration path:
* Enable kafka ingest, block-builder, etc. It is now flushing RF1 blocks.
* Grace period for monitoring and validation
* Set `rf1_after` to move the read path over to the new RF1 blocks (recent searches/tags/tracelookups are still directed at ingesters)
* Grace period for monitoring and validation
* Disable flushes on ingesters

Note:  Global config is non-trivial due to Tempo's modular setup, so this adds redundant config in multiple places. Not ideal.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/ingester/flush.go
```diff
@@ -286,11 +286,13 @@ func (i *Ingester) handleComplete(ctx context.Context, op *flushOp) (retry bool,
 
 	// add a flushOp for the block we just completed
 	// No delay
-	i.enqueue(&flushOp{
-		kind:    opKindFlush,
-		userID:  instance.instanceID,
-		blockID: op.blockID,
-	}, false)
+	if i.cfg.FlushObjectStorage {
```
mapno, line 289: We need to mark blocks flushed or change some logic so complete blocks are cleared at some point.

mapno (commented):
Nice, so close to RF1 :rocket:

PR #4480, opened 2024-12-19
[Rhythm] Block-builder consumption loop

mdisibio:
**What this PR does**:
This is an alternate block-builder consumption loop that I think has benefits and would like to get feedback on.  

The curent loop can be thought of as top-down. It calculates the total time range that the block builder is lagging, splits it into smaller sections (i.e. 5 minutes), consume/flush/commit each section.

This new loop is:  while more data, start at last commit and consume/flush/commit another chunk of data (i.e. 5 minutes).

The benefits are:
* less state - Each consume/flush/commit cycle is independent, doesn't require knowledge of the overall state of the queue, or side effects from previous loops
* fewer kafka apis are involved, (i.e. no CalculateGroupLag)
* doesn't require custom commit metadata
* I think core loop is simpler and will make it easier to iterate.  For example there is a TODO to round-robin when the block-builder is assigned to multiple partitions and they are all lagging.  I think this is more complex to try in the current design. 

The drawbacks are:
~* Mainly around how we want to metric "lag".  The current metric is number of messages, but I think finding a way to determine length of time (i.e. "the block-builder is 15 minutes behind") is more useful.  However you have to read a record to determine that, so this needs more work.~

TODO
- ~This is a draft, and needs more finishing touches before it can be merged.  What I want is to think about the overall loop structure.  Logs/cleanup/finishing touches would be added before any merge.~
- ~Depends on test updates in https://github.com/grafana/tempo/pull/4510 which setup support for consumer groups.~

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/blockbuilder/blockbuilder.go
```diff
@@ -167,6 +172,165 @@ func (b *BlockBuilder) running(ctx context.Context) error {
...
 
+func (b *BlockBuilder) running(ctx context.Context) error {
+	// Initial delay
+	waitTime := 0 * time.Second
+	for {
+		select {
+		case <-time.After(waitTime):
+			err := b.consume(ctx)
+			if err != nil {
+				b.logger.Log("msg", "consumeCycle failed", "err", err)
```
javiermolinar, line 183:
```suggestion
				level.Error(b.logger).Log("msg", "consumeCycle failed", "err", err)
```

modules/blockbuilder/blockbuilder.go
```diff
@@ -167,6 +172,165 @@ func (b *BlockBuilder) running(ctx context.Context) error {
...
+		lastRec     *kgo.Record
+		begin       time.Time
+	)
+
+	commits, err := b.kadm.FetchOffsetsForTopics(ctx, group, topic)
+	if err != nil {
+		return false, err
+	}
+
+	lastCommit, ok := commits.Lookup(topic, partition)
```
javiermolinar, line 241:
```suggestion
	lastCommit, exists := commits.Lookup(topic, partition)
	if exists && lastCommit.At >= 0 {
		startOffset = startOffset.At(lastCommit.At)
	} else {
		startOffset = kgo.NewOffset().AtStart()
	}
```

https://pkg.go.dev/github.com/twmb/franz-go/pkg/kadm@v1.14.0#OffsetResponses.Lookup

modules/blockbuilder/blockbuilder.go
```diff
@@ -167,6 +172,165 @@ func (b *BlockBuilder) running(ctx context.Context) error {
...
+
+				writer = newPartitionSectionWriter(b.logger, int64(partition), rec.Offset, b.cfg.BlockConfig, b.overrides, b.wal, b.enc)
+			}
+
+			if rec.Timestamp.Before(begin) || rec.Timestamp.After(end) {
+				break outer
+			}
+
+			err := b.pushTraces(rec.Key, rec.Value, writer)
+			if err != nil {
```
javiermolinar, line 305: What will happen if something is wrong with the WAL? I guess it will enter in a loop

javiermolinar (commented):
Looks good!

modules/blockbuilder/blockbuilder.go
```diff
@@ -27,14 +28,15 @@ import (
...
+	pollTimeout             = 2 * time.Second
 )
 
 var (
 	metricPartitionLag = promauto.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: "tempo",
 		Subsystem: "block_builder",
-		Name:      "partition_lag",
-		Help:      "Lag of a partition.",
+		Name:      "partition_lag_s",
```
mapno, line 38:
```suggestion
		Name:      "partition_lag_seconds",
```

https://prometheus.io/docs/practices/naming/#metric-names

mapno: IMO the number of pending records is also relevant. I think we should keep the original metric and add one for time.

modules/blockbuilder/blockbuilder.go
```diff
@@ -167,6 +172,165 @@ func (b *BlockBuilder) running(ctx context.Context) error {
...
+
+			// Initialize if needed
+			if writer == nil {
+
+				// Determine begin and end time range, which is -/+ cycle duration.
+				// But don't exceed the given overall end time.
+				begin = rec.Timestamp.Add(-dur)
+				if rec.Timestamp.Add(dur).Before(end) {
+					end = rec.Timestamp.Add(dur)
+				}
```
mapno, line 293: This feels like a strange side-effect of the `writer` being `nil`. Cycle initialisation could be consolidated in one place instead.

mapno (commented):
Nice job on reducing the number of API calls. The way we calculate cycle sections is complex and should be simplified.

I have some concerns on losing visibility on the state of partitions. Before it was easy to see exactly how many pending messages there were and what was being consumed. I feel we lose that a bit with this PR.
I believe lag can still be periodically polled with fewer lines than currently and we'd keep that functionality. We might be able to delete the fallback time entirely.

modules/blockbuilder/blockbuilder.go
```diff
@@ -167,6 +172,165 @@ func (b *BlockBuilder) running(ctx context.Context) error {
...
+
+				metricPartitionLag.WithLabelValues(strconv.Itoa(int(partition))).Set(time.Since(rec.Timestamp).Seconds())
+
+				writer = newPartitionSectionWriter(b.logger, int64(partition), rec.Offset, b.cfg.BlockConfig, b.overrides, b.wal, b.enc)
+			}
+
+			if rec.Timestamp.Before(begin) || rec.Timestamp.After(end) {
```
mapno, line 300: Unsure if we should break for records being too old. The record ts is set by the produced—the distributor in this case.

If a record was so old as to be outside of the cycle's start time, I think we'd want to put it in the current block. Otherwise it could impact consumption too much, creating too small blocks.

modules/blockbuilder/blockbuilder.go
mdisibio: 👍 re-added, and it is polled in a separate go routine.

modules/blockbuilder/blockbuilder.go
mdisibio: This is the cycle initialization. Swapped to use an `init` bool which should be clearer.

modules/blockbuilder/blockbuilder.go
mdisibio: I was thinking that if something is too old (older than the cycle time, i.e. minutes), then it likely has old traces too.   But don't have strong opinions.  Removed the check.

mdisibio:
> lag can still be periodically polled with fewer lines than currently and we'd keep that functionality. We might be able to delete the fallback time entirely.

Good call, re-added the original metric, polled it in a separate goroutine, and removed the fallback logic.

modules/blockbuilder/blockbuilder.go
mdisibio: I think `ok` is more idiomatic, and the lib is reading from a map internally anyway.

mapno (commented):
`commit_meta.go` and `commit_meta_test.go` can also be deleted now

modules/blockbuilder/blockbuilder.go
```diff
@@ -142,265 +145,231 @@ func (b *BlockBuilder) starting(ctx context.Context) (err error) {
...
 
-	// TODO - Commit with backoff
-	adm := kadm.NewClient(b.kafkaClient)
-	err := adm.CommitAllOffsets(ctx, b.cfg.IngestStorageConfig.Kafka.ConsumerGroup, offsets)
-	if err != nil {
-		return fmt.Errorf("failed to commit offsets: %w", err)
+	for {
+		select {
+		case <-time.After(waitTime):
+			metricPartitionLag.Reset()
```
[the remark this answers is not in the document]
mdisibio: Not strictly necessary, but I thought it would help metrics quality when a partition is goes inactive.  Without a reset it would keep reporting the last lag. But thinking about it closer, that should mean the last reported lag is zero.  If you are good with it, will remove it.  (Didn't do it for the other metrics either)

mapno (approved):
LGTM

PR #4483, opened 2024-12-20
[Ingester] Create one goroutine per tenant to flush traces to disk

joe-elliott:
**What this PR does**:
We have identified a failure mode in the ingesters due to using a [single goroutine](https://github.com/grafana/tempo/blob/main/modules/ingester/ingester.go#L143-L159) to flush all live traces to disk. If there is a heavy query or other event that causes resource starvation this goroutine will fall behind and an ingester's memory will balloon and it will start refusing traces with a LIVE_TRACES_EXCEEDED error. In an extreme case it will OOM.

This PR creates one goroutine per tenant that manages the flushing of traces to disk for that tenant. This is similar to the [local blocks](https://github.com/grafana/tempo/blob/main/modules/generator/processor/localblocks/processor.go#L120-L121) in the metrics generator. I would have preferred to have the goroutine lifecycle managed by the instance itself, but that would be counter to the way things are currently designed and would have required more changes.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/ingester/flush.go
```diff
@@ -72,20 +72,6 @@ const (
 	opKindFlush
 )
 
-// Flush triggers a flush of all in memory traces to disk.  This is called
-// by the lifecycler on shutdown and will put our traces in the WAL to be
-// replayed.
-func (i *Ingester) Flush() {
```
joe-elliott, line 78: this is the old way that we flushed traces to disk on shutdown. it was difficult to find and driven through an obsolete ring mechanic "FlushTransfer" so I removed it and moved it to "stopping". I think the new way is more easily discoverable and clear.

https://github.com/grafana/tempo/pull/4483/files#diff-b17cd433ae9859f67f0056e452237ddf70a44e9821c6128c8990005eaf7decd1R176

modules/ingester/flush.go
```diff
@@ -124,9 +110,9 @@ func (i *Ingester) FlushHandler(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		level.Info(log.Logger).Log("msg", "flushing instance", "instance", instance.instanceID)
-		i.sweepInstance(instance, true)
+		i.cutOneInstanceToWal(instance, true)
 	} else {
-		i.sweepAllInstances(true)
+		i.cutAllInstancesToWal()
```
joe-elliott, line 115: renamed these funcs for clarity

modules/ingester/flush.go
```diff
@@ -151,16 +137,47 @@ func (o *flushOp) Priority() int64 {
 	return -o.at.Unix()
 }
 
-// sweepAllInstances periodically schedules series for flushing and garbage collects instances with no series
-func (i *Ingester) sweepAllInstances(immediate bool) {
+// cutToWalLoop kicks off a goroutine for the passed instance that will periodically cut traces to WAL.
+// it signals completion through cutToWalWg, waits for cutToWalStart and stops on cutToWalStop.
+func (i *Ingester) cutToWalLoop(instance *instance) {
```
joe-elliott, line 142: the new, per tenant loop, that drives flushing live traces to disk

modules/ingester/ingester.go
```diff
@@ -183,49 +190,38 @@ func (i *Ingester) starting(ctx context.Context) error {
...
+	// accept traces
 	i.pushErr.Store(nil)
 
+	// start flushing traces to wal
+	close(i.cutToWalStart)
+
 	return nil
 }
 
 func (i *Ingester) loop(ctx context.Context) error {
```
mapno, line 202:
```suggestion
func (i *Ingester) running(ctx context.Context) error {
```

`loop` never truly looped, but now it even makes less sense than before

mapno (approved):
LGTM

modules/ingester/ingester.go
joe-elliott: good call. i think that also makes it more consistent with the other modules

PR #4494, opened 2024-12-26
Add endpoint to distributor to push tempopb.Trace

yvrhdn:
**What this PR does**:

Adds a new endpoint to the distributor that accepts `tempopb.Trace` directly.

This makes it possible to push tempopb directly without passing through the OTel Collector receiver. It avoids work associated with converting `ptrace.Traces` to `tempopb.Trace`.

I have also extended `tempopb/trace_utils.go` with handy functions to count spans in a trace and to convert tempopb to OTLP and back.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


yvrhdn:
I think the idea of adding an endpoint for pushing tempopb directly is fine, but if we have otlp converting it to tempopb to send it ends up being the same work, just in a different place.

PR #4525, opened 2025-01-07
enforce max series for metrics queries

### LINKED ISSUE #4219: Limit series produced by TraceQL Metrics
Create a global limit for the number of series produced by TraceQL metrics. If the number of metrics are trimmed it would be ideal to include a message about the cutoff.

I don't think this needs to be per tenant. Tenants of any size can easily generate extremely high series TraceQL metrics queries.

ie-pham:
**What this PR does**: Add config to enforce max time series returned in a metrics query. This is enforced at 4 levels: front-end combiner, querier combiner, metrics-generator local blocks, and metrics evaluation. The configuration is set in the query-frontend config and is passed to all levels as `maxSeries` in the `QueryRangeRequest` proto. 

new config: max_response_series <default 1000>
Setting the value to 0 will disable this feature. 

approach : Keep track of number of series by calling Length() until limit is reached. Whatever series were generated up to this point will be truncated at the limit and returned as partial results. This may mean that the partial response will contain values that are inaccurate. 

Fixes #4219

**Checklist**
- [x] Tests updated
- [x] Documentation added
- [ ] Once merged please let know @grafana/observability-traces-and-profiling that https://github.com/grafana/grafana/issues/103595 can be worked on

<details>
<summary>benchmarks</summary>

```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
                                                                                                       │  before.txt  │            after1000.txt            │
BackendBlockQueryRange/{}_|_rate()/5                                                                     201.5m ± 11%   205.9m ± 16%       ~ (p=0.971 n=10)
BackendBlockQueryRange/{}_|_rate()_by_(span.http.status_code)/5                                          443.8m ±  2%   446.0m ±  3%       ~ (p=0.739 n=10)
BackendBlockQueryRange/{}_|_rate()_by_(resource.service.name)/5                                          251.8m ±  1%   246.8m ±  6%       ~ (p=0.123 n=10)
BackendBlockQueryRange/{}_|_rate()_by_(span.http.url)/5                                                  433.3m ±  2%   440.9m ±  8%       ~ (p=0.529 n=10)
BackendBlockQueryRange/{resource.service.name=`loki-ingester`}_|_rate()/5                                3.721m ±  3%   3.710m ±  4%       ~ (p=0.739 n=10)
BackendBlockQueryRange/{span.http.host_!=_``_&&_span.http.flavor=`2`}_|_rate()_by_(span.http.flavor)/5   136.1m ±  6%   146.8m ±  9%       ~ (p=0.247 n=10)
BackendBlockQueryRange/{status=error}_|_rate()/5                                                         35.54m ±  9%   33.54m ±  3%       ~ (p=0.063 n=10)
BackendBlockQueryRange/{}_|_quantile_over_time(duration,_.99,_.9,_.5)/5                                  357.0m ±  4%   364.6m ±  3%       ~ (p=0.190 n=10)
BackendBlockQueryRange/{}_|_quantile_over_time(duration,_.99)_by_(span.http.status_code)/5               596.0m ±  4%   600.7m ±  8%       ~ (p=0.631 n=10)
BackendBlockQueryRange/{}_|_histogram_over_time(duration)/5                                              361.7m ± 13%   365.7m ±  3%       ~ (p=0.912 n=10)
BackendBlockQueryRange/{}_|_avg_over_time(duration)_by_(span.http.status_code)/5                         528.3m ±  2%   519.9m ±  1%       ~ (p=0.105 n=10)
BackendBlockQueryRange/{}_|_max_over_time(duration)_by_(span.http.status_code)/5                         524.4m ±  5%   519.4m ±  3%       ~ (p=0.247 n=10)
BackendBlockQueryRange/{}_|_min_over_time(duration)_by_(span.http.status_code)/5                         528.1m ±  3%   529.5m ±  3%       ~ (p=0.912 n=10)
BackendBlockQueryRange/{_name_!=_nil_}_|_compare({status=error})/5                                        9.137 ±  2%    8.954 ±  3%       ~ (p=0.075 n=10)
geomean                                                                                                  279.8m         280.5m        +0.25%

                                                                                                       │ before.txt  │            after1000.txt             │
                                                                                                       │  MB_IO/op   │  MB_IO/op    vs base                 │
BackendBlockQueryRange/{}_|_rate()/5                                                                      3.733 ± 0%    3.733 ± 0%       ~ (p=1.000 n=10)
BackendBlockQueryRange/{}_|_rate()_by_(span.http.status_code)/5                                           3.984 ± 0%    3.984 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_rate()_by_(resource.service.name)/5                                           3.763 ± 0%    3.763 ± 0%       ~ (p=1.000 n=10)
BackendBlockQueryRange/{}_|_rate()_by_(span.http.url)/5                                                   3.904 ± 0%    3.904 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{resource.service.name=`loki-ingester`}_|_rate()/5                                289.5m ± 0%   289.5m ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{span.http.host_!=_``_&&_span.http.flavor=`2`}_|_rate()_by_(span.http.flavor)/5    15.71 ± 0%    15.71 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{status=error}_|_rate()/5                                                          3.905 ± 0%    3.904 ± 0%       ~ (p=0.070 n=10)
BackendBlockQueryRange/{}_|_quantile_over_time(duration,_.99,_.9,_.5)/5                                   6.663 ± 0%    6.663 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_quantile_over_time(duration,_.99)_by_(span.http.status_code)/5                6.913 ± 0%    6.913 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_histogram_over_time(duration)/5                                               6.663 ± 0%    6.663 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_avg_over_time(duration)_by_(span.http.status_code)/5                          6.913 ± 0%    6.913 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_max_over_time(duration)_by_(span.http.status_code)/5                          6.913 ± 0%    6.913 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_min_over_time(duration)_by_(span.http.status_code)/5                          6.913 ± 0%    6.913 ± 0%       ~ (p=1.000 n=10)
BackendBlockQueryRange/{_name_!=_nil_}_|_compare({status=error})/5                                        43.85 ± 0%    43.85 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                                                                   5.385         5.385       -0.00%
¹ all samples are equal

                                                                                                       │  before.txt   │            after1000.txt             │
                                                                                                       │   spans/op    │  spans/op    vs base                 │
BackendBlockQueryRange/{}_|_rate()/5                                                                     623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_rate()_by_(span.http.status_code)/5                                          623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_rate()_by_(resource.service.name)/5                                          623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_rate()_by_(span.http.url)/5                                                  623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{resource.service.name=`loki-ingester`}_|_rate()/5                                 0.000 ± 0%      0.000 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{span.http.host_!=_``_&&_span.http.flavor=`2`}_|_rate()_by_(span.http.flavor)/5    0.000 ± 0%      0.000 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{status=error}_|_rate()/5                                                         1.754k ± 0%     1.754k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_quantile_over_time(duration,_.99,_.9,_.5)/5                                  623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_quantile_over_time(duration,_.99)_by_(span.http.status_code)/5               623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_histogram_over_time(duration)/5                                              623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_avg_over_time(duration)_by_(span.http.status_code)/5                         623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_max_over_time(duration)_by_(span.http.status_code)/5                         623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_min_over_time(duration)_by_(span.http.status_code)/5                         623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{_name_!=_nil_}_|_compare({status=error})/5                                       623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                                                                              ²                +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean

                                                                                                       │   before.txt   │             after1000.txt             │
                                                                                                       │    spans/s     │   spans/s     vs base                 │
BackendBlockQueryRange/{}_|_rate()/5                                                                     3.093M ± 10%     3.027M ± 14%       ~ (p=0.971 n=10)
BackendBlockQueryRange/{}_|_rate()_by_(span.http.status_code)/5                                          1.404M ±  2%     1.397M ±  3%       ~ (p=0.739 n=10)
BackendBlockQueryRange/{}_|_rate()_by_(resource.service.name)/5                                          2.475M ±  1%     2.526M ±  6%       ~ (p=0.123 n=10)
BackendBlockQueryRange/{}_|_rate()_by_(span.http.url)/5                                                  1.438M ±  2%     1.414M ±  8%       ~ (p=0.529 n=10)
BackendBlockQueryRange/{resource.service.name=`loki-ingester`}_|_rate()/5                                 0.000 ±  0%      0.000 ±  0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{span.http.host_!=_``_&&_span.http.flavor=`2`}_|_rate()_by_(span.http.flavor)/5    0.000 ±  0%      0.000 ±  0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{status=error}_|_rate()/5                                                         49.38k ±  8%     52.30k ±  3%       ~ (p=0.063 n=10)
BackendBlockQueryRange/{}_|_quantile_over_time(duration,_.99,_.9,_.5)/5                                  1.746M ±  4%     1.710M ±  3%       ~ (p=0.190 n=10)
BackendBlockQueryRange/{}_|_quantile_over_time(duration,_.99)_by_(span.http.status_code)/5               1.046M ±  4%     1.037M ±  7%       ~ (p=0.631 n=10)
BackendBlockQueryRange/{}_|_histogram_over_time(duration)/5                                              1.723M ± 12%     1.704M ±  3%       ~ (p=0.912 n=10)
BackendBlockQueryRange/{}_|_avg_over_time(duration)_by_(span.http.status_code)/5                         1.180M ±  2%     1.199M ±  1%       ~ (p=0.105 n=10)
BackendBlockQueryRange/{}_|_max_over_time(duration)_by_(span.http.status_code)/5                         1.189M ±  4%     1.200M ±  3%       ~ (p=0.247 n=10)
BackendBlockQueryRange/{}_|_min_over_time(duration)_by_(span.http.status_code)/5                         1.180M ±  3%     1.177M ±  3%       ~ (p=0.912 n=10)
BackendBlockQueryRange/{_name_!=_nil_}_|_compare({status=error})/5                                       68.21k ±  2%     69.61k ±  3%       ~ (p=0.075 n=10)
geomean                                                                                                               ²                 +0.26%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean
```

</details>


ie-pham:
the way this is implemented, tempo will truncate the final results at the frontend level. We can implement it in a way that will return as soon as 1000 series is reached regardless of how many data points are in each series to exit early. Not sure which we prefer.

modules/frontend/combiner/metrics_query_range.go
```diff
@@ -43,6 +44,11 @@ func NewQueryRange(req *tempopb.QueryRangeRequest, trackDiffs bool) (Combiner, e
 			if resp == nil {
 				resp = &tempopb.QueryRangeResponse{}
 			}
+			if setMaxSeries && len(resp.Series) > maxSeries {
+				resp.Series = resp.Series[:maxSeries]
```
electron0zero, line 48: IIUC, we are still collecting all the series and the dropping the extra data before we return them, and not exiting early here? right?

it would be great if exited early when we hit this limit.

It would be very useful in the cases where is q metrics query is returning high cardinality results, for example: `{} | rate by (span:id)`

Just the work of pulling the series from the blocks will be resource intensive, and can OOM all generators.

pkg/api/http.go
```diff
@@ -86,9 +86,11 @@ const (
 	// PathOverrides user configurable overrides
 	PathOverrides = "/api/overrides"
 
-	PathSearchTagValuesV2 = "/api/v2/search/tag/{" + MuxVarTagName + "}/values"
-	PathSearchTagsV2      = "/api/v2/search/tags"
-	PathTracesV2          = "/api/v2/traces/{traceID}"
+	PathSearchTagValuesV2     = "/api/v2/search/tag/{" + MuxVarTagName + "}/values"
+	PathSearchTagsV2          = "/api/v2/search/tags"
+	PathTracesV2              = "/api/v2/traces/{traceID}"
+	PathMetricsQueryInstantV2 = "/api/v2/metrics/query"
```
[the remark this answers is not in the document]
joe-elliott: why are we adding a v2 metrics api?

pkg/tempopb/tempo.proto
```diff
@@ -58,11 +58,7 @@ message TraceByIDRequest {
 message TraceByIDResponse {
   Trace trace = 1;
   TraceByIDMetrics metrics = 2;
-  enum Status {
-    COMPLETE = 0;
-    PARTIAL = 1;
-  }
-  Status status = 3;
+  PartialStatus status = 3;
```
joe-elliott, line 61: nice work reusing this

modules/frontend/combiner/metrics_query_range.go
ie-pham: i raised the question here https://github.com/grafana/tempo/pull/4525#issuecomment-2605057481
the problem is the results come in indeterministically - so if we exit as soon as we hit the max series, we could have a response where each series just has one data point which isn't very uesful. but yes it would save us for out performing.

pkg/tempopb/tempo.proto
ie-pham: we don't have to do a V2 - it's just extra precautions for when we roll out and Grafana is not yet using the new protos.

modules/frontend/combiner/metrics_query_range.go
electron0zero: cool, I am okay with current impl, and we can leve a note or todo that this can improved by making the results deterministic and exiting early.

sidenote: I think Joe's ordered results work might help here.

pkg/tempopb/tempo.proto
joe-elliott: adding a field to a proto object does not break current readers. let's not add a `v2` of an endpoint unless we have to. can we update the PR to just change the current one?

knylander-grafana (commented):
Thank you for adding docs.

modules/frontend/combiner/metrics_query_range.go
```diff
@@ -55,15 +61,18 @@ func NewQueryRange(req *tempopb.QueryRangeRequest, trackDiffs bool) (Combiner, e
 			sortResponse(resp)
 			return resp, nil
 		},
+		quit: func(resp *tempopb.QueryRangeResponse) bool {
```
electron0zero, line 64: can we add a test for early exit in the combiner.

pkg/tempopb/tempo.proto
```diff
@@ -58,11 +58,7 @@ message TraceByIDRequest {
 message TraceByIDResponse {
   Trace trace = 1;
   TraceByIDMetrics metrics = 2;
-  enum Status {
```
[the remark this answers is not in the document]
ie-pham: tested locally with grafana and it did not break

integration/e2e/query_range_test.go
```diff
@@ -65,6 +71,69 @@ sendLoop:
...
+	require.NotNil(t, jaegerClient)
+
+	ticker := time.NewTicker(500 * time.Millisecond)
+	defer ticker.Stop()
+	timer := time.NewTimer(5 * time.Second)
+	defer timer.Stop()
+sendLoop:
+	for {
+		select {
+		case <-ticker.C:
```
joe-elliott, line 94: curious as to why you chose this loop structure. the goal seems to be loop 10 times and send data?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 * [CHANGE] **BREAKING CHANGE** Enforce max attribute size at event, link, and instrumentation scope. Make config per-tenant.
   Renamed max_span_attr_byte to max_attribute_bytes
   [#4633](https://github.com/grafana/tempo/pull/4633) (@ie-pham)
+* [CHANGE] Enforce max series in response for metrics queries [#4525](https://github.com/grafana/tempo/pull/4525) (@ie-pham)
```
joe-elliott, line 20: i'd mention the addition of the config param(s) to control behavior.

docs/sources/tempo/configuration/_index.md
```diff
@@ -838,6 +841,10 @@ querier:
         # the address of the query frontend to connect to, and process queries
         # Example: "frontend_address: query-frontend-discovery.default.svc.cluster.local:9095"
         [frontend_address: <string>]
+
+    metrics:
+        # Maximum number of time series returned for a metrics query.
+        [max_response_series: <int> | default = 1000]
```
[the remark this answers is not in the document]
joe-elliott: this does make me wonder if we should have a shared section of the config for querying like we do for storage. that feels like overkill for one setting tho.

modules/frontend/combiner/metrics_query_range.go
```diff
@@ -45,6 +46,13 @@ func NewQueryRange(req *tempopb.QueryRangeRequest) (Combiner, error) {
 			if resp == nil {
 				resp = &tempopb.QueryRangeResponse{}
 			}
+			if maxSeries > 0 && len(resp.Series) >= maxSeries {
```
joe-elliott, line 49: i believe we need a similar check in the diff function?

modules/frontend/metrics_query_handler.go
```diff
@@ -51,14 +51,17 @@ func newQueryInstantStreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTri
...
+		c, err := combiner.NewTypedQueryRange(qr, cfg.Metrics.Sharder.MaxResponseSeries)
 		if err != nil {
 			return err
 		}
 
 		collector := pipeline.NewGRPCCollector(next, cfg.ResponseConsumers, c, func(qrr *tempopb.QueryRangeResponse) error {
 			// Translate each diff into the instant version and send it
 			resp := translateQueryRangeToInstant(*qrr)
+			// series already limited by the query range combiner just need to copy the status and message
+			resp.Status = qrr.Status
```
joe-elliott, line 63: why don't we do this in translateQueryRangeToInstant? seems like we need similar logic in the http handler

modules/querier/querier_query_range.go
```diff
@@ -32,13 +32,16 @@ func (q *Querier) queryRangeRecent(ctx context.Context, req *tempopb.QueryRangeR
...
 
-	c, err := traceql.QueryRangeCombinerFor(req, traceql.AggregateModeSum)
+	c, err := traceql.QueryRangeCombinerFor(req, traceql.AggregateModeSum, q.cfg.Metrics.MaxResponseSeries)
 	if err != nil {
 		return nil, err
 	}
 
 	mtx := sync.Mutex{} // combiner doesn't lock, so take lock before calling Combine to make is safe
 	forEach := func(ctx context.Context, client tempopb.MetricsGeneratorClient) error {
+		if c.MaxSeriesReached() {
```
joe-elliott, line 42: should we not enforce this in the generators to prevent `{} | rate() by (span:id)` or whatever from overwhelming them?

have you tested such a query on this branch?

pkg/traceql/combine.go
```diff
@@ -149,29 +149,42 @@ type QueryRangeCombiner struct {
...
 	eval    *MetricsFrontendEvaluator
 	metrics *tempopb.SearchMetrics
+
+	maxSeries        int
+	maxSeriesReached bool
+
+	// used to track which series were updated since the previous diff
+	// todo: it may not be worth it to track the diffs per series. it would be simpler (and possibly nearly as effective) to just calculate a global
+	//  max/min for all series
+	seriesUpdated map[string]tsRange
```
joe-elliott, line 159: i think this was from a previous feature that's been removed

modules/querier/querier_query_range.go
joe-elliott: also, don't we need similar code on the backend path? we also need to be thoughtful about the situation where the output series are different than the intermediate series.

for instance a `quantile_over_time()` calculation will pass up intermediate histograms that are then turned into quantiles in the frontend. so the queriers/generators may actually be handling more series than the output result.  i'm wondering if we want a 2 tier limit where the queriers/generators that are doing the intermediate work have a higher limit than the frontend? can you push this discussion internally?

pkg/traceql/combine.go
```diff
@@ -149,29 +149,42 @@ type QueryRangeCombiner struct {
...
 func (q *QueryRangeCombiner) Combine(resp *tempopb.QueryRangeResponse) {
-	if resp == nil {
+	if resp == nil || q.maxSeriesReached {
 		return
 	}
 
 	// Here is where the job results are reentered into the pipeline
 	q.eval.ObserveSeries(resp.Series)
 
+	if q.maxSeries > 0 && len(q.eval.Results()) >= q.maxSeries {
```
joe-elliott, line 184: i'm guessing "Results()" can get quite expensive. it might be easier to count and limit input series? this would be similar in concept to limiting input streams in a loki or prometheus query. can you do some research here to determine the cost?

docs/sources/tempo/configuration/_index.md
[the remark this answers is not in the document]
ie-pham: switching to passing the config in the request proto since it needs to get passed all the way to the metrics evaluator

integration/e2e/config-query-range-max-series.yaml
```diff
@@ -0,0 +1,81 @@
...
+server:
+  http_listen_port: 3200
+  log_level: debug
+
+query_frontend:
+  search:
+    query_backend_after: 0 # setting these both to 0 will force all range searches to hit the backend
+    query_ingesters_until: 0
+  metrics:
+    max_response_series: 3
```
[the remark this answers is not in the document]
ie-pham: now it is set to pass in the request proto

modules/querier/querier_query_range.go
ie-pham: With the new approach, we are setting the same limit at the query-frontend level all the way down to the metrics evaluator level - this means at the default limit of 1000 - each intermediate step - each block would have 1000 series limit, each metrics-generator would have 1000 series limit before the responses converge at the query-frontend

pkg/traceql/combine.go
ie-pham: switching to return series count in `ObserveSeries`

javiermolinar (commented):
How a metric query looks like now? Are the series evenly distributed?

docs/sources/tempo/configuration/_index.md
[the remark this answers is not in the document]
joe-elliott: should this be removed from the docs then?

modules/frontend/metrics_query_range_sharder.go
```diff
@@ -115,6 +116,7 @@ func (s queryRangeSharder) RoundTrip(pipelineRequest pipeline.Request) (pipeline
 		}
 	}
 	req.Exemplars = maxExemplars
+	req.MaxSeries = uint32(s.cfg.MaxResponseSeries)
```
joe-elliott, line 119: i find the code that is setting the value on the request confusing. we are doing it in the handler, again here, and in the backend requests below. i'm finding this confusing.

modules/generator/instance.go
```diff
@@ -540,16 +541,24 @@ func (i *instance) QueryRange(ctx context.Context, req *tempopb.QueryRangeReques
 	for _, p := range processors {
 		err = p.QueryRange(ctx, req, rawEval, jobEval)
 		if err != nil {
+			fmt.Printf("error in query range: %v\n", err)
```
joe-elliott, line 544: remove? 

bunch of similar test lines below

modules/generator/processor/localblocks/query_range.go
```diff
@@ -62,13 +63,40 @@ func (p *Processor) QueryRange(ctx context.Context, req *tempopb.QueryRangeReque
...
 
+	maxSeries := int(req.MaxSeries)
+	seriesCount := int32(0)
+	seriesCountCh := make(chan int)
+	seriesCtx, seriesCancel := context.WithCancel(ctx) // Use context for cancellation
+	defer seriesCancel()
+
+	// listens for series counts and updates the total series count.
+	go func() {
+		for count := range seriesCountCh {
```
joe-elliott, line 74: can we find a simpler way to do this? there's a lot of moving parts here to just count the current series.

i'm also concerned about the cost of repeatedly calling `.Results()` on the series. does this require a lot of work? how does it compare for different functions like rates/quantiles/avgs/etc?

also, the metrics aggregated at this level are different than those in the frontend. for instance `quantile_over_time` will produce considerably more series here than at the frontend b/c it is a histogram here that is aggregated into quantiles in the frontend. how does this work into our solution?

also also, if series are truncated here do we pass that back in the response?

pkg/traceql/combine.go
```diff
@@ -326,6 +326,10 @@ type QueryRangeCombiner struct {
 	req     *tempopb.QueryRangeRequest
 	eval    *MetricsFrontendEvaluator
 	metrics *tempopb.SearchMetrics
+
+	maxSeries        int
```
joe-elliott, line 330: you only need 2 of these 3, right?

why not remove maxSeriesReached and  change the function below to
```
func (q *QueryRangeCombiner) MaxSeriesReached() bool {
	return q.seriesCount >= q.maxSeries
}
```

pkg/traceql/engine_metrics_compare.go
```diff
@@ -167,6 +167,7 @@ func (m *MetricsCompare) observe(span Span) {
 		}
 		totals[i]++
 	})
+	return len(m.seriesAgg.Results())
```
joe-elliott, line 170: can we get tests on these all these series aggregators observe's returning total series?

modules/generator/processor/localblocks/query_range.go
ie-pham: Correct me if I'm wrong but I thought the series from "jobEval.ObserveSeries" are already aggregated series. So at worst, we would have the truncated amount of series. But if the environment has more than one metrics-generator, that is additional series that would get aggregated again. I think this would be suffice but what do you think? 

The .Results() is calling this https://github.com/grafana/tempo/blob/main/pkg/traceql/engine_metrics.go#L716 

Since these are the raw not-yet aggregated results, it's a lot harder to count them. Since (I think) we avoided calling them for performance and calling it only once at the end https://github.com/grafana/tempo/blob/main/modules/generator/instance.go#L548 do you think maybe we should just not keep track of the series on the head and wal blocks?

modules/frontend/metrics_query_range_handler.go
```diff
@@ -34,6 +34,7 @@ func newQueryRangeStreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTripp
 		if req.Step == 0 {
 			req.Step = traceql.DefaultQueryRangeStep(req.Start, req.End)
 		}
+		req.MaxSeries = uint32(cfg.Metrics.Sharder.MaxResponseSeries)
```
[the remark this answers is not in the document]
ie-pham: this makes sense

modules/generator/processor/localblocks/query_range.go
joe-elliott: > Correct me if I'm wrong but I thought the series from "jobEval.ObserveSeries" are already aggregated series. So at worst, we would have the truncated amount of series.

It's dependent on the metric type and aggregation level. So when you aggregate quantiles at the lower levels you are actually aggregating histograms and passing those up to the frontend. A histogram has considerably more series than a quantile. Up to N times more where N is the number of buckets in the histogram

When you aggregate a simple rates at all levels it is the same number of series.

> The .Results() is calling this https://github.com/grafana/tempo/blob/main/pkg/traceql/engine_metrics.go#L716

yeah, that looks potentially expensive. I did track down all the .Observe calls at the series level and the count they are returning seems a lot cheaper

modules/frontend/metrics_query_range_sharder.go
ie-pham: the backend request at the bottom is wrong, but this is necessary because in the handler, the value is only passed to the combiner https://github.com/grafana/tempo/blob/ba3fb50572d413af790a74a06d3f3cbd4adc6f76/modules/frontend/metrics_query_range_handler.go#L93 

but the http req was never modified before being passed to the sharder https://github.com/grafana/tempo/blob/ba3fb50572d413af790a74a06d3f3cbd4adc6f76/modules/frontend/metrics_query_range_handler.go#L118

joe-elliott: i'm good on this 👍 

i just realized i basically put this comment 2x.

ie-pham:
benchmark shows no noticeable difference with added call to Length()/length()
```
                                                                                                       │  before.txt  │            after1000.txt            │
BackendBlockQueryRange/{}_|_rate()/5                                                                     201.5m ± 11%   205.9m ± 16%       ~ (p=0.971 n=10)
BackendBlockQueryRange/{}_|_rate()_by_(span.http.status_code)/5                                          443.8m ±  2%   446.0m ±  3%       ~ (p=0.739 n=10)
BackendBlockQueryRange/{}_|_rate()_by_(resource.service.name)/5                                          251.8m ±  1%   246.8m ±  6%       ~ (p=0.123 n=10)
BackendBlockQueryRange/{}_|_rate()_by_(span.http.url)/5                                                  433.3m ±  2%   440.9m ±  8%       ~ (p=0.529 n=10)
BackendBlockQueryRange/{resource.service.name=`loki-ingester`}_|_rate()/5                                3.721m ±  3%   3.710m ±  4%       ~ (p=0.739 n=10)
BackendBlockQueryRange/{span.http.host_!=_``_&&_span.http.flavor=`2`}_|_rate()_by_(span.http.flavor)/5   136.1m ±  6%   146.8m ±  9%       ~ (p=0.247 n=10)
BackendBlockQueryRange/{status=error}_|_rate()/5                                                         35.54m ±  9%   33.54m ±  3%       ~ (p=0.063 n=10)
BackendBlockQueryRange/{}_|_quantile_over_time(duration,_.99,_.9,_.5)/5                                  357.0m ±  4%   364.6m ±  3%       ~ (p=0.190 n=10)
BackendBlockQueryRange/{}_|_quantile_over_time(duration,_.99)_by_(span.http.status_code)/5               596.0m ±  4%   600.7m ±  8%       ~ (p=0.631 n=10)
BackendBlockQueryRange/{}_|_histogram_over_time(duration)/5                                              361.7m ± 13%   365.7m ±  3%       ~ (p=0.912 n=10)
BackendBlockQueryRange/{}_|_avg_over_time(duration)_by_(span.http.status_code)/5                         528.3m ±  2%   519.9m ±  1%       ~ (p=0.105 n=10)
BackendBlockQueryRange/{}_|_max_over_time(duration)_by_(span.http.status_code)/5                         524.4m ±  5%   519.4m ±  3%       ~ (p=0.247 n=10)
BackendBlockQueryRange/{}_|_min_over_time(duration)_by_(span.http.status_code)/5                         528.1m ±  3%   529.5m ±  3%       ~ (p=0.912 n=10)
BackendBlockQueryRange/{_name_!=_nil_}_|_compare({status=error})/5                                        9.137 ±  2%    8.954 ±  3%       ~ (p=0.075 n=10)
geomean                                                                                                  279.8m         280.5m        +0.25%

                                                                                                       │ before.txt  │            after1000.txt             │
                                                                                                       │  MB_IO/op   │  MB_IO/op    vs base                 │
BackendBlockQueryRange/{}_|_rate()/5                                                                      3.733 ± 0%    3.733 ± 0%       ~ (p=1.000 n=10)
BackendBlockQueryRange/{}_|_rate()_by_(span.http.status_code)/5                                           3.984 ± 0%    3.984 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_rate()_by_(resource.service.name)/5                                           3.763 ± 0%    3.763 ± 0%       ~ (p=1.000 n=10)
BackendBlockQueryRange/{}_|_rate()_by_(span.http.url)/5                                                   3.904 ± 0%    3.904 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{resource.service.name=`loki-ingester`}_|_rate()/5                                289.5m ± 0%   289.5m ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{span.http.host_!=_``_&&_span.http.flavor=`2`}_|_rate()_by_(span.http.flavor)/5    15.71 ± 0%    15.71 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{status=error}_|_rate()/5                                                          3.905 ± 0%    3.904 ± 0%       ~ (p=0.070 n=10)
BackendBlockQueryRange/{}_|_quantile_over_time(duration,_.99,_.9,_.5)/5                                   6.663 ± 0%    6.663 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_quantile_over_time(duration,_.99)_by_(span.http.status_code)/5                6.913 ± 0%    6.913 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_histogram_over_time(duration)/5                                               6.663 ± 0%    6.663 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_avg_over_time(duration)_by_(span.http.status_code)/5                          6.913 ± 0%    6.913 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_max_over_time(duration)_by_(span.http.status_code)/5                          6.913 ± 0%    6.913 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_min_over_time(duration)_by_(span.http.status_code)/5                          6.913 ± 0%    6.913 ± 0%       ~ (p=1.000 n=10)
BackendBlockQueryRange/{_name_!=_nil_}_|_compare({status=error})/5                                        43.85 ± 0%    43.85 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                                                                   5.385         5.385       -0.00%
¹ all samples are equal

                                                                                                       │  before.txt   │            after1000.txt             │
                                                                                                       │   spans/op    │  spans/op    vs base                 │
BackendBlockQueryRange/{}_|_rate()/5                                                                     623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_rate()_by_(span.http.status_code)/5                                          623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_rate()_by_(resource.service.name)/5                                          623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_rate()_by_(span.http.url)/5                                                  623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{resource.service.name=`loki-ingester`}_|_rate()/5                                 0.000 ± 0%      0.000 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{span.http.host_!=_``_&&_span.http.flavor=`2`}_|_rate()_by_(span.http.flavor)/5    0.000 ± 0%      0.000 ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{status=error}_|_rate()/5                                                         1.754k ± 0%     1.754k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_quantile_over_time(duration,_.99,_.9,_.5)/5                                  623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_quantile_over_time(duration,_.99)_by_(span.http.status_code)/5               623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_histogram_over_time(duration)/5                                              623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_avg_over_time(duration)_by_(span.http.status_code)/5                         623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_max_over_time(duration)_by_(span.http.status_code)/5                         623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{}_|_min_over_time(duration)_by_(span.http.status_code)/5                         623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{_name_!=_nil_}_|_compare({status=error})/5                                       623.2k ± 0%     623.2k ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                                                                              ²                +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean

                                                                                                       │   before.txt   │             after1000.txt             │
                                                                                                       │    spans/s     │   spans/s     vs base                 │
BackendBlockQueryRange/{}_|_rate()/5                                                                     3.093M ± 10%     3.027M ± 14%       ~ (p=0.971 n=10)
BackendBlockQueryRange/{}_|_rate()_by_(span.http.status_code)/5                                          1.404M ±  2%     1.397M ±  3%       ~ (p=0.739 n=10)
BackendBlockQueryRange/{}_|_rate()_by_(resource.service.name)/5                                          2.475M ±  1%     2.526M ±  6%       ~ (p=0.123 n=10)
BackendBlockQueryRange/{}_|_rate()_by_(span.http.url)/5                                                  1.438M ±  2%     1.414M ±  8%       ~ (p=0.529 n=10)
BackendBlockQueryRange/{resource.service.name=`loki-ingester`}_|_rate()/5                                 0.000 ±  0%      0.000 ±  0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{span.http.host_!=_``_&&_span.http.flavor=`2`}_|_rate()_by_(span.http.flavor)/5    0.000 ±  0%      0.000 ±  0%       ~ (p=1.000 n=10) ¹
BackendBlockQueryRange/{status=error}_|_rate()/5                                                         49.38k ±  8%     52.30k ±  3%       ~ (p=0.063 n=10)
BackendBlockQueryRange/{}_|_quantile_over_time(duration,_.99,_.9,_.5)/5                                  1.746M ±  4%     1.710M ±  3%       ~ (p=0.190 n=10)
BackendBlockQueryRange/{}_|_quantile_over_time(duration,_.99)_by_(span.http.status_code)/5               1.046M ±  4%     1.037M ±  7%       ~ (p=0.631 n=10)
BackendBlockQueryRange/{}_|_histogram_over_time(duration)/5                                              1.723M ± 12%     1.704M ±  3%       ~ (p=0.912 n=10)
BackendBlockQueryRange/{}_|_avg_over_time(duration)_by_(span.http.status_code)/5                         1.180M ±  2%     1.199M ±  1%       ~ (p=0.105 n=10)
BackendBlockQueryRange/{}_|_max_over_time(duration)_by_(span.http.status_code)/5                         1.189M ±  4%     1.200M ±  3%       ~ (p=0.247 n=10)
BackendBlockQueryRange/{}_|_min_over_time(duration)_by_(span.http.status_code)/5                         1.180M ±  3%     1.177M ±  3%       ~ (p=0.912 n=10)
BackendBlockQueryRange/{_name_!=_nil_}_|_compare({status=error})/5                                       68.21k ±  2%     69.61k ±  3%       ~ (p=0.075 n=10)
geomean                                                                                                               ²                 +0.26%                ²
```

pkg/traceql/engine_metrics.go
```diff
@@ -313,15 +313,15 @@ type VectorAggregator interface {
 // RangeAggregator sorts spans into time slots
 // TODO - for efficiency we probably combine this with VectorAggregator (see todo about CountOverTimeAggregator)
 type RangeAggregator interface {
-	Observe(s Span)
+	Observe(s Span) int
```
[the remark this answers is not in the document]
ie-pham: I have replaced this approach with adding a Length() function instead

modules/generator/processor/localblocks/query_range.go
ie-pham: Refactored to not call .Results() anymore.

integration/e2e/api/query_range_test.go
```diff
@@ -103,6 +110,70 @@ sendLoop:
...
+		case <-timer.C:
+			break sendLoop
+		}
+	}
+
+	// Wait for traces to be flushed to blocks
+	require.NoError(t, tempo.WaitSumMetricsWithOptions(e2e.GreaterOrEqual(1), []string{"tempo_metrics_generator_processor_local_blocks_spans_total"}, e2e.WaitMissingMetrics))
+	require.NoError(t, tempo.WaitSumMetricsWithOptions(e2e.GreaterOrEqual(1), []string{"tempo_metrics_generator_processor_local_blocks_cut_blocks"}, e2e.WaitMissingMetrics))
+
+	query := "{} | rate() by (span:id)"
```
mdisibio, line 144: I love that this is our test case 💯

modules/frontend/metrics_query_handler.go
```diff
@@ -43,6 +43,11 @@ func newQueryInstantStreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTri
 			End:   req.End,
 			Step:  req.End - req.Start,
 		}
+		// if a limit is being enforced, honor the request if it is less than the limit
+		// else set it to max limit
+		if cfg.Metrics.Sharder.MaxResponseSeries > 0 && (qr.MaxSeries > uint32(cfg.Metrics.Sharder.MaxResponseSeries) || qr.MaxSeries == 0) {
```
mdisibio, line 48: Do the handlers need to do this too, when the sharder is already doing it?  It looks like doing in the 1 place in the sharder should work.

modules/generator/instance.go
```diff
@@ -553,13 +553,21 @@ func (i *instance) QueryRange(ctx context.Context, req *tempopb.QueryRangeReques
...
 	jobEval.ObserveSeries(walResults)
 
 	r := jobEval.Results()
 	rr := r.ToProto(req)
+
+	maxSeries := int(req.MaxSeries)
+	if maxSeries > 0 && len(rr) > maxSeries {
+		return &tempopb.QueryRangeResponse{
+			Series: rr[:maxSeries],
+			Status: tempopb.PartialStatus_PARTIAL,
```
mdisibio, line 567: Does this need to set message also?

pkg/traceql/combine.go
```diff
@@ -361,8 +370,16 @@ func (q *QueryRangeCombiner) Combine(resp *tempopb.QueryRangeResponse) {
 }
 
 func (q *QueryRangeCombiner) Response() *tempopb.QueryRangeResponse {
-	return &tempopb.QueryRangeResponse{
+	response := &tempopb.QueryRangeResponse{
 		Series:  q.eval.Results().ToProto(q.req),
 		Metrics: q.metrics,
 	}
+	if q.maxSeriesReached {
+		response.Status = tempopb.PartialStatus_PARTIAL
```
mdisibio, line 378: Does this need to set message?

modules/generator/processor/localblocks/query_range.go
```diff
@@ -101,17 +111,20 @@ func (p *Processor) QueryRange(ctx context.Context, req *tempopb.QueryRangeReque
...
+			go func(b *ingester.LocalBlock) {
+				defer wg.Done()
+				resp, err := p.queryRangeCompleteBlock(ctx, b, *req, timeOverlapCutoff, unsafe, int(req.Exemplars))
+				if err != nil {
+					jobErr.Store(err)
+					return
+				}
+				jobEval.ObserveSeries(resp)
+				seriesCount := jobEval.Length()
+				syncAtomic.AddInt32(&totalSeriesCount, int32(seriesCount))
```
mdisibio, line 125: Can you clarify this part a bit - The atomic is used to skip remaining blocks when we exceed the limit, but should it be a `Set` instead of `Add`?  It looks like if we have 100 complete blocks all returning the same 10 series, it can still hit the (default) 1000 limit.

modules/generator/instance.go
ie-pham: we can add it for consistency but this message is only relevant when returning to the user. The query-frontend combiner will check for the partial status and add it only at the very end. or do you think we should add it anyway?

modules/generator/processor/localblocks/query_range.go
ie-pham: Ah nice catch. The atomic should just be used to keep track of the raw eval for wal and head blocks. The jobEval.length() is already tracking the series for the completed blocks. The check should now be 

raw series count + jobEval.length() > max series

modules/frontend/metrics_query_handler.go
mdisibio: Ah ok.  Let's still consolidate it to 1 function?

modules/generator/instance.go
mdisibio: Ok, I misread and thought the message was already set in multiple places.  In that case a comment isn't really necessary either.

pkg/traceql/engine_metrics_test.go
```diff
@@ -512,9 +512,10 @@ func TestQuantileOverTime(t *testing.T) {
 		},
 	}
 
-	result, err := runTraceQLMetric(req, in)
+	result, seriesCount, err := runTraceQLMetric(req, in)
 	require.NoError(t, err)
 	require.Equal(t, out, result)
+	assert.Equal(t, len(result), seriesCount)
```
mdisibio, line 518: Use `require.Equal`?   Actually this might be worth talking about more broadly, but is there ever a reason to use `assert`?  I haven't found a case where continuing a test is helpful after failing and I always prefer `require`.

modules/frontend/combiner/metrics_query_range.go
```diff
@@ -16,11 +17,13 @@ var _ GRPCCombiner[*tempopb.QueryRangeResponse] = (*genericCombiner[*tempopb.Que
 // NewQueryRange returns a query range combiner.
 func NewQueryRange(req *tempopb.QueryRangeRequest) (Combiner, error) {
 	combiner, err := traceql.QueryRangeCombinerFor(req, traceql.AggregateModeFinal)
+	maxSeries := int(req.MaxSeries)
```
joe-elliott, line 20: are we handling Status = Partial being returned from the queriers here? Shouldn't we pass that back to the caller even if `combiner.MaxSeriesReached() == false` here?

modules/frontend/metrics_query_handler.go
joe-elliott: i think it only needs to be done here? the sharder can assume that the value is set correctly?

although looking at search it looks like we repeat the logic 2x. once in the sharder and once in the handler:

https://github.com/grafana/tempo/blob/main/modules/frontend/search_sharder.go#L82
https://github.com/grafana/tempo/blob/main/modules/frontend/search_handlers.go#L128

it seems like we should establish a better pattern and only do this once, but agree with marty that consolidating the logic in one func is the right choice

modules/generator/instance.go
joe-elliott: this code is weird. it assumes there can be multiple local blocks processors, queries each and combines the results. is that possible?

modules/generator/processor/localblocks/query_range.go
```diff
@@ -62,15 +63,21 @@ func (p *Processor) QueryRange(ctx context.Context, req *tempopb.QueryRangeReque
 		jobErr = atomic.Error{}
 	)
 
+	maxSeries := int(req.MaxSeries)
+	totalRawResultsSeriesCount := int32(0)
```
joe-elliott, line 67: use the atomic int32 struct instead. it forces concurrent access to the var and is harder to make a mistake with. you can also use uber's atomic package which is more common in the codebase, but that matters less. 

perhaps one day we will standardize.

modules/generator/processor/localblocks/query_range.go
```diff
@@ -101,17 +111,18 @@ func (p *Processor) QueryRange(ctx context.Context, req *tempopb.QueryRangeReque
...
-			defer wg.Done()
-			resp, err := p.queryRangeCompleteBlock(ctx, b, *req, timeOverlapCutoff, unsafe, int(req.Exemplars))
-			if err != nil {
-				jobErr.Store(err)
-				return
-			}
-
-			jobEval.ObserveSeries(resp)
-		}(b)
+		if maxSeries == 0 || (int(syncAtomic.LoadInt32(&totalRawResultsSeriesCount))+jobEval.Length() < maxSeries) {
```
joe-elliott, line 114: i don't think this check is correct. it assumes that all series in "jobEval" are distinct from all series in "eval".

modules/generator/processor/localblocks/query_range.go
```diff
@@ -62,15 +63,21 @@ func (p *Processor) QueryRange(ctx context.Context, req *tempopb.QueryRangeReque
...
 	if p.headBlock != nil && withinRange(p.headBlock.BlockMeta()) {
-		wg.Add(1)
-		go func(w common.WALBlock) {
-			defer wg.Done()
-			err := p.queryRangeWALBlock(ctx, w, rawEval)
-			if err != nil {
-				jobErr.Store(err)
-			}
-		}(p.headBlock)
+		if maxSeries == 0 || (syncAtomic.LoadInt32(&totalRawResultsSeriesCount) < int32(maxSeries)) {
```
joe-elliott, line 70: all of this code external to the evaluator to track max series feels strange to me. should we push this down into the evaluator and just ask it if max series has been exceeded? 

wdyt @mdisibio ?

modules/querier/querier_query_range.go
```diff
@@ -121,13 +124,19 @@ func (q *Querier) queryBlock(ctx context.Context, req *tempopb.QueryRangeRequest
...
 		Series: queryRangeTraceQLToProto(res, req),
 		Metrics: &tempopb.SearchMetrics{
 			InspectedBytes: inspectedBytes,
 			InspectedSpans: spansTotal,
 		},
-	}, nil
+	}
+
+	if len(res) > int(req.MaxSeries) {
+		response.Status = tempopb.PartialStatus_PARTIAL
```
joe-elliott, line 136: truncate the series instead of marshalling them all and passing them back? or really truncate it before we `queryRangeTraceQLToProto`?

similar to above the evaluator could just truncate it internally and let us know that it had truncated results? it looks like we're already passing max series in when we call Do()? 

maybe we require that parameter at creation time and Do returns if the limit was exceeded?

cc @mdisibio

pkg/traceql/combine.go
```diff
@@ -335,19 +338,25 @@ func QueryRangeCombinerFor(req *tempopb.QueryRangeRequest, mode AggregateMode) (
 	}
 
 	return &QueryRangeCombiner{
-		req:     req,
-		eval:    eval,
-		metrics: &tempopb.SearchMetrics{},
+		req:       req,
+		eval:      eval,
```
joe-elliott, line 342: Combiner is also using an evaluator. Is it possible to push the limiting logic to only exist in the evaluator and have calling code just ask it to see if it limited or not?

cc @mdisibio

modules/frontend/combiner/metrics_query_range.go
ie-pham: the combiner will set MaxSeriesReached() to true if Status = Partial is returned from the querier

modules/generator/instance.go
mdisibio: Yes it is part of the new rhythm architecture. There are 2 copies of local blocks, 1 is receiving grpc traffic and flushing to object storage (i.e. previous arch), the other is receiving queue traffic and not flushing (i.e. new arch). This is necessary to handle initial cutover and allows revert at any time. We can chat about this offline too.

joe-elliott (approved):
LGTM.

Some minor comments on truncating before doing marshalling work, but i think this is good to go.

modules/generator/processor/localblocks/query_range.go
ie-pham: updated to use atomic bool for maxSeriesReached which will be set to true if either rawEval.length() > maxSeries or jobEval.length() > maxSeries

mdisibio (approved):

PR #4539, opened 2025-01-13
Fix TraceQL results caching bug for floats ending in .0

### LINKED ISSUE #4528: TraceQL results caching bug for floats ending in `.0`
**Describe the bug**
TraceQL string roundtrip isn't accurate for queries with floating point values ending in `.0`, and unintentionally converts them to an integer.    

For example: `{ span.foo != 3.0}` is parsed ok, but then converting back to canonical string gives `{ span.foo != 3 }`, which is an integer and a completely different query.

This is a problem for results caching in the frontend, because it uses the canonical query form. This means that it can show the wrong results if you only change the numeric from integer to float, and vice versa.

**To Reproduce**
Add the above query to `pkg/traceql/test_examples.yaml` and verify that `TestStringer` now fails.  The StaticType is wrong:

```
$ go test -run=TestStringer 
--- FAIL: TestStringer (0.01s)
    --- FAIL: TestStringer/{_span.a_!=_3.0_} (0.00s)
        ast_stringer_test.go:45: 
            	Error Trace:	/Users/marty/src/tempo/pkg/traceql/ast_stringer_test.go:45
            	Error:      	Not equal: 
            	            	expected: &traceql.RootExpr{Pipeline:traceql.Pipeline{Elements:[]traceql.pipelineElement{(*traceql.SpansetFilter)(0x140003aff50)}}, MetricsPipeline:traceql.metricsFirstStageElement(nil), Hints:(*traceql.Hints)(nil)}
            	            	actual  : &traceql.RootExpr{Pipeline:traceql.Pipeline{Elements:[]traceql.pipelineElement{(*traceql.SpansetFilter)(0x14000470090)}}, MetricsPipeline:traceql.metricsFirstStageElement(nil), Hints:(*traceql.Hints)(nil)}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -13,4 +13,4 @@
            	            	      RHS: (traceql.Static) {
            	            	-      Type: (traceql.StaticType) 4,
            	            	-      valScalar: (uint64) 4613937818241073152,
            	            	+      Type: (traceql.StaticType) 3,
            	            	+      valScalar: (uint64) 3,
            	            	       valBytes: ([]uint8) <nil>,
            	Test:       	TestStringer/{_span.a_!=_3.0_}
```

**Expected behavior**
Floating point type should be preserved.   We need to update this logic to preserve the `.0` I think?  The `-1` means use as few digits as possible, which is 99% correct except this one case.  Technically just the decimal point `.` is required, i.e. `3.` is the same as `3.0`.

```
	case TypeFloat:
		return strconv.FormatFloat(s.Float(), 'g', -1, 64)
```

Don't think we want to change the frontend, using canonical form is the right choice.

**Environment:**

**Additional Context**

carles-grafana:
**What this PR does**:

Fixes #4528 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/traceql/engine_metrics_test.go
```diff
@@ -352,7 +352,7 @@ func TestQuantileOverTime(t *testing.T) {
 	// Output series with quantiles per foo
 	// Prom labels are sorted alphabetically, traceql labels maintain original order.
 	out := SeriesSet{
-		`{p="0", span.foo="bar"}`: TimeSeries{
+		`{p="0.0", span.foo="bar"}`: TimeSeries{
```
carles-grafana, line 355: not sure if this is wanted @mdisibio

mdisibio: Ah interesting side effect.  It's ok. The actual label value is the same `NewStaticFloat(0)`

pkg/traceql/ast_test.go
```diff
@@ -131,9 +131,9 @@ func TestStatic_String(t *testing.T) {
 		{arg: nil, want: "nil"},
 		{arg: 101, want: "101"},
 		{arg: -10, want: "-10"},
-		{arg: -1.0, want: "-1"},
-		{arg: 0.0, want: "0"},
-		{arg: 10.0, want: "10"},
+		{arg: -1.0, want: "-1.0"},
+		{arg: 0.0, want: "0.0"},
+		{arg: 10.0, want: "10.0"},
```
mdisibio, line 136: Can you add a few more cases:  not ending in .0, and something with the `e` scientific notation?

mdisibio (commented):
Let's add a few more test cases to be 100% of all effects, and then lgtm.

pkg/traceql/ast_test.go
carles-grafana: done

mdisibio: Sorry my comment wasn't clear.   The cases I meant were a value like `12.34` to test the `.`, and something huge that outputs the `e` to test that case.   12.34 is halfway covered by the second case, but let's get clear cases where input == output.  The `0` case isn't needed cause that's testing the int path which is already covered.

carles-grafana: I've added the 2 test cases and another one I think wasn't covered

mdisibio: `12.34e+2` Nice catch 👍

PR #4568, opened 2025-01-16
feat: update minio to version 7.0.83

javiermolinar:
**What this PR does**:
It updates minio to version 7.0.83. 
This is a minor version but it has a breaking change, any struct implementing `Provider` interface must include the RetrieveWithCredContext method.[https://github.com/minio/minio-go/pull/2041]( https://github.com/minio/minio-go/pull/2041)

**Checklist**
- [x] Tests updated
- [x] Documentation added


joe-elliott (approved):
Given the complexity of s3 auth I think we should add a changelog for this. In case this does break something it would be nice for folks who scan the versions changelog to see a change here.

PR #4593, opened 2025-01-22
[rhythm] Implement group-consuming in the metrics-generator

mapno:
**What this PR does**:

This PR introduces group consuming to the generators, instead of the previously used direct consuming. Group consuming delegates partitions assignment to Kafka, which greatly simplifies operations for the generators.

It uses a custom balancer that combines Kafka's cooperative sticky balancing with partition ring awareness. This implementation is [forked from Loki](https://github.com/grafana/loki/blob/889404791d9125d2d17a7da963ae0fc2b268e2e7/pkg/kafka/partitionring/consumer/balancer.go).

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/generator_kafka.go
+++ b/modules/generator/generator_kafka.go
@@ -3,17 +3,11 @@ package generator
 import (
 	"context"
 	"errors"
-	"fmt"
-	"sync"
 	"time"
 
 	"github.com/go-kit/log/level"
 	"github.com/grafana/tempo/pkg/ingest"
 	"github.com/grafana/tempo/pkg/tempopb"
-
-	"github.com/twmb/franz-go/pkg/kadm"
-	"github.com/twmb/franz-go/pkg/kerr"
-	"github.com/twmb/franz-go/pkg/kgo"
 )
 
 func (g *Generator) startKafka() {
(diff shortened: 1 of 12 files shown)


modules/generator/generator_kafka.go
```diff
@@ -47,65 +44,13 @@ func (g *Generator) listenKafka(ctx context.Context) {
...
-}
-
-func (g *Generator) consumePartition(ctx context.Context, partition int32, lag kadm.GroupMemberLag) error {
 	d := ingest.NewDecoder()
 
-	// We always rewind the partition's offset to the commit offset by reassigning the partition to the client (this triggers partition assignment).
-	// This is so the cycle started exactly at the commit offset, and not at what was (potentially over-) consumed previously.
-	// In the end, we remove the partition from the client (refer to the defer below) to guarantee the client always consumes
-	// from one partition at a time. I.e. when this partition is consumed, we start consuming the next one.
-	g.kafkaClient.AddConsumePartitions(map[string]map[int32]kgo.Offset{
```
mdisibio, line 102: Can you help clarify for my understanding - since this client update is removed, how does PollFetches know which partitions are assigned?  Is that now handled by the server (broker, etc)?

mapno: Yes, when joining a group (ie. `kgo.ConsumerGroup(kafkaCfg.ConsumerGroup)`) Kafka handles partitions assignment. The client will talk to the broker which will share the partition balance for them. In each pull, the client will receive fetches for the partitions that have been assigned to it.

mdisibio (approved):

Issue #2188, opened 2023-03-09
[TraceQL] Support attribute existence filter


joe-elliott:
**Is your feature request related to a problem? Please describe.**
There should be a way to query for the existence of an attribute without asserting a value. Perhaps something like:

```
{ span.att = nil }
{ exists(span.att) }
{ span.att }
```

Note that the last one overloads querying a boolean attribute for true or false values. If we want to go that route we should think through it carefully.

lanapouney:
Hi, I am looking for a way to query that an attribute does not exist for a span, could this be achieved via { !exists(span.att) } or would you propose another method for this?

joe-elliott:
Unfortunately this does not currently exist. I want it too, but haven't found the time to add it. Do you have a preferred syntax above?

If you know the type of your attribute you can do something like `{ span.att != ""}` or  `{ span.att != 0 }'. I know that's not great, but it's the only current work around.

cyrille-leclerc:
I prefer `{ span.att = nil }` and `{ span.att = nil }` because they reuse the existing TraceQL [comparison operators](https://grafana.com/docs/tempo/latest/traceql/#comparison-operators) and work well IMO for auto-completion

cedricziel:
I'm putting my money on `{ .attr = nil }` as well because it's simple and intuitive

adirmatzkin:
Just making sure I don't miss out on something here...
By querying for a value of "nil" for some attribute `x` (syntax as `{ span.x = nil }`), we would assume `x` doesn't even exist?
How would we differentiate between the 2 cases:
1. x is not even a field in a span
2. x **is** a field in a span, but **its value is** None/Null/Nil.

If we're about to use "nil", "none", or "null" for different meanings here - let me raise a red flag, because it would be confusing and not intuitive for users...

Unless I'm missing something - I feel like the `{ exists(span.x) }` syntax would be better here due to the need to differentiate between the above 2 cases.

cyrille-leclerc:
Great verification.

From what I understand, null value is not allowed for hotel attributes and
it means to suppress the attribute.
I have to read in depth
https://github.com/open-telemetry/opentelemetry-specification/issues/797

cyrille-leclerc:
Confirmed, otel attributes MUST be non null.

See
https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/common

"Attribute values of null are not valid and attempting to set a null value
is undefined behavior."

adirmatzkin:
[...]

PR #4600, opened 2025-01-23
WIP: [TraceQL] Fix to put all conditions following a select clause into the second pass

mdisibio:
**What this PR does**:
Found some incorrect behavior from https://github.com/grafana/tempo/issues/2188#issuecomment-2608751948 and this fixes it.  The idea is that all clauses after a `select()` operation should be considered in the second pass and optional, but it was being put in the first pass and required.  Rare, but think it's worth fixing as this is a work-around for attribute == nil check.

Relates to #2188 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


ie-pham (approved):
cool beans

PR #4603, opened 2025-01-23
vParquet4 WAL mapping fix

mdisibio:
**What this PR does**:
vParquet4 WAL computes the dedicated column definitions for every trace that is written. The mapping doesn't change once the block is created, so this fixes it to be done once.

This is about a 10% perf improvement in the block builder.  Probably will be noticeable in ingesters also but didn't check.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


joe-elliott (approved):
also cool beans

mapno (approved):

PR #4611, opened 2025-01-24
rhythm: set ingestion slack for partition consumption

javiermolinar:
**What this PR does**:

It adjusts the ingestion slack based on the first record's timestamp, addressing an issue where the slack was previously calculated using the current time. This change prevents spans from falling outside the ingestion slack due to partition lag.

The calculation of the slack time range is done outside the headblock. Every partition writer is initialized with the timestamp of the first record in the partition. We can use that timestamp as a pivot to control the slack [rec.timestamp - ingestionSlack, rec.timestamp + ingestionSlack + cycleDuration]. 

**Checklist**
- [X] Tests updated


modules/blockbuilder/tenant_store.go
```diff
@@ -230,6 +232,27 @@ func (s *tenantStore) CutIdle(since time.Time, immediate bool) error {
...
+		end = uint32(startSectionTime.Unix())
+	}
+
+	if warn {
+		dataquality.WarnOutsideIngestionSlack(s.headBlock.BlockMeta().TenantID)
+	}
+
+	return start, end
+}
+
```
mapno, line 255: We have now 4 implementations of the same function. IMO this should be moved to a different pkg and reused.

tempodb/encoding/vparquet4/wal_block.go
```diff
@@ -331,10 +331,14 @@ func (b *walBlock) Append(id common.ID, buff []byte, start, end uint32) error {
 	if err != nil {
 		return fmt.Errorf("error preparing trace for read: %w", err)
 	}
-
+	start, end = b.adjustTimeRangeForSlack(start, end)
```
mapno, line 334: Having a different behavior in `Append` and `AppendTrace` is very unexpected. They're the same func but for the data in different formats—bytes or decoded in a struct.

IMO, that should be explicitly stated in the signature with a boolean or a callback, or by code comments very clearly.

mapno (commented):
Should the range be `[rec.timestamp - ingestionSlack, rec.timestamp + cycleDuration + ingestionSlack]` instead?

How it is right now, spans towards the end of the cycle will be outside the range if `ingestionSlack < cycleDuration`. We'll need to keep `ingestionSlack` high just to keep up with the cycle's duration.

javiermolinar:
> Should the range be `[rec.timestamp - ingestionSlack, rec.timestamp + cycleDuration + ingestionSlack]` instead?
> 
> How it is right now, spans towards the end of the cycle will be outside the range if `ingestionSlack < cycleDuration`. We'll need to keep `ingestionSlack` high just to keep up with the cycle's duration.

We can add the cycleDuration sure,  it won't hurt

tempodb/encoding/vparquet4/wal_block.go
javiermolinar: Yeah I though about that, I didn't want to change the signature due to the amount of changes required

javiermolinar: I think that the best way is explicitly dictate if the ingestionSlack must be recomputed or not. So I've added a new boolean for that

modules/blockbuilder/tenant_store.go
javiermolinar: I've moved it to a common place

modules/blockbuilder/partition_writer.go
```diff
@@ -34,17 +35,18 @@ type writer struct {
 	m   map[string]*tenantStore
 }
 
-func newPartitionSectionWriter(logger log.Logger, partition, cycleEndTs uint64, blockCfg BlockConfig, overrides Overrides, wal *wal.WAL, enc encoding.VersionedEncoding) *writer {
+func newPartitionSectionWriter(logger log.Logger, partition, cycleTs uint64, startSectionTime time.Time, blockCfg BlockConfig, overrides Overrides, wal *wal.WAL, enc encoding.VersionedEncoding) *writer {
```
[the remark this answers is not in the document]
javiermolinar: Done

mapno (approved):
LGTM. Thanks for addressing all my comments

PR #4624, opened 2025-01-27
[Query-frontend] Fix metrics streaming

joe-elliott:
**What this PR does**:
Metrics diff streaming relied on using the start/end timestamps of input series and labels to determine which output series were impacted. In hindsight is quite obvious that this will only work in the most trivial of examples. Consider a metrics quantile stream: The input series are all labeled like histograms, but the output series are labeled like quantiles.

The engine can technically calculate which output series are impacted by which input series, but this would be increasingly complicated as we added additional metrics function. It would also be quite brittle and require constant upkeep for every function. Instead we have settled on doing a simple diff in the frontend between the previous metrics and the current one.

**Notes for the reviewer:**

I am currently using [prom label string equality](https://github.com/grafana/tempo/pull/4624/files#diff-9c36f8a75f46932610d4c582e1450885ef34b3c2651da7ba92c8f4092bf01c94R116) to match previous and current series. I believe this works, but looking for confirmation from someone with more experience.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/frontend/combiner/metrics_query_range.go
```diff
@@ -88,6 +96,67 @@ func sortResponse(res *tempopb.QueryRangeResponse) {
...
+			diff.Series = append(diff.Series, s)
+			continue
+		}
+
+		// promlabels are the same, have to check individual samples
+		// just copy forward labels and exemplars
+		diffSeries := &tempopb.TimeSeries{
+			Labels:     s.Labels,
+			PromLabels: s.PromLabels,
+			Exemplars:  s.Exemplars,                                  // taking all current exemplars. improve?
```
mdisibio, line 126: It probably makes sense to diff the exemplars.  Is there a reason not to?  They should be stable, but even if they aren't, there would still be some savings.

mdisibio (approved):
Looking good to me.  Good approach for diffing the sorted slices.   Using prom labels is ok, internally it is the key in the [SeriesSet](https://github.com/grafana/tempo/blob/main/pkg/traceql/engine_metrics.go#L231) and used for other combine/compare operations like [here](https://github.com/grafana/tempo/blob/main/pkg/traceql/engine_metrics.go#L1198).  If there is an issue with it, would rather have the consistent usage that can all be fixed together than try something different here.

modules/frontend/combiner/metrics_query_range.go
joe-elliott: i wasn't sure about exemplars so i did the most straightforward thing. will diff them as well 👍

mdisibio (approved):

PR #4633, opened 2025-01-28
switch max attribute bytes to per tenant and also check event links a…

ie-pham:
**What this PR does**: Enforce max attribute byte size at the event, link, and instrumentation level. Make config per tenant a well. 

BREAKING CHANGE *** 
renamed `max_span_attr_byte` => `max_attribute_bytes`

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


modules/distributor/distributor.go
```diff
@@ -986,3 +1006,11 @@ func logSpan(s *v1.Span, allAttributes bool, logger log.Logger) {
 func startEndFromSpan(span *v1.Span) (uint32, uint32) {
 	return uint32(span.StartTimeUnixNano / uint64(time.Second)), uint32(span.EndTimeUnixNano / uint64(time.Second))
 }
+
+func getMaxAttributeBytes(d Distributor, userID string) int {
```
joe-elliott, line 1010: why not make this a function on the struct pointer?

modules/overrides/config_legacy.go
```diff
@@ -149,6 +151,7 @@ func (l *LegacyOverrides) toNewLimits() Overrides {
 			MaxLocalTracesPerUser:  l.MaxLocalTracesPerUser,
 			MaxGlobalTracesPerUser: l.MaxGlobalTracesPerUser,
 			TenantShardSize:        l.IngestionTenantShardSize,
+			MaxAttributeBytes:      l.IngestionMaxAttributeBytes,
```
joe-elliott, line 154: do we have tests that lock this copying behavior down?

ie-pham: i am adding a test to make sure the conversion between the legacy and new format is working. is that what you're looking for?

joe-elliott: yeah, also just kind of asking if we have these tests generally?

joe-elliott (approved):

electron0zero (approved):

knylander-grafana (commented):

PR #4642, opened 2025-01-30
Update block builder rollout strategy

mapno:
**What this PR does**:

Updates block-builder sts rollout strategy to ensure simultaneous updates. Otherwise it run the risk of ownership conflicts in partitions.

Example:
```
Say there are 6 partitions and 3 block-builders. Partition assignment is as follows:
- block-builder-0 has assigned partitions 0 and 3.
- block-builder-1 has assigned partitions 1 and 4.
- block-builder-2 has assigned partitions 2 and 5.

If a new block-builder is deployed, partition assignment will change to
- block-builder-0 has assigned partitions 0 and 4.
- block-builder-1 has assigned partitions 1 and 5.
- block-builder-2 has assigned partitions 2.
- block-builder-3 has assigned partitions 3.

As soon as block-builders are rolled out, they'll have the new config, while older replicas will have the old config—ie. at one point, block-builder-3 and block-builder-0 both will own partition 3
```

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


mapno:
Marking as draft while I investigate the viability of this change, given that it depends on an [feature gate in alpha](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#maximum-unavailable-pods)

PR #4646, opened 2025-01-31
Add support for topk and bottomk functions for TraceQL metrics

### LINKED ISSUE #4217: [TraceQL] Add topk/bottomk to TraceQL Metrics
TraceQL metrics brings the promise of querying infinite cardinality, but in a practical sense both Tempo and Grafana suffer once we start returning series in the 10s of thousands. It can also be quite difficult to parse this amount of information looking at a time series graph.

Add `topk`/`bottomk` style functionality to the language to give users control over cardinality and return the series they are most interested in. Syntax would look something like:

```
{} | rate() by (span.client_ip) | topk(10)
```

electron0zero:
**What this PR does**:

Add support for topk and bottomk functions on top of TraceQL Metrics.

you can now write TraceQL Queries like `{} | rate() by (span.client_ip) | topk(10)` or `{} | rate() by (span.client_ip) | bottomk(10)`, and only get top or bottomk series from the underlaying TraceQL Metrics Queries.

topk and bottomk behaves like Prometheus topk and bottomk functions. 

checkout this demo video to see it in action:

https://github.com/user-attachments/assets/279e1816-ce97-4e37-9232-919e9c4cf827

---

Topk and Bottomk are implemented as first set of second stage functions (functions that operate on series generated by TraceQL Metrics). code in this PR lays the ground for second stage functions and opens the door for other second stage functions like min, max and avg on top of TraceQL Metrics.

Fixes #4217 
Fixes https://github.com/grafana/tempo-squad/issues/601

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/traceql/ast_metrics.go
```diff
@@ -0,0 +1,360 @@
...
+
+// metricsSecondStageElement represents operations that can be performed
+// after the first stage metrics pipeline, such as topK/bottomK, etc.
+type metricsSecondStageElement interface {
+	Element
+	process(input SeriesSet) SeriesSet
+}
+
+// MetricsSecondStage handles second stage metrics operations (topK/bottomK)
+type MetricsSecondStage struct {
```
mdisibio, line 303: The metrics first stage elements all shared the same struct, and now I wish we had broken them out into dedicated structs.  Let's go ahead and break these out into separate Topk/BottomKElement struct and `newTopK()/newBottomK()` funcs?  Or at least rename to something like `TopKBottomKElement` so that we aren't tempted to make another mega struct for all second stage metrics elements.

pkg/traceql/engine.go
```diff
@@ -40,17 +40,17 @@ func Compile(query string) (*RootExpr, SpansetFilterFunc, metricsFirstStageEleme
 
 	err = expr.validate()
 	if err != nil {
-		return nil, nil, nil, nil, err
+		return nil, nil, nil, nil, nil, err
 	}
 
-	return expr, expr.Pipeline.evaluate, expr.MetricsPipeline, req, nil
+	return expr, expr.Pipeline.evaluate, expr.MetricsPipeline, expr.MetricsSecondStage, req, nil
```
mdisibio, line 46: As cumbersome as this is, I think it's the best way to start, since there is a lot of specialty in which pieces of the pipeline are executed by which modules.  The whole expression is never executed together.   Ideally we would create a holistic "metrics pipeline" that doesn't differentiate the stages so much, but it would be some work, and possibly introduce some wasted overhead (i.e. initializing span-level aggregation in the frontend that will never use it).

pkg/traceql/engine_metrics.go
```diff
@@ -1112,6 +1126,14 @@ func (m *MetricsFrontendEvaluator) Results() SeriesSet {
 	m.mtx.Lock()
 	defer m.mtx.Unlock()
 
+	if m.metricsSecondStage != nil {
```
mdisibio, line 1129: I find this logic order a little confusing since it checks second stage first?  

Maybe something like:

```
result := m.metricsPipeline.result()

if m.metricsSecondStage != nil {
result = m.metricsSecondStage.process(result)
}

return result
```

(also in MetricsEvaluator.Result())

pkg/traceql/engine_metrics.go
```diff
@@ -1097,8 +1110,9 @@ func (e *MetricsEvalulator) sampleExemplar(id []byte) bool {
 // MetricsFrontendEvaluator pipes the sharded job results back into the engine for the rest
 // of the pipeline.  i.e. This evaluator is for the query-frontend.
 type MetricsFrontendEvaluator struct {
-	mtx             sync.Mutex
-	metricsPipeline metricsFirstStageElement
+	mtx                sync.Mutex
+	metricsPipeline    metricsFirstStageElement
+	metricsSecondStage metricsSecondStageElement
```
mdisibio, line 1115: nit: these names stutter, we could do `firstStage`, `secondStage`

mdisibio (commented):
Looking good so far.  Can review again after the process functions are implemented.  To make sure we have the same understanding:  I believe the result of our discussion was that they need to work like Prometheus and operate on each timestamp interval independently.  For example `topk(10)` can return 10 different series at each interval.   In addition to the design, was the performance concerns.  This is so that we can push down some of the top/bottom filtering to the job/block level.  Each job can return it's top 10, and the frontend then chooses the final top 10.  If it is designed the other way, the 10 series with the highest values anywhere in the query, then nothing could be pushed down.  All data would have to gathered up in the frontend.

knylander-grafana:
Should we add doc for this?

pkg/traceql/ast_metrics.go
electron0zero: done

pkg/traceql/engine_metrics.go
electron0zero: done

pkg/traceql/engine_metrics.go
electron0zero: renamed and trimmed metrics prefix

docs/sources/tempo/traceql/metrics-queries/_index.md
```diff
@@ -74,6 +74,8 @@ Refer to [Solve problems using metrics queries](ref:solve-problems-mq) for some
 TraceQL metrics queries currently include the following functions for aggregating over groups of spans: `rate`, `count_over_time`, `sum_over_time`, `max_over_time`, `min_over_time`, `avg_over_time`, `quantile_over_time`, `histogram_over_time`, and `compare`.
 These functions can be added as an operator at the end of any TraceQL query.
 
+`topk` and `bottomk` functions are supported on TraceQL metrics functions.
```
knylander-grafana, line 77:
```suggestion
The `topk` and `bottomk` functions are supported on TraceQL metrics functions.
```

pkg/traceql/engine_metrics.go
```diff
@@ -1484,3 +1505,119 @@ func FloatizeAttribute(s Span, a Attribute) (float64, StaticType) {
...
+					value: value,
+				})
+			}
+		}
+
+		// we have iterated over all series for this timestamp
+		// empty the heap and record these series in result set
+		for h.Len() > 0 {
+			sv := heap.Pop(h).(seriesValue)
+			result[sv.key] = input[sv.key]
```
mdisibio, line 1560: This is adding the entire series to the result, instead of just the 1 timestamp value.  I think it should be something like `result[sv.key].Values[i] = input[sv.key].Values[i]`. The first time the series is added to the result it needs to be initialized with valueLength NaNs, and any still unset at the end are dropped further up the chain.

pkg/traceql/engine_metrics.go
```diff
@@ -1484,3 +1505,119 @@ func FloatizeAttribute(s Span, a Attribute) (float64, StaticType) {
...
+				heap.Push(h, seriesValue{
+					key:   key,
+					value: value,
+				})
+				continue
+			}
+
+			// If new value is less than largest in heap, replace it
+			largest := (*h)[0]
+			if value < largest.value {
```
mdisibio, line 1605: Since these are so similar, it would probably be worth consolidating.

pkg/traceql/ast_metrics.go
```diff
@@ -0,0 +1,376 @@
...
+	return "unknown"
+}
+
+// TopKBottomK handles second stage topK/bottomK operations
+type TopKBottomK struct {
+	op    SecondStageOp
+	limit int
+}
+
+func newTopKBottomK(op SecondStageOp, limit int) *TopKBottomK {
```
mdisibio, line 338: Debating on if all of the second stage ops need to be part of the same enum.  For example when we add another op, it would be possible to call `newTopKBottomK( somethingElse)` which shouldn't be valid.  Maybe this could be `newTopKBottomK(top bool)`

pkg/traceql/engine_metrics.go
```diff
@@ -1484,3 +1505,119 @@ func FloatizeAttribute(s Span, a Attribute) (float64, StaticType) {
 	}
 	return f, v.Type
 }
+
+// processTopK implements TopKBottomK topk method
+func processTopK(input SeriesSet, limit int) SeriesSet {
```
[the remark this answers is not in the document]
mdisibio: In this case it means that avg_over_time() is not shardable and must be computed at the query-frontend.  Whatever follows an unshardable element is also unshardable.  So it means that `{ } | avg_over_time(...) | topk(...)`, the topk must be done only at the query-frontend.

Some queries are shardable, for example min/max_over_time, the top/bottomk can also be pushed down to the earliest level.

Starting with always making the secondStage unshardable and computed only in the frontend sounds good.  That is more straightforward and we can prioritize correctness first.  Adding the shardable/unshardable aspect to the AST will be more involved.

docs/sources/tempo/traceql/metrics-queries/functions.md
```diff
@@ -187,6 +187,17 @@ This is unlikely to tell you anything useful (what does a median status code of
...
+## `topk` and `bottomk` functions
+
+We support the `topk` and `bottomk` aggregation functions that allows you to aggregate the TraceQL metrics and further
+process the metrics produced by TraceQL Metrics.
+
+TraceQL Metrics topk and bottomk functions behave similar to PromQL topk and bottomk functions. example usage:
+
+- `{ } | rate() by(resource.service.name) | bottomk(5)`
+- `{ } | rate() by(resource.service.name) | topk(5)`
+
```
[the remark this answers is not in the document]
electron0zero: @knylander-grafana this looks good to me but I want to know how we have `The max that will ever be returned in 10,000` that 10,000 limit? is where is this coming from?

pkg/traceql/engine_metrics.go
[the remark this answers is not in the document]
electron0zero: had a chat with @javiermolinar about this and updated the code to handle this.

I am only handling topk and bottomk in frontend for now, and we can handle pushdown later when first stage is shardable.

pkg/traceql/engine_metrics_test.go
```diff
@@ -1234,6 +1234,357 @@ func TestHistogramOverTime(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			fmt.Printf("== input seriesSet: %v\n", tt.input)
+			result := processBottomK(tt.input, tt.limit)
+			fmt.Printf("== result seriesSet: %v\n", result)
+			expectSeriesSet(t, tt.expected, result)
+		})
+	}
+}
+
 func runTraceQLMetric(req *tempopb.QueryRangeRequest, inSpans ...[]Span) (SeriesSet, error) {
```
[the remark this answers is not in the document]
electron0zero: included this test.

pkg/traceql/engine_metrics.go
```diff
@@ -1484,3 +1505,119 @@ func FloatizeAttribute(s Span, a Attribute) (float64, StaticType) {
...
+	var valueLength int
+	for _, series := range input {
```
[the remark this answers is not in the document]
electron0zero: updated, and removed this loop

docs/sources/tempo/traceql/metrics-queries/functions.md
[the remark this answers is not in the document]
electron0zero: updated to include these changes.

pkg/traceql/ast_metrics.go
```diff
@@ -0,0 +1,376 @@
...
+func (m *TopKBottomK) process(input SeriesSet) SeriesSet {
+	// if input size is less than limit, return input as is
+	if len(input) <= m.limit {
+		return input
+	}
+
+	// limit can't be zero or negative and will fail query validation
+	// if input is empty, return empty SeriesSet
+	if len(input) == 0 {
+		return SeriesSet{}
```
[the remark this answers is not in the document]
electron0zero: removed, as validation will catch this.

pkg/traceql/ast_metrics.go
```diff
@@ -0,0 +1,376 @@
...
+	}
+
+	switch m.op {
+	case OpTopK:
+		return processTopK(input, m.limit)
+	case OpBottomK:
+		return processBottomK(input, m.limit)
+	default:
+		// unknown operation, return empty SeriesSet, we shouldn't reach here
+		return SeriesSet{}
```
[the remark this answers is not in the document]
electron0zero: done.

pkg/traceql/ast_metrics.go
```diff
@@ -0,0 +1,387 @@
...
+	if len(a.by) > maxGroupBys {
+		return newUnsupportedError(fmt.Sprintf("metrics group by %v values", len(a.by)))
+	}
+
+	return nil
+}
+
+var _ firstStageElement = (*MetricsAggregate)(nil)
+
+// secondStageElement represents operations that are performed
```
electron0zero, line 300: new code for second stage is from here.

docs/sources/tempo/traceql/metrics-queries/functions.md
```diff
@@ -181,21 +196,51 @@ To demonstrate this flexibility, consider this nonsensical quantile on `span.htt
...
+## `topk` and `bottomk` functions
+
+TraceQL supports the `topk` and `bottomk` functions that let you aggregate and process TraceQL metrics.
+These functions are similar to their equivalent PromQL functions. For example:
+
+- `{ } | rate() by(resource.service.name) | bottomk(5)`
+- `{ } | rate() by(resource.service.name) | topk(5)`
+
+When a query response is larger than the maximum, you can use to return only the specified number
+from 1 through `k` of the number of the top or bottom results.
```
knylander-grafana, line 216: I think we have a missing word in here: "you can use to return..." 
What word should be here?

docs/sources/tempo/traceql/metrics-queries/functions.md
```diff
@@ -181,21 +196,51 @@ To demonstrate this flexibility, consider this nonsensical quantile on `span.htt
...
+Because it's evaluated at each data point, you'll get the top series for each data point.
 
 ## The `compare` function
 
 The `compare` function is used to split a set of spans into two groups: a selection and a baseline.
 It returns time-series for all attributes found on the spans to highlight the differences between the two groups.
 
-This is a powerful function that's best understood by using the [**Comparison** tab in Traces Drilldown](https://grafana.com/docs/grafana/<GRAFANA_VERSION>/explore/simplified-exploration/traces/investigate/#comparison).
+This is a powerful function that's best understood by using the [**Comparison
+** tab in Traces Drilldown](https://grafana.com/docs/grafana/<GRAFANA_VERSION>/explore/simplified-exploration/traces/investigate/#comparison).
```
knylander-grafana, line 239:
```suggestion
This powerful function is best understood by using the [**Comparison** tab in Traces Drilldown](https://grafana.com/docs/grafana/<GRAFANA_VERSION>/explore/simplified-exploration/traces/investigate/#comparison).
```

knylander-grafana (approved):
Thank you for updating docs. :) I have two minor suggestions to correct a missing word and remove an extra space.

modules/querier/querier_query_range.go
```diff
@@ -158,6 +159,12 @@ func queryRangeTraceQLToProto(set traceql.SeriesSet, req *tempopb.QueryRangeRequ
 
 		exemplars := make([]tempopb.Exemplar, 0, len(s.Exemplars))
 		for _, e := range s.Exemplars {
+			// skip exemplars that has NaN value
```
electron0zero, line 162: decided to cleanup `queryRangeTraceQLToProto` in it's own PR: https://github.com/grafana/tempo/pull/4978, so forn now I made the same change here as well.

mdisibio: We need to add bounds checking on `i` here and in the other spot. Reliably reproduce panics both with `{} | compare(...)` query (not even involving topk/bottomk).  I'm guessing this is already fixed by the exemplar filtering in #4962 , but let's guard here too.

```
github.com/grafana/tempo/modules/querier.queryRangeTraceQLToProto(0x1402e1d3860, 0x14010896280)
	/Users/marty/src/tempo/modules/querier/querier_query_range.go:164 +0x994
github.com/grafana/tempo/modules/querier.(*Querier).queryBlock(0x14000576588, {0x1075c7858, 0x14001252150}, 0x14010896280)
	/Users/marty/src/tempo/modules/querier/querier_query_range.go:126 +0xa8c
github.com/grafana/tempo/modules/querier.(*Querier).QueryRange(0x14000576588, {0x1075c7858, 0x14001252150}, 0x14010896280)
	/Users/marty/src/tempo/modules/querier/querier_query_range.go:25 +0x1e4
github.com/grafana/tempo/modules/querier.(*Querier).QueryRangeHandler(0x14000576588, {0x1075bd6b0, 0x14000986c40}, 0x14000994dc0)
	/Users/marty/src/tempo/modules/querier/http.go:406 +0x8f4
```

electron0zero: added in [`b4169be` (#4646)](https://github.com/grafana/tempo/pull/4646/commits/b4169be91730bafaea504af3ee9c15cf23c83919)

mdisibio (approved):

alexbikfalvi:
@electron0zero @knylander-grafana I think the documentation didn't publish to next:
https://github.com/grafana/tempo/actions/runs/14386797134/job/40344098510

Is there a way to retrigger?

electron0zero:
@alexbikfalvi next run that triggered on another PR took care of this so it's published now: https://grafana.com/docs/tempo/next/traceql/metrics-queries/functions/#topk-and-bottomk-functions

so we don't need to re-trigger it now.

tho we should look into this failure, cc @knylander-grafana @jdbaldry

09jvilla:
@electron0zero nice work on this! I checked out the demo video since you mentioned it in your presentation and enjoyed watching it :) My only question was why we were seeing more than 5 series for a topk(5) query but your docs answered it perfectly when I checked them.

PR #4660, opened 2025-02-05
[rhythm] Add rollout-operator and allow simultaneous update of block-builder pods

mapno:
**What this PR does**:

Introduced the rollout-operator so the pods in the block-builder's statefulset can simultaneously update. Otherwise there is the risk of ownership conflicts in partitions during rollouts.

### It solves the following scenario:

There are 6 partitions and 3 block-builders.

Partition assignment is as follows:
- `block-builder-0` has assigned partitions `0` and `3`.
- `block-builder-1` has assigned partitions `1` and `4`.
- `block-builder-2` has assigned partitions `2` and `5`.

If a new block-builder is deployed, partition assignment will change to
- `block-builder-0` has assigned partitions `0` and `4`.
- `block-builder-1` has assigned partitions `1` and `5`.
- `block-builder-2` has assigned partitions `2`.
- `block-builder-3` has assigned partitions `3`.

As soon as block-builders are rolled out, they'll have the new config, while older replicas will have the old config—ie. at one point, `block-builder-3` and `block-builder-0` both will own partition `3`.

### Alternatives

A previous attempt was made in https://github.com/grafana/tempo/pull/4642, but in the end discarded. It required Kubernetes >=v1.24 and `MaxUnavailableStatefulSet` feature gate enabled, which is in alpha and not supported in many environments.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/operations/jsonnet/microservices/rollout-operator.libsonnet
+++ b/operations/jsonnet/microservices/rollout-operator.libsonnet
@@ -39,7 +39,7 @@
     rollout_operator_replica_template_access_enabled: true,
 
     // Automatically add groups based on enabled features
-    rollout_operator_enabled_groups+:: 
+    rollout_operator_enabled_groups+::
       (if $._config.block_builder_concurrent_rollout_enabled then ['block-builder'] else []) +
       (if $._config.multi_zone_ingester_enabled then ['ingester'] else []),
 
--- a/operations/jsonnet/microservices/rollout-operator.libsonnet
+++ b/operations/jsonnet/microservices/rollout-operator.libsonnet
@@ -1,54 +1,114 @@
 {
   local k = import 'ksonnet-util/kausal.libsonnet',
 
+  local clusterRole = k.rbac.v1.clusterRole,
+  local clusterRoleBinding = k.rbac.v1.clusterRoleBinding,
   local container = k.core.v1.container,
   local deployment = k.apps.v1.deployment,
+  local mutatingWebhook = k.admissionregistration.v1.mutatingWebhook,
+  local mutatingWebhookConfiguration = k.admissionregistration.v1.mutatingWebhookConfiguration,
   local policyRule = k.rbac.v1.policyRule,
   local role = k.rbac.v1.role,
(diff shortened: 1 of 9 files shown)


operations/jsonnet/microservices/rollout-operator.libsonnet
```diff
@@ -0,0 +1,73 @@
...
+
+  _config+:: {
+    rollout_operator_enabled: $._config.block_builder_concurrent_rollout_enabled,
+  },
+
+  rollout_operator_args:: {
+    'kubernetes.namespace': $._config.namespace,
+  },
+
+  rollout_operator_container::
```
javiermolinar, line 20: There is already a rollout-operator deployment in 

https://github.com/grafana/tempo/blob/545ab11c71f11f363b2f4a5b86fd4766ba9a52c7/operations/jsonnet/microservices/multi-zone.libsonnet#L146

Should we merge both?

operations/jsonnet/microservices/rollout-operator.libsonnet
```diff
@@ -0,0 +1,73 @@
...
+    container.new('rollout-operator', $._images.rollout_operator) +
```
javiermolinar, line 21: We can bump the existing version wich is outdated: 

https://github.com/grafana/tempo/blob/545ab11c71f11f363b2f4a5b86fd4766ba9a52c7/operations/jsonnet/microservices/config.libsonnet#L6

operations/jsonnet/microservices/rollout-operator.libsonnet
mapno: Yes! I had forgotten, thanks!

operations/jsonnet/microservices/rollout-operator.libsonnet
mapno: Ha, latest is `v0.23.0` 🙈

Logiraptor (approved):

PR #4673, opened 2025-02-07
Return the operand as the only value if the tag is already filtered in the query

mapno:
**What this PR does**:

Return the operand as the only value if the tag is already filtered in the query. I think this is more in line with what one would expect.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/engine.go
+++ b/pkg/traceql/engine.go
@@ -181,9 +181,11 @@ func (e *Engine) ExecuteTagValues(
 	// I.e. we are autocompleting resource.service.name and the query was {resource.service.name="foo"}
 	for _, c := range autocompleteReq.Conditions {
 		if c.Attribute == tag && c.Op == OpEqual {
-			// If the tag is already filtered in the query, then we can just return the operand
-			// as the only value.
-			cb(c.Operands[0])
+			// If the tag is already filtered in the query,
+			// then we can just return the operand as the only value.
+			if len(c.Operands) > 0 {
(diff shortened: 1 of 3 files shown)


pkg/traceql/engine.go
```diff
@@ -181,6 +181,9 @@ func (e *Engine) ExecuteTagValues(
 	// I.e. we are autocompleting resource.service.name and the query was {resource.service.name="foo"}
 	for _, c := range autocompleteReq.Conditions {
 		if c.Attribute == tag && c.Op == OpEqual {
+			// If the tag is already filtered in the query, then we can just return the operand
+			// as the only value.
+			cb(c.Operands[0])
```
joe-elliott, line 186: Operands is always guaranteed to have a length?

joe-elliott (approved):
agreed this is a better return. i think the code is fine, but one simple question.

pkg/traceql/engine.go
mapno: No, it could be an empty string. I'll cover that case

joe-elliott: oh, sorry. i mean is the c.Operands slice always guaranteed to have something at index 0?

mapno: IMO it needs both checks

mapno: Actually, I don't see a good logic to skip values like empty string.

I've done a check for non-empty `c.Operands` only

PR #4686, opened 2025-02-12
[Metrics Generator] Allow running on a different source of data

flxbk:
**What this PR does**:

This PR makes it possible to run the metrics-generation features of metrics-generator (i.e. `spanmetrics` and `service-graph` processors, but not `localblocks`) on a different Kafka-compatible data source. It introduces ~two~ three new configuration options to achieve this:

1. `DisableLocalBlocks`: this defaults to `false`, when set to `false` there's no change in behavior. When this is set to `true`, metrics-generator will not run the `localblocks` processor even when it's configured for a tenant.

1. `DisableGRPC`: this defaults to `false` as well. When it's `true` the gRPC server setup is skipped and metrics generator doesn't join the generator ring (which it only needs to do to make itself available to be queried via gRPC, so when gRPC is disabled joining the ring makes no sense).

1. `Codec`: this controls what data type metrics generators expect to find in the Kafka records they consume. This PR adds support for consuming records that contain `ptrace.Traces` instead of the `tempopb.PushBytesRequest` that's already supported.

This PR also introduces a way of making metrics generator watch a configurable partition ring (currently watching the ingester ring is hardcoded).

Fixes n/a

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/ingester/ingester.go
```diff
@@ -53,10 +53,6 @@ var tracer = otel.Tracer("modules/ingester")
 
 const (
 	ingesterRingKey = "ring"
-
-	// PartitionRingKey is the key under which we store the partitions ring used by the "ingest storage".
-	PartitionRingKey  = "ingester-partitions"
-	PartitionRingName = "ingester-partitions"
```
flxbk, line 59: I had to move this because not doing so would create an import cycle.

pkg/ingest/reader_client.go
```diff
@@ -57,7 +57,7 @@ func NewGroupReaderClient(kafkaCfg KafkaConfig, partitionRing ring.PartitionRing
 		kgo.SessionTimeout(3*time.Minute),
 		kgo.RebalanceTimeout(5*time.Minute),
 		kgo.Balancers(NewCooperativeActiveStickyBalancer(partitionRing)),
-		kgo.ConsumeResetOffset(kgo.NewOffset().AtStart()),
+		kgo.ConsumeResetOffset(kgo.NoResetOffset().AtEnd()),
```
flxbk, line 60: This makes the Kafka consumer start at the end of the topic (i.e. when no offsets are present) but stop consuming if it ever tries to read an offset that's out of range (which would happen if it fell behind the _entire_ backlog). In this case an operator needs to step in and reset offsets to what they think makes sense at that point.

`NewOffset().AtStart()` makes it start at the beginning of the topic, which is not reasonable when there's already a day worth of data present at first startup.

modules/generator/config.go
```diff
@@ -40,6 +48,13 @@ type Config struct {
 	QueryTimeout          time.Duration `yaml:"query_timeout"`
 	OverrideRingKey       string        `yaml:"override_ring_key"`
 
+	// JoinRing controls whether to join the metrics-generator ring and become
+	// available to serve requests.
+	JoinRing bool `yaml:"join_ring"`
```
yvrhdn, line 53: I feel like `JoinRing` doesn't fully cover what this toggle does. Maybe it's more something like `HostLocalBlocks` or `RunLocalBlocks`?

yvrhdn (commented):
Overall this looks good to me but I'll let someone with more experience on the new Kafka ingest and local-blocks processor look into it 🙂

modules/generator/config.go
flxbk: You're right, I think what you're suggesting would be a better name. Lets wait to see what others think and then I'll rename according to the consensus.

modules/generator/encoding_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/tempo/pkg/tempopb"
+	v1 "github.com/grafana/tempo/pkg/tempopb/trace/v1"
+	"github.com/grafana/tempo/pkg/util/test"
+)
+
+func BenchmarkOTLPDecoder(b *testing.B) {
```
joe-elliott, line 13: how does the otlp decoder compare to the tempo one?

modules/generator/encoding.go
```diff
@@ -0,0 +1,63 @@
...
+	d.dec.Reset()
+	spanBytes, err := d.dec.Decode(data)
+	if err != nil {
+		return nil, err
+	}
+
+	trace := tempopb.Trace{}
+	return func(yield func(*tempopb.PushSpansRequest, error) bool) {
+		for _, tr := range spanBytes.Traces {
+			trace.Reset()
```
joe-elliott, line 36: `Reset()` is disappointing :), but i liked that you called it

modules/generator/encoding.go
```diff
@@ -0,0 +1,63 @@
...
+// decode implements codec.
+func (d *otlpDecoder) decode(data []byte) (iter.Seq2[*tempopb.PushSpansRequest, error], error) {
+	d.trace.ResourceSpans = d.trace.ResourceSpans[:0]
+	err := d.trace.Unmarshal(data)
+	if err != nil {
+		return nil, err
+	}
+
+	return func(yield func(*tempopb.PushSpansRequest, error) bool) {
+		yield(&tempopb.PushSpansRequest{Batches: d.trace.ResourceSpans}, nil)
```
joe-elliott, line 61: so we are returning a slice owned and reused by this struct into the bowels of the metrics generator. we are certain that we will not reuse it here while it's being processed in the generator?

modules/generator/generator.go
```diff
@@ -369,6 +383,10 @@ func (g *Generator) createInstance(id string) (*instance, error) {
 }
 
 func (g *Generator) CheckReady(_ context.Context) error {
+	if !g.cfg.JoinRing {
```
joe-elliott, line 386: there's a lot of small behavior changes scattered around due to this "JoinRing" boolean. obviously this is related to one of my previous comments. I'm fine with things being messy in the interim before we land on a final pattern that both adaptive traces and rhythm can use, but we really need comments for lines like this to help us understand why we would behave this way when JoinRing = false

pkg/ingest/reader_client.go
joe-elliott: definitely would like someone from rhythm to review this line

pkg/ingest/config.go
```diff
@@ -32,6 +32,10 @@ const (
 	// in the worst case scenario, which is expected to be way above the actual one.
 	maxProducerRecordDataBytesLimit = producerBatchMaxBytes - 16384
 	minProducerRecordDataBytesLimit = 1024 * 1024
+
+	// PartitionRingKey is the key under which we store the partitions ring used by the "ingest storage".
+	PartitionRingKey  = "ingester-partitions"
```
mdisibio, line 37: @mapno Can you give thoughts here?  I think this is not maintaining the distinction between `ingester` and `ingest`, and what "owns" the partitions, that you would know more about.

cmd/tempo/app/modules.go
```diff
@@ -186,16 +188,34 @@ func (t *App) initPartitionRing() (services.Service, error) {
...
+
+	return t.partitionRingWatcher, nil
+}
+
+func (t *App) initPartitionRingWatcher() (services.Service, error) {
+	if !t.cfg.Ingest.Enabled {
+		return services.NewIdleService(nil, nil), nil
+	}
+
+	kvClient, err := kv.NewClient(t.cfg.Generator.Ring.KVStore, ring.GetPartitionRingCodec(), kv.RegistererWithKVName(prometheus.DefaultRegisterer, t.cfg.Ingest.OverrideRingKey+"-watcher"), util_log.Logger)
```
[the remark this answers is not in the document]
flxbk: Right, the only goal of this change was to remove the generator's dependency on the ingester ring. This can be a subservice of the generator instead, I'll change that accordingly.

cmd/tempo/app/modules.go
```diff
@@ -303,11 +323,13 @@ func (t *App) initGenerator() (services.Service, error) {
 	}
 
 	t.cfg.Generator.Ring.ListenPort = t.cfg.Server.GRPCListenPort
-
 	t.cfg.Generator.Ingest = t.cfg.Ingest
-	t.cfg.Generator.Ingest.Kafka.ConsumerGroup = generator.ConsumerGroup
 
-	genSvc, err := generator.New(&t.cfg.Generator, t.Overrides, prometheus.DefaultRegisterer, t.partitionRing, t.store, log.Logger)
+	if t.cfg.Generator.Ingest.Kafka.ConsumerGroup == "" {
+		t.cfg.Generator.Ingest.Kafka.ConsumerGroup = generator.ConsumerGroup
```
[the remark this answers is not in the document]
flxbk: Using a different consumer group name is not technically required to use a second set of generators (as in the second set could also use the same group name to consume a different topic), but I think it definitely makes sense for the group name to be configurable to have a clearer separation of the two consumers. If you're making changes in this area in #4721 it would be nice if you can make it configurable there.

cmd/tempo/app/modules.go
```diff
@@ -317,13 +339,15 @@ func (t *App) initGenerator() (services.Service, error) {
 	}
 	t.generator = genSvc
 
-	spanStatsHandler := t.HTTPAuthMiddleware.Wrap(http.HandlerFunc(t.generator.SpanMetricsHandler))
-	t.Server.HTTPRouter().Handle(path.Join(api.PathPrefixGenerator, addHTTPAPIPrefix(&t.cfg, api.PathSpanMetrics)), spanStatsHandler)
+	if t.cfg.Generator.JoinRing {
```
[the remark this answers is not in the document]
flxbk: @yvrhdn proposed something along the lines of `HostLocalBlocks`. That makes sense to me because (iiuc) the `localblocks` processor is the only one that creates the need for generators to be queryable, and therefore join the ring. My understanding is that in Rhythm that processor will eventually go away and be replaced by the block builder, and at that point we could just drop that flag again and make the non-localblocks mode the only supported one. Does that sound right?

pkg/ingest/config.go
mapno: What owns the partitions is whatever component is in the partition ring. This change allows to use a different ring name/key, so multiple partition rings can co-exist and the generator can watch one or the other.

From this PR I understand the idea is to isolate the two parts: the ring and the ring watcher. Tempo's partition ring will continue to be owned by the ingesters. The generator now should be able to watch another ring that's not the one owned by the ingesters. It's confusing.

mapno (commented):
I haven't done an in-depth review. Still have some comments.

cmd/tempo/app/modules.go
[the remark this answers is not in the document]
yvrhdn: Yes, that's how I understand it. We only use the metrics-generator ring to write from distributor -> metrics-generator and query from querier -> metrics-generator (for local-blocks). So if we 1) only ingest from Kafka and 2) don't run local-blocks we don't need the ring.

To make this explicit we could also split it up into two booleans? `KafkaIngestOnly` and `DisableLocalBlocks`. If both are set we don't join the ring.

pkg/ingest/reader_client.go
flxbk: I have removed this change from the diff. @mdisibio's changes in #4721 seems to be enough to let generators catch up even when there's substantial backlog.

cmd/tempo/app/modules.go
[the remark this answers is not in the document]
flxbk: I have taken up @mdisibio's suggestion to add a new target for running metrics generator without the localblocks processor and the config appropriate for that. To resolve the dual (write) use of `t.partitionRingWatcher` I have added a new invisible target for the generator partition ring watcher that the new target depends on.

modules/generator/encoding.go
```diff
@@ -0,0 +1,63 @@
...
+	dec *ingest.Decoder
+}
+
+func newTempoDecoder() *tempoDecoder {
+	return &tempoDecoder{dec: ingest.NewDecoder()}
+}
+
+// decode implements codec.
+func (d *tempoDecoder) decode(data []byte) (iter.Seq2[*tempopb.PushSpansRequest, error], error) {
+	d.dec.Reset()
```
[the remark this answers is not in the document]
flxbk: I have rebased on top of the concurrency change and made the decoder local to the goroutine.

modules/generator/encoding_test.go
flxbk: I have added a benchmark for that decoder as well.

modules/generator/encoding.go
flxbk: Pretty certain from what I can discern. This is also following the pattern of what the code is already doing before this PR, so I think it should be fine.

modules/generator/generator.go
flxbk: I have tried to find better names for the config flags and introduced a helper `shouldJoinRing` to hopefully make this kind of behaviour change this less confusing.

cmd/tempo/app/modules.go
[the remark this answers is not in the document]
flxbk: I tried making this a subservice of the generator, but if we want to do that we also need to make the ingester partition ring watcher a subservice (or else things would be very awkward, i.e. we pass an external service in once case but use a subservice in another).

For now my resolution of this is to create a new (invisible) target for the generator's ring watcher that can be passed to the generator. This is used by the new `generator-no-local-blocks` and could also be used by the regular generator target (but I haven't included that last change in this diff to keep it smaller).

pkg/ingest/config.go
flxbk: > The generator now should be able to watch another ring that's not the one owned by the ingesters.

Exactly, that's one of the two high-level things this PR is trying to achieve (the other being running a subset of processors only).

modules/generator/encoding.go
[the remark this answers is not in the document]
flxbk: 👍 Good point, moved there.

modules/generator/config.go
```diff
@@ -98,9 +115,19 @@ func (cfg *Config) Validate() error {
...
+		return fmt.Errorf("invalid codec: %s, valid choices are %s", cfg.Codec, validCodecs)
+	}
+
 	return nil
 }
 
+func (cfg *Config) shouldJoinRing() bool {
+	// If both gRPC and the localblocks processor are disabled, generators don't need
+	// to join the ring.
+	return !(cfg.DisableLocalBlocks && cfg.DisableGRPC)
```
joe-elliott, line 128: what does disabling local blocks have to do with joining the ring or not? are these not independent? if DisableGRPC=true it makes sense to me to not join the ring and skip all that logic, but I don't see the connection with disabling local blocks.

cmd/tempo/app/modules.go
```diff
@@ -667,17 +724,19 @@ func (t *App) setupModuleManager() error {
...
-		Ingester:         {Common, Store, MemberlistKV, PartitionRing},
-		MetricsGenerator: {Common, OptionalStore, MemberlistKV, PartitionRing},
-		Querier:          {Common, Store, IngesterRing, MetricsGeneratorRing, SecondaryIngesterRing},
-		Compactor:        {Common, Store, MemberlistKV},
-		BlockBuilder:     {Common, Store, MemberlistKV, PartitionRing},
+		QueryFrontend:                 {Common, Store, OverridesAPI},
+		Distributor:                   {Common, IngesterRing, MetricsGeneratorRing, PartitionRing},
+		Ingester:                      {Common, Store, MemberlistKV, PartitionRing},
+		MetricsGenerator:              {Common, OptionalStore, MemberlistKV, PartitionRing},
+		MetricsGeneratorNoLocalBlocks: {Common, GeneratorRingWatcher},
```
joe-elliott, line 736: The dependence between "no local blocks" and the partition ring watcher don't make sense to me. Shouldn't this be something like `MetricsGeneratorPartitionRing`

modules/generator/config.go
flxbk: I believe you're right and these two should be independent, I have changed it in https://github.com/grafana/tempo/pull/4686/commits/846e38008e6f2dd09653a5d3a0688dd325410c62.
My initial thinking was that localblocks' function was to produce data that can be queried via gRPC, but I think that understanding is incorrect.

cmd/tempo/app/modules.go
flxbk: In a different PR [comment](https://github.com/grafana/tempo/pull/4686#discussion_r1960544177), @mdisibio suggested we could use a separate target for this metrics generator mode. This is what I did here, however that new mode still needs a way to watch a (configurable) partition ring, hence the dependency on `GeneratorRingWatcher`.

mdisibio (approved):
Thanks for the amount of work put into this. The final result with the dedicated target is working well to keep all of the unique bits together, and I think it gives a clean separation between the modes, so that we can continue to iterate on naming or other things without breaking config or deployment changes. LGTM

flxbk:
Thanks for the review and all of the suggestions that have helped me make this better!

I haven't added a changelog entry yet to avoid eternal rebases/conflicts with other entries, but happy to do that before this goes in.

flxbk:
Added changelog and rebased to resolve conflicts in `changelog.md`.

PR #4690, opened 2025-02-13
Rewrite traces using rebatching

stoewer:
**What this PR does**:
Rebatch traces in order to remove redundant `ResourceSpans` and `ScopeSpans` to reduces the number of rows on resource and scope level. This has shown to speed up queries with matching resource attributes and also reduces the overall file size.

Fixes grafana/tempo-squad#548

**Checklist**
- [x] Tests updated
- [ ] Documentation added


mapno (approved):
Very nice code, LGTM

tempodb/encoding/vparquet4/schema.go
```diff
@@ -547,6 +547,8 @@ func traceToParquetWithMapping(id common.ID, tr *tempopb.Trace, ot *Trace, dedic
 		}
 	}
 
+	// modify trace by rebatching spans and assigning nested set values
+	rebatchTrace(ot)
```
joe-elliott, line 551: should we also call this after combining here-ish?

https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet4/combiner.go#L139

we seem to have three things we do every time we build a trace:
- rebatch
- nested set model
- service stats

should we create some holistic function like "finalizeTrace" that accomplishes this? that way if we add more of this kind of processing we will have an easy place to put it?

```
func finalizeTrace() {
  rebatch()
  assignNestedSetModel()
  ...
}
```

tempodb/encoding/vparquet4/schema_test.go
```diff
@@ -919,10 +923,86 @@ func BenchmarkExtendReuseSlice(b *testing.B) {
...
+	for range 1_000 {
+		nb := 20 + rand.Intn(5)
+		id := test.ValidTraceID(nil)
+		traces = append(traces, test.AddDedicatedAttributes(test.MakeTrace(nb, test.ValidTraceID(id))))
+	}
+
+	b.ResetTimer()
+	b.ReportAllocs()
+
+	for i := range b.N {
```
joe-elliott, line 944: go 1.24 has the cool new `b.Loop()`!

you're welcome to use or ignore, just learned about this myself.

tempodb/encoding/vparquet4/rebatch_trace_test.go
```diff
@@ -0,0 +1,208 @@
+package vparquet4
+
+import (
+	"crypto/rand"
```
joe-elliott, line 4: nit: math/rand good enough?

tempodb/encoding/vparquet4/rebatch_trace.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+// rebatchTrace removes redundant ResourceSpans and ScopeSpans from the trace through rebatching.
+func rebatchTrace(trace *Trace) {
+	if len(trace.ResourceSpans) == 0 {
+		return
+	}
+
+	// preallocate a map and a slice to collect the indices of unique ResourceSpans and ScopeSpans
+	uniqueIndexes := make([]int, 0, max(len(trace.ResourceSpans), len(trace.ResourceSpans[0].ScopeSpans)))
```
joe-elliott, line 18: took me a bit to figure out what you were using this. appreciate the effort on the in place alg

joe-elliott (commented):
nice implementation. couple nits and only one "real" question.

thanks!

tempodb/encoding/vparquet4/schema_test.go
stoewer: > just learned about this myself

Good suggestion! Definitely looking forward to using `b.Loop()` more often instead of the old `for i := 0; i < b.N; i++`. However, in this case I'm actually using the `i` to select one of the pre-allocated traces

tempodb/encoding/vparquet4/rebatch_trace_test.go
stoewer: The `Read()` function from `math/rand` is deprecated. There are two alternatives:
* Use `Read()` from `crypto/math`
* Create a `ChaCha8` random number generator from `math/rand/v2` and use it's `Read()` method

I think it would be great to migrate from `math/rand` to `math/rand/v2` but it might be better to do this package wide, or even project wide. There are subtle differences between `math/rand` and `math/rand/v2` and I'm not sure if it creates confusion to use both at the same time.

PR #4698, opened 2025-02-14
add host info processor

rlankfo:
**What this PR does**:
Introduces first draft of host info processor

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/processor/hostinfo/config.go
+++ b/modules/generator/processor/hostinfo/config.go
@@ -1,21 +1,33 @@
 package hostinfo
 
 import (
+	"errors"
 	"flag"
-	"time"
+
+	"github.com/prometheus/common/model"
 )
 
 type Config struct {
 	// HostIdentifiers defines the list of resource attributes used to derive
 	// a unique `grafana.host.id` value. In most cases, this should be [ "host.id" ]
-	HostIdentifiers []string      `yaml:"host_identifiers"`
-	StaleDuration   time.Duration `yaml:"stale_duration"`
+	HostIdentifiers []string `yaml:"host_identifiers"`
 	// MetricName defines the name of the metric that will be generated
 	MetricName string `yaml:"metric_name"`
 }
 
 func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
 	cfg.HostIdentifiers = []string{"k8s.node.name", "host.id"}
-	cfg.StaleDuration = 2 * time.Minute
 	cfg.MetricName = hostInfoMetric
+}
+
+func (cfg *Config) Validate() error {
+	if len(cfg.HostIdentifiers) == 0 {
+		return errors.New("at least one value must be provided in host_identifiers")
+	}
+
+	if !model.IsValidMetricName(model.LabelValue(cfg.MetricName)) {
+		return errors.New("metric_name is invalid")
+	}
+
+	return nil
 }
--- a/modules/generator/processor/hostinfo/config.go
+++ b/modules/generator/processor/hostinfo/config.go
@@ -1,0 +1,21 @@
+package hostinfo
+
+import (
+	"flag"
+	"time"
+)
+
+type Config struct {
+	// HostIdentifiers defines the list of resource attributes used to derive
+	// a unique `grafana.host.id` value. In most cases, this should be [ "host.id" ]
+	HostIdentifiers []string      `yaml:"host_identifiers"`
+	StaleDuration   time.Duration `yaml:"stale_duration"`
+	// MetricName defines the name of the metric that will be generated
+	MetricName string `yaml:"metric_name"`
+}
+
+func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
+	cfg.HostIdentifiers = []string{"k8s.node.name", "host.id"}
+	cfg.StaleDuration = 2 * time.Minute
+	cfg.MetricName = hostInfoMetric
+}
--- a/modules/generator/processor/hostinfo/config.go
+++ b/modules/generator/processor/hostinfo/config.go
@@ -7,6 +7,10 @@ import (
 	"github.com/prometheus/common/model"
 )
 
+const (
+	defaultHostInfoMetric = "traces_host_info"
+)
+
 type Config struct {
 	// HostIdentifiers defines the list of resource attributes used to derive
 	// a unique `grafana.host.id` value. In most cases, this should be [ "host.id" ]
@@ -17,7 +21,7 @@ type Config struct {
 
 func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
 	cfg.HostIdentifiers = []string{"k8s.node.name", "host.id"}
-	cfg.MetricName = hostInfoMetric
+	cfg.MetricName = defaultHostInfoMetric
 }
 
 func (cfg *Config) Validate() error {
--- a/modules/generator/processor/hostinfo/processor.go
+++ b/modules/generator/processor/hostinfo/processor.go
@@ -68,8 +68,6 @@ func (p *Processor) Shutdown(_ context.Context) {}
 func New(cfg Config, reg registry.Registry, logger log.Logger) (*Processor, error) {
 	labels := make([]string, 1)
 	labels[0] = hostIdentifierAttr
-	myGauge := reg.NewGauge(cfg.MetricName)
-	myGauge.SetExpiration(cfg.StaleDuration)
 	p := &Processor{
 		Cfg:        cfg,
 		logger:     logger,
--- a/modules/generator/processor/hostinfo/processor.go
+++ b/modules/generator/processor/hostinfo/processor.go
@@ -36,7 +36,7 @@ func (p *Processor) PushSpans(ctx context.Context, req *tempopb.PushSpansRequest
 	_, span := tracer.Start(ctx, "hostinfo.PushSpans")
 	defer span.End()
 
-	values := make([]string, 1, 1)
+	values := make([]string, 1)
 
 outer:
 	for i := range req.Batches {
@@ -66,7 +66,7 @@ outer:
 func (p *Processor) Shutdown(_ context.Context) {}
 
 func New(cfg Config, reg registry.Registry, logger log.Logger) (*Processor, error) {
-	labels := make([]string, 1, 1)
+	labels := make([]string, 1)
 	labels[0] = hostIdentifierAttr
 	myGauge := reg.NewGauge(cfg.MetricName)
 	myGauge.SetExpiration(cfg.StaleDuration)
--- a/modules/generator/processor/hostinfo/processor.go
+++ b/modules/generator/processor/hostinfo/processor.go
@@ -1,0 +1,82 @@
+package hostinfo
+
+import (
+	"context"
+
+	"github.com/go-kit/log"
+	"github.com/grafana/tempo/modules/generator/registry"
+	"github.com/grafana/tempo/pkg/tempopb"
+	"go.opentelemetry.io/otel"
+)
+
+const (
+	Name = "host-info"
+
+	hostInfoMetric     = "traces_host_info"
+	hostIdentifierAttr = "grafana.host.id"
+)
+
+var tracer = otel.Tracer("modules/generator/processor/hostinfo")
+
+type Processor struct {
+	Cfg    Config
+	logger log.Logger
+
+	gauge      registry.Gauge
+	registry   registry.Registry
+	metricName string
+	labels     []string
+}
+
+func (p *Processor) Name() string {
+	return Name
+}
+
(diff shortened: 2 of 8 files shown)


modules/generator/processor/hostinfo/processor.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/go-kit/log"
+	"github.com/grafana/tempo/modules/generator/registry"
+	"github.com/grafana/tempo/pkg/tempopb"
+	"go.opentelemetry.io/otel"
+)
+
+const (
+	Name = "host-info"
+
+	hostInfoMetric     = "traces_host_info"
```
cedricziel, line 16: i See a problem here - if we start with the same metric name, we'll blow up the invoice for a lot of customers that dont expect it.

rlankfo: made this configurable

modules/generator/processor/hostinfo/config.go
```diff
@@ -0,0 +1,21 @@
...
+	// a unique `grafana.host.id` value. In most cases, this should be [ "host.id" ]
+	HostIdentifiers []string      `yaml:"host_identifiers"`
+	StaleDuration   time.Duration `yaml:"stale_duration"`
+	// MetricName defines the name of the metric that will be generated
+	MetricName string `yaml:"metric_name"`
+}
+
+func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
+	cfg.HostIdentifiers = []string{"host.id"}
+	cfg.StaleDuration = 2 * time.Minute
```
rlankfo, line 19: Attempting to set the default stale duration as low as possible here as we depend on the absence of these metrics to get accurate counts. I noticed the ManagedRegistry only removes stale series on a 5 minute interval though, so this might be confusing if set to anything less than 5 minutes.

https://github.com/grafana/tempo/blob/main/modules/generator/registry/registry.go#L134

modules/generator/processor/hostinfo/processor.go
```diff
@@ -0,0 +1,71 @@
...
+		resourceSpan := req.Batches[i]
+		attrs := resourceSpan.GetResource().GetAttributes()
+
+	idAttrLoop:
+		for _, idAttr := range p.Cfg.HostIdentifiers {
+			for _, attr := range attrs {
+				if attr.GetKey() == idAttr {
+					labelValues := p.registry.NewLabelValueCombo(
+						[]string{hostIdentifierAttr},
+						[]string{attr.GetValue().GetStringValue()},
```
joe-elliott, line 48: any concern about this not being a string? seems unlikely

modules/generator/processor/hostinfo/processor.go
```diff
@@ -0,0 +1,71 @@
...
+				}
+			}
+		}
+	}
+}
+
+func (p *Processor) Shutdown(_ context.Context) {}
+
+func New(cfg Config, reg registry.Registry, logger log.Logger) (*Processor, error) {
+	myGauge := reg.NewGauge(cfg.MetricName)
```
joe-elliott, line 61: we should fail on invalid config. negative stale duration, no host identifiers, etc.

modules/generator/processor/hostinfo/processor.go
```diff
@@ -0,0 +1,71 @@
...
+
+	idAttrLoop:
+		for _, idAttr := range p.Cfg.HostIdentifiers {
+			for _, attr := range attrs {
+				if attr.GetKey() == idAttr {
+					labelValues := p.registry.NewLabelValueCombo(
+						[]string{hostIdentifierAttr},
+						[]string{attr.GetValue().GetStringValue()},
+					)
+					p.gauge.Set(labelValues, 1)
```
joe-elliott, line 50: `.SetForTargetInfo` better matches the functionality you are trying to achieve here. Perhaps we should rename that method and use it here?

modules/generator/processor/hostinfo/processor.go
```diff
@@ -0,0 +1,71 @@
...
+	for i := 0; i < len(req.Batches); i++ {
+		resourceSpan := req.Batches[i]
+		attrs := resourceSpan.GetResource().GetAttributes()
+
+	idAttrLoop:
+		for _, idAttr := range p.Cfg.HostIdentifiers {
+			for _, attr := range attrs {
+				if attr.GetKey() == idAttr {
+					labelValues := p.registry.NewLabelValueCombo(
+						[]string{hostIdentifierAttr},
```
joe-elliott, line 47: both of these string slices can be alloc'ed once and stored on the struct. the first one can be alloc'ed and never changed. the second you would just set the string at index 0 each time before you call NewLabelValueCombo.

modules/generator/registry/gauge.go
```diff
@@ -164,10 +165,20 @@ func (g *gauge) removeStaleSeries(staleTimeMs int64) {
 	g.seriesMtx.Lock()
 	defer g.seriesMtx.Unlock()
 
+	expTimeMs := staleTimeMs
```
joe-elliott, line 168: not psyched about this custom expiration logic. we run with default stale duration of 15m, but we could likely bring that down.

joe-elliott (commented):
added some thoughts. don't we need to reference this processor somewhere as well? here?

https://github.com/grafana/tempo/blob/main/modules/generator/instance.go#L33-L40

Perhaps try running it using this example to confirm it works the way you want:

https://github.com/grafana/tempo/tree/main/example/docker-compose/local

modules/generator/processor/hostinfo/processor.go
```diff
@@ -0,0 +1,71 @@
...
+func (p *Processor) PushSpans(ctx context.Context, req *tempopb.PushSpansRequest) {
+	_, span := tracer.Start(ctx, "hostinfo.PushSpans")
+	defer span.End()
+
+	for i := 0; i < len(req.Batches); i++ {
+		resourceSpan := req.Batches[i]
+		attrs := resourceSpan.GetResource().GetAttributes()
+
+	idAttrLoop:
+		for _, idAttr := range p.Cfg.HostIdentifiers {
```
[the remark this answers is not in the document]
jcreixell: it should be a priority list, and I think we should also add `k8s.node.name` to the list, probably as first priority @rlankfo

modules/generator/processor/hostinfo/config.go
```diff
@@ -0,0 +1,21 @@
...
+	// HostIdentifiers defines the list of resource attributes used to derive
+	// a unique `grafana.host.id` value. In most cases, this should be [ "host.id" ]
+	HostIdentifiers []string      `yaml:"host_identifiers"`
+	StaleDuration   time.Duration `yaml:"stale_duration"`
+	// MetricName defines the name of the metric that will be generated
+	MetricName string `yaml:"metric_name"`
+}
+
+func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
+	cfg.HostIdentifiers = []string{"host.id"}
```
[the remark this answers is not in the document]
jcreixell: I think we do, it should be possible to override or even disable this logic entirely on a tenant per tenant basis (in fact, I would argue that the default state should be "disabled", and the setting should be managed externally via overrides API)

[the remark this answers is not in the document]
rlankfo: Yes, this is per tenant. My plan was to follow this up with another PR adding support in the user-configurable overrides API.

modules/generator/processor/hostinfo/processor.go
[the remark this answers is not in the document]
rlankfo: Correct, this should be a priority list. It would be ideal if the resource attributes were accessible as a map.

There should typically only be 2 or 3 possible attributes used for host identification though, so I don't think this will be too inefficient.

modules/generator/processor/hostinfo/processor.go
rlankfo: `SetForTargetInfo` will pass `updateIfAlreadyExist` as false. This isn't the behavior we want as we intend to reset the `lastUpdated` value stored on the gauge every time a host identifier is matched.

I could add `SetForHostInfo` but it would essentially be a clone of `Set`.

modules/generator/registry/gauge.go
rlankfo: Given that each metric type already had separate implementations to remove stale series, this seemed like it would have less overall impact than reducing stale duration across the board. The logic should remain the same unless `SetExpiration` is called explicitly on the gauge. 

If it's more desirable to lower stale duration for all metrics, I can remove this and we can look into the impact of that update.

modules/generator/processor/hostinfo/config.go
rlankfo: Thanks Joe, makes sense!

modules/generator/processor/hostinfo/processor.go
rlankfo: Thanks for the suggestion. I'll make this update. I haven't looked too closely, but I assume this method could be called concurrently? If so, I can add the necessary synchronization mechanisms for writing into the value slice.

modules/generator/processor/hostinfo/processor.go
joe-elliott: good catch. now i'm wondering why SetForTargetInfo exists? I feel like it might be a bug that it's not setting the lastUpdated field.

no need to make a `SetForHostInfo`. `Set` is fine.

modules/generator/processor/hostinfo/processor.go
joe-elliott: that's a good point. rather than adding concurrency mechanisms maybe just alloc these once at the top of the func?

i'm also no wondering if go is smart enough to just put this on the stack and this whole conversation is moot :)

modules/generator/processor/hostinfo/config.go
cyrille-leclerc, line 18: Did we consider adding other resource attributes like `k8s.node.uid`, `k8S.node.name`, or `host.name`?

modules/generator/registry/gauge.go
joe-elliott: yeah, i just don't like that all series are expired in the same way on the same cutoff time except for a few gauges that internally calc their own cutoff time based on time.Now()

this feels more like a business decision then a technical one. can we not do what we need to do with a standard stale duration? perhaps we should sync offline

modules/generator/processor/hostinfo/config.go
rlankfo: We can add `k8s.node.name` to this default list as well but this is a configurable property so we can define what we want at the time of usage.

modules/generator/processor/hostinfo/processor.go
rlankfo: unlikely but I made sure to check for this

modules/generator/processor/hostinfo/processor.go
rlankfo: sounds good! I added the labels string slice to the struct as that never changes. I alloc'd values slice at the top of the func.

modules/generator/registry/gauge.go
rlankfo: thanks for the feedback Joe. I removed the custom expiration logic

modules/generator/processor/hostinfo/processor.go
t00mas: I think now we should rename this to defaultHostInfoMetric and move it to config.go

modules/generator/processor/hostinfo/processor.go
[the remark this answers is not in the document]
t00mas: I see nothing wrong about this logic, but just to make it clearer I'd extract the function to find the first matching of the host identifier out. 

So something like

```
func (p *Processor) PushSpans(ctx context.Context, req *tempopb.PushSpansRequest) {
	_, span := tracer.Start(ctx, "hostinfo.PushSpans")
	defer span.End()

	values := make([]string, 1)

	for i := range req.Batches {
		resourceSpan := req.Batches[i]
		if hostID := p.findHostIdentifier(resourceSpan); hostID != "" {
			values[0] = hostID
			labelValues := p.registry.NewLabelValueCombo(
				p.labels,
				values,
			)
			p.gauge.Set(labelValues, 1)
		}
	}
}

func (p *Processor) findHostIdentifier(resourceSpan *v1.ResourceSpans) string {
	attrs := resourceSpan.GetResource().GetAttributes()

	for _, idAttr := range p.Cfg.HostIdentifiers {
		for _, attr := range attrs {
			if attr.GetKey() == idAttr {
				if val := attr.GetValue(); val != nil {
					if strVal := val.GetStringValue(); strVal != "" {
						return strVal
					}
				}
			}
		}
	}

	return ""
}
```

As a plus, you won't need the label break after this

t00mas (approved):

joe-elliott (approved):
lgtm. let's go!

PR #4718, opened 2025-02-17
fix trace:id search for ids less than 128 bit

### LINKED ISSUE #4437: Traces with 64 bit IDs cannot be queried by trace:id
**Describe the bug**
Queries such as `{ trace:id = "26592c9ca6385" }` return "0 series returned" even though the trace does exist.
This only happens when the trace id is 64 bit. When it is 128 bit, such as `{ trace:id = "fa4781118759fea1e7226976e0ef5c14" }`, the result is as expected.

**To Reproduce**
1. Open Grafana Cloud -- Explore a Tempo data source (tried on branch r178)
2. Query `{}`, find a short trace id and then query it with `{trace:id = "<traceid>"}`
3. It doesn't return anything.

**Expected behavior**
The trace is found and returned.

**Environment:**
 - Infrastructure: Kubernetes
 - Deployment tool: jsonnet

ie-pham:
**What this PR does**: 

`trace:id` queries were not working because of inconsistent trimming in bytes predicate caused by this [PR](https://github.com/grafana/tempo/pull/4198) 

Also force left pad `span:id`, `link:spanID`, `span:parentID` queries for IDs that are provided that are shorter than 8 bytes. 

Fixes #4437

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,7 @@ configurable via the throughput_bytes_slo field, and it will populate op="traces
 * [BUGFIX] Correctly cache frontend jobs for query range (TraceQL Metrics). [#4771](https://github.com/grafana/tempo/pull/4771) (@joe-elliott)
 * [BUGFIX] Various edge case fixes for query range (TraceQL Metrics) [#4962](https://github.com/grafana/tempo/pull/4962) (@ruslan-mikhailov)
 * [BUGFIX] Fix `TempoBlockListRisingQuickly` alert grouping. [#4876](https://github.com/grafana/tempo/pull/4876) (@mapno)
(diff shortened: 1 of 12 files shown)


mdisibio:
This seems like a case where it's not much code, but it's hard to keep know if everything is aligned (i.e. this mentions trace:id but the change involves link:spanid), let's add some tests in tempodb_search_test.go.

sam-mcbr:
Just curious, I saw some issues querying some 128 bit trace ids in our deployment and was wondering if these changes would fix those querying issues as well #4967?

joe-elliott:
@ie-pham what's happening with this PR?

PR #4728, opened 2025-02-21
Allow skipping spans for metric generation

flxbk:
**What this PR does**:

This PR makes it possible for distributor clients to indicate that the spans in a write request should not be considered for metrics generation (e.g. because metrics have been generated for these spans elsewhere already, and reprocessing them would lead to incorrect overall metric values).

The spans contained in requests marked this way will still be forwarded to generators so they can be ingested into the localblocks processor. The spanmetrics and service-graph processors will however be skipped.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/distributor/distributor.go
```diff
@@ -672,6 +686,12 @@ func (d *Distributor) sendToKafka(ctx context.Context, userID string, keys []uin
 	}, ring.DoBatchOptions{})
 }
 
+// shouldGenerateSpanMetrics determines whether a request's context indicates
+// that metrics should not be generated for the spans contained in the request.
+func shouldGenerateSpanMetrics(ctx context.Context) bool {
+	return len(metadata.ValueFromIncomingContext(ctx, NoGenerateMetricsContextKey)) == 0
```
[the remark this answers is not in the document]
yvrhdn: This makes me think we do but just wondering if you've tested it as well
https://github.com/grafana/tempo/blob/681c531b3519ed544dc8e234a3578166480fa183/modules/distributor/receiver/shim.go#L259

modules/generator/generator.go
```diff
@@ -38,6 +38,12 @@ const (
 	// We use a safe default instead of exposing to config option to the user
 	// in order to simplify the config.
 	ringNumTokens = 256
+
+	// NoGenerateMetricsContextKey is used in request contexts/headers to signal to
+	// the metrics generator that it should not generate metrics for the spans
+	// contained in the requests. This is intended to be used by clients that send
+	// requests for which span metrics have already been generated elsewhere.
+	NoGenerateMetricsContextKey = ingest.NoGenerateMetricsContextKey
```
yvrhdn, line 46: Out of curiosity, what's the reason for redefining `NoGenerateMetricsContextKey` in each pakage?

modules/generator/generator.go
```diff
@@ -38,6 +38,12 @@ const (
 	// We use a safe default instead of exposing to config option to the user
 	// in order to simplify the config.
 	ringNumTokens = 256
+
+	// NoGenerateMetricsContextKey is used in request contexts/headers to signal to
+	// the metrics generator that it should not generate metrics for the spans
+	// contained in the requests. This is intended to be used by clients that send
+	// requests for which span metrics have already been generated elsewhere.
```
yvrhdn, line 45: This will also skip generating service graph metrics:

```suggestion
	// contained in the requests. This is intended to be used by clients that send
	// requests for which span and service graph metrics have already been generated elsewhere.
```

modules/distributor/distributor.go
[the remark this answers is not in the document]
flxbk: Good point, I don't think it works out of the box, will take a closer look.

modules/generator/generator.go
flxbk: My main reason for this was to keep dependencies "logical", i.e. a client pushing data to the generator needs to import the generator package to figure out what special value to use when skipping data, etc. If you think it's better I'm happy to make them all import `ingest` directly instead.

modules/generator/generator.go
flxbk: I was thinking of "span metrics" as a general term for all metrics derived from spans (so including service graph metrics), but I see how this is confusing with the terminology that already exists. Will change.

modules/distributor/distributor.go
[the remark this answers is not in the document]
flxbk: I have changed the logic to account for the different receivers and added test coverage for this in `shim_test.go`

modules/generator/generator.go
flxbk: With the latest changes there are only two places left that define this key (but now independently), and only the generator package exports it (so clients can use that value by importing the package).

flxbk:
Rebasing to get CI to run (via #4751).

flxbk:
After discussion with @mdisibio I have switched to using the request body instead of the context to propagate this flag wherever possible.

The new approach will cause a diff-level conflict with #4686, I'll rebase on top of that PR once it's gone in.

flxbk:
I have rebased on top of #4686 and resolved conflicts.

mdisibio (approved):

PR #4743, opened 2025-02-24
rhythm: fail safe when no partitions are assigned to a blockbuilder instance

javiermolinar:
**What this PR does**:
Handle gracefully when a block-builder is initialized and it has no partitions assigned.

It also relies on the existing ring partition Ids to connect on instead of just looking on the number of active partitions

**Checklist**
- [X] Tests updated
- [ ] Documentation added


modules/blockbuilder/blockbuilder_test.go
```diff
@@ -78,6 +78,37 @@ func TestBlockbuilder_lookbackOnNoCommit(t *testing.T) {
...
+	require.NoError(t, services.StartAndAwaitRunning(ctx, b))
+	t.Cleanup(func() {
+		require.NoError(t, services.StopAndAwaitTerminated(ctx, b))
+	})
+
+	client := newKafkaClient(t, cfg.IngestStorageConfig.Kafka)
+	sendReq(t, ctx, client)
+
+	// Wait for record to be consumed and committed.
+	require.Eventually(t, func() bool {
```
mdisibio, line 107: I think this won't test things as expected, since the atomic starts at zero it would be satisfied immediately.

modules/blockbuilder/blockbuilder.go
```diff
@@ -201,6 +201,7 @@ func (b *BlockBuilder) running(ctx context.Context) error {
 		waitTime, err := b.consume(ctx)
 		if err != nil {
 			level.Error(b.logger).Log("msg", "consumeCycle failed", "err", err)
+			waitTime = b.cfg.ConsumeCycleDuration
```
mapno, line 204: Shouldn't `consume` decide how long to wait for? We may want to retry immediately.

mapno (changes requested):
`getAssignedActivePartitions` also needs to be fixed. That function assumes that if there are 5 instances in the partition ring, partitions `0`, `1`, `2`, `3` and `4` are the ones active-which is wrong, it can be any other existing partition in Kafka.

modules/blockbuilder/blockbuilder.go
javiermolinar: Well in this case I think that it makes sense to let the runner to decide what to do if there are no partitions

modules/blockbuilder/blockbuilder.go
```diff
@@ -472,14 +477,20 @@ func (b *BlockBuilder) pushTraces(ts time.Time, tenantBytes, reqBytes []byte, p
...
+// Gets assigned partitions, these can be active or inactive. Pending partitions won't be included
+func (b *BlockBuilder) getAssignedPartitions() []int32 {
+	partitions := b.partitionRing.PartitionRing().Partitions()
+	ringAssignedPartitions := make(map[int32]bool, len(partitions))
+	for _, p := range partitions {
+		if p.IsActive() || p.IsInactive() {
+			ringAssignedPartitions[p.Id] = true
+		}
+	}
+	assignedActivePartitions := make([]int32, 0, len(ringAssignedPartitions))
```
mapno, line 489: Nit: not sure this is worth it. The final length will be `1/N`, with `N` the number of generators

modules/blockbuilder/blockbuilder.go
mapno: But `consume` can fail for other reasons (eg. failed to commit). In such case, we probably want to retry immediately, not wait a full cycle.

javiermolinar: I see your point, I will change it

mapno (approved):

PR #3954, opened 2024-08-09
fix: correct end time when a trace is outside the ingestion slack range

javiermolinar:
**What this PR does**:
This PR addresses a bug where traces older than the ingestion slack are ingested. When that happened, the block start time exceeded the block end time. This led the compactor to remove those blocks, and the inability to search them by id.

**Example of the current error**
Trace data:

```json
{
          "traceId": "5B8EFFF798038103D269B633813FC703",
          "spanId": "EEE19B7EC3C1B100",
          "name": "I am a span!",
          "startTimeUnixNano": 1689969302000000000,
          "endTimeUnixNano": 1689970000000000000,
          "kind": 2,
          "status": {
            "code": "STATUS_CODE_OK"
          },
          "attributes": [
          {
              "key": "my.span.attr",
              "value": {
                  "stringValue": "some value"
              }
}
```
The start and date are from 2023

Resulting block:
```
window: 2023-07-21T20:00:00Z
start: 2024-08-07T17:21:44+02:00
end: 2023-07-21T22:06:40+02:00
duration:  -9187h15m4s
age: 9203h48m32s
```

The start date is being set as now() whereas the end date is left as it is in the span data.

**Example after the PR changes**
```json
{"format":"vParquet4","blockID":"be564c9d-f76d-49ee-9668-24b6cc024eed","minID":"W47/95gDgQPSabYzgT/HAw==","maxID":"W47/95gDgQPSabYzgT/HAw==","tenantID":"single-tenant","startTime":"2024-08-09T15:45:16.837762+02:00","endTime":"2024-08-09T15:47:18+02:00","totalObjects":2,"size":0,"compactionLevel":0,"encoding":"none","indexPageSize":0,"totalRecords":0,"dataEncoding":"","bloomShards":0,"footerSize":0}
```
Now, in the wall block, the start and end times are set to now() and there are no inconsistencies.

**Checklist**
- [X] Tests updated
- [ ] Documentation added
- [X] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #4746, opened 2025-02-25
rhythm: fix block time range adjustment

javiermolinar:
**What this PR does**:
This fixes the previous behavior, where the end date was always set to the start time. This was a translation of our behavior in the ingesters, where the start and end dates are set to now() if they are outside the time range.

 **How it works now**:
It checks independently the start and the end.
If the end is before the start it will set it to the start
If the end is after the end range it will set it to the end range

**Checklist**
- [X] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/blockbuilder/tenant_store.go
+++ b/modules/blockbuilder/tenant_store.go
@@ -157,6 +157,8 @@ func (s *tenantStore) Flush(ctx context.Context, store tempodb.Writer) error {
 	return nil
 }
 
+// Adjust the time range based on when the record was added to the partition, factoring in slack and cycle duration.
+// Any span with a start or end time outside this range will be constrained to the valid limits.
 func (s *tenantStore) adjustTimeRangeForSlack(start, end time.Time) (time.Time, time.Time) {
 	startOfRange := s.startTime.Add(-s.slackDuration)
 	endOfRange := s.startTime.Add(s.slackDuration + s.cycleDuration)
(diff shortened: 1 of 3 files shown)


modules/blockbuilder/tenant_store.go
```diff
@@ -157,18 +157,32 @@ func (s *tenantStore) Flush(ctx context.Context, store tempodb.Writer) error {
...
 		warn = true
-		end = s.startTime
+		start = endOfRange
+	}
+	// this can happen with old spans added to Tempo
+	// setting it to start to not jump forward unexpectedly
+	if end.Before(start) {
+		warn = true
+		end = start
+	}
```
mapno, line 181: This is just wrong data, no? I don't think we have to fix it, or at least it's not related to ingestion slack.

mapno: Did you mean `end.Before(startOfRange)` instead?

```suggestion
	if end.Before(startOfRange) {
		warn = true
		end = start
	}
```

javiermolinar: Not wrong data, old data. Imagine someone wants to add old spans into Tempo
I did a very similar fix a time ago:
https://github.com/grafana/tempo/pull/3954
We need to correct the end date or the block will be wrong

It's checked against the ``start`` which can be already corrected before. That way we ensure that the start is never greater than the end. If we check it against ``startOfRange`` we may miss an edge case where startDate < end < start

javiermolinar: using `start` instead of `startOfRange` makes sense to me because it addresses both scenarios. If the `start` is before the `startOfRange` it has been already adjusted. In the edge case where `end < start < startOfRange` it fixes it as well

mapno: >In the edge case where end < startOfRange < start it fixes it as well

This is what I meant by wrong data, `end < start` is logically incorrect.

mapno (approved):
LGTM. I think `end.Before(start)` is safe but feels a bit strange.

modules/blockbuilder/tenant_store.go
mdisibio: Internally during a test block builders flushed several blocks in a row with meta.start==meta.end.  I'm not sure that this change is the fix, because the only related change is the case `end after endOfRange`. Previously it was set to `start` and now it's `endOfRange`.  For this to be the issue it would have to mean that every trace in the block was a long-running one that exceeds the buffer `s.startTime.Add(s.slackDuration + s.cycleDuration)`.  I'd like to dig more and see if there is another explanation.

PR #4748, opened 2025-02-25
Convert throughput SLO metrics from histogram to counter

carles-grafana:
For performance

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


javiermolinar (commented):
This is a breaking change, what's the reason to convert it to a counter? We are losing the ability to calculate throughput  percentiles

carles-grafana:
> This is a breaking change, what's the reason to convert it to a counter? We are losing the ability to calculate throughput percentiles

it's expected to perform better, I have discussed this with Joe, and I don't think we or anyone else is actually using the buckets, but I'll add the changelog line

joe-elliott:
I don't believe the histogram gives us any value. The goal here is just to count the bytes processed per tenant.

In the case that we do need quantiles on this metric it can be calculated from logs.

joe-elliott (approved):

PR #4756, opened 2025-02-26
[rhythm] blockbuilder: fix cycle handling

mapno:
**What this PR does**:

Fix offset handling in BlockBuilder to prevent double consumption:

Align the return value of `consumePartition` with the committed offset (`lastRec.Offset + 1`). This ensures consistency between what's committed to Kafka and what's tracked in memory. This prevents records from being processed multiple times within the same consumption cycle.

Taken from https://github.com/grafana/tempo/pull/4750

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/blockbuilder/blockbuilder.go
+++ b/modules/blockbuilder/blockbuilder.go
@@ -375,9 +375,10 @@ outer:
 		return time.Time{}, -1, err
 	}
 
-	// TODO - Retry commit
-
-	resp, err := b.kadm.CommitOffsets(ctx, group, kadm.OffsetsFromRecords(*lastRec))
+	offset := kadm.NewOffsetFromRecord(lastRec)
+	offsets := make(kadm.Offsets)
+	offsets.Add(offset)
+	resp, err := b.kadm.CommitOffsets(ctx, group, offsets) // TODO - Retry commit
 	if err != nil {
 		return time.Time{}, -1, err
 	}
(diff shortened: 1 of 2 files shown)


modules/blockbuilder/blockbuilder.go
```diff
@@ -387,11 +387,11 @@ outer:
 	level.Info(b.logger).Log(
 		"msg", "successfully committed offset to kafka",
 		"partition", ps.partition,
-		"last_record", lastRec.Offset,
+		"commit_offset", lastRec.Offset+1,
```
javiermolinar, line 390: Should this be ` lastRec.Offset` ?. This is the committed one. A bit philosophical, nobody is going to inspect this log statement :P

javiermolinar (approved):
+1!!!!

modules/blockbuilder/blockbuilder.go
mapno: Commit is actually `lastRec.Offset+1`, that's the tricky part. Commit is the first unread message, rather than the last read message.

In terms of logging, not sure what's more relevant to the operator tbh.

javiermolinar: Cool I like the offset.At, it abstracts all of this!

mdisibio (approved):

PR #4765, opened 2025-02-27
[Rhythm] Block-builder metrics and code cleanup

mdisibio:
**What this PR does**:
Some small cleanup. This came about while writing tenantStore tests to verify handling of historical data, and decided to get a few more things:

* Block-builder move to the common `partition_lag_seconds` metric in pkg/ingest
* Clear old metrics when partitions are revoked
* Record lag at the end of the block builder loop so that we can see both the high and low-water marks, and fix to record zero lag when it knows the partition is caught up.
* Reorganize liveTracesIter and other consts

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/blockbuilder/tenant_store_test.go
+++ b/modules/blockbuilder/tenant_store_test.go
@@ -7,8 +7,7 @@ import (
 	"time"
 
 	"github.com/go-kit/log"
-	"github.com/grafana/tempo/pkg/tempopb"
-	v1 "github.com/grafana/tempo/pkg/tempopb/trace/v1"
+	"github.com/grafana/tempo/pkg/util/test"
 	"github.com/grafana/tempo/tempodb/backend"
 	"github.com/grafana/tempo/tempodb/encoding"
 	"github.com/grafana/tempo/tempodb/wal"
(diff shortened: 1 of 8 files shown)


modules/blockbuilder/tenant_store_test.go
```diff
@@ -92,3 +98,113 @@ func TestAdjustTimeRangeForSlack(t *testing.T) {
...
+
+			// Verify other properties of the block
+			require.EqualValues(t, count, meta.TotalObjects)
+			require.EqualValues(t, 1, meta.TotalRecords)
+			require.Greater(t, meta.Size_, uint64(0))
+		})
+	}
+}
+
+func writeHistoricalData(t *testing.T, count int, startTime time.Time, cycleDuration, slackDuration time.Duration, traceStart, traceEnd time.Time) *backend.BlockMeta {
```
javiermolinar, line 163: Maybe you can reuse this:

https://github.com/grafana/tempo/blob/1b998e9437031f82888f95025f586ad28b7f3a37/modules/blockbuilder/blockbuilder_test.go#L562

PR #4770, opened 2025-02-28
Begin backend scheduler/worker for consistent compaction

zalegrala:
**What this PR does**:

Here we add two new target modules `BackendScheduler` and `BackendWorker`.   At present, these modules work together to replace the function of the current `Compactor` module.  These modules are not included in any of the default targets and require enablement in order to function.

The `BackendScheduler` is responsible for scheduling compaction jobs, tracking their status and persisting that status to the backend for reload at startup.  The `BackendWorker` is responsible for picking up these jobs and executing them.  The `BackendWorker` is also responsible for writing the tenant index to the object storage.  Sharding for which tenants are owned by which worker is handled by the ring, which is copied from the current `Compactor` module.  Small duplication of effort during scaling events for the tenant index is a non-issue.

Tenant fairness is acheived with a new `tenantselector` package which aims to ensure that all tenants get compacted, but also not neglect any tenant for too long.  This is a simple approach which is not perfect, but should work for our use case.  With the current implementation, the blocklist, outstanding blocks, and the last compaction time are taken into account to determmine which tenants need to have jobs scheduled.  Eventually, if a tenant has not been compacted for a long time, it will become the priority no matter the lengh of the blocklist.

**Why we need it**:

During sclaing events of the compactors, the ring may not be fully propogated and there is a race between compactors that may already be executing a compaction job for a given block and new compactors just joining the ring.  This can lead to a small amount of duplicated data in the backend until the ring is fully propgated and stable.  For small environments, this may never surface as an issue, but in large environments that wish to autoscale their compactors with load, this can be problematic.  Additionally, we want to rely on RF1 data in the backend for future work.

Completed and failed jobs are dropped from the state after 1 hour.

**Known issues to follow up**:

The output blocks are not idempotent, since the desitntion block ID is not known until the compaction is complete.  This is a smallish change to the `encoding` package to include the target block ID in the output block.  This is not included in this PR, but we can follow up.

The worker does not wait to complete the in flight jobs before shutting down.  This is something I would like to resolve and is not included in this PR.

Tenant retention jobs are not included in this PR, but I expect this to be a fast follow with the current pattern.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/config.go
```diff
@@ -137,8 +138,34 @@ type CompactorConfig struct {
...
+	// 	CompactionCycle:         DefaultCompactionCycle,
+	// }
+
+	f.DurationVar(&cfg.BlockRetention, util.PrefixConfig(prefix, "compaction.block-retention"), 14*24*time.Hour, "Duration to keep blocks/traces.")
+	f.IntVar(&cfg.MaxCompactionObjects, util.PrefixConfig(prefix, "compaction.max-objects-per-block"), 6000000, "Maximum number of traces in a compacted block.")
+	f.Uint64Var(&cfg.MaxBlockBytes, util.PrefixConfig(prefix, "compaction.max-block-bytes"), 100*1024*1024*1024 /* 100GB */, "Maximum size of a compacted block.")
+	f.DurationVar(&cfg.MaxCompactionRange, util.PrefixConfig(prefix, "compaction.compaction-window"), time.Hour, "Maximum time window across which to compact blocks.")
+}
+
+func (cfg *CompactorConfig) validate() error {
```
zalegrala, line 167: Validating on the pointer receiver.

cmd/tempo/app/app.go
```diff
@@ -125,6 +129,8 @@ func (t *App) setupAuthMiddleware() {
 		noGRPCAuthOn := []string{
 			"/frontend.Frontend/Process",
 			"/frontend.Frontend/NotifyClientShutdown",
+			"/tempopb.BackendScheduler/Next",
```
zalegrala, line 132: This was insufficient to prevent this endpoint from requiring auth.  Currently, the workerID is passed as the tenant to satisfy the call requirements, but it is not used on the server.

cmd/tempo/app/modules.go
```diff
@@ -676,6 +680,41 @@ func (t *App) initCacheProvider() (services.Service, error) {
...
+
+	scheduler, err := backendscheduler.New(t.cfg.BackendScheduler, t.store, t.Overrides)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create backend scheduler: %w", err)
+	}
+
+	// Register the GRPC service
+	tempopb.RegisterBackendSchedulerServer(t.Server.GRPC(), scheduler)
+
+	t.Server.HTTPRouter().Path("/status/backendscheduler").HandlerFunc(scheduler.StatusHandler)
```
zalegrala, line 696:
```
+--------+-----------------------------------------------------------------------------+--------+------------------+-----------------------------------------------------------+-----------------------------------------------------------+----------------------------------------------------------+
| TENANT | BLOCKS                                                                      | STATUS | WORKER           | CREATED                                                   | START                                                     | END                                                      |
+--------+-----------------------------------------------------------------------------+--------+------------------+-----------------------------------------------------------+-----------------------------------------------------------+----------------------------------------------------------+
| ops    | [0d3144e4-14c2-498d-bf63-e3fa95e0dd44 7541231b-d6c2-4628-a282-e101efa1750e] |      2 | backend-worker-0 | 2025-03-06 18:45:58.314324489 +0000 UTC m=+1373.027027032 | 2025-03-06 18:45:58.563568518 +0000 UTC m=+1373.276271091 | 2025-03-06 18:45:58.57645014 +0000 UTC m=+1373.289152695 |
+--------+-----------------------------------------------------------------------------+--------+------------------+-----------------------------------------------------------+-----------------------------------------------------------+----------------------------------------------------------+
```

modules/backendscheduler/backendscheduler.go
```diff
@@ -0,0 +1,391 @@
...
+	store     storage.Store
+	overrides overrides.Interface
+
+	work *work.Queue
+}
+
+var _ JobProcessor = (*BackendScheduler)(nil)
+
+// Interface that for future API work to create and manage jobs.
+type JobProcessor interface {
```
joe-elliott, line 39: a lot of these aren't used. i gather this is for future use, but it would help the readability of this change a lot to get it down to only what is necessary to accomplish current goals

modules/backendscheduler/backendscheduler.go
```diff
@@ -0,0 +1,391 @@
...
+
+// BackendScheduler manages scheduling and execution of backend jobs
+type BackendScheduler struct {
+	services.Service
+
+	cfg       Config
+	store     storage.Store
+	overrides overrides.Interface
+
+	work *work.Queue
```
joe-elliott, line 33: these don't seem to be persisted anywhere. we need to do that in the initial impl

modules/backendscheduler/backendscheduler.go
```diff
@@ -0,0 +1,391 @@
...
+	metricJobsActive.WithLabelValues(j.JobDetail.Tenant, j.Type.String()).Dec()
+
+	level.Info(log.Logger).Log("msg", "failed job", "job_id", id, "tenant", j.JobDetail.Tenant, "type", j.Type.String())
+	return nil
+}
+
+// Next implements the BackendSchedulerServer interface
+func (s *BackendScheduler) Next(_ context.Context, req *tempopb.NextJobRequest) (*tempopb.NextJobResponse, error) {
+	// Find jobs that already exist for this worker
+	for id, j := range s.work.Jobs() {
```
joe-elliott, line 172: might be nicer to just make a func on s.work?  `JobForWorkerID(workerID)`

modules/backendscheduler/backendscheduler.go
```diff
@@ -0,0 +1,391 @@
...
+				}
+
+				level.Info(log.Logger).Log("msg", "assigned previous job to worker", "job_id", id, "worker", req.WorkerId)
+				return resp, nil
+			}
+		}
+	}
+
+	// Find next available job
+	for id, j := range s.work.Jobs() {
```
joe-elliott, line 188: similar thought here. this func would be super clear if it read:
```
func() {

j := jobForWorker()

if j == nil {
  j = jobForType()
}

...

}
```

modules/backendscheduler/backendscheduler.go
```diff
@@ -0,0 +1,391 @@
...
+		return nil, fmt.Errorf("job not found: %s", req.JobId)
+	}
+
+	switch req.Status {
+	case tempopb.JobStatus_JOB_STATUS_SUCCEEDED:
+		j.Complete()
+		metricJobsCompleted.WithLabelValues(j.JobDetail.Tenant, j.JobDetail.Tenant).Inc()
+		level.Info(log.Logger).Log("msg", "job completed", "job_id", req.JobId)
+
+		// TODO: update the blocklist?
```
joe-elliott, line 227: we may have to do this? pretty sure the current implementation does this to prevent constantly picking jobs that don't actually exist. good catch.

modules/backendscheduler/backendscheduler.go
```diff
@@ -0,0 +1,391 @@
...
+
+	// No jobs available
+	return &tempopb.NextJobResponse{}, nil
+}
+
+// UpdateJob implements the BackendSchedulerServer interface
+func (s *BackendScheduler) UpdateJob(_ context.Context, req *tempopb.UpdateJobStatusRequest) (*tempopb.UpdateJobStatusResponse, error) {
+	j := s.work.GetJob(req.JobId)
+	if j == nil {
+		return nil, fmt.Errorf("job not found: %s", req.JobId)
```
joe-elliott, line 218: should it return error or should it just log something? is the scheduler able to do something different if this fails?

also, if we do return error we should wrap it in a gRPC code. maybe INVALID_ARGUMENT? this would clearly tell the caller not to try again with the same value.

modules/backendscheduler/backendscheduler.go
```diff
@@ -0,0 +1,391 @@
...
+
+		// TODO: update the blocklist?
+
+	case tempopb.JobStatus_JOB_STATUS_FAILED:
+		j.Fail()
+		metricJobsFailed.WithLabelValues(j.JobDetail.Tenant, j.JobDetail.Tenant).Inc()
+		level.Error(log.Logger).Log("msg", "job failed", "job_id", req.JobId, "error", req.Error)
+
+	default:
+		return nil, fmt.Errorf("invalid job status: %v", req.Status)
```
joe-elliott, line 235: this should be an invalid arg as well

modules/backendscheduler/backendscheduler.go
```diff
@@ -0,0 +1,391 @@
...
+// ListJobs returns all jobs for a given tenant
+func (s *BackendScheduler) ListJobs(_ context.Context) []*work.Job {
+	return s.work.Jobs()
+}
+
+// CreateCompactionJob creates a new compaction job for the given tenant and blocks.
+// Must be called under jobsMtx lock.
+func (s *BackendScheduler) createCompactionJob(_ context.Context, tenantID string, input []string) error {
+	// Skip blocks which already have a job
+	for _, blockID := range input {
```
joe-elliott, line 252: similar thought here. the tracker could have a function: "JobExistsForBlock"? "Conflicts"?

encapsulating this functionality in the tracker will also make it a lot easier to change the tracker's internal methods of tracking jobs later down the road.

modules/backendscheduler/backendscheduler.go
```diff
@@ -0,0 +1,391 @@
...
+				}
+			}
+		}
+	}
+
+	jobID := uuid.New().String()
+
+	job := &work.Job{
+		ID:   jobID,
+		Type: tempopb.JobType_JOB_TYPE_COMPACTION,
```
joe-elliott, line 274: specify "pending" or whatever?

modules/backendscheduler/backendscheduler.go
```diff
@@ -0,0 +1,391 @@
...
+
+	"github.com/go-kit/log/level"
+	"github.com/google/uuid"
+	"github.com/grafana/dskit/services"
+	"github.com/grafana/tempo/modules/backendscheduler/work"
+	"github.com/grafana/tempo/modules/overrides"
+	"github.com/grafana/tempo/modules/storage"
+	"github.com/grafana/tempo/pkg/tempopb"
+	"github.com/grafana/tempo/pkg/util/log"
+	"github.com/grafana/tempo/tempodb/blockselector"
```
joe-elliott, line 19: +1 to this refactor

modules/backendscheduler/backendscheduler.go
```diff
@@ -0,0 +1,391 @@
...
+
+	x.AppendSeparator()
+
+	// x.SetOutputMirror(w)
+	w.WriteHeader(http.StatusOK)
+	// w.Header().Set("Content-Type", "plain/text")
+	_, _ = io.WriteString(w, x.Render())
+}
+
+func (s *BackendScheduler) compactions(ctx context.Context) []tempopb.JobDetail {
```
joe-elliott, line 317: reading this func and they way its used in above in ScheduleOnce we have to be careful. this is going to create A LOT of jobs and repeatedly throw them away. i'm also not convinced of the value of storing every pending compaction job in the job tracker struct. 

i think we either need to:

- generate the job on demand. so create a job when `Next` is called. this to me feels the cleanest
- find a way to store a smaller number of "todo" jobs.

I prefer the former b/c we need to persist the work in flight to object storage and we don't want to persist "pending" jobs. only the ones that are in flight.

modules/backendworker/backendworker.go
```diff
@@ -0,0 +1,254 @@
...
+// var tracer = otel.Tracer("modules/backendworker")
+
+func New(cfg Config, schedulerClientCfg backendscheduler_client.Config, store storage.Store, overrides overrides.Interface) (*BackendWorker, error) {
+	s := &BackendWorker{
+		cfg:       cfg,
+		store:     store,
+		overrides: overrides,
+	}
+
+	workerID, err := os.Hostname()
```
joe-elliott, line 47: log this as info?

modules/backendworker/backendworker.go
```diff
@@ -0,0 +1,254 @@
...
+	if err != nil {
+		return fmt.Errorf("error getting next job: %w", err)
+	}
+
+	if resp == nil || resp.JobId == "" {
+		return fmt.Errorf("no jobs available")
+	}
+
+	if resp.Detail.Tenant == "" {
+		return w.failJob(ctx, resp.JobId, "received job with empty tenant")
```
joe-elliott, line 114: it's a little odd to tell the scheduler a job is failed b/c it was a bad job, but i'm also not really sure what the best way to handle this edge case. it basically a suggests bug in the system.

i do think we need a metric for "asked for a job but didn't get one" that we alert on. i think "didn't get one" should be holistic and mean everything:

1) network call failed
2) "bad job" i.e. no tenant or other job invalid reasons
3) other crap

modules/backendworker/backendworker.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+func (w *BackendWorker) stopping(_ error) error {
+	return nil
+}
+
+func (w *BackendWorker) failJob(ctx context.Context, jobID string, errMsg string) error {
+	level.Error(log.Logger).Log("msg", "job failed", "job_id", jobID, "error", errMsg)
+
+	_, err := w.backendScheduler.UpdateJob(user.InjectOrgID(ctx, w.workerID), &tempopb.UpdateJobStatusRequest{
```
joe-elliott, line 157: i would make a wrapper func that you use for all uses of UpdateJob and Next that handles backoff. Something like:

```
func callSchedulerWithBackoff(ctx, func() err {
  w.backendScheduler.UpdateJob()
})
```

that way the same back off settings apply to all calls to the scheduler and all interactions between the worker and scheduler tolerate small outages in the same way.

it would use a loop like this one from mimir:  
https://github.com/grafana/mimir/blob/769906aa8f9d789759e783d972f0454238639e1f/pkg/usagestats/seed.go#L55

modules/backendworker/backendworker.go
```diff
@@ -0,0 +1,254 @@
...
+func (w *BackendWorker) starting(ctx context.Context) error {
+	if w.cfg.Poll {
+		w.store.EnablePolling(ctx, w)
+	}
+
+	return nil
+}
+
+func (w *BackendWorker) running(ctx context.Context) error {
+	// TODO: consider dropping the ticker and rely only on the backoff.
```
joe-elliott, line 72: see the comment below. personally i think you should rely on the backoff only but encapsulate it in a func suggested below. it feels fine to immediately reenter `processCompactionJobs` as soon as it returns but I wouldn't be opposed to ticker either with a very short timeout.

modules/backendworker/backendworker.go
```diff
@@ -0,0 +1,254 @@
...
+	s.workerID = workerID
+
+	schedulerClient, err := backendscheduler_client.New(cfg.BackendSchedulerAddr, schedulerClientCfg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create backend scheduler client: %w", err)
+	}
+	s.backendScheduler = schedulerClient
+
+	s.Service = services.NewBasicService(s.starting, s.running, s.stopping)
+	return s, nil
```
joe-elliott, line 60: should return an error in the event of bad config. i.e. 0s in backoff setting or something

pkg/tempopb/tempo.proto
```diff
@@ -46,6 +49,71 @@ service Metrics {
   rpc QueryRange(QueryRangeRequest) returns (QueryRangeResponse) {}
 }
 
+service BackendScheduler {
```
joe-elliott, line 52: maybe add to a new proto file in the same folder? scheduler.proto or something? wdyt?

tempodb/compactor.go
```diff
@@ -235,6 +236,10 @@ func (rw *readerWriter) compactWhileOwns(ctx context.Context, blockMetas []*back
 }
 
 func (rw *readerWriter) compactOneJob(ctx context.Context, blockMetas []*backend.BlockMeta, tenantID string) error {
+	return rw.CompactWithConfig(ctx, blockMetas, tenantID, rw.compactorCfg, rw.compactorSharder, rw.compactorOverrides)
+}
+
+func (rw *readerWriter) CompactWithConfig(ctx context.Context, blockMetas []*backend.BlockMeta, tenantID string, compactorCfg *CompactorConfig, compactorSharder CompactorSharder, compactorOverrides CompactorOverrides) error {
```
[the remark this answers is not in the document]
joe-elliott: I think this is a great compromise to keep the old compactor pattern working and support the new one

joe-elliott (commented):
Lot of comments but I think this is in on the right path. The major things I see missing:

- Tenant fairness
- Persisting jobs to object storage and reading them out
- Output block id on Job to pin the ID
- tenant index polling. do the scheduler do that and i'm missing it?
- scheduler expiration of jobs and marking complete jobs based on backend state

cmd/tempo/app/modules.go
```diff
@@ -676,6 +680,41 @@ func (t *App) initCacheProvider() (services.Service, error) {
...
+
+	t.backendScheduler = scheduler
+
+	return scheduler, nil
+}
+
+func (t *App) initBackendWorker() (services.Service, error) {
+	if t.cfg.Target == BackendWorker {
+		t.cfg.BackendWorker.Poll = true
```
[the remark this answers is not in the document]
zalegrala: Currently this is required because we want the blocklist.  The job only contains the UUID, but we need the meta for each of the blocks.
https://github.com/zalegrala/tempo/blob/55a63af1b12c4db8c66d209625f004f3a0678a1d/modules/backendworker/backendworker.go#L121
Alternatively we could update the job proto to include the blockmeta, but I think we want to put the tenant index writing on the workers, since there is more than 1 and they can share the load of keeping the tenant indexes up to date.

pkg/tempopb/tempo.proto
zalegrala: Simple enough, I've updated.

modules/backendworker/backendworker.go
zalegrala: Good call, I'll add some config validation.

modules/backendworker/backendworker.go
zalegrala: Yep, I like relying on just the backoff here.

zalegrala:
Thanks for the review.  I've made some updates to your comments.

* tenant fairness: there is a small approach which I think will work for our use case to ensure that all tenants get a little love, but not too much.
* job persistence: I've added caching to the backend at convent locations.
* output block id in the status: this is is implemented, but only possible right now on completed jobs.  In flight jobs don't have the target ID without more changes to the `encoding` for each version.  A smallish change, but perhaps a later PR.

  Things on my list currently:
* review "last compacted date" on priority queue
* add compactor ring to the worker and move tenant-index writing to the worker
* update the worker to wait for in flight jobs to complete before shutdown

cmd/tempo/app/config.go
```diff
@@ -39,23 +42,26 @@ type Config struct {
...
+	Generator             generator.Config               `yaml:"metrics_generator,omitempty"`
+	Ingest                ingest.Config                  `yaml:"ingest,omitempty"`
+	BlockBuilder          blockbuilder.Config            `yaml:"block_builder,omitempty"`
+	StorageConfig         storage.Config                 `yaml:"storage,omitempty"`
+	Overrides             overrides.Config               `yaml:"overrides,omitempty"`
+	MemberlistKV          memberlist.KVConfig            `yaml:"memberlist,omitempty"`
+	UsageReport           usagestats.Config              `yaml:"usage_report,omitempty"`
+	CacheProvider         cache.Config                   `yaml:"cache,omitempty"`
+	BackendScheduler      backendscheduler.Config        `yaml:"backend_scheduler,omitempty"`
+	BackenSchedulerClient backendscheduler_client.Config `yaml:"backend_scheduler_client,omitempty"`
```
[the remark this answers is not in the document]
joe-elliott: Fair. I'm good on the current location.

modules/backendscheduler/work/queue.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"errors"
+	"sort"
+	"sync"
+	"time"
+)
+
+var ErrJobAlreadyExists = errors.New("job already exists")
+
+type Queue struct {
```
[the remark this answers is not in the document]
zalegrala: This has been updated.  Let me know how it feels now.

modules/backendscheduler/backendscheduler.go
```diff
@@ -0,0 +1,391 @@
...
+	return nil
+}
+
+func (s *BackendScheduler) running(ctx context.Context) error {
+	level.Info(log.Logger).Log("msg", "backend scheduler running")
+
+	ticker := time.NewTicker(s.cfg.ScheduleInterval)
+	defer ticker.Stop()
+
+	if len(s.work.Jobs()) == 0 {
```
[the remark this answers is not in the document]
zalegrala: This has been adjusted.  We now have a min/max work depth setting, and only refill when we've dropped below the min.  This will allow work to go into the queue before it gets picked up and has a nice balance of not over/under queuing to allow a bit more fairness.

modules/backendscheduler/backendscheduler.go
zalegrala: This has been updated to flush and load the work from the backend.

modules/backendscheduler/backendscheduler.go
zalegrala: This has been updated.

cmd/tempo/app/app.go
zalegrala: I tried again and it worked.  :shrug:

modules/backendscheduler/backendscheduler.go
zalegrala: This has been updated and reads a little clearer now.

modules/backendscheduler/backendscheduler.go
zalegrala: I'm leveraging "UNSPECIFIED" for this case currently.  The worker will then report the status as RUNNING when it starts the work.

modules/backendworker/backendworker.go
zalegrala: I agree its a little strange.  Currently I don't see a way we end up with an empty tenant on the worker, but this was a little future proofing, since future code might push a job which has an empty tenant.  I've added a metric for this here, but open to other thoughts as we grow this functionality.

modules/backendscheduler/backendscheduler.go
zalegrala: I've dropped all that isn't used now.  Good call out.

modules/backendworker/backendworker.go
```diff
@@ -0,0 +1,533 @@
...
+
+	level.Info(log.Logger).Log("msg", "backend worker starting", "worker_id", w.workerID)
+
+	schedulerClient, err := backendscheduler_client.New(cfg.BackendSchedulerAddr, schedulerClientCfg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create backend scheduler client: %w", err)
+	}
+	w.backendScheduler = schedulerClient
+
+	if w.isSharded() {
```
joe-elliott, line 90: i really wish you could get rid of this logic. it makes this module seem so much more complicated than it is, but i think we need it as long as we do tenant index building outside the job system?

or can we do a ring of 1 and it just owns all tokens? i can't remember why we did this in the compactor tbh

modules/backendworker/backendworker.go
```diff
@@ -0,0 +1,533 @@
...
+	})
+
+	if resp == nil || resp.JobId == "" {
+		return fmt.Errorf("no jobs available")
+	}
+
+	metricWorkerJobsTotal.WithLabelValues().Inc()
+
+	if resp.Detail.Tenant == "" {
+		metricWorkerBadJobsRecieved.WithLabelValues("no_tenant").Inc()
```
joe-elliott, line 260: spelling: metricWorkerBadJobsRecieved -> metricWorkerBadJobsReceived

joe-elliott (commented):
nits and thoughts only at this point. out of band we agreed that job expiration was still to come and retention.

i'm ready to merge when you are.

modules/backendscheduler/work/work.go
```diff
@@ -0,0 +1,162 @@
...
+		switch j.GetStatus() {
+		case tempopb.JobStatus_JOB_STATUS_FAILED, tempopb.JobStatus_JOB_STATUS_SUCCEEDED:
+			// Keep the completed jobs around a while so as not to recreate them
+			// before the blocklist has been updated.
+			if time.Since(j.GetEndTime()) > q.cfg.PruneAge {
+				delete(q.Jobs, id)
+			}
+		case tempopb.JobStatus_JOB_STATUS_RUNNING:
+			// Fail jobs which have been running for too long
+			if time.Since(j.GetStartTime()) > q.cfg.DeadJobTimeout {
```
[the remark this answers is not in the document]
zalegrala: I think it got lost in the refactor.  I've pushed a commit for this.

modules/backendscheduler/backendscheduler.go
```diff
@@ -0,0 +1,421 @@
...
+	s.rpcMtx.Lock()
+	defer s.rpcMtx.Unlock()
+
+	j := s.work.GetJob(req.JobId)
+	if j == nil {
+		return &tempopb.UpdateJobStatusResponse{}, status.Error(codes.NotFound, work.ErrJobNotFound.Error())
+	}
+
+	switch req.Status {
+	case tempopb.JobStatus_JOB_STATUS_RUNNING:
```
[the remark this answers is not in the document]
zalegrala: Yeah, now that we're not queued, I agree.  Good call.

modules/backendscheduler/backendscheduler.go
```diff
@@ -0,0 +1,421 @@
...
+// BackendScheduler manages scheduling and execution of backend jobs
+type BackendScheduler struct {
+	services.Service
+
+	cfg       Config
+	store     storage.Store
+	overrides overrides.Interface
+
+	work    *work.Work
+	workMtx sync.RWMutex
```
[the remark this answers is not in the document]
zalegrala: I've dropped to 1 on the backendscheduler.  This leaves the work mutex and job with one each.

modules/backendscheduler/cache.go
```diff
@@ -0,0 +1,54 @@
...
+	"io"
+
+	jsoniter "github.com/json-iterator/go"
+
+	"github.com/go-kit/log/level"
+	"github.com/grafana/tempo/pkg/util/log"
+	"github.com/grafana/tempo/tempodb/backend"
+)
+
+func (s *BackendScheduler) flushWorkCache(ctx context.Context) error {
```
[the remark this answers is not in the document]
zalegrala: I've pushed a test for this.  Good call.

joe-elliott (approved):
let's go!

PR #4771, opened 2025-02-28
Fix query range frontend caching

joe-elliott:
**What this PR does**:
Cache keys were not being generated for query range backend jobs due to a mismatch between seconds and nanoseconds. Block metas are only included if their start/end times are completely encapsulated in the query start/end times. The start/ends were ns but the blocks were seconds which made it so no cache keys were ever generated.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


mattdurham (approved):
Lgtm

mdisibio (approved):

PR #4783, opened 2025-03-03
BugFix: Blocks whose end second matches a query range throw an erroneous error

joe-elliott:
**What this PR does**:
Currently if you perform a query range search and the end of the time range matches the end second of a block it will fail with:

`end must be greater than start`

This is due a consolidation of code when [adding the most_recent=true query hint](https://github.com/grafana/tempo/pull/4238). A bug was introduced when consolidating `blockMetasForSearch`. The query range pipeline respects a nanosecond range but blockMetasForSearch was not. As a result it was possible for a block to be incorrectly included in the search if it's end time shared the same second as the end time of the requested range.

- Applied the same fix to the cache key function.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/cache_keys.go
+++ b/modules/frontend/cache_keys.go
@@ -3,6 +3,7 @@ package frontend
 import (
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/grafana/tempo/tempodb/backend"
 )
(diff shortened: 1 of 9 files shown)


modules/frontend/search_sharder.go
```diff
@@ -136,7 +136,9 @@ func (s *asyncSearchSharder) backendRequests(ctx context.Context, tenantID strin
 		return
 	}
 
-	blocks := blockMetasForSearch(s.reader.BlockMetas(tenantID), start, end, backend.DefaultReplicationFactor)
+	startT := time.Unix(int64(start), 0)
```
javiermolinar, line 139: Maybe we could move this to `backendRange`
```go
func backendRange(start, end uint32, queryBackendAfter time.Duration) (time.Time, time.Time) 
```
It could also simplify the logic inside that method:

```go
if end.after(backendAfter)
```

modules/frontend/cache_keys.go
```diff
@@ -15,25 +16,29 @@ const (
...
 func searchJobCacheKey(tenant string, queryHash uint64, start int64, end int64, meta *backend.BlockMeta, startPage, pagesToSearch int) string {
-	return cacheKey(cacheKeyPrefixSearchJob, tenant, queryHash, start, end, meta, startPage, pagesToSearch)
+	startTime := time.Unix(start, 0)
+	endTime := time.Unix(end, 0)
+	return cacheKey(cacheKeyPrefixSearchJob, tenant, queryHash, startTime, endTime, meta, startPage, pagesToSearch)
 }
 
 func queryRangeCacheKey(tenant string, queryHash uint64, start int64, end int64, meta *backend.BlockMeta, startPage, pagesToSearch int) string {
-	return cacheKey(cacheKeyPrefixQueryRange, tenant, queryHash, start, end, meta, startPage, pagesToSearch)
+	startTime := time.Unix(0, start)
```
javiermolinar, line 25: How do we know we are handling ms here? Maybe we can rename the arguments to

startMs and endMs

Could we use time.Time on all the cache methods?

modules/frontend/cache_keys.go
```diff
@@ -15,25 +16,29 @@ const (
 )
 
 func searchJobCacheKey(tenant string, queryHash uint64, start int64, end int64, meta *backend.BlockMeta, startPage, pagesToSearch int) string {
-	return cacheKey(cacheKeyPrefixSearchJob, tenant, queryHash, start, end, meta, startPage, pagesToSearch)
+	startTime := time.Unix(start, 0)
```
javiermolinar, line 19: how do we know we are using seconds here? Mabe we can rename the arguments to

startSecods and endSeconds or smth similar

joe-elliott: pushed time.Time out one more level. i did the translation here b/c it seems like it has enough context to know if it's unix epoch seconds or nanos, but agree it's nicer to go one more level.

modules/frontend/cache_keys.go
joe-elliott: same as above

modules/frontend/search_sharder.go
joe-elliott: i'd rather leave `backendRange` alone. it's internally consistent b/c it only ever deals with epoch seconds as its owned by the search path. we could chase this time.Time change quite a ways, but i'd like to keep the changes small to address the issue.

javiermolinar (approved):

PR #4784, opened 2025-03-03
Move all intrinsic tag lookup to the query-frontend and prioritize them in the results

### LINKED ISSUE #4769: Search Tags endpoint doesn't return intrinsics when the response size limit is hit
**Describe the bug**
We have a default 1MB limit on the response size from tag lookups, /api/v2/search/tags.   When this limit is hit we stop adding new values to the response and return the first encountered tags.  However because we [add intrinsics last](https://github.com/grafana/tempo/blob/main/modules/ingester/instance_search.go#L324) the response is unlikely to include them, i.e. `duration`, `name`, etc.   This causes unexpected usability issues within Grafana because it no longer recognizes them as intrinsics and treats them like regular attributes, leading to faulty queries that will return no results.  Example `{ .duration > 1s }` (notice the dot which means an attribute of this name instead of the span field).   

**To Reproduce**

**Expected behavior**
We need to add intrinsics first to ensure they are always in the response.  We should also consider moving this code up a in the request pipeline to be performed at the query-frontend instead of the ingester and querier since these are static values.

**Environment:**

**Additional Context**

mdisibio:
**What this PR does**:
Fixes #4769 by ensuring intrinsic tags are added first to the value collectors.  Because they are static, there is also no need to return them from queriers and ingesters, and instead the query-frontend adds them statically.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/tag_handlers_test.go
+++ b/modules/frontend/tag_handlers_test.go
@@ -10,6 +10,7 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"regexp"
+	"sort"
 	"testing"
 	"time"
 
@@ -208,6 +209,12 @@ func TestSearchTagsV2AlwaysReturnsIntrinsics(t *testing.T) {
 
 	require.NoError(t, err)
 	require.Equal(t, 2, len(resp.Scopes))
+
+	// Sort scopes to give stable comparison
+	sort.Slice(resp.Scopes, func(i, j int) bool {
+		return resp.Scopes[i].Name < resp.Scopes[j].Name
+	})
+
 	require.Equal(t, api.ParamScopeIntrinsic, resp.Scopes[0].Name)
 	require.Equal(t, mockScope, resp.Scopes[1].Name)
 	require.ElementsMatch(t, search.GetVirtualIntrinsicValues(), resp.Scopes[0].Tags)
--- a/modules/frontend/tag_handlers_test.go
+++ b/modules/frontend/tag_handlers_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/gogo/status"
 	"github.com/gorilla/mux"
 	"github.com/grafana/dskit/user"
+	"github.com/grafana/tempo/modules/overrides"
 	"github.com/grafana/tempo/pkg/api"
 	"github.com/grafana/tempo/pkg/cache"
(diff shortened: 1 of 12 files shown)


modules/frontend/tag_handlers_test.go
```diff
@@ -167,6 +170,57 @@ func runnerTagValuesV2ClientCancelContext(t *testing.T, f *QueryFrontend) {
 	require.Equal(t, status.Error(codes.Canceled, "context canceled"), err)
 }
 
+func TestSearchTagsV2AlwaysReturnsIntrinsics(t *testing.T) {
```
javiermolinar, line 173: Does this test check that the intrinsics are returned even when we set low MaxBytesPerTagValuesQuery ?

javiermolinar (approved):
LGTM

modules/querier/querier.go
```diff
@@ -876,19 +871,10 @@ func (q *Querier) SearchBlock(ctx context.Context, req *tempopb.SearchBlockReque
 }
 
 func (q *Querier) internalTagsSearchBlockV2(ctx context.Context, req *tempopb.SearchTagsBlockRequest) (*tempopb.SearchTagsV2Response, error) {
-	// check if it's the special intrinsic scope
-	// note that every block search passes the same values up. this could be handled in the frontend and be far more efficient
+	// For the intrinsic scope there is nothing to do in the querier,
```
electron0zero, line 874: let's add a TODO that we can optimize this by never sending intrinsic only requests to querier and ingesters, and answer them directly from frontend

modules/frontend/tag_handlers_test.go
electron0zero: +1, I would like to a see a test that makes sure that we return intrinsics even when limit is hit, it would catch this bug if we introduce it again, and imo it would be very easy to introduce this bug again during a frontend refactor.

electron0zero (approved):
just two nit, rest lgtm ✔️

modules/frontend/tag_handlers_test.go
mdisibio: Updated the test to check both within or over the limit.

javiermolinar (approved):

PR #4787, opened 2025-03-04
fix: flaky TestWalBlockFindTraceByID test

javiermolinar:
**What this PR does**:
It fix a test with a non deterministic behaviour 

**Checklist**
- [X] Tests updated


carles-grafana (approved):

tempodb/encoding/vparquet4/wal_block_test.go
```diff
@@ -224,7 +224,7 @@ func TestWalBlockFindTraceByID(t *testing.T) {
 			require.NoError(t, err)
 			require.NotNil(t, found)
 			require.True(t, proto.Equal(trs[i], found.Trace))
-			require.Greater(t, found.Metrics.InspectedBytes, uint64(40000)) // approximate value
```
electron0zero, line 227: can you make it something like 100, or a low value so we test that InspectedBytes has a value?

electron0zero (approved):

PR #4802, opened 2025-03-07
TraceQL metrics tests

ruslan-mikhailov:
**What this PR does**:
Improves coverage of TraceQL metrics

Fixes #<issue number>

**Checklist**

- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/integration/e2e/api/query_range_test.go
+++ b/integration/e2e/api/query_range_test.go
@@ -55,6 +55,14 @@ sendLoop:
 	for _, query := range []string{
 		"{} | rate()",
 		"{} | compare({status=error})",
+		"{} | count_over_time()",
+		"{} | min_over_time(duration)",
+		"{} | max_over_time(duration)",
+		"{} | avg_over_time(duration)",
+		"{} | sum_over_time(duration)",
+		"{} | quantile_over_time(duration, .5)",
+		"{} | quantile_over_time(duration, .5, 0.9, 0.99)",
+		"{} | histogram_over_time(duration)",
 	} {
 		t.Run(query, func(t *testing.T) {
 			callQueryRange(t, tempo.Endpoint(tempoPort), query, debugMode)
--- a/integration/e2e/api/query_range_test.go
+++ b/integration/e2e/api/query_range_test.go
@@ -63,17 +63,47 @@ sendLoop:
 		"{} | quantile_over_time(duration, .5)",
 		"{} | quantile_over_time(duration, .5, 0.9, 0.99)",
 		"{} | histogram_over_time(duration)",
+		"{} | count_over_time() by (status)",
+		"{status != error} | count_over_time() by (status)",
 	} {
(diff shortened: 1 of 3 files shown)


integration/e2e/api/query_range_test.go
```diff
@@ -55,6 +55,14 @@ sendLoop:
 	for _, query := range []string{
 		"{} | rate()",
 		"{} | compare({status=error})",
+		"{} | count_over_time()",
```
javiermolinar, line 58: Can we include aggregations here?

`{span.http.status_code = 500} | count_over_time() by (resource.service.name)`

We should also test metrics with no matches

`{span.foo = bar} | count_over_time() by (resource.service.name)`

ruslan-mikhailov: \+

tempodb/tempodb_metrics_test.go
```diff
@@ -22,18 +26,31 @@ import (
...
-			End:   50 * uint64(time.Second),
-			Step:  15 * uint64(time.Second),
-			Query: "{ } | rate()",
-		},
-		expected: []*tempopb.TimeSeries{
+		name: "rate",
+		req:  requestWithDefaultRange("{ } | rate()"),
+		expectedL1: []*tempopb.TimeSeries{
 			{
 				PromLabels: `{__name__="rate"}`,
```
ruslan-mikhailov, line 55: This PR https://github.com/grafana/tempo/pull/4805 will affect PromLabels behavior. I will need to rebase later from master and fix accordingly

integration/e2e/api/query_range_test.go
```diff
@@ -55,17 +55,55 @@ sendLoop:
...
+		"{} | sum_over_time(duration)",
+		"{} | quantile_over_time(duration, .5)",
+		"{} | quantile_over_time(duration, .5, 0.9, 0.99)",
+		"{} | histogram_over_time(duration)",
+		"{} | count_over_time() by (status)",
+		"{status != error} | count_over_time() by (status)",
 	} {
 		t.Run(query, func(t *testing.T) {
-			callQueryRange(t, tempo.Endpoint(tempoPort), query, debugMode)
+			queryRangeRes := callQueryRange(t, tempo.Endpoint(tempoPort), query, debugMode)
```
javiermolinar, line 70: I think this test should be rewritten to a deterministic table-driven test asserting that the data returned is correct:

- MakeThriftBatch generates random data; this should be changed to something we can control
- We should ensure we have traces in the backend and in recent data to test the different aggregation layers
- We should test a comprehensive list of metrics with and without aggregation and check the correctnes of the returned series

joe-elliott (approved):
@mdisibio i'm good on this. it's more detailed than expected (queries at multiple aggregation layers).

you have any thoughts?

mdisibio (approved):
Very comprehensive, thanks for tackling this @ruslan-mikhailov.

PR #4805, opened 2025-03-08
chore: upgrade prometheus to version 3.1.0

javiermolinar:
**What this PR does**:
It upgrades prometheus to version 3.1.0

It fixes the following warning we get now on mod tidy:

```zsh
❯ go mod tidy
go: finding module for package github.com/prometheus/common/promlog
go: github.com/grafana/tempo/cmd/tempo/build imports
        github.com/prometheus/prometheus/web/api/v1 tested by
        github.com/prometheus/prometheus/web/api/v1.test imports
        github.com/prometheus/common/promlog: module github.com/prometheus/common@latest found (v0.62.0), but does not contain package github.com/prometheus/common/promlog
```

**Breaking changes**

Prometheus has deprecated go-kit/log, and  now we need to pass the built-in slog:
https://github.com/prometheus/prometheus/issues/14355

Now, the metric name labels are also quoted:
https://github.com/prometheus/prometheus/pull/15531

```json
`{"span.foo"="baz"}`
```
This can be seen in the `promLabels`, but it doesn't affect the displayed label
```json
 "promLabels": "{\"resource.service.name\"=\"article-service\"}",
    {
      "labels": [
        {
          "key": "resource.service.name",
          "value": {
            "stringValue": "auth-service"
          }
        }
      ],
```

**Checklist**
- [X] Tests updated


pkg/traceql/engine_metrics_test.go
```diff
@@ -364,7 +364,7 @@ func TestQuantileOverTime(t *testing.T) {
 				0,
 			},
 		},
-		`{p="0.5", span.foo="bar"}`: TimeSeries{
+		`{p="0.5", "span.foo"="bar"}`: TimeSeries{
```
javiermolinar, line 367: We can still keep the name label unquoted. The simplest way to do it would be to use a modified label version.Labels.String() without quoting the name

https://github.com/prometheus/prometheus/blob/e32d89af7fb8baee53770de74793e8b2fb828f52/model/labels/labels_common.go#L43-L65

modules/generator/storage/instance.go
```diff
@@ -50,26 +50,29 @@ type storageImpl struct {
...
 
 var _ Storage = (*storageImpl)(nil)
 
 // New creates a metrics WAL that remote writes its data.
-func New(cfg *Config, o Overrides, tenant string, reg prometheus.Registerer, logger log.Logger) (Storage, error) {
-	logger = log.With(logger, "tenant", tenant)
+// TODO the passed logger does not include any other context attribute
+// Should we standarize slog and deprecate go-kit/log too?
+func New(cfg *Config, o Overrides, tenant string, reg prometheus.Registerer, _ log.Logger) (Storage, error) {
+	logger := slog.New(slog.NewTextHandler(os.Stdout, nil)).With("tenant", tenant)
```
mapno, line 62: With this we lose some logger functionality, like adding the logging line.

This is how Thanos has adapted go-kit/log to slog, maybe we can do the same: https://github.com/SungJin1212/thanos/blob/c280c664de7e7ba591ca0f46908b80a5039b1bc3/pkg/logutil/logutil.go

javiermolinar: Not sure if adding a new dependency to deprecate an existing one is a wise choice 

`sloggk "github.com/tjhop/slog-gokit"`

javiermolinar: I can do a similar method to get the log level from the go-kit/log and include the contextual information for this PR. Eventually, we should pass the logger from upstream code

mapno: >Not sure if adding a new dependency to deprecate an existing one is a wise choice

We're not deprecating one right now, no?

javiermolinar: I've added the ability to include the source (line of code and method name). I think it's ok for this PR

mapno (approved):
LGTM

PR #4830, opened 2025-03-11
[rhythm] Add partition lag alerts for Kafka consumers

mapno:
**What this PR does**:

This PR adds new alerts to monitor Kafka partition lag in consumer groups. It introduces two alert levels:
* Warning alert when a partition lags by more than 5 minutes
* Critical alert when a partition lags by more than 15 minutes

### DIFF
--- a/operations/tempo-mixin/config.libsonnet
+++ b/operations/tempo-mixin/config.libsonnet
@@ -28,6 +28,8 @@
       // Partition lag thresholds in seconds
       partition_lag_warning_seconds: 300,  // 5 minutes
       partition_lag_critical_seconds: 900,  // 15 minutes
+      // Filter for consumer groups to monitor for partition lag
+      partition_lag_group_filter: 'metrics-generator|block-builder',
     },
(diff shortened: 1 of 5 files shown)


electron0zero (approved):

operations/tempo-mixin/config.libsonnet
```diff
@@ -25,6 +25,9 @@
       p99_request_exclude_regex: 'metrics|/frontend.Frontend/Process|debug_pprof',
       outstanding_blocks_warning: 100,
       outstanding_blocks_critical: 250,
+      // Partition lag thresholds in seconds
```
mdisibio, line 28: Can we add a setting to filter the groups?  For the case that there are other consumer groups running components in the same namespace, but unrelated.  i.e. Maybe a regex like `partition_lag_group_filter: "(metrics-generator|block-builder)"`

mapno: Done!

PR #4837, opened 2025-03-12
rhythm: implement MaxBytesPerCycle to prevent excessive memory usage

javiermolinar:
**What this PR does**:
Implements a mechanism to ensure that the processing of records is bounded by the `maxBytesPerCycle` limit, preventing excessive memory usage or processing time in a single cycle. Configured by default to 50 MB

**Checklist**
- [X] Tests updated
- [ ] Documentation added


modules/blockbuilder/blockbuilder.go
```diff
@@ -359,12 +361,22 @@ outer:
 				break outer
 			}
 
+			if maxBytesPerCycle > 0 && consumedBytes+recordSizeBytes >= maxBytesPerCycle {
```
mdisibio, line 364: What do you think about moving this check after pushTraces? That means it is guaranteed to always allow at least 1 record.   I am thinking about the case where the max is set to a value less than the size of the next record, the check would prevent the BB from ever making progress.

javiermolinar: Makes sense to me :)

modules/blockbuilder/config.go
```diff
@@ -74,6 +75,7 @@ func (c *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet) {
 	f.StringVar(&c.InstanceID, "block-builder.instance-id", hostname, "Instance id.")
 	f.Var(newPartitionAssignmentVar(&c.AssignedPartitions), prefix+".assigned-partitions", "List of partitions assigned to this block builder.")
 	f.DurationVar(&c.ConsumeCycleDuration, prefix+".consume-cycle-duration", 5*time.Minute, "Interval between consumption cycles.")
+	f.Uint64Var(&c.MaxBytesPerCycle, prefix+".max-bytes-per-cycle", 50e6, "Maximum number of bytes that can be consumed in a single cycle. 0 to disable")
```
mdisibio, line 78: I think this is too low for a default, I was thinking more like 5GB.  Enough to prevent a run away but not be triggered in most cases.

mdisibio (approved):

PR #4852, opened 2025-03-13
Pass context to the poller.Do method for clean shutdown in tests

zalegrala:
**What this PR does**:

Sometimes in tests, I want to populate the `local` backend and poll it as a dependency for whatever I'm testing.  I occasionally see strange behavior when the go routine is still running when the test directory is cleaning up after a test.  To address this situation, we can pass a context into the `poller.Do` to allow for clean shutdown.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/integration/poller/poller_test.go
+++ b/integration/poller/poller_test.go
@@ -177,8 +177,11 @@ func TestPollerOwnership(t *testing.T) {
 					require.Equal(t, 0, len(cmResults))
 				}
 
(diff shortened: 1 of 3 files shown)


joe-elliott (approved):
I don't believe the local backend honors context cancellation, but I think this is a good change regardless.

Nice one

PR #4853, opened 2025-03-14
WIP: generator registry stale markers

rlankfo:
**What this PR does**:
This PR adds stale marker support to the metrics generator registry. A series is marked stale when it's stale duration has elapsed. During the next collection interval, any series marked stale will append a stale marker value before removal from the metric's series map. 

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/registry/counter.go
+++ b/modules/generator/registry/counter.go
@@ -33,6 +33,7 @@ type counterSeries struct {
 	// to the desired value.  This avoids Prometheus throwing away the first
 	// value in the series, due to the transition from null -> x.
 	firstSeries *atomic.Bool
+	stale       *atomic.Bool
 }
 
 var (
@@ -119,12 +120,14 @@ func (c *counter) newSeries(labelValueCombo *LabelValueCombo, value float64) *co
 		value:       atomic.NewFloat64(value),
 		lastUpdated: atomic.NewInt64(time.Now().UnixMilli()),
 		firstSeries: atomic.NewBool(true),
+		stale:       atomic.NewBool(false),
 	}
 }
 
 func (c *counter) updateSeries(s *counterSeries, value float64) {
 	s.value.Add(value)
 	s.lastUpdated.Store(time.Now().UnixMilli())
+	s.stale.Store(false)
 }
 
 func (c *counter) name() string {
@@ -133,11 +136,11 @@ func (c *counter) name() string {
 
 func (c *counter) collectMetrics(appender storage.Appender, timeMs int64) (activeSeries int, err error) {
 	c.seriesMtx.RLock()
-	defer c.seriesMtx.RUnlock()
 
 	activeSeries = len(c.series)
+	staleSeries := []uint64{}
 
-	for _, s := range c.series {
+	for hash, s := range c.series {
 		// If we are about to call Append for the first time on a series, we need
 		// to first insert a 0 value to allow Prometheus to start from a non-null
 		// value.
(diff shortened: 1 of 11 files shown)


modules/generator/registry/config.go
```diff
@@ -15,6 +15,9 @@ type Config struct {
 	// Defaults to 15m.
 	StaleDuration time.Duration `yaml:"stale_duration"`
 
+	// EnableStaleMarkers controls whether stale markers should be sent.
+	EnableStaleMarkers bool `yaml:"enable_stale_markers"`
```
joe-elliott, line 19: is there any reason to make this configurable? if this is correct behavior why don't we just do it.

going to summon @mattdurham who likely knows far more about staleness markers and when/if they are worth sending than myself.

mattdurham: Perhaps an ignorant question, are the same series being generated from the same instance always? Side question out of the way, to my knowledge its always 5m.

rlankfo: > is there any reason to make this configurable? if this is correct behavior why don't we just do it.
I didn't want to automatically change default behavior w/o providing a way to change it back. No real reason here besides "just in case".  I can get rid of this config.

> are the same series being generated from the same instance always?
The same series might be generated by different instances however they will always be differentiated by the `__metrics_gen_instance` label, so mimir ingesters should see them as different series iiuc.

mattdurham: For what its worth Alloy generally hard codes 5m, otel makes it configurable. For max compatibility configuration is reasonable but a comment that 5m is the accepted default that is almost never changed might go a long way.

rlankfo: One thing to note here is that there are default adaptive metrics rules in place to aggregate series on the `__metrics_gen_instance` label and I'm not sure what impact that might have.

rlankfo: Thanks @mattdurham. Setting the `stale_duration` to 5m makes sense to me. The probably is configurable. 

Do you think it makes sense to make sending stale markers configurable or rather we should just always send them when cleaning up stale series?

mattdurham: Off the top of my head I cant think of a downside to sending staleness marker.

modules/generator/registry/gauge.go
```diff
@@ -166,8 +180,12 @@ func (g *gauge) removeStaleSeries(staleTimeMs int64) {
 
 	for hash, s := range g.series {
 		if s.lastUpdated.Load() < staleTimeMs {
```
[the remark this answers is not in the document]
joe-elliott: wups. i'm definitely wrong here. misread it as a "lastUpdated" at the metric level and not the series level.

modules/generator/registry/gauge.go
```diff
@@ -148,8 +151,19 @@ func (g *gauge) collectMetrics(appender storage.Appender, timeMs int64) (activeS
 
 	activeSeries = len(g.series)
 
-	for _, s := range g.series {
+	for hash, s := range g.series {
 		t := time.UnixMilli(timeMs)
+
+		if g.enableStaleMarkers && s.stale {
```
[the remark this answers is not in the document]
rlankfo: yes, definitely. this makes a lot of sense. thanks Joe!

modules/generator/registry/counter.go
```diff
@@ -147,30 +150,55 @@ func (c *counter) collectMetrics(appender storage.Appender, timeMs int64) (activ
 			endOfLastMinuteMs := getEndOfLastMinuteMs(timeMs)
 			_, err = appender.Append(0, s.labels, endOfLastMinuteMs, 0)
 			if err != nil {
+				c.seriesMtx.RUnlock()
```
mattdurham, line 153: Do you think we can avoid this manual unlocking? Makes it very easy to have a danging deadlock.

mattdurham:
Poked at this a bit on thursday. Histograms make consolidating to one locking pattern hard. Going to look to ensure histograms are handled and fix the unit test failures.

mattdurham:
I created https://github.com/grafana/tempo/pull/4933 but I think it is to large to pull into this. I think  the only thing this is missing is Native Histograms use the Sum field for staleness markers, which makes it considerably easier.

image

PR #4854, opened 2025-03-14
add host info to userconfigurable overrides api

rlankfo:
**What this PR does**:
This PR adds support for the host info processor to the user configurable overrides api.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/sources/tempo/operations/manage-advanced-systems/user-configurable-overrides.md
+++ b/docs/sources/tempo/operations/manage-advanced-systems/user-configurable-overrides.md
@@ -81,6 +81,10 @@ metrics_generator:
       ]
       [enable_target_info: <bool>]
       [target_info_excluded_dimensions: <list of string>]
+
+    host_info:
+      [metric_name: <string>]
+      [host_identifiers: <list of string>]
 ```
 
 ## API
@@ -98,12 +102,13 @@ If the tenant is run in distributed mode, only the query-frontend will accept AP
 Returns the current overrides and it's version.
 
 Query-parameters:
+
 - `scope`: whether to return overrides from the API only `api` or merge it with the runtime overrides `merged`. Defaults to `api`.
 
 Example:
 
 ```shell
-$ curl -X GET -v -H "X-Scope-OrgID: 3" http://localhost:3100/tempo/api/overrides\?scope=merged`
+curl -X GET -v -H "X-Scope-OrgID: 3" http://localhost:3100/tempo/api/overrides\?scope=merged`
 ```
 
 #### POST /api/overrides
@@ -156,13 +161,13 @@ $ curl -v http://localhost:3100/api/overrides
 Requests that modify or delete overrides need to pass the current version using the `If-Match` header:
 
 ```shell
-$ curl -X POST -H "If-Match: 1697726795401423" http://localhost:3100/api/overrides --data "..."
+curl -X POST -H "If-Match: 1697726795401423" http://localhost:3100/api/overrides --data "..."
 ```
 
 This example uses overrides in the `overrides.json` file with the location in `pwd`:
 
 ```shell
-$ curl -X POST -H "X-Scope-OrgID: 3" -H "If-Match: 1697726795401423" http://localhost:3100/api/overrides --data @overrides.json
+curl -X POST -H "X-Scope-OrgID: 3" -H "If-Match: 1697726795401423" http://localhost:3100/api/overrides --data @overrides.json
 ```
 
 If the version doesn't match the version in the backend, the request is rejected with HTTP error 412.
@@ -190,5 +195,5 @@ overrides:
 You can bypass this check by setting the query parameter `skip-conflicting-overrides-check=true`:
 
 ```shell
-$ curl -X POST -H "If-Match: 1697726795401423" http://localhost:3100/api/overrides?skip-conflicting-overrides-check=true --data "..."
+curl -X POST -H "If-Match: 1697726795401423" http://localhost:3100/api/overrides?skip-conflicting-overrides-check=true --data "..."
 ```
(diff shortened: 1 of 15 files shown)


modules/overrides/userconfigurable/api/api_test.go
```diff
@@ -62,7 +62,7 @@ func Test_UserConfigOverridesAPI_overridesHandlers(t *testing.T) {
 			name:           "GET",
 			handler:        overridesAPI.GetHandler,
 			req:            prepareRequest(tenant, "GET", nil),
-			expResp:        `{"forwarders":["my-other-forwarder"],"cost_attribution":{},"metrics_generator":{"processor":{"service_graphs":{},"span_metrics":{}}}}`,
```
jcreixell, line 65: we can add a test here, set some pre-defined values in `overridesAPI` and extend the postJSON payload to include  different values. I was particularly interested on whether setting an empty list of host identifiers and an empty metric name would overrides previous values. Seems like it does but having test coverage would be nice. (I was interested in that particular case because it would allow us to disable host info metric generation when app o11y is disabled).

jcreixell (commented):
added a small comment but overall LGTM

modules/overrides/config.go
```diff
@@ -114,12 +114,16 @@ type LocalBlocksOverrides struct {
 	CompleteBlockTimeout time.Duration `yaml:"complete_block_timeout,omitempty" json:"complete_block_timeout,omitempty"`
 }
 
+type HostInfoOverrides struct {
+	HostIdentifiers []string `yaml:"host_identifiers,omitempty" json:"host_identifies,omitempty"`
+	MetricName      string   `yaml:"metric_name,omitempty" json:"metric_name,omitempty"`
```
joe-elliott, line 119: does this need to be per tenant configurable? user overrides are already quite complex and it would be better to only put in them what is necessary

rlankfo: it would be ideal if we could configure this per tenant re: @cedricziel's comment on the processor PR here: https://github.com/grafana/tempo/pull/4698#discussion_r1955904886

joe-elliott (commented):
do we need to add code here so that the host info processor reloads correctly?

https://github.com/grafana/tempo/blob/main/modules/generator/instance.go#L277

here as well?

https://github.com/grafana/tempo/blob/main/modules/generator/config.go#L157

modules/generator/instance_test.go
```diff
@@ -203,6 +204,41 @@ func Test_instance_updateProcessors(t *testing.T) {
...
+	})
+
+	t.Run("replace hostinfo processor", func(t *testing.T) {
+		overrides.processors = map[string]struct{}{
+			servicegraphs.Name: {},
+			spanmetrics.Name:   {},
+			hostinfo.Name:      {},
+		}
+		overrides.hostInfoHostIdentifiers = []string{"host.id"}
+		overrides.hostInfoMetricName = "traces_host_info"
```
t00mas, line 229: (nit because I've tested and still works) but this test uses the defaults and it should be using non-defaults just in case

t00mas (approved):

joe-elliott (commented):
one concern, but other than that i think this is ready to merge

modules/generator/config.go
```diff
@@ -223,6 +223,10 @@ func (cfg *ProcessorConfig) copyWithOverrides(o metricsGeneratorOverrides, userI
 
 	copyCfg.ServiceGraphs.EnableVirtualNodeLabel = o.MetricsGeneratorProcessorServiceGraphsEnableVirtualNodeLabel(userID)
 
+	copyCfg.HostInfo.HostIdentifiers = o.MetricsGeneratorProcessorHostInfoHostIdentifiers(userID)
```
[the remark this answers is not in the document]
rlankfo: Thanks Joe. I added the checks here similar to other overrides I saw in this method. Let me know if this is what you had in mind

knylander-grafana:
Should we add host info to the doc for the [user-configurable overrides](https://grafana.com/docs/tempo/latest/operations/manage-advanced-systems/user-configurable-overrides/)?

joe-elliott (approved):
LGTM! Thanks Robbie.

Will wait on @knylander-grafana 's doc request before merging.

rlankfo:
> Should we add host info to the doc for the [user-configurable overrides](https://grafana.com/docs/tempo/latest/operations/manage-advanced-systems/user-configurable-overrides/)?

updated the docs -- my editor automatically removed the `$` shell indicator in the shell blocks. I didn't add it back as I noticed some blacks had it, but others did not. It makes sense to me to remove it as when you click copy you wouldn't want this character included.

docs/sources/tempo/operations/manage-advanced-systems/user-configurable-overrides.md
```diff
@@ -98,12 +102,13 @@ If the tenant is run in distributed mode, only the query-frontend will accept AP
...
 
 Query-parameters:
+
 - `scope`: whether to return overrides from the API only `api` or merge it with the runtime overrides `merged`. Defaults to `api`.
 
 Example:
 
 ```shell
-$ curl -X GET -v -H "X-Scope-OrgID: 3" http://localhost:3100/tempo/api/overrides\?scope=merged`
+curl -X GET -v -H "X-Scope-OrgID: 3" http://localhost:3100/tempo/api/overrides\?scope=merged`
```
knylander-grafana, line 111: Is the ` at the end of the example supposed to be there?

knylander-grafana (approved):
Thank you for updating the doc!

docs/sources/tempo/operations/manage-advanced-systems/user-configurable-overrides.md
rlankfo: good catch! I don't think this should be there.

rlankfo:
@joe-elliott are we good to merge?

joe-elliott:
yup

PR #4855, opened 2025-03-14
TraceQL: fix queries like {.foo && true} and {.foo || false}

### LINKED ISSUE #4839: TraceQL fails on &&ing attributes to true
When &&ing or ||ing a single attribute with a boolean value Tempo fails with the following error:

```
conditions invalid: unknown operation. condition: {Attribute:span.foo Op:&& Operands:[true] CallBack:<nil>} 
```

Image

stoewer:
**What this PR does**:
Fix behavior for queries single attributes that are &&ed or ||ed with a boolean.

Fixes #4839

**Checklist**
- [x] Tests updated
- [ ] Documentation added


javiermolinar (approved):
LGTM

PR #4857, opened 2025-03-14
[Rhythm] Full RF1 read path

mapno:
**What this PR does**:

Updates https://github.com/grafana/tempo/pull/4478 to actual state of `main`. Original text:

These are the changes needed to migrate the backend read path to RF1.  New config options `rf1_after` which is a timestamp, after that only blocks with rf==1 are included in searches and trace by ID lookups.  And an option to discontinue flushes to object storage from the ingesters.

How each api is handled:

Search: the frontend determines the blocks:
```yaml
query_frontend:
  search:
    rf1_after: "2024-12-18T00:00:00Z"
```

Trace lookup: the querier determines the blocks:
```yaml
querier:
  trace_by_id:
    rf1_after: "2024-12-18T00:00:00Z"
```

Tags: this is actually not filtering on replication factor, so no changes needed
Metrics:  already limited to rf1 blocks

The migration path:
* Enable kafka ingest, block-builder, etc. It is now flushing RF1 blocks.
* Grace period for monitoring and validation
* Set `rf1_after` to move the read path over to the new RF1 blocks (recent searches/tags/tracelookups are still directed at ingesters)
* Grace period for monitoring and validation
* Disable flushes on ingesters

Note:  Global config is non-trivial due to Tempo's modular setup, so this adds redundant config in multiple places. Not ideal.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/querier/config.go
```diff
@@ -31,6 +31,7 @@ type SearchConfig struct {
 
 type TraceByIDConfig struct {
 	QueryTimeout time.Duration `yaml:"query_timeout"`
+	RF1After     time.Time     `yaml:"rf1_after"`
```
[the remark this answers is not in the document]
mapno: Now `rf1After` is passed via query params to the querier.

modules/frontend/tag_sharder.go
```diff
@@ -257,7 +257,9 @@ func (s searchTagSharder) backendRequests(ctx context.Context, tenantID string,
 	// get block metadata of blocks in start, end duration
 	startT := time.Unix(int64(start), 0)
 	endT := time.Unix(int64(end), 0)
-	blocks := blockMetasForSearch(s.reader.BlockMetas(tenantID), startT, endT, backend.DefaultReplicationFactor)
+	blocks := blockMetasForSearch(s.reader.BlockMetas(tenantID), startT, endT, func(m *backend.BlockMeta) bool {
+		return m.ReplicationFactor == backend.DefaultReplicationFactor
```
javiermolinar, line 261: Can we not use the same logic here as in the search sharder?

mapno: Actually, we need to. Otherwise these requests will fail when we stop flushing ingester blocks. Great catch!

modules/frontend/traceid_handlers.go
```diff
@@ -43,7 +43,7 @@ func newTraceIDHandler(cfg Config, next pipeline.AsyncRoundTripper[combiner.Pipe
 		}
 
 		// validate start and end parameter
-		_, _, _, _, _, reqErr := api.ValidateAndSanitizeRequest(req)
+		_, _, _, _, _, _, reqErr := api.ValidateAndSanitizeRequest(req)
```
mdisibio, line 46: This function signature is getting lengthy.  Do we want to introduce a TraceByIDRequest object and use it as the output from `api.ParseTraceID(req)` ?

mapno: Yes, I thought of that as well. What didn't convince me was that it also contains extra params `start` and `end` that are not part of `TraceByIDRequest` and it requires extra changes.

I don't have a strong opinion, but initially favored introducing only the necessary changes.

javiermolinar (approved):

PR #4864, opened 2025-03-17
TraceQL: Fix lookups .attr != nil with special columns

stoewer:
**What this PR does:**  
Improves attribute existence checks using `{ .attr != nil }`. Queries involving well-known attributes and attributes stored in dedicated columns currently don't check generic attribute columns for attributes with the same name but a different type.  

For example, the query `{ span.http.status_code != nil }` will not match spans where `http.status_code` exists as a string (e.g. `"200"`).

**Which issue(s) this PR fixes**:
Fixes #2968

**Checklist**
- [x] Tests updated
- [ ] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #4869, opened 2025-03-18
TraceQL: make comparison to nil symmetric

stoewer:
**What this PR does**:
Make comparison to `nil` such as `{ nil != .attr }` symmetric.

Fixes https://github.com/grafana/tempo/issues/3989

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/traceql/ast_execute.go
```diff
@@ -250,7 +250,7 @@ func (a Aggregate) evaluate(input []*Spanset) (output []*Spanset, err error) {
 			count := 0
 			for _, s := range ss.Spans {
 				val, err := a.e.execute(s)
-				if err != nil {
+				if err != nil || val.Equals(&StaticNil) {
```
joe-elliott, line 253: is this related to symmetric traceql? i'm not opposed. it makes sense to ignore nils here but i wonder if this is part of a larger change to have better nil handling across traceql.

joe-elliott (approved):
changelog?

one question but this looks simple enough to approve. did it really only require a single line in isMatchingOperand?

pkg/traceql/ast_execute.go
stoewer: After making the nil check in `isMatchingOperand()` symmetric the tests for the query `{{} | avg(.dne) != 0}` failed, because the aggregation over the non existing attribute returned results. The explanation is that parts of the logic relied on the asymmetric nil check in `isMatchingOperand()`

> i wonder if this is part of a larger change to have better nil handling across traceql

Good point. I'll run some more tests and look out for other missing nil checks.

stoewer:
> did it really only require a single line in isMatchingOperand?

Yes, the nil check in `isMatchingOperand()` was not symmetric. The other type check in `BinaryOperation.execute()` take both lhs and rhs into account. Therefor if one operand is nil, it always leads to a call of `lhs.Equals(&rhs)` which is symmetric. Since we are relying on this I'll add some more test cases to check for symmetry to `TestStatic_Equals`

pkg/traceql/ast_execute.go
stoewer: I've checked the aggregates `min`, `max`, and `sum` and they now all return results when they are applied to non existing attributes like this `{} | sum(.dne) != 0}`.

image

Since there is a test for this in `searchesThatDontMatch` [here](https://github.com/grafana/tempo/blob/d6b6e2d8383c9fdee4a549e7b6cd30d9d555570f/tempodb/tempodb_search_test.go#L276), I assume that the correct behavior is that all those aggregations don't return results.

On the other hand: looking at the results above the `sum(.dne)` column is populated with `nil` and technically `nil != 0` is true. Therefore returning results is technically also not wrong. I would just like to confirm the correct behavior here...

stoewer: The behavior that I've now implemented for `avg`, `sum`, `min`, and `max` in [this commit](https://github.com/grafana/tempo/pull/4869/commits/d2517d6e8954ba16f223aa1bc1264dfa6560ed4c) is as follows:
* Spans without the attributes are skipped from the aggregate
* Spansets without any any span with the aggregated attribute are removed from the results

tempodb/encoding/vparquet4/block_traceql_test.go
```diff
@@ -165,6 +165,11 @@ func TestBackendBlockSearchTraceQL(t *testing.T) {
 		{"instrumentation:name", traceql.MustExtractFetchSpansRequestWithMetadata(`{instrumentation:name = "scope-1"}`)},
 		{"instrumentation:version", traceql.MustExtractFetchSpansRequestWithMetadata(`{instrumentation:version = "version-1"}`)},
 		{"instrumentation.attr-str", traceql.MustExtractFetchSpansRequestWithMetadata(`{instrumentation.scope-attr-str = "scope-attr-1"}`)},
+		// Operations containing nil
+		{".foo != nil", traceql.MustExtractFetchSpansRequestWithMetadata(`{.foo != nil}`)},
+		{"nil != .foo", traceql.MustExtractFetchSpansRequestWithMetadata(`{nil != .foo}`)},
+		{"span.http.status_code != nil", traceql.MustExtractFetchSpansRequestWithMetadata(`{span.http.status_code != nil}`)},
+		{"nil != span.http.status_code", traceql.MustExtractFetchSpansRequestWithMetadata(`{nil != span.http.status_code}`)},
```
stoewer, line 172: I'll consolidate those tests with similar test cases from #4864 once the other PR is merged

pkg/traceql/ast_execute.go
joe-elliott: @electron0zero was asking questions about nil handling awhile back and we never answered that Q. I personally like the behavior you've chosen. I think it's the most consistent with the language.

PR #4870, opened 2025-03-18
[rhythm] Block-builder graceful shutdown

mapno:
**What this PR does**:

Added proper graceful shutdown to the block builder component to ensure operational safety during service termination. Instead of abruptly terminating, it now:

1. Uses a separate cancelable context for consume operations
2. Properly signals when consumption cycles are complete
3. Waits for the current cycle to finish during shutdown

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/blockbuilder/blockbuilder.go
+++ b/modules/blockbuilder/blockbuilder.go
@@ -89,6 +89,8 @@ type BlockBuilder struct {
 	enc       encoding.VersionedEncoding
 	wal       *wal.WAL // TODO - Shared between tenants, should be per tenant?
 	writer    tempodb.Writer
+
+	consumeStopped chan struct{}
 }
 
 type partitionState struct {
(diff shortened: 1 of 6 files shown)


modules/blockbuilder/blockbuilder.go
```diff
@@ -471,6 +482,12 @@ func (b *BlockBuilder) getPartitionState(partition int32, commits kadm.OffsetRes
 }
 
 func (b *BlockBuilder) stopping(err error) error {
+	select {
+	case <-b.consumeStopped:
+	case <-time.After(2 * b.cfg.ConsumeCycleDuration):
```
mapno, line 487: Maybe this is too long? Wasn't sure what a reasonable timeout is. I'd prefer for this to not be configurable, a good default should be good enough.

javiermolinar: We can set the terminationGracePeriod

https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-terminating-with-grace

mdisibio (approved):

modules/blockbuilder/blockbuilder.go
mapno: Good call. Extended the timeout and `TerminationGracePeriod` to 60s.

PR #4866, opened 2025-03-17
[rhythm] Query param for rf1_after config

mapno:
**What this PR does**:

Follow-up to https://github.com/grafana/tempo/pull/4857. 

Adds a new `rf1After` query parameter to search endpoints, allowing clients to dynamically filter blocks based on replication factor and time constraints.

It also updates Docker build images for protobuf generation to newer versions and fixes timestamp protobuf imports.

**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #4871, opened 2025-03-18
[rhythm] Add support for passing rf1_after query param in Vulture

mapno:
**What this PR does**:

Follow-up to https://github.com/grafana/tempo/pull/4866

Added support for filtering search results to only include blocks with RF=1 after a specified timestamp, helping with rhythm consistency testing. Implemented via a new CLI flag in Vulture and an underlying query parameter propagation mechanism in the HTTP client.


joe-elliott (commented):
Why are we passing this as a query param? Shouldn't this be configured in the query frontend and it makes the decision about whether to use rf1 blocks or not?

mapno:
> Why are we passing this as a query param? Shouldn't this be configured in the query frontend and it makes the decision about whether to use rf1 blocks or not?

We may want to simultaneously run both options during the transition period to gain confidence. For example, we may want to run two sets of vultures, each querying a different set of blocks.

I don't expect the query param to be used in the long-term. It'll probably be removed before we finish all the changes.

javiermolinar:
> > Why are we passing this as a query param? Shouldn't this be configured in the query frontend and it makes the decision about whether to use rf1 blocks or not?
> 
> We may want to simultaneously run both options during the transition period to gain confidence. For example, we may want to run two sets of vultures, each querying a different set of blocks.
> 
> I don't expect the query param to be used in the long-term. It'll probably be removed before we finish all the changes.

We can skip the changelog

PR #4872, opened 2025-03-18
[Rhythm] Block builder not mutate blocks across replays

mdisibio:
**What this PR does**:
Blocks in Tempo are immutable.  Once written, they are never changed, and the compactor and polling rely on this.  The current implementation of the block builder violates this by reflushing a new data.parquet to the same ID, with possibly different contents.    This gives the situation where the info in the block index no longer matches the contents and queries fail.

This changes the block builder to detect when a block from a previous attempt exists, and instead of overwriting it, mark it compacted and flush a new block.   Eventually the blocks will be repolled and effectively the old block is replaced with the new one.   Any number of replays are tolerated, because we use the deterministic block ID generator to continue forward until the next unused ID is found.

There are multiple efforts in this area, including #4870. Expect more follow up PRs.

Examples of handled scenarios:
* Block builder OOMs after flushing half of the tenants, restarts and begins another consumption cycle on the same commit
* Commit ultimately fails and the cycle must be restarted
* Cycle duration is changed, causing pod to restart. In this case the new block contents will always be different.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/blockbuilder/tenant_store.go
+++ b/modules/blockbuilder/tenant_store.go
@@ -88,15 +88,41 @@ func (s *tenantStore) AppendTrace(traceID []byte, tr []byte, ts time.Time) error
 	return nil
 }
 
-func (s *tenantStore) Flush(ctx context.Context, store tempodb.Writer) error {
+func (s *tenantStore) Flush(ctx context.Context, r tempodb.Reader, w tempodb.Writer, c tempodb.Compactor) error {
 	if s.liveTraces.Len() == 0 {
 		// This can happen if the tenant instance was created but
 		// no live traces were successfully pushed. i.e. all exceeded max trace size.
 		return nil
 	}
 
+	blockID, existingBlocksToBeCompacted, err := s.determineBlockIDs(ctx, r)
+	if err != nil {
+		return err
+	}
+
+	// TODO - Check if the existing block can be reused and exit early
+
+	if len(existingBlocksToBeCompacted) > 0 {
+		level.Info(s.logger).Log(
+			"msg", "Found existing blocks, marking them compacted",
+			"tenant", s.tenantID,
+			"existing_blocks", existingBlocksToBeCompacted,
+		)
+
+		for _, blockID := range existingBlocksToBeCompacted {
+			level.Info(s.logger).Log(
+				"msg", "Marking existing block compacted",
+				"tenant", s.tenantID,
+				"blockid", blockID,
+			)
+			if err := c.MarkBlockCompacted(s.tenantID, blockID); err != nil {
+				return err
+			}
+		}
+	}
+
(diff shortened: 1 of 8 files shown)


modules/blockbuilder/tenant_store.go
```diff
@@ -88,15 +88,41 @@ func (s *tenantStore) AppendTrace(traceID []byte, tr []byte, ts time.Time) error
...
 
+	blockID, existingBlocksToBeCompacted, err := s.determineBlockIDs(ctx, r)
+	if err != nil {
+		return err
+	}
+
+	// TODO - Check if the existing block can be reused and exit early
+
+	if len(existingBlocksToBeCompacted) > 0 {
+		level.Info(s.logger).Log(
```
javiermolinar, line 106: Do we need this log? Maybe we could have this or the one below

tempodb/tempodb.go
```diff
@@ -274,6 +277,20 @@ func (rw *readerWriter) WAL() *wal.WAL {
 	return rw.wal
 }
 
+func (rw *readerWriter) BlockMeta(ctx context.Context, tenantID string, blockID backend.UUID) (*backend.BlockMeta, *backend.CompactedBlockMeta, error) {
+	meta, err := rw.r.BlockMeta(ctx, (uuid.UUID)(blockID), tenantID)
```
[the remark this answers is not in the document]
mdisibio: Technically both files can exist, but you're right.  We should go ahead and exit early, and the benefit of this is that it only makes 1 object store read when meta.json exists instead of always making 2.

javiermolinar (approved):

modules/blockbuilder/tenant_store.go
```diff
@@ -88,15 +88,35 @@ func (s *tenantStore) AppendTrace(traceID []byte, tr []byte, ts time.Time) error
...
+	blockID, existingBlocksToBeCompacted, err := s.determineBlockIDs(ctx, r)
+	if err != nil {
+		return err
+	}
+
+	// TODO - Check if the existing block can be reused and exit early
+
+	if len(existingBlocksToBeCompacted) > 0 {
+		for _, blockID := range existingBlocksToBeCompacted {
+			level.Info(s.logger).Log(
```
javiermolinar, line 107: one last thing. should this be a warning instead? I wonder if we should add a metric to alert us when this happens. This should be a really edge case, otherwise the bug is on our side

mdisibio: Wouldn't go so far as to call it a bug, since any restart mid-way through a cycle can cause it.  My guess is that even with changes #4870 and #4874, it will be less frequent, but still a regular occurrence. But good with changing to warning.

PR #4895, opened 2025-03-24
[rhythm] Add support for `rf1After` query param in vulture.

mapno:
**What this PR does**:

Follow-up to https://github.com/grafana/tempo/pull/4857

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/httpclient/client.go
+++ b/pkg/httpclient/client.go
@@ -17,7 +17,6 @@ import (
 
 	userconfigurableoverrides "github.com/grafana/tempo/modules/overrides/userconfigurable/client"
 	"github.com/grafana/tempo/pkg/api"
-	tempo_api "github.com/grafana/tempo/pkg/api"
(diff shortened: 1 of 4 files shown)


pkg/httpclient/client.go
```diff
@@ -89,6 +89,13 @@ func (c *Client) SetHeader(key string, value string) {
 	c.headers[key] = value
 }
 
+func (c *Client) SetQueryParam(key, value string) {
+	if c.queryParams == nil {
+		c.queryParams = make(map[string]string)
+	}
+	c.queryParams[key] = value
+}
```
mapno, line 97: I'm not particularly happy with this implementation, but I wanted to do as little changes as possible given this is likely a temporary feature of Vulture.

mdisibio: 👍 This approach to set global query params fine to me.

mdisibio (approved):

PR #4897, opened 2025-03-24
Finish polling current tenants before shut down

zalegrala:
**What this PR does**:

Currently, the module context is passed to the poller.  In the event of a termination request, this context is used on the backend reader/writer and fail because of the cancellation.  Here we use a background context for the polling so that once a poll for a tenant has started to ensure the index is written.  The store module is updated to block shutdown until the poller is done.  This is necessary because the polling process is run in a routine, and during rollout of the index-writers, we see alerts due to stale index.  Allowing the poller to complete tenants which it has started should improve this siutaiton.
 

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


mapno (approved):
LGTM. It's missing the changelog entry.

PR #4899, opened 2025-03-25
Extend artificial delay with default configuration

mapno:
**What this PR does**:

This PR extends the artificial delay feature with default configuration in the distributor.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/artificial_delay.go
+++ b/modules/distributor/artificial_delay.go
@@ -6,7 +6,11 @@ import (
 )
 
 func (d *Distributor) padWithArtificialDelay(reqStart time.Time, userID string) {
-	artificialDelay := d.overrides.IngestionArtificialDelay(userID)
+	artificialDelay := d.cfg.ArtificialDelay
+	if artificialDelayOverride, ok := d.overrides.IngestionArtificialDelay(userID); ok {
+		artificialDelay = artificialDelayOverride
+	}
+
 	if artificialDelay <= 0 {
 		return
 	}
--- a/modules/distributor/artificial_delay_test.go
+++ b/modules/distributor/artificial_delay_test.go
@@ -14,37 +14,53 @@ func Test_outerMaybeDelayMiddleware(t *testing.T) {
 	tests := []struct {
 		name          string
 		userID        string
-		delay         time.Duration
+		delayCfg      time.Duration
+		delayOverride *time.Duration
 		reqDuration   time.Duration
 		expectedSleep time.Duration
 	}{
 		{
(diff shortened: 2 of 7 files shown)


modules/distributor/artificial_delay.go
```diff
@@ -6,7 +6,11 @@ import (
 )
 
 func (d *Distributor) padWithArtificialDelay(reqStart time.Time, userID string) {
-	artificialDelay := d.overrides.IngestionArtificialDelay(userID)
+	artificialDelay := d.cfg.ArtificialDelay
+	if artificialDelayOverride, ok := d.overrides.IngestionArtificialDelay(userID); ok {
```
javiermolinar, line 10: Should not check that` d.overrides.IngestionArtificialDelay(userID)` is not 0?. If not it can overwrite the default value (not sure if that is the intended behaviour).

We do something similar here:

https://github.com/grafana/tempo/blob/ff83648dc7415e109c9649a977c9215f0e2f0bf3/modules/frontend/metrics_query_range_sharder.go#L326-L334

modules/distributor/artificial_delay_test.go
```diff
@@ -14,37 +14,53 @@ func Test_outerMaybeDelayMiddleware(t *testing.T) {
...
 		},
+		{
+			name:          "Global delay configured",
+			userID:        "user4",
+			delayCfg:      500 * time.Millisecond,
+			reqDuration:   50 * time.Millisecond,
+			expectedSleep: 450 * time.Millisecond,
+		},
+		{
+			name:          "Global delay is overridden",
```
javiermolinar, line 51: There is one missing case:

```go
{
  name:          "Global delay configured",
  userID:        "user4",
  delayCfg:      500 * time.Millisecond,
  delayOverride: durationPtr(0 * time.Millisecond),
  reqDuration:   50 * time.Millisecond,
  expectedSleep: 450 * time.Millisecond,
},
```

javiermolinar (commented):
Nice, just a comment

modules/distributor/artificial_delay.go
javiermolinar: Do we allow a user override with value 0 to override the default one?. If that's the case, then this is fine. In the test case that I propose below we find that scenario

modules/distributor/artificial_delay_test.go
mapno: Added, good catch!

javiermolinar (approved):

PR #4904, opened 2025-03-25
Initial retention job

zalegrala:
**What this PR does**:

Here we leverage the backend scheduler/worker model to implement a retention job which will retain all tenants.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/tempopb/backendwork.proto
```diff
@@ -46,8 +48,7 @@ message JobDetail {
   string tenant = 1;
```
zalegrala, line 48: Currently the retention job retains all tenants, and so doesn't need the tenant field.  Perhaps `tenant` could move to the job detail on the jobs which need it.

modules/backendscheduler/backendscheduler.go
```diff
@@ -222,9 +247,14 @@ func (s *BackendScheduler) Next(ctx context.Context, req *tempopb.NextJobRequest
 		return resp, nil
 	}
 
-	// only one job type right now, just grab the next comapction job. later we can add code to grab the next whatever type of job
-	j = s.nextCompactionJob(ctx)
-	if j != nil {
+	select {
+	case j = <-s.jobChan:
```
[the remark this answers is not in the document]
zalegrala: I'd like to review the structure here again.  We can likely add a little consistency between the job types.

zalegrala:
Taking this back to draft while testing and cleaning up.

modules/backendscheduler/backendscheduler.go
```diff
@@ -82,116 +112,96 @@ func (s *BackendScheduler) starting(ctx context.Context) error {
...
+				case <-ctx.Done():
+					return
+				}
+			}
+		}(s.providers[i].jobs)
+	}
+
+	wg.Add(1)
+	// Start a goroutine to merge jobs from all providers
+	go func() {
```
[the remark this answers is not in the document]
zalegrala: This was an oversight that I've cleaned up a bit.

modules/backendscheduler/backendscheduler.go
```diff
@@ -82,116 +112,96 @@ func (s *BackendScheduler) starting(ctx context.Context) error {
...
+					select {
+					case job, ok := <-p.jobs:
+						if !ok {
+							continue // Provider channel closed
+						}
+						select {
+						case s.mergedJobs <- job:
+							jobFound = true
+						default:
+							// Merged channel full, try next time
```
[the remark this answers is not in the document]
zalegrala: This code was dropped.  Initially I had two approaches to this and left them both, but we only need one.  To your comment above too, I think we want to check if we can put an item in the channel and then return to the top so we can handle the context cancellation.

modules/backendscheduler/backendscheduler.go
```diff
@@ -82,116 +112,96 @@ func (s *BackendScheduler) starting(ctx context.Context) error {
...
+		// Start a goroutine to forward jobs from each provider to the merged channel
+		go func(jobs <-chan *work.Job) {
+			defer wg.Done()
+			for job := range jobs {
+				select {
+				case s.mergedJobs <- job:
+				case <-ctx.Done():
+					return
+				}
+			}
```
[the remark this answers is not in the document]
zalegrala: I think if we do something like this we avoid this edge case.
```go
			for job := range jobs {
				select {
				case <-ctx.Done():
					level.Info(log.Logger).Log("msg", "stopping provider forwarder", "provider", i)
					return
				default:
					select {
					case s.mergedJobs <- job:
						metricProviderJobsMerged.WithLabelValues(fmt.Sprintf("%d", i)).Inc()
					default:
						time.Sleep(100 * time.Millisecond)
					}
				}
			}
```

Thoughts?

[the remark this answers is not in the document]
zalegrala: I've pushed a change for this.

modules/backendscheduler/backendscheduler.go
```diff
@@ -59,20 +64,47 @@ func New(cfg Config, store storage.Store, overrides overrides.Interface, reader
...
-		reader:      reader,
-		writer:      writer,
-		curPriority: tenantselector.NewPriorityQueue(),
+		cfg:        cfg,
+		store:      store,
+		overrides:  overrides,
+		work:       work.New(cfg.Work),
+		reader:     reader,
+		writer:     writer,
+		mergedJobs: make(chan *work.Job, 100),
```
joe-elliott, line 73: would prefer an unbuffered channel here or at least very lightly buffered.  we shouldn't have any starvation issues on the workers b/c of how quick we can generate jobs and completely unbuffered would make us the most responsive to changes. the provider channels could be unbuffered as well as this would make a very easy to understand "machine" that delivered jobs from the providers to the Next() function

modules/backendscheduler/backendscheduler.go
```diff
@@ -82,123 +114,98 @@ func (s *BackendScheduler) starting(ctx context.Context) error {
...
+		s.providers[i].jobs = s.providers[i].provider.Start(ctx)
+
+		wg.Add(1)
+		// Start a goroutine to forward jobs from each provider to the merged channel
+		go func(jobs <-chan *work.Job) {
+			defer wg.Done()
+
+			var job *work.Job
+
+			for {
```
joe-elliott, line 132: suggested cleanup:

```
for {
  select {
    case job = <-jobs:
    case <-ctx.Done():
      return
   }
   
   select {
     case s.mergedJobs <- job:
     case <-ctx.Done():
       return
   } 
}
```

would remove the timer and the job nil handling and remove the potential to spin tightly if s.mergedJobs is full

modules/backendscheduler/provider/compaction.go
```diff
@@ -0,0 +1,395 @@
...
+		}
+
+		for {
+			select {
+			case <-ctx.Done():
+				level.Info(p.logger).Log("msg", "compaction provider stopping")
+				return
+			case <-measureTicker.C:
+				// Measure the tenants to get their current compaction status
+				p.measureTenants()
```
joe-elliott, line 114: i'm seeing this done here and below (line 173) should we remove the below goroutine and only do it here?

modules/backendscheduler/provider/retention.go
```diff
@@ -0,0 +1,86 @@
...
+			}
+		}
+	}()
+
+	return jobs
+}
+
+func (p *RetentionProvider) nextRetentionJob() *work.Job {
+	// Check if we already have a retention job running
+	for _, j := range p.sched.ListJobs() {
```
joe-elliott, line 69: i'm unsure on putting this here. imo this kind of "control" or "business" logic should be handled entirely in the backendscheduler. this would give us one place to look to see how jobs are prioritized/scheduled.

i think this is fine for now, but as we expand the number of providers, job types etc, i believe we will see a need to consolidate this kind of logic in one spot.

zalegrala: I see this also.  Let's see if we can clean this up with the next provider we add.

joe-elliott (approved):

modules/backendscheduler/provider/compaction.go
zalegrala: This ended up not being an issue only in the smaller environments.  Once we get to a large environment, the block here while measuring can take a considerable amount of time.  I've raised https://github.com/grafana/tempo/pull/5321 to move the measurement back out to its own routine.

PR #4905, opened 2025-03-25
{.a = Nil } 

ie-pham:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/parquetquery/iters.go
+++ b/pkg/parquetquery/iters.go
@@ -1163,7 +1163,6 @@ func (c *SyncIterator) next() (RowNumber, *pq.Value, error) {
 			c.curr.Next(v.RepetitionLevel(), v.DefinitionLevel())
 			c.currBufN++
 			c.currPageN++
-			fmt.Printf("sync iterator rn: %v, value %s\n", c.curr, *v)
 			if c.filter != nil && !c.filter.KeepValue(*v) {
 				continue
 			}
--- a/pkg/parquetquery/iters.go
+++ b/pkg/parquetquery/iters.go
@@ -1163,7 +1163,7 @@ func (c *SyncIterator) next() (RowNumber, *pq.Value, error) {
 			c.curr.Next(v.RepetitionLevel(), v.DefinitionLevel())
 			c.currBufN++
 			c.currPageN++
-
+			fmt.Printf("sync iterator rn: %v, value %s\n", c.curr, *v)
 			if c.filter != nil && !c.filter.KeepValue(*v) {
 				continue
 			}
--- a/pkg/parquetquery/nilIter.go
+++ b/pkg/parquetquery/nilIter.go
@@ -1,0 +1,533 @@
+package parquetquery
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"sync"
+
+	"github.com/grafana/tempo/pkg/parquetquery/intern"
+
+	pq "github.com/parquet-go/parquet-go"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
+
+type NilAttributeIteratorOpt func(*NilAttributeIterator)
+
+// NilAttributeIteratorOptIntern enables interning of string values.
+// This is useful when the same string value is repeated many times.
+// Not recommended with (very) high cardinality columns, such as UUIDs (spanID and traceID).
+func NilAttributeIteratorOptIntern() NilAttributeIteratorOpt {
+	return func(i *NilAttributeIterator) {
+		i.intern = true
+		i.interner = intern.New()
+	}
+}
+
+// NilAttributeIterator is like ColumnIterator but synchronous. It scans through the given row
+// groups and column, and applies the optional predicate to each chunk, page, and value.
+// Results are read by calling Next() until it returns nil.
+type NilAttributeIterator struct {
+	// Config
+	column     int
+	columnName string
+	rgs        []pq.RowGroup
+	rgsMin     []RowNumber
+	rgsMax     []RowNumber // Exclusive, row number of next one past the row group
+	readSize   int
+	filter     Predicate
+
+	// Status
+	span                     trace.Span
+	curr                     RowNumber
+	currRowGroup             pq.RowGroup
+	currRowGroupMin          RowNumber
+	currRowGroupMax          RowNumber
+	currChunk                *ColumnChunkHelper
+	currPage                 pq.Page
+	currPageMin              RowNumber
+	currPageMax              RowNumber
+	currValues               pq.ValueReader
+	currBuf                  []pq.Value
+	currBufN                 int
+	currPageN                int
+	at                       IteratorResult // Current value pointed at by iterator. Returned by call Next and SeekTo, valid until next call.
+	lastNonMatchingRowNumber RowNumber
+
+	intern   bool
+	interner *intern.Interner
+}
+
+var _ Iterator = (*NilAttributeIterator)(nil)
+
+func NewNilAttributeIterator(ctx context.Context, rgs []pq.RowGroup, column int, columnName string, readSize int, filter Predicate, selectAs string, opts ...NilAttributeIteratorOpt) *NilAttributeIterator {
+	// Assign row group bounds.
+	// Lower bound is inclusive
+	// Upper bound is exclusive, points at the first row of the next group
+	rn := EmptyRowNumber()
+	rgsMin := make([]RowNumber, len(rgs))
+	rgsMax := make([]RowNumber, len(rgs))
+	for i, rg := range rgs {
+		rgsMin[i] = rn
+		rgsMax[i] = rn
+		rgsMax[i].Skip(rg.NumRows() + 1)
+		rn.Skip(rg.NumRows())
+	}
+
+	_, span := tracer.Start(ctx, "NilAttributeIterator", trace.WithAttributes(
+		attribute.Int("columnIndex", column),
+		attribute.String("column", columnName),
+	))
+
+	at := IteratorResult{}
+	if selectAs != "" {
+		// Preallocate 1 entry with the given name.
+		at.Entries = []struct {
+			Key   string
+			Value pq.Value
+		}{
+			{Key: selectAs},
+		}
+	}
+
+	// Create the iterator
+	i := &NilAttributeIterator{
+		span:       span,
+		column:     column,
+		columnName: columnName,
+		rgs:        rgs,
+		readSize:   readSize,
+		rgsMin:     rgsMin,
+		rgsMax:     rgsMax,
+		filter:     filter,
+		curr:       EmptyRowNumber(),
+		at:         at,
+	}
+
+	// Apply options
+	for _, opt := range opts {
+		opt(i)
+	}
+
+	return i
+}
+
+func (c *NilAttributeIterator) String() string {
+	filter := "nil"
+	if c.filter != nil {
+		filter = c.filter.String()
+	}
+	return fmt.Sprintf("NilAttributeIterator: %s : %s", c.columnName, filter)
+}
+
+func (c *NilAttributeIterator) Next() (*IteratorResult, error) {
+	rn, v, err := c.next()
+	if err != nil {
+		return nil, err
+	}
+	if !rn.Valid() {
+		return nil, nil
+	}
+	return c.makeResult(rn, v), nil
+}
+
+// SeekTo moves this iterator to the next result that is greater than
+// or equal to the given row number (and based on the given definition level)
+func (c *NilAttributeIterator) SeekTo(to RowNumber, definitionLevel int) (*IteratorResult, error) {
+	if c.seekRowGroup(to, definitionLevel) {
+		return nil, nil
+	}
+
+	done, err := c.seekPages(to, definitionLevel)
+	if err != nil {
+		return nil, err
+	}
+	if done {
+		return nil, nil
+	}
+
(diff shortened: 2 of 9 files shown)


pkg/parquetquery/nilIter.go
```diff
@@ -0,0 +1,201 @@
...
+type NilAttributeIterator struct {
+	syncIterator          SyncIterator
+	lastRowNumberReturned RowNumber
+	attrFound             bool
+	lastBuff              bool
+}
+
+var _ Iterator = (*NilAttributeIterator)(nil)
+
+func NewNilAttributeIterator(ctx context.Context, rgs []pq.RowGroup, column int, columnName string, readSize int, filter Predicate, selectAs string, maxDefinitionLevel int, opts ...SyncIteratorOpt) *NilAttributeIterator {
```
stoewer: The iterator has the parameters `columnName`, `readSize`, `filter`, `selectAs`, and `maxDefinitionLevel` but they are not use. In `makeNilIterFunc` you are also passing them down via `opts`. I think they should be removed

pkg/parquetquery/nilIter.go
```diff
@@ -0,0 +1,201 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"io"
+
+	pq "github.com/parquet-go/parquet-go"
+)
+
+// NilAttributeIterator copies all functions of the sync iterator with just the next() function being different
+type NilAttributeIterator struct {
```
stoewer: I'm wondering if it would be better to implement the behavior of `NilAttributeIterator` in `SyncIterator` such that it can be turned on/off via a `SyncIteratorOption`?

pkg/parquetquery/nilIter.go
```diff
@@ -0,0 +1,201 @@
...
+	lastBuff              bool
+}
+
+var _ Iterator = (*NilAttributeIterator)(nil)
+
+func NewNilAttributeIterator(ctx context.Context, rgs []pq.RowGroup, column int, columnName string, readSize int, filter Predicate, selectAs string, maxDefinitionLevel int, opts ...SyncIteratorOpt) *NilAttributeIterator {
+	// Create the sync iterator
+	syncIterator := NewSyncIterator(ctx, rgs, column, opts...)
+	// Apply options
+	for _, opt := range opts {
```
stoewer: This should not be necessary as the options are already applied in `NewSyncIterator`, right?

pkg/parquetquery/predicates.go
```diff
@@ -478,3 +478,94 @@ func (m *CallbackPredicate) KeepColumnChunk(*ColumnChunkHelper) bool { return m.
 func (m *CallbackPredicate) KeepPage(pq.Page) bool { return m.cb() }
 
 func (m *CallbackPredicate) KeepValue(pq.Value) bool { return m.cb() }
+
+var _ Predicate = (*NilKeyPredicate)(nil)
+
+type NilKeyPredicate struct {
```
stoewer: It seems `NilKeyPredicate` has the same `KeepValue` logic as the `ByteEqualPredicate`, but omits the logic skipping column chunks and pages that don't contain the value. Why isn't it possible to use `ByteEqualPredicate` instead of `NilKeyPredicate`

pkg/parquetquery/predicates.go
```diff
@@ -478,3 +478,94 @@ func (m *CallbackPredicate) KeepColumnChunk(*ColumnChunkHelper) bool { return m.
...
+	// 	for i := 0; i < ci.NumPages(); i++ {
+	// 		min := ci.MinValue(i).IsNull()
+	// 		max := ci.MaxValue(i).IsNull()
+
+	// 		if min || max {
+	// 			return true
+	// 		}
+	// 	}
+	// 	return false
+	// }
```
stoewer: Please remove commented code

pkg/parquetquery/predicates.go
```diff
@@ -478,3 +478,94 @@ func (m *CallbackPredicate) KeepColumnChunk(*ColumnChunkHelper) bool { return m.
...
+	return true
+}
+
+func (p NilValuePredicate) KeepValue(v pq.Value) bool {
+	if v.IsNull() {
+		return true
+	}
+	switch v.Kind() {
+	case pq.ByteArray:
+		return bytes.Equal(v.ByteArray(), []byte{})
```
stoewer:
```suggestion
		return len(v.ByteArray()) == 0
```

pkg/parquetquery/nilIter.go
ie-pham: yes but the nil iter needs an additional 2 variable in the struct. do you think it would impact performance much?

stoewer: Maybe

stoewer: When I was working on the array support I implemented a version of [SyncIterator](https://github.com/grafana/tempo/compare/main...stoewer:tempo:traceql-collect-arrays-again) that was capable of reading arrays from the generic attribute columns. Technically this would also be able to detect nils, as they would come back as empty arrays

I'm wondering if we should resurrect the array support and use it for nil search as well?

pkg/parquetquery/nilIter.go
```diff
@@ -0,0 +1,185 @@
...
+	"errors"
+	"fmt"
+	"io"
+
+	pq "github.com/parquet-go/parquet-go"
+)
+
+// NilAttributeIterator copies all functions of the sync iterator with just the next() function being different
+type NilAttributeIterator struct {
+	syncIterator          SyncIterator
```
stoewer: If `SyncIterator` was embedded it would not be necessary to implement methods where you simply forward the call to the corresponding `SyncIterator` method like `SeekTo()`, `popRowGroup()`, `popRowGroup()` etc

pkg/parquetquery/nilIter.go
```diff
@@ -0,0 +1,185 @@
...
+
+func (c *NilAttributeIterator) SeekTo(to RowNumber, definitionLevel int) (*IteratorResult, error) {
+	return c.syncIterator.SeekTo(to, definitionLevel)
+}
+
+func (c *NilAttributeIterator) popRowGroup() (pq.RowGroup, RowNumber, RowNumber) {
+	return c.syncIterator.popRowGroup()
+}
+
+func (c *NilAttributeIterator) next2() (RowNumber, *pq.Value, error) {
```
stoewer: Any particular reason why it should not be named `next()`?

pkg/parquetquery/nilIter.go
```diff
@@ -0,0 +1,185 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"io"
+
+	pq "github.com/parquet-go/parquet-go"
+)
+
+// NilAttributeIterator copies all functions of the sync iterator with just the next() function being different
+type NilAttributeIterator struct {
```
stoewer: Naming:
* iterators in `parquetquery` are agnostic of Tempo's parquet schema and are therefore not aware of concepts like "attribute": please rename to `NilIterator`, `NilSyncIterator` or something similar
* file names in Go projects are typically in snake_case: please rename file to `nil_iter.go`

pkg/parquetquery/predicates.gen.go
```diff
@@ -701,9 +693,9 @@ func (p StringEqualPredicate) String() string {
 }
 
 func (p StringEqualPredicate) KeepColumnChunk(c *ColumnChunkHelper) bool {
-	if d := c.Dictionary(); d != nil {
-		return keepDictionary(d, p.KeepValue)
-	}
+	// if d := c.Dictionary(); d != nil {
+	// 	return keepDictionary(d, p.KeepValue)
+	// }
```
stoewer: Is this commented out on purpose?

pkg/parquetquery/predicates.go
```diff
@@ -478,3 +478,43 @@ func (m *CallbackPredicate) KeepColumnChunk(*ColumnChunkHelper) bool { return m.
...
+func NewNilValuePredicate() NilValuePredicate {
+	return NilValuePredicate{}
+}
+
+func (p NilValuePredicate) String() string {
+	return "NilValuePredicate{}"
+}
+
+func (p NilValuePredicate) KeepColumnChunk(_ *ColumnChunkHelper) bool {
+	return true
```
stoewer, line 505: The column chunk has access to the [ColumnIndex](https://pkg.go.dev/github.com/parquet-go/parquet-go#ColumnIndex) which has access to the number of nulls per page. It should be possible to use this in order to optimize `KeepColumnChunk()` and `KeepPage()`

pkg/parquetquery/predicates.go
```diff
@@ -478,3 +478,43 @@ func (m *CallbackPredicate) KeepColumnChunk(*ColumnChunkHelper) bool { return m.
...
+		return len(v.ByteArray()) == 0
+	case pq.Int32:
+		return v.Int32() == 0
+	case pq.Int64:
+		return v.Int64() == 0
+	case pq.Float:
+		return v.Float() == 0
+	case pq.Double:
+		return v.Double() == 0
+	default:
```
stoewer: For number types 0 is actually a valid non null value. Is there a reason why it isn't enough to just check for `v.IsNull()`?

pkg/traceql/ast_validate.go
```diff
@@ -198,12 +206,23 @@ func (o *BinaryOperation) validate() error {
...
 
 	if rhsT == TypeNil && o.Op == OpEqual {
-		return newUnsupportedError("{.a = nil}")
+		if o.LHS.impliedType() == TypeAttribute {
+			switch o.LHS.(Attribute).Scope {
+			case AttributeScopeEvent:
+				return newUnsupportedError("{event.a = nil}")
+			case AttributeScopeLink:
+				return newUnsupportedError("{link.a = nil}")
+			}
```
stoewer: Just curious: why is this not supported for event and link scope?

tempodb/encoding/vparquet4/block_traceql.go
```diff
@@ -2495,11 +2563,19 @@ func createPredicate(op traceql.Operator, operands traceql.Operands) (parquetque
 	}
 }
 
+func createNilPredicate(attribute string) parquetquery.Predicate {
```
stoewer: The name is a bit misleading as this creates a `StringEqualPredicate` the function also feels a bit unnecessary, can it be removed?

pkg/parquetquery/nilIter.go
```diff
@@ -0,0 +1,185 @@
...
+
+func (c *NilAttributeIterator) String() string {
+	filter := "nil"
+	if c.syncIterator.filter != nil {
+		filter = c.syncIterator.filter.String()
+	}
+	return fmt.Sprintf("NilAttributeIterator: %s : %s", c.syncIterator.columnName, filter)
+}
+
+func (c *NilAttributeIterator) Next() (*IteratorResult, error) {
```
stoewer: It would be great to have a test similar to `TestColumnIterator` that verifies the behavior in a controlled scenario

pkg/traceql/ast_validate.go
ie-pham: I wanted to have a discussion about this before implementing on the implications of this on levels lower than spans. If there are two events, one has the attribute and the other does, do we still return the span?

pkg/parquetquery/predicates.go
ie-pham: you are right - I was testing with status_code and duration and they had a nil value of 0. I thought that was how it was set but I just found that we set those values to 0, not that the nil value is 0. Will revert this and handle our intrinsics separately.

pkg/parquetquery/predicates.go
ie-pham: not sure i understand this. can you explain?

pkg/parquetquery/nilIter.go
ie-pham: not sure i understand this. can you explain?

pkg/parquetquery/predicates.gen.go
ie-pham: ah yes, will create a new predicate for this

tempodb/encoding/vparquet4/block_traceql.go
ie-pham: i created a new predicate in place of this - called `IncludeNilStringEqualPredicate` since the NilIterator needs to go through every value without skipping to make sure a key does not exist. Unless there is another way

pkg/parquetquery/iters.go
```diff
@@ -112,6 +112,10 @@ func (t *RowNumber) Valid() bool {
 	return t[0] >= 0
 }
 
+func RowNumberValidAtDefinitionLevel(t RowNumber, definitionLevel int) bool {
```
stoewer: Only used internally and does not need to be exported

pkg/parquetquery/nil_iter_test.go
```diff
@@ -0,0 +1,66 @@
...
+	type Attr struct {
+		Key string `parquet:",snappy,dict"`
+	}
+	type T struct {
+		Attrs []Attr `parquet:"attrs,list"`
+	}
+	genericKeys := []string{"key-1", "key-2", "key-3", "key-4", "key-5"}
+
+	rows := []T{}
+	for i := 0; i < count; i++ {
```
stoewer: `nitpick` you can just write `for i := range count`

pkg/parquetquery/predicates.go
stoewer: Please also add code to skip over complete column chunks in `KeepColumnChunk()` as suggested above

PR #4912, opened 2025-03-27
[rhythm] Add tracing to block-builder

mapno:
**What this PR does**:

Adds tracing to the block-builder

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/blockbuilder/blockbuilder.go
+++ b/modules/blockbuilder/blockbuilder.go
@@ -76,7 +76,7 @@ var (
 		Help:      "Total number of errors while fetching by the consumer.",
 	}, []string{"partition"})
 
-	tracer = otel.Tracer("modules/ingester")
+	tracer = otel.Tracer("modules/blockbuilder")
 )
 
 type BlockBuilder struct {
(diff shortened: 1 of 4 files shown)


modules/blockbuilder/blockbuilder.go
```diff
@@ -280,8 +291,13 @@ func (b *BlockBuilder) consume(ctx context.Context) (time.Duration, error) {
 }
 
 func (b *BlockBuilder) consumePartition(ctx context.Context, ps partitionState) (lastTs time.Time, commitOffset int64, err error) {
+	ctx, span := tracer.Start(ctx, "blockbuilder.consumePartition",
```
javiermolinar, line 294: We are detaching the link between the distributor and the blockbuilder.
I was wondering if we could take the context from the first record in the consumption loop and create a span from it:

https://github.com/twmb/franz-go/blob/master/plugin/kotel/README.md#processing-records

That way we could track the whole journey

mapno: I've added a link to the producer span, so block-builder traces are linked to producing traces.

Producing traces and block-builder traces is a many-to-many problem, if we were to build them as a single trace I'm not sure how we would build them without them being chaotic.

javiermolinar (approved):

PR #4913, opened 2025-03-27
[rhythm] Add endpoint for partition downscaling

mapno:
**What this PR does**:

Introduce `/ingester/prepare-partition-downscale` endpoint to manage ingester partition states. This enables transitions between ACTIVE and INACTIVE states, supporting partition downscaling.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/ingester/downscale.go
+++ b/modules/ingester/downscale.go
@@ -1,0 +1,105 @@
+package ingester
+
+import (
+	"net/http"
+
+	kitlog "github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+
+	"github.com/grafana/dskit/ring"
+	"github.com/grafana/dskit/services"
+	"github.com/grafana/tempo/pkg/util"
+	"github.com/grafana/tempo/pkg/util/log"
+)
+
+// PreparePartitionDownscaleHandler prepares the ingester's partition downscaling. The partition owned by the
+// ingester will switch to INACTIVE state (read-only).
+//
+// Following methods are supported:
+//
+//   - GET
+//     Returns timestamp when partition was switched to INACTIVE state, or 0, if partition is not in INACTIVE state.
+//
+//   - POST
+//     Switches the partition to INACTIVE state (if not yet), and returns the timestamp when the switch to
+//     INACTIVE state happened.
+//
+//   - DELETE
+//     Sets partition back from INACTIVE to ACTIVE state, and returns 0 signalling the partition is not in INACTIVE state
+func (i *Ingester) PreparePartitionDownscaleHandler(w http.ResponseWriter, r *http.Request) {
+	logger := kitlog.With(log.Logger, "partition", i.ingestPartitionID)
+
+	// Don't allow callers to change the shutdown configuration while we're in the middle
+	// of starting or shutting down.
+	if i.State() != services.Running {
+		w.WriteHeader(http.StatusServiceUnavailable)
+		return
+	}
+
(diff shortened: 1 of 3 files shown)


modules/ingester/downscale.go
```diff
@@ -0,0 +1,105 @@
+package ingester
+
+import (
+	"net/http"
+
+	kitlog "github.com/go-kit/log"
```
javiermolinar, line 6: Could we use the standard slog instead?

modules/ingester/downscale.go
```diff
@@ -0,0 +1,105 @@
...
+
+	state, stateTimestamp, err := i.ingestPartitionLifecycler.GetPartitionState(r.Context())
+	if err != nil {
+		level.Error(logger).Log("msg", "failed to check partition state in the ring", "err", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	if state == ring.PartitionInactive {
+		util.WriteJSONResponse(w, map[string]any{"timestamp": stateTimestamp.Unix()})
```
javiermolinar: Maybe we could return the status the of the partition and the timestamp

```
{
 "status": "active"
  "timestamp": 0
}

```

javiermolinar (commented):
LGTM just some minor suggestions

modules/ingester/downscale.go
mapno: Done!

modules/ingester/downscale.go
mapno: I don't a strong opinion on loggers, but `slog` is not widely used in the codebase. IMO for consistency the go-kit logger makes more sense.

javiermolinar (approved):
LGTM

modules/ingester/downscale.go
```diff
@@ -0,0 +1,104 @@
...
+	case http.MethodDelete:
+		state, _, err := i.ingestPartitionLifecycler.GetPartitionState(r.Context())
+		if err != nil {
+			level.Error(logger).Log("msg", "failed to check partition state in the ring", "err", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		// If partition is inactive, make it active. We ignore other states Active and especially Pending.
+		if state == ring.PartitionInactive {
```
mdisibio, line 77: The comment says other states are ignored - but it's not clear what happens in the else condition - are they noops?  It might be more clear to do an early return instead?

modules/ingester/downscale.go
```diff
@@ -0,0 +1,104 @@
...
+		level.Error(logger).Log("msg", "failed to check partition state in the ring", "err", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	if state == ring.PartitionInactive {
+		util.WriteJSONResponse(w, map[string]any{"timestamp": stateTimestamp.Unix(), "state": state.String()})
+	} else {
+		util.WriteJSONResponse(w, map[string]any{"timestamp": 0, "state": state.String()})
+	}
```
ruslan-mikhailov: (very minor) To make it more explicit, I'd put this part under switch `case http.MethodGet`, and add default with `w.WriteHeader(some4xxStatus)`

mapno: Done!

modules/ingester/downscale.go
mapno: It's a noop, yes. It still returns with a response with the state of the partition.

PR #4915, opened 2025-03-27
Fix compactor panic

joe-elliott:
**What this PR does**:
Compactors rarely panic with a stack like:

```
        /enterprise-traces/vendor/github.com/parquet-go/parquet-go/page.go:129 +0xfd
created by github.com/parquet-go/parquet-go.AsyncPages in goroutine 33664
        /enterprise-traces/vendor/github.com/parquet-go/parquet-go/page.go:264 +0x178
github.com/parquet-go/parquet-go.readPages({0x3bc5390, 0xc06ed2f360}, 0xc0b09b2f50, 0xc0acc4b780, 0x3bea488?, 0xc0b09b3030)
        /enterprise-traces/vendor/github.com/parquet-go/parquet-go/file.go:801 +0x9f
github.com/parquet-go/parquet-go.(*FilePages).ReadPage(0xc06ed2f360)
        /enterprise-traces/vendor/github.com/parquet-go/parquet-go/file.go:916 +0x65
github.com/parquet-go/parquet-go.(*FilePages).readPage(0xc06ed2f360, 0xc096766f30, 0xc0795b8000)
        /enterprise-traces/vendor/github.com/parquet-go/parquet-go/buffer.go:532 +0xd7
github.com/parquet-go/parquet-go.(*bufferPool).get(0x5a1ff60?, 0xfffffffffffffffc?)
        /enterprise-traces/vendor/github.com/parquet-go/parquet-go/buffer.go:510 +0x2d
github.com/parquet-go/parquet-go.(*bufferPool).newBuffer(0x5a1ff60, 0xfffffffffffffffc, 0x1000)
goroutine 176675 [running]:   
        
panic: runtime error: makeslice: len out of range
```

This was caused by the `io.BufferedReaderAt` not invalidating buffers when an error was received here:

https://github.com/grafana/tempo/blob/c5e5978f1ba9bb0819df509f4de74625fc1eb5a4/pkg/io/buffered.go#L64-L68

Later it was possible to return data from this buffer as if it were correct even though the buffer was in an unknown state.

**Other Changes**

- Removed `BufferedWriterWithQueue` b/c it was unused.
- Added a test to replicate the described issue.
- Renamed open -> openForIteration and added comments to help clarify how the two open funcs are used.
- Simplified the locking in `BufferedReaderAt`. Previously it would drop the primary lock for per-buffer locking before it attempted to read from the backend. This was designed to prevent this struct from serializing calls to object storage. I believe this could be maintained with the fix, but it was getting quite complicated. Consider that you would need to handle a buffer being invalid after it had been already selected b/c the read might invalidate it. Since this code was only used in the compactors I favored simplicity over performance. There is a small, but reasonable performance hit.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/io/buffered.go
+++ b/pkg/io/buffered.go
@@ -68,7 +68,7 @@ func (r *BufferedReaderAt) populate(buf *readerBuffer) (int, error) {
 
 	// if err is fatal we need to invalidate the buffer by setting it back to 0s (uninitialized)
 	if isFatalError(err) {
-		buf.buf = buf.buf[:n]
+		buf.buf = buf.buf[:0]
 		buf.count = 0
 		buf.off = 0
 	}
--- a/pkg/io/buffered.go
+++ b/pkg/io/buffered.go
@@ -9,6 +9,10 @@ import (
 // Subsequent reads are returned from the buffers. Additionally it supports concurrent readers
 // by maintaining multiple buffers at different offsets, and matching up reads with existing
 // buffers where possible. When needed the least-recently-used buffer is overwritten with new reads.
+//
+// Note that the locking effectively serializes reads to the underlying io.ReaderAt. This reader
+// is not suitable for high-concurrency workloads, but does nicely reduce memory usage and backend calls
+// for batch workloads.
 type BufferedReaderAt struct {
 	mtx     sync.Mutex
 	ra      io.ReaderAt
--- a/pkg/io/buffered.go
+++ b/pkg/io/buffered.go
@@ -63,10 +63,10 @@ func (r *BufferedReaderAt) prep(buf *readerBuffer, offset, length int64) {
 }
 
 func (r *BufferedReaderAt) populate(buf *readerBuffer) (int, error) {
-	// Read
+	// read
 	n, err := r.ra.ReadAt(buf.buf, buf.off)
 
-	// if err != nil we need to invalidate the buffer by setting it back to 0s (uninitialized)
+	// if err is fatal we need to invalidate the buffer by setting it back to 0s (uninitialized)
 	if isFatalError(err) {
 		buf.buf = buf.buf[:n]
 		buf.count = 0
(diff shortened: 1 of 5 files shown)


pkg/io/buffered.go
```diff
@@ -128,29 +124,19 @@ func (r *BufferedReaderAt) ReadAt(b []byte, offset int64) (int, error) {
...
 
-		if _, err := r.populate(buf); err != nil {
+		var err error
+		if _, err = r.populate(buf); isFatalError(err) {
 			return 0, err
 		}
 
 		r.read(buf, b, offset)
-		return len(b), nil
+		return len(b), err
```
mattdurham, line 136: Is this a change of behavior in the case of EOF error? ie buf will be zeroed out and the additional call to r.read?

joe-elliott: yeah, this is [technically correct ReaderAt](https://pkg.go.dev/io#ReaderAt) behavior. it's possible for the nested ReaderAt to return both valid data and io.EOF and this code attempts to pass this back to the caller.

now the way parquet-go uses ReaderAt this should never happen b/c we're just reading known ranges out of an object in object storage somewhere. i'm tempted to leave it the way it was to reduce complexity. i also noticed these locations that behave incorrectly while debugging this panic:

https://github.com/grafana/tempo/blob/c5e5978f1ba9bb0819df509f4de74625fc1eb5a4/tempodb/encoding/vparquet4/readers.go#L45-L47

https://github.com/grafana/tempo/blob/c5e5978f1ba9bb0819df509f4de74625fc1eb5a4/tempodb/encoding/vparquet4/readers.go#L150-L152

i feel like i should fix these as well even though it's added complexity and i'm sure we break this contract in other spots.

mattdurham: Sounds entirely reasonable, has long as we internally arent adversely affected.

mapno (approved):
LGTM

PR #4936, opened 2025-04-02
distributor: add counter for spans in the future

carles-grafana:
**What this PR does**:

This will allow us to find customers that send spans in the future,
which can't be found using the Search API.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/distributor.go
+++ b/modules/distributor/distributor.go
@@ -36,6 +36,7 @@ import (
 	generator_client "github.com/grafana/tempo/modules/generator/client"
 	ingester_client "github.com/grafana/tempo/modules/ingester/client"
 	"github.com/grafana/tempo/modules/overrides"
+	"github.com/grafana/tempo/pkg/dataquality"
 	"github.com/grafana/tempo/pkg/ingest"
 	"github.com/grafana/tempo/pkg/model"
 	"github.com/grafana/tempo/pkg/tempopb"
@@ -430,6 +431,7 @@ func (d *Distributor) PushTraces(ctx context.Context, traces ptrace.Traces) (*te
 	if spanCount == 0 {
 		return &tempopb.PushResponse{}, nil
 	}
+
 	// check limits
 	// todo - usage tracker include discarded bytes?
 	err = d.checkForRateLimits(size, spanCount, userID)
@@ -461,6 +463,7 @@ func (d *Distributor) PushTraces(ctx context.Context, traces ptrace.Traces) (*te
 
 	metricBytesIngested.WithLabelValues(userID).Add(float64(size))
 	metricSpansIngested.WithLabelValues(userID).Add(float64(spanCount))
+
 	statBytesReceived.Inc(int64(size))
 	statSpansReceived.Inc(int64(spanCount))
 
@@ -705,7 +708,7 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount, ma
 	const tracesPerBatch = 20 // p50 of internal env
 	tracesByID := make(map[uint64]*rebatchedTrace, tracesPerBatch)
 	truncatedAttributeCount := 0
-
+	currentTime := uint32(time.Now().Unix())
 	for _, b := range batches {
 		spansByILS := make(map[uint64]*v1.ScopeSpans)
 		// check resource for large attributes
@@ -793,6 +796,13 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount, ma
 
 				// increase span count for trace
 				existingTrace.spanCount = existingTrace.spanCount + 1
+
+				// Count spans with timestamps in the future
+				if end > currentTime {
+					dataquality.MetricSpanInFuture.WithLabelValues(userID).Observe(float64(end - currentTime))
+				} else {
+					dataquality.MetricSpanInPast.WithLabelValues(userID).Observe(float64(currentTime - end))
+				}
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


javiermolinar (commented):
This is roughly the same as we have in the data quality package. What do you think of adding it there?. It would be good to get all the related code in the same place. The name of the metrics and the help could be improved, though..

https://github.com/grafana/tempo/blob/911e7e053ad85bd3499fa79a04c1cd338caad3c1/pkg/dataquality/warnings.go#L19-L23

modules/distributor/distributor.go
javiermolinar, line 1: Currently, we are tracking the data quality warning here:

https://github.com/grafana/tempo/blob/a2949c33412e9b62bba709b0c16901fb41a895f3/tempodb/encoding/common/slackTimeRange.go#L24

This is called when the trace is appended to the WAL

With this implementation, the counter will be incremented twice. Once for the `span_in_future` and another one for the `outside_ingestion_time_slack`

Since the `span_in_future` is a subset of `outside_ingestion_time_slack`, I'm not quite sure what the correct behavior here is.

carles-grafana: I discussed with Joe the possibility of discarding future spans at some point, so this will help us understand how often it happens.

javiermolinar: Yes, my point is if we should add the code in the  `AdjustTimeRangeForSlack` by checking if the dates are in the future and, in that case, just increment the counter for the  `span_in_future`.

carles-grafana: I believe AdjustTimeRangeForSlack is per Trace while we are trying to count spans, I think the implementation in the PR is more direct and clearer.

carles-grafana:
As discussed with Joe, we should use the slack time to determine whether a span is in the past or in the future. However, this configuration field is only available in the Ingesters, and these have RF3, which means we cannot reliably count ingested spans there.

I will try instead to implement it in the block builder for Tempo 3.

modules/distributor/distributor.go
```diff
@@ -773,6 +776,13 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount, ma
 
 				// increase span count for trace
 				existingTrace.spanCount = existingTrace.spanCount + 1
+
+				// Count spans with timestamps in the future
+				if end > currentTime {
+					dataquality.MetricSpanInFuture.WithLabelValues(userID).Observe(float64(end - currentTime))
+				} else {
```
carles-grafana, line 783: all correct spans will be in the past, so I'm doubting whether it's worth having the histogram for spans in the past, unless we want to have a histogram of how far in the past they are.

pkg/dataquality/warnings.go
```diff
@@ -37,3 +37,17 @@ func WarnDisconnectedTrace(tenant string, phase string) {
 func WarnRootlessTrace(tenant string, phase string) {
 	metric.WithLabelValues(tenant, reasonRootlessTrace+phase).Inc()
 }
+
+var MetricSpanInFuture = promauto.NewHistogramVec(prometheus.HistogramOpts{
+	Namespace: "tempo",
+	Name:      "span_in_future_total",
```
joe-elliott, line 43: the suffix on prometheus metrics [should be the unit](https://prometheus.io/docs/practices/naming/). something like?

```
tempo_spans_distance_in_future_seconds
```

wdyt about putting these in the dataquality package:

https://github.com/grafana/tempo/blob/main/pkg/dataquality/warnings.go

modules/distributor/distributor.go
```diff
@@ -773,6 +776,13 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount, ma
...
+					dataquality.MetricSpanInPast.WithLabelValues(userID).Observe(float64(currentTime - start))
```
joe-elliott, line 784: we use span end time to enforce ingestion slack in other spots. should we use the end time here? what are we attempting to do with this metric?

if the goal is to measure lag due to the tracing pipeline i think end makes more sense, but maybe our goals here are different?

carles-grafana: I don't think it's worth having the metric in the past, only for spans in the future, unless you have use case in mind

carles-grafana: which makes me thing a histogram is really worth it. Maybe we should go for the original idea of counting spans in the future and that's it.

joe-elliott: A spans in the future counter seems useful as it suggests something wildly wrong like a system clock misconfigured or a script that's generating bad data.

A spans in the past histogram seems useful as it can give information about the impact of delays due to trace pipelining.

@mapno how does rhythm impact ingestion slack in the block builder/generator? does the time in queue impact this behavior or do we propagate the ingestion time forward somehow?

mapno: >@mapno how does rhythm impact ingestion slack in the block builder/generator? does the time in queue impact this behavior or do we propagate the ingestion time forward somehow?

Traces get longer to get to metrics-generators (and block-builders) than with the classic architecture. Delay should go from milliseconds in normal operation to whatever lag it accumulates—minutes, hours.

In the distributors there is no difference.

carles-grafana: I've changed it to the end time, I agree it makes more sense.

knylander-grafana:
Are there any changes that impact Tempo users? If so, we should consider adding docs.

carles-grafana:
@joe-elliott  are we still interested in having this?

PR #4964, opened 2025-04-07
My ubuntu version did not have eth0 or any of the other named interfaces

mattdurham:
Relevant links https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/ and https://superuser.com/questions/1053003/what-is-the-difference-between-eth1-and-eno1

### DIFF
--- a/modules/backendworker/backendworker_test.go
+++ b/modules/backendworker/backendworker_test.go
@@ -72,7 +72,7 @@ func setupDependencies(ctx context.Context, t *testing.T, limits overrides.Confi
 	workerConfig.BackendSchedulerAddr = "localhost:1234"
 	workerConfig.Ring.KVStore.Store = "inmemory"
 	workerConfig.Ring.KVStore.Mock = nil
-	workerConfig.Ring.InstanceInterfaceNames = []string{"eth0", "en0", "lo0", "enp102s0f4u1u3"}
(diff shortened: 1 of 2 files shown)


modules/backendworker/backendworker_test.go
```diff
@@ -72,7 +72,7 @@ func setupDependencies(ctx context.Context, t *testing.T, limits overrides.Confi
 	workerConfig.BackendSchedulerAddr = "localhost:1234"
 	workerConfig.Ring.KVStore.Store = "inmemory"
 	workerConfig.Ring.KVStore.Mock = nil
-	workerConfig.Ring.InstanceInterfaceNames = []string{"eth0", "en0", "lo0", "enp102s0f4u1u3"}
+	workerConfig.Ring.InstanceInterfaceNames = []string{"eth0", "en0", "lo0", "enp102s0f4u1u3", "eno1"}
```
joe-elliott, line 75: i'm guessing an empty slice doesn't check all interfaces? or maybe we could write a func that just grabbed all interfaces and added it to the config for tests?

fine with this if it just unblocks.

also lol at `enp102s0f4u1u3`

joe-elliott (approved):

electron0zero (approved):

modules/backendworker/backendworker_test.go
mattdurham: Also you will need to push the merge button, I cant.

zalegrala (approved):
Nice!

PR #4970, opened 2025-04-08
Reduce memcached allocations

joe-elliott:
**What this PR does**:
I discovered this opportunity while researching panics in the querier. This PR reduces allocations related to using memcached caching by pooling buffers.

- Use the memcached allocator option to pull buffers from a pool on read
- Return buffers to the pool in ~frontend and~ object storage cache
- Moved the prealloc buffer pool to `pkg/pool` to be used by cache
- Removed the logic to copy before setting in cache. This is synchronous in both memcached and redis so there is no reason to copy.
- Added a changelog for #4963 b/c I forgot!

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/cache/memcached.go
```diff
@@ -24,6 +25,10 @@ func (cfg *MemcachedConfig) RegisterFlagsWithPrefix(prefix, description string,
 	f.DurationVar(&cfg.Expiration, prefix+"memcached.expiration", 0, description+"How long keys stay in the memcache.")
 }
 
+var cacheBufAllocator = &allocator{
+	pool: pool.New("cache", 0, 100, 40*1024),
```
javiermolinar, line 29: In `prealloc.go' I see that the number of buckets and their sizes is configurable. Do we want the same here? What's the rationale behind the 40KB?

javiermolinar (approved):
Nice clean up

pkg/cache/memcached.go
joe-elliott: > Do we want the same here?

Sure. Not a bad idea.

> What's the rationale behind the 40KB?

Untuned. I just chose 100 buckets and wanted the cap to be 4MB to catch all value sizes in memcached.

PR #4962, opened 2025-04-07
Issue 4896 traceql metrics fixes

ruslan-mikhailov:
## What this PR does

Start and end of the trace in generator are converted to seconds, dropping the fractional part.
```go
// Convert from unix nanos to unix seconds
startSeconds := uint32(start / uint64(time.Second))
endSeconds := uint32(end / uint64(time.Second))

err := p.headBlock.AppendTrace(id, tr, startSeconds, endSeconds, p.Cfg.AdjustTimeRangeForSlack)
```
Additionally, blocks' start and end times are in seconds.

image

### Bug 1
Search in blocks was exclusive, but it should have included both edges. See https://github.com/grafana/tempo/blob/a1ab3f911f9656fc5a82e7f1bd76eea4494c7bc0/modules/frontend/frontend.go#L361 as a reference
### Bug 2. Losing the last trace in the block. 

image

It filters here:
```
			if e.checkTime {
				st := s.StartTimeUnixNanos()
				if st < e.start || st >= e.end {
					continue
				}
			}
```

To fix it, need to include the right border, but _only_ when request's time range fully includes the right edge of the block

### Bug 3. Lonely trace in a block. 
It is possible to have a block with start==end, it can be reproduced if only one trace was pushed and flushed.

This results in skipping this block
```go
func (e *MetricsEvaluator) Do(ctx context.Context, f SpansetFetcher, fetcherStart, fetcherEnd uint64) error {
	// Make a copy of the request so we can modify it.
	storageReq := *e.storageReq

	if fetcherStart > 0 && fetcherEnd > 0 &&
		fetcherStart != fetcherEnd {
			overlap := timeRangeOverlap(e.start, e.end, fetcherStart, fetcherEnd)
			if overlap == 0.0 { // <------ overlap will be zero
				// This shouldn't happen but might as well check.
				// No overlap == nothing to do
				return nil
			}
```
 
### Bug 4. Exemplars are not filtered
```go
		for _, s := range ss.Spans {
			if e.checkTime {
				st := s.StartTimeUnixNanos()
				if st < e.start || st >= e.end {
					continue
				}
			}

			e.spansTotal++

			e.metricsPipeline.observe(s)
		}

		if len(ss.Spans) > 0 && e.sampleExemplar(ss.TraceID) {
			e.metricsPipeline.observeExemplar(ss.Spans[rand.Intn(len(ss.Spans))])
		}
```
This results in including traces outside of the requested time range

image

## Which issue(s) this PR fixes
[...]

PR #4976, opened 2025-04-09
Issue 4632: TraceQL Metrics: respect exemplars parameter

### LINKED ISSUE #4632: Query exemplars field not respected
**Describe the bug**

**To Reproduce**
1. Start Tempo (2.7.0)

**Expected behavior**

ruslan-mikhailov:
**What this PR does**:
Respect exemplars parameters by limiting returned exemplars on query frontend side.
I decided to leave the latest traces as most probably they have more value.

Fixes #4632

**How it has been tested:**

1. Added check in e2e tests. The test fails with master tempo image
2. Unit test to reflect changes
3. Manually tested using `example/docker-compose/local`:
    A. Reproduce the issue with master tempo image by sending query with the following query parameters:
```json
{
    "q": "{} | rate()",
    "since": "3h",
    "step": "1m",
    "exemplars": 5,
}
```

B. Apply the fix. Expected result: number of exemplars returned is limited to 5

**Checklist**

- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/integration/e2e/api/query_range_test.go
+++ b/integration/e2e/api/query_range_test.go
@@ -33,6 +33,7 @@ var debugMode = false
 func TestQueryRangeExemplars(t *testing.T) {
 	t.Parallel()
 
+	maxExemplars := 1
 	s, err := e2e.NewScenario("tempo_e2e")
 	require.NoError(t, err)
 	defer s.Close()
@@ -79,19 +80,21 @@ sendLoop:
 		"{status != error} | count_over_time() by (status)",
 	} {
 		t.Run(query, func(t *testing.T) {
-			queryRangeRes := callQueryRange(t, tempo.Endpoint(tempoPort), query, debugMode)
+			queryRangeRes := callQueryRange(t, tempo.Endpoint(tempoPort), query, maxExemplars, debugMode)
 			require.NotNil(t, queryRangeRes)
 			require.GreaterOrEqual(t, len(queryRangeRes.GetSeries()), 1)
 			exemplarCount := 0
 			for _, series := range queryRangeRes.GetSeries() {
-				exemplarCount += len(series.GetExemplars())
+				exemplars := len(series.GetExemplars())
+				exemplarCount += exemplars
+				assert.LessOrEqual(t, exemplars, maxExemplars)
 			}
-			require.GreaterOrEqual(t, exemplarCount, 1)
+			assert.GreaterOrEqual(t, exemplarCount, 1)
 		})
 	}
 
 	// invalid query
-	res := doRequest(t, tempo.Endpoint(tempoPort), "{. a}")
+	res := doRequest(t, tempo.Endpoint(tempoPort), "{. a}", maxExemplars)
 	require.Equal(t, 400, res.StatusCode)
 
 	// query with empty results
@@ -102,7 +105,7 @@ sendLoop:
 		`{span.randomattr = "doesnotexist"} | count_over_time()`,
 	} {
 		t.Run(query, func(t *testing.T) {
-			queryRangeRes := callQueryRange(t, tempo.Endpoint(tempoPort), query, debugMode)
+			queryRangeRes := callQueryRange(t, tempo.Endpoint(tempoPort), query, maxExemplars, debugMode)
 			require.NotNil(t, queryRangeRes)
 			// it has time series but they are empty and has no exemplars
 			require.GreaterOrEqual(t, len(queryRangeRes.GetSeries()), 1)
@@ -142,7 +145,7 @@ func TestQueryRangeSingleTrace(t *testing.T) {
 
 	// Query the trace by count. As we have only one trace, we should get one dot with value 1
 	query := "{} | count_over_time()"
-	queryRangeRes := callQueryRange(t, tempo.Endpoint(tempoPort), query, debugMode)
+	queryRangeRes := callQueryRange(t, tempo.Endpoint(tempoPort), query, 100, debugMode)
 	require.NotNil(t, queryRangeRes)
 	require.Equal(t, len(queryRangeRes.GetSeries()), 1)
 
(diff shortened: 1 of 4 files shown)


integration/e2e/api/query_range_test.go
```diff
@@ -129,13 +133,14 @@ func doRequest(t *testing.T, endpoint, query string) *http.Response {
 	return res
 }
 
-func buildURL(endpoint, query string) string {
+func buildURL(endpoint, query string, exemplars int) string {
```
ruslan-mikhailov, line 136: Note to myself: if this PR https://github.com/grafana/tempo/pull/4962 is merged sooner, need to rebase and resolve conficts

modules/frontend/combiner/metrics_query_range.go
```diff
@@ -179,8 +179,13 @@ func diffResponse(prev, curr *tempopb.QueryRangeResponse) *tempopb.QueryRangeRes
 // have NaNs, and we can't attach them until the very end.
 func attachExemplars(req *tempopb.QueryRangeRequest, res *tempopb.QueryRangeResponse) {
 	for _, ss := range res.Series {
+		if ss == nil {
+			continue
+		}
+		if len(ss.Exemplars) > int(req.Exemplars) {
```
mdisibio, line 185: What is the condition that leads to a series having more exemplars than requested, is there maybe a bug upstream that could be addressed instead?

ruslan-mikhailov: From my understanding, query frontend tries to predict number of exemplars per page here: https://github.com/grafana/tempo/blob/main/modules/frontend/metrics_query_range_sharder.go#L224-L241 , https://github.com/grafana/tempo/blob/main/modules/frontend/metrics_query_range_sharder.go#L161-L165
As we cannot predict the exact number of traces we will find, there could be situations where we have more traces than needed or less.

ruslan-mikhailov:
\+ rebase from latest master to resolve conflicts

modules/frontend/combiner/metrics_query_range.go
mdisibio: This is operating per-series, but I think the request parameter is intended to be global.  Wondering if this enforcement is really needed, and what the consequence of returning too many exemplars is, other than to satisfy tests (which could be written differently to handle more exemplars).    At the minimum we should move this in it's own function and clarify with a comment how we are prioritizing the most recent exemplars.

mapno: >This is operating per-series, but I think the request parameter is intended to be global.

Agree. Otherwise it'd be too difficult to tune effectively. Distributing exemplars across series might be difficult though.

>At the minimum we should move this in it's own function and clarify with a comment how we are prioritizing the most recent exemplars.

Are we doing this?

mdisibio: > Are we doing this?

Sort of, by the line `ss.Exemplars = ss.Exemplars[len(ss.Exemplars)-int(req.Exemplars):]` it is limiting to the last N exemplars.  I think this should be the most recent ones by job ordering. Either way we should clarify the intent with a comment for when we want to tune this behavior later.

ruslan-mikhailov:
\+ review fixes

ruslan-mikhailov:
\+ rebase from latest main

modules/frontend/combiner/metrics_query_range.go
ruslan-mikhailov: If the parameter is a global limit, I see 2 ways on how to limit time series. Let me try and benchmark both approaches, and then we can decide which one is better.

#### Approach 1
Straightforward approach. Much easier to implement.

```go
		{
			input: [][]int{
				{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
				{11, 12, 13, 14, 15},
				{16, 17, 18, 19, 20, 21, 22, 23, 24, 25},
			},
			limit: 15,
			expected: [][]int{
				{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
				{11, 12, 13, 14, 15},
				{},
			},
		},
```

#### Approach 2

A fairer approach, trying to distribute exemplars evenly. It is better from UX standpoint, but it may affect performance.

```go
		{
			input: [][]int{
				{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
				{11, 12, 13, 14, 15},
				{16, 17, 18, 19, 20, 21, 22, 23, 24, 25},
			},
			limit: 15,
			expected: [][]int{
				{6, 7, 8, 9, 10},
				{11, 12, 13, 14, 15},
				{21, 22, 23, 24, 25},
			},
		},
```

ruslan-mikhailov: I've pushed these two versions for the history.
Approach one: https://github.com/grafana/tempo/pull/4976/commits/34ecfc49368b5c88505bc46751e014a5ddc8f595
Approach two: https://github.com/grafana/tempo/pull/4976/commits/9ca85ac219f3742d5eb64e060f1764db4076d267

Benchmarks:
First version:
```
goos: darwin
goarch: arm64
pkg: github.com/grafana/tempo/modules/frontend/combiner
cpu: Apple M4 Pro
BenchmarkAttachExemplars
BenchmarkAttachExemplars-12    	23190048	        52.30 ns/op	       0 B/op	       0 allocs/op
```

Second version:
```
goos: darwin
goarch: arm64
pkg: github.com/grafana/tempo/modules/frontend/combiner
cpu: Apple M4 Pro
BenchmarkAttachExemplars
BenchmarkAttachExemplars-12    	17842572	        66.69 ns/op	      80 B/op	       1 allocs/op
```

According to pprof, the second approach has ~10% (~5 ns) overhead on creating a slice with a total ~15 ns difference.

ruslan-mikhailov:
I think we've reached consensus. As the parameter's main goal is to not overheat Tempo, in order to not bring additional complexity, we are going to change the documentation for Exemplar to state the parameter's meaning and its behavior that the returned number of exemplars is not guaranteed. Therefore, I close the PR

PR #4978, opened 2025-04-09
metrics: cleanup duplicate code in QueryRange endpoint

electron0zero:
**What this PR does**:

noticed this duplicate code while working on https://github.com/grafana/tempo/pull/4646, so sending this standalone PR to cleanup this code.

no change log entry as this is an internal cleanup, just made it's own PR to keep https://github.com/grafana/tempo/pull/4646 clean.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/querier/querier_query_range.go
+++ b/modules/querier/querier_query_range.go
@@ -3,14 +3,12 @@ package querier
 import (
 	"context"
 	"fmt"
-	"math"
 	"time"
 
 	"github.com/go-kit/log/level"
 	"github.com/grafana/dskit/ring"
 	"github.com/grafana/dskit/user"
 	"github.com/grafana/tempo/pkg/tempopb"
-	v1 "github.com/grafana/tempo/pkg/tempopb/common/v1"
 	"github.com/grafana/tempo/pkg/traceql"
 	"github.com/grafana/tempo/pkg/util/log"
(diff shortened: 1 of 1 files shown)


modules/querier/querier_query_range.go
```diff
@@ -122,67 +121,10 @@ func (q *Querier) queryBlock(ctx context.Context, req *tempopb.QueryRangeRequest
 	inspectedBytes, spansTotal, _ := eval.Metrics()
 
 	return &tempopb.QueryRangeResponse{
-		Series: queryRangeTraceQLToProto(res, req),
+		Series: res.ToProto(req),
```
joe-elliott, line 124: there are subtle differences between these two methods. are we sure they are equivalent?

electron0zero: yeah, me and @mdisibio reviewed them today and res.ToProto is just better version of `queryRangeTraceQLToProto`.

res.ToProto has some extra checks and allocates less, other then that the logic is same.

feel free to audit it in case we missed something: https://github.com/grafana/tempo/blob/68786618c9bd6f986ce494d35b9c1dc84eaf5640/pkg/traceql/engine_metrics.go#L233-L304

joe-elliott (approved):

PR #4979, opened 2025-04-09
Add support for pprof mutex/blocking

mattdurham:
**What this PR does**:

This adds default value for blocking and mutex contention. They are very low defaults to have minimal impact on cpu. In practice alloy has been running these same defaults for years. There is no documentation, I will add that in a separate PR since it will be a larger new section on starting tempo.

**Checklist**
- [x] Tests updated
- [ ] Documentation added
- [x]  `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/cmd/tempo/main.go
+++ b/cmd/tempo/main.go
@@ -58,8 +58,8 @@ func init() {
 func main() {
 	printVersion := flag.Bool("version", false, "Print this builds version information")
 	ballastMBs := flag.Int("mem-ballast-size-mbs", 0, "Size of memory ballast to allocate in MBs.")
-	mutexProfileFraction := flag.Int("mutex-profile-fraction", 0, "Enable mutex profiling.")
-	blockProfileThreshold := flag.Int("block-profile-threshold", 0, "Enable block profiling.")
+	mutexProfileFraction := flag.Int("mutex-profile-fraction", 0, "Override default mutex profiling fraction.")
+	blockProfileThreshold := flag.Int("block-profile-threshold", 0, "Override default block profiling threshold.")
 
 	config, configVerify, err := loadConfig()
 	if err != nil {
--- a/cmd/tempo/main.go
+++ b/cmd/tempo/main.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"reflect"
 	"runtime"
+	"strconv"
 	"time"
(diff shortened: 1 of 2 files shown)


cmd/tempo/main.go
```diff
@@ -283,3 +282,35 @@ type otelErrorHandlerFunc func(error)
...
+			runtime.SetMutexProfileFraction(100 / rate)
+		} else {
+			// This is 1/1000 sampling rate.
+			runtime.SetMutexProfileFraction(1000)
+		}
+	} else {
+		// Why 1000 because that is what istio defaults to and that seemed reasonable to start with.
+		runtime.SetMutexProfileFraction(1000)
+	}
+	blockRate := os.Getenv("PPROF_BLOCK_PROFILING_RATE")
```
mattdurham, line 304: Need to add a command line flag so its consistent

cmd/tempo/main.go
```diff
@@ -283,3 +282,35 @@ type otelErrorHandlerFunc func(error)
 func (f otelErrorHandlerFunc) Handle(err error) {
 	f(err)
 }
+
+func setMutexBlockProfiling(mutexFraction int) {
+	mutexPercent := os.Getenv("PPROF_MUTEX_PROFILING_PERCENT")
```
joe-elliott, line 287: search the codebase for `intFromEnv` perhaps we consolidate this logic?

mattdurham: I am going to remove the env vars and add a new command flag since one already exists for mutex.

cmd/tempo/main.go
```diff
@@ -59,6 +59,7 @@ func main() {
 	printVersion := flag.Bool("version", false, "Print this builds version information")
 	ballastMBs := flag.Int("mem-ballast-size-mbs", 0, "Size of memory ballast to allocate in MBs.")
 	mutexProfileFraction := flag.Int("mutex-profile-fraction", 0, "Enable mutex profiling.")
+	blockProfileThreshold := flag.Int("block-profile-threshold", 0, "Enable block profiling.")
```
mattdurham, line 62: I should change this description.

PR #4986, opened 2025-04-11
Improve rate limit error message for traces size exceeds rate limit

electron0zero:
**What this PR does**:

I saw errors that reads as  `pusher failed to consume trace data" err="rpc error: code = ResourceExhausted desc = RATE_LIMITED: ingestion rate limit (local: 50000 bytes, global: 500000 bytes) exceeded while adding 966953 bytes for user xxxxxx"`

This error message had me confused because the ingestion rate for this user was way under the ingestion rate limit.

After some investigation I noticed that this push failed because the batch size of **966953** bytes is more then the global limit of **500000** bytes (my cluster is using `global` ingestion rate limiting strategy)

This PR improves the error message when the push failure is due to batch size being more then rate limit and not because user is hitting the ingestion rate limit.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/distributor.go
+++ b/modules/distributor/distributor.go
@@ -393,12 +393,12 @@ func (d *Distributor) checkForRateLimits(tracesSize, spanCount int, userID strin
 		// batch size is too big if it's more than the limit and burst both
 		if tracesSize > limit && tracesSize > burst {
 			return status.Errorf(codes.ResourceExhausted,
-				"%s: batch size (%d bytes) exceeds ingestion limit (local: %d bytes/s, global: %d bytes/s, burst: %d bytes/s) while adding %d bytes for user %s. Consider reducing batch size or increasing rate limit.",
+				"%s: batch size (%d bytes) exceeds ingestion limit (local: %d bytes/s, global: %d bytes/s, burst: %d bytes) while adding %d bytes for user %s. consider reducing batch size or increasing rate limit.",
 				overrides.ErrorPrefixRateLimited, tracesSize, limit, globalLimit, burst, tracesSize, userID)
 		}
 
 		return status.Errorf(codes.ResourceExhausted,
-			"%s: ingestion rate limit (local: %d bytes/s, global: %d bytes/s, burst: %d bytes/s) exceeded while adding %d bytes for user %s. Consider increasing the limit or reducing ingestion rate.",
+			"%s: ingestion rate limit (local: %d bytes/s, global: %d bytes/s, burst: %d bytes) exceeded while adding %d bytes for user %s. consider increasing the limit or reducing ingestion rate.",
 			overrides.ErrorPrefixRateLimited, limit, globalLimit, burst, tracesSize, userID)
 	}
 
--- a/modules/distributor/distributor_test.go
+++ b/modules/distributor/distributor_test.go
@@ -2066,7 +2066,7 @@ func TestCheckForRateLimits(t *testing.T) {
 			tracesSize:      1100,
 			rateLimitBytes:  500,
 			burstLimitBytes: 500,
-			expectError:     "RATE_LIMITED: batch size (1100 bytes) exceeds ingestion limit (local: 500 bytes/s, global: 0 bytes/s, burst: 500 bytes/s) while adding 1100 bytes for user test-user. Consider reducing batch size or increasing rate limit.",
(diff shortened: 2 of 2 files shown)


modules/distributor/distributor.go
```diff
@@ -374,6 +374,16 @@ func (d *Distributor) checkForRateLimits(tracesSize, spanCount int, userID strin
 		if d.overrides.IngestionRateStrategy() == overrides.GlobalIngestionRateStrategy {
 			globalLimit = limit * d.DistributorRing.InstancesCount()
 		}
+
+		if tracesSize > limit || tracesSize > globalLimit {
```
joe-elliott, line 378: does this mean that their batch size is too large? should we advise they reduce it?

joe-elliott: i think it's fine to give more details here. i don't find the new error message any more useful than the old one. you could mention that a single batch exceeded the rate limit.

also, i don't think this check is correct. the push is limited if it exceeds local limit or global limit depending on how tempo is configured. the local limit is just the limit. the global limit is limit / numDistributors. the logic is bit tough to unwind but you can see some if it in ingestion_rate_strategy.go.

mdisibio: I'm wondering if what we need is to include the remaining burst size in the error message.  I think that is an area that is unclear, because a push can exceed the local limit as a burst. But not when it's consistent.

electron0zero: updated to add some more tests and comments to better explain the local and global ingestion rate strategy.

> also, i don't think this check is correct. the push is limited if it exceeds local limit or global limit depending on how tempo is configured. the local limit is just the limit. the global limit is limit / numDistributors. the logic is bit tough to unwind but you can see some if it in ingestion_rate_strategy.go.

limit is fetched from ingestionRateLimiter so it will be the local or global limit based on the configured IngestionRateStrategy.

I also updated the check to not include globalLimit because that's not correct, instead we should be checking on limit AND burst to only return the batch size error when the batch is bigger then both limit.

if the batch size is more then limit but less then burst, it would allowed by burst and we should return the ingestion rate limit instead of batch size error.

modules/distributor/distributor.go
```diff
@@ -364,21 +364,42 @@ func (d *Distributor) stopping(_ error) error {
...
 		if d.overrides.IngestionRateStrategy() == overrides.GlobalIngestionRateStrategy {
+			// note: global limit should be calculated using healthy distributors count,
+			// but we are using it in logs, instance count is good enough.
 			globalLimit = limit * d.DistributorRing.InstancesCount()
 		}
+
+		// batch size is too big if it's more than the limit and burst both
+		if tracesSize > limit && tracesSize > burst {
+			return status.Errorf(codes.ResourceExhausted,
+				"%s: batch size (%d bytes) exceeds ingestion limit (local: %d bytes/s, global: %d bytes/s, burst: %d bytes/s) while adding %d bytes for user %s. Consider reducing batch size or increasing rate limit.",
```
mdisibio, line 396: nit: burst is total bytes, not per second `burst: %d bytes`.  Message here and below

modules/distributor/distributor.go
```diff
@@ -364,21 +364,42 @@ func (d *Distributor) stopping(_ error) error {
...
+		// globalLimit will be 0 when using local ingestion rate strategy
 		var globalLimit int
 		if d.overrides.IngestionRateStrategy() == overrides.GlobalIngestionRateStrategy {
+			// note: global limit should be calculated using healthy distributors count,
+			// but we are using it in logs, instance count is good enough.
 			globalLimit = limit * d.DistributorRing.InstancesCount()
 		}
+
+		// batch size is too big if it's more than the limit and burst both
+		if tracesSize > limit && tracesSize > burst {
```
zalegrala, line 394: Is burst always greater than limit?  If so, we can just check burst here, correct?

modules/distributor/distributor.go
electron0zero: fixed, now burst is logged as `burst: %d bytes`

zalegrala (approved):
Looks reasonable to me.

modules/distributor/distributor.go
electron0zero: I see a case where burst can be lower then the limit in case you are using global rate limit strategy

limit: 1000, burst 100, but you have 20 distributors but if you are using global limit so per instance limit is 1000/20 = 50, 

but as a rule of thumb per instance limit is usually same or lower then burst.

PR #4997, opened 2025-04-14
Refactor endpoint metrics collection with cache

carles-grafana:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/frontend/pipeline/sync_handler_cache.go
```diff
@@ -100,6 +134,120 @@ func (c cachingWare) RoundTrip(req Request) (*http.Response, error) {
 	return resp, nil
 }
 
+func resetMetricsInResponse(body []byte, responseType CacheableRespType, contentType string) ([]byte, error) {
```
joe-elliott, line 137: why don't we move this code back into the combiners? this requires a double unmarshal of all responses.

modules/frontend/combiner/search.go
```diff
@@ -61,8 +63,10 @@ func NewSearch(limit int, keepMostRecent bool) Combiner {
 			}
 
 			if partial.Metrics != nil {
+				if !cache.IsCacheHit(resp.HTTPResponse()) {
+					inspectedBytes += partial.Metrics.InspectedBytes
+				}
 				final.Metrics.CompletedJobs++
-				final.Metrics.InspectedBytes += partial.Metrics.InspectedBytes
 				final.Metrics.InspectedTraces += partial.Metrics.InspectedTraces
```
carles-grafana, line 70: not sure if we want to reset all metrics

modules/frontend/pipeline/sync_handler_cache.go
carles-grafana: @joe-elliott  I've pushed a new proposed which checks the cache header in the combiner, please check

pkg/cache/header.go
```diff
@@ -0,0 +1,18 @@
+package cache
```
joe-elliott, line 1: these headers are produced and consumed entirely within the frontend. i think they belong there.

modules/frontend/pipeline/sync_handler_cache.go
```diff
@@ -39,30 +39,42 @@ func (c cachingWare) RoundTrip(req Request) (*http.Response, error) {
...
+
+			resp := &http.Response{
+				Header:        http.Header{},
+				StatusCode:    http.StatusOK,
+				Status:        http.StatusText(http.StatusOK),
+				Body:          io.NopCloser(bytes.NewBuffer(body)),
+				ContentLength: int64(len(body)),
 			}
+
+			resp.Header.Add(api.HeaderContentType, contentType)
```
joe-elliott, line 63: nit: these could be declared inline when you create `http.Header{}` on line 56

it might preallocate the map nicely and reduce allocs?

modules/frontend/pipeline/sync_handler_cache.go
```diff
@@ -39,30 +39,42 @@ func (c cachingWare) RoundTrip(req Request) (*http.Response, error) {
...
+			// Determine content type based on first byte
+			var contentType string
 			// TODO - Cache should capture all of the relevant parts of the
 			// original response including both content-type and content-length headers, possibly more.
 			// But upgrading the cache format requires migration/detection of previous format either way.
 			// It's tempting to use https://pkg.go.dev/net/http#DetectContentType but it doesn't detect
 			// json or proto.
 			if body[0] == '{' {
-				resp.Header.Add(api.HeaderContentType, api.HeaderAcceptJSON)
+				contentType = api.HeaderAcceptJSON
```
joe-elliott, line 50: please add a test to lock down this behavior of adding the cache header

modules/frontend/combiner/search.go
```diff
@@ -61,8 +63,10 @@ func NewSearch(limit int, keepMostRecent bool) Combiner {
 			}
 
 			if partial.Metrics != nil {
+				if !cache.IsCacheHit(resp.HTTPResponse()) {
```
joe-elliott, line 66: there's a lot of opportunity for cleanup and consolidation here. there are 3 metrics objects: TraceByIDMetrics, SearchMetrics, and MetadataMetrics. They are all handled a bit differently.

Why don't we have a set of combiner structs for these metadata objects where we have the code to accumulate them? One for each type. Then we can make all the logic in these combiners feel very consistent in how they combine the actual data and the metadata.

modules/frontend/pipeline/sync_handler_cache.go
carles-grafana: but how? The Add method does some transformations

pkg/cache/header.go
carles-grafana: moved

modules/frontend/combiner/search.go
carles-grafana: I've pushed a new proposal to combine the response metrics

modules/frontend/combiner/response_metrics.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+}
+
+func (mc *QueryRangeMetricsCombiner) Combine(newMetrics *tempopb.SearchMetrics, resp PipelineResponse) {
+	if newMetrics != nil && !IsCacheHit(resp.HTTPResponse()) {
+		// this is a coordination between the sharder and combiner. the sharder returns one response with summary metrics
+		// only. the combiner correctly takes and accumulates that job. however, if the response has no jobs this is
+		// an indicator this is a "real" response so we set CompletedJobs to 1 to increment in the combiner.
+		if newMetrics.TotalJobs == 0 {
+			newMetrics.CompletedJobs = 1
```
carles-grafana, line 75: do this even with cache hit

pkg/traceql/combine.go
```diff
@@ -358,21 +356,11 @@ func (q *QueryRangeCombiner) Combine(resp *tempopb.QueryRangeResponse) {
 		q.maxSeriesReached = true
 	}
 
-	if resp.Metrics != nil {
```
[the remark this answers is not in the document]
carles-grafana: this is also used in the querier, I suppose it's needed there to collect metrics in the querier jobs, so I'm reverting it.

modules/frontend/pipeline/sync_handler_cache.go
```diff
@@ -100,6 +102,18 @@ func (c cachingWare) RoundTrip(req Request) (*http.Response, error) {
 	return resp, nil
 }
 
+func determineContentType(body []byte) string {
```
joe-elliott, line 105: +1 love improvements like this with the test

pkg/traceql/combine.go
```diff
@@ -358,6 +358,8 @@ func (q *QueryRangeCombiner) Combine(resp *tempopb.QueryRangeResponse) {
 		q.maxSeriesReached = true
 	}
 
+	// TODO: this is here only for the querier
```
joe-elliott, line 361: long term we need to use the same metric combiner objects everywhere. i don't like that the metrics accumulation logic is still here and is now duplicated in

perhaps something to follow up on?

modules/frontend/combiner/response_metrics.go
```diff
@@ -0,0 +1,95 @@
...
+	return &SearchMetricsCombiner{
+		Metrics: &tempopb.SearchMetrics{},
+	}
+}
+
+func (mc *SearchMetricsCombiner) Combine(newMetrics *tempopb.SearchMetrics, resp PipelineResponse) {
+	if newMetrics != nil && !IsCacheHit(resp.HTTPResponse()) {
+		mc.Metrics.InspectedTraces += newMetrics.InspectedTraces
+		mc.Metrics.InspectedBytes += newMetrics.InspectedBytes
+		mc.Metrics.CompletedJobs++
```
joe-elliott, line 23: i think we should always increment this. otherwise these values might not line up in the frontend.

image

modules/frontend/combiner/response_metrics.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+}
+
+func (mc *SearchMetricsCombiner) CombineMetadata(newMetrics *tempopb.SearchMetrics, resp PipelineResponse) {
+	if newMetrics != nil && !IsCacheHit(resp.HTTPResponse()) {
+		mc.Metrics.TotalBlocks += newMetrics.TotalBlocks
```
joe-elliott, line 29: what is the meaning of the "total" fields? should these be incremented regardless? are they conceptually different than the "inspected" fields?

modules/frontend/combiner/response_metrics.go
```diff
@@ -0,0 +1,95 @@
...
+		// only. the combiner correctly takes and accumulates that job. however, if the response has no jobs this is
+		// an indicator this is a "real" response so we set CompletedJobs to 1 to increment in the combiner.
+		if newMetrics.TotalJobs == 0 {
+			newMetrics.CompletedJobs = 1
+			mc.Metrics.CompletedJobs += newMetrics.CompletedJobs
+		}
+		if !IsCacheHit(resp.HTTPResponse()) {
+			mc.Metrics.TotalJobs += newMetrics.TotalJobs
+			mc.Metrics.TotalBlocks += newMetrics.TotalBlocks
+			mc.Metrics.TotalBlockBytes += newMetrics.TotalBlockBytes
```
joe-elliott, line 89: same question here about total vs inspected.

modules/frontend/combiner/response_metrics.go
```diff
@@ -0,0 +1,95 @@
...
+
+func NewQueryRangeMetricsCombiner() *QueryRangeMetricsCombiner {
+	return &QueryRangeMetricsCombiner{
+		Metrics: &tempopb.SearchMetrics{},
+	}
+}
+
+func (mc *QueryRangeMetricsCombiner) Combine(newMetrics *tempopb.SearchMetrics, resp PipelineResponse) {
+	if newMetrics != nil {
+		// this is a coordination between the sharder and combiner. the sharder returns one response with summary metrics
```
joe-elliott, line 79: this could be removed using the same "meta data" pattern that the search pipeline uses. perhaps something you can tackle in a future PR?

modules/frontend/combiner/response_metrics.go
carles-grafana: Looking at the code, https://github.com/carles-grafana/tempo/blob/refactor-endpoint-metrics-collection-with-cache/modules/frontend/metrics_query_range_sharder.go#L215, TotalBlockBytes is set in the sharder and calculates the total size of blocks that are searched, while InspectedBytes are set in the querier as bytes read from Parquet files: https://github.com/carles-grafana/tempo/blob/refactor-endpoint-metrics-collection-with-cache/tempodb/encoding/vparquet4/block_search.go#L112

I don't think we use the Block metrics other than logging the, while we use the InspectedBytes for the throughput SLO exposed as a metric.

joe-elliott (approved):
can we leave off the linter change for now and address that separately?

regardless, lgtm!

joe-elliott (commented):
changelog?

.golangci.yml
```diff
@@ -29,6 +29,9 @@ linters:
         - (github.com/go-kit/log.Logger).Log
     errorlint:
       errorf: false
+    gosec:
```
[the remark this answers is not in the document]
carles-grafana: I've ignored instead the relevant lines, as otherwise I would need to change a lot of types. The problems stem from the protobufs definitions being uint32 and some conversion from `len`, which always returns an int.

PR #5005, opened 2025-04-15
[Rhythm] Fix ingester to correctly delete old blocks even if not flushing to object storage

mdisibio:
**What this PR does**:
Fixes a bug in #4857 where the ingester doesn't delete blocks off disk when flush_to_storage=false.   Additionally, updated the time comparison to look at the block meta to determine the age of the block.  Previously it was based on flushed time, which meant that for a retention 1h it wouldn't delete the block off disk until 1h after it was flushed to object storage.  

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/ingester/instance.go
+++ b/modules/ingester/instance.go
@@ -376,27 +376,33 @@ func (i *instance) GetBlockToBeFlushed(blockID uuid.UUID) *LocalBlock {
 	return nil
 }
 
-func (i *instance) ClearFlushedBlocks(completeBlockTimeout time.Duration) error {
+func (i *instance) ClearOldBlocks(flushObjectStorage bool, completeBlockTimeout time.Duration) error {
 	var err error
 
 	i.blocksMtx.Lock()
 	defer i.blocksMtx.Unlock()
 
+	cutoff := time.Now().Add(-completeBlockTimeout)
+
 	for idx, b := range i.completeBlocks {
-		flushedTime := b.FlushedTime()
-		if flushedTime.IsZero() {
+		// If we are flushing to object storage, never delete blocks that have not been flushed.
+		if flushObjectStorage && b.FlushedTime().IsZero() {
 			continue
 		}
 
-		if flushedTime.Add(completeBlockTimeout).Before(time.Now()) {
-			i.completeBlocks = append(i.completeBlocks[:idx], i.completeBlocks[idx+1:]...)
-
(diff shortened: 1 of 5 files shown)


modules/ingester/instance.go
```diff
@@ -376,27 +376,33 @@ func (i *instance) GetBlockToBeFlushed(blockID uuid.UUID) *LocalBlock {
...
+			continue
+		}
 
-			err = i.local.ClearBlock((uuid.UUID)(b.BlockMeta().BlockID), i.instanceID)
-			if err == nil {
-				metricBlocksClearedTotal.Inc()
-			}
-			break
+		// This block can be deleted.
+		i.completeBlocks = append(i.completeBlocks[:idx], i.completeBlocks[idx+1:]...)
```
javiermolinar, line 399: Should this line be after the clearBlock try? If we remove the id from the list of complete blocks and then we get an error, the block will be dangling forever

mdisibio: Good catch. I guess that has been an issue for a long time. Also, the previous code was only deleting 1 block per pass, and I attemped to delete all matching blocks. This was broken and needed some other updates because we are modifying the slice while iterating it. Thoughts?  There are several ways to do this, and none are particularly clear.

modules/ingester/instance.go
```diff
@@ -376,30 +376,36 @@ func (i *instance) GetBlockToBeFlushed(blockID uuid.UUID) *LocalBlock {
...
 
-			err = i.local.ClearBlock((uuid.UUID)(b.BlockMeta().BlockID), i.instanceID)
-			if err == nil {
-				metricBlocksClearedTotal.Inc()
-			}
-			break
+		// This block can be deleted.
+		err := i.local.ClearBlock((uuid.UUID)(b.BlockMeta().BlockID), i.instanceID)
+		if err != nil {
+			return err
```
mattdurham, line 400: Does this mean a bad block will prevent any flush from ever happening?

mdisibio: True, if it failed repeatedly it would mean that no blocks later in the slice would be deleted off the local disk, if the sliced maintained the same order (I think it does).  Flushing to object storage and persisting data off the ingester wouldn't be affected.  Can say that I've never seen this happen in 4 years of operations, so I'm ok with this as-is.

mattdurham: Sounds reasonable!

joe-elliott (approved):
a glorious act of software development

PR #5007, opened 2025-04-15
TraceQL Bugfix: Disambiguate existence from equality

joe-elliott:
### The problem

Currently the TraceQL engine treats all != comparisons to nil as existence checks (as opposed to equality). This is different from the fetch layer which will not collect spans where an attribute is not present. For example consider:

```
{ span.foo != "bar" } 
```

If foo is not present on a span the fetch layer will not return it. In this case the fetch layer is effectively evaluating `nil != "bar"` as false. However, the engine will currently evaluate `nil != "bar"` as true. This is apparent when running queries like:

```
{} >> { span.foo != "bar" }
```

In this case b/c the engine is evaluating the condition and not the fetch layer, Tempo will return spans that do not have an attribute "foo" unlike the first example.

### The fix

This PR disambiguates the concept of existence from equality in the engine. This allows it to have one set of behaviors for existence and a different set of behaviors (consistent with the fetch layer) for equality. This is accomplished by rewriting the BinaryOperations `!= nil` and `nil !=` as a Unary "existence" operation. It might be possible to parse these directly in the grammar as well. Looking for feedback.

**Note:** As is, this PR further cements the following equality behaviors w/r to nil in TraceQL. 

value = nil -> false
value != nil -> false
nil != nil -> false
nil = nil -> true

Should nil = nil be true or false? Should this match spans?

`{ span.does-not-exist-attr = span.does-not-exist-attr-2 }`

### Checklist
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/ast.go
+++ b/pkg/traceql/ast.go
@@ -491,12 +491,12 @@ func newBinaryOperation(op Operator, lhs, rhs FieldExpression) FieldExpression {
 		}
 	}
 
-	// rewrite != nil and nil != to be just existence
-	if binop.Op == OpNotEqual && !binop.LHS.referencesSpan() && binop.LHS.impliedType() == TypeNil { // jpe better nil check?
-		return newUnaryOperation(OpNotExists, binop.RHS)
+	// rewrite != nil and nil != to be a unary existence
+	if binop.Op == OpNotEqual && !binop.LHS.referencesSpan() && binop.LHS.impliedType() == TypeNil {
+		return newUnaryOperation(OpExists, binop.RHS)
 	}
-	if binop.Op == OpNotEqual && !binop.RHS.referencesSpan() && binop.RHS.impliedType() == TypeNil { // jpe better nil check?
-		return newUnaryOperation(OpNotExists, binop.LHS)
+	if binop.Op == OpNotEqual && !binop.RHS.referencesSpan() && binop.RHS.impliedType() == TypeNil {
+		return newUnaryOperation(OpExists, binop.LHS)
 	}
 
 	if (op == OpAnd || op == OpOr) && binop.referencesSpan() {
@@ -553,11 +553,11 @@ func newUnaryOperation(op Operator, e FieldExpression) FieldExpression {
 func (UnaryOperation) __fieldExpression() {}
 
 func (o UnaryOperation) impliedType() StaticType {
-	if o.Op == OpNotExists {
+	if o.Op == OpExists {
 		return TypeBoolean
 	}
 
-	// both operators (opPower and opNot) will just be based on the operand type
+	// opPower and opNot will just be based on the operand type
 	return o.Expression.impliedType()
 }
 
@@ -780,7 +780,7 @@ func (s Static) Equals(o *Static) bool {
 	}
 }
 
-func (s Static) NotEquals(o *Static) bool { // jpe - test
+func (s Static) NotEquals(o *Static) bool {
 	if s.Type == TypeNil || o.Type == TypeNil {
 		return false
 	}
--- a/pkg/traceql/ast.go
+++ b/pkg/traceql/ast.go
@@ -727,10 +727,6 @@ func (s Static) MapKey() StaticMapKey {
 }
 
 func (s Static) Equals(o *Static) bool {
-	if s.Type == TypeNil && o.Type == TypeNil {
-		return true
-	}
-
 	// if one is nil, they are not equal
 	if s.Type == TypeNil || o.Type == TypeNil {
 		return false
--- a/pkg/traceql/ast.go
+++ b/pkg/traceql/ast.go
@@ -484,10 +484,19 @@ func newBinaryOperation(op Operator, lhs, rhs FieldExpression) FieldExpression {
 		RHS: rhs,
 	}
 
+	// AST rewrite for simplification
 	if !binop.referencesSpan() && binop.validate() == nil {
 		if simplified, err := binop.execute(nil); err == nil {
 			return simplified
 		}
+	}
+
+	// rewrite != nil and nil != to be just existence
+	if binop.Op == OpNotEqual && !binop.LHS.referencesSpan() && binop.LHS.impliedType() == TypeNil { // jpe better nil check?
+		return newUnaryOperation(OpNotExists, binop.RHS)
+	}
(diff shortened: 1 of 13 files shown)


mdisibio:
Overall, I am supportive of this change because it is fixing the obviously broken query `{} >> { span.foo != "bar" }`, and there looks to be no impact to any others, so it is a safe change.

Stepping back through things and thinking out loud:

**1. Fetch existence**
`{ span.foo != "bar" }`   Instead let's compare to something more common:  

`{span.http.status_code != 200}`  This only makes sense to return spans with the http status code attribute.  It would not make sense to return every span unrelated to http traffic. Therefore this behavior seems "correct".  

**2. Engine**
I think this is where the difference occurs:
 
```go
func (a Attribute) execute(span Span) (Static, error) {
	static, ok := span.AttributeFor(a)
	if ok {
		return static, nil
	}

	return StaticNil, nil
}
```

The engine lookup for the attribute is using Nil to mean not-present, but Nil is a real value that is comparable. 

The parent BinOp sees `nil != 200` and evaluated this as true, because it is treating Nil as a normal value and not as a failed AttributeOp.  

**Conclusion**
Yes I agree what we need is for the engine to evaluate `<failed attribute lookup> != 200` as false.  Special case handling of nil comparisons works, which is what was done here.    An alternative idea, not necessarily a good one, is BinOp could have special handling for when LHS/RHS are AttributeOps that return Nil.

So yes overall supportive, but I want to dig more into the repercussions of `nil = nil`, etc.

joe-elliott:
I was thinking about this issue as conflating two different operations: existence and equality. @mdisibio's comments have me thinking about it as conflating two different types: no value in parquet (TypeDNE) and static value in the language (TypeNil).

I've pushed a final commit that I like. It removes the concept of "static value in the language" by only parsing NIL in prescribed ways to test existence. This collapses all TypeNil values to mean the same thing (no value in parquet) which simplifies all other code. It also tightly controls the valid uses of nil as a static in the language which I like.

I prefer this to adding a new value returned from AttributeFor that means "does not exist". If we add a new type then the engine code will need special handling for both TypeNil and TypeDNE throughout the codebase.

As written this does change the behavior of the query `{ nil = nil }` which feels ok to me. This used to return everything and now it returns nothing to be consistent with `{ span.DNE1 = span.DNE2 }`.

electron0zero:
> This collapses all TypeNil values to mean the same thing (no value in parquet) which simplifies all other code. It also tightly controls the valid uses of nil as a static in the language which I like.

I also like this

> The complexity of this addition would not be worth the value imo.

agree, not worth it.

pkg/traceql/parse_test.go
```diff
@@ -924,8 +923,7 @@ func TestSpansetFilterOperators(t *testing.T) {
 		{in: "{ .a = true }", expected: newBinaryOperation(OpEqual, NewAttribute("a"), NewStaticBool(true)), alsoTestWithoutSpace: true},
```
electron0zero, line 923: let's also add test case for `NIL NEQ NIL` and  `NIL EQL NIL`?

pkg/traceql/test_examples.yaml
```diff
@@ -237,7 +237,11 @@ parse_fails:
   - '{} | bottomk(10) with(sample=0.1)'
   - '{} | rate() | topk(-1)'
   - '{} | rate() | bottomk(-1)'
-
+  - '{ parent = nil }'
```
electron0zero, line 240: let's add valid nil queries under valid:?, we can add `{ nil = nil }, { nil != nil }, { span:name != nil }, and { nil != span:name }`.

pkg/traceql/ast.go
```diff
@@ -721,6 +727,15 @@ func (s Static) MapKey() StaticMapKey {
...
 func (s Static) Equals(o *Static) bool {
+	if s.Type == TypeNil && o.Type == TypeNil {
+		return true
+	}
+
+	// if one is nil, they are not equal
+	if s.Type == TypeNil || o.Type == TypeNil {
+		return false
+	}
+
```
electron0zero, line 738: just to check my understanding, is this correct?

`{ nil = nil }` -> true
`{ span.foo = nil }` -> false
`{ nil = span.foo }` -> false
`{ nil != nil }` -> false
`{ span.food != nil }` -> false
`{ nil != span.food }` -> false

and now TypeNil only comes from the fetch layer and can't be the Nil from the query because we translate that into `OpExists`.

pkg/traceql/ast.go
```diff
@@ -751,14 +766,20 @@ func (s Static) Equals(o *Static) bool {
...
-	case TypeNil:
-		return o.Type == TypeNil
 	default:
 		// should not be reached
 		return false
 	}
 }
 
+func (s Static) NotEquals(o *Static) bool {
+	if s.Type == TypeNil || o.Type == TypeNil {
```
electron0zero, line 776: let's add a test case for this as well because we are handling it in a special way.

mdisibio (commented):
👍 I like the last change which effectively changes foo != nil into exists(foo).   It tightens down the grammar and behavior while still leaving open some possibilities for general nil handling.

It looks like it should not be affected, but could double check a metrics query can still group by a nil value?   I.e. `{ } | rate() by (.partially_exists)`. That should have a nil series for all spans that matched `{ }` but without the attribute.  Metrics use the `FastValues` maps with `StaticMapKey`, which should maintain equality checks on TypeNil.

pkg/traceql/ast.go
joe-elliott: Removing 
```
	if s.Type == TypeNil && o.Type == TypeNil {
		return true
	}
```

so that all comparisons with the value nil result in false. This will make all nil comparisons have consistent behavior in the fetch layer and the engine.

pkg/traceql/ast.go
joe-elliott: extended `TestStatic_Equals` to handle

pkg/traceql/parse_test.go
joe-elliott: done

pkg/traceql/test_examples.yaml
joe-elliott: done. good call on this one. it caught a stringer error i didn't know existed b/c i hadn't handled the OpExists in UnaryOperation.String()

electron0zero (approved):
nice change, it makes nil simple and easy to wrap my head around.

lgtm :shipit:

PR #5016, opened 2025-04-16
[bugfix] fix avg_over_time & compare queries panic with max series

ie-pham:
**What this PR does**: 
1. Max series PR introduced a bug for avg_over_time metrics queries. The metrics calculation relies on a pair of series per actual value and when the max series limit truncated the series, in some occurrences, one of the pair is no longer in the resulting time series - leading to a panic during aggregations. 
2. Truncating the series caused issues for the frontend because it expects at least one _total series per key (attribute key). This PR add different truncating logic for compare queries. 

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/combiner/metrics_query_range.go
+++ b/modules/frontend/combiner/metrics_query_range.go
@@ -58,7 +58,9 @@ func NewQueryRange(req *tempopb.QueryRangeRequest, maxSeriesLimit int) (Combiner
 			if combiner.MaxSeriesReached() {
 				// Truncating the final response because even if we bail as soon as len(resp.Series) >= maxSeries
 				// it's possible that the last response pushed us over the max series limit.
-				resp.Series = resp.Series[:maxSeries]
+				if len(resp.Series) > maxSeries {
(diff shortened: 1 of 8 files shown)


modules/frontend/combiner/metrics_query_range.go
```diff
@@ -58,7 +58,9 @@ func NewQueryRange(req *tempopb.QueryRangeRequest, maxSeriesLimit int) (Combiner
 			if combiner.MaxSeriesReached() {
 				// Truncating the final response because even if we bail as soon as len(resp.Series) >= maxSeries
 				// it's possible that the last response pushed us over the max series limit.
-				resp.Series = resp.Series[:maxSeries]
+				if len(resp.Series) > maxSeries {
```
mdisibio, line 61: Does this also need to check `&& maxSeries > 0` to handle the case of unlimited?  Not sure if there is a case where we could get `combiner.MaxSeriesReached()` and maxSeries==0.

PR #5019, opened 2025-04-17
Tighten up the compaction loops to avoid dead spots

zalegrala:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/backendscheduler/backendscheduler.go
+++ b/modules/backendscheduler/backendscheduler.go
@@ -127,9 +127,9 @@ func (s *BackendScheduler) starting(ctx context.Context) error {
 		go func(jobs <-chan *work.Job) {
 			defer wg.Done()
 
-			for {
-				var job *work.Job
+			var job *work.Job
 
+			for {
 				select {
 				case job = <-jobs:
 				case <-ctx.Done():
@@ -219,6 +219,9 @@ func (s *BackendScheduler) Next(ctx context.Context, req *tempopb.NextJobRequest
 		return resp, nil
 	}
 
+	timeoutCtx, cancel := context.WithTimeout(ctx, s.cfg.JobTimeout)
+	defer cancel()
+
 	// Try to get a job from the merged channel
 	select {
 	case j := <-s.mergedJobs:
@@ -256,7 +259,7 @@ func (s *BackendScheduler) Next(ctx context.Context, req *tempopb.NextJobRequest
 		level.Info(log.Logger).Log("msg", "assigned job to worker", "job_id", j.ID, "worker", req.WorkerId)
 
 		return resp, nil
-	default:
+	case <-timeoutCtx.Done():
 		span.SetAttributes(attribute.Int("job_q_depth", len(s.mergedJobs)))
 		metricJobsNotFound.WithLabelValues(req.WorkerId).Inc()
 
--- a/modules/backendscheduler/provider/compaction.go
+++ b/modules/backendscheduler/provider/compaction.go
@@ -98,11 +98,31 @@ func (p *CompactionProvider) Start(ctx context.Context) <-chan *work.Job {
 		)
 
 		reset := func() {
-			job = nil
-			b.Reset()
+			p.curSelector = nil
+			p.curTenant = nil
+			p.curTenantJobCount = 0
 		}
 
 		for {
+			if p.curSelector == nil {
+				if !p.prepareNextTenant(ctx) {
+					b.Wait()
+				}
+				continue
+			}
+
+			if p.curTenantJobCount >= p.cfg.MaxJobsPerTenant {
+				reset()
+				continue
+			}
+
+			job = p.createJob(ctx)
+			if job == nil {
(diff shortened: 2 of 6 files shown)


modules/backendscheduler/provider/compaction.go
```diff
@@ -263,10 +192,6 @@ func (p *CompactionProvider) getNextBlockIDs(_ context.Context) ([]string, bool)
 	}
 
 	for _, b := range toBeCompacted {
-		if p.sched.HasBlocks([]string{b.BlockID.String()}) {
```
zalegrala, line 266: Skipping blocks which are in flight is covered by the `newBlockSelector`. No need to duplicate it here.

modules/backendscheduler/work/job.go
```diff
@@ -120,13 +120,19 @@ func (j *Job) GetWorkerID() string {
 func (j *Job) OnBlock(id string) bool {
 	for _, b := range j.GetCompactionInput() {
 		if b == id {
-			return true
+			switch j.GetStatus() {
+			case tempopb.JobStatus_JOB_STATUS_RUNNING, tempopb.JobStatus_JOB_STATUS_UNSPECIFIED:
```
zalegrala, line 124: In low block throughput environments, it was enough for any status.  Observations in a higher throughput environment mean that the output blocks of level 0 don't get fed into the next compaction level 1, etc jobs, resulting in lower compaction throughput.

I believe we will only care about the running and unspecified when deciding work for future jobs as well.

modules/backendscheduler/backendscheduler.go
```diff
@@ -256,7 +259,7 @@ func (s *BackendScheduler) Next(ctx context.Context, req *tempopb.NextJobRequest
 		level.Info(log.Logger).Log("msg", "assigned job to worker", "job_id", j.ID, "worker", req.WorkerId)
 
 		return resp, nil
-	default:
+	case <-timeoutCtx.Done():
```
zalegrala, line 262: This dramatically lowered the instances of "no jobs found" results and increased compaction throughput.

modules/backendscheduler/backendscheduler.go
```diff
@@ -219,6 +219,9 @@ func (s *BackendScheduler) Next(ctx context.Context, req *tempopb.NextJobRequest
 		return resp, nil
 	}
 
+	timeoutCtx, cancel := context.WithTimeout(ctx, s.cfg.JobTimeout)
```
joe-elliott, line 222: this is a timeout imposed by the server. should the caller instead set a timeout and bail after N seconds when it has decided to give up? we have timeouts like this in our queriers and i've always wondered if it should be set caller side. fine either way

modules/backendscheduler/backendscheduler.go
joe-elliott, line 262: even if we move the timeout client side we need a line like this. the previous code did not respect the client cancelling context. i missed this in review.

modules/backendscheduler/provider/compaction.go
```diff
@@ -98,119 +98,47 @@ func (p *CompactionProvider) Start(ctx context.Context) <-chan *work.Job {
 		)
 
 		reset := func() {
-			job = nil
-			b.Reset()
+			p.curSelector = nil
```
joe-elliott, line 101: i find this code a bit confusing. for instance p.curTenantJob is reset on line 103, tested on 114 and incremented in createJob on 173 which makes it a bit tough to track how it's handled.

however, this PR is definitely moving things in the right direction. seeing some really solid simplification here. nice work. i think you're getting down to the core of it.

modules/backendscheduler/work/job.go
joe-elliott, line 124: some simple tests to lock down this behavior might be nice.

joe-elliott (commented):
some thoughts but i'll go ahead with the green check. if you make some more changes and want a second pass just let me know.

line 262 was a miss from before. we always should have had the case where we bailed on context cancelled. good catch.

joe-elliott (approved):

modules/backendscheduler/backendscheduler.go
joe-elliott: noted in my review, but this was a miss from before. good catch!

modules/backendscheduler/backendscheduler.go
zalegrala: I think I want to watch it with this on the server.  We can set a timeout on the grpc caller if it ends up being an issue.

modules/backendscheduler/provider/compaction.go
zalegrala: In a couple PRs this might get a little cleaner also. :)

modules/backendscheduler/backendscheduler.go
zalegrala: Me too.

PR #5023, opened 2025-04-18
Reduce lock contention in backendscheduler

zalegrala:
**What this PR does**:

Previous to #5019 a lock for the RPC calls was sufficient, but we only need it for the work, which has its own lock.  Here we drop the locking from the backendworker entirely to rely on the job and work mutex.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/backendscheduler/backendscheduler.go
+++ b/modules/backendscheduler/backendscheduler.go
@@ -38,8 +38,6 @@ type BackendScheduler struct {
 
 	work *work.Work
 
-	mtx sync.Mutex
-
 	reader backend.RawReader
 	writer backend.RawWriter
 
--- a/modules/backendscheduler/backendscheduler.go
+++ b/modules/backendscheduler/backendscheduler.go
@@ -186,9 +186,6 @@ func (s *BackendScheduler) stopping(_ error) error {
 
 // Next implements the BackendSchedulerServer interface.  It returns the next queued job for a worker.
 func (s *BackendScheduler) Next(ctx context.Context, req *tempopb.NextJobRequest) (*tempopb.NextJobResponse, error) {
-	s.mtx.Lock()
-	defer s.mtx.Unlock()
-
 	ctx, span := tracer.Start(ctx, "Next")
 	defer span.End()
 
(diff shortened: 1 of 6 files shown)


electron0zero (approved):

modules/backendscheduler/backendscheduler.go
```diff
@@ -186,9 +186,6 @@ func (s *BackendScheduler) stopping(_ error) error {
 
 // Next implements the BackendSchedulerServer interface.  It returns the next queued job for a worker.
 func (s *BackendScheduler) Next(ctx context.Context, req *tempopb.NextJobRequest) (*tempopb.NextJobResponse, error) {
```
joe-elliott, line 188: are these the only 2 uses of this mtx? remove from the struct?

modules/backendscheduler/cache.go
```diff
@@ -12,6 +12,9 @@ import (
 )
 
 func (s *BackendScheduler) flushWorkCache(ctx context.Context) error {
+	s.mtx.Lock()
```
joe-elliott, line 15: why add locking here? s.work seems to protect itself. as long as we're not worried about racy access to s.work i don't think we need this?

modules/backendscheduler/work/job.go
```diff
@@ -118,24 +118,28 @@ func (j *Job) GetWorkerID() string {
...
 
-	for _, b := range j.GetCompactionOutput() {
-		if b == id {
-			switch j.GetStatus() {
-			case tempopb.JobStatus_JOB_STATUS_RUNNING, tempopb.JobStatus_JOB_STATUS_UNSPECIFIED:
-				return true
+		for _, b := range j.GetCompactionOutput() {
+			if b == id {
+				switch j.GetStatus() {
+				case tempopb.JobStatus_JOB_STATUS_RUNNING, tempopb.JobStatus_JOB_STATUS_UNSPECIFIED:
```
joe-elliott, line 135: small code cleanup. these cases could be tested once on line 121:

```
status := j.GetStatus()
if status != running && status != unspecified {
  return false
}

// check other things
```

zalegrala: I pushed a commit to tidy this up a bit.  It avoids some processing now, good call.

PR #5025, opened 2025-04-18
[bugfix] fix bug when max size is set to zero

ie-pham:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/integration/e2e/api/config-query-range-max-series-disabled.yaml
+++ b/integration/e2e/api/config-query-range-max-series-disabled.yaml
@@ -1,0 +1,81 @@
+target: all
+stream_over_http_enabled: true
+
+server:
+  http_listen_port: 3200
+  log_level: debug
+
+query_frontend:
+  search:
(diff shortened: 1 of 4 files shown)


integration/e2e/api/config-query-range-max-series-disabled.yaml
```diff
@@ -0,0 +1,81 @@
+target: all
+stream_over_http_enabled: true
+
+server:
+  http_listen_port: 3200
+  log_level: debug
+
+query_frontend:
+  search:
+    query_backend_after: 0 # setting these both to 0 will force all range searches to hit the backend
```
mdisibio, line 10: I think this e2e test is still querying the generators because this needs to be moved to under `metrics:` instead.

ie-pham: added another test that targets the querier

mdisibio (approved):

PR #5034, opened 2025-04-21
fix: measure the write latency on append errors

javiermolinar:
**What this PR does**:
This PR fixes an issue measuring Kafka write latency when all append record operations fail.
It also improves the accuracy of the latency measurement by moving the start time and duration calculation closer to the actual append operation, avoiding interference from the encoding step.

Additionally, the PR changes the error handling to return the first encountered error (if any), rather than the last. This simplifies the logic and more clearly indicates that the response is partially erroneous.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/distributor.go
+++ b/modules/distributor/distributor.go
@@ -24,7 +24,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/prometheus/util/strutil"
 	"github.com/segmentio/fasthash/fnv1a"
-	"github.com/twmb/franz-go/pkg/kgo"
 	"go.opentelemetry.io/collector/pdata/ptrace"
 	"go.opentelemetry.io/otel"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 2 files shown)


modules/distributor/distributor.go
```diff
@@ -645,48 +644,39 @@ func (d *Distributor) sendToKafka(ctx context.Context, userID string, keys []uin
...
+		if count > 0 {
+			metricKafkaWriteBytesTotal.WithLabelValues(partitionLabel).Add(float64(sizeBytes))
+			metricKafkaAppends.WithLabelValues(partitionLabel, "success").Add(float64(count))
 		}
-	}
 
-	return count, sizeBytes
+		_ = level.Debug(d.logger).Log("msg", "kafka write success stats", "count", count, "size_bytes", sizeBytes, "partition", partitionLabel)
+
+		return produceResults.FirstErr()
```
mapno, line 678: Is there a difference between returning the first error vs the last error?

mapno (approved):

PR #5035, opened 2025-04-21
Fix artificial delay override

mapno:
**What this PR does**:

This PR contains two changes:
1) Correctly handles conversion of ingestion_artificial_delay field between config formats
2) Fixes other conversion issues and adds comprehensive conversion tests

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/overrides/config_test.go
+++ b/modules/overrides/config_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"flag"
 	"reflect"
+	"slices"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


modules/overrides/config_test.go
```diff
@@ -306,3 +311,199 @@ func TestOverrides_AssertUserConfigurableOverridesAreASubsetOfRuntimeOverrides(t
...
+	structType := v.Type()
+
+	// Iterate through all fields
+	for i := 0; i < v.NumField(); i++ {
+		field := v.Field(i)
+		fieldName := structType.Field(i).Name
+
+		// Skip certain fields that can be zero in valid configs
+		switch fieldName {
+		case "IngestionArtificialDelay": // Can legitimately be nil
```
javiermolinar, line 345: what do you think about creating a static list of fieldNames, that way it would be easier to add new nil values to skip

```go
	skip := []string{"IngestionArtificialDelay"}

	if slices.Contains(skip, fieldName) {
		continue
	}
```

javiermolinar (approved):
Nice!

modules/overrides/config_test.go
mapno: For now it's just that field, but sounds good 👍

PR #5041, opened 2025-04-23
Updated memcached client to honor a cancelled context

joe-elliott:
**What this PR does**:
Updates the memcached client to honor a cancelled context. This prevents a somewhat rare panic from occurring:

```
	/Users/joe/dev/joe-elliott/tempo/vendor/github.com/parquet-go/parquet-go/page.go:129 +0xfd
created by github.com/parquet-go/parquet-go.AsyncPages in goroutine 610894
	/Users/joe/dev/joe-elliott/tempo/vendor/github.com/parquet-go/parquet-go/page.go:269 +0x178
github.com/parquet-go/parquet-go.readPages({0x3fdf8f0, 0xc0bf919ae0}, 0xc10c1dc310, 0xc095e75880, 0x4007b18?, 0xc10c1dc3f0)
	/Users/joe/dev/joe-elliott/tempo/vendor/github.com/parquet-go/parquet-go/file.go:801 +0x9f
github.com/parquet-go/parquet-go.(*FilePages).ReadPage(0xc0bf919ae0)
	/Users/joe/dev/joe-elliott/tempo/vendor/github.com/parquet-go/parquet-go/file.go:916 +0x65
github.com/parquet-go/parquet-go.(*FilePages).readPage(0xc0bf919ae0, 0xc07d7b5740, 0xc096ba61e0)
	/Users/joe/dev/joe-elliott/tempo/vendor/github.com/parquet-go/parquet-go/buffer.go:534 +0x12d
github.com/parquet-go/parquet-go.(*bufferPool).get(0x623bf20?, 0xfffffffffffffffc?)
goroutine 611543 [running]:

panic: runtime error: slice bounds out of range [:-4]
```

The believed series of events that create this panic are:

1. While [decoding a header](https://github.com/parquet-go/parquet-go/blob/main/file.go#L798) in FilePages.ReadPage the thrift code will read short segments and sometimes [individual bytes](https://github.com/parquet-go/parquet-go/blob/main/encoding/thrift/binary.go#L207).
1. One of these very short reads misses the [bufio.Reader's](https://github.com/parquet-go/parquet-go/blob/ba969aa4929b45459259ea5c2c4a09876a881c04/file.go#L759) internal buffer triggering a read from the backend.
1. Cache misses and the backend read fails with `context.Canceled` b/c the job was canceled for any number of reasons.
1. Something else successfully reads that exact parquet page and stores it in cache.
1. The [AsyncPages](https://github.com/parquet-go/parquet-go/blob/ba969aa4929b45459259ea5c2c4a09876a881c04/page.go#L263-L289) loop comes back around and calls `ReadPage` again.
1. [Decoder.decode](https://github.com/parquet-go/parquet-go/blob/main/file.go#L798) is called again but now the underlying buffer is misaligned. We are N bytes into the header instead of at the start of the header. This wouldn't matter if the backend would just return `context.Canceled` again but...
1. This time cache hits and returns successfully because memcached [doesn't honor context](https://github.com/grafana/tempo/blob/4ab3c173af81c037af9e5d77f30309e78d57eed7/pkg/cache/memcached.go#L98-L100).

Other notes:
- The redis client seems to already respect context? It takes one but I've found it difficult to trace exactly how it's using it.
- Long term we should probably swap to the [dskit cache clients](https://github.com/grafana/dskit/tree/main/cache)

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/cache/memcached.go
+++ b/pkg/cache/memcached.go
@@ -153,13 +153,13 @@ func (c *Memcached) FetchKey(ctx context.Context, key string) ([]byte, bool) {
 
 // Store stores the key in the cache.
 func (c *Memcached) Store(ctx context.Context, keys []string, bufs [][]byte) {
-	select {
-	case <-ctx.Done():
-		return
-	default:
-	}
-
 	for i := range keys {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+		}
+
 		err := measureRequest(ctx, "Memcache.Put", c.requestDuration, memcacheStatusCode, func(_ context.Context) error {
 			item := memcache.Item{
 				Key:        keys[i],
--- a/pkg/cache/memcached.go
+++ b/pkg/cache/memcached.go
@@ -74,46 +74,18 @@ func NewMemcached(cfg MemcachedConfig, client MemcachedClient, name string, maxI
 	return c
 }
 
-func memcacheStatusCode(err error) string {
-	// See https://godoc.org/github.com/bradfitz/gomemcache/memcache#pkg-variables
-	if errors.Is(err, memcache.ErrCacheMiss) {
-		return "404"
-	}
-	if errors.Is(err, memcache.ErrMalformedKey) {
-		return "400"
-	}
-	if err != nil {
-		return "500"
-	}
-	return "200"
-}
-
-// Fetch gets keys from the cache. The keys that are found must be in the order of the keys requested.
+// Fetch gets keys from the cache. The keys that are found must be in the order of the keys requested. jpe - just remove
 func (c *Memcached) Fetch(ctx context.Context, keys []string) (found []string, bufs [][]byte, missed []string) {
+	select {
+	case <-ctx.Done():
+		found = nil
+		bufs = nil
+		missed = keys
+		return
+	default:
+	}
 	found, bufs, missed = c.fetch(ctx, keys)
 	return
-}
-
-// FetchKey gets a single key from the cache
-func (c *Memcached) FetchKey(ctx context.Context, key string) ([]byte, bool) {
-	const method = "Memcache.Get"
-	var item *memcache.Item
-	err := measureRequest(ctx, method, c.requestDuration, memcacheStatusCode, func(_ context.Context) error {
-		var err error
-		item, err = c.memcache.Get(key, memcache.WithAllocator(cacheBufAllocator))
-		if err != nil {
-			if errors.Is(err, memcache.ErrCacheMiss) {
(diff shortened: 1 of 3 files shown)


pkg/cache/memcached.go
```diff
@@ -90,12 +90,24 @@ func memcacheStatusCode(err error) string {
 
 // Fetch gets keys from the cache. The keys that are found must be in the order of the keys requested.
 func (c *Memcached) Fetch(ctx context.Context, keys []string) (found []string, bufs [][]byte, missed []string) {
+	select {
```
mattdurham, line 93: Can the same error still occur in the underlying `fetch` call, it looks like around line 113 it is throwing away the context.

joe-elliott: fetch() is only called in one place which is here. i think i'll just collapse these into one function.

i was tempted to simply remove `Fetch()` b/c it's not used in our codebase. I started down this path but it broke a lot of tests. I decided to just leave this call alone in order to keep the PR simple.

done!

pkg/cache/memcached.go
```diff
@@ -143,8 +123,42 @@ func (c *Memcached) fetch(ctx context.Context, keys []string) (found []string, b
...
+	})
+	if err != nil {
+		return nil, false
+	}
+	return item.Value, true
+}
+
 // Store stores the key in the cache.
 func (c *Memcached) Store(ctx context.Context, keys []string, bufs [][]byte) {
+	select {
```
zalegrala, line 156: Do you want to put this inside the loop?  Not sure how long the Store calls take generally, but checking at the top of the loop could possibly help.

zalegrala (approved):
Nice one.

pkg/cache/memcached.go
```diff
@@ -143,9 +123,43 @@ func (c *Memcached) fetch(ctx context.Context, keys []string) (found []string, b
...
+		if err != nil {
+			if errors.Is(err, memcache.ErrCacheMiss) {
+				level.Debug(c.logger).Log("msg", "Failed to get key from memcached", "err", err, "key", key)
+			} else {
+				level.Error(c.logger).Log("msg", "Error getting key from memcached", "err", err, "key", key)
+			}
+		}
+		return err
+	})
+	if err != nil {
```
mattdurham, line 148: Not familiar with the nuances of the error, but should we also check the context before returning, IE if it has been cancelled simulate a nil/false response?

joe-elliott: in the case of the panic in question this won't matter. i copied this pattern of "check once at the beginning" from dskits memcached client.

Not sure if checking with more frequency is worth it.

PR #5042, opened 2025-04-23
feature: expose a metric for blockbuilder owned partitions

javiermolinar:
**What this PR does**:
It adds a new metric to track the partitions owned by a blockbuilder. This will allow us to identify problems with unassigned partitions. 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/blockbuilder/blockbuilder.go
```diff
@@ -232,6 +238,10 @@ func (b *BlockBuilder) running(ctx context.Context) error {
 func (b *BlockBuilder) consume(ctx context.Context) (time.Duration, error) {
 	partitions := b.getAssignedPartitions()
 
+	for _, p := range partitions {
+		metricOwnedPartitions.WithLabelValues(strconv.Itoa(int(p))).Set(1)
```
zalegrala, line 242: If a partition becomes un-owned, do we have a place to set this back to 0?

javiermolinar: I may be wrong, but I think't we can leverage on Prometheus staleness. If the metric has not been updated in the past 5 minutes then it won't show up

mapno (commented):
LGTM. We could measure it inside `b.getAssignedPartitions()` and add partition state too.

modules/blockbuilder/blockbuilder.go
```diff
@@ -75,6 +75,12 @@ var (
 		Name:      "fetch_errors_total",
 		Help:      "Total number of errors while fetching by the consumer.",
 	}, []string{"partition"})
+	metricOwnedPartitions = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "tempo",
+		Subsystem: "block_builder",
+		Name:      "owned_partitions",
+		Help:      "Indicates partition ownership by this block-builder (1 = owned).",
+	}, []string{"partition"})
```
mapno, line 83: Adding the partition state would be very nice.

javiermolinar: Good idea

mapno (approved):

PR #5054, opened 2025-04-24
Remove ColumnIterator

joe-elliott:
**What this PR does**:

The old async ColumnIterator has long since been abandoned. It was only used in one spot and that was simple to replace with the SyncIterator. I'm noodling some iterator level improvements and this cleanup makes it easier to try them.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


zalegrala (approved):
Looks reasonable.

PR #5058, opened 2025-04-27
feat(export): Add OTLP export support replacing Jaeger agent

### LINKED ISSUE #4928: Replace jaeger agent by otlp exporter
The Jaeger agent was deprecated some time ago. Two weeks ago, the code was removed from Jaeger

https://github.com/jaegertracing/jaeger/pull/6868

Currently, we use it in two places:

[Tempo vulture](https://github.com/grafana/tempo/blob/4c489a4e94c22312330ea68349753eee12832684/cmd/tempo-vulture/main.go#L19)

[Integration tests util](https://github.com/grafana/tempo/blob/4c489a4e94c22312330ea68349753eee12832684/integration/util/util.go#L24)

The otlp exporter can replace it 

https://github.com/grafana/tempo/blob/4c489a4e94c22312330ea68349753eee12832684/integration/util/util.go#L360

iamrajiv:
**What this PR does**:

Fixes #4928

**Checklist**
- [x] Tests updated
- [x] Documentation added


integration/util/util.go
```diff
@@ -671,3 +674,75 @@ func NewPrometheus() *e2e.HTTPService {
...
+}
+
+func (c *OtlpJaegerClient) EmitBatch(ctx context.Context, b *thrift.Batch) error {
+	resourceSpans := thriftBatchToResourceSpans(b)
+	traces := ptrace.NewTraces()
+	resourceSpans.CopyTo(traces.ResourceSpans().AppendEmpty())
+	return c.exporter.ConsumeTraces(ctx, traces)
+}
+
+func thriftBatchToResourceSpans(b *thrift.Batch) ptrace.ResourceSpans {
```
javiermolinar: Can we use `translator/jaeger` for this? :

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/078a85d9fe74d331c7949de92d118e3720d2182d/pkg/translator/jaeger/jaegerthrift_to_traces.go#L22

javiermolinar (commented):
Thank you for your patience, we need to rebase with main but it's looking good

integration/util/util.go
iamrajiv: yeah sure @javiermolinar i made the changes can you review again?

integration/util/util.go
```diff
@@ -692,3 +694,27 @@ func NewPrometheus() *e2e.HTTPService {
 		9090,
 	)
 }
+
+type OtlpJaegerClient struct {
```
javiermolinar, line 698: What do you think if we move this into 

`github.com/grafana/tempo/pkg/util`

That way, we don't need to reference the integration package from Vulture.

javiermolinar (commented):
Looking good, there are some errors from the rebase we need to fix.

javiermolinar:
**Breaking changes**

Vulture client now sends traces to the OTLP grpc endpoint instead of the Jaeger one. The new default port is 4317

**Other changes**
There was a bug in the `getGRPCEndpoint` function in Vulture that trimmed the protocol from the URL

cmd/tempo-vulture/main.go
```diff
@@ -155,11 +149,18 @@ func main() {
...
 	}
+	pushEndpoint, err := getGRPCEndpoint(vultureConfig.tempoPushURL)
+	if err != nil {
+		panic(err)
+	}
+
+	logger.Info("Tempo Vulture starting", zap.String("tempoQueryURL", vultureConfig.tempoQueryURL), zap.String("tempoPushURL", pushEndpoint))
 
-	jaegerClient, err := newJaegerGRPCClient(vultureConfig, logger)
+	jaegerClient, err := utilpkg.NewOtlpThriftClient(pushEndpoint)
```
joe-elliott: is this still a jaeger client? it's not clear to me if we're using the otel jaeger exporter or the otel otlp exporter.

integration/util/util.go
```diff
@@ -692,3 +695,27 @@ func NewPrometheus() *e2e.HTTPService {
 		9090,
 	)
 }
+
+type OtlpThriftClient struct {
+	exporter exporter.Traces
+}
+
+func NewOtlpThriftClient(endpoint string) (*OtlpThriftClient, error) {
```
joe-elliott: i don't think this is thrift. i think this is grpc/proto

joe-elliott (approved):
i'm good on the change, but i think the "thrift" name is not correct.

integration/util/util.go
javiermolinar: Yes, you are right, I was confused because we import:
"github.com/jaegertracing/jaeger/thrift-gen/jaeger" as thrift

So this is a client that takes a Jaeger model and converts it to OTEL traces

javiermolinar:
@iamrajiv could you sign the license so we can merge it? Thank you!

iamrajiv:
done @javiermolinar

joe-elliott:
thx @iamrajiv for getting this started 🙏

iamrajiv:
thank you @javiermolinar and @joe-elliott

PR #5071, opened 2025-04-28
Secure CI: set explicit top-level permissions

ruslan-mikhailov:
**What this PR does**:  sets permissions explicitly. It might break CI as it sets default permissions to none

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -11,6 +11,8 @@ on:
       - closed
       - labeled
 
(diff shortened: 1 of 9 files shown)


electron0zero:
I am going to merge this because it's better to merge this, see what breaks and fix that in followup PRs. we won't know what's actually breaking without merging it.

PR #5079, opened 2025-04-29
add zizmor analysis to workflow

ie-pham:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


stoewer (approved):

.github/workflows/zizmor-analysis.yml
```diff
@@ -0,0 +1,19 @@
...
+    permissions:
+      actions: read
+      contents: read
+
+      pull-requests: write
+      security-events: write
+
+    uses: grafana/shared-workflows/.github/workflows/reusable-zizmor.yml@c0944351c68e107649cb9c9a86204b400b607564
+    with:
+      fail-severity: high
```
electron0zero, line 19: default is `high`, but should we fail on `any` so catch new issues we introduce into the actions? and we can go back to medium or low if `any` is too noisy.

PR #5084, opened 2025-05-01
Improve backendscheduler UpdateJob performance

zalegrala:
**What this PR does**:

Here we improve performance of the UpdateJob method when searching for the blocks which have been compacted by comparing the UUID instead of the strings.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


zalegrala:
```
❯ benchstat main.txt wip.txt
goos: linux
goarch: amd64
pkg: github.com/grafana/tempo/modules/backendscheduler
                                      │   main.txt   │               wip.txt               │
BackendScheduler_blockLoopPatterns-16   9232.3µ ± 2%   828.1µ ± 12%  -91.03% (p=0.002 n=6)

                                      │    main.txt     │              wip.txt               │
                                      │      B/op       │    B/op     vs base                │
BackendScheduler_blockLoopPatterns-16   9600171.00 ± 0%   48.00 ± 2%  -100.00% (p=0.002 n=6)

                                      │    main.txt     │              wip.txt               │
                                      │    allocs/op    │ allocs/op   vs base                │
BackendScheduler_blockLoopPatterns-16   200003.000 ± 0%   3.000 ± 0%  -100.00% (p=0.002 n=6)

```

zalegrala:
```go
func BenchmarkBackendScheduler_blockLoopPatterns(b *testing.B) {
	count := 100000
	metas := make([]*backend.BlockMeta, count)

	for i := 0; i < count; i++ {
		metas[i] = &backend.BlockMeta{
			BlockID: backend.NewUUID(),
		}
	}

	j := &work.Job{
		ID:          uuid.New().String(),
		CreatedTime: time.Now(),
		StartTime:   time.Now(),
		EndTime:     time.Now(),
		Type:        tempopb.JobType_JOB_TYPE_COMPACTION,
		JobDetail: tempopb.JobDetail{
			Compaction: &tempopb.CompactionDetail{
				Input:  []string{metas[0].BlockID.String(), metas[1].BlockID.String()},
				Output: []string{uuid.New().String()},
			},
		},
	}
	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		var oldBlocks []*backend.BlockMeta

		var bb uuid.UUID
		for _, b := range j.GetCompactionInput() {
			bb = (uuid.UUID)(backend.MustParse(b))

			for _, m := range metas {
				if (uuid.UUID)(m.BlockID) == bb {
					oldBlocks = append(oldBlocks, m)
				}
			}
		}

		// for _, b := range j.GetCompactionInput() {
		// 	for _, m := range metas {
		// 		if m.BlockID.String() == b {
		// 			oldBlocks = append(oldBlocks, m)
		// 		}
		// 	}
		// }

		require.Len(b, oldBlocks, 2)

	}
}
```

modules/backendscheduler/backendscheduler.go
```diff
@@ -290,11 +291,14 @@ func (s *BackendScheduler) UpdateJob(ctx context.Context, req *tempopb.UpdateJob
 		var (
 			metas     = s.store.BlockMetas(j.Tenant())
 			oldBlocks []*backend.BlockMeta
+			bb        uuid.UUID
 		)
 
 		for _, b := range j.GetCompactionInput() {
+			bb = (uuid.UUID)(backend.MustParse(b))
```
mattdurham, line 298: Are there concerns over a panic here?

mattdurham: Calling parse and checking the error seems more reliable.

zalegrala: I wasn't concerned until you asked. :)

I believe this won't be an issue, because we convert the blockIDs to strings [here](https://github.com/zalegrala/tempo/blob/4c7d3f91f2bed4eab301a5d336c7cfe5cfa64e2e/modules/backendscheduler/provider/compaction.go#L197) which is then converted to proto [here](https://github.com/zalegrala/tempo/blob/4c7d3f91f2bed4eab301a5d336c7cfe5cfa64e2e/modules/backendscheduler/provider/compaction.go#L182).  Then it's json in/out to the backend, and proto between the components.

To be cautious we could convert this to a Parse call and log the resulting error if we had one.  What do you think?

modules/backendscheduler/backendscheduler.go
```diff
@@ -290,11 +291,21 @@ func (s *BackendScheduler) UpdateJob(ctx context.Context, req *tempopb.UpdateJob
...
+				continue
+			}
+
+			bb = (uuid.UUID)(u)
+
 			for _, m := range metas {
-				if m.BlockID.String() == b {
+				if (uuid.UUID)(m.BlockID) == bb {
 					oldBlocks = append(oldBlocks, m)
 				}
```
stoewer, line 310: If i'm not mistaken both `u` and `m.BlockID` have the type `backend.UUID`. Therefore it should be possible to simplify the code and omit the double casting to `uuid.UUID`.

Removing L294 and L305 and write the loop like this should work as well:
```go
			for _, m := range metas {
				if m.BlockID == u {
					oldBlocks = append(oldBlocks, m)
				}
			}
 ```

File: modules/backendscheduler/backendscheduler.go [hunk 3]
```diff
@@ -290,11 +290,18 @@ func (s *BackendScheduler) UpdateJob(ctx context.Context, req *tempopb.UpdateJob
...
 		for _, b := range j.GetCompactionInput() {
+			u, err = backend.ParseUUID(b)
+			if err != nil {
+				level.Error(log.Logger).Log("msg", "failed to parse block ID", "block_id", b, "error", err)
+				continue
+			}
+
 			for _, m := range metas {
-				if m.BlockID.String() == b {
+				if m.BlockID == u {
```
mapno, line 304: Shouldn't it exit the loop if it finds a match?

stoewer (approved):
LGTM

modules/backendscheduler/backendscheduler.go
mapno: Right, but `u` will only match a single meta `m.BlockID`, no?

zalegrala: I'll follow this up in the next PR.

PR #5093, opened 2025-05-02
chore: add extra linters to golangci

javiermolinar:
It adds
- forbidigo: https://github.com/ashanbrown/forbidigo
- gocritic: https://github.com/go-critic/go-critic
- gosec: https://github.com/securego/gosec

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,6 +14,9 @@ output:
 linters:
   enable:
     - errorlint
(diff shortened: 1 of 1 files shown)


mdisibio (commented):
+1 Generally in favor of this, and since everything is already passing now is a good time to add.  But I would like to hear - is there any reason not to add them?

PR #5094, opened 2025-05-02
fix: stop early when a  partial trace is reached

### LINKED ISSUE #5080: v2 trace lookup doesn't exit when reaching max bytes
**Describe the bug**
V2 trace lookup returns the partial trace data, that is under the max trace size limit.  However it doesn't look like it exits early upon reaching the limit. Instead it keeps processing all remaining jobs.

I think the required change is for the [traceid v2 combiner](https://github.com/grafana/tempo/blob/main/modules/frontend/combiner/trace_by_id_v2.go#L18) to implement the genericCombiner's [quit callback](https://github.com/grafana/tempo/blob/main/modules/frontend/combiner/common.go#L285) 

**To Reproduce**

**Expected behavior**
The API should return the first N bytes of partial data as soon as it reaches the limit.

**Environment:**

**Additional Context**

javiermolinar:
**What this PR does**:
This change bails early when a partial trace is detected. Instead of waiting for the entire query to complete, the query frontend combiner now stops as soon as a partial trace is found.

Fixes #5080

**Checklist**
- [X] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/combiner/trace_by_id_v2_test.go
+++ b/modules/frontend/combiner/trace_by_id_v2_test.go
@@ -36,17 +36,9 @@ func TestNewTraceByIdV2ReturnsAPartialTrace(t *testing.T) {
 		Trace:   test.MakeTrace(2, []byte{0x01, 0x02}),
 		Metrics: &tempopb.TraceByIDMetrics{},
 	}
-	resBytes, err := proto.Marshal(traceResponse)
-	require.NoError(t, err)
-	response := http.Response{
-		StatusCode: 200,
-		Header: map[string][]string{
-			"Content-Type": {"application/protobuf"},
-		},
-		Body: io.NopCloser(bytes.NewReader(resBytes)),
-	}
+
 	combiner := NewTraceByIDV2(10, api.HeaderAcceptJSON)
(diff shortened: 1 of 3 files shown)


modules/frontend/combiner/trace_by_id_v2_test.go
```diff
@@ -36,17 +36,9 @@ func TestNewTraceByIdV2ReturnsAPartialTrace(t *testing.T) {
...
-	require.NoError(t, err)
-	response := http.Response{
-		StatusCode: 200,
-		Header: map[string][]string{
-			"Content-Type": {"application/protobuf"},
-		},
-		Body: io.NopCloser(bytes.NewReader(resBytes)),
-	}
+
 	combiner := NewTraceByIDV2(10, api.HeaderAcceptJSON)
```
mdisibio, line 40: Is this test working as expected?  The combiner has a limit of 10 bytes, but we only input 2 bytes, but it still returns partial status.

mdisibio (approved):
Changes LGTM.  If possible let's follow up with testing this in a larger cluster that issues a few thousand jobs, so we can see if there early exit functionality has any remaining gaps.

PR #5106, opened 2025-05-05
Improve backendscheduler performance and integrity

zalegrala:
**What this PR does**:

Here we build on #5084 to improve the performance and integrity of the backendscheduler.

* Further reduce lock contention.  After reviewing the profiles, `HasBlocks()` showed up near the top of the list for lock contention.  After review and consideration, it wasn't clear that this would be useful in the future, or what it would mean to "have a block" in other contexts.  Refactoring this into the compaction provider helped reduce the lock contention by not locking jobs more than necessary.

* A change with the above refactor is to only consult the job input for compaction jobs when creating a new blocklist selector.  We want the output of a previous job to feed into the input of a new job.  As such, we should only be the input blocks of a compaction job, and not the output, when skipping blocks in the `newBlockSelector` method. 

* Reload the work list on top of the blocklist at startup.  This is to ensure that the jobs handed out to the workers take into consideration the blocks which have already been operated on.  This happens on startup after the polling has completed, and is necessary because the tenant indexes may not be 100% up to date.

* New metrics and logging for simpler debugging and alerting.

* Fix CPU spin when all tenants are compacted and correctly handle nil job case during shutdown.  In a case where all tenants were completely compacted, a nil job meant that we would continue at the top of the loop and fall through repeatedly with no delay.  This resulted in a large amount of memory allocations.  Adding a simple backoff handles this case much more gracefully.  This was a miss in a previous PR and was only noticed when the compaction process had reduce the outstanding blocks to almost nothing.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/backendscheduler/provider/compaction.go
```diff
@@ -93,33 +92,46 @@ func (p *CompactionProvider) Start(ctx context.Context) <-chan *work.Job {
...
+				level.Info(p.logger).Log("msg", "max jobs per tenant reached, skipping to next tenant")
 				reset()
 				continue
 			}
 
 			job = p.createJob(ctx)
 			if job == nil {
+				level.Info(p.logger).Log("msg", "tenant exhausted, skipping to next tenant after delay", "waiting", b.NextDelay())
 				// we don't have a job, reset the curTenant and try again
+				metricTenantNilJob.Inc()
```
electron0zero, line 132: this is basically indicating the workers to backoff because we have no jobs, let's rename this to something like BackoffJob because we are telling workers to backoff.

modules/backendscheduler/provider/metrics.go
```diff
@@ -16,4 +16,14 @@ var (
...
 		Help:      "The number of times the tenant is changed",
 	}, []string{"tenant"})
+	metricTenantBackoff = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: "tempo_backend_scheduler",
+		Name:      "compaction_tenant_backoff_total",
+		Help:      "The number of times the backoff is triggered",
+	})
+	metricTenantNilJob = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: "tempo_backend_scheduler",
+		Name:      "compaction_tenant_nil_job_total",
```
electron0zero, line 26: let's rename this to something like `compaction_tenant_backoff_job_total` or something that indicates that we have no jobs and workers are not getting jobs because we have no work to handout.

electron0zero: maybe call it NoJobs instead of backoff, because NoJob is more clear??

zalegrala: Good callout.  I added a commit to rename it to `empty` which I think reads true enough.

modules/backendscheduler/provider/compaction.go
zalegrala: I pushed a commit to call it EmptyJob.  We're indicating that work was not found to hand out.  The backoff is a side-effect in this case, but its the job creation engine we want to measure here.

electron0zero (approved):

PR #5115, opened 2025-05-07
Issue 4632: fix exemplars calculation

### LINKED ISSUE #4632: Query exemplars field not respected

ruslan-mikhailov:
**What this PR does**:

- Request at least 1 exemplar per shard
- Correct exemplars calculation: with totalRecords > pages, page should be the numerator and total records the denominator
- Correct calculation of required pages per request (fix for case `small block`)

Fixes #4632

**Checklist**

- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/metrics_query_range_sharder_test.go
+++ b/modules/frontend/metrics_query_range_sharder_test.go
@@ -1,0 +1,25 @@
+package frontend
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func FuzzExemplarsPerShard(f *testing.F) {
+	f.Add(uint32(1), uint32(10))  // total = 1, exemplars = 10
+	f.Add(uint32(100), uint32(1)) // total = 100, exemplars = 1
+	f.Add(uint32(10), uint32(0))  // total = 10, exemplars = 0
+
+	s := &queryRangeSharder{}
+
+	f.Fuzz(func(t *testing.T, total uint32, exemplars uint32) {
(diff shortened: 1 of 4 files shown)


modules/frontend/metrics_query_range_sharder_test.go
```diff
@@ -0,0 +1,194 @@
...
+func TestBuildBackendRequestsExemplarsOneBlock(t *testing.T) {
+	// Create the test sharder with exemplars enabled
+	sharder := &queryRangeSharder{
+		logger: log.NewNopLogger(),
+		cfg: QueryRangeSharderConfig{
+			MaxExemplars: 100,
+		},
+	}
+	tenantID := "test-tenant"
+	targetBytesPerRequest := 1000
```
zalegrala, line 47: Leaving this static I suppose is desired because the test cases have blocksizes and records above and below this number to test each side of the calculation.

zalegrala (approved):
This looks correct to me.  Nice work on the tests.

mapno (approved):
LGTM

modules/frontend/metrics_query_range_sharder_test.go
ruslan-mikhailov: Exactly

ruslan-mikhailov:
\+ rebase from latest master to resolve merge conflicts

PR #5122, opened 2025-05-10
spanmetrics: drop invalid prometheus label names

KyriosGN0:

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * [CHANGE] **BREAKING CHANGE** Upgrade OTEL Collector to v0.122.1 [#4893](https://github.com/grafana/tempo/pull/4893) (@javiermolinar)
   The `name` dimension from `tempo_receiver_accepted_span` and `tempo_receiver_refused_spans` changes from `tempo/jaeger_receiver` to `jaeger/jaeger_receiver`
 * [CHANGE] **BREAKING CHANGE** Convert SLO metric `query_frontend_bytes_processed_per_second` from a histogram to a counter as it's more performant. [#4748](https://github.com/grafana/tempo/pull/4748) (@carles-grafana)
+* [ENHANCEMENT] Drop invalid prometheus label names in spanmetrics processor
 * [CHANGE] **BREAKING CHANGE** Remove tempo serverless [#4599](https://github.com/grafana/tempo/pull/4599/) @electron0zero
   Following config options are no longer valid, please remove them if you are using these in your tempo config:
   ```
@@ -39,7 +40,7 @@ configurable via the throughput_bytes_slo field, and it will populate op="traces
 * [FEATURE] Added most_recent=true query hint to TraceQL to return most recent results. [#4238](https://github.com/grafana/tempo/pull/4238) (@joe-elliott)
 * [FEATURE] Add ability to add artificial delay to push requests [#4716](https://github.com/grafana/tempo/pull/4716) [#4899](https://github.com/grafana/tempo/pull/4899) [#5035](https://github.com/grafana/tempo/pull/5035) (@yvrhdn, @mapno)
 * [FEATURE] TraceQL metrics: sum_over_time [#4786](https://github.com/grafana/tempo/pull/4786) (@javiermolinar)
-* [ENHANCEMENT] Add default mutex and blocking values. [#4979](https://github.com/grafana/tempo/pull/4979) (@mattdurham) 
+* [ENHANCEMENT] Add default mutex and blocking values. [#4979](https://github.com/grafana/tempo/pull/4979) (@mattdurham)
 * [ENHANCEMENT] Improve Tempo build options [#4755](https://github.com/grafana/tempo/pull/4755) (@stoewer)
 * [ENHANCEMENT] Rewrite traces using rebatching [#4690](https://github.com/grafana/tempo/pull/4690) (@stoewer @joe-elliott)
 * [ENHANCEMENT] Reorder span iterators [#4754](https://github.com/grafana/tempo/pull/4754) (@stoewer)
@@ -58,7 +59,7 @@ configurable via the throughput_bytes_slo field, and it will populate op="traces
 * [ENHANCEMENT] Rhythm: add block builder to resources dashboard [#4556](https://github.com/grafana/tempo/pull/4669) (@javiermolinar)
 * [ENHANCEMENT] Upgrade prometheus to version 3.1.0 [#4805](https://github.com/grafana/tempo/pull/4805) (@javiermolinar)
 * [ENHANCEMENT] Update dskit to latest version [#4681](https://github.com/grafana/tempo/pull/4681) (@javiermolinar) [#4865](https://github.com/grafana/tempo/pull/4865) (@javiermolinar)
-* [ENHANCEMENT] Increase query-frontend default batch size [#4844](https://github.com/grafana/tempo/pull/4844) (@javiermolinar) 
+* [ENHANCEMENT] Increase query-frontend default batch size [#4844](https://github.com/grafana/tempo/pull/4844) (@javiermolinar)
 * [ENHANCEMENT] Improve TraceQL perf by reverting EqualRowNumber to an inlineable function.[#4705](https://github.com/grafana/tempo/pull/4705) (@joe-elliott)
 * [ENHANCEMENT] Rhythm: Implement MaxBytesPerCycle [#4835](https://github.com/grafana/tempo/pull/4835) (@javiermolinar)
 * [ENHANCEMENT] Rhythm: fair partition consumption in blockbuilders [#4655](https://github.com/grafana/tempo/pull/4655) (@javiermolinar)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
   The `name` dimension from `tempo_receiver_accepted_span` and `tempo_receiver_refused_spans` changes from `tempo/jaeger_receiver` to `jaeger/jaeger_receiver`
 * [CHANGE] **BREAKING CHANGE** Convert SLO metric `query_frontend_bytes_processed_per_second` from a histogram to a counter as it's more performant. [#4748](https://github.com/grafana/tempo/pull/4748) (@carles-grafana)
 * [CHANGE] **BREAKING CHANGE** Remove tempo serverless [#4599](https://github.com/grafana/tempo/pull/4599/) (@electron0zero)
+* [ENHANCEMENT] Drop invalid prometheus label names in spanmetrics processor [#5122](https://github.com/grafana/tempo/pull/5122)
```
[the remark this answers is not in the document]
KyriosGN0: moved it to the unreleased section

KyriosGN0:
> thank you for the PR, but i believe we are already sanitizing label names with `SanitizeLabelNameWithCollisions` ?
> 
> is this necessary as well?

from reading what that functions does, it seems that it  check is labels are also used in intrinsics and then it adds `__` to those labels

KyriosGN0:
> thank you for the PR, but i believe we are already sanitizing label names with `SanitizeLabelNameWithCollisions` ?
> 
> is this necessary as well?

from reading what that functions does, it seems that it  check is labels are also used in intrinsics and then it adds `__` to those labels

joe-elliott:
they're also being passed through here: https://github.com/grafana/tempo/blob/main/modules/generator/processor/spanmetrics/spanmetrics.go#L73

which is Prometheus's sanitize func:
https://github.com/prometheus/prometheus/blob/main/util/strutil/strconv.go#L45-L47

KyriosGN0:
> they're also being passed through here: [`main`/modules/generator/processor/spanmetrics/spanmetrics.go#L73](https://github.com/grafana/tempo/blob/main/modules/generator/processor/spanmetrics/spanmetrics.go?rgh-link-date=2025-05-29T20%3A58%3A11Z#L73)
> 
> which is Prometheus's sanitize func: [prometheus/prometheus@`main`/util/strutil/strconv.go#L45-L47](https://github.com/prometheus/prometheus/blob/main/util/strutil/strconv.go?rgh-link-date=2025-05-29T20%3A58%3A11Z#L45-L47)

I see, in that case we make the regex simpler and cover the case that this functions doesn't (since it states that it doesn't cover all of the prometheus's requirement, the scenario is very limited and only happens if we enable InfoTarget metrics from what i understand, (or if someone on purpose request adds a dimension that is against Prometheus standard) 
What do you think ?

joe-elliott:
I'm concerned about running another regex on every label that goes through metrics generator. I see two options, but there are likely more:

1) We could try [SanitizeFullLabelName](https://github.com/prometheus/prometheus/blob/main/util/strutil/strconv.go?rgh-link-date=2025-05-29T20%3A58%3A11Z#L51) which seems to handle the "can't be prefixed with numbers" situation. We'd need benchmarks over different strings (valid, invalid, etc) to confirm it wouldn't be a nasty regression.

2) Since the only rule not covered by the above sanitize is the "can't be prefixed with numbers" we could simply check if the first rune of the string is a number and, if so, reject it. This would be cheaper than running a regex.

KyriosGN0:
@joe-elliott i have addressed  your comments

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -213,12 +213,14 @@ func (p *Processor) aggregateMetricsForSpan(svcName string, jobName string, inst
...
 	spanMultiplier := processor_util.GetSpanMultiplier(p.Cfg.SpanMultiplierKey, span, rs)
 
-	err := validateLabelValues(labelValues)
+	err := validateUTF8LabelValues(labelValues)
 	if err != nil {
 		p.invalidUTF8Counter.Inc()
 		return
 	}
 
+	validatePromLabelNames(&labels, &labelValues)
```
joe-elliott: it looks like we are doing this to labels, but these are manually configured in dimensions yes? so a user would have to configure an invalid label

would it be better to perform this on targetInfoLabels which are blindly copied from the resource? looking at the original issue it seems that is what it was written up for.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -270,6 +272,29 @@ func validateLabelValues(v []string) error {
 	return nil
 }
 
+func validatePromLabelNames(labels *[]string, labelValues *[]string) {
+	validLabels := make([]string, 0, len(*labels))
```
joe-elliott: can we do this inline instead of alloc'ing new slices?

KyriosGN0:
@joe-elliott fixed your comments

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -258,10 +258,11 @@ func (p *Processor) aggregateMetricsForSpan(svcName string, jobName string, inst
 		if resourceAttributesCount > 0 && len(targetInfoLabels) > resourceAttributesCount {
 			p.spanMetricsTargetInfo.SetForTargetInfo(targetInfoRegistryLabelValues, 1)
 		}
+		validatePromLabelNames(&labels, &labelValues)
```
joe-elliott: shouldn't this be done on targetInfoLabels and values? actually, let's dig deeper than that.

targetInfoLabels are copied from [resourceLabels](https://github.com/grafana/tempo/blob/main/modules/generator/processor/spanmetrics/spanmetrics.go#L164-L165) which are retrieved from [GetTargetInfoAttributeValues](https://github.com/grafana/tempo/blob/main/modules/generator/processor/spanmetrics/spanmetrics.go#L138).

Why don't we just skip these invalid labels in the GetTargetInfoAttributeValues [implementation](https://github.com/grafana/tempo/blob/main/modules/generator/processor/spanmetrics/spanmetrics.go#L273-L292)?

KyriosGN0: seems fair, but if we are doing that, that functions also passes labels through the sanitize functions mentioned before, if we are touching that i think it would be better to switch the functions to use the full `SanitizeFullLabelName` functions from the prometheus lib

i will try to run the benchmark test locally with that change.

KyriosGN0: i have tested locally, and test pass, but i can't make the benchmark test work on my machine, (docker-compose ain't working since the load test image is missing) 
is there some way /updated guide on how to run the benchmark on macos ?

joe-elliott: You shouldn't need to docker to run benchmarks. Just create a new benchmark that focuses on the change:
```
func BenchFoo(b *testing.B) {
  ...
}
```
and run it:
```
go test -run xxx -bench BenchFoo -benchmem -count=5 > after.txt
```
generate a before and after.txt and then you can use benchstat to get a summary
```
benchstat before.txt after.txt
```

I will say this change is a lot more complex and it would be a lot safer to just ignore the integer prefixed metrics in the `GetTargetInfoAttributeValues` func

KyriosGN0: Aside from performance degradation what else is at risk here? Its not like the other rules could be ingested to prometheus
EDIT: never mind after reading the prometheus functions i see it had more behavioral changes, i won't use it right now

KyriosGN0: @joe-elliott address your comments, i have a added a simple check in the if condition to skip the keys that start with a number

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -70,7 +70,7 @@ func New(cfg Config, reg registry.Registry, filteredSpansCounter, invalidUTF8Cou
 		labels = append(labels, dimStatusMessage)
 	}
 
-	c := reclaimable.New(strutil.SanitizeLabelName, 10000)
+	c := reclaimable.New(strutil.SanitizeFullLabelName, 10000)
```
joe-elliott: this is so close!

my last concern is this change here. `SanitizieFullLabelName` creates a string builder and reallocs the string every time. `SanitizeLabelName` uses a regex to replace all invalid chars with `_`. we can either change it back or i'd need to see a benchmark of these two methods to confirm there isn't a regression here.

if you do write a benchmark let's make sure to bench the case where both substitutions and no substitutions are made on both funcs.

KyriosGN0: @joe-elliott forgot to revert it, fixed

joe-elliott:
That jsonnet check thing is failing on my branch too. I'm currently looking into it.

joe-elliott:
this should fix it: https://github.com/grafana/tempo/pull/5349

joe-elliott (approved):
let's go!

joe-elliott:
Nice fix 🙏 Thank you for your patience and responsiveness to reviews.

PR #5130, opened 2025-05-12
TraceQL: Allow comparison duration to float

### LINKED ISSUE #4046: TraceQL: Durations can't be compared to floats

ndk:
**What this PR does**

This PR enables comparisons between `span:duration` and floating-point values.

Previously, the query engine only supported integer comparisons.
The new helper `createIntPredicateFromFloat` rewrites float predicates into
equivalent integer predicates while preserving semantics. Because some floats
cannot be represented exactly as integers, the function adjusts the operator
accordingly. For example:

- `>= 10.1`   ->  `> 11`
- `< 10.1`    ->  `<= 10`

It seems that `v2` already handles floats internally due to different design. So no changes are required there.

**Which issue(s) this PR fixes**

Fixes #4046

**Checklist**

- [x] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated — entries should be ordered `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/tempodb/encoding/vparquet2/block_traceql.go
+++ b/tempodb/encoding/vparquet2/block_traceql.go
@@ -1372,7 +1372,7 @@ func createSpanIterator(makeIter makeIterFn, primaryIter parquetquery.Iterator, 
 			continue
 
 		case traceql.IntrinsicDuration:
-			pred, err := createIntPredicate(cond.Op, cond.Operands)
+			pred, err := createDurationPredicate(cond.Op, cond.Operands)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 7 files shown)


joe-elliott (commented):
Apologies for the silence on this end. I've been quite busy lately. This PR fixes a strange corner of TraceQL likely not to be run into by an average user, but I appreciate it b/c it improves the correctness of the language.

These parse and work as expected:
```
{ 1 < 1.1 }
{ span:duration < 1 }
``` 

This failed with a strange iterator error:
```
{ span:duration < 1.1 }
```

I also like this PR b/c it adds the `createIntPredicateFromFloat` which has some code I think we will want in the future.

One change please! You have fixed the place where an iterator for `span:duration` is created but can you add the same fix to `trace:duration`:

https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet4/block_traceql.go#L2364

Thank you for your time!

ndk:
> This failed with a strange iterator error:
> 
> ```
> { span:duration < 1.1 }
> ```

Could you provide more details on how to reproduce this error? I'm not seeing any errors when I run this expression.

joe-elliott:
> Could you provide more details on how to reproduce this error? I'm not seeing any errors when I run this expression.

Sorry, I mean in Tempo currently that query gives a strange iterator error. Something like 
```
creating fetch iter: error creating iterator: creating span iterator: operand is not int, duration, status or kind: 1.1
```

Your PR nicely fixes it.

joe-elliott (approved):
Thanks for the fix! cc @ruslan-mikhailov who is looking into traceql correctness this Q

PR #5135, opened 2025-05-16
Abort appending during "unexpected" errors in compactor

### LINKED ISSUE #4878: Corrupt parquet files accumulate to a lots of dangling s3 multipart uploads in minio during compactor work
**Describe the bug**
For some reason we managed to have some corrupted parquet blocks in s3/minio, I guess somehow during ingest or maybe magical bit flips in our minio drive.

We noticed that as the compactor was logging this with this error in when it tried to compact it
```
error iterating input blocks: error iterating through block 31b9669c-b103-4f8a-9152-9bff6d8e8e34: crc32 checksum mismatch in page of column "rs.list.element.ss.list.element.Spans.list.element.Attrs.list.element.Value.list.element": want=0xA97290CE got=0x42327D70: corrupted parquet page
```
In the same time we started see a massive increase of disk space in minio (~15gb/h)
This was not the tempo bucket itself but the minio's multipart upload directory where it stores ongoing multipart uploads.
In Minio's Access logs we saw a NewMultiPartUploads followed by PutObjectPart but missing Complete or AbortsMultipart calls.

When digging through tempo's code I kind of came to the conclusion that in the loop where it [iterates through the blocks to compact](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet4/compactor.go#L145-L151) and stores it in [several parts by appending data via a multipart upload](https://github.com/grafana/tempo/blob/main/tempodb/backend/s3/s3.go#L204-L218) it never reaches the [closing part of the appending](https://github.com/grafana/tempo/blob/main/tempodb/backend/s3/s3.go#L253) as the loop never reaches the end where [it finishes the block](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet4/compactor.go#L220) due to the direct return statement

After we deleted the corrupted block parquet files the issue stopped.

**To Reproduce**
I'm struggeling a bit in creating a reproducable code in a test.
I tried to setup a test in `tempodb/encoding/vparquet4/compactor_test.go` where I corrupt the file by flipping a bit but the iterator fails in the first step not reachting the backend.Writer at all.
Unfortunatelly I have not understand all the logic of this iteration code to create the proper setup. Any pointers?

**Expected behavior**
It seems that the compactor is not able to handle those corrupt files.
1. The Compact method should somehow handle errors and abort the "appending". In s3/minio case somehow resulting in an `AbortMultipartUpload` call.
2. Make sure the corrupted files will not be retried by deleting or quarantine them for a further inspection

**Environment:**
 - Infrastructure: Nomad
 - Deployment tool: microservice mode

mxab:
**What this PR does**:
This PR mitigates the issue when "unexpected" error happens during the appending part in the compactor phase.

We do this by introducing a `AbortAppend` on the `bakend.Writer` / `backend.RawWriter` interface, as those interface has a lot of different implementation we leave them all unimplemented expect the `s3` one. Here we call the abort multipart upload on the minio client with the values from the tracker.

Setting up the tests for that verify the call to the `AbortAppend` was a bit challanging and currently covers 2 of the 4 potential places where they need to be handled.

Therefore we

1. Introduce a factory for multiblockIterator to have the creation for those better under control 4d4cd49375f913e3663203ed8c8246ba0e1f07bb
2. same for the streamingBlock 04ab66da56a7f2bf0faba45d8db67b1095540a75
3. by default we use the current logic 81dd85ce4ed75e99e0e1b02ead9b50f7472b088e
4. But introduce another `NewCompactorWithFactories` to override them for testing purposes eebe3900a5936dbbeb3fbf1cd77d2e93fa2edce2
5. We introduc the `AbortAppend` method on the `backend.Writer` interface which produces a lot of noise df36a0c41506013c4dd4c5ab2f2f1f726e047c88
6. A mock `backend.Writer` to verify the call of the Abort methpd 796101a810238ab4184cbc9063ccaaca1dae859f
7.  we then inject a "faulty" iterator, which is a bit tricky but solved by letting the combine method fail on the second call and then implement the call to the abort method 712ee1d7df5a9a57c10efb21c69d14e8e2c9a0a0
8. concrete s3 abort (no tests as the whole append thing hasn't any there) bafdcc7f6e20a65d6a23ef9aa3c561ab44b38323
9.  catch another fail 68d4056119d75aba7cc00fcce55ea23134a82a62

todo: handle the other 2 unhandle errors

the streamingBlockFactory thing is currently not used but I think I might need it for those other errors

Overall the whole way how it is injected doesnt feel really right but I didnt see any other way without to much refactoring

Fixes #4878

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/backend/azure/azure.go
+++ b/tempodb/backend/azure/azure.go
@@ -147,6 +147,12 @@ func (rw *Azure) Append(ctx context.Context, name string, keypath backend.KeyPat
 	return a, nil
 }
 
+// AbortAppend implements backend.Writer
+func (rw *Azure) AbortAppend(ctx context.Context, tracker backend.AppendTracker) error {
+	// TODO: implement AbortAppend
+	return nil
+}
+
 // CloseAppend implements backend.Writer
 func (rw *Azure) CloseAppend(context.Context, backend.AppendTracker) error {
 	return nil
--- a/tempodb/backend/backend.go
+++ b/tempodb/backend/backend.go
@@ -48,6 +48,8 @@ type Writer interface {
 	WriteBlockMeta(ctx context.Context, meta *BlockMeta) error
 	// Append starts or continues an Append job. Pass nil to AppendTracker to start a job.
 	Append(ctx context.Context, name string, blockID uuid.UUID, tenantID string, tracker AppendTracker, buffer []byte) (AppendTracker, error)
+	// AbortAppend aborts an Append job. This is a no-op if the AppendTracker is nil.
+	AbortAppend(ctx context.Context, tracker AppendTracker) error
 	// CloseAppend closes any resources associated with the AppendTracker
 	CloseAppend(ctx context.Context, tracker AppendTracker) error
 	// WriteTenantIndex writes the two meta slices as a tenant index
--- a/tempodb/backend/cache/cache.go
+++ b/tempodb/backend/cache/cache.go
@@ -158,6 +158,12 @@ func (r *readerWriter) Append(ctx context.Context, name string, keypath backend.
 	return r.nextWriter.Append(ctx, name, keypath, tracker, buffer)
 }
 
+// AbortAppend implements backend.Writer
+func (r *readerWriter) AbortAppend(ctx context.Context, tracker backend.AppendTracker) error {
+	// TODO: implement AbortAppend
+	return nil
+}
+
 // CloseAppend implements backend.Writer
 func (r *readerWriter) CloseAppend(ctx context.Context, tracker backend.AppendTracker) error {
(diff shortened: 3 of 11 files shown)


joe-elliott (commented):
thank you for this PR! i want to apologize for the slow response. I've been on PTO and and been swamped with other responsibilities.

the change looks quite good. i'll say i'm not a fan of the factories. i generally prefer more readable code over perfect testing. is this testable at all without the factories?

> todo: handle the other 2 unhandle errors

+1 !

mxab:
Hi, thanks for your feedback, this time I'm the slow one. I try to complete it asap but I'm also dragged into multiple projects :)

PR #5150, opened 2025-05-21
Add jitter in backendworker to avoid thundering herd from workers

electron0zero:
**What this PR does**:

Right now we can have a case where the scheduler comes up, all the workers connect to it and wait for jobs and fails due to issues or when the scheduler has no jobs or failed to handout jobs to workers. all the workers will start backing off with retry with the same backoff durations.

This can cause waves of thundering herd on the scheduler if we have a sizeable amount of workers.

This PR add a jitter in workers so they are offset by jitter and won't cause waves of thundering herd on the scheduler.

This is even more important because backend scheduler is a singleton component and can be easily overwhelmed in bigger clusters.

Jitter code in play: https://go.dev/play/p/qb5GU11lKF5

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,12 @@
 ## main / unreleased
-* [FEATURE] Add support for topk and bottomk functions for TraceQL metrics [#4646](https://github.com/grafana/tempo/pull/4646/) @electron0zero
+* [ENHANCEMENT] Add jitter in backendworker to avoid thundering herd from workers [#5150](https://github.com/grafana/tempo/pull/5150) (@electron0zero)
+* [FEATURE] Add support for topk and bottomk functions for TraceQL metrics [#4646](https://github.com/grafana/tempo/pull/4646/) (@electron0zero)
 * [CHANGE] **BREAKING CHANGE** Change default http-listen-port from 80 to 3200 [#4960](https://github.com/grafana/tempo/pull/4960) (@martialblog)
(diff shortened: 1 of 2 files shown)


modules/backendscheduler/backendscheduler.go
```diff
@@ -255,6 +259,11 @@ func (s *BackendScheduler) Next(ctx context.Context, req *tempopb.NextJobRequest
 
 		return resp, nil
 	case <-timeoutCtx.Done():
+		// Add a jitter after timeout so all workers don't all retry at the same time
```
zalegrala, line 262: In thinking about the conditions for this to help, I think in both the scheduelr crash condition and the normal operations, jitter would be better applied to the backendworker.

The only think the scheduler can do if it runs out of memory or panics is crash, which would signal a disconnect at the worker and immediately retry (after the 100ms initial delay).  I believe the backoff is deterministic in its delay incrementing, so jitter here would help the crash case, but also potentially the normal operations.  

What do you think about moving this to the worker?

electron0zero: yeah, we can move this to worker.

electron0zero: @zalegrala updated to move the jitter into the worker so now all calls to backendscheduler are with jitter.

zalegrala (approved):
Nice turnaround. Let's hope for trickling herds.

PR #5153, opened 2025-05-22
fix: apply the artificial latency even on pushTrace errors

javiermolinar:
**What this PR does**:
The artificial latency was not being applied on errors. For Tempo clusters with lot of refused spans this means that the artificial latency is not always used

**Checklist**
- [X] Tests updated


mapno (approved):

mapno:
cc/ @yvrhdn

PR #5155, opened 2025-05-22
tempo-query: Use Storage v2 (WIP)

denysvitali:
**What this PR does**:

Starting from Jaeger v2.6.0 the v1 GRPC remote storage is not supported anymore and v2 should be used instead.

This PR implements the new API and fixes a few bugs:
- Operation Name is now dependant on the selected service
- tempo-query now exports the traces via `autoexport`

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


knylander-grafana:
Greetings. Thank you for your additions! Do any of your updates provide changes that a Tempo user could access? For example, does the API provide new endpoints? If so, do we need to add docs to cover those new capabiliities?

denysvitali:
> Do any of your updates provide changes that a Tempo user could access?

To the user this change is mostly transparent. `tempo-query` is just a glue between Tempo's backend and Jaeger. This is possible thanks to the [Remote Storage API](https://www.jaegertracing.io/docs/next-release-v2/storage/#remote-storage) which, since [v2.6.0](https://github.com/jaegertracing/jaeger/releases/tag/v1.69.0) (see the breaking change PR https://github.com/jaegertracing/jaeger/pull/6969) only supports the Storage API v2.  
  
`tempo-query`, before this PR, was exposing the Storage API v1 and the implementation was also rather broken (e.g: operations weren't scoped to the service, filtering was partially broken, tempo-query wasn't set up to send its internal traces back to any backend, ...).

This PR is pretty much a complete overhaul of the component. I've been using this internally at our company for a few days now and it seems to work great.  
  
The code lacks E2E tests and unit tests, but so did the original code too 😅.

joe-elliott:
@pavolloffay @andreasgerstmayr 

since RH primarily use tempo-query are one of you all able to provide a review here?

denysvitali:
@joe-elliott Can you please remove the `stale` label?

joe-elliott:
> @joe-elliott Can you please remove the stale label?

You just did :). Thank you for the ping. I had forgotten about this.

It seems no one from RH is able to review this. I've also caught wind that they are also removing support for tempo-query. It seems you use it still, yes? I can try to find some time to review it soon. Does it break compatibility with any version of Jaeger?

denysvitali:
Yes, newer versions of Jaeger do not support the Storage API v1 anymore. Without this, you won't be able to use the latest version of Jaeger

joe-elliott:
Thank you for this PR it seems very comprehensive. I've noticed you added a docker-compose and other supporting files in `/cmd/tempo-query`. If possible I would prefer to add an docker-compose example in `/example/docker-compose`. You could copy the `/cmd/docker-compose/local` example and add tempo-query and jaeger-query. This would simplify my review quite a bit as well.

Also, we need to list this as a breaking change with whatever version of Jaeger tempo-query would no longer be compatible with.

Thank you for this work! 🙏

PR #5145, opened 2025-05-20
TraceQL Metrics: right exemplars for histogram and quantiles

ruslan-mikhailov:
## What this PR does

1. Fixes a bug with unmarshalling of TraceQL Metrics response in tests.
2. Fixes a bug when for histogram and quantiles exemplars were put in wrong series.

### Root cause
In this loop: https://github.com/ruslan-mikhailov/tempo/blob/17e20a43af0881ef0841d2e33eaeb190c422e549/pkg/traceql/engine_metrics.go#L1477
it iterates over the series set while putting all the exemplars into each resulting series without any filtration: https://github.com/ruslan-mikhailov/tempo/blob/17e20a43af0881ef0841d2e33eaeb190c422e549/pkg/traceql/engine_metrics.go#L1488

For example, if in total, we have the following exemplars:
E1 - span_attr=val_A
E2 - span_attr=val_A
E3 - span_attr=val_B
E4 - span_attr=val_C
For the following query:
```
{span.span_attr!=nil} | quantile_over_time(span:duration, 0.9) by(span.span_attr)
```

we have:
val_A time series exemplars: E1,E2,E3,E4
val_B time series exemplars: E1,E2,E3,E4
val_C time series exemplars: E1,E2,E3,E4

Expected results:
val_A time series exemplars: E1,E2
val_B time series exemplars: E3
val_C time series exemplars: E4

### Impact
In Grafana UI, after filtering by series, it provides wrong exemplars from another series, misleading users and providing a bad UX.

### Unit tests (in the PR)
Check that all exemplars have an attribute label that matches the corresponding time series label.

### Manual testing
1. Run local tempo using local docker compose from `example/docker-compose/local`
2. Run the following script to generate random traces

<details>
  <summary>Python script</summary>
  

```python
import random
import time
from datetime import datetime
from opentelemetry import trace
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import BatchSpanProcessor
from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
from opentelemetry.sdk.resources import Resource
from opentelemetry.semconv.resource import ResourceAttributes

# Initialize trace provider
resource = Resource.create({
    ResourceAttributes.SERVICE_NAME: "fake-service",
    ResourceAttributes.SERVICE_VERSION: "0.1.0",
})

provider = TracerProvider(resource=resource)
[...]

PR #5158, opened 2025-05-23
Exemplars UX improvements

### LINKED ISSUE #4632: Query exemplars field not respected

### LINKED ISSUE #4856: TraceQL Metrics exemplar distribution is skewed
**Describe the bug**
Exemplars from TraceQL metrics are not homogeneously distributed across the series but instead cluster towards the right/left side, leading to potential biases in analysis.

Exemplars sometimes appear on the left side and other times on the right, meaning the sampling decision directly impacts their distribution

Image

Showing all the series:

Image

**To Reproduce**

1. Run a TraceQL metrics query that returns many series with exemplars
2. Check the exemplars distribution

**Expected behavior**
Exemplars are distributed homogeneously

ruslan-mikhailov:
**What this PR does**:

1. Distribute exemplars over time.
a. Change the approach to calculating requested exemplars from frontend to querier. Instead of equal share, calculate it based on block time range. The biggest block will receive more exemplars than a small block
b. Change sampling algorithm by fixing number of sampling buckets

Before the change:
image

After the change:
image

**Results:**
Before:
Pasted image 20250523145320

After:
Pasted image 20250523144347

2. Hard limit returned exemplars from query-frontend during final aggregation.

Fixes #4856
Fixes #4632

**Checklist**

- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/metrics_query_range_sharder_test.go
+++ b/modules/frontend/metrics_query_range_sharder_test.go
@@ -280,6 +280,7 @@ func FuzzExemplarsPerShard(f *testing.F) {
 	})
 }
 
+// nolint: gosec // G115
 func TestExemplarsCutoff(t *testing.T) {
 	s := &queryRangeSharder{}
 	now := time.Now()
--- a/pkg/traceql/engine_metrics_test.go
+++ b/pkg/traceql/engine_metrics_test.go
@@ -1910,7 +1910,7 @@ func BenchmarkHistogramAggregator_Combine(b *testing.B) {
 			series := generateTestTimeSeries(seriesCount, bm.samplesCount, bm.exemplarCount, req.Start, req.End)
 
 			for b.Loop() {
-				agg := NewHistogramAggregator(req, []float64{0.5, 0.9, 0.99}, uint32(bm.exemplarCount))
+				agg := NewHistogramAggregator(req, []float64{0.5, 0.9, 0.99}, uint32(bm.exemplarCount)) // nolint: gosec // G115
 				agg.Combine(series)
 			}
 		})


modules/frontend/metrics_query_range_sharder.go
```diff
@@ -228,8 +263,8 @@ func (s *queryRangeSharder) buildBackendRequests(ctx context.Context, tenantID s
 	queryHash := hashForQueryRangeRequest(&searchReq)
 	colsToJSON := api.NewDedicatedColumnsToJSON()
 
-	exemplarsPerBlock := s.exemplarsPerShard(uint32(len(metas)), searchReq.Exemplars)
-	for _, m := range metas {
+	exemplarsPerBlock := s.exemplarsPerShard(metas, searchReq.Exemplars)
```
mdisibio, line 266: Should we also scale the number of exemplars for generatorRequest based on the time range?  I think it sends over the whole limit, which was also a reason exemplars seemed to clump in recent data.

knylander-grafana:
@ruslan-mikhailov  Do we need to update the docs so people understand the changes you've made in this PR? Maybe this doc? https://grafana.com/docs/tempo/latest/traceql/metrics-queries/#exemplars

integration/e2e/api/query_range_test.go
```diff
@@ -64,34 +64,67 @@ sendLoop:
...
+			"{} | count_over_time() by (status)",
+			"{status != error} | count_over_time() by (status)",
+		} {
+			t.Run(fmt.Sprintf("%s: %s", exeplarsCase.name, query), func(t *testing.T) {
+				queryRangeRes := callQueryRange(t, tempo.Endpoint(tempoPort), query, exeplarsCase.exemplars, debugMode)
+				require.NotNil(t, queryRangeRes)
+				require.GreaterOrEqual(t, len(queryRangeRes.GetSeries()), 1)
+				if query == "{} | quantile_over_time(duration, .5, 0.9, 0.99)" {
+					t.Skip("Bug in quantile_over_time in calculating exemplars")
+				}
```
ruslan-mikhailov, line 113: This is not https://github.com/grafana/tempo/pull/5145 , it's another issue. I need to create a bug and add it in code as a comment

ruslan-mikhailov: https://github.com/grafana/tempo/issues/5167

integration/e2e/limits_test.go
```diff
@@ -239,6 +239,9 @@ func TestQueryLimits(t *testing.T) {
 
 	require.Eventually(t, func() bool {
 		_, err = client.QueryTrace(tempoUtil.TraceIDToHexString(traceID[:]))
+		if err == nil {
+			return false
+		}
```
mapno, line 244: Is this correct? The trace might not be immediately too big because it's sent in batches.

modules/frontend/metrics_query_range_sharder.go
ruslan-mikhailov: Good point! With bucket fix, it still should distribute fairly, but by cutting the exemplars parameter, we can reduce the load to generators.

I pushed exemplar splitting between generator and querier in a separate commit. Could you please take a look?

mapno (approved):
LGTM

ruslan-mikhailov:
\+ rebase from latest main

ruslan-mikhailov:
\+ linter fixes

integration/e2e/api/query_range_test.go
ruslan-mikhailov: \+

PR #5170, opened 2025-05-28
Assert max live traces limits in local-blocks processor

mapno:
**What this PR does**:

Asserts max live traces limit in in the non-queue implementation of the localblocks processor.

This helps protect the generator from excessive memory and CPU consumption.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


ruslan-mikhailov (approved):

joe-elliott (approved):
nice, i like the simplification of removing the config option and making both paths behave the same.

PR #5185, opened 2025-05-29
linter: disable gosec rule g115

carles-grafana:
G115: Potential integer overflow when converting between integer types

This rule introduced a lot of noise as we have a lot of cases where we truncate or change the sign of an integer type. In addition, in many cases protobuf definitions are involved, which require regeneration. But most importantly, `len` always returns `int`, which means we should add a conditional every time we need to convert it, potentially affecting performance.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -29,6 +29,9 @@ linters:
         - (github.com/go-kit/log.Logger).Log
     errorlint:
       errorf: false
+    gosec:
+      excludes:
+        - G115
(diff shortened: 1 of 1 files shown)


mapno (approved):
LGTM

carles-grafana:
There's more support now, so reopening

PR #5186, opened 2025-05-29
Fix distributor rebatch bug

mdisibio:
**What this PR does**:
Distributors rebatch incoming writes by trace ID, in order to ensure that all spans for a given trace end up on the same ingesters over time.  It's using a 32-bit hash which can lead to collisions.  When 2 traces in the incoming write request have the same hash (collide), their spans will get intermixed, i.e. bad data.

The core issue is that the distributor logic was conflating hashing for the ring, and hashing for dedupe.  These don't need to be the same hash, nor should they be.  Because the ring requires 32-bit hashes and collisions there don't matter, it just means 2 traces go the same ingesters (normal and unavoidable).   

But collisions for dedupe must be avoided, and we do this by swapping that part to a 64-bit hash.  Same approach as in trace combiner for spans, and there are test cases for known trace ID collisions under the old method and a test to check the collision rate of the new method.  

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/distributor.go
+++ b/modules/distributor/distributor.go
@@ -42,8 +42,7 @@ import (
 	v1_common "github.com/grafana/tempo/pkg/tempopb/common/v1"
 	v1 "github.com/grafana/tempo/pkg/tempopb/trace/v1"
 	"github.com/grafana/tempo/pkg/usagestats"
-	tempo_util "github.com/grafana/tempo/pkg/util"
-
+	"github.com/grafana/tempo/pkg/util"
 	"github.com/grafana/tempo/pkg/validation"
 )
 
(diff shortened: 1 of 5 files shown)


modules/distributor/distributor.go
```diff
@@ -800,15 +799,15 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount, ma
...
 
-	keys := make([]uint32, 0, len(tracesByID))
+	ringTokens := make([]uint32, 0, len(tracesByID))
 	traces := make([]*rebatchedTrace, 0, len(tracesByID))
 
-	for k, r := range tracesByID {
-		keys = append(keys, k)
-		traces = append(traces, r)
+	for _, tr := range tracesByID {
+		ringTokens = append(ringTokens, util.TokenFor(userID, tr.id))
```
joe-elliott, line 806: nice. this prevents traces from shuffling on rollout

modules/distributor/distributor.go
```diff
@@ -472,7 +471,7 @@ func (d *Distributor) PushTraces(ctx context.Context, traces ptrace.Traces) (*te
 
 	maxAttributeBytes := d.getMaxAttributeBytes(userID)
 
-	keys, rebatchedTraces, truncatedAttributeCount, err := requestsByTraceID(batches, userID, spanCount, maxAttributeBytes)
+	ringTokens, rebatchedTraces, truncatedAttributeCount, err := requestsByTraceID(batches, userID, spanCount, maxAttributeBytes)
```
joe-elliott, line 474: +1 to the rename

joe-elliott (approved):

modules/distributor/distributor.go
mdisibio: Yes this preserves the existing trace ID sharding logic. There is no reason to change it.

PR #5196, opened 2025-06-02
Exclude parent span id from selectAll

mdisibio:
**What this PR does**:
When we added the intrinsic for parent span ID, it got automatically included in selectAll for the compare() function. But it isn't useful and excluding it will save significant cpu/mem.

```
                                                                      │ before.txt │             after.txt             │
                                                                      │   sec/op   │   sec/op    vs base               │
BackendBlockQueryRange/{_name_!=_nil_}_|_compare({status=error})/5-14   3.353 ± 2%   2.879 ± 3%  -14.12% (p=0.002 n=6)

                                                                      │ before.txt │             after.txt             │
                                                                      │  MB_IO/op  │  MB_IO/op   vs base               │
BackendBlockQueryRange/{_name_!=_nil_}_|_compare({status=error})/5-14   26.21 ± 0%   19.64 ± 0%  -25.07% (p=0.002 n=6)

                                                                      │ before.txt  │           after.txt            │
                                                                      │  spans/op   │  spans/op    vs base           │
BackendBlockQueryRange/{_name_!=_nil_}_|_compare({status=error})/5-14   819.6k ± 0%   819.6k ± 0%  ~ (p=1.000 n=6) ¹
¹ all samples are equal

                                                                      │   spans/s   │   spans/s    vs base               │
BackendBlockQueryRange/{_name_!=_nil_}_|_compare({status=error})/5-14   244.5k ± 2%   284.7k ± 3%  +16.45% (p=0.002 n=6)

                                                                      │  before.txt   │              after.txt               │
                                                                      │     B/op      │     B/op       vs base               │
BackendBlockQueryRange/{_name_!=_nil_}_|_compare({status=error})/5-14   807.4Mi ± 27%   403.0Mi ± 42%  -50.08% (p=0.002 n=6)

BackendBlockQueryRange/{_name_!=_nil_}_|_compare({status=error})/5-14   6.359M ± 3%   2.746M ± 4%  -56.83% (p=0.002 n=6)
```

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/encoding/vparquet3/block_traceql.go
+++ b/tempodb/encoding/vparquet3/block_traceql.go
@@ -1660,9 +1660,11 @@ func createSpanIterator(makeIter makeIterFn, primaryIter parquetquery.Iterator, 
 			if entry.scope != intrinsicScopeSpan {
 				continue
 			}
-			// These intrinsics aren't included in select all because I say so.
+			// These intrinsics aren't included in select all because they
+			// aren't useful for compare().
 			switch intrins {
 			case traceql.IntrinsicSpanID,
+				traceql.IntrinsicParentID,
 				traceql.IntrinsicSpanStartTime,
 				traceql.IntrinsicStructuralDescendant,
 				traceql.IntrinsicStructuralChild,
--- a/tempodb/encoding/vparquet4/block_traceql.go
+++ b/tempodb/encoding/vparquet4/block_traceql.go
@@ -2036,9 +2036,11 @@ func createSpanIterator(makeIter makeIterFn, innerIterators []parquetquery.Itera
 			if entry.scope != intrinsicScopeSpan {
 				continue
 			}
-			// These intrinsics aren't included in select all because I say so.
+			// These intrinsics aren't included in select all because they
+			// aren't useful for compare().
 			switch intrins {
 			case traceql.IntrinsicSpanID,
+				traceql.IntrinsicParentID,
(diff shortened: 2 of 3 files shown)


joe-elliott:
this may be beyond the scope of your change, but i'm seeing other suspect returned values:

image

and

image

mdisibio:
> this may be beyond the scope of your change, but i'm seeing other suspect returned values:

The switch statement already excludes [nestedSetParent](https://github.com/grafana/tempo/pull/5196/files#diff-a201423ab0b50d4455a497bf1804b1a9f596394413c28b7702710f89237c49c1R2050), strange, yeah can take a look.

mdisibio:
Pushed 2 fixes:

1. nestedSetParent - This was being returned because it was part of the first pass.  Drilldown runs queries like `{nestedSetParent<0 && true} | compare({status = error}, 10)`. So the fix was to exclude grouping on it in the compare() function itself.

2. instrumentation/resource. unnamed attributes - This was because a nil unnamed attribute was being created by attributeCollector unintentionally, for any scopes that had zero generic attributes. Because if all keys and values are nil, it's encoded as a single nil at the parent definition level.  I think this has been around a while, but rare for span and resource, and really first shown after we added the instrumentation scope.  There are a couple ways to filter this out, but I think using the skip nils predicate in createAttributeIterator is the best.

PR #5198, opened 2025-06-02
Add nil check to `partitionAssignmentVar`

mapno:
**What this PR does**:

Adds nil check for `partitionAssignmentVar.String()`.

It used to panic when Tempo was run with `--help` and the config wasn't set.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/blockbuilder/config.go
```diff
@@ -106,5 +106,8 @@ func (p *partitionAssignmentVar) Set(s string) error {
 }
 
 func (p *partitionAssignmentVar) String() string {
+	if p.p == nil {
+		return "map[]"
```
carles-grafana, line 110: would "{}" be clearer?

mapno: The string representation of an empty map in Go is `map[]` (https://go.dev/play/p/RNM_DdmfxbS), IMO that's clearest.

PR #5202, opened 2025-06-02
fix tempo configuration options that are always overrided with config overrides section

### LINKED ISSUE #4492: traces_service_graph_request_messaging_system_seconds metrics is not produced, while enabled.
I am running Tempo 2.6.1.
My full Tempo config can be found here: https://github.com/cbos/observability-toolkit/blob/main/config/tempo/tempo-config.yaml

For the service graph processor I have I this config:

```yaml
  processor:
    service_graphs:
      wait: 30s
      max_items: 10000
      workers: 10
      dimensions: []
      #enable_client_server_prefix: false
      peer_attributes:
        - peer.service
        - db.name
        - db.system
        - messaging.destination.name
      # Enables additional labels for services and virtual nodes. Default false
      enable_virtual_node_label: true
      # If enabled another histogram will be produced for interactions over messaging systems middlewares
      # If this feature is relevant over long time ranges (high latencies) - consider increasing
      # `wait` value for this processor. Default false
      # enable_messaging_system_latency_histogram: false
      enable_messaging_system_latency_histogram: true
```

All metrics are generated:
Image

But `traces_service_graph_request_messaging_system_seconds` is missing. That is introduced with #3453 

My test setup is a modified setup of OpenTelemetry demo:
https://github.com/cbos/opentelemetry-demo/tree/jms-support

Producer
https://github.com/cbos/opentelemetry-demo/tree/jms-support/src/adservice
JMS system is ActiveMQ (messagequeue in the setup)
Consumer
https://github.com/cbos/opentelemetry-demo/tree/jms-support/src/analytics

Image
So the client-server matching is done correctly and the connection_type is 'messaging_system'.

https://github.com/grafana/tempo/blob/deece3c9084a56988361be9d0c44b73911a2210e/modules/generator/processor/servicegraphs/servicegraphs.go#L380

I checked the logging of Tempo, but I don't see the warning "producerSpanEndTime must be smaller than consumerSpanStartTime. maybe the peers clocks are not synced"

Based on what I can see all conditions are met to create the metrics, but these are not produced.
What can I do to find the problem to get it fixed?

KyriosGN0:

**What this PR does**:
fixes some configuration options that were always overrode with Config Override (meaning if we wanted to enable them we had to do it with the overrides configs

Fixes #4492

**Checklist**
- [X] Tests updated
- [ ] Documentation added


knylander-grafana:
Greetings. Does your PR add or modify options that a Tempo user can access? i.e., is there user-facing impact? If so, please consider updating the docs.

KyriosGN0:
> Greetings. Does your PR add or modify options that a Tempo user can access? i.e., is there user-facing impact? If so, please consider updating the docs.

Hi, no it doesn't it only fixes an issue that sometimes users had to use the override config to set certain options

mapno (approved):
LGTM. If you can solve the conflict I'll merge

modules/generator/config.go
```diff
@@ -211,17 +211,29 @@ func (cfg *ProcessorConfig) copyWithOverrides(o metricsGeneratorOverrides, userI
...
 
-	copyCfg.SpanMetrics.DimensionMappings = o.MetricsGeneratorProcessorSpanMetricsDimensionMappings(userID)
+	if dimensionMappings := o.MetricsGeneratorProcessorSpanMetricsDimensionMappings(userID); dimensionMappings != nil {
+		copyCfg.SpanMetrics.DimensionMappings = dimensionMappings
+	}
 
-	copyCfg.SpanMetrics.EnableTargetInfo = o.MetricsGeneratorProcessorSpanMetricsEnableTargetInfo(userID)
+	if enableTargetInfo := o.MetricsGeneratorProcessorSpanMetricsEnableTargetInfo(userID); enableTargetInfo {
+		copyCfg.SpanMetrics.EnableTargetInfo = enableTargetInfo
+	}
```
mapno: @joe-elliott just noticed this doesn't work well because it's a boolean—this means that `target_info` couldn't be disabled via overrides.

For this to work, you'll need to use a pointer for the value like in `artificial_latency` ([code](https://github.com/grafana/tempo/blob/main/modules/overrides/runtime_config_overrides.go#L332-L338))

mapno (changes requested):

modules/generator/config.go
KyriosGN0: @mapno not sure i understand the required change, do i need to update the  `target_info` and other boolean option that i updated in this PR to use a similiar logic to `artificial_latency` ? and if so where, in this file or in https://github.com/grafana/tempo/blob/main/modules/overrides/runtime_config_overrides.go

mapno: For those overrides to be handled correctly, yes, that's needed. You'll need to change them in to a pointer in `modules/overrides/config.go` and change the method when they are called in `modules/overrides/runtime_config_overrides.go`.

You can check the changes to the `overrides` pkg in this PR https://github.com/grafana/tempo/pull/4899/files#diff-b9c8b28c611e5914bcd7772a763148dcbc631c4260ee1c946cfc37af8efdb616 as reference.

KyriosGN0:
@mapno i have address your requested changes, i also removed the dimension mapping changes, since i wanted to focus on booleans options in this PR, 
i also wanted to ping/ask about the micro-service docker-compose in the integration folder, the synthetic-load-generator container can't start since its image doesn't exist in dockerhub as far as i can see, it is working for you guys ?

KyriosGN0:
Hey @mapno, would it be possible to get another look at my PR?

mapno (commented):
It still has the same problem that you can't differentiate if the parameter was set to false or not set. You need to change the signature of the overrides methods (eg. `MetricsGeneratorProcessorSpanMetricsEnableTargetInfo`) to return another boolean to signal presence—ie. it was set or not.

KyriosGN0:
hey @mapno, i have addressed your comments (took me a bit to understand it), i could not setup the specific issue in the original issue, so instead i added  a E2E test to make sure this works, LMK if this is ok. 
Thanks

modules/overrides/interface.go
```diff
@@ -67,10 +67,10 @@ type Interface interface {
 	MetricsGeneratorProcessorLocalBlocksFlushCheckPeriod(userID string) time.Duration
 	MetricsGeneratorProcessorLocalBlocksCompleteBlockTimeout(userID string) time.Duration
 	MetricsGeneratorProcessorSpanMetricsDimensionMappings(userID string) []sharedconfig.DimensionMappings
-	MetricsGeneratorProcessorSpanMetricsEnableTargetInfo(userID string) bool
+	MetricsGeneratorProcessorSpanMetricsEnableTargetInfo(userID string) (bool, bool)
 	MetricsGeneratorProcessorServiceGraphsEnableClientServerPrefix(userID string) bool
-	MetricsGeneratorProcessorServiceGraphsEnableMessagingSystemLatencyHistogram(userID string) bool
-	MetricsGeneratorProcessorServiceGraphsEnableVirtualNodeLabel(userID string) bool
+	MetricsGeneratorProcessorServiceGraphsEnableMessagingSystemLatencyHistogram(userID string) (bool, bool)
+	MetricsGeneratorProcessorServiceGraphsEnableVirtualNodeLabel(userID string) (bool, bool)
```
mapno: Could you add a comment explaining what each value means? Ie. which is the value and which is the existence check

mapno (approved):
LGTM, just one minor comment. Thank you!

modules/overrides/interface.go
KyriosGN0: @mapno Thanks, added a comment per function

KyriosGN0:
@mapno can i get another round of CI? the lint passes locally for me, guess it was some temp issue ? 
Thanks in advance

KyriosGN0:
Really weird, the lint failure is not reproducing for me locally (running the same `make lint base=origin/main`), same for e2e test, maybe its some local cache on the github runners?

mapno:
I've merged `main`, let's see if that fixes the issue. I have no idea what happened.

integration/e2e/metrics_generator_test.go
```diff
@@ -22,10 +22,110 @@ import (
...
+	isServiceActiveMatcher := func(service string) []*labels.Matcher {
+		return []*labels.Matcher{
+			labels.MustNewMatcher(labels.MatchEqual, "name", service),
+			labels.MustNewMatcher(labels.MatchEqual, "state", "ACTIVE"),
+		}
+	}
+	require.NoError(t, tempoDistributor.WaitSumMetricsWithOptions(e2e.Equals(1), []string{"tempo_ring_members"}, e2e.WithLabelMatchers(isServiceActiveMatcher("ingester")...), e2e.WaitMissingMetrics))
+	require.NoError(t, tempoDistributor.WaitSumMetricsWithOptions(e2e.Equals(1), []string{"tempo_ring_members"}, e2e.WithLabelMatchers(isServiceActiveMatcher("metrics-generator")...), e2e.WaitMissingMetrics))
+
+	c, err := util.NewJaegerGRPCClient(tempoDistributor.Endpoint(14250))
```
mapno: The signature changed

```suggestion
	c, err := util.NewJaegerToOTLPExporter(tempoDistributor.Endpoint(14250))
```

KyriosGN0:
> I've merged `main`, let's see if that fixes the issue. I have no idea what happened.

Thanks, i will look into the lint and E2E failure

KyriosGN0:
i fixed the E2E test, but i have no idea why the lint fails, the same line exist in 2 other test in the same file but the linter doesn't complain about them.

KyriosGN0:
@mapno i will fix the linter issue and let u know

KyriosGN0:
Im thinking of just adding `#nosec G404` to this line, i don't see a the point trying to generate via the `crypto/rand` as its just suppose to give me a trace id for testing, 
if its ok with you guys I will commit this, if not I will switch it to `crypto/rand`

mapno:
> Im thinking of just adding `#nosec G404` to this line, i don't see a the point trying to generate via the `crypto/rand` as its just suppose to give me a trace id for testing, if its ok with you guys I will commit this, if not I will switch it to `crypto/rand`

Sounds good to me. That linter is too strict

KyriosGN0:
@mapno Thank, done!

mapno (approved):
LGTM. Thank you for sticking with it!

PR #5212, opened 2025-06-04
Tempo MCP Server 🤖 

joe-elliott:
**What this PR does**:
Adds an MCP server shim in the query frontend using [mcp-go](https://github.com/mark3labs/mcp-go). This API layer allows LLMs and other agents access to tracing data. Due to privacy/security concerns it is disabled by default and warns the user if enabled.

Also, I reorganized the TraceQL docs to serve them directly from MCP. As seen in the screenshot it can give up to date information to an LLM to correct syntax errors in its queries. Unsure if what I did is brittle and awful or dynamic and powerful. You decide!

Allows for fun interactions like:
claude-code-tempo-mcp

**To test locally**:
This can be tested with local test data using Claude code and our docker examples. Refer to the [documentation in PR](https://github.com/grafana/tempo/pull/5212/files#diff-832f7cd7cd7daaf25828a410a6bfec68bfbb0ca0e013c911348334273eb59646R30-R32)

**Checklist**
- [x] Tests updated
- [x] Documentation added


pkg/api/http.go
```diff
@@ -80,6 +80,7 @@ const (
 	PathSpanMetricsSummary  = "/api/metrics/summary"
 	PathMetricsQueryInstant = "/api/metrics/query"
 	PathMetricsQueryRange   = "/api/metrics/query_range"
+	PathMCP                 = "/mcp"
```
electron0zero: should we have  MCP server under a path like `/mcp/tempo` or `/mcp/traceql` because this is a traceql MCP server? and it keeps the door open to for more MCP servers into the tempo under one API path.

joe-elliott: The mcp server represents all actions we want to expose to an LLM agent so it doesn't just represent traceql or trace by id lookup. The path change I am considering is whether or not should be offered as `api/mcp`

joe-elliott: Decided to change the path to `/api/mcp`. This allows us to rev the mcp path if we need to. For instance if mcp updates it's protocol we may want to continue serving the old on `/api/mcp` and serve the newer protocol on `/api/v2/mcp`

docs/sources/tempo/shared/traceql-main.md
```diff
@@ -0,0 +1,576 @@
...
+labels:
+  products:
+    - enterprise
+    - oss
+---
+
+<!-- WARNING: This file is loaded by /modules/frontend/mcp.go and served to LLMs through the MCP protocol. It -->
+<!-- should be kept terse and to the point. Links and videos and such are bad. Examples are good! The below "mcp-cutoff" 
+<!-- is used to remove everything above. -->
+<!-- TODO: Anthropic suggests using xml tags to organize content delivered to an LLM. Can we tag this content with xml tags  -->
```
knylander-grafana: @grafsean @jdbaldry  Do you the answer to lines 13-14?

jdbaldry: I believe any XML tags in Markdown source will be treated as raw HTML according to the CommonMark spec. This means we'll be injecting extra, non-standard HTML into the website and that's generally considered bad for accessibility and can cause oddities in browser rendering.

If possible, it would be better for the MCP loader to insert the XML tags if it can. Perhaps we could use HTML comments to include them in the source file in a way that won't affect rendering:

```
<!-- <instructions> -->
Some instructions
<!-- </instructions> -->
```

cbos:
@joe-elliott 
How is this related to https://github.com/grafana/mcp-grafana?
For Prometheus, Loki and Grafana itself it is added as separate MCP, not part of Loki for example.
To me it would sound logical to add Tempo support the the same MCP to have 1 MCP configured to access the whole LGTM stack. 

In a few days I will be having a Hackathon where I will be exploring the Grafana MCP to see how it can help to do analysis on problems in a software stack.
If you have some suggestions on how to use this, I can try this Tempo MCP as well and let you know the outcome.

joe-elliott:
For databases willing to support their own MCP server mcp-grafana should simply read tools, append them to its internal list and proxy them back to the caller. By supporting this protocol directly Tempo can more easily update and test new tools as we add new endpoints and capabilities.

docs/sources/tempo/shared/traceql-main.md
joe-elliott: thanks for the insight. definitely a future topic as we improve here 🙏

docs/sources/tempo/api_docs/mcp-server.md
```diff
@@ -0,0 +1,36 @@
...
+
+To experiment with the MCP server using dummy data and Claude Code:
+
+1. Run the local docker-compose example in `/example/docker-compose`. This will expose the mcp server at http://localhost:3200/api/mcp
+1. Run `claude mcp add -t stdio -s user tempo npx mcp-remote http://localhost:3200/api/mcp` to add a reference to Claude Code.
+1. Run `claude` and ask some questions
+
+This MCP server has also been tested succesfully in cursor using the [`mcp-remote`](https://www.npmjs.com/package/mcp-remote) package.
+
+![Claude Code interacting with the Tempo MCP server](/static/img/docs/tempo/claude-code-tempo-mcp.png)
```
ruslan-mikhailov, line 36: How images are uploaded? I do not see them in the PR

modules/frontend/mcp_tools.go
```diff
@@ -0,0 +1,361 @@
...
+	parsed, err := traceql.Parse(query)
+	if err != nil {
+		return mcp.NewToolResultError(fmt.Sprintf("query parse error. Consult TraceQL docs tools: %v", err)), nil
+	}
+
+	if parsed.MetricsPipeline != nil || parsed.MetricsSecondStage != nil {
+		return mcp.NewToolResultError("TraceQL metrics query received on traceql-search tool. Use the traceql-metrics-instant or traceql-metrics-range tool instead"), nil
+	}
+
+	// nolint:gosec // G115
```
ruslan-mikhailov: We excluded G115, should be safe to remove nolint

modules/frontend/mcp_tools.go
```diff
@@ -0,0 +1,361 @@
...
+	body, err := handleHTTP(ctx, s.frontend.SearchTagsValuesV2Handler, req)
+	if err != nil {
+		return mcp.NewToolResultError(err.Error()), nil
+	}
+
+	return toolResult(body, MetaTypeAttributeValues, "json", "2"), nil
+}
+
+func handleHTTP(ctx context.Context, handler http.Handler, req *http.Request) (string, error) {
+	rw := httptest.NewRecorder()
```
ruslan-mikhailov: Is it ok to use httptest outside of tests?

modules/frontend/mcp_tools.go
```diff
@@ -0,0 +1,361 @@
...
+}
+
+func (s *MCPServer) handleTraceQLMetrics(_ context.Context, _ mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+	level.Info(s.logger).Log("msg", "traceql metrics tool requested")
+
+	return toolResult(trimDocs(docs.TraceQLMetrics), MetaTypeDocumentation, "markdown", "1"), nil
+}
+
+// handleSearch handles the traceql-search tool
+func (s *MCPServer) handleSearch(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
```
electron0zero, line 56: should we log or capture the user-agent or some other metrics to know which Tempo API calls (search, metadata, etc) are coming from MCP and which ones are from grafana

docs/sources/tempo/api_docs/mcp-server.md
joe-elliott: they are uploaded out of band. i can share the process internally.

modules/frontend/mcp_tools.go
joe-elliott: good catch. removed!

modules/frontend/mcp_tools.go
joe-elliott: i don't see a whole lot of value in logging the user agent here. these are just http calls. we could do that farther up the chain if desired.

added a metric to count mcp tool calls by type

modules/frontend/mcp_tools.go
joe-elliott: i don't know?  we do it here: https://github.com/grafana/dskit/blob/main/httpgrpc/server/server.go#L82 but that doesn't mean it's a good choice.

Since the mcp response is a different shape than the response provided by the frontend http pipeline I think we have to buffer the response fully and then send it back into the mcp-go framework. I could write a buffered ResponseWriter implementation but I think it would basically be the same as the httptest one?

wdyt?

joe-elliott: Replaced the httptest response recorder with a "responseBuffer" which implements the http.ResponseWriter interface.

docs/sources/tempo/api_docs/_index.md
```diff
@@ -53,6 +53,7 @@ For externally supported gRPC API, [refer to Tempo gRPC API](#tempo-grpc-api).
 | [Compactor ring status](#compactor-ring-status) | Compactor |  HTTP | `GET /compactor/ring` |
 | [Status](#status) | Status |  HTTP | `GET /status` |
 | [List build information](#list-build-information) | Status |  HTTP | `GET /api/status/buildinfo` |
+| [MCP Server](https://grafana.com/docs/tempo/<TEMPO_VERSION>/api_docs/mcp-server) (*) | MCP |   | `/api/mcp` |
```
electron0zero, line 56: Question/nit: should we have a section in docs page for MCP endpoint like we do for others? or it doesn't make sense?

electron0zero (approved):
lgtm, have a Q, but that's a nit, so approving :shipit:

docs/sources/tempo/api_docs/_index.md
joe-elliott: I don't know. I wasn't sure what to say about it b/c I didn't intend to document parameters or request/response bodies and I thought it needed more space than just a simple section like the rest of our endpoints.

PR #5220, opened 2025-06-06
new action to approve and auto-merge dependabot pull requests

javiermolinar:
it will auto-approve only if the update is a patch update


zalegrala (approved):
This looks like it might work :)  LGTM

Issue #5164, opened 2025-05-26
TraceQL metrics count_over_time yields different results for different time boundaries (and instant vs. range queries)


alexbikfalvi:
### Describe the bug

Results are different for `count_over_time()` when running an range or instant queries with different time boundaries

An example is the following, where the query is `{}|count_over_time()with(exemplars=0)` and the time range is:

* for the range query a 10 min (`1748254200`..`1748254800`) with steps of 1 min
* for the instant query a 1 min range:

|min|start|end|query_range (step=1m)|query (in 1 min steps)|
|---|---|---|---|---|
|0..1|1748254200|1748254260|11538|7102|
|1..2|1748254260|1748254320|12343|9403|
|2..3|1748254320|1748254380|13408|8722|
|3..4|1748254380|1748254440|11959|8314|
|4..5|1748254440|1748254500|11840|11327|
|5..6|1748254500|1748254560|13210|5145|
|6..7|1748254560|1748254620|12422|8473|
|7..8|1748254620|1748254680|12126|7296|
|8..9|1748254680|1748254740|12338|8419|
|9..10|1748254740|1748254800|12037|5294|
|sum|||**123221**|**79495**|

**Note 1:** For the above the range query yields a different result for the whole range with a step of 10min

* query_range: `sum(count_over_time())` for 200..800 step=1m: `123221`
* query_range: `count_over_time()` for 200..800 step=10m: `118635`
* query: `count_over_time()` for 200..800: `118635`

**Note 2:** Other than this difference, the results are deterministic, where Tempo always returns the same results for the same input parameters.

**Note 3:** Extending the range query doesn't alter it's result for the overlapping range.

|min|start|end|query_range (1748254200..1748254200)|query_range (1748254020..1748254980)|
|---|---|---|---|---|
|-3..-2|1748254020|1748254080|-|11942|
|-2..-1|1748254080|1748254140|-|13169|
|-1..0|1748254140|1748254200|-|12079|
|0..1|1748254200|1748254260|11538|11538|
|1..2|1748254260|1748254320|12343|12343|
|2..3|1748254320|1748254380|13408|13408|
|3..4|1748254380|1748254440|11959|11959|
|4..5|1748254440|1748254500|11840|11840|
|5..6|1748254500|1748254560|13210|13210|
|6..7|1748254560|1748254620|12422|12422|
|7..8|1748254620|1748254680|12126|12126|
|8..9|1748254680|1748254740|12338|12338|
[...]

PR #5252, opened 2025-06-10
[bugfix] Correct instant query calculation 

### LINKED ISSUE #5221: topk/bottomk returns more then k results in TraceQL metrics instant query
**Describe the bug**
TraceQL metrics Instant query like `{span.http.client_ip!=""} | count_over_time() by (span.http.client_ip) | topk(10)` should return max of 10 results but it returns more then 10 results.

as you can see the screenshot below, it has more then 10 results.

Image

**To Reproduce**
1. Run tempo main
2. Run `{span.http.client_ip!=""} | count_over_time() by (span.http.client_ip) | topk(10)` and select `Instant` in the query type.
3. see more then 10 results

**Expected behavior**
TraceQL metrics Instant queries should never return more then k results.

**Environment:**
 - Infrastructure: Grafana Cloud
 - Deployment tool: Grafana Cloud

**Additional Context**
None

ruslan-mikhailov:
**What this PR does**: TODO

Fixes #5221 #5164

**Checklist**

- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/engine_metrics.go
+++ b/pkg/traceql/engine_metrics.go
@@ -52,8 +52,8 @@ func DefaultQueryRangeStep(start, end uint64) uint64 {
 
 // IntervalCount is the number of intervals in the range with step.
 func IntervalCount(start, end, step uint64) int {
-	start = alignStart(start, step)
-	end = alignEnd(end, step)
+	start = alignStart(start, end, step)
+	end = alignEnd(start, end, step)
 
 	intervals := (end - start) / step
 	intervals++
@@ -61,15 +61,15 @@ func IntervalCount(start, end, step uint64) int {
 }
 
 // TimestampOf the given interval with the start and step.
-func TimestampOf(interval, start, step uint64) uint64 {
-	start = alignStart(start, step)
+func TimestampOf(interval, start, end, step uint64) uint64 {
+	start = alignStart(start, end, step)
 	return start + interval*step
 }
 
 // IntervalOf the given timestamp within the range and step.
 func IntervalOf(ts, start, end, step uint64) int {
-	start = alignStart(start, step)
-	end = alignEnd(end, step) + step
+	start = alignStart(start, end, step)
+	end = alignEnd(start, end, step) + step
 
 	if ts < start || ts > end || end == start || step == 0 {
 		// Invalid
@@ -142,7 +142,11 @@ func TrimToAfter(req *tempopb.QueryRangeRequest, before time.Time) {
 }
 
 func IsInstant(req tempopb.QueryRangeRequest) bool {
-	return req.End-req.Start == req.Step
+	return isInstant(req.Start, req.End, req.Step)
+}
+
+func isInstant(start, end, step uint64) bool {
+	return end-start == step
 }
 
 // AlignRequest shifts the start and end times of the request to align with the step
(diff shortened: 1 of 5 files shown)


pkg/traceql/engine_metrics.go
```diff
@@ -155,23 +159,30 @@ func AlignRequest(req *tempopb.QueryRangeRequest) {
...
 }
 
 // End time is rounded up to next step
-func alignEnd(end, step uint64) uint64 {
+func alignEnd(start, end, step uint64) uint64 {
 	if step == 0 {
 		return 0
 	}
+	if isInstant(start, end, step) {
+		return start
```
ruslan-mikhailov: I made a copy-paste mistake here, it should return end. Now I wonder, how it works well with that

ruslan-mikhailov: It worked because with end == start, the number of intervals is 1 due to https://github.com/ruslan-mikhailov/tempo/blob/main/pkg/traceql/engine_metrics.go#L59
If `return end` as it should be, then the number of samples will be 2. See https://github.com/ruslan-mikhailov/tempo/blob/main/tempodb/tempodb_metrics_test.go#L62

ruslan-mikhailov: To fix properly, I force instant queries to always use only one interval

pkg/traceql/engine_metrics.go
```diff
@@ -155,23 +173,30 @@ func AlignRequest(req *tempopb.QueryRangeRequest) {
...
+	}
+
 	return start - start%step
 }
 
 // End time is rounded up to next step
-func alignEnd(end, step uint64) uint64 {
+func alignEnd(start, end, step uint64) uint64 {
```
mdisibio, line 201: We chatted about this offline - this is the root cause of too many intervals both for instant and query_range.  Because we round start down, and end up, it always creates 1 more step.  This is modeling behavior of prometheus where start and end timestamps are inclusive.  

At some point we want to fix this for query_range too.  I.e. 1h with 1m steps should return 60 data points not 61.  Is it worth fixing in this PR, or are you thinking to address this in a separate PR?   I think we could revisit and simplify some of these changes after doing that.

pkg/traceql/engine_metrics_test.go
```diff
@@ -1443,6 +1450,33 @@ func TestSecondStageTopK(t *testing.T) {
 	require.Equal(t, []float64{5, 5, 5, 5, 5, 5, 5, 5}, resultBaz.Values)
 }
 
+func TestSecondStageTopKInstant(t *testing.T) {
```
mdisibio, line 1613: +1 Nice tests

electron0zero: thank you, and we caught a panic with these tests.

pkg/traceql/engine_metrics.go
electron0zero: I feel like we should decide on the a behaviour that makes sense of us, even if it's slightly different from prometheus.

we don't promise to be prometheus format compliant, we are prometheus inspired and similar so it's okay to diverge where it makes sense for us to diverge.

ruslan-mikhailov:
\+ additional fix

pkg/traceql/engine_metrics.go
ruslan-mikhailov: My suggestion is to fix it with this workaround so far to not be blocked by discussions as it fixes at least 3 issues. Then, when decision is made, we can refactor the code.

mdisibio (commented):
I'm ok with the changes as-is because the fixes are good, but I want to mention another possible direction:

The original vision I had for instant queries was to use a different range aggregator instead of StepAggregator here:  https://github.com/grafana/tempo/blob/main/pkg/traceql/ast_metrics.go#L156  StepAggregator does the work to sort values into intervals in its [Observe](https://github.com/grafana/tempo/blob/main/pkg/traceql/engine_metrics.go#L463) function.  If we made a new `InstantAggregator` it would always have and return 1 interval, and it wouldn't need to do time calculations at all.  I think this is the ideal fix, because it is simpler but also more efficient.  Similar to how we swap between Grouping or [Ungrouped aggregators](https://github.com/grafana/tempo/blob/main/pkg/traceql/engine_metrics.go#L563) based on presence of any by() clause.  

Not sure why this wasn't done originally when introducing instant queries. I think I just didn't realize the number of edge cases being introduced without it.

If you are interested in exploring this direction a bit, I think it would be worthwhile.  It should be less churn overall.

ruslan-mikhailov:
\+ linter fixes

ruslan-mikhailov:
\+ rebase from latest main in order to resolve conflicts

mdisibio (approved):

electron0zero (approved):
lgtm :shipit:

PR #5253, opened 2025-06-10
feat: handle kafka errors that require to refresh metadata

javiermolinar:
**What this PR does**:
Some Kafka errors require to refresh the client metadata:

`error
level=erro caller=metrics.go:58 msg="metric lag failed:" err="request ListOffsets has 1 separate shard errors, first: unknown broker"`

or

```
level=error caller=blockbuilder.go:224 msg="consumeCycle failed" err="request ListOffsets has 1 separate shard errors, first: unknown broker"
```

We want to ensure that the client metadata is not stale and in the case of "unknownBroker" retry the operation. It adds a backoff strategy to handle the retries.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/blockbuilder/blockbuilder.go
```diff
@@ -206,7 +206,10 @@ func (b *BlockBuilder) starting(ctx context.Context) (err error) {
 		b.kadm,
 		b.logger,
 		b.cfg.IngestStorageConfig,
-		b.getAssignedPartitions)
+		b.getAssignedPartitions,
+		func() {
+			b.kafkaClient.ForceMetadataRefresh()
+		})
```
mapno: You can pass the function directly without wrapping it.
```suggestion
		b.kafkaClient.ForceMetadataRefresh
```

modules/blockbuilder/blockbuilder.go
```diff
@@ -507,33 +510,62 @@ func formatActivePartitions(partitions []int32) string {
...
+	for boff.Ongoing() {
+		commits, endsOffsets, err = b.getPartitionOffsets(ctx)
+		if err == nil {
+			break
+		}
+		retryable := ingest.HandleKafkaError(err, b.kafkaClient.ForceMetadataRefresh)
+		if retryable {
+			boff.Wait()
+		}
+	}
```
mapno: I don't understand this loop and a similar one in `metrics.go`, if the error is non-retryable it retries it immediately anyway while the back-off is still ongoing?

javiermolinar: right, I have simplified the logic, and it always retries until there is no error or the max number of attempts is reached

mapno (approved):

PR #5257, opened 2025-06-10
[backend-scheduler] flush locally, fall back to backend

zalegrala:
**What this PR does**:

After observing the backend scheduler/worker pattern in a large environment, rate limits on the single backend file started to become an issue.  Once rate limited, most jobs would start to fail because the work flush to the backend is on the hot-path before the workers could do their jobs.

Here we update the scheduler to keep a local storage, which replaces the backend hot-path for the work storage.  A periodic flush to the backend has been included using a ticker, and a new configuration option to tune the duration.

Additionally, now that we have a process which reads the work cache out of band, it is necessary to avoid races, and so a few methods have moved to the `Work` object to allow the mutex protection.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/backendscheduler/cache.go
+++ b/modules/backendscheduler/cache.go
@@ -61,7 +61,9 @@ func (s *BackendScheduler) loadWorkCache(ctx context.Context) error {
 	workPath := filepath.Join(s.cfg.LocalWorkPath, backend.WorkFileName)
 	data, err := os.ReadFile(workPath)
 	if err != nil {
-		level.Error(log.Logger).Log("msg", "failed to read work cache from local path", "path", workPath, "error", err)
+		if !os.IsNotExist(err) {
+			level.Error(log.Logger).Log("msg", "failed to read work cache from local path", "path", workPath, "error", err)
+		}
 		return s.loadWorkCacheFromBackend(ctx)
 	}
 


modules/backendscheduler/cache.go
```diff
@@ -17,15 +19,59 @@ func (s *BackendScheduler) flushWorkCache(ctx context.Context) error {
...
 }
 
-// readSeedFile reads the cluster seed file from the object store.
+// loadWorkCache loads the work cache from the local filesystem
 func (s *BackendScheduler) loadWorkCache(ctx context.Context) error {
+	// Try to load the local work cache first, falling back to the backend if it doesn't exist.
+	workPath := filepath.Join(s.cfg.LocalWorkPath, backend.WorkFileName)
+	data, err := os.ReadFile(workPath)
+	if err != nil {
+		level.Error(log.Logger).Log("msg", "failed to read work cache from local path", "path", workPath, "error", err)
```
electron0zero: we can have file doesn't exist as a valid error here, so we need to filter that out like we do in `LocalBlock`'s GetDiskCache method that does local caching on the disk.

modules/backendscheduler/cache.go
```diff
@@ -17,15 +19,59 @@ func (s *BackendScheduler) flushWorkCache(ctx context.Context) error {
...
+	workPath := filepath.Join(s.cfg.LocalWorkPath, backend.WorkFileName)
+	err = os.WriteFile(workPath, b, 0o600)
+	if err != nil {
+		return fmt.Errorf("error writing %q: %w", workPath, err)
+	}
+
+	return nil
+}
+
+func (s *BackendScheduler) flushWorkCacheToBackend(ctx context.Context) error {
```
electron0zero, line 40: we should probably logs or expose a counter for succesful flushes and failures to flush the work cache to backend 🤔

zalegrala: I added a couple counters.  One for total and one for failed, which is enough to give us a rate we can add in an alert.

modules/backendscheduler/cache.go
zalegrala: I may have refactored this a bit since the comment.  Now if we fail to load the local file for any reason, we fall back to the backend.  How does that read to you now?

electron0zero: I think the comment applies, we should be logging an into when we don't find a file and fall back to backend? or log nothing instead of logging an error.

this will log an error each time we fallback to backend.

zalegrala: Right, we'll log an error each time we fallback, but this should happen exactly once, unless the file gets corrupted due to OOMs or some other.  Are you saying just skip the log message if the file doesn't exist?

electron0zero: ah, we are deploying it as a Statefulset where the same disk will be attached to it.

i was thinking of this as a deployment where you can have a different disk attached to it on each rollout and then it has to fallback to backend on each rollout.

I am okay with this being an error in that case.

electron0zero (approved):
lgtm :shipit:

electron0zero (approved):

Issue #5246, opened 2025-06-09
tempo_backend_scheduler_jobs_active gauge goes negative


electron0zero:
**Describe the bug**
`tempo_backend_scheduler_jobs_active` is being reported as negative value, when it should never be below zero because we don't have negative active jobs

**To Reproduce**
1. Start Tempo (SHA or version): run tempo:main
2. Perform Operations (Read/Write/Others):
- start the cluster and send data, and enable scheduler and workers
- let scheduler dispatch few jobs and let the workers start work on it.
- kill the scheduler and start it again
- watch `tempo_backend_scheduler_jobs_active` go to negative when workers report that jobs are done or failed.

See `sum(tempo_backend_scheduler_jobs_active{namespace=~"$namespace"})`
![Image](https://github.com/user-attachments/assets/ee4e7cda-e5dc-46db-ba9e-b8d8d7b73faf)

See `sum(rate(tempo_backend_scheduler_jobs_completed_total{namespace=~"$namespace"}[$__rate_interval]))`

![Image](https://github.com/user-attachments/assets/cb2d5273-8362-469b-8cfd-4dc3d5aed72d)

**Expected behavior**
tempo_backend_scheduler_jobs_active gauge goes should show the correct number of active jobs and should never be below 0

**Environment:**
 - Infrastructure: [e.g., Kubernetes, bare-metal, laptop]
 - Deployment tool: [e.g., helm, jsonnet]

**Additional Context**

we are probably not set the correct value to gauge when we reload the active jobs and boot up from a crash:

few ways to solve it:
1: gauge should go to 0 -> active jobs counter after active jobs are loaded -> inc or dec based on the jobs created, finished, failed.
2: gauge should be populate by the in memory state of the scheduler that has up to date active jobs value.
3: replace gauge with counters so we don't have this but it would be a more involved promql query to know the active jobs with the counters.

PR #5262, opened 2025-06-10
compaction: correctly load active jobs gauge on startup

electron0zero:
**What this PR does**:
Fix https://github.com/grafana/tempo/issues/5246 by loading active jobs during startup of backendscheduler

Fixes https://github.com/grafana/tempo/issues/5246

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/backendscheduler/backendscheduler.go
+++ b/modules/backendscheduler/backendscheduler.go
@@ -323,7 +323,6 @@ func (s *BackendScheduler) replayWorkOnBlocklist() {
 	// Get all the input blocks which have been successfully compacted
 	for _, j := range s.work.ListJobs() {
 		tenant = j.Tenant()
-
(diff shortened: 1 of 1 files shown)


modules/backendscheduler/backendscheduler.go
```diff
@@ -322,8 +322,12 @@ func (s *BackendScheduler) replayWorkOnBlocklist() {
 
 	// Get all the input blocks which have been successfully compacted
 	for _, j := range s.work.ListJobs() {
-		tenant = j.Tenant()
+		// count the active jobs as we are replaying the work cache on the blocklist
+		if j.GetStatus() == tempopb.JobStatus_JOB_STATUS_RUNNING {
```
zalegrala: There is a lock for each one of these Get calls.  Can we store the value so we don't call it more often than we need?

electron0zero: done

PR #5276, opened 2025-06-13
ingester: use a 64-bit hash to avoid collisions

carles-grafana:
We had this issue in the distributor, it it can happen in the ingester too with live traces.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


carles-grafana:
@mdisibio  I've added a test to check for collisions. I've also run the same test in main and the same span IDs where found in multiple traces, so I can confirm that this fixes the issue for live traces.

mdisibio (approved):

PR #5278, opened 2025-06-13
Improve traceql exmeplar quantile assignment in quantile_over_time calculation

### LINKED ISSUE #5167: quantile_over_time with multiple quantiles returns copied exemplars for all series
**Describe the bug**
In this loop:
https://github.com/ruslan-mikhailov/tempo/blob/17e20a43af0881ef0841d2e33eaeb190c422e549/pkg/traceql/engine_metrics.go#L1477-L1480 each quantile receives a copy of exemplars [here](https://github.com/grafana/tempo/blob/17e20a43af0881ef0841d2e33eaeb190c422e549/pkg/traceql/engine_metrics.go#L1488). It leads to bad UX. For example, each dot in the plot is an exemplar for each of the quantiles. 

Image

In particular, q50 has exemplars that are much higher than time series data:

Image

Because of this behaviour, it also ignores `exemplars` query parameter.

**To Reproduce**
1. Run Docker Compose from `example/docker-compose/local`
2. Wait some time (5+ minutes) until k6 pushes enough traces
3. Query `{} | quantile_over_time(span:duration, 0.5, 0.9, 0.99)`

**Expected behavior**

I would expect exemplar values from each series to be less than the quantile's value. 

**Environment:**
Can be reproduced in:
 - Infrastructure: docker compose

**Additional Context**

zalegrala:
**What this PR does**:

Here we change the exemplar selection from global to per-series, per-quantile.  This should result in an improved user experience, such that exemplars are attached to the series based on the quantile form which they were found.

Fixes #5167

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/traceql/engine_metrics.go
```diff
@@ -1513,34 +1519,137 @@ func (h *HistogramAggregator) Combine(in []*tempopb.TimeSeries) {
...
+
+	quantileValues := make([]float64, len(h.qs))
+	for i, q := range h.qs {
+		quantileValues[i] = Log2Quantile(q, buckets)
+	}
+
+	// Build results using the calculated quantile values
+	for _, in := range h.ss {
+		// Pre-sort all interval buckets once for efficiency
+		sortedIntervalBuckets := make([][]HistogramBucket, len(in.hist))
```
zalegrala: TODO: hoist the allocation?

pkg/traceql/engine_metrics.go
```diff
@@ -1485,14 +1491,14 @@ func (h *HistogramAggregator) Combine(in []*tempopb.TimeSeries) {
 			}
 		}
 
+		// Collect exemplars per series, not globally
 		for _, exemplar := range ts.Exemplars {
-			if h.exemplarBuckets.testTotal() {
-				break
+			if existing.exemplarBuckets.testTotal() {
+				continue
```
ruslan-mikhailov: Should it be `break` here? If I'm not mistaken, if testTotal is true for a given exemplarBuckets, on next call it still will be true

pkg/traceql/engine_metrics.go
```diff
@@ -1513,34 +1519,137 @@ func (h *HistogramAggregator) Combine(in []*tempopb.TimeSeries) {
...
+		for _, hist := range in.hist {
+			for _, bucket := range hist.Buckets {
+				aggregatedBuckets[bucket.Max] += bucket.Count
+			}
+		}
+	}
+
+	// Calculate quantile values from aggregated distribution
+	// Convert map to sorted slice
+	var buckets []HistogramBucket
```
ruslan-mikhailov: We can pre-allocate memory here

pkg/traceql/engine_metrics_test.go
```diff
@@ -2001,6 +2002,77 @@ func TestTiesInBottomK(t *testing.T) {
 	checkEqualForTies(t, result[`{label="c"}`].Values, []float64{10, 3, math.NaN()})
 }
 
+func TestHistogramAggregator(t *testing.T) {
+	// nolint:gosec // G115
```
ruslan-mikhailov: We excluded G115, should be safe to remove

pkg/traceql/engine_metrics_test.go
```diff
@@ -2001,6 +2002,77 @@ func TestTiesInBottomK(t *testing.T) {
...
+	const seriesCount = 6
+
+	cases := []struct {
+		name          string
+		samplesCount  int
+		exemplarCount int
+	}{
+		{"Small", 10, 5},
+		// {"Medium", 100, 20},
+		// {"Large", 1000, 100},
```
ruslan-mikhailov: Forgotten comment?

ruslan-mikhailov: Plus here as well: https://github.com/grafana/tempo/pull/5278/files#diff-d41cf3fc78887a85ec63bf1db9ee7fad9b6e4e7535fb400dac607629e9bf75fcR2050-R2052

pkg/traceql/engine_metrics_test.go
```diff
@@ -2001,6 +2002,77 @@ func TestTiesInBottomK(t *testing.T) {
...
+				require.LessOrEqual(t, len(ts.Exemplars), tc.exemplarCount*5, "Per-series exemplars should be reasonable")
+				// Note: With aggregated semantic matching, exemplars are distributed more efficiently
+				// across quantile series based on comprehensive quantile thresholds
+
+				// t.Logf("Series: %s", ts.Labels)
+				// t.Logf("Values: %v", ts.Values)
+				// t.Logf("Exemplars: %v", ts.Exemplars)
+
+				// check that the values are within the expected histogram buckets
+				for _, value := range ts.Values {
```
ruslan-mikhailov, line 2255: Need also to check that len of `ts.Values` not 0, otherwise in case of such a bug, the test will not catch it.

pkg/traceql/engine_metrics.go
zalegrala: Ah you're right, good catch.

pkg/traceql/engine_metrics_test.go
zalegrala: I left these commented because in its current state, we can actually receive more exemplars than what was requested.  We had a thread about it a few weeks ago, but basically, I think its okay to have a few extra exemplars.  There is additional complexity that arises since we don't know how early in the distribution a limit is reached, and identifying that is tricky.  If you see a shortcut to this issue, perhaps we can follow up after this PR.  The test was originally copied, but the later two cases failed for the extra exemplars.

pkg/traceql/engine_metrics.go
```diff
@@ -1470,6 +1471,11 @@ func (h *HistogramAggregator) Combine(in []*tempopb.TimeSeries) {
 			existing = histSeries{
 				labels: withoutBucket,
 				hist:   make([]Histogram, h.len),
+				exemplarBuckets: newBucketSet(
+					h.exemplarLimit,
+					h.start,
+					h.end,
```
mapno: Don't you need to align the start and end like in the other types?

pkg/traceql/engine_metrics.go
mapno: This is still `continue`

mapno (commented):
Do we know what's the performance impact of this change? It feels like a lot more work for quantile queries.

pkg/traceql/engine_metrics.go
zalegrala: Good catch, I missed this.

zalegrala:
I'll take a look at some benchmarks and review performance.

zalegrala:
Here is where we are currently.  I'll take a look where we might be able to improve this.

image


```
❯ benchstat main.txt branch.txt
goos: linux
goarch: amd64
                                      │  main.txt   │             branch.txt              │
                                      │   sec/op    │    sec/op     vs base               │
HistogramAggregator_Combine/Small-16    10.46µ ± 3%    13.04µ ± 3%  +24.68% (p=0.002 n=6)
HistogramAggregator_Combine/Medium-16   30.71µ ± 2%    42.54µ ± 3%  +38.53% (p=0.002 n=6)
HistogramAggregator_Combine/Large-16    90.69µ ± 3%   161.85µ ± 1%  +78.46% (p=0.002 n=6)
geomean                                 30.77µ         44.78µ       +45.53%

                                      │   main.txt   │              branch.txt               │
                                      │     B/op     │     B/op       vs base                │
HistogramAggregator_Combine/Small-16    41.89Ki ± 0%    47.16Ki ± 0%   +12.57% (p=0.002 n=6)
HistogramAggregator_Combine/Medium-16   55.32Ki ± 0%    84.40Ki ± 0%   +52.56% (p=0.002 n=6)
HistogramAggregator_Combine/Large-16    90.35Ki ± 0%   228.90Ki ± 0%  +153.34% (p=0.002 n=6)
geomean                                 59.38Ki         96.94Ki        +63.25%

                                      │  main.txt   │             branch.txt             │
HistogramAggregator_Combine/Small-16     109.0 ± 0%    156.0 ± 0%  +43.12% (p=0.002 n=6)
HistogramAggregator_Combine/Medium-16    669.0 ± 0%    804.0 ± 0%  +20.18% (p=0.002 n=6)
HistogramAggregator_Combine/Large-16    1.597k ± 0%   2.148k ± 0%  +34.50% (p=0.002 n=6)
geomean                                  488.3         645.9       +32.26%
```

zalegrala:
Today I learned about `❯ go tool pprof -list=".*HistogramAggregator.*Combine" branch.mem`.

With the last commit, I'm seeing this improvement over the last.
```
❯ benchstat main.txt next.txt
goos: linux
goarch: amd64
                                      │  main.txt   │              next.txt               │
                                      │   sec/op    │    sec/op     vs base               │
HistogramAggregator_Combine/Small-16    9.833µ ± 3%   13.127µ ± 3%  +33.50% (p=0.000 n=8)
HistogramAggregator_Combine/Medium-16   29.25µ ± 2%    39.96µ ± 2%  +36.63% (p=0.000 n=8)
HistogramAggregator_Combine/Large-16    87.75µ ± 1%   149.42µ ± 2%  +70.28% (p=0.000 n=8)
geomean                                 29.33µ         42.80µ       +45.90%

                                      │   main.txt   │               next.txt                │
                                      │     B/op     │     B/op       vs base                │
HistogramAggregator_Combine/Small-16    41.89Ki ± 0%    46.94Ki ± 0%   +12.05% (p=0.000 n=8)
HistogramAggregator_Combine/Medium-16   55.32Ki ± 0%    73.96Ki ± 0%   +33.70% (p=0.000 n=8)
HistogramAggregator_Combine/Large-16    90.45Ki ± 0%   183.22Ki ± 0%  +102.57% (p=0.000 n=8)
geomean                                 59.40Ki         86.00Ki        +44.78%

                                      │  main.txt   │              next.txt              │
HistogramAggregator_Combine/Small-16     109.0 ± 0%    145.0 ± 0%  +33.03% (p=0.000 n=8)
HistogramAggregator_Combine/Medium-16    669.0 ± 0%    775.0 ± 0%  +15.84% (p=0.000 n=8)
HistogramAggregator_Combine/Large-16    1.603k ± 0%   2.107k ± 0%  +31.44% (p=0.000 n=8)
geomean                                  488.9         618.6       +26.53%
```

zalegrala:
Here is a screenshot showing the results of the most recent commit.
image

Here is the benchstat results from main, including the benchmark addition here for the `Results` call.
```
❯ benchstat main.txt branch.txt
goos: linux
goarch: amd64
                                                 │   main.txt   │              branch.txt              │
                                                 │    sec/op    │    sec/op      vs base               │
HistogramAggregator_Combine/Small-16                10.02µ ± 2%    13.29µ ±  2%  +32.53% (p=0.000 n=8)
HistogramAggregator_Combine/Medium-16               30.29µ ± 2%    37.65µ ±  5%  +24.29% (p=0.000 n=8)
HistogramAggregator_Combine/Large-16                89.51µ ± 1%   122.95µ ±  1%  +37.36% (p=0.000 n=8)
HistogramAggregator_Results/Small_3Quantiles-16     84.70µ ± 1%    27.60µ ±  7%  -67.41% (p=0.000 n=8)
HistogramAggregator_Results/Medium_3Quantiles-16   298.48µ ± 1%    52.64µ ±  7%  -82.36% (p=0.000 n=8)
HistogramAggregator_Results/Large_3Quantiles-16    1279.1µ ± 2%    152.4µ ±  6%  -88.08% (p=0.000 n=8)
HistogramAggregator_Results/Small_5Quantiles-16    143.56µ ± 2%    42.77µ ± 10%  -70.21% (p=0.000 n=8)
HistogramAggregator_Results/Medium_5Quantiles-16   534.51µ ± 1%    80.04µ ±  6%  -85.03% (p=0.000 n=8)
HistogramAggregator_Results/Large_5Quantiles-16    2160.8µ ± 1%    229.0µ ±  2%  -89.40% (p=0.000 n=8)
HistogramAggregator_Results/High_Exemplars-16      301.62µ ± 2%    77.52µ ±  2%  -74.30% (p=0.000 n=8)
geomean                                             183.8µ         61.91µ        -66.31%

                                                 │   main.txt   │              branch.txt               │
                                                 │     B/op     │     B/op       vs base                │
HistogramAggregator_Combine/Small-16               41.89Ki ± 0%    47.04Ki ± 0%   +12.29% (p=0.000 n=8)
HistogramAggregator_Combine/Medium-16              55.32Ki ± 0%    74.01Ki ± 0%   +33.79% (p=0.000 n=8)
HistogramAggregator_Combine/Large-16               90.35Ki ± 0%   183.04Ki ± 0%  +102.58% (p=0.000 n=8)
HistogramAggregator_Results/Small_3Quantiles-16    54.91Ki ± 0%    53.50Ki ± 0%    -2.56% (p=0.000 n=8)
HistogramAggregator_Results/Medium_3Quantiles-16   155.7Ki ± 0%    113.8Ki ± 0%   -26.94% (p=0.000 n=8)
HistogramAggregator_Results/Large_3Quantiles-16    509.7Ki ± 0%    407.6Ki ± 0%   -20.03% (p=0.000 n=8)
HistogramAggregator_Results/Small_5Quantiles-16    92.46Ki ± 0%    88.51Ki ± 0%    -4.26% (p=0.000 n=8)
HistogramAggregator_Results/Medium_5Quantiles-16   255.5Ki ± 0%    167.2Ki ± 0%   -34.56% (p=0.000 n=8)
HistogramAggregator_Results/Large_5Quantiles-16    841.4Ki ± 0%    514.4Ki ± 0%   -38.87% (p=0.000 n=8)
HistogramAggregator_Results/High_Exemplars-16      155.7Ki ± 0%    203.9Ki ± 0%   +30.90% (p=0.000 n=8)
geomean                                            139.7Ki         137.7Ki         -1.40%

                                                 │   main.txt   │             branch.txt             │
                                                 │  allocs/op   │  allocs/op   vs base               │
HistogramAggregator_Combine/Small-16                 109.0 ± 0%    149.0 ± 0%  +36.70% (p=0.000 n=8)
HistogramAggregator_Combine/Medium-16                669.0 ± 0%    779.0 ± 0%  +16.44% (p=0.000 n=8)
HistogramAggregator_Combine/Large-16                1.597k ± 0%   2.111k ± 0%  +32.19% (p=0.000 n=8)
HistogramAggregator_Results/Small_3Quantiles-16      346.0 ± 0%    216.0 ± 0%  -37.57% (p=0.000 n=8)
HistogramAggregator_Results/Medium_3Quantiles-16    3274.0 ± 0%    386.0 ± 0%  -88.21% (p=0.000 n=8)
HistogramAggregator_Results/Large_3Quantiles-16    15004.0 ± 0%    808.0 ± 0%  -94.61% (p=0.000 n=8)
HistogramAggregator_Results/Small_5Quantiles-16      574.0 ± 0%    336.0 ± 0%  -41.46% (p=0.000 n=8)
HistogramAggregator_Results/Medium_5Quantiles-16    5454.0 ± 0%    586.0 ± 0%  -89.26% (p=0.000 n=8)
HistogramAggregator_Results/Large_5Quantiles-16    25.004k ± 0%   1.209k ± 0%  -95.16% (p=0.000 n=8)
HistogramAggregator_Results/High_Exemplars-16       3274.0 ± 0%    407.0 ± 0%  -87.57% (p=0.000 n=8)
geomean                                             1.864k         525.3       -71.82%

```

zalegrala:
I also pushed a branch `quantilOverTimeExemplarSpreadLog2QuantileWithBuckets` which included the suggested change to use the `Log2Quantile` and wrap a method `Log2QuantileWithBucket` to return the bucket and the index, but this ended up with strange results.  It looked like two exemplars could end up in the same bucket, but best fit different exemplars.  Claude helped show this in a test to confirm, which is committed on that branch.  In the final image, most exemplars were assigned to the lower quantile.  I attempted to invert this but then it didn't seem to help the actual spread of the exemplar assignment.

With this in mind, I think the approach here is reasonable enough for another review.  In the screenshot above, you can see exemplars closely relate to their series.

pkg/traceql/engine_metrics.go
```diff
@@ -1409,50 +1409,71 @@ func (h *Histogram) Record(bucket float64, count int) {
...
-			alignEnd(req.Start, req.End, req.Step),
-		),
+		ss:            make(map[string]histSeries),
+		qs:            qs,
+		len:           IntervalCount(req.Start, req.End, req.Step),
+		start:         req.Start,
+		end:           req.End,
+		step:          req.Step,
+		exemplarLimit: exemplarLimit,
+		labelBuffer:   make(Labels, 0, 8), // Pre-allocate with reasonable capacity
```
ruslan-mikhailov, line 1544: How often `Combine` is called through `HistogramAggregator` lifetime? If once, I think, we don't need to allocate it as a part of the structure and make it local in `Combine` function scope

pkg/traceql/engine_metrics.go
```diff
@@ -1513,34 +1564,129 @@ func (h *HistogramAggregator) Combine(in []*tempopb.TimeSeries) {
...
+	for _, series := range h.ss {
+		if len(series.hist) > 0 {
+			// Clear the slice and return it to the pool
+			histSlice := series.hist
+			for i := range histSlice {
+				histSlice[i] = Histogram{}
+			}
+			histogramSlicePool.Put(histSlice[:0]) //nolint:all //SA6002
+		}
+	}
```
ruslan-mikhailov: Hmm, it looks dangerous: `Results` cannot be called twice as it not only returns results but also changes the structure.

ruslan-mikhailov: How big is the improvement from the pool?

pkg/traceql/engine_metrics.go
```diff
@@ -1769,3 +1921,26 @@ func initSeriesInResult(result SeriesSet, key string, input SeriesSet, valueLeng
 		result[key].Values[j] = math.NaN()
 	}
 }
+
+// fastIntervalOfMs is an optimized version of IntervalOfMs that uses pre-computed values
+// to avoid expensive calculations in the hot path
+func (h *HistogramAggregator) fastIntervalOfMs(tsmills int64) int {
```
ruslan-mikhailov: jfyi I have PoC that aims to solve the same problem: https://github.com/ruslan-mikhailov/tempo/pull/1/files#diff-1b84537244dd74c3ba117a9181d6b27c2ba7d632b26c355ad5e5b428b6d00d19R1395

Feel free to leave a comment

pkg/traceql/engine_metrics.go
zalegrala: My understanding is that its called often, since once we discover results, we call combine to merge them.  The labels showed up as the largest source of memory allocations in our benchmarks, and pre-allocating them meant a significant reduction.

pkg/traceql/engine_metrics.go
zalegrala: Hmm, this is a good question, help me understand.  Is it necessary that we call Results twice?  My understanding is that Combine above is called many times to merge the data, but then Results is called once.

The improvement from the polling was reasonable as tested from the benchmarks.  I coped this pattern (and the nolint) from other parts of the code where we do memory polling.  Since the pool itself is a package level variable, I expect the outcome to improve even though a single HistogramAggregator only gets called once, we can save the slice allocations for later use.

In the last benchmarks, you can see the results from main, and including the new benchmark which is added with this PR.

pkg/traceql/engine_metrics.go
zalegrala: I'm happy to replace this with something centralized.  I noticed this looking at the profile that we spent a decent amount of CPU time here, and this function and its short-circuits helped alleviate that a bit.

zalegrala:
I've moved the pool to the HistograramAggregator, but I'm working up a test to cover the memory pooling and ensure we're not leaking.

zalegrala:
I've dropped the pooling change from this PR.  Perhaps we can follow up.

pkg/traceql/engine_metrics.go
zalegrala: I've dropped the pooling change from this PR to avoid the risk of data contamination.  Perhaps we can follow up specifically with a test.

pkg/traceql/engine_metrics.go
```diff
@@ -1513,34 +1550,117 @@ func (h *HistogramAggregator) Combine(in []*tempopb.TimeSeries) {
...
+
+// determineTargetQuantile determines which quantile an exemplar should be assigned to
+// by comparing its value against the calculated quantile thresholds.
+func (h *HistogramAggregator) determineTargetQuantile(exemplarValue float64, quantileValues []float64) int {
+	// Assign to the smallest quantile that can contain this exemplar
+	// e.g., if exemplar is 1.5s and p50=0.5s, p90=2.0s, p99=5.0s
+	// then 1.5s goes to p90 (since 1.5s > p50 but <= p90)
+
+	for i := 0; i < len(quantileValues); i++ {
+		if exemplarValue <= quantileValues[i] {
```
mdisibio: I wonder if we should adjust this logic to find the nearest quantile, instead of finding the upper bound?    In this example, there are exemplars in blue labelled for p99 that are closer to the yellow p95 line.  And likewise for the yellow exemplars.  Some are circled in the screenshot for clarity, but it's common across the results.  Realize this might not be logical, but it seemed confusing at first.   


image


Additionally, we might be doing something different than prometheus, and this is why the exemplars aren't tightly related to the line.   We are choosing random spans as exemplars, which means by definition they cover the whole vertical range. Whereas prometheus is only selects exemplars from the `le` buckets that matched the final quantile.  Maybe we should also reject exemplars that didn't source exactly from any quantile bucket? 

In this screenshot, the exemplar was attached to the p50, which is ~40ms.  But the exemplar value is 97 microseconds. These are different buckets internally.

image

knylander-grafana:
Does your PR have user impact? Reading through, it seems like ti might, but wanted to verify.

zalegrala:
@knylander-grafana :wave: It will impact users, but do my knowledge we don't say anything in the docs specific to how exemplars are assigned to the particular quantile.  Let me know if I'm wrong about that.  Otherwise the change should be only positive.

pkg/traceql/engine_metrics.go
zalegrala: Thanks for the review, I've pushed a commit for your feedback.

We're now checking the bucket index of the exemplar values and excluding exemplars which are outside the n+-1 range, which should handle the outliers you point out.

Additionally, the logic is updated to assign exemplars to the closest quantile, which requires us to calculate the distance to each.  Visually there are still some oddball, but I think the logic is correct for the timeslice.

image

Let me know what you think.

knylander-grafana:
> @knylander-grafana 👋 It will impact users, but do my knowledge we don't say anything in the docs specific to how exemplars are assigned to the particular quantile. Let me know if I'm wrong about that. Otherwise the change should be only positive.

Interesting. if we don't have that in the docs, do you think we should? How would you use this capability?

zalegrala:
Its somewhat automatic if enable exemplars on the query.  Check the linked issue https://github.com/grafana/tempo/issues/5167 to see how the exemplars look compared to what we're changing here.

zalegrala:
Thanks for the review.  I'll resolve the conflict, test once more and then merge.

PR #5291, opened 2025-06-16
fix not a git repository by checked out the repository first

javiermolinar:


.github/workflows/dependabot-auto-approve.yml
```diff
@@ -11,6 +11,10 @@ jobs:
       contents: write # Needed to merge PRs via gh CLI
     if: github.event.pull_request.user.login == 'dependabot[bot]' && github.repository == 'grafana/tempo'
     steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
```
joe-elliott, line 15: does this need to be pinned to a commit hash to make zizmor happy?

javiermolinar: good call

javiermolinar: ah, I think it's not needed it since the v4 is already the version

PR #5293, opened 2025-06-16
Stop method

stephanos:
**What this PR does**:

Adds a `Stop` method to `app.App`.

see https://github.com/grafana/tempo/discussions/5225

tl;dr it's helpful when embedding Tempo inside a Go application to be able to control it's full lifecycle, not just the start.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo/app/app.go
+++ b/cmd/tempo/app/app.go
@@ -218,7 +218,7 @@ func (t *App) Run() error {
 		// let's find out which module failed
 		for m, s := range serviceMap {
 			if s == service {
-				err = service.FailureCase()
+				err := service.FailureCase()
 				if errors.Is(err, modules.ErrStopProcess) {
 					level.Info(log.Logger).Log("msg", "received stop signal via return error", "module", m, "err", err)
 				} else if errors.Is(err, context.Canceled) {
--- a/cmd/tempo/app/app.go
+++ b/cmd/tempo/app/app.go
@@ -84,6 +84,7 @@ type App struct {
 	MemberlistKV         *memberlist.KVInitService
 	backendScheduler     *backendscheduler.BackendScheduler
 	backendWorker        *backendworker.BackendWorker
+	signalsHandler       *signals.Handler
 
 	HTTPAuthMiddleware       middleware.Interface
 	TracesConsumerMiddleware receiver.Middleware
@@ -167,7 +168,7 @@ func (t *App) setupAuthMiddleware() {
 	}
 }
 
-// Run starts, and blocks until a signal is received.
+// Run starts, and blocks until a signal is received or Stop is called.
 func (t *App) Run() error {
 	if !t.ModuleManager.IsUserVisibleModule(t.cfg.Target) {
 		level.Warn(log.Logger).Log("msg", "selected target is an internal module, is this intended?", "target", t.cfg.Target)
@@ -234,9 +235,9 @@ func (t *App) Run() error {
 	sm.AddListener(services.NewManagerListener(healthy, stopped, serviceFailed))
 
 	// Setup signal handler. If signal arrives, we stop the manager, which stops all the services.
-	handler := signals.NewHandler(t.Server.Log())
+	t.signalsHandler = signals.NewHandler(t.Server.Log())
 	go func() {
-		handler.Loop()
+		t.signalsHandler.Loop()
 
 		shutdownRequested.Store(true)
 		t.Server.SetKeepAlivesEnabled(false)
@@ -256,6 +257,14 @@ func (t *App) Run() error {
 	}
 
 	return sm.AwaitStopped(context.Background())
+}
+
+// Stop the app. It panics if the app is not running.
+func (t *App) Stop() {
+	if t.signalsHandler == nil {
+		panic("app is not running")
+	}
+	t.signalsHandler.Stop()
 }
 
 func (t *App) writeStatusVersion(w io.Writer) error {
--- a/cmd/tempo/app/app_test.go
+++ b/cmd/tempo/app/app_test.go
@@ -1,12 +1,14 @@
 package app
 
 import (
+	"fmt"
 	"net/http"
 	"os"
 	"path/filepath"
 	"testing"
 	"time"
 
+	"github.com/grafana/tempo/pkg/util"
 	"github.com/grafana/tempo/tempodb/backend"
 	"github.com/stretchr/testify/require"
 )
@@ -21,6 +23,8 @@ func TestApp_RunStop(t *testing.T) {
 	}()
 
 	config := NewDefaultConfig()
+	config.Server.HTTPListenPort = util.MustGetFreePort()
+	config.Server.GRPCListenPort = util.MustGetFreePort() // not used in the test; set to ensure conflict-free start
 	config.StorageConfig.Trace.Backend = backend.Local
 	config.StorageConfig.Trace.Local.Path = filepath.Join(tempDir, "tempo")
 	config.StorageConfig.Trace.WAL.Filepath = filepath.Join(tempDir, "wal")
@@ -35,7 +39,7 @@ func TestApp_RunStop(t *testing.T) {
 	}()
 
 	// check health endpoint is reachable
-	healthCheckURL := "http://localhost:3200/ready"
+	healthCheckURL := fmt.Sprintf("http://localhost:%d/ready", config.Server.HTTPListenPort)
 	require.Eventually(t, func() bool {
 		t.Log("Checking Tempo is up...")
 		resp, httpErr := http.Get(healthCheckURL)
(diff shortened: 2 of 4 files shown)


cmd/tempo/app/app.go
```diff
@@ -258,6 +259,13 @@ func (t *App) Run() error {
 	return sm.AwaitStopped(context.Background())
 }
 
+func (t *App) Stop() {
+	if t.signalsHandler == nil {
+		panic("app is not running")
```
stephanos, line 264: Happy to change this to return an error instead if that's more fitting here.

stephanos:
Note that when I tested this locally, I had to add `config.UsageReport.Enabled = false` for the Stop to take effect immediately. Otherwise it would wait 30+ seconds and eventually continue with:
```
level=warn ts=2025-06-16T15:27:09.78514Z caller=module_service.go:118 msg="module failed with error" module=usage-report err="context canceled"
```

stoewer (commented):
Thanks for opening this PR. The code looks like it should work. It would be great to have a test for the new method. Do you think you could add one?

stephanos:
@stoewer I was looking at that, but I didn't know where to start? I might have missed it, but I couldn't find a single test for that package to build upon.

stoewer:
We don't have a test that calls `App.Run()` directly. Therefore this needs to be a new test. Out of the top of my head this could be a test in `app_test.go` (does not exist yet) that may have the following structure:

```go
func TestApp_RunStop(t *testing.T) {
	var config Config
	// assign config e.g. via NewDefaultConfig()

	app, err := New(config)
	go func() {
		app.Run()
	}()
	// check health endpoint is reachable

	app.Stop()
	// check health endpoint is not reachable
}
```

cmd/tempo/app/app_test.go
```diff
@@ -0,0 +1,54 @@
...
+	app, err := New(*config)
+	require.NoError(t, err)
+
+	// start Tempo
+	go func() {
+		require.NoError(t, app.Run())
+	}()
+
+	// check health endpoint is reachable
+	healthCheckURL := "http://localhost:3200/ready"
```
stephanos: I didn't see any utilities to obtain a random free port in the repo, so I'm using the default. Since I can't run the GHA tests, though, I don't know if this might clash with an existing test that might use the port.

stoewer: Thanks for adding the test. It looks good and works fine locally. I’d be happy to accept it as is.

That said, if you’d like to improve it further by addressing the fixed port usage, you could obtain a free port like this:

```go
func getFreePort() (int, error) {
	addr, err := net.ResolveTCPAddr("tcp", "localhost:0")

	if err != nil {
		return 0, err
	}

	l, err := net.ListenTCP("tcp", addr)
	if err != nil {
		return 0, err
	}

	defer l.Close()
	return l.Addr().(*net.TCPAddr).Port, nil
}
```

Once you have the free port, you can set it in the [`config` struct](https://grafana.com/docs/tempo/latest/configuration/#server) and use it in the `healthCheckURL`.

stephanos: 👍 We've been using [this one for a while now in Temporal](https://github.com/temporalio/temporal/blob/main/common/testing/freeport/freeport.go), I'll use that unless you object.

stephanos: Okay, I pushed my changes. But the problem is the test is failing now. I can see that if I run the healthcheck on 3200 it succeeds. So that must mean the port isn't taking; even though there's log line print the port. Odd! Do you have some guidance?

cmd/tempo/app/app.go
```diff
@@ -217,7 +218,7 @@ func (t *App) Run() error {
 		// let's find out which module failed
 		for m, s := range serviceMap {
 			if s == service {
-				err = service.FailureCase()
+				err := service.FailureCase()
```
stephanos, line 221: Drive-by change; I ran my test with `-race` and it showed me a data race here with the other `err` outside of this goroutine.

cmd/tempo/app/app_test.go
stoewer: > We've been using [this one for a while now in Temporal](https://github.com/temporalio/temporal/blob/main/common/testing/freeport/freeport.go), I'll use that unless you object.

Yes this is OK. AFAIK the MIT license permits copying of small, non-substantial parts of the code without using the same license. 
As a courtesy we typically add a comment mentioning the code was copied, including a link to the original source. Could you please add such a comment to `freeport.go`? Thanks

stoewer: > I can see that if I run the healthcheck on 3200 it succeeds. So that must mean the port isn't taking; even though there's log line print the port. Odd!

Yes, that does seem odd. I’ll take a look and get back to you shortly.

stoewer: @stephanos I tried to reproduce the failing test but in my environment (Fedoara 41) everything works fine. 

<details>
<summary>Test cmd and logs</summary>

```
go test -mod=vendor -v -run='TestApp_RunStop' -count=1 ./cmd/tempo/app
```

```
=== RUN   TestApp_RunStop
level=info msg="server listening on addresses" http=[::]:39795 grpc=[::]:43919
ts=2025-07-01T01:39:16Z level=info msg="OTel Shim Logger Initialized" component=tempo
ts=2025-07-01T01:39:16Z level=info msg="Starting HTTP server for Jaeger Thrift" component=tempo endpoint=localhost:14268
ts=2025-07-01T01:39:16Z level=info msg="Starting gRPC server for Jaeger Protobuf" component=tempo endpoint=localhost:14250
ts=2025-07-01T01:39:16Z level=info msg="Starting GRPC server" component=tempo endpoint=localhost:4317
    app_test.go:44: Checking Tempo is up...
    app_test.go:44: Checking Tempo is up...
    app_test.go:44: Checking Tempo is up...
    app_test.go:44: Checking Tempo is up...
    app_test.go:44: Checking Tempo is up...
    app_test.go:44: Checking Tempo is up...
    app_test.go:44: Checking Tempo is up...
    app_test.go:44: Checking Tempo is up...
    app_test.go:44: Checking Tempo is up...
    app_test.go:44: Checking Tempo is up...
    app_test.go:44: Checking Tempo is up...
    app_test.go:44: Checking Tempo is up...
    app_test.go:44: Checking Tempo is up...
    app_test.go:44: Checking Tempo is up...
    app_test.go:44: Checking Tempo is up...
    app_test.go:44: Checking Tempo is up...
    app_test.go:44: Checking Tempo is up...
level=info msg="=== Handler.Stop()'d ==="
level=warn method=/frontend.Frontend/Process duration=16.983895258s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.984015555s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.983845896s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.983706553s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.984040682s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.983877315s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.984082972s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.984120583s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.98377426s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.984103081s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.984014333s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.984113921s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.984015545s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.983989235s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.984135361s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.983900028s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.984209511s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.984090216s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.984157543s msg=gRPC err="queue is stopped"
level=warn method=/frontend.Frontend/Process duration=16.984226263s msg=gRPC err="queue is stopped"
    app_test.go:54: Checking Tempo is down...
--- PASS: TestApp_RunStop (18.01s)
PASS
ok  	github.com/grafana/tempo/cmd/tempo/app	18.037s
```
</details>

The test also completes successfully in our [CI env](https://github.com/grafana/tempo/actions/runs/15880756117/job/44887385314?pr=5293). What environment and commands are you using to run the test?

stephanos: You're right! I must have mixed something up, it's passing for me locally now, too. Thank you :)

stoewer (approved):
LGTM!
Thanks for submitting the PR and for going the extra mile with the tests - really appreciate the effort!

PR #5324, opened 2025-06-25
Backend scheduler tests

zalegrala:
**What this PR does**:

Here we include some integration tests for the `backend-scheduler` component, and put them into CI.  These tests check the metric state of the scheduler, based on the work in the system at various points of operation against a set of expectations.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] CI updated
- [ ] Documentation added

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,7 @@ jobs:
             test-e2e,
             test-e2e-api,
             test-integration-poller,
-            test-integration-backendscheduler
+            test-integration-backendscheduler,
             test-e2e-deployments,
             test-e2e-ingest,
           ]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,6 +99,7 @@ jobs:
             test-e2e,
             test-e2e-api,
             test-integration-poller,
+            test-integration-backendscheduler
             test-e2e-deployments,
             test-e2e-ingest,
           ]
--- a/integration/backendscheduler/backendscheduler_test.go
+++ b/integration/backendscheduler/backendscheduler_test.go
@@ -255,7 +255,7 @@ func setupBackendWithEndpoint(t testing.TB, cfg *tempodb.Config, endpoint string
 		DedicatedColumns:     backend.DedicatedColumns{{Scope: "span", Name: "key", Type: "string"}},
 	}
 	cfg.WAL = &wal.Config{
-		Filepath: "/var/tempo/wal",
+		Filepath: t.TempDir(),
 	}
 	cfg.S3.Endpoint = endpoint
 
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -99,6 +99,7 @@ jobs:
             test-e2e,
             test-e2e-api,
             test-integration-poller,
+            test-integration-backendscheduler
```
electron0zero: fyi: we need to add CI check into the Rulesets to enforce this CI check.

integration/backendscheduler/backendscheduler_test.go
```diff
@@ -64,26 +64,183 @@ func TestBackendScheduler(t *testing.T) {
...
 
 	// // Setup tempodb with local backend
 	// tempodbWriter := setupBackend(t, tmpDir)
 	tempodbWriter := setupBackendWithEndpoint(t, &cfg.StorageConfig.Trace, e)
 
-	// // Push some data to a few tenants
-	for i := 0; i < tenantCount; i++ {
-		testTenant := tenant + strconv.Itoa(i)
-		populateBackend(ctx, t, tempodbWriter, testTenant)
+	time.Sleep(5 * time.Second)
```
electron0zero: any reasons to add sleep over `StartAndWaitReady` for all services we want to wait on?


it's okay but I would like us to avoid using sleep when we can avoid.

integration/backendscheduler/backendscheduler_test.go
```diff
@@ -64,26 +64,183 @@ func TestBackendScheduler(t *testing.T) {
...
+	}{
+		{
+			name:                      "a bunch of tenants with 1 block each is 0 outstanding",
+			tenantCount:               11,
+			blockCount:                1,
+			expectedBlocks:            1,
+			expectedOutstandingBlocks: 0,
+		},
+		{
+			name:                      "a bunch of tenants with 12 blocks each 12 3 outstanding",
```
electron0zero: looks like a typo here? 
```suggestion
			name:                      "a bunch of tenants with 12 blocks each 12 outstanding",
```

electron0zero (approved):
few nits, rest of the test lgtm.

I like that we are testing replay as well, nice coverage 💯

.github/workflows/ci.yml
zalegrala: This is post-merge in the config of the repo?

integration/backendscheduler/backendscheduler_test.go
zalegrala: Testing without it still seems to pass.  We're using `WaitMissingMetrics` below, so I think its okay to drop.

.github/workflows/ci.yml
electron0zero: yup, in Settings -> Rule -> Ruleset or just go to https://github.com/grafana/tempo/settings/rules

PR #5325, opened 2025-06-25
Compaction provider improvements

zalegrala:
**What this PR does**:

Here we make a few small improvements to the compaction provider.
* skip tenants which have too low a priority to create a compaction job
* relocate a backoff to make testing easier and simplify
* allow jobs of any status which have input blocks to reduce what is visible to the block selector

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/backendscheduler/provider/compaction.go
+++ b/modules/backendscheduler/provider/compaction.go
@@ -317,8 +317,11 @@ func (p *CompactionProvider) prioritizeTenants(ctx context.Context) {
 
 	for _, tenant := range tenants {
 		priority = ts.PriorityForTenant(tenant.ID)
-		item = tenantselector.NewItem(tenant.ID, priority)
-		heap.Push(p.curPriority, item)
+
+		if priority >= blockselector.DefaultMinInputBlocks {
+			item = tenantselector.NewItem(tenant.ID, priority)
+			heap.Push(p.curPriority, item)
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


electron0zero (approved):

modules/backendscheduler/provider/compaction.go
```diff
@@ -344,29 +347,19 @@ func (p *CompactionProvider) newBlockSelector(tenantID string) (blockselector.Co
...
-		}
-
-		if jobType = job.GetType(); jobType == tempopb.JobType_JOB_TYPE_UNSPECIFIED {
+		if job.Tenant() != tenantID {
 			continue
 		}
 
-		jobStatus = job.GetStatus()
-		if jobStatus != tempopb.JobStatus_JOB_STATUS_RUNNING && jobStatus != tempopb.JobStatus_JOB_STATUS_UNSPECIFIED {
+		if job.GetType() == tempopb.JobType_JOB_TYPE_UNSPECIFIED {
```
mapno, line 362: When can a job have unspecified type?

mapno (approved):

modules/backendscheduler/provider/compaction.go
zalegrala: It shouldn't ever happen, but its in the proto, so this is a protection.

zalegrala: I think technically we could drop this here, since the check is performed in the GetCompactionInput method.

PR #5330, opened 2025-06-27
New bufferer component

mapno:
**What this PR does**:

The bufferer is a Kafka-based ingestion service that consumes trace data from Kafka topics instead of receiving it via gRPC like the traditional ingester. It reads serialized traces from Kafka, processes them through the same data pipeline as the ingester (live traces → head blocks → WAL → complete blocks).

```
                      ┌─────────────────────────────┐
                      │        Bufferer             │
                      │ • Manages instances map     │
                      │ • Shared WAL coordination   │
                      │ • Flush queue management    │
                      └─────────────┬───────────────┘
                                    │
                      ┌─────────────┴───────────────┐
          ┌─────────────────────┐       ┌─────────────────────┐
          │  PartitionReader    │       │   instances map     │
          │                     │       │   [tenantID]        │
          │ • Kafka consumer    │       │                     │
          │ • Watermark commits │       │ ┌─────────────────┐ │
          │ • Routes to consume │       │ |   instance A    │ │
          │   function          │       │ │   (tenant A)    │ │
          └─────────────────────┘       │ │                 │ │
                                        │ │ • Live traces   │ │
                                        │ │ • Head block    │ │
                                        │ │ • WAL blocks    │ │
                                        │ │ • Complete      │ │
                                        │ │   blocks        │ │
                                        │ └─────────────────┘ │
                                        │                     │
                                        │ ┌─────────────────┐ │
                                        │ │   instance B    │ │
                                        │ │   (tenant B)    │ │
                                        │ │                 │ │
                                        │ │ • Live traces   │ │
                                        │ │ • Head block    │ │
                                        │ │ • WAL blocks    │ │
                                        │ │ • Complete      │ │
                                        │ │   blocks        │ │
                                        │ └─────────────────┘ │
                                        └─────────────────────┘
                                                   │
                                                   ▼
                                        ┌─────────────────────┐
                                        │     Shared WAL      │
                                        │                     │
                                        │ • Single WAL for    │
                                        │   all instances     │
                                        │ • Coordinated       │
                                        │   block cutting     │
                                        └─────────────────────┘
```

**Checklist**
- [x] Tests updated
- [ ] Documentation added


mattdurham:
In general does it make sense to move responsibility of timing and actions from the Bufferer to the individual instances? This would make the instances more self contained and potentially avoid a good chunk of the locking. Would be harder to do global *cuts* but we could do tenant specific then? Or trigger via a channel with at timer in the Bufferer.

mattdurham:
From chatting would calling `Bufferer` `ReadCache` make more sense?

mattdurham:
Can we document the failure modes and what happens? Do we need 100% data fidelity or can 99.9%, since the data will be correctly handled by long term storage regardless.

mattdurham:
Assuming failures happen is it better to return incorrect data or no data?

mapno:
Development has been moved to a dev branch. See https://github.com/grafana/tempo/pull/5430

PR #5343, opened 2025-07-01
[Bugfix] query span id no leading zeros

ruslan-mikhailov:
**What this PR does**:
Fixes search by span id without leading zeros

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/traceql/ast_conditions.go
```diff
@@ -61,10 +79,16 @@ func (o *BinaryOperation) extractConditions(request *FetchSpansRequest) {
 					Operands:  nil,
 				})
 			} else {
+				// Normalize span ID operands
+				operand := o.RHS.(Static)
+				if o.LHS.(Attribute).Intrinsic == IntrinsicSpanID {
+					operand = normalizeSpanIDOperand(operand)
+					o.RHS = operand // Update the operation with the normalized operand
```
ruslan-mikhailov: I'm not sure whether it's the right way to do to change LHS/RHS inside `extractConditions`

pkg/parquetquerygen/predicates.go
```diff
@@ -261,6 +261,24 @@ func (p {{ $structName }}) KeepValue(v pq.Value) bool {
...
+		{
+			Name:           "ByteIgnoreLeadingZeroes",
+			Type:           "[]byte",
+			ParquetFunc:    "ByteArray()",
+			FormatModifier: "%s",
+			Ops: []op{
+				{
+					Op:          "Equal",
+					CompareCond: `bytes.Equal(bytes.TrimLeft(vv, "\x00"), p.value)`,
+					RangeCond:   "",
```
ruslan-mikhailov, line 273: I don't really like that we have to skip `RangeCond`. Maybe it would be better to add/remove leading zeroes on write as we do for trace id?

PR #5346, opened 2025-07-01
BugFix: Correctly apply trace idle period in ingesters and add a max live trace period

joe-elliott:
**What this PR does**:
Trace idle period has never worked in Tempo ingesters. Every "flush sweep" has always flushed all traces to the wal. Traces were primarily made contiguous during wal flush when all wal segments are rewritten into one large block.

This PR:

1. Fixes trace idle period in tempo ingesters
1. Sets trace idle default to 5s in ingesters and generators
1. Adds the concept of trace live period to both ingesters and metrics generators. This is the max amount of time a trace can be in the live traces map before being flushed. This is to combat write patterns where a very large trace is drip fed a span at a time into Tempo which results in very large traces being held in memory. This is defaulted to 30s.

Naturally such a change as actually honoring the trace idle period upset some timer based e2e tests. There are a handful of patches and improvements there.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/integration/e2e/api/api_test.go
+++ b/integration/e2e/api/api_test.go
@@ -264,13 +264,6 @@ func TestSearchTagsV2(t *testing.T) {
 	// wait for 2 objects to be written to the backend
 	require.NoError(t, tempo.WaitSumMetricsWithOptions(e2e.Equals(2), []string{"tempodb_backend_objects_total"}, e2e.WaitMissingMetrics))
 
-	// Assert no more on the ingester
-	for _, tc := range testCases {
-		t.Run(tc.name, func(t *testing.T) {
-			callSearchTagsV2AndAssert(t, tempo, tc.scope, tc.query, searchTagsV2Response{}, 0, 0)
-		})
-	}
-
 	// Assert tags on storage backend
 	now := time.Now()
 	start := now.Add(-2 * time.Hour)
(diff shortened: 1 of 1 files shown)


integration/e2e/api/api_test.go
```diff
@@ -34,8 +34,8 @@ const (
 	queryableCheckEvery = 100 * time.Millisecond // check every 100ms for traces to be queryable
 
 	// Wait to block flushed to backend, 5 seconds is the complete_block_timeout configuration on all in one, we add
-	// 100ms for security.
-	blockFlushTimeout = 5*time.Second + 100*time.Millisecond
+	// 1s for security.
```
mapno, line 37:
```suggestion
	// 6s for security.
```

mapno (approved):

integration/e2e/api/api_test.go
joe-elliott: its 5s + 1s for security, but it doesn't even really matter anymore. i removed the check that nothing is found in the ingesters b/c there was no good way to detect they were no longer on the ingesters (expect for querying til they disappeared). it was flaky and it didn't seem important. the test to make sure that they are queryable from the backend after flush remains and that feels more important.

PR #5366, opened 2025-07-07
[Breaking changes] Calculate Metrics based on data in the past

ruslan-mikhailov:
**Goal:**
1. Improve UX
2. Align behaviour with Prometheus
3. Solve problem with the last bucket that is always empty (e.g. https://github.com/grafana/tempo/blob/main/tempodb/tempodb_metrics_test.go#L62)

**What this PR does**:
1. It changes the way Tempo calculates buckets. Before the change, it counted data to the right, now to the left
4. Change intervals from `[;)` to `(;]`
5. Solves problem with empty last (now first) bucket. Now frontend is responsible for adding additional first bucket

image

image

How frontend divides incoming request into parts:
image

**Before and after**
image
image

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/metrics_query_range_handler.go
+++ b/modules/frontend/metrics_query_range_handler.go
@@ -29,6 +29,7 @@ func newQueryRangeStreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTripp
 		ctx := srv.Context()
 
 		headers := headersFromGrpcContext(ctx)
+		traceql.AlignRequest(req)
 
 		// default step if not set
 		if req.Step == 0 {
@@ -88,11 +89,9 @@ func newMetricsQueryRangeHTTPHandler(cfg Config, next pipeline.AsyncRoundTripper
 				Body:       io.NopCloser(strings.NewReader(err.Error())),
 			}, nil
 		}
-		if queryRangeReq.Start > queryRangeReq.Step {
-			queryRangeReq.Start -= queryRangeReq.Step
-		}
-
 		logQueryRangeRequest(logger, tenant, queryRangeReq)
+
+		traceql.AlignRequest(queryRangeReq)
 
 		// build and use roundtripper
 		combiner, err := combiner.NewTypedQueryRange(queryRangeReq, cfg.Metrics.Sharder.MaxResponseSeries)
--- a/modules/frontend/metrics_query_range_sharder.go
+++ b/modules/frontend/metrics_query_range_sharder.go
@@ -74,9 +74,6 @@ func (s queryRangeSharder) RoundTrip(pipelineRequest pipeline.Request) (pipeline
 	if err != nil {
 		return pipeline.NewBadRequest(err), nil
 	}
-	if !s.instantMode && req.Start > req.Step {
-		req.Start -= req.Step // force to have additional initial bucket
-	}
 
 	expr, _, _, _, _, err := traceql.Compile(req.Query)
 	if err != nil {
--- a/pkg/traceql/engine_metrics.go
+++ b/pkg/traceql/engine_metrics.go
@@ -203,6 +203,10 @@ func AlignRequest(req *tempopb.QueryRangeRequest) {
 	// It doesn't really matter but the request fields are expected to be in nanoseconds.
 	req.Start = alignStart(req.Start, req.End, req.Step)
 	req.End = alignEnd(req.Start, req.End, req.Step)
+
+	if req.Start > req.Step { // to avoid overflow
+		req.Start -= req.Step // force to have additional initial bucket
+	}
 }
 
 // Start time is rounded down to next step


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * [CHANGE] Adjust the definition of `tempo_metrics_generator_processor_service_graphs_expired_edges` to exclude edges that are counted in the service graph. [#5319](https://github.com/grafana/tempo/pull/5319) (@joe-elliott)
 * [CHANGE] Update Go to version 1.24.4 [#5323](https://github.com/grafana/tempo/pull/5323) (@stoewer)
 * [CHANGE] **BREAKING CHANGE** Drop unused `backend_scheduler.tenant_measurement_interval`, use `backend_scheduler.compaction.measure_interval` instead. [#5328](https://github.com/grafana/tempo/pull/5328) (@zalegrala)
+* [CHANGE] **BREAKING CHANGE** TraceQL Metrics buckets are calculated based on data in past [#5366](https://github.com/grafana/tempo/pull/5366) (@ruslan-mikhailov)
```
ruslan-mikhailov, line 15: Probably need to update documentation

modules/frontend/metrics_query_range_handler.go
```diff
@@ -88,6 +88,9 @@ func newMetricsQueryRangeHTTPHandler(cfg Config, next pipeline.AsyncRoundTripper
 				Body:       io.NopCloser(strings.NewReader(err.Error())),
 			}, nil
 		}
+		if queryRangeReq.Start > queryRangeReq.Step {
```
mdisibio: Does this need to be in the handler and the sharder?  It seems like that would do it twice.

modules/frontend/metrics_query_range_sharder.go
```diff
@@ -74,6 +74,9 @@ func (s queryRangeSharder) RoundTrip(pipelineRequest pipeline.Request) (pipeline
 	if err != nil {
 		return pipeline.NewBadRequest(err), nil
 	}
+	if !s.instantMode && req.Start > req.Step {
+		req.Start -= req.Step // force to have additional initial bucket
```
mdisibio: Does this need to be done after alignment?  The start may not require entire step to cover the first bucket.  I understand the intent:  if we are looking at an hour of data 12:00-1:00 at 1 minute intervals, we potentially need to back the start time to 11:59, to fulfill the bucket at 12:00.

pkg/traceql/engine_metrics.go
```diff
@@ -59,38 +59,50 @@ func IntervalCount(start, end, step uint64) int {
...
 
 	if !isTsValidForInterval(ts, start, end, step) {
 		return -1
 	}
+	if ts <= start { // to avoid overflow
+		return 0 // if pass validation and less than start, always first interval
+	}
 
-	return int((ts - start) / step)
+	interval := (ts - start - 1) / step
```
mdisibio: Is there a clearer way to write this or add comments?   I find subtracting a single nanosecond confusing.  It's done this way because we are upper-bound inclusive.

modules/frontend/metrics_query_range_handler.go
ruslan-mikhailov: Yep, need to do twice. I doubled checked by removing one of them and running e2e tests.

This is necessary, because it does not change origin `req *http.Request` and both call `api.ParseQueryRangeRequest(req)` independently

modules/frontend/metrics_query_range_sharder.go
ruslan-mikhailov: But won't it be the same as before the alignment?

image

pkg/traceql/engine_metrics.go
ruslan-mikhailov: Changed this code to make it a bit more readable and added comments

modules/frontend/metrics_query_range_handler.go
mdisibio: Ah got it.  The handler needs it for the combiner, which must be knowledgeable of the intervals for various reasons, including attaching exemplars.

But what's sticking out is the asymmetry between these areas:   does the combiner not need full alignment too?  And also the grpc streaming handler?  Think we need to dig into these a bit more, because I would expect that all mutation of `req` is identical for each entry point.

mdisibio (commented):
Ran this locally and it does look good.  Can confirm that it no longer always returns a 0 bucket for the last data point (because it was always looking forward where no data exists).  Also looking good for avg_over_time and compare.

But mentioned in the comment, need clarity about the slightly different handling in each case.

(1) Grpc streaming handler probably needs to do the same adjustment
(2) The handler/combiner probably needs to do full alignment (wondering how it works without it)
(3) Now if we are doing the same thing everywhere, can we collect it in one method, perhaps `traceql.AlignRequest` ?

ruslan-mikhailov:

PR #5373, opened 2025-07-08
Allow configuration of min/max input blocks for compaction provider

zalegrala:
**What this PR does**:

With the new scheduled compactions, we're observing that jobs handed out in sequence of difficulty often mean that the jobs are too easy for workers to complete, which results in lower throughput because the scheduler has to spend more time scheduling and handling the jobs.  Eventually the jobs handed out are more difficult, but it can take some time before this happens, which results in a valley of low throughput.  This may happen at the beginning of working on a new tenant, for example, and in a large environment with many tenants, the performance is sub-optimal.  I believe this warrants an allowable configuration on the compaction provider, proposed here.

Historically, the min/max input blocks have been hard-coded to match the parquet version performance.  It would be easy to try and compact too many blocks and drive memory use while combining.  Looking at the memory and some recent improvements in the compaction process, I think we can begin to think about allowing this to be configurable.

Users should go slow and think hard about changing these defaults.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


mapno (approved):
Can you add a chlog entry?

PR #5378, opened 2025-07-09
[metrics]: fix rate() calculation

zalegrala:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/engine_metrics.go
+++ b/pkg/traceql/engine_metrics.go
@@ -1272,6 +1272,7 @@ type SimpleAggregator struct {
 	aggregationFunc  func(existingValue float64, newValue float64) float64
 	start, end, step uint64
 	initWithNaN      bool
+	resultMultiplier float64 // Optional multiplier applied to final results (1.0 = no change)
 }
 
 func NewSimpleCombiner(req *tempopb.QueryRangeRequest, op SimpleAggregationOp, exemplars uint32) *SimpleAggregator {
@@ -1303,12 +1304,13 @@ func NewSimpleCombiner(req *tempopb.QueryRangeRequest, op SimpleAggregationOp, e
 			alignStart(req.Start, req.End, req.Step),
 			alignEnd(req.Start, req.End, req.Step),
 		),
-		len:             l,
-		start:           req.Start,
-		end:             req.End,
-		step:            req.Step,
-		aggregationFunc: f,
-		initWithNaN:     initWithNaN,
+		len:              l,
+		start:            req.Start,
+		end:              req.End,
+		step:             req.Step,
+		aggregationFunc:  f,
+		initWithNaN:      initWithNaN,
+		resultMultiplier: 1.0, // Default: no multiplier
 	}
 }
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 * [BUGFIX] Correct instant query calculation [#5252](https://github.com/grafana/tempo/pull/5252) (@ruslan-mikhailov)
 * [BUGFIX] Fix tracing context propagation in distributor HTTP write requests [#5312](https://github.com/grafana/tempo/pull/5312) (@mapno)
 * [BUGFIX] Correctly apply trace idle period in ingesters and add the concept of trace live period. [#5346](https://github.com/grafana/tempo/pull/5346/files) (@joe-elliott)
+* [BUGFIX] Fix rate() calculation to apply only on the final result. [#5378](https://github.com/grafana/tempo/pull/5378) (@zalegrala)
```
ruslan-mikhailov: Maybe we can highlight it fixes instant queries?

pkg/traceql/engine_metrics.go
```diff
@@ -1378,7 +1380,27 @@ func (b *SimpleAggregator) aggregateExemplars(ts *tempopb.TimeSeries, existing *
...
+			multipliedValues[i] = value * b.resultMultiplier
+		}
+
+		results[k] = TimeSeries{
+			Labels:    ts.Labels,
+			Values:    multipliedValues,
+			Exemplars: ts.Exemplars,
+		}
+	}
+	return results
```
ruslan-mikhailov, line 1518: It is better to apply the multiplier during Combine to avoid additional allocation and copying data 

Probably, also need to apply the multiplier to exemplar values as well

tempodb/tempodb_metrics_test.go
```diff
@@ -55,25 +55,39 @@ var queryRangeTestCases = []struct {
 				PromLabels: `{__name__="rate"}`,
```
ruslan-mikhailov: Could you please add a test case for instant query that reproduces the initial problem?

pkg/traceql/engine_metrics.go
zalegrala: I think its okay in the results, since its called last.  Only when the multiplier is not 1.0 as is the case on the final do we make the extra copy.  I looked at moving this to the Combine, but I think it adds some complexity.

The exemplars look okay to me here, and on an instant query we don't care about exemplars, so I think we don't need to modify them.
image

tempodb/tempodb_metrics_test.go
zalegrala: Okay, I've included a test to compare the `count_over_time/step_seconds` results matches the `rate` results with a small amount of tolerance.  Hopefully not too small of a tolerance to make the test noisy.

CHANGELOG.md
zalegrala: Does it fix both?  Instant makes it easier to compare, but I think the logic change is being applied to the range results also.

mdisibio:
> incorrect rate result being applied

Can you clarify this a bit more?  Is this intent of this change to minimize floating point error, or was a rate being miscalculated somewhere?

zalegrala:
I've updated the PR summary with some detail.  The original issue is written about instant queries, but the logic change here is for both instant and range, and comparing results from `main` shows an improvement in accuracy of the calculation.

zalegrala:
After some internal discussion, it looks like this might not be the right approach.  The concern is that with this PR we might be masking a bug in the `TrimToBlockOverlap` method where we handle the instant queries, and might close doors to future kinds of operations we might wish to at the lower levels.  I don't have a change prepared for the `TrimToBlockOverlap` method, and it could use some testing to confirm the concerns.

zalegrala:
I believe I've addressed all the feedback on this PR.  Would anyone like to review again?

PR #5380, opened 2025-07-10
Command tempo-cli analyse block(s) skips over array values

stoewer:
**What this PR does**:
The analyse block sub command is used to find candidates for dedicate attribute columns. So far the sub commands `analyse block` and `analyse blocks` included attributes containing string arrays in this analysis. As a result those attributes were configured as dedicated attribute columns.

Dedicated attribute columns only support string attributes. When attributes with string array values are configured as dedicated attribute column, it is no longer possible to query for those attributes.

With this change the respective sub commands skip over attributes with array values in order to avoid making misleading suggestions for dedicated attribute columns.

*Note: Reading the `IsArray` column would have been difficult with the `parquetinspect` package. Therefore the aggregation was re-implemented using iterators from `parquetquery`. The `parquetinspect` was no longer used and therefore removed.*

Contributes to https://github.com/grafana/tempo-squad/issues/724

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


ie-pham (approved):
lgtm

PR #5390, opened 2025-07-11
[Rhythm] nocompact.flg

ruslan-mikhailov:
**What this PR does**: 
For block-builder, before writing a block's meta.json, create `nocompact.flg` file. When compactor sees the file, it ignores the block. The flag is removed only after block-builder commits to Kafka. This must ensure that compaction can never be initiated on a block that is not committed.  

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/blockbuilder/blockbuilder_test.go
+++ b/modules/blockbuilder/blockbuilder_test.go
@@ -786,10 +786,10 @@ type ownEverythingSharder struct{}
 func (o *ownEverythingSharder) Owns(string) bool { return true }
 
 func newStore(ctx context.Context, t testing.TB) storage.Store {
-	return newStoreWithLogger(ctx, t, test.NewTestingLogger(t))
+	return newStoreWithLogger(ctx, t, test.NewTestingLogger(t), false)
 }
 
-func newStoreWithLogger(ctx context.Context, t testing.TB, log log.Logger) storage.Store {
+func newStoreWithLogger(ctx context.Context, t testing.TB, log log.Logger, skipNoCompactBlocks bool) storage.Store {
 	tmpDir := t.TempDir()
 
 	s, err := storage.NewStore(storage.Config{
@@ -814,7 +814,7 @@ func newStoreWithLogger(ctx context.Context, t testing.TB, log log.Logger) stora
 	}, nil, log)
 	require.NoError(t, err)
 
-	s.EnablePolling(ctx, &ownEverythingSharder{}, false)
+	s.EnablePolling(ctx, &ownEverythingSharder{}, skipNoCompactBlocks)
 	return s
 }
 
@@ -959,7 +959,7 @@ func BenchmarkBlockBuilder(b *testing.B) {
 		ctx        = context.Background()
 		logger     = log.NewNopLogger()
 		_, address = testkafka.CreateCluster(b, 1, testTopic)
-		store      = newStoreWithLogger(ctx, b, logger)
+		store      = newStoreWithLogger(ctx, b, logger, false)
 		cfg        = blockbuilderConfig(b, address, []int32{0})
 		client     = newKafkaClient(b, cfg.IngestStorageConfig.Kafka)
 		o          = &mockOverrides{
--- a/modules/blockbuilder/tenant_store_test.go
+++ b/modules/blockbuilder/tenant_store_test.go
@@ -162,7 +162,7 @@ func writeHistoricalData(t *testing.T, count int, startTime time.Time, cycleDura
 	var (
 		ctx   = t.Context()
 		log   = log.NewNopLogger()
-		store = newStoreWithLogger(ctx, t, log)
+		store = newStoreWithLogger(ctx, t, log, false)
 	)
 
 	ts, err := getTenantStore(t, startTime, cycleDuration, slackDuration)
@@ -199,7 +199,7 @@ func TestTenantStoreNoCompactFlag(t *testing.T) {
 		traceEnd      = startTime.Add(time.Minute)
 
 		log   = log.NewNopLogger()
-		store = newStoreWithLogger(ctx, t, log)
+		store = newStoreWithLogger(ctx, t, log, true)
 	)
 
 	ts, err := getTenantStore(t, startTime, cycleDuration, slackDuration)
--- a/tempodb/blocklist/poller_test.go
+++ b/tempodb/blocklist/poller_test.go
@@ -397,6 +397,7 @@ func TestPollBlockWithNoCompactFlag(t *testing.T) {
 		name                   string
 		hasNoCompactFlag       bool
 		noCompactFlagError     error
+		skipNoCompactBlocks    bool
 		expectedMeta           *backend.BlockMeta
 		expectedNoCompactCalls int
 		expectedBlockMetaCalls int
@@ -405,6 +406,7 @@ func TestPollBlockWithNoCompactFlag(t *testing.T) {
 		{
 			name:                   "block without nocompact flag is included",
 			hasNoCompactFlag:       false,
+			skipNoCompactBlocks:    true,
 			expectedMeta:           &backend.BlockMeta{BlockID: blockID, TenantID: tenantID},
 			expectedNoCompactCalls: 1,
 			expectedBlockMetaCalls: 1,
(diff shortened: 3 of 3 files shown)


tempodb/backend/raw.go
```diff
@@ -311,6 +324,19 @@ func (r *reader) Shutdown() {
 	r.r.Shutdown()
 }
 
+// HasNoCompactFlag implements backend.Reader
+func (r *reader) HasNoCompactFlag(ctx context.Context, blockID uuid.UUID, tenantID string) (bool, error) {
+	// can be optimized by not creating a descriptor and just checking file stat
```
ruslan-mikhailov: There is a third way how to do it. We can return such blocks in
```
	// Blocks returns the blockIDs, compactedBlockIDs and an error from the backend.
	Blocks(ctx context.Context, tenantID string) (blockIDs []uuid.UUID, compactedBlockIDs []uuid.UUID, err error)
```
that just checks files in blocks without opening them.

Here for the PR, I picked the simplest solution that requires minimum changes to interfaces

tempodb/backend/backend.go
```diff
@@ -54,6 +54,10 @@ type Writer interface {
 	WriteTenantIndex(ctx context.Context, tenantID string, meta []*BlockMeta, compactedMeta []*CompactedBlockMeta) error
 	// Delete deletes an object.
 	Delete(ctx context.Context, name string, keypath KeyPath) error
+	// WriteNoCompactFlag writes the nocompact flag to prevent a block from being compacted
+	WriteNoCompactFlag(ctx context.Context, blockID uuid.UUID, tenantID string) error
+	// DeleteNoCompactFlag deletes the nocompact flag to allow a block to be compacted
+	DeleteNoCompactFlag(ctx context.Context, blockID uuid.UUID, tenantID string) error
```
ruslan-mikhailov, line 60: Maybe set/unset naming style fits better for these two functions?

tempodb/blocklist/poller.go
```diff
@@ -477,6 +477,14 @@ func (p *Poller) pollBlock(
 	var compactedBlockMeta *backend.CompactedBlockMeta
 
 	if !compacted {
+		noCompact, flagErr := p.reader.HasNoCompactFlag(derivedCtx, blockID, tenantID)
+		if flagErr != nil {
+			return nil, nil, fmt.Errorf("failed to check nocompact flag: %w", flagErr)
+		}
+		if noCompact {
+			span.AddEvent("block excluded due to nocompact flag")
```
javiermolinar: Should this be a log instead? I don't see the point of tracing in the compactor tbh

javiermolinar (commented):
Should we do this just for the blockbuilder blocks? The goal is to ensure that the compactor does not pick blocks that Kafka offsets haven't been successfully committed

ruslan-mikhailov:
\+ rebase from latest main

tempodb/blocklist/poller.go
```diff
@@ -477,6 +477,13 @@ func (p *Poller) pollBlock(
 	var compactedBlockMeta *backend.CompactedBlockMeta
 
 	if !compacted {
+		noCompact, flagErr := p.reader.HasNoCompactFlag(derivedCtx, blockID, tenantID)
+		if flagErr != nil {
+			return nil, nil, fmt.Errorf("failed to check nocompact flag: %w", flagErr)
+		}
```
ruslan-mikhailov, line 484: As an alternative, we can just skip such a block. In the PR I replicated the way compactor handles errors with block meta

tempodb/blocklist/poller.go
ruslan-mikhailov: Removed this line

tempodb/tempodb.go
```diff
@@ -284,6 +285,10 @@ func (rw *readerWriter) CompleteBlockWithBackend(ctx context.Context, block comm
 	return backendBlock, nil
 }
 
+func (rw *readerWriter) DeleteNoCompactFlag(ctx context.Context, tenantID string, blockID backend.UUID) error {
+	return rw.w.DeleteNoCompactFlag(ctx, (uuid.UUID)(blockID), tenantID)
```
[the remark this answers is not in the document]
ruslan-mikhailov: All good here

tempodb/tempodb.go
```diff
@@ -75,6 +75,7 @@ type Writer interface {
 	WriteBlock(ctx context.Context, block WriteableBlock) error
 	CompleteBlock(ctx context.Context, block common.WALBlock) (common.BackendBlock, error)
 	CompleteBlockWithBackend(ctx context.Context, block common.WALBlock, r backend.Reader, w backend.Writer) (common.BackendBlock, error)
+	DeleteNoCompactFlag(ctx context.Context, tenantID string, blockID backend.UUID) error
```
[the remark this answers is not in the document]
ruslan-mikhailov: It is consistent with other interfaces

modules/blockbuilder/blockbuilder.go
```diff
@@ -456,6 +456,8 @@ outer:
 		return time.Time{}, -1, err
 	}
 
+	writer.allowCompaction(ctx, b.writer)
```
mapno: Nit: this should go below the log about committing

modules/blockbuilder/partition_writer.go
```diff
@@ -108,6 +108,31 @@ func (p *writer) flush(ctx context.Context, r tempodb.Reader, w tempodb.Writer,
...
+
+	wg := sync.WaitGroup{}
+	wg.Add(len(p.m))
+	for _, i := range p.m {
+		go func() {
+			defer wg.Done()
+			level.Info(p.logger).Log("msg", "allowing compaction", "tenant", i.tenantID)
+			err := i.AllowCompaction(ctx, w)
+			if err != nil {
+				level.Error(p.logger).Log("msg", "cannot allow compaction, the block will remain uncompacted", "tenant", i.tenantID, "err", err)
```
mapno: "cannot allow compaction" sounds like there is an impediment for it happening vs the operation failed.

```suggestion
				level.Error(p.logger).Log("msg", "failed to remove nocompact.flag, the block will remain uncompacted", "tenant", i.tenantID, "err", err)
```

modules/blockbuilder/partition_writer.go
```diff
@@ -108,6 +108,31 @@ func (p *writer) flush(ctx context.Context, r tempodb.Reader, w tempodb.Writer,
...
+		trace.WithAttributes(
+			attribute.Int("partition", int(p.partition)),
+			attribute.String("section_start_time", p.startTime.String()),
+		),
+	)
+	defer span.End()
+
+	wg := sync.WaitGroup{}
+	wg.Add(len(p.m))
+	for _, i := range p.m {
```
mapno, line 123: It should take the map's mutex here

tempodb/backend/raw.go
mapno: Leaving a TODO for future optimization is good enough IMO.

tempodb/blocklist/poller.go
```diff
@@ -477,6 +477,13 @@ func (p *Poller) pollBlock(
...
+		if noCompact {
+			return nil, nil, nil
+		}
```
mapno, line 487: Blocks can't be skipped here. This is called for building the index, which would exclude them in the read path.

I think the blocklist needs to either return that info when `List.Metas()` is called or to conditionally include/exclude them.

modules/blockbuilder/blockbuilder.go
ruslan-mikhailov: \+

modules/blockbuilder/partition_writer.go
ruslan-mikhailov: \+

This layer does not know about the flag implementation, so I changed the log to `failed to allow compaction, the block will remain uncompacted`.

ruslan-mikhailov:
\+ review fixes

modules/blockbuilder/partition_writer.go
mapno: Taking a second look, it doesn't seem necessary

tempodb/blocklist/poller.go
ruslan-mikhailov: I've fixed it, now only compactor ignores such blocks. This test can help to understand the new behaviour: https://github.com/grafana/tempo/pull/5390/files#diff-3329701faa3228c7f00c2b01a586fa8d2976c632050a1523cded9cecc4f3d66fR859-R886

mapno (approved):
Let's go!

modules/backendscheduler/backendscheduler.go
```diff
@@ -108,7 +108,7 @@ func (s *BackendScheduler) starting(ctx context.Context) error {
 	level.Info(log.Logger).Log("msg", "backend scheduler starting")
 
 	if s.cfg.Poll {
-		s.store.EnablePolling(ctx, blocklist.OwnsNothingSharder)
```
zalegrala, line 111: I think we need this to be true here as well.

To make sure I understand the approach, we're trying to keep blocks out of the compactor's blocklist which we want to avoid compacting, correct?  All other pollers will continue to ignore?  And each polling cycle, we make an additional request to the backend per-block to determine if the special object is in place.  Once the object is removed from the backend, the poller will read the block into the blocklist and allow the compactor to compact it, correct?

What's the reason this change is made to the poller, and not only the compactor?

mapno: >I think we need this to be true here as well.

You're right, we missed this 🙏 

>What's the reason this change is made to the poller, and not only the compactor?

It's only wanted on the compactor, but the compactor also runs the poller. Whether the poller is run for compaction or for reading, the code is shared.

Also, the scheduler doesn't run the compactor, only the poller, no?

ruslan-mikhailov: Thanks!

Fix: https://github.com/grafana/tempo/pull/5440

zalegrala: > Also, the scheduler doesn't run the compactor, only the poller, no?

Yes, the scheduler runs the read-only poller, but since it is handing out jobs for which blocks to compact, I think the desire is to make those blocks invisible to the compaction process until the "right" time, so if it is excluded on the scheduler, then it won't know about the blocks, and won't hand out jobs which contain those blocks.  If the scheduler handed out a job which contained a block we didn't want to compact, then the compactor would skip it, but it would report back the job status that it had in fact compacted the block, which is strange.

PR #5398, opened 2025-07-13
feat(service-graphs): add database_name_attributes config to service graph processor 

### LINKED ISSUE #5243: update servicegraph database request handling

Currently, when a service graph connection type is identified as a database, we still prefer to set the server service to either `peer.service`, `server.address`, or `network.peer.address`.  This results in service graph database nodes showing ip addresses rather than their database name in most cases. 

When connection type is identified as a database, we should prefer to use the database name attribute as the server service name.

This could be configurable to preserve the old behavior by adding a `database_name_attributes` list to the config. If set, ideally we would prefer to set the dbName to the first matching attribute from this list.

This is similar to the way the otel service graph connector works:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/connector/servicegraphconnector/connector.go#L276-L282

A couple of alternatives:
- Do nothing. This prevents us from being able to control the name of database nodes in the service graph while maintaining additional the peer attributes on spans.
- Replace existing behavior to match otel connector. This would likely be better than what we do today but it could come as a surprise for existing users of the service graph processor in tempo.

**Additional context**

KyriosGN0:

**What this PR does**:

this PR introduces a array of database name attributes that the service graph processor looks for, if it finds one of them it uses them for edge node name, this is similiar to the way the service graph processor behaves in the otel collector. 
the list is currently empty but if we want we can set the `db.name` as default (like in the collector) but that would require changes to the other test 

Fixes #5243

**Checklist**
- [X] Tests updated
- [X] Documentation added

### DIFF
--- a/docs/sources/tempo/configuration/manifest.md
+++ b/docs/sources/tempo/configuration/manifest.md
@@ -641,7 +641,10 @@ metrics_generator:
                 - db.system
             span_multiplier_key: ""
             enable_virtual_node_label: false
-            database_name_attributes: []
+            database_name_attributes:
+                - db.namespace
+                - db.name
+                - db.system
         span_metrics:
             histogram_buckets:
                 - 0.002


KyriosGN0:
Not stale

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -307,6 +307,14 @@ func (p *Processor) upsertDatabaseRequest(e *store.Edge, resourceAttr []*v1_comm
 	e.ServerLatencySec = spanDurationSec(span)
 
 	// Set the service name by order of precedence
+	if len(p.Cfg.DatabaseNameAttributes) > 0 {
```
joe-elliott: what do you think about replacing this logic:

https://github.com/grafana/tempo/blob/main/modules/generator/processor/servicegraphs/servicegraphs.go#L290-L298

with a list of attributes that tempo checks for the db name? the configured list could default to the two we have currently but people could configure it to whatever they want.

docs/sources/tempo/configuration/_index.md
```diff
@@ -461,6 +461,9 @@ metrics_generator:
             # Enables additional labels for services and virtual nodes.
             [enable_virtual_node_label: <bool> | default = false]
 
+            # list of attribute names used to identify the database name from span attributes. if not set the order is peer.service -> server.address -> network.peer.address -> db.name
```
knylander-grafana:
```suggestion
            # List of attribute names used to identify the database name from span attributes. If it isn't set, the order is peer.service -> server.address -> network.peer.address -> db.name.
```

modules/generator/processor/servicegraphs/servicegraphs.go
KyriosGN0: sounds good to me, will update the PR

KyriosGN0: @joe-elliott Done!

docs/sources/tempo/configuration/_index.md
KyriosGN0: done!

modules/generator/processor/servicegraphs/config.go
```diff
@@ -62,4 +67,10 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
 	cfg.PeerAttributes = peerAttr
 
 	cfg.EnableMessagingSystemLatencyHistogram = false
+
+	cfg.DatabaseNameAttributes = []string{
+		string(semconvnew.DBNamespaceKey),
```
joe-elliott, line 72: i nearly commented that these are out of order with the previous checks, but these are in fact correct. the previous logic is just awfully convoluted. nice catch

KyriosGN0:
@joe-elliott added changelog entry, Thanks!

KyriosGN0:
Weird, the linter complains about files I didn't touch

joe-elliott:
Yeah, I've seen this happen before after rebase. Humorously the files that are failing lint have actually been deleted from main.

I'm going to drop the lint requirement real quick and merge.

PR #5405, opened 2025-07-14
[backend-scheduler] fix duplicate work handout during race

zalegrala:
**What this PR does**:

A race exists between the compaction provider and the Next call of the scheduler where the provider may detect the same block IDs for a job for which the cache has not yet been updated.  This results in the failed jobs and duplicate data in the backend due to two compactions happening on the same blocks.

Here we work around the race by checking the work list right before a job is handed out to a worker to ensure that we have not already assigned a job with the same block IDs.  This does not solve the underlying race, but prevents the conditions whereby this becomes a problem.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/backendscheduler/backendscheduler.go
+++ b/modules/backendscheduler/backendscheduler.go
@@ -252,6 +252,17 @@ func (s *BackendScheduler) Next(ctx context.Context, req *tempopb.NextJobRequest
 			attribute.String("job_id", j.GetID()),
 		))
 
+		// Check for overlapping compaction jobs before assigning this job
+		// This prevents the race condition where duplicate jobs are created
+		// for the same block IDs before the work cache is updated
+		if s.work.HasOverlappingCompactionJob(j) {
+			span.AddEvent("job rejected due to overlapping blocks")
+
+			level.Info(log.Logger).Log("msg", "rejected job due to overlapping compaction blocks", "job_id", j.ID, "worker", req.WorkerId)
+			metricJobsRejectedDuplicate.WithLabelValues(j.JobDetail.Tenant).Inc()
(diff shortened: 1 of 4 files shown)


modules/backendscheduler/metrics.go
```diff
@@ -38,6 +38,11 @@ var (
 		Name:      "backend_scheduler_jobs_not_found_total",
 		Help:      "The number of calls to get a job that were not found",
 	}, []string{"worker_id"})
+	metricJobsRejectedDuplicate = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "tempo",
+		Name:      "backend_scheduler_jobs_rejected_duplicate_total",
```
electron0zero, line 43: my understanding is that in ideal case, this metric should be zero, if you are creating duplicate jobs, something is wrong? is that correct?

if yes, we should add an alert on this metric, maybe in a follow up PR.

modules/backendscheduler/backendscheduler.go
```diff
@@ -252,6 +253,17 @@ func (s *BackendScheduler) Next(ctx context.Context, req *tempopb.NextJobRequest
 			attribute.String("job_id", j.GetID()),
 		))
 
+		// Check for overlapping compaction jobs before assigning this job
+		// This prevents the race condition where duplicate jobs are created
```
electron0zero, line 257: is it possible to solve this by making the job creation & update of work cache an atomic operation? or no?

I don't mean to say that those can operations can be made atomic, just want to understand how hard it would be to make them atomic or if we can't make them atomic, what would be the tradeoff?

zalegrala:
I've put up https://github.com/grafana/tempo/pull/5409 as an alternative to this approach, which is a cleaner and more performant.

PR #5407, opened 2025-07-15
[Feature] s3 checksum flag

ruslan-mikhailov:
**What this PR does**:
Minio library we use for S3 backend uses CRC32C for checksum. Some object storages do not support it. The PR adds a configration parameter that allows to set what checksum to use.

Fixes #<issue number>

**How this PR has been tested:**
**TODO!**

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


tempodb/backend/s3/config.go
```diff
@@ -81,6 +93,27 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
...
+	case ChecksumSHA1:
+		return minio.ChecksumSHA1
+	case ChecksumCRC32:
+		return minio.ChecksumCRC32
+	case ChecksumCRC32C:
+		return minio.ChecksumCRC32C
+	case ChecksumCRC64NVME:
+		return minio.ChecksumCRC64NVME
+	default:
+		return minio.ChecksumCRC32C
```
ruslan-mikhailov, line 122: Should I write warn log in this case or even panic?

joe-elliott: i would panic. personally prefer loud, obvious errors on bad config like this

ruslan-mikhailov: Added validation. Now application does not run if config value is invalid.

docs/sources/tempo/configuration/manifest.md
```diff
@@ -825,6 +825,7 @@ storage:
                 type: ""
                 kms_key_id: ""
                 kms_encryption_context: ""
+            checksum_type: CRC32C
```
joe-elliott: instead of defaulting to a specific value can we default to minio's default? presumably they would do a better job of keeping this up to date

integration/e2e/checksum_test.go
```diff
@@ -0,0 +1,121 @@
...
+	tempoUtil "github.com/grafana/tempo/pkg/util"
+	"github.com/grafana/tempo/pkg/util/test"
+)
+
+const (
+	configS3Checksum = "./config-s3-checksum.tmpl.yaml"
+)
+
+func TestS3Checksum(t *testing.T) {
+	checksumTypes := []string{
```
joe-elliott, line 30: this e2e test is interesting. does minio work with all these checksum types? the s3 client requests a specific checksum from s3 which is honored or not?

i suppose there's no way to validate that the actual type is being requested/honored?

tempodb/backend/s3/s3.go
```diff
@@ -119,6 +119,10 @@ func internalNew(cfg *Config, confirm bool) (*readerWriter, error) {
 		return nil, fmt.Errorf("config is nil")
 	}
 
+	if err := cfg.Validate(); err != nil {
```
javiermolinar, line 122: I'm ok with this, but we already have another piece of code validating the config. 

https://github.com/ruslan-mikhailov/tempo/blob/d1f2f7f38a4a5087c335b76069be4bcdafe54809/tempodb/backend/s3/s3.go#L146C2-L149C3

docs/sources/tempo/configuration/manifest.md
ruslan-mikhailov: I checked the library. When minio.ChecksumNone is set, although it says `ChecksumNone indicates no checksum.`, it overwrites it with default in some cases. I've removed "None" option and replaced with empty string, which will be default.

integration/e2e/checksum_test.go
ruslan-mikhailov: I started digging and found out that in our use cases, minio does not use checksums. I temporarily put `-1` to size [here](https://github.com/ruslan-mikhailov/tempo/blob/cbb62d6e2d8494c4cc59e7a300334926f98bf91e/tempodb/backend/s3/s3.go#L196) to force the use of checksums, and run the test again. It worked well.

From a white box perspective, this test does nothing, but I think we can leave it as a regression test in case implementation changes

> i suppose there's no way to validate that the actual type is being requested/honored?

I used debug logs to ensure checksum were used. Except for mitm traffic inspection, no idea how else to check in e2e

tempodb/backend/s3/s3.go
ruslan-mikhailov: I did not change it because it does not really simplify code in my opinion. Here how it looks like: https://github.com/grafana/tempo/commit/5a6d023d3cf85710db34fc34b436d0b4af92a6d1 , only `cfg.SSE.KMSKeyID == ""` was removed.

Let me know if you have better ideas how to reorganise the code

joe-elliott (approved):

ruslan-mikhailov:
We shouldn't have sent a checksum in the first place, there is a bug in minio-go library. Closing the PR so far. If we do need to send and configure the checksum, need to reopen the PR, update the library with fix for the issue and retest.

PR #5409, opened 2025-07-15
[backend-scheduler] fix race condition between scheduler and compaction provider

zalegrala:
**What this PR does**:

This is an alternative approach to the same issue described in #5405.

Here we fix the race by keeping track of the recent jobs which were created and their blocks.  This cache is used to skip blocks on the block selector from being considered.  Once the next blockselector is called, we use the job list to identify what cache needs to be pruned.  This allows a small delay between when the compaction provider created the job and when the scheduler assigns the job, but without the additional overhead of looping through the work list as proposed in #5405.

Fixes #5405

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/backendscheduler/config.go
+++ b/modules/backendscheduler/config.go
@@ -2,6 +2,7 @@ package backendscheduler
 
 import (
 	"flag"
+	"fmt"
 	"time"
 
 	"github.com/grafana/tempo/modules/backendscheduler/provider"
@@ -40,6 +41,15 @@ func ValidateConfig(cfg *Config) error {
 
 	if err := provider.ValidateConfig(&cfg.ProviderConfig); err != nil {
 		return err
+	}
+
+	// Validate the measurement interval is twice the speed of the prune interval
+	// so that the when newBlockSelector is called it has enough time to delete
+	// the temporary entry and know that it has been persisted to the work cache.
+	// If the prune age is too short, work could get deleted before the
+	// newBlockSelector is able to delete the temporary entry.
+	if cfg.ProviderConfig.Compaction.MeasureInterval > cfg.Work.PruneAge/2 {
+		return fmt.Errorf("provider.compaction.measure_interval must be at least twice the work.prune_age, got %s and %s", cfg.ProviderConfig.Compaction.MeasureInterval, cfg.Work.PruneAge/2)
 	}
 
 	return nil
--- a/modules/backendscheduler/config.go
+++ b/modules/backendscheduler/config.go
@@ -49,7 +49,7 @@ func ValidateConfig(cfg *Config) error {
 	// If the prune age is too short, work could get deleted before the
 	// newBlockSelector is able to delete the temporary entry.
 	if cfg.ProviderConfig.Compaction.MeasureInterval > cfg.Work.PruneAge/2 {
-		return fmt.Errorf("provider.compaction.measure_interval must be at least twice the work.prune_age, got %s and %s", cfg.ProviderConfig.Compaction.MeasureInterval, cfg.Work.PruneAge/2)
+		return fmt.Errorf("provider.compaction.measure_interval must be no more than half of work.prune_age; tenant measurement should happen at least twice as often as the work prune, got %s and %s", cfg.ProviderConfig.Compaction.MeasureInterval, cfg.Work.PruneAge/2)
 	}
 
 	return nil
--- a/modules/backendscheduler/provider/compaction.go
+++ b/modules/backendscheduler/provider/compaction.go
@@ -4,6 +4,7 @@ import (
 	"container/heap"
 	"context"
 	"flag"
+	"sync"
 	"time"
 
 	"github.com/go-kit/log"
@@ -68,7 +69,8 @@ type CompactionProvider struct {
 	lastPrioritizeTime time.Time
 
 	// Recent jobs cache for duplicate block ID prevention.
-	recentJobs map[string][]string
+	recentJobs    map[string][]string
+	recentJobsMtx sync.Mutex
 }
 
 func NewCompactionProvider(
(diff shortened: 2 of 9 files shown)


modules/backendscheduler/provider/compaction.go
```diff
@@ -68,6 +69,10 @@ type CompactionProvider struct {
 	curPriority *tenantselector.PriorityQueue
 	curTenant   *tenantselector.Item
 	curSelector blockselector.CompactionBlockSelector
+
+	// Recent jobs cache for duplicate block ID prevention.
+	recentJobs    map[string][]string
```
mdisibio: Might rename this to `outstandingJobs` or something since it is to cover the jobs written to the channel that don't show up in `ListJobs()` yet?  It took me a minute to understand that distinction, versus any recently-issued job.

modules/backendscheduler/provider/compaction.go
```diff
@@ -381,9 +393,23 @@ func (p *CompactionProvider) newBlockSelector(tenantID string) (blockselector.Co
...
+	// Also include blocks from recent jobs cache to prevent duplicate job creation
+	p.recentJobsMtx.Lock()
+	for _, recentBlockIDs := range p.recentJobs {
+		for _, blockID := range recentBlockIDs {
+			bid, err = backend.ParseUUID(blockID)
+			if err != nil {
+				level.Error(p.logger).Log("msg", "failed to parse recent job block ID", "block_id", blockID, "err", err)
+				continue
+			}
+			perTenantBlockIDs[bid] = struct{}{}
```
mdisibio: Might rename `perTenantBlockIDs` to `inProgressBlocks` ?  Since this function is already per-tenant, somewhat redundant to keep that in the name.

modules/backendscheduler/provider/compaction.go
mdisibio: It might make sense to make this cache `map[string][]uuid` instead of string.  I think that would be more efficient since then we only need to parse the UUID once on write, instead of every read.

modules/backendscheduler/provider/compaction.go
```diff
@@ -361,6 +368,11 @@ func (p *CompactionProvider) newBlockSelector(tenantID string) (blockselector.Co
 	)
 
 	for _, job := range p.sched.ListJobs() {
+		// Clean up recent jobs cache when we find a job which has been persisted
+		p.recentJobsMtx.Lock()
+		delete(p.recentJobs, job.ID)
```
mdisibio: Could there still be a gap where a block is no longer in ListJobs, and the new compacted status hasn't been polled yet?

modules/backendscheduler/provider/compaction.go
zalegrala: Good call. The name here was something left over from previous.

modules/backendscheduler/provider/compaction.go
zalegrala: Hmm, good idea.  I see now that the same could be said for the job itself, but perhaps not in this PR.  I'll push a commit for the map change.

modules/backendscheduler/provider/compaction.go
zalegrala: This is a good point.  It would be possible to configure the polling cycle and maintenance interval of the scheduler to be at odds with each-other.  I think we can solve this with a configuration validation to avoid this.  In practice, I don't expect this to come up and the default values are sane for the case you mention.

zalegrala: Hmm, the config isn't plumbed in a way that makes this simple, since the store has the polling cycle.  I'll think about this a bit more.

zalegrala: I just needed to validate at a higher level than the scheduler.  I've pushed a commit for this.

Reading this closely, there is another gap where the newBlockSelector is not called faster than the prune age.  I think we can make sure that the measurement interval is 2x faster than the prune age, which would cover this case.  I'll push another validation commit.

modules/backendscheduler/config.go
```diff
@@ -42,5 +43,14 @@ func ValidateConfig(cfg *Config) error {
 		return err
 	}
 
+	// Validate the measurement interval is twice the speed of the prune interval
+	// so that the when newBlockSelector is called it has enough time to delete
+	// the temporary entry and know that it has been persisted to the work cache.
+	// If the prune age is too short, work could get deleted before the
+	// newBlockSelector is able to delete the temporary entry.
+	if cfg.ProviderConfig.Compaction.MeasureInterval > cfg.Work.PruneAge/2 {
```
mdisibio, line 51: Is the error message backward?  This is checking that `prune_age` is at least twice the `measurement interval`.

modules/backendscheduler/provider/compaction.go
```diff
@@ -377,13 +389,22 @@ func (p *CompactionProvider) newBlockSelector(tenantID string) (blockselector.Co
...
+		for _, blockID := range recentBlockIDs {
+			inProgressBlockIDs[blockID] = struct{}{}
+		}
+	}
+	p.outstandingJobsMtx.Unlock()
+
 	for _, block := range fullBlocklist {
-		if _, ok := perTenantBlockIDs[block.BlockID]; ok {
-			// Skip blocks that are already in the input list from another job
+		if _, ok := inProgressBlockIDs[block.BlockID]; ok {
```
mdisibio: nite: This loop could be simplified:
```
if ... ; !ok {
   append...
}
```

modules/backendscheduler/config.go
zalegrala: Hmm, yeah this is a funny word game.  I see I didn't put "interval" in there, so yeah reads a little strange.  Half the duration is twice the interval.
To reword, perhaps we say it both ways?
`provider.compaction.measure_interval must be no more than half of work.prune_age; tenant measurement should happen at least twice as often as the work pruning, got %s and %s`

zalegrala: I'll push a commit, but let me know if that still doesn't ring to you and we'll update.

modules/backendscheduler/provider/compaction.go
zalegrala: I've pushed a commit for the additional validation.

PR #5410, opened 2025-07-15
[backend-scheduler] require minimum time between tenant sorting

zalegrala:
**What this PR does**:

Here we use a minimum cycle time to ensure we move fast when we can, but not too fast when we don't need to.

New blocks arrive at the end of a poll cycle, and we shouldn't expect to find new work on a tenant which has been drained until then.

**BREAKING CHANGE**
The config setting `provider.compaction.backoff` has been removed in favor of this simplified approach.

Additionally the `compaction_tenant_backoff_total` metric has been renamed to `compaction_empty_tenant_cycle_total` for clarity.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/backendscheduler/provider/compaction.go
+++ b/modules/backendscheduler/provider/compaction.go
@@ -30,10 +30,9 @@ type CompactionConfig struct {
 	MeasureInterval  time.Duration           `yaml:"measure_interval"`
 	Compactor        tempodb.CompactorConfig `yaml:"compaction"`
 	MaxJobsPerTenant int                     `yaml:"max_jobs_per_tenant"`
-	// Backoff          backoff.Config          `yaml:"backoff"`
-	MinInputBlocks   int           `yaml:"min_input_blocks"`
-	MaxInputBlocks   int           `yaml:"max_input_blocks"`
-	MinCycleInterval time.Duration `yaml:"min_cycle_interval"`
+	MinInputBlocks   int                     `yaml:"min_input_blocks"`
+	MaxInputBlocks   int                     `yaml:"max_input_blocks"`
+	MinCycleInterval time.Duration           `yaml:"min_cycle_interval"`
 }
 
 func (cfg *CompactionConfig) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet) {
@@ -129,8 +128,8 @@ func (p *CompactionProvider) Start(ctx context.Context) <-chan *work.Job {
 			if p.curSelector == nil {
 				if !p.prepareNextTenant(loopCtx) {
 					level.Info(p.logger).Log("msg", "received empty tenant")
-					metricTenantBackoff.Inc()
-					span.AddEvent("tenant not prepared")
+					metricEmptyTenantCycle.Inc()
+					span.AddEvent("no tenant selected")
 				}
 
 				if p.curTenant != nil {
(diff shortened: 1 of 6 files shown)


modules/backendscheduler/provider/compaction.go
```diff
@@ -49,6 +50,9 @@ func (cfg *CompactionConfig) RegisterFlagsAndApplyDefaults(prefix string, f *fla
 	f.DurationVar(&cfg.Backoff.MaxBackoff, prefix+".backoff-max-period", 10*time.Second, "Maximum delay when backing off.")
 	cfg.Backoff.MaxRetries = 0
```
zalegrala: I think we can completely replace the backoff with this approach, which is cleaner and more specific about its desire not to run wild with memory allocations.

modules/backendscheduler/provider/config.go
```diff
@@ -23,8 +23,8 @@ func ValidateConfig(cfg *Config) error {
 		return fmt.Errorf("max_jobs_per_tenant must be greater than 0")
 	}
 
-	if cfg.Compaction.Backoff.MaxRetries != 0 {
-		return fmt.Errorf("max_retries must be 0, since it is not respected")
+	if cfg.Compaction.MinCycleInterval <= 0 {
+		return fmt.Errorf("min_cycle_interval must be greater than 0")
```
zalegrala: Should we raise our minimum guidance here?

modules/backendscheduler/provider/compaction.go
```diff
@@ -129,12 +126,16 @@ func (p *CompactionProvider) Start(ctx context.Context) <-chan *work.Job {
...
 				if !p.prepareNextTenant(loopCtx) {
-					level.Info(p.logger).Log("msg", "received empty tenant", "waiting", b.NextDelay())
-					metricTenantBackoff.Inc()
-					span.AddEvent("tenant not prepared")
+					level.Info(p.logger).Log("msg", "received empty tenant")
+					metricEmptyTenantCycle.Inc()
+					span.AddEvent("no tenant selected")
 				}
+
+				if p.curTenant != nil {
```
joe-elliott: nit: we could log this in prepareNextTenant() right before we call p.newBlockSelector() and avoid the if.

modules/backendscheduler/provider/compaction.go
```diff
@@ -200,7 +198,22 @@ func (p *CompactionProvider) prepareNextTenant(ctx context.Context) bool {
 	defer span.End()
 
 	if p.curPriority.Len() == 0 {
+		// Rate limit calls to prioritizeTenants to prevent excessive CPU usage
+		// when cycling through tenants with no available work.  We only expect new
+		// work for tenants after a the next blocklist poll.
+		if elapsed := time.Since(p.lastPrioritizeTime); elapsed < p.cfg.MinCycleInterval {
```
joe-elliott, line 200: i suppose this works on the first pass b/c the 0 value of time.Time is very very far in the past?

modules/backendscheduler/provider/config.go
joe-elliott: no harm + potential small benefit in adding some guidance here. something like:

"min_cycle_interval must be greater than 0 and it's advised to be at lesat ... "

joe-elliott (approved):
nice simplification. likely easier to predict behavior here?

modules/backendscheduler/provider/compaction.go
zalegrala: Nice, good eyes.

modules/backendscheduler/provider/compaction.go
zalegrala: Right, zero value start and we only move it forward.

zalegrala:
More predictable and improved behavior.  Thanks for the review.

PR #5412, opened 2025-07-16
[backend-scheduler] initial work file sharding for more targeted marshaling

zalegrala:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/backendscheduler/backendscheduler.go
```diff
@@ -69,7 +69,7 @@ func New(cfg Config, store storage.Store, overrides overrides.Interface, reader
 		cfg:        cfg,
 		store:      store,
 		overrides:  overrides,
-		work:       work.New(cfg.Work),
+		work:       work.NewWork(cfg.Work),
```
zalegrala: I don't love this being called `NewWork` when we also have a `New` method.  I want to change this, but I'm concerned about the size of the diff here. Perhaps we can submit a simple rename after this PR.

zalegrala:
A lot of the complexity of this PR comes from migrating the state back and forth, and the special handling in there.  After some internal conversations, I'm thinking about dropping the old path entirely and skipping all of the migration and backwards compatibility so that we only have one path, which would be sharded.

I'd like to hear some perspectives from the team.  If this sounds like a better approach, I'll reduce the complexity here and remove the old path.  Thoughts from the group?

zalegrala:
I'm going to take this back to draft while I work up the diff to remove the dual implementation approach.  I think this will simplify the approach and make review easier.

modules/backendscheduler/work/atomic.go
```diff
@@ -0,0 +1,44 @@
+package work
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// atomicWriteFile writes data to a file atomically using a temporary file and rename.
+// This prevents race conditions and ensures readers never see partial data.
+func atomicWriteFile(data []byte, targetPath, tempPrefix string) error {
```
mdisibio, line 10: What is concurrently reading the local work files?  It looks like the scheduler is always writing to them (as jobs are created and updated), and the flush to object storage is a remarshal from memory, and not reading the shard files.

zalegrala: Nothing is reading concurrently.  The race I'm mentioning here is that we have a mutex on the marshal, but not on the write.  If we have multiple Next or UpdateJob calls happening concurrently, the writes of the local files are racing to put the bytes on disk.

zalegrala: The message should say "readers never in a broken state".
The other condition I'd like to guard against is a crash right in the middle of a write.  If we take the atomic approach, I think we ensure that the work files on disk are always readable.

mdisibio: We discussed and the main advantage is to prevent corrupted files on unexpected shutdown (i.e. panic/oom). So let's update the comment to reflect that. There doesn't actually appear to be any concurrent reads, because of the locking.

modules/backendscheduler/work/work.go
```diff
@@ -13,186 +17,449 @@ import (
...
+// flushAffectedShards writes only the shards that contain the affected jobs using atomic operations
+func (w *Work) flushAffectedShards(localPath string, affectedJobIDs []string) error {
+	affectedShards := make(map[uint8]bool)
+	for _, jobID := range affectedJobIDs {
+		shardID := w.GetShardID(jobID)
+		affectedShards[shardID] = true
+	}
+
+	for shardID := range affectedShards {
+		shardData, err := w.MarshalShard(shardID)
```
mdisibio: Discussed and we need to lock the shard here:  If there are concurrent updates from Next or UpdateJob, they need to be serialized. The second write to the file can't lose the updates made by the first write.  Therefore both marshaling and writing have to be done under the lock for the shard.

modules/backendscheduler/work/work.go
```diff
@@ -13,186 +17,446 @@ import (
...
+
+	for shardID := range shards {
+		shard = w.Shards[shardID]
+		funcErr = func(shard *Shard) error {
+			shard.mtx.Lock()
+			defer shard.mtx.Unlock()
+
+			clear(shardData)
+
+			shardData, err = jsoniter.Marshal(shard)
```
joe-elliott: this is alloc'ing/overwriting shardData so there's no reason to hold this like a buffer. just create it locally in the function.

modules/backendscheduler/work/work.go
```diff
@@ -13,186 +17,446 @@ import (
...
-	if j, ok := q.Jobs[id]; ok {
-		j.SetCompactionOutput(output)
+// getShard returns the shard for a given job ID
+func (w *Work) getShard(jobID string) *Shard {
+	return w.Shards[w.getShardID(jobID)]
+}
+
+// flushAllShards writes all shards to individual files using atomic operations
+func (w *Work) flushAllShards(localPath string) error {
+	shards := make(map[uint8]bool, ShardCount)
```
joe-elliott: uint8 optimization probably not worth it (and may even be slower) and something we may trip over later. i'd just make it `map[int]bool`

modules/backendscheduler/work/work.go
```diff
@@ -13,186 +17,446 @@ import (
...
 
-	delete(q.Jobs, id)
+	delete(shard.Jobs, id)
 }
 
-func (q *Work) ListJobs() []*Job {
-	q.mtx.Lock()
+// ListJobs returns all jobs across all shards, sorted by creation time
+func (w *Work) ListJobs() []*Job {
+	var allJobs []*Job
```
joe-elliott: if this is called infrequently than this implementation is good. if this is called regularly/constantly it maybe faster to iterate the shards, count the total jobs and pre-alloc this slice

modules/backendscheduler/work/work.go
```diff
@@ -13,186 +17,446 @@ import (
...
 			}
-		case tempopb.JobStatus_JOB_STATUS_RUNNING:
-			// Fail jobs which have been running for too long
-			if time.Since(j.GetStartTime()) > q.cfg.DeadJobTimeout {
-				j.Fail()
+
+			switch j.GetStatus() {
+			case tempopb.JobStatus_JOB_STATUS_UNSPECIFIED, tempopb.JobStatus_JOB_STATUS_RUNNING:
+				shard.mtx.Unlock()
+				return j
```
joe-elliott: this can return and leave a mutex locked?

modules/backendscheduler/work/work.go
```diff
@@ -13,186 +17,446 @@ import (
...
+	defer shard.mtx.Unlock()
+
+	return jsoniter.Unmarshal(data, shard)
 }
 
-func (q *Work) FailJob(id string) {
-	q.mtx.Lock()
-	defer q.mtx.Unlock()
+// GetShardStats returns statistics about job distribution across shards
+func (w *Work) GetShardStats() map[string]any {
```
joe-elliott, line 369: not seeing this used?

modules/backendscheduler/work/work.go
```diff
@@ -13,186 +17,446 @@ import (
...
+			defer shard.mtx.Unlock()
+
+			clear(shardData)
+
+			shardData, err = jsoniter.Marshal(shard)
+			if err != nil {
+				return err
+			}
+
+			filename = fmt.Sprintf("shard_%03d.json", shardID)
```
joe-elliott: this is defined in two different places which is a danger. once here and once where we check for shards. let's collapse to one

modules/backendscheduler/cache.go
```diff
@@ -9,56 +9,11 @@ import (
 	"path/filepath"
 
 	"github.com/go-kit/log/level"
+	"github.com/grafana/tempo/modules/backendscheduler/work"
 	"github.com/grafana/tempo/pkg/util/log"
 	"github.com/grafana/tempo/tempodb/backend"
-	"go.opentelemetry.io/otel/attribute"
 )
 
-func (s *BackendScheduler) flushWorkCache(ctx context.Context) error {
```
joe-elliott, line 17: it was very nice organizing all the flush/load code into this file. i gather we moved a lot of the code into the work folder b/c of the complexity of adding shards. if possible it would be nice to keep this all in one spot tho

modules/backendscheduler/work/work.go
zalegrala: Good call, I see the `make` call in there now.

modules/backendscheduler/work/work.go
zalegrala: Interesting, can you elaborate on the performance?  I put this in to constrain to the size of the hash space more than thinking about speed.

modules/backendscheduler/work/work.go
zalegrala: Good call.  This was on my mind but forgot to come back to clean up.  Should we name it `work_*.json` while we're here?

modules/backendscheduler/work/work.go
zalegrala: I was using this in tests to confirm the spread of the shards with the thought we might metric around it later.  I can remove it and add it back later if want it.

zalegrala: Hmm, let me come back to this, since we're verifying job counts in the tests I need to work that out if its not part of the interface.

modules/backendscheduler/cache.go
zalegrala: The flushing moved to `work.go` through the interface method `FlushToLocal`, so its still reasonably clean.  This way the scheduler doesn't need to know about how to handle the sharding.  Maybe later we move the backend flushing to the interface and to the Work object as well.

modules/backendscheduler/work/work.go
zalegrala: Right above is the Unlock.  I can wrap this in a local func() so the defer is more visible.

zalegrala: I've moved this into a `func` which should clarify the locking.

modules/backendscheduler/work/work.go
zalegrala: This is called pretty much always.  I added the two loops; count -> alloc -> collect.  Good callout.

modules/backendscheduler/work/work.go
zalegrala: I left it as shard*.json, but let me know if you have feelings.  Now is the time to name the file :)

modules/backendscheduler/work/work.go
joe-elliott: I'm no cpu/machine code expert, but a processor is often most efficient using its native int width. Also, depending on compiler choices, a uint8 might actually be stored in an int width to align reads from memory.

There's a lot of factors. I honestly don't know, but even if there are minor gains I would personally still just choose an int for clarity.

zalegrala: Okay cool, I'll update.  Thanks for the explanation.

zalegrala: I've pushed a commit for this.

modules/backendscheduler/work/work.go
joe-elliott: file name looks good. i was just concerned it was defined in 2 spots. like the fix 👍

Issue #5392, opened 2025-07-11
Add ability to override signature when connecting to object storage backends.


mattdurham:
By default the Minio client uses `CRC32C` some S3 compatible backends may not allow `CRC32C`. Add the ability to override the with a given allowed value. 

Will need to ensure it works with GCS,Azure and AWS S3 backends, and added to the helm chart. By default will use the `CRC32C` signature.

mattdurham:
All the minio commands have a `*ObjectOptions` that have a `AutoChecksum` that defaults to `CRC32C`. By passing in the configuration value (default CRC32C) it should resolve this.

knylander-grafana:
We'll need to update docs for this as well.

PR #5413, opened 2025-07-17
[Enhancement] put actual size for writing to backend

ruslan-mikhailov:
**What this PR does**:
It passes actual size for all writes to backend.

From minio-go library:
```
// PutObject creates an object in a bucket.
// <...>
//   - For size input as -1 PutObject does a multipart Put operation
//     until input stream reaches EOF. Maximum object size that can
//     be uploaded through this operation will be 5TiB.
//
//     WARNING: Passing down '-1' will use memory and these cannot
//     be reused for best outcomes for PutObject(), pass the size always.
//
```
As we know objects' size before write, there are no obstacles on passing the parameter.

Bonus. If I'm not mistaken, write with negative size is the only case when minio-go library passes default (CRC32C) checksum which should solve the problem https://github.com/grafana/tempo/issues/5392 for object storages that don't support CRC32C.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/usagestats/reporter_test.go
```diff
@@ -81,7 +81,7 @@ func Test_LeaderElectionWithBrokenSeedFile(t *testing.T) {
 	// Ensure that leader election succeeds even when the seed file has been
 	// corrupted.  This means that we don't need to extend the interface of the
 	// backend in order to delete a corrupted seed file.
-	err = objectClient.Write(context.Background(), backend.ClusterSeedFileName, []string{}, bytes.NewReader([]byte("{")), -1, nil)
+	err = objectClient.Write(context.Background(), backend.ClusterSeedFileName, []string{}, bytes.NewReader([]byte("{")), 1, nil)
```
javiermolinar, line 84: What's the difference here?

javiermolinar (commented):
There is a comment in the minio PutObject about multipart uploads:

> 
For size smaller than 16MiB PutObject automatically does a single atomic PUT operation.
For size larger than 16MiB PutObject automatically does a multipart upload operation.
For size input as -1 PutObject does a multipart Put operation until input stream reaches EOF. Maximum object size that can be uploaded through this operation will be 5TiB.
WARNING: Passing down ‘-1’ will use memory and these cannot be reused for best outcomes for PutObject(), pass the size always.

https://pkg.go.dev/github.com/minio/minio-go/v7@v7.0.94#Client.PutObject


Maybe the checksum problem is related to multipart uploads?

pkg/usagestats/reporter_test.go
ruslan-mikhailov: The line imitates write with partial data. I thought if I changed the file write function, it would be nice to be close to it. I don't think it's a big difference, it is just a test anyway, both functions will eventually write the same data

ruslan-mikhailov:
> Maybe the checksum problem is related to multipart uploads?

It's a bit complicated. There is decision tree what function to use for Put. In some function, it sets checksum only if TrailingHeaders is enabled:
```
	if c.trailingHeaderSupport {
		opts.AutoChecksum.SetDefault(ChecksumCRC32C)
		addAutoChecksumHeaders(&opts)
	}
```
trailingHeaderSupport is `opts.TrailingHeaders && clnt.overrideSignerType.IsV4()`, TrailingHeaders is disabled by default. That makes it skip checksum.

In some cases (for example, in case size is -1), it does not have such checks.

The issue was addressed here: https://github.com/minio/minio-go/pull/2130/files#diff-d63a6c6b6013f9e4432b05d4f81f4a7b9037fc91453bfb9fc658d6cfa412e926R448

mattdurham (approved):
Looks straightforward and more accurate!

PR #5415, opened 2025-07-17
feature: implement a listOffset by partition client

javiermolinar:
**What this PR does**:

The goal of this PR is to replace the 

kadm.ListStartOffsets and kadm.ListEndOffsets by a built-in compatible methods that accept a list of partitions. This is important to reduce the number of ListOffsets requests. Tests show a 50% reduction in requests

Screenshot 2025-07-17 at 16 21 06

Screenshot 2025-07-17 at 16 21 27

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingest/config.go
+++ b/pkg/ingest/config.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	"github.com/grafana/dskit/backoff"
 	"github.com/grafana/dskit/flagext"
 	"github.com/twmb/franz-go/pkg/kadm"
 	"github.com/twmb/franz-go/pkg/kgo"
@@ -84,6 +85,15 @@ type KafkaConfig struct {
 
 	TargetConsumerLagAtStartup time.Duration `yaml:"target_consumer_lag_at_startup"`
 	MaxConsumerLagAtStartup    time.Duration `yaml:"max_consumer_lag_at_startup"`
+
+	FetchMaxWait                      time.Duration `yaml:"fetch_max_wait"`
+	FetchConcurrencyMax               int           `yaml:"fetch_concurrency_max"`
+	UseCompressedBytesAsFetchMaxBytes bool          `yaml:"use_compressed_bytes_as_fetch_max_bytes"`
+	MaxBufferedBytes                  int           `yaml:"max_buffered_bytes"`
+
+	// The fetch backoff config to use in the concurrent fetchers (when enabled). This setting
+	// is just used to change the default backoff in tests.
+	concurrentFetchersFetchBackoffConfig backoff.Config `yaml:"-"`
 }
 
 func (cfg *KafkaConfig) RegisterFlags(f *flag.FlagSet) {
--- a/pkg/ingest/config.go
+++ b/pkg/ingest/config.go
@@ -86,8 +86,6 @@ type KafkaConfig struct {
 	TargetConsumerLagAtStartup time.Duration `yaml:"target_consumer_lag_at_startup"`
 	MaxConsumerLagAtStartup    time.Duration `yaml:"max_consumer_lag_at_startup"`
 
-	FetchConcurrencyMax int `yaml:"fetch_concurrency_max"`
-
 	// The fetch backoff config to use in the concurrent fetchers (when enabled). This setting
 	// is just used to change the default backoff in tests.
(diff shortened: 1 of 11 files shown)


pkg/ingest/config.go
```diff
@@ -84,6 +85,12 @@ type KafkaConfig struct {
 
 	TargetConsumerLagAtStartup time.Duration `yaml:"target_consumer_lag_at_startup"`
 	MaxConsumerLagAtStartup    time.Duration `yaml:"max_consumer_lag_at_startup"`
+
+	FetchConcurrencyMax int `yaml:"fetch_concurrency_max"`
```
mapno: This seems unused

pkg/ingest/partition_offset_client.go
```diff
@@ -0,0 +1,118 @@
...
+	if len(partitionIDs) == 0 {
+		return nil, nil
+	}
+
+	return p.fetchPartitionsOffsets(ctx, kafkaOffsetStart, partitionIDs)
+}
+
+// fetchPartitionsEndOffsets returns the end offset of each partition in input. This function returns
+// error if fails to get the offset of any partition (no partial result is returned).
+func (p *PartitionOffsetClient) fetchPartitionsOffsets(ctx context.Context, timestamp int64, partitionIDs []int32) (kadm.ListedOffsets, error) {
```
mapno: The mix of offset and timestamp is a bit confusing. It would be nice if we could chose one or link to where this is documented (https://github.com/twmb/franz-go/blob/master/pkg/kmsg/generated.go#L5781-L5808)

pkg/ingest/partition_offset_client.go
```diff
@@ -0,0 +1,118 @@
...
+	kafkaOffsetEnd = int64(-1)
+
+	// ReaderMetricsPrefix is the reader metrics prefix used by the ingest storage.
+	ReaderMetricsPrefix = "cortex_ingest_storage_reader"
+)
+
+// partitionOffsetClient is a client used to read partition offsets.
+type PartitionOffsetClient struct {
+	client *kgo.Client
+	admin  *kadm.Client
```
mapno: The admin client looks unused.

pkg/ingest/partition_offset_client.go
```diff
@@ -0,0 +1,118 @@
...
+
+const (
+	// kafkaOffsetStart is a special offset value that means the beginning of the partition.
+	kafkaOffsetStart = int64(-2)
+
+	// kafkaOffsetEnd is a special offset value that means the end of the partition.
+	kafkaOffsetEnd = int64(-1)
+
+	// ReaderMetricsPrefix is the reader metrics prefix used by the ingest storage.
+	ReaderMetricsPrefix = "cortex_ingest_storage_reader"
```
mapno:
```suggestion
```

pkg/ingest/partition_offset_client.go
javiermolinar: good call, we can do both

pkg/ingest/config.go
javiermolinar: It is used here https://github.com/grafana/tempo/pull/5415/files#diff-00affafa4a8638d736d38d765591201b3c1150440840d6418b3edda28d7c1e16R197, but removing it doesn't affect the test outcome, so I will remove it

pkg/ingest/metrics.go
```diff
@@ -43,7 +43,7 @@ var (
 // Call ResetLagMetricsForRevokedPartitions when partitions are revoked to prevent exporting
 // stale data. For efficiency this is not detected automatically from changes inthe assigned
 // partition callback.
-func ExportPartitionLagMetrics(ctx context.Context, admClient *kadm.Client, log log.Logger, cfg Config, getAssignedActivePartitions func() []int32, forceMetadataRefresh func()) {
+func ExportPartitionLagMetrics(ctx context.Context, admClient *kadm.Client, partitionClient *PartitionOffsetClient, log log.Logger, cfg Config, getAssignedActivePartitions func() []int32, forceMetadataRefresh func()) {
```
mdisibio: Is there a benefit to making callers pass the partitionClient?  I think we could replace both client params with just one `client kgo.Client` and create the admin and partition clients internally.

javiermolinar: Sure, sounds good!

mapno (approved):

Issue #5395, opened 2025-07-11
Trace IDs with leading zeros are trimmed in the Search API response (non compliant with w3c or Otel)


sblack79:
**Describe the bug**
Resurrecting [this issue](https://github.com/grafana/tempo/issues/4723) marked stale 

When a trace with leading zeros in the trace ID (ex: 008efff798038103d269b633813fc703) is ingested to Tempo, the trace ID in the search API results is trimmed to omit the leading zeros.

The [TraceIDToHexString](https://github.com/grafana/tempo/blob/main/pkg/util/traceid.go#L17) function is being used to generate the string representation for the trace ID and this removes the leading zeros in the trace ID. It looks like this function is being used in a few places including:

1. the [parquet module to generate TraceID Text](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet4/schema.go#L356) (the trace ID bytestring in the corresponding column in parquet contains the leading zeros)
2. the [traceql engine module](https://github.com/grafana/tempo/blob/main/pkg/traceql/engine.go#L266) to generate the trace search metadata.

We have observed that for trace IDs with leading zeros, the search API response includes the trimmed trace IDs instead of the full trace ID (16 byte hexstring). The trace ID API succeeds with either the trimmed trace ID or the zero prefixed trace IDs.

Because of the trimming of leading zeros, it's not straightforward to correlate trace IDs with external sources as the API consumer has to left pad the trimmed IDs to make them compatible with the [trace context spec](https://www.w3.org/TR/trace-context/#trace-id).

A similar [fix](https://github.com/grafana/tempo/issues/2060) was rolled out for span IDs.

**Why this is important** 

The W3C spec [mandates compliant TraceIds be padded with zeroes](https://www.w3.org/TR/trace-context/#interoperating-with-existing-systems-which-use-shorter-ide)   

>  **_When a system creates an outbound message and needs to generate a fully compliant 16 bytes trace-id from a shorter identifier, it SHOULD left pad the original identifier with zeroes. For example, the identifier 53ce929d0e0e4736, SHOULD be converted to trace-id value 000000000000000053ce929d0e0e4736._** 

The Opentelemetry Spec [functionally mandates the same](https://opentelemetry.io/docs/specs/otel/trace/api/#retrieving-the-traceid-and-spanid)

> **The API MUST allow retrieving the TraceId and SpanId in the following forms:  
[...]

PR #5416, opened 2025-07-17
Pad Zeroes to the left of a traceId

sblack79:
**What this PR does**:

Pads 0s to the left of a string formatted trace id according to W3C standards.

I didn't see any particular existing documentation regarding the behavior of leading 0s in Tempo so I didn't add any here. 

Addresses this issue: https://github.com/grafana/tempo/issues/5395

**Checklist**
- [x] Tests updated
- [x] Documentation added


sblack79:
I believe the test should succeed now if I could get a workflow approval @mdisibio @zalegrala @ie-pham @carles-grafana @joe-elliott

sblack79:
Fixed a linting error, could I get another workflow approval @mdisibio @zalegrala @ie-pham @carles-grafana @joe-elliott

joe-elliott:
Thank you for the work, but we are discussing internally whether or not we want this PR. We are unsure if it would break anything in Grafana. Also 64 bit traces ids feel like they'd be significantly harder to read prefixed with so many 0s.

sblack79:
> Thank you for the work, but we are discussing internally whether or not we want this PR. We are unsure if it would break anything in Grafana. Also 64 bit traces ids feel like they'd be significantly harder to read prefixed with so many 0s.

Gotcha on the Grafana end I realize this is a pretty broad change so it could have larger scale ramifications and that testing/confirmation could take a while.

Regarding if it should be done (and let me know if I should save this discussion for the issue itself) I would say that keeping in line with Open source standards is very important for a service like Tempo and I know from personal experience we've already run into quite a few instances of this representation being assumed to be one way in Tempo and one way by another service following OpenTelemetry assumptions. A very common one is a trace not matching to a log or vice versa because the log representation is assuming there are 0's on the end of the trace id and Tempo is not. I know this issue has impacted other users (though not sure how broadly) by the interest this issue has garnered in the past and I tend to think that being able to make assumptions more broadly about a system is very important for preventing friction over time. And given that these are the baseline assumptions both W3C and OpenTelemetry are making now I think the problems posed by not following this representation will only grow with time. 

I think the UI notion makes sense to consider but it does seem to me like something that could be handled on the Grafana end when displaying the data as an option.

joe-elliott:
> A very common one is a trace not matching to a log or vice versa because the log representation is assuming there are 0's on the end of the trace id and Tempo is not. 

Can you give more details here. I think (from our side) this is what we are struggling to see. We are concerned that if we add the 0s we're going to break an unknown number of users and processes. Understanding what is not working now due to sending the trimmed trace id string would help.

sblack79:
> Can you give more details here. I think (from our side) this is what we are struggling to see. We are concerned that if we add the 0s we're going to break an unknown number of users and processes. Understanding what is not working now due to sending the trimmed trace id string would help.

Formerly it was both ways (this has since been fixed) but for our logging system regex's will assume (based on OTel standards and how we display trace ids as a string representation) that the start of a trace in a log correlates to the start of one we're searching for. So for example if you use `^` to begin your regex match it'll fail if your trace id is "01" but the one you're trying to match on is "1" or more commonly you search for "01" and the only existing trace ids from Tempo have "1" (since fixed on the Tempo end but certain services log out the data received from Tempo). We also have certain user services that pull in trace data via Tempo and attempt to make correlations with their logging data that run into this same issue. 

I understand that this change will probably adversely impact some users in the same way we are experiencing now, but that is borne from the current representation not aligning with the existing standards and I think that should take precedence to avoid this kind of situation in the future. 

These sorts of issues are only going to become more frequent with time as more systems assume interoperability with OTel and make these sorts of assumptions on the basis that data is OTel compatible.

joe-elliott:
As [mentioned in the issue](https://github.com/grafana/tempo/issues/5395#issuecomment-3109041911) we will take this PR if we add it as a config option and leave the default behavior alone.

PR #5430, opened 2025-07-21
New live-store

mapno:
**What this PR does**:

The live-store is a Kafka-based ingestion service that consumes trace data from Kafka topics instead of receiving it via gRPC like the traditional ingester. It reads serialized traces from Kafka, processes them through the same data pipeline as the ingester (live traces → head blocks → WAL → complete blocks).

```
                      ┌─────────────────────────────┐
                      │        LiveStore            │
                      │ • Manages instances map     │
                      │ • Shared WAL coordination   │
                      │ • Flush queue management    │
                      └─────────────┬───────────────┘
                                    │
                      ┌─────────────┴───────────────┐
          ┌─────────────────────┐       ┌─────────────────────┐
          │  PartitionReader    │       │   instances map     │
          │                     │       │   [tenantID]        │
          │ • Kafka consumer    │       │                     │
          │ • Watermark commits │       │ ┌─────────────────┐ │
          │ • Routes to consume │       │ |   instance A    │ │
          │   function          │       │ │   (tenant A)    │ │
          └─────────────────────┘       │ │                 │ │
                                        │ │ • Live traces   │ │
                                        │ │ • Head block    │ │
                                        │ │ • WAL blocks    │ │
                                        │ │ • Complete      │ │
                                        │ │   blocks        │ │
                                        │ └─────────────────┘ │
                                        │                     │
                                        │ ┌─────────────────┐ │
                                        │ │   instance B    │ │
                                        │ │   (tenant B)    │ │
                                        │ │                 │ │
                                        │ │ • Live traces   │ │
                                        │ │ • Head block    │ │
                                        │ │ • WAL blocks    │ │
                                        │ │ • Complete      │ │
                                        │ │   blocks        │ │
                                        │ └─────────────────┘ │
                                        └─────────────────────┘
                                                   │
                                                   ▼
                                        ┌─────────────────────┐
                                        │     Shared WAL      │
                                        │                     │
                                        │ • Single WAL for    │
                                        │   all instances     │
                                        │ • Coordinated       │
                                        │   block cutting     │
                                        └─────────────────────┘
```

Contributes to https://github.com/grafana/tempo-squad/issues/734

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/livestore/live_store.go
+++ b/modules/livestore/live_store.go
@@ -228,7 +228,6 @@ func (s *LiveStore) flushRemaining() {
 			return // shutdown timeout reached
 		}
 	}
-
 }
 
 func (s *LiveStore) consume(_ context.Context, rs []record) error {
--- a/modules/livestore/partition_reader.go
+++ b/modules/livestore/partition_reader.go
@@ -95,7 +95,7 @@ func (r *PartitionReader) running(ctx context.Context) error {
 	go r.commitLoop(ctx)
 
 	for ctx.Err() == nil {
-		fetches := r.client.PollFetches(ctx)
+		fetches := r.client.PollFetches(consumeCtx)
 		if fetches.Err() != nil {
 			if errors.Is(fetches.Err(), context.Canceled) {
 				return nil


mapno:
From https://github.com/grafana/tempo/pull/5330

> In general does it make sense to move responsibility of timing and actions from the Bufferer to the individual instances? This would make the instances more self contained and potentially avoid a good chunk of the locking. Would be harder to do global cuts but we could do tenant specific then? Or trigger via a channel with at timer in the Bufferer.

This approach has been since dropped in favor of periodic global committing.

>From chatting would calling Bufferer ReadCache make more sense?

It's been since changed to `LiveStore`. I think that capture bests the responsibility of this new component. But I'm happy to consider others.

My concern with `ReadCache` is that it might be mistaken with the other caches (memcached) that we run.

>Can we document the failure modes and what happens? Do we need 100% data fidelity or can 99.9%, since the data will be correctly handled by long term storage regardless.

That's not currently covered in this PR. What if we defined the requirements for this component instead?

>Assuming failures happen is it better to return incorrect data or no data?

We've opted for a RF=2, Quorum=1 design, which favors failing fast and returning no data.

modules/livestore/live_store.go
```diff
@@ -0,0 +1,421 @@
...
+	// Stop consuming
+	err := services.StopAndAwaitTerminated(context.Background(), s.reader)
+	if err != nil {
+		level.Warn(s.logger).Log("msg", "failed to stop reader", "err", err)
+		return err
+	}
+
+	// Cancel and wait for async loops to return
+	s.cancel()
+	s.flushqueues.Close()
```
mattdurham, line 205: Should this be `DiscardAndClose`? The `Dequeue` call in the `completeLoop` looks like it would still try consuming all the items from the queue which could ultimately exceed the grace period shutdown and prevent a clean teardown.

modules/livestore/instance.go
```diff
@@ -0,0 +1,333 @@
...
+		return nil, err
+	}
+
+	return i, nil
+}
+
+func (i *instance) pushBytes(ts time.Time, req *tempopb.PushBytesRequest) {
+	// For each pre-marshalled trace, we need to unmarshal it and push to live traces
+	for j, traceBytes := range req.Traces {
+		if j >= len(req.Ids) {
```
mattdurham: Can this be moved outside this loop? `if len(req.Traces) != len(req.Ids)` or even higher in the chain?

modules/livestore/instance.go
```diff
@@ -0,0 +1,333 @@
...
+	writer := backend.NewWriter(i.wal.LocalBackend())
+
+	iter, err := walBlock.Iterator()
+	if err != nil {
+		level.Error(i.logger).Log("msg", "failed to get WAL block iterator", "id", id, "err", err)
+		return err
+	}
+	defer iter.Close()
+
+	newMeta, err := i.enc.CreateBlock(context.Background(), &common.BlockConfig{}, walBlock.BlockMeta(), iter, reader, writer)
```
mattdurham: Should this use the context from `s.ctx` from the `completeLoop` func?

mattdurham: And maybe on NewLocalBlock below?

modules/livestore/live_store.go
```diff
@@ -0,0 +1,421 @@
...
+				return
+			}
+			err = inst.completeBlock(op.blockID)
+			if err != nil {
+				level.Error(s.logger).Log("msg", "failed to complete block", "tenant", op.tenantID, "block", op.blockID, "err", err)
+
+				delay := op.backoff()
+				op.at = time.Now().Add(delay)
+
+				go func() {
```
mattdurham: Am not sure I fully understand this go func. Looks like it will try to wrap around but also try to enqueue a message at the same time? Is this possibly an infinite loop? I dont see anything in the backoff about a failsafe or way to handle poisoned message.

mattdurham: Sorry, I see the maxFlushAttempts now.

cmd/tempo/app/config.go
```diff
@@ -144,6 +146,7 @@ func (c *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet) {
 	c.CacheProvider.RegisterFlagsAndApplyDefaults(util.PrefixConfig(prefix, "cache"), f)
 	c.BackendScheduler.RegisterFlagsAndApplyDefaults(util.PrefixConfig(prefix, "backend-scheduler"), f)
 	c.BackendWorker.RegisterFlagsAndApplyDefaults(util.PrefixConfig(prefix, "backend-worker"), f)
+	c.LiveStore.RegisterFlagsAndApplyDefaults(util.PrefixConfig(prefix, "liveStore"), f)
```
mattdurham: Should this be `live-store` for consistency?

modules/livestore/live_store.go
```diff
@@ -0,0 +1,421 @@
...
+
+	// Flush all data to disk
+	s.flushRemaining()
+
+	return nil
+}
+
+func (s *LiveStore) flushRemaining() {
+	s.cutAllInstancesToWal()
+	for s.flushqueues.Length() > 0 {
```
mattdurham, line 223: Should this also take into account a context or max amount to wait?

modules/livestore/partition_reader.go
```diff
@@ -0,0 +1,307 @@
...
+	r.Service = services.NewBasicService(r.start, r.running, r.stop)
+	return r, nil
+}
+
+func (r *PartitionReader) start(context.Context) error {
+	return nil
+}
+
+func (r *PartitionReader) running(ctx context.Context) error {
+	consumeCtx, cancel := context.WithCancel(context.Background())
```
mattdurham, line 83: Should this use the passed in context and not background? Background will mess with the retry logic in the fetchLastCommittedOffsetWithRetries

modules/livestore/partition_reader.go
```diff
@@ -0,0 +1,307 @@
...
+	defer cancel()
+
+	offset, err := r.fetchLastCommittedOffsetWithRetries(consumeCtx)
+	if err != nil {
+		return fmt.Errorf("failed to fetch last committed offset: %w", err)
+	}
+	r.client.AddConsumePartitions(map[string]map[int32]kgo.Offset{r.topic: {r.partitionID: offset}})
```
mattdurham, line 90: I hate this map in map but looks like its what the library expects.

modules/livestore/partition_reader.go
```diff
@@ -0,0 +1,307 @@
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			// Final commit on shutdown
+			r.commitCurrentWatermark(lastCommittedOffset)
+			return
+		case <-ticker.C:
+			// Periodic commit
+			lastCommittedOffset = r.commitCurrentWatermark(lastCommittedOffset)
```
mattdurham, line 242: How do the offsets interact? Since we only sync the watermark every X time, that there is the possibility for replay in regards for consuming more records. But we dont use the offset in the records structure in the live store consume function and there doesnt appear to be an logic to check for duplicate entries?

modules/ingester/ingester.go
```diff
@@ -151,6 +151,9 @@ func New(cfg Config, store storage.Store, overrides overrides.Interface, reg pro
 
 	i.subservicesWatcher = services.NewFailureWatcher()
 	i.subservicesWatcher.WatchService(i.lifecycler)
+	if ingestCfg := cfg.IngestStorageConfig; ingestCfg.Enabled {
```
ruslan-mikhailov: (very minor, matter of taste) Why not `if cfg.IngestStorageConfig.Enabled` for simplicity?

modules/livestore/instance.go
```diff
@@ -0,0 +1,333 @@
...
+		// Push each batch in the trace to live traces
+		for _, batch := range trace.ResourceSpans {
+			if len(batch.ScopeSpans) == 0 || len(batch.ScopeSpans[0].Spans) == 0 {
+				continue
+			}
+
+			// Apply tenant limits
+			maxTracesLimit := uint64(maxTraces)
+			if maxTraces <= 0 {
+				maxTracesLimit = 10000 // Default limit
```
ruslan-mikhailov: Why don't we cap it in MaxLocalTracesPerUser() 🤔?  Anyway, I suggest moving `10000` to global constants

modules/livestore/instance.go
```diff
@@ -0,0 +1,333 @@
...
+
+	i.blocksMtx.Lock()
+	defer i.blocksMtx.Unlock()
+	if i.headBlock != nil {
+		err := i.headBlock.Flush()
+		if err != nil {
+			return err
+		}
+
+		// Update watermark after successful flush, passing the specific trace IDs
```
ruslan-mikhailov: To be honest, I did not fully understand what it means

modules/livestore/partition_reader.go
```diff
@@ -0,0 +1,307 @@
...
+	return lastCommittedOffset
+}
+
+func (r *PartitionReader) commitOffset(ctx context.Context, offset int64) error {
+	// Use the admin client to commit the offset
+	offsets := make(kadm.Offsets)
+	offsets.Add(kadm.Offset{
+		Topic:     r.topic,
+		Partition: r.partitionID,
+		At:        offset + 1,
```
ruslan-mikhailov, line 272: Why do you add 1 to the offset? It means, you're moving the offset to the next record after last processed message, right?

mapno: That's a Kafka quirk:

> It's important to note that the committed offset refers to the next offset the consumer intends to read

Source: https://www.confluent.io/blog/guide-to-consumer-offsets/

modules/livestore/live_store.go
mapno: Discussed internally: we're leaving it this way for now to simplify shutdown and replay—it tries to complete all work before stopping.

modules/livestore/instance.go
mapno: Yes.

modules/livestore/instance.go
mapno: True. I've fixed a couple of bad context propagations.

cmd/tempo/app/config.go
mapno: Yes

modules/livestore/live_store.go
mapno: Changed it to wait max 30s while trying to complete all pending work, after it quits.

modules/livestore/partition_reader.go
mapno: I preferred to not abruptly cancel consumption if possible and quit when a full cycle is complete, so it needs to be detached.

modules/livestore/partition_reader.go
mapno: Yes

modules/livestore/partition_reader.go
mapno: Added a comment mentioning the tradeoffs of this approach

modules/ingester/ingester.go
mapno: Yes, it was a silly choice.

modules/livestore/instance.go
mapno: Removed the limit altogether like in the block-builder

modules/livestore/instance.go
mapno: Leftover comment from a past implementation, removed.

PR #5440, opened 2025-07-25
[Rhythm] Scheduler: ignore blocks with nocompact flag

ruslan-mikhailov:
**What this PR does**: ignore blocks with nocompact flag in backend scheduler

Fixes https://github.com/grafana/tempo/pull/5390#discussion_r2228928185

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/backendscheduler/backendscheduler.go
+++ b/modules/backendscheduler/backendscheduler.go
@@ -108,7 +108,7 @@ func (s *BackendScheduler) starting(ctx context.Context) error {
 	level.Info(log.Logger).Log("msg", "backend scheduler starting")
 
 	if s.cfg.Poll {
-		s.store.EnablePolling(ctx, blocklist.OwnsNothingSharder, false)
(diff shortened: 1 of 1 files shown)


mapno (approved):

modules/backendscheduler/backendscheduler.go
```diff
@@ -108,7 +108,7 @@ func (s *BackendScheduler) starting(ctx context.Context) error {
 	level.Info(log.Logger).Log("msg", "backend scheduler starting")
 
 	if s.cfg.Poll {
-		s.store.EnablePolling(ctx, blocklist.OwnsNothingSharder, false)
```
zalegrala, line 111: It might be worth some test coverage for the cases where we want to skip or include the detection of the block.

ruslan-mikhailov: Discussed that it does not block merging and will be resolved separately.

PR #5441, opened 2025-07-25
Step validation

ruslan-mikhailov:
**What this PR does**:
It checks if step is too small and generates too many time series

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/metrics_query_range_handler.go
+++ b/modules/frontend/metrics_query_range_handler.go
@@ -34,6 +34,9 @@ func newQueryRangeStreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTripp
 		if req.Step == 0 {
 			req.Step = traceql.DefaultQueryRangeStep(req.Start, req.End)
 		}
+		if err := validateQueryRangeReq(req); err != nil {
+			return err
+		}
 		traceql.AlignRequest(req)
 
 		httpReq := api.BuildQueryRangeRequest(&http.Request{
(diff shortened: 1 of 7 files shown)


modules/frontend/metrics_query_range_handler.go
```diff
@@ -172,3 +174,21 @@ func logQueryRangeRequest(logger log.Logger, tenantID string, req *tempopb.Query
...
+		Body:       io.NopCloser(strings.NewReader(err.Error())),
+	}
+}
+
+func validateQueryRangeReq(req *tempopb.QueryRangeRequest) error {
+	if req.Start > req.End {
+		return errors.New("end must be greater than start")
+	}
+	if (req.End-req.Start)/req.Step > maxInterval {
+		return errors.New("step is too granular")
```
mdisibio: Since this is user facing, let's include more information.  Maybe something like:  "step of 1s is too small. The minimum step for query range of 36h is 1m"    Should be straightforward to do this using time.Durations.

modules/frontend/metrics_query_range_handler.go
```diff
@@ -20,6 +21,8 @@ import (
 	"github.com/grafana/tempo/pkg/traceql"
 )
 
+const maxInterval = 10_000
```
mdisibio: This probably needs to be a config option.

ruslan-mikhailov:
\+ review fixes

modules/frontend/metrics_query_range_handler.go
ruslan-mikhailov: \+

Example of an error: `step of 35ms is too small, minimum step for given range is 36ms`

modules/frontend/metrics_query_range_handler.go
ruslan-mikhailov: \+

Added config + docs for it

ruslan-mikhailov:

PR #5444, opened 2025-07-25
Move cache handling to Work

zalegrala:
**What this PR does**:

Here we move the cache handling to the Work struct.  Reviewing the current code, it seems that ever time we make an update to the job, the cache is flushed, but the two events are separated.  This change unifies the two events, and gives authority for the cache handling to the Work struct, which should improve clarity in the code structure.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/backendscheduler/config.go
+++ b/modules/backendscheduler/config.go
@@ -19,15 +19,12 @@ type Config struct {
 	// Provider configs
 	ProviderConfig provider.Config `yaml:"provider"`
 	JobTimeout     time.Duration   `yaml:"job_timeout"`
-	// LocalWorkPath  string          `yaml:"local_work_path,omitempty"` // Path to store local work cache
 }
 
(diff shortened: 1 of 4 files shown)


modules/backendscheduler/backendscheduler.go
```diff
@@ -188,7 +188,7 @@ func (s *BackendScheduler) running(ctx context.Context) error {
 }
 
 func (s *BackendScheduler) stopping(_ error) error {
-	err := s.work.FlushToLocal(context.Background(), s.cfg.LocalWorkPath, nil) // flush all shards
+	err := s.work.FlushToLocal(context.Background(), nil) // flush all shards
 	if err != nil {
 		return fmt.Errorf("failed to flush work cache on shutdown: %w", err)
```
zalegrala: Consider continuing to attempt a flush to the backend.

knylander-grafana:
@zalegrala Does this need an entry in the changelog?

PR #4916, opened 2025-03-28
feat(servicegraphs): Add support for db.system.name

iamrajiv:
**What this PR does**:

This PR adds support for the `db.system.name` attribute introduced in OpenTelemetry Semantic Conventions v1.30.0. The change maintains backward compatibility while supporting the new attribute name.

**Which issue(s) this PR fixes**:
Fixes #4640

**Checklist**

- [x] Tests updated
- [x] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #4656, opened 2025-02-03
feat(servicegraphs): Add support for `db.system.name`

iamrajiv:
**What this PR does**:

This PR adds support for the `db.system.name` attribute introduced in OpenTelemetry Semantic Conventions v1.30.0. The change maintains backward compatibility while supporting the new attribute name.

**Which issue(s) this PR fixes**:
Fixes #4640

**Checklist**

- [x] Tests updated
- [x] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #5445, opened 2025-07-26
feat(servicegraphs): Add support for db.system.name

### LINKED ISSUE #4640: Breaking changes to `db.system` attribute in OpenTelemetry Semantic Conventions v1.30.0

[v1.30.0 of OpenTelemetry Semantic Conventions has two breaking changes](https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.30.0) to the `db.system` attribute. This will impact the ability for the metrics-generator to create virtual nodes in service graphs.

The breaking changes are:

* Attribute renamed from `db.system` to `db.system.name`
* List of constants for recognized databases updated

`db.system.name` recognized as a peer attribute for service graph metrics generation.

**Additional context**

[Diff between releases v1.29.0 and v1.30.0](https://github.com/open-telemetry/semantic-conventions/compare/v1.29.0...v1.30.0#diff-8aa7bbea8c8d64909c5d9abdff0ed7f4841b7a47d650741f5da348d88d50374a)

iamrajiv:
**What this PR does**:

This PR adds support for the `db.system.name` attribute introduced in OpenTelemetry Semantic Conventions v1.30.0. The change maintains backward compatibility while supporting the new attribute name.

Fixes #4640

**Checklist**

- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/configuration/_index.md
+++ b/docs/sources/tempo/configuration/_index.md
@@ -459,6 +459,10 @@ metrics_generator:
             # resouce and span level attributes
             [span_multiplier_key: <string> | default = ""]
 
+            # database name attributes is the attribute name list used to identify the databasename from span attributes
+            # the  deafault value is ["db.name", "db.system.name"]
+            [database_name_attributes: <list of string> | default = ["db.name", "db.system.name"]]
+
             # Enables additional labels for services and virtual nodes.
             [enable_virtual_node_label: <bool> | default = false]
 
@@ -1789,6 +1793,7 @@ overrides:
           [histogram_buckets: <list of float>]
           [dimensions: <list of string>]
           [peer_attributes: <list of string>]
+          [database_name_attributes: <list of string>]
           [enable_client_server_prefix: <bool>]
           [enable_messaging_system_latency_histogram: <bool>]
 
--- a/modules/generator/processor/servicegraphs/config.go
+++ b/modules/generator/processor/servicegraphs/config.go
@@ -47,6 +47,10 @@ type Config struct {
 	// If enabled attribute value will be used for metric calculation
 	SpanMultiplierKey string `yaml:"span_multiplier_key"`
 
+	// DatabaseNameAttributes is the attribute name list used to identify the database name from span attributes
+	// The default value is {"db.name", "db.system.name"}
+	DatabaseNameAttributes []string `yaml:"database_name_attributes"`
+
 	// EnableVirtualNodeLabel enables additional labels for uninstrumented services
 	EnableVirtualNodeLabel bool `yaml:"enable_virtual_node_label"`
 }
@@ -64,6 +68,9 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
 		peerAttr = append(peerAttr, string(attr))
 	}
 	cfg.PeerAttributes = peerAttr
+
+	// Set default database name attributes to support both old and new semantic conventions
+	cfg.DatabaseNameAttributes = []string{"db.name", "db.system.name"}
 
 	cfg.EnableMessagingSystemLatencyHistogram = false
 }
--- a/modules/generator/processor/servicegraphs/servicegraphs.go
+++ b/modules/generator/processor/servicegraphs/servicegraphs.go
@@ -270,14 +270,14 @@ func (p *Processor) upsertPeerNode(e *store.Edge, spanAttr []*v1_common.KeyValue
 }
 
 // upsertDatabaseRequest handles the logic of adding a database edge on the
-// graph.  If we have a db.name or db.system attribute, we assume this is a
+// graph.  If we have any of the configured database name attributes, we assume this is a
 // database request.  The name of the edge is determined by the following
 // order:
 //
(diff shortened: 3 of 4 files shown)


iamrajiv:
hi @joeelliott @zalegrala @rlankfo thanks for reviewing the earlier prs (#4656 #4774 and #4916) on this issue sorry for the trouble caused so i created this new pr i made the changes that were asked in the previous ones let me know if this looks good i ran the tests i added but didnt check coverage not sure if theres any coverage criteria

iamrajiv:
btw while digging through the code i noticed some hardcoded attribute checks that could cause trouble later maybe
like

- service name detection only checks for servicename but what if otel adds serviceinstancename or something else in future
- server address resolution seems fixed too

maybe we should refactor these to use something like what we did here?

cc: @joeelliott @zalegrala @rlankfo

iamrajiv:
btw while digging through the code i noticed some hardcoded attribute checks that could cause trouble later maybe
like

- service name detection only checks for servicename but what if otel adds serviceinstancename or something else in future
- server address resolution (same issue as above)

maybe we should refactor these to use something like what we did here?

cc: @joeelliott @zalegrala @rlankfo

modules/generator/processor/servicegraphs/config.go
```diff
@@ -65,5 +69,8 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
 	}
 	cfg.PeerAttributes = peerAttr
 
+	// Set default database name attributes to support both old and new semantic conventions
+	cfg.DatabaseNameAttributes = []string{"db.name", "db.system.name"}
```
zalegrala: These strings should refer to upstream constants.  Additionally, I believe this is a breaking change in the order of which keys are consulted, and should be mentioned in the CHANGELOG entry and probably an update to the documentation.

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -286,20 +286,35 @@ func (p *Processor) upsertDatabaseRequest(e *store.Edge, resourceAttr []*v1_comm
...
+				dbName = name
+			}
+
+			isDatabase = true
+			break
+		}
 	}
 
-	// Check for db.system only if we don't have db.name above
 	if !isDatabase {
```
zalegrala, line 303: Can we collapse this logic into the loop above?

iamrajiv:
thanks for the review @zalegrala i fixed the if else block and also updated the config since `DBSystemNameKey` is not available in semconv v1.25.0 i kept it as a string literal with a comment explaining why and updated the changelog too can you review again?

docs/sources/tempo/configuration/_index.md
```diff
@@ -458,6 +458,11 @@ metrics_generator:
             # resouce and span level attributes
             [span_multiplier_key: <string> | default = ""]
 
+            # database name attributes is the attribute name list used to identify the database name from span attributes
+            # attributes are searched in the order they are provided - the first matching attribute will be used
+            # the default value is ["db.name", "db.system.name"]
```
knylander-grafana:
```suggestion
            # database_name_attributes is the attribute name list used to identify the database name from span.
            # attributes. Attributes are searched in the order they are provided. The first matching attribute is used.
            # The default value is ["db.name", "db.system.name"]
```

iamrajiv:
thanks @knylander-grafana i did the fixes

modules/generator/processor/servicegraphs/config.go
```diff
@@ -65,5 +70,9 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
 	}
 	cfg.PeerAttributes = peerAttr
 
+	// Set default database name attributes to support both old and new semantic conventions
+	// Note: "db.system.name" is not yet available as a constant in semconv v1.25.0
```
zalegrala: Thanks for the update here.  My suggestion was to update the version we import at the top of this file, so that we can refer to the upstream constant.

zalegrala:
I believe the docs [here](https://github.com/zalegrala/tempo/blob/97a244012fb84feaf2b75eb9e9b2811ddf57292b/docs/sources/tempo/metrics-from-traces/service_graphs/_index.md#L63-L64) should also get double checked for accuracy with this change.

modules/generator/processor/servicegraphs/config.go
iamrajiv: thanks for the feedback @zalegrala so i checked and updating to a newer semconv version to use DBSystemNameKey so DBSystemNameKey constant is available in v1.34.0 but this version has breaking changes where DBNameKey used elsewhere in the code was replaced with DBNamespaceKey as part of the semantic conventions evolution dbname → dbnamespace shoud i do it?

zalegrala: I think the change you are targeting is against 1.30 if I see correctly, so I don't think we need to go to 1.34.0 for the breaking change.  Perhaps a follow up PR could include just the breaking change if we need it.

iamrajiv:
thanks for review @zalegrala i did the changes can you check again?

PR #5457, opened 2025-07-29
[Bugfix] Default step for grpc

ruslan-mikhailov:
**What this PR does**: Align request in grpc query range call only when default step is applied.

The bug was introduced in https://github.com/grafana/tempo/pull/5366 in commit https://github.com/grafana/tempo/pull/5366/commits/c26423c7a824d45affd96b583f2b3b9c2f8de4cf

**How it has been tested:**
1. http in e2e tests
2. Because we don't have e2e tests for it, grpc was checked manually, by calling MetricsQueryRange grpc method with step=0.
Before the fix (main branch):
```
StatusCode.UNKNOWN, start required
```

After the fix: returned 301 intervals, no errors

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/integration/e2e/api/query_range_test.go
+++ b/integration/e2e/api/query_range_test.go
@@ -34,9 +34,13 @@ type queryRangeRequest struct {
 	End       time.Time `json:"end"`       // default: now + 1m
 	Step      string    `json:"step"`      // default: 5s
 	Exemplars int       `json:"exemplars"` // default: 100
+	noDefault bool      `json:"-"`         // if true, SetDefaults() will not set defaults`
 }
 
 func (r *queryRangeRequest) SetDefaults() {
+	if r.noDefault {
+		return
+	}
 	if r.Start.IsZero() {
 		r.Start = time.Now().Add(-5 * time.Minute)
 	}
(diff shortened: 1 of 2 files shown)


ruslan-mikhailov:
I did not add changelog, because the bug was introduced very recently on a change that was not yet released, therefore from changelog point of view, nothing's changed from the last version

integration/e2e/api/query_range_test.go
```diff
@@ -278,12 +282,26 @@ sendLoop:
...
-		req := queryRangeRequest{Query: "{. a}"}
-		req.SetDefaults()
-		res := doRequest(t, tempo.Endpoint(tempoPort), "api/metrics/query_range", queryRangeRequest{Query: "{. a}"})
-		require.Equal(t, 400, res.StatusCode)
-	})
+	for name, req := range map[string]queryRangeRequest{
+		"invalid query": {
+			Query: "{. a}",
+		},
+		"step=0 (default step)": {
```
joe-elliott, line 289: this test seems to be calling tempo with a step of 0 and expecting a 400 back, but from the code below it looks like we choose a default step if none is provided?

I'm probably just misunderstanding something.

https://github.com/grafana/tempo/pull/5457/files#diff-63daee0e74e34bd236730fe0ca7066249b40db452f70ed128c4d883540b53751R33-R36

ruslan-mikhailov: The code you're pointing is from grpc, it accepts numbers. For http, it parses string here: https://github.com/ruslan-mikhailov/tempo/blob/main/pkg/api/http.go#L592
"0" as well as "0s" != "" -> default can not be applied -> only when you send in query `?step=` or do not send it at all.

This test covers http. Although, the fix is for grpc I added the test to state such behaviours and check nothing crazy happens.

joe-elliott:
+1 to changelog decision

PR #5458, opened 2025-07-29
Implement poller hook on compaction provider for accurate delay

zalegrala:
**What this PR does**:

This change addresses two issues.

First, on startup, the scheduler hands out jobs on a potentially out of date blocklist, which means the scheduler possibly hands out jobs for blocks which don't exist.

Second, when we cycle for the min period, we only sleep the period, which is only useful if we expect new blocks to show up on this interval.  In reality, it will cycle here several times until the blocklist is polled.

In both cases, I think we want to know the moment a blocklist is updated, so that we can a) start the compaction provider and hand out accurate jobs and b) sleep until new work shows up once we've drained all tenants.

An implicit assumption here is that the poller on the worker is on the same interval as the poller on the scheduler.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/tempodb.go
+++ b/tempodb/tempodb.go
@@ -105,8 +105,10 @@ type Reader interface {
 	// PollNow does an immediate poll of the blocklist and is for testing purposes. Must have already called EnablePolling.
 	PollNow(ctx context.Context)
 
-	// Receive a context which is closed when the blocklist has been updated.
-	PollNotification(ctx context.Context) context.Context
+	// PollNotification returns a channel that will be closed when the blocklist
+	// is updated.  The received context is used as the parent, to avoid
+	// deadlocks.
+	PollNotification(ctx context.Context) <-chan struct{}
 
 	Shutdown()
 }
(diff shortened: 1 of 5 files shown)


tempodb/tempodb.go
```diff
@@ -651,6 +658,17 @@ func (rw *readerWriter) PollNow(ctx context.Context) {
 	rw.pollBlocklist(ctx)
 }
 
+func (rw *readerWriter) PollNotification(ctx context.Context) context.Context {
+	ctx, cancel := context.WithCancel(ctx)
```
zalegrala, line 668: The received context may not be useful, and the interface could drop this argument.

zalegrala: The received context here is required to avoid deadlock, since the parent context cancellation will cancel the children, any blocking calls waiting on the channel closure are only properly handled with this in place.

tempodb/tempodb.go
```diff
@@ -651,6 +661,17 @@ func (rw *readerWriter) PollNow(ctx context.Context) {
...
 
+func (rw *readerWriter) PollNotification(ctx context.Context) <-chan struct{} {
+	ctx, cancel := context.WithCancel(ctx)
+
+	rw.pollerNotificationLock.Lock()
+	defer rw.pollerNotificationLock.Unlock()
+
+	rw.pollerNotificationFuncs = append(rw.pollerNotificationFuncs, cancel)
+
+	return ctx.Done()
```
mdisibio, line 675: This is clever. I expected something to be sent on the channel at first.  Function comment would be good.

mdisibio (approved):

PR #5460, opened 2025-07-30
[backend-worker] add config option to finish current job before shutdown

zalegrala:
**What this PR does**:

Currently, when the worker shutdown is requested, the current task is given up and the worker shuts down.  This leads to possibly failed jobs and duplicate data.  Here we add a configuration option to allow the worker to finish the current job before shutting down, and another option to set the duration that we should wait before giving up.  In large rollouts, this may take a considerable amount of time depending on the workloads, so the default timeout here is scoped to 30s.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


docs/sources/tempo/configuration/manifest.md
```diff
@@ -1103,4 +1103,6 @@ backend_worker:
         instance_addr: ""
         enable_inet6: false
         wait_active_instance_timeout: 10m0s
+    finish_on_shutdown: false
+    job_finish_timeout: 30s
```
mdisibio: Is this timeout used for anything other than shutdown? Else I might call it `finish_on_shutdown_timeout`.

Or maybe 1 setting can accomplish both?  if timeout==0 then shutdown immediately, else wait for the timeout.

PR #5464, opened 2025-07-31
Fix/spanid zero index out of range

### LINKED ISSUE #4573: Span Id with all zeros cause index out of range
**Describe the bug**

When Tempo queries for a traces that contains a span_id `0000000000000000`
cause 
` panic serving 172.23.0.3:47754: runtime error: index out of range [7] with length 0`

**To Reproduce**
1. Start Tempo 2.7.0
2. Push a trace to the `/v1/traces` endpoint a that contains a span with a span_id `0000000000000000`
3. Query the trace 

**Expected behavior**
Tempo should return an error if this against the OTEL spec

**Environment:**
 - Infrastructure: laptop
 - Deployment tool: Docker

**Additional Context**

[panic_output.log](https://github.com/user-attachments/files/18452877/panic_output.log)

tempo config 
```yaml
server:
  http_listen_port: 3200
  log_level: info

distributor:
  receivers:
    otlp:
      protocols:
        grpc:
          endpoint: "tempo:4317"
        http:
          endpoint: "tempo:4318"

storage:
  trace:
    backend: local 
    wal:
      path: /var/tempo/wal 
    local:
      path: /var/tempo/blocks

```

mexirica:
**What this PR does**: Fix panic error from empty span id, filtering the span in the ingestion.

Fixes #4573

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/traceql/metrics-queries/functions.md
+++ b/docs/sources/tempo/traceql/metrics-queries/functions.md
@@ -16,14 +16,14 @@ keywords:
 <!-- If you add a new function to this page, make sure you also add it to the _index.md#functions section.-->
 
 TraceQL supports `rate`, `count_over_time`, `sum_over_time`, `min_over_time`, `avg_over_time`, `quantile_over_time`,
-`histogram_over_time`, and `compare` functions. These methods can be appended to any TraceQL query to calculate and 
+`histogram_over_time`, and `compare` functions. These methods can be appended to any TraceQL query to calculate and
 return the desired metrics like:
 
 ```
 {} | rate()
 ```
 
-Note that `topk` and `bottomk` are also supported to only return a subset of series. These can only be added 
+Note that `topk` and `bottomk` are also supported to only return a subset of series. These can only be added
 after a metrics query like:
 
 ```
@@ -250,7 +250,7 @@ from 1 through `k` of the number of the top or bottom results.
 
 For example:
 ```
-`{ resource.service.name = "foo" } | rate() by (span.http.url)  | topk(10)`
+{ resource.service.name = "foo" } | rate() by (span.http.url)  | topk(10)
 ```
 
 The first part, `{ resource.service.name = "foo" }`, takes all spans in the service `foo`.


javiermolinar:
Hi, thank you for giving this a try!

javiermolinar (commented):
I believe we should discard the batch if the span ID is not valid, as filtering out can break the trace.

According to the OTEL spec, that span is not valid: https://github.com/open-telemetry/opentelemetry-proto-go/blob/695a196a9adc4768feffa5397f4984ab8495a84e/otlp/trace/v1/trace.pb.go#L469

Another question is where to do this check. Tempo currently sends the traces to three possible destinations: ingesters, metric generators, and a Kafka queue. Probably the best place to check this would be in the distributor:

https://github.com/grafana/tempo/blob/9874d459b52d25d691eb56919866bfeb850d315e/modules/distributor/distributor.go#L418

Second thing is that we want to avoid decoding the traces more than needed since this is a costly operation; we already do that here:

https://github.com/grafana/tempo/blob/9874d459b52d25d691eb56919866bfeb850d315e/modules/distributor/distributor.go#L477

The third thing is that we need to communicate this back to the clients, looking at the `requestsByTraceID`, I see we are validating the trace ID and refusing the batch if any of them are incorrect. We could do the same for the spans ID

https://github.com/grafana/tempo/blob/9874d459b52d25d691eb56919866bfeb850d315e/modules/distributor/distributor.go#L742-L745

pkg/validation/validate.go
```diff
@@ -1,10 +1,16 @@
...
 
+import "bytes"
+
 // ValidTraceID confirms that trace ids are 128 bits
 func ValidTraceID(id []byte) bool {
 	return len(id) == 16
 }
 
+func ValidSpanID(id []byte) bool {
+	return len(id) == 8 && bytes.Equal(id, []byte{0, 0, 0, 0, 0, 0, 0, 0})
```
javiermolinar: Shouldn't this be?:

```go
return len(id) == 8 && !bytes.Equal(id, []byte{0, 0, 0, 0, 0, 0, 0, 0})
```

mexirica: @javiermolinar  You are right, just fixed it.

javiermolinar (approved):
This looks good to me. Thank you for your contribution!

PR #5466, opened 2025-08-01
[PoC] [WIP] Marshal Trace IDs and Span IDs in hex format

### LINKED ISSUE #5437: Return Trace ID and Span ID hex-encoded
Hi,

the `/api/search` endpoint returns trace IDs hex-encoded, whereas the `/api/traces/<traceid>` endpoint returns trace IDs and span IDs base64-encoded. I think it'd be good to always return hex-encoded IDs.

The OTLP JSON Protobuf Encoding also mentions that trace ID and span ID should be hex-encoded, not base64-encoded: https://opentelemetry.io/docs/specs/otlp/#json-protobuf-encoding
Example: https://github.com/open-telemetry/opentelemetry-proto/blob/v1.7.0/examples/trace.json

Related question, the `/api/traces/<traceid>` returns `batches` as a top-level attribute, whereas the OTLP/JSON spec calls it `resourceSpans`. Except for that, I think the response confirms to the OTLP/JSON spec.

As this is a potentially breaking change, I guess it'd be a 3.0 candidate maybe. What do you think?

andreasgerstmayr:
**What this PR does**:
Use a custom type to marshal Trace IDs and Span IDs in hex format, according to the OTLP spec at https://opentelemetry.io/docs/specs/otlp/#json-protobuf-encoding

> The traceId and spanId byte arrays are represented as [case-insensitive hex-encoded strings](https://tools.ietf.org/html/rfc4648#section-8); they are not base64-encoded as is defined in the standard [Protobuf JSON Mapping](https://developers.google.com/protocol-buffers/docs/proto3#json). Hex encoding is used for traceId and spanId fields in all OTLP Protobuf messages, e.g., the Span, Link, LogRecord, etc. messages.

The protobuf patch and TraceID and SpanID types are adapted from the opentelemetry-collector (https://github.com/open-telemetry/opentelemetry-collector/blob/8b73477b981084947edbb878957e1976707484b2/proto_patch.sed, https://github.com/open-telemetry/opentelemetry-collector/blob/8b73477b981084947edbb878957e1976707484b2/pdata/internal/data/traceid.go).

Fixes #5437

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott:
does this marshal cleanly back into otel proto? i'm thinking this as well as the trace id with 0s prepended change might have to be optional somehow. perhaps a request header? 

also @adrapereira what would this break in grafana :)?

PR #5468, opened 2025-08-01
[Refactoring] interval operations

ruslan-mikhailov:
**What this PR does**:
1. Additional tests
2. Aggregator for instant queries instead of StepAggregator to simplify logic
3. Introduce IntervalMapper to delegate responsibility of interval calculations and bucket assignments.
4. Calculate aligned values only once and store them in IntervalMapper to reduce CPU usage in engine metrics.
5. To reduce spread of align* function calls, delegate alignment and instant query logic to bucket set

**Benchmarks** show ~5% improvement.
`go test -benchmem -run=^$  -bench "^(BenchmarkSumOverTime|BenchmarkHistogramAggregator_Combine)$" -benchtime=10s github.com/grafana/tempo/pkg/traceql`

<details>
  <summary>Benchmarks</summary>
  

```
goos: darwin
goarch: arm64
cpu: Apple M4 Pro
                                      │ /tmp/main.txt │         /tmp/refactoring.txt         │
                                      │    sec/op     │    sec/op     vs base                │
SumOverTime-12                           200.3m ± ∞ ¹   201.2m ± ∞ ¹       ~ (p=1.000 n=1) ²
HistogramAggregator_Combine/Small-12     6.770µ ± ∞ ¹   6.908µ ± ∞ ¹       ~ (p=1.000 n=1) ²
HistogramAggregator_Combine/Medium-12    15.37µ ± ∞ ¹   14.57µ ± ∞ ¹       ~ (p=1.000 n=1) ²
HistogramAggregator_Combine/Large-12     53.74µ ± ∞ ¹   44.56µ ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                                  182.9µ         173.3µ        -5.26%
² need >= 4 samples to detect a difference at alpha level 0.05

                                      │ /tmp/main.txt │         /tmp/refactoring.txt          │
                                      │     B/op      │     B/op       vs base                │
SumOverTime-12                          92.20Ki ± ∞ ¹   92.53Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
HistogramAggregator_Combine/Small-12    41.89Ki ± ∞ ¹   41.92Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
HistogramAggregator_Combine/Medium-12   55.32Ki ± ∞ ¹   55.35Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
HistogramAggregator_Combine/Large-12    90.45Ki ± ∞ ¹   90.48Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                                 66.30Ki         66.39Ki        +0.13%
² need >= 4 samples to detect a difference at alpha level 0.05

                                      │ /tmp/main.txt │         /tmp/refactoring.txt         │
                                      │   allocs/op   │  allocs/op    vs base                │
SumOverTime-12                            340.0 ± ∞ ¹    347.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
HistogramAggregator_Combine/Small-12      109.0 ± ∞ ¹    110.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
HistogramAggregator_Combine/Medium-12     669.0 ± ∞ ¹    670.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
HistogramAggregator_Combine/Large-12     1.603k ± ∞ ¹   1.604k ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                                   446.5          450.0        +0.79%
² need >= 4 samples to detect a difference at alpha level 0.05
```

</details>

`go test -benchmem -run=^$  -bench "^(BenchmarkSumOverTime|BenchmarkHistogramAggregator_Combine)$" -benchtime=1s github.com/grafana/tempo/pkg/traceql --count=10`

<details>
  <summary>Benchmarks</summary>
  
```
goos: darwin
goarch: arm64
cpu: Apple M4 Pro
                                      │ /tmp/main.txt │        /tmp/refactoring.txt         │
                                      │    sec/op     │   sec/op     vs base                │
SumOverTime-12                            208.6m ± 6%   203.3m ± 3%   -2.50% (p=0.002 n=10)
HistogramAggregator_Combine/Small-12      6.773µ ± 1%   6.841µ ± 1%        ~ (p=0.075 n=10)
HistogramAggregator_Combine/Medium-12     15.24µ ± 2%   14.46µ ± 2%   -5.14% (p=0.000 n=10)
HistogramAggregator_Combine/Large-12      52.77µ ± 3%   44.51µ ± 2%  -15.66% (p=0.000 n=10)
geomean                                   183.6µ        173.0µ        -5.78%

                                      │ /tmp/main.txt │        /tmp/refactoring.txt         │
                                      │     B/op      │     B/op      vs base               │
SumOverTime-12                           92.21Ki ± 0%   92.53Ki ± 0%  +0.36% (p=0.000 n=10)
HistogramAggregator_Combine/Small-12     41.89Ki ± 0%   41.92Ki ± 0%  +0.07% (p=0.000 n=10)
HistogramAggregator_Combine/Medium-12    55.32Ki ± 0%   55.31Ki ± 0%  -0.03% (p=0.000 n=10)
HistogramAggregator_Combine/Large-12     90.35Ki ± 0%   90.38Ki ± 0%  +0.03% (p=0.000 n=10)
geomean                                  66.29Ki        66.36Ki       +0.11%

                                      │ /tmp/main.txt │         /tmp/refactoring.txt         │
                                      │   allocs/op   │  allocs/op   vs base                 │
SumOverTime-12                             340.0 ± 0%    347.0 ± 0%  +2.06% (p=0.000 n=10)
HistogramAggregator_Combine/Small-12       109.0 ± 0%    110.0 ± 0%  +0.92% (p=0.000 n=10)
HistogramAggregator_Combine/Medium-12      669.0 ± 0%    669.0 ± 0%       ~ (p=1.000 n=10) ¹
HistogramAggregator_Combine/Large-12      1.597k ± 0%   1.598k ± 0%  +0.06% (p=0.000 n=10)
geomean                                    446.1         449.4       +0.76%
¹ all samples are equal
```

</details>

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/traceql/engine_metrics.go
```diff
@@ -1432,34 +1378,145 @@ func (h *Histogram) Record(bucket float64, count int) {
...
+	return i.interval(ts, i.start, i.end, i.step)
+}
+
+func (i *IntervalMapperQueryRange) IntervalMs(tsmill int64) int {
+	ts := uint64(time.Duration(tsmill) * time.Millisecond)
+	if !i.isTsValid(ts, i.startMs, i.endMs, i.step) {
+		return -1
+	}
+
+	return i.interval(ts, i.startMs, i.endMs, i.step)
```
ruslan-mikhailov, line 1520: I noticed here that start and end are truncated to ms, while step remains the same. I left origin behaviour, but it does look like a bug.

pkg/traceql/engine_metrics.go
```diff
@@ -312,21 +237,16 @@ func (set SeriesSet) ToProto(req *tempopb.QueryRangeRequest) []*tempopb.TimeSeri
...
-		intervals := IntervalCount(start, end, req.Step)
+		intervals := mapper.IntervalCount()
 		samples := make([]tempopb.Sample, 0, intervals)
 		for i, value := range s.Values {
-			ts := TimestampOf(uint64(i), req.Start, req.End, req.Step)
-
 			// todo: this loop should be able to be restructured to directly pass over
 			// the desired intervals
-			if ts < start || ts > end || math.IsNaN(value) {
+			if math.IsNaN(value) {
```
ruslan-mikhailov: Maybe I'm making a mistake here? I don't see how it can be possible that TimestampOf will be out of the range

integration/e2e/api/query_range_test.go
```diff
@@ -395,6 +395,33 @@ sendLoop:
 		})
 	}
 
+	t.Run("avg_over_time instan query", func(t *testing.T) {
```
joe-elliott:
```suggestion
	t.Run("avg_over_time instant query", func(t *testing.T) {
```

pkg/traceql/engine_metrics.go
joe-elliott: w/o deep diving all the code i could see `i` exceeding a value where its timestamp is `> end`. 

`< start` doesn't seem possible unless there's some shenanigans with the alignStart function above

pkg/traceql/engine_metrics.go
```diff
@@ -557,6 +470,47 @@ func (s *StepAggregator) Length() int {
...
+func (i *InstantAggregator) Observe(span Span) {
+	// if outside the range, skip
+	if ts := span.StartTimeUnixNanos(); ts < i.start || ts > i.end {
+		return
+	}
+	i.vector.Observe(span)
+}
+
+func (i *InstantAggregator) ObserveExemplar(float64, uint64, Labels) {
+	// instant query does not have exemplars, so we skip
```
joe-elliott, line 514: any particular reason we don't? i don't think you should add it here or anything but i would have guessed we return exemplars with instant queries if you had asked. doesn't seem like there's any technical blockers to doing it.

pkg/traceql/engine_metrics.go
```diff
@@ -1358,15 +1312,14 @@ const (
 
 type SimpleAggregator struct {
 	ss               SeriesSet
-	exemplarBuckets  *bucketSet
-	len              int
+	exemplarBuckets  bucketSet
```
joe-elliott, line 1348: curious as to why you went from a pointer to a value?

pkg/traceql/engine_metrics.go
joe-elliott: agree it looks like a bug. maybe it's partially working b/c it's consistently generating the same buckets even if they are intuitively wrong?

may as well fix it if we can show its a bug

pkg/traceql/engine_metrics.go
ruslan-mikhailov: To an interface, actually. For instant queries to avoid unnecessary action, I added alwaysFullBucketSet implementation

pkg/traceql/engine_metrics.go
ruslan-mikhailov: "Feature, not a bug". We intentionally do not support exemplars in instant query. But from technical point of view, it is possible, yes, not a big deal. Do we have such a need?

integration/e2e/api/query_range_test.go
ruslan-mikhailov: \+

pkg/traceql/engine_metrics.go
ruslan-mikhailov: Hmm. Ok, I returned the check but in form of checking interval number: `if i > intervals`. The timestamp anyway is calculated from `i`

pkg/traceql/engine_metrics.go
ruslan-mikhailov: I'll take a look, but so far I will leave it as is to not introduce changes to internal logic. Left TODO in code to state the problem

pkg/traceql/engine_metrics.go
joe-elliott: Nah, just wondering. Definitely not something to change in this PR.

pkg/traceql/engine_metrics.go
joe-elliott: oops! good call out.

joe-elliott:
this PR lgtm but I believe @mdisibio wants a look before merging so i will hold off the approval. nice work @ruslan-mikhailov!

ruslan-mikhailov:
\+ fix for edge case

pkg/traceql/engine_metrics.go
```diff
@@ -557,6 +470,47 @@ func (s *StepAggregator) Length() int {
...
+	return &InstantAggregator{
+		start:  start,
+		end:    end,
+		vector: innerAgg(),
+	}
+}
+
+func (i *InstantAggregator) Observe(span Span) {
+	// if outside the range, skip
+	if ts := span.StartTimeUnixNanos(); ts < i.start || ts > i.end {
```
mdisibio, line 507: Should these bounds be half open/closed?

pkg/traceql/engine_metrics.go
```diff
@@ -1499,34 +1445,145 @@ func (h *Histogram) Record(bucket float64, count int) {
 	})
 }
 
+type IntervalMapper interface {
```
mdisibio, line 1469: Was a little concerned about the interface here impacting performance, but it looks like the wins of saving the precomputed values ends a net benefit.  Possibly also from passing fewer params in each call.

mdisibio (approved):
One small question and LGTM. We chatted and reviewed a draft of these changes awhile back. This final PR looks even better than I expected.  The new name `IntervalMapper` is great, liking the instant vs step range aggregator, and great cleanup.

pkg/traceql/engine_metrics.go
ruslan-mikhailov: I preserved behaviour for instant query: https://github.com/ruslan-mikhailov/tempo/blob/d19bb112d5c68ddfb457cb1921c1e1e6cd161ab8/pkg/traceql/engine_metrics.go#L111-L115. Currently, instant query includes both borders. If need to change, it shouldn't be hard to do.

Side note: I think it makes sense to document the behaviour. I'll add it to my backlog

ruslan-mikhailov:

ruslan-mikhailov:
After updating to latest main, I noticed two problems:
1. vParquet5 uses interval functions. I fixed it by introducing back the functions with IntervalMapper initialisation inside. The functions are marked as deprecated.
2. [This change](https://github.com/ruslan-mikhailov/tempo/commit/2c125a140d1dcf0b118ba7e11545fb3bf9570868#diff-1b84537244dd74c3ba117a9181d6b27c2ba7d632b26c355ad5e5b428b6d00d19R148-R153) introduced a case when after TrimToBlockOverlap for a small block (end-start < step), new time range (end-start) is exactly == step. This PR has optimisation for instant queries: it does not return exemplars. The fix I applied is dirty, maybe there is a better fix for this that does not require significant effort?

@mdisibio, could you please review last two commits with the fixes?

pkg/traceql/engine_metrics.go
```diff
@@ -50,83 +50,23 @@ func DefaultQueryRangeStep(start, end uint64) uint64 {
...
-}
-
 // TimestampOf the given interval with the start and step.
+//
+// Deprecated: use IntervalMapper instead.
 func TimestampOf(interval, start, end, step uint64) uint64 {
-	start = alignStart(start, end, step)
-	// start as initial offset plus interval's offset
-	return start + (interval+1)*step
+	intervalMapper := NewIntervalMapper(start, end, step)
```
mdisibio: Since this is called several times for each span written, I think this will cause too much overhead.    I think we could instead init a few static intervalmappers inside vp5 and call them.   The start/end/steps are fixed, to 0 and ~maxint32, and 15/60/300/3600 for each column.  Then performance will return to be close to straight function call.  

The only concern is concurrency - it should be no problem to call the IntervalMapper functions concurrently right?    

Now I notice the signature is pointer `(i *IntervalMapperQueryRange)`, that could be value and would remove all uncertainty.

pkg/traceql/engine_metrics.go
```diff
@@ -1584,22 +1585,29 @@ type histSeries struct {
 	labels          Labels
 	hist            []Histogram
 	exemplars       []Exemplar
-	exemplarBuckets *bucketSet
+	exemplarBuckets bucketSet
```
zalegrala, line 1593: I believe this should remain a pointer.

pkg/traceql/engine_metrics.go
ruslan-mikhailov: Good idea! Done 

> The only concern is concurrency - it should be no problem to call the IntervalMapper functions concurrently right?

interval mapper is designed as immutable. Once created, struct values shouldn't be changed. 

> Now I notice the signature is pointer (i *IntervalMapperQueryRange), that could be value and would remove all uncertainty.

Changed to value ✅

pkg/traceql/engine_metrics.go
ruslan-mikhailov: This is actually an interface, but under the hood it is still a pointer

tempodb/encoding/vparquet5/block_traceql_test.go
```diff
@@ -698,10 +698,10 @@ func fullyPopulatedTestTraceWithOption(id common.ID, parentIDTest bool) *Trace {
 								SpanID:                 []byte("spanid"),
 								Name:                   "hello",
 								StartTimeUnixNano:      uint64(100 * time.Second),
-								StartTimeRounded15:     roundSpanStartTime(uint64(100*time.Second), 15),
-								StartTimeRounded60:     roundSpanStartTime(uint64(100*time.Second), 60),
-								StartTimeRounded300:    roundSpanStartTime(uint64(100*time.Second), 300),
-								StartTimeRounded3600:   roundSpanStartTime(uint64(100*time.Second), 3600),
+								StartTimeRounded15:     roundSpanStartTime(uint64(100*time.Second), precision15Seconds),
```
mdisibio: nit:  Simplify and reference the interval mappers directly?  This would avoid the switch statement and extra consts.

tempodb/encoding/vparquet5/schema.go
```diff
@@ -502,10 +518,10 @@ func traceToParquetWithMapping(id common.ID, tr *tempopb.Trace, ot *Trace, dedic
 					ss.StatusMessage = ""
 				}
 				ss.StartTimeUnixNano = s.StartTimeUnixNano
-				ss.StartTimeRounded15 = roundSpanStartTime(s.StartTimeUnixNano, 15)
-				ss.StartTimeRounded60 = roundSpanStartTime(s.StartTimeUnixNano, 60)
-				ss.StartTimeRounded300 = roundSpanStartTime(s.StartTimeUnixNano, 300)
-				ss.StartTimeRounded3600 = roundSpanStartTime(s.StartTimeUnixNano, 3600)
+				ss.StartTimeRounded15 = roundSpanStartTime(s.StartTimeUnixNano, precision15Seconds)
```
mdisibio: Here as well.

ruslan-mikhailov:
\+ review fixes

tempodb/encoding/vparquet5/block_traceql_test.go
ruslan-mikhailov: Good call. No idea why I decided to leave roundSpanStartTime in the first place.

tempodb/encoding/vparquet5/schema.go
ruslan-mikhailov: \+

PR #5469, opened 2025-08-01
TraceQL metrics sampling for faster performance

mdisibio:
**What this PR does**:
This adds a new query hint `with(sample=true)` that can be used with any TraceQL metrics query to return an approximate result faster.  It works by applying probabilistic sampling at the storage layer, i.e. only inspecting xx% spans, or xx% traces, depending on the needs of the query.

It uses an adaptive probabilistic approach that responds to how common spans and traces matching the query are.  When there is a lot of data, it lowers the sampling rate.  When matches are rare it keeps the sampling rate higher, possibly never going below 100%.  Therefore the performance gain depends on the query.  Heavy queries such as `{ } | rate()` are showing improvements of 2-4x.   Needle in the haystack queries won't show an improvement, but also won't lose signal.  The goal is for `with(sample=true)` to be safe to include in virtually any query, regardless of scale or selectivity.

The trick here is the new `Sampler` interface that has knowledge of the low-level parquet column activity, and the matched spansets being fed to the engine.

There are also 2 other hints added while we explore the viability of this feature:
* `with(span_sample=0.xx)`  <-- Perform span-level sampling at the fixed percent.  i.e. `span_sample=0.01` only inspects 1% of spans. 
* `with(trace_sample=0.xx)` <-- Same thing but for trace-level queries such as descendant or count()s

Example screenshot showing the level of accuracy I'm targeting with this:  green is raw results, and yellow is sampled.  The sample version was ~3x faster (21s vs 6s)

image

**Additional Context**
This is bringing back a feature https://github.com/grafana/tempo/pull/3275 that was lost when we changed the metrics job sharding to be block/rowgroup based.  However, that was never very good and this new implementation has several advantages:

* It doesn't rely on trace or span IDs so it more efficient by not reading those columns, and isn't impacted by trace ID distribution.
* Span-level sampling for supported queries gives better accuracy than trace-level sampling for the same amount of work.
* It is done locally within the job, so it is transparent to the frontend. 

I also explored using statistical methods to determine sample sizes targeting specific margins of error.  You'll see that in the [commit history](https://github.com/grafana/tempo/pull/5469/commits/99c6ccfb23dbd1e0e17a8330eb5fb8ed8425287d#diff-a471b46d703d93af65159f14165cbc92e3872c925739a6b79f64a576b5814a93R81), but there were too many edge cases, and the simple formula `int(math.Ceil(proportion * 10))` outperformed in all cases.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


knylander-grafana:
Marty, do we need to update the docs for this?

pkg/parquetquery/iters.go
```diff
@@ -1171,6 +1171,15 @@ func (j *LeftJoinIterator) SeekTo(t RowNumber, d int) (*IteratorResult, error) {
 
 func (j *LeftJoinIterator) seek(iterNum int, t RowNumber, d int) (err error) {
 	if j.peeksRequired[iterNum] == nil || CompareRowNumbers(d, j.peeksRequired[iterNum].RowNumber, t) == -1 {
+		// Release peeks if present
+		if j.peeksRequired[iterNum] != nil {
```
joe-elliott, line 1206: seems there are a handful of other places we could add this optimization. maybe we should make this a method on the IteratorResult to iterate through other entries and release anything release-able?

also, it seems like it's possible for these OtherEntries to be passed out of the iterator through the `collect()` method and this code accidentally releases them as well?

pkg/traceql/ast.go
```diff
@@ -35,6 +35,34 @@ type RootExpr struct {
 	Hints              *Hints
 }
 
+func NeedsFullTrace(e ...Element) bool {
```
joe-elliott, line 38: here's a weird one. this technically parses correctly but fails validation. i think we intended to include queries like this at one point.

```
({} | count()) = ({} | count())
```

if we ever implemented this it would need the full trace but the function returns false. it took me awhile to find this so i believe this method is generally quite sound.

pkg/traceql/ast_metrics.go
```diff
@@ -266,9 +266,24 @@ func (a *MetricsAggregate) observeSeries(ss []*tempopb.TimeSeries) {
...
-		return a.agg.Series()
+		ss := a.agg.Series()
+
+		// These operations don't get scaled by the multiplier.
+		switch a.op {
+		case metricsAggregateMinOverTime, metricsAggregateMaxOverTime:
+			return ss
+		}
+
+		if multiplier > 0 && multiplier != 1.0 {
```
joe-elliott: is a multiplier < 1.0 valid?

pkg/traceql/ast_metrics.go
```diff
@@ -266,9 +266,24 @@ func (a *MetricsAggregate) observeSeries(ss []*tempopb.TimeSeries) {
...
 
-func (a *MetricsAggregate) result() SeriesSet {
+func (a *MetricsAggregate) result(multiplier float64) SeriesSet {
 	if a.agg != nil {
-		return a.agg.Series()
+		ss := a.agg.Series()
+
+		// These operations don't get scaled by the multiplier.
+		switch a.op {
+		case metricsAggregateMinOverTime, metricsAggregateMaxOverTime:
```
joe-elliott, line 275: also should we lock this behavior with a test?

pkg/traceql/engine_metrics_average.go
```diff
@@ -67,9 +67,31 @@ func (a *averageOverTimeAggregator) observeSeries(ss []*tempopb.TimeSeries) {
...
 	if a.agg != nil {
-		return a.agg.Series()
+		ss := a.agg.Series()
+		if multiplier > 0 && multiplier != 1.0 {
+			countLabel := NewStaticString(internalMetaTypeCount)
+			for _, s := range ss {
+				// Skip non-count series.
+				found := false
+				for _, l := range s.Labels {
+					if l.Name == internalLabelMetaType && l.Value.Equals(&countLabel) {
```
joe-elliott, line 79: i know this is not new to this PR, but what does reserving `__meta_type` as a special label mean in the real world? 

```
{} | avg_over_time(span:duration) by (span.__meta_type)
```

if `span.__meta_type` happens to have a value `__count` it just returns wrong results? asking for my understanding only

pkg/traceql/engine_metrics_compare.go
```diff
@@ -286,6 +286,15 @@ func (m *MetricsCompare) result() SeriesSet {
 		addExemplar(internalLabelTypeSelectionTotal, e)
 	}
 
+	// Multiply to account for sampling as needed.
+	if multiplier > 0 && multiplier != 1.0 {
```
joe-elliott: is the `__meta_error` series impacted in anyway here

pkg/traceql/engine_metrics.go
```diff
@@ -922,6 +922,45 @@ func (e *Engine) CompileMetricsQueryRange(req *tempopb.QueryRangeRequest, exempl
 		exemplars = v
 	}
 
+	// Debug sampling hints, remove once we settle on approach.
```
joe-elliott, line 925: we could just move these to unsafe if we felt other shouldn't use them. they could remain nice debugging tools once merged.

i do really like `sample=true` or `sample=0.5` support below as well.

pkg/traceql/engine_metrics.go
```diff
@@ -1148,19 +1187,28 @@ func (e *MetricsEvaluator) Do(ctx context.Context, f SpansetFetcher, fetcherStar
...
 		for i, s := range ss.Spans {
+
 			if e.checkTime {
 				st := s.StartTimeUnixNanos()
 				if st <= e.start || st > e.end {
 					continue
 				}
 			}
 
+			if e.storageReq.SpanSampler != nil {
```
joe-elliott, line 1208: how does calling Measured() after checking the start/end time impact the final results? does this mean that blocks that barely overlap the time range will be downsampled less? so the edges of the graph will have more accurate results?

should we call measured before we filter due to time range?

tempodb/encoding/vparquet4/block_traceql.go
```diff
@@ -1900,6 +1909,13 @@ func createSpanIterator(makeIter makeIterFn, innerIterators []parquetquery.Itera
 			if err != nil {
 				return nil, err
 			}
+
+			if sampler != nil {
```
joe-elliott, line 1913: does the below check in `if len(required) == 0` need to exist?

there are some weird assumptions here that this will be used for metrics and so the span start time column must be pulled. should we return err if sampler == nil at the end of this func? it's not clear to me when lines 2147 would ever be ran.

pkg/parquetquery/iters.go
mdisibio: Yeah this is a tricky part so the question is fair. This does not double-release results, because when `seek()` is called, the values in the `peeks` slices have never been returned. They always point to the next row that we stopped at, i.e. the next span, next attribute, etc.   Inside `collect()` you can see that when it returns a result upstream, the value of `peek` is the next one in holding, or nil or the iterator was exhausted.

We never saw this before because all results were either returned upstream to be released in metrics evaluator, or inside batch/span collector and released manually.

+1 I can add the Release() method/comments and catch the other iterators too.  Was holding off until we wanted to move this PR forward.

pkg/traceql/ast.go
mdisibio: The current implementation of NeedsFullTrace is a top-down, easy mode.  Was debating on whether to add this to the AST element signature throughout, which would eliminate gaps like this, but for increased noise.  Thoughts?

pkg/traceql/ast_metrics.go
mdisibio: Hmm, I can't think of a case where the results are smaller.  This was also trying to avoid work for the common case that multiplier == 1.0.  I can change this to `if multiplier > 1.0`

pkg/traceql/engine_metrics.go
mdisibio: Good catch, this was something I tried both ways.   Calling Measured() after the check is more accurate, because it's capturing the true rate of data being found.  The edges of a block are where this most often occurs.   It prevents the false signal of data looking more common than it really is.  So this way the sampling probability stays high (starts at 100%), until there are enough spans passing the time check to make a real determination.

tempodb/encoding/vparquet4/block_traceql.go
mdisibio: I think it does need to exist. The sampler always has to be attached to the first pass's driving column:

Metrics queries have two modes related to first/second pass and performance:

(1) AllConditions and all first pass intrinsics:  i.e. `{ } | rate() by (resource.service.name)`    We eliminate the second pass and put StartTime in the first pass.  That becomes the driving column and gets the sampler.

(2) The case you are asking about:  all other queries, if we can't avoid the second pass, then we leave the start time in the second pass and it's more efficient (only processed for matching rows). In that case Status is the first pass's driving column, and we attach the sampler there.  An example is `{ resource.service.name="foo" } >> { status=error } | rate()`

pkg/traceql/engine_metrics_compare.go
mdisibio: They are added as empty series without values, so they aren't impacted.

pkg/traceql/engine_metrics_average.go
mdisibio: An attribute of that name should work fine because everything is scoped when we call the string function here: https://github.com/grafana/tempo/blob/main/pkg/traceql/engine_metrics_compare.go#L247  It would show up as `span.__meta_type="..."` and not conflict with the other which is `__meta_type="baseline|selection"`

image

pkg/traceql/ast_metrics.go
mdisibio: Cleaned up

pkg/traceql/engine_metrics.go
mdisibio: They are safe as in they don't present vulnerability or impact to cell stability. But it is possible to get misleading results from them, i.e. using span_sample on a query that needs the full trace like {} >> {}.

mdisibio:
@knylander-grafana 
> Marty, do we need to update the docs for this?

Yes, we will want to document the main hints `sample=true` and `sample=0.xx`.   Leaning towards the bottom of page https://grafana.com/docs/tempo/latest/traceql/metrics-queries/functions/, since they are (1) specific to metrics queries and (2) only understood after learning about the metrics functions themselves.

mdisibio:
For reference, there is one gap that I've identified where this approach to sampling can introduce significant error:  When a query involves a lot of data, and there is a large difference in the occurrence of the top and bottom series:  i.e. `rate() by (resource.service.name)`. This is present on every span, so the sampling rate will go low. Services with lots of data will remain accurate. But services that are rare, end up below a proper sampling rate and lose signal.

If you query the rare service directly `{ resource.service.name="rare"} | rate()`, it will sample at a higher rate and return a high quality signal.

pkg/parquetquery/iters.go
mdisibio: Added comments and addressed the other iterators.

pkg/traceql/ast.go
mdisibio: Fixed

pkg/parquetquery/iters.go
joe-elliott: oh, i see. collect calls Next after using `j.peeks` so it's safe to Release here. do we know why we call Next in collect? it seems like that could be throwaway work if the caller is done with the iterator.

pkg/traceql/enum_hints.go
```diff
@@ -8,16 +8,20 @@ import (
...
 	HintExemplars         = "exemplars"
 	HintMostRecent        = "most_recent" // traceql search hint to return most recent results ordered by time
+	HintDebug             = "debug"
+	HintInfo              = "info"
 )
 
 func isUnsafe(h string) bool {
 	switch h {
-	case HintSample, HintExemplars, HintMostRecent:
+	case HintSample, HintTraceSample, HintSpanSample, HintExemplars, HintMostRecent:
```
joe-elliott, line 24: add debug and info?

tempodb/encoding/vparquet4/block_traceql.go
joe-elliott: i don't know about you but the fetch layer is feeling quite complicated :)

thx for the explanation! makes sense

pkg/traceql/enum_hints.go
mdisibio: This is the list of safe hints.  The return statement is inverted so all hints are unsafe by default.

joe-elliott: oof. thank you

PR #5476, opened 2025-08-04
Add access handling functionality

ie-pham:
**What this PR does**: This PR adds an `AccessHandler` interface that can be used to handle customized access handling. 

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/frontend/combiner/span_matcher.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+	}
+	return true
+}
+
+func NewSpanMatcher(matcherValue string) (*SpanMatcher, error) {
+	if matcherValue == "" {
+		return nil, errors.New("span matcher header value is empty")
+	}
+	matcherValue = strings.TrimSpace(matcherValue)
```
mattdurham: Not in love with all this manual string mangling but if that is the best way then :shrug: .

modules/frontend/combiner/span_matcher.go
```diff
@@ -0,0 +1,247 @@
...
+				matchType = config.Regex
+				shouldMatch = false
+			default:
+				return nil, fmt.Errorf("unsupported match type: %v", matcher.Type)
+			}
+			policyMatcher := &PolicyMatcher{
+				shouldMatch: shouldMatch,
+			}
+			if attr.Intrinsic > 0 {
+				intrinsictMatcher, err := makeIntrinsicMatcher(matcher.Value, matchType)
```
mattdurham: nit: intrinsict spelling

modules/frontend/combiner/span_matcher.go
```diff
@@ -0,0 +1,247 @@
...
+		if !strings.HasSuffix(policyStr, "}") {
+			policyStr += "}"
+		}
+		matchers, err := parser.ParseMetricSelector(policyStr)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing policy selector '%s': %v", policyStr, err)
+		}
+		policyMatchers := make([]*PolicyMatcher, 0, len(matchers))
+		for _, matcher := range matchers {
+			shouldMatch := true
```
mattdurham: Thoughts on bouncing this to its own function? We are several layers deep by the time we hit the nested ifs below this.

modules/frontend/combiner/span_matcher.go
```diff
@@ -0,0 +1,247 @@
...
+				matchType = config.Strict
+				shouldMatch = false
+			case labels.MatchRegexp:
+				matchType = config.Regex
+				shouldMatch = true
+			case labels.MatchNotRegexp:
+				matchType = config.Regex
+				shouldMatch = false
+			default:
+				return nil, fmt.Errorf("unsupported match type: %v", matcher.Type)
```
mattdurham: Throwing the matcher name and value in there would be nice.

modules/frontend/combiner/span_matcher.go
```diff
@@ -0,0 +1,247 @@
...
+			Matchers: policyMatchers,
+		})
+	}
+	return &SpanMatcher{
+		Policies: policies,
+	}, nil
+}
+
+// so we don't have to pass the resource attributes every time we check a span
+func (sm *SpanMatcher) MatchResource(rsAttrs []*commonv1.KeyValue, rs *v1.Resource) []bool {
```
mattdurham: Not sure I understand this func. `rs` is unused and it doesnt exit early.

modules/frontend/combiner/span_matcher.go
```diff
@@ -0,0 +1,247 @@
...
+	if matchType == config.Strict {
+		return policymatch.NewNameIntrinsicFilter(value), nil
+	}
+	return policymatch.NewRegexpIntrinsicFilter(traceql.IntrinsicName, value)
+}
+
+func ProcessUnmatchedSpans(span *tracev1.Span) {
+	span.Attributes = []*commonv1.KeyValue{}
+	span.DroppedAttributesCount = 0
+	span.Name = "redacted"
```
mattdurham: Had we talked about having an allowed list of attributes? Or did we drop that and only going with redact it all approach.

modules/frontend/combiner/span_matcher.go
```diff
@@ -0,0 +1,247 @@
...
+	for i, policy := range sm.Policies {
+		if matchedResourceIndex[i] && policy.MatchSpan(span) && policy.MatchIntrinsic(span) {
+			return true
+		}
+	}
+	return false
+}
+
+func (sm *SpanMatcher) ProcessTrace(trace *tempopb.Trace) {
+	for _, b := range trace.ResourceSpans {
```
mattdurham: Would love a big comment on what ProcessTrace is doing and what the desired outcome is here with regards to redaction. Maybe even name it RedactTrace? Or some other more specific name?

modules/frontend/combiner/trace_by_id.go
```diff
@@ -30,24 +30,27 @@ type TraceByIDCombiner struct {
...
 	MetricsCombiner *TraceByIDMetricsCombiner
+
+	spanMatcher *SpanMatcher
 }
 
 // NewTraceByID returns a trace id combiner. The trace by id combiner has a few different behaviors then the others
 // - 404 is a valid response code. if all downstream jobs return 404 then it will return 404 with no body
 // - translate tempopb.TraceByIDResponse to tempopb.Trace. all other combiners pass the same object through
 // - runs the zipkin dedupe logic on the fully combined trace
 // - encode the returned trace as either json or proto depending on the request
```
mattdurham, line 41: May want to add what the span matcher does to this comment block.

modules/frontend/combiner/span_matcher.go
```diff
@@ -0,0 +1,247 @@
...
+	commonv1 "github.com/grafana/tempo/pkg/tempopb/common/v1"
+	v1 "github.com/grafana/tempo/pkg/tempopb/resource/v1"
+	tracev1 "github.com/grafana/tempo/pkg/tempopb/trace/v1"
+	"github.com/grafana/tempo/pkg/traceql"
+)
+
+const SpanMatcherHeader = "X-Span-Matcher"
+
+type SpanMatcher struct {
+	Policies []*FilterPolicy
```
mattdurham: Thoughts on unexporting both Policies and Matchers. It doesnt look to be used outside the struct and is shared resource across multiple threads. This should reduce risk of someone misusing it in the future.

mattdurham: Reviewing it further, it may not be reused across goroutines.

modules/frontend/combiner/span_matcher.go
```diff
@@ -0,0 +1,247 @@
...
+			continue
+		}
+		if matcher.shouldMatch != matcher.resourceFilter.Matches(rsAttrs) {
+			return false
+		}
+	}
+	return true
+}
+
+func NewSpanMatcher(matcherValue string) (*SpanMatcher, error) {
```
mattdurham: future thought but dont implement yet. If the expectation that we get a lot of the same matcherValues a static lru of string hash to SpanMatcher cache might be a reasonable idea if this is an expensive operation.

knylander-grafana:
@ie-pham Will we need docs for this?

modules/frontend/combiner/span_matcher.go
ie-pham: Some of these values are required by OTEL standards like span.name for example. The rest was requested by the FE team so they know to correctly display the information.

modules/frontend/combiner/common.go
```diff
@@ -289,3 +290,7 @@ func (c *genericCombiner[R]) shouldQuit() bool {
 	// 2xx
 	return false
 }
+
+type TraceRedactor interface {
+	ProcessTraceToRedactAttributes(t *tempopb.Trace)
```
mattdurham, line 295: Is the assumption here that it modifies the input? Is there a case where we would ever return an error? And if we do encounter an error what would be the state of the passed in trace?

modules/frontend/combiner/trace_by_id.go
```diff
@@ -125,6 +128,9 @@ func (c *TraceByIDCombiner) HTTPFinal() (*http.Response, error) {
 	// dedupe duplicate span ids
 	deduper := newDeduper()
 	traceResult = deduper.dedupe(traceResult)
+	if c.traceRedactor != nil {
```
mattdurham, line 131: Does it make sense to always have a valid trace redactor, even if its a noop one? That way we can remove any checks like this one.

modules/frontend/frontend.go
```diff
@@ -229,13 +241,13 @@ func New(cfg Config, next pipeline.RoundTripper, o overrides.Interface, reader t
...
 
 		// grpc/streaming
-		streamingSearch:       newSearchStreamingGRPCHandler(cfg, searchPipeline, apiPrefix, logger),
-		streamingTags:         newTagsStreamingGRPCHandler(cfg, searchTagsPipeline, apiPrefix, o, logger),
-		streamingTagsV2:       newTagsV2StreamingGRPCHandler(cfg, searchTagsPipeline, apiPrefix, o, logger),
-		streamingTagValues:    newTagValuesStreamingGRPCHandler(cfg, searchTagValuesPipeline, apiPrefix, o, logger),
-		streamingTagValuesV2:  newTagValuesV2StreamingGRPCHandler(cfg, searchTagValuesV2Pipeline, apiPrefix, o, logger),
-		streamingQueryRange:   newQueryRangeStreamingGRPCHandler(cfg, queryRangePipeline, apiPrefix, logger),
-		streamingQueryInstant: newQueryInstantStreamingGRPCHandler(cfg, queryRangePipeline, apiPrefix, logger), // Reuses the same pipeline
+		streamingSearch:       newSearchStreamingGRPCHandler(cfg, searchPipeline, apiPrefix, logger, accessHandler),
```
mattdurham, line 244: Kind of hate that these are different patterns, ala its easy to see the handler is on all of them on grpc whereas http you need to check the above. Probably not a big deal but made me double check that http was correct.

modules/frontend/combiner/trace_by_id.go
ie-pham: The default for this is nil in the config at the app level so we don't want to throw a panic if a tempo user does not want to use access handling of any kind

modules/frontend/metrics_query_handler.go
```diff
@@ -33,6 +33,13 @@ func newQueryInstantStreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTri
 		}
 
 		headers := headersFromGrpcContext(ctx)
+		if accessHandler != nil {
```
mattdurham, line 36: As before can we always have an accesshandler even if its a noop?

modules/frontend/metrics_query_range_handler.go
```diff
@@ -23,23 +21,27 @@ import (
...
+			if err != nil {
+				level.Error(logger).Log("msg", "search streaming: add filter failed", "err", err)
+				return err
+			}
+		}
 
 		// default step if not set
 		if req.Step == 0 {
 			req.Step = traceql.DefaultQueryRangeStep(req.Start, req.End)
 		}
```
mattdurham, line 44: Did we lose validateQueryRange?

mattdurham: And align?

modules/frontend/frontend.go
```diff
@@ -62,10 +63,21 @@ type QueryFrontend struct {
...
 }
 
+type AccessHandler interface {
+	AddFilterHttp(r *http.Request) error
+	AddFilterSearch(c context.Context, r *tempopb.SearchRequest) (context.Context, error)
+	AddFilterTags(c context.Context, r *tempopb.SearchTagsRequest) (context.Context, error)
+	AddFilterTagValues(c context.Context, r *tempopb.SearchTagValuesRequest) (context.Context, error)
+	AddFilterQueryRange(c context.Context, r *tempopb.QueryRangeRequest) (context.Context, error)
+	AddFilterQueryInstant(c context.Context, r *tempopb.QueryInstantRequest) (context.Context, error)
+	SwitchToV2(r *http.Request) error
```
mattdurham, line 73: What does this mean?

ie-pham:
too many changes have been made and this PR has become messy. will open another one

PR #5481, opened 2025-08-05
Add test for cost_attribution in user-configurable overrides API

electron0zero:
**What this PR does**:

Adds test for the `cost_attribution` section of the user configurable overrides API.

Also updates the docs to remove the If-Match Header from example, PATCH calls doesn't required If-Match header because that's being handled internally by the Patch Handler.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


docs/sources/tempo/operations/manage-advanced-systems/user-configurable-overrides.md
```diff
@@ -129,7 +139,7 @@ It follows the JSON merge patch protocol ([RFC 7386](https://datatracker.ietf.or
 Example:
 
 ```shell
-curl -X PUT -v -H "X-Scope-OrgID: 3" -H "If-Match: 1697726795401423" http://localhost:3100/api/overrides --data "{\"forwarders\":null}"
```
carles-grafana, line 132: You added a comment below saying that PATCH doesn't need the Etag, but this is a PUT example.
Isn't PUT supposed to require it?

carles-grafana (changes requested):

docs/sources/tempo/operations/manage-advanced-systems/user-configurable-overrides.md
electron0zero: oops, my bad. let me fix it.

electron0zero: looking at the code, I see that we don't support PUT.

also this section is titled `PATCH /api/overrides` so this curl command should be using PATCH instead of PUT

docs/sources/tempo/operations/manage-advanced-systems/user-configurable-overrides.md
```diff
@@ -129,7 +139,7 @@ It follows the JSON merge patch protocol ([RFC 7386](https://datatracker.ietf.or
...
+curl -X PUT -v -H "X-Scope-OrgID: 3" http://localhost:3100/api/overrides --data "{\"forwarders\":null}"
```
electron0zero:
```suggestion
curl -X PATCH -v -H "X-Scope-OrgID: 3" http://localhost:3100/api/overrides --data "{\"forwarders\":null}"
```

docs/sources/tempo/operations/manage-advanced-systems/user-configurable-overrides.md
electron0zero: updated.

PR #5483, opened 2025-08-05
feat: New /api/traces-check endpoint

adrapereira:
**What this PR does**:

This PR adds a new API endpoint `POST /api/traces-check` which receives a list of trace IDs and returns a list of traces that exist in Tempo. 
The main use case driving this work is for Grafana to check whether a list of trace IDs, derived from logs, exist in Tempo so it knows whether to show the link/button. It's currently frustrating to always see the button, click on it and seeing a Not Found message at the end. Users stopped trusting the button and avoid clicking on it so we're trying to only show it when we know the trace exists. 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/combiner/traces_check.go
+++ b/modules/frontend/combiner/traces_check.go
@@ -37,7 +37,7 @@ func NewTracesCheck(logger log.Logger) Combiner {
 			for _, traceID := range final.TraceIDs {
 				existingSet[traceID] = true
 			}
-			
+
 			newTraceCount := 0
 			for _, traceID := range partial.TraceIDs {
 				if !existingSet[traceID] {
--- a/modules/frontend/frontend.go
+++ b/modules/frontend/frontend.go
@@ -50,7 +50,7 @@ type QueryFrontend struct {
 	TraceByIDHandler, TraceByIDHandlerV2, SearchHandler, MetricsSummaryHandler                 http.Handler
 	SearchTagsHandler, SearchTagsV2Handler, SearchTagsValuesHandler, SearchTagsValuesV2Handler http.Handler
 	MetricsQueryInstantHandler, MetricsQueryRangeHandler                                       http.Handler
-	TracesCheckHandler                                                     					   http.Handler
+	TracesCheckHandler                                                                         http.Handler
 	MCPHandler                                                                                 http.Handler
 	cacheProvider                                                                              cache.Provider
 	streamingSearch                                                                            streamingSearchHandler
--- a/modules/frontend/traces_check_sharder.go
+++ b/modules/frontend/traces_check_sharder.go
@@ -88,13 +88,13 @@ func (s *asyncTracesCheckSharder) buildShardedRequests(parent pipeline.Request) 
 
 	reqs[0], err = cloneRequestforQueriers(parent, userID, func(r *http.Request) (*http.Request, error) {
 		params[querier.QueryModeKey] = querier.QueryModeIngesters
-		
+
 		// Copy the request body
 		if len(bodyBytes) > 0 {
 			r.Body = io.NopCloser(bytes.NewReader(bodyBytes))
 			r.ContentLength = int64(len(bodyBytes))
 		}
-		
+
 		return api.BuildQueryRequest(r, params), nil
 	})
 	if err != nil {
--- a/modules/ingester/ingester.go
+++ b/modules/ingester/ingester.go
@@ -374,7 +374,7 @@ func (i *Ingester) TracesCheck(ctx context.Context, req *tempopb.TracesCheckRequ
 	if len(req.TraceIDs) == 0 {
 		return &tempopb.TracesCheckResponse{
 			TraceIDs: []string{},
-			Metrics: &tempopb.TraceByIDMetrics{},
+			Metrics:  &tempopb.TraceByIDMetrics{},
 		}, nil
 	}
 
@@ -382,13 +382,13 @@ func (i *Ingester) TracesCheck(ctx context.Context, req *tempopb.TracesCheckRequ
 	if err != nil {
 		return nil, err
 	}
-	
+
 	inst, ok := i.getInstanceByID(instanceID)
 	if !ok || inst == nil {
 		// Return empty list if instance doesn't exist (no traces found)
 		return &tempopb.TracesCheckResponse{
 			TraceIDs: []string{},
-			Metrics: &tempopb.TraceByIDMetrics{},
+			Metrics:  &tempopb.TraceByIDMetrics{},
 		}, nil
 	}
 
@@ -408,7 +408,7 @@ func (i *Ingester) TracesCheck(ctx context.Context, req *tempopb.TracesCheckRequ
 	if len(validTraceIDs) == 0 {
 		return &tempopb.TracesCheckResponse{
 			TraceIDs: []string{},
-			Metrics: &tempopb.TraceByIDMetrics{},
+			Metrics:  &tempopb.TraceByIDMetrics{},
 		}, nil
 	}
 
--- a/modules/querier/querier.go
+++ b/modules/querier/querier.go
@@ -388,7 +388,7 @@ func (q *Querier) TracesCheck(ctx context.Context, req *tempopb.TracesCheckReque
 	for _, traceID := range foundTraceIDs {
 		foundSet[traceID] = true
 	}
-	
+
 	unfoundTraces := make([][]byte, 0)
 	for _, traceIDHex := range req.TraceIDs {
 		if !foundSet[traceIDHex] {
@@ -425,7 +425,7 @@ func (q *Querier) TracesCheck(ctx context.Context, req *tempopb.TracesCheckReque
 
 	return &tempopb.TracesCheckResponse{
 		TraceIDs: foundTraceIDs,
-		Metrics: &tempopb.TraceByIDMetrics{InspectedBytes: inspectedBytes},
+		Metrics:  &tempopb.TraceByIDMetrics{InspectedBytes: inspectedBytes},
 	}, nil
 }
 
@@ -1217,4 +1217,3 @@ func protoToTraceQLStatic(kv *tempopb.KeyValue) traceqlmetrics.KeyValue {
 		Value: val,
 	}
 }
-
--- a/pkg/api/http.go
+++ b/pkg/api/http.go
@@ -826,7 +826,7 @@ func ParseTracesCheckRequest(r *http.Request) (*tempopb.TracesCheckRequest, erro
 	}
 
 	var req tempopb.TracesCheckRequest
-	
+
 	// Try JSON first, then protobuf
 	if err := json.Unmarshal(body, &req); err != nil {
 		// If JSON parsing fails, try protobuf
--- a/tempodb/encoding/vparquet4/block_traces_check.go
+++ b/tempodb/encoding/vparquet4/block_traces_check.go
@@ -51,7 +51,6 @@ func checkTraceExistsInParquet(ctx context.Context, traceID common.ID, pf *parqu
 		min := buf[0].Clone().ByteArray()
 		return bytes.Compare(traceID, min), nil
 	})
-
 	if err != nil {
 		return false, fmt.Errorf("error binary searching row groups: %w", err)
 	}
@@ -92,7 +91,7 @@ func (b *backendBlock) TracesCheck(ctx context.Context, traceIDs []common.ID, op
 
 	results := make(map[string]bool, len(traceIDs))
 	var totalInspectedBytes uint64
-	
+
 	// Initialize all results as false
 	for _, traceID := range traceIDs {
 		results[string(traceID)] = false
@@ -147,7 +146,7 @@ func (b *backendBlock) TracesCheck(ctx context.Context, traceIDs []common.ID, op
 	if err != nil {
 		return nil, 0, fmt.Errorf("unexpected error opening parquet file: %w", err)
 	}
-	
+
 	// Check all remaining candidates in the parquet file
 	for _, traceID := range indexCandidates {
 		exists, err := checkTraceExistsInParquet(derivedCtx, traceID, pf)
@@ -156,9 +155,9 @@ func (b *backendBlock) TracesCheck(ctx context.Context, traceIDs []common.ID, op
 		}
 		results[string(traceID)] = exists
 	}
-	
+
 	totalInspectedBytes = rr.BytesRead()
 	span.SetAttributes(attribute.Int64("totalInspectedBytes", int64(totalInspectedBytes)))
-	
+
 	return results, totalInspectedBytes, nil
 }
--- a/cmd/tempo/app/modules.go
+++ b/cmd/tempo/app/modules.go
@@ -524,7 +524,7 @@ func (t *App) initQueryFrontend() (services.Service, error) {
 	// http trace by id endpoint
 	t.Server.HTTPRouter().Handle(addHTTPAPIPrefix(&t.cfg, api.PathTraces), base.Wrap(queryFrontend.TraceByIDHandler))
 	t.Server.HTTPRouter().Handle(addHTTPAPIPrefix(&t.cfg, api.PathTracesV2), base.Wrap(queryFrontend.TraceByIDHandlerV2))
-	
+
 	// http traces check endpoint
 	t.Server.HTTPRouter().Handle(addHTTPAPIPrefix(&t.cfg, api.PathTracesCheck), base.Wrap(queryFrontend.TracesCheckHandler))
 
--- a/modules/frontend/traces_check_handler.go
+++ b/modules/frontend/traces_check_handler.go
@@ -16,7 +16,6 @@ import (
 
 // newTracesCheckHandler creates a http.handler for traces check requests
 func newTracesCheckHandler(cfg Config, next pipeline.AsyncRoundTripper[combiner.PipelineResponse], o overrides.Interface, logger log.Logger) http.RoundTripper {
-
 	return RoundTripperFunc(func(req *http.Request) (*http.Response, error) {
 		tenant, err := user.ExtractOrgID(req.Context())
 		if err != nil {
--- a/modules/ingester/instance.go
+++ b/modules/ingester/instance.go
@@ -505,7 +505,7 @@ func (i *instance) TracesCheck(ctx context.Context, ids [][]byte) (*tempopb.Trac
 	metrics := tempopb.TraceByIDMetrics{}
 	foundTraceIDs := make([]string, 0, len(ids))
 	remainingIDs := make([]common.ID, 0, len(ids))
-	
+
 	// Convert byte slices to common.ID and check live traces
 	i.tracesMtx.Lock()
 	for _, id := range ids {
@@ -525,7 +525,7 @@ func (i *instance) TracesCheck(ctx context.Context, ids [][]byte) (*tempopb.Trac
 		// All traces found in live traces
 		return &tempopb.TracesCheckResponse{
 			TraceIDs: foundTraceIDs,
-			Metrics: &metrics,
+			Metrics:  &metrics,
 		}, nil
 	}
 
@@ -536,7 +536,7 @@ func (i *instance) TracesCheck(ctx context.Context, ids [][]byte) (*tempopb.Trac
 	if err != nil {
 		return nil, fmt.Errorf("headBlock.TracesCheckBatch failed: %w", err)
 	}
-	
+
 	// Add found traces from headBlock and update remaining
 	nextRemainingIDs := make([]common.ID, 0, len(remainingIDs))
 	for _, id := range remainingIDs {
@@ -551,7 +551,7 @@ func (i *instance) TracesCheck(ctx context.Context, ids [][]byte) (*tempopb.Trac
 	if len(remainingIDs) == 0 {
 		return &tempopb.TracesCheckResponse{
 			TraceIDs: foundTraceIDs,
-			Metrics: &metrics,
+			Metrics:  &metrics,
 		}, nil
 	}
 
@@ -563,12 +563,12 @@ func (i *instance) TracesCheck(ctx context.Context, ids [][]byte) (*tempopb.Trac
 		if len(remainingIDs) == 0 {
 			break
 		}
-		
+
 		results, err := i.checkTracesExistInBlockBatch(ctx, c, remainingIDs, &metrics)
 		if err != nil {
 			return nil, fmt.Errorf("completingBlock.TracesCheckBatch failed: %w", err)
 		}
-		
+
 		// Add found traces and update remaining
 		nextRemainingIDs := make([]common.ID, 0, len(remainingIDs))
 		for _, id := range remainingIDs {
@@ -586,12 +586,12 @@ func (i *instance) TracesCheck(ctx context.Context, ids [][]byte) (*tempopb.Trac
 		if len(remainingIDs) == 0 {
 			break
 		}
-		
+
 		results, err := i.checkTracesExistInLocalBlockBatch(ctx, c, remainingIDs, &metrics)
 		if err != nil {
 			return nil, fmt.Errorf("completeBlock.TracesCheckBatch failed: %w", err)
 		}
-		
+
 		// Add found traces and update remaining
 		nextRemainingIDs := make([]common.ID, 0, len(remainingIDs))
 		for _, id := range remainingIDs {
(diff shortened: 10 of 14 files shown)


knylander-grafana:
Should we update the API docs? /docs/sources/tempo/api_docs/_index.md?

adrapereira:
> Should we update the API docs? /docs/sources/tempo/api_docs/_index.md?

Yes, however this was done without input from the Tempo team so let's wait a bit for the API to be finalized and then we'll work together on the docs!

pkg/api/http.go
```diff
@@ -810,3 +812,19 @@ func ReadBodyToBuffer(resp *http.Response) (*bytes.Buffer, error) {
 
 	return buffer, nil
 }
+
+// ParseTraceLookupRequest takes an http.Request and decodes the body to create a tempopb.TraceLookupRequest
+func ParseTraceLookupRequest(r *http.Request) (*tempopb.TraceLookupRequest, error) {
+	body, err := io.ReadAll(r.Body)
```
joe-elliott: `io.ReadAll` is memory inefficient. if we can get a body size from the http request we should use [ReadAllWithEstimate](https://github.com/grafana/tempo/blob/97a244012fb84feaf2b75eb9e9b2811ddf57292b/pkg/io/read.go#L10).

An even better solution would be to unmarshal directly from the io.Reader (r.Body) if possible.

pkg/api/http.go
```diff
@@ -70,6 +71,7 @@ const (
 	PathPrefixGenerator = "/generator"
 
 	PathTraces              = "/api/traces/{traceID}"
+	PathTraceLookup         = "/api/trace-lookup"
```
joe-elliott: might need to talk about this. i don't like `/api/trace-lookup` but i can't think of anything off the top of my head i do like.

i'm also wondering if this intended to be used as a POST? since we seem to be reading the body below?

joe-elliott (commented):
So far so good, but we have some work to do to get this mergeable. We should probably sync offline for api ergonomics (path / arguments / HTTP method / etc)

Some things missing. I know some of this is on your radar, but let's just list them here.

- changelog
- api docs (and potentially other docs like config if we add something)
- e2e tests

modules/querier/querier.go
```diff
@@ -315,6 +315,97 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
+	if err != nil {
+		return nil, fmt.Errorf("error extracting org id in Querier.TraceLookup: %w", err)
+	}
+
+	ctx, span := tracer.Start(ctx, "Querier.TraceLookup")
+	defer span.End()
+
+	span.SetAttributes(attribute.Int("traceCount", len(req.TraceIDs)))
+
+	results := make(map[string]bool)
```
joe-elliott: we know the total traces so we should prealloc the map

modules/querier/querier.go
```diff
@@ -315,6 +315,97 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
+	var inspectedBytes uint64
+
+	// Initialize all trace IDs as not found
+	for _, traceID := range req.TraceIDs {
```
joe-elliott: instead of `map[string]bool` we could do `map[string]struct{}`. this is a common "flag" map in go. an unset string means "does not exist" and a set string means "does exist".

although i see you have a `map[string]bool` as part of your proto response. let's talk about the response format offline which may help inform this decision

modules/querier/querier.go
```diff
@@ -315,6 +315,97 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
+		results[fmt.Sprintf("%x", traceID)] = false
```
joe-elliott: we should use: https://github.com/grafana/tempo/blob/main/pkg/util/traceid.go#L17

modules/querier/querier.go
```diff
@@ -315,6 +315,97 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
+	}
+
+	// Check each trace ID
+	for _, traceID := range req.TraceIDs {
+		traceIDStr := fmt.Sprintf("%x", traceID)
+		
+		if !validation.ValidTraceID(traceID) {
+			continue // Skip invalid trace IDs, they remain false
```
joe-elliott: probably should 400. also we should validate the request and 400 in the query frontend before we farm out to queriers.

modules/querier/querier.go
```diff
@@ -315,6 +315,97 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
+		var getRSFn replicationSetFn
+		if q.cfg.QueryRelevantIngesters {
+			traceKey := util.TokenFor(userID, traceID)
+			getRSFn = func(r ring.ReadRing) (ring.ReplicationSet, error) {
+				return r.Get(traceKey, ring.Read, nil, nil, nil)
+			}
+		}
+
+		forEach := func(funcCtx context.Context, client tempopb.QuerierClient) (any, error) {
+			resp, err := client.FindTraceByID(funcCtx, &tempopb.TraceByIDRequest{
```
joe-elliott: yeah, we should definitely not do this. individually requesting all traces by id in the ingesters is very expensive compared to what we really need to do. we need to push the "TraceLookup" request as is all the way down to the ingesters.

modules/querier/querier.go
```diff
@@ -315,6 +315,97 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
+				}
+			}
+		}
+
+		// If not found in ingesters, check the store
+		if !found {
+			maxBytes := q.limits.MaxBytesPerTrace(userID)
+			opts := common.DefaultSearchOptionsWithMaxBytes(maxBytes)
+
+			partialTraces, _, err := q.store.Find(ctx, userID, traceID, "", "", 0, 0, opts)
```
joe-elliott: ok, similar thought here. we need to have the store support "trace id lookup" so we can efficiently test for a trace's existence w/o actually individually requesting and marshalling every one.

this will need to be pushed all the way down to the block itself. the existing [find by trace id](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet4/block_findtracebyid.go#L147) function finds and unmarshals the trace from parquet -> parquet schema trace -> proto trace which is quite expensive.

we only need to scan the trace id column to confirm/disconfirm if the block contains the trace so we only need through here:

https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet4/block_findtracebyid.go#L147-L253

modules/querier/http.go
```diff
@@ -126,6 +127,38 @@ func (q *Querier) TraceByIDHandlerV2(w http.ResponseWriter, r *http.Request) {
...
+func (q *Querier) TraceLookupHandler(w http.ResponseWriter, r *http.Request) {
+	// Enforce the query timeout while querying backends
+	ctx, cancel := context.WithDeadline(r.Context(), time.Now().Add(q.cfg.TraceByID.QueryTimeout))
+	defer cancel()
+
+	ctx, span := tracer.Start(ctx, "Querier.TraceLookupHandler")
+	defer span.End()
+
+	// Parse request body
+	body, err := io.ReadAll(r.Body)
```
joe-elliott: should we use the `api.ParseTraceLookupRequest` below?

modules/ingester/ingester.go
```diff
@@ -357,6 +357,76 @@ func (i *Ingester) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDRequ
...
+	for _, traceID := range req.TraceIDs {
+		traceIDStr := fmt.Sprintf("%x", traceID)
+		
+		if !validation.ValidTraceID(traceID) {
+			results[traceIDStr] = false
+			continue
+		}
+
+		// Check if trace exists by calling FindTraceByID
+		resp, err := inst.FindTraceByID(ctx, traceID, true)
```
joe-elliott: similar thoughts here as below. we need to push trace lookup all the way into the block.

modules/ingester/ingester.go
```diff
@@ -357,6 +357,76 @@ func (i *Ingester) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDRequ
 	return res, nil
 }
 
+// TraceLookup implements tempopb.Querier.
+func (i *Ingester) TraceLookup(ctx context.Context, req *tempopb.TraceLookupRequest) (res *tempopb.TraceLookupResponse, err error) {
```
joe-elliott: is this used? it looks like the querier calls FindTraceByID on the ingester

modules/frontend/trace_lookup_handler.go
```diff
@@ -0,0 +1,122 @@
...
+			level.Error(logger).Log("msg", "trace lookup: failed to extract tenant id", "err", err)
+			return &http.Response{
+				StatusCode: http.StatusBadRequest,
+				Status:     http.StatusText(http.StatusBadRequest),
+				Body:       io.NopCloser(strings.NewReader(err.Error())),
+			}, nil
+		}
+
+		// Parse request body
+		body, err := io.ReadAll(req.Body)
```
joe-elliott: should we use the api.ParseTraceLookupRequest below? Also we need to do any validation here

modules/frontend/trace_lookup_handler.go
```diff
@@ -0,0 +1,122 @@
...
+		if len(lookupReq.IDs) == 0 {
+			return &http.Response{
+				StatusCode: http.StatusBadRequest,
+				Body:       io.NopCloser(strings.NewReader("no trace IDs provided")),
+				Header:     http.Header{},
+			}, nil
+		}
+
+		// Convert trace IDs to bytes
+		traceIDs := make([][]byte, len(lookupReq.IDs))
```
joe-elliott: why are we converting this? the frontend should take the same request as the querier

modules/frontend/trace_lookup_handler.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/grafana/tempo/pkg/util"
+)
+
+// TraceLookupRequest represents the request body for trace lookup
+type TraceLookupRequest struct {
+	IDs []string `json:"ids"`
+}
+
+// newTraceLookupHandler creates a http.handler for trace lookup requests
+func newTraceLookupHandler(cfg Config, next pipeline.AsyncRoundTripper[combiner.PipelineResponse], o overrides.Interface, logger log.Logger) http.RoundTripper {
```
joe-elliott: it looks like this is being setup as a simple pass through the frontend to the querier, but we need to implement sharding so we can break the request into a lot of small pieces. check out the find by trace id path to see how we do sharding there.

modules/frontend/frontend.go
```diff
@@ -195,6 +196,7 @@ func New(cfg Config, next pipeline.RoundTripper, o overrides.Interface, reader t
 
 	traces := newTraceIDHandler(cfg, tracePipeline, o, combiner.NewTypedTraceByID, logger)
 	tracesV2 := newTraceIDV2Handler(cfg, tracePipeline, o, combiner.NewTypedTraceByIDV2, logger)
+	traceLookup := newTraceLookupHandler(cfg, tracePipeline, o, logger)
```
joe-elliott: i'm honestly not sure what this is going to do. it seems you are passing this through the trace pipeline which should shard the request into a lot of small jobs for the queriers, but your trace lookup handler on the querier doesn't seem to respect job sharding.

modules/frontend/combiner/trace_lookup.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"github.com/grafana/tempo/pkg/api"
+	"github.com/grafana/tempo/pkg/tempopb"
+)
+
+var _ GRPCCombiner[*tempopb.TraceLookupResponse] = (*genericCombiner[*tempopb.TraceLookupResponse])(nil)
+
+// NewTraceLookup returns a trace lookup combiner
+func NewTraceLookup() Combiner {
```
joe-elliott: are we going to support streaming here? trace id and metrics summary are the only endpoints that currently do not.

modules/frontend/combiner/trace_lookup.go
```diff
@@ -0,0 +1,58 @@
...
+		current:        &tempopb.TraceLookupResponse{Results: make(map[string]bool), Metrics: &tempopb.SearchMetrics{}},
+
+		combine: func(partial *tempopb.TraceLookupResponse, final *tempopb.TraceLookupResponse, resp PipelineResponse) error {
+			if final.Results == nil {
+				final.Results = make(map[string]bool)
+			}
+
+			// Merge the results - if any partial response indicates a trace exists, mark it as found
+			for traceID, exists := range partial.Results {
+				if exists || final.Results[traceID] == false {
```
joe-elliott: should this be &&?

modules/querier/querier.go
adrapereira: Will update to use `map[string]struct{}` and return a list of strings with trace IDs that Tempo found.

modules/querier/querier.go
adrapereira: Should we discard the full request if one ID is invalid? The current approach treats it as if it didn't exist in Tempo which is more forgiving.

ifrost:
> The main use case driving this work is for Grafana to check whether a list of trace IDs, derived from logs, exist in Tempo so it knows whether to show the link/button

Just to double check - how this API is going to be consumed in the frontend? At the moment links are calculated only when someone opens the log details. Are we going to call the API after opening log details (that would mean probably just one traceId) or we're gonna change Logs Panel to calculate and validate all links after loading the data? If we're gonna process all links, might be worth checking with Logs squad on possible performance impact?

Also - is there way to tell why the trace doesn't exist? E.g. it just was not ingested yet or dropped by adaptive tracing. I'm just thinking about UX - are we gonna hide/gray out the button or give more info to the user?

adrapereira:
@ifrost The ideal integration is this: 
1. Logs Panel computes the trace ID list for the visible window of logs 
2. makes a request to Tempo or to the Tempo datasource to check which traces exist
3. stores the result in state and renders the button only for the traces that exist
4. when the user scrolls and a new window of logs is visible make a new request only for the new trace IDs

modules/frontend/combiner/trace_lookup.go
adrapereira: No streaming in my opinion

modules/querier/querier.go
joe-elliott: yeah, i think so, the message should indicate why it was 400'ed. it feels better to give a clear error code when the request params are incorrect than a partial result.

adrapereira:
@joe-elliott Update the branch to use a new sharder that clones the body for each shard request. Seems to be working well now, thank you for the help!

It might still be missing a few tests but do you mind re-reviewing the approach? Once we're good on that I'll finish adding the tests.

modules/frontend/combiner/traces_check.go
```diff
@@ -0,0 +1,106 @@
...
+					}
+					return 0
+				}(),
+				"final_trace_count_before", len(final.TraceIDs))
+
+			if final.TraceIDs == nil {
+				final.TraceIDs = []string{}
+			}
+
+			// Merge trace IDs - add any new found traces (avoid duplicates)
```
joe-elliott, line 35: this work doesn't need to be done in len(partial.TraceIDs) == 0

also, if the strings were sorted, you could do a binary search which might be faster and would definitely be more memory efficient than converting to a map.

modules/frontend/frontend.go
```diff
@@ -123,6 +129,17 @@ func New(cfg Config, next pipeline.RoundTripper, o overrides.Interface, reader t
...
 
+	tracesCheckPipeline := pipeline.Build(
+		[]pipeline.AsyncMiddleware[combiner.PipelineResponse]{
+			headerStripWare,
+			urlDenyListWare,
+			pipeline.NewWeightRequestWare(pipeline.TracesCheck, cfg.Weights),
+			multiTenantMiddleware(cfg, logger),
+			newAsyncTracesCheckSharder(&cfg.TracesCheck, logger),
+		},
+		[]pipeline.Middleware{traceIDStatusCodeWare, retryWare},
```
joe-elliott, line 140: what does traceIDStatusCodeWare do? does it make sense here?

modules/frontend/config.go
```diff
@@ -65,6 +66,16 @@ type TraceByIDConfig struct {
 	RF1After time.Time `yaml:"-"`
 }
 
+type TracesCheckConfig struct {
+	QueryShards      int       `yaml:"query_shards,omitempty"`
+	ConcurrentShards int       `yaml:"concurrent_shards,omitempty"`
+	SLO              SLOConfig `yaml:",inline"`
```
joe-elliott, line 72: this seems unused? should we use the same settings as trace by id lookup? or use this?

modules/frontend/config.go
```diff
@@ -65,6 +66,16 @@ type TraceByIDConfig struct {
...
+
+	// RF1After specifies the time after which RF1 logic is applied, injected by the configuration
+	// or determined at runtime based on search request parameters.
+	RF1After time.Time `yaml:"-"`
```
joe-elliott, line 76: this field only exists for the transition to rhythm/tempo 3.0. it might be better to remove this and just always query rf1 blocks

modules/frontend/traces_check_sharder.go
```diff
@@ -0,0 +1,134 @@
...
+func (s *asyncTracesCheckSharder) buildShardedRequests(parent pipeline.Request) ([]pipeline.Request, error) {
+	userID, err := user.ExtractOrgID(parent.Context())
+	if err != nil {
+		return nil, err
+	}
+
+	// Read the request body once and store it for reuse
+	var bodyBytes []byte
+	if parent.HTTPRequest().Body != nil {
+		bodyBytes, err = io.ReadAll(parent.HTTPRequest().Body)
```
joe-elliott, line 79: use our io.ReadAllWithEstimate() and use ContentLength from the http request.

modules/frontend/traces_check_sharder.go
```diff
@@ -0,0 +1,134 @@
...
+// block boundaries
+func (s *asyncTracesCheckSharder) buildShardedRequests(parent pipeline.Request) ([]pipeline.Request, error) {
+	userID, err := user.ExtractOrgID(parent.Context())
+	if err != nil {
+		return nil, err
+	}
+
+	// Read the request body once and store it for reuse
+	var bodyBytes []byte
+	if parent.HTTPRequest().Body != nil {
```
joe-elliott, line 78: if body is nil or empty shouldn't we just immediately return since its an empty query?

modules/ingester/ingester.go
```diff
@@ -357,6 +358,75 @@ func (i *Ingester) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDRequ
 	return res, nil
 }
 
+// TracesCheck implements tempopb.Querier.
+func (i *Ingester) TracesCheck(ctx context.Context, req *tempopb.TracesCheckRequest) (res *tempopb.TracesCheckResponse, err error) {
```
joe-elliott, line 362: this code will need to be replicated in the new livestore module. in tempo 3.0 ingesters go away and livestores become the new "recent data" storage process.

modules/ingester/ingester.go
```diff
@@ -357,6 +358,75 @@ func (i *Ingester) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDRequ
...
+			level.Error(log.Logger).Log("msg", "recover in TracesCheck", "stack", r, string(debug.Stack()))
+			err = errors.New("recovered in TracesCheck")
+		}
+	}()
+
+	// tracing instrumentation
+	ctx, span := tracer.Start(ctx, "Ingester.TracesCheck")
+	defer span.End()
+
+	if len(req.TraceIDs) == 0 {
```
joe-elliott, line 374: i don't mind this guard code here, but really it should be in the query frontend

modules/ingester/ingester.go
```diff
@@ -357,6 +358,75 @@ func (i *Ingester) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDRequ
...
+		// Return empty list if instance doesn't exist (no traces found)
+		return &tempopb.TracesCheckResponse{
+			TraceIDs: []string{},
+			Metrics:  &tempopb.TraceByIDMetrics{},
+		}, nil
+	}
+
+	// Filter valid trace IDs
+	validTraceIDs := make([][]byte, 0, len(req.TraceIDs))
+	for _, traceIDHex := range req.TraceIDs {
```
joe-elliott, line 397: same thoughts here. we should do this in the query-frontend and 400 if any trace ids are invalid.

modules/querier/http.go
```diff
@@ -126,6 +126,62 @@ func (q *Querier) TraceByIDHandlerV2(w http.ResponseWriter, r *http.Request) {
...
+func (q *Querier) TracesCheckHandler(w http.ResponseWriter, r *http.Request) {
+	// Enforce the query timeout while querying backends
+	ctx, cancel := context.WithDeadline(r.Context(), time.Now().Add(q.cfg.TraceByID.QueryTimeout))
+	defer cancel()
+
+	ctx, span := tracer.Start(ctx, "Querier.TracesCheckHandler")
+	defer span.End()
+
+	// Parse request body
+	body, err := api.ParseTracesCheckRequest(r)
```
joe-elliott, line 138: nit: name is strange. body => req

modules/querier/querier.go
```diff
@@ -315,6 +316,119 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
+	ctx, span := tracer.Start(ctx, "Querier.TracesCheck")
+	defer span.End()
+
+	span.SetAttributes(attribute.Int("traceCount", len(req.TraceIDs)))
+
+	var inspectedBytes uint64
+	var foundTraceIDs []string
+
+	// Query ingesters first (only if queryMode allows)
+	if req.QueryMode == "" || req.QueryMode == QueryModeAll || req.QueryMode == QueryModeIngesters {
```
joe-elliott, line 334: queryMode empty should likely be a 400? this cases shouldn't happen

modules/querier/querier.go
```diff
@@ -315,6 +316,119 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
+		for _, traceIDHex := range req.TraceIDs {
+			// Convert hex string to bytes for validation
+			traceID, err := hex.DecodeString(traceIDHex)
+			if err != nil || !validation.ValidTraceID(traceID) {
+				continue
```
joe-elliott, line 339: we shouldn't ignore invalid trace ids, we should fail with 400. also, as above, this should be done by the frontend but can be repeated here

modules/querier/querier.go
```diff
@@ -315,6 +316,119 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
+			var getRSFn replicationSetFn
+			if q.cfg.QueryRelevantIngesters {
+				traceKey := util.TokenFor(userID, traceID)
+				getRSFn = func(r ring.ReadRing) (ring.ReplicationSet, error) {
+					return r.Get(traceKey, ring.Read, nil, nil, nil)
+				}
+			}
+
+			// Create a single trace request for each trace ID
+			singleReq := &tempopb.TracesCheckRequest{
```
joe-elliott, line 351: we can't do this. the full request with all trace ids has to be pushed down to the ingesters.

modules/querier/querier.go
```diff
@@ -315,6 +316,119 @@ func (q *Querier) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReque
...
+		}
+	}
+
+	if len(unfoundTraces) > 0 && (req.QueryMode == "" || req.QueryMode == QueryModeAll || req.QueryMode == QueryModeBlocks) {
+		maxBytes := q.limits.MaxBytesPerTrace(userID)
+		opts := common.DefaultSearchOptionsWithMaxBytes(maxBytes)
+
+		// Check each unfound trace in storage with block filtering
+		for _, traceID := range unfoundTraces {
+			partialTraces, _, err := q.store.Find(ctx, userID, traceID, req.BlockStart, req.BlockEnd, int64(req.TimeStart), int64(req.TimeEnd), opts)
```
joe-elliott, line 409: same here. there can be a lot of shared work if we push all trace ids all the way down into the storage layer.

also, `q.store.Find()` fully marshals the trace from parquet which is very expensive. we need a new function that does not do this and only confirms/denies a trace exists.

tempodb/encoding/vparquet4/block_traces_check.go
```diff
@@ -0,0 +1,163 @@
+package vparquet4
+
+import (
```
joe-elliott, line 3: this will all need to be copied into the new vParquet5 as well.

tempodb/encoding/vparquet4/block_traces_check.go
```diff
@@ -0,0 +1,163 @@
...
+	// Initialize all results as false
+	for _, traceID := range traceIDs {
+		results[string(traceID)] = false
+	}
+
+	// First pass: check bloom filter for all trace IDs
+	// This can quickly eliminate many traces that definitely don't exist
+	candidateIDs := make([]common.ID, 0, len(traceIDs))
+	for _, traceID := range traceIDs {
+		found, err := b.checkBloom(derivedCtx, traceID)
```
joe-elliott, line 104: we should share work here. this will repeatedly pull the same bloom filters over and over again. we need to only pull each one once.

tempodb/encoding/vparquet4/block_traces_check.go
```diff
@@ -0,0 +1,163 @@
...
+
+	// Third pass: check parquet file for remaining candidates
+	pf, rr, err := b.openForSearch(derivedCtx, opts)
+	if err != nil {
+		return nil, 0, fmt.Errorf("unexpected error opening parquet file: %w", err)
+	}
+
+	// Check all remaining candidates in the parquet file
+	for _, traceID := range indexCandidates {
+		exists, err := checkTraceExistsInParquet(derivedCtx, traceID, pf)
```
joe-elliott, line 152: same thought here. we shouldn't check each trace individually b/c it will repeatedly open and reopen the parquet file which is costly. traceids are sorted in the parquet file. if we sort the indexCandidates slice than we can do a single pass through the parquet file to confirm/disconfirm all trace ids

knylander-grafana:
Do we need to update the API docs for this endpoint? i asked earliear, and it looks like we have input from the Tempo team. :)  https://grafana.com/docs/tempo/latest/api_docs/

PR #5493, opened 2025-08-07
wip

carles-grafana:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## main / unreleased
 
+* [CHANGE] Return Bad Request from frontend if the provided tag is invalid in SearchTagValuesV2 endpoint [#5493](https://github.com/grafana/tempo/pull/5493/) (@carles-grafana)
```
joe-elliott: i'd call this a bugfix. it should have always returned 400 for an invalid attr name

modules/frontend/tag_handlers_test.go
```diff
@@ -108,6 +109,23 @@ func runnerTagValuesV2BadRequestOnOrgID(t *testing.T, f *QueryFrontend) {
 	require.Equal(t, status.Error(codes.InvalidArgument, "no org id"), err)
 }
 
+func runnerTagValuesV2BadRequestOnInvalidTagName(t *testing.T, f *QueryFrontend) {
```
joe-elliott, line 112: nice test! can we add one more in integration/e2e/api? we have a few tests in there that test 400s. we should honestly expand it.

modules/frontend/tag_handlers.go
```diff
@@ -188,6 +188,11 @@ func newTagValuesV2StreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTrip
 			return err
 		}
 
+		if _, err := traceql.ParseIdentifier(req.TagName); err != nil {
```
joe-elliott: validation is starting to feel everywhere. i think long term we need to start considering where to put it and try to consolidate all logic to either the sharder or handler for all signals.

in this case i think this would be better to implement once in the sharder around here:

https://github.com/grafana/tempo/blob/main/modules/frontend/tag_sharder.go#L206-L209

can you try that and see if your tests still pass?

joe-elliott (commented):
nice! i'd like to try the validation logic in a different spot but this looks good overall

modules/frontend/tag_handlers.go
```diff
@@ -188,6 +188,11 @@ func newTagValuesV2StreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTrip
...
+			level.Error(logger).Log("msg", "failed to parse tagName", "err", err)
```
javiermolinar: I see this same validation logic is used in several places, in the ingesters and the queriers:

https://github.com/grafana/tempo/blob/4ba2cb9ae113a8cbb6e064c2536fc303d1be45ad/modules/ingester/instance_search.go#L405

https://github.com/grafana/tempo/blob/4ba2cb9ae113a8cbb6e064c2536fc303d1be45ad/modules/querier/querier.go#L1007

If we are already validating this at query-frontend and querier level, I don't see a reason to do it again in the ingesters.

There is also a nice method that validates the traceQL identifier and more:

https://github.com/grafana/tempo/blob/4ba2cb9ae113a8cbb6e064c2536fc303d1be45ad/pkg/api/search_tags.go#L381C6-L381C34

Maybe we could reuse it

modules/frontend/tag_handlers.go
carles-grafana: I've moved the validation to the sharder in the last commit and it also works.

But I'm wondering, shouldn't we catch this as early as possible in the frontend?

Also, in either case, should we remove the validation from the querier, because it will be duplicated and always converted to 500 in the frontend?

modules/frontend/tag_handlers_test.go
carles-grafana: added

CHANGELOG.md
carles-grafana: changed

modules/frontend/tag_handlers.go
joe-elliott: > But I'm wondering, shouldn't we catch this as early as possible in the frontend?

That's a good Q. There's generally not much of anything occurs between the handlers and the sharders so it's not a big of a deal to put it in the sharder. I asked for it there only b/c that's the location of most validation logic for most endpoints.

I do like the idea of bringing this up for discussion. Now that the handlers exist maybe we should push all validation logic to that layer? I'm not opposed. Note that some validation naturally happens at the sharder or handler layer b/c they have to parse the request for different reasons.

> Also, in either case, should we remove the validation from the querier

The querier and ingester endpoints can be called independently so I think it makes sense to have validation logic in there as well. It's also possible for the frontend to mutate the request in a way that makes it invalid and so it's good for the querier/ingester to double check.

> because it will be duplicated and always converted to 500 in the frontend?

this is intentional. if the frontend produces a job for a querier that 400s it is assumed that a bug in the frontend created a bad query and the whole request should 500.

modules/frontend/tag_handlers.go
joe-elliott: > If we are already validating this at query-frontend and querier level, I don't see a reason to do it again in the ingesters.

some reasons given in https://github.com/grafana/tempo/pull/5493#discussion_r2263322119

carles-grafana: In the end I'm basically calling the function you mentioned in the sharder

PR #5494, opened 2025-08-07
Initial search functions for the live_store to mirror what ingester does.

mattdurham:
Add search functionality that is mostly copied from the ingester. Some things in the old dont pass the linter in the new tests so there are sadly subtle but meaningless differences.

### DIFF
--- a/modules/livestore/instance_search_test.go
+++ b/modules/livestore/instance_search_test.go
@@ -44,7 +44,7 @@ const (
 )
 
 func TestInstanceSearch(t *testing.T) {
-	i, _, _ := defaultInstanceAndTmpDir(t)
+	i, _ := defaultInstanceAndTmpDir(t)
 
 	tagKey := foo
 	tagValue := bar
@@ -91,7 +91,7 @@ func TestInstanceSearchTraceQL(t *testing.T) {
 
 	for _, query := range queries {
 		t.Run(fmt.Sprintf("Query:%s", query), func(t *testing.T) {
-			i, _, _ := defaultInstanceAndTmpDir(t)
+			i, _ := defaultInstanceAndTmpDir(t)
 
 			_, ids := pushTracesToInstance(t, i, 10)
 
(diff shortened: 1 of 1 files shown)


modules/livestore/instance.go
```diff
@@ -228,6 +229,13 @@ func (i *instance) cutBlocks(immediate bool) (uuid.UUID, error) {
 	return id, nil
 }
 
+var blockConfig = common.BlockConfig{}
+
+func init() {
+	// TODO MRD this is a hack until we roll the config into the livestore
+	blockConfig.RegisterFlagsAndApplyDefaults("", &flag.FlagSet{})
```
mattdurham, line 236: If we dont do this bloom filters try to apply int64 max transformations.

modules/livestore/instance_search.go
```diff
@@ -0,0 +1,661 @@
+/*
```
mattdurham, line 1: Most of this file is copied from instance_search, limits are not yet applied but TODOs are sprinkled in the code where appropriate.

modules/livestore/instance.go
```diff
@@ -29,7 +30,7 @@ type instance struct {
 	enc encoding.VersionedEncoding
 
 	// Block management
-	blocksMtx      sync.Mutex
+	blocksMtx      sync.RWMutex
```
mattdurham, line 33: Needed more fine grained mutex when dealing with blocks.

mattdurham:
test-e2e-api runs fine locally...le sigh. But good to review at the moment.

modules/ingester/instance_search.go
```diff
@@ -1,3 +1,6 @@
+/*
+* ingester is used as the template for livestore/instance_search.go any changes here should be reflected there and vice versa.
```
joe-elliott, line 2: agree with the fork. hopefully we will delete this soon.

modules/livestore/instance.go
joe-elliott: rwmutex's are slower than plain ol' mutexes. unless it's heavy on reads a lot of times a plain mutex is faster. no idea what the ratio is.

modules/livestore/instance.go
joe-elliott: odd we're passing in a defaulted block config. should we not add that to the config struct in config.go?

joe-elliott (commented):
couple Qs but generally looks good. we can approve this quick.

modules/livestore/instance.go
mattdurham: The ingester code used RWMutex so assuming we thought about that then :)

joe-elliott (approved):
onward!

Issue #4694, opened 2025-02-13
Create vParquet5


joe-elliott:
We have not yet begun on the next iteration of the vParquet format, but I would like to start collecting ideas for what we would include. The past two iterations have been focused on adding columns and features and so I propose we focus on removing columns and cleanup.

I would like to:

- Remove "well known" columns and rely instead on dedicated columns
  - This would remove a significant amount of complexity in our read/write paths as well reduce footer size
- Remove the [ServiceStats](https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet4/schema.go#L251) columns in favor of a marshalled proto or json representation. We only return these on search and don't need them broken out into individual columns.

Everything is up for discussion. Please put ideas here!

cc @stoewer, @mdisibio, @ndk

mdisibio:
> Dictionary-less dedicated columns

Agree our current handling of high-cardinality/blobs is a limiting factor which is driving up cpu and memory to encode/decode the dictionaries.   Custom columns could be a good first step.  Long-term I would like to see an ability to turn dictionary encoding on or off based on the cardinality of data.  Kind of best of both worlds.   The traceql engine/storage reads could work seamlessly, I think all we need is to toggle the parquet schema during write time.

mdisibio:
More dedicated columns, and at other levels like events would probably be a net gain.  There's an inflection point where too many columns becomes negative, we don't have good numbers of where that is or how it relates to the composition of data.

ndk:
~~Hey folks, I have a somewhat *dilettante/naive* question, so please don't pelt me with rotten tomatoes! :)~~

In my current understanding, **the offset (or index) for a given `key`** in one column **matches** the offset for its corresponding **`value`** in another column, especially in a repeated structure like `rs.list.element.ss.list.element.Spans.list.element.Attrs.list.element.Key` vs `rs.list.element.ss.list.element.Spans.list.element.Attrs.list.element.ValueDouble.list.element`. 

Is there any way to exploit that 1:1 offset alignment to optimize how Tempo retrieves key-value pairs? Would be it Parquet-friendly approach?

mdisibio:
Implementing the proposed `root` and `parent` scopes would be very useful.  Details still to be worked out, but likely need new columns so considering for the next format.

mdisibio:
[...]

PR #5495, opened 2025-08-07
vParquet5 Preview 1 - low resolution timestamp columns

mdisibio:
**What this PR does**:
Adds new block format vParquet5 with low-resolution timestamp columns mentioned [here](https://github.com/grafana/tempo/issues/4694#issuecomment-3089664736).  

This isn't the entire set of ideas for the new format, but I want to iterate instead of doing a big bang.  The approach I want to try is:
* Set the version string to `vParquet5-preview1`
* Each time we make column changes, increment the preview number.
* Update Tempo to gracefully handle unrecognized formats, in the same way that `v2` blocks don't honor traceql queries.

This would let us deploy iterative format changes without having to clear existing installs which is quite cumbersome:  object storage, ingesters, block builders, generators, etc.  Data become unqueryable on the prior preview formats which is acceptable for dev installs.

**Low-resolution timestamp columns**
These are copies of the span start time but rounded to common granularities used in metrics queries.  These compress much better and significantly reduce i/o, opening up better performance and longer range.  Even on small scale (< 1K spans/s) reduction is 75%.   On larger scale it is 90-96%.  Compression gets better as span density increases because value repetition increases.

Not much rounding is needed:  just 15s already provides the 75% and 90% savings.  This level of granularity is used for queries as short as 1h (this is our default granularity in [DefaultQueryRangeStep](https://github.com/grafana/tempo/blob/main/pkg/traceql/engine_metrics.go#L31)).  The other granularities keep increasing savings, but not as drastically. But at the same time they are effectively free (block size increase is negligible).  The 1h column is 96% smaller at high scale.  They are to help longer time ranges.

There is a trick to getting these huge savings:  We aren't storing rounded 64-bit nanos.  Instead it is an integer offset for that interval from the unix epoch.  I.e. value 1 means "00:00:30" since the epoch. These can be represented with fewer bits and is required for maximum compression.  Rounding the actual 64-bit nano value only helps by 50% and less.  This integer offset is calculated on write, and inverted on read.

Example:
```
StartTimeUnixNano-> Size In Bytes: 20038592 Size In Ratio: 0.03902404
StartTimeRounded15-> Size In Bytes: 1784741 Size In Ratio: 0.0034756837
StartTimeRounded60-> Size In Bytes: 1381463 Size In Ratio: 0.0026903222
StartTimeRounded300-> Size In Bytes: 1108503 Size In Ratio: 0.0021587478
StartTimeRounded3600-> Size In Bytes: 871148 Size In Ratio: 0.0016965122
```

**Limitations**
There are cases where we must fallback to the full nanosecond column in order to provide accurate data:
* The step isn't an even multiple of any of the preselected values
* Blocks that overlap the handoff between generators and object storage (where we must split a block in half). This happens for both object storage and inside the generators.
* Blocks that overlap the edges of the query range

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Additionally the `compaction_tenant_backoff_total` metric has been renamed to `c
 * [CHANGE] Remove .005s and add a 1.5s bucket to all request duration histograms [#5492](https://github.com/grafana/tempo/pull/5492) (@joe-elliott)
 * [CHANGE] Improve tempo writes dashboard [#5500](https://github.com/grafana/tempo/pull/5500) (@javiermolinar)
 * [CHANGE] **BREAKING CHANGE** TraceQL Metrics buckets are calculated based on data in past [#5366](https://github.com/grafana/tempo/pull/5366) (@ruslan-mikhailov)
+* [CHANGE] Change worker to shutdown after the current job, waiting 30s by default [#5460](https://github.com/grafana/tempo/pull/5460) (@zalegrala)
 * [FEATURE] New block encoding vParquet5-preview1 with low-resolution timestamp columns for better TraceQL metrics performance. This format is in development and breaking changes are expected before final release. [#5495](https://github.com/grafana/tempo/pull/5495) (@mdisibio)
 * [FEATURE] Add histograms `spans_distance_in_future_seconds` / `spans_distance_in_past_seconds` that count spans with end timestamp in the future / past. While spans in the future are accepted, they are invalid and may not be found using the Search API. [#4936](https://github.com/grafana/tempo/pull/4936) (@carles-grafana)
 * [FEATURE] Add MCP Server support. [#5212](https://github.com/grafana/tempo/pull/5212) (@joe-elliott)
@@ -48,6 +49,7 @@ Additionally the `compaction_tenant_backoff_total` metric has been renamed to `c
 * [ENHANCEMENT] Make block ordering deterministic [#5411](https://github.com/grafana/tempo/pull/5411) (@rajiv-singh)
 * [BUGFIX] Fix race condition between compaction provider and backend-scheduler [#5409](https://github.com/grafana/tempo/pull/5409) (@zalegrala)
 * [BUGFIX] Do not allow very small steps [#5441](https://github.com/grafana/tempo/pull/5441) (@ruslan-mikhailov)
+* [BUGFIX] Fix incorrect results in TraceQL compare() for spans with array attributes [#5519](https://github.com/grafana/tempo/pull/5519) (@ruslan-mikhailov)
 
 # v2.8.2
 
--- a/docs/sources/tempo/api_docs/mcp-server.md
+++ b/docs/sources/tempo/api_docs/mcp-server.md
@@ -27,8 +27,8 @@ Be aware that using this feature will likely cause tracing data to be passed to 
 
 To experiment with the MCP server using dummy data and Claude Code:
 
-1. Run the local docker-compose example in `/example/docker-compose`. This will expose the mcp server at http://localhost:3200/api/mcp
-1. Run `claude mcp add -t stdio -s user tempo npx mcp-remote http://localhost:3200/api/mcp` to add a reference to Claude Code.
+1. Run the local docker-compose example in `/example/docker-compose/local`. This will expose the mcp server at http://localhost:3200/api/mcp
+1. Run `claude mcp add --transport=http tempo http://localhost:3200/api/mcp` to add a reference to Claude Code.
 1. Run `claude` and ask some questions
 
 This MCP server has also been tested succesfully in cursor using the [`mcp-remote`](https://www.npmjs.com/package/mcp-remote) package.
--- a/docs/sources/tempo/configuration/manifest.md
+++ b/docs/sources/tempo/configuration/manifest.md
@@ -1103,4 +1103,5 @@ backend_worker:
         instance_addr: ""
         enable_inet6: false
         wait_active_instance_timeout: 10m0s
+    finish_on_shutdown_timeout: 30s
 ```
--- a/integration/backendscheduler/config.yaml
+++ b/integration/backendscheduler/config.yaml
@@ -41,6 +41,7 @@ backend_scheduler:
       measure_interval: 1s
 backend_worker:
   backend_scheduler_addr: tempo-integration-backend-scheduler:9095
+  finish_on_shutdown_timeout: 1s
 backend_scheduler_client:
   grpc_client_config:
     max_send_msg_size: 134217728
--- a/modules/backendscheduler/backendscheduler.go
+++ b/modules/backendscheduler/backendscheduler.go
@@ -296,6 +296,8 @@ func (s *BackendScheduler) UpdateJob(ctx context.Context, req *tempopb.UpdateJob
 		return &tempopb.UpdateJobStatusResponse{}, status.Error(codes.NotFound, work.ErrJobNotFound.Error())
 	}
 
+	metricJobDuration.WithLabelValues(j.GetType().String()).Observe(time.Since(j.GetCreatedTime()).Seconds())
+
 	switch req.Status {
 	case tempopb.JobStatus_JOB_STATUS_RUNNING:
 	case tempopb.JobStatus_JOB_STATUS_SUCCEEDED:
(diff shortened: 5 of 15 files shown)


stoewer:
For reviewers using Goland: there is a diff tool that allows you to view [differences between directories](https://www.jetbrains.com/help/idea/comparing-files-and-folders.html#comparing_folders). Select the vparquet4 dir `Ctrl+D` select the vparquet5 dir. This can be very helpful when git shows all code as new code.

tempodb/encoding/vparquet4/schema.go
```diff
@@ -754,6 +754,10 @@ func parquetToProtoEvents(parquetEvents []Event, spanStartTimeNano uint64) []*v1
 	return protoEvents
 }
 
+func ParquetTraceToTempopbTrace(meta *backend.BlockMeta, parquetTrace *Trace) *tempopb.Trace {
```
stoewer, line 757: Why not just capitalize the first letter of `parquetTraceToTempopbTrace()`?

tempodb/encoding/vparquet5/block_findtracebyid.go
```diff
@@ -0,0 +1,306 @@
...
+		rowGroupMins := make([]common.ID, numRowGroups+1)
+		// todo: restore using meta min/max id once it works
+		//    https://github.com/grafana/tempo/issues/1903
+		rowGroupMins[0] = bytes.Repeat([]byte{0}, 16)
+		rowGroupMins[numRowGroups] = bytes.Repeat([]byte{255}, 16) // This is actually inclusive and the logic is special for the last row group below
+
+		// Gets the minimum trace ID within the row group. Since the column is sorted
+		// ascending we just read the first value from the first page.
+		getRowGroupMin := func(rgIdx int) (common.ID, error) {
+			mn := rowGroupMins[rgIdx]
```
stoewer: `nitpick` I personally find `mn` and `mx` not very intuitive as a abbreviation for minimum, maybe `minRG` and `maxRG`?

tempodb/encoding/vparquet5/block_findtracebyid_test.go
```diff
@@ -0,0 +1,165 @@
...
+		wantProto := parquetTraceToTempopbTrace(meta, tr)
+
+		gotProto, err := b.FindTraceByID(ctx, tr.TraceID, common.DefaultSearchOptions())
+		require.NoError(t, err)
+		require.Equal(t, wantProto, gotProto.Trace)
+		require.Greater(t, gotProto.Metrics.InspectedBytes, uint64(60000)) // approximate value
+	}
+}
+
+/*func TestBackendBlockFindTraceByID_TestData(t *testing.T) {
```
stoewer: Just curious: why is this test now commented out?

tempodb/encoding/vparquet5/block_search.go
```diff
@@ -0,0 +1,487 @@
...
+	if len(resourceIters) == 1 {
+		traceIters = append(traceIters, resourceIters[0])
+	}
+	if len(resourceIters) > 1 {
+		traceIters = append(traceIters, pq.NewJoinIterator(DefinitionLevelTrace, resourceIters, nil))
+	}
+
+	// Duration filtering?
+	if req.MinDurationMs > 0 || req.MaxDurationMs > 0 {
+		mn := int64(0)
```
stoewer: `nitpick` Still not finding `mn` and `mx` very intuitive names. Maybe `lower`/`upper` or `start`/`end`?

tempodb/encoding/vparquet5/block_traceql.go
```diff
@@ -0,0 +1,3672 @@
...
+			}
+
+			if sampler != nil {
+				pred = newSamplingPredicate(sampler, pred)
+				// Removed so that it's not used down below.
+				sampler = nil
+			}
+
+			// Choose the least precise column possible.
+			// The step interval must be an even multiple of the pre-rounded precision.
```
stoewer, line 1925: I assume `cond.Precision` is the one calculated in `engine_metrics.go:L992`, right? I'm curious about the edge cases here: is it possible that `cond.Precision` is larger 1m but still falls through to the high precision column due to misalignment?

tempodb/encoding/vparquet5/block_traceql_test.go
```diff
@@ -0,0 +1,2301 @@
...
+
+	b := makeBackendBlockWithTraces(t, traces)
+	ctx := context.Background()
+	traceIDText := util.TraceIDToHexString(wantTraceID)
+
+	searchesThatMatch := []struct {
+		name string
+		req  traceql.FetchSpansRequest
+	}{
+		// {"empty request", traceql.FetchSpansRequest{}},
```
stoewer: Do we no longer need this test?

tempodb/encoding/versioned.go
```diff
@@ -72,6 +73,8 @@ func FromVersion(v string) (VersionedEncoding, error) {
 		return vparquet3.Encoding{}, nil
 	case vparquet4.VersionString:
 		return vparquet4.Encoding{}, nil
+	case vparquet5.VersionString:
```
stoewer, line 76: For the next PR: if someone creates a `vparquet5-preview2` and rolls it out to dev / ops where we could have blocks with `vparquet5-preview1`, Tempo would not be able to load those older preview versions. Should we maybe add a function `vparquet5.IsValidVersion(v string) bool` that returns true for all preexisting preview versions?

stoewer (commented):
Just a couple of minor comments and questions. Looks good otherwise :+1:

stoewer:
Now that we have a vparquet5 source tree, we should probably remind the team that any changes made in vparquet4 need to be replicated in vparquet5 as well

tempodb/encoding/versioned.go
mdisibio: I am working on a separate PR that will make Tempo gracefully handle unknown block versions.  Similar to how v2 doesn't support traceql and other new apis, but the remaining features are useable.   Blocks from old previews would be unqueryable, but otherwise harmless and eventually deleted from retention. Things like compaction and wal completion need to be skipped on them.  So it is more involved.

tempodb/encoding/vparquet4/schema.go
mdisibio: I think I was trying to reduce change in vp4 folder by not having to update all existing calls to it.  But I can go ahead and do this.

tempodb/encoding/vparquet5/block_findtracebyid.go
mdisibio: This was caught by the linter since all of vp5 looks like new code.  No strong opinions, will find something more readable for this and the comment below.

tempodb/encoding/vparquet5/block_findtracebyid_test.go
mdisibio: Thanks, is unintentional.  All of the testdata tests were commented out for a bit until the test-data block was created. Will fix.

tempodb/encoding/vparquet5/block_traceql.go
mdisibio: The engine now only sets Precision when the request is perfectly aligned:  https://github.com/grafana/tempo/pull/5495/files#diff-1b84537244dd74c3ba117a9181d6b27c2ba7d632b26c355ad5e5b428b6d00d19R993    Previously it only looked at the Step, which did have those gaps.

tempodb/encoding/vparquet4/schema.go
mdisibio: Exported the 1 method and cleaned up, for both vp4 and vp5.

tempodb/encoding/vparquet5/block_findtracebyid.go
mdisibio: Updated

tempodb/encoding/vparquet5/block_search.go
mdisibio: Updated

tempodb/encoding/vparquet5/block_traceql_test.go
mdisibio: Fixed

tempodb/encoding/versioned.go
stoewer: Handling unknown block versions is definitely a nice improvement. But that's not quite what I meant: If the next preview version e.g. `vparquet5-preview2` is built on top of `vparquet5-preview1` it would work just fine with `vparquet5-preview1` blocks, but just would not open them because of the string mismatch.

But that's of course something that could also be addressed in the vparquet5-preview2 PR

mdisibio: Agree, that is something to consider depending on the breaking nature of the column changes in the each preview version.

PR #5496, opened 2025-08-07
WIP: Tests and mock kafka implementation.

mattdurham:


mattdurham:
I hate I had to run the e2e tests 3 times for them to pass.

modules/livestore/kafka_client_franz.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/go-kit/log"
+	"github.com/grafana/tempo/pkg/ingest"
+	"github.com/twmb/franz-go/pkg/kadm"
+	"github.com/twmb/franz-go/pkg/kgo"
+	"github.com/twmb/franz-go/plugin/kprom"
+)
+
+// FranzKafkaClient wraps a kgo.Client and admin client to implement the KafkaClient interface
+type FranzKafkaClient struct {
```
joe-elliott: franz kafka was the writer and has nothing to do with kafka the queue technology :). probably should rename this.

modules/livestore/kafka_client_franz.go
```diff
@@ -0,0 +1,71 @@
...
+	client  *kgo.Client
+	adminCl *kadm.Client
+}
+
+// NewFranzKafkaClient creates a new Franz/kgo-based Kafka client
+func NewFranzKafkaClient(cfg ingest.KafkaConfig, metrics *kprom.Metrics, logger log.Logger) (KafkaClient, error) {
```
joe-elliott: surprised we need this. aren't we already consuming traces from kafka in the block builder/metrics generator. do we need a new client?

modules/livestore/kafka_client_inmemory.go
```diff
@@ -0,0 +1,329 @@
...
+	"github.com/grafana/tempo/pkg/ingest"
+	"github.com/twmb/franz-go/pkg/kadm"
+	"github.com/twmb/franz-go/pkg/kgo"
+	"github.com/twmb/franz-go/plugin/kprom"
+)
+
+// InMemoryMessage represents a message in the in-memory queue
+// Technically there is a fake cluster that listens on a port that is used for other tests.
+// But that has more weight and ties us to actually using Kafka instead of using something similar at a go api level.
+type InMemoryMessage struct {
```
joe-elliott: this is a mock? testing code? we've placed other testing code here and maybe it belongs there to be useful elsewhere?

https://github.com/grafana/tempo/tree/main/pkg/util/test

again curious what we did in live store/metrics generator given we've not needed this til now.

modules/livestore/live_store.go
```diff
@@ -52,8 +51,9 @@ type LiveStore struct {
 	ingestPartitionID         int32
 	ingestPartitionLifecycler *ring.PartitionInstanceLifecycler
 
-	client  *kgo.Client
-	decoder *ingest.Decoder
+	client        KafkaClient
+	clientFactory KafkaClientFunc
```
joe-elliott: oof, generally would prefer not, but understand if it aids testing.

modules/livestore/kafka_client_franz.go
```diff
@@ -0,0 +1,71 @@
...
+func (c *FranzKafkaClient) RemoveConsumePartitions(partitions map[string][]int32) {
+	c.client.RemoveConsumePartitions(partitions)
+}
+
+func (c *FranzKafkaClient) PollFetches(ctx context.Context) kgo.Fetches {
+	return c.client.PollFetches(ctx)
+}
+
+func (c *FranzKafkaClient) Close() {
+	c.client.Close()
```
joe-elliott: close admin client?

modules/livestore/live_store.go
mattdurham: I could probably get rid of the javaesque ClientFunc, but it needs some items from the constructor. I am more than happy to revisit this once we get it rolled out and refined a bit more.

modules/livestore/kafka_client_inmemory.go
mattdurham: Good place for it. Let me bounce it there.

mattdurham: Previously used kfake which simulates kafka over the wire. By moving to a golang api it allows non kafka backed implementations for more flexibility. Especially given the communities preference in single binary to reduce external dependencies.

modules/livestore/live_store.go
```diff
@@ -71,19 +71,20 @@ type LiveStore struct {
...
+		cfg:           cfg,
+		logger:        logger,
+		reg:           reg,
+		decoder:       ingest.NewDecoder(),
+		ctx:           ctx,
+		cancel:        cancel,
+		instances:     make(map[string]*instance),
+		overrides:     overrides,
+		flushqueues:   flushqueues.NewPriorityQueue(metricCompleteQueueLength),
+		clientFactory: clientFactory,
```
joe-elliott: does this need to be a struct var?

modules/livestore/kafka_client_franz.go
```diff
@@ -0,0 +1,71 @@
...
+// FranzKafkaClient wraps a kgo.Client and admin client to implement the KafkaClient interface
+type FranzKafkaClient struct {
+	client  *kgo.Client
+	adminCl *kadm.Client
+}
+
+// NewFranzKafkaClient creates a new Franz/kgo-based Kafka client
+func NewFranzKafkaClient(cfg ingest.KafkaConfig, metrics *kprom.Metrics, logger log.Logger) (KafkaClient, error) {
+	// Create client with topic consumption configured
+	opts := []kgo.Opt{
```
joe-elliott: why are we adding these? the ingest.NewReaderClient was called with default opts before?

modules/livestore/live_store.go
joe-elliott: no that's fine. i think we can/should remove it from the struct but the factory pattern is nbd

modules/livestore/kafka_client_franz.go
mattdurham: It doesnt need it here removed.

modules/livestore/live_store.go
mattdurham: I can bounce it out and create the client during the constructor instead of on startup.

PR #5503, opened 2025-08-10
remove hardcoded limit of 100 examplers

### LINKED ISSUE #5166: Use query_frontend.metrics.max_exemplars instead of hardcoded limit
`query_frontend.metrics.max_exemplars` limits the maximum number of returned exemplars. Currently, we use it here: https://github.com/grafana/tempo/blob/17e20a43af0881ef0841d2e33eaeb190c422e549/modules/frontend/metrics_query_range_sharder.go#L112-L117

But we also have have hardcoded limit here: https://github.com/grafa a/tempo/blob/17e20a43af0881ef0841d2e33eaeb190c422e549/pkg/traceql/engine_metrics.go#L25 . As a result, when `query_frontend.metrics.max_exemplars` < maxExemplars, it works as intended, but if it is more, then it will be capped to 100.

As `maxExemplars` is an implementation-specific variable, it is better to remove it and replace it with the configuration variable.

KyriosGN0:

**What this PR does**:

This PR removed the hardcoded limit of 100 examplers in traceql

Fixes #5166

**Checklist**
- [X] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,6 @@
 * [FEATURE] Add SSE-C encryption support to S3 backend [#5789](https://github.com/grafana/tempo/pull/5789) (@steffsas)
 * [FEATURE] Add support for `Accept: application/vnd.grafana.llm` to Tempo endpoints. Currently supported directly by trace by id and tag values [#5961](https://github.com/grafana/tempo/pull/5961) (@joe-elliott)
   This response is subject to change and should not be relied on. It is intended for LLM consumption only. Even a fundamental change to its representation (yaml? markdown?) would not be considered breaking.
-* [FEATURE] add database_name_attributes config to service graph processor  [#5398](https://github.com/grafana/tempo/pull/5398) (@KyriosGN0)
 * [ENHANCEMENT] docs: Add explicit notes about authentication [#5735](https://github.com/grafana/tempo/pull/5735) (@electron0zero)
 * [ENHANCEMENT] On startup, first record for live store to consume is not older than two complete block timeouts [#5693](https://github.com/grafana/tempo/pull/5693) (@ruslan-mikhailov)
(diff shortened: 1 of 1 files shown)


modules/querier/querier_query_range.go
```diff
@@ -107,7 +107,7 @@ func (q *Querier) queryBlock(ctx context.Context, req *tempopb.QueryRangeRequest
 		timeOverlapCutoff = v
 	}
 
-	eval, err := traceql.NewEngine().CompileMetricsQueryRange(req, int(req.Exemplars), timeOverlapCutoff, unsafe)
```
javiermolinar, line 103: Is this not breaking the possibility to define the maximum number of exemplars when doing a metrics query?

https://grafana.com/docs/tempo/latest/api_docs/#traceql-metrics

KyriosGN0: @javiermolinar in the `CompileMetricsQueryRange` function we just take the examplers from the `QueryRangeRequest` is this query not covered in the api E2E tests?
i will also spin up this build locally to test this endpoint later today and let you know

KyriosGN0: @javiermolinar i tested locally and this example works 
image

knylander-grafana:
Does this impact any configuration options that users can set?

KyriosGN0:
> Does this impact any configuration options that users can set?

No, it Should not

KyriosGN0:
will take care of this in the next couple of days

zhxiaogg:
Hello @KyriosGN0, thanks for the work to make this possible to happen. Just check in to see if you might still find time for it? I'm thinking to give it a try next.

KyriosGN0:
> Hello @KyriosGN0, thanks for the work to make this possible to happen. Just check in to see if you might still find time for it? I'm thinking to give it a try next.

hey @zhxiaogg, probably not, feel free to try it

PR #5507, opened 2025-08-11
[Bugfix] TraceQL Metrics fix for edge case

ruslan-mikhailov:
**What this PR does**: Fixed an edge case bug: spans in a block with a single trace could not be queried if span's timestamp is rounded to seconds.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/engine_metrics.go
+++ b/pkg/traceql/engine_metrics.go
@@ -136,7 +136,9 @@ func IntervalOfMs(tsmills int64, start, end, step uint64) int {
 func TrimToBlockOverlap(start, end, step uint64, blockStart, blockEnd time.Time) (uint64, uint64, uint64) {
 	wasInstant := end-start == step
 
-	start2 := uint64(blockStart.UnixNano())
+	// We subtract 1 nanosecond from the block's start time
+	// to make sure we include the left border of the block.
+	start2 := uint64(blockStart.UnixNano()) - 1
 	// Block's endTime is rounded down to the nearest second and considered inclusive.
 	// In order to include the right border with nanoseconds, we add 1 second
 	blockEnd = blockEnd.Add(time.Second)


pkg/traceql/engine_metrics.go
```diff
@@ -136,7 +136,9 @@ func IntervalOfMs(tsmills int64, start, end, step uint64) int {
 func TrimToBlockOverlap(start, end, step uint64, blockStart, blockEnd time.Time) (uint64, uint64, uint64) {
 	wasInstant := end-start == step
 
-	start2 := uint64(blockStart.UnixNano())
+	// We subtract 1 nanosecond from the block's start time
+	// to make sure we include the left border of the block.
+	start2 := uint64(blockStart.UnixNano()) - 1
```
ruslan-mikhailov, line 141: Is it possible to have block's start time 0? In that case, it will have overflow here. Not sure, the case worths handling.

joe-elliott: i think the answer is technically yes, but it would be quite difficult.

I would guess that a block start of 0 is more likely related to a bug then actually starting a block on second 0 of the unix epoch.

pkg/traceql/engine_metrics_test.go
```diff
@@ -221,21 +221,21 @@ func TestTrimToBlockOverlap(t *testing.T) {
...
+			"2024-01-01 01:30:59", "2024-01-01 02:01:00", 5 * time.Minute,
 		},
 		{
 			// Instant query
 			// Original range is 1h
 			// Inner overlap is only 30m and step is updated to match
 			"2024-01-01 01:00:00", "2024-01-01 02:00:00", time.Hour,
 			"2024-01-01 01:30:00", "2024-01-01 02:30:00",
-			"2024-01-01 01:30:00", "2024-01-01 02:00:00", 30 * time.Minute,
+			"2024-01-01 01:29:59", "2024-01-01 02:00:00", 30*time.Minute + time.Nanosecond,
```
mdisibio, line 238: Hmm this was unexpected, but it does seem correct.   01:30-02:00, is only 30 minutes when half open half closed borders.  Since we want the block start to be included, it really is + 1 ns.

ruslan-mikhailov: Plus one ns I believe comes from changing step value for an instant query.

PR #5514, opened 2025-08-13
[bugfix] TraceQL Metrics: panic in compare

ruslan-mikhailov:
**What this PR does**: fixes panic in TraceQL Metrics for Grafana Traces Drilldown query `{nestedSetParent>0 && true} | compare({})`

The bug is introduced by the following change https://github.com/grafana/tempo/pull/5419/files#diff-eb271e624dd1b0e372f056c46bbf31ed659448234bad1c80d87ccfe6e7a60f60L224-R236
Before the change, if a value not found under the given key, it would use empty struct, while here it becomes a nil pointer, which results in panic in `values[v.MapKey()].counts`

**How to reproduce the bug**:
Test case 1:

1. Run `examples/docker-compose/local`
2. In grafana (localhost:3000), open Traces Drilldown
3. Click on Duration histogram
4. Open Comparison Tab

Expected result: no error
Actual result: error `Query error An error occurred in the query`

Test case 2:
1. Run `examples/docker-compose/local` 
2. Execute query `{} | compare({})` in `/api/metrics/query_range`

Expected result: no error, returns 200 status code
Actual result: application crashed, panic in logs

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/engine_metrics_compare.go
+++ b/pkg/traceql/engine_metrics_compare.go
@@ -234,6 +234,7 @@ func (m *MetricsCompare) result(multiplier float64) SeriesSet {
 	}
 
 	addValues := func(prefix Label, data map[Attribute]map[StaticMapKey]*staticWithCounts) {
+		var value *staticWithCounts
 		for a, values := range data {
 			// Compute topN values for this attribute
 			top.reset()
@@ -242,10 +243,15 @@ func (m *MetricsCompare) result(multiplier float64) SeriesSet {
 			}
 
 			top.get(m.topN, func(v Static) {
+				value = values[v.MapKey()]
+				var counts []float64
+				if value != nil {
+					counts = value.counts
+				}
 				add(Labels{
 					prefix,
 					{Name: a.String(), Value: v},
-				}, values[v.MapKey()].counts)
+				}, counts)
 			})
 
 			if len(values) > m.topN {


pkg/traceql/engine_metrics_compare.go
```diff
@@ -242,10 +243,15 @@ func (m *MetricsCompare) result(multiplier float64) SeriesSet {
 			}
 
 			top.get(m.topN, func(v Static) {
+				value = values[v.MapKey()]
```
javiermolinar, line 246: Cannot we just declare the variable here?

```go
value := values[v.MapKey()]
```

It would be great to have a test for this scenario

ruslan-mikhailov: It's a small optimisation to avoid additional allocations.

The problem with that approach that if value assignment goes under if condition, we might accidentally use value from the previous iteration. As value assignment and usage are very close and scoped, I decided to leave it.

mdisibio, line 246: Can we dig into understanding this a bit more?  How can a value be in the topN but not in the map? Since we just populated topN from `values` in the previous loop.

ruslan-mikhailov:
The root cause is fixed here: https://github.com/grafana/tempo/pull/5519

PR #5519, opened 2025-08-14
[bugfix] TraceQL Metrics: panic in compare

ruslan-mikhailov:
**What this PR does**: 
This PR fixes origin issue that was unnoticed until it introduced a panic. Explanation why it panics can be found here: https://github.com/grafana/tempo/pull/5514

**The bug explanation.**
https://go.dev/play/p/P78nVSN44BP
The reason the value's changed in both slices is because they both point to the same array.

If we have two slices that point to one array:
image

then we append to the first slice:
image

The same happens here.
1. We init struct with buffer [here](https://github.com/ruslan-mikhailov/tempo/blob/c73964fc471cdc2d10edb2f268e9bf90e9fce337/tempodb/encoding/vparquet4/block_traceql.go#L3337)
1. We reset buffer and append [here](https://github.com/ruslan-mikhailov/tempo/blob/c73964fc471cdc2d10edb2f268e9bf90e9fce337/tempodb/encoding/vparquet4/block_traceql.go#L3298-L3323)
1. We use the slice [here](https://github.com/ruslan-mikhailov/tempo/blob/c73964fc471cdc2d10edb2f268e9bf90e9fce337/pkg/traceql/ast.go#L755)

As the object is created per request, it does not expose data from other tenants, it just can mix attributes values from queried data.

In the end, at the moment request is processing [this line](https://github.com/ruslan-mikhailov/tempo/blob/c73964fc471cdc2d10edb2f268e9bf90e9fce337/pkg/traceql/engine_metrics_compare.go#L240), the value under the struct's already changed and calling `v.MapKey()` differs, which results in incorrect calculations.

**Note:** this change will increase number of allocated objects and increase pressure to GC which can lead to increased resource usage.

**How to reproduce the bug**:
Test case 1:

1. Run `examples/docker-compose/local`
2. In grafana (localhost:3000), open Traces Drilldown
3. Click on Duration histogram
4. Open Comparison Tab

Expected result: no error
Actual result: error `Query error An error occurred in the query`

Test case 2:
1. Run `examples/docker-compose/local` 
2. Execute query `{} | compare({})` in `/api/metrics/query_range`

Expected result: no error, returns 200 status code
Actual result: application crashed, panic in logs

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/util/slices.go
+++ b/pkg/util/slices.go
@@ -1,0 +1,12 @@
+package util
+
+// Clone returns a copy of the slice.
+// The elements are copied using assignment, so this is a shallow clone.
+// The difference with slices.Clone is that the result will not have additional unused capacity.
+func Clone[S ~[]E, E any](s S) S {
+	// Preserve nilness in case it matters.
+	if s == nil {
+		return nil
+	}
+	return append(make(S, 0, len(s)), s...)
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ Additionally the `compaction_tenant_backoff_total` metric has been renamed to `c
 * [ENHANCEMENT] Make block ordering deterministic [#5411](https://github.com/grafana/tempo/pull/5411) (@rajiv-singh)
 * [BUGFIX] Fix race condition between compaction provider and backend-scheduler [#5409](https://github.com/grafana/tempo/pull/5409) (@zalegrala)
 * [BUGFIX] Do not allow very small steps [#5441](https://github.com/grafana/tempo/pull/5441) (@ruslan-mikhailov)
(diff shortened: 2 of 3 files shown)


pkg/util/slices.go
```diff
@@ -0,0 +1,12 @@
+package util
+
+// Clone returns a copy of the slice.
+// The elements are copied using assignment, so this is a shallow clone.
+// The difference with slices.Clone is that the result will not have additional unused capacity.
```
ruslan-mikhailov, line 5: I forked the function from slices package to optimise memory usage:

https://go.dev/play/p/LhxS2IZAPC5

mdisibio (approved):

PR #5520, opened 2025-08-14
Support exemplars with native-only histograms

zalegrala:
**What this PR does**:

When native histograms were added to the metrics-generator, exemplars were not included in the feature-set.  Here we rectify that by adding support for sending native histograms to the remote-write endpoint.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/registry/native_histogram.go
+++ b/modules/generator/registry/native_histogram.go
@@ -140,13 +140,11 @@ func (h *nativeHistogram) newSeries(labelValueCombo *LabelValueCombo, value floa
 
 	var buckets []float64
 
+	// The native histogram only uses the static buckets when the classic histograms are enabled.
 	hasClassic := hasClassicHistograms(h.histogramOverride)
 	if hasClassic {
 		// Hybrid "both" mode: include classic buckets for compatibility
 		buckets = h.buckets
-	} else {
-		// native-only: no classic buckets to force native exemplar format
-		buckets = nil
 	}
 
 	// Configure native histogram options based on mode
@@ -205,9 +203,11 @@ func (h *nativeHistogram) updateSeries(s *nativeHistogramSeries, value float64, 
 	// Use Prometheus native exemplar handling
 	exemplarObserver := s.promHistogram.(prometheus.ExemplarObserver)
 
+	labels := prometheus.Labels{h.traceIDLabelName: traceID}
+
 	for i := 0.0; i < multiplier; i++ {
 		// Let Prometheus handle exemplars natively
-		exemplarObserver.ObserveWithExemplar(value, prometheus.Labels{h.traceIDLabelName: traceID})
+		exemplarObserver.ObserveWithExemplar(value, labels)
 	}
 
 	s.lastUpdated = time.Now().UnixMilli()
@@ -321,10 +321,8 @@ func (h *nativeHistogram) nativeHistograms(appender storage.Appender, lbls label
 	//
 	// Use native exemplars when available, falling back to bucket exemplars
 
-	nativeExemplarCount := 0
 	for _, ex := range s.histogram.Exemplars {
 		if ex != nil && len(ex.Label) > 0 {
-			nativeExemplarCount++
 			_, err = appender.AppendExemplar(ref, lbls, exemplar.Exemplar{
 				Labels: convertLabelPairToLabels(ex.GetLabel()),
 				Value:  ex.GetValue(),
@@ -340,7 +338,7 @@ func (h *nativeHistogram) nativeHistograms(appender storage.Appender, lbls label
 	// client_golang package handles the expiration of exemplars internally, and
 	// we don't have control over clearing the native histogram exemplars in the
 	// same way we do for the class histogram exemplars.
-	if nativeExemplarCount > 0 {
+	if len(s.histogram.Exemplars) > 0 {
 		clear(s.histogram.Exemplars)
 		s.histogram.Exemplars = s.histogram.Exemplars[:0]
 	}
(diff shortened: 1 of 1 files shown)


modules/generator/registry/native_histogram.go
```diff
@@ -301,11 +339,67 @@ func (h *nativeHistogram) nativeHistograms(appender storage.Appender, lbls label
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	// Clear all native exemplars after emission to prevent accumulation
+	// This should match the behavior of classic histograms: s.exemplars[i].Store("")
+	//
+	// FIXME: this is not working as expected.  It appears that the internal
```
zalegrala: Bug?

zalegrala: I see now that `client_golang` has a TTL which expires old exemplars.  Is this good enough for us to ignore the possibility of sending the same exemplar twice?  Posted in an SME channel and waiting for review.

modules/generator/registry/native_histogram_test.go
```diff
@@ -587,3 +587,71 @@ func expectedSeriesLen(samples []sample) int {
...
+		require.NoError(t, err)
+
+		t.Logf("Native-only - Active series: %d", activeSeries)
+		t.Logf("Native-only - Captured samples: %d", len(appender.samples))
+		t.Logf("Native-only - Captured exemplars: %d", len(appender.exemplars))
+
+		// This test might still use bucket exemplars due to Prometheus's default behavior,
+		// but it helps us understand the difference
+		for i, ex := range appender.exemplars {
+			t.Logf("Native-only exemplar %d: labels=%v, value=%f, trace=%v", i, ex.l, ex.e.Value, ex.e.Labels)
```
zalegrala, line 664: Let's validate the contents here, not just print it out.

modules/generator/registry/native_histogram.go
```diff
@@ -179,12 +202,14 @@ func (h *nativeHistogram) newSeries(labelValueCombo *LabelValueCombo, value floa
 }
 
 func (h *nativeHistogram) updateSeries(s *nativeHistogramSeries, value float64, traceID string, multiplier float64) {
+	// Use Prometheus native exemplar handling
+	exemplarObserver := s.promHistogram.(prometheus.ExemplarObserver)
```
mdisibio, line 204: It looks like this is the only reference to `s.promHistogram`, could we instead cast it once in `newSeries` ?

modules/generator/registry/native_histogram.go
```diff
@@ -133,19 +133,42 @@ func (h *nativeHistogram) ObserveWithExemplar(labelValueCombo *LabelValueCombo,
...
+
+	var buckets []float64
+
+	hasClassic := hasClassicHistograms(h.histogramOverride)
+	if hasClassic {
+		// Hybrid "both" mode: include classic buckets for compatibility
+		buckets = h.buckets
+	} else {
+		// native-only: no classic buckets to force native exemplar format
+		buckets = nil
```
mdisibio: This line isn't necessary. Remove and combine/clarify the comments?

modules/generator/registry/native_histogram.go
```diff
@@ -179,12 +202,14 @@ func (h *nativeHistogram) newSeries(labelValueCombo *LabelValueCombo, value floa
...
+	// Use Prometheus native exemplar handling
+	exemplarObserver := s.promHistogram.(prometheus.ExemplarObserver)
+
 	for i := 0.0; i < multiplier; i++ {
-		s.promHistogram.(prometheus.ExemplarObserver).ObserveWithExemplar(
-			value,
-			map[string]string{h.traceIDLabelName: traceID},
-		)
+		// Let Prometheus handle exemplars natively
+		exemplarObserver.ObserveWithExemplar(value, prometheus.Labels{h.traceIDLabelName: traceID})
```
mdisibio: I know it was like this already, but should we move the labels out of the loop too?

modules/generator/registry/native_histogram.go
```diff
@@ -197,6 +222,7 @@ func (h *nativeHistogram) collectMetrics(appender storage.Appender, timeMs int64
 	defer h.seriesMtx.Unlock()
 
 	activeSeries = 0
+	var totalNativeExemplars, totalBucketExemplars, totalSeries int
```
mdisibio: Are these used? I see them incremented only.

modules/generator/registry/native_histogram.go
```diff
@@ -301,11 +319,64 @@ func (h *nativeHistogram) nativeHistograms(appender storage.Appender, lbls label
...
+	// we don't have control over clearing the native histogram exemplars in the
+	// same way we do for the class histogram exemplars.
+	if nativeExemplarCount > 0 {
+		clear(s.histogram.Exemplars)
+		s.histogram.Exemplars = s.histogram.Exemplars[:0]
+	}
+
+	// For pure native mode, never emit bucket exemplars - only native ones
+	// For hybrid mode, fallback to bucket exemplars if no native exemplars available
+	bucketExemplarCount := 0
```
mdisibio: Is `bucketExemplarCount` used? I see it incremented only.

modules/generator/registry/native_histogram.go
```diff
@@ -301,11 +319,64 @@ func (h *nativeHistogram) nativeHistograms(appender storage.Appender, lbls label
...
+				return err
+			}
+		}
+	}
+
+	// NOTE: We clear the native exemplar slice to prevent accumulation, but the
+	// client_golang package handles the expiration of exemplars internally, and
+	// we don't have control over clearing the native histogram exemplars in the
+	// same way we do for the class histogram exemplars.
+	if nativeExemplarCount > 0 {
```
mdisibio: It might be better to have this check `len(s.histogram.Exemplars) > 0` instead, because `nativeExemplarCount` is only incremented if Label is present. Not sure if that matters, but this way it would always reset a non-empty slice.

modules/generator/registry/native_histogram.go
zalegrala: Its also called at collection in `collectMetrics`, which needs the `Write` method that is not on the `ExemplarObserver` interface.  It might get a little messy.

modules/generator/registry/native_histogram.go
zalegrala: Good call, this is cleaner.

mdisibio (approved):

PR #5524, opened 2025-08-15
Live store ring/other goodies

joe-elliott:
Adds support for a live store ring and wires up the complete read and write path for live stores.

Done:
- New root level config option to use the live stores to drive the partition ring
- New distributor config option stop writing to ingesters over gRPC
- ingest-storage docker-compose has ingesters completely removed. it is technically rhythm in its final form (but full of bugs)
- copied in mimir partition ring query code

Todo:
- Should the live stores use a new partition ring name or continue with the ingester one? This will likely be decided by how we orchestrate the migration. Currently using the ingester ring name/key.
- **NOTE** querying is still wonky due to a bug in live store kafka consumption where it 
- Confirm live store ring querying behavior is sane
- Add config options that the ring exposes for partition ring querying


cmd/tempo/app/modules.go
```diff
@@ -837,7 +861,7 @@ func (t *App) setupModuleManager() error {
 		Ingester:                      {Common, Store, MemberlistKV, PartitionRing},
 		MetricsGenerator:              {Common, OptionalStore, MemberlistKV, PartitionRing},
 		MetricsGeneratorNoLocalBlocks: {Common, GeneratorRingWatcher},
-		Querier:                       {Common, Store, IngesterRing, MetricsGeneratorRing, SecondaryIngesterRing},
+		Querier:                       {Common, Store, IngesterRing, MetricsGeneratorRing, SecondaryIngesterRing, PartitionRing},
```
mattdurham, line 864: Is there any harm in having both the ingester/partition ring here? I suppose if we want dual checking read/write we will need it to be both available. Though we would need more code changes above, probably not a concern at the moment

modules/distributor/config.go
```diff
@@ -42,7 +42,8 @@ type Config struct {
 	Forwarders          forwarder.ConfigList      `yaml:"forwarders"`
 	Usage               usage.Config              `yaml:"usage,omitempty"`
 
-	// Kafka
+	// Migration to Kafka write path
```
mattdurham, line 45: No totally in love with this name but so is the life of bikeshedding, thoughts on `ingest_write_path_enabled`, seems more straightforward?

joe-elliott: mimir (and us) use "ingest" often to mean the new kafka write path so i'd steer clear of that.

`ingester_write_path_enabled` ?

no issues with renaming. this config option will be removed as soon as the rhythm migration is complete.

cmd/tempo/app/modules.go
joe-elliott: nope. no harm.

the querier needs the ingester and generator read rings for the current query path and the partition ring for the new shiny rhythm query path. once we are migrated we can remove the ingester and generator rings.

modules/livestore/live_store.go
```diff
@@ -26,6 +28,12 @@ import (
 
 const (
 	liveStoreServiceName = "live-store"
+
+	ringNumTokens = 1 // we only need 1 token in the read ring per instance b/c it's for service discovery only. sharding is done with the parttition ring.
```
mattdurham, line 32: Can you explain this to my newb brain? I assume its meant to handle ownership in the ring, but how many total tokens are there?

joe-elliott: i'm not totally convinced this is the right way to do this and have a followup todo to review how mimir uses these rings. what this is doing is creating a ring that is used for discovery only. all we need is a way to map instance name -> instance address.

since we don't actually shard anything with this ring we are adding only 1 token to reduce the overhead of propagating it. i have a hunch there is a "more correct" way to do this which is why i want to dig deeper in the next week or so

modules/livestore/live_store.go
```diff
@@ -419,3 +465,124 @@ func (s *LiveStore) completeLoop() {
 		}
 	}
 }
+
+// OnRingInstanceRegister implements ring.BasicLifecyclerDelegate
+func (s *LiveStore) OnRingInstanceRegister(_ *ring.BasicLifecycler, _ ring.Desc, _ bool, _ string, _ ring.InstanceDesc) (ring.InstanceState, ring.Tokens) {
+	// tokens don't matter for the livestore ring, we just need to be in the ring for service discovery
+	token := rand.Uint32() //nolint: gosec // G404 we don't need a cryptographic random number here
```
mattdurham, line 472: Is token here for uniqueness? Does this link to ringnumtoken = 1, ie if it was two we would pass in two tokens on the return?

joe-elliott: yes, part of the same set of questions as above.

mattdurham: Long term should this be a dynamically done via the number of ringnumtokens, so that it is more "associated" between the two concepts. IE `for ringnumtokens build random array`

joe-elliott: no, b/c this ring never needs tokens. it's really only here to map instance name -> address. things to test/research:

- can we just return 0 tokens?
- is there a specialized ring i'm unaware of that does this mapping and has no concept of tokens?

PR #5539, opened 2025-08-18
Remove risky unsafe in regexp

carles-grafana:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/regexp/regexp.go
+++ b/pkg/regexp/regexp.go
@@ -2,7 +2,6 @@ package regexp
 
 import (
 	"fmt"
-	"regexp"
 	"unsafe"
 
 	"github.com/prometheus/prometheus/model/labels"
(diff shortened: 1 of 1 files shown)


pkg/regexp/regexp.go
```diff
@@ -93,28 +92,8 @@ func (r *Regexp) String() string {
...
-
 	// TODO: research and improve this. we're making a guess on whether an optimization will improve the regex
 	// performance enough that its faster to not memoize. See compileMatchStringFunction() in the labels
 	// package. maybe there's even a way to do this dynamically?
-	return cheat.stringMatcher == nil && // a stringMatcher definitively rejects or accepts. if a string matcher is present the regex will never be executed
-		len(cheat.setMatches) == 0 && // setMatches definitively reject or accept. if len != 0 the regex will never be executed, but perhaps if there are a very large # of setMatches we prefer memoizing anyway?
-		cheat.prefix == "" && // prefix and suffix _do not_ prevent the regex from executing, but they are quick to evaluate and tend to nicely filter down.
-		cheat.suffix == "" // perhaps a length requirement would be an improvement? i.e. require a prefix or suffix of at least 3 chars?
-	// len(cheat.contains) == 0 // in testing, it was faster to memoize with a contains filter. perhaps if the filters are long enough we don't memoize?
+	return !m.IsOptimized()
```
carles-grafana, line 98:
```
func (m *FastRegexMatcher) IsOptimized() bool {
	return len(m.setMatches) > 0 || m.stringMatcher != nil || m.prefix != "" || m.suffix != "" || len(m.contains) > 0
}
```

the only difference is the `len(m.contains) > 0`

joe-elliott: please benchmark the change. the `shouldMemoize` filters were chosen purposefully b/c it had the best perf characteristics. `len(m.contains) > 0` was left out on purpose.

also, how risky is this?

PR #5540, opened 2025-08-18
Hook up query by range.

mattdurham:


modules/livestore/live_store.go
```diff
@@ -583,6 +584,58 @@ func (s *LiveStore) GetMetrics(_ context.Context, _ *tempopb.SpanMetricsRequest)
 }
 
 // QueryRange implements tempopb.MetricsGeneratorServer
-func (s *LiveStore) QueryRange(_ context.Context, _ *tempopb.QueryRangeRequest) (*tempopb.QueryRangeResponse, error) {
-	return nil, fmt.Errorf("QueryRange not implemented in livestore") // todo: implement. seems to be a disconnect between this and instance.QueryRange
+func (s *LiveStore) QueryRange(ctx context.Context, req *tempopb.QueryRangeRequest) (*tempopb.QueryRangeResponse, error) {
+	instanceID, err := user.ExtractOrgID(ctx)
```
joe-elliott, line 587: longer term it may be nice to push this into the instance if possible. all the other calls have are a very simple pass through except this one. not a blocker.

modules/livestore/config.go
```diff
@@ -17,18 +17,22 @@ type Config struct {
...
 	// This config is dynamically injected because defined outside the ingester config.
 	IngestConfig ingest.Config `yaml:"-"`
 
 	// WAL is non-configurable and only uses defaults
 	WAL wal.Config `yaml:"-"`
-}
 
-type MetricsConfig struct {
-	ConcurrentBlocks  uint    `yaml:"concurrent_blocks,omitempty"`
+	// TimeOverlapCutoff is a tuning factor that controls whether the trace-level
```
joe-elliott, line 31: apologies. reading the code this setting is specific to metrics. i thought this was something that was more universal.

concurrent blocks, however, could be applied to all endpoints by using the boundedwaitgroup. should we apply this config to all endpoints before merging?

PR #5541, opened 2025-08-18
Add FindByTraceID

mattdurham:
The logic on looping on the blocks makes sense to me, but def someone with more knowledge should give eyes. No tests since the  actual important code is tested in the ingester/local_block.go.

### DIFF
--- a/modules/livestore/instance.go
+++ b/modules/livestore/instance.go
@@ -367,53 +367,33 @@ func (i *instance) FindByTraceID(ctx context.Context, traceID []byte) (*tempopb.
 
 	// TODO MRD Add limits
 
-	// Loop over all the complete blocks looking for a specific ID. The implementation looks like it will return nil if the trace is not found.
-	for _, b := range i.completeBlocks {
+	loopBlock := func(b common.Finder) error {
 		trace, err := b.FindTraceByID(ctx, traceID, common.DefaultSearchOptions())
 		if err != nil {
-			return nil, err
+			return err
 		}
 		if trace == nil {
-			continue
+			return nil
 		}
 		_, err = combiner.Consume(trace.Trace)
 		if err != nil {
-			return nil, err
+			return err
 		}
 		if trace.Metrics != nil {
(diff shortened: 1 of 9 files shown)


modules/livestore/instance.go
```diff
@@ -334,3 +338,89 @@ func (i *instance) deleteOldBlocks() error {
...
+	defer i.blocksMtx.RUnlock()
+
+	var completeTrace *tempopb.Trace
+	metrics := tempopb.TraceByIDMetrics{}
+	combiner := trace.NewCombiner(math.MaxInt64, true)
+	_, err := combiner.Consume(completeTrace)
+	if err != nil {
+		return nil, err
+	}
+	// TODO MRD this code looks ripe for simplification. Its the same pattern repeated several times.
```
joe-elliott, line 353: the live traces section just below this is unique to FindTraceByID. the rest of it, however, could likely be consolidated. (check head block, loop over wal blocks, loop over complete blocks).

modules/livestore/instance.go
```diff
@@ -334,3 +338,89 @@ func (i *instance) deleteOldBlocks() error {
...
+		if err != nil {
+			i.liveTracesMtx.Unlock()
+			return nil, fmt.Errorf("unable to unmarshal liveTrace: %w", err)
+		}
+	}
+	i.liveTracesMtx.Unlock()
+
+	// TODO MRD Add limits
+
+	// Loop over all the complete blocks looking for a specific ID. The implementation looks like it will return nil if the trace is not found.
```
joe-elliott: why shouldn't we trust nil = not found?

joe-elliott (approved):

modules/livestore/instance.go
mattdurham: I can remove the comment, I am not checking for it thoroughly enough like I previously was. Combiner will handle if `trace.Trace` is nil.

modules/livestore/instance.go
mattdurham: Done with the loopblock. Good call!

PR #5542, opened 2025-08-18
Livestore metrics

ruslan-mikhailov:
**What this PR does**: adds metrics to live-store. Metrics are similar to ingester's

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/livestore/instance.go
+++ b/modules/livestore/instance.go
@@ -20,6 +20,49 @@ import (
 	"github.com/grafana/tempo/tempodb/encoding"
 	"github.com/grafana/tempo/tempodb/encoding/common"
 	"github.com/grafana/tempo/tempodb/wal"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
+	"go.opentelemetry.io/otel/attribute"
+	oteltrace "go.opentelemetry.io/otel/trace"
+)
+
+const traceDataType = "trace"
+
+var (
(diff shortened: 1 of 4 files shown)


modules/livestore/instance.go
```diff
@@ -147,6 +204,13 @@ func (i *instance) cutIdleTraces(immediate bool) error {
 	return nil
 }
 
+func (i *instance) measureReceivedBytes(traceBytes []byte) {
```
joe-elliott: is it worth having a func to do this one liner?

modules/livestore/instance.go
```diff
@@ -315,6 +397,7 @@ func (i *instance) deleteOldBlocks() error {
 				return err
 			}
 			delete(i.walBlocks, id)
+			metricBlocksClearedTotal.Inc()
```
joe-elliott: this is a weird case, can we add a "type=wal|complete". i'd be interested if this line ever hits

joe-elliott (approved):
two Qs, but lgtm

modules/livestore/instance.go
ruslan-mikhailov: I just copied it from ingester, but I'm ok to remove it

ruslan-mikhailov: \+

modules/livestore/instance.go
ruslan-mikhailov: \+

joe-elliott (approved):

PR #5543, opened 2025-08-18
Gracefully handle unsupported block encodings

mdisibio:
**What this PR does**:
Currently many parts of Tempo will fail with an error when encountering an unknown block encoding. In order to iterate on more changes after #5495 , this adds more graceful handling. Tempo is now able to start, and skip over blocks of unknown format without impacting operations of other supported blocks. 

It does this by creating a new "Unsupported" encoding which is used for any unknown block version string with one restriction:  the version must include the word "preview".  This is safeguard against possibly deleting unrelated things from WAL and object storage.

> [!NOTE]
> Lots of churn from deduplicating `ErrUnsupported` from the common package, and moving towards `pkg/util/errors.go`

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/blockselector/compaction_block_selector_test.go
+++ b/tempodb/blockselector/compaction_block_selector_test.go
@@ -915,10 +915,12 @@ func TestTimeWindowBlockSelectorBlocksToCompact(t *testing.T) {
 					b.Version = encoding.DefaultEncoding().Version()
 				}
 			}
-
-			// For testing, assign all data to valid encoding
-			// if not specified.
-			for _, b := range tt.blocklist {
+			for _, b := range tt.expected {
+				if b.Version == "" {
+					b.Version = encoding.DefaultEncoding().Version()
+				}
+			}
+			for _, b := range tt.expectedSecond {
 				if b.Version == "" {
 					b.Version = encoding.DefaultEncoding().Version()
 				}


tempodb/blockselector/compaction_block_selector.go
```diff
@@ -59,6 +60,15 @@ func NewTimeWindowBlockSelector(blocklist []*backend.BlockMeta, maxCompactionRan
 	activeWindow := twbs.windowForTime(now.Add(-activeWindowDuration))
 
 	for _, b := range blocklist {
+
+		enc, err := encoding.FromVersion(b.Version)
+		if err != nil {
+			continue
+		}
+		if !enc.CompactionSupported() {
```
mdisibio, line 71: Not 100% settled on this. It seems a bit over-complicated, and open to suggestions.  Returning ErrUnsupported from `encoding.NewCompactor()` is more straightforward, but too late in the process. The blocks would cause overhead in the backend scheduler/worker/etc, and show in outstanding block metrics.

tempodb/wal/wal.go
```diff
@@ -83,7 +84,10 @@ func (w *WAL) RescanBlocks(additionalStartSlack time.Duration, log log.Logger) (
 		return nil, err
 	}
 
-	encodings := encoding.AllEncodings()
+	// For wal ownership, add Unsupported to the list at the end.
+	// It will catch any removed preview encodings.
+	encodings := append(encoding.AllEncodings(), unsupported.Encoding{})
```
mdisibio, line 89: This could use a bit more cleanup.  We want to allow Unsupported to take ownership of a WAL block and allow it to be deleted, but adding it to the global list of `AllEncodings()` has massive impacts to tests and validation.

mattdurham:
Nice markdown!

tempodb/wal/wal.go
stoewer: I agree, adding the unsupported encoding to `AllEncodings()` is probably not desirable

tempodb/blockselector/compaction_block_selector.go
stoewer: I think adding `CompactionSupported()` is fine - even if we will likely never have a production ready encoding  where `CompactionSupported()` returns false

stoewer:
Looks good to me so far :+1:

PR #5544, opened 2025-08-18
Fix duplicate ingestion in live-store

joe-elliott:
- Traces were being duplicated in livestore b/c the decoder was not being reset. Fixed.
- Removed 10 second out of band offset commit and just committed synchronously

### DIFF
--- a/modules/livestore/partition_reader.go
+++ b/modules/livestore/partition_reader.go
@@ -122,7 +122,7 @@ func collectFetchErrs(fetches kgo.Fetches) (_ error) {
 }
 
 func (r *PartitionReader) consumeFetches(ctx context.Context, fetches kgo.Fetches) *kadm.Offset {
-	records := make([]record, 0, len(fetches.Records()))
+	records := make([]record, 0, fetches.NumRecords())
 
 	var lastRecord *kgo.Record
 


modules/livestore/live_store.go
```diff
@@ -271,6 +271,8 @@ func (s *LiveStore) flushRemaining() {
 }
 
 func (s *LiveStore) consume(_ context.Context, rs []record) error {
+	s.decoder.Reset()
```
joe-elliott: this was the real fix

mattdurham: Does this work? We are reusing the decoder on each fetch from RS?

mattdurham: I guess it does because we never reuse pushReq in any way after the push bytes, but it feels fragile.

joe-elliott: yup, it's fine. the BB does the same thing. this was just missing the "Reset" call. it's done to reuse buffers.

mattdurham: Thoughts on making it a defer so that the decoder is in a ready to use position after the consume is done? Right now we dont use it anywhere but it feels easy to mess up.

joe-elliott: done. also reset in the loop. which is where it should have been.

PR #5549, opened 2025-08-19
[bugfix] Fix cache collision for incomplete query in SearchTagValuesV2

ruslan-mikhailov:
**What this PR does**: flaky test `TestSearchTagValuesV2/second_batch_with_incomplete_query_-_name` detected a bug in SearchTagValuesV2: https://github.com/grafana/tempo/actions/runs/17066085086/job/48384561937?pr=5542
When queries are incomplete, they all generate the same cache key which results in cache collision.

**How to test:** run e2e test `TestSearchTagValuesV2` with `-count=10`

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ Additionally the `compaction_tenant_backoff_total` metric has been renamed to `c
 * [BUGFIX] Fix race condition between compaction provider and backend-scheduler [#5409](https://github.com/grafana/tempo/pull/5409) (@zalegrala)
 * [BUGFIX] Do not allow very small steps [#5441](https://github.com/grafana/tempo/pull/5441) (@ruslan-mikhailov)
 * [BUGFIX] Fix incorrect results in TraceQL compare() for spans with array attributes [#5519](https://github.com/grafana/tempo/pull/5519) (@ruslan-mikhailov)
+* [BUGFIX] Fix cache collision for incomplete query in SearchTagValuesV2 [#5549](https://github.com/grafana/tempo/pull/5549) (@ruslan-mikhailov)
 
 # v2.8.2
 


modules/ingester/instance_search.go
```diff
@@ -606,23 +606,23 @@ func includeBlock(b *backend.BlockMeta, req *tempopb.SearchRequest) bool {
 // searchTagValuesV2CacheKey generates a cache key for the searchTagValuesV2 request
 // cache key is used as the filename to store the protobuf data on disk
 func searchTagValuesV2CacheKey(req *tempopb.SearchTagValuesRequest, limit int, prefix string) string {
```
ruslan-mikhailov, line 608: Note to myself @ruslan-mikhailov : need to copy the solution to live-trace in `main-new-ingester` branch

modules/ingester/instance_search.go
```diff
@@ -606,23 +606,23 @@ func includeBlock(b *backend.BlockMeta, req *tempopb.SearchRequest) bool {
...
-		} else {
-			// In case of a bad TraceQL query, we ignore the query and return unfiltered results.
-			// if we fail to parse the query, we will assume query is empty and compute the cache key.
-			query = ""
+			cacheKey = ast.String()
+		} else { // if parsing fails (e.g., incomplete/invalid TraceQL), fall back to extracted matchers
+			extracted := traceql.ExtractMatchers(req.Query)
+			if !traceql.IsEmptyQuery(extracted) {
+				cacheKey = extracted
+			}
```
ruslan-mikhailov: As you can see from tests, the solution is not perfect: for some queries that should hit the same key, keys are different.

Maybe there is a better way to fix it?

joe-elliott: looking at the other tag values paths it looks like we can safely do something like:

```
cacheKey := traceql.ExtractMatchers(req.query)
```

i don't think we need to call parse at all and it would be more consistent with handling of this query param elsewhere. also wondering if we need to call `IsEmptyQuery`

> for some queries that should hit the same key, keys are different.

is this due to ordering differences? like:
```
{ span.foo = "bar" && span.baz = "bat" && span.qux = }
```
will have a different key than:
```
{ span.baz = "bat" && span.foo = "bar" && span.qux = }
```

if so we could have extract matchers return ordered keys as long as it doesn't break things elsewhere. we should note that in comments and pin the behavior with a test so its known why.

ruslan-mikhailov: Discussed internally. The difference comes from how `ExtractMatchers` works. 

Changed it in the following way:
1. ExtractMatchers
2. Parse the resulted query and use canonical form

modules/ingester/instance_search.go
```diff
@@ -606,23 +606,21 @@ func includeBlock(b *backend.BlockMeta, req *tempopb.SearchRequest) bool {
...
+		q := traceql.ExtractMatchers(req.Query)
+		if ast, err := traceql.Parse(q); err == nil {
 			// forces the query into a canonical form
-			query = ast.String()
-		} else {
-			// In case of a bad TraceQL query, we ignore the query and return unfiltered results.
-			// if we fail to parse the query, we will assume query is empty and compute the cache key.
-			query = ""
+			cacheKey = ast.String()
+		} else { // if parsing fails, fall back to extracted matchers
```
ruslan-mikhailov: `ExtractMatchers` can return invalid queries, for example, `{ < not valid traceql > }` -> `{ < not}`

joe-elliott (approved):

PR #5552, opened 2025-08-20
Rhythm/Live Store: Wal replay

joe-elliott:
**What this PR does**:
Adds wal replay to the live store + other things:

- Adds a config flag `holdAllBackgroundProcesses` to prevent block lifecycle processes from running. I normally don't like config like this as it complicates normal logic, but it's extremely difficult to test block lifecycle if these processes are running.
- Made some helper functions start/stop background processes to give some structure to the 3 background processes and related struct vars.
- Fixed 2 bugs in deleteOldBlocks: it was adding complete block timeout to create a future cut off time and it was checking to see if complete blocks had been flushed before deleting.
- Added tests for block lifecycle
- Renamed the "Concurrent" config option "QueryBlockConcurrency" to distinguish between it and "CompleteBlockConcurrency"

### DIFF
--- a/modules/livestore/live_store_test.go
+++ b/modules/livestore/live_store_test.go
@@ -77,7 +77,8 @@ func TestLiveStoreFullBlockLifecycleCheating(t *testing.T) {
 	requireInstanceState(t, inst, instanceState{liveTraces: 0, walBlocks: 0, completeBlocks: 1})
 
 	// stop gracefully
-	services.StopAndAwaitTerminated(t.Context(), liveStore)
+	err = services.StopAndAwaitTerminated(t.Context(), liveStore)
+	require.NoError(t, err)
 }
 
 func TestLiveStoreReplaysTraceInLiveTraces(t *testing.T) {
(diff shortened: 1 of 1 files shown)


modules/livestore/live_store.go
```diff
@@ -259,8 +265,15 @@ func (s *LiveStore) starting(ctx context.Context) error {
 		return fmt.Errorf("failed to start partition reader: %w", err)
 	}
 
-	s.wg.Add(1)
-	go s.completeLoop()
+	for i := range s.cfg.CompleteBlockConcurrency {
+		idx := i
```
mattdurham, line 269: protecting against the ye olde reassignment issue? Though I think its fixed now.

modules/livestore/live_store.go
```diff
@@ -302,14 +306,16 @@ func (s *LiveStore) flushRemaining() {
...
 	defer timeout.Stop()
 
-	for s.flushqueues.Length() > 0 {
+	for !s.completeQueues.IsEmpty() {
 		select {
 		case <-ticker.C:
 		case <-timeout.C:
-			level.Error(s.logger).Log("msg", "flush remaining blocks timed out", "remaining", s.flushqueues.Length())
-			return // shutdown timeout reached
+			level.Error(s.logger).Log("msg", "flush remaining blocks timed out", "isEmpty", s.completeQueues.IsEmpty())
```
mattdurham, line 313: Is the IsEmpty here valuable? Feels like it might be confusing if in between the for and timeout tick it does become empty this could lead to debug confusion.

mattdurham (approved):

PR #5557, opened 2025-08-20
fix: fix identify root spans with child_of links

### LINKED ISSUE #3634: trace root span incorrectly identified due to instrumentation
Consider the following trace:

```
[
    Span{
        traceId: "39343434",
        spanId: "aaa"
        parentSpanId: "",
        links: [
            link{
                traceId: "39343434"
                spanId: "bbb"
                atributes: {
                    "opentracing.ref_type": "child_of"
                }
            }
        ]
    },
    Span{
        traceId: "39343434",
        spanId: "bbb"
        parentSpanId: "",
        links: []
    },
    Span{
        traceId: "39343434",
       spanId: "ccc"
        parentSpanId: "",
        links: [
            link{
                traceId: "39343434"
                spanId: "aaa"
                atributes: {
                    "opentracing.ref_type": "child_of"
                }
            }
        ]
    }
 
 ]

```

Currently, Tempo labels any spans with an empty parentSpanId field as the root span. In the above situation, Tempo would label whichever of the three spans is ingested last. Tempo should also check whether there is a link with attribute `child of` before labeling the span as the root span. The correct root span is `bbb` as it has no `child of` links.

mexirica:
**What this PR does**:
 Fixes incorrect root span detection in Parquet trace conversion (v2, v3, v4, v5). Now, a span is only considered root if it has no parent and no child_of link

Fixes #3634

**Checklist**
- [X] Tests updated
- [ ] Documentation added


knylander-grafana:
Greetings. Do you have any doc that you need reviewed for this PR? The Documentation box is checked but I didn't see any doc files in the PR. 

BTW, nice changelog entry. It's very clear and concise.

mexirica:
Sorry, checked by mistaken, no docs added.

Already changed the field @knylander-grafana 

Thanks for the changelog!

joe-elliott:
Nice! Thank you for taking this on.

The example in [the original issue](https://github.com/grafana/tempo/issues/3634) had child of links that linked to the same trace. I'm wondering if we want to check for both child of links and if they are internal to the trace. @ie-pham do you have input on this?

tempodb/encoding/vparquet3/schema.go
```diff
@@ -370,7 +370,20 @@ func traceToParquet(meta *backend.BlockMeta, id common.ID, tr *tempopb.Trace, ot
 				if s.EndTimeUnixNano > traceEnd {
 					traceEnd = s.EndTimeUnixNano
 				}
-				if len(s.ParentSpanId) == 0 {
+				var hasChildOfLink bool
+				for _, spanLink := range s.Links {
+					for _, attr := range spanLink.GetAttributes() {
+						if attr.Key == "opentracing.ref_type" && attr.GetValue().GetStringValue() == "child_of" {
```
joe-elliott: getting back into this. i'm noticing that this is an opentracing attribute. does otel not have a concept of "link type"? do we need to look for an otel attribute as well?

also, are these constants we can reference somewhere? i'd hope the semantic convs  are defined somewhere in code.

mexirica: I will check this, thanks for the tip @joe-elliott

mexirica: @joe-elliott I took a look at the code and found the traceql.Span that matches, which has a ChildOf field, but I don’t think that’s the right place to use this structure.

I didn’t find any constant for "opentracing.ref_type" or "child_of", only found "link.opentracing.ref_type" in the tests for vparquet4.

ie-pham: I believe this was an opentracing instrumentation issue. otel has many emphasis on span links being links outside of parent-child relationships.

mexirica:
@joe-elliott @ie-pham Could you please check the pipelines or manually re-run the checks?

Thank you!

ie-pham (approved):
LGTM! Amazing job. Thank you for contributing

PR #5559, opened 2025-08-21
Add tests to make unsafe usages more robust

carles-grafana:
**What this PR does**:

We have a couple of unsafe usages which could use more verification
to ensure a change in upstream dependencies don't cause panics or
wrong results.

This is the case of pq.Value in parquet-go, and labels.FastRegexMatcher in prometheus

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/parquetquery/intern/intern_test.go
+++ b/pkg/parquetquery/intern/intern_test.go
@@ -21,7 +21,7 @@ func TestInterner_internBytes(t *testing.T) {
 		t.Errorf("expected 2, got %d", len(i.m))
 	}
 	interned1, interned2 := i.internBytes([]byte("hello")), i.internBytes([]byte("hello"))
-	if interned1[0] != interned2[0] {
+	if unsafe.SliceData(interned1) != unsafe.SliceData(interned2) {
 		// Values are interned, so the memory address should be the same
 		t.Error("expected same memory address")
 	}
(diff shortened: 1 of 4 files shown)


pkg/parquetquery/intern/intern_test.go
```diff
@@ -21,7 +22,7 @@ func TestInterner_internBytes(t *testing.T) {
 		t.Errorf("expected 2, got %d", len(i.m))
 	}
 	interned1, interned2 := i.internBytes([]byte("hello")), i.internBytes([]byte("hello"))
-	if interned1[0] != interned2[0] {
+	if unsafe.SliceData(interned1) != unsafe.SliceData(interned2) {
```
carles-grafana, line 24: this test was wrong. It was comparing the value of the first byte instead of the address

pkg/regexp/regexp_test.go
```diff
@@ -23,6 +24,145 @@ func TestRegexpMatch(t *testing.T) {
...
+				require.NoError(t, err, "should create matcher for %s", tc.regex)
+
+				cheat := (*cheatToSeeInternals)(unsafe.Pointer(matcher))
+
+				matcherSize := unsafe.Sizeof(*matcher)
+				cheatSize := unsafe.Sizeof(*cheat)
+				require.Equal(t, matcherSize, cheatSize, "struct size mismatch")
+
+				matcherAlign := unsafe.Alignof(*matcher)
+				cheatAlign := unsafe.Alignof(*cheat)
```
carles-grafana: Sizeof and Alignof could detect big changes in the structure

pkg/regexp/regexp_test.go
```diff
@@ -23,6 +24,137 @@ func TestRegexpMatch(t *testing.T) {
...
+	t.Run("field_has_correct_order", func(t *testing.T) {
+		// This test validates that critical fields are at expected offsets to detect field reordering.
+		// Field reordering would pass size/alignment checks but cause reading wrong memory locations.
+
+		var cheat cheatToSeeInternals
+
+		require.Equal(t, uintptr(0), unsafe.Offsetof(cheat.reString),
+			"reString should be at offset 0 - struct field order may have changed")
+
+		stringMatcherOffset := unsafe.Offsetof(cheat.stringMatcher)
```
joe-elliott, line 40: very cool, i didn't know this existed.

PR #5568, opened 2025-08-22
ingest:  make Enconding sync.Pool more robust

carles-grafana:
We were resetting the length or the underlying slices but not clearing the contents.
This makes a potential future data leak due to a bug less likely.

Also add a test to ensure the data is reset when reusing it.

Benchmark in this branch:
```
goos: darwin
goarch: arm64
pkg: github.com/grafana/tempo/pkg/ingest
cpu: Apple M3 Pro
BenchmarkEncoderPoolPutDifferentSizes/Small_Original-11         	80590712	        15.56 ns/op	       0 B/op	       0 allocs/op
BenchmarkEncoderPoolPutDifferentSizes/Small_WithClear-11        	48544262	        25.38 ns/op	       0 B/op	       0 allocs/op
BenchmarkEncoderPoolPutDifferentSizes/Medium_Original-11        	19074123	        62.93 ns/op	       0 B/op	       0 allocs/op
BenchmarkEncoderPoolPutDifferentSizes/Medium_WithClear-11       	10028392	       117.1 ns/op	       0 B/op	       0 allocs/op
BenchmarkEncoderPoolPutDifferentSizes/Large_Original-11         	 1676859	       702.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkEncoderPoolPutDifferentSizes/Large_WithClear-11        	 1000000	      1120 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/grafana/tempo/pkg/ingest	9.330s
```

Clearing the slices is significantly slower. (60-85% slower). Is the extra safety worth it?

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingest/encoding_test.go
+++ b/pkg/ingest/encoding_test.go
@@ -2,8 +2,10 @@ package ingest
 
 import (
 	"math/rand"
+	"reflect"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/grafana/tempo/pkg/tempopb"
@@ -113,6 +115,43 @@ func BenchmarkEncodeDecode(b *testing.B) {
 	}
 }
 
+func TestResetPushBytesRequest(t *testing.T) {
+	// Create a request with all fields set to non-zero values
+	traces := []tempopb.PreallocBytes{
+		{Slice: []byte("trace1")},
+		{Slice: []byte("trace2")},
+		{Slice: []byte("trace3")},
+	}
+	ids := [][]byte{
+		[]byte("id1"),
+		[]byte("id2"),
+		[]byte("id3"),
+	}
+	req := &tempopb.PushBytesRequest{
+		Traces:                traces,
+		Ids:                   ids,
+		SkipMetricsGeneration: true,
+	}
+
+	resetPushBytesRequest(req)
+
+	// Verify all fields are properly reset
+	assert.NotNil(t, req.Traces, "Traces should not be nil")
+	assert.NotNil(t, req.Ids, "Ids should not be nil")
+	assert.Equal(t, 0, len(req.Traces), "Traces should be empty after reset")
+	assert.Equal(t, 0, len(req.Ids), "Ids should be empty after reset")
+	assert.Equal(t, false, req.SkipMetricsGeneration, "SkipMetricsGeneration should be reset to false")
+
+	// Verify slices are reused (not reallocated) by comparing pointers
+	originalTracesPtr := reflect.ValueOf(traces).Pointer()
+	newTracesPtr := reflect.ValueOf(req.Traces).Pointer()
+	assert.Equal(t, originalTracesPtr, newTracesPtr, "Traces slice should be reused, not reallocated")
+
+	originalIDsPtr := reflect.ValueOf(ids).Pointer()
+	newIDsPtr := reflect.ValueOf(req.Ids).Pointer()
+	assert.Equal(t, originalIDsPtr, newIDsPtr, "Ids slice should be reused, not reallocated")
+}
+
 // Helper function to generate a test trace
 func generateRequest(entries, lineLength int) *tempopb.PushBytesRequest {
 	stream := &tempopb.PushBytesRequest{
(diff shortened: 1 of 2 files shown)


carles-grafana:
@mdisibio Inlining the reset function didn't provide any significant speedup

joe-elliott:
For the benchmarks to be meaningful you need to run them a decent number of times in this branch and in main and compare the results.

Something like:
```
go test -run xxx -bench <your bench> -count=10 -benchmem > before.txt
go test -run xxx -bench <your bench> -count=10 -benchmem > after.txt
benchstat before.txt after.txt
```

carles-grafana:
> For the benchmarks to be meaningful you need to run them a decent number of times in this branch and in main and compare the results.
> 
> Something like:
> 
> ```
> go test -run xxx -bench <your bench> -count=10 -benchmem > before.txt
> go test -run xxx -bench <your bench> -count=10 -benchmem > after.txt
> ```

Note that in the new benchmark I'm already comparing the new and old implementation of `encoderPoolPut` (the original is reimplemented in the test file so I don't have to switch branches).

For the existing `EncodeDecode` benchmark, the effect is not visible:
```
  benchstat old.txt new.txt                                                    
  goos: darwin
  goarch: arm64
  pkg: github.com/grafana/tempo/pkg/ingest
  cpu: Apple M3 Pro
                  │   old.txt    │            new.txt            │
                  │    sec/op    │   sec/op     vs base          │
  EncodeDecode-11   139.0µ ± 23%   136.0µ ± 8%  ~ (p=0.684 n=10)

                  │   old.txt    │            new.txt             │
                  │     B/op     │     B/op      vs base          │
  EncodeDecode-11   1.242Mi ± 1%   1.255Mi ± 3%  ~ (p=0.796 n=10)

                  │   old.txt   │             new.txt             │
                  │  allocs/op  │  allocs/op   vs base            │
  EncodeDecode-11   2.004k ± 0%   2.004k ± 0%  ~ (p=1.000 n=10) ¹
  ¹ all samples are equal
```

pkg/ingest/encoding_test.go
```diff
@@ -184,3 +223,48 @@ func BenchmarkGeneratorDecoderPushBytes(b *testing.B) {
 		}
 	}
 }
+
+// Original implementation without clear() for comparison
+func encoderPoolPutOriginal(req *tempopb.PushBytesRequest) {
```
joe-elliott, line 228: cleanup?

tempodb/encoding/vparquet2/compactor.go
```diff
@@ -297,9 +297,7 @@ func (r *rowPool) Put(row parquet.Row) {
 	// Clear before putting into the pool.
 	// This is important so that pool entries don't hang
 	// onto the underlying buffers.
-	for i := range row {
-		row[i] = parquet.Value{}
-	}
+	clear(row)
```
joe-elliott: are these changes supposed to be here? if they are there are likely similar lines in parquet 4 and 5.

pkg/ingest/encoding_test.go
carles-grafana: This is an intentional implementation of the original function for the benchmark below.
Do you think it's not worth keeping?

tempodb/encoding/vparquet2/compactor.go
carles-grafana: it was on purpose but not really needed, because the clear is equivalent to the loop, so I'll just remove these changes

pkg/ingest/encoding_test.go
joe-elliott: that's not unprecedented. here we keep the original implementation of truncateRowNumber:

https://github.com/grafana/tempo/blob/main/pkg/parquetquery/iters.go#L80-L88

but we also do this b/c it allows us to validate the more complex implementation against the original:

https://github.com/grafana/tempo/blob/main/pkg/parquetquery/iters_test.go#L28

i'm not opposed to this and will go ahead and approve, but imo this is a simple enough change that we are good removing the old impl

joe-elliott (approved):

PR #5577, opened 2025-08-25
tempo: add tests for vparquet pools

carles-grafana:
To ensure data is cleaned and not leaked.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/encoding/vparquet5/block_traceql.go
+++ b/tempodb/encoding/vparquet5/block_traceql.go
@@ -819,7 +819,7 @@ func (s *span) attributesMatched() int {
 // can return a slice of dropped and kept spansets?
 var spanPool = sync.Pool{
 	New: func() interface{} {
-		return &span{}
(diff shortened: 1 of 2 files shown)


tempodb/encoding/vparquet5/block_traceql.go
```diff
@@ -819,7 +819,7 @@ func (s *span) attributesMatched() int {
 // can return a slice of dropped and kept spansets?
 var spanPool = sync.Pool{
 	New: func() interface{} {
-		return &span{}
+		return &span{rowNum: parquetquery.EmptyRowNumber()}
```
carles-grafana, line 822: we already set the EmptyRowNumber when we put the span back to the pool, so for consistency new Span should also start with the default.

stoewer (approved):

PR #5578, opened 2025-08-25
[Bugfix] Fix for not-parent operation

### LINKED ISSUE #4804: Unexpected behavior for TraceQL query using !< (not-parent) operator
**Describe the bug**

When using the !< (not-parent) operator in a query to find parent spans that do not have any child that satisfies a certain condition, the query returns no results, even when there are spans that should be returned.

**To Reproduce**

I'm creating multiple spans named `AppSession`. Some of these spans have a child span named `Navigation to ConfirmationScreen`.

If I run the following TraceQL query: 
`{ span:name = "Navigation to ConfirmationScreen" } < { span:name = "AppSession" }`

I get all the parent spans that have the `Navigation to ConfirmationScreen` child. That works as expected, as the documentation states: 

> `{condA} &< {condB}` - The parent operator (<) looks for spans matching {condB} that are direct parent spans of a child matching {condA}.

However, if I run:
 `{ span:name = "Navigation to ConfirmationScreen" } !< { span:name = "AppSession" }`

this query returns nothing. 

The docs for the not-parent operator state: 
> {condA} !< {condB} - The not-parent operator (!<) looks for spans matching {condB} that are not direct parent spans of a child matching {condA}

If my understanding is correct, any span with the name `AppSession` that is not a parent of a child with the name `Navigation to ConfirmationScreen` should be returned, but this is not the case.

**Expected behavior**
I expect the query to return parent spans that do not have any child with the specified name.

**Environment:**
I'm running the `grafana-lgtm` docker image locally and sending spans to Tempo from an Android app.

ruslan-mikhailov:
**What this PR does**: fixes queries like `{ span:name = "not-exist" } !< { span:name = "exist" }` and `{ span:name = "not-exist" } !> { span:name = "exist" }`. The bug was in early exit if the first part of the query has no spans.

Fixes #4804

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/ast_execute.go
+++ b/pkg/traceql/ast_execute.go
@@ -168,7 +168,7 @@ func (o SpansetOperation) evaluate(input []*Spanset) (output []*Spanset, err err
 // where the eval callback returns true.  For now the behavior is only defined when there is exactly one
 // spanset on both sides and will return an error if multiple spansets are present.
 func (o *SpansetOperation) joinSpansets(lhs, rhs []*Spanset, eval func(s Span, l, r []Span) []Span) ([]Span, error) {
-	if len(lhs) < 1 || len(rhs) < 1 {
+	if len(rhs) < 1 {
 		return nil, nil
 	}
 
@@ -181,7 +181,12 @@ func (o *SpansetOperation) joinSpansets(lhs, rhs []*Spanset, eval func(s Span, l
 		return nil, nil
 	}
 
-	return eval(rhs[0].Spans[0], lhs[0].Spans, rhs[0].Spans), nil
+	var lspans []Span
+	if len(lhs) >= 1 {
+		lspans = lhs[0].Spans
+	}
+
+	return eval(rhs[0].Spans[0], lspans, rhs[0].Spans), nil
 }
 
 // addSpanset is a helper function that adds a new spanset to the output. it clones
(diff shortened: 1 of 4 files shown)


pkg/traceql/ast_execute.go
```diff
@@ -168,7 +168,7 @@ func (o SpansetOperation) evaluate(input []*Spanset) (output []*Spanset, err err
 // where the eval callback returns true.  For now the behavior is only defined when there is exactly one
 // spanset on both sides and will return an error if multiple spansets are present.
 func (o *SpansetOperation) joinSpansets(lhs, rhs []*Spanset, eval func(s Span, l, r []Span) []Span) ([]Span, error) {
-	if len(lhs) < 1 || len(rhs) < 1 {
+	if len(rhs) < 1 {
```
mapno, line 171: Can't this result in the same error if the `rhs` is empty?

ruslan-mikhailov: If I understand correctly how it works, empty `rhs` always means empty end results. "All spans that satisfy a condition (rhs) that has `N` relation (a structural operation) to another set of spans (`lhs`)". If there is no spans for rhs, the final set will always be empty. 

"Pink cats that don't like green tea":
- lhs - "green tea"
- operation - "don't like"
- rhs - "pink cats"

If no pink cats exist, it does not matter what's in lhs and what operator is, there will be no such cats anyway.

mapno: Haha, amazing analogy 😆

mapno: Thanks for the explanation

PR #5581, opened 2025-08-27
[Enhancement] Allow block-builder to operate over empty partitions

ruslan-mikhailov:
**What this PR does**:
In case partition is empty during consumption cycle, it is possible that previously empty partition started to receive data

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/blockbuilder/blockbuilder.go
+++ b/modules/blockbuilder/blockbuilder.go
@@ -127,7 +127,7 @@ func (p partitionState) getStartOffset() kgo.Offset {
 }
 
 func (p partitionState) hasRecords() bool {
-	return p.endOffset >= -1
+	return p.endOffset > 0
 }
 
 func New(
@@ -265,7 +265,10 @@ func (b *BlockBuilder) consume(ctx context.Context) (time.Duration, error) {
 
 	// First iteration over all the assigned partitions to get their current lag in time
 	for i, p := range ps {
-		if !p.hasRecords() { // No records, we can skip the partition
+		if !p.hasRecords() { // No records, skip for the first iteration
+			ps[i].lastRecordTs = time.Now()
+			ps[i].commitOffset = 0 // always start at beginning
+			level.Info(b.logger).Log("msg", "partition has no records", "partition", p.partition)
 			continue
 		}
 		lastRecordTs, commitOffset, err := b.consumePartition(ctx, p)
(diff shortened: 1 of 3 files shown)


modules/blockbuilder/blockbuilder_test.go
```diff
@@ -539,7 +539,7 @@ func TestBlockBuilder_honor_maxBytesPerCycle(t *testing.T) {
 		{
 			name:             "Limited to 1 bytes per cycle",
 			maxBytesPerCycle: 1,
-			expectedCommits:  1,
+			expectedCommits:  2,
```
ruslan-mikhailov, line 542: This change passes in main. It is because second commit comes after `require.Eventually` and if wait a little more, it becomes 2. The behaviour seems legit: consumePartition has only one loop iteration, it writes a block and commits, then repeats on the next call. Correct me if I'm wrong.

modules/blockbuilder/blockbuilder.go
```diff
@@ -127,7 +127,7 @@ func (p partitionState) getStartOffset() kgo.Offset {
 }
 
 func (p partitionState) hasRecords() bool {
-	return p.endOffset >= -1
+	return p.endOffset > 0
```
mdisibio: We default to the value of endOffset to -2 here:  `ps    = partitionState{partition: partition, commitOffset: -1, endOffset: -2}`   Can we check why that is, is -1 is a valid value, if the partition contains 1 message?

modules/blockbuilder/blockbuilder.go
```diff
@@ -265,7 +265,10 @@ func (b *BlockBuilder) consume(ctx context.Context) (time.Duration, error) {
 
 	// First iteration over all the assigned partitions to get their current lag in time
 	for i, p := range ps {
-		if !p.hasRecords() { // No records, we can skip the partition
+		if !p.hasRecords() { // No records, skip for the first iteration
+			ps[i].lastRecordTs = time.Now()
```
mdisibio, line 288: Let's leave a comment here like:  `We treat the partition as updated through now, and will check it again after ConsumeCycleDuration has elapsed`

modules/blockbuilder/blockbuilder.go
ruslan-mikhailov: There are two entities in `partitionState`: `commitOffset` (committed offset) which is a consumer property and `endOffset` (last offset in partition) which is a partition property. 

`commitOffset` can have values:
- `-2` means consumer points to the earliest record
- `-1` means consumer points to the latest record
- `>= 0` - absolute offset number

For `endOffset` 0 represents an empty partition with no records. As soon as it has `1` record, it becomes `1` as well.


I added global vars and comments to make it more clear in codebase

modules/blockbuilder/blockbuilder.go
ruslan-mikhailov: \+

ruslan-mikhailov:
\+ rebase from latest main to resolve conflicts in changelogs

mdisibio (approved):

ruslan-mikhailov:
\+ rebase from latest main to resolve conflicts in changelogs

PR #5583, opened 2025-08-27
Livestore: Consolidate query logic

joe-elliott:
This PR consolidates query logic in the livestore to remove redundant code and standardize the way we iterate through blocks in an instance. It is a combination of behaviors from the ingester and metrics generator read paths.

- Created an [iterateBlocks](https://github.com/grafana/tempo/pull/5583/files#diff-708a69e1ac83ae4c6c95163e2ef2a54174a348aa2c350864e928d65419d83f58R48) function and used it through all 5 search paths (search tags, tag values, search/traceql, query range and trace id lookup). This function standardizes
  - use of the concurrency setting
  - error handling
  - exit early behavior and context cancelling
  - logging/tracing
  - testing block range for inclusion in the query
- Cleaned up [some places](https://github.com/grafana/tempo/pull/5583/files#diff-708a69e1ac83ae4c6c95163e2ef2a54174a348aa2c350864e928d65419d83f58R347) where we were using the global logger instead of the instance one
- [Fixed ingest-storage](https://github.com/grafana/tempo/pull/5583/files#diff-ee30af768fd59c456189ca40bb715e2a581a9158db86d4dc16c315a0107601b1R64) docker-compose so that search queries would correctly find historical blocks
- [Adjusted SearchTagValues](https://github.com/grafana/tempo/pull/5583/files#diff-db1208c4a3db264a9c4e8c51890a05d09e42db7ce150e447c14c515031e300f3R534) to follow the same pattern as all other read path calls.

After this PR is merged I want to follow up with a final cleanup step where all read path functions are moved to the same file instead of scattered across 3. I left them where they are for now to help the reviewer.

### DIFF
--- a/modules/livestore/query_range.go
+++ b/modules/livestore/query_range.go
@@ -8,14 +8,15 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync/atomic"
 	"time"
 
+	"github.com/go-kit/log/level"
 	"github.com/google/uuid"
 	"github.com/segmentio/fasthash/fnv1a"
 
 	"github.com/gogo/protobuf/proto"
 	"github.com/grafana/tempo/modules/ingester"
-	"github.com/grafana/tempo/pkg/boundedwaitgroup"
 	tempo_io "github.com/grafana/tempo/pkg/io"
 	"github.com/grafana/tempo/pkg/tempopb"
 	"github.com/grafana/tempo/pkg/traceql"
@@ -23,7 +24,6 @@ import (
 	"github.com/grafana/tempo/tempodb/encoding/common"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
-	"go.uber.org/atomic"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


modules/livestore/query_range.go
```diff
@@ -73,100 +73,45 @@ func (i *instance) QueryRange(ctx context.Context, req *tempopb.QueryRangeReques
...
-	for _, w := range i.walBlocks {
-		if jobErr.Load() != nil {
-			break
-		}
-
-		if !withinRange(w.BlockMeta()) {
-			continue
-		}
-
-		if maxSeries == 0 || !maxSeriesReached.Load() {
```
mapno, line 120: We're losing this check to avoid queueing more work once max series has been reached, should we add it to `search`?

joe-elliott: if we hit max series then search will return errComplete. this will cause the iterateBlocks function to cancel context and stop queueing up any jobs. so i think this check still exists effectively.

Cancel context here:
https://github.com/grafana/tempo/pull/5583/files/64cbc82f00f2a4c1f9a148128ce3bfefb0730f9b#diff-708a69e1ac83ae4c6c95163e2ef2a54174a348aa2c350864e928d65419d83f58R58

And then we check context at the start of the loop here:
https://github.com/grafana/tempo/pull/5583/files/64cbc82f00f2a4c1f9a148128ce3bfefb0730f9b#diff-708a69e1ac83ae4c6c95163e2ef2a54174a348aa2c350864e928d65419d83f58R98-R100

PR #5585, opened 2025-08-27
Allow max compaction level configuration in compaction block selector

zalegrala:
**What this PR does**:

Here we allow a configuration of the maximum compaction level a block can reach and still be considered for compaction.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/sources/tempo/configuration/manifest.md
+++ b/docs/sources/tempo/configuration/manifest.md
@@ -1049,6 +1049,7 @@ backend_scheduler:
             max_jobs_per_tenant: 1000
             min_input_blocks: 2
             max_input_blocks: 4
+            max_compaction_level: 0
             min_cycle_interval: 30s
     job_timeout: 15s
     local_work_path: /var/tempo
--- a/modules/backendscheduler/provider/compaction.go
+++ b/modules/backendscheduler/provider/compaction.go
@@ -28,12 +28,13 @@ import (
 var tracer = otel.Tracer("modules/backendscheduler/provider/compaction")
 
 type CompactionConfig struct {
-	MeasureInterval  time.Duration           `yaml:"measure_interval"`
-	Compactor        tempodb.CompactorConfig `yaml:"compaction"`
-	MaxJobsPerTenant int                     `yaml:"max_jobs_per_tenant"`
-	MinInputBlocks   int                     `yaml:"min_input_blocks"`
-	MaxInputBlocks   int                     `yaml:"max_input_blocks"`
-	MinCycleInterval time.Duration           `yaml:"min_cycle_interval"`
+	MeasureInterval    time.Duration           `yaml:"measure_interval"`
+	Compactor          tempodb.CompactorConfig `yaml:"compaction"`
+	MaxJobsPerTenant   int                     `yaml:"max_jobs_per_tenant"`
+	MinInputBlocks     int                     `yaml:"min_input_blocks"`
+	MaxInputBlocks     int                     `yaml:"max_input_blocks"`
+	MaxCompactionLevel int                     `yaml:"max_compaction_level"`
+	MinCycleInterval   time.Duration           `yaml:"min_cycle_interval"`
 }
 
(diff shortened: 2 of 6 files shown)


docs/sources/tempo/configuration/manifest.md
```diff
@@ -1049,6 +1049,7 @@ backend_scheduler:
             max_jobs_per_tenant: 1000
             min_input_blocks: 2
             max_input_blocks: 4
+            max_compaction_level: 0
```
electron0zero, line 1052: should we default to something like 5?

zalegrala: I'm thinking its going to be dependent on too many factors to have it enabled by default.

joe-elliott (commented):
the value of compacting 4 blocks to 1 is always basically the same. the cost of compacting 4 blocks to 1 is a function of the size of the input and output files. b/c of this it makes more sense to limit compactions based on max bytes/objects since these are the things that drive cost.

can you give some reasoning on why this is being added? my best guess is it's a way to limit the number of compaction jobs found for a given tenant to help move on from a tenant and service all equally.

zalegrala:
Right, I'm in agreement that the value to the query path and poller are the same for each compaction, and that the cost of doing so rises over time, as a function of the block size, etc.
Primarily I'm thinking about the number of times we have to touch a block before we reach the final state.  In our largest and most diverse (tenant sizing) environments, I'm seeing some level 8 blocks which are still tiny, and so we can expect even more compaction if other blocks are found in the same time window since they won't reach their max block size (set at the cluster level).  For these high level blocks, this could have been a single compaction with say 20 input blocks, and we only have to read/write those bytes once.
Some of this might be a feeling that we lack certain controls around how we compact.  If we're autoscaling on the number of outstanding blocks, and we only get to those level 8 blocks after all the level 0, 1, etc, we'er scaling unnecessarily for tiny blocks which take milliseconds to compact, grow the job list, which we then have to clean up in the scheduler.
For the large tenants this isn't an issue since they reach the max block size well before the levels get too high.  So say we go nominally from 256 original level 0 blocks to 1 final block at level 4, say and we stop because because of max block size.
Take for example a recent rollout.
image
There is a difference in measurement, since the compactors (on the left) only report the outstanding blocks after they have worked on a tenant for 15 minutes and then moved on.  This is showing a relatively stable 20k outstanding blocks which are the highest level blocks that compactors are not making progress against, yet we maintain scale for.
In this environment, if we set a level limit like 4 or 5, we'd shed all of that excess load and I think we'd be able to scale down.  The smallest tenants would have an elevated block list because we didn't fully compact them to reach the max block size, but since they are the smallest tenants, I expect this to be relatively low impact.

Okay so I wrote my thoughts, but now I'm looking for the number of objects written at level 5 and above.
image
Some, yes, but not as many as I had in my head.  Maybe I'm wrong and that any level of compaction is worth it.  There are also other changes I can think of which might scale the input blocks based on final estimated size or some such, but more complexity than the config param here, but also likely more effective.
I'm open either way.
Is your assessment that any level of compaction is worth the cost?

joe-elliott:
Nice details. Thank you. I didn't consider it as a way to limit the total work to be done in the system which impacts our scaling.

> Is your assessment that any level of compaction is worth the cost?

I think the value of compaction is really hard to lock down. obviously it's reducing the blocklist which reduces overall TCO. it also helps stabilize trace by id search since the amount of work grows linearly with the blocklist. but in a multi tenant system it's better to spend resources on larger blocklists b/c we know that the cell is scaled enough to handle the smaller ones. from that perspective this PR makes a lot of sense to me b/c it cuts off any jobs that makes it to levels 5,6,7+ and is still under the max bytes. this is likely a smaller tenant. i do wonder if this can be indirectly controlled through compaction_window, but to do that well it would have to be some kind of dynamic system based on tenant ingestion.

i do worry a bit about the number of options to control compaction jobs, but i better understand the need for this one. thx!

PR #5586, opened 2025-08-28
fix(metrics-generator): prevent WAL deletion when tenant is empty

sienna011022:
**What this PR does:**
Prevents metrics-generator from deleting the entire WAL root directory when an empty tenant is encountered. This issue could cause data loss for all tenants

**Root Cause:**
• Empty tenant ("") causes filepath.Join(cfg.Path, tenant) to resolve to the WAL root directory
• os.RemoveAll(walDir) then deletes the entire root, affecting all tenant WAL data
• This occurs when X-Scope-OrgID header is missing or internal metrics lack tenant labels

**Solution:**
• Throw Error when tenant is  empty when  multi tenancy environment enabled

**Testing:**
• Added TestInstance_emptyTenantValidation with 2 test cases
• All existing tests pass
• Verified WAL directory structure remains intact for existing tenants

Which issue(s) this PR fixes:
Fixes potential WAL data loss in multi-tenant environments (no existing issue number)

Checklist
* [x] Tests updated
* [x] Documentation added  
* [x] CHANGELOG.md updated - the order of entries should be [CHANGE], [FEATURE], [ENHANCEMENT], [BUGFIX]

### DIFF
--- a/modules/generator/storage/instance.go
+++ b/modules/generator/storage/instance.go
@@ -61,10 +61,9 @@ var _ Storage = (*storageImpl)(nil)
 func New(cfg *Config, o Overrides, tenant string, reg prometheus.Registerer, _ log.Logger) (Storage, error) {
 	// TODO move this to the generator.go
 
-	// Validate empty tenant to prevent WAL directory deletion
-	if tenant == "" {
-		// Use Tempo standard default for single-tenant mode
-		tenant = "single-tenant"
+	// Validate empty tenant to prevent WAL directory deletion only when org ID header is required
+	if cfg.RemoteWriteAddOrgIDHeader && tenant == "" {
+		return nil, errors.New("tenant cannot be empty")
 	}
 
 	logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{
--- a/modules/generator/storage/instance.go
+++ b/modules/generator/storage/instance.go
@@ -60,14 +60,22 @@ var _ Storage = (*storageImpl)(nil)
 // Should we standarize slog and deprecate go-kit/log too?
 func New(cfg *Config, o Overrides, tenant string, reg prometheus.Registerer, _ log.Logger) (Storage, error) {
 	// TODO move this to the generator.go
+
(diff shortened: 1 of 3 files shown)


modules/generator/storage/instance.go
```diff
@@ -60,14 +60,22 @@ var _ Storage = (*storageImpl)(nil)
 // Should we standarize slog and deprecate go-kit/log too?
 func New(cfg *Config, o Overrides, tenant string, reg prometheus.Registerer, _ log.Logger) (Storage, error) {
 	// TODO move this to the generator.go
+
+	// Validate empty tenant to prevent WAL directory deletion
+	if tenant == "" {
```
joe-elliott, line 65: i like this fix b/c of how catastrophic the result is if it occurs, but technically this scenario should never happen.

wdyt about just returning an error instead of using "single-tenant"?

sienna011022: @joe-elliott 
I agree thx

Finally, I used `RemoteWriteAddOrgIDHeader` as the decision flag because it already maps directly to the deployment mode:

- true (multi-tenant) → tenant ID is required for isolation, so we throw an error if it’s missing
- false (single-tenant) → no isolation needed, so we allow empty tenant IDs to pass through

This way we don’t introduce new config knobs or break existing single-tenant flows. 
The validation logic stays aligned with the original flag’s semantics, making the behavior both clear and consistent.

Backend-sungyoon:
@CLAassistant check

Backend-sungyoon:
@cla-assistant check

modules/generator/storage/instance.go
joe-elliott: RemoteWriteOrgIDHeader doesn't indicate whether or not Tempo is in multitenant mode. Honestly an empty tenant here is always an error whether it's running as single or multi tenant. I'd just change the code to always return an error if tenant is an empty string.

sienna011022: @joe-elliott I understood I changed my code always return error when tenant comes empty string

joe-elliott (approved):
nice find and thx for the contribution!

Issue #5370, opened 2025-07-08
Search by traceID and trace:id in traceQL returns different list of spans


pjastrzabek:
I am sending traces with telemetrygen to locally running otel collector and later to tempo. 
Each trace  has 10 spans. Next I am listing recent traces and pick some of them using two methods

1. by traceID "7820b37938dfa5743829fb7f8c133b7"
2. by traceQL
```
{trace:id="7820b37938dfa5743829fb7f8c133b7"}
```

Number of spans returned sometimes is different.

**To Reproduce**

1. Send traces using telementrygen
```
telemetrygen traces --otlp-insecure --rate 1  --service bugtest  --telemetry-attributes bugtest=\"debug\" --child-spans 10 --span-duration 1h --traces 20
```

2. List recent traces with. 
```
{} with(most_recent=true)
```

Sometimes some traces are listed with "root span not yet received"

<img width="1353" height="459" alt="Image" src="https://github.com/user-attachments/assets/ee002ac1-2b9f-4a0d-84e6-6f1732a5001d" />

If I search for such trace by traceID it returns all spans (10 of them as specified in telemetrygen)

<img width="715" height="753" alt="Image" src="https://github.com/user-attachments/assets/184f359b-886b-4e16-bc09-b85463458f34" />

If I search for it with traceQL like this
```
{trace:id="1db4de36a545f42f59e9ae42e9b996c0"}
```

sometimes It returns subset of spans
Number of spans returned varies, sometimes it's 5, sometimes it's 9, sometimes all of them are there.

<img width="1407" height="593" alt="Image" src="https://github.com/user-attachments/assets/65588ba6-488a-4c57-a71c-a6cac990a92b" />

But once not all of them are returned via traceQL 
The response **stays like this**.

This is production issue for us reported by internal customer.
Otel collector between telemetry gen and tempo is configured with batch processor, I think it's not relevant as the trace is fully sent to tempo (all spans are visible when searched by traceID)

batch processor config is
```
processors:
  batch:
    send_batch_max_size: 100
    send_batch_size: 10
    timeout: 10s
```

**Expected behaviour**
List of spans returned is the same.

**Environment:**
 - Infrastructure:  tempo distributed installed with helm chart in kubernetes

**Additional Context**
Tempo version is 2.8.0, but issue existed before (for sure on 2.7.x)

[...]

Issue #4967, opened 2025-04-08
Issue querying some trace IDs using `trace:id`


sam-mcbr:
**Describe the bug**
Running TraceQL queries `{trace:id = "<some trace id>"}` does not return data for some trace ids.

I have this trace id in Tempo: `fbb99bfc59174ccefa0338d4886914`. I can run the TraceQL for the trace (`fbb99bfc59174ccefa0338d4886914`) and the data for the trace is returned. Additionally, if I run a query for all traces (`{}`) I see it in the returned table. But if I run `{trace:id = "fbb99bfc59174ccefa0338d4886914"}` I get "0 series returned".

This only seems to affect some trace ids though, as this one `7fbea980725df26097c73beeaa8ba8`  returns the expected trace in the table view when running `{trace:id = "7fbea980725df26097c73beeaa8ba8"}`.

**To Reproduce**

I'm not exactly sure how to reproduce this as it appears only some trace ids are affected. I left information under the bug description.

**Expected behavior**
I expect a query for a known trace id like `{trace:id = "fbb99bfc59174ccefa0338d4886914"}` to return data.

**Environment:**
 - Infrastructure: Kubernetes
 - Deployment tool: helm
 - Tempo version: 2.7.1
 - Grafana version: 11.6.0

**Additional Context**
Looked at logs for the query-frontend and didn't see anything obvious for why one would not return data.

Running TraceQL query `{trace:id = "7fbea980725df26097c73beeaa8ba8"}`:
```
level=info ts=2025-04-08T00:18:18.684153195Z caller=search_handlers.go:191 msg="search request" tenant=single-tenant query="{trace:id = \"7fbea980725df26097c73beeaa8ba8\"}" range_seconds=10800 limit=20 spans_per_spanset=3
level=info ts=2025-04-08T00:18:18.688416552Z caller=search_handlers.go:172 msg="search response" tenant=single-tenant query="{trace:id = \"7fbea980725df26097c73beeaa8ba8\"}" range_seconds=10800 duration_seconds=0.004263983 request_throughput=2.186875510526191e+07 total_requests=6 total_blockBytes=95262 total_blocks=3 completed_requests=6 inspected_bytes=93248 inspected_traces=0 inspected_spans=0 status_code=200 error=null
[...]

PR #5587, opened 2025-08-28
[bugfix] Fix search by trace:id for short IDs with leading zeros

ruslan-mikhailov:
**What this PR does**: fixed trace:id queries with leading zeroes like `{ trace:id = "00000000000000008bb87d091822a2f9" }`

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/ast_test.go
+++ b/pkg/traceql/ast_test.go
@@ -1100,3 +1100,67 @@ func TestNeedsFullTrace(t *testing.T) {
 		})
 	}
 }
+
+func TestNewBinaryOperation_TraceIDNormalization(t *testing.T) {
+	traceIDAttr := NewIntrinsic(IntrinsicTraceID)
+	op := OpEqual
+
+	tests := []struct {
+		name       string
+		expression FieldExpression
+		expected   FieldExpression
+	}{
+		{
+			name:       "with leading zeros",
+			expression: NewStaticString("0000123"),
+			expected:   NewStaticString("123"),
+		},
+		{
+			name:       "no leading zeros",
+			expression: NewStaticString("123456789abcdef123456789abcdef"),
+			expected:   NewStaticString("123456789abcdef123456789abcdef"),
+		},
+		{
+			name:       "int",
+			expression: NewStaticInt(123),
+			expected:   NewStaticInt(123),
+		},
+		{
+			name:       "float",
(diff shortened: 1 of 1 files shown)


joe-elliott:
I believe this fixes: 
https://github.com/grafana/tempo/issues/5370
https://github.com/grafana/tempo/issues/4967

Edit:
Well maybe I'm wrong. It's strange that both of those issues are examples of trace ids that have leading 0s that are not in the traceql query string.

pkg/traceql/ast_conditions.go
```diff
@@ -61,10 +73,16 @@ func (o *BinaryOperation) extractConditions(request *FetchSpansRequest) {
 					Operands:  nil,
 				})
 			} else {
+				operand := o.RHS.(Static)
```
joe-elliott: it feels strange to me to perform this normalization in these very specific places. should we instead do this in newBinaryOperation and then we can feel confident that trace id's are represented the same everywhere in the ast regardless?

joe-elliott:
I was thinking about this and realized that it doesn't fix cases like:

```
{ trace:id = span.foo }
```

Unfortunately, to fix this completely I feel like we'd need some kind of `IdsEqual(a, b)` function in the `traceql` package that is used by both the engine and the fetch layers. Right now we're just treating it as a string type, but obviously that's not quite correct.

joe-elliott:
One more thought! Should we stop comparing strings and start comparing byte slices?

pkg/traceql/ast_conditions.go
ruslan-mikhailov: That's a very nice idea, thank you for that! Fixed

ruslan-mikhailov:
\+ rebase from latest main to revive the branch and resolve conflicts

pkg/traceql/ast.go
```diff
@@ -519,6 +520,17 @@ func newBinaryOperation(op Operator, lhs, rhs FieldExpression) FieldExpression {
 		RHS: rhs,
 	}
 
+	if attr, ok := lhs.(Attribute); ok && attr.Intrinsic == IntrinsicTraceID {
```
joe-elliott, line 523: a unit test on newBinaryOperation to lock this down would be nice.

ruslan-mikhailov:
\+ unit test

PR #5563, opened 2025-08-21
Rhythm: Livestore

joe-elliott:
**What this PR does**:

Adds a new component, the live store, to Tempo. This new component is part of the Rhythm RF1 queue based rearchitecture. It is designed to consume and store recent data for querying only. The live store "works", but it is far from production ready. We are PRing this long running branch to main so that we can start experimenting with the new component in dev cells and more cleanly iterate on it.

This is a large PR and deserves a decent review, but the most important thing as a reviewer is to feel confident this does not impact current Tempo operations. The livestore itself is still being heavily debugged and tested so things missed in that code are less important, but we need to be certain this won't impact our existing architecture. Having said that I'd encourage the reviewer to focus on changes made in:

- ./cmd/tempo/app (especially modules.go)
- ./modules/distributor
- ./modules/querier
- ./modules/generator

PR #5590, opened 2025-08-28
Livestore: Consume improvements

joe-elliott:
Primarily this PR correctly persists the span start/end timestamps to the block start/end meta while using ingestion slack and record.timestamp. The blockbuilder already correctly did this.

Other changes:
- Adds a test to confirm that record.Timestamp, ingestion slack and span start/end are correctly used to generate block metas in the livestore
- Adds the same test in blockbuilder (no changes to blockbuilder logic. just locking down this behavior)
- Changed the consume loop to take a kgo.RecordIter to remove unnecessary allocs/transformations. [this was a change request on the original PR](https://github.com/grafana/tempo/pull/5563#discussion_r2298876135)
- Preferred dropping records to perfectly persisting data in the face of errors. Added drop reasons to metric failed records. I'm willing to discuss but it seems to me that it would be better in the livestore to favor "best effort" behavior to provide recent data as effectively as possible while the blockbuilder should favor "perfect" behavior.

### DIFF
--- a/modules/blockbuilder/blockbuilder_test.go
+++ b/modules/blockbuilder/blockbuilder_test.go
@@ -603,6 +603,110 @@ func TestBlockBuilder_honor_maxBytesPerCycle(t *testing.T) {
 	}
 }
 
+func TestBlockbuilder_usesRecordTimestampForBlockStartAndEnd(t *testing.T) {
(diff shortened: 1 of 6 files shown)


mapno (approved):
LGTM

PR #5591, opened 2025-08-29
Add live-stores to jsonnet lib

mapno:
**What this PR does**:

Adds required resources to run live-stores from the jsonnet library. Also updates the k3d example to run a full rhythm setup.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


operations/jsonnet/microservices/configmap.libsonnet
```diff
@@ -109,8 +109,13 @@
 
   tempo_query_frontend_config:: $.tempo_config {},
   tempo_block_builder_config:: $.tempo_config {},
+  tempo_live_store_config:: $.tempo_config {},
   tempo_backend_scheduler_config:: $.tempo_config {},
-  tempo_backend_worker_config:: $.tempo_config {},
+  tempo_backend_worker_config:: $.tempo_config {
+    backend_worker+: {
+      backend_scheduler_addr: 'backend-scheduler.%s.svc.cluster.local.:9095' % [$._config.namespace],
```
mapno, line 116: Noticed this was missing, but the `backend_worker` can't start without this address. Is it okay to include by default @zalegrala? It follows the same pattern of the query-frontend / querier.

zalegrala: Looks good to me. :+1:

operations/jsonnet-compiled/ConfigMap-tempo-live-store.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: v1
+data:
+  tempo.yaml: |
+    compactor: {}
+    distributor: {}
+    http_api_prefix: ""
+    ingester:
```
joe-elliott, line 7: do we need this block? not sure what it does

operations/jsonnet-compiled/StatefulSet-live-store-zone-a.yaml
```diff
@@ -0,0 +1,96 @@
...
+          - labelSelector:
+              matchExpressions:
+              - key: rollout-group
+                operator: In
+                values:
+                - live-store
+              - key: name
+                operator: NotIn
+                values:
+                - live-store-zone-a
```
joe-elliott, line 38: not 100% sure what these scheduling affinity rules do, but it's ok if multiple zone-a livestores are on the same node. we need to make sure that zone-a livestores never share a node with zone-b

operations/jsonnet-compiled/ConfigMap-tempo-live-store.yaml
mapno: It's inherited from the base config, it's set in all components. I thought of cleaning it up but didn't want to change too many things outside the scope of this PR

operations/jsonnet-compiled/StatefulSet-live-store-zone-a.yaml
mapno: Yes, that's exactly what it does. If `allow_multiple_replicas_on_same_node` is `false`, then it excludes nodes where other zones are running in.

PR #5592, opened 2025-08-29
Proposal: [bugfix] Do not start from start if block-builder is at the end of partition

ruslan-mikhailov:
**What this PR does**: addresses issue from
```go
func (p partitionState) getStartOffset() kgo.Offset {
	if p.commitOffset > commitOffsetAtEnd {
		return kgo.NewOffset().At(p.commitOffset)
	}
	// If commit offset is AtEnd (-1), it nevertheless will consume from the start.
	// This is a workaround for franz-go and default Kafka behaviour:
	// in case consumer is new and has no committed offsets, it will start consuming from the end,
	// while for block builder, it should consume from the earliest record.
	// The workaround is dirty and can break the consumer if it starts returning AtEnd (-1) for
	// already running consumer.
	// TODO: replace the workaround with proper new consumer offset initialization
	// if p.commitOffset == commitOffsetAtEnd {
	// 	return kgo.NewOffset().AtEnd()
	// }
	return kgo.NewOffset().AtStart()
}
```
So, the PR introduces 2 changes:
1. If kafka lib returns offset at end, it points the consumer to at end
2. It sets initial offset to -2 (at start):
- fetch all offsets from consumer group
- lookup for given partitions. If they are not found, set it to `AtStart`

I wrap the PR as a Proposal, because I'm not sure it is better or less hacky than the current solution. Another downside is that if we receive `AtEnd` from libs for a not-new consumer (timestamp 1), and we start consuming from the latest record (timestamp 2), there is a chance that between these two timestamps we can receive new records that will be lost. 

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/blockbuilder/blockbuilder.go
+++ b/modules/blockbuilder/blockbuilder.go
@@ -568,12 +568,29 @@ func (b *BlockBuilder) getPartitionOffsets(ctx context.Context, partitionIDs []i
 		topic = b.cfg.IngestStorageConfig.Kafka.Topic
 		group = b.cfg.IngestStorageConfig.Kafka.ConsumerGroup
 	)
-	commits, err := b.kadm.FetchOffsetsForTopics(ctx, group, topic)
+
+	// Use FetchOffsets directly instead of FetchOffsetsForTopics to avoid automatic -1 backfill
+	commits, err := b.kadm.FetchOffsets(ctx, group)
 	if err != nil {
 		return nil, nil, err
 	}
 	if err := commits.Error(); err != nil {
 		return nil, nil, err
+	}
+
+	// Manually handle missing partitions by setting them to -2 (start from beginning)
+	// instead of franz-go's default -1 backfill
+	for _, partitionID := range partitionIDs {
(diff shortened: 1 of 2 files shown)


modules/blockbuilder/blockbuilder.go
```diff
@@ -575,7 +608,7 @@ func (b *BlockBuilder) getPartitionOffsets(ctx context.Context, partitionIDs []i
 func (b *BlockBuilder) getPartitionState(partition int32, commits kadm.OffsetResponses, endsOffsets kadm.ListedOffsets) partitionState {
 	var (
 		topic = b.cfg.IngestStorageConfig.Kafka.Topic
-		ps    = partitionState{partition: partition, commitOffset: -1, endOffset: -2}
+		ps    = partitionState{partition: partition, commitOffset: commitOffsetAtEnd, endOffset: emptyPartitionEndOffset}
```
ruslan-mikhailov: This line also looks suspicious and should be `commitOffsetAtStart`

PR #5594, opened 2025-08-29
Drop tracing bridges, include tracing SDK and service version

zalegrala:
**What this PR does**:

Here we dtop the tracing bridges and rely on the OTEL SDK for all tracing needs.  We had kept the bridges for a while to ease the transition, but with upstream changes in `dskit` I believe this is no longer necessary.  Also here is a change to include the service version in the trace provider, which is useful for debugging.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


electron0zero (approved):

mapno (approved):
🙈

PR #5598, opened 2025-09-01
pool: reset slice length when putting them to the bucketed pool

carles-grafana:
Before, it was the responsibility of the user to manage the length. In all existing usages the length was reset by the caller or the slice was fully overwritten, so this wasn't a problem. But for extra safety in the future, it's worth doing to ensure future callers don't introduce a bug that could leak data.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


javiermolinar (approved):
LGTM

PR #5599, opened 2025-09-01
blockbuilder: fix data race in test

carles-grafana:
Before, the poller goroutine attempted to use to the test log during the test cleanup, causing a data race.

Now we stop the poller goroutine before the cleanup finishes.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/blockbuilder/blockbuilder_test.go
```diff
@@ -766,6 +766,8 @@ func TestBlockbuilder_marksOldBlocksCompacted(t *testing.T) {
 	require.NoError(t, services.StartAndAwaitRunning(ctx, b))
 	t.Cleanup(func() {
 		require.NoError(t, services.StopAndAwaitTerminated(ctx, b))
+		cancel(errors.New("test cleanup"))
+		store.Shutdown()
```
javiermolinar, line 770: There is a cleanup function registered at the top of the test cancelling the context. We could remove that one to favor this one

carles-grafana: I left it on purpose because if there is a failure before this other cleanup, we still want to cancel the context.
And a double cancel is a no-op.

javiermolinar (approved):

modules/blockbuilder/blockbuilder_test.go
stoewer: Or we can start using `T.Context` that was [added in Go 1.24](https://pkg.go.dev/testing#T.Context)

PR #5600, opened 2025-09-02
[Rhythm] Live-store: downscale endpoint

ruslan-mikhailov:
**What this PR does**: adds endpoint `/live-store/prepare-partition-downscale` to prepare for downscaling. Live-store downscale flow:
1. Trigger POST request to `/live-store/prepare-partition-downscale`
2. Partition's set to `INACTIVE` state. It is in read-only mode and does not receive new records.
3. Live-store lives N minutes until requests for traces that it stores start hitting backend, and it is safe to remove it from load.
4. Trigger POST request to `/live-store/prepare-downscale` which allows live-store to remove itself from partition owners
5. Shutdown the pod
6. When no owners of the partition left, remove it from the ring

The implementation of the endpoint is similar to Mimir ingester's endpoint with slight changes

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/operations/jsonnet/microservices/live-store.libsonnet
+++ b/operations/jsonnet/microservices/live-store.libsonnet
@@ -123,7 +123,7 @@
     statefulSet.mixin.spec.selector.withMatchLabels({ name: name, 'rollout-group': 'live-store' }) +
     statefulSet.mixin.spec.updateStrategy.withType('OnDelete') +
     statefulSet.mixin.spec.template.spec.withTerminationGracePeriodSeconds(1200) +
-    statefulSet.mixin.spec.withReplicas($._config.live_store.replicas) +
+    statefulSet.mixin.spec.withReplicas(0) +
     (if !std.isObject($._config.node_selector) then {} else statefulSet.mixin.spec.template.spec.withNodeSelectorMixin($._config.node_selector)) +
     statefulSet.spec.template.spec.securityContext.withFsGroup(10001) +  // 10001 is the UID of the tempo user
     self.liveStoreZoneAntiAffinity(name),


modules/livestore/downscale.go
```diff
@@ -0,0 +1,128 @@
...
+//   - GET
+//     Returns timestamp when partition was switched to INACTIVE state, or 0, if partition is not in INACTIVE state.
+//
+//   - POST
+//     Switches the partition to INACTIVE state (if not yet), and returns the timestamp when the switch to
+//     INACTIVE state happened.
+//
+//   - DELETE
+//     Sets partition back from INACTIVE to ACTIVE state.
+func (s *LiveStore) PreparePartitionDownscaleHandler(w http.ResponseWriter, r *http.Request) {
```
mattdurham, line 33: Is there a risk of multiple parallel invocations causing issues here? IE in the HTTP POST where the partition state might change between line 53 and line 60?

mapno: This endpoint is called by the [rollout-operator](https://github.com/grafana/rollout-operator) which is run as a singleton. This handler is not thread-safe, but I'm not sure it's a real problem

mapno (approved):

operations/jsonnet/microservices/config.libsonnet
```diff
@@ -143,6 +143,7 @@
       pvc_storage_class: error 'Must specify a live-store pvc storage class',
       replicas: 0,
       max_unavailable: 25,
+      downscale_delay: '30m',
```
ruslan-mikhailov: I picked the value based on default in query_backend_after (30 min). Possibly, need to increase to 45 min to avoid edge cases?

ruslan-mikhailov: Updated to 35m

operations/jsonnet/microservices/configmap.libsonnet
```diff
@@ -8,6 +8,11 @@
     server: {
       http_listen_port: $._config.port,
     },
+    live_store: {
+      partition_ring: {
+        delete_inactive_partition_after: $._config.live_store.downscale_delay,
+      },
+    },
```
mapno: Can you change this to `tempo_live_store_config` (line 112) so the blast radius is smaller? Otherwise this is propagated to all components

ruslan-mikhailov:
\+ review and doc fixes
\+ rebase from main

operations/jsonnet/microservices/configmap.libsonnet
ruslan-mikhailov: Changed, should be better now

operations/jsonnet/microservices/live-store.libsonnet
```diff
@@ -85,17 +85,45 @@
...
+      'grafana.com/prepare-downscale-http-path': 'live-store/prepare-downscale',
+      'grafana.com/prepare-downscale-http-port': '%(port)s' % $._config,
+      'grafana.com/rollout-delayed-downscale': $._config.live_store.downscale_delay,
+      // Endpoint for telling live-store that it's going to be scaled down later. Partition changes state to read-only.
+      'grafana.com/rollout-prepare-delayed-downscale-url': 'http://pod:%(port)s/live-store/prepare-partition-downscale' % $._config,
+    } + (
+      if !primary then {
+        // Disable updates of ReplicaTemplate/live-store from non-primary (zone-b) statefulsets.
+        'grafana.com/rollout-mirror-replicas-from-resource-write-back': 'false',
+      }
```
mapno, line 119: Can you briefly explain how this works? I thought that replicas were updated from the `ReplicaTemplate` to the `StatefulSet`. Does the `ReplicaTemplate` follow the replicas in one zone instead?

ruslan-mikhailov: I copied this from mimir's implementation.

Here PR where it was introduced: https://github.com/grafana/rollout-operator/pull/169

My understanding is that if it sets to true (default), then changes in statefulset's replica count will write back to replicatemplate. If understand correctly, then changes to zone-a's sts (due to autoscaling, for example) changes replicatemplate -> changes also zone-b

mapno: What's the strategy for updating replicas then? Update the primary sts?

ruslan-mikhailov:

PR #5601, opened 2025-09-03
Measure bytes before limits

mapno:
**What this PR does**:

Adds a new metric that measures bytes received before limits

Fixes https://github.com/grafana/tempo-squad/issues/848

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/distributor/distributor.go
```diff
@@ -87,6 +87,11 @@ var (
 		Name:      "distributor_bytes_received_total",
 		Help:      "The total number of proto bytes received per tenant",
 	}, []string{"tenant"})
+	metricIngressBytes = promauto.NewCounterVec(prometheus.CounterOpts{
```
javiermolinar, line 90: Being a bit picky, it seems we have a convention where the unit of measurement goes first:

`metricBytesIngested`

Although this seems incorrect, following Prometheus naming conventions. https://prometheus.io/docs/practices/naming/#metric-names


Instead of Ingress and received which may be confusing maybe we could use something like:


`metricBytesRecievedRaw`
`distributor_bytes_received_raw_total`

modules/distributor/distributor.go
```diff
@@ -87,6 +87,11 @@ var (
...
+		Namespace: "tempo",
+		Name:      "distributor_ingress_bytes_total",
+		Help:      "The total number of bytes received per tenant, before limits",
```
electron0zero, line 93: can we update the help text on `distributor_bytes_received_total` to mention that it's after limits (actual bytes ingested into the system).

also a nit: "before limits" is bit open ended, maybe "before rate limits" or "before ingestion limits" would be more clear here

modules/distributor/distributor.go
electron0zero: along with `distributor_bytes_received_raw_total` I want to throw in few more names into the mix:
- `tempo_distributor_bytes_attempted_total` since these are the bytes that are attempted (may or may not be ingested)
- `distributor_bytes_received_unfiltered_total` since we are capturing the bytes before ingestion limits are applied (i.e. filtering the data)

mapno: Being even _nitpickier_, `ingress bytes` can be considered the measurement unit in this metric. If we wanted a verb after the unit like in `distributor_bytes_received_total` it could be `distributor_ingress_bytes_received_total` or `distributor_bytes_ingressed_total`.

Ideally, we would rename `distributor_bytes_received_total` to `ingested` and use `distributor_bytes_received_total` for this new metric.

I don't have a strong opinion on this, I'm ok to go with whichever name sounds more clear for the team.

modules/distributor/distributor.go
mapno: Done!

modules/distributor/distributor.go
electron0zero: agree, I don't want to bikeshade on the name and okay with `tempo_distributor_ingress_bytes_total` as well.

PR #5602, opened 2025-09-03
service-graph: Support the new `db.namespace` attribute

gouthamve-but-enterprise:
**What this PR does**:

The latest DB semantic conventions have renamed the `db.name` attribute to `db.namespace`.

This updates the service graph to support the new attribute.

**Checklist**
- [x] Tests updated
- [x] Documentation added


mapno:
Could you add a changelog entry? 🙏

joe-elliott:
can you add a test here to confirm this string doesn't drift: 

https://github.com/grafana/tempo/blob/main/modules/generator/processor/servicegraphs/servicegraphs_test.go#L28-L35

gouthamve-but-enterprise:
I've updated the `CHANGELOG` and docs. 

@joe-elliott I already updated that test with the attribute, is there something that is missing?

joe-elliott:
Nope. I think i just missed it when I glanced at the PR. Apologies!

Thanks, Goutham 🙏

PR #5603, opened 2025-09-03
feat: improve truncated attributes visibility by logging them

javiermolinar:
**What this PR does**:

Currently we have a lack of visibility when big attributes (or attributes names) are trimmed due to the `max_attribute_bytes` limit. 

This pr aims to fix that by logging the truncated attributes once per batch.

```bash

level=warn ts=2025-09-03T14:37:25.753695Z caller=distributor.go:704 msg="attributes values truncated in batch" tenant=single-tenant truncated_total=5 truncated_attributes="[{\"attributeName\":\"my.span.attr\",\"originalSize\":85},{\"attributeName\":\"my.span.add\",\"originalSize\":85},{\"attributeName\":\"my.span.gg\",\"originalSize\":85},{\"attributeName\":\"my.span.ww\",\"originalSize\":85},{\"attributeName\":\"my.span.tt\",\"originalSize\":85}]"

```

It discriminates between attribute values and attributes names. The original code includes both of them in the same bucket. 

Fixes #<issue number>

**Checklist**
- [X] Tests updated

### DIFF
--- a/modules/distributor/distributor.go
+++ b/modules/distributor/distributor.go
@@ -694,8 +694,8 @@ func (d *Distributor) sendToKafka(ctx context.Context, userID string, keys []uin
 }
 
 type truncatedAttribute struct {
-	AttributeName string `json:"attributeName"`
-	OriginalSize  int    `json:"originalSize"`
+	AttributeName     string `json:"attributeName"`
+	OriginalSizeBytes int    `json:"originalSizeBytes"`
 }
 
 func logTruncatedAttributes(truncatedAttr []truncatedAttribute, message string, tenant string, l log.Logger) {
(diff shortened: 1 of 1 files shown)


modules/distributor/distributor.go
```diff
@@ -692,44 +693,55 @@ func (d *Distributor) sendToKafka(ctx context.Context, userID string, keys []uin
 	}, ring.DoBatchOptions{})
 }
 
+type truncatedAttribute struct {
+	AttributeName     string `json:"attributeName"`
+	OriginalSizeBytes int    `json:"originalSizeBytes"`
+}
+
+func logTruncatedAttributes(truncatedAttr []truncatedAttribute, message string, tenant string, l log.Logger) {
+	b, _ := json.Marshal(truncatedAttr)
```
mapno, line 702: Isn't this a lot of overhead in (potentially) every write for a log?

joe-elliott: yeah, i like this idea in theory, but i think this would destroy our distributors. we also need to consider that we may be logging sensitive information if we do this. 

perhaps some kind of temporary way to turn this on per tenant paired with some rate limiting?

PR #5605, opened 2025-09-03
feat: Add option to run tempo-vulture in validation mode

davidham:
**What this PR does**:
This PR adds a `validation` mode to tempo-vulture that allows it to be used as a post-deployment validation test. It does similar work to tempo-vulture, but instead of running continuously and reporting its results as metrics, it will create `n` traces (determined by the `--validation-cycles` flag) and attempt to read them back by ID, and query for them, and then exit. It will exit `0` if all operations succeed and exit `1` if any fail. 

I tested this against the `example/docker-compose/distributed` Docker Compose file. The search validations did not pass but I intend to try them against our federal dev environment and see if they work there.

Relates to grafana/deployment_tools#294598.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/integration/util/util.go
+++ b/integration/util/util.go
@@ -47,13 +47,13 @@ import (
 )
 
 const (
-	image            = "tempo:latest"
-	debugImage       = "tempo-debug:latest"
-	queryImage       = "tempo-query:latest"
-	jaegerImage      = "jaegertracing/jaeger-query:1.64.0"
-	prometheusImage  = "prom/prometheus:latest"
-	hdrXScopeOrgID   = "x-scope-orgid"
-	hdrAuthorization = "authorization"
+	image               = "tempo:latest"
+	debugImage          = "tempo-debug:latest"
+	queryImage          = "tempo-query:latest"
+	jaegerImage         = "jaegertracing/jaeger-query:1.64.0"
+	prometheusImage     = "prom/prometheus:latest"
+	xScopeOrgIDHeader   = "x-scope-orgid"
+	authorizationHeader = "authorization"
 )
 
 // GetExtraArgs returns the extra args to pass to the Docker command used to run Tempo.
@@ -413,10 +413,10 @@ func NewOtelGRPCExporterWithAuth(endpoint, orgID, basicAuthToken string, useTLS 
 	// Configure headers for authentication (gRPC metadata format)
 	headers := make(map[string]configopaque.String)
 	if orgID != "" {
-		headers[hdrXScopeOrgID] = configopaque.String(orgID)
+		headers[xScopeOrgIDHeader] = configopaque.String(orgID)
 	}
 	if basicAuthToken != "" {
-		headers[hdrAuthorization] = configopaque.String("Basic " + basicAuthToken)
+		headers[authorizationHeader] = configopaque.String("Basic " + basicAuthToken)
 	}
 
 	otlpCfg.ClientConfig = configgrpc.ClientConfig{


davidham:
Changing back to draft so I can add basic auth.

cmd/tempo-vulture/main.go
```diff
@@ -186,17 +197,83 @@ func main() {
...
+			return newStart, ts, skip
 		}
 		newStart, ts = selectPastTimestamp(startTime, newest, interval, vultureConfig.tempoRetentionDuration, r)
-		return
+		return newStart, ts, skip
+	}
+
+	// Check if we're running in validation mode
+	if validationMode {
+		accessPolicyToken := os.Getenv("TEMPO_ACCESS_POLICY_TOKEN")
```
zalegrala: Consider logging a warning if this variable is set and the validationMode is false.

Additionally, consider describing what kind of token is required for validation.

cmd/tempo-vulture/main.go
```diff
@@ -186,17 +197,83 @@ func main() {
...
-			return
+			return newStart, ts, skip
 		}
 		newStart, ts = selectPastTimestamp(startTime, newest, interval, vultureConfig.tempoRetentionDuration, r)
-		return
+		return newStart, ts, skip
+	}
+
+	// Check if we're running in validation mode
+	if validationMode {
```
zalegrala, line 218: Consider moving this out to its own function to clean up the `main()` a bit.

cmd/tempo-vulture/main.go
```diff
@@ -186,17 +197,83 @@ func main() {
...
+			zap.Int("cycles", validationCycles),
+			zap.Duration("timeout", validationTimeout),
+			zap.String("org_id", vultureConfig.tempoOrgID),
+		)
+
+		ctx := context.Background()
+
+		validationConfig := ValidationConfig{
+			Cycles:                validationCycles,
+			Timeout:               validationTimeout,
```
zalegrala: Timeout looks unused.  Consider using a context with timeout so that when you call RunValidation with the context, the cancellation handles the desired timeout.

cmd/tempo-vulture/main.go
```diff
@@ -186,17 +197,83 @@ func main() {
...
+		logger.Info("Validation completed",
+			zap.Int("total_traces", result.TotalTraces),
+			zap.Int("validations_passed", result.SuccessCount),
+			zap.Int("validations_failed", len(result.Failures)),
+			zap.Duration("duration", result.Duration),
+		)
+
+		// Optionally log each failure for debugging
+		for _, failure := range result.Failures {
+			logger.Error("Validation failure",
```
zalegrala: Consider lower-case errors everywhere.

cmd/tempo-vulture/validation.go
```diff
@@ -0,0 +1,286 @@
...
+				Cycle:     cycle,
+				Phase:     "search",
+				Error:     err,
+				Timestamp: traceInfo.timestamp,
+			})
+			continue
+		}
+
+		// Get a random attribute from the expected trace (same as original vulture)
+		attr := util.RandomAttrFromTrace(expected)
```
zalegrala: Is this sufficient for validation mode?  Consider executing a search for each attribute to confirm results.

zalegrala (commented):
Overall this is looking reasonable.  I left a few thoughts.

cmd/tempo-vulture/validation.go
```diff
@@ -0,0 +1,426 @@
...
+	"go.uber.org/zap"
+)
+
+func RunValidationMode(
+	ctx context.Context,
+	vultureConfig vultureConfiguration,
+	pushEndpoint string,
+	logger *zap.Logger,
+) int {
+	accessPolicyToken := os.Getenv("TEMPO_ACCESS_POLICY_TOKEN")
```
zalegrala: Consider making `TEMPO_ACCESS_POLICY_TOKEN` a constant.

cmd/tempo-vulture/validation.go
```diff
@@ -0,0 +1,426 @@
...
+	if accessPolicyToken == "" {
+		logger.Error("TEMPO_ACCESS_POLICY_TOKEN environment variable is required in validation mode")
```
zalegrala, line 27: Consider making this optional for use and testing without auth.

cmd/tempo-vulture/mocks.go
```diff
@@ -101,13 +102,17 @@ func (m *MockHTTPClient) QueryTrace(id string) (*tempopb.Trace, error) {
 
 //nolint:all
 func (m *MockHTTPClient) QueryTraceWithRange(id string, start int64, end int64) (*tempopb.Trace, error) {
+	return m.QueryTraceWithRangeContext(context.Background(), id, start, end)
+}
+
+func (m *MockHTTPClient) QueryTraceWithRangeContext(_ context.Context, _ string, _ int64, _ int64) (*tempopb.Trace, error) {
+	m.requestsCount++
```
zalegrala, line 107: This should be called under lock.  Consider moving the lock/defer to the top above this.

cmd/tempo-vulture/validation.go
```diff
@@ -0,0 +1,426 @@
...
+
+	writeErr := traceInfo.EmitAllBatches(writer)
+	if writeErr != nil {
+		return nil, nil, fmt.Errorf("failed to write trace, error: %w", err) // Any write failure is critical
+	}
+
+	return traceInfo, traceStructure, nil
+}
+
+func (vs *ValidationService) validateTraceRetrieval(
```
zalegrala, line 288: Consider dropping the context  from these methods if we don't need it.

cmd/tempo-vulture/main_test.go
```diff
@@ -420,7 +420,7 @@ func TestDoReadForAnErroredRead(t *testing.T) {
 	// Assert a read with errors
 	mockHTTPClient := MockHTTPClient{err: errors.New("an error"), traceResp: trace}
 	doRead(&mockHTTPClient, config, traceInfo, logger)
-	assert.Equal(t, 0, mockHTTPClient.GetRequestsCount())
+	assert.Equal(t, 1, mockHTTPClient.GetRequestsCount())
```
zalegrala, line 423: This is incrementing now because we count the erroring call, correct?

cmd/tempo-vulture/validation.go
```diff
@@ -0,0 +1,426 @@
...
+				Timestamp: vs.clock.Now(),
+			})
+		}
+
+		// sleep 1 sec to guarantee different timestamps on traces
+		if err := vs.sleepWithContext(ctx, 1*time.Second); err != nil {
+			// Context cancelled - add failure and return early
+			result.Failures = append(result.Failures, ValidationFailure{
+				Phase:     "timeout",
+				Error:     ctx.Err(),
```
zalegrala: I think you want `err` here since it is what's being returned.

cmd/tempo-vulture/validation.go
```diff
@@ -0,0 +1,426 @@
...
+					return fmt.Sprintf("%x", span.TraceId)
+				}
+			}
+		}
+	}
+	return ""
+}
+
+func (vs *ValidationService) logTraceAttributes(trace *tempopb.Trace, traceID string) {
+	vs.logger.Info("=== TRACE ATTRIBUTES DEBUG ===", zap.String("traceID", traceID))
```
zalegrala: Consider making this `Debug` level?  I'm not sure.

integration/util/util.go
```diff
@@ -402,16 +403,28 @@ func TempoBackoff() backoff.Config {
...
-func NewOtelGRPCExporter(endpoint string) (exporter.Traces, error) {
+func NewOtelGRPCExporterWithAuth(endpoint, orgID, basicAuthToken string, useTLS bool) (exporter.Traces, error) {
 	factory := otlpexporter.NewFactory()
 	exporterCfg := factory.CreateDefaultConfig()
 	otlpCfg := exporterCfg.(*otlpexporter.Config)
+
+	// Configure headers for authentication (gRPC metadata format)
+	headers := make(map[string]configopaque.String)
+	if orgID != "" {
+		headers["x-scope-orgid"] = configopaque.String(orgID)
```
zalegrala: Is there a constant for these header keys we an reference?

integration/util/util.go
```diff
@@ -402,16 +403,28 @@ func TempoBackoff() backoff.Config {
...
+	}
+	if basicAuthToken != "" {
+		headers["authorization"] = configopaque.String("Basic " + basicAuthToken)
+	}
+
 	otlpCfg.ClientConfig = configgrpc.ClientConfig{
 		Endpoint: endpoint,
 		TLS: configtls.ClientConfig{
-			Insecure: true,
+			Insecure: !useTLS,
```
zalegrala, line 425: I think there are two concepts here.  One is weather to enable TLS.  The other is to trust it.  It might be nice to implement both if we are going to be using TLS at all.

pkg/util/trace_info.go
```diff
@@ -105,8 +105,12 @@ func (t *TraceInfo) Done() {
 }
 
 func (t *TraceInfo) EmitBatches(c JaegerClient) error {
+	return t.EmitBatchesWithContext(context.Background(), c)
+}
+
+func (t *TraceInfo) EmitBatchesWithContext(ctx context.Context, c JaegerClient) error {
```
zalegrala: It appears that this is only called from above.  Do we need two methods?

For further consideration, I think adjusting the existing methods to receive a context is perfectly reasonable.  If its too much churn to update the existing calls, I understand, but I don't expect that we need methods which are identical but one receives a context and the other doesn't.  Callers who don't have a context can always pass a new one.

cmd/tempo-vulture/main_test.go
davidham: Correct.

pkg/util/trace_info.go
davidham: I merged all of my new methods and used the old names, and I updated the callers to pass in `context.Background()`.

integration/util/util.go
davidham: Would it be OK to leave as-is for now? For our use case we will need both or neither. I can add if you feel it's important though.

zalegrala: Its fine to come back.

PR #5607, opened 2025-09-04
[Bugfix] 

ruslan-mikhailov:
## What this PR does

How to reproduce
1. send request to `api/v2/search/tags` with invalid query. Example:
```
--> GET http://localhost:3100/api/v2/search/tags
--> query params: {'q': '{ a="test" }'}
<-- status: 500
<-- body: error querying ingesters in Querier.SearchTags: failed to execute f() for x.x.x.x:x: rpc error: code = Unknown desc = recovered in SearchTagsV2
2. Try to search by trace id, it will always return 500. Example:

```
--> GET http://localhost:3100/api/traces/19a8654f9aaa456b7e45d19ea770ac03
<-- status: 500
<-- body: error querying ingesters in Querier.FindTraceByID: context deadline exceeded
```

## Which issue(s) this PR fixes
Fixes #<issue number>

## Checklist
- [ ] Tests updated
- [ ] Documentation added

### CODE REVIEW

File: modules/ingester/instance_search.go [hunk 1]
```diff
@@ -269,10 +269,12 @@ func (i *instance) SearchTagsV2(ctx context.Context, req *tempopb.SearchTagsRequ
 		}, fetcher)
 	}
 
-	i.headBlockMtx.RLock()
-	span.AddEvent("acquired headblock mtx")
-	err = searchBlock(ctx, i.headBlock, "headBlock")
-	i.headBlockMtx.RUnlock()
+	func() {
```
electron0zero: I am not sure if I understand why we need to wrap this in func()?  can you explain why we need this?

also let's add a comment explaining this so we capture the context.

pkg/traceql/engine_test.go
```diff
@@ -562,6 +562,29 @@ func TestExamplesInEngine(t *testing.T) {
 	}
 }
 
+func TestExecuteTagNames_InvalidQuery(t *testing.T) {
```
electron0zero, line 565: let's add an e2e test for this as well, to fully exercise the whole pipeline.

modules/ingester/instance_search.go
ruslan-mikhailov: That's an approach to defer Unlock after certain operations, not after the whole function execution. 

Example: https://go.dev/play/p/t8xioJ3eojs 
Same approach in live-store: https://github.com/ruslan-mikhailov/tempo/blob/main/modules/livestore/instance_search.go#L79-L82

ruslan-mikhailov: Added a comment

pkg/traceql/engine_test.go
ruslan-mikhailov: Added a test for invalid query

PR #5616, opened 2025-09-05
Fix live-store shutdown logic

mapno:
**What this PR does**:

1. Clears complete operations after they are finished.
1. Fixes a deadlock in `stopping` that impedes a clean shutdown.

```go

func (s *LiveStore) stopping(error) error {
	// stop consuming
	...

	// flush all data to disk
	...

	s.stopAllBackgroundProcesses() // stops the workers that pull from completeQueues
        
	...

	for !s.completeQueues.IsEmpty() { // this creates a deadlock - the queue can't become empty because the workers processing it have been stopped
		...
	}
```

3. Adds an extra condition during shutdown so it exits in tests as before.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/livestore/live_store.go
+++ b/modules/livestore/live_store.go
@@ -308,6 +308,10 @@ func (s *LiveStore) stopping(error) error {
 	// Flush all data to disk
 	s.cutAllInstancesToWal()
 
+	if s.cfg.holdAllBackgroundProcesses { // nothing to do
+		return nil
+	}
+
 	ticker := time.NewTicker(100 * time.Millisecond)
 	defer ticker.Stop()
 


modules/livestore/live_store.go
```diff
@@ -325,6 +327,8 @@ func (s *LiveStore) stopping(error) error {
 		}
 	}
 
+	s.stopAllBackgroundProcesses()
```
joe-elliott, line 330: should we call `s.completeQueues.Stop()`? 

looking at the code I actually don't know if it would do anything which is odd, but it seems appropriate.

modules/livestore/live_store_background.go
```diff
@@ -117,6 +117,7 @@ func (s *LiveStore) globalCompleteLoop(idx int) {
 			}()
 		} else {
 			metricBlocksCompleted.Inc()
+			s.completeQueues.Clear(op)
```
joe-elliott, line 120: that's a miss. good find.

joe-elliott (approved):

modules/livestore/live_store.go
mapno: It's called inside that function 😄

joe-elliott: i'm losing my mind! thx mario 🙏

PR #5617, opened 2025-09-05
Add metric for livestore

mattdurham:
Add metric for trying to determine the lag in pulling from kafka.

### DIFF
--- a/modules/livestore/live_store.go
+++ b/modules/livestore/live_store.go
@@ -92,6 +92,12 @@ var (
 		Name:      "records_dropped_total",
 		Help:      "The total number of kafka records dropped per tenant.",
 	}, []string{"tenant", "reason"})
+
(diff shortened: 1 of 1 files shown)


modules/livestore/live_store.go
```diff
@@ -383,6 +389,9 @@ func (s *LiveStore) consume(ctx context.Context, rs recordIter, now time.Time) (
 		return nil, nil
 	}
 
+	// Update metric with the timestamp of the most recent kafka record
+	metricMostRecentKafkaRecord.Set(float64(lastRecord.Timestamp.Unix()))
```
joe-elliott, line 393: i think partition lag would be more interesting. that's what the blockbuilder does:

https://github.com/grafana/tempo/blob/main/modules/blockbuilder/blockbuilder.go#L387

PR #5625, opened 2025-09-08
feat: implement trace total size logging

sienna011022:
**What this PR does**:

Adds total size logging functionality to track and log the cumulative size of traces processed by the ingester. 
This enhancement provides better observability into trace data volume and helps with monitoring resource usage.

**Checklist**
- [X] Tests updated
- [ ] Documentation added


pkg/tracesizes/traceSizes.go
```diff
@@ -63,3 +63,12 @@ func (s *Tracker) ClearIdle(idleSince time.Time) {
 		}
 	}
 }
+
+func (s *Tracker) GetCurrentSize(traceID []byte) int {
+	token := s.token(traceID)
```
joe-elliott, line 68: you need to take the lock in this func to prevent concurrent map access. wdyt about just returning the trace size from `.Allow()` instead? it may be worth benchmarking the two approaches if you feel comfortable doing that.

sienna011022: @joe-elliott thanks I made another [PR](https://github.com/grafana/tempo/pull/5628) and I did benchmarking.
It will be efficient if Allow() return the size

joe-elliott:
Closing in favor of https://github.com/grafana/tempo/pull/5628/files

PR #5626, opened 2025-09-08
operations: make TempoDistributorUnhealthy alert critical

electron0zero:
**What this PR does**:

makes TempoDistributorUnhealthy a critical alerts instead of warning.

this alert should be critical, like the similar alert for other components. 

I found this when I saw that distributor was Unhealthy for long time and I was not paged for it because this alert was not at critical.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


electron0zero:
talked about it, and Unhealthy Distributor doesn't impact the tempo's operations so we don't need to make it critical and wake someone up in the middle of night.

PR #5627, opened 2025-09-08
Add context to the global loop

mattdurham:


joe-elliott:
The queues being empty should signal this loop to exit. On shutdown the queues are closed and then the livestores draw down their queues before exiting. This is the same design as the ingesters. Are we looking to do something different here?

https://github.com/grafana/tempo/blob/main/modules/livestore/live_store_background.go#L77-L80

mattdurham:
I am unsure if its the same, the ingester looks to have explicit retry conditions that are triggered instead of relying only on max flushes. Currently a pod takes 10m to shutdown from sigint. Its possible its not this code, but it does feel to pivot around the completeQueues in some way.

joe-elliott:
Possibly this code in stopping:
```
	for !s.completeQueues.IsEmpty() {
		select {
		case <-ticker.C:
		case <-timeout.C:
			level.Error(s.logger).Log("msg", "flush remaining blocks timed out")
			return nil // shutdown timeout reached
		}
	}

	s.stopAllBackgroundProcesses()
```

we are waiting for completeQueues to be empty before we are calling queues.Stop. we should stop the queues then wait for them to be empty. also heads up that @mapno just made changes here:

https://github.com/grafana/tempo/commit/68f8e2af315524d39db3bb22e62977ac0a6fe3c1

PR #5630, opened 2025-09-09
[live-store] Read from snapshot

mapno:
**What this PR does**:

Allow for concurrent reads from a snapshot instead of juggling RWMutex locks

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/livestore/instance.go
+++ b/modules/livestore/instance.go
@@ -86,8 +86,7 @@ type instance struct {
 	wal *wal.WAL
 	enc encoding.VersionedEncoding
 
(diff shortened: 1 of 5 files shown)


joe-elliott:
this adds a lot of complexity and we've never needed anything like this before. any particular reason we are going this path?

mapno:
Closing. We're not pursuing this path unless locking becomes an unsolvable bottleneck. While this approach is memory safe, it doesn't address potential races of the data in disk.

PR #5631, opened 2025-09-09
Add backpressure mechanism in live-store

mapno:
**What this PR does**:

Adds a backpressure mechanism, almost identical to the one in the [generators](https://github.com/grafana/tempo/blob/main/modules/generator/processor/localblocks/processor.go#L168-L206). It waits for "room in the pipeline" by pushing back on pushes if 1) live traces are at the limit for the tenant or 2) there is more than 1 WAL block waiting to be completed.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/sources/tempo/configuration/manifest.md
+++ b/docs/sources/tempo/configuration/manifest.md
@@ -1235,6 +1235,7 @@ live_store:
     flush_op_timeout: 5m0s
     max_trace_live: 30s
     max_trace_idle: 5s
+    max_live_traces_bytes: 250000000
     max_block_duration: 30m0s
     max_block_bytes: 524288000
     block_config:


joe-elliott (approved):
@mattdurham should review as well, but since this was copied as is from live store i'm good.

i do think long term we will likely tweak this but you gotta love the simplicity

modules/livestore/instance.go
```diff
@@ -125,12 +135,64 @@ func newInstance(instanceID string, cfg Config, wal *wal.WAL, overrides override
 	return i, nil
 }
 
-func (i *instance) pushBytes(ts time.Time, req *tempopb.PushBytesRequest) {
+func (i *instance) backpressure(ctx context.Context) bool {
```
mattdurham, line 138: Bikeshedding, thoughts on making the name something like needsBackpressure? Overall not to hung up on the naming though.

modules/livestore/instance.go
```diff
@@ -125,12 +135,64 @@ func newInstance(instanceID string, cfg Config, wal *wal.WAL, overrides override
...
+	return false
+}
+
+func (i *instance) pushBytes(ctx context.Context, ts time.Time, req *tempopb.PushBytesRequest) {
 	if len(req.Traces) != len(req.Ids) {
 		level.Error(i.logger).Log("msg", "mismatched traces and ids length", "IDs", len(req.Ids), "traces", len(req.Traces))
 		return
 	}
 
+	// Wait for room in pipeline if needed
```
mattdurham, line 186: Thoughts on moving the select block from inside the func to inside this for?  That way the backpressure func is only determining if backpressure should be applied and the for loop doesnt look nearly as odd.

mdisibio: We are metric'ing the time asleep separately for each case.  So it would need to return a flag to indicate the reason to the outer loop too.  Then it ended up just simpler to do it this way, trivial loops in either case.

modules/livestore/instance.go
joe-elliott: i like backpressure. also we should paint the shed blue ;)

mapno:
Changed to backpressure on bytes, it translates better to memory

PR #5633, opened 2025-09-09
Commit changes to prevent mutex locking

mattdurham:
Removed the blocking out of iterateBlocks and depend on the parent blocking. Its not the most intuitive but this is a relatively low rate of code change to fix the issue.

### DIFF
--- a/modules/livestore/instance_search.go
+++ b/modules/livestore/instance_search.go
@@ -77,9 +77,6 @@ func (i *instance) iterateBlocks(ctx context.Context, reqStart, reqEnd time.Time
 	}
 
 	func() {
-		i.blocksMtx.RLock()
-		defer i.blocksMtx.RUnlock()
-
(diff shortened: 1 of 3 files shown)


modules/livestore/instance_search.go
```diff
@@ -251,6 +246,9 @@ func (i *instance) Search(ctx context.Context, req *tempopb.SearchRequest) (*tem
 		return nil
 	}
 
+	i.blocksMtx.RLock()
```
joe-elliott: we should leave this in `iterateBlocks()` and just remove the lock in `QueryRange`. are you saying the livestore still locks with that change?

mapno: +1, I think it's simpler code if locking is in `iterateBlocks()`

mapno (approved):

PR #5634, opened 2025-09-09
Move tracing install to pkg/tracing

zalegrala:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


cmd/tempo/build/build.go
```diff
@@ -5,26 +5,6 @@ import (
...
-
-func init() {
-	version.Version = Version
-	version.Revision = Revision
-	version.Branch = Branch
-	version.BuildUser = BuildUser
-	version.BuildDate = BuildDate
-}
-
 func GetVersion() prom.PrometheusVersion {
```
zalegrala, line 8: There is only one call which is in the `/status/version` handler.  This could also be replaced with the `version.Info()` call we see in `main()` and then this entire package could get deleted.  Thoughts?

zalegrala: I overlooked the usage stats handler, and I don't want to take that on in this PR.

PR #5638, opened 2025-09-10
Allow per-tenant configuration of native histograms

zalegrala:
**What this PR does**:

Here we add three new override configurations which allow the tuning of the buckets in the native histograms implementaion in the metrics-generator.  This allows tenants of vartying needs to keep their histograms more relevant to their use cases.
The three new configurations are:
- `native_histogram_bucket_factor` - the minimum required distance between buckets
- `native_histogram_max_bucket_number` - the maximum number of buckets to create
- `native_histogram_min_reset_duration` - the minimum duration before a histogram is reset

The value of these configurations are passed directly to the prometheus client library when creating a new histogram.  More detail on these parameters can be found in the [client documentation](https://github.com/prometheus/client_golang/blob/14ccb93091c00f86b85af7753100aa372d63602b/prometheus/histogram.go?plain=1#L405).

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/registry/native_histogram.go
+++ b/modules/generator/registry/native_histogram.go
@@ -306,7 +306,6 @@ func (h *nativeHistogram) hashOverrides() (uint64, float64, uint32, time.Duratio
 	)
 
 	hsh := fnv.New64a()
-	hsh.Write([]byte(h.tenant))
 
 	_ = binary.Write(hsh, binary.LittleEndian, bucketFactor)
 	_ = binary.Write(hsh, binary.LittleEndian, maxBucketNum)
--- a/modules/generator/registry/native_histogram.go
+++ b/modules/generator/registry/native_histogram.go
@@ -273,6 +273,10 @@ func (h *nativeHistogram) removeStaleSeries(staleTimeMs int64) {
 	h.seriesMtx.Lock()
 	defer h.seriesMtx.Unlock()
 	for hash, s := range h.series {
+		// TODO: native histogram bucket configuration is only checked when the
+		// series is created.  We need a way to determine if the overrides has
+		// changed, and recreate the series if needed.  Without this, the bucket
+		// config is only set at startup.
 		if s.lastUpdated < staleTimeMs {
 			delete(h.series, hash)
 			h.onRemoveSerie(h.activeSeriesPerHistogramSerie())
--- a/modules/generator/registry/native_histogram.go
+++ b/modules/generator/registry/native_histogram.go
@@ -41,6 +41,10 @@ type nativeHistogram struct {
 	// reload of the process, and a new instance of the histogram to be created.
 	histogramOverride HistogramMode
 
+	overrides Overrides
+	// The tenant for this registry instance is received at create time and does not change.
+	tenant string
+
 	externalLabels map[string]string
 
 	// classic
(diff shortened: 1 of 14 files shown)


modules/generator/registry/native_histogram.go
```diff
@@ -260,20 +272,44 @@ func (h *nativeHistogram) collectMetrics(appender storage.Appender, timeMs int64
...
+func (h *nativeHistogram) hashOverrides() (uint64, float64, uint32, time.Duration) {
+	var (
+		bucketFactor = h.overrides.MetricsGeneratorNativeHistogramBucketFactor(h.tenant)
+		maxBucketNum = h.overrides.MetricsGeneratorNativeHistogramMaxBucketNumber(h.tenant)
+		minResetDur  = h.overrides.MetricsGeneratorNativeHistogramMinResetDuration(h.tenant)
+	)
+
+	s := fmt.Sprintf("%s%f%d%s", h.tenant, bucketFactor, maxBucketNum, minResetDur)
+	hsh := fnv.New64a()
+	hsh.Write([]byte(s))
```
carles-grafana: is the tenant necessary for the hash if it's always the same?

Also, if I recall correctly it's better to write the hash elements one by one instead of using Sprintf and concatenating them to avoid collisions.

zalegrala: The tenant isn't strictly required since the we're keeping the hash as part of the object, but I threw it in there since it was available and I'm not aware of a cost penalty.

I'm not sure what you mean about writing the objects one at a time.  Is there a `[]byte` conversion for float which is better?  I see I could use `encoding/binary`, but the implementation looks more involved than the simple string printing here.

Are you seeing a simpler approach to this?

carles-grafana: The potential problem is concatenation ambiguity, although not sure how likely it is in this instance.

It's true that the safest alternatively is more involved, here's an AI-generated alternative that is more deterministic:
```
  func (h *nativeHistogram) hashOverrides() (uint64, float64, uint32, time.Duration) {
  	bucketFactor := h.overrides.MetricsGeneratorNativeHistogramBucketFactor(h.tenant)
  	maxBucketNum := h.overrides.MetricsGeneratorNativeHistogramMaxBucketNumber(h.tenant)
  	minResetDur := h.overrides.MetricsGeneratorNativeHistogramMinResetDuration(h.tenant)

  	hsh := fnv.New64a()
  	hsh.Write([]byte(h.tenant))

  	var b8 [8]byte
  	binary.LittleEndian.PutUint64(b8[:], math.Float64bits(bucketFactor))
  	hsh.Write(b8[:])

  	var b4 [4]byte
  	binary.LittleEndian.PutUint32(b4[:], uint32(maxBucketNum))
  	hsh.Write(b4[:])

  	binary.LittleEndian.PutUint64(b8[:], uint64(int64(minResetDur)))
  	hsh.Write(b8[:])

  	return hsh.Sum64(), bucketFactor, maxBucketNum, minResetDur
  }
```

This ensures 100% hash correctness but might be overkill, up to you.

zalegrala: Interesting.  Thanks for the feedback.  I've pushed a change for this which is a little simpler than the above, what do you think?

joe-elliott:
iirc we talked about adjusting defaults so that the number of series produced by the native histogram is roughly the same as produced by the classic histogram defaults. is that change included here?

zalegrala:
I didn't change the defaults here, since now its an override, which can be adjusted for the cluster.  I was thinking for our internal environments we'd make that adjustment, but I can include a lower default if we prefer.

modules/generator/registry/native_histogram.go
zalegrala: I've pushed a change based on this feedback and we chatted.

carles-grafana (approved):

PR #5639, opened 2025-09-11
vParquet5: dedicated columns for int values

### LINKED ISSUE #5636: vParquet5: dedicated attribute columns for integer values
Add support for dedicated attribute columns for integer values on resource and span scope.

stoewer:
**What this PR does**:

Add dedicated columns for integer values. I've added 5 spare columns on resource level and 5 on span level, but it needs to be discussed how many we spare integer columns we want to have.

Fixes #5636

**Checklist**
- [x] Tests updated
- [ ] Documentation added


stoewer:
> Did we want address tempo-cli to choose ints in this PR or a followup?

At this point I'm not quite sure what the best metrics are for choosing good int attributes for dedicated columns. Therefore I'd prefer doing this in a separate PR.

PR #5640, opened 2025-09-11
Support querying live-stores via header

mapno:
**What this PR does**:

It adds support to query the live stores for recent data instead of ingesters by passing the header `Recent-Data-Target` with value `live-store`. This allows to run both modules at the same time and query one or the other per request.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/querier/http.go
+++ b/modules/querier/http.go
@@ -137,7 +137,7 @@ func (q *Querier) SearchHandler(w http.ResponseWriter, r *http.Request) {
 	ctx, cancel := context.WithDeadline(r.Context(), time.Now().Add(q.cfg.Search.QueryTimeout))
 	defer cancel()
 
-	ctx = context.WithValue(ctx, recentDataTargetContextKey, api.ParseRecentDataTargetHeader(r))
+	ctx = injectRecentDataTarget(ctx, api.ParseRecentDataTargetHeader(r))
 
 	ctx, span := tracer.Start(ctx, "Querier.SearchHandler")
 	defer span.End()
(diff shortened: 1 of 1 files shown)


pkg/api/http.go
```diff
@@ -61,10 +61,12 @@ const (
...
-	HeaderAccept         = "Accept"
-	HeaderContentType    = "Content-Type"
-	HeaderAcceptProtobuf = "application/protobuf"
-	HeaderAcceptJSON     = "application/json"
+	HeaderAccept                    = "Accept"
+	HeaderContentType               = "Content-Type"
+	HeaderAcceptProtobuf            = "application/protobuf"
+	HeaderAcceptJSON                = "application/json"
+	HeaderRecentDataTarget          = "X-Recent-Data-Target"
+	HeaderRecentDataTargetLiveStore = "live-store"
```
mattdurham: Is this value used anywhere?

mapno: Nope, removed

pkg/api/http.go
```diff
@@ -61,10 +61,11 @@ const (
...
 
-	HeaderAccept         = "Accept"
-	HeaderContentType    = "Content-Type"
-	HeaderAcceptProtobuf = "application/protobuf"
-	HeaderAcceptJSON     = "application/json"
+	HeaderAccept           = "Accept"
+	HeaderContentType      = "Content-Type"
+	HeaderAcceptProtobuf   = "application/protobuf"
+	HeaderAcceptJSON       = "application/json"
+	HeaderRecentDataTarget = "X-Recent-Data-Target"
```
mattdurham: nit: Isn't the X prefix against the preferences these days for headers?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers

mapno: Yeah, but we use that convention within Tempo and Grafana. I wasn't sure what to do in this case. It's a temporary header, so I didn't think much about it. Removed.

modules/querier/http.go
```diff
@@ -133,6 +137,8 @@ func (q *Querier) SearchHandler(w http.ResponseWriter, r *http.Request) {
 	ctx, cancel := context.WithDeadline(r.Context(), time.Now().Add(q.cfg.Search.QueryTimeout))
 	defer cancel()
 
+	ctx = context.WithValue(ctx, recentDataTargetContextKey, api.ParseRecentDataTargetHeader(r))
+
```
ruslan-mikhailov: Here and below, why not `injectRecentDataTarget`?

PR #5641, opened 2025-09-11
Add jitter to queuing blocks in the live-store

mapno:
**What this PR does**:

Adds jitter to enqueuing blocks for completition in the live stores. This eases the memory pressure at startup and makes live-stores more stable in high-load scenarios.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/livestore/live_store.go
+++ b/modules/livestore/live_store.go
@@ -446,11 +446,7 @@ func (s *LiveStore) cutOneInstanceToWal(inst *instance, immediate bool) {
 
 	// If head block is cut, enqueue complete operation
 	if blockID != uuid.Nil {
-		err = s.enqueueCompleteOp(inst.tenantID, blockID, false)
-		if err != nil {
-			level.Error(s.logger).Log("msg", "failed to enqueue complete operation", "tenant", inst.tenantID, "err", err)
-			return
-		}
+		s.enqueueCompleteOp(inst.tenantID, blockID)
 	}
 }
 
--- a/modules/livestore/live_store.go
+++ b/modules/livestore/live_store.go
@@ -446,7 +446,11 @@ func (s *LiveStore) cutOneInstanceToWal(inst *instance, immediate bool) {
 
 	// If head block is cut, enqueue complete operation
 	if blockID != uuid.Nil {
-		s.enqueueCompleteOp(inst.tenantID, blockID)
+		err = s.enqueueCompleteOp(inst.tenantID, blockID, false)
+		if err != nil {
+			level.Error(s.logger).Log("msg", "failed to enqueue complete operation", "tenant", inst.tenantID, "err", err)
+			return
+		}
 	}
 }
 
--- a/modules/livestore/live_store.go
+++ b/modules/livestore/live_store.go
@@ -446,7 +446,7 @@ func (s *LiveStore) cutOneInstanceToWal(inst *instance, immediate bool) {
 
 	// If head block is cut, enqueue complete operation
 	if blockID != uuid.Nil {
-		err = s.enqueueCompleteOp(inst.tenantID, blockID)
+		err = s.enqueueCompleteOp(inst.tenantID, blockID, false)
 		if err != nil {
 			level.Error(s.logger).Log("msg", "failed to enqueue complete operation", "tenant", inst.tenantID, "err", err)
 			return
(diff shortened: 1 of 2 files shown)


modules/livestore/live_store.go
```diff
@@ -446,7 +446,7 @@ func (s *LiveStore) cutOneInstanceToWal(inst *instance, immediate bool) {
 
 	// If head block is cut, enqueue complete operation
 	if blockID != uuid.Nil {
-		err = s.enqueueCompleteOp(inst.tenantID, blockID)
```
javiermolinar, line 449: What about passing the jitter from the outside? That way it's more understandable and testable. 0 means no jitter

modules/livestore/live_store_background.go
```diff
@@ -156,14 +158,31 @@ func (s *LiveStore) perTenantCleanupLoop(inst *instance) {
 	}
 }
 
-func (s *LiveStore) enqueueCompleteOp(tenantID string, blockID uuid.UUID) error {
-	return s.enqueueOp(&completeOp{
+func (s *LiveStore) enqueueCompleteOp(tenantID string, blockID uuid.UUID, jitter bool) error {
```
mattdurham, line 160: Should we always have jitter? IE can we remove this and always use the jitter? Theoretically could hit our throughput but we likely want this under heavy load also, not just on startup?

modules/livestore/live_store.go
mapno: I've changed to always add jitter

modules/livestore/live_store_background.go
mapno: Done!

joe-elliott: sry, but i'm concerned about this change. the jitter function waits up to 10s to enqueue a message and i think its important that we do it synchronously during normal operations. with this change we could:

1) call enqueueOperation
2) background goroutine starts up with long jitter wait
3) SIGTERM received 
4) process shuts down even though there should have been something in queue


this is fine during startup b/c it allows for a clean startup w/ a lot of jobs, but i think we should leave the old code normally.

these patterns are well established from years of operational experience in generator/ingester. in an effort to get livestore out and functional i think we should just copy as directly as possible in most cases.

mapno: That situation is fine as it'll be queued again during replay.

I don't have a strong opinion on this. I dislike spawning a goroutine, but some jitter is nice.

mattdurham: Isnt this jitter waiting in the neighborhood of nano to microseconds? Though should probably be something like `math.Intn(1_000) * time.Millisecond` so its more constrained. The above can even overflow depending on the rand value.

joe-elliott: > Isnt this jitter waiting in the neighborhood of nano to microseconds? 

nice catch. @mapno we should correct flushJitter const in this PR to be like the ingesters:

https://github.com/grafana/tempo/blob/main/modules/ingester/flush.go#L65

my opinion is that jitter is important at startup to prevent memory spikes due to synchronized completions, but after things have settled natural jitter in the process works well. this pattern has worked for ingesters for years. is there a subtle bug with async enqueues during normal operation? probably not but maybe. is there a strong reason to add jitter to enqueues during normal operation? we don't need it in ingesters so i don't think so.

but i will not block this PR if that's what you all want.

> I dislike spawning a goroutine

+1

mapno: Fixed both things: 1) reverted the changes to jitter only during startup and 2) changed the delay calculation

modules/livestore/live_store_background.go
```diff
@@ -156,14 +157,31 @@ func (s *LiveStore) perTenantCleanupLoop(inst *instance) {
...
+
+	if jitter {
+		return s.enqueueOpWithJitter(op)
+	}
+
+	return s.enqueueOp(op)
+}
+
+func (s *LiveStore) enqueueOpWithJitter(op *completeOp) error {
+	delay := time.Duration(rand.Int64N(1_000) * int64(time.Millisecond)) //gosec:disable G404 — It doesn't require strong randomness
```
joe-elliott: why don't we use 10s like the ingesters?

PR #5647, opened 2025-09-12
Allow multiple compaction provider instances

zalegrala:
**What this PR does**:

Here we modify the compaction provider to operate on multiple tenants at once.

Currently, the scheduler only hands out jobs for a single tenant at a time.  To avoid a single tenant monopolizing the workers, we have a `max_jobs_per_tenant` setting which limits the number of sequential jobs for a single tenant before moving on.  However, in high-tenant count environments, the cycle time back around means that the outstanding blocks can grow for a large number of tenants and their level 0 blocks can grow very large.  When combined with autoscale based on the outstanding blocks, this can lead to overscale.

By working on multiple tenants at once, we can ensure that the level 0 blocks for all tenants are kept in check, allowing the autoscale to be more effective.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/sources/tempo/configuration/manifest.md
+++ b/docs/sources/tempo/configuration/manifest.md
@@ -1059,6 +1059,7 @@ backend_scheduler:
             max_input_blocks: 4
             max_compaction_level: 0
             min_cycle_interval: 30s
+            max_concurrent_tenants: 1
     job_timeout: 15s
     local_work_path: /var/tempo
 backend_scheduler_client:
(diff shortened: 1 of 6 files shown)


docs/sources/tempo/configuration/manifest.md
```diff
@@ -1059,6 +1059,7 @@ backend_scheduler:
             max_input_blocks: 4
             max_compaction_level: 0
             min_cycle_interval: 30s
+            max_concurrent_tenants: 1
```
knylander-grafana, line 1062: Do we need to add more doc for this? 

Do we need a changelog entry?

PR #5650, opened 2025-09-15
[live-store] Decouple consuming and committing from/to Kafka

mapno:
**What this PR does**:

Adds the ability to decouple committing and consuming from Kafka in the live-stores.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingest/testkafka/cluster.go
+++ b/pkg/ingest/testkafka/cluster.go
@@ -70,7 +70,7 @@ func (c *Cluster) ensureConsumerGroupExists(consumerGroup string) {
 	}
 	c.committedOffsets[consumerGroup] = make([]int64, c.numPartitions+1)
 
-	// Initialise the Partition offsets with the special value -1 which means "no offset committed".
+	// Initialise the partition offsets with the special value -1 which means "no offset committed".
 	for i := 0; i < len(c.committedOffsets[consumerGroup]); i++ {
 		c.committedOffsets[consumerGroup][i] = -1
 	}
@@ -93,7 +93,7 @@ func (c *Cluster) offsetCommit(request kmsg.Request) (kmsg.Response, error, bool
 	require.Len(c.t, commitR.Topics, 1, "test only has support for one topic per request")
 	topic := commitR.Topics[0]
 	require.Equal(c.t, c.topic, topic.Topic)
-	require.Len(c.t, topic.Partitions, 1, "test only has support for one Partition per request")
+	require.Len(c.t, topic.Partitions, 1, "test only has support for one partition per request")
 
 	partitionID := topic.Partitions[0].Partition
 	c.committedOffsets[consumerGroup][partitionID] = topic.Partitions[0].Offset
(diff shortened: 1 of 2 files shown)


modules/livestore/partition_reader.go
```diff
@@ -91,21 +102,59 @@ func (r *PartitionReader) running(ctx context.Context) error {
...
 }
 
+func (r *PartitionReader) commitLoop(ctx context.Context) {
+	defer r.wg.Done()
+
+	if r.commitInterval == 0 { // Sync commits
+		return
+	}
+
+	t := time.NewTicker(5 * time.Second)
```
javiermolinar: Should not be 

```go
t := time.NewTicker(r.commitInterval)
```

modules/livestore/partition_reader.go
```diff
@@ -91,21 +102,59 @@ func (r *PartitionReader) running(ctx context.Context) error {
...
+	defer t.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			// Commit one last time before shutting down
+			func() {
+				// Detach context with a deadline
+				ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(30*time.Second))
```
javiermolinar, line 149: We could use a constant for the deadline time

modules/livestore/partition_reader.go
```diff
@@ -91,21 +102,59 @@ func (r *PartitionReader) running(ctx context.Context) error {
...
+
 	r.client.Close()
 
 	return nil
 }
 
+func (r *PartitionReader) commitLoop(ctx context.Context) {
+	defer r.wg.Done()
+
+	if r.commitInterval == 0 { // Sync commits
```
javiermolinar, line 134: Can't this goroutine to store a copy of the last offset? That way we won't need a second atomic bool to know if the value was updated.

javiermolinar (commented):
I think this is good to increase the throughput by reducing the number of Kafka requests as long as we can live with duplication in case of an error

modules/livestore/partition_reader.go
mapno: Yes, forgot to update this

modules/livestore/partition_reader.go
mapno: I'm not sure what you mean by this, can you elaborate?

javiermolinar: Sure, the `hasUncommittedOffsets atomic.Bool` is used to flag if there are uncommitted offsets or not. It could be possible to store a copy of the last watermark locally

```go
offset := r.offsetWatermark.Load()
if offset == nil || offset == lastOffset{
   return
}
```

That way, `hasUncommittedOffsets` wouldn't be necessary

mapno: Ah, I see. Done!

pkg/ingest/testkafka/cluster.go
```diff
@@ -70,7 +70,7 @@ func (c *Cluster) ensureConsumerGroupExists(consumerGroup string) {
 	}
 	c.committedOffsets[consumerGroup] = make([]int64, c.numPartitions+1)
 
-	// Initialise the partition offsets with the special value -1 which means "no offset committed".
+	// Initialise the Partition offsets with the special value -1 which means "no offset committed".
```
mattdurham: nit: Is this capitalization intentional?

mapno: lol no, prob from a refactor in Goland

mapno: Fixed

PR #5652, opened 2025-09-15
Add support for metrics

mattdurham:
This adds several metrics to tracking live store including using the partition lag like ingest/blockbuilder. Followup PR will switch ingest/blockbuilder to this style but it was a bit to large for one pr.

### DIFF
--- a/modules/livestore/partition_reader_test.go
+++ b/modules/livestore/partition_reader_test.go
@@ -100,14 +100,16 @@ func defaultPartitionReaderWithCommitInterval(t *testing.T, address string, comm
 	cfg.Topic = testTopic
 	cfg.ConsumerGroup = testConsumerGroup
 
+	reg := prometheus.NewRegistry()
+
 	client, err := ingest.NewReaderClient(
 		cfg,
-		ingest.NewReaderClientMetrics(liveStoreServiceName, prometheus.NewRegistry()),
+		ingest.NewReaderClientMetrics(liveStoreServiceName, reg),
 		l,
 	)
 	require.NoError(t, err)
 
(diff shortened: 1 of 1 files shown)


joe-elliott:
Since we will have 3 processes that are all consuming from kafka we should consolidate these metrics in one place and confirm that all 3 produce the exact same metrics (related to kafka consumption).

mattdurham:
I noted that in the WIP part :)

joe-elliott:
> I noted that in the WIP part :)

Good call! I didn't understand what what you meant by "kafka components"

mattdurham:
Ended up making this only apply to live store, the full change introduced a lot of noise. Will followup with that one after this is merged.

modules/livestore/partition_reader.go
```diff
@@ -169,6 +176,12 @@ func collectFetchErrs(fetches kgo.Fetches) (_ error) {
 
 func (r *PartitionReader) consumeFetches(ctx context.Context, fetches kgo.Fetches) *kadm.Offset {
 	// Pass offset and byte information to the live-store
+	partLabel := strconv.Itoa(int(r.partitionID))
```
mattdurham, line 179: I hate the fact we have to do this, but changing it is relatively painful and holding a string for the id in addition also feels bad.

pkg/ingest/metrics.go
```diff
@@ -38,6 +38,125 @@ var (
 	}, []string{labelGroup, labelPartition})
 )
 
+type Metrics struct {
```
joe-elliott, line 41: shouldn't we move all kafka ingesting things to use this and not just livestore?

also not sure live trace/trace bytes and some other metrics are meaningful for all 3 queue consumers.

PR #5656, opened 2025-09-16
wip

carles-grafana:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/registry/counter.go
+++ b/modules/generator/registry/counter.go
@@ -174,3 +174,7 @@ func (c *counter) removeStaleSeries(staleTimeMs int64) {
 		}
 	}
 }
+
+func (c *counter) countTotalSeries() int {
+	return 0
+}
--- a/modules/generator/registry/gauge.go
+++ b/modules/generator/registry/gauge.go
@@ -21,6 +21,8 @@ type gauge struct {
 	// seriesMtx is used to sync modifications to the map, not to the data in series
 	seriesMtx sync.RWMutex
 	series    map[uint64]*gaugeSeries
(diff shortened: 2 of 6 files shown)


modules/generator/registry/counter.go
```diff
@@ -15,8 +15,9 @@ type counter struct {
 	metricName string
 
 	// seriesMtx is used to sync modifications to the map, not to the data in series
-	seriesMtx sync.RWMutex
-	series    map[uint64]*counterSeries
+	seriesMtx      sync.RWMutex
+	series         map[uint64]*counterSeries
+	rejectedSeries map[uint64]int64
```
carles-grafana: Adrian mentioned we can guard this to prevent this from growing infinitely

carles-grafana: Joe mentioned if we can track rejectedSeries globally instead of per metric type. But we still need to identify metric type to know the actual generated series (e.g. histogram depends on bucket number)

carles-grafana: If we add a maximum of rejected series we can estimate the maximum increase in memory per tenant.
For example, it probably doesn't make sense to store more than 500k rejected series as won't be able to support them any way. So this would be 500k * 16 bytes =~ 7.6 MB

It's still worth comparing with hyperloglog though.

mdisibio: We chatted about this, and generators don't OOM in dryrun mode even with 1M+ active series. Because we are using map of series hash, and deduping labels, the memory pressure of the active series is actually not bad. Therefore tracking the rejected series, and their staleness, for real should be minimal overhead.

PR #5658, opened 2025-09-16
[TraceQL] Fix string comparison bug of strings starting with numbers

mdisibio:
**What this PR does**:
Fixes a bug in the traceql engine comparing strings starting with numbers.  `"5" > ""` should be true, but the engine bug was making this false.  Internally it was quoting both hand sides first, so it became ```"`5`" > "``"```, the backtick is before numbers in the ascii table.  The fix is not quote both hand sides, which is both more correct and more performant. 

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@
 * [BUGFIX] Fix invalid YAML output from /status/runtime_config endpoint by adding document separator. [#5146](https://github.com/grafana/tempo/issues/5146)
 * [BUGFIX] Fix race condition between compaction provider and backend-scheduler [#5409](https://github.com/grafana/tempo/pull/5409) (@zalegrala)
 * [BUGFIX] Do not allow very small steps [#5441](https://github.com/grafana/tempo/pull/5441) (@ruslan-mikhailov)
+* [BUGFIX] Fix incorrect TraceQL string comparison of strings starting with numbers [$5658](https://github.com/grafana/tempo/pull/5658) (@mdisibio)
 * [BUGFIX] Fix incorrect results in TraceQL compare() for spans with array attributes [#5519](https://github.com/grafana/tempo/pull/5519) (@ruslan-mikhailov)
 * [BUGFIX] Fix cache collision for incomplete query in SearchTagValuesV2 [#5549](https://github.com/grafana/tempo/pull/5549) (@ruslan-mikhailov)
 * [BUGFIX] Fix for structural operator with empty left-hand spanset [#5578](https://github.com/grafana/tempo/pull/5578) (@ruslan-mikhailov)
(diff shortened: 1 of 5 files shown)


pkg/traceql/ast_execute.go
```diff
@@ -419,34 +419,38 @@ func (o *BinaryOperation) execute(span Span) (Static, error) {
 	}
```
ruslan-mikhailov, line 419: (minor) I see String misusage also here: https://github.com/ruslan-mikhailov/tempo/blob/de50bf1eb8f51a1b90fa6726e317525edfc91ba3/pkg/traceql/ast_validate.go#L215
But it is used only for regexp validation, does not affect behaviour.

pkg/traceql/ast_execute_test.go
```diff
@@ -1766,6 +1766,39 @@ func TestBinaryOperationsWorkAcrossNumberTypes(t *testing.T) {
 	}
 }
 
+func TestBinOp(t *testing.T) {
```
ruslan-mikhailov, line 1769: It would be nice to add a test case to `tempodb/tempodb_search_test.go`, for example, to `searchesThatMatch`

pkg/traceql/ast_execute.go
mdisibio: Good catch, it should validate on the raw value. To do that we would need to cast RHS as a Static and call EncodeToString(false).    Hmm is there ever a case where RHS won't be a fixed string?    Do we support regex against another attribute `.foo ~= .bar` ?

mdisibio: Added the catch/cast of Static and use the unquoted value.  Left fallback behavior to catch anything else.

CHANGELOG.md
```diff
@@ -62,6 +62,7 @@
 * [BUGFIX] Fix invalid YAML output from /status/runtime_config endpoint by adding document separator. [#5146](https://github.com/grafana/tempo/issues/5146)
 * [BUGFIX] Fix race condition between compaction provider and backend-scheduler [#5409](https://github.com/grafana/tempo/pull/5409) (@zalegrala)
 * [BUGFIX] Do not allow very small steps [#5441](https://github.com/grafana/tempo/pull/5441) (@ruslan-mikhailov)
+* [BUGFIX] Fix incorrect TraceQL string comparison of strings starting with numbers [$5658](https://github.com/grafana/tempo/pull/5658) (@mdisibio)
```
ruslan-mikhailov:
```suggestion
* [BUGFIX] Fix incorrect TraceQL string comparison of strings starting with numbers [#5658](https://github.com/grafana/tempo/pull/5658) (@mdisibio)
```

PR #5659, opened 2025-09-16
[Metrics] Fix series handling of conflicting string and integer values

mdisibio:
**What this PR does**:
TraceQL metrics has incorrect results when grouping by an attribute with overlapping string and integer values.  I.e. `rate() by (span.http.status_code)` where the data contains 500 (int) and "500" (string). Because we internally down-cast to strings for inserting into `SeriesSet` maps via `promLabels`.  This fix updates SeriesSet to preserve the types in the map key, and eliminates promLabels entirely (to be sure there are no remaining gaps), and return two distinct time series.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/traceql/engine_metrics.go
```diff
@@ -233,15 +280,22 @@ type TimeSeries struct {
 	Exemplars []Exemplar
 }
 
+type SeriesMapLabel struct {
+	Name  string
+	Value StaticMapKey
+}
+
+type SeriesMapKey [maxGroupBys]SeriesMapLabel
```
mdisibio, line 288: This is the core change.  SeriesSet changed to use a map key that is `comparable` and preserves the static type. This is the same approach used in the `by()` operator itself (GroupingAggregator), but with also the label name.

pkg/tempopb/tempo.proto
```diff
@@ -485,9 +485,8 @@ message TimeSeries {
   // Sorted by time, oldest sample first.
   repeated Sample samples = 2 [(gogoproto.nullable) = false];
 
-  // prom_labels are a flattened string-only version of the typed labels.
-  // They are used internally and may differ from official prometheus conventions.
-  string prom_labels = 3;
+  // prom_labels is obsolete, but left here to ensure field ID isn't reused.
+  // string prom_labels = 3;
```
ruslan-mikhailov: You can use `reserved` for this and above case:

https://protobuf.dev/programming-guides/proto3/#fieldreserved

pkg/traceql/engine_metrics.go
```diff
@@ -212,6 +249,8 @@ func (ls Labels) String() string {
 			} else {
 				promValue = s
 			}
+		case l.Value.Type == TypeInt:
+			promValue = "int(" + l.Value.EncodeToString(false) + ")"
```
ruslan-mikhailov, line 253: Nice idea!

ruslan-mikhailov (approved):

PR #5664, opened 2025-09-18
Add access control interface to query-frontend

ie-pham:
**What this PR does**: Adds a data access controller interface to query-frontend to allow custom permission handling of tracing data.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/frontend.go
+++ b/modules/frontend/frontend.go
@@ -1,6 +1,7 @@
 package frontend
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"net/http"
@@ -62,10 +63,29 @@ type QueryFrontend struct {
 	logger                                                                                     log.Logger
 }
 
+type DataAccessController interface {
+	HandleHTTPSearchReq(r *http.Request) error
+	HandleHTTPTagsReq(r *http.Request) error
+	HandleHTTPTagValuesReq(r *http.Request) error
+	HandleHTTPQueryRangeReq(r *http.Request) error
+	HandleHTTPQueryInstantReq(r *http.Request) error
+
+	HandleGRPCSearchReq(c context.Context, r *tempopb.SearchRequest) error
+	HandleGRPCTagsReq(c context.Context, r *tempopb.SearchTagsRequest) error
+	HandleGRPCTagValuesReq(c context.Context, r *tempopb.SearchTagValuesRequest) error
+	HandleGRPCQueryRangeReq(c context.Context, r *tempopb.QueryRangeRequest) error
+	HandleGRPCQueryInstantReq(c context.Context, r *tempopb.QueryInstantRequest) error
+
+	HandleHTTPUnsupportedEndpoint(r *http.Request) error
+	HandleGRPCUnsupportedEndpoint(c context.Context) error
+
+	TraceByIDResponseRedactor(r *http.Request) (combiner.TraceRedactor, error)
+}
+
 var tracer = otel.Tracer("modules/frontend")
 
 // New returns a new QueryFrontend
-func New(cfg Config, next pipeline.RoundTripper, o overrides.Interface, reader tempodb.Reader, cacheProvider cache.Provider, apiPrefix string, authMiddleware middleware.Interface, logger log.Logger, registerer prometheus.Registerer) (*QueryFrontend, error) {
+func New(cfg Config, next pipeline.RoundTripper, o overrides.Interface, reader tempodb.Reader, cacheProvider cache.Provider, apiPrefix string, authMiddleware middleware.Interface, dataAccessController DataAccessController, logger log.Logger, registerer prometheus.Registerer) (*QueryFrontend, error) {
 	level.Info(logger).Log("msg", "creating middleware in query frontend")
 
 	if cfg.TraceByID.QueryShards < minQueryShards || cfg.TraceByID.QueryShards > maxQueryShards {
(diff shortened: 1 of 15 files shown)


modules/frontend/tag_handlers.go
```diff
@@ -221,6 +257,13 @@ func newTagsHTTPHandler(cfg Config, next pipeline.AsyncRoundTripper[combiner.Pip
 			return errResp, nil
 		}
 
+		if dataAccessController != nil {
+			if err := dataAccessController.HandleHTTPUnsupportedEndpoint(req); err != nil {
+				level.Error(logger).Log("msg", "SearchTags http: access control handling failed", "err", err)
+				return httpInvalidRequest(err), nil
```
stoewer: When `HandleHTTPUnsupportedEndpoint()` is called and it returns a `errUnsupportedEndpoint` this can be interpreted as accessing the endpoint is forbidden (status code 403), but `httpInvalidRequest()` maps the error to 400. Not 100% sure about this though and I guess this can be handled in a later PR.

stoewer (approved):
LGTM

modules/frontend/frontend.go
```diff
@@ -62,10 +63,29 @@ type QueryFrontend struct {
...
+	HandleGRPCSearchReq(c context.Context, r *tempopb.SearchRequest) error
+	HandleGRPCTagsReq(c context.Context, r *tempopb.SearchTagsRequest) error
+	HandleGRPCTagValuesReq(c context.Context, r *tempopb.SearchTagValuesRequest) error
+	HandleGRPCQueryRangeReq(c context.Context, r *tempopb.QueryRangeRequest) error
+	HandleGRPCQueryInstantReq(c context.Context, r *tempopb.QueryInstantRequest) error
+
+	HandleHTTPUnsupportedEndpoint(r *http.Request) error
+	HandleGRPCUnsupportedEndpoint(c context.Context) error
+
+	TraceByIDResponseRedactor(r *http.Request) (combiner.TraceRedactor, error)
```
mdisibio: Might fit the pattern more if we rename this to `HandleHTTPTraceByIDReq(r *http.Request) (combiner.TraceRedator, error)`?  Otherwise it's not clear why that one is different (it isn't really).

modules/frontend/metrics_query_handler.go
```diff
@@ -90,15 +97,18 @@ func newMetricsQueryInstantHTTPHandler(cfg Config, next pipeline.AsyncRoundTripp
...
 		// Parse request
 		i, err := api.ParseQueryInstantRequest(req)
 		if err != nil {
 			level.Error(logger).Log("msg", "query instant: parse search request failed", "err", err)
-			return &http.Response{
-				StatusCode: http.StatusBadRequest,
-				Status:     http.StatusText(http.StatusBadRequest),
-				Body:       io.NopCloser(strings.NewReader(err.Error())),
-			}, nil
+			return httpInvalidRequest(err), nil
```
mdisibio, line 111: Agree with Adrian's question about the mapping of error to http response.  Could there be other errors where we should return 500 internal server error?  gRPC should be ok, we can embed the needed response within the error like `grpcErr = status.Error(codes.InvalidArgument, <message>)`   It's ok to proceed and address this up later if we aren't sure.

modules/frontend/tag_handlers.go
```diff
@@ -137,10 +153,21 @@ func newTagsV2StreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTripper[c
...
-func newTagValuesStreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTripper[combiner.PipelineResponse], apiPrefix string, o overrides.Interface, logger log.Logger) streamingTagValuesHandler {
+func newTagValuesStreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTripper[combiner.PipelineResponse], apiPrefix string, o overrides.Interface, logger log.Logger, dataAccessController DataAccessController) streamingTagValuesHandler {
 	postSLOHook := metadataSLOPostHook(cfg.Search.MetadataSLO)
 
 	return func(req *tempopb.SearchTagValuesRequest, srv tempopb.StreamingQuerier_SearchTagValuesServer) error {
+		ctx := srv.Context()
+		var err error
+
+		if dataAccessController != nil {
+			err = dataAccessController.HandleGRPCUnsupportedEndpoint(ctx)
```
mdisibio: Should we make this a real named function and let the data access controller decide that it's unsupported and when to error?   It looks like maybe it should always be failing because it's known to be unsupported, but that isn't the case. It's only if policies are present, which is an internal concern in the data access controller.

modules/frontend/frontend.go
ie-pham: This is different because it redacts information after the roundtrip whereas the other one changes the req before the roundtrip takes place

modules/frontend/tag_handlers.go
ie-pham: Do you mean like keep this check inside HandleGRPCSearchTagsReq? 

We wanted to make it abundantly clear that the endpoints where the unsportedendpoints handler functions are called that those endpoints are not supported for our access control implementation. [Here](https://github.com/grafana/cloud-traces/pull/84#discussion_r2350924023) is the discussion that led to the creation of those handlers.

mdisibio: Right, it is still kind of an implementation detail, for the current goals.  Tempo OSS doesn't need to assume that this endpoint will never work for anything.

modules/frontend/frontend.go
mdisibio: Sure, I am thinking the pattern `Handle<Protocol><API>` is good, I like it. It would be nice to make this one match.  It's ok if this returns `(thing_you_must_use, error)` instead of just `(error)` like the others.  It's an implementation detail of the current goals and not necessarily always true.  Wouldn't be surprised if others end up returning ones too.

modules/frontend/traceid_handlers.go
```diff
@@ -60,7 +49,16 @@ func newTraceIDHandler(cfg Config, next pipeline.AsyncRoundTripper[combiner.Pipe
 			"tenant", tenant,
 			"path", req.URL.Path)
 
-		comb := combinerFn(o.MaxBytesPerTrace(tenant), marshallingFormat)
+		var traceRedactor combiner.TraceRedactor
+		if dataAccessController != nil {
+			traceRedactor, err = dataAccessController.TraceByIDResponseRedactor(req)
+			if err != nil {
+				level.Error(logger).Log("msg", "trace id: failed to get trace redactor", "err", err)
+				return httpInvalidRequest(err), nil
```
joe-elliott, line 57: wouldn't this be a 500 instead of a 400?

modules/frontend/traceid_handlers.go
```diff
@@ -130,7 +120,16 @@ func newTraceIDV2Handler(cfg Config, next pipeline.AsyncRoundTripper[combiner.Pi
 			"tenant", tenant,
 			"path", req.URL.Path)
 
-		comb := combinerFn(o.MaxBytesPerTrace(tenant), marshallingFormat)
+		var traceRedactor combiner.TraceRedactor
+		if dataAccessController != nil {
+			traceRedactor, err = dataAccessController.TraceByIDResponseRedactor(req)
+			if err != nil {
+				level.Error(logger).Log("msg", "trace id v2: failed to get trace redactor", "err", err)
+				return httpInvalidRequest(err), nil
```
joe-elliott, line 128: same Q here

modules/frontend/traceid_handlers.go
ie-pham: this is tricky because for our implementation, the only errors that get tripped are due to bad request- but if we were to make it more generic a 500 is more fitting. what do you think?

PR #5677, opened 2025-09-24
Add kafka metrics

mattdurham:
This sets up the alerts, the dashboards I will likely copy block builder but want to play with it some first. During testing the context.Background tended to hang if something bad happened so switched out most of the problematic ones to t.Context.

### DIFF
--- a/modules/livestore/partition_reader_test.go
+++ b/modules/livestore/partition_reader_test.go
@@ -107,7 +107,7 @@ func defaultPartitionReaderWithCommitInterval(t *testing.T, address string, comm
 	)
 	require.NoError(t, err)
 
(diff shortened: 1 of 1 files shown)


modules/livestore/live_store.go
```diff
@@ -123,6 +123,7 @@ type LiveStore struct {
 	wg              sync.WaitGroup
 	completeQueues  *flushqueues.ExclusiveQueues
 	startupComplete chan struct{} // channel to signal that the starting function has finished. allows background processes to block until the service is fully started
+	lagCancel       context.CancelFunc
```
mattdurham, line 126: The partition lag cant use the normal context and needs to be explicitly closed first before stopping else logs leak into tests.

mapno (approved):

PR #5680, opened 2025-09-25
Tempo: Assert limits in livestore

joe-elliott:
**What this PR does**:
Correctly support req.AllowPartialTrace and tenant limits in the live store when requesting trace by id.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/livestore/instance_search.go
+++ b/modules/livestore/instance_search.go
@@ -580,14 +580,13 @@ func (i *instance) FindByTraceID(ctx context.Context, traceID []byte, allowParti
 	var (
 		metricsMtx sync.Mutex
 		metrics    = tempopb.TraceByIDMetrics{}
+		maxBytes   = i.overrides.MaxBytesPerTrace(i.tenantID)
+		searchOpts = common.DefaultSearchOptions()
+		combiner   = trace.NewCombiner(maxBytes, allowPartialTrace)
 	)
 
-	maxBytes := i.overrides.MaxBytesPerTrace(i.tenantID)
-	searchOpts := common.DefaultSearchOptions()
-	combiner := trace.NewCombiner(maxBytes, allowPartialTrace)
-
 	if !allowPartialTrace {
-		searchOpts.MaxBytes = maxBytes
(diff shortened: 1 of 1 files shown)


modules/livestore/instance_search.go
```diff
@@ -577,13 +576,20 @@ func (i *instance) SearchTagValuesV2(ctx context.Context, req *tempopb.SearchTag
...
 		metricsMtx sync.Mutex
 		metrics    = tempopb.TraceByIDMetrics{}
-		combiner   = trace.NewCombiner(math.MaxInt64, true)
 	)
 
+	maxBytes := i.overrides.MaxBytesPerTrace(i.tenantID)
+	searchOpts := common.DefaultSearchOptions()
+	combiner := trace.NewCombiner(maxBytes, allowPartialTrace)
+
+	if !allowPartialTrace {
```
mattdurham, line 588: Can this be simplified to DefaultSearchOptionsWithMaxBytes?

joe-elliott: yup, good call. pushed the change along with some other declaration cleanup

mattdurham: Sorry I meant can the searchOpts instantiaton be changed to common.DefaultSearchOptionsWithMaxBytes(maxBytes) and then remove this if check. That looks more closely mirror what the ingester/instance.go does in FindTraceByID

joe-elliott: oh, there's a small and mostly innocuous bug in ingesters that this change fixes. if req.AllowPartialTraces = true and a single block exceeds the max trace size when calling .FindTraceByID() then the function call will return an error instead of the partial trace.

by only setting the max trace size on search opts if allowPartialTrace is false we fix this issue.

now, it's extremely difficult to trigger this issue b/c the ingester will refuse to consume a trace larger than the configured max trace size. to trip it you have to consume a large trace, the max trace size on the tenant has to be decreased and then you request the trace.

mattdurham: Gotcha, that makes sense!

PR #5682, opened 2025-09-26
feat: limit recent queries

javiermolinar:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


modules/frontend/frontend.go
```diff
@@ -172,6 +178,7 @@ func New(cfg Config, next pipeline.RoundTripper, o overrides.Interface, reader t
 	metricsPipeline := pipeline.Build(
 		[]pipeline.AsyncMiddleware[combiner.PipelineResponse]{
 			urlDenyListWare,
+			adjustEndWareSeconds,
```
electron0zero: should this be `adjustEndWareNanos` ?

javiermolinar: yes, good catch

mattdurham: I swear we should alias our time types

modules/frontend/config.go
```diff
@@ -88,8 +88,10 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
 	cfg.ResponseConsumers = 10
 	cfg.Search = SearchConfig{
 		Sharder: SearchSharderConfig{
-			QueryBackendAfter:     15 * time.Minute,
-			QueryIngestersUntil:   30 * time.Minute,
+			QueryBackendAfter:   15 * time.Minute,
+			QueryIngestersUntil: 30 * time.Minute,
+			// DefaultQueryStart:  0,  // Leave as 0 to indicate "not set"
```
joe-elliott: since these values are based on flush times in the live store should we just have one pair of settings for all endpoints? instead of separate ones for serach/metrics?

modules/frontend/frontend.go
```diff
@@ -121,10 +121,26 @@ func New(cfg Config, next pipeline.RoundTripper, o overrides.Interface, reader t
...
+			return nil, fmt.Errorf("metrics default query start (%v) must be greater than default query end buffer (%v)",
+				cfg.Metrics.Sharder.DefaultQueryStart, cfg.Metrics.Sharder.DefaultQueryEndBuffer)
+		}
+	}
+
 	// Propagate RF1After to search and traceByID sharders
 	cfg.Search.Sharder.RF1After = cfg.RF1After
 	cfg.TraceByID.RF1After = cfg.RF1After
 
+	adjustEndWareSeconds := pipeline.NewAdjustStartEndWare(cfg.Search.Sharder.DefaultQueryStart, cfg.Search.Sharder.DefaultQueryEndBuffer, false) // jpe - these should not be hardcoded but based on config: max live traces + flush cycle for the endbuffer and QueryBackendAfter for the defStart?
```
joe-elliott:
```suggestion
	adjustEndWareSeconds := pipeline.NewAdjustStartEndWare(cfg.Search.Sharder.DefaultQueryStart, cfg.Search.Sharder.DefaultQueryEndBuffer, false) 
```

modules/frontend/search_sharder.go
```diff
@@ -33,8 +33,16 @@ type SearchSharderConfig struct {
 	DefaultLimit          uint32        `yaml:"default_result_limit"`
 	MaxLimit              uint32        `yaml:"max_result_limit"`
 	MaxDuration           time.Duration `yaml:"max_duration"`
-	QueryBackendAfter     time.Duration `yaml:"query_backend_after,omitempty"`
-	QueryIngestersUntil   time.Duration `yaml:"query_ingesters_until,omitempty"`
+	// QueryBackendAfter determines when to query backend storage vs ingesters only.
```
joe-elliott, line 36: remove? seems unused?

modules/livestore/config.go
```diff
@@ -73,7 +73,7 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
 	// Set defaults for timing configuration (based on ingester defaults)
 	cfg.InstanceFlushPeriod = 10 * time.Second
 	cfg.InstanceCleanupPeriod = 5 * time.Minute
-	cfg.MaxTraceLive = 30 * time.Second
+	cfg.MaxTraceLive = 30 * time.Second // jpe - add docs that this limits querying
```
joe-elliott:
```suggestion
	cfg.MaxTraceLive = 30 * time.Second
```

pkg/api/http.go
```diff
@@ -545,52 +566,139 @@ func determineBounds(now time.Time, startString, endString, sinceString string)
...
+// Parameter precedence and behavior:
+//  1. If 'since' is provided: end = now - endBuffer, start = end - since
+//  2. If neither start nor end provided: end = now - endBuffer, start = now - defStart
+//  3. If both start and end provided: parse both and clamp end to not exceed (now - endBuffer)
+//  4. If only one of start/end provided: returns validation error
+//
+// Validation rules when both start and end are provided:
+//   - Both must have the same string length (seconds vs nanoseconds format)
+//   - Both must be either fractional or integer (consistent decimal point usage)
+func ClampDateRangeReq(req *http.Request, defStart, endBuffer time.Duration) (start, end time.Time, p TimestampPrecision, err error) {
```
joe-elliott: timestamp precision seems ignored by the caller of this func? i can't find anywhere its used. remove all TimestampPrecision code?

joe-elliott (commented):
a few comments. also i really don't like the default_query_start config option since you're correct that we should just use query backend after.

i understand we did this for some tests? is there any other way around this? maybe we can meet sync and look closely at the problem?

pkg/api/http.go
javiermolinar: Yeah, I used it in the tests, but can remove it

modules/frontend/config.go
javiermolinar: That's a very good point. I'll try to simplify the config

javiermolinar: What do you think if we let that to a follow-up pr?

javiermolinar:
@joe-elliott I was able to make it work without the extra config `DefaultQueryStart` in fact I've simplified most of the tests configs. Could you take it another look?

modules/frontend/config.go
joe-elliott: PR overall looks good, but we should move this config now before it gets merged. once you start using config it can be more difficult to move it. This is definitely the time to make sure it is what we want it to be.

Also, i'm going to nit the name. Can we make it:
```
query_frontend:
  query_end_cutoff: <duration>
```

joe-elliott (commented):
PR looking good close to merging. Can we go ahead and make the config change now? Also needs a changelog.

modules/frontend/frontend.go
```diff
@@ -121,10 +121,19 @@ func New(cfg Config, next pipeline.RoundTripper, o overrides.Interface, reader t
 		return nil, fmt.Errorf("frontend metrics interval should be greater than 0")
 	}
 
+	if cfg.QueryEndCutoff > cfg.Search.Sharder.QueryBackendAfter {
+		return nil, fmt.Errorf("QueryBackendAfter (%v) must be greater than query end cutoff (%v)", cfg.Search.Sharder.QueryBackendAfter, cfg.QueryEndCutoff)
+	}
+	if cfg.QueryEndCutoff > cfg.Metrics.Sharder.QueryBackendAfter {
+		return nil, fmt.Errorf("QueryBackendAfter (%v) must be greater than query end cutoff (%v)", cfg.Metrics.Sharder.QueryBackendAfter, cfg.QueryEndCutoff)
```
joe-elliott: same check 2x? remove the second one?

javiermolinar: It´s not really duplicated since the QueryBackendAfter can be different for search and metrics. Nevertheless we already discussed moving this property up. I will remove the second one

joe-elliott: oh, good call. i didn't notice. i'm good with the double check

PR #5686, opened 2025-09-26
Include new metric for generator ring size at the distributor

zalegrala:
**What this PR does**:

Here we include a new metric `tempo_distributor_metrics_generator_tenant_ring_size` to increase observability in our metrics-generator pipeline.  When tenant metrics are being discarded, we often want to know more details about the tenant, and this looks like a gap in our metrics.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/distributor.go
+++ b/modules/distributor/distributor.go
@@ -72,6 +72,11 @@ var (
 		Name:      "distributor_metrics_generator_pushes_failures_total",
 		Help:      "The total number of failed span pushes sent to metrics-generators.",
 	}, []string{"metrics_generator"})
+	metricGeneratorTenantRingSize = promauto.NewGaugeVec(prometheus.GaugeOpts{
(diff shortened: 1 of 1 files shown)


modules/distributor/distributor.go
```diff
@@ -72,6 +72,11 @@ var (
 		Name:      "distributor_metrics_generator_pushes_failures_total",
 		Help:      "The total number of failed span pushes sent to metrics-generators.",
 	}, []string{"metrics_generator"})
+	metricGeneratorTenantRingSize = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
electron0zero, line 75: missing changelog.md entry

zalegrala: I've pushed a commit for the changelog.

PR #5693, opened 2025-09-29
Live-store: jump offset to skip old records

ruslan-mikhailov:
**What this PR does**: if no commits found, do not start from beginning and skip up no now-lookbackPeriod. If commit is found, compare offsets between the commit and `now-lookbackPeriod` offset, and pick the biggest.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/sources/tempo/configuration/manifest.md
+++ b/docs/sources/tempo/configuration/manifest.md
@@ -1242,7 +1242,6 @@ live_store:
     max_live_traces_bytes: 250000000
     max_block_duration: 30m0s
     max_block_bytes: 524288000
-    max_lookback: 1h0m0s
     block_config:
         bloom_filter_false_positive: 0.01
         bloom_filter_shard_size_bytes: 102400


modules/livestore/partition_reader.go
```diff
@@ -228,9 +230,11 @@ func (r *PartitionReader) fetchLastCommittedOffset(ctx context.Context) (kgo.Off
...
-	if !found {
-		// No committed offset found for this partition, start from the end
-		return kgo.NewOffset().AtStart(), nil
+	if !found { // No committed offset found for this partition
+		if r.lookbackPeriod == 0 {
+			return kgo.NewOffset().AtEnd(), nil
+		}
+		return kgo.NewOffset().AfterMilli(time.Now().Add(-r.lookbackPeriod).UnixMilli()), nil
 	}
 	return kgo.NewOffset().At(offset.At), nil
```
ruslan-mikhailov: Also drafted the optimisation for case when it has offsets: https://github.com/ruslan-mikhailov/tempo/commit/19d52e13b183a42cf405dd07a641a33083a6e31a

The problem is that it requires additional listOffset operation, so I dropped it

mapno: This operation is done once during startup. I think it's a nice optimisation.

ruslan-mikhailov: \+

Included the commit, updated the test.

Issue #5665, opened 2025-09-19
vParquet5: replace well-known attributes with default dedicated columns


stoewer:
Remove the [well-known attributes](https://grafana.com/docs/tempo/latest/operations/schema/#well-known-attributes) such as `http.status_code` and replace them with defaults for dedicated columns.

The well-known attribute columns and dedicated attribute columns have very similar performance characteristics. The main difference is that well-known attributes are predefined and align with the OTEL semantic conventions. However, semantic conventions have changed over time, this lead to a growing mismatch (e.g. #5172).

PR #5696, opened 2025-09-30
vParquet5: remove well-known attribute columns

stoewer:
**What this PR does**:

This PR consists of the following changes:
* Remove the existing columns for well-known attribute columns
* Apply a default dedicated attribute column config when no dedicated columns are present
* Encodings vp4 and vp3 ignore dedicated column configs that match existing well-known columns. This helps to avoid conflicts with the new default dedicated columns in older encodings
* Use LRU cache for unmarshaled dedicated columns and avoid allocations for cache lookup: This is a small optimization to compensate for the larger and more frequently used dedicated column configs. I plan to implement further optimizations in a separate PR. 

Contributes to #5665

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/encoding/vparquet3/dedicated_columns.go
```diff
@@ -188,3 +212,30 @@ func dedicatedColumnsToColumnMapping(dedicatedColumns backend.DedicatedColumns,
 
 	return mapping
 }
+
+func filterDedicatedColumns(columns backend.DedicatedColumns) backend.DedicatedColumns {
+	return slices.Collect(func(yield func(c backend.DedicatedColumn) bool) {
```
mdisibio: Is there a benefit to using the iter.Seq approach?  It looks like it doesn't actually yield and can just build a new slice like usual.  Question applies to all vp3/4/5 instances of this method.

tempodb/encoding/vparquet3/dedicated_columns.go
```diff
@@ -188,3 +212,30 @@ func dedicatedColumnsToColumnMapping(dedicatedColumns backend.DedicatedColumns,
...
+			if isIgnoredDedicatedColumn(c.Scope, c.Type, c.Name) {
+				continue
+			}
+			yield(c)
+		}
+	})
+}
+
+func isIgnoredDedicatedColumn(scope backend.DedicatedColumnScope, typ backend.DedicatedColumnType, attr string) bool {
+	if _, found := DedicatedResourceColumnPaths[scope]; !found {
```
mdisibio: The double-lookup here I think shows a small gap.  The type information is included in DedicatedResourceColumnPaths, but the attribute name is in ignoredAttributes, and they aren't directly connected.  Maybe we need to include both type and name in ignoredAttributes? So that we ignore `int(http.status_code)` but not `string(http.status_code)`. Question applies to both vp3 and vp4.

tempodb/backend/dedicated_columns_cache.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func putDedicatedColumnsToCache(marshalled []byte, cols DedicatedColumns) {
+	if len(marshalled) == 0 {
+		return
+	}
+	_ = dedicatedColumnsCache.Add(string(marshalled), cols)
+}
+
+func clearDedicatedColumnsCache() {
```
mdisibio: Just checking, the only reference is for testing.  Since it's LRU it's not require to purge it like the old cache, and the poller doesn't call it anymore.

tempodb/backend/dedicated_columns_cache.go
```diff
@@ -0,0 +1,48 @@
...
+	"unsafe"
+
+	lru "github.com/hashicorp/golang-lru/v2"
+)
+
+var (
+	dedicatedColumnsCacheNull  = []byte("null")
+	dedicatedColumnsCacheEmpty = []byte("[]")
+	dedicatedColumnsCacheSize  = 1024
+	dedicatedColumnsCache      *lru.Cache[string, DedicatedColumns]
```
mdisibio, line 14: We also have:  https://github.com/grafana/tempo/blob/main/pkg/cache/reclaimable/reclaimable.go

Probably negligible different trade offs and not critical.  But differences that stick out to me is it would allow scaling to more than 1024 configs, and also can be GCed.

tempodb/encoding/vparquet3/dedicated_columns.go
stoewer: It's just a matter of code style preference, I can change it

tempodb/backend/dedicated_columns_cache.go
stoewer: Yes, was only used in tests. I removed it

tempodb/backend/dedicated_columns_cache.go
stoewer: The reclaimable could also be an option. We currently see way less than 1000 distinct dedicated column configurations at the same time. This means the cache only fills over time, which makes LRU a good choice in my opinion.

tempodb/encoding/vparquet3/dedicated_columns.go
```diff
@@ -188,3 +216,23 @@ func dedicatedColumnsToColumnMapping(dedicatedColumns backend.DedicatedColumns,
...
+			continue
+		}
+		filtered = append(filtered, c)
+	}
+	return filtered
+}
+
+func isIgnoredDedicatedColumn(scope backend.DedicatedColumnScope, typ backend.DedicatedColumnType, attr string) bool {
+	if _, found := DedicatedResourceColumnPaths[scope][typ]; !found {
+		return true
```
mdisibio, line 233: Should this return `false`?  I think the only remaining case is resource/int.  It will be `!found`, and returning true here means it will get ignored.  We can probably remove this check too since now ignoredAttributes contains all of the information.

stoewer: The purpose is to remove dedicated columns with unsupported type or scope as well (e.g. event or bool). If there is no column path, the condition is `!found` and returns `true` (i.e. ignore the column)

I'm aware that this now makes the `http.status_code` entry in `ignoredAttributes` redundant, as all int columns are already filtered by the first check. The reasons why I still kept `http.status_code` in `ignoredAttributes` was:
* Otherwise it looks like a bug if `ignoredAttributes` contain all well-known columns except http.status_code
* It can easily lead to bugs if someone changes the first condition without paying attention to the second one

mdisibio (approved):

PR #5704, opened 2025-10-03
Add rpm deb

mattdurham:
Created a draft release v0.0.999-rc-testdeb2 (which I will delete after merge). I was able to follow our internal pipeline to the point where it would publish the package. Manually running some of the commands with include-drafts-enabled it was able to find the packages though I could not test beyond that without pushing an actual package :)

### DIFF
--- a/tools/packaging/verify-deb-install.sh
+++ b/tools/packaging/verify-deb-install.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+# Install tempo and check it's running
+dpkg -i ./tempo_*_linux_amd64.deb
+[ "$(systemctl is-active tempo)" = "active" ] || (echo "tempo is inactive" && exit 1)
+
+# Wait for tempo to be ready.
+apt update && apt install -y curl
+./wait-for-ready.sh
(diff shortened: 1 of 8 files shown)


tools/packaging/verify-deb-install.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+# Install tempo and check it's running
+dpkg -i ./tempo_*_linux_amd64.deb
```
joe-elliott: these 2 verify scripts are no longer used. they used to be referenced by release.yml as a testing step. do we want to restore the verify? they may require modifications to accept an unsigned package.

i'm good on just removing if we have high confidence in the packages.

tools/packaging/wait-for-ready.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env sh
+
+status="$(curl -s http://localhost:3200/ready)"
```
joe-elliott: similar comment on this script since it's used by the above.

joe-elliott (approved):
lgtm

PR #5706, opened 2025-10-03
Add ability to drop instance label from span metrics

### LINKED ISSUE #4873: High Cardinality Issue with service.instance.id in Span Metrics Processor
**Describe the bug**

The span metrics processor currently adds `service.instance.id` resource attribute as `instance` label if `EnableTargetInfo` is enabled. This leads to high cardinality in generated metrics even when `service.instance.id` is not part of the configured dimensions. There should be special handling to only include `service.instance.id` as `instance` when it is explicitly configured as a dimension.

**To Reproduce**
* configure span metrics processor with EnableTargetInfo: true.
* generate traces that include `service.instance.id` but do not configure it as a dimension in the processor.
* generated metrics will include an `instance` label derived from `service.instance.id`, even when it is not part of the configured dimensions.

**Expected behavior**

The instance label should only be added if `service.instance.id` is explicitly configured as part of the dimensions.

**Actual behavior**
`instance` is always added when `EnableTargetInfo` is enabled, even if `service.instance.id` is not in the configured dimensions, causing high metric cardinality.

ie-pham:
**What this PR does**: Add a config to allow dropping of the `instance` label from span_metrics timeseries since it tends to be very high in cardinality

Fixes #4873 

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/processor/spanmetrics/spanmetrics.go
+++ b/modules/generator/processor/spanmetrics/spanmetrics.go
@@ -206,7 +206,7 @@ func (p *Processor) aggregateMetricsForSpan(svcName string, jobName string, inst
 		labelValues = append(labelValues, jobName)
 	}
 	//  add instance label only if job is not blank
-	if instanceID != "" && p.Cfg.EnableTargetInfo {
+	if instanceID != "" && p.Cfg.EnableTargetInfo && !p.Cfg.DropInstanceLabel {
 		labels = append(labels, dimInstance)
 		labelValues = append(labelValues, instanceID)
 	}
@@ -246,7 +246,7 @@ func (p *Processor) aggregateMetricsForSpan(svcName string, jobName string, inst
 			targetInfoLabelValues = append(targetInfoLabelValues, jobName)
 		}
 		//  add instance label to target info only if job is not blank
-		if instanceID != "" {
+		if instanceID != "" && !p.Cfg.DropInstanceLabel {
 			targetInfoLabels = append(targetInfoLabels, dimInstance)
 			targetInfoLabelValues = append(targetInfoLabelValues, instanceID)
 		}
--- a/integration/e2e/overrides_test.go
+++ b/integration/e2e/overrides_test.go
@@ -141,6 +141,11 @@ func TestOverrides(t *testing.T) {
 			patch := &client.Limits{
 				MetricsGenerator: client.LimitsMetricsGenerator{
 					DisableCollection: boolPtr(true),
+					Processor: client.LimitsMetricsGeneratorProcessor{
+						SpanMetrics: client.LimitsMetricsGeneratorProcessorSpanMetrics{
+							DropInstanceLabel: boolPtr(true),
+						},
+					},
 				},
 			}
 
@@ -154,6 +159,9 @@ func TestOverrides(t *testing.T) {
 			processors, ok = limits.GetMetricsGenerator().GetProcessors()
 			assert.True(t, ok)
 			assert.ElementsMatch(t, keys(processors.GetMap()), []string{"span-metrics"})
+			dropInstanceLabel, dropInstanceLabelIsSet := limits.GetMetricsGenerator().GetProcessor().GetSpanMetrics().GetDropInstanceLabel()
+			assert.True(t, dropInstanceLabel)
+			assert.True(t, dropInstanceLabelIsSet)
 
 			// Delete overrides
 			if !tc.skipVersioning && tc.name != "gcs" {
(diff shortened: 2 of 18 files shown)


docs/sources/tempo/metrics-from-traces/span-metrics/span-metrics-metrics-generator.md
```diff
@@ -83,7 +83,7 @@ By default, the metrics processor adds the following labels to each metric: `ser
   - `STATUS_CODE_ERROR` - The span operation completed with an error
 - `status_message` (optionally enabled) - The message that details the reason for the `status_code` label
 - `job` - The name of the job, a combination of namespace and service; only added if `metrics_generator.processor.span_metrics.enable_target_info: true`
-- `instance` - The instance ID; only added if `metrics_generator.processor.span_metrics.enable_target_info: true`
+- `instance` - The instance ID; only added if `metrics_generator.processor.span_metrics.enable_target_info: true` and can be dropped by setting `metrics_generator.processor.span_metrics.drop_instance_label: true
```
mattdurham: @rlankfo isn't instance semi required for `target_info` to work correctly?

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -206,7 +206,7 @@ func (p *Processor) aggregateMetricsForSpan(svcName string, jobName string, inst
 		labelValues = append(labelValues, jobName)
 	}
 	//  add instance label only if job is not blank
```
mattdurham: Not related to your code but this comment looks like its wrong.

modules/generator/processor/spanmetrics/spanmetrics.go
```diff
@@ -246,7 +246,7 @@ func (p *Processor) aggregateMetricsForSpan(svcName string, jobName string, inst
 			targetInfoLabelValues = append(targetInfoLabelValues, jobName)
 		}
 		//  add instance label to target info only if job is not blank
-		if instanceID != "" {
+		if instanceID != "" && !p.Cfg.DropInstanceLabel {
```
mattdurham: same comment issue here.

docs/sources/tempo/metrics-from-traces/span-metrics/span-metrics-metrics-generator.md
mattdurham: IE when we go to set the value on the Gauge your likely to override/not override appropriately in SetForTargetInfo

modules/generator/processor/spanmetrics/spanmetrics.go
ie-pham: Yeah this is a chance to the original requirement. I will change the comments.

docs/sources/tempo/metrics-from-traces/span-metrics/span-metrics-metrics-generator.md
ie-pham: can you explain why it would not work appropriately?

ie-pham: the original requirement also states that if instance is empty to skip it entirely, which sounds to me like it's not a hard requirement.

mattdurham: If there is not a job or instance then the metric will never get recorded, which feels like different behavior than reducing cardinality. The change in behavior would be if there is no job label and drop instance is true then the metric will not be recorded, whereas previously it would be. Though granted that should be pretty rare that a `service.name` (job) is not declared.

docs/sources/tempo/metrics-from-traces/span-metrics/span-metrics-metrics-generator.md
```diff
@@ -83,7 +83,7 @@ By default, the metrics processor adds the following labels to each metric: `ser
   - `STATUS_CODE_ERROR` - The span operation completed with an error
 - `status_message` (optionally enabled) - The message that details the reason for the `status_code` label
 - `job` - The name of the job, a combination of namespace and service; only added if `metrics_generator.processor.span_metrics.enable_target_info: true`
-- `instance` - The instance ID; only added if `metrics_generator.processor.span_metrics.enable_target_info: true`
+- `instance` - The instance ID; only added if `metrics_generator.processor.span_metrics.enable_target_info: true` and can be dropped by setting `metrics_generator.processor.span_metrics.enable_instance_label: false
```
mattdurham: Thoughts on changing this to say 
```
The instance ID; only added if `metrics_generator.processor.span_metrics.enable_target_info: true` 
and `metrics_generator.processor.span_metrics.enable_instance_label: true` 
```

PR #5710, opened 2025-10-06
metrics-generator: estimate active series demand

carles-grafana:
**What this PR does**:

Add `tempo_metrics_generator_registry_active_series_demand_estimate` that estimates the active series demand.

This is especially useful when the active series reaches the max limit.

To calculate the actual cardinality, we would need 16 bytes for each series.
- **For 500k series, that would be 8MB.**

**With HyperLogLog, we only need 5KB (99.9% less)** per tenant, without having to store rejected series in memory. 
- with only a 3.25% standard error vs the actual

In practice we have seen that the deviation against the actual cardinality stays below 3% most of the time, with some very short spikes due to the Prometheus scrape interval that converge quickly.

Deviation percentage of the estimate vs actual cardinality in development environment:

unknown

Absolute estimate vs actual cardinality of 1 tenant:

unknown

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/generator/registry/counter.go
```diff
@@ -78,26 +80,25 @@ func (c *counter) Inc(labelValueCombo *LabelValueCombo, value float64) {
 	s, ok := c.series[hash]
 	c.seriesMtx.RUnlock()
 
+	c.seriesDemand.Insert(hash)
 	if ok {
 		c.updateSeries(s, value)
 		return
 	}
 
-	if !c.onAddSeries(1) {
```
carles-grafana, line 86: this reordering is to prevent race condition if the same series is added concurrently between locks

modules/generator/registry/counter.go
```diff
@@ -131,12 +132,10 @@ func (c *counter) name() string {
 	return c.metricName
 }
 
-func (c *counter) collectMetrics(appender storage.Appender, timeMs int64) (activeSeries int, err error) {
+func (c *counter) collectMetrics(appender storage.Appender, timeMs int64) error {
 	c.seriesMtx.RLock()
 	defer c.seriesMtx.RUnlock()
 
-	activeSeries = len(c.series)
```
carles-grafana, line 132: I've removed active series counting from these functions because it's redundant. We already hold the series in the map, so at any time we know the exact number.

joe-elliott:
very cool. i will try to find time to review, but in the meantime can you share how close the demand estimate is to the true number of series?

modules/generator/registry/cardinality.go
```diff
@@ -0,0 +1,137 @@
...
+	window := c.cachedMerge.Clone()
+	current := c.sketches[c.current].Clone()
+	c.mu.RUnlock()
+	_ = window.Merge(current)
+	return window.Estimate()
+}
+
+// Advance should be called by an external ticker every sketchDuration to advance the ring.
+// It advances at least 1 step per call.
+func (c *Cardinality) Advance() {
```
mdisibio, line 107: Is there a case where Advance isn't called in regular intervals?  I would be ok with always advancing exactly one step in this function, even if it's called 5-30s too late.  Because it would be much simpler, and this is already an estimate with some error anyway.

modules/generator/registry/registry.go
```diff
@@ -83,12 +90,16 @@ type ManagedRegistry struct {
...
-	collectMetrics(appender storage.Appender, timeMs int64) (activeSeries int, err error)
+	collectMetrics(appender storage.Appender, timeMs int64) error
+	countActiveSeries() int
+	countSeriesDemand() int
 	removeStaleSeries(staleTimeMs int64)
 }
 
 const highestAggregationInterval = 1 * time.Minute
 
+const removeStaleSeriesInterval = 5 * time.Minute
```
mdisibio, line 102: Should we make this configurable, or maybe based on `stale_duration` ?   I.e. if stale duration is 15m, then stale_duration/3 or any constant you think is best.  My understanding is this provides better accuracy, there isn't a constraint that requires 5 minutes.

modules/generator/registry/gauge.go
```diff
@@ -57,6 +58,7 @@ func newGauge(name string, onAddSeries func(uint32) bool, onRemoveSeries func(co
 	return &gauge{
 		metricName:     name,
 		series:         make(map[uint64]*gaugeSeries),
+		seriesDemand:   NewCardinality(10, staleDuration, removeStaleSeriesInterval),
```
mdisibio: Let's make the default precision `10` a const, with a comment reclarifying the expected accuracy and memory usage would be great. i.e. +/- 3.25%.

modules/generator/registry/cardinality.go
carles-grafana: not sure how reliable go tickers are, this is just extra safety in case we miss 1 tick.

but if the ticker is reliable then yes, we could simplify it

modules/generator/registry/registry.go
carles-grafana: I'm reusing the hard-coded 5 minutes that is used for the existing ticker that removes stale series, as it makes sense the same is used for active series and the estimate. We could make it configurable indeed but I would leave it like this for the moment.

modules/generator/registry/gauge.go
carles-grafana: changed

modules/generator/registry/registry.go
```diff
@@ -184,33 +196,50 @@ func (r *ManagedRegistry) registerMetric(m metric) {
...
+	maxActiveSeries := r.overrides.MetricsGeneratorMaxActiveSeries(r.tenant)
+	r.metricMaxActiveSeries.Set(float64(maxActiveSeries))
+
+	if seriesDemand > int(maxActiveSeries) {
+		r.metricSeriesDemand.Set(float64(seriesDemand))
+	} else {
+		r.metricSeriesDemand.Set(float64(activeSeries))
+	}
+
 	if r.overrides.MetricsGeneratorDisableCollection(r.tenant) {
```
javiermolinar, line 236: Why not keeping this before locking?

modules/generator/registry/registry.go
```diff
@@ -83,12 +90,16 @@ type ManagedRegistry struct {
 // metric is the interface for a metric that is managed by ManagedRegistry.
 type metric interface {
 	name() string
-	collectMetrics(appender storage.Appender, timeMs int64) (activeSeries int, err error)
+	collectMetrics(appender storage.Appender, timeMs int64) error
+	countActiveSeries() int
```
javiermolinar, line 94: Could you document what countSeriesDemand is measuring?

modules/generator/registry/registry.go
```diff
@@ -184,33 +196,50 @@ func (r *ManagedRegistry) registerMetric(m metric) {
...
+		seriesDemand += m.countSeriesDemand()
+		activeSeries += m.countActiveSeries()
+	}
+
+	r.activeSeries.Store(uint32(activeSeries))
+	r.metricActiveSeries.Set(float64(activeSeries))
+	maxActiveSeries := r.overrides.MetricsGeneratorMaxActiveSeries(r.tenant)
+	r.metricMaxActiveSeries.Set(float64(maxActiveSeries))
+
+	if seriesDemand > int(maxActiveSeries) {
```
javiermolinar: This can happen due to an estimation error?

modules/generator/registry/cardinality.go
```diff
@@ -0,0 +1,142 @@
...
+		current:        0,
+		sketches:       sketches,
+		cachedMerge:    cachedMerge,
+		lastAdvance:    time.Now(),
+		precision:      precision,
+	}
+}
+
+func (c *Cardinality) Insert(hash uint64) {
+	c.mu.Lock()
```
javiermolinar, line 90: Since we are locking here, do we also need to lock in the caller function?

modules/generator/registry/cardinality.go
```diff
@@ -0,0 +1,142 @@
...
+
+// NewCardinality creates a new Cardinality estimate with a 3.25 standard error.
+func NewCardinality(staleTime, sketchDuration time.Duration) *Cardinality {
+	return newCardinality(10, staleTime, sketchDuration)
+}
+
+func newCardinality(precision uint8, staleTime, sketchDuration time.Duration) *Cardinality {
+	// If parameters are out of bounds, set defaults.
+	if precision < 4 || precision > 18 {
+		precision = 14
```
javiermolinar, line 60: Should not be 10 the default? Maybe we could pass it from the caller method

javiermolinar: Or create a constant

modules/generator/registry/cardinality.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	if staleTime <= 1*time.Minute || sketchDuration < 1*time.Minute || staleTime < sketchDuration {
+		staleTime = 15 * time.Minute
```
javiermolinar, line 63: Same, these should be constants

modules/generator/registry/registry.go
carles-grafana: when the activeSeries is less than the max, I set the estimate to the activeSeries so they match, to avoid confusion as can be slightly different

modules/generator/registry/registry.go
carles-grafana: we still want to count the estimate when the collection is disabled

modules/generator/registry/cardinality.go
carles-grafana: this mutex is only for the cardinality struct

modules/generator/registry/cardinality.go
carles-grafana: I've added the NewCardinality method that creates a new one with precision 10.

The conditionals here are just to prevent panics. 14 is the default in the HyperLogLog implementation we using.

modules/generator/registry/registry.go
javiermolinar: Ah right, good point

modules/generator/registry/cardinality.go
carles-grafana: it's just a sane default in case the provided times don't make sense

carles-grafana:
@javiermolinar  I included a few benchmarks, it's very fast:
```
go test ./modules/generator/registry -run=^$ -bench=BenchmarkCardinality -benchmem
goos: darwin
goarch: arm64
pkg: github.com/grafana/tempo/modules/generator/registry
cpu: Apple M3 Pro
BenchmarkCardinalityInsert-11      	166722549	         7.181 ns/op	       0 B/op	       0 allocs/op
BenchmarkCardinalityEstimate-11    	 1840472	       651.2 ns/op	    1184 B/op	      18 allocs/op
BenchmarkCardinalityAdvance-11     	 7457814	       163.2 ns/op	     320 B/op	       8 allocs/op
PASS
ok  	github.com/grafana/tempo/modules/generator/registry	5.561s
```

modules/generator/registry/registry.go
carles-grafana: added

carles-grafana:
after a discussion with @joe-elliott  I've renamed the metric to `tempo_metrics_generator_registry_active_series_demand_estimate` so it is explicit that it's an estimate, and the metric will always show the estimate even if it's below the max limit

PR #5712, opened 2025-10-07
test: fix test race condition in test logger

carles-grafana:
**What this PR does**:

Fix race conditions when calling the test log after the test has finished by adding a "done" flag in the test logger.
This can happen in tests with services that are not shut down correctly.

I've seen it in different test in the past months, and recently again in a blockbuilder tests such as: https://github.com/grafana/tempo/actions/runs/18311278340/job/52140208404?pr=5710#step:4:12769

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/util/test/logger.go
+++ b/pkg/util/test/logger.go
@@ -2,6 +2,7 @@ package test
 
 import (
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


joe-elliott (approved):
yuppers

PR #5716, opened 2025-10-08
vParquet5: fix issues related to integer dedicated columns

stoewer:
**What this PR does**:
This addresses two issues with dedicated attribute columns with integer values introduced in `vParquet5-preview2`. The functions `DedicatedColumns.Validate()` and `resourceSpanHash()` were not taking taking the added integer columns into account.

Fixes issues introduced by #5639

**Checklist**
- [x] Tests updated
- [ ] Documentation added


tempodb/encoding/vparquet5/rebatch_trace.go
```diff
@@ -176,18 +182,25 @@ func attributeHash(attr *Attribute, hash uint64) uint64 {
...
 	}
-	hash = addHash(attr.ValueUnsupported, hash)
+	hash = addHashStr(attr.ValueUnsupported, hash)
 
 	return hash
 }
 
-func addHash(s *string, hash uint64) uint64 {
+func addHashStr(s *string, hash uint64) uint64 {
 	if s == nil {
```
mdisibio, line 191: I know this wasn't changed and it must be working, but it seems like a potential bug to me.  If we skip nils, then do these layouts have the same hash?

(1) String01: foo, String02: nil
(2) String01: nil, String02: foo

In both cases it yields the same result: hash(foo)?  One fix is to add a separator in between like AddUint64(prime number), and always add it even if nil.

Then they become distinct:
(1) foo, sep, sep
(2) sep, foo, sep

Issue #5558, opened 2025-08-21
Provide a distro-based Tempo Docker image with networking tools for health checks


kilian-goetz:
**Is your feature request related to a problem? Please describe.**

Currently, the Tempo Docker images are “distro-less” and do not include basic tools like wget or curl. This makes it impossible to perform health checks or network tests directly within the container, which is problematic for automated monitoring.

**Describe the solution you'd like**

I would like Tempo to provide an additional set of Docker images based on a lightweight Linux distribution (e.g., Alpine or Debian), which includes essential networking tools like wget or curl. This would allow users to perform health checks and network diagnostics inside the container.

**Describe alternatives you've considered**

- Building custom images on top of the distro-less image and adding the missing tools manually.
- Running health checks from outside the container (host or sidecar container), though this adds complexity to monitoring setups.

**Additional context**

Using a “distro-based” Tempo image would simplify performing container health checks, debugging, and running Tempo in Docker Compose setups. It would also align with common practices for other observability tools that provide both minimal and distro-based images.

joe-elliott:
I'm not opposed to this, but the reason we use distro-less images are because they reduce the number of CVEs and, presumably, improve our security posture. We also have internal SLOs on addressing CVEs in our OSS images and will cut patch releases if we cross certain boundaries.

If we were to provide images with these tools, they would be provided as is. We would not track CVEs on the image or make an effort to remediate them.

kilian-goetz:
Hi @joe-elliott,

Thanks a lot for your quick answer, and apologies for the delay in getting back to you, the last few weeks have been a bit overwhelming.

I completely understand the security considerations for the Tempo team. However, this limitation can be quite restrictive for companies relying solely on Docker Compose, as we can’t easily perform health checks in this setup.

Would it be possible to provide a distro-based image, with a clear disclaimer stating that the distroless image is the recommended one and that the Tempo team won’t be responsible for any potential security issues in the distro-based version?

As a reference, I noticed that the Mimir team also provides a mimir-alpine image on Docker Hub [https://hub.docker.com/r/grafana/mimir-alpine](https://hub.docker.com/r/grafana/mimir-alpine). The latest push was around 3 months ago, so I’m not sure if it’s still actively maintained or if Grafana has changed its policy on this.

cmcgalliard:
Hey!
[...]

PR #5717, opened 2025-10-08
dockerfile: remove busybox

carles-grafana:
**What this PR does**:

Months ago, when we switched to distroless images we decided to use the "debug" version, which includes busybox, which could help for debugging purposes.

However, recently busybox had vulnerabilities and the 2.8.2 image is affected.

I propose to remove it to make the image even more minimal and prevent future vulnerabilities. Updated docker-compose files accordingly.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo/Dockerfile
+++ b/cmd/tempo/Dockerfile
@@ -1,19 +1,21 @@
 FROM alpine:latest AS ca-certificates
 RUN apk add --update --no-cache ca-certificates
 
-FROM gcr.io/distroless/static-debian12:debug
-
-# we need this because some docker-compose files call chown assuming there's a shell
-SHELL ["/busybox/sh", "-c"]
+# Use the same image with busybox for the setup to ensure the /etc files are compatible.
+FROM gcr.io/distroless/static-debian12:debug AS tempo-setup
 
 RUN ["/busybox/addgroup", "-g", "10001", "-S", "tempo"]
 RUN ["/busybox/adduser", "-u", "10001", "-S", "tempo", "-G", "tempo"]
-RUN ["/busybox/mkdir", "-p", "/var/tempo", "-m", "0700"]
-RUN ["/busybox/chown", "-R", "tempo:tempo", "/var/tempo"]
+RUN ["/busybox/mkdir", "-p", "/var/tempo"]
+
+FROM gcr.io/distroless/static-debian12
 
 ARG TARGETARCH
 COPY bin/linux/tempo-${TARGETARCH} /tempo
 COPY --from=ca-certificates /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+COPY --from=tempo-setup /etc/passwd /etc/passwd
+COPY --from=tempo-setup /etc/group /etc/group
+COPY --from=tempo-setup --chown=10001:10001 /var/tempo /var/tempo
 
 USER 10001:10001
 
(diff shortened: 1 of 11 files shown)


javiermolinar (commented):
This creates friction within the community. For example:
https://github.com/grafana/tempo/issues/5558

Do we provide any guidance on how to debug Tempo issues in running containers?

carles-grafana:
It's possible to debug a distroless container using Orbstack or Docker Desktop to open a shell. For Orbstack:
image

cmd/tempo/Dockerfile
```diff
@@ -1,19 +1,20 @@
 FROM alpine:latest AS ca-certificates
 RUN apk add --update --no-cache ca-certificates
 
-FROM gcr.io/distroless/static-debian12:debug
-
-# we need this because some docker-compose files call chown assuming there's a shell
-SHELL ["/busybox/sh", "-c"]
+FROM gcr.io/distroless/static-debian12:debug AS tempo-setup
```
javiermolinar, line 5: Why don't we use busybox base image directly? It states clearly that the goal of this stage is to create the users and groups. I had to look twice to understand why the gcr.io/distroless/static-debian12:debug was used to. It's also shorter

```
FROM busybox AS user-setup
RUN addgroup -g 10001 tempo && \
    adduser -D -u 10001 -G tempo tempo
```

carles-grafana: The reason is that in the final stage the /etc/passwd and /etc/group is replaced by those prepared in this stage. 
These files in busybox are different to the static-debian12.
I've verified it:
```
docker run -it busybox                                                  
/ # cat /etc/passwd
root:x:0:0:root:/root:/bin/sh
daemon:x:1:1:daemon:/usr/sbin:/bin/false
bin:x:2:2:bin:/bin:/bin/false
sys:x:3:3:sys:/dev:/bin/false
sync:x:4:100:sync:/bin:/bin/sync
mail:x:8:8:mail:/var/spool/mail:/bin/false
www-data:x:33:33:www-data:/var/www:/bin/false
operator:x:37:37:Operator:/var:/bin/false
nobody:x:65534:65534:nobody:/home:/bin/false
/ # exit

docker run -it gcr.io/distroless/static-debian12:debug                   
/ # cat /etc/passwd
root:x:0:0:root:/root:/sbin/nologin
nobody:x:65534:65534:nobody:/nonexistent:/sbin/nologin
nonroot:x:65532:65532:nonroot:/home/nonroot:/sbin/nologin
```

And in static-debian12:latest, by attaching a shell:
```
❯ cat /etc/passwd
root:x:0:0:root:/root:/sbin/nologin
nobody:x:65534:65534:nobody:/nonexistent:/sbin/nologin
nonroot:x:65532:65532:nonroot:/home/nonroot:/sbin/nologin
```

But it's a good question, I'll add a comment to the Dockerfile

cmd/tempo/Dockerfile
```diff
@@ -1,19 +1,21 @@
...
-
-# we need this because some docker-compose files call chown assuming there's a shell
-SHELL ["/busybox/sh", "-c"]
+# Use the same image with busybox for the setup to ensure the /etc files are compatible.
+FROM gcr.io/distroless/static-debian12:debug AS tempo-setup
 
 RUN ["/busybox/addgroup", "-g", "10001", "-S", "tempo"]
 RUN ["/busybox/adduser", "-u", "10001", "-S", "tempo", "-G", "tempo"]
-RUN ["/busybox/mkdir", "-p", "/var/tempo", "-m", "0700"]
-RUN ["/busybox/chown", "-R", "tempo:tempo", "/var/tempo"]
```
mapno, line 12: Is `chown` not needed with this change?

carles-grafana: chown is now done by the COPY instruction below

PR #5721, opened 2025-10-08
Enable histogram tuning in the user-configurable overrides API

zalegrala:
**What this PR does**:

Here we include two settings in the user-configurable overrides API to allow enablement of the native histograms, and also the max bucket configuration of the native histograms.

Related #5638

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/operations/manage-advanced-systems/user-configurable-overrides.md
+++ b/docs/sources/tempo/operations/manage-advanced-systems/user-configurable-overrides.md
@@ -63,6 +63,8 @@ metrics_generator:
   [processors: <list of strings>]
   [collection_interval: <duration>]
   [disable_collection: <bool> | default = false]
+  [generate_native_histograms: <classic|native|both> | default = classic]
+  [native_histogram_max_bucket_number: <int> | default = 100]
 
   processor:
 


modules/generator/registry/overrides.go
```diff
@@ -4,13 +4,14 @@ import (
...
 )
 
 type Overrides interface {
 	MetricsGeneratorMaxActiveSeries(userID string) uint32
 	MetricsGeneratorCollectionInterval(userID string) time.Duration
 	MetricsGeneratorDisableCollection(userID string) bool
-	MetricsGeneratorGenerateNativeHistograms(userID string) overrides.HistogramMethod
+	MetricsGeneratorGenerateNativeHistograms(userID string) histograms.HistogramMethod
 	MetricsGenerationTraceIDLabelName(userID string) string
 	MetricsGeneratorNativeHistogramBucketFactor(userID string) float64
```
electron0zero, line 16: does it make sense to expose `MetricsGeneratorNativeHistogramBucketFactor` and `MetricsGeneratorNativeHistogramMinResetDuration` as well? so that all native histogram settings are expose via the user configurable overrides?

electron0zero: discussed it offline, better to leave them out for now, and add it later when we need them.

PR #5723, opened 2025-10-09
Proposal: introduce hints for frontend performance tests

ruslan-mikhailov:
**What this PR does**: it introduces the following hints:
`{} with (debug_return_in=100s, debug_std_dev=10s, debug_data_factor=0.1)`

- `debug_return_in` - adds latency for queriers (disabled by default)
- `debug_std_dev` - adds dispersion to wait time (0 by default)
- ~~`fake` - makes querier return fake traces instead of querying actual data.~~ removed
- `debug_data_factor` - probability of non-empty response from a querier (0 by default)

In order to enable it, need to allow tenant unsafe queries by setting `overrides.tenant_id.read.unsafe_query_hints`.

These hints help to pick optimal query-frontend settings

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/engine.go
+++ b/pkg/traceql/engine.go
@@ -63,7 +63,11 @@ func (e *Engine) ExecuteSearch(ctx context.Context, searchReq *tempopb.SearchReq
 		}
 		simulateLatency(returnIn, stdDev)
 
-		return generateFakeSearchResponse(1), nil
+		var probability float64
+		if p, ok := rootExpr.Hints.GetFloat(HintDebugDataFactor, allowUnsafeQueryHints); ok {
+			probability = p
+		}
+		return generateFakeSearchResponse(probability), nil
 	}
 
 	var mostRecent, ok bool
--- a/pkg/traceql/enum_hints.go
+++ b/pkg/traceql/enum_hints.go
@@ -17,8 +17,9 @@ const (
 	HintMostRecent        = "most_recent" // traceql search hint to return most recent results ordered by time
 	HintDebug             = "debug"
 	HintInfo              = "info"
-	HintDebugReturnIn     = "debug_return_in" // performance testing hint to simulate query latency and return fake data
-	HintDebugStdDev       = "debug_std_dev"   // standard deviation for debug_return_in latency simulation
+	HintDebugReturnIn     = "debug_return_in"   // performance testing hint to simulate query latency and return fake data
+	HintDebugStdDev       = "debug_std_dev"     // standard deviation for debug_return_in latency simulation
+	HintDebugDataFactor   = "debug_data_factor" // performance testing hint to control the possibility of non-empty fake data
 )
 
 func isUnsafe(h string) bool {
--- a/pkg/traceql/fake_data_test.go
+++ b/pkg/traceql/fake_data_test.go
@@ -18,7 +18,7 @@ func TestPerformanceTestingHints_Search(t *testing.T) {
 		engine := NewEngine()
 
 		req := &tempopb.SearchRequest{
-			Query: `{} with (debug_return_in=100s)`,
+			Query: `{} with (debug_return_in=100s, debug_data_factor=1)`,
 			Start: uint32(time.Now().Add(-1 * time.Hour).Unix()),
 			End:   uint32(time.Now().Unix()),
 		}
@@ -39,7 +39,7 @@ func TestPerformanceTestingHints_SearchWithStdDev(t *testing.T) {
 		engine := NewEngine()
 
 		req := &tempopb.SearchRequest{
-			Query: `{} with (debug_return_in=100s, debug_std_dev=10s)`,
+			Query: `{} with (debug_return_in=100s, debug_std_dev=100ms, debug_data_factor=1)`,
 			Start: uint32(time.Now().Add(-1 * time.Hour).Unix()),
 			End:   uint32(time.Now().Unix()),
 		}


joe-elliott:
Should we prefix them in a way to communicate they are meant to be used together:
```
test_?
something_better_?
```

also, do we need the `fake` param? if `return_in` is present you can assume you should return fake data. if `std_dev` is present but `return_in` is not present i think you just ignore the params and execute normally.

pkg/traceql/enum_hints.go
```diff
@@ -17,6 +17,9 @@ const (
 	HintMostRecent        = "most_recent" // traceql search hint to return most recent results ordered by time
 	HintDebug             = "debug"
 	HintInfo              = "info"
+	HintDebugReturnIn     = "debug_return_in" // performance testing hint to simulate query latency
+	HintDebugStdDev       = "debug_std_dev"   // standard deviation for debug_return_in latency simulation
+	HintDebugFakeData     = "debug_fake_date" // return fake data instead of querying actual data
```
ruslan-mikhailov: @joe-elliott, answering you here https://github.com/grafana/tempo/pull/5723#issuecomment-3385616415

I have no strong opinion whether we should remove `fake_data` param or not.

Initially, I thought it is not obvious that `return_in` also forces returning fake data, but `debug_` prefix indicates it is not for regular use anyway. So far, I don't see cases where real data is needed when artificial latency is enabled.

To reduce the number of entities, I'll remove it.

pkg/traceql/fake_data.go
```diff
@@ -0,0 +1,97 @@
...
+	"math/rand"
+	"time"
+
+	"github.com/grafana/tempo/pkg/tempopb"
+	v1 "github.com/grafana/tempo/pkg/tempopb/common/v1"
+)
+
+// GenerateFakeSearchResponse creates a fake SearchResponse.
+// It must be used only for testing purposes.
+func GenerateFakeSearchResponse() *tempopb.SearchResponse {
```
joe-elliott: make these funcs internal to the package?

pkg/traceql/fake_data.go
```diff
@@ -0,0 +1,97 @@
...
+	r := rand.New(rand.NewSource(time.Now().UnixNano())) //nolint:gosec // G404
+
+	numTraces := 3 + r.Intn(8)
```
joe-elliott: we need to be careful about the fake data generated. if we have a limit of 20 in the frontend and every job returns 3->11 responses then the whole query will be basically instantly satisfied.

i apologize for not thinking about this before, but data recombination is probably more important in metrics and traceql searches with most_recent=true b/c these are exhaustive. in normal traceql search returning any data will basically make it instantly finish.

if you'd like to keep the fake data then we probably need ways to control it. so we can force searches to be completely exhaustive.

pkg/traceql/fake_data.go
```diff
@@ -0,0 +1,97 @@
...
+			InspectedTraces: uint32(numTraces),
+			InspectedBytes:  uint64(r.Intn(1000000)),
+		},
+	}
+}
+
+// SimulateLatency sleeps for the specified duration with optional standard deviation
+func SimulateLatency(duration time.Duration, stdDev time.Duration) {
+	if stdDev > 0 {
+		variation := time.Duration(rand.NormFloat64() * float64(stdDev)) //nolint:gosec // G404
```
joe-elliott, line 89: `NormFloat64` is awesome! i was wondering how gnarly the code would be to correctly implement the stdDev hint. this is way better than i thought it would be.

pkg/traceql/fake_data.go
```diff
@@ -0,0 +1,97 @@
...
+		// possibility of having a crazy big number is never a zero
+		// capping to 3 sigmas (0.3% possibility)
```
joe-elliott, line 91: good catch. not sure i would have thought of this.

pkg/traceql/engine.go
```diff
@@ -55,8 +55,19 @@ func (e *Engine) ExecuteSearch(ctx context.Context, searchReq *tempopb.SearchReq
...
 
+	// Check for performance testing hints
+	if returnIn, ok := rootExpr.Hints.GetDuration(HintDebugReturnIn, allowUnsafeQueryHints); ok {
+		var stdDev time.Duration
+		if stdDevDuration, ok := rootExpr.Hints.GetDuration(HintDebugStdDev, allowUnsafeQueryHints); ok {
+			stdDev = stdDevDuration
+		}
+		SimulateLatency(returnIn, stdDev)
+
+		return GenerateFakeSearchResponse(), nil
```
javiermolinar: Should the fake data be generated before the latency is included? That way we can subtract the time it takes to generate it

ruslan-mikhailov: Checked with a simple benchmark, the function takes ~0.1ms to execute. I think, it's too small to take into account.

pkg/traceql/fake_data.go
ruslan-mikhailov: I propose one more hint `debug_data_factor`(not the best naming, maybe you have better ideas?), added in the last 2 commits. It can control the possibility of a success of a query. This way we can return no data at all (default, 0), always return (1) or sometimes based on the given number.

wdyt?

ruslan-mikhailov: Additionally, I reduced number of returned traces to 1-3

pkg/traceql/fake_data.go
ruslan-mikhailov: \+

fixed

pkg/traceql/fake_data.go
joe-elliott: love it. 👍

ruslan-mikhailov:

ruslan-mikhailov:
\+ very minor fix for new test

PR #5725, opened 2025-10-10
Improve shutdown time in the first 30 seconds

ldufr:
**What this PR does**:

Trying to shutdown (SIGINT) tempo in the first 30 seconds would block until we got a stable key which currently takes 30 seconds. With this change, it should be instantaneous.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


zalegrala (approved):
Looks good to me.

Issue #3414, opened 2024-02-20
Metrics Generator:: Include server/server spans as well


dkrizic:
**Is your feature request related to a problem? Please describe.**

I have an ingress-nginx that emits spans of type SERVER and calls a backend where the first span is of type SERVER as well. The Metrics Generator does not create a connection between ingress-nginx. I presume that is due to the fact that the Metrics Generator only evaluates CLIENT->SERVER spans.

**Describe the solution you'd like**

A span of type SERVER->SERVER should be handled identically to CLIENT->SERVER

**Describe alternatives you've considered**

An alternative would be to change the ingress-nginx OTEL integration to emit two spans, first a SERVER span and a child span of type CLIENT.

joe-elliott:
Agree this would be a nice improvement to the generator. It might be difficult to pull off because the current code uses the client/server labels to track edges:

https://github.com/grafana/tempo/blob/main/modules/generator/processor/servicegraphs/servicegraphs.go#L174

One option would be to track all spans regardless of the span kind and note any cross service calls, but obviously this would be more costly. It's also tough b/c if a server->server edge is found we still need to keep the edge around in case the matching client->server shows up. There are some complexities but this does happen in the real world and it would be nice to handle it.

Does client->client ever occur?

msvechla:
Hi, I stumbled across this issue. Is it true that tempo metrics generator only considers `CLIENT->SERVER` spans? Does it also consider `SERVER->CLIENT` spans?

Because I am currently comparing tempo metrics generator with the otel metrics generator and it looks like `SERVER->CLIENT` span metrics are missing

dkrizic:
Hi @msvechla, I did not test it, but according to their documentation, they only consider CLIENT->SERVER and PRODUCER->CONSUMER spans. Others do the same, e.g. Datadog or Azure Monitor. I am also talking to the ingress-nginx guys to create a second span of type CLIENT in order to complete my diagram. https://github.com/kubernetes/ingress-nginx/issues/11002#

PR #5727, opened 2025-10-11
servicegraphs: support SERVER→SERVER edges (dual-upsert, parent synth…

brelandj:
servicegraphs: Adding support for SERVER→SERVER edges (dual-upsert, parent synthesis

**What this PR does**:
Adding support for server to server edges

Fixes #Metrics Generator:: Include server/server spans as well
[#3414](https://github.com/grafana/tempo/issues/3414)

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/processor/servicegraphs/servicegraphs.go
+++ b/modules/generator/processor/servicegraphs/servicegraphs.go
@@ -186,7 +186,6 @@ func (p *Processor) consume(resourceSpans []*v1_trace.ResourceSpans) (err error)
 						p.upsertPeerNode(e, span.Attributes)
 						p.upsertDatabaseRequest(e, rs.Resource.Attributes, span)
 					})
-
 				case v1_trace.Span_SPAN_KIND_CONSUMER:
 					// override connection type and continue processing as span kind server
 					connectionType = store.MessagingSystem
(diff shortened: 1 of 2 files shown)


modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -170,6 +170,35 @@ func (p *Processor) consume(resourceSpans []*v1_trace.ResourceSpans) (err error)
...
 
+	// Pre-index server-like spans (SERVER and CONSUMER) to enable SERVER->SERVER edges
+	type serverSpanInfo struct {
+		serviceName   string
+		resourceAttrs []*v1_common.KeyValue
+		spanAttrs     []*v1_common.KeyValue
+		span          *v1_trace.Span
+	}
+
+	serverIndex := make(map[string]serverSpanInfo, 1024)
```
joe-elliott: why are we building this map? why the value 1024?

the code using it to pair below seems to assume that the parent will be in the same batch as the child which is not necessarily true.

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -217,6 +245,53 @@ func (p *Processor) consume(resourceSpans []*v1_trace.ResourceSpans) (err error)
...
+							e.ClientService = svcName
+							e.ClientLatencySec = spanDurationSec(span)
+							e.ClientEndTimeUnixNano = span.EndTimeUnixNano
+							e.Failed = e.Failed || p.spanFailed(span)
+							p.upsertDimensions("client_", e.Dimensions, rs.Resource.Attributes, span.Attributes)
+							p.upsertPeerNode(e, span.Attributes)
+							// keep multiplier consistent
+							e.SpanMultiplier = spanMultiplier
+						}
+					}); err2 != nil {
```
joe-elliott, line 223: would be better to accumulate and count all at once below instead of splitting this code.

joe-elliott (commented):
you have a failing test you need to correct. also we would definitely need additional tests for this feature.

can you talk through the approach and explain what you're trying to accomplish. draw out some different trace structures and explain how your code handles them.

it would help me review.

anglerfishlyy:
Hey @brelandj, looks like the `make test-with-cover-others` check is failing on CI , you might want to run it locally to see what’s causing it.

modules/generator/processor/servicegraphs/servicegraphs.go
brelandj: The map was from old testing. Removed it

The approach is upserting the server side at key(traceId, parentSpanId). Additionally, upserting the client side at key(traceId, spanId) from the parent SERVER span. When the child SERVER arrives, both upserts meet on the same key and the edge completes. And only set client fields if they’re empty, so a real client span always wins. Removed the temporary same-batch serverIndex and parent synthesis. The dual-upsert works across separate OTLP requests within the store wait window.

modules/generator/processor/servicegraphs/servicegraphs.go
brelandj: Updated - Metrics are only emitted on onComplete (when both sides are present). The extra upsert only populates missing fields

PR #5741, opened 2025-10-14
Allow RetryInfo to be disabled in per tenant overrides

electron0zero:
**What this PR does**:
Add a flag to enable or disable RetryInfo at cluster level, and per tenant level overrides.

This allow end users to enable RetryInfo, and be more compliant with spec: https://opentelemetry.io/docs/specs/otlp/#failures

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 ## main / unreleased
 
 * [CHANGE] Remove busybox from Tempo image to make it more minimal and prevent future vulnerabilities [#5717](https://github.com/grafana/tempo/pull/5717) (@carles-grafana)
+* [CHANGE] Allow RetryInfo to be disabled in per tenant overrides [#5741](https://github.com/grafana/tempo/pull/5741) (@electron0zero)
 * [FEATURE] Add `tempo_metrics_generator_registry_active_series_demand_estimate` that estimates metrics-generator active series demand even when the active series limit is reached [#5710](https://github.com/grafana/tempo/pull/5710) (@carles-grafana)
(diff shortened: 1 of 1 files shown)


modules/distributor/config.go
```diff
@@ -54,6 +54,8 @@ type Config struct {
 	// configures the distributor to indicate to the client that it should retry resource exhausted errors after the
 	// provided duration
 	RetryAfterOnResourceExhausted time.Duration `yaml:"retry_after_on_resource_exhausted"`
+	// control retryInfo at cluster level
+	RetryInfoEnabled bool `yaml:"retry_info_enabled"`
```
joe-elliott: why are we adding a second config option. retry_after_on_resource_exhausted can already be used to turn this feature on/off

electron0zero: not really needed, will remove, added it because I wanted to match the overrides.

we can keep using `RetryAfterOnResourceExhausted > 0` as a way to see if it's enabled or not.

carles-grafana:
The code looks good but can you remind me why we want this to be optionally turned off?

electron0zero:
> The code looks good but can you remind me why we want this to be optionally turned off?

@carles-grafana One case where this would cause problems is if you are sending a huge amount of data into tempo, and most of the writes are being refused due to rate limits. in that case, if you have Retries configured in the collector, collector will hold on to data and retry after the RetryDelay. 

In that case where the collector is getting tons of rate limit errors from tempo, these retryInfo responses will cause the collector to hold onto more and more data for retries, and collectors will starting OOMing if they are collector queue or retry polices are not configured to expect RetryInfo. 

in that case, having a per tenant toggle gives us the control over retryInfo at tenant level, you can disable it for tenant who don't want this enabled.

PR #5762, opened 2025-10-17
wip

carles-grafana:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/util/tenantid_test.go
+++ b/pkg/util/tenantid_test.go
@@ -1,0 +1,90 @@
+package util
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/grafana/dskit/tenant"
+	"github.com/grafana/dskit/user"
+	"github.com/stretchr/testify/require"
+)
+
+func TestExtractValidOrgID(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name    string
+		orgID   string
+		want    string
+		wantErr bool
+	}{
+		{
+			name:  "valid single tenant",
+			orgID: "tenant-123",
+			want:  "tenant-123",
+		},
+		{
+			name:    "invalid single tenant",
+			orgID:   "tenant#123",
+			wantErr: true,
+		},
+		{
+			name:    "invalid single tenant unsupported character slash",
+			orgID:   "tenant/123",
+			wantErr: true,
+		},
+		{
+			name:    "invalid single tenant unsafe path segment",
+			orgID:   "..",
+			wantErr: true,
+		},
+		{
+			name:    "invalid single tenant too long",
+			orgID:   strings.Repeat("a", tenant.MaxTenantIDLength+1),
+			wantErr: true,
+		},
+		{
+			name:  "valid multi tenant",
+			orgID: "tenantA|tenantB",
+			want:  "tenantA|tenantB",
+		},
+		{
+			name:    "invalid multi tenant",
+			orgID:   "tenantA|tenant#B",
+			wantErr: true,
+		},
+		{
+			name:    "invalid multi tenant with unsafe path segment",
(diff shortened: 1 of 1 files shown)


pkg/util/tenantid.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// ExtractValidOrgID extracts and validates tenant string from context.
+func ExtractValidOrgID(ctx context.Context) (string, error) {
+	id, err := tenant.TenantID(ctx)
+	if err == nil {
+		return id, nil
+	}
+
+	if !errors.Is(err, user.ErrTooManyOrgIDs) {
```
electron0zero, line 18: we can skip this logic and just use `tenant.TenantIDs` and it would validate all tenant IDs.

pkg/util/tenantid.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"context"
+	"errors"
+
+	"github.com/grafana/dskit/tenant"
+	"github.com/grafana/dskit/user"
+)
+
+// ExtractValidOrgID extracts and validates tenant string from context.
+func ExtractValidOrgID(ctx context.Context) (string, error) {
```
electron0zero, line 12: let's add some units test for it.

modules/distributor/distributor.go
```diff
@@ -396,7 +396,7 @@ func (d *Distributor) checkForRateLimits(tracesSize, spanCount int, userID strin
 }
 
 func (d *Distributor) extractBasicInfo(ctx context.Context, traces ptrace.Traces) (userID string, spanCount, tracesSize int, err error) {
-	user, e := user.ExtractOrgID(ctx)
+	user, e := util.ExtractValidOrgID(ctx)
```
electron0zero, line 399: I would love to see e2e tests for write path and read path. just to be sure that we don't break it in the future.

pkg/util/tenantid.go
carles-grafana: I've added a benchmark to compare both implementations

pkg/util/tenantid_benchmark_test.go
```diff
@@ -0,0 +1,90 @@
...
+		result = id
+	}
+
+	benchmarkExtractValidOrgID2 = result
+}
+
+// ./pkg/util/compare_benchmarks.sh
+//                      │ /tmp/BenchmarkExtractValidOrgID.txt │ /tmp/BenchmarkExtractValidOrgID2.txt │
+//                      │               sec/op                │   sec/op     vs base                 │
+// ExtractValidOrgID-11                           15.28n ± 1%   37.16n ± 1%  +143.31% (p=0.000 n=10)
```
carles-grafana, line 52: @electron0zero  This compares the implementation that assumes it gets a single tenant with the other that assumes multiple tenants. We can see the second is slower and allocates memory. Although we are talking about nanoseconds, this will run for every single request so I think the first implementation is worth it even if it has a few extra lines, as single-tenant is the most common request.

I will remove this benchmark once approved.

pkg/util/tenantid.go
carles-grafana: done

modules/frontend/search_handlers.go
```diff
@@ -51,7 +51,7 @@ func newSearchStreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTripper[c
 		}
 
 		httpReq = httpReq.WithContext(ctx)
-		tenant, _ := user.ExtractOrgID(ctx)
+		tenant, _ := util.ExtractValidOrgID(ctx)
```
carles-grafana, line 54: we are not checking the err in some handlers currently, so it's not as trivial as replacing the call

PR #5771, opened 2025-10-20
Add block size to head block cut log message

### LINKED ISSUE #5770: Add block size to head block cut log message

I recently experienced repeated OOMKilled issues in the ingester due to simultaneous flushing of large blocks. The root cause was that multiple oversized head blocks were being cut and flushed concurrently, causing memory spikes that exceeded Kubernetes resource limits.

After extensive troubleshooting and log analysis, I resolved the issue by tuning `max_block_bytes` and `max_block_duration`. However, it took significant time to identify that abnormally large blocks were the culprit because the current log message doesn't include the block size:

```
level=info msg="head block cut. enqueueing flush op" tenant=*** block=0e5151dd-4a5f-4fbe-b24e-d946f6faf894
```

Without size information, it's hard to tell:
- Whether blocks are being cut due to size limit or duration limit
- If certain tenants are generating abnormally large blocks
- The correlation between block sizes and memory pressure/OOMKilled events

Add a `size` field to the "head block cut" log message to show the actual size of the block being cut:

```
level=info msg="head block cut. enqueueing flush op" tenant=*** block=0e5151dd-4a5f-4fbe-b24e-d946f6faf894 size=134217728
```

This would provide immediate visibility into:
- Real-time block sizes for each tenant
- Whether blocks are hitting the configured `max_block_bytes` limit (e.g., 128 MiB = 134217728 bytes)
- Patterns that lead to memory pressure and OOMKilled events
- Data-driven decisions for tuning block size limits

**Implementation approach:**
Modify `modules/ingester/flush.go` to retrieve the block size from the completing blocks list after `CutBlockIfReady()` returns and include it in the log message.

1. **Using existing metrics**: The `tempo_ingester_flush_size_bytes` histogram exists, but it only records sizes after blocks are flushed to backend storage, which is too late for correlating with OOMKilled events that occur during the completion phase.

2. **Enabling debug logging**: This generates excessive noise in production and doesn't provide structured data for analysis.

3. **Modifying `CutBlockIfReady()` return signature**: Returning both block ID and size would require updating all callers and test code, making it more invasive.

4. **Adding size to "block completed" log**: By the time completion finishes, the memory spike has already occurred, making it less useful for real-time incident diagnosis.

**Additional context**

This enhancement is particularly valuable for:
- **Incident response**: Quickly identifying memory pressure causes during OOMKilled events
- **Multi-tenant environments**: Understanding which tenants generate large blocks that could impact stability
- **Configuration optimization**: Making data-driven decisions for `max_block_bytes` and `max_block_duration` settings per tenant
- **Capacity planning**: Understanding typical block size patterns across different workloads

When oversized blocks are cut simultaneously across multiple tenants, they consume significant memory during the completion and flush process, which can trigger OOMKilled in resource-constrained Kubernetes environments. Having block size in the logs provides the critical data needed to tune ingester configuration and prevent such issues.

MinhoJJang:
## What this PR does / why we need it

This PR adds a `size` field to the "head block cut" log message in the ingester. Currently, when a block is cut, we only log the tenant and block ID, making it difficult to diagnose memory-related issues and understand block cutting behavior.

## Changes

- Modified `modules/ingester/flush.go` to retrieve and log the block size when cutting a head block
- The block size is obtained from the completing blocks list after `CutBlockIfReady()` returns

## Example log output

**Before:**
`level=info msg="head block cut. enqueueing flush op" tenant=*** block=0e5151dd-4a5f-4fbe-b24e-d946f6faf894`

**After:**
`level=info msg="head block cut. enqueueing flush op" tenant=*** block=0e5151dd-4a5f-4fbe-b24e-d946f6faf894 size=134217728`

## Benefits

- Better observability for block cutting behavior
- Easier debugging of OOMKilled issues caused by concurrent flushing of large blocks
- Quick identification of whether blocks are cut by size limits or duration limits
- Data-driven decisions for tuning `max_block_bytes` and `max_block_duration` settings

## Testing

- All existing tests pass
- No new tests added as this is a logging-only change

## Checklist

- [x] Code follows the project's coding standards
- [x] Ran `make fmt`
- [x] Ran `make lint`
- [x] Tests pass locally
- [ ] Updated documentation (if needed)
- [ ] Added changelog entry (if needed)

Fixes #5770

### DIFF
--- a/modules/ingester/instance.go
+++ b/modules/ingester/instance.go
@@ -290,7 +290,7 @@ func (i *instance) CutBlockIfReady(maxBlockLifetime time.Duration, maxBlockBytes
 		blockID := (uuid.UUID)(completingBlock.BlockMeta().BlockID)
 		blockSize := completingBlock.DataLength()
 
-		level.Info(i.logger).Log("msg", "head block cut. enqueueing flush op", "block", blockID, "size", blockSize)
+		level.Info(i.logger).Log("msg", "head block cut", "block", blockID, "size", blockSize)
 
 		// Now that we are adding a new block take the blocks mutex.
 		// A warning about deadlocks!!  This area does a hard-acquire of both mutexes.
--- a/modules/ingester/flush.go
+++ b/modules/ingester/flush.go
@@ -193,6 +193,7 @@ func (i *Ingester) cutOneInstanceToWal(instance *instance, immediate bool) {
 	}
 
 	if blockID != uuid.Nil {
+		level.Info(log.Logger).Log("msg", "enqueueing flush op", "tenant", instance.instanceID, "block", blockID)
 		// jitter to help when flushing many instances at the same time
 		// no jitter if immediate (initiated via /flush handler for example)
 		i.enqueue(&flushOp{


joe-elliott:
Nice PR! Thanks for the improvement. 

Two things:

- Let's move this log message into the instance itself. I think it will be cleaner. Here: https://github.com/grafana/tempo/blob/main/modules/ingester/instance.go#L270
- In the new architecture we are moving this functionality to the livestore. Can you add a similar log message here so it will be available in Tempo 3.0:  https://github.com/grafana/tempo/blob/main/modules/livestore/instance.go#L378

MinhoJJang:
> Nice PR! Thanks for the improvement.
> 
> Two things:
> 
> * Let's move this log message into the instance itself. I think it will be cleaner. Here: https://github.com/grafana/tempo/blob/main/modules/ingester/instance.go#L270
> * In the new architecture we are moving this functionality to the livestore. Can you add a similar log message here so it will be available in Tempo 3.0:  https://github.com/grafana/tempo/blob/main/modules/livestore/instance.go#L378

Thank you for the quick review! I've updated the PR with your suggestions:

**Changes made:**
1. ✅ Moved the log message into [`modules/ingester/instance.go:293`](https://github.com/grafana/tempo/blob/main/modules/ingester/instance.go#L293) within the `CutBlockIfReady()` function  
2. ✅ Added the same log message to [`modules/livestore/instance.go:402`](https://github.com/grafana/tempo/blob/main/modules/livestore/instance.go#L402) within the `cutBlocks()` function for Tempo 3.0 compatibility  
3. ✅ Removed the logging code from `modules/ingester/flush.go` (remove duplicate logs)

**Example outputs:**

**Ingester (current):**
`level=info tenant=* msg="head block cut. enqueueing flush op" block=0e5151dd-4a5f-4fbe-b24e-d946f6faf894 size=134217728`

**Livestore (Tempo 3.0):**
`level=info tenant=* msg="queueing wal block for completion" block=0e5151dd-4a5f-4fbe-b24e-d946f6faf894 size=134217728`


****Note:** The `tenant` field is automatically included in the logs because both `i.logger` (ingester) and `i.logger` (livestore) are initialized with the tenant ID using `kitlog.With()` during instance creation.**

The implementation is now cleaner, and the size information is logged precisely at the point where the block is cut, ensuring consistency across both architectures.
This placement also aligns with the original goal of improving visibility into block-cutting behavior.

modules/ingester/instance.go
```diff
@@ -287,6 +287,10 @@ func (i *instance) CutBlockIfReady(maxBlockLifetime time.Duration, maxBlockBytes
 		}
 
 		completingBlock := i.headBlock
+		blockID := (uuid.UUID)(completingBlock.BlockMeta().BlockID)
+		blockSize := completingBlock.DataLength()
+
+		level.Info(i.logger).Log("msg", "head block cut. enqueueing flush op", "block", blockID, "size", blockSize)
```
joe-elliott: this log message is not necessarily true. we could return an err on line 308 and the flush op would not be enqueued.

- keep the log message in flush.go that indicates the flush op is enqueued
- change this log message to simply indicate the head block has been cut

MinhoJJang: Thanks for the feedback! The changes have been applied here:
[5005aa6](https://github.com/grafana/tempo/pull/5771/commits/5005aa668f90df3aac2360c1f40543e6424b7700)

Please let me know if you have any further suggestions.

PR #5781, opened 2025-10-22
Add metric to track added latency to write requests

mapno:
**What this PR does**:

Add metric to track added latency to write requests

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/artificial_delay.go
+++ b/modules/distributor/artificial_delay.go
@@ -3,7 +3,19 @@ package distributor
 import (
 	"math/rand"
 	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 )
+
+var metricAddedLatency = promauto.NewHistogramVec(prometheus.HistogramOpts{
+	Namespace:                       "tempo",
+	Subsystem:                       "distributor",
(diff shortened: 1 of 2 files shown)


modules/distributor/artificial_delay.go
```diff
@@ -3,8 +3,20 @@ package distributor
...
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
+var metricAddedLatency = promauto.NewHistogramVec(prometheus.HistogramOpts{
+	Namespace:                       "tempo",
+	Subsystem:                       "distributor",
+	Name:                            "added_latency_seconds",
+	NativeHistogramBucketFactor:     1.1,
```
javiermolinar, line 15: It's not 1.1 to fine grained?

mapno: TBH I don't know, I used what's recommended in the docs

```
	// [...] A generally good trade-off
	// between cost and accuracy is a value of 1.1 (each bucket is at most
	// 10% wider than the previous one), which will result in each power of
	// two divided into 8 buckets (e.g. there will be 8 buckets between 1
	// and 2, same as between 2 and 4, and 4 and 8, etc.).
```

PR #5782, opened 2025-10-22
[Enhancement] heavy query weights

ruslan-mikhailov:
**What this PR does**: It adds three more conditions when to add +1 to weight:
1. If AllConditions is false
2. Full trace is needed
3. SecondPassSelectAll is true

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * [ENHANCEMENT] Improve shutdown time in the first 30 seconds [#5725](https://github.com/grafana/tempo/pull/5725) (@ldufr)
 * [ENHANCEMENT] Add metric for tracking added latency to write requests [#5781](https://github.com/grafana/tempo/pull/5781) (@mapno)
 * [ENHANCEMENT] Improve error message when overrides fail to parse [#5787](https://github.com/grafana/tempo/pull/5787) (@mapno)
+* [ENHANCEMENT] Increase weight for heavy TraceQL queries [#5782](https://github.com/grafana/tempo/pull/5782) (@ruslan-mikhailov)
 * [BUGFIX] Fix compactor to properly consider SSE-KMS information during metadata copy [#5774](https://github.com/grafana/tempo/pull/5774) (@steffsas)
 * [BUGFIX] Correctly track and reject too large traces in live stores. [#5757](https://github.com/grafana/tempo/pull/5757) (@joe-elliott)
 * [BUGFIX] Fix issues related to integer dedicated columns in vParquet5-preview2 [#5716](https://github.com/grafana/tempo/pull/5716) (@stoewer)


modules/frontend/pipeline/async_weight_middleware_test.go
```diff
@@ -63,48 +65,143 @@ func TestWeightMiddlewareForTraceQLRequest(t *testing.T) {
...
+		},
+		{
+			query:    "{span.a = \"1\"} ~ {span.b = \"2\"}",
+			expected: TraceQLSearchWeight + 2,
+		},
+
+		{
+			// Aggregate requiring full trace
+			query:    "{} | count() > 5",
+			expected: TraceQLSearchWeight + 2, // +1 for AllConditions is false, +1 for NeedsFullTrace
```
ruslan-mikhailov, line 142: To be sure I understand well what part of the code added weight, I modified it to "weight += 10, 100, 1000` and so on (different on each line), run the test and double-checked comments.

modules/frontend/pipeline/async_weight_middleware.go
```diff
@@ -120,6 +121,24 @@ func (c weightRequestWare) setTraceQLWeight(req Request) {
...
+
+	// Queries with OR operations (AllConditions=false) are more expensive
+	// because they can't be optimized in the storage layer
+	if !spanRequest.AllConditions {
+		weight++
+	}
+
+	// Query that requere full trace scanning, e.g. with structural operators
+	if rootExpr != nil && rootExpr.NeedsFullTrace() {
+		weight++
```
ruslan-mikhailov, line 135: This feels like we need to increase weight even more?

joe-elliott: i forgot about `NeedsFullTrace()`!! nice find.

i don't have a good answer for your question, but yes queries like `{} >> {}` are generally more expensive than `{ foo || bar }`. if `NeedsFullTrace()` is true is `AllConditions` is always false? we may already be incrementing by 2. `{ span.foo && span.bar } | count() > 10` may be an example of a query that has `AllConditions` true and `NeedsFullTrace()` true, but maybe this is ok and it only gets a +1?

you're welcome to:
- make the weight additions configurable
- do some basic testing to determine if you should go to `weight += 2`
- go with your instincts and choose +1 or +2
- spin a wheel to decide

ruslan-mikhailov:
\+ rebase from main to resolve conflicts in Changelog

modules/frontend/pipeline/async_weight_middleware.go
ruslan-mikhailov: I did some e2e load test, couldn't prove +2 gives better results (though, results had large variation). I can't convince myself it's worth to add +2, I leave +1.

Thanks!

ruslan-mikhailov:
I ran internal e2e load tests, no degradation noticed

PR #5785, opened 2025-10-22
[DOC] Tune traceQL queries draft doc

knylander-grafana:
**What this PR does**:
Adds a Tune TraceQL queries doc. 

Fixes https://github.com/grafana/tempo-squad/issues/946
**Checklist**
- [ ] Tests updated
- [X] Documentation added

### DIFF
--- a/docs/sources/tempo/traceql/tune-traceql-queries.md
+++ b/docs/sources/tempo/traceql/tune-traceql-queries.md
@@ -60,7 +60,7 @@ The following query scopes the attribute to the span and runs faster than an equ
 Avoid forms without a scope that force extra lookups:
 
 ```traceql
-{ http.status_code = 500 }
+{ .http.status_code = 500 }
 ```
 
 ## Access as few attributes as possible


docs/sources/tempo/traceql/tune-traceql-queries.md
```diff
@@ -0,0 +1,128 @@
...
+The following query scopes the attribute to the span and runs faster than an equivalent without a scope:
+
+```traceql
+{ span.http.status_code = 500 }
+```
+
+Avoid forms without a scope that force extra lookups:
+
+```traceql
+{ http.status_code = 500 }
```
joe-elliott:
```suggestion
{ .http.status_code = 500 }
```

docs/sources/tempo/traceql/tune-traceql-queries.md
```diff
@@ -0,0 +1,128 @@
...
+Also avoid splitting these conditions across multiple selectors joined by `&&`, which matches them on different spans and changes semantics:
+
+```traceql
+{ span.http.url = "/api" } && { span.http.status_code != 200 }
+```
+
+Use a single selector when you want both conditions to be true on the same span.
+
+```traceql
+{ span.http.url = "/api" && span.http.status_code != 200 }
```
joe-elliott, line 45: we should add an example where we collapse `span.foo = "a" || span.foo = "b"` into a regex

knylander-grafana: We're going to fix this, so it will be automatically better.

ayah-el, line 45: This is the same query as the one in line 33. Instead of using it twice, maybe move the description up higher to when the query was first use and remove the second occurrence of the query.

PR #5786, opened 2025-10-23
frontend, distributor: Add tenant ID validation

carles-grafana:
Use the same validation than in Mimir for consistency. In addition, reject empty IDs ("").

See https://grafana.com/docs/mimir/latest/configure/about-tenant-ids/

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/pipeline/async_tenant_validator_middleware.go
+++ b/modules/frontend/pipeline/async_tenant_validator_middleware.go
@@ -9,6 +9,9 @@ type tenantValidatorRoundTripper struct {
 	next AsyncRoundTripper[combiner.PipelineResponse]
 }
 
+// NewTenantValidatorMiddleware returns a middleware that validates the tenant ID in requests.
+// It assumes there's only one tenant ID per request, otherwise it returns an error.
+// For this reason it must run after the multiTenantMiddleware.
 func NewTenantValidatorMiddleware() AsyncMiddleware[combiner.PipelineResponse] {
 	return AsyncMiddlewareFunc[combiner.PipelineResponse](func(next AsyncRoundTripper[combiner.PipelineResponse]) AsyncRoundTripper[combiner.PipelineResponse] {
 		return &tenantValidatorRoundTripper{next: next}


pkg/validation/tenantid.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"github.com/grafana/dskit/tenant"
+	"github.com/grafana/dskit/user"
+)
+
+func TempoValidTenantID(tenantID string) error {
+	if tenantID == "" {
+		return user.ErrNoOrgID
+	}
+	if err := tenant.ValidTenantID(tenantID); err != nil {
```
javiermolinar: Why not return directly?

```go
if tenantID == "" {
	return user.ErrNoOrgID
}
return tenant.ValidTenantID(tenantID)

```

modules/frontend/frontend.go
```diff
@@ -132,13 +132,15 @@ func New(cfg Config, next pipeline.RoundTripper, o overrides.Interface, reader t
...
 	headerStripWare := pipeline.NewStripHeadersWare(cfg.AllowedHeaders)
+	tenantValidatorWare := pipeline.NewTenantValidatorMiddleware()
 
 	tracePipeline := pipeline.Build(
 		[]pipeline.AsyncMiddleware[combiner.PipelineResponse]{
 			headerStripWare,
 			urlDenyListWare,
 			pipeline.NewWeightRequestWare(pipeline.TraceByID, cfg.Weights),
 			multiTenantMiddleware(cfg, logger),
+			tenantValidatorWare,
```
javiermolinar, line 149: We should place it first. The order of execution matter. We can save some time if the middleware is placed at the begining of the chain

javiermolinar (commented):
Just a couple of comments

modules/frontend/frontend.go
carles-grafana: it has to go after multiTenantMiddleware because then we can ensure there's only 1 tenant per request. Otherwise, we would have to check if there's one or more tenants in each validation.

pkg/validation/tenantid.go
carles-grafana: updated

modules/frontend/frontend.go
javiermolinar: Them maybe it makes sense to include it in the multiTenantMiddleware itself?

This code does some validation from the tenant MultiResolver

https://github.com/grafana/tempo/blob/a3e0bf459a41ebc927c4774b59b7ddc5f8e26f2e/modules/frontend/pipeline/async_handler_multitenant.go#L37

carles-grafana:
> Have we checked if this tenant validation should be done in the queriers too? It is possible to query them directly

I've decided to do it at the edge for simplicity. Queriers shouldn't be exposed, right?

javiermolinar:
> > Have we checked if this tenant validation should be done in the queriers too? It is possible to query them directly
> 
> I've decided to do it at the edge for simplicity. Queriers shouldn't be exposed, right?

Is up to the administrator

carles-grafana:
> > > Have we checked if this tenant validation should be done in the queriers too? It is possible to query them directly
> > 
> > 
> > I've decided to do it at the edge for simplicity. Queriers shouldn't be exposed, right?
> 
> Is up to the administrator

We document this endpoint is only for debugging:

> This API isn’t meant to be used directly unless for debugging the sharding functionality of the query frontend.

https://grafana.com/docs/tempo/latest/api_docs/#query

carles-grafana:
I've added another commit to verify it everywhere where we extract it except the frontend handlers

We can't do it in frontend handlers because we need to wait for the multitenant middleware,
but we rely on the validation middleware.

Validating everywhere follows the same pattern as Mimir and ensures we don't miss any edge case.

modules/frontend/frontend.go
carles-grafana: I think they should be separate because have different responsibilities. The multitenant middleware uses this function mostly to split the tenants, the validation is an extra. And not all endpoints have this middleware.

javiermolinar (commented):
Taking a second look at this, I think it would be much more simple if we move the logic to the authenticate user middleware. This middleware is only used when the multitenant is enabled and is used in most of the endpoints like query frontend and queriers. That way it is centralized and the amount of changes minimal


https://github.com/grafana/tempo/blob/c9343050df94418ff3497b0c466e9be79e3bc744/cmd/tempo/app/app.go#L161


We could either copy the code from dskit or modifiying the upstream code

```go
var AuthenticateUser = Func(func(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		_, ctx, err := user.ExtractOrgIDFromHTTPRequest(r)
		if err != nil {
			http.Error(w, err.Error(), http.StatusUnauthorized)
			return
		}
		tenants, err := tenant.TenantIDs(ctx)
		if err != nil {
			http.Error(w, err.Error(), http.StatusUnauthorized)
			return
		}
		if slices.Contains(tenants, "") {
			http.Error(w, "tenant cannot be empty", http.StatusUnauthorized)
			return
		}
		next.ServeHTTP(w, r.WithContext(ctx))
	})
})

```

carles-grafana:
The problem of HTTPAuthMiddleware is that it only affects HTTP requests, not GRPC.

So I'm happy with the latest solution: add a middleware for explicit validation in the frontend (and in PushTraces), and also change most ExtractOrgID calls so they also do the validation to ensure we don't miss any path.

@joe-elliott  thoughts?

joe-elliott:
I'm good on whatever direction you all agree on as long as its well tested. I will point out that grpc has a similar concept with "interceptors" and so you could apply a solution to gRPC similar to the HTTPAuthMiddleware

https://github.com/grafana/tempo/blob/main/cmd/tempo/app/modules.go#L128-L129
https://github.com/grafana/tempo/blob/main/cmd/tempo/app/app.go#L145-L160

carles-grafana:
@javiermolinar Are you OK with my latest solution?

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## main / unreleased
 
+* [CHANGE] **BREAKING CHANGE** Validate tenant ID in frontend and distributor [#5786](https://github.com/grafana/tempo/pull/5786) (@carles-grafana)
```
electron0zero, line 4: can you mention the chars that are allowed and makes a valid tenant ID, and maybe we should add a docs page for it, similar to https://grafana.com/docs/mimir/latest/configure/about-tenant-ids/

electron0zero: @knylander-grafana we need docs for this, this page can be similar to what mimir docs I liked above.

carles-grafana: yes please, the docs can be identical, with the explicit addition that the tenant ID should not be empty.

modules/frontend/frontend.go
electron0zero: nit: can we add a comment in here to say that this middleware must be after multiTenantMiddleware with a line on why?

carles-grafana: added

PR #5788, opened 2025-10-23
[Metrics Generator] Add support for entity limiting

Logiraptor:
This PR adds a new type of limiter to the metrics generator. It is
designed to limit on "entities" rather than series. In this case an
entity is a single labelset across multiple metrics, excluding external
labels. In effect, it allows the limiter to always generate the full set
of data for a given entity, rather than limiting randomly once the
series limit is triggered.


modules/generator/localentitylimiter/local_entity_limiter.go
```diff
@@ -0,0 +1,105 @@
...
+func NewLocalEntityLimiter(maxEntityFunc func(tenant string) uint32, staleDuration time.Duration) *LocalEntityLimiter {
+	return &LocalEntityLimiter{
+		maxEntityFunc:     maxEntityFunc,
+		entityLastUpdated: make(map[uint64]time.Time),
+		staleDuration:     staleDuration,
+		demand:            cardinality.NewCardinality(staleDuration, removeStaleSeriesInterval),
+	}
+}
+
+func (l *LocalEntityLimiter) TrackEntities(_ context.Context, tenant string, hashes []uint64) (rejected []uint64, err error) {
```
joe-elliott: taking a context b/c you expect a future implementation to make a network call?

modules/generator/registry/registry.go
```diff
@@ -208,20 +240,48 @@ func (r *ManagedRegistry) onAddMetricSeries(count uint32) bool {
...
 
+func (r *ManagedRegistry) onRemoveEntity(count uint32) {
+	r.activeSeries.Sub(count)
 	r.metricTotalSeriesRemoved.Add(float64(count))
 }
 
 func (r *ManagedRegistry) CollectMetrics(ctx context.Context) {
+	r.updatedSinceLastCollectionMtx.RLock()
+	hashes := maps.Clone(r.updatedSinceLastCollection)
+	r.updatedSinceLastCollection = make(map[uint64]struct{})
```
joe-elliott: clear(r.updatedSinceLastCollection) is likely more efficient

modules/generator/registry/registry.go
```diff
@@ -208,20 +240,48 @@ func (r *ManagedRegistry) onAddMetricSeries(count uint32) bool {
...
+}
 
+func (r *ManagedRegistry) onRemoveEntity(count uint32) {
+	r.activeSeries.Sub(count)
 	r.metricTotalSeriesRemoved.Add(float64(count))
 }
 
 func (r *ManagedRegistry) CollectMetrics(ctx context.Context) {
+	r.updatedSinceLastCollectionMtx.RLock()
+	hashes := maps.Clone(r.updatedSinceLastCollection)
```
joe-elliott: clone is concerning. do we expect CollectMetrics to block? previously we simply locked the entire func which would be more efficient

modules/generator/localentitylimiter/local_entity_limiter.go
```diff
@@ -0,0 +1,105 @@
...
+
+		_, exists := l.entityLastUpdated[hash]
+
+		if exists {
+			l.entityLastUpdated[hash] = now
+			continue
+		}
+
+		if shouldLimit && uint32(len(l.entityLastUpdated)) >= maxEntities {
+			rejected = append(rejected, hash)
```
joe-elliott: might get costly in allocs. i reviewed the below code where `.TrackEntities` is called and i'm not sure why the slice of rejected is necessary. can you clarify?

modules/generator/registry/registry.go
```diff
@@ -208,20 +240,48 @@ func (r *ManagedRegistry) onAddMetricSeries(count uint32) bool {
...
+	r.updatedSinceLastCollection = make(map[uint64]struct{})
+	r.updatedSinceLastCollectionMtx.RUnlock()
+
 	r.metricsMtx.RLock()
 	defer r.metricsMtx.RUnlock()
 
+	rejected, err := r.entityLimiter.TrackEntities(ctx, r.tenant, slices.Collect(maps.Keys(hashes)))
```
joe-elliott: pulling the hashes out of the map keys into a slice seems pretty inefficient. i think using the iterators introduced in go 1.23 we can pass an iterator into TrackEntities to range over with 0 allocs.

modules/generator/registry/registry.go
```diff
@@ -208,20 +240,48 @@ func (r *ManagedRegistry) onAddMetricSeries(count uint32) bool {
...
+	if err != nil {
+		r.limitLogger.Log("msg", "tracking entities failed", "err", err)
+	}
+
+	if len(rejected) > 0 {
+		r.limitLogger.Log("msg", "max active entities reached", "rejected", len(rejected), "updated_since_last_collection", len(hashes))
+	}
+
+	rejectedMap := make(map[uint64]struct{})
```
joe-elliott: i'm still digging on what this deleteFunc does but why not have TrackEntities return a `map[uint64]struct{}` to save allocs?

modules/generator/registry/registry.go
```diff
@@ -208,20 +240,48 @@ func (r *ManagedRegistry) onAddMetricSeries(count uint32) bool {
...
+	for _, hash := range rejected {
+		rejectedMap[hash] = struct{}{}
+	}
+
 	var activeSeries int
 	var seriesDemand int
 
 	for _, m := range r.metrics {
+		m.deleteFunc(func(hash uint64, _ int64) bool {
```
joe-elliott: ok, I think I'm seeing. this is another way to cleanup metrics. we still have the time based cleanup below, but this is how you are enforcing the "max active entities" concept. on each collection loop you are removing all series whose "entity" is not in the kept group.

modules/generator/localentitylimiter/local_entity_limiter.go
```diff
@@ -0,0 +1,105 @@
...
+		maxEntityFunc:     maxEntityFunc,
+		entityLastUpdated: make(map[uint64]time.Time),
+		staleDuration:     staleDuration,
+		demand:            cardinality.NewCardinality(staleDuration, removeStaleSeriesInterval),
+	}
+}
+
+func (l *LocalEntityLimiter) TrackEntities(_ context.Context, tenant string, hashes []uint64) (rejected []uint64, err error) {
+	maxEntities := l.maxEntityFunc(tenant)
+	shouldLimit := maxEntities != 0
```
joe-elliott: if shouldLimit is false why are we doing any of this?

modules/generator/localentitylimiter/local_entity_limiter.go
Logiraptor: Correct, I have experimented with using mimir's usage-tracker module as the limiter, which requires a grpc call with the same format. This is the answer for several of your questions questions:

* returning a slice from trackEntities because the grpc call will allocate a slice anyway
* taking a context here
* extending onAddMaxSeries would involve grpc on series creation, which is much slower than batching the hashes until collection

modules/generator/registry/registry.go
Logiraptor: 3aac0d562674f0e756e1277d46e5b7acf960ba8b

modules/generator/registry/registry.go
Logiraptor: 3aac0d562674f0e756e1277d46e5b7acf960ba8b

modules/generator/localentitylimiter/local_entity_limiter.go
Logiraptor: The reason is so we can still track metrics. This means it can be deployed without limiting, and then we can check metrics to determine what a good initial limit would be.

modules/generator/registry/registry.go
Logiraptor: That's right, it could also be another function, but this felt more right to me. It's the same interface as `slices.DeleteFunc` and is the pattern we use in adaptive metrics' series tracking to manage garbage collection.

Logiraptor: One alternative would be to keep `removeStaleSeries` and instead add another `delete(hash)` method. WDYT?

modules/generator/registry/registry.go
Logiraptor: I'm happy to do that way too. In practice the grpc call I am thinking off took around 20ms p99, so on a collection interval of 15s for example, it should still be fine.

modules/generator/registry/registry.go
Logiraptor: This would also remove the need for a `rejectedMap`, so the seq.Iter[uint64] can truly be an iterator

Logiraptor: I went ahead and implemented that, so there are two separate methods, each with a clear purpose. It allows getting rid of the rejectedMap and rejected slice now when local limiting is happening.

Logiraptor:
@joe-elliott Based on our conversations in Grafana Labs internal slack, I've made the following changes:

* Simplified the Limiter interface to the minimum needed for _this PR_ only
* Extracted series limiting into an implementation of Limiter
* Added a configuration option to pick which limiter applies

I opted to keep the demand metrics for both limiters always on, this way it's possible to estimate what a good entity limit would be when migrating. I've also moved the series limiter metrics into local_series_limiter.go and added equivalents for entity limiting, so it should all feel familiar for operators when switching.

modules/generator/instance.go
```diff
@@ -109,12 +112,21 @@ type instance struct {
 func newInstance(cfg *Config, instanceID string, overrides metricsGeneratorOverrides, wal storage.Storage, logger log.Logger, traceWAL, rf1TraceWAL *wal.WAL, writer tempodb.Writer) (*instance, error) {
 	logger = log.With(logger, "tenant", instanceID)
 
+	limitLogger := tempo_log.NewRateLimitedLogger(1, level.Warn(logger))
+	var limiter registry.Limiter
+	switch cfg.LimiterType {
+	case LimiterTypeSeries:
+		limiter = localserieslimiter.New(overrides.MetricsGeneratorMaxActiveSeries, instanceID, limitLogger)
+	case LimiterTypeEntity:
+		limiter = localentitylimiter.New(overrides.MetricsGeneratorMaxActiveEntities, instanceID, limitLogger)
```
joe-elliott, line 112: default case that returns an error?

modules/generator/localentitylimiter/local_entity_limiter.go
```diff
@@ -0,0 +1,125 @@
...
+			Help:      "The total amount of entities removed after they have become stale per tenant",
+		}, []string{"tenant"}),
+	}
+}
+
+var metrics = newMetrics(prometheus.DefaultRegisterer)
+
+type LocalEntityLimiter struct {
+	tenant            string
+	entityLastUpdated map[uint64]uint32
```
joe-elliott: entityLastUpdated -> entityActiveSeries? or something similar?

modules/generator/localentitylimiter/local_entity_limiter.go
```diff
@@ -0,0 +1,125 @@
...
+	if ok {
+		activeSeries += seriesCount
+		l.entityLastUpdated[labelHash] = activeSeries
+		return true
+	}
+
+	maxEntities := l.maxEntityFunc(l.tenant)
+	if maxEntities != 0 && uint32(len(l.entityLastUpdated))+1 > maxEntities {
+		l.limitLogger.Log("msg", "reached max active entities", "active_entities", len(l.entityLastUpdated), "max_active_entities", maxEntities)
+		l.metricTotalEntitiesLimited.Add(float64(1))
```
joe-elliott, line 97: we could track series limited as well? i don't know if that's important to the way you want to operate this. up to you!

modules/generator/localentitylimiter/local_entity_limiter.go
```diff
@@ -0,0 +1,125 @@
...
+		activeSeries -= seriesCount
+		if activeSeries <= 0 {
+			delete(l.entityLastUpdated, labelHash)
+		} else {
+			l.entityLastUpdated[labelHash] = activeSeries
+		}
+	}
+
+	l.metricActiveEntities.Set(float64(len(l.entityLastUpdated)))
+	l.metricTotalEntitiesRemoved.Add(float64(seriesCount))
```
joe-elliott: .Add(1) ?

modules/generator/localentitylimiter/local_entity_limiter.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func (l *LocalEntityLimiter) OnDelete(labelHash uint64, seriesCount uint32) {
+	l.mtx.Lock()
+	defer l.mtx.Unlock()
+
+	activeSeries, ok := l.entityLastUpdated[labelHash]
+	if ok {
+		activeSeries -= seriesCount
+		if activeSeries <= 0 {
```
joe-elliott: activeSeries is a uint. compare to 0 only? we could also store series count as an int in case there are edge cases where activeSeries < 0.

modules/generator/localentitylimiter/local_entity_limiter.go
```diff
@@ -0,0 +1,125 @@
...
+	l.metricTotalEntitiesAdded.Add(float64(1))
+	return true
+}
+
+func (l *LocalEntityLimiter) OnDelete(labelHash uint64, seriesCount uint32) {
+	l.mtx.Lock()
+	defer l.mtx.Unlock()
+
+	activeSeries, ok := l.entityLastUpdated[labelHash]
+	if ok {
```
joe-elliott: should we do:
```
if !ok {
  return
}
```

the way this is currently written we will increment 	`l.metricTotalEntitiesRemoved` below if the hash is not found in the map. not sure what that means? may never happen?

i suppose we just need to figure out how we want to handle unexpected input in this function. similar style comment on line 116.

not super concerned either way. i think the only real fix is the increment on line 124

joe-elliott (commented):
added some thoughts. i really like the reorg with the limiter interface and i think this will be np to merge. can we get a changelog entry?

also it might be nice to add a few lines to a doc somewhere that this mode exists and why. unfortunately i couldn't find a good place to add it. maybe just note it in the [metrics generator config](https://grafana.com/docs/tempo/v2.5.x/configuration/#metrics-generator)? a full explanation of the feature and all the details is likely a bit much (unless you are super excited to add all that), but just documenting its existence might be nice to see if we get community adoption.

modules/generator/localentitylimiter/local_entity_limiter.go
Logiraptor: In practice I think these metrics aren't super helpful other than asking "are they > 0". The reason is because any entity which is rejected will likely retry very soon, so there's no way to tell from the rate of rejections how many entities are actually in the demand. For that we need to rely on the HLL sketch.

Logiraptor:
@joe-elliott I believe I've implemented all your feedback. Regarding docs, I think for now the config docs are sufficient. Once I get some experience with this in practice, and also work on a few other improvements, it might be worth adding / updating a docs page specifically for managing span metrics cardinality, we'll see.

joe-elliott (approved):
Loki Grafana Tempo Mimir

looks like some minor conflicts to cleanup then we can merge?

Logiraptor:
@joe-elliott done! thanks

PR #5789, opened 2025-10-23
feat(s3): Support S3 SSE-C encryption

### LINKED ISSUE #5761: Tempo S3 with SSE-C
Currently, the only valid S3 SSE options in Tempo are SSE-KMS and SSE-S3.
These options are not available in my cloud offering (Scaleway) at this time.

I'd love if support for SSE-C was added into Tempo.

I've seen some folks are working on it for Loki: 
- https://github.com/grafana/loki/issues/9501
- https://github.com/grafana/loki/pull/19393

Some other for Mimir:
- https://github.com/grafana/mimir/issues/3381
- https://github.com/grafana/mimir/pull/3383

SSE-S3, SSE-KMS, Encryption at rest: not available at my clouder (Scaleway).

steffsas:
**What this PR does**:
Adds support for SSE-C (server-side encryption with customer-provided keys) alongside existing SSE-S3 and SSE-KMS implementations. The `SSE` struct is now propagated to all relevant MinIO S3 operations.

I have verified the SSE-C compatibility with a docker compose example, similar to those in `example/docker-compose`. If desired, I can add this example as well.

Unit tests are not included. Adding them would require mocking the `minio.Client` by introducing an interface layer to capture SSE parameters in calls (e.g. `minio.PutObjectOptions`). See https://github.com/grafana/tempo/blob/main/tempodb/backend/s3/s3.go#L42-L43. If desired, I can implement them either in another PR or in this one.

Fixes #5761

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/tempodb/backend/s3/s3.go
+++ b/tempodb/backend/s3/s3.go
@@ -171,6 +171,12 @@ func getObjectOptions(rw *readerWriter) minio.GetObjectOptions {
 	}
 }
 
+func getPutObjectPartOptions(rw *readerWriter) minio.PutObjectPartOptions {
+	return minio.PutObjectPartOptions{
+		SSE: rw.sse,
+	}
+}
+
 // Write implements backend.Writer
 func (rw *readerWriter) Write(ctx context.Context, name string, keypath backend.KeyPath, data io.Reader, size int64, _ *backend.CacheInfo) error {
 	derivedCtx, span := tracer.Start(ctx, "s3.Write")
@@ -211,7 +217,7 @@ func (rw *readerWriter) Append(ctx context.Context, name string, keypath backend
 	keypath = backend.KeyPathWithPrefix(keypath, rw.cfg.Prefix)
 	objectName := backend.ObjectFileName(keypath, name)
 
-	options := getPutObjectOptions(rw)
+	putObjOptions := getPutObjectOptions(rw)
 	if tracker != nil {
 		a = tracker.(appendTracker)
 	} else {
@@ -219,7 +225,7 @@ func (rw *readerWriter) Append(ctx context.Context, name string, keypath backend
 			ctx,
 			rw.cfg.Bucket,
 			objectName,
-			options,
+			putObjOptions,
 		)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


tempodb/backend/s3/s3.go
```diff
@@ -239,7 +239,9 @@ func (rw *readerWriter) Append(ctx context.Context, name string, keypath backend
 		a.partNum,
 		bytes.NewReader(buffer),
 		int64(len(buffer)),
-		minio.PutObjectPartOptions{},
+		minio.PutObjectPartOptions{
```
javiermolinar: Can we extract this as a new method in the same way we have the getPutObjectOptions?

tempodb/backend/s3/s3.go
```diff
@@ -744,6 +748,14 @@ func buildSSEConfig(cfg *Config) (encrypt.ServerSide, error) {
 		}
 	case SSES3:
 		return encrypt.NewSSE(), nil
+	case SSEC:
+		if cfg.SSE.CustomerEncryptionKey == "" {
+			return nil, errors.New("SSE-C EncryptionKey is missing")
+		}
+		if len(cfg.SSE.CustomerEncryptionKey) != 32 {
+			return nil, errors.New("SSE-C EncryptionKey must be 32 bytes long")
```
javiermolinar: Minio already validates the key. Maybe we can rely on that

https://github.com/minio/minio-go/blob/785b638cc4afc979cb92a1358f172d05d32a18bf/pkg/encrypt/server-side.go#L115

javiermolinar (commented):
Two nitpicks, I think the manifest has to be updated

tempodb/backend/s3/s3.go
```diff
@@ -270,7 +277,9 @@ func (rw *readerWriter) CloseAppend(ctx context.Context, tracker backend.AppendT
 		a.objectName,
 		a.uploadID,
 		completeParts,
-		minio.PutObjectOptions{},
+		minio.PutObjectOptions{
+			ServerSideEncryption: rw.sse,
```
steffsas, line 281: I am unsure whether we can use `getPutObjectOptions` here as well as it additionally adds headers which may not be needed in this case.

javiermolinar: I don't know either, let's keep it like that for now

steffsas: fine for me, fixed merge conflicts in changelog btw

PR #5815, opened 2025-10-28
feat: expose new owned partition metric

javiermolinar:
**What this PR does**:
This PR introduces a new metric to expose partition ownership information from live-store instances.
The metric allows us to accurately determine which consumer (in which zone) currently owns each Kafka partition, enabling better visibility and alerting for ingestion coverage issues.

This addition is required to create an actionable and stable alert for ingestion coverage, replacing existing heuristics that rely on indirect or noisy signals (e.g., replica counts, lag series presence)

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 * [ENHANCEMENT] Add metric for tracking added latency to write requests [#5781](https://github.com/grafana/tempo/pull/5781) (@mapno)
 * [ENHANCEMENT] Improve error message when overrides fail to parse [#5787](https://github.com/grafana/tempo/pull/5787) (@mapno)
 * [ENHANCEMENT] Increase weight for heavy TraceQL queries [#5782](https://github.com/grafana/tempo/pull/5782) (@ruslan-mikhailov)
+* [ENHANCEMENT] Add TraceQL support for nonexistential attribute queries [#4950](https://github.com/grafana/tempo/pull/4905) (@ie-pham)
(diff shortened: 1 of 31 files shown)


modules/livestore/partition_reader.go
```diff
@@ -159,6 +161,25 @@ func (r *PartitionReader) commitLoop(ctx context.Context) {
...
+	r.metrics.ownedPartition.WithLabelValues(partition, r.consumerGroup).Set(1)
+
+	for {
+		select {
+		case <-ctx.Done():
+			r.metrics.ownedPartition.WithLabelValues(partition, r.consumerGroup).Set(0)
+			return
+		case <-t.C:
+			r.metrics.ownedPartition.WithLabelValues(partition, r.consumerGroup).Set(1)
+		}
```
mapno: Does it need to be set on a ticker? Once it's initially set it should be in the registry until the live-store shuts down

javiermolinar: Right setting it once is enough to keep it alive, but the ticker is useful as keep-alive. I was thinking that we could also set it in the commitLoop but I find it more explicit this way

mapno: That's my question I guess: why do we need this keepalive mechanism? Aren't series kept in the Prometheus registry throughout its lifecycle?

javiermolinar: Just as a precaution but you are right, we can test it like you are suggestion first

PR #5817, opened 2025-10-28
Fix nil and filter processor

mattdurham:


modules/querier/http.go
```diff
@@ -442,6 +443,16 @@ func handleError(w http.ResponseWriter, err error) {
 }
 
 func writeFormattedContentForRequest(w http.ResponseWriter, req *http.Request, m proto.Message, span oteltrace.Span) {
+	// Check for both explicit nil and typed nil pointers (e.g., (*tempopb.SearchResponse)(nil))
+	// A typed nil pointer is not equal to nil when passed as an interface, so we need reflection
+	if m == nil || (reflect.ValueOf(m).Kind() == reflect.Ptr && reflect.ValueOf(m).IsNil()) {
```
joe-elliott, line 448: wut

modules/distributor/forwarder/ottl_ismatch_test.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"context"
+	"testing"
+
+	dslog "github.com/grafana/dskit/log"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+)
+
+func TestFilterForwarder_IsMatch(t *testing.T) {
```
joe-elliott, line 12: nice fix and test 🙏

modules/querier/http.go
mattdurham, line 448: This is most assuredly due to a bug in livestore

mattdurham: Nil and interface checking is hard because an interface is a type and value so you have to check both of them. Regardless this was a bandaid fix and I think I fixed the root cause in livestore.

mattdurham: Ugh not resolved, but this seems to fix the issue

modules/livestore/live_store.go
```diff
@@ -580,23 +580,23 @@ func (s *LiveStore) GetMetrics(_ context.Context, _ *tempopb.SpanMetricsRequest)
...
 // withInstance extracts the tenant ID from the context, gets the instance,
 // and calls the provided function if the instance exists. If the instance
-// doesn't exist, it returns a new zero-valued instance of T.
-func withInstance[T any](ctx context.Context, s *LiveStore, fn func(*instance) (*T, error)) (*T, error) {
+// doesn't exist, it returns the provided default value.
+func withInstance[T any](ctx context.Context, s *LiveStore, defaultValue *T, fn func(*instance) (*T, error)) (*T, error) {
 	instanceID, err := user.ExtractOrgID(ctx)
 	if err != nil {
-		return nil, err
+		return defaultValue, err
```
mapno:
```suggestion
		var defaultValue T
		return &defaultValue, err
```

Why not something like this? That way there is no need to pass a default value to the function.

mapno (approved):
Nice fix!

modules/livestore/live_store.go
mattdurham: I originally had something similar to that but was a bit worried it was somehow causing a nil to come back.

mattdurham: Though I dont think it ended up being the root cause since it only stopped with the very explicit nil check, so changing back.

PR #5819, opened 2025-10-29
fix: deduplicate prom labels in metric generator registry again

javiermolinar:
**What this PR does**:
Label deduplication was removed on a previous commit when updating Prometheus:
https://github.com/grafana/tempo/commit/3c9b1386ee7448998a0a11681bca800db41988d0

This PR bring back the deduplication by not initializing the builder with duplicated labels, instead we initialized empty and use the Set method to append the pairs. This method internally overwrite existing values

Fixes #<issue number>

**Checklist**
- [X] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/registry/util.go
+++ b/modules/generator/registry/util.go
@@ -5,8 +5,10 @@ import "github.com/prometheus/prometheus/model/labels"
 // newSeriesLabelsBuilder creates a labels builder with user labels and external labels pre-populated.
 func newSeriesLabelsBuilder(labelValueCombo *LabelValueCombo, externalLabels map[string]string) *labels.Builder {
 	builder := labels.NewBuilder(labels.Labels{})
-	for i := range labelValueCombo.labels.names {
-		builder.Set(labelValueCombo.labels.names[i], labelValueCombo.labels.values[i])
+	if labelValueCombo != nil {
+		for i := range labelValueCombo.labels.names {
+			builder.Set(labelValueCombo.labels.names[i], labelValueCombo.labels.values[i])
+		}
 	}
 	for name, value := range externalLabels {
 		builder.Set(name, value)


modules/generator/registry/util.go
```diff
@@ -2,11 +2,23 @@ package registry
```
carles-grafana, line 2: I was thinking something like this to log duplicates:
```
package registry

import (
	"github.com/go-kit/log/level"
	"github.com/grafana/tempo/pkg/util/log"
	"github.com/prometheus/prometheus/model/labels"
)

// newSeriesLabelsBuilder creates a labels builder with user labels and external labels pre-populated.
func newSeriesLabelsBuilder(labelValueCombo *LabelValueCombo, externalLabels map[string]string) *labels.Builder {
	builder := labels.NewBuilder(labels.Labels{})
	uniqueLabelNames := make(map[string]struct{})

	if labelValueCombo != nil {
		for i := range labelValueCombo.labels.names {
			if _, exists := uniqueLabelNames[labelValueCombo.labels.names[i]]; exists {
				_ = level.Warn(log.Logger).Log(
					"msg", "duplicate label name detected while building metric series labels",
					"label", labelValueCombo.labels.names[i],
				)
			} else {
				uniqueLabelNames[labelValueCombo.labels.names[i]] = struct{}{}
			}
			builder.Set(labelValueCombo.labels.names[i], labelValueCombo.labels.values[i])
		}
	}
	for name, value := range externalLabels {
		if _, exists := uniqueLabelNames[name]; exists {
			_ = level.Warn(log.Logger).Log(
				"msg", "duplicate label name detected while building metric series labels",
				"label", name,
			)
		} else {
			uniqueLabelNames[name] = struct{}{}
		}
		builder.Set(name, value)
	}
	return builder
}
```

javiermolinar: Yes, better to do it on another pr, so I'll create a new feature for that

PR #5827, opened 2025-10-29
Rewrite conditions on same attribute to equivalent array operation

stoewer:
**What this PR does**:
Rewrite queries like `{ .a="a" || .a="b" }` into the equivalent, faster form `{ .a=["a", "b"] }`.

This AST optimization is applied by default, but it can be disabled using the query hint `skip_optimization=true`.

Initially, I added a dedicated `IN` / `NOT IN` operator (see [commit](https://github.com/grafana/tempo/pull/5827/commits/12667b0640e2fce7e7416be83127501c6bdfbbd7)), but later realized that the existing `IN` / `NOT IN` semantics for the operators `=`, `!=`, `=~`, and `!~` with array operands already provided nearly all the functionality needed for this feature.

**Limitations**
* Does not optimize conditions with types that don't have directly corresponding array type such as: TypeKind, TypeStatus, TypeDuration
* Does not optimize conditions not adjacent to each other like `{ .a="a" && .b="b" || .a="c" }`
* The syntax of the array form e.g. `{ .a=["a", "b"] }` only works internally, the syntax is not actually available in TraceQL

**Breaking Changes**
Implementing this feature required a few minor breaking changes to operator semantics:

* `!=` now means NOT IN (previously: CONTAINS NOT EQUAL)
* `!~` now means MATCH NONE (previously: CONTAINS NON-MATCH)
* Regexes must now appear on the right-hand side e.g. `{"on.*" =~ .a}` is no longer valid
* Regex operands must be of type string or string array (previously, any value convertible to a valid regex was accepted)

While these changes technically break existing TraceQL behavior, the affected behavior was undocumented and limited to edge cases.

Fixes grafana/tempo-squad#874

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 * [FEATURE] Add query_end_cutoff setting to clamp near-“now” queries (default 30s) for consistent Tempo results [#5682](https://github.com/grafana/tempo/pull/5682) (@javiermolinar)
 * [FEATURE] New block encoding vParquet5-preview4 with array support for dedicated columns. This is a preview, breaking changes are expected. [#5760](https://github.com/grafana/tempo/pull/5760) (@stoewer)
 * [FEATURE] Add SSE-C encryption support to S3 backend [#5789](https://github.com/grafana/tempo/pull/5789) (@steffsas)
+* [FEATURE] **BREAKING CHANGE** Optimize TraceQL AST by rewriting conditions on the same attribute to their array equivalent [#5827](https://github.com/grafana/tempo/pull/5827) (@stoewer)
+  Slightly changes the array matching semantics of != and !~ operators and introduces stricter rules for regex literals. 
 * [ENHANCEMENT] docs: Add explicit notes about authentication [#5735](https://github.com/grafana/tempo/pull/5735) (@electron0zero)
 * [ENHANCEMENT] On startup, first record for live store to consume is not older than two complete block timeouts [#5693](https://github.com/grafana/tempo/pull/5693) (@ruslan-mikhailov)
 * [ENHANCEMENT] Add secure connection support to tempo-cli [#5692](https://github.com/grafana/tempo/pull/5692) (@TheoBrigitte)


stoewer:
Local benchmark results of the following queries with and without optimizations:

* spanAttValMatch: `{ span.component="net/http" || span.component="database/sql" }`
* spanAttIntrinsicMatch: `{ name="distributor.ConsumeTraces" || name="grpcutils.Authenticate" }`
* resourceAttValMatch: `{ resource.module.path="github.com/grafana/stack-state-service" || resource.module.path="github.com/grafana/pdc" }`
* resourceAttIntrinsicMatch: `{ resource.service.name="loki-querier" || resource.service.name="loki-gateway" }`


<details><summary>Benchmark opt vs no opt</summary>

```
goos: linux
goarch: amd64
cpu: AMD Ryzen AI 9 HX PRO 370 w/ Radeon 890M       
                                                 │ bench-traceql-vp4-01-no-opt.txt │    bench-traceql-vp4-01-opt.txt     │
                                                 │             sec/op              │    sec/op     vs base               │
BackendBlockTraceQL/spanAttValMatch-12                                756.9m ±  2%   304.0m ±  3%  -59.83% (p=0.000 n=8)
BackendBlockTraceQL/spanAttIntrinsicMatch-12                          711.9m ±  3%   169.5m ±  6%  -76.19% (p=0.000 n=8)
BackendBlockTraceQL/resourceAttValMatch-12                            731.0m ± 10%   412.5m ±  4%  -43.58% (p=0.000 n=8)
BackendBlockTraceQL/resourceAttIntrinsicMatch-12                     1016.5m ±  7%   913.9m ± 10%  -10.09% (p=0.015 n=8)
geomean                                                               795.5m         373.3m        -53.07%

                                                 │ bench-traceql-vp4-01-no-opt.txt │      bench-traceql-vp4-01-opt.txt      │
                                                 │               B/s               │      B/s        vs base                │
BackendBlockTraceQL/spanAttValMatch-12                               129.9Mi ± 13%    319.1Mi ± 12%  +145.72% (p=0.000 n=8)
BackendBlockTraceQL/spanAttIntrinsicMatch-12                         139.8Mi ± 12%    579.5Mi ± 13%  +314.55% (p=0.000 n=8)
BackendBlockTraceQL/resourceAttValMatch-12                           94.41Mi ± 13%   165.66Mi ± 10%   +75.47% (p=0.000 n=8)
BackendBlockTraceQL/resourceAttIntrinsicMatch-12                     96.76Mi ± 10%   107.19Mi ± 13%   +10.78% (p=0.015 n=8)
geomean                                                              113.5Mi          239.4Mi        +110.95%

                                                 │ bench-traceql-vp4-01-no-opt.txt │   bench-traceql-vp4-01-opt.txt   │
                                                 │            MB_io/op             │  MB_io/op   vs base              │
BackendBlockTraceQL/spanAttValMatch-12                                  103.1 ± 0%   101.7 ± 0%  -1.36% (p=0.000 n=8)
BackendBlockTraceQL/spanAttIntrinsicMatch-12                            104.3 ± 0%   103.0 ± 0%  -1.25% (p=0.000 n=8)
BackendBlockTraceQL/resourceAttValMatch-12                              72.37 ± 0%   71.62 ± 0%  -1.04% (p=0.000 n=8)
BackendBlockTraceQL/resourceAttIntrinsicMatch-12                        103.1 ± 0%   102.7 ± 0%  -0.39% (p=0.000 n=8)
geomean                                                                 94.64        93.69       -1.01%

                                                 │ bench-traceql-vp4-01-no-opt.txt │    bench-traceql-vp4-01-opt.txt     │
                                                 │            spans/op             │  spans/op    vs base                │
BackendBlockTraceQL/spanAttValMatch-12                                 19.41k ± 0%   19.41k ± 0%       ~ (p=1.000 n=8) ¹
BackendBlockTraceQL/spanAttIntrinsicMatch-12                           2.258k ± 0%   2.258k ± 0%       ~ (p=1.000 n=8) ¹
BackendBlockTraceQL/resourceAttValMatch-12                             3.323k ± 0%   3.323k ± 0%       ~ (p=1.000 n=8) ¹
BackendBlockTraceQL/resourceAttIntrinsicMatch-12                       11.38k ± 0%   11.38k ± 0%       ~ (p=1.000 n=8) ¹
geomean                                                                6.381k        6.381k       +0.00%
¹ all samples are equal

                                                 │ bench-traceql-vp4-01-no-opt.txt │     bench-traceql-vp4-01-opt.txt     │
                                                 │              B/op               │     B/op       vs base               │
BackendBlockTraceQL/spanAttValMatch-12                               134.8Mi ±  2%    170.5Mi ± 0%  +26.54% (p=0.000 n=8)
BackendBlockTraceQL/spanAttIntrinsicMatch-12                         30.26Mi ± 25%    21.22Mi ± 3%  -29.88% (p=0.000 n=8)
BackendBlockTraceQL/resourceAttValMatch-12                           355.6Mi ±  1%    609.1Mi ± 0%  +71.30% (p=0.000 n=8)
BackendBlockTraceQL/resourceAttIntrinsicMatch-12                     723.1Mi ±  1%   1276.7Mi ± 0%  +76.56% (p=0.000 n=8)
geomean                                                              180.0Mi          230.3Mi       +27.99%

                                                 │ bench-traceql-vp4-01-no-opt.txt │    bench-traceql-vp4-01-opt.txt     │
                                                 │            allocs/op            │  allocs/op    vs base               │
BackendBlockTraceQL/spanAttValMatch-12                                 1.425M ± 0%    1.889M ± 0%  +32.59% (p=0.000 n=8)
BackendBlockTraceQL/spanAttIntrinsicMatch-12                           201.5k ± 4%    222.1k ± 0%  +10.23% (p=0.000 n=8)
BackendBlockTraceQL/resourceAttValMatch-12                             4.426M ± 0%    7.090M ± 0%  +60.20% (p=0.000 n=8)
BackendBlockTraceQL/resourceAttIntrinsicMatch-12                       9.304M ± 0%   15.079M ± 0%  +62.06% (p=0.000 n=8)
geomean                                                                1.854M         2.588M       +39.57%
```

</details>

knylander-grafana:
Will we need to update our TraceQL docs for this?

stoewer:
> Will we need to update our TraceQL docs for this?

Good question. I don't think there is much to document as the TraceQL syntax isn't actually changing - it's all happening under the hood. 

With regards to the breaking changes: the behavior that is changing isn't documented at the moment and it's also such a subtle change that I doubt users will notice it. Anyway, we should add a short note about the breaking changes to the next releasenotes

pkg/parquetquerygen/predicates.go
```diff
@@ -90,6 +90,110 @@ func (p {{ $structName }}) KeepValue(v pq.Value) bool {
 }
 {{- end }}
 {{- end }}
+
+{{- /* Generate In/NotIn for Int, Float, Bool */}}
```
joe-elliott, line 94: predicates gen is a nice perf boost over the generic predicate we had before, but i still hate it. do you think we can do this with go generics?

i don't expect that change in this PR. just wondering.

pkg/traceql/parse.go
```diff
@@ -45,6 +52,12 @@ func Parse(s string) (expr *RootExpr, err error) {
 		return nil, fmt.Errorf("unknown parse error: %d", e)
 	}
 
+	hintSkipOptimization, _ := l.expr.Hints.GetBool(HintSkipOptimization, true)
+	if astOptimization && !hintSkipOptimization {
+		l.expr = ApplyDefaultASTRewrites(l.expr)
+		level.Debug(log.Logger).Log("msg", "optimize AST for TraceQL query", "query", s, "optimizedQuery", l.expr.String(), "optimizationCount", l.expr.OptimizationCount)
```
joe-elliott: this is only once per Q. maybe log as info? no strong feelings

pkg/traceql/ast_rewriter.go
```diff
@@ -0,0 +1,267 @@
...
+		if !slices.Contains(restrictTypes, valLHS.Type) {
+			return fe, 0
+		}
+		if !slices.Contains(restrictTypes, valRHS.Type) {
+			return fe, 0
+		}
+	}
+
+	// create an equivalent array operation
+	array, ok := valLHS.append(valRHS)
```
joe-elliott, line 236: so internally this is rewriting:

```
{ span.foo = 1 || span.foo = 2 }
```

to this which is technically not parseable but works in the way we want:

```
{ span.foo = [1, 2] }
```

which logically evaluates like this when span foo is an integer:
```
{ span.foo IN (1, 2) } 
```

what happens if span.foo is an integer array? does this become array equality?

pkg/traceql/ast_conditions_test.go
```diff
@@ -346,3 +350,44 @@ func TestMetricsAggregate_extractConditions(t *testing.T) {
 		})
 	}
 }
+
+func TestBinaryOperation_extractConditions(t *testing.T) {
```
joe-elliott, line 354: nit: `TestBinaryOperation_extractConditionsWithArrays` or something similar ?

pkg/traceql/ast_execute.go
```diff
@@ -433,24 +433,26 @@ func (o *BinaryOperation) execute(span Span) (Static, error) {
 		case OpLessEqual:
 			return NewStaticBool(strings.Compare(lhsS, rhsS) <= 0), nil
 		case OpRegex:
-			if o.compiledExpression == nil {
-				o.compiledExpression, err = regexp.NewRegexp([]string{rhsS}, true)
+			if len(o.compiledExpressions) == 0 {
+				compiled, err := regexp.NewRegexp([]string{rhsS}, true)
```
joe-elliott: insight into why we made this a slice? it looks like we can still are only adding one compiled expression

pkg/traceql/ast_rewriter.go
```diff
@@ -0,0 +1,267 @@
+package traceql
+
+import "slices"
+
+func ApplyDefaultASTRewrites(r *RootExpr) *RootExpr {
```
joe-elliott, line 5: love this!

if you have time can you move some of our existing rewrites to this logic in a follow up PR:

https://github.com/grafana/tempo/blob/main/pkg/traceql/ast.go#L523-L539

https://github.com/grafana/tempo/blob/main/pkg/traceql/ast.go#L591-L596

pkg/traceql/parse_test.go
```diff
@@ -1566,3 +1566,50 @@ func TestMetricsSecondStageErrors(t *testing.T) {
...
+		},
+		{
+			name:  "query with rewrite OR",
+			query: "{ .attr = `foo` || .attr = `bar` } | rate()",
+			want:  "{ .attr = [`foo`, `bar`] } | rate()",
+		},
+		{
+			name:  "query with rewrite AND",
+			query: "{ .attr != `foo` && .attr != `bar` }",
+			want:  "{ .attr != [`foo`, `bar`] }",
```
joe-elliott, line 1589: does this parse? or does it just come out of `.String()`?

mdisibio:
> { .a=["a", "b"] }

Does this prevent the ability to add full array support in the future, for example if we wanted to find all `.a` where it is that string array?

> don't have directly corresponding array type such as: TypeKind

It would be nice to handle this, and it seems like `kind = [server, client]` would work in same way.   Adding a type of KindArray to Static doesn't seem right, because it can't be expressed from/to real data.  Does this gap mean we should be distinguishing these two concepts?  i.e. a list of statics, versus a static that is a list?

I was kind of envisioning to use this for the IN operator:  https://github.com/grafana/tempo/blob/main/pkg/traceql/storage.go#L13  Each condition can take multiple operands. To compare:
- This pr:  OpEqual, StringArray["a","b"]
- As multiple operands would be:  OpIn, "a", "b"

I would expect the first example to tell the fetch layer to look for that array exactly, right?  And the second example is more clear to look for any of these values.

pkg/traceql/ast_execute.go
```diff
@@ -535,31 +537,67 @@ func (o *BinaryOperation) execute(span Span) (Static, error) {
...
-		err := arraySide.GetElements(func(elem Static) bool {
-			elemOp.LHS = elem
-			res, err = elemOp.execute(span)
+		// operators OpNotEqual and OpNotRegex have the semantics of 'not in' / 'match none': must check all elements
+		matchAll := o.Op == OpNotEqual || o.Op == OpNotRegex
+
+		var elemCount, matchCount int
+		for e := range elem {
+			elemCount++
+			elemOp := BinaryOperation{Op: op, LHS: scalar, RHS: e}
```
joe-elliott, line 568: not necessarily asking for a change here, but i do find this logic strange. we're creating a temporary binary operation which we then execute on the span. this comes with some unfortunate compiledExpression juggling to prevent repeatedly parsing a regex. this condition feels particularly brittle:

```
if len(elemOp.compiledExpressions) == 1 && len(o.compiledExpressions) == elemCount-1 {
```

instead of creating a temporary binary operation would it be possible to extract the appropriate comparison logic into helper functions and call it from above and here? i'm good with the direction you want to go here. you've been deeper in this code more recently.

pkg/traceql/ast_rewriter.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+// rewriteAndToMatchNone is a fieldExpressionRewriteFn that rewrites a single { .a !~ "a" && .b !~ "b" } to { .a !~ ["a", "b"]}
+func rewriteAndToMatchNone(fe FieldExpression) (FieldExpression, int) {
+	return rewriteBinaryOperationsToArrayEquivalent(fe, OpAnd, OpNotRegex, TypeString, TypeStringArray)
+}
+
+// rewriteBinaryOperationsToArrayEquivalent rewrites binary operations to an equivalent array form
+func rewriteBinaryOperationsToArrayEquivalent(fe FieldExpression, opOuter, opInner Operator, restrictTypes ...StaticType) (FieldExpression, int) {
+	opOr, ok := fe.(*BinaryOperation)
```
joe-elliott, line 201: opOr is a strange name here since we don't know what it is. binOp?

joe-elliott (commented):
I love this change. I think we will be doing more of this in the future. Some comments but I'm eager to get this in.

ie-pham:
I've tested this and it works well. I've also tested it with the nil operations and I think the rebase was fine. 

I guess this can be added in the future but I have two scenarios I feel would make this even better

1. mixed with different attributes in between the matching attribute
```
{ ((.a = `val1`) || (.b = `val2`)) || (.a = `val3`) }

to

{ (.a = [`val1`, `val3`] ) || (.b = `val2`) }
```

2. remove repeated values from the array
```
"{ .a = `val1` || .a = `val1` }"

to

"{ .a = `val1`}" 

instead of

"{ .a = `val1` || .a = `val1` }"
```

pkg/traceql/ast_rewriter.go
stoewer: Yes, that was my plan. I'll create an issue for it

pkg/traceql/parse_test.go
stoewer: No, this does not parse, it's just the output of String()

stoewer:
> Does this prevent the ability to add full array support in the future, for example if we wanted to find all .a where it is that string array?

That's a good point. Adding distinct `IN` / `NOT IN` operators keeps more options open for array comparisons with `=`. I'll add them again.

> It would be nice to handle this, and it seems like kind = [server, client] would work in same way. Adding a type of KindArray to Static doesn't seem right, because it can't be expressed from/to real data. Does this gap mean we should be distinguishing these two concepts? i.e. a list of statics, versus a static that is a list?

If it's possible, I'd rather avoid adding/using the concept of a list of statics. While this could work for `traceql.Condition`, it does not really work in other places. For example a `BinaryOperation` with hypothetical `IN` operator:

```go
op := BinaryOperation{
  Op:  OpIn,
  LHS: NewStaticKind(KindClient),
  RHS: ??, // <--- how do we express multiple kinds here?
}
```

pkg/traceql/ast_rewriter.go
stoewer: @mdisibio  raised a similar point that I tried to answer [here](https://github.com/grafana/tempo/pull/5827#issuecomment-3580815220)

pkg/parquetquerygen/predicates.go
stoewer: I think this could work. However, it would still require to pass some kind of function `f(parquet.Value) T` or similar into the predicate in order to convert the parquet.Value to the parameterized type. Logically this works, but it could have unwanted side effects like preventing inlining. But I guess it would be worth trying

PR #5835, opened 2025-10-31
Fix potential hash collisions of traceql StringArray + compare() improvement

mdisibio:
**What this PR does**:
This fixes a potential hash collision of StringArray static MapKey.  Similar to https://github.com/grafana/tempo/pull/5716#discussion_r2417557871 we need to write a separator between values, so that `"foo","bar"` hashes to a different value than `"foobar"`.  Also ports the missing [clone of attribute buffers](https://github.com/grafana/tempo/pull/5522) to vparquet5, which was causing panics that caused me to come across this potential hash collision.

A few other changes:
* Removes unnecessary computations of MapKey in the compare() function (we only need to do it once).  This is about a 4% improvement in memory.
* Replaces unsafe.Slice/unsafe.StringData with `[]byte(string)` which the compiler can optimize in this case to be zero alloc (confirmed with benchmark). 

Fixes n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/ast.go
+++ b/pkg/traceql/ast.go
@@ -755,10 +755,10 @@ func (s Static) MapKey() StaticMapKey {
 		}
 
 		h := xxhash.New()
-		h.Write(seedBytes)
+		_, _ = h.Write(seedBytes)
 		for _, str := range s.valStrings {
-			h.Write([]byte(str))
-			h.Write(separatorByte)
+			_, _ = h.Write([]byte(str))
+			_, _ = h.Write(separatorByte)
 		}
 
 		return StaticMapKey{typ: s.Type, code: h.Sum64()}


pkg/traceql/ast.go
```diff
@@ -751,10 +751,11 @@ func (s Static) MapKey() StaticMapKey {
 		}
 
 		h := fnv.New64a()
-		_, _ = h.Write(seedBytes) // avoid collisions with values like []string{""}
 		for _, str := range s.valStrings {
-			_, _ = h.Write(unsafe.Slice(unsafe.StringData(str), len(str)))
+			_, _ = h.Write(seedBytes)
+			_, _ = h.Write([]byte(str))
```
stoewer: We use a similar approach for avoiding hash collisions for the dedicated columns in [block_meta.go:L377](https://github.com/grafana/tempo/blob/a97f399545bef66afc25c039cf8e492b1723cef2/tempodb/backend/block_meta.go#L377)
I'm not sure what the best method is, but it would be nice to use the same in both places

mdisibio: Can you clarify - are you thinking the same seed/separator bytes is enough?  Or perhaps shared code.  There is a small difference here in that an empty string is ok for attribute values, but not for attribute name in the dedicated columns config.

stoewer: Yeah, just suggesting to use the same separator and seed for consistency, not necessarily sharing the same code. It could also be a nice touch to switch from xxhash to fnv (as we don't use xxhash anywhere else).
Mainly suggesting this to make things a bit more consistent. Totally fine if you think it’s out of scope for this PR - don’t want to block it

Issue #4099, opened 2024-09-19
Tempo Compactions Failing, CloudFlare R2 Bucket Always Increasing, Compactor throwing error: "error completing block: error completing multipart upload"


alextricity25:
**Describe the bug**
About a week ago, our Tempo alerts for compaction failures started firing.

This is the query that we use. If it's above 1, the alerts fire.
```
sum by (cluster, namespace) (increase(tempodb_compaction_errors_total{}[1h]))
```
![Image](https://github.com/user-attachments/assets/05ab9f75-1760-41fc-bc07-a07e9d24f29e)

I checked the compactor logs (we only have one compactor running), and the only suspicious thing that I've noticed is messages like this:
```
level=error ts=2024-09-17T16:39:39.029886052Z caller=compactor.go:162 msg="error during compaction cycle" err="error shipping block to backend, blockID da530417-5510-4f22-996f-c35336eea572: error completing block: error completing multipart upload, object: single-tenant/da530417-5510-4f22-996f-c35336eea572/data.parquet, obj etag: : All non-trailing parts must have the same length."
level=error ts=2024-09-17T16:41:23.423707238Z caller=compactor.go:162 msg="error during compaction cycle" err="error shipping block to backend, blockID 5b745471-a889-40e9-8152-006625469f6c: error completing block: error completing multipart upload, object: single-tenant/5b745471-a889-40e9-8152-006625469f6c/data.parquet, obj etag: : All non-trailing parts must have the same length."
level=error ts=2024-09-17T16:43:08.882899458Z caller=compactor.go:162 msg="error during compaction cycle" err="error shipping block to backend, blockID 055d0b14-7789-492f-99ca-bd0b74aa1f2c: error completing block: error completing multipart upload, object: single-tenant/055d0b14-7789-492f-99ca-bd0b74aa1f2c/data.parquet, obj etag: : All non-trailing parts must have the same length."
```

On top of that, I also noticed that the r2 bucket that we use for traces is ever increasing in size, which didn't happen before observing the compaction errors.
![Image](https://github.com/user-attachments/assets/8379403d-673c-4495-a464-b9cd36efd657)

**To Reproduce**
1. Start Tempo 2.6.0
2. Configure Tempo to use a CloudFlare R2 Bucket as the s3 backend
3. Perform Operations (Read/Write/Others)
4. Observe compaction errors

**Expected behavior**
I expect that there would be no compaction errors

**Environment:**
 - Infrastructure: Kubernetes GKE
[...]

PR #5838, opened 2025-11-02
Respect configured S3 part size for compatibility with Cloudflare R2

constantins2001:
**Respect configured S3 part size for compatibility with Cloudflare R2**:
Upload compaction parts in uniform size except the last chunk to allow usage of Grafana Tempo with Cloudflare R2.

Fixes https://github.com/grafana/tempo/issues/4099

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/backend/s3/config.go
+++ b/tempodb/backend/s3/config.go
@@ -87,7 +87,6 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
 	f.StringVar(&cfg.SSE.KMSEncryptionContext, util.PrefixConfig(prefix, "s3.sse.kms-encryption-context"), "", "KMS Encryption Context used for object encryption. It expects JSON formatted string.")
 	f.StringVar(&cfg.SSE.CustomerEncryptionKey, util.PrefixConfig(prefix, "s3.sse.encryption-key"), "", "SSE-C Encryption Key used for object encryption.")
 	cfg.HedgeRequestsUpTo = 2
-	cfg.PartSize = 100 * 1024 * 1024 // 100MB default, matches parquet row group size (R2 compliant, min 5MB)
 }
 
 func (cfg *Config) PathMatches(other *Config) bool {
--- a/tempodb/backend/s3/s3.go
+++ b/tempodb/backend/s3/s3.go
@@ -233,36 +233,57 @@ func (rw *readerWriter) Append(ctx context.Context, name string, keypath backend
 		}
 		a.uploadID = id
 		a.objectName = objectName
-		a.buffer = make([]byte, 0, rw.cfg.PartSize)
 	}
 
-	a.buffer = append(a.buffer, buffer...)
-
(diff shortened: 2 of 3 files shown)


tempodb/backend/s3/config.go
```diff
@@ -87,6 +87,7 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
 	f.StringVar(&cfg.SSE.KMSEncryptionContext, util.PrefixConfig(prefix, "s3.sse.kms-encryption-context"), "", "KMS Encryption Context used for object encryption. It expects JSON formatted string.")
 	f.StringVar(&cfg.SSE.CustomerEncryptionKey, util.PrefixConfig(prefix, "s3.sse.encryption-key"), "", "SSE-C Encryption Key used for object encryption.")
 	cfg.HedgeRequestsUpTo = 2
+	cfg.PartSize = 100 * 1024 * 1024 // 100MB default, matches parquet row group size (R2 compliant, min 5MB)
```
joe-elliott: this may cause a small increase in s3 API calls. did you see that in your testing?

constantins2001:
> nice fix with minimal allocs. i hadn't considered this approach.
> 
> 
> 
> it feels a little weird to reuse the existing PartSize option. do you know what this config option does?

Thanks! As far as I know this is the right config value for the intended behaviour.
It is used to control the maximum size of a S3 multipart upload part.

tempodb/backend/s3/s3.go
```diff
@@ -232,26 +233,37 @@ func (rw *readerWriter) Append(ctx context.Context, name string, keypath backend
 		}
 		a.uploadID = id
 		a.objectName = objectName
+		a.buffer = make([]byte, 0, rw.cfg.PartSize)
 	}
 
-	level.Debug(rw.logger).Log("msg", "appending object to s3", "objectName", objectName)
+	a.buffer = append(a.buffer, buffer...)
```
joe-elliott: i love that you've kept the allocs to a minimum but this will still do a lot of memory copying. can we make the default 0 and have it observe the old behavior in that case?

constantins2001: Good idea, I'll implement that!

tempodb/backend/s3/config.go
constantins2001: The value I've chosen should prevent this, as the Parquet writes accumulates data until it hits [RowGroupSizeBytes](https://github.com/grafana/tempo/blob/c5b4b556221610357425e6dc5c0a5114d0716979/tempodb/encoding/common/config.go#L52).
Nevertheless, let's make the old behaviour the default to prevent any issues.

joe-elliott (approved):
Really nice fix for a problem I thought was significantly more complex.

Thanks!

tiffanyfay:
@constantins2001 Would you have any interest in talking about this PR in a community call? Thanks!

constantins2001:
> @constantins2001 Would you have any interest in talking about this PR in a community call? Thanks!

Sounds interesting, feel free to contact me using the email address linked in my profile!

PR #5846, opened 2025-11-03
Make topk/bottom deterministic by tie breaking with labels

joe-elliott:
**What this PR does**:

Topk/bottomk are currently non-deterministic when values are equal. This PR uses the labels to break ties in these cases.

Before:

https://github.com/user-attachments/assets/399fcc81-d738-4b60-94ba-685881add2bc

After:

https://github.com/user-attachments/assets/16be562b-11dd-464f-a286-8ea5b116595d

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/engine_metrics_test.go
+++ b/pkg/traceql/engine_metrics_test.go
@@ -2229,40 +2229,199 @@ func TestProcessBottomK(t *testing.T) {
 }
 
 func TestTiesInTopK(t *testing.T) {
-	input := createSeriesSet(map[string][]float64{
-		"a": {10, 5, 1},
-		"b": {10, 4, 2},
-		"c": {10, 3, 3},
-	})
-	result := processTopK(input, 3, 2)
+	t.Run("three-way tie at first position", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/traceql/series_heap.go
```diff
@@ -1,17 +1,75 @@
...
+
 // seriesValue keeps a value from a time series with its key
 type seriesValue struct {
 	key   SeriesMapKey
 	value float64
 }
 
+// compareSeriesMapKey compares two SeriesMapKey values deterministically.
+// Returns -1 if a < b, 0 if a == b, 1 if a > b
+func compareSeriesMapKey(a, b SeriesMapKey) int {
```
electron0zero, line 13: nit: this code is mostly exercised in metrics tests but it would be nice to have a few tests for compareSeriesMapKey and compareSeriesValues

electron0zero (approved):
one nit for extra unit test but rest lgtm.

pkg/traceql/series_heap.go
joe-elliott: done

PR #5854, opened 2025-11-05
generator: extract validation to reusable package

carles-grafana:
Centralize validation to improve readability and reusability. 

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/processor/processor_names.go
+++ b/modules/generator/processor/processor_names.go
@@ -1,5 +1,7 @@
 package processor
 
+// Span Metrics Subprocessor Names
+// these are used when only specific metrics are wanted
 const (
 	SpanMetricsLatencyName = "span-metrics-latency"
 	SpanMetricsCountName   = "span-metrics-count"


modules/generator/processor/processor_names.go
```diff
@@ -0,0 +1,14 @@
+package processor
+
```
electron0zero, line 2: add a comment to mention that these are subprocesros for the span-metics and can be enabled when you only want subset of span metrics instead of the all span metrics.

modules/generator/processor/dimension_names.go
```diff
@@ -0,0 +1,11 @@
+package processor
```
electron0zero, line 1: nit: does it make sense to put all constants into a single file, something like `cost.go` instead of two files?

modules/generator/validation/fields.go
```diff
@@ -0,0 +1,99 @@
...
+	"fmt"
+	"time"
+
+	"github.com/grafana/tempo/modules/generator/processor"
+	"github.com/grafana/tempo/modules/generator/registry"
+	"github.com/grafana/tempo/pkg/sharedconfig"
+	"github.com/prometheus/common/model"
+)
+
+var SupportedProcessors = []string{
```
electron0zero, line 14: nit: any reasons to put it here instead of `modules/generator/processor/processor_names.go`

modules/generator/config.go
```diff
@@ -144,7 +145,10 @@ func (cfg *ProcessorConfig) Validate() error {
 	if err := cfg.LocalBlocks.Validate(); err != nil {
 		errs = append(errs, err)
 	}
-	if err := cfg.HostInfo.Validate(); err != nil {
```
electron0zero, line 147: `RegisterFlagsAndApplyDefaults` and `.Validate()` is a pattern we follow so removing it makes it odd one out.

we can keep the Validate method to keep this pattern and move the actual validation into the `Validate` method call instead of doing it here.

modules/generator/processor/hostinfo/config.go
```diff
@@ -23,15 +20,3 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(string, *flag.FlagSet) {
 	cfg.HostIdentifiers = []string{"k8s.node.name", "host.id"}
 	cfg.MetricName = defaultHostInfoMetric
 }
-
-func (cfg *Config) Validate() error {
```
electron0zero, line 27: keep this method around and do the actual validation inside this method.

modules/generator/config.go
carles-grafana: I did this change on purpose to avoid a circular dependency. Since we are extracting validation in a new package, we pass the objects to validate to the independent functions instead of having methods.

Only localblocks follows the same pattern, and it will be delete it, so I don't consider it a problem.

modules/generator/processor/dimension_names.go
carles-grafana: I think this is more explicit

modules/generator/processor/processor_names.go
carles-grafana: done

modules/generator/validation/fields.go
carles-grafana: semantically SupportedProcessors is more related to validation

PR #5855, opened 2025-11-05
[bugfix] Wrong sleep duration if one of partitions is inactive

ruslan-mikhailov:
**What this PR does**: fixes a bug when block-builder has multiple partitions assigned with at least one is inactive or has no lag, which causes b.cfg.ConsumeCycleDuration cycle sleep even when other partitions have big lag. This is happening, because [here](https://github.com/ruslan-mikhailov/tempo/blob/8842bf0a56194d79f5cf93f934ddf4daa7439185/modules/blockbuilder/blockbuilder.go#L454) consumePartition exits with lastRecordTs 0. This leads to returning full sleep duration [here](https://github.com/ruslan-mikhailov/tempo/blob/8842bf0a56194d79f5cf93f934ddf4daa7439185/modules/blockbuilder/blockbuilder.go#L294-L296)

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/blockbuilder/blockbuilder.go
+++ b/modules/blockbuilder/blockbuilder.go
@@ -299,6 +299,11 @@ func (b *BlockBuilder) consume(ctx context.Context) (time.Duration, error) {
 		if lagTime < b.cfg.ConsumeCycleDuration {
 			return b.cfg.ConsumeCycleDuration - lagTime, nil
 		}
+		// If we don't know exact offset, we need to start over on next cycle to fetch offsets again.
+		// This can happen when pull timeouted or the consumer had no lag.
+		if laggiestPartition.commitOffset == commitOffsetAtEnd {
+			return 0, nil
+		}
 		lastRecordTs, lastRecordOffset, err := b.consumePartition(ctx, laggiestPartition)
 		if err != nil {
 			return 0, err
--- a/modules/blockbuilder/blockbuilder_test.go
+++ b/modules/blockbuilder/blockbuilder_test.go
@@ -1110,6 +1110,7 @@ func (s *slowStore) unlock() {
 // - committed offsets equal number of sent records.
 // The test highly coupled with the blockbuilder implementation. In case it stuck
 // due to channel, it is advised to debug consume cycle step by step.
+// If the test hangs, it is most likely because storage received unexpected writes.
 func TestBlockbuilder_twoPartitions_secondEmpty(t *testing.T) {
 	ctx, cancel := context.WithCancelCause(context.Background())
 	t.Cleanup(func() { cancel(errors.New("test done")) })
@@ -1122,7 +1123,7 @@ func TestBlockbuilder_twoPartitions_secondEmpty(t *testing.T) {
 	ch := make(chan struct{})
 	store := &slowStore{Store: newStore(ctx, t), wait: ch}
 	cfg := blockbuilderConfig(t, address, []int32{0, 1})
-	cfg.ConsumeCycleDuration = 7 * time.Second
+	cfg.ConsumeCycleDuration = time.Second
 	partitionRing := newPartitionRingReaderWithPartitions(map[int32]ring.PartitionDesc{
 		0: {Id: 0, State: ring.PartitionActive},
 		1: {Id: 1, State: ring.PartitionActive},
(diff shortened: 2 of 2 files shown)


modules/blockbuilder/blockbuilder.go
```diff
@@ -451,7 +456,7 @@ outer:
 		span.AddEvent("no data")
 		// No data means we are caught up
 		ingest.SetPartitionLagSeconds(group, ps.partition, 0)
-		return time.Time{}, commitOffsetAtEnd, nil
+		return time.Now(), commitOffsetAtEnd, nil
```
ruslan-mikhailov, line 459: Pay attention, that calling getStartOffset() with `commitOffsetAtEnd` will force starting from `AtStart()`: [link](https://github.com/ruslan-mikhailov/tempo/blob/8842bf0a56194d79f5cf93f934ddf4daa7439185/modules/blockbuilder/blockbuilder.go#L129-L145)

It has protection against it here: https://github.com/grafana/tempo/pull/5855/files#diff-1f801af33df7e19afcfef8d30d584b9fb0c98833230e37095b29a91bbd6e29a0R302-R306, where I force refetching and get exact values

modules/blockbuilder/blockbuilder_test.go
```diff
@@ -658,7 +674,8 @@ func TestBlockbuilder_usesRecordTimestampForBlockStartAndEnd(t *testing.T) {
 				return nil, nil, false
 			})
 
-			store := newStore(ctx, t)
+			store, stopStore := newStore(ctx, t)
+			t.Cleanup(stopStore)
```
mapno: Can this be added to `newStore`?

modules/blockbuilder/blockbuilder_test.go
```diff
@@ -235,7 +240,12 @@ func TestBlockbuilder_receivesOldRecords(t *testing.T) {
 		return nil, nil, false
 	})
 
-	store := newStore(ctx, t)
+	store, stopStore := newStore(ctx, t)
+	t.Cleanup(stopStore)
+	t.Cleanup(func() {
+		require.NoError(t, services.StopAndAwaitTerminated(ctx, store))
+	})
```
mapno: Duplicated with `stopStore`

modules/blockbuilder/blockbuilder.go
```diff
@@ -292,13 +292,18 @@ func (b *BlockBuilder) consume(ctx context.Context) (time.Duration, error) {
 
 		laggiestPartition := ps[0]
 		if laggiestPartition.lastRecordTs.IsZero() {
-			return b.cfg.ConsumeCycleDuration, nil
+			return b.cfg.ConsumeCycleDuration, errors.New("partition has no last record timestamp")
```
mapno, line 295: I think this line is not reachable now 🤔 

Every time `consumePartition` return with `time.Time{}` it also returns with an error

1. Line 375: `return time.Time{}, commitOffsetAtEnd, err` - Returns zero time with error
2. Line 426: `return time.Time{}, commitOffsetAtEnd, err` - Returns zero time with error
3. Line 459: `return time.Now(), commitOffsetAtEnd, nil` - Returns non-zero time, no error (when no data received)
4. Line 467: `return time.Time{}, commitOffsetAtEnd, err` - Returns zero time with error (flush failed)
5. Line 473: `return time.Time{}, commitOffsetAtEnd, err` - Returns zero time with error (commit failed)
6. Line 485: `return lastRec.Timestamp, offset.At, nil` - Returns non-zero time, no error (success case)

I guess it's fine as a safe-check. I just wanted to point it out.

modules/blockbuilder/blockbuilder.go
```diff
@@ -292,13 +292,18 @@ func (b *BlockBuilder) consume(ctx context.Context) (time.Duration, error) {
...
 
 		lagTime := time.Since(laggiestPartition.lastRecordTs)
 		if lagTime < b.cfg.ConsumeCycleDuration {
 			return b.cfg.ConsumeCycleDuration - lagTime, nil
 		}
+		// If we don't know exact offset, we need to start over on next cycle to fetch offsets again.
+		// This can happen when pull timeouted or the consumer had no lag.
+		if laggiestPartition.commitOffset == commitOffsetAtEnd {
+			return 0, nil
+		}
```
mapno, line 306: Doesn't the loop exit in the previous condition? When `consumePartition` returns with `commitOffsetAtEnd` and no error, it also returns `time.Now()`, so `lagTime` will be less than `ConsumeCycleDuration`.

modules/blockbuilder/blockbuilder.go
ruslan-mikhailov: Yes! We shouldn't have empty time at all, that's why I return an error. The error will be logged and the consumer will be forced to sleep full cycle duration.

I was thinking of panics, but I thought it would be too brutal.

modules/blockbuilder/blockbuilder_test.go
ruslan-mikhailov: \+

Moved it to `newStore`

modules/blockbuilder/blockbuilder_test.go
ruslan-mikhailov: Oh, indeed! That's much simpler now, thanks!

modules/blockbuilder/blockbuilder.go
ruslan-mikhailov: It can happen only when there is a lag to catch up, e.g. if block builder was down for 30 minutes. In that case, it should not sleep and has to continue until it reaches recent traces again.

PR #5858, opened 2025-11-05
feat: Add default_spans_per_span_set config and treat spss=0 as unlimited

### LINKED ISSUE #5435: spss in search does not respect configured default
**Describe the bug**
The `max_spans_per_span_set: 0` configuration setting in the query-frontend is not respected when the `spss` parameter is not explicitly provided in search API requests. When no `spss` parameter is specified, the API automatically sets `SpansPerSpanSet = 3` (hardcoded default), which overrides the configuration setting and limits search results to 3 spans per span set.

**To Reproduce**
Steps to reproduce the behaviour:
1. Start Tempo with configuration:
   ```yaml
   query_frontend:
     search:
       max_spans_per_span_set: 0
   ```
2. Perform a search request without the `spss` parameter:
   ```bash
   curl -G -s "http://localhost:3200/api/search" \
     --data-urlencode 'q={ status=error }'
   ```
3. Observe that only 3 spans are returned per span set, despite the configuration setting `max_spans_per_span_set: 0`

**Expected behaviour**
When `max_spans_per_span_set: 0` is configured, search requests without an explicit `spss` parameter should return all available spans per span set, not be limited to the hardcoded default of 3 spans.

**Environment:**

**Additional Context**
The issue occurs because the API layer in `pkg/api/http.go` sets a hardcoded default value of 3 for `SpansPerSpanSet` when no `spss` parameter is provided:

```go
req := &tempopb.SearchRequest{
    Tags:            map[string]string{},
    SpansPerSpanSet: defaultSpansPerSpanSet,  // Always 3
}
```

The `max_spans_per_span_set` configuration only affects frontend validation in `modules/frontend/search_sharder.go`, but doesn't influence the API default value. This creates a disconnect between the configuration intent and the actual behaviour.

**Workaround:**
Currently, users must explicitly specify a high `spss` value in their requests:
```bash
curl -G -s "http://localhost:3200/api/search" \
  --data-urlencode 'q={ status=error }' \
  --data-urlencode 'spss=10000'
```

**Related Files:**
- `pkg/api/http.go` (lines 121-124)
- `modules/frontend/search_sharder.go` (lines 100-102)
- `pkg/traceql/engine.go` (lines 102-107)

iamrajiv:
**What this PR does**:

Add default_spans_per_span_set config and treat spss=0 as unlimited

Fixes #5435

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/frontend/search_sharder.go
+++ b/modules/frontend/search_sharder.go
@@ -105,7 +105,7 @@ func (s asyncSearchSharder) RoundTrip(pipelineRequest pipeline.Request) (pipelin
 	// Validate SpansPerSpanSet against MaxSpansPerSpanSet
 	// If MaxSpansPerSpanSet is 0, it means unlimited spans are allowed
 	// If MaxSpansPerSpanSet is non-zero, enforce the limit
-	if s.cfg.MaxSpansPerSpanSet >= 0 && searchReq.SpansPerSpanSet > s.cfg.MaxSpansPerSpanSet {
+	if s.cfg.MaxSpansPerSpanSet != 0 && searchReq.SpansPerSpanSet > s.cfg.MaxSpansPerSpanSet {
 		return pipeline.NewBadRequest(fmt.Errorf("spans per span set exceeds %d. received %d", s.cfg.MaxSpansPerSpanSet, searchReq.SpansPerSpanSet)), nil
 	}
 


anglerfishlyy (commented):
Hi @iamrajiv , overall implementation looks solid, good addition with `default_spans_per_span_set` , and the docs and tests are covered. a few CI checks are failing  (lint , unit , e2e ), likely just needs  a `make fmt`   and ensuring `ParseSearchRequestWithDefault()`  is consistently used. once those are fixed , it should be good to go.

modules/frontend/search_sharder.go
```diff
@@ -99,6 +102,9 @@ func (s asyncSearchSharder) RoundTrip(pipelineRequest pipeline.Request) (pipelin
 		return pipeline.NewBadRequest(fmt.Errorf("range specified by start and end exceeds %s. received start=%d end=%d", maxDuration, searchReq.Start, searchReq.End)), nil
 	}
 
+	// Validate SpansPerSpanSet against MaxSpansPerSpanSet
+	// If MaxSpansPerSpanSet is 0, it means unlimited spans are allowed
+	// If MaxSpansPerSpanSet is non-zero, enforce the limit
 	if s.cfg.MaxSpansPerSpanSet != 0 && searchReq.SpansPerSpanSet > s.cfg.MaxSpansPerSpanSet {
```
javiermolinar, line 108:
```suggestion
	if s.cfg.MaxSpansPerSpanSet >= 0 && searchReq.SpansPerSpanSet > s.cfg.MaxSpansPerSpanSet {
```

Just to avoid weird behaviors in case a negative number is used

anglerfishlyy:
Thanks for the update @iamrajiv,changes look good after formatting. Agree with @javiermolinar’s suggestion on handling negative values for safety. You can also run `make generate-manifest` and `make test-e2e-api` locally to catch any errors early and keep the commit history clean before merge 👍

modules/frontend/search_sharder.go
```diff
@@ -99,7 +102,10 @@ func (s asyncSearchSharder) RoundTrip(pipelineRequest pipeline.Request) (pipelin
 		return pipeline.NewBadRequest(fmt.Errorf("range specified by start and end exceeds %s. received start=%d end=%d", maxDuration, searchReq.Start, searchReq.End)), nil
 	}
 
-	if s.cfg.MaxSpansPerSpanSet != 0 && searchReq.SpansPerSpanSet > s.cfg.MaxSpansPerSpanSet {
+	// Validate SpansPerSpanSet against MaxSpansPerSpanSet
+	// If MaxSpansPerSpanSet is 0, it means unlimited spans are allowed
+	// If MaxSpansPerSpanSet is non-zero, enforce the limit
+	if s.cfg.MaxSpansPerSpanSet >= 0 && searchReq.SpansPerSpanSet > s.cfg.MaxSpansPerSpanSet {
```
javiermolinar: Sorry I didn't notice this is an unsigned int so it cannot be negative. Hence the linter error

iamrajiv:
@javiermolinar @knylander-grafana can we run the ci again?

javiermolinar (approved):
LGTM thank you for your contribution

PR #5865, opened 2025-11-06
feat: Add support for protobuf format in all endpoints

oleg-kozlyuk-grafana:
**What this PR does**:

* Adds support for protobuf encoding based on presence of "Accept: application/protobuf" header
* Extends tests appropriately to cover both "application/protobuf" and "application/json" cases

**Checklist**
- [x] Tests updated
- [x] Documentation added


modules/frontend/traceid_handlers.go
```diff
@@ -58,7 +56,7 @@ func newTraceIDHandler(cfg Config, next pipeline.AsyncRoundTripper[combiner.Pipe
 			}
 		}
 
-		comb := combinerFn(o.MaxBytesPerTrace(tenant), marshallingFormat, traceRedactor)
+		comb := combinerFn(o.MaxBytesPerTrace(tenant), string(marshallingFormat), traceRedactor)
```
electron0zero: we change the signature of the combinerFn and avoid casting it to string?

I assumed that we defined MarshallingFormat type so we can strongly type it? right?

modules/frontend/tag_handlers.go
```diff
@@ -223,9 +233,11 @@ func newTagValuesV2StreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTrip
 		if err != nil {
 			return err
 		}
+		// Use protobuf for internal queries by default
+		marshallingFormat := util.MarshallingFormatProtobuf
```
electron0zero: any reasons to define it as a var, instead of just passing `util.MarshallingFormatProtobuf` in the comb? I see this is only used once?

modules/frontend/tag_handlers.go
```diff
@@ -178,8 +185,11 @@ func newTagValuesStreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTrippe
 			return err
 		}
 
+		// Use protobuf for internal queries by default
+		marshallingFormat := util.MarshallingFormatProtobuf
```
electron0zero: same comment as the newTagValuesV2StreamingGRPCHandler

modules/frontend/tag_handlers.go
```diff
@@ -55,8 +56,11 @@ func newTagsStreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTripper[com
...
```
electron0zero: and here as well.

modules/frontend/pipeline/responses_test.go
```diff
@@ -304,7 +304,7 @@ func TestAsyncResponsesDoesNotLeak(t *testing.T) {
 				bridge := &pipelineBridge{
 					next: tc.finalRT(cancel),
 				}
-				httpCollector := NewHTTPCollector(sharder{next: bridge}, 0, combiner.NewSearch(0, false))
+				httpCollector := NewHTTPCollector(sharder{next: bridge}, 0, combiner.NewSearch(0, false, "application/json"))
```
electron0zero: can we use `api.HeaderAcceptJSON` in here instead of string?

same for the lines below?

pkg/util/http.go
```diff
@@ -22,6 +22,30 @@ import (
 
 const messageSizeLargerErrFmt = "received message larger than max (%d vs %d)"
 
+const (
```
javiermolinar: We have that already defined here

https://github.com/grafana/tempo/blob/303c7ef9e086955071312522de5c1bc26565e733/pkg/api/http.go#L66

oleg-kozlyuk-grafana: Good point, moved this to same file to be able to reuse without circular dependency

modules/frontend/tag_handlers.go
oleg-kozlyuk-grafana: Fixed

modules/frontend/tag_handlers.go
oleg-kozlyuk-grafana: Fixed

modules/frontend/pipeline/responses_test.go
oleg-kozlyuk-grafana: Fixed

modules/frontend/tag_handlers.go
oleg-kozlyuk-grafana: Fixed

modules/frontend/traceid_handlers.go
oleg-kozlyuk-grafana: Yes, looks like we can change it here as well, done

PR #5868, opened 2025-11-07
chore: Switch from Dependabot to Renovate

oleg-kozlyuk-grafana:
**What this PR does**:

Switches from Dependabot to Renovate due to following advantages:

* Automatic detection of hidden dependencies
* Pinning of dependency versions making builds more reproducible
* Renovate support within Grafana

**Checklist**

N/A


javiermolinar (commented):
Could we remove this? https://github.com/grafana/tempo/blob/main/.github/workflows/dependabot-auto-approve.yml

it never worked and now if we change to renovate we should create a new one

PR #5936, opened 2025-11-12
feat: improve observability of collection failures in the metrics generator with error categorization

javiermolinar:
**What this PR does**:
It improves the observability of collection errors in the metrics generator by categorize Prometheus errors. Right now the current metric lacks of proper detail.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 * [ENHANCEMENT] Add metric for tracking added latency to write requests [#5781](https://github.com/grafana/tempo/pull/5781) (@mapno)
 * [ENHANCEMENT] Improve error message when overrides fail to parse [#5787](https://github.com/grafana/tempo/pull/5787) (@mapno)
 * [ENHANCEMENT] Increase weight for heavy TraceQL queries [#5782](https://github.com/grafana/tempo/pull/5782) (@ruslan-mikhailov)
+* [ENHANCEMENT] Improve observability of collection failures in the metrics generator with error categorization [#5936](https://github.com/grafana/tempo/pull/5936) (@javiermolinar)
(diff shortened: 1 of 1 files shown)


modules/generator/registry/errors.go
```diff
@@ -0,0 +1,34 @@
...
+		return "none"
+	}
+
+	switch {
+	case errors.Is(err, storage.ErrNotFound):
+		return "not_found"
+	case errors.Is(err, tsdb.ErrInvalidSample),
+		errors.Is(err, tsdb.ErrInvalidExemplar):
+		return "invalid"
+	case errors.Is(err, storage.ErrOutOfOrderSample),
```
carles-grafana, line 22: sounds good but why group multiple error types with the same label?

javiermolinar: Mostly to have an actionable observability. Matching 1-1 the errors in Prometheus can end in cognitive overload, this also maintains forward compatibly. Also, this keeps the cardinality low. We can always go to the logs to know the exact error

modules/generator/registry/errors.go
```diff
@@ -0,0 +1,34 @@
...
+		errors.Is(err, storage.ErrOutOfBounds),
+		errors.Is(err, storage.ErrTooOldSample),
+		errors.Is(err, storage.ErrOutOfOrderCT),
+		errors.Is(err, storage.ErrCTNewerThanSample):
+		return "out_of_order"
+	case errors.Is(err, storage.ErrDuplicateSampleForTimestamp),
+		errors.Is(err, storage.ErrDuplicateExemplar):
```
electron0zero, line 29: nit: Exemplar seems like a fit for dedicated ErrType instead of being clubbed with duplicate_sample

javiermolinar: The idea is having behavioral categories, not type-based

PR #5945, opened 2025-11-14
fix: ignore out of order samples when registering first series seen

javiermolinar:
**What this PR does**:

We were experiencing frequent out-of-order sample errors. These errors occurred when time series were removed from the metrics generator registry due to staleness cleanup, but still existed in Prometheus's TSDB and we try to initialize the first series to 0 to avoid downsampling in adaptive metrics.

The metrics generator and Prometheus have different retention policies for time series:

- **Metrics Generator Registry**: Removes stale series every 15 minutes (`removeStaleSeriesInterval`)
- **Prometheus TSDB**: Retains series data for much longer

This mismatch creates the following scenario:

1. A series is created and samples are written to Prometheus
2. The series goes idle for more than 15 minutes
3. Metrics generator removes it from the registry due to staleness
4. The same series becomes active again
5. Metrics generator treats it as a "new" series (sets `firstSeries = true`)
6. Attempts to add an initialization point with value `0` at the end of the previous minute
7. **Prometheus rejects it with `ErrOutOfOrderSample`** because it already has more recent data for that series

I've modified the `collectMetrics` method in `counter.go`, `histogram.go`, and `native_histogram.go` to gracefully handle out-of-order errors when inserting the initialization point.

**Why This Is Safe**:

1. The initialization point (`0` value at end of previous minute) is only needed to avoid downsampling for truly new series
2. If Prometheus returns an out-of-order error, it means the series already exists in TSDB
3. The current sample value will be appended normally and will be in the correct order
4. This doesn't affect data correctness - we're just skipping a redundant initialization point

 
**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/registry/counter.go
+++ b/modules/generator/registry/counter.go
@@ -40,12 +40,12 @@ var (
 	_ metric  = (*counter)(nil)
 )
 
-func (co *counterSeries) wasNew() bool {
-	return co.firstSeries.CompareAndSwap(true, false)
+func (co *counterSeries) isNew() bool {
+	return co.firstSeries.Load()
 }
 
-func (co *counterSeries) resetSeenSeries() {
-	co.firstSeries.Store(true)
+func (co *counterSeries) registerSeenSeries() {
+	co.firstSeries.Store(false)
 }
 
 func newCounter(name string, lifecycler Limiter, externalLabels map[string]string, staleDuration time.Duration) *counter {
@@ -117,19 +117,18 @@ func (c *counter) collectMetrics(appender storage.Appender, timeMs int64) error 
 		// If we are about to call Append for the first time on a series, we need
 		// to first insert a 0 value to allow Prometheus to start from a non-null
 		// value.
-		if s.wasNew() {
+		if s.isNew() {
 			// We set the timestamp of the init serie at the end of the previous minute, that way we ensure it ends in a
 			// different aggregation interval to avoid be downsampled.
 			endOfLastMinuteMs := getEndOfLastMinuteMs(timeMs)
 			_, err := appender.Append(0, s.labels, endOfLastMinuteMs, 0)
-
 			// Out-of-order errors occur when a series is deleted from our registry due to staleness cleanup,
 			// but still exists in Prometheus's TSDB. When the series reappears, we attempt to add the initial 0
 			// but Prometheus already has more recent data. We ignore this error and continue with the current value.
 			if err != nil && !isOutOfOrderError(err) {
-				s.resetSeenSeries()
 				return err
 			}
+			s.registerSeenSeries()
 		}
 
 		_, err := appender.Append(0, s.labels, timeMs, s.value.Load())
--- a/modules/generator/registry/histogram.go
+++ b/modules/generator/registry/histogram.go
@@ -54,12 +54,12 @@ type histogramSeries struct {
 	firstSeries *atomic.Bool
 }
 
-func (co *histogramSeries) wasNew() bool {
-	return co.firstSeries.CompareAndSwap(true, false)
+func (hs *histogramSeries) isNew() bool {
+	return hs.firstSeries.Load()
 }
 
-func (co *histogramSeries) resetSeenSeries() {
-	co.firstSeries.Store(true)
+func (hs *histogramSeries) registerSeenSeries() {
+	hs.firstSeries.Store(false)
 }
 
 var (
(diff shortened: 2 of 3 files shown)


modules/generator/registry/counter_test.go
```diff
@@ -279,19 +298,18 @@ func collectMetricAndAssert(t *testing.T, m metric, collectionTimeMs int64, expe
...
 	assert.ElementsMatch(t, expectedExemplars, appender.exemplars)
-	assert.Equal(t, expectedActiveSeries, m.countActiveSeries())
 }
 
 func Test_counter_demandTracking(t *testing.T) {
-	c := newCounter("my_counter", nil, nil, nil, 15*time.Minute)
+	c := newCounter("my_counter", noopLimiter, map[string]string{}, 15*time.Minute)
 
 	// Initially, demand should be 0
 	assert.Equal(t, 0, c.countSeriesDemand())
```
mapno, line 297: Why is this removed?

modules/generator/registry/errors.go
```diff
@@ -32,3 +32,7 @@ func getErrType(err error) string {
 		return "other"
 	}
 }
+
+func isOutOfOrderError(err error) bool {
+	return err != nil && errors.Is(err, storage.ErrOutOfOrderSample)
+}
```
mapno, line 38: `errors.Is` already checks if the error is `nil` (plus it's also checked before all the calls to this function. Can it be simplified to `errors.Is(err, storage.ErrOutOfOrderSample)` to remove clutter?

modules/generator/registry/counter_test.go
```diff
@@ -338,30 +358,61 @@ func Test_counter_demandVsActiveSeries(t *testing.T) {
...
-		c.removeStaleSeries(time.Now().Add(time.Hour).UnixMilli())
+		c.removeStaleSeries(time.Now().UnixMilli())
 	}
 
 	// Demand should have decreased or be zero
 	finalDemand := c.countSeriesDemand()
 	assert.LessOrEqual(t, finalDemand, initialDemand/2, "demand should significantly decay")
 }
 
+func Test_counter_onUpdate(t *testing.T) {
```
mapno, line 371: I'm confused by this test, is this testing new functionality or something that wasn't covered before?

modules/generator/registry/errors.go
javiermolinar: TIL, sure good catch

modules/generator/registry/counter_test.go
javiermolinar: agree, that was more for me to test that after remove stale series "old" series  are added a new ones. I'll remove those lines

modules/generator/registry/counter.go
```diff
@@ -40,12 +40,12 @@ var (
 	_ metric  = (*counter)(nil)
 )
 
-func (co *counterSeries) isNew() bool {
-	return co.firstSeries.Load()
+func (co *counterSeries) wasNew() bool {
+	return co.firstSeries.CompareAndSwap(true, false)
```
oleg-kozlyuk-grafana: Does having to resort to CAS indicate a deeper issue? If we have multiple threads that are attempting to simultaneously write same counter series, can we end up with multiple concurrent time series _with different values_ for same name/labels?

javiermolinar: No since we have this lock https://github.com/grafana/tempo/blob/df20e3b44ecce0ce9ca147d81f569647fc484fb9/modules/generator/registry/counter.go#L113

This is more to be extra precautious

oleg-kozlyuk-grafana: I have a couple of design concerns here:

1.The lock you mention is a read lock, not write lock, which means that multiple threads can enter this function simultaneously - and the function seems to change state of `c.series`
2.Performance - a function should either be atomic or under mutex. If there is both, it is possible some assumptions are incorrect
3.Other operations seem to not use same mutex - newSeries, updateSeries

Not a PR blocker or criticism - just suggesting it may be worth to iterate on the design here

javiermolinar: You have a point. I'll revert the CAS since it adds more confusion and in the end does not solve the "out of order" issue

javiermolinar: Nevertheless I think the race condition still can happen. If two threads call collectMetrics simultaneously (read lock allow it), both could see firstSeries = true

Both would append the 0 value on the first collection
Duplicate initialization entries

But I agree that maybe the CAS is just a band-aid

javiermolinar: On the other hand, switching to a write lock can result in a performance degradation

> 2.Performance - a function should either be atomic or under mutex. If there is both, it is possible some assumptions are incorrect

Since the firsts series initialization only happens once the combination of a read lock and the atomic struct doesn't lock that bad

javiermolinar: > 3.Other operations seem to not use same mutex - newSeries, updateSeries

updateSeries and newSeries are called from the Inc in the write lock section

PR #5953, opened 2025-11-17
[Generator] Use prometheus labels instead of LabelValueCombo

Logiraptor:
This PR replaces registry.LabelValueCombo with labels.Labels from prometheus. We're already transforming LabelValueCombo into labels.Labels, so my intention is to remove unnecessary transformations, gain access to recent optimizations in prometheus, and setup the `registry.Registry` interface for further improvements. 

I also just find it much easier to follow the code when it looks like this:

```
	if p.Cfg.IntrinsicDimensions.SpanName {
	-	labelValues = append(labelValues, span.GetName())
	+	builder.Add(gen.DimSpanName, span.GetName())
	}
```

The name + value are together instead of split between the constructor and the implementation code.

**Here's the background I've pieced together on LabelValueCombo:**

### (1) Original LabelValues implementation

#1340 added the ManagedRegistry and with it, a `LabelValues` struct. At the time, this struct was significantly more efficient than labels.Labels for two reasons: 
* It stored only values and relied on constant label names stored elsewhere
* It cached the label hash across metrics

### (2) Addition of label names

The first point was nullified in #2261, when `LabelValues` became `LabelValueCombo` and now stores both label names and values. This leaves just the hash improvement. At this time, the Hash implementation in prometheus involved loops, a byte buffer, etc. This made it expensive: 

```
func (ls Labels) Hash() uint64 {
	// Use xxhash.Sum64(b) for fast path as it's faster.
	b := make([]byte, 0, 1024)
	for i, v := range ls {
		if len(b)+len(v.Name)+len(v.Value)+2 >= cap(b) {
			// If labels entry is 1KB+ do not allocate whole entry.
			h := xxhash.New()
			_, _ = h.Write(b)
			for _, v := range ls[i:] {
				_, _ = h.WriteString(v.Name)
				_, _ = h.Write(seps)
				_, _ = h.WriteString(v.Value)
				_, _ = h.Write(seps)
			}
			return h.Sum64()
		}

		b = append(b, v.Name...)
		b = append(b, sep)
		b = append(b, v.Value...)
		b = append(b, sep)
	}
	return xxhash.Sum64(b)
}
```

### (3) Prometheus optimizations

Since then, #5671 upgraded the vendored prometheus version to one which uses the new `stringlabels` implementation. This version is not only more memory efficient in general, it also has a Hash implementation which turns out to be extremely fast:

```
func (ls Labels) Hash() uint64 {
	return xxhash.Sum64(yoloBytes(ls.data))
}
```

Note `xxhash.Sum64` is written entirely in assembly and is extremely fast. (https://github.com/prometheus/prometheus/pull/10991 shows about 2x improvement). Since the hash caching done in LabelValueCombo only saves work across multiple metrics for the same span, we're at a point where IMO the additional complexity is no longer worth it.

### Results from profiling

Running this image in a dev cell shows approximately net-zero change in resource usage. This aligns with my expectation, since we're doing roughly equivalent work: the `labels.Builder` type stores all names + values in a slice, just like LabelValueCombo did. It's only after converting to a labels.Labels value that we get the neat optimizations.

### Changes to registry.Registry

Along with the change, I've modified `registry.Registry` to provide a method `NewLabelBuilder` instead of `NewLabelValueCombo`. This was necessary to keep the max length truncation behavior inside the registry. Now that it exists, however, we have the opportunity to do some interesting things:

* We could move label name sanitization into the builder, so it's not duplicated into each processor
* We could implement string interning in a central location without changing processor code
* We could implement filtering or compaction rules to reduce label cardinality.

I've intentionally kept the LabelBuilder interface as small as possible to leave options open.

### DIFF
--- a/modules/generator/processor/hostinfo/processor.go
+++ b/modules/generator/processor/hostinfo/processor.go
@@ -23,7 +23,6 @@ type Processor struct {
 	gauge      registry.Gauge
 	registry   registry.Registry
 	metricName string
-	labels     []string
 }
 
 func (p *Processor) Name() string {
@@ -62,16 +61,12 @@ func (p *Processor) PushSpans(_ context.Context, req *tempopb.PushSpansRequest) 
 func (p *Processor) Shutdown(_ context.Context) {}
 
 func New(cfg Config, reg registry.Registry, logger log.Logger) (*Processor, error) {
-	labels := make([]string, 2)
-	labels[0] = hostIdentifierAttr
-	labels[1] = hostSourceAttr
 	p := &Processor{
 		Cfg:        cfg,
 		logger:     logger,
 		registry:   reg,
 		metricName: cfg.MetricName,
 		gauge:      reg.NewGauge(cfg.MetricName),
-		labels:     labels,
 	}
 	return p, nil
 }
--- a/modules/generator/processor/servicegraphs/servicegraphs.go
+++ b/modules/generator/processor/servicegraphs/servicegraphs.go
@@ -71,7 +71,6 @@ type Processor struct {
 	Cfg Config
 
 	registry registry.Registry
-	labels   []string
 	store    store.Store
 
 	closeCh chan struct{}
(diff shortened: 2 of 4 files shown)


joe-elliott (approved):
Very nice improvement in readability. Also like the consolidation of adding label adding logic in one spot. There's future potential there for label name/value validation/redaction/etc.

PR #5954, opened 2025-11-17
[Generator] Produce overflow series instead of dropping

Logiraptor:
**What this PR does**:

This PR establishes a pattern for Limiters to control cardinality by combining series, instead of just dropping them. A limiter that does this would ideally never drop data and instead would find an optimal strategy for combination that keeps all data available _somewhere_ while maximizing useful information.

As a first step, this PR is the most basic form: once the limit is hit, all new series are collapsed into a single "overflow" series. Future implementations could do this at a per-service level or any other level of granularity, including dynamically finding high cardinality span names and template-izing them on the fly - PR incoming later for that 😉 

Since this occurs completely outside the lifecycle tracking done inside the metrics, it's enough to simply move new series into those overflow buckets, and the data under the more granular labels will automatically age out over time. As stale series are pruned, new ones will be split out from the overflow series until the limit is reached again, and the cycle repeats. 

Happy to hear feedback on the specific labels used in the overflow series. I took inspiration from the otel metric sdk spec here: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#cardinality-limits. So that means each metric will get a single series with the label `metric_overflow="true"`. I would like to add other labels, but as is it's a bit tricky. The limiter is used for multiple processors, and it doesn't have a mechanism to return different labels to each processor. For example spanmetrics might want service and span_name, but service graphs would want something else. Any ideas welcome!

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 * [FEATURE] Add SSE-C encryption support to S3 backend [#5789](https://github.com/grafana/tempo/pull/5789) (@steffsas)
 * [FEATURE] Add support for `Accept: application/vnd.grafana.llm` to Tempo endpoints. Currently supported directly by trace by id and tag values [#5961](https://github.com/grafana/tempo/pull/5961) (@joe-elliott)
   This response is subject to change and should not be relied on. It is intended for LLM consumption only. Even a fundamental change to its representation (yaml? markdown?) would not be considered breaking.
+* [FEATURE] Metrics generator will now produce overflow series to capture new data once limits are hit. These series have the label `metric_overflow="true"`. [#5954](https://github.com/grafana/tempo/pull/5954) (@Logiraptor)
 * [ENHANCEMENT] docs: Add explicit notes about authentication [#5735](https://github.com/grafana/tempo/pull/5735) (@electron0zero)
 * [ENHANCEMENT] On startup, first record for live store to consume is not older than two complete block timeouts [#5693](https://github.com/grafana/tempo/pull/5693) (@ruslan-mikhailov)
 * [ENHANCEMENT] Add secure connection support to tempo-cli [#5692](https://github.com/grafana/tempo/pull/5692) (@TheoBrigitte)


modules/generator/localentitylimiter/local_entity_limiter.go
```diff
@@ -77,13 +81,47 @@ func New(maxEntityFunc func(tenant string) uint32, tenant string, limitLogger *t
 		metricMaxActiveEntities:    metrics.maxActiveEntities.WithLabelValues(tenant),
 		metricTotalEntitiesAdded:   metrics.totalEntitiesAdded.WithLabelValues(tenant),
 		metricTotalEntitiesRemoved: metrics.totalEntitiesRemoved.WithLabelValues(tenant),
+
+		// The overflow entity implements the otel metric sdk spec for cardinality control.
+		// See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#cardinality-limits
+		overflowEntity: labels.FromStrings("otel_metric_overflow", "true"),
```
joe-elliott: i don't like this name but only b/c the span metrics connector has drifted from tempo metrics over the years. it's almost like i'd rather have it clearly different since everything else is. wdyt?

we've discussed having an optional "otel compatibility mode" for the metrics generator where the generator would produce otel collector compatible metrics where we could set this to the otel name.

modules/generator/localserieslimiter/local_series_limiter.go
```diff
@@ -76,6 +77,10 @@ func New(maxSeriesFunc func(tenant string) uint32, tenant string, limitLogger *t
 	}
 }
 
+func (l *LocalSeriesLimiter) SanitizeLabels(lbls labels.Labels) labels.Labels {
+	return lbls
```
joe-elliott: Should the above logic be separate from the local entity limiter so it can be used here as well?

modules/generator/localentitylimiter/local_entity_limiter.go
```diff
@@ -77,13 +81,47 @@ func New(maxEntityFunc func(tenant string) uint32, tenant string, limitLogger *t
...
+		return lbls
+	}
+
+	maxEntities := l.maxEntityFunc(l.tenant)
+	if maxEntities != 0 && uint32(len(l.entityActiveSeries))+1 >= maxEntities {
+		// Since we will never reject an entity in OnAdd, we need to capture our telemetry here.
+		l.limitLogger.Log("msg", "reached max active entities", "active_entities", len(l.entityActiveSeries), "max_active_entities", maxEntities)
+		l.metricTotalEntitiesLimited.Add(float64(1))
+		return l.overflowEntity
+	}
```
joe-elliott: there's some strange overlap in functionality between this and OnAdd. i think there's also a race condition in that SanitizeLabel could "pass" a set of labels that OnAdd later rejects.

i was trying to think of a way to add this functionality directly in `OnAdd`. that way the same set of code that rejects also choose to use the overflow labels. Perhaps `OnAdd` returns the mutated labels and "never fails"?

```
func OnAdd(labelHash uint64, seriesCount, lbls labels.Labels) (labels.Labels, uint64)
```

and then e.g. counter code becomes something like:

```
	lbls, hash := c.lifecycler.OnAdd(hash, 1, lbls)

	c.series[hash] = c.newSeries(lbls, value)
```

modules/generator/localserieslimiter/local_series_limiter.go
Logiraptor: It actually can't be used here, not without increasing the cost of the local series limiter. Since we don't track which series were allowed, we can't do the same logic of "don't change it if we already accepted it"

modules/generator/localentitylimiter/local_entity_limiter.go
Logiraptor: I thought of doing it this way, but had a few reservations:

1. Most importantly, it leaks label transformation into all the metric implementations instead of keeping it in the builder. 
2. As I noted above, the series limiter can't do this, so it has to be able to fail in OnAdd

You are right about the rate condition though. Note for myself: the race condition is as follows (G for goroutine)

1. G1 calls SanitizeLabels while 1 away from the limit and gets back the same labels L1
2. G2 calls SanitizeLabels while 1 away from the limit and gets back the same labels L2
3. G1 calls OnAdd with L1 and is accepted
4. G2 calls OnAdd with L2 and is rejected

Logiraptor: The consequence of that race condition is that some data is lost for L2, _but_ the next time a goroutine attempts to send data for L2 it will be transformed to the overflow entity as expected. So concretely, it's arbitrarily large dataloss for any new labelsets arriving between SanitizeLabels and OnAdd, which in practice is fast, but still - I should fix it.

modules/generator/localentitylimiter/local_entity_limiter.go
Logiraptor: Yeah I also don't like this name, any suggestions on what would fit better? What I'd really like is to have something like `service="*"` for span metrics, and something different for service graphs, but I haven't found the right way to make that clean.

modules/generator/localentitylimiter/local_entity_limiter.go
Logiraptor: What if I do the opposite: make the existing SanitizeLabels do the actual limiting, and then OnAdd is purely a lifecycle callback like OnUpdate? Does that give the best of both?

Logiraptor: Wait no, that still doesn't work with the series limiter. Series limiter is just an atomic int, so it can't do these more interesting things

modules/generator/localserieslimiter/local_series_limiter.go
joe-elliott: the individual metrics handle this logic by only calling "OnAdd" if they want to add a new series that they're not already tracking. so i think it would work?

basically if `OnAdd` is called and this would push us over our limits just return the dummy labels, else return the original labels. 

kind of responding to the below as well :).

modules/generator/localentitylimiter/local_entity_limiter.go
joe-elliott: `metric_overflow=true`? i do kind of hate that that "stutters" since its already a metric, but it works.

modules/generator/localserieslimiter/local_series_limiter.go
Logiraptor: Moved the logic into OnAdd. It's a bit more complicated since we need to check the series map twice, but I hid that in a helper function and it's not too bad. WDYT?

modules/generator/localentitylimiter/local_entity_limiter.go
```diff
@@ -77,33 +81,45 @@ func New(maxEntityFunc func(tenant string) uint32, tenant string, limitLogger *t
 		metricMaxActiveEntities:    metrics.maxActiveEntities.WithLabelValues(tenant),
 		metricTotalEntitiesAdded:   metrics.totalEntitiesAdded.WithLabelValues(tenant),
 		metricTotalEntitiesRemoved: metrics.totalEntitiesRemoved.WithLabelValues(tenant),
+
+		overflowEntity: labels.FromStrings("metric_overflow", "true"),
```
joe-elliott: make a var or const that both this and local series can use?

modules/generator/registry/builder.go
```diff
@@ -61,6 +66,7 @@ func (b *labelBuilder) Add(name, value string) {
 
 func (b *labelBuilder) CloseAndBuildLabels() (labels.Labels, bool) {
 	labels := b.builder.Labels()
+	b.demandTracker(labels.Hash())
```
joe-elliott: did something change here? there's probably an obvious reason for this i'm missing

Logiraptor: Thanks, now that the logic is in OnAdd, I no longer need to do this here. Removing!

modules/generator/localentitylimiter/local_entity_limiter.go
Logiraptor: Added registry.OverflowEntity

joe-elliott (commented):
oof, changelog please. i should have noticed last pass.

other than that we are good to merge.

joe-elliott (approved):

PR #5966, opened 2025-11-18
[registry.LabelBuilder] pool builders

Logiraptor:
This PR adds object pooling to the label builder code. It also defensively drops the reference to the pooled object once it is returned to the pool, to prevent memory corruption.

The allocations were the only thing left annoying me after #5953, so I can be at peace now 😄 

In a dev env, this reduced GC costs by around 50% and total CPU almost halved.

Screenshot 2025-11-18 at 3 47 52 PM
Screenshot 2025-11-18 at 3 42 11 PM
Screenshot 2025-11-18 at 3 41 22 PM

### DIFF
--- a/modules/generator/registry/builder.go
+++ b/modules/generator/registry/builder.go
@@ -6,11 +6,30 @@ import (
 	"github.com/prometheus/prometheus/model/labels"
 )
 
-var builderPool = sync.Pool{
-	New: func() interface{} {
-		return labels.NewBuilder(labels.New())
-	},
+type safeBuilderPool struct {
+	pool sync.Pool
 }
+
+func newSafeBuilderPool() *safeBuilderPool {
+	return &safeBuilderPool{
+		pool: sync.Pool{
+			New: func() interface{} {
+				return labels.NewBuilder(labels.New())
+			},
+		},
+	}
+}
+
+func (p *safeBuilderPool) Get() *labels.Builder {
+	return p.pool.Get().(*labels.Builder)
+}
+
+func (p *safeBuilderPool) Put(builder *labels.Builder) {
+	builder.Reset(labels.New())
+	p.pool.Put(builder)
+}
+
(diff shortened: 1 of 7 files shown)


modules/generator/registry/builder.go
```diff
@@ -28,6 +40,13 @@ func (b labelBuilder) Add(name, value string) {
 	b.builder.Set(name, value)
 }
 
-func (b labelBuilder) Labels() labels.Labels {
-	return b.builder.Labels()
+func (b *labelBuilder) Labels() labels.Labels {
+	labels := b.builder.Labels()
+	// it's no longer safe to use the builder after this point, so we drop our
+	// reference to it. this may cause a nil panic if the builder is used after
+	// this point, but it's better than memory corruption.
```
joe-elliott, line 65: so this func assumes that "Labels" is essentially the end of the lifecycle of the labelBuilder struct and it's about to be thrown away?

two nits:
- can we rename it something that indicates it destructive? `Close()`? `Complete()`?
- can we reset the builder here before putting it in the pool? currently this code relies on the Getter knowing to reset the label builder to prevent data leakage. if we reset the builder before we put it in the pool then the Getter can't make a mistake.

wdyt?

Logiraptor: Went explicit with `CloseAndBuildLabels`, and also wrapped the sync.Pool in a small struct so it's less likely to be misused.

modules/generator/registry/builder.go
```diff
@@ -1,6 +1,35 @@
...
+			},
+		},
+	}
+}
+
+func (p *safeBuilderPool) Get() *labels.Builder {
+	return p.pool.Get().(*labels.Builder)
+}
+
+func (p *safeBuilderPool) Put(builder *labels.Builder) {
```
joe-elliott, line 27: +1, i like this. was thinking of suggesting it

PR #5974, opened 2025-11-19
generator: add intrinsic_dimensions to user API

carles-grafana:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## main / unreleased
 
+* [CHANGE] Expose metrics_generator.intrinsic_dimensions to user-configurable overrides API [#5974](https://github.com/grafana/tempo/pull/5974) (@carles-grafana)
 * [CHANGE] Expose metrics_generator.trace_id_label_name to user-configurable overrides API [#5972](https://github.com/grafana/tempo/pull/5972) (@carles-grafana)
 * [CHANGE] Expose metrics_generator.ingestion_time_range_slack to user-configurable overrides API [#5958](https://github.com/grafana/tempo/pull/5958) (@carles-grafana)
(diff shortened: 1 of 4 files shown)


modules/generator/processor/spanmetrics/config.go
```diff
@@ -12,7 +12,16 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-var targetInfoIntrinsicLabels = append(validation.SupportedIntrinsicDimensions, processor.DimJob, processor.DimInstance)
+var targetInfoIntrinsicLabelsSet map[string]struct{}
```
electron0zero, line 15: I am bit confused by this change? and don't understand why?

I liked the []string as it was simple and reads better and doesn't need the setup we do in the `init()`

modules/generator/validation/sanitize.go
```diff
@@ -2,19 +2,18 @@ package validation
...
 
 type SanitizeFn func(string) string
 
-func SanitizeLabelNameWithCollisions(name string, dimensions []string, sanitizeFn SanitizeFn) string {
+func SanitizeLabelNameWithCollisions(name string, dimensions map[string]struct{}, sanitizeFn SanitizeFn) string {
 	sanitized := sanitizeFn(name)
 
 	// check if same label as intrinsics
-	if slices.Contains(dimensions, sanitized) {
+	if _, ok := dimensions[sanitized]; ok {
```
electron0zero, line 16: same question about map[string]struct{} instead of the []string in here as well.

modules/generator/processor/spanmetrics/config.go
carles-grafana: I've added SupportedIntrinsicDimensionsSet, following the same pattern as SupportedProcessorsSet and therefore doing the same with these. It's a small optimization to avoid traversing the slice everytime.

electron0zero: ahh, got it.

so we build a Set (using maps because go doesn't have a Set), and then check that instead of looping over the slice for each check.

PR #5980, opened 2025-11-19
[Metrics Generator] Handle invalid utf8

joe-elliott:
**What this PR does**:
Extends invalid utf8 handling from just span metrics to all metrics produced by the metrics generator. Also adds a new label to `tempo_metrics_generator_spans_discarded_total` to indicate the processor that dropped the spans. This is to differentiate between dropped data in the service graph processor vs span metrics processor. `invalid_utf8` is now the first drop reason that occurs independently in different processors. Note that in service graphs this metric is not truly 1:1 incremented with spans that are not counted, but it's directionally correct (and simple).

It's generally difficult to get invalid utf8 into these labels. If you use standard go gRPC tooling the client will fail when marshaling before even attempting to write to Tempo. However, specially crafted HTTP payloads with invalid data can be sent which creates these scenarios.

Finally, I decided to do the utf8 validation in `CloseAndBuildLabels()`. Originally I had the validation code outside this call, but since the two were so tightly coupled I decided to move it in, but this is up for discussion.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/registry/builder.go
+++ b/modules/generator/registry/builder.go
@@ -67,7 +67,6 @@ func (b *labelBuilder) CloseAndBuildLabels() (labels.Labels, bool) {
 	builderPool.Put(b.builder)
 	b.builder = nil
 
-	// Validate UTF-8. Return false if invalid.
 	if !labels.IsValid(model.UTF8Validation) {
 		return labels, false
 	}


modules/generator/registry/builder.go
```diff
@@ -58,13 +59,18 @@ func (b *labelBuilder) Add(name, value string) {
...
 	labels := b.builder.Labels()
 	// it's no longer safe to use the builder after this point, so we drop our
 	// reference to it. this may cause a nil panic if the builder is used after
 	// this point, but it's better than memory corruption.
 	builderPool.Put(b.builder)
 	b.builder = nil
 
-	return labels
+	// Validate UTF-8. Return false if invalid.
+	if !labels.IsValid(model.UTF8Validation) {
```
mattdurham, line 70: Looks good, eyeballed any other validations and looks like the only other one is legacy so this seems legit!

mattdurham (approved):
lgtm

PR #5981, opened 2025-11-19
Remove MustNewConstMetric and validate config in usage tracker

electron0zero:
**What this PR does**:

`prometheus.MustNewConstMetric` will panic when you try to create an metric with invalid label name. in prometheus world any labels that start with `__` are invalid because they are reserved.

registering same labels will also cause prometheus.MustNewConstMetric to panic.

you can configure overrides to map to invalid label names or use label names that are sanitized into `__` and it will cause distributors to panic when you scrape `/usage_metrics` to collect usage tracker metrics.

there are also ways to cause panic by pushing data into tempo that results in an invalid metric.

`prometheus.MustNewConstMetric` is designed to panic and is expected to panic when you have invalid config. 

in this case, we should use `prometheus.NewConstMetric` (it's what prometheus.MustNewConstMetric uses internally) and handle errors when we can't guarantee valid label names.

This PR replaces usage of `prometheus.MustNewConstMetric` in usage tracker with `prometheus.NewConstMetric` so never panic.

Along with that, this PR does few more things:

- validate per tenant runtime overrides and fail if overrides contains a usage tracker config that results in invalid label names (note: this will cause tempo to fail at startup with errors like `level=error ts=2025-11-19T18:23:43.440930939Z caller=app.go:231 msg="module failed" module=overrides-api err="failed to start overrides-api, because it depends on module overrides, which has failed: invalid service state: Failed, expected: Running, failure: starting module overrides: invalid service state: Failed, expected: Running, failure: failed to start subservices: not healthy, 0 terminated, 1 failed: [failed to load runtime config: load file: validating overrides for * failed: cost_attribution.dimensions has duplicate label name: 'name', both 'service.name' and 'span.http.target' map to it]"`
- validate usage tracker config in user configurable overrides API so we don't allow invalid config into tempo. this will reject al
- log the errors with a rate limited logger from `prometheus.NewConstMetric` in case of an invalid label name in usage tracker config.
- Add `tempo_distributor_usage_tracker_errors_total` counter to capture the errors with reason for the failures in the the usage tracker

<details>
<summary>More Details</summary>

I used a config like ⏬ to cause a panic:
```
    cost_attribution:
      dimensions:
        service.name: "__name__"
        span.http.target: "service_route"
```

here is panic log, I used local docker-compose in this test
```
tempo-1       | panic: "__name__" is not a valid label name for metric "tempo_usage_tracker_bytes_received_total"
tempo-1       |
tempo-1       | goroutine 7167 [running]:
tempo-1       | github.com/prometheus/client_golang/prometheus.MustNewConstMetric(...)
tempo-1       | 	/Users/suraj/wd/grafana/tempo/vendor/github.com/prometheus/client_golang/prometheus/value.go:129
tempo-1       | github.com/grafana/tempo/modules/distributor/usage.(*Tracker).Collect(0x4000e2ff80?, 0x401fc4d340)
tempo-1       | 	/Users/suraj/wd/grafana/tempo/modules/distributor/usage/tracker.go:397 +0x18c
tempo-1       | github.com/prometheus/client_golang/prometheus.(*Registry).Gather.func1()
tempo-1       | 	/Users/suraj/wd/grafana/tempo/vendor/github.com/prometheus/client_golang/prometheus/registry.go:458 +0xac
tempo-1       | created by github.com/prometheus/client_golang/prometheus.(*Registry).Gather in goroutine 7165
tempo-1       | 	/Users/suraj/wd/grafana/tempo/vendor/github.com/prometheus/client_golang/prometheus/registry.go:467 +0x4b4
```

</details>

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/usage/tracker.go
+++ b/modules/distributor/usage/tracker.go
@@ -108,7 +108,7 @@ func (t *tenantUsage) GetBuffersForDimensions(dimensions map[string]string) ([]m
 	for k, v := range dimensions {
 		// parse the dimensions key to get the attribute name and the scope
 		// and build the keys for the lookup
-		attr, scope := parseDimensionKey(k)
+		attr, scope := ParseDimensionKey(k)
 		dimensionsScope[k] = scope
 		dimensionsAttr[k] = attr
 
@@ -491,12 +491,12 @@ func protoLengthMath(x int) (n int) {
 	return 1 + (bits.Len64(uint64(x)|1)+6)/7
 }
 
-// parseAttribute take the dimensions key, and gives the scope and the attribute with scope trimmed
+// ParseDimensionKey take the dimensions key, and gives the scope and the attribute with scope trimmed
 //
 // we are not using traceql.ParseIdentifier because it needs `.` for unscoped attributes
 // and errors out for string without `.` or the scope prefix, also it supports other scopes that we don't support
 // in cost attribution config
-func parseDimensionKey(key string) (attribute string, scope Scope) {
+func ParseDimensionKey(key string) (attribute string, scope Scope) {
 	switch {
 	case strings.HasPrefix(key, "resource."):
 		return strings.TrimPrefix(key, "resource."), ScopeResource
--- a/modules/generator/validation/fields.go
+++ b/modules/generator/validation/fields.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/grafana/tempo/modules/distributor/usage"
 	"github.com/grafana/tempo/modules/generator/processor"
 	"github.com/grafana/tempo/modules/generator/registry"
 	"github.com/grafana/tempo/pkg/sharedconfig"
@@ -128,19 +129,20 @@ func ValidateCostAttributionDimensions(dimensions map[string]string) error {
 
 	// map is with key=tempo attribute, value=prometheus labelName
 	for k, v := range dimensions {
-		// build labelName the same way as usage.GetBuffersForDimensions in the usage tracker
+		// build labelName in the similar way as usage.GetBuffersForDimensions
+		attr, _ := usage.ParseDimensionKey(k) // extract attr so validate the duplicates with scope prefix
 		labelName := v
 		if labelName == "" {
-			labelName = k // The dimension is using default mapping, we map it to attribute
+			labelName = attr // The dimension is using default mapping, we map it to attribute
 		}
 		labelName = strutil.SanitizeFullLabelName(labelName) // sanitize label name
 
-		// Check for duplicate prometheus label names.
+		// check for duplicate prometheus label names.
 		// when we have duplicate labelNames, we randomly pick one so validate and don't allow duplicates.
 		if originalKey, exists := seenLabels[labelName]; exists {
 			return fmt.Errorf("cost_attribution.dimensions has duplicate label name: '%s', both '%s' and '%s' map to it", labelName, originalKey, k)
 		}
-		seenLabels[labelName] = k
+		seenLabels[labelName] = k // put k as value so we can show configured keys in the error
 
 		// creating a desc do the complete labelName validation
 		desc := prometheus.NewDesc("test_desc", "test desc created for validation", []string{labelName}, nil)
--- a/modules/distributor/usage/tracker_test.go
+++ b/modules/distributor/usage/tracker_test.go
@@ -601,7 +601,7 @@ func TestParseDimensionKey(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.input, func(t *testing.T) {
-			attr, scope := parseDimensionKey(tt.input)
+			attr, scope := ParseDimensionKey(tt.input)
 			require.Equal(t, tt.expectedAttr, attr)
 			require.Equal(t, tt.expectedScope, scope)
 		})
--- a/modules/generator/validation/fields_test.go
+++ b/modules/generator/validation/fields_test.go
@@ -142,7 +142,33 @@ func TestValidateCostAttributionDimensions(t *testing.T) {
 				"span.name": "name",
 			},
 			expErr:     true,
-			expErrText: "duplicate label name: 'name'",
+			expErrText: "cost_attribution.dimensions has duplicate label name: 'name', both",
+		},
+		{
+			name: "same attribute across two scopes in dimensions will error",
+			dimensions: map[string]string{
+				"resource.attribute": "",
+				"span.attribute":     "",
+			},
+			expErr:     true,
+			expErrText: "cost_attribution.dimensions has duplicate label name: 'attribute', both",
+		},
+		{
+			name: "same attribute scoped and unscoped scopes in dimensions will error",
+			dimensions: map[string]string{
+				"resource.attribute": "",
+				"attribute":          "",
+			},
+			expErr:     true,
+			expErrText: "cost_attribution.dimensions has duplicate label name: 'attribute', both",
+		},
+		{
+			name: "same attribute across two scopes in dimensions but with mapping",
+			dimensions: map[string]string{
+				"resource.attribute": "res_name",
+				"span.attribute":     "span_name",
+			},
+			expErr: false,
 		},
 	}
 


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
 * [ENHANCEMENT] Add partition ownership metric to live-store [#5815](https://github.com/grafana/tempo/pull/5815) (@javiermolinar, @mapno)
 * [ENHANCEMENT] Update list of intrinsics returned by search tags endpoint [#5857](https://github.com/grafana/tempo/pull/5857) (@andreasgerstmayr)
 * [ENHANCEMENT] Add "Requests Executed" panel for querier metrics in the operational dashboard. [#5848](https://github.com/grafana/tempo/pull/5848) (@anglerfishlyy)
+* [ENHANCEMENT] Remove MustNewConstMetric and validate config in usage tracker [#5981](https://github.com/grafana/tempo/pull/5981) (@electron0zero)
```
joe-elliott: way too much changelog churn. let's get it down to just the chagne you made

cmd/tempo/app/overrides_validation.go
```diff
@@ -36,6 +36,12 @@ func (r *runtimeConfigValidator) Validate(config *overrides.Overrides) error {
 		}
 	}
 
+	if config.CostAttribution.Dimensions != nil {
```
joe-elliott, line 46: does this allow the user config overrides to accept data that then prevents tempo from starting? i do find this concerning.

modules/distributor/usage/tracker.go
```diff
@@ -384,17 +392,25 @@ func (u *Tracker) Handler() http.Handler {
...
-	for _, t := range u.tenants {
+	for tenantID, t := range u.tenants {
 		for _, b := range t.series {
-			ch <- prometheus.MustNewConstMetric(b.descr, prometheus.CounterValue, float64(b.bytes), b.labels...)
+			metric, err := prometheus.NewConstMetric(b.descr, prometheus.CounterValue, float64(b.bytes), b.labels...)
+			// we have invalid config that's causing an error when registering a metric, log it and move on to next series
+			if err != nil {
+				// this will be logged for every invalid metric, for each scrape of `/usage_metrics`
+				// so we use a rate limited logger to avoid log span and perf overhead of logging
+				u.logger.Log("msg", "failed to collect usage tracker metric", "tenantID", tenantID, "err", err)
```
joe-elliott, line 421: should we log any details here like the data that may be causing the issue?

cmd/tempo/app/overrides_validation.go
electron0zero: that's on line 100 in same file as part of `overridesValidator`

this code is part of runtimeConfigValidator, and it's only used to validate the overrides loaded via the `per_tenant_override_config` config and only validates the per tenant overrides.

ps: it doesn't validate the overrides loaded by the `overrides.defaults` key.

modules/distributor/usage/tracker.go
electron0zero: we could log the dimensions configured by the user and `b.labels` that's causing the error

tenantID is enough to find the offending config but I guess more data is always useful.

CHANGELOG.md
electron0zero: fixed, my IDE's markdown formatter is too aggressive.

modules/distributor/usage/tracker.go
electron0zero: err will mention what caused the error, it will be something like `err="\"__name__\" is not a valid label name for metric \"tempo_usage_tracker_bytes_received_total\"`.

I added labels and dimensions and here is what log lines will look like with labels and dimensions.

```
level=error ts=2025-11-19T21:25:51.538038063Z caller=rate_limited_logger.go:38 component=usage-tracker msg="failed to collect usage tracker metric" tenantID=single-tenant dimensions="map[service.name:__name__ span.http.target:name]" labels=postgres;__missing__ err="\"__name__\" is not a valid label name for metric \"tempo_usage_tracker_bytes_received_total\""
level=error ts=2025-11-19T21:25:49.001481198Z caller=rate_limited_logger.go:38 component=usage-tracker msg="failed to collect usage tracker metric" tenantID=single-tenant dimensions="map[service.name:__name__ span.http.target:name]" labels=shop-backend;__missing__ err="\"__name__\" is not a valid label name for metric \"tempo_usage_tracker_bytes_received_total\""

```

modules/distributor/usage/tracker.go
```diff
@@ -384,17 +393,25 @@ func (u *Tracker) Handler() http.Handler {
...
-	for _, t := range u.tenants {
+	for tenantID, t := range u.tenants {
 		for _, b := range t.series {
-			ch <- prometheus.MustNewConstMetric(b.descr, prometheus.CounterValue, float64(b.bytes), b.labels...)
+			metric, err := prometheus.NewConstMetric(b.descr, prometheus.CounterValue, float64(b.bytes), b.labels...)
+			// we have invalid config that's causing an error when registering a metric, log it and move on to next series
+			if err != nil {
+				// this will be logged for every invalid metric, for each scrape of `/usage_metrics`
+				// so we use a rate limited logger to avoid log span and perf overhead of logging
+				u.logger.Log("msg", "failed to collect usage tracker metric", "tenantID", tenantID, "dimensions", fmt.Sprintf("%v", t.dimensions), "labels", strings.Join(b.labels, ";"), "err", err)
```
joe-elliott: do we need the custom formatting? what's the data look like if we pass the vars in directly? 

the reason i ask is b/c we're doing a lot of work here that will be thrown away a huge %age of the time due to the rate limited logger

modules/distributor/usage/tracker.go
joe-elliott: apologies, if the tenant id is all we need then it's fine to leave it how it was. i thought perahps the span data could play a role.

electron0zero: thinking more about this error. 

it's probably a good candidate for a counter metric.

metric can be useful in alerting on these errors.

cmd/tempo/app/overrides_validation.go
joe-elliott: sure, but the general question still stands. is it possible for tempo to fail to start b/c of user supplied config in a multi tenant environment? should it be that way?

modules/generator/validation/fields.go
```diff
@@ -104,3 +106,35 @@ func ValidateTraceIDLabelName(traceIDLabelName string) error {
...
+		// Check for duplicate prometheus label names, we silently pick one so validate and don't allow it.
+		if originalKey, exists := seenLabels[labelName]; exists {
+			return fmt.Errorf("cost_attribution.dimensions has duplicate label name: '%s', both '%s' and '%s' map to it", labelName, originalKey, k)
+		}
+		seenLabels[labelName] = k
+
+		// creating a desc just so can do the validation
+		desc := prometheus.NewDesc("test_desc", "test desc created for validation", []string{labelName}, nil)
+		// try to create a metric and see if there are any error, we use same method in usage.Collect
+		_, err := prometheus.NewConstMetric(desc, prometheus.CounterValue, float64(1), labelName)
```
carles-grafana, line 150: this seems redundant, I think essentially we only need to check that it's valid utf-8

electron0zero: > I think essentially we only need to check that it's valid utf-8

prometheus has support for utf-8 but it's not [there yet](https://prometheus.io/docs/guides/utf8/), so we still need to follow the Legacy validation to support most of the users.

in Legacy validation, prometheus is much more strict and won't allowed reserved labels (anything that starts with `__`), or something like `__name__` so I am using the prometheus's code to do the validation instead of re-implementing the validations.

also, you can't use the same label twice or it will err

cmd/tempo/app/overrides_validation.go
carles-grafana: I don't think it's possible. If the config is not valid, it's not written: https://github.com/grafana/tempo/blob/main/modules/overrides/userconfigurable/api/api.go#L96

modules/distributor/usage/tracker.go
electron0zero: without custom formatting logger failed to log the dimensions and labels and shows `unsupported value type` for both.

example log lines:
```
level=error ts=2025-11-20T13:11:53.533074579Z caller=rate_limited_logger.go:38 component=usage-tracker msg="failed to collect usage tracker metric" tenantID=single-tenant dimensions="unsupported value type" labels="unsupported value type" err="\"__name__\" is not a valid label name for metric \"tempo_usage_tracker_bytes_received_total\""
```

I think we can drop these attributes, they are not wroth extra overhead.

cmd/tempo/app/overrides_validation.go
electron0zero: > does this allow the user config overrides to accept data that then prevents tempo from starting?

nope, overrides API will not accept invalid config, and then the bad config it not loaded into tempo.


> is it possible for tempo to fail to start b/c of user supplied config in a multi tenant environment?

Yes if we add that bad config in the runtime overrides via the `per_tenant_override_config` file. but we never allowed bad config via the user configurable overrides API.

> should it be that way?

I think yes, you should not be allowed the load the bad config via `per_tenant_override_config` as well. In case of k8s deployments, your rollouts will save you but I prefer to fail fast instead of failing silently.

one thing to note is that `per_tenant_override_config` is only modified by the operators and not via overrides API so an operator has to put in the bad config, and in that case I think we should fail and let the operator know that config is bad.

modules/distributor/usage/tracker.go
electron0zero: Added `tempo_distributor_usage_tracker_errors_total`

cmd/tempo/app/overrides_validation.go
joe-elliott: i'll leave it to you two. i generally want to set the expectation that user supplied config should never be able to prevent tempo startup. we should always "fail open" in these cases.

modules/distributor/usage/tracker.go
joe-elliott: nice change. i like this 🙏

modules/generator/validation/fields.go
```diff
@@ -104,3 +106,35 @@ func ValidateTraceIDLabelName(traceIDLabelName string) error {
...
+	seenLabels := make(map[string]string)
+
+	// map is with key=tempo attribute, value=prometheus labelName
+	for k, v := range dimensions {
+		// build labelName the same way as usage.GetBuffersForDimensions in the usage tracker
+		labelName := v
+		if labelName == "" {
+			labelName = k // The dimension is using default mapping, we map it to attribute
+		}
+		labelName = strutil.SanitizeLabelName(labelName) // sanitize final label name the label name
```
carles-grafana: in GetBuffersForDimensions we use strutil.SanitizeFullLabelName, is this difference on purpose?

carles-grafana: we also use parseDimensionKey there

modules/generator/validation/fields.go
carles-grafana: My point is that creating the metric here is unnecessary, because looking at the function definition the only relevant validation is UTF-8:

```
func NewConstMetric(desc *Desc, valueType ValueType, value float64, labelValues ...string) (Metric, error) {
	if desc.err != nil {
		return nil, desc.err
	}
	if err := validateLabelValues(labelValues, len(desc.variableLabels.names)); err != nil {
		return nil, err
	}

	metric := &dto.Metric{}
	if err := populateMetric(valueType, value, MakeLabelPairs(desc, labelValues), nil, metric, nil); err != nil {
		return nil, err
	}

	return &constMetric{
		desc:   desc,
		metric: metric,
	}, nil
}
```

We only care about validating the label name, and we can see that the only relevant validation is the UTF-8, so we just need to check that and we avoid doing extra work:

```
func validateLabelValues(vals []string, expectedNumberOfValues int) error {
	if len(vals) != expectedNumberOfValues {
		// The call below makes vals escape, copy them to avoid that.
		vals := append([]string(nil), vals...)
		return fmt.Errorf(
			"%w: expected %d label values but got %d in %#v",
			errInconsistentCardinality, expectedNumberOfValues,
			len(vals), vals,
		)
	}

	for _, val := range vals {
		if !utf8.ValidString(val) {
			return fmt.Errorf("label value %q is not valid UTF-8", val)
		}
	}

	return nil
}
```

As far as I know the version of the Prometheus library we use uses UTF-8 validation by default, not legacy.

modules/generator/validation/fields.go
carles-grafana: It's a bit hard to follow if we are doing the same validation as GetBuffersForDimensions. Can it be done more similarly?

electron0zero: > in GetBuffersForDimensions we use strutil.SanitizeFullLabelName, is this difference on purpose?

good catch, nope, it should be `strutil.SanitizeFullLabelName`

> parseDimensionKey

we do that to parse the scope and attribute name from the traces attributes. `parseDimensionKey` doesn't do much because it would either trim the `span.` or `resource.` or return the input as is so it doesn't add anything in the validation.

In this case, we want to validate the labels we will use the Prometheus metrics,  they are either in the mapping, and if not, we will use the sanitised version of traces attribute name if the mapping is not supplied.


> It's a bit hard to follow if we are doing the same validation as GetBuffersForDimensions. Can it be done more similarly?

It's similar but not same, I will reword the comment to make it clear.

modules/generator/validation/fields.go
electron0zero: Checking if the label is Valid UTF-8 string will not catch it things like `__name__` and labels that start with `__`, we need to rejected those as well.

If you replace this metric creation code with just utf-8 validation, and you will see that we will allow bunch of invalid label names.

one easy way to see it is to comment out the `NewConstMetric` the code and add the  UTF-8 Validation, and run the `TestValidateCostAttributionDimensions` unit test.

Creating the metrics allows us to reusing the validations that are sprinkled across `client_golang` code without re-implementing them.

we will need to also do the validations that are done in https://github.com/prometheus/client_golang/blob/c316de06dea0caf0631f9eca944bc74cca1eb19c/prometheus/desc.go#L92 if want to avoid using `NewConstMetric`

`NewConstMetric` should be fine since we only do call this code path on overrides config reload or when we get call the overrides API and modify the cost attribution configs. because of that, I am not too worried about performance of this code.

also, NewConstMetric just creates an ad-hoc metric so it's not registered or exposed anywhere, and won't be collected as well.

modules/generator/validation/fields.go
carles-grafana: > parseDimensionKey doesn't do much because it would either trim the span. or resource. or return the input as is so it doesn't add anything in the validation.

but what if the same attribute is in both scopes? what is supposed to happen?

Issue #4623, opened 2025-01-27
TraceQL doesn't parse `math.MinInt` properly


ndk:
**Describe the bug**
When adding test cases to [`TestSpansetFilterStatics`](https://github.com/grafana/tempo/blob/f895567299dd0dd012726dd9e5d20d84350882a2/pkg/traceql/parse_test.go#L852):
```golang
func TestSpansetFilterStatics(t *testing.T) {
	tests := []struct {
		in       string
		expected FieldExpression
	}{
		...
		{in: "{ 9223372036854775807 }", expected: NewStaticInt(math.MaxInt)},
		{in: "{ -9223372036854775808 }", expected: NewStaticInt(math.MinInt)},
```

the second case (`-9223372036854775808`) triggers an out-of-range error:
```
Error:      	Received unexpected error:
            	parse error at line 1, col 4: strconv.Atoi: parsing "9223372036854775808": value out of range
Test:       	TestSpansetFilterStatics/{_-9223372036854775808_}
```
It appears the parser handles the numeric part (`9223372036854775808`) as a positive integer before applying the minus sign, causing the overflow.

The relevant parsing rule is in [expr.y](https://github.com/grafana/tempo/blob/f895567299dd0dd012726dd9e5d20d84350882a2/pkg/traceql/expr.y#L280):
```yacc
scalarExpression: // shares the same operators as scalarPipelineExpression. split out for readability
    OPEN_PARENS scalarExpression CLOSE_PARENS  { $$ = $2 }                                   
  | scalarExpression ADD scalarExpression      { $$ = newScalarOperation(OpAdd, $1, $3) }
  | scalarExpression SUB scalarExpression      { $$ = newScalarOperation(OpSub, $1, $3) }
  | scalarExpression MUL scalarExpression      { $$ = newScalarOperation(OpMult, $1, $3) }
  | scalarExpression DIV scalarExpression      { $$ = newScalarOperation(OpDiv, $1, $3) }
  | scalarExpression MOD scalarExpression      { $$ = newScalarOperation(OpMod, $1, $3) }
  | scalarExpression POW scalarExpression      { $$ = newScalarOperation(OpPower, $1, $3) }
  | aggregate                                  { $$ = $1 }
  | INTEGER                                    { $$ = NewStaticInt($1)              }
  | FLOAT                                      { $$ = NewStaticFloat($1)            }
  | DURATION                                   { $$ = NewStaticDuration($1)         }
[...]

PR #5982, opened 2025-11-19
traceql: fix signed-int parsing to correctly handle math.MinInt64 (issue #4623)

Syedowais312:
# 🔧 Fix: Proper Handling of Signed Integers in TraceQL Lexer  
Closes: #4623

## Summary

This PR fixes incorrect parsing of signed integers in the TraceQL lexer.

Previously, expressions such as:

    { -9223372036854775808 }
    { +10 }
    { -5s }

were incorrectly parsed or caused unexpected failures due to improper unary sign handling and integer overflow behavior.

### ✔ 1. Correct signed integer parsing
- Supports unary `+` and `-` in valid contexts.
- Handles the `MinInt64` edge case (-9223372036854775808) safely.
- Validates numeric ranges to prevent silent overflows.

### ✔ 2. Correct signed duration parsing
Signed durations are now parsed consistently:

    -5s
    +10ms

using the existing `tryScanDuration` logic.

### ✔ 3. Fixes unary parsing context rules
Unary signs are now only allowed:
- at expression start,
- after `{`,
- after comparison operators,
- after commas,
- after unary `!`.

(Prevents incorrect parsing inside function argument lists.)

## Tests

Added and updated tests in `parse_test.go`:
- `{ -9223372036854775808 }`
- `{ +9223372036854775807 }`
- `{ -5s }`
- `{ +10ms }`

All related TraceQL tests pass.

## Why This Fix?

The previous logic failed in edge cases and mixed incorrectly with duration parsing.  
This fix is:
- simple,
- safe,
- correct,
- aligned with TraceQL grammar expectations.

## Result

TraceQL signed integer parsing is now robust, predictable, and fully tested.

## 🙏 Notes

If additional signed numeric types (float, duration, hex, etc.) need first-class support, I’d be happy to extend this work.

### DIFF
--- a/modules/frontend/mcp_tools_test.go
+++ b/modules/frontend/mcp_tools_test.go
@@ -101,7 +101,7 @@ func TestHandleSearch(t *testing.T) {
 				"query": "{ foo bar baz }",
 			}),
 			expected: expectedResult{
-				err: "query parse error. Consult TraceQL docs tools: parse error at line 1, col 3: syntax error: unexpected IDENTIFIER",
+				err: "query parse error. Consult TraceQL docs tools: parse error at line 1, col 3: unknown identifier: foo",
 			},
 		},
 		{
@@ -181,7 +181,7 @@ func TestHandleInstantQuery(t *testing.T) {
 				"query": "{ foo bar baz }",
 			}),
 			expected: expectedResult{
-				err: "query parse error. Consult TraceQL docs tools: parse error at line 1, col 3: syntax error: unexpected IDENTIFIER",
+				err: "query parse error. Consult TraceQL docs tools: parse error at line 1, col 3: unknown identifier: foo",
 			},
 		},
 		{
@@ -261,7 +261,7 @@ func TestHandleRangeQuery(t *testing.T) {
 				"query": "{ foo bar baz }",
 			}),
 			expected: expectedResult{
-				err: "query parse error. Consult TraceQL docs tools: parse error at line 1, col 3: syntax error: unexpected IDENTIFIER",
+				err: "query parse error. Consult TraceQL docs tools: parse error at line 1, col 3: unknown identifier: foo",
 			},
 		},
 		{


Syedowais312:
Hi maintainers!  @joe-elliott 
This PR fixes issue #4623 regarding correct parsing of math.MinInt64.  
All local tests pass. Please let me know if any changes are needed!

joe-elliott:
So this is quite a lot of complex code for such an intense edge case. Tests should keep it together, but we still need to weigh the value of the feature against the complexity of the code.

What do you think about an alternative solution like adding constants like `minInt`, `maxInt`? Could we do this in a cleaner way?

Syedowais312:
> So this is quite a lot of complex code for such an intense edge case. Tests should keep it together, but we still need to weigh the value of the feature against the complexity of the code.
> 
> What do you think about an alternative solution like adding constants like `minInt`, `maxInt`? Could we do this in a cleaner way?
Hey @joe-elliott 👋

Thanks for the feedback!

I've updated the PR to use the simpler and cleaner approach you suggested.  
Instead of complex unary parsing logic, I added built-in integer constants:

    minInt → math.MinInt
    maxInt → math.MaxInt

This keeps the lexer minimal, avoids lookahead complexity, and removes the risk of overflow handling code.  
I’ve also updated the tests to use `{ minInt }` and `{ maxInt }`, and all TraceQL tests are passing.

Whenever you get time, please take another look and let me know if anything else needs refinement. Happy to iterate on it!

Syedowais312:
Thanks for the review, @joe-elliott 
implementing `minInt`  /  `maxInt`  directly in the parser makes complete sense. It keeps the lexer simpler and aligns better with how other literals are handled


I'll update the PR shortly with:
-`minInt`  /  `maxInt` evaluate inside the parser
-The corresponding changelog entry
-Updated tests to match the parser-based approach

Thanks again for the guidance - I'll ping you once the new changes are pushed

Syedowais312:
Thanks for the review @joe-elliott!

I’ve updated the PR with the requested changes. The constant handling for `minInt`
and `maxInt` is now implemented cleanly in the parser, directly inside the `integer`
rule as suggested.

All generated parser files were updated using `make gen-traceql-local` and I’ve added
the corresponding tests and CHANGELOG entry.

Local build & full TraceQL test suite also pass successfully.

Please let me know if you’d like any follow-up changes!

pkg/traceql/lexer.go
```diff
@@ -247,6 +253,12 @@ func (l *lexer) Lex(lval *yySymType) int {
 		return tok
 	}
 
+	// builtin integer constants (e.g. minInt, maxInt)
+	if value, ok := builtinIntegerConstants[l.TokenText()]; ok {
```
joe-elliott: is this code still necessary?

Syedowais312: Thanks for the follow-up
This block is now no longer needed since `minInt` and `maxInt` are handled directly in the parser via the updated `integer` rule. The lexer no longer needs to check for builtin integer constants

I'll remove this secction and push an update.

Please let me know if you'd like me to include anything else in this cleanup!

Syedowais312:
@joe-elliott 

I’ve moved the minInt/maxInt handling into the parser and removed the redundant lexer logic. The branch is now rebased and ready for another look.

CHANGELOG.md
```diff
@@ -1,8 +1,5 @@
 ## main / unreleased
 
-* [CHANGE] Add stricter validation to user-configurable overrides API, such as refusing duplicate labels [#6008](https://github.com/grafana/tempo/pull/6008) (@carles-grafana)
```
joe-elliott, line 3: deleted changelog entries?

Syedowais312:
Thanks for pointing this out 
I've reset `CHANGELOG.md`  to upstream, restored the removed entries, and added a single changing entry for this TraceQL change under `main / unreleased` .

Please let me know if the wording looks good.

joe-elliott:
Seeing some CI issues

- Don't worry about the linter. I think it's a bug or perhaps a miss configuration on our part. Those are invalid errors.
- The unit test failures need to be corrected. You changed the error return string which is breaking some tests. I prefer your error return so let's keep it and just update the tests to match the new strings.
- The vendor check is due to unexpected differences in the compiled .y file. This might be due to a version difference between what's on your machine and what the CIs using. If you are unable to correct I should be able to rebuild the file and push a commit. `make gen-traceql` should pull a docker container and use the same version as CI.

Syedowais312:
Thanks for the detailed review — this is very helpful.

Acknowledged on all points 👍

• I’ll revert the error return strings back to their original values and update the implementation so existing unit tests continue to pass without modifying expected messages.

• I’ll investigate the vendor diff. Most likely this is coming from a generator/version mismatch on my local setup. I’ll try regenerating via `make gen-traceql` using the same tooling version as CI.  
If I still can’t get a clean diff, I’ll leave the generated file unchanged and only include the minimal necessary changes.

• Understood on the linter errors — I’ll ignore those for now as suggested.

I’ll push an updated commit shortly addressing the test failures and vendor differences.  
Please let me know if there’s anything else you’d like me to be careful about here.

Thanks again for the guidance!

joe-elliott:
> • I’ll revert the error return strings back to their original values and update the implementation so existing unit tests continue to pass without modifying expected messages.

No, please keep your error strings. Update the tests to match.

Syedowais312:
@joe-elliott 
Thanks for the clarification 👍  
I’ve kept the new error strings, updated the affected tests, and regenerated the TraceQL files using the CI image.  
Fixed remaining frontend unit tests to align with updated TraceQL parse error messages.
All relevant unit tests now pass locally.

Syedowais312:
Thanks for the review and guidance! learned alot through parser and test expectation. Looking forward to contribute more

PR #5983, opened 2025-11-19
[jsonnet/microservices]: use upstream rollout-operator

zalegrala:
**What this PR does**:

Without this change, we rely on a CRD which is not installed by the jsonnet.  Using the [upstream](https://github.com/grafana/rollout-operator/tree/main/operations/rollout-operator) jsonnet for the rollout-operator allows us to move with the community and install the CRDs into the cluster when necessary.

Additional changes:
* add jsonnetfile.json to microservices for dependency tracking
* Include ksonnet-util dependency
* Include memcahced dependency
* Include rollout-operator dependency
* Update k8s lib version in tests

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


operations/jsonnet/microservices/rollout-operator.libsonnet
```diff
@@ -1,256 +0,0 @@
...
-  local serviceAccount = k.core.v1.serviceAccount,
-  local servicePort = k.core.v1.servicePort,
-  local validatingWebhook = k.admissionregistration.v1.validatingWebhook,
-  local validatingWebhookConfiguration = k.admissionregistration.v1.validatingWebhookConfiguration,
-  local podDisruptionBudget = k.policy.v1.podDisruptionBudget,
-
-  local replica_template = (import 'replica-template.libsonnet').replica_template,
-
-  _config+:: {
-    rollout_operator_enabled: $._config.multi_zone_ingester_enabled || $._config.block_builder_concurrent_rollout_enabled || $._config.live_store.replicas > 0,
```
zalegrala, line 23: These enablement flags need some consideration.  I expect that live-store will not be optional for much longer.

zalegrala: I'd like to hear if folks think we should make this optional given that we will start requiring it soon.

PR #5989, opened 2025-11-20
generator: add dimension mappings to user API

carles-grafana:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/overrides/userconfigurable/api/limits_test.go
+++ b/modules/overrides/userconfigurable/api/limits_test.go
@@ -40,7 +40,7 @@ func Test_limitsFromOverrides(t *testing.T) {
 					SpanMetrics: overrides.SpanMetricsOverrides{
 						Dimensions:          []string{"your-dim1", "your-dim2"},
 						IntrinsicDimensions: map[string]bool{"service": true, "span_name": false},
-						DimensionMappings:   []sharedconfig.DimensionMappings{{Name: "env", SourceLabel: []string{"k8s.namespace"}, Join: ""}},
+						DimensionMappings:   []sharedconfig.DimensionMappings{{Name: "env", SourceLabel: []string{"k8s.namespace", "foo"}, Join: "/"}},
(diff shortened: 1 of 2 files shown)


modules/overrides/userconfigurable/client/limits_test.go
```diff
@@ -51,6 +52,15 @@ func TestLimits_parseJson(t *testing.T) {
         "intrinsic_dimensions": {
           "service": true
         },
+        "dimension_mappings": [
```
electron0zero, line 55: can we add test that has multiple source_labels and make use of `join` or maybe modify this to add those.

I want to ensure we test both cases:
- case 1: plain remapping, one source label and empty join
- case 2: join and remapping, two source labels, join is not empty

modules/overrides/userconfigurable/api/limits_test.go
```diff
@@ -39,6 +40,7 @@ func Test_limitsFromOverrides(t *testing.T) {
 					SpanMetrics: overrides.SpanMetricsOverrides{
 						Dimensions:          []string{"your-dim1", "your-dim2"},
 						IntrinsicDimensions: map[string]bool{"service": true, "span_name": false},
+						DimensionMappings:   []sharedconfig.DimensionMappings{{Name: "env", SourceLabel: []string{"k8s.namespace"}, Join: ""}},
```
electron0zero: let's add multiple source labels and set the join so we are testing all configs.

modules/overrides/userconfigurable/client/limits_test.go
carles-grafana: done

modules/overrides/userconfigurable/api/limits_test.go
carles-grafana: done

electron0zero (approved):

PR #6004, opened 2025-11-25
Fixes for intrinsic = nil and other stuff != nil

### LINKED ISSUE #5998: != nil checks broken in TraceQL Metrics
**Describe the bug**

`!= nil` show some strange behaviors with dedicated and well known columns when used in TraceQL metrics queries. The behavior appears to be the same at resource and span scopes but this should be double checked.

#### `!= nil` on well known column
Incorrectly returns nil but only for recent data. Works on backend blocks.

Image

#### `!= nil` on dedicated column
Incorrectly includes nil for all data.

Image

### LINKED ISSUE #5999: Fix = nil handling with intrinsic columns
`= nil` is currently conflated with the zero value for intrinsic columns.

```
{ span:name = nil }
```

is evaluated as:

```
{ span:name = "" }
```

This was [done purposefully](https://github.com/grafana/tempo/blob/4b613b698183ef258e5ef3cb576daac330074813/tempodb/encoding/vparquet5/block_traceql.go#L1911-L1927) but should be removed and instead we should return an "invalid query" error.

Intrinsics like `span:name` can never be nil by definition. All spans have a value for these fields. Only custom attributes can be nil.

mdisibio:
**What this PR does**:
This PR fixes general handling of =nil and !=nil for intrinsics, and metrics queries.  

1.  As discussed offline, the best way to handle a query with `intrinsic=nil` is to not run it at all.  These are now validation errors and return a message like in the screenshot below.  When these are present in autocomplete requests, we silently return nothing.  Internally this is implemented as a new query validation check inside autocomplete.
 
Alternatively we also considered running the query acting like these clause are always false (they are), but there are other behavior to satisfy like how we return all attributes referenced in the query, impacts in tag autocomplete/etc.  Overall not worth it for now, and since we just introduced the =nil feature recently anyway, this breaking change should have minimal impact.  This also gives us an opening to re-allow these queries in the future after we define the behavior.

image

4. Metrics queries - This is the majority of the change.  The issue was that these 2 queries were identical at the fetch layer: 

```
{ well.known.column != nil } | rate() by (well.known.column)
{                          } | rate() by (well.known.column)
```

They both executed the fetch `OpNone` which is **no** filtering on the column, which means all values even nils were fetched.  The first query relies on the engine to do the last filter step, but the second query requires no engine work. Metrics query optimization wasn't calling the engine because it is tuned for the second case.

How do solve this?

Instead of slowing metrics down by always calling the engine, we need to fetch better.   Now `attr != nil` is pushed down as `OpExists`, and the first query won't return nils, and requires no engine.  The language parser was already converting every `attr != nil` statement as that, we just needed to honor it.

This required a decent number of changes in every vp2/3/4/5 folder, but I tried to minimize it.  Happy for suggestions on ways to further simplify.

Screenshots:

(a) correctly filtering out spans w/o http.status_code (no longer returning nil):
image

(b) Filtering correctly on resource.pod. I don't have any in my data and it correctly returns nothing:

image

Fixes #5998
Fixes #5999

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/encoding/vparquet5/block_traceql.go
+++ b/tempodb/encoding/vparquet5/block_traceql.go
@@ -2338,19 +2338,7 @@ func createResourceIterator(makeIter, makeNilIter makeIterFn, instrumentationIte
 	for _, cond := range conditions {
 		// Well-known selector?
 		if entry, ok := wellKnownColumnLookups[cond.Attribute.Name]; ok && entry.level != traceql.AttributeScopeSpan {
-			// Operands that need special handling.
-			switch cond.Op {
-			case traceql.OpNone:
-				addPredicate(entry.columnPath, nil) // No filtering
-				columnSelectAs[entry.columnPath] = cond.Attribute.Name
-				continue
-			case traceql.OpExists:
-				addPredicate(entry.columnPath, &parquetquery.SkipNilsPredicate{})
-				columnSelectAs[entry.columnPath] = cond.Attribute.Name
-				continue
-			case traceql.OpNotExists:
-				pred := parquetquery.NewNilValuePredicate()
-				iters = append(iters, makeIter(entry.columnPath, pred, cond.Attribute.Name))
+			if specialCase(cond, entry.columnPath) {
 				continue
 			}
 


tempodb/encoding/vparquet5/block_traceql.go
```diff
@@ -2345,26 +2316,39 @@ func createResourceIterator(makeIter, makeNilIter makeIterFn, instrumentationIte
...
+		default:
+			return false
+		}
+	}
 
+	for _, cond := range conditions {
 		// Well-known selector?
 		if entry, ok := wellKnownColumnLookups[cond.Attribute.Name]; ok && entry.level != traceql.AttributeScopeSpan {
-			if cond.Op == traceql.OpNone {
+			// Operands that need special handling.
```
ie-pham: nit: any reason why we're not using the specialCase function like the other cases?

mdisibio: Good catch, fixed.

PR #6010, opened 2025-11-27
Fix panic when topk() reduces series below maxSeries in QueryRange

### LINKED ISSUE #6000: Metrics query panics if using topK() and the internal series limit is reached
**Describe the bug**
A panic occurs in the querier if a metrics query hits the internal series limit and is also using topk()/bottomk().   Tempo has a server-side series limit that controls the maximum number of series returned from a metrics query, for example 10K series.  This can be triggered by a high cardinality query like `{ } | rate() by (some.uuid)`.  If you then try to reduce this, for example to the top 10 using `{ } | rate() by (some.uuid) | topk(10)`, it panics.

Stack trace:
```
http: panic serving ...: runtime error: slice bounds out of range [:10000] with capacity 10
goroutine 422519998 [running]:
net/http.(*conn).serve.func1()
	/opt/hostedtoolcache/go/1.25.1/x64/src/net/http/server.go:1943 +0xd3
github.com/grafana/tempo/modules/frontend/combiner.NewQueryRange.func3(0x483c8e?)
traces/vendor/github.com/grafana/tempo/modules/frontend/combiner/metrics_query_range.go:52 +0x113
```

The line in question is:

```go
if combiner.MaxSeriesReached() {
  // Truncating the final response because even if we bail as soon as len(resp.Series) >= maxSeries
  // it's possible that the last response pushed us over the max series limit.
  resp.Series = resp.Series[:maxSeries]
```

In this panic maxSeries=10000 and topk=10, and the bug is that we are reslicing the last 10000 entries, but there are only 10 since this happens after topk.

**To Reproduce**

**Expected behavior**

**Environment:**

**Additional Context**

Syedowais312:
Fixes #6000

# Summary

Fixes a crash in QueryRange when the internal series limit is reached while using `topk()` or ` bottomk()`.
These functions reduce the series count before truncation, causing an out-of-bounds slice.

## Root Cause

`MaxSeriesReached()` triggers truncation using:

``resp.Series = resp.Series[:maxSeries]
``

But after topk(10), the response may contain fewer series than maxSeries, leading to:

``panic: slice bounds out of range
``

## Fix

Added a conditional check:

```
if len(resp.Series) > maxSeries {
    resp.Series = resp.Series[:maxSeries]
}
```

This prevents slicing beyond available series.

## Tests

- Verified build: go build ./pkg/traceql
- Ran combiner tests: go test ./modules/frontend/combiner
- Tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@
 * [ENHANCEMENT] Add support for application/protobuf in frontend endpoints [#5865](https://github.com/grafana/tempo/pull/5865) (@oleg-kozliuk-grafana)
 * [ENHANCEMENT] Validate metrics-generator histogram buckets [#5991](https://github.com/grafana/tempo/pull/5991) (@carles-grafana)
 * [ENHANCEMENT] Removed MustNewConstMetric to prevent panic and added validation for usage tracker config. Added `tempo_distributor_usage_tracker_errors_total` to surface errors in usage tracker. [#5981](https://github.com/grafana/tempo/pull/5981) (@electron0zero)
-* [BUGFIX] Prevent slice panic when truncating series after topk() by adding bounds check in metrics query-range combiner
+* [BUGFIX] Prevent slice panic when truncating series after topk() by adding bounds check in metrics query-range combiner [#6010](https://github.com/grafana/tempo/pull/6010) (@Syedowais312)
 * [BUGFIX] Fix compactor to properly consider SSE-KMS information during metadata copy [#5774](https://github.com/grafana/tempo/pull/5774) (@steffsas)
 * [BUGFIX] Fix incorrect results in TraceQL compare() caused by potential hash collision of string array attributes [#5835](https://github.com/grafana/tempo/pull/5835) (@mdisibio)
 * [BUGFIX] Correctly track and reject too large traces in live stores. [#5757](https://github.com/grafana/tempo/pull/5757) (@joe-elliott)


mdisibio:
Thanks for taking care of this.  Fix looks good, can you add a changelog.md entry? And we'll merge.

Syedowais312:
> Thanks for taking care of this. Fix looks good, can you add a changelog.md entry? And we'll merge.

Changelog entry added. Let me know if anything else is needed!
Thank you for the review!

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@
 * [ENHANCEMENT] Add support for application/protobuf in frontend endpoints [#5865](https://github.com/grafana/tempo/pull/5865) (@oleg-kozliuk-grafana)
 * [ENHANCEMENT] Validate metrics-generator histogram buckets [#5991](https://github.com/grafana/tempo/pull/5991) (@carles-grafana)
 * [ENHANCEMENT] Removed MustNewConstMetric to prevent panic and added validation for usage tracker config. Added `tempo_distributor_usage_tracker_errors_total` to surface errors in usage tracker. [#5981](https://github.com/grafana/tempo/pull/5981) (@electron0zero)
+* [BUGFIX] Prevent slice panic when truncating series after topk() by adding bounds check in metrics query-range combiner
```
mdisibio:
```suggestion
* [BUGFIX] Prevent slice panic when truncating series after topk() by adding bounds check in metrics query-range combiner [#6010](https://github.com/grafana/tempo/pull/6010) (@Syedowais312)
```

mdisibio (approved):

PR #6011, opened 2025-11-28
Fix delete implementation for s3/gcs/azure backends to account for prefix

### LINKED ISSUE #5993: Tempo doesn't seem to delete indexes for tenants with empty blocklist if S3 bucket is configured with a prefix
**Describe the bug**
Tempo documentation states that for tenants with empty blocklists, the compactor will delete the tenant indexes as part of the poll. However, we are seeing that on our deployments where the S3 backend is configured with a prefix, the tenant indexes are never deleted.

It looks like the [S3 readerWriter's Delete](https://github.com/grafana/tempo/blob/main/tempodb/backend/s3/s3.go#L346) implementation doesn't prepend the configured prefix while generating the index filepath for deletion. 
```
func (rw *readerWriter) Delete(ctx context.Context, name string, keypath backend.KeyPath, _ *backend.CacheInfo) error {
	filename := backend.ObjectFileName(keypath, name)
	return rw.core.RemoveObject(ctx, rw.cfg.Bucket, filename, minio.RemoveObjectOptions{})
}
```
As a result, when the poller is attempting to [write the tenant index](https://github.com/grafana/tempo/blob/main/tempodb/blocklist/poller.go#L303C17-L303C33) for tenants with empty blocklist, the backend writer is unable to [delete the index](https://github.com/grafana/tempo/blob/main/tempodb/backend/raw.go#L135) because of the missing prefix in index filepath. Also, it seems this would result in silent exit since invalid file path could be considered as a non-existent/deleted file. 

**To Reproduce**
1. Configue Tempo (we are using 2.7) with S3 backend and prefix
2. Start Tempo
3. Publish Traces for a Tenant
4. Let the blocklist be cleared post retention in S3
5. The index for this tenant is not deleted

**Expected behavior**
The indexes for tenants with empty blocklists are deleted if S3 backend is configured with a prefix.

**Environment:**

**Additional Context**

kaustubhkurve:
**What this PR does**:
The PR updates the delete implementation for s3/gcs/azure backends to account for prefix if it is configured for the storage backend.

This fixes the issue where tenant indexes are never deleted if a prefix is used. When the poller attempts to write the tenant index for tenants with empty blocklist, the backend writer is unable to delete the index  because of the missing prefix in index filepath.

Fixes #5993 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/backend/s3/s3.go
+++ b/tempodb/backend/s3/s3.go
@@ -519,7 +519,7 @@ func (rw *readerWriter) Find(ctx context.Context, keypath backend.KeyPath, f bac
 			if len(res.Contents) > 0 {
 				for _, c := range res.Contents {
 					opts := backend.FindMatch{
-						Key:      strings.TrimPrefix(c.Key, prefix),
+						Key:      strings.TrimPrefix(c.Key, rw.cfg.Prefix),
 						Modified: c.LastModified,
 					}
 					f(opts)
--- a/tempodb/backend/azure/azure.go
+++ b/tempodb/backend/azure/azure.go
@@ -288,7 +288,7 @@ func (rw *Azure) Find(ctx context.Context, keypath backend.KeyPath, f backend.Fi
 
 			o = strings.TrimSuffix(*b.Name, dir)
 			opts := backend.FindMatch{
-				Key:      strings.TrimPrefix(o, prefix),
+				Key:      strings.TrimPrefix(o, rw.cfg.Prefix),
 				Modified: *b.Properties.LastModified,
 			}
 			f(opts)
--- a/tempodb/backend/gcs/gcs.go
+++ b/tempodb/backend/gcs/gcs.go
@@ -350,7 +350,7 @@ func (rw *readerWriter) Find(ctx context.Context, keypath backend.KeyPath, f bac
 		}
 
 		opts := backend.FindMatch{
-			Key:      strings.TrimPrefix(attrs.Name, prefix),
+			Key:      strings.TrimPrefix(attrs.Name, rw.cfg.Prefix),
 			Modified: attrs.Updated,
 		}
 		f(opts)
--- a/tempodb/blocklist/poller.go
+++ b/tempodb/blocklist/poller.go
@@ -576,7 +576,7 @@ func (p *Poller) deleteTenant(ctx context.Context, tenantID string) error {
 	for _, object := range foundObjects {
 		dir, name := path.Split(object)
 		level.Info(p.logger).Log("msg", "deleting", "tenant", tenantID, "object", object)
-		err = p.writer.Delete(ctx, name, backend.KeyPath{tenantID, dir})
+		err = p.writer.Delete(ctx, name, backend.KeyPath{dir})
 		if err != nil {
 			return err
 		}


joe-elliott:
Thanks for the fix! It seems like this test is failing:

https://github.com/grafana/tempo/blob/main/integration/poller/poller_test.go#L214

Can you review?

kaustubhkurve:
Thanks @joe-elliott, I have pushed a fix for the test failure.

It seems the test was failing due to the presence of prefix in the [readWriter.Find function result set](https://github.com/grafana/tempo/blob/2b4f143081d708c316ac00cc062f0c2b0053f54e/tempodb/backend/s3/s3.go#L522).
 
The delete function is being called from two places: [backend.writer.WriteTenantIndex](https://github.com/grafana/tempo/blob/2b4f143081d708c316ac00cc062f0c2b0053f54e/tempodb/backend/raw.go#L137) and [poller.deleteTenant](https://github.com/grafana/tempo/blob/2b4f143081d708c316ac00cc062f0c2b0053f54e/tempodb/blocklist/poller.go#L579). The first call passes a non prefixed path for deletion, while the second one passes the full path including the prefix.  

In the deleteTenant function, the result of readWriter.Find call is used to generate a list of files to be deleted for a tenant. The Find function was populating the prefixed key in backend.FindMatch (full path to the object including the prefix), and the key from this result was being passed to the Delete function which would again prepend the prefix, causing the test to fail.

I have changed the readWriter.Find implementation for s3/gcs/azure backends to populate non prefixed key in the FindMatch resultset. This is in accordance with the List implementation for these backends, where the prefix is trimmed from the result.

tempodb/backend/s3/s3.go
```diff
@@ -519,7 +519,7 @@ func (rw *readerWriter) Find(ctx context.Context, keypath backend.KeyPath, f bac
 			if len(res.Contents) > 0 {
 				for _, c := range res.Contents {
 					opts := backend.FindMatch{
-						Key:      c.Key,
+						Key:      strings.TrimPrefix(c.Key, prefix),
```
joe-elliott: trimming this prefix seems too aggressive. this will remove the tenant as well? i think the Find function should return the full path to the object minus `rw.cfg.Prefix`. this way the storage layer abstracts away the configured prefix only but otherwise gives full information to the caller. i think the way this is written will break this functionality on the local backend

maybe this should be:

```
Key: strings.TrimPrefix(c.Key, rw.cfg.Prefix),
```

and we should remove the change in poller.go? wdyt?

kaustubhkurve: I agree that returning the full path without the bucket prefix would be better. The first iteration to fix the test involved just removing the bucket prefix using `strings.TrimPrefix(c.Key, rw.cfg.Prefix)`. But then I went through the List implementation, and modified Find to be consistent with List.

List and Find looked functionally similar, with the exception that Find takes a FindFunc callback to be executed for a match. If we only remove the bucket prefix in Find, the (List + Delete) and (Find + Delete) flows would differ. This was the primary reason to prefer the same consistent result for both methods (returning relative paths over absolute paths). Another option I had evaluated was to augment the `FindMatch` struct to include both relative and absolute paths, however I was unsure if that would be needed.

If it's okay for Find and List to differ and are supposed to serve different use cases, I do concur that it would be better to return the full path.

Thank you for pointing out the issue with the local backend, apologies, I missed it.

kaustubhkurve: @joe-elliott I have updated the backend Find methods to only trim bucket prefix.

joe-elliott: All fair points. Now I'm wondering why we even have both Find and List? They seem so similar I don't know why we didn't just have one implementation.

Regardless, your fix feels safe and minimal. I do agree there are some discrepancies here, but I like that we have a nice targetted fix for the issue. We can do some cleanup later.

MikaelElkiaer:
Has this not been part of a release yet?

PR #6021, opened 2025-12-04
Single binary 3.0

joe-elliott:
**What this PR does**:

Adds a new target: `all-3.0` that bundles all Tempo 3.0 components in a single binary. Note that this setup requires kafka but only b/c this unblocks me for adding Tempo 3.0 e2e tests. Once we do the 3.0 cleanup then this mode will become the default single binary setup.

**Other changes**

- Added an `IsSingleBinary()` helper func to clean up a lot of hard to read boolean logic checking for two modes
- Added a new docker-compose example for `all-3.0`
- Updated the `TestIngest` e2e test to test the new 3.0 single binary
- Removed BlockBuilder from the current single binary. Likely added to make the above work.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/integration/e2e/ingest/config-kafka.yaml
+++ b/integration/e2e/ingest/config-kafka.yaml
@@ -1,9 +1,13 @@
 target: all-3.0
 stream_over_http_enabled: true
+partition_ring_live_store: true
 
 server:
   http_listen_port: 3200
   log_level: info
+
+querier:
+  query_live_store: true
 
 query_frontend:
   rf1_after: "2025-01-01T00:00:00Z"
@@ -13,6 +17,8 @@ query_frontend:
     query_ingesters_until: 30s
 
 distributor:
+  ingester_write_path_enabled: false
+  kafka_write_path_enabled: true
   receivers:
     jaeger:
       protocols:


example/docker-compose/ingest-storage-single-binary/tempo.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+distributor:
+  ingester_write_path_enabled: false
+  kafka_write_path_enabled: true
+  receivers:
+    otlp:
+      protocols:
+        grpc:
+          endpoint: "tempo:4317"
+#  log_received_spans:
```
javiermolinar, line 16: Do we want to leave this commented?

example/docker-compose/ingest-storage-single-binary/tempo.yaml
```diff
@@ -0,0 +1,74 @@
...
+      processors: ['span-metrics', 'service-graphs']
+      generate_native_histograms: both
+
+ingest:
+  enabled: true
+  kafka:
+    address: kafka:9092
+    topic:   tempo-ingest
+
+block_builder:
```
javiermolinar, line 70: It is needed to define the default partition?

assigned_partitions:
   block-builder-0: [0]

javiermolinar: Ah disregard the comment it is already set in the blockbuilder initialization

example/docker-compose/ingest-storage-single-binary/tempo.yaml
joe-elliott: i could go either way. i found the constant logging of spans to make it difficult to debug other issues. leave it uncommented? just delete it? wdyt?

joe-elliott: going to merge so i can move forward, but let me know what you think. i'd be glad to clean this up next PR

PR #6022, opened 2025-12-04
block-builder: simplify partition assigment

mapno:
**What this PR does**:

Adds a new config parameter `partitions_per_instance` that can be used to simplify partition assignment. When `assigned_partitions` is used, it takes precedence over `partitions_per_instance`. When only `partitions_per_instance` is configured partitions are assigned based on the instance's ring ID and the number of configured partitions per instance.

Example:

```
partitions_per_instance: 3
...
block-builder-0: [0,1,2]
block-builder-1: [3,4,5]
block-builder-2: [6,7,8]
```

Fixes https://github.com/grafana/tempo-squad/issues/984

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/blockbuilder/config.go
+++ b/modules/blockbuilder/config.go
@@ -35,7 +35,7 @@ func (c *BlockConfig) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagS
 type Config struct {
 	InstanceID            string             `yaml:"instance_id" doc:"Instance id."`
 	AssignedPartitionsMap map[string][]int32 `yaml:"assigned_partitions" doc:"List of partitions assigned to this block builder."`
-	PartitionsPerInstance int                `yaml:"partitions_per_instance" doc:"Number of partitions assigned to this block."`
+	PartitionsPerInstance int                `yaml:"partitions_per_instance" doc:"Number of partitions assigned to this block builder."`
 	ConsumeCycleDuration  time.Duration      `yaml:"consume_cycle_duration" doc:"Interval between consumption cycles."`
 	MaxBytesPerCycle      uint64             `yaml:"max_consuming_bytes" doc:"Maximum number of bytes that can be consumed in a single cycle.  0 to disable"`
 


modules/blockbuilder/config.go
```diff
@@ -45,6 +46,23 @@ type Config struct {
...
+
+	id, err := ingest.IngesterPartitionID(c.InstanceID)
+	if err != nil {
+		return c.AssignedPartitionsMap[c.InstanceID]
+	}
+
+	assignedPartitions := make([]int32, 0, c.PartitionsPerInstance)
+	for i := 0; i < c.PartitionsPerInstance; i++ {
+		assignedPartitions = append(assignedPartitions, id*int32(c.PartitionsPerInstance)+int32(i))
+	}
```
oleg-kozlyuk-grafana, line 62: What would happen if we change `PartitionsPerInstance` and execute a rolling restart (or any other operation where a single BB ends up with different config)?

mapno: Great Q. That's an existing problem with `assigned_partitions` as well and it's catastrophic. We run the rollout-operator to force a restart of all pods at the same time—but there is still the possibility of a race condition or misconfiguration.

mattdurham:
If we enable this will it reshuffle partition assignments? Should `PartitionAssigment` be the default if both are set for backwards compat?

cmd/tempo/app/config.go
```diff
@@ -258,6 +258,10 @@ func (c *Config) CheckConfig() []ConfigWarning {
 		warnings = append(warnings, warnBackendSchedulerPruneAgeLessThanBlocklistPoll)
 	}
 
+	if len(c.BlockBuilder.AssignedPartitionsMap) == 0 && len(c.BlockBuilder.AssignedPartitionsMap) > 0 {
```
javiermolinar: This condition seems off. It cannot be at the same time 0 and bigger than 0

modules/blockbuilder/config.go
```diff
@@ -45,6 +46,23 @@ type Config struct {
...
 }
 
+func (c *Config) AssignedPartitions() []int32 {
+	if c.PartitionsPerInstance <= 0 {
+		return c.AssignedPartitionsMap[c.InstanceID]
+	}
+
+	id, err := ingest.IngesterPartitionID(c.InstanceID)
+	if err != nil {
+		return c.AssignedPartitionsMap[c.InstanceID]
```
javiermolinar, line 56: Is this possible? if we cannot extract the id we cannot operate and we should return an error

cmd/tempo/app/config.go
mapno: That was a bug, it's fixed now.

modules/blockbuilder/config.go
mapno: Yes, instances can be named anything actually. For example in monolithic mode.

mapno:
> If we enable this will it reshuffle partition assignments? Should `PartitionAssigment` be the default if both are set for backwards compat?

Yes, but that's fine as long as all block-builders update at the same time. And that's taken care of by the rollout-operator.

javiermolinar (approved):

modules/blockbuilder/config.go
```diff
@@ -33,10 +33,11 @@ func (c *BlockConfig) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagS
 }
 
 type Config struct {
-	InstanceID           string             `yaml:"instance_id" doc:"Instance id."`
-	AssignedPartitions   map[string][]int32 `yaml:"assigned_partitions" doc:"List of partitions assigned to this block builder."`
-	ConsumeCycleDuration time.Duration      `yaml:"consume_cycle_duration" doc:"Interval between consumption cycles."`
-	MaxBytesPerCycle     uint64             `yaml:"max_consuming_bytes" doc:"Maximum number of bytes that can be consumed in a single cycle.  0 to disable"`
+	InstanceID            string             `yaml:"instance_id" doc:"Instance id."`
+	AssignedPartitionsMap map[string][]int32 `yaml:"assigned_partitions" doc:"List of partitions assigned to this block builder."`
+	PartitionsPerInstance int                `yaml:"partitions_per_instance" doc:"Number of partitions assigned to this block."`
```
jordanbertasso: This looks like a typo.

```suggestion
	PartitionsPerInstance int                `yaml:"partitions_per_instance" doc:"Number of partitions assigned to this block builder."`
```

PR #6032, opened 2025-12-05
cleanup: drop dependency on github.com/prometheus/statsd_exporter

electron0zero:
**What this PR does**:

It was probably included by accident, we were only using `level` from github.com/prometheus/statsd_exporter.

we use go-kit across our codebase so makes sense to keep using go-kit and dropping statsd_exporter dependency

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


oleg-kozlyuk-grafana:
LGTM, always great to reduce number of dependencies

PR #6033, opened 2025-12-05
Fix leading zero handling in native histograms implementations

zalegrala:
**What this PR does**:

Here we ensure that the native histogram impelemtation apples the leading zero logic to both the classic (when using "both") and native implementations.  This is to avoid an issue when new series are created with a null value, the transition from null -> x does not account for the first value written.  In environments in which metrics are created and then age out, this can cause instability in the results.

Related #2068
Related #4140

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/generator/registry/native_histogram.go
```diff
@@ -488,10 +507,6 @@ func (h *nativeHistogram) classicHistograms(appender storage.Appender, timeMs in
 	// drop "le" label again
 	s.lb.Del(labels.BucketLabel)
 
-	if s.isNew() {
```
javiermolinar, line 486: Why deleting this? We need it to flag that the init series has been appended

zalegrala: The registration moves to collectMetrics.  Both implementations for classic and native now check at the beginning to see if they need to handle the init leading zero, but we only record the series as seen upon collect.

zalegrala: @javiermolinar Did my comment make sense?  Check the code flow if you can to see that we're covering both implementations. :pray:

javiermolinar: Ah I see now. Looks good to me

PR #6034, opened 2025-12-05
Fix many lint errors

mdisibio:
**What this PR does**:
Currently lint errors only get caught when changing a file.  This goes ahead and fixes a bunch in /pkg.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


pkg/traceql/engine_metrics_test.go
```diff
@@ -2643,7 +2643,7 @@ func BenchmarkSumOverTime(b *testing.B) {
 	for range totalSpans {
 		s := time.Duration(randInt(1, 3)) * time.Second
 		v := randFloat(minimum, maximun)
-		in = append(in2, newMockSpan(nil).WithStartTime(uint64(s)).WithSpanString("foo", "bar").WithSpanFloat("kafka.lag", v).WithDuration(100))
+		in = append(in, newMockSpan(nil).WithStartTime(uint64(s)).WithSpanString("foo", "bar").WithSpanFloat("kafka.lag", v).WithDuration(100))
```
joe-elliott, line 2646: i don't know what in2 was but is this a behavior change?

pkg/tempopb/prealloc.go
```diff
@@ -33,7 +33,7 @@ func (r *PreallocBytes) Unmarshal(dAtA []byte) error {
 // MarshalTo implements proto.Marshaller.
 // returned int is not used
 func (r *PreallocBytes) MarshalTo(dAtA []byte) (int, error) {
-	copy(dAtA[:], r.Slice[:])
+	copy(dAtA, r.Slice)
```
joe-elliott, line 36: it's really strange we were reslicing the entire slice before? i'm wondering if this was a holdover from a previous array -> slice conversion

pkg/traceql/engine_metrics_test.go
mdisibio: This is constructing test data for a benchmark, and it looks like it was doing it incorrectly before.  Oh you mean in regards to linting:  yes.  But it was caught by the linter

pkg/tempopb/prealloc.go
mdisibio: That makes sense, as `[:]` is required to convert an array to a slice.

PR #6037, opened 2025-12-08
Tighten up block builder encoding validation, fix tests to actually run

mdisibio:
**What this PR does**:

(1) The block builder validates the encoding on startup, but it's not restrictive enough.  Encodings like "vParquet5-preview1" wouldn't be caught on startup, but much later only when it attempted to flush a block.   This tightens the check to only allow encodings that can be used to write new blocks, so deprecated and other preview versions will cause a startup error.

(2) Block-builder tests have been accidentally not running because we were missing the call to `m.Run()` in `TestMain()`.  Whoops. This updates it so they run again, but also removes the flood of log statements which made running and understanding the tests quite cumbersome. We can substitute the console logger back in as needed when a regression is introduced.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/blockbuilder/blockbuilder.go
+++ b/modules/blockbuilder/blockbuilder.go
@@ -220,6 +220,9 @@ func (b *BlockBuilder) running(ctx context.Context) error {
 	defer close(b.consumeStopped)
 	for {
 		// Create a detached context for consume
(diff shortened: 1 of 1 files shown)


mdisibio:
BB tests are flakey in CI, so going to look for a few more fixes.

PR #6039, opened 2025-12-09
Fix many lint errors in /tempodb

mdisibio:
**What this PR does**:
Similar to #6034 this is fixing many lint errors in /tempodb/*

Fixes n/a

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/config_test.go
+++ b/tempodb/config_test.go
@@ -243,7 +243,7 @@ func TestValidateCompactorConfig(t *testing.T) {
 		MaxCompactionRange: 0,
 	}
 
-	expected := errors.New("Compaction window can't be 0")
+	expected := errors.New("compaction window can't be 0")
 	actual := compactorConfig.validate()
 
 	require.Equal(t, expected, actual)


tempodb/encoding/vparquet4/block_traceql_test.go
```diff
@@ -2289,8 +2277,10 @@ func TestChildOf(t *testing.T) {
 			t.Run(tc.name+"-disconnected", func(t *testing.T) {
 				s := &span{}
 
-				lhs := append(tc.lhs, allDisconnected...)
-				rhs := append(tc.rhs, allDisconnected...)
+				lhs := tc.lhs
```
joe-elliott, line 2280: i prefer the way it was, but i also enjoy not being yelled at by the linter :)

mdisibio: Agree, another option is `tc.lhs = append(tc.lhs...`, but modifying the test inputs is also not great.

PR #6046, opened 2025-12-09
feat: Added flag to disable KIP-714 Kafka client metrics

oleg-kozlyuk-grafana:
**What this PR does**:

Disabled Kafka API client metrics as currently unused

https://github.com/grafana/tempo-squad/issues/992

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/ingest/writer_client.go
+++ b/pkg/ingest/writer_client.go
@@ -138,8 +138,8 @@ func commonKafkaClientOptions(cfg KafkaConfig, metrics *kprom.Metrics, logger lo
 		}),
 	}
 
-	// Disable client metrics unless explicitly enabled to reduce noise.
-	if !cfg.EnableKafkaTelemetry {
+	// Disable client metrics if explicitly disabled to reduce noise.
+	if cfg.DisableKafkaTelemetry {
 		opts = append(opts, kgo.DisableClientMetrics())
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/ingest/writer_client.go
```diff
@@ -104,6 +104,7 @@ func commonKafkaClientOptions(cfg KafkaConfig, metrics *kprom.Metrics, logger lo
 		kgo.ClientID(cfg.ClientID),
 		kgo.SeedBrokers(cfg.Address),
 		kgo.DialTimeout(cfg.DialTimeout),
+		kgo.DisableClientMetrics(),
```
mattdurham: Are we disabling here because we trust that the reader will set them up?

oleg-kozlyuk-grafana: No, we're disabling [KIP-714](https://cwiki.apache.org/confluence/display/KAFKA/KIP-714%3A+Client+metrics+and+observability) completely as WarpStream does not support it, and there is no current Tempo features that use it

mdisibio: Does it make sense for this to be configurable, or is there a way to determine if the server supports it?  It might be useful for users of different kafka implementations.

oleg-kozlyuk-grafana: Fair point, added configuration flag

oleg-kozlyuk-grafana:
Flipped the meaning of the flag so it's opt-out

javiermolinar (approved):

PR #5688, opened 2025-09-28
Deprecating vParquet2 block format

ie-pham:
**What this PR does**: Deprecating vParquet2 and disabling writing of vParquet2. 

**Which issue(s) this PR fixes**:
Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #6071, opened 2025-12-11
Delete vParquet2 format

mdisibio:
**What this PR does**:
Following #5688 and #6062 , we can now remove vParquet2 for the next release.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/operations/tempo_cli.md
+++ b/docs/sources/tempo/operations/tempo_cli.md
@@ -410,24 +410,24 @@ Arguments:
 ```bash
 tempo-cli parquet convert data.parquet out.parquet
 ```
-## Parquet convert 2 to 3 command
-Converts a vParquet2 file (actual data.parquet) to a vParquet3 block with an optional list of dedicated attribute columns.
-This utility command is useful when testing the impact of different combinations of dedicated columns. Currently,
-all listed columns are assumed to be at the span scope.
+## Parquet convert A to B command
+Converts a vParquet file (actual data.parquet) of format A to a block of newer format B with an optional list of dedicated attribute columns.
+Actual supported versions for A and B vary by Tempo release.  This utility command is useful when testing the impact of different combinations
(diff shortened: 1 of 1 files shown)


tempodb/encoding/common/config.go
```diff
@@ -75,9 +75,9 @@ func ValidateConfig(b *BlockConfig) error {
 	// TODO - Cyclic dependency makes this awkward to improve by using the
 	// deprecation information in the encoding itself, in the versioned logic
 	// in the parent folder. So we are checking raw strings here.
-	if b.Version == "vParquet2" {
+	/*if b.Version == "vParquet2" {
 		return fmt.Errorf(DeprecatedError, "vParquet2", "vParquet3")
-	}
+	}*/
```
mattdurham, line 80: Thoughts on removing vs commenting out?

tempodb/config_test.go
```diff
@@ -142,7 +140,7 @@ func TestDeprecatedVersions(t *testing.T) {
 		{
 			cfg: &Config{
 				WAL: &wal.Config{
-					Version: vparquet2.VersionString,
+					Version: "vParquet5-preview1", // Turned into unsupported
```
mattdurham, line 143: Is this a different change/related?

tempodb/encoding/common/config.go
mdisibio: I left that there because tracking down that logic check next time we deprecate a version.  The issue is that the location of this check can't reference anything in encoding/ or encoding/vParquetX because then it would create a cyclic dependency.  Is there a different way we could comment this to be clearer?

tempodb/config_test.go
mdisibio: We are deleting the vparquet2 module so we cannot use that string from it anymore.  All we need here is a version that is parseable/recognized, but doesn't support writes. Since there isn't another deprecated version, that is the next best thing.

mattdurham (approved):
Lgtm

PR #6088, opened 2025-12-16
Tempo 3.0: Integration Tests Rewrite

joe-elliott:
Complete rewrite of all Tempo integration tests using the new 3.0 architecture. This is an enormous PR and would be quite difficult to review line by line and so let's discuss the most critical elements here. In addition I'll overview the new Tempo integration test harness that simplifies and standardizes the way we run integration tests ... to a point.  Read below for details! I also considered creating a readme.md in the ./integration folder that describes different common testing patterns and how to use test harness to accomplish them, but perhaps this is a case where the code stands for itself? If desired I will put something together.

## Test harness

The [new test harness](https://github.com/grafana/tempo/pull/6088/files#diff-ba114e0cc50926a7feabeccfb680f1d19a45fb0d25c699ffb8335fb57337e746R150) is a simplified and standardized way to run Tempo integration tests. It has three main configuration options that control the way the test is run:

- DeploymentMode: Allows starting in microservices or single binary mode. Default is microservices.
- Backends: Allows configuration the backend to use and will run a test multiple times for each specified backend. Default is local.
- Components: Allows for starting different Tempo components depending on what you want to test. Default is "recent data querying".

The harness also supports a [handful of other options](https://github.com/grafana/tempo/pull/6088/files#diff-ba114e0cc50926a7feabeccfb680f1d19a45fb0d25c699ffb8335fb57337e746R114-R129) that allow for fine tuning the tests and were added to support migration of all existing integration tests. The most important of which is the ConfigOverlay option. This allows the caller to only specify settings that are specific to the integration test they are working on. By default the test is run with a [standardized config](https://github.com/grafana/tempo/pull/6088/files#diff-bc06192ddd0c0454011a16f2c5c02575f72b525f970208e20d40f676b5e134f1) but the caller can overlay [specific settings](https://github.com/grafana/tempo/pull/6088/files#diff-975034aa4a6f4902450e3767bf336b1708b123906ee5e60b18d64d5a2553f0e0) for their test if desired.

All boilerplate of starting up Tempo services has been abstracted away and helper methods have been provided for common patterns. time.Sleep() was avoided at all costs in favor of polling metrics or watching for other state changes. A trace by id lookup test may look something like:

```go
func TestTraceByID(t *testing.T) {
	util.RunIntegrationTests(t, util.TestHarnessConfig{
		Components: util.ComponentsRecentDataQuerying | util.ComponentsBackendQuerying, // start what's necessary for recent data and backend querying
		Backends:   util.BackendObjectStorageAll, // runs basic querying against all 3 object storage backends. no need to replicate for every test.
	}, func(h *util.TempoHarness) {
        // watches a metric to confirm Tempo is ready to accept writes
		h.WaitTracesWritable(t) 

        // write trace
        info := tempoUtil.NewTraceInfo(time.Now(), "")
        require.NoError(t, h.WriteTraceInfo(info, ""))

        // watches a metric to confirm the trace is ready to be queried
		h.WaitTracesQueryable(t, countTraces)

		apiClient := h.APIClientHTTP("")
        util.QueryAndAssertTrace(t, apiClient, info)

        // watches a metric to confirm the trace has been written to object storage
		h.WaitTracesWrittenToBackend(t, countTraces)
        // restarts the query frontend with a special config that forces it to query the backend
		h.ForceBackendQuerying(t)

        // query again! we must get a new API client b/c the query-frontend ports changed
		apiClient = h.APIClientHTTP("")
		util.QueryAndAssertTrace(t, apiClient, i)
	})
}
```

For the sake of sanity all tests have not been completely collapsed into one pattern. For instance there are still [multiple ways](https://github.com/grafana/tempo/pull/6088/files#diff-c2c63c31455b9e6d436377eff744178285d5e82ab308247bea30d71b5ece3b99R83-R101) to write data supported by the test harness. Maybe we can reduce this even further in the future?

## Test structure

Tests were divided into 5 subcategories both for organization purposes and to improve CI run times:

- api               : read, write, mcp, management/status, overrides, etc. all things http and grpc api related
- limits            : tests the application of limits on reading/writing. the smallest sub category and could be expanded or combined into something else
- metrics-generator : what do _you_ think this tests?
- operations        : various alternative operational modes and configurations: https, failure modes, partition scaling, jaeger query, all kvstores, etc.
- storage           : polling and backend scheduler/worker tests

Note that [pathing](https://github.com/grafana/tempo/pull/6088/files#diff-baa837b9ab3c5a10ab140e6e6084d73d6fd2f5893ebaa6c1ddb10e456b2b64ddR25-R28) in the test harness requires all tests to live in folders next to the ./integration/util folder.

## Test migration

Tests were migrated! Some adjusted, deleted and a few created. A lot of tests were checking random metrics, querying /api/echo, or testing seeming unrelated behavior. Most of this has been cleaned up and each test should be focused on testing what it says on the tin. As much of the original test coverage was preserved by creating new tests more focused on specific behavior.

Here's an overview of the changes.

### migrations

- TestSearchTagsV2, TestSearchTagValuesV2
  - went through a rewrite. they were unmarshalling responses into custom structs declared locally instead of tempopb response structs. this has all been cleaned up. in addition all tests now go through both http and streaming endpoints.
- TestSingleTenantSearch, TestWildCardTenantSearch, TestTwoTenantsSearch, TestThreeTenantsSearch
  - simplified. still tests multiple endpoints, but no longer checks a bunch of query metrics. 
- TestOverrides
  - Expanded to run on all 3 object storage backends
- TestQueryRange*
  - all query range tests were expanded to test on both http and streaming 
- TestReceivers
  - expanded to actually test all 4 receivers. was only doing 2
- TestSearchUsingJaegerPlugin
  - combination of TestSearchUsingJaegerPlugin and TestSearchUsingBackendTagsService
- TestMetricsGeneratorTargetInfoEnabled
  - Fixed to actually test target info was previously just a copy of the TestMetricsGeneratorRemoteWrite test
- TestStreamingSearch_badRequest, TestSearchTagValuesV2_badRequest, TestOverridesAPI_GET, TestOverridesAPI_POST, TestOverridesAPI_PATCH, TestOverridesAPI_DELETE, TestInvalidTenants, TestMCP, TestCompression, TestHTTPS, TestOTLPLimits, TestOTLPLimitsVanillaClient, TestQueryLimits, TestLimitsPartialSuccess, TestQueryRateLimits, TestMetricsGeneratorMessagingSystemLatencyHistogramEnabled, TestEncodings, TestPollerOwnership, TestTenantDeletion, TestLiveStoreDownscaleHappyPath, TestLiveStoreLookback, TestLiveStorePartitionDownscale, TestLiveStoreRestart
  - migrated directly

### renames

Essentially migrated directly but renamed.

- TestLimits -> TestIngestionLimits
- TestMetricsGenerator -> TestMetricsGeneratorRemoteWrite
- TestBackendSchedulerConfigurations -> TestBackendScheduler
- TestMicroservicesWithKVStores -> TestKVStores

### removed tests

- TestSearchTags, TestSearchTagValues 
  - dropped b/c 3.0 will not support the v1 endpoints.
- TestCrossClusterReads
  - a deployment mode that doesn't exist in 3.0
- TestAllInOne
  - too generic? a lot of tests run the single binary for different reasons. seemed good enough
- TestShutdownDelay
  - too simple? maybe worth keeping?
- TestScalableSingleBinary
  - not supported in 3.0!
- TestQueryRangeMaxSeriesDisabledQuerier
  - could not figure out what it was meant to be testing.
- TestIngest, TestLiveStore, TestSmokeLiveStoreAPI
  - covered by plenty of other tests
- TestTracePool
  - removed trace pool in favor of a helper func to generate multiple trace infos
- TestPartitionDownscale
  - did some ingester crap. out! also covered by TestLivestorePartitionDownscale

### new tests

- TestTraceByIDandTraceQL
  - consolidation of behavior of a _lot_ of tests that semi-randomly tested TraceQL and trace by id lookup
- TestStatusAndRingAPIs
  - basic smoke test of management apis. could be improved by looking for actual content
- TestWriteMetrics, TestReadMetrics
  - a fair number of tests would write or read data and then check random metrics to make sure they had expected values. these tests are meant to consolidate that and allow all other tests to focus on testing specific APIs or other behavior.

## Additional changes

- Added a [CI step](https://github.com/grafana/tempo/pull/6088/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR104) to detect if new folders were added under ./integration but weren't added to CI to prevent accidentally adding tests that are not run.
- Changed the vulture to use a Nano seed instead of a second seed. This is to prevent multiple traces in a tight loop from generating the same id.
- Added a [few parameters](https://github.com/grafana/tempo/pull/6088/files#diff-933673f9b9ed499c2b5585bff550d0dcab0800e6b8ab63fefe3dbe592c662422R44) to the HTTP client necessary for the tests.

## How to review

Yeah, it's a big one. I'd ask the reviewer to focus on:

- Any changes outside of ./integration
- The test harness itself. Particularly the design choices and ergonomics of creating integration tests.
- Spot check a handful of tests to get a sense of how they were migrated.

EDIT:

Ignore the pending checks in CI! Since the CI checks changed slightly we will need to adjust our branch protection rules after this is approved but before we merge

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,12 +18,12 @@
   have expired and been deleted from the backend, or else it will encounter read errors after upgrading to this release.
 * [CHANGE] Remove busybox from Tempo image to make it more minimal and prevent future vulnerabilities [#5717](https://github.com/grafana/tempo/pull/5717) (@carles-grafana)
 * [CHANGE] Allow RetryInfo to be disabled in per tenant overrides [#5741](https://github.com/grafana/tempo/pull/5741) (@electron0zero)
+* [CHANGE] **BREAKING CHANGE** Added `scope` and `query` params while rewriting integration tests to the Tempo HTTP client `SearchTagsV2WithRange` function.
 * [CHANGE] Added a single binary 3.0 mode `--target=all-3.0` to begin testing single binary 3.0 and updating integration tests. [#6021](https://github.com/grafana/tempo/pull/6021) (@joe-elliott)
   This will be removed in 3.0 and become the standard single binary mode.
 * [CHANGE] Upgrade Tempo to Go 1.25.4 [#5939](https://github.com/grafana/tempo/pull/5939) [#6001](https://github.com/grafana/tempo/pull/6001) (@ruslan-mikhailov)
 * [CHANGE] Upgrade Tempo to Go 1.25.5 [#6096](https://github.com/grafana/tempo/pull/6096) [#6089](https://github.com/grafana/tempo/pull/6089) (@joe-elliott)
 * [CHANGE] Kafka KIP-714 telemetry is enabled by default. Added `disable_kafka_telemetry` config flag to opt-out [#6046](https://github.com/grafana/tempo/pull/6046) (@oleg-kozlyuk-grafana)
-* [CHANGE] **BREAKING CHANGE** Added `scope` and `query` params while rewriting integration tests to the Tempo HTTP client `SearchTagsV2WithRange` function.
 * [FEATURE] Add `tempo_metrics_generator_registry_active_series_demand_estimate` that estimates metrics-generator active series demand even when the active series limit is reached [#5710](https://github.com/grafana/tempo/pull/5710) (@carles-grafana)
 * [FEATURE] Added validation mode and tests for tempo-vulture [#5605](https://github.com/grafana/tempo/pull/5605) (@davidham)
 * [FEATURE] New block encoding vParquet5-preview3 replacing well-known attributes with dedicated column defaults. This format is in development and breaking changes are expected before final release. [#5696](https://github.com/grafana/tempo/pull/5696) (@stoewer)


pkg/httpclient/client.go
```diff
@@ -421,6 +416,29 @@ func (c *Client) MetricsQueryRange(query string, start, end int, step string, ex
 	return m, nil
 }
 
+func (c *Client) MetricsQueryInstant(query string, start, end int64, exemplars int) (*tempopb.QueryInstantResponse, error) {
```
javiermolinar, line 419: Should we refactor this? It just change the endpoint

Makefile
```diff
@@ -144,37 +144,42 @@ test-with-cover-others: tools ## Run other tests with code coverage
...
+test-e2e-operations: tools docker-tempo docker-tempo-query ## Run end to end tests for operations
+	$(GOTEST) -v $(GOTEST_OPT) ./integration/operations
 
 # runs only api e2e tests
 .PHONY: test-e2e-api
 test-e2e-api: tools docker-tempo docker-tempo-query ## Run end to end tests for api
-	$(GOTEST) -v $(GOTEST_OPT) ./integration/e2e/api
+	$(GOTEST) -v $(GOTEST_OPT) ./integration/api
 
 # runs only poller integration tests
```
javiermolinar: Could we change the doc descriptions?. Those are displayed when you type make:

```
❯ make


 Testin' and Lintin'
  test             Run tests
  benchmark        Run benchmarks
  test-with-cover  Run tests with code coverage
  test-with-cover-pkg   Run Tempo packages' tests with code coverage
  test-with-cover-tempodb  Run tempodb tests with code coverage
  test-with-cover-tempodb-wal  Test tempodb/wal with code coverage
  test-with-cover-others  Run other tests with code coverage
  test-integration-poller  Run poller integration tests
  test-integration-backendscheduler  Run backend-scheduler integration tests
  fmt              Check fmt

```

integration/util/config-base.yaml
```diff
@@ -0,0 +1,115 @@
...
+        grpc:
+          endpoint: "0.0.0.0:4317"
+        http:
+          endpoint: "0.0.0.0:4318"
+    zipkin:
+      endpoint: "0.0.0.0:9411"
+    jaeger:
+      protocols:
+        grpc:
+          endpoint: "0.0.0.0:14250"
```
javiermolinar: Is the zipkin receiver used in the tests? 

If I recall jaeger receiver is deprecated and we use now the OTEL exporter internally

integration/util/harness.go
```diff
@@ -0,0 +1,583 @@
...
+
+	// ReadinessProbe is a function to use a custom readiness probe for the Tempo services.
+	// It's used by the https tests to swap the readiness port to 3201.
+	ReadinessProbe e2e.ReadinessProbe
+
+	// PreStartHook is called before starting Tempo services to perform any setup or adjustments.
+	// It's used by a few tests to copy in required files, start prerequisite services, or adjust the template data for overlays.
+	PreStartHook func(*e2e.Scenario, map[string]any) error
+
+	// DeploymentMode specifies whether to run as single binary, microservices, or both
```
javiermolinar: // DeploymentMode specifies whether to run as single binary, microservices, or both

Should be

	// DeploymentMode specifies whether to run as single binary, microservices, or none

?

integration/util/harness.go
```diff
@@ -0,0 +1,583 @@
...
+		if err := s.StartAndWaitReady(liveStoreZoneB); err != nil {
+			return fmt.Errorf("failed to start live store zone b: %w", err)
+		}
+	}
+
+	// Start Distributor - needed by ComponentsRecentDataQuerying, ComponentsBackendQuerying, ComponentsMetricsGeneration
+	if config.Components&(ComponentsRecentDataQuerying|ComponentsBackendQuerying|ComponentsMetricsGeneration) != 0 {
+		h.Services[ServiceDistributor] = NewTempoService("distributor", "distributor",
+			readinessProbe,
+			14250, // jaeger grpc ingest
```
javiermolinar: I think this is not needed

javiermolinar (commented):
Great, nice refactor. I like the new test harness approach. Most of the tests work fine with the default values removing all th boilerplate

integration/util/harness_api_writes.go
```diff
@@ -0,0 +1,251 @@
...
+					VType: thrift.TagType_STRING,
+					VStr:  &resourceValue,
+				},
+			},
+		},
+		Spans: spans,
+	}
+}
+
+func (h *TempoHarness) WriteTraceInfo(traceInfo *util.TraceInfo, tenant string) error {
```
javiermolinar, line 83: Eventually we should switch to tempopb.Trace everywhere

pkg/httpclient/client.go
joe-elliott: do you mean consolidate it with query range? the endpoint changes and then query range also takes a step param. honestly basically every single one of these could be consolidated to something like:

```
func (c *Client) callThing(endpoint string, params map[string]any, out proto.Message) error {}
```

but that seems like a refactor that might be better in its own PR given how big this one has gotten. wdyt? should i attempt this?

integration/util/config-base.yaml
joe-elliott: I don't think we've deprecated the receivers? I've been tempted to, but I don't think we made a decision yet. Currently I'm leaning towards keeping them all.

All 4 ingestions protocols are tested here: https://github.com/grafana/tempo/pull/6088/files#diff-c0059be3dcb97cf5060467548153d1f1a7ecce44e3667776995e7ba2b6e1eb5c

Other than this test everything only uses otel proto. I'm going to remove the ports from config-base.yaml and create an overlay for this test.

integration/util/harness.go
joe-elliott: Good catch! For awhile I was attempting to run all tests in single binary and monolithic mode, but they were taking forever so I just had each test pick one or the other with a preference for microservices.

integration/util/harness.go
joe-elliott: we need it for the receivers test mentioned above. so we either need to just leave it here to always open the port or we need the distributor ports to be a configurable list and use the option in the receivers test. wdyt?

EDIT: on closer look the receivers test uses the single binary. so we could technically remove these ports but leave them in the single binary? i kind of like the symmetry of always opening the same ports tho.

integration/util/harness_api_writes.go
joe-elliott: yes! i was tempted to do this, but it would have required major rewrites of a lot of tests. Currently I think most are using the "trace info" structs, some are writing jaeger batches and very few are using tempopb.Proto but I do think it's the ideal.

I would like to do this refactor but I think we should get this in first so we can see it more clearly.

Makefile
joe-elliott: Nice catch! Forgot about these.

`make help` wasn't working correctly for me. Claude puttered around a bit and made a change to the command so it works for me now? I'm concerned this may work on OSX awk but fail on other distributions. ¯\_(ツ)_/¯

```
@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[[:alnum:]_-]+:.*##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
```

Updated/Fixed all e2e commands and did some mild reorganization.

joe-elliott: The output looks like this to me now:

<details>

<summary>make help output</summary>

```
> make help

Usage:
  make <target>
  help             Display this help

Building
  tempo            Build tempo
  tempo-query      Build tempo-query
  tempo-cli        Build tempo-cli

Unit Tests
  test             Run tests
  benchmark        Run benchmarks
  test-with-cover  Run tests with code coverage
  test-with-cover-pkg  Run Tempo packages' tests with code coverage
  test-with-cover-tempodb  Run tempodb tests with code coverage
  test-with-cover-tempodb-wal  Test tempodb/wal with code coverage
  test-with-cover-others  Run other tests with code coverage

End to End Tests
  test-e2e         Run all e2e tests
  test-e2e-operations  Run operations e2e tests
  test-e2e-api     Run api e2e tests
  test-e2e-limits  Run limits e2e tests
  test-e2e-metrics-generator  Run metrics-generator e2e tests
  test-e2e-storage  Run storage e2e tests
  test-e2e-util    Run unit tests on the e2e test harness
  test-all         Run all tests

Linters/Formatters
  fmt              Format codebase with gofumpt and goimports
  jsonnetfmt       Format jsonnet codebase with jsonnetfmt
  lint             Lint codebase with golangci-lint

Code Coverage
  coverage-clean   Clean coverage files

Docker Images
  docker-tempo     Build tempo docker image
  docker-tempo-multi  Build multiarch image locally, requires containerd image store
  docker-tempo-debug  Build tempo debug docker image
  docker-tempo-cli  Build tempo cli docker image
  docker-tempo-query  Build tempo query docker image
  docker-tempo-vulture  Build tempo vulture docker image

Code Generation
  gen-proto        Generate proto files
  gen-traceql      Generate traceql
  gen-traceql-local  Generate traceq local
  gen-parquet-query  Generate Parquet query

Tempo tools
  vendor-check     Keep up to date vendorized files
  update-mod       Update module
  release          Release
  release-snapshot  Release snapshot

Docs
  docs             Generate docs
  generate-manifest  Generate manifest.md file

jsonnet
  jsonnet          Generate jsonnet
  jsonnet-check    Check jsonnet
  jsonnet-test     Test jsonnet
```

</details>

pkg/util/trace_info_test.go
```diff
@@ -93,7 +93,7 @@ func TestGenerateRandomInt(t *testing.T) {
 }
 
 func TestConstructTraceFromEpoch(t *testing.T) {
-	seed := time.Unix(1632146180, 0)
+	seed := time.Unix(0, 1632146180)
```
oleg-kozlyuk-grafana: Given it's repeated in 3 places and changed anyway - constant?

pkg/util/trace_info_test.go
```diff
@@ -47,7 +47,7 @@ func TestGenerateRandomTags(t *testing.T) {
 }
 
 func TestGenerateRandomString(t *testing.T) {
-	seed := time.Unix(1632146180, 0)
+	seed := time.Unix(0, 1632146180)
```
oleg-kozlyuk-grafana: Nit: Why this change? This was a valid 2020 timestamp before, now it's `Mon Jan 19 1970 21:22:26` 

func Unix(sec [int64](https://pkg.go.dev/builtin#int64), nsec [int64](https://pkg.go.dev/builtin#int64)) [Time](https://pkg.go.dev/time#Time)

pkg/httpclient/client.go
```diff
@@ -42,10 +41,10 @@ type TempoHTTPClient interface {
 	SearchTags() (*tempopb.SearchTagsResponse, error)
 	SearchTagsV2() (*tempopb.SearchTagsV2Response, error)
 	SearchTagsWithRange(start int64, end int64) (*tempopb.SearchTagsResponse, error)
-	SearchTagsV2WithRange(start int64, end int64) (*tempopb.SearchTagsV2Response, error)
+	SearchTagsV2WithRange(scope, query string, start int64, end int64) (*tempopb.SearchTagsV2Response, error)
```
oleg-kozlyuk-grafana, line 44: Is there possibility that someone uses this API? Potential breaking change - at the very least needs CHANGELOG entry

.github/workflows/ci.yml
```diff
@@ -101,20 +101,59 @@ jobs:
...
+            test-e2e-operations,
             test-e2e-api,
-            test-integration-poller,
-            test-integration-backendscheduler,
-            test-e2e-deployments,
-            test-e2e-ingest,
+            test-e2e-limits,
+            test-e2e-metrics-generator,
+            test-e2e-storage,
+            test-e2e-util,
```
oleg-kozlyuk-grafana, line 156: Given you're adding a job dependency anyways, you can use a script to generate the list of folders, pipe it to job output and use it in next job as matrix - rather than validating against hardcoded list. The functionality of ensuring that all folders are covered would be preserved by virtue of make error

integration/util/config-backend-azure.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+overrides:
+  user_configurable_overrides:
+    client:
+      backend: azure
+      azure:
+        container_name: tempo
+        endpoint_suffix: azurite:10000
+        storage_account_name: "devstoreaccount1"
+        storage_account_key: "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=="
```
oleg-kozlyuk-grafana, line 18: Q: what's the idea behind having same settings in config _and_ override?

integration/limits/config-query-rate.yaml
```diff
@@ -0,0 +1,2 @@
+query_frontend:
+  max_outstanding_per_tenant: 0 # forces everything to 429
```
oleg-kozlyuk-grafana, line 2: Nit: does linter not trigger on lack of trailing LF?

integration/util/harness.go
```diff
@@ -0,0 +1,580 @@
...
+
+const (
+	// ComponentsRecentDataQuerying starts the distributor, kafka, query-frontend, querier, and livestores for recent data querying. (default)
+	ComponentsRecentDataQuerying ComponentsMask = 1 << iota
+	// ComponentsBackendQuerying starts the distributor, kafka, query-frontend, querier, and block-builder for backend querying.
+	ComponentsBackendQuerying
+	// ComponentsMetricsGeneration starts the distributor, kafka, metrics generator and prometheus for metrics generation.
+	ComponentsMetricsGeneration
+	// ComponentsBackendWork starts the backend scheduler and worker for testing backend work.
+	ComponentsBackendWork
```
oleg-kozlyuk-grafana, line 81: More of a nitpick, but I would expect a ComponentMask to have one bit per component and then have composite values for use cases, e.g. 
```
ComponentsScheduler = 0x01
ComponentsWorker = 0x02
ComponentsBackendWork = ComponentsScheduler | ComponentsWorker
```
similar to `BackendObjectStorageAll` below

integration/util/harness.go
```diff
@@ -0,0 +1,580 @@
...
+	Backends BackendsMask
+
+	// Components is a bitmask controlling which optional Tempo components are started.
+	// Recent data components are always started. Use this to add optional components.
+	// Defaults to ComponentsRecentDataQuerying
+	Components ComponentsMask
+
+	// DisableParallelism is a flag to disable parallelism for the test.
+	// Defaults to false
+	DisableParallelism bool
```
oleg-kozlyuk-grafana: Nit: not sure of Go paradigm, but it's usually preferred not to have negatives in settings, unless unavoidable

integration/util/harness.go
```diff
@@ -0,0 +1,580 @@
...
+		config.Backends = BackendLocal
+	}
+
+	if config.DeploymentMode == 0 {
+		config.DeploymentMode = DeploymentModeMicroservices
+	}
+
+	if config.Components == 0 {
+		config.Components = ComponentsRecentDataQuerying
+	}
```
oleg-kozlyuk-grafana, line 161: Q: Is there no idiomatic way in Go to have a meaningful default enum value?

integration/util/harness.go
```diff
@@ -0,0 +1,580 @@
...
+
+	backendTCs := backendTestCases(config.Backends)
+
+	// Run tests for each deployment mode and backend combination
+	for _, be := range backendTCs {
+		// t.Run() with t.Parallel() here will cause some tests to run faster, but others like TestKVStores will fail for unknown reasons.
+		runTempoHarness(t, config, be.name, testFunc)
```
oleg-kozlyuk-grafana, line 168: Can parallelism be driven off `DisableParallelism` as well here then?

integration/util/harness.go
```diff
@@ -0,0 +1,580 @@
...
+
+	// Run the test function
+	testFunc(harness)
+}
+
+func requiresKafka(c ComponentsMask) bool {
+	return c&ComponentsRecentDataQuerying != 0 ||
+		c&ComponentsMetricsGeneration != 0 ||
+		c&ComponentsBackendQuerying != 0
+}
```
oleg-kozlyuk-grafana, line 259: ^ exactly why I mentioned bitflags above - `ComponentsKafka` should just be one of bitflag values

integration/util/harness.go
```diff
@@ -0,0 +1,580 @@
...
+	return nil
+}
+
+// normalizeTestName creates a valid Docker service name from a test name
+func normalizeTestName(testName string) string {
+	// max docker name length is 63. otherwise dns fails silently
+	// max test name length is 40 to leave room prefix and suffix. the full container name will be e2e_<test name>_<service name>
+	// this means that if two tests have the same first 40 characters in their names they will conflict!!
+	maxNameLen := 40
+	name := testName[len("Test"):] // strip "Test" prefix
```
oleg-kozlyuk-grafana: `TrimPrefix`?

integration/api/management_test.go
```diff
@@ -0,0 +1,165 @@
...
+			},
+
+			// ring endpoints
+			{
+				name:           "livestore ring",
+				service:        util.ServiceQuerier,
+				endpoint:       "/live-store/ring",
+				expectedStatus: http.StatusOK,
+			},
+			// { 200s but should 404!
```
oleg-kozlyuk-grafana, line 82: Suggestion: explicitly mark as TODO so it can be found later

integration/limits/config-query-rate.yaml
oleg-kozlyuk-grafana, line 2: Q: would it make more sense to set it to 1 and push 2 values? This test case won't catch e.g. if the outstanding counter is broken

pkg/util/trace_info_test.go
joe-elliott: `TraceInfo` is a way to generate consistent traces from a time seed. this is used by the vulture to generate known traces, query tempo and test that they return as expected. the integration tests also use `TraceInfo` as a convenient way to generate traces, push them into Tempo and query them back.

This resulted in some flakiness b/c the tests were using `time.Now()` and if you generated two traces in the same second they were the exact same trace which the tests did not expect. In this PR I [changed the seed](https://github.com/grafana/tempo/pull/6088/files#diff-6491a88bbd3e1cbc71121a030a156d8d82d7275c1a606afcfe6c1c7da7b5ddf0R396) from using seconds to nanos to reduce the occurrence of this flakiness. The tests you see here in trace_info_test.go were using a known seed, generating a trace and testing some properties of the created trace. The simplest fix for the tests was simply to move the timestamp from seconds to nanoseconds to match the change linked above.

pkg/util/trace_info_test.go
joe-elliott: done!

pkg/httpclient/client.go
joe-elliott: fair. i honestly have no idea if this in use outside the tempo codebase but it's certainly possible. adding to the changelog!

integration/util/config-backend-azure.yaml
joe-elliott: in actual production deployments we use different buckets for actual trace data and overrides. the different settings blocks just allow for more flexibility when storing the different data types. agree that it looks strange here to just have everything duplicated.

integration/limits/config-query-rate.yaml
joe-elliott: i suppose not. do you want a newline?

integration/util/harness.go
joe-elliott: i wanted to make the "components" mask more about the logical behaviors that the developer wanted to test than the actual tempo components that were started. offering complete flexibility of starting components would be difficult (but not impossible) b/c we'd need to dynamically choose memberlist seeds and likely other things i'm not considering.

however, by the end i wondered if it was worth it at all. it adds complexity and it may just be simpler to start all tempo components every time. that would have a negative impact on test duration but maybe it may be worth it. i have not researched how much longer tests would take.

integration/util/harness.go
joe-elliott: i prefer the "positive" setting as well, but for simplicity with defaults and wanting parallelism by default i chose this direction. 

i added this setting b/c i thought a test was flaky due to being run in parallel with other tests and i was using this setting to stabilize it. i later found the flake and this setting is no longer in use. i will remove.

integration/util/harness.go
joe-elliott: some people use the "functional options pattern" [like this](https://github.com/parquet-go/parquet-go/blob/main/config.go#L107-L136) but I've always found it to be verbose and reduce discoverability. i prefer the code i wrote b/c it's very straightforward, but ultimately do not care.

integration/util/harness.go
joe-elliott: option removed! adding `t.Parallel()` here would require a level of effort i don't think is worth it at this point.

integration/util/harness.go
joe-elliott: making every component individually startable is more difficult than it seems and i don't think it's worth the effort atm. if you'd like to tackle this after merge, have at it!

integration/util/harness.go
joe-elliott: done

integration/limits/config-query-rate.yaml
joe-elliott: i didn't write this test. only ported it. this limit is on the number of outstanding jobs the query-frontend will keep in its per tenant queue before 429ing. it could be improved, yes, but would require some thought to avoid flakiness.

at this point i'd prefer merging what we have as long as there are not egregious issues and improving iteratively. the PR has already gotten quite large!

integration/api/management_test.go
joe-elliott: done

.github/workflows/ci.yml
joe-elliott: i used an LLM to generate this and have very low investment. not quite sure what advantages your proposal has. if we want more architecture surrounding our integration tests i would prefer saving that for a future PR. i have no vision there :)

integration/util/harness.go
oleg-kozlyuk-grafana: It's not about making components individually startable but rather about code structuring. This piece of code is not on critical path or expected to be changed often, so it's up to you how far you want to go with improving maintainability of this. I'm certainly fine either way.

integration/util/harness.go
oleg-kozlyuk-grafana: I'm fine either way, just interested in patterns here

integration/util/harness.go
oleg-kozlyuk-grafana: Fair enough

integration/util/harness.go
oleg-kozlyuk-grafana: I think there is still value in launching the minimal set of components, to uncover accidental dependencies between them. However, it doesn't necessarily have to happen within the scope of this PR.

integration/limits/config-query-rate.yaml
oleg-kozlyuk-grafana: The usual Unix standard is to have LF at the end of text files, I'm surprised that rather nitpicky CI linter does not flag this. It's a complete nitpick however and not a topic for this PR.

pkg/util/trace_info_test.go
oleg-kozlyuk-grafana: Hm, and adding a nanosecond time delta to a second-based timestamp did not produce diff enough to generate a different seed? That would be an interesting quirk of Go. This is not an issue however, just curious.

integration/util/config-backend-azure.yaml
oleg-kozlyuk-grafana: I am curious about the test config itself - it is indeed not a blocker, but worth a question. Not in any way critical.

javiermolinar (approved):

PR #6098, opened 2025-12-17
[Generator] Add DRAIN span_name sanitizer

Logiraptor:
This PR introduces a DRAIN-based span_name sanitizer. Basically that means we can automatically replace span_name label values like `GET /users/123` with `GET /users/<_>`. The implementation is forked from loki's pattern ingesters, which is originally forked from https://github.com/faceair/drain. This version has been almost fully rewritten at this point to conform to general Go idioms and better match Tempo's code style. The PR is huge, so I'll break down the major points:

1. I've added the `otter` caching library as a dependency. It is an improvement over the `lru` version used in loki for one main reason: it supports both time and size based eviction at once. For more, read the excellent post [The Evolution of Caching Libraries in Go](https://maypok86.github.io/otter/blog/cache-evolution/)
2. DRAIN can be hard to predict. To make it easier to tell what has changed over time, I've created a set of test fixtures which simulate real life span names found internally at Grafana Labs and in a few of our customers. These test fixtures record DRAIN's exact behavior across a diverse input population.
3. In order to avoid leaking sensitive data, the test fixtures are generated by a small tool based on simple patterns. See the doc in pkg/drain/testdata/README.md for more info there.
4. DRAIN requires constant training in order to work, so we can't fit this into the existing Limiter.OnAdd interface. The reason is that metrics only call OnAdd when a series is new, then never again. So this doesn't allow DRAIN to dynamically move series around.

To get an intuition for how drain works, think of it as follows:

For each input string, we first break it into N tokens.  
We then construct a prefix-tree of those tokens.  
Nodes are merged under two conditions and replaced with the `ParamString` (<_>). 
1. If the number of children exceeds `MaxChildren`
2. If the `IsDataHeuristic` returns true for a token.

### DIFF
--- a/docs/sources/tempo/configuration/manifest.md
+++ b/docs/sources/tempo/configuration/manifest.md
@@ -1020,7 +1020,6 @@ overrides:
             native_histogram_bucket_factor: 1.1
             native_histogram_max_bucket_number: 100
             native_histogram_min_reset_duration: 15m0s
-            span_name_sanitization: disabled
         global:
             max_bytes_per_trace: 5000000
     per_tenant_override_config: ""


modules/generator/registry/drain_sanitizer_test.go
```diff
@@ -0,0 +1,152 @@
...
+	// Sanitize multiple times
+	sanitizer.Sanitize(lbls1)
+	sanitizer.Sanitize(lbls2)
+	sanitizer.Sanitize(lbls3)
+
+	// Demand should be tracked (we can't easily verify exact count without exposing internals,
+	// but we can verify the sanitizer doesn't crash and processes all labels)
+	// The demand gauge will be updated periodically via doPeriodicMaintenance
+	demandEstimate := sanitizer.demand.Estimate()
+	assert.GreaterOrEqual(t, demandEstimate, uint64(0))
```
mdisibio: Should this be >= 1 ?

pkg/drain/line_tokenizer.go
```diff
@@ -0,0 +1,182 @@
...
+		currentState := lexAny
+		for currentState != nil {
+			currentState = currentState(l)
+		}
+		if l.token != "" {
+			tokens = append(tokens, l.token)
+		}
+	}
+
+	tokens = append(tokens, "<END>")
```
mdisibio, line 42: Let's add a comment here.  The <END> token is added because the drain algorithm often will wildcard the last token, and this doesn't work well for many inputs like `GET /users/123/data` and `GET /users/123/settings`, we want to preserve the `/data` and `/settings` parts.  Adding the extra token helps reduce this behavior.

pkg/drain/line_tokenizer_test.go
```diff
@@ -0,0 +1,92 @@
+package drain
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestLineTokenizer(t *testing.T) {
```
mdisibio, line 9: Let's add some tests with multi-byte characters and emojis?

modules/generator/registry/drain_sanitizer_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+import (
+	"sync"
+	"testing"
+
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDrainSanitizer_PatternDetection(t *testing.T) {
```
mdisibio, line 12: What do you think about some tests against the full sanitized output?  i.e. that `GET /api/users/123`,etc is sanitized to `GET /api/users/<_>`.   I like the idea of locking in the behavior more than a `.Contains("<_>")` check.

Also, thinking down the road, when we want to try to replicate / debug some behavior, it would be an easy place to start from.

pkg/drain/testdata/dev1.drain
```diff
@@ -0,0 +1,4520 @@
...
+    "PATCH /honeydew-fig-apple-fig-cherry-mango-honeydew-cherry/ECcPhYj2/vbhOA==": "PATCH /honeydew-fig-apple-fig-cherry-mango-honeydew-cherry/ECcPhYj2/vbhOA==",
+    "PATCH /honeydew-grape-honeydew/LEfMM8HX/G4uNA==": "PATCH /honeydew-grape-honeydew/LEfMM8HX/G4uNA==",
+    "PATCH /honeydew-honeydew-nectarine-elderberry-pineapple-apple-fig-cherry/72744445": "PATCH /honeydew-honeydew-nectarine-elderberry-pineapple-apple-fig-cherry/72744445",
+    "PATCH /honeydew-kiwi-fig/nOiqHvsX54uYyA==": "PATCH /honeydew-kiwi-fig/nOiqHvsX54uYyA==",
+    "PATCH /honeydew-kiwi?h=9fd75f188c7039e2fc98d41d8cd98f00b204e9800998ecf8427e": "PATCH /honeydew-kiwi?h=9fd75f188c7039e2fc98d41d8cd98f00b204e9800998ecf8427e",
+    "PATCH /honeydew-lime-banana-mango/18008503": "PATCH /honeydew-lime-banana-mango/18008503",
+    "PATCH /honeydew-mango-kiwi-apple-grape-lemon-pear-pineapple?h=5086bc2e8cb3cefba79dd41d8cd98f00b204e9800998ecf8427e": "PATCH /honeydew-mango-kiwi-apple-grape-lemon-pear-pineapple?h=5086bc2e8cb3cefba79dd41d8cd98f00b204e9800998ecf8427e",
+    "PATCH /honeydew-
```
mdisibio: Is there a reason the placeholder is printed like this?  It looks like it's standard printable ascii `<_>`.

mdisibio, line 4519: My understanding from the readme is this is input from each line in the json, and its matching output value.  Should there be roughly the same number of lines, ~5k?  Unless the input file contains ~500 duplicates?

pkg/drain/config.go
```diff
@@ -0,0 +1,47 @@
...
+		// Similarity of these sequences is 0.75 (the distance)
+		// Both SimTh and LogClusterDepth impact branching factor: the greater
+		// LogClusterDepth and SimTh, the less the chance that there will be
+		// "similar" clusters, but the greater the footprint.
+		SimTh:       0.3,
+		MaxChildren: 1000,
+		ParamString: `<_>`,
+		MinTokens:   3, // const + number suffix + <END>
+		MaxTokens:   80,
+		MaxClusters: 100000,
```
mdisibio, line 45: How's the memory looking with these defaults, is there any concern about them being too large?  I know the variety of span names is pretty high and it takes values like this to get good results, just curious to see what you found.

modules/generator/registry/drain_sanitizer.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+type drainSanitizerMetrics struct {
+	totalSpansCompacted *prometheus.CounterVec
+	demand              *prometheus.GaugeVec
+}
+
+func newMetrics(reg prometheus.Registerer) drainSanitizerMetrics {
+	return drainSanitizerMetrics{
+		totalSpansCompacted: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
```
mdisibio: I'm not seeing other references to this method or the drainSanitizerMetrics struct.  Could we the metrics flat like in other places?

pkg/drain/line_tokenizer.go
```diff
@@ -0,0 +1,182 @@
...
+	case r == '%':
+		return lexURLEncoded
+	case unicode.IsSpace(r):
+		return lexSpace
+	case unicode.Is(unicode.Hex_Digit, r):
+		return maybeHex
+	case unicode.IsLetter(r) || unicode.IsNumber(r):
+		return lexAlphaNumeric
+	default:
+		return l.emit()
```
mdisibio, line 107: I think the emit call here means that every delimiter is emitted separately, i.e. slashes, commas, etc. Not sure how common or worthwhile it is, but maybe a future todo to combine them into one token?

pkg/drain/testdata/dev1.drain
Logiraptor: That's right, there are some duplicates in some of the input files, which leads to a slightly shorter output

pkg/drain/config.go
Logiraptor: I was surprised to find that it didn't add much to memory, but after thinking about it, it made more sense. We already store all these span names multiple times between the spans being processed, the registry itself, and the WAL. Adding another copy which is strictly a subset of all span names doesn't impact memory as much as I originally expected. I don't have profiles anymore since I wrote this a while ago. I can get some concrete profiling data from a dev env if that would help, just let me know!

modules/overrides/config.go
```diff
@@ -158,6 +158,7 @@ type MetricsGeneratorOverrides struct {
 	NativeHistogramBucketFactor     float64       `yaml:"native_histogram_bucket_factor,omitempty" json:"native_histogram_bucket_factor,omitempty"`
 	NativeHistogramMaxBucketNumber  uint32        `yaml:"native_histogram_max_bucket_number,omitempty" json:"native_histogram_max_bucket_number,omitempty"`
 	NativeHistogramMinResetDuration time.Duration `yaml:"native_histogram_min_reset_duration,omitempty" json:"native_histogram_min_reset_duration,omitempty"`
+	DrainMode                       string        `yaml:"drain_mode,omitempty" json:"drain_mode,omitempty"`
```
electron0zero: can this be bit more descriptive so users don't have to know the internal details. we named it `drain_mode` because we are using DRAIN but it doesn't tell anything to end users on what this config does.

modules/overrides/runtime_config_overrides.go
```diff
@@ -463,6 +463,13 @@ func (o *runtimeConfigOverridesManager) MetricsGeneratorNativeHistogramMinResetD
 	return o.defaultLimits.MetricsGenerator.NativeHistogramMinResetDuration
 }
 
+func (o *runtimeConfigOverridesManager) MetricsGeneratorDrainMode(userID string) string {
+	if mode := o.getOverridesForUser(userID).MetricsGenerator.DrainMode; mode != "" {
+		return mode
+	}
+	return o.defaultLimits.MetricsGenerator.DrainMode
```
electron0zero: can we also add it into user configurable overrides as well.

modules/generator/registry/overrides.go
```diff
@@ -17,6 +17,7 @@ type Overrides interface {
 	MetricsGeneratorNativeHistogramBucketFactor(userID string) float64
 	MetricsGeneratorNativeHistogramMaxBucketNumber(userID string) uint32
 	MetricsGeneratorNativeHistogramMinResetDuration(userID string) time.Duration
+	MetricsGeneratorDrainMode(userID string) string
```
electron0zero: can we also add validation for this so only allowed values can be configured here.

checkout `modules/generator/validation/fields.go` for existing validations.

modules/generator/registry/drain_sanitizer.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/grafana/tempo/pkg/drain"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/prometheus/prometheus/model/labels"
+)
+
+var (
+	metricTotalSpansCompacted = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "tempo",
+		Name:      "metrics_generator_registry_drain_spans_compacted_total",
```
electron0zero: if we update the config name, maybe we should rename the metrics as well to remove reference to drain and use a more generic name.

modules/generator/registry/drain_sanitizer.go
```diff
@@ -0,0 +1,102 @@
...
+	demandUpdateChan <-chan time.Time
+	pruneChan        <-chan time.Time
+}
+
+func NewDrainSanitizer(tenant string, dryRun bool) *DrainSanitizer {
+	return &DrainSanitizer{
+		drain:                     drain.New(tenant, drain.DefaultConfig()),
+		dryRun:                    dryRun,
+		metricTotalSpansCompacted: metricTotalSpansCompacted.WithLabelValues(tenant),
+		demand:                    NewCardinality(15*time.Minute, 5*time.Minute),
```
electron0zero: in other places we use the stale duration that's configured in the registry.stale_duration.

checkout usage of the NewCardinality in `newCounter` method in modules/generator/registry/counter.go

modules/generator/registry/drain_sanitizer.go
```diff
@@ -0,0 +1,102 @@
...
+		demandGauge:               metricDemand.WithLabelValues(tenant),
+		demandUpdateChan:          time.Tick(15 * time.Second),
```
electron0zero, line 51: does it make sense to make these timers configurable?

modules/generator/registry/drain_sanitizer.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+}
+
+func (s *DrainSanitizer) Sanitize(lbls labels.Labels) labels.Labels {
+	s.doPeriodicMaintenance()
+
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	spanName := lbls.Get("span_name")
```
electron0zero: nit: move `span_name` into a const?

modules/generator/registry/registry.go
```diff
@@ -113,6 +127,11 @@ func New(cfg *Config, overrides Overrides, tenant string, appendable storage.App
 		externalLabels[cfg.InjectTenantIDAs] = tenant
 	}
 
+	var sanitizer Sanitizer = &noopSanitizer{}
```
electron0zero, line 130: does it make sense to take  sanitizer as an argument for New? similar to how we are doing it for NewLabelBuilder.

then we can pass either noopSanitizer or NewDrainSanitizer when we are creating a ManagedRegistry?

modules/overrides/config.go
electron0zero, line 270: can we add a default for drain_mode and disable it via by default here?

modules/overrides/config_test.go
```diff
@@ -476,6 +476,7 @@ func generateTestLegacyOverrides() LegacyOverrides {
 		MetricsGeneratorNativeHistogramBucketFactor:                      1.5,
 		MetricsGeneratorNativeHistogramMaxBucketNumber:                   200,
 		MetricsGeneratorNativeHistogramMinResetDuration:                  10 * time.Minute,
+		MetricsGeneratorDrainMode:                                        "disabled",
```
electron0zero: once we default to `disabled` in modules/overrides/config.go, we should be able to remove this line and this test should work fine.

electron0zero: once we default to `disabled` in modules/overrides/config.go, we should be able to remove this line and this test should work fine.

modules/generator/registry/drain_sanitizer.go
```diff
@@ -0,0 +1,102 @@
...
+	"time"
+
+	"github.com/grafana/tempo/pkg/drain"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/prometheus/prometheus/model/labels"
+)
+
+var (
+	metricTotalSpansCompacted = promauto.NewCounterVec(prometheus.CounterOpts{
```
electron0zero: nit: we using the word "sanitised" at most other places but we are referring the same process as `compacted`.  can we use the term sanitised here?

it makes the term `compaction` overloaded because tempo has a compaction component that compacts blocks.

pkg/drain/config.go
mdisibio: No need, sounds like nothing noteworthy which is great.

modules/generator/registry/registry.go
mdisibio: Don't have strong feelings either way.  I might say the pattern of having r.sanitizer nil when not needed, instead a no-op thing, is more common in Tempo.   Making it a required non-nil argument means every caller and test have to be aware of the noop sanitizier to get back to previous behavior.

modules/overrides/config.go
```diff
@@ -269,6 +270,7 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 func (c *Config) RegisterFlagsAndApplyDefaults(f *flag.FlagSet) {
 	// Generator
 	c.Defaults.MetricsGenerator.GenerateNativeHistograms = histograms.HistogramMethodClassic
+	c.Defaults.MetricsGenerator.SpanNameSanitization = "disabled"
```
mdisibio: I kind of want to change it to be empty string == no sanitization, instead of "disabled".  A couple thoughts:  

(1) this matches how tenant overrides will be, their overrides will be parsed as having empty strings.  Then we don't have to rely on the fallback back to default limits having something.  

(2) I think it's currently possible to configure defaults.generator.span_name_sanitization to be empty, which enables it imo unintentionally?  Based on the logic in MangedRegistry New().  

```go
if sanitizationMode := overrides.MetricsGeneratorSpanNameSanitization(tenant); sanitizationMode != SpanNameSanitizationDisabled {
		sanitizer = NewDrainSanitizer(tenant, sanitizationMode == SpanNameSanitizationDryRun, cfg.StaleDuration)
}
```

Logiraptor: Good points!

[d2ec130](https://github.com/grafana/tempo/pull/6098/commits/d2ec1308910a5292d080b1f7235495b4c2b99137)

CHANGELOG.md
```diff
@@ -52,6 +52,7 @@
 * [FEATURE] Add entity-based limiting mode for metrics-generator as an alternative to series-based limiting. [#5788](https://github.com/grafana/tempo/pull/5788) (@Logiraptor)
 * [FEATURE] Add `tempo_metrics_generator_registry_active_series_demand_estimate` that estimates metrics-generator active series demand even when the active series limit is reached [#5710](https://github.com/grafana/tempo/pull/5710) (@carles-grafana)
 * [FEATURE] Metrics generator will now produce overflow series to capture new data once limits are hit. These series have the label `metric_overflow="true"`. [#5954](https://github.com/grafana/tempo/pull/5954) (@Logiraptor)
+* [FEATURE] Add experimental drain limiter to the metrics generator. This can reduce metrics cardinality by clustering similar span names. [](https://github.com/grafana/tempo/pull/...) (@Logiraptor)
```
electron0zero:
```suggestion
* [FEATURE] Add experimental drain limiter to the metrics generator. This can reduce metrics cardinality by clustering similar span names. [#6098](https://github.com/grafana/tempo/pull/6098) (@Logiraptor)
```

mdisibio:
For the docs build error, I think we just need to run `make generate-manifest`

mdisibio (approved):

electron0zero (approved):
lgtm.

PR #6106, opened 2025-12-19
Ensure that number of partitions are correct

mattdurham:
Sometimes we see topics with one partition, which causes issues. This code is very similar to other systems used in Grafana to ensure the number of topics is correct. Whomever is review this let me know and I can send you where I pulled it from.


pkg/ingest/config.go
```diff
@@ -162,34 +163,91 @@ func (cfg *KafkaConfig) GetConsumerGroup(instanceID string, partitionID int32) s
...
 
-	defaultNumberOfPartitions := fmt.Sprintf("%d", cfg.AutoCreateTopicDefaultPartitions)
-	_, err = adm.AlterBrokerConfigsState(context.Background(), []kadm.AlterConfig{
-		{
-			Op:    kadm.SetConfig,
-			Name:  "num.partitions",
-			Value: &defaultNumberOfPartitions,
-		},
-	})
+	ctx := context.Background()
```
mattdurham, line 187: Not entirely in love with using background here but its taken from internal sources so seems legit

pkg/ingest/config.go
```diff
@@ -162,34 +163,91 @@ func (cfg *KafkaConfig) GetConsumerGroup(instanceID string, partitionID int32) s
...
+
+	// Try to create the topic. As of Kafka 2.4, we can pass -1 for replication factor
+	// and the broker will use its default configuration.
+	const defaultReplication = -1
+	resp, err := adm.CreateTopic(ctx, int32(cfg.AutoCreateTopicDefaultPartitions), defaultReplication, nil, cfg.Topic)
+	if err == nil {
+		err = resp.Err
+	}
```
mapno, line 195: Nit: This error handling is a bit strange, the exit for `err !=nil` is 40 lines below. Why not 

```go
if err != nil {
    return fmt.Errorf("failed to create topic %s: %w", cfg.Topic, err)
}
```

and continue with

```go
if resp.Err != nil {}
```

mapno (commented):
LGTM

pkg/ingest/config.go
mattdurham: Its actually a bit stranger than that since the functions called returns the response.Err as err so its mostly duplicating that behavior.

mattdurham: Though some of the other functions behave differently on the client so its hard to always trust.

mattdurham: Changing it made it look slightly more off since the code as is  follows a pattern. Ideally making all the client commands return errors more aligned would allow it to clean up where it all makes sense but that would be outside the scope of this.

mattdurham:
The E2E tests arent happy with this change, digging into the call order for the tests.

integration/util/config-base.yaml
```diff
@@ -73,6 +73,7 @@ ingest:
   kafka:
     address: kafka:9092
     topic: tempo-ingest
+    auto_create_topic_default_partitions: 2
```
mattdurham, line 76: Changed this since 1000 took a while to consolidate in the test harness, with these changes.

PR #6110, opened 2025-12-21
cmd/tempo: fail fast on configuration warnings

Ocheretovich:
This change makes Tempo fail fast when configuration warnings are detected.

Previously, invalid or partially incorrect configurations could start successfully with warnings, potentially leading to undefined behavior at runtime. This ensures misconfigurations are surfaced immediately as errors.


joe-elliott:
Thank you for the effort here, but we'd like Tempo to still start even if there are configuration warnings. Some of the warnings accompany perfectly valid production configurations like `mcp.enabled = true`. Other warnings are for strange configuration combinations that, even though invalid, we use. For instance the integration tests often use settings combinations that trip the warnings.

Ocheretovich:
Closing this in favor of a new PR that aligns with maintainer feedback
and keeps configuration warnings non-fatal

PR #6111, opened 2025-12-21
Log configuration warnings without blocking startup

Ocheretovich:
Configuration validation warnings are now returned and logged separately,
and no longer prevent Tempo from starting.

This preserves existing production configurations and integration tests
that intentionally trigger warnings.

### DIFF
--- a/cmd/tempo/main.go
+++ b/cmd/tempo/main.go
@@ -50,7 +50,8 @@ func main() {
 	mutexProfileFraction := flag.Int("mutex-profile-fraction", 0, "Override default mutex profiling fraction.")
 	blockProfileThreshold := flag.Int("block-profile-threshold", 0, "Override default block profiling threshold.")
 
-	config, _, err := loadConfig()
+	config, configVerify, err := loadConfig()
+	// configVerify is kept for backward compatibility with the CLI-based config verification flow.
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "failed parsing config: %v\n", err)
 		os.Exit(1)


Ocheretovich:
Thanks for the clarification on the previous PR https://github.com/grafana/tempo/pull/6110

Ocheretovich:
So I understand you  I have restored the configuration check flag and limited this configuration check request to internal cleanup only. No change in behavior is expected

cmd/tempo/main.go
```diff
@@ -51,6 +51,7 @@ func main() {
 	blockProfileThreshold := flag.Int("block-profile-threshold", 0, "Override default block profiling threshold.")
 
 	config, configVerify, err := loadConfig()
+	// configVerify is kept for backward compatibility with the CLI-based config verification flow.
```
bboreham, line 54: The variable is "kept", but hard-coded to `false`.
How does that help?

cmd/tempo/main.go
```diff
@@ -68,14 +69,17 @@ func main() {
 	log.InitLogger(&config.Server)
 
 	// Verifying the config's validity and log warnings now that the logger is initialized
-	isValid := configIsValid(config)
+	warnings, err := configIsValid(config)
+	if err != nil {
```
bboreham, line 73: How could this happen?  Looking at the code of `configIsValid`, the second return value is hard-coded to `nil`.

bboreham (commented):
Can you explain your thinking here please?

PR #6125, opened 2026-01-05
[Jsonnet] Add gossip member label to block-builders and live-stores

joe-elliott:
**What this PR does**:

Updated the jsonnet so that block-builders and live-stores will be included in the headless gossip service in k8s. This will increase cell stability if other components are down.

Jsonnet Changes:
- Added gossip label to the block builder pod template and removed it from the service
- Swapped to use the "label mixin" function in the livestores to prevent clobbering existing labels
- Moved the gossip label from the statefulset to the pod template where it belongs. This was a bug
- Added the "app" label to the live store pods for consistency with other deployments and statefulsets. I don't know why we have both "app" and "name" on everything but we do.

Output Changes:

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,7 @@
 * [BUGFIX] Fix GetTrace() in tempo-query. [#5864](https://github.com/grafana/tempo/pull/5864) (@andreasgerstmayr)
 * [BUGFIX] Fix unsupported nonexistence nil operator for tags lookup. Fixed issue where some tag values/names were not returned due to distinctAttrCollector optimization [#5967](https://github.com/grafana/tempo/pull/5967) (@ie-pham) (@joe-elliott)
 * [BUGFIX] Fix delete implementation for s3/gcs/azure backends to account for prefix. [#6011](https://github.com/grafana/tempo/pull/6011) (@kaustubhkurve)
+* [BUGFIX] Jsonnet: Correctly add tempo-gossip-member: true labels to block-builders and live-stores [#6125](https://github.com/grafana/tempo/pull/6125) (@joe-elliott)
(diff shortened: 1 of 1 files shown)


operations/jsonnet-compiled/microservices/gen/Service-live-store-zone-a.yaml
```diff
@@ -12,5 +12,6 @@ spec:
     port: 3200
     targetPort: 3200
   selector:
+    app: live-store-zone-a
```
oleg-kozlyuk-grafana, line 15: Nit: semantically this should be `app: tempo`

operations/jsonnet/microservices/live-store.libsonnet
```diff
@@ -119,7 +119,7 @@
       }
       else {}
     )) +
-    statefulSet.mixin.spec.template.metadata.withLabels({ name: name, 'rollout-group': 'live-store' }) +
+    statefulSet.mixin.spec.template.metadata.withLabelsMixin({ name: name, 'rollout-group': 'live-store' }) +
```
oleg-kozlyuk-grafana, line 122: Discussion: would it make sense to separate zone a/zone b into 2 different rollout groups and redeploy sequentially?

operations/jsonnet-compiled/microservices/gen/Service-live-store-zone-a.yaml
joe-elliott: i agree, but all resources in tempo are using this same pattern.

i think we should change this, but we should do it in a different PR and all at once.

operations/jsonnet/microservices/live-store.libsonnet
joe-elliott: i believe the rollout-group label is used by the [rollout operator](https://github.com/grafana/rollout-operator). my guess is that its important that both zones have the same label value here but @mapno would know for sure.

mapno: Right, the rollout operator has zone awareness and rolls each zone sequentially

PR #6126, opened 2026-01-05
Support span:childCount intrinsic with new column in vp5

mdisibio:
**What this PR does**:
Adds support for the `span:childCount` intrinsic, backed by a new column in vParquet5 (still in development).  This is the number of direct children on a span.  It works for search, metrics, autocomplete, and drilldown/compare.

Removed the unscoped `childCount` variant from the parser. This should be fine as it always returned an `unsupported` error so there is no impact to existing queries.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


oleg-kozlyuk-grafana:
This looks comprehensive and well-rounded, LGTM

PR #6131, opened 2026-01-06
[tempo-cli] add suggest dedicated columns cli command

ie-pham:
**What this PR does**: Add new command to generate dedicated columns with different outputs. Options for outputs are: json, jsonnet, yaml.

json vs jsonnet option:
`json` will use this [DedicatedColumn](https://github.com/grafana/tempo/blob/main/tempodb/backend/block_meta.go#L157) struct. `jsonnet` will print out with keys that can be translated to yaml (scope vs s, type vs t etc). `yaml` is just `yaml`. 

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo-cli/cmd-migrate-overrides-config.go
+++ b/cmd/tempo-cli/cmd-migrate-overrides-config.go
@@ -20,8 +20,8 @@ import (
 type migrateOverridesConfigCmd struct {
 	ConfigFile string `arg:"" help:"Path to tempo config file"`
 
-	OverridesDest string `type:"path" short:"o" help:"Path to tempo overrides file. If not specified, output to stdout"`
 	ConfigDest    string `type:"path" short:"d" help:"Path to tempo config file. If not specified, output to stdout"`
+	OverridesDest string `type:"path" short:"o" help:"Path to tempo overrides file. If not specified, output to stdout"`
 }
 
 func (cmd *migrateOverridesConfigCmd) Run(*globalOptions) error {


cmd/tempo-cli/cmd-suggest-dedicated-columns.go
```diff
@@ -0,0 +1,141 @@
...
+
+	"github.com/grafana/tempo/tempodb/backend"
+)
+
+type suggestDedicatedColumnsCmd struct {
+	backendOptions
+
+	TenantID string `arg:"" help:"tenant-id within the bucket"`
+
+	Format              string  `help:"Output format (json/jsonnet/yaml)" enum:"json,jsonnet,yaml" default:"jsonnet"`
```
mdisibio: Do any of these formats allow for easy copy and paste into the tenant runtime config?  Thinking about how OSS users with helm are expected to configure their installs, i.e. Tempo is running and receiving data -> analyse blocks/suggest columns -> ???.  For vParquet5 there are no default well-known columns, so configuring them is more important, and ideally it should be streamlined and easy.  For us internally we use custom jsonnet, so there will be an extra translation layer, so not really worried about that.

ie-pham: jsonnet can be copy pasted directly into anyone using jsonnet and yaml should be helm chart compatible. are there any other ones you're thinking of that we should add?

cmd/tempo-cli/cmd-analyse-block.go
```diff
@@ -439,6 +442,128 @@ func (s blockSummary) print(settings heuristicSettings, printSettings printSetti
 	return nil
 }
 
+func (s blockSummary) printDedicatedColumns(settings heuristicSettings, format string, outPath string) error {
```
mdisibio: What do you think about splitting this into 2 function layers:  blockSummary.ToDedicatedColumns(settings), and then the existing `printDedicatedColumnSuggestions<Format>` ?

cmd/tempo-cli/cmd-analyse-block.go
```diff
@@ -902,6 +1027,98 @@ func printDedicatedColumnOverridesJsonnet(summary blockSummary, settings heurist
 	fmt.Println("")
 }
 
+type DedicatedColumnYAMLReadableKeys struct {
+	// The Scope of the attribute
+	Scope string `yaml:"scope" json:"scope,omitempty"`
```
mdisibio: This looks mostly the same, except not using enums.  Does marshaling the existing struct o yaml not work?

cmd/tempo-cli/cmd-analyse-block.go
```diff
@@ -902,6 +1027,98 @@ func printDedicatedColumnOverridesJsonnet(summary blockSummary, settings heurist
...
+
+	outBytes, err := json.MarshalIndent(printDedCols, "", "  ")
+	if err != nil {
+		fmt.Println("error marshaling dedicated column suggestions to json:", err)
+		return
+	}
+	outBytes = append(outBytes, byte('\n'))
+
+	if outPath != "" {
+		err = os.WriteFile(outPath, outBytes, 0o644)
```
mdisibio: The common write ops could be extracted out of these layers and put in the main command?

cmd/tempo-cli/cmd-suggest-dedicated-columns.go
mdisibio: In that case does it make sense to remove json support? I guess it could be helpful for other arbitrary automation, but yaml should work just the same.

knylander-grafana:
Do we need to update anything in the tempo-cli docs?

cmd/tempo-cli/cmd-analyse-block.go
ie-pham: oh you're right

cmd/tempo-cli/cmd-suggest-columns.go
```diff
@@ -0,0 +1,228 @@
...
+	printDedCols := backendDedicatedColumnsToYAMLReadableKeys(dedCol)
+
+	outBytes, err := json.MarshalIndent(printDedCols, "", "  ")
+	if err != nil {
+		fmt.Println("error marshaling dedicated column suggestions to json:", err)
+		return
+	}
+	outBytes = append(outBytes, byte('\n'))
+
+	if outPath != "" {
```
mdisibio, line 200: Getting an error when running with default outPath.   Debugging, the default is set to the current working directory instead of empty.

Image

Looking at the docs: https://github.com/alecthomas/kong?tab=readme-ov-file#custom-named-decoders  I think it needs to be defaulted `-` and then check for that.

cmd/tempo-cli/main.go
```diff
@@ -25,7 +25,9 @@ const (
 )
 
 type globalOptions struct {
-	ConfigFile string `type:"path" short:"c" help:"Path to tempo config file"`
+	ConfigFile   string `type:"path" short:"c" help:"Path to tempo config file"`
+	OutputFile   string `type:"path" short:"o" help:"File to write output to, instead of stdout" default:""`
```
mdisibio: A little concerned about them being misleading global params, since most commands won't support them.  I would lean towards putting them back in the 2 commands for now, unless you still prefer them here?

cmd/tempo-cli/cmd-suggest-columns.go
```diff
@@ -0,0 +1,228 @@
...
+
+type suggestColumnsCmd struct {
+	backendOptions
+
+	TenantID string `arg:"" help:"tenant-id within the bucket"`
+
+	BlockID             string  `help:"specific block ID to analyse"`
+	MinCompactionLevel  int     `help:"Min compaction level to analyse" default:"3"`
+	MaxBlocks           int     `help:"Max number of blocks to analyse" default:"10"`
+	NumAttr             int     `help:"Number of attributes to display" default:"15"`
```
mdisibio: Let's change the default to 10 attributes.   Otherwise a default copy and paste will result in an error in Tempo, which doesn't prevent startup and could go unnoticed: `msg="Unable to apply overrides for dedicated attribute columns. Columns invalid." error="invalid dedicated attribute columns: number of columns with type 'string' and scope 'span' must be <= 10 but was 15"`

cmd/tempo-cli/main.go
```diff
@@ -25,7 +25,9 @@ const (
...
+	OutputFormat string `help:"Output format (jsonnet/yaml)" enum:"jsonnet,yaml" default:"jsonnet"`
```
mdisibio: Since most people use the helm chart, I think defaulting to yaml for suggest columns would be best.  I confirmed that it works to copy and paste the output straight into the runtime overrides file.

PR #6133, opened 2026-01-06
Implement GCS delete retry when marking a block compacted

zalegrala:
**What this PR does**:

Retry the delete calls of the compaction and retention operations.  A failure to delete this object results in repeated failures to compact or query the block.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/backend/gcs/gcs_test.go
+++ b/tempodb/backend/gcs/gcs_test.go
@@ -154,14 +154,13 @@ func TestRetry_MarkBlockCompacted(t *testing.T) {
 		case "/b/blerg":
 			_, _ = w.Write([]byte(`{}`))
 		default:
-			// Increment the request count for this path
-			val, _ := reqCounts.LoadOrStore(r.URL.Path, new(int32))
-			countPtr := val.(*int32)
-			count := atomic.AddInt32(countPtr, 1)
+			val, _ := reqCounts.LoadOrStore(r.URL.Path, &atomic.Int32{})
+			if atomicCounter, ok := val.(*atomic.Int32); ok {
+				atomicCounter.Add(1)
+			}
 
 			// First two requests fail, third succeeds for each path.
-			if count <= 2 {
-				atomic.AddInt32(&count, 1)
+			if val.(*atomic.Int32).Load() <= 2 {
 				w.WriteHeader(503)
 				return
 			}
@@ -186,8 +185,7 @@ func TestRetry_MarkBlockCompacted(t *testing.T) {
 
 	reqCounts.Range(func(key, value any) bool {
 		urlPath := key.(string)
-		countPtr := value.(*int32)
-		require.Equal(t, int32(3), atomic.LoadInt32(countPtr), "should attempt 3 times for %s", urlPath)
+		require.Equal(t, int32(3), value.(*atomic.Int32).Load(), "should attempt 3 times for %s", urlPath)
 		return true
 	})
 }


tempodb/backend/gcs/gcs_test.go
```diff
@@ -146,14 +147,20 @@ func TestObjectConfigAttributes(t *testing.T) {
...
+
 	server := httptest.NewUnstartedServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Path {
 		case "/b/blerg":
 			_, _ = w.Write([]byte(`{}`))
 		default:
-			// First two requests fail, third succeeds.
-			if atomic.LoadInt32(&count) < 2 {
+			// Increment the request count for this path
+			val, _ := reqCounts.LoadOrStore(r.URL.Path, new(int32))
```
mattdurham: nit: Would an atomic counter look cleaner?

tempodb/backend/gcs/gcs_test.go
```diff
@@ -175,7 +182,13 @@ func TestRetry_MarkBlockCompacted(t *testing.T) {
 	require.NoError(t, err)
 
 	require.NoError(t, c.MarkBlockCompacted(id, "tenant"))
-	require.Equal(t, int32(2), atomic.LoadInt32(&count))
+
+	reqCounts.Range(func(key, value any) bool {
+		urlPath := key.(string)
+		countPtr := value.(*int32)
+		require.Equal(t, int32(3), atomic.LoadInt32(countPtr), "should attempt 3 times for %s", urlPath)
```
mattdurham: Ignorance here: Is 3 times defined somewhere? Ala with ```	storage.WithBackoff(gax.Backoff{}),
		storage.WithPolicy(storage.RetryAlways),```

zalegrala: Not that I see.  One failure followed by one success for each path would be enough to cover the code.

tempodb/backend/gcs/compactor.go
```diff
@@ -19,7 +19,10 @@ func (rw *readerWriter) MarkBlockCompacted(blockID uuid.UUID, tenantID string) e
 	metaFilename := backend.MetaFileName(blockID, tenantID, rw.cfg.Prefix)
 	compactedMetaFilename := backend.CompactedMetaFileName(blockID, tenantID, rw.cfg.Prefix)
 
-	src := rw.bucket.Object(metaFilename)
+	src := rw.bucket.Object(metaFilename).Retryer(
+		storage.WithBackoff(gax.Backoff{}),
+		storage.WithPolicy(storage.RetryAlways),
+	)
```
oleg-kozlyuk-grafana, line 26: How many times will this code retry until giving up?

zalegrala: It looks like it will only give up when the context is canceled, but seeing that we use context.TODO, I think it will retry forever.

tempodb/backend/gcs/gcs_test.go
zalegrala: I'm not sure, since we're storing this as a map we'd need to protect access to the map, which might get worse.  What do you think?

tempodb/backend/gcs/compactor.go
zalegrala: Here is the logic for the retry.  Notice that the 404 we chatted abut in the call is not handled, so fall through to the default false of the return.
https://github.com/zalegrala/tempo/blob/dc6ba5ddba3c18664c49fb9a22292228f3066324/vendor/cloud.google.com/go/storage/invoke.go?plain=1#L147

zalegrala:
I've pushed a commit to follow the same pattern on the Delete call within `ClearBlock`, since it could be another source of left over objects.  If we think we need a maxium retry config, I'm happy to plumb that in, but I'm struggling with thinking of a good default.  If we give up a delete call in either method, we leave data behind that causes undesirable behavior.  Perhaps making it configurable can be a follow up PR.

I do think as a follow up, I think it makes sense to plumb in the context in the `tempodb.Compactor` interface.  This will cause some churn, but each of the CSP backends make use of a `context.TODO`, so giving the context we have makes sense to me.

zalegrala:
I pushed a commit to include a configuration which will allow tuning the retries.

tempodb/backend/gcs/gcs_test.go
zalegrala: I've made this more explicit in the code and set a default config in the test.

tempodb/backend/gcs/compactor.go
zalegrala: I've added a configuration which allows tuning the number of retries.

tempodb/backend/gcs/gcs_test.go
zalegrala: I took another look and pushed a commit.  Perhaps a little smaller, but about the same.

tempodb/backend/gcs/gcs_test.go
```diff
@@ -168,14 +174,95 @@ func TestRetry_MarkBlockCompacted(t *testing.T) {
...
+		require.Equal(t, int32(3), atomic.LoadInt32(&info.count), "should attempt 3 times for %s", urlPath)
+
+		if strings.HasSuffix(urlPath, "meta.json") {
+			require.Equal(t, http.MethodDelete, info.method, "should be delete method for %s", urlPath)
+		} else {
+			require.Equal(t, http.MethodGet, info.method, "should be delete method for %s", urlPath)
+		}
+
+		return true
+	})
```
oleg-kozlyuk-grafana, line 265: As an optional suggestion, it would be nice to also test "fake failure" scenario where the object is actually deleted but the service returns 500.

zalegrala: I'm going to merge, but will look at this again.

zalegrala:
Thanks for the review.

PR #6142, opened 2026-01-07
generator: back off when instance creation fails to avoid resource exhaustion

carles-grafana:
**What this PR does**:

When a processor validation fails and the instance creation fails, the generator will attempt again to create it indefinitely,
potentially causing OOM errors.

This change caches failed instances and backs off to prevent the issue.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/generator.go
+++ b/modules/generator/generator.go
@@ -45,13 +45,17 @@ const (
 	// contained in the requests. This is intended to be used by clients that send
 	// requests for which span-derived metrics have already been generated elsewhere.
 	NoGenerateMetricsContextKey = "no-generate-metrics"
+
+	// failureBackoff is the duration to wait before retrying failed tenant instance creation.
+	failureBackoff = 1 * time.Minute
 )
 
 var tracer = otel.Tracer("modules/generator")
 
 var (
-	ErrUnconfigured = errors.New("no metrics_generator.storage.path configured, metrics generator will be disabled")
-	ErrReadOnly     = errors.New("metrics-generator is shutting down")
+	ErrUnconfigured            = errors.New("no metrics_generator.storage.path configured, metrics generator will be disabled")
+	ErrReadOnly                = errors.New("metrics-generator is shutting down")
+	errInstanceCreationBackoff = errors.New("instance creation in backoff")
 )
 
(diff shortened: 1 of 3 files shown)


modules/generator/generator.go
```diff
@@ -45,13 +45,17 @@ const (
 	// contained in the requests. This is intended to be used by clients that send
 	// requests for which span-derived metrics have already been generated elsewhere.
 	NoGenerateMetricsContextKey = "no-generate-metrics"
+
+	// failureBackoff is the duration to wait before retrying failed tenant instance creation.
+	failureBackoff = 10 * time.Minute
```
mdisibio: I think this means that after fixing the config issue, the operator must wait up to 10 minutes to know it's working.  That seems like a long time, what do you think about something shorter like 1 or 5 minutes?  When we found this, the failure rate was tens/hundreds of times per second (every message received from the queue), so I think even a 1 minute backoff is a huge improvement and guarantees stability.

Or is there a way to respond when the configuration is reloaded, and we could clear `failedInstances` to make the fix quicker?

modules/generator/generator.go
```diff
@@ -62,8 +66,9 @@ type Generator struct {
 
 	ringLifecycler *ring.BasicLifecycler
 
-	instancesMtx sync.RWMutex
-	instances    map[string]*instance
+	instancesMtx  sync.RWMutex
+	instances     map[string]*instance
+	failedTenants map[string]time.Time // tenant -> when creation last failed
```
mdisibio: Although I like the word `tenant` the best, everything else in this the generator is called `instance`. Rename to `failedInstances`?

modules/generator/generator.go
carles-grafana: when the processor is created successfully, the config is reloaded every 10 seconds: https://github.com/carles-grafana/tempo/blob/fix-generator-oom/modules/generator/instance.go#L144

so 1 minute for failed instances sounds good, changed

modules/generator/generator.go
carles-grafana: agree, changed

PR #6179, opened 2026-01-08
fix special operands search tags handling bug

ie-pham:
**What this PR does**: Fix bug in special operand tags handling.

The last change bypassed this function for well known and dedicated columns 
```
addSelectAs := func(attr traceql.Attribute, columnPath string, selectAs string) {
		if attr == tr.tag {
			columnSelectAs[columnPath] = selectAs
		} else {
			columnSelectAs[columnPath] = "" // Don't select, just filter
		}
	}
```

This causes a panic in this [line](https://github.com/grafana/tempo/tree/main/tempodb/encoding/vparquet4/wal_block.go#L851) because it is expecting an traceql.Attribute which well known and dedicated columns do not have. 

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/encoding/vparquet5/block_autocomplete_test.go
+++ b/tempodb/encoding/vparquet5/block_autocomplete_test.go
@@ -404,8 +404,8 @@ func TestFetchTagNames(t *testing.T) {
 			if scope == traceql.AttributeScopeEvent || scope == traceql.AttributeScopeNone {
 				if len(expectedEventValues) > 0 {
 					expectedValues["event"] = append(expectedValues["event"], expectedEventValues...)
-					expectedValues["event"] = append(expectedValues["event"], dedicatedEventValues...)
 				}
+				expectedValues["event"] = append(expectedValues["event"], dedicatedEventValues...)
 			}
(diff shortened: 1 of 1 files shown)


joe-elliott:
can we get a more detailed description of what this fixes?

mdisibio:
The panic bug was introduced in https://github.com/grafana/tempo/pull/6004.  It is triggered by a tag lookup with the parameter `q={ <well-known-column> != nil }`, in one case `span.http.status_code`.

tempodb/encoding/vparquet5/block_autocomplete.go
```diff
@@ -615,15 +616,16 @@ func createDistinctSpanIterator(
 			switch cond.Op {
 			case traceql.OpNone:
 				addPredicate(entry.columnPath, nil) // No filtering
-				columnSelectAs[entry.columnPath] = cond.Attribute.Name
+				addSelectAs(cond.Attribute, entry.columnPath, cond.Attribute.Name)
```
mdisibio: Did we want to go ahead and remove this entire block?  Since there are no more well-known columns for spans in vp5.

mdisibio (approved):

PR #6185, opened 2026-01-09
[querier] Support external mode for traceID lookups

Logiraptor:
**What this PR does**:

This PR adds a new `external` mode to the querier. In this mode, an HTTP request is sent to an external endpoint to fetch trace contents in addition to the normal ingester + block storages.

The external API is expected to support an API matching Tempo's TraceIDV2 API. `/api/v2/traces/{traceID}?start=<start>&end=<end>`. This endpoint should respond with an OTLP proto encoded trace.

This can be used to augment tempo's internal traceID lookup with additional data that may not be ingested to tempo directly.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/configuration/manifest.md
+++ b/docs/sources/tempo/configuration/manifest.md
@@ -341,7 +341,6 @@ querier:
     trace_by_id:
         query_timeout: 10s
         external:
-            enabled: false
             endpoint: ""
             timeout: 10s
     metrics:


modules/querier/querier.go
```diff
@@ -182,6 +185,14 @@ func New(
 		limits: limits,
 	}
 
+	if cfg.TraceByID.External.Enabled {
+		externalClient, err := external.NewClient(cfg.TraceByID.External.Endpoint, cfg.TraceByID.External.Timeout)
```
electron0zero, line 191: can we include some metrics and spans on this so we can see latency, total requests, failed requests, etc to external endpoints.

it would help us pin down the reasons for slow queries when external endpoint is slow.

Logiraptor: Done. Used otelhttp to get standard client spans and metrics, and also added a metric that looks more "tempo native" to track the performance.

docs/sources/tempo/configuration/_index.md
```diff
@@ -900,6 +905,20 @@ querier:
         # Timeout for trace lookup requests
         [query_timeout: <duration> | default = 10s]
 
+        # External trace source configuration. When enabled, trace-by-ID queries
+        # will also fetch trace data from an external HTTP endpoint that implements
+        # Tempo's TraceIDV2 API (/api/v2/traces/{traceID}).
+        external:
+            # Enable querying an external endpoint for trace data.
+            [enabled: <bool> | default = false]
```
electron0zero, line 913: do we need the enabled flag in here as well? I see that we have `external_enabled` config under the frontend section as well?

modules/querier/external/client.go
```diff
@@ -0,0 +1,110 @@
...
+	return &Client{
+		httpClient: &http.Client{
+			Timeout:   timeout,
+			Transport: otelhttp.NewTransport(http.DefaultTransport),
+		},
+		externalURL: externalURL,
+	}, nil
+}
+
+// TraceByID forwards a trace-by-ID request to the external endpoint
```
electron0zero:
```suggestion
// TraceByID forwards a trace-by-ID v2 request to the external endpoint
```

we are doing a v2 request.

modules/querier/external/client.go
```diff
@@ -0,0 +1,110 @@
...
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read external response body: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusNotFound {
+		return nil, fmt.Errorf("external endpoint returned status %d: %s", resp.StatusCode, string(body))
+	}
+
+	var trace tempopb.Trace
```
electron0zero, line 103: we expect and otel trace proto in the response, can we add some comments in this code so that assumption is documented.

docs/sources/tempo/configuration/_index.md
Logiraptor: It's not strictly necessary. Is this what you had in mind? 7d1c11610

electron0zero (approved):

PR #6193, opened 2026-01-11
feat: add support for querying Trace by Id through multiple Tempo instances 

Apoorva64:
**What this PR does**:

This PR introduces federation in the Tempo using Query Frontend for TraceById queries. The implementation works by creating a new Middleware(NewAsyncFederationSharder) based on the sharding middlewares, but instead of sharding the requests in a single tempo instance it "queries" the external tempo instances.

### Example Federation configuration

```
target: federated-query-frontend
stream_over_http_enabled: true

server:
  http_listen_port: 3200

query_frontend:
  max_batch_size: 10


federated_frontend:
  concurrent_requests: 4
  instances:
  - name: tempo1
    endpoint: http://e2e_Federation-tempo1:3200
    timeout: 30s
  - name: tempo2
    endpoint: http://e2e_Federation-tempo2:3200
    timeout: 30s
storage:
  trace:
    backend: local
    local:
      path: /tmp/tempo/traces

 ```
 
 
**Checklist**
- [x] Tests updated
- [x] Documentation added

### CODE REVIEW

**Comment by joe-elliott** [comment]
I don't think we should make the frontend have a federated mode. Instead let's make new module in modules.go, a new target, etc. This independence will prevent federated frontend changes from causing issues with our existing frontend and establish it as an independent target. Since we are currently trying to reuse the frontend pipelines I'm fine with keeping the code in `./modules/frontend/frontend_federated.go` even though targets normally have their own folders under `./modules`.

I do prefer the path we are heading now over https://github.com/grafana/tempo/pull/6127. Thank you for trying something different. 

Two things we will need before merging:
- An integration test in our [operations](https://github.com/grafana/tempo/tree/main/integration/operations) folder.
- Docs that very clearly mark this as experimental, community driven, partially implemented, etc. Perhaps include a table with currently supported features and we can slowly check boxes as we add them. I would probably put a doc [here](https://github.com/grafana/tempo/tree/main/docs/sources/tempo/operations) that describes this new component.

Longer term thoughts:
- The two most difficult to combine responses will be `/api/metrics/query_range` and `/api/metrics/query`. Doing this correctly will require changes to the query-frontend. For instance if you request a `{} | quantile_over_time(span:duration, .99)` the query frontend will need to return histograms so we can correctly calculate quantiles.
- Most endpoints support gRPC streaming modes. This may be more difficult to pull off, but if we reuse the frontend architecture it feels possible. A longer term goal. Supporting HTTP only is a fine start.
- The [combiners](https://github.com/grafana/tempo/tree/main/modules/frontend/combiner) are the main component I'm concerned about attempting to reuse in the federated frontend. The search combiner has special logic to handle "most recent" flag and the metrics combiners have similar logic when streaming results. Look at the relationship between the sharders and combiners to better understand. This complexity is not necessary for just trace by id. For HTTP only it would probably be easier just to have custom, simpler, combiners, but I'm interested in your thoughts there.
- Ignore the "metrics summary" endpoint. It is deprecated and we will remove it in 3.0.

Thankfully the trace by id endpoints dodge most of these concerns. So they are definitely a good place to start.

**Comment by Apoorva64** [comment]
@joe-elliott added docs, tests and moved the federation to a dedicated target 
notes:
- I had to do a small change to the traceById(v1) combiner to also handle json

File: docs/sources/tempo/operations/federation.md [hunk 1]
```diff
@@ -0,0 +1,150 @@
+---
+title: Federated query frontend
+menuTitle: Federation
+description: Learn how to query traces across multiple Tempo instances using the federated query frontend.
+weight: 650
+---
+
+# Federated query frontend
+
+{{< admonition type="warning" >}}
```
knylander-grafana, line 10: Thank you for adding the note for this feature. We have [two shortcodes](https://grafana.com/docs/writers-toolkit/write/shortcodes/#docsexperimental-deployment) that insert admonitions that use standard wording. I think some variation of this might be useful. 

Experimental deployment shortcode has this usage:  `{{< docs/experimental-deployment >}}`

and results in: 
>This deployment is provided only as an example; engineering and on-call support is not available. Documentation is either limited or not provided outside of code comments. No SLA is provided. This example is primarily intended for open source engineers. Because this is provided as an example, it is not meant to be used in production environments, and the risks are unknown/high.

and experimental shortcode:  `{{< docs/experimental product="experimental-feature" featureFlag="its-feature-flag" >}}`

which results in: 
> experimental-feature is an [experimental feature](https://grafana.com/docs/release-life-cycle/). Engineering and on-call support is not available. Documentation is either limited or not provided outside of code comments. No SLA is provided. Enable the its-feature-flag feature toggle in Grafana to use this feature. Do not enable this feature in production environments.

docs/sources/tempo/operations/federation.md
```diff
@@ -0,0 +1,150 @@
...
+The federated query frontend is an **experimental, community-driven feature**.
```
knylander-grafana, line 11: Love community driven features. :)

docs/sources/tempo/operations/federation.md
```diff
@@ -0,0 +1,150 @@
...
+It's partially implemented and under active development.
+The API and configuration may change in future releases.
+Use in production at your own risk.
+{{< /admonition >}}
+
+The federated query frontend is a new Tempo component that enables querying traces across multiple independent Tempo instances from a single endpoint.
```
knylander-grafana, line 17: You may want to explain how this is different from [Cross-tenant query federation](https://grafana.com/docs/tempo/latest/operations/manage-advanced-systems/cross_tenant_query/) and provide a note about when you'd use this versus when you'd use the other.

docs/sources/tempo/operations/federation.md
```diff
@@ -0,0 +1,150 @@
...
+This is useful for multi-region, multi-cluster, or multi-tenant deployments where traces are distributed across separate Tempo backends.
+
+## Feature status
+
+The following table shows the current implementation status of federation features:
+
+| Feature | Status | Description |
```
knylander-grafana, line 24: Consider removing the icons. We don't usually use the X and check mark in the docs.

docs/sources/tempo/operations/federation.md
```diff
@@ -0,0 +1,150 @@
...
+| Weighted routing | ❌ Not implemented | Route requests based on instance capacity |
+| Caching | ❌ Not implemented | Cache responses from federated queries |
+
+## Use cases
+
+Federation addresses common scenarios in large-scale distributed systems:
+
+- **Multi-region deployments**: Services in different regions send traces to local Tempo instances
+- **Multi-cluster setups**: Each Kubernetes cluster has its own Tempo instance
+- **Organizational boundaries**: Different teams operate independent Tempo instances
```
knylander-grafana, line 43: In general, we only bold UI elements or in special cases. Something like doesn't need to be bolded. You may want to add a note to your LLM to not use this type of bolding. Refer to the [Writers' Toolkit Style guide](https://grafana.com/docs/writers-toolkit/write/style-guide/) for additional information.

docs/sources/tempo/operations/federation.md
```diff
@@ -0,0 +1,150 @@
...
+
+Without federation, querying a trace ID requires knowing which instance contains it, or querying each instance manually.
+Federation solves this by providing a unified query layer.
+
+## Architecture
+
+The federated query frontend acts as a proxy that fans out requests to multiple Tempo instances and combines the results.
+
+When you query a trace ID, the federated frontend:
```
knylander-grafana, line 52: Optional: This could be an opportunity for a diagram to illustrate the process.

docs/sources/tempo/operations/federation.md
```diff
@@ -0,0 +1,150 @@
...
+
+1. Sends the request to all configured Tempo instances in parallel
+1. Collects responses from each instance
+1. Combines the trace spans from all instances into a single response
+1. Returns the merged trace to the client
```
knylander-grafana, line 57:
```suggestion
1. Sends the request to all configured Tempo instances in parallel.
1. Collects responses from each instance.
1. Combines the trace spans from all instances into a single response.
1. Returns the merged trace to the client.
```

docs/sources/tempo/operations/federation.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Configuration
```
knylander-grafana, line 59: For headings for procedures or tasks, use an action verb. 
```suggestion
## Configure the federated query frontend
```

docs/sources/tempo/operations/federation.md
```diff
@@ -0,0 +1,150 @@
...
+You must set the `target` to `federated-query-frontend` as a command-line argument when starting Tempo.
+This starts Tempo as a dedicated federated query frontend instead of the standard all-in-one or microservices mode.
+
+```sh
+tempo -target=federated-query-frontend -config.file=/etc/tempo.yaml
+```
+
+### Basic configuration
+
+The following example shows a minimal federation configuration:
```
knylander-grafana, line 72: I don't think you need this heading. You provide the configuration start instructions, the example, and then available options. Also, does someone add this configuration into their standard Tempo configuration file or is this a completely separate configuration file? Would they append this to an existing configuration file? 
```suggestion

The example `tempo.yaml` shows a minimal federation configuration:
```

docs/sources/tempo/operations/federation.md
```diff
@@ -0,0 +1,150 @@
...
+### Configuration options
+
+The following table describes the available configuration options:
+
+| Option | Summary | Required | Type | Default |
+| ------ | ------- | -------- | ---- | ------- |
+| `concurrent_requests` | Maximum number of concurrent requests to send to instances | No | Integer | `4` |
+| `instances` | List of Tempo instances to query | Yes | List | N/A |
+
+### Instance configuration
```
knylander-grafana, line 96: This heading is not needed. The introductory sentence above the table is enough. Someone looking for configuration options will find the information they need by using the other heading. 
```suggestion
```

docs/sources/tempo/operations/federation.md
```diff
@@ -0,0 +1,150 @@
...
+The federated query frontend is a community-driven feature.
+Contributions are welcome to help expand its capabilities.
+
+Refer to the [Tempo GitHub repository](https://github.com/grafana/tempo) to:
+
+- Report issues or request features
+- Submit pull requests to implement missing functionality
+- Discuss design decisions in GitHub issues
+
+## Related resources
```
knylander-grafana, line 146: Should we link to the Cross-tenant query federation? https://grafana.com/docs/tempo/latest/operations/manage-advanced-systems/cross_tenant_query/

knylander-grafana (commented):
Thank you for adding docs for your project! Much appreciated! I can't approve the other parts, since I'm docs-only.  I do have some feedback and suggestions.

chenlujjj:
Is this still in progress?

PR #6197, opened 2026-01-12
Remove workflow that updates make docs procedure

jdbaldry:
This workflow was useful when the procedure changed frequently but has become a burden as it has sprawled across repositories.

It can fail without obvious notfication and requires more permissions than I'd like to perform the task.

It's replaced by a centralized script that updates the procedure in a push fashion on an ad-hoc basis.


knylander-grafana:
@jdbaldry Resolve conflicts?

jdbaldry:
I think it's probably not worth resolving until it's going to be merged because it's a delete only, it trumps any changes in the file. I can resolve if a maintainer with write access merging wants me to but it's probably easier for them to just manually merge the PR picking my side.

oleg-kozlyuk-grafana:
@jdbaldry I'm happy to merge this once you rebase it

PR #6205, opened 2026-01-13
[Bugfix] instant rate

ruslan-mikhailov:
## What this PR does: 

it fixes instant queries for rates. Currently, rate is calculated in level 1 and then assembles into final results from different sources at level 2 and 3. As we change step for instant queries for jobs, it changes denominator.

$$\sum_i interval_i = total\\_duration$$

$$rate = \frac{\sum_{i}sample_i}{\sum_i interval_i} = \sum_{i} (\frac{sample_i}{total\\_duration}) \neq \sum_{i} (\frac{sample_i}{interval_i})$$ 

This explains why rate is accurate for query range and inaccurate for instant queries.

image

To solve this problem, we need to preserve origin step. To do so, we need to pass a param to query range that indicates the request is instant.

As a bonus, it fixes a rare case for small blocks that fully matches step and a.

### Version update
Propagating the new param during version update leaves Tempo in state where some components do send and accept the param, and some not. This means that during update instant queries may return inaccurate results.
To allow graceful update, we can use optional bool that has 3 states: true, false, undefined. If it is undefined, we fallback to previous logic.

However, during rollout when only query-frontend is rolled out, it cannot know rollout state of other components and will not keep origin step after `TrimToBlockOverlap`. To allow a smooth rollout, it is possible to stop using Instant param by setting `send_instant_param` to false (default is true).

In future release we will be able to remove the old logic and assume the param is always sent correctly.

## How this PR has been tested
- e2e and unit tests
- manual check:

1. deploy tempo in distributed mode
2. populate data from vulture for at least 2 hours
3. for different time ranges:
    a. run instant query `{} | count_over_time()`
    b. divide result from `a` to interval range
    c. run instant query `{} | rate()`
Expected result: `b` == `c`

Code for the test (part of it, with test case only; source code is available on demand):
```python
    @pytest.mark.parametrize("_range", [
        pytest.param(LastHour, id="last hour"),
        pytest.param(LastHourRoundedToMinute,  id="last hour rounded to minute"),
        pytest.param(LastHourRounded, id="last hour rounded to hour"),
        pytest.param(LastMinute, id="last minute"),
    ])
    @pytest.mark.parametrize("shift, ", [
        pytest.param(0, id="recent and backend"),
        pytest.param(
            60,
            id="recent and backend, one minute ago",
        ),
        pytest.param(10*60, id="backend only"),
    ])
    def test_rate(self, shift, _range):
        start = _range.start - shift
        end = _range.end - shift
        query = "{} | rate()"
        count_query = "{} | count_over_time()"

        res = api.get_instant_metrics(query, start, end, step='')
        res.raise_for_status()
        res = res.json()
        if not res.get('series'):
            pytest.skip(f"No series found in response: {res}")
        assert res.get('series'), f"No series found in response: {res}"
        actual_value = res['series'][0]['value']
        print(f"instant metrics response: {pretty_json(res)}")

        count_value = api.get_instant_metrics(
            count_query, start, end, step='',
        ).json()['series'][0]['value']
        assert count_value != 0, "No samples found"
        expected_rate = count_value / (_range.end - _range.start)  # rate is count / duration
        print(f"actual value: {actual_value}, expected rate: {expected_rate}")
        assert in_acceptable_range(
            actual_value, expected_rate), f"inaccurancy level: {inaccurancy(actual_value, expected_rate)}\n actual: {actual_value}, expected: {expected_rate}"
```
 
Rollout test:

| test case                    | query frontend                               | querier and live-store | result                               |
| ---------------------------- | -------------------------------------------- | ---------------------- | ------------------------------------ |
| rollout simulation           | bugfix/instant-rate with config flag **off** | main                   | ✅ no regressions                     |
| rollout with default configs | bugfix/instant-rate with config flag **on**  | main                   | ✅ regressions (expected)             |
| rollout simulation           | main                                         | bugfix/instant-rate    | ✅ no regressions                     |
| All rolled out, flag is off  | bugfix/instant-rate with config flag **off** | bugfix/instant-rate    | ✅ no regressions                     |
| All rolled out               | bugfix/instant-rate with config flag **on**  | bugfix/instant-rate    | ✅ no regressions, fixes instant rate |

## Which issue(s) this PR fixes:
Fixes #<issue number>

## Checklist
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/metrics_query_range_handler.go
+++ b/modules/frontend/metrics_query_range_handler.go
@@ -119,6 +119,7 @@ func newMetricsQueryRangeHTTPHandler(cfg Config, next pipeline.AsyncRoundTripper
 		if err := validateQueryRangeReq(cfg, queryRangeReq); err != nil {
 			return httpInvalidRequest(err), nil
 		}
+		req = api.BuildQueryRangeRequest(req, queryRangeReq, "")
 		traceql.AlignRequest(queryRangeReq)
 
 		// build and use roundtripper


pkg/traceql/engine_metrics.go
```diff
@@ -70,21 +73,16 @@ func TrimToBlockOverlap(start, end, step uint64, blockStart, blockEnd time.Time)
...
 	// they will get trimmed back down.
-	start2 = alignStart(start2, end2, step)
-	end2 = alignEnd(start2, end2, step)
-	// if wasn't instant and may become instant,
-	// add one nanosecond. This usually happens to small blocks.
-	if !wasInstant && isInstant(start2, end2, step) {
-		end2++
-	}
+	start2 = alignStart(start2, end2, step, false)
+	end2 = alignEnd(start2, end2, step, false)
```
ruslan-mikhailov: Am I right always putting here "not instant"?

joe-elliott:
Please rebase before merging and make sure the changelog entry is above 2.10-rc.0 🙏

pkg/tempopb/tempo.proto
```diff
@@ -457,6 +457,7 @@ message QueryRangeRequest {
   // Exemplars are optional and can be empty.
   uint32 exemplars = 16;
   uint32 maxSeries = 17; // max response serie before bailing early
+  optional bool instant = 18; // whether this is an instant query
```
oleg-kozlyuk-grafana, line 460: You don't have to explicitly set field as `optional`. Boolean fields default to false, and if a message is sent without this field, it is assumed to have default value: https://protobuf.dev/programming-guides/proto3/#default

Consequently, you can avoid the complexity of having optional field with gogo/protobuf and just add `bool instant = 18`

pkg/tempopb/utils.go
```diff
@@ -28,3 +28,21 @@ func MakeKeyValueDouble(key string, value float64) v1.KeyValue {
 		},
 	}
 }
+
+// gogo/protobuf has poor support for optional field
+// TODO: remove these functions after migration from gogo/protobuf
```
oleg-kozlyuk-grafana, line 33: See comment in `pkg/tempopb/tempo.proto`

pkg/traceql/engine_metrics.go
```diff
@@ -125,11 +126,11 @@ func IsInstant(req *tempopb.QueryRangeRequest) bool {
...
-	return isInstant(req.Start, req.End, req.Step)
-}
-
-func isInstant(start, end, step uint64) bool {
-	return end-start == step
+	if req.HasInstant() {
+		return req.GetInstant()
+	}
+	// false or no field defined, fallback to old logic.
+	return req.End-req.Start == req.Step
```
oleg-kozlyuk-grafana, line 140: Minor: if possible to change to `req.IsInstant()` instead of `IsInstant(req)`, the code would flow a tiny bit better

pkg/traceql/engine_metrics.go
```diff
@@ -493,7 +494,8 @@ type StepAggregator struct {
 var _ RangeAggregator = (*StepAggregator)(nil)
 
 func NewStepAggregator(start, end, step uint64, innerAgg func() VectorAggregator) *StepAggregator {
-	mapper := NewIntervalMapper(start, end, step)
+	const instant = false // never used for instant queries
```
oleg-kozlyuk-grafana, line 504: Why separate constant?

pkg/traceql/engine_metrics.go
```diff
@@ -1547,13 +1549,13 @@ type IntervalMapper interface {
...
-	return NewIntervalMapper(req.Start, req.End, req.Step)
+	return NewIntervalMapper(req.Start, req.End, req.Step, IsInstant(req))
 }
 
-func NewIntervalMapper(start, end, step uint64) IntervalMapper {
+func NewIntervalMapper(start, end, step uint64, instant bool) IntervalMapper {
 	startMs := start - start%uint64(time.Millisecond)
 	endMs := end - end%uint64(time.Millisecond)
-	if isInstant(start, end, step) {
+	if instant || step == 0 {
```
oleg-kozlyuk-grafana: Do we want to fall back to old logic here as well just in case?

pkg/traceql/engine_metrics.go
```diff
@@ -70,11 +72,11 @@ func TrimToBlockOverlap(start, end, step uint64, blockStart, blockEnd time.Time)
...
 	// they will get trimmed back down.
-	start2 = alignStart(start2, end2, step)
-	end2 = alignEnd(start2, end2, step)
-	// if wasn't instant and may become instant,
+	start2 = alignStart(start2, end2, step, false)
+	end2 = alignEnd(start2, end2, step, false)
+	// if had no instant flag set, had no instant step and become instant,
 	// add one nanosecond. This usually happens to small blocks.
-	if !wasInstant && isInstant(start2, end2, step) {
+	if !req.HasInstant() && !IsInstant(req) && (end2-start2 == step) {
```
oleg-kozlyuk-grafana: Does this need to handle 3 states (false, not set, true)? Can we handle just 2 cases (not set = as before, set = new)?

pkg/tempopb/utils.go
oleg-kozlyuk-grafana: Discussed over call

pkg/tempopb/tempo.proto
oleg-kozlyuk-grafana: Discussed over call. Given this is a temporary situation for rollout, need to add an explicit TODO here with a warning that this field will change - to avoid someone using this in protocol.

pkg/traceql/engine_metrics.go
ruslan-mikhailov: Just to state it's instant, for better readability. Const is just to allocate it once.

Discussed: all good here

pkg/traceql/engine_metrics.go
ruslan-mikhailov: If has no instant flag -> we fallback to previous behaviour.
Now we check if req is not instant and may become instant.

In future, this condition will be removed

pkg/traceql/engine_metrics.go
ruslan-mikhailov: I think it may confuse: req has HasInstant() and GetInstant(), and we add IsInstant. I prefer to leave the function, but after we remove `HasInstant` and old logic, we can remove all these functions and just rely on `GetInstant()`

pkg/traceql/engine_metrics.go
ruslan-mikhailov: Maybe I did not get what you mean here. Detection of "instantness" is encapsulated in IsInstant, so here we rely on a `NewIntervalMapper` caller to pass it. All callers use `IsInstant`, which contains the fallback

oleg-kozlyuk-grafana: My point is that if we rely on callers to correctly pass `instant` flag, we likely should not check for `step = 0` as well? Or is it for backward compatibility?

pkg/tempopb/tempo.proto
ruslan-mikhailov: I've added comments.

> to avoid someone using this in protocol.

I'm thinking we can leave it optional. In a future release with cleanup, we will just use default (false) if the param is not provided

mdisibio: It looks like the optional bool compiles to a nilable sub-struct.  I was expecting a `*bool`, but it seems that is not possible?

pkg/traceql/engine_metrics.go
mdisibio: The main reason to trim the job is to slim down the response for better caching hit rate.  I.e. if I query range `Last 24 Hours` repeatedly, and the time range for the block's job is constantly changing 24h, it'll never be hit twice. We trim the range to let that block be reused the whole time.

Thinking about it, that same thought does apply to instant.  However since this is taking a QueryRangeRequest - let's see if it is even called for instant.

pkg/traceql/engine_metrics.go
mdisibio: Let's see if we can improve readability:
wasAssumedRange := !req.HasInstant() && !IsIstant(req)
willBecomeInstant := (end2-start2==step)

pkg/tempopb/tempo.proto
ruslan-mikhailov: I also expected a pointer, but unfortunetly, I could not get better results. I tried
```
  optional bool instant = 18  [(gogoproto.nullable) = true]; // whether this is an instant query
```

but it generates invalid code that does not compile:
```
 invalid operation: m.Instant != nil (mismatched types bool and untyped nil)
```

I added sugar for better readibility and test to be sure `optional` works as expected.

pkg/traceql/engine_metrics.go
mdisibio: +1 I don't think checking `step == 0` hurts anything, but it should be unnecessary.  The new param can/should be relied on as authoritative.  And I think if a step==0 case gets in here, better to fail and detect it then change query modes.

pkg/tempopb/tempo.proto
oleg-kozlyuk-grafana: > I'm thinking we can leave it optional. In a future release with cleanup, we will just use default (false) if the param is not provided

I think as a general practice leaving "migration" plumbing in the final protocol is undesirable, but this should be weighted against other concerns.

pkg/traceql/engine_metrics_test.go
mdisibio, line 865: What do you think about adding a new TestRate() function, that verifies the new behavior across range and instant?

pkg/traceql/engine_metrics.go
ruslan-mikhailov: > However since this is taking a QueryRangeRequest - let's see if it is even called for instant.

Yes, it's also called for instant query as instant forms `QueryRangeRequest` and is being processed as query range with `instant=true`

pkg/traceql/engine_metrics.go
ruslan-mikhailov: Yea, it should never receive step 0

I've removed this check, assuming query engine always gets valid request.

pkg/traceql/engine_metrics.go
ruslan-mikhailov: \+

Added

pkg/traceql/engine_metrics_test.go
ruslan-mikhailov: Added a test case, including wrong rate calculation if no instant param provided

ruslan-mikhailov:
\+ rebase from lastest main

pkg/traceql/engine_metrics_test.go
```diff
@@ -104,26 +117,53 @@ func TestTimestampOf(t *testing.T) {
...
+			end:      15,
+			step:     100,
+			instant:  true,
+			expected: 100,
+		},
 	}
 
 	for _, c := range tc {
-		mapper := NewIntervalMapper(c.start, c.end, c.step)
+		mapper := NewIntervalMapper(c.start, c.end, c.step, false)
```
mdisibio: Should be passing `tc.instant` ?

pkg/traceql/engine_metrics_test.go
```diff
@@ -182,47 +243,53 @@ func TestIntervalOf(t *testing.T) {
...
 			end:      50,
 			step:     10,
+			instant:  false,
 			expected: 3, // 4th interval: (40;50]
 		},
 	}
 
 	for _, c := range tc {
-		mapper := NewIntervalMapper(c.start, c.end, c.step)
+		mapper := NewIntervalMapper(c.start, c.end, c.step, false)
```
mdisibio: Should be passing `tc.instant`, or if this test is always for ranges, then we don't need the field in the test cases?

pkg/traceql/engine_metrics_test.go
ruslan-mikhailov: Oh no 😱! Yes, I missed it here. + added a few more test cases

pkg/traceql/engine_metrics_test.go
ruslan-mikhailov: Fixed too, thank you!

modules/frontend/metrics_query_handler.go
```diff
@@ -50,6 +50,9 @@ func newQueryInstantStreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTri
 			End:   req.End,
 			Step:  req.End - req.Start,
 		}
+		if cfg.Metrics.SendInstantParam {
```
mdisibio: One last thing for me in this PR:  why does this need to be configurable?  I would expect that the new query-frontend always populates this field.  

In all case during rollout it seems ok:

|    | New querier | Old querier |
| ---- | ------------- | ------------- |
| New frontend | true/false, use the field  | ignores the field, existing logic  |
| Old frontend  | nil, fallback logic | existing logic  |

mdisibio: Also, shouldn't we `SetInstant(false)` in the query_range handlers?

ruslan-mikhailov: Yes! Fixed for both. Double checked with debug logs and e2e that the flag is not missing

modules/frontend/metrics_query_range_handler.go
```diff
@@ -120,6 +120,7 @@ func newMetricsQueryRangeHTTPHandler(cfg Config, next pipeline.AsyncRoundTripper
 			return httpInvalidRequest(err), nil
 		}
 		traceql.AlignRequest(queryRangeReq)
+		req = api.BuildQueryRangeRequest(req, queryRangeReq, "")
```
ruslan-mikhailov, line 122: Bug it fixes: we adjust the request but never change it back and pass origin http req to roundtrip

ruslan-mikhailov:
\+ rebase from main to resolve conflicts

PR #6209, opened 2026-01-14
compactor: return error if encoding is unsupported

carles-grafana:
Otherwise it will panic.
This happens when trying to compact preview blocks.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/compactor_test.go
+++ b/tempodb/compactor_test.go
@@ -989,3 +989,62 @@ func benchmarkCompaction(b *testing.B, targetBlockVersion string) {
 	err = rw.compactOneJob(ctx, metas, testTenantID)
 	require.NoError(b, err)
 }
+
+func TestCompactWithConfigUnsupportedVersion(t *testing.T) {
+	tempDir := t.TempDir()
+
+	// Create backend reader/writer directly to write the block meta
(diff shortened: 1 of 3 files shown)


tempodb/compactor_test.go
```diff
@@ -989,3 +989,62 @@ func benchmarkCompaction(b *testing.B, targetBlockVersion string) {
...
+	}, nil, log.NewNopLogger())
+	require.NoError(t, err)
+
+	ctx := context.Background()
+
+	// Create a block meta with an unsupported preview version and write it to storage
+	meta := &backend.BlockMeta{
+		BlockID:  backend.NewUUID(),
+		TenantID: testTenantID,
+		Version:  "vParquet5-preview6",
```
oleg-kozlyuk-grafana, line 1030: Nit: for test might be more illustrative to have `vParquet5-unsupported` if that works

carles-grafana: it's only versions with "preview" that are considered unsupported. others are considered invalid and already returned an error. see https://github.com/carles-grafana/tempo/blob/fix-compactor-panic/tempodb/encoding/versioned.go#L71

PR #6211, opened 2026-01-14
feature: Add live-store.max-partition-lag parameter to prevent stale query res…

oleg-kozlyuk-grafana:
**What this PR does**:

Adds new `max-partition-lag` parameter to live-store that allows to automatically fall back to another zone if one zone lags behind

N/A

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/livestore/live_store.go
+++ b/modules/livestore/live_store.go
@@ -123,13 +123,13 @@ type LiveStore struct {
 	overrides    overrides.Interface
 
 	// Background processing
-	ctx             context.Context // context for the service. all background processes should exit if this is cancelled
-	cancel          func()
-	wg              sync.WaitGroup
-	completeQueues  *flushqueues.ExclusiveQueues[*completeOp]
-	startupComplete chan struct{} // channel to signal that the starting function has finished. allows background processes to block until the service is fully started
-	lagCancel       context.CancelFunc
-	currentLag      atomic.Int64 // stored as nanoseconds, updated by consume()
+	ctx                 context.Context // context for the service. all background processes should exit if this is cancelled
+	cancel              func()
+	wg                  sync.WaitGroup
+	completeQueues      *flushqueues.ExclusiveQueues[*completeOp]
+	startupComplete     chan struct{} // channel to signal that the starting function has finished. allows background processes to block until the service is fully started
+	lagCancel           context.CancelFunc
+	lastRecordTimestamp atomic.Int64 // stored as unix nanoseconds, updated by consume(). zero indicates no data has been received
 }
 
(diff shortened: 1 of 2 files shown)


modules/livestore/config.go
```diff
@@ -45,6 +45,11 @@ type Config struct {
 	MaxBlockDuration      time.Duration `yaml:"max_block_duration"`
 	MaxBlockBytes         uint64        `yaml:"max_block_bytes"`
 
+	// MaxPartitionLag is the maximum acceptable lag for the partition consumer.
+	// If the lag exceeds this value, query endpoints will return Unavailable errors.
+	// A value of 0 disables this check.
+	MaxPartitionLag time.Duration `yaml:"max_partition_lag"`
```
mattdurham, line 51: How often does this occur in practice? Is it better to return partial results over unavailable? How do the upstream queriers handle this?

oleg-kozlyuk-grafana: 1. This is an improvement in handling disaster recovery - when one of the zones significantly lags behind due to rebuild (or network conditions). I.e. very rare
2. The source of the change is that right now, if we were to completely rebuild one zone, the results would be randomly incorrect for the time of rebuild
3. query-frontend is expected to handle this by faling back on same partition in the other zone

There is a subtle tradeoff here that we are exchanging graceful degradation in case of uniform warpstream-caused lag to "all-or-nothing" approach. However, on the other side of the equation is having completely incorrect query result if we ever have to rebuild livestores (e.g. due to low-level changes), or if one side lags due to network issues

modules/livestore/live_store.go
```diff
@@ -126,6 +129,7 @@ type LiveStore struct {
 	completeQueues  *flushqueues.ExclusiveQueues[*completeOp]
 	startupComplete chan struct{} // channel to signal that the starting function has finished. allows background processes to block until the service is fully started
 	lagCancel       context.CancelFunc
+	currentLag      atomic.Int64 // stored as nanoseconds, updated by consume()
```
mapno: Using only the lag in seconds doesn't cover the edge case of the partition not receiving new data because 1) there is none or 2) the partition is in read-only mode during downscale.

mapno (changes requested):
It's a good improvement, but it should cover partitions with no data too.

modules/livestore/live_store.go
oleg-kozlyuk-grafana: That's a fair point. I think switching to absolute timestamps would be a better match here

oleg-kozlyuk-grafana: Changed the code to use last record timestamp as reference

joe-elliott:
Please rebase and make sure the changelog entry is above the 2.10 line 🙏

PR #6212, opened 2026-01-14
Bug: Fix tag values handling with special characters

joe-elliott:
**What this PR does**:
Tempo doesn't handle tag value lookups when special characters are present in the tag value very well. This PR adds support for special characters and adds an integration test to pin this behavior. There are some ticky-tacky changes in this PR to tag path parsing but the integration test should hold everything in place.

Also changed our http client to mimic the way Grafana builds urls when calling Tempo. This will help keep our integration tests "honest" with our most important client. See the integration tests and http client changes for details.

Finally, I was not able to find a solution for attributes containing a double forward slash `//`. The router itself seemed to eat one of the slashes and normalize the path to only have one. This will require more digging.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/api/search_tags_test.go
+++ b/pkg/api/search_tags_test.go
@@ -54,12 +54,12 @@ func TestParseSearchTagValuesRequest(t *testing.T) {
 	}
 
 	for _, tc := range tcs {
-		testUrl := fmt.Sprintf("http://tempo/api/v2/search/tag/%s/values", tc.tagName)
+		testURL := fmt.Sprintf("http://tempo/api/v2/search/tag/%s/values", tc.tagName)
 		if tc.query != "" {
-			testUrl = fmt.Sprintf("%s?q=%s", testUrl, tc.query)
+			testURL = fmt.Sprintf("%s?q=%s", testURL, tc.query)
 		}
 
-		httpReq := httptest.NewRequest("GET", testUrl, nil)
+		httpReq := httptest.NewRequest("GET", testURL, nil)
 		escapedTagName, err := url.PathUnescape(tc.tagName)
 		require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


pkg/httpclient/client.go
```diff
@@ -466,12 +460,28 @@ func (c *Client) buildTagsV2QueryURL(scope, query string, start int64, end int64
...
+	tagsPath := path.Join("/api/search/tag/", key, "/values")
+
+	joinURL, _ := url.Parse(c.BaseURL)
+	joinURL.Path = path.Join(joinURL.Path, tagsPath)
+
+	return joinURL.String()
 }
 
+// the use of path.Join, not escaping the key and the specific way the url is parsed is all very purposeful here
+// to mimic how Grafana builds URLs when calling Tempo so our integration tests mimic our most important client
```
oleg-kozlyuk-grafana, line 478: This is rather unfortunate

modules/frontend/tag_handlers.go
```diff
@@ -29,8 +29,8 @@ import (
 
 // regex patterns for tag values endpoints, precompile for performance
 var (
-	tagNameRegexV1 = regexp.MustCompile(`.*/api/search/tag/([^/]+)/values`)
-	tagNameRegexV2 = regexp.MustCompile(`.*/api/v2/search/tag/([^/]+)/values`)
+	tagNameRegexV1 = regexp.MustCompile(`.*/api/search/tag/(.+)/values`)
+	tagNameRegexV2 = regexp.MustCompile(`.*/api/v2/search/tag/(.+)/values`)
```
oleg-kozlyuk-grafana, line 33: What if I construct an attribute with `key=maliciousAttribute/api/search/tag/`?

PR #6219, opened 2026-01-14
Promote vParquet5 to stable preview

### LINKED ISSUE #4694: Create vParquet5

mdisibio:
**What this PR does**:
In my list we have finished all schema changes, so now we can promote vParquet5 to stable preview.   It can be optionally enabled.  vParquet4 is still the default.  This unblocks the 2.10 release as I think this was the last thing we were waiting for.

There are a few more logic/non-schema nice to haves we want to try, but they won't require breaking column changes.  These include trying out higher numbers of dedicated columns at the span level, query rewrite/optimizations, and blobs in the shared string columns, in case of blob spillover (i.e. you have more than 10 blob attributes and run out of dedicated columns).

Fixes #4694

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


joe-elliott (approved):

oleg-kozlyuk-grafana (approved):

joe-elliott:
oooooooooooooooooooooooooooooooooooo

PR #6223, opened 2026-01-15
[BUGFIX] Fix tag handling when non-attribute runes are present

joe-elliott:
**What this PR does**:
Follow up to https://github.com/grafana/tempo/pull/6212. This PR further improves Tempo's attribute parsing by handling additional cases when "non attribute" runes are present in the attribute name.

In order to fix this I needed to improve the `ParseIdentifier` function. It was not correctly handling quoted attributes like `span."foo bar"`. There were several paths to making this change that all had pros and cons. I chose the one that felt the least terrible, but please let me know if you have a better idea.

This PR correctly mimics the way Grafana requests attribute values from Tempo.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/parse_test.go
+++ b/pkg/traceql/parse_test.go
@@ -1317,6 +1317,7 @@ func TestParseIdentifier(t *testing.T) {
 		".http+":       NewAttribute("http+"),
 		".foo3":        NewAttribute("foo3"),
 		".foo.3":       NewAttribute("foo.3"),
+		".\"0\"":       NewAttribute("0"),
 
 		// Scoped attributes - resource
 		"resource.foo":         NewScopedAttribute(AttributeScopeResource, false, "foo"),


pkg/traceql/parse.go
```diff
@@ -49,26 +49,35 @@ func Parse(s string) (expr *RootExpr, err error) {
 }
 
 func ParseIdentifier(s string) (Attribute, error) {
-	if i := intrinsicFromString(s); i != IntrinsicNone {
-		return NewIntrinsic(i), nil
+	// Wrap the identifier in curly braces to create a valid spanset filter expression
+	attr := "{" + s + "}"
```
mdisibio, line 53: Very clever.  Agree at this point it seems better to rely on the real language parser rather than custom string code.  The tons of added parse tests look good.

Thinking if this changes any edge cases... The only one I could come up with is float-like strings like `.0` `.0abc` which would work under the old code, but I think not under the new one.  `{ .0 }` isn't valid traceql since it doesn't evaluate to a bool, and `{ .0abc }` has a general parse error.  

This edge case might be ok though, it does work if quoted `."0abc".  So in that in that sense maybe this is even more correct.

PR #6236, opened 2026-01-16
[external mode] Use v1 trace query

Logiraptor:
**What this PR does**:
While writing #6217, my intention was to make the external contract match an existing tempo API, that way it would at least be _possible_ to chain multiple tempo clusters together. That's why I chose the path `/api/v2/traces/{traceID}`. I misunderstood that tempo's v2 endpoint returns an OTLP trace directly. It actually wraps the OTLP trace in a small protobuf message. The v1 endpoint actually does return OTLP directly, so I should have used v1 originally.

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/configuration/_index.md
+++ b/docs/sources/tempo/configuration/_index.md
@@ -906,13 +906,13 @@ querier:
         [query_timeout: <duration> | default = 10s]
 
         # External trace source configuration. When enabled, trace-by-ID queries
-        # will also fetch trace data from an external HTTP endpoint that implements
-        # Tempo's TraceIDV1 API (/api/traces/{traceID}).
+        # will also fetch trace data from an external HTTP endpoint that returns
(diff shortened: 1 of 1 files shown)


javiermolinar (approved):
The reason we added the v2 endpoint was to return a proper OTLP response.

V1 returns the OTLP `Trace` directly (with the legacy JSON `batches` rename), while V2 returns the Tempo envelope `TraceByIDResponse` containing the OTLP `Trace`.

The `batches` rename hack is here for backward compatibility:
https://github.com/grafana/tempo/blob/513907521fa36584ef0bab25fedb229760bf1498/pkg/tempopb/trace_utils.go#L14

V1 works for you because it returns the OTLP payload directly at the querier level. 

We want to deprecate the v1 endpoint eventually, but keeping it available at the queriers could still be useful.

electron0zero (approved):

PR #6238, opened 2026-01-16
enhancement: Improve live store readiness state management

oleg-kozlyuk-grafana:
**What this PR does**:

This PR improves live-store behavior on initialization in 2 dimensions:

* Live store will not register in ring until it is ready to accept queries, 
* Live store will now correctly report "ready" only when correctly caught up and initialized, preventing rollout-operator from starting second zone rollout early

**Checklist**
- [x] Tests updated
- [x] Documentation added


modules/livestore/config.go
```diff
@@ -48,6 +48,16 @@ type Config struct {
...
 	BlockConfig common.BlockConfig `yaml:"block_config"`
 
+	// ReadinessTargetLag is the target consumer lag threshold before the live-store
+	// is considered ready to serve queries. The live-store will wait until lag drops
+	// below this value. Set to 0 to disable readiness waiting (default, backward compatible).
+	ReadinessTargetLag time.Duration `yaml:"readiness_target_lag"`
+
+	// ReadinessMaxWait is the maximum time to wait for catching up at startup.
+	// If this timeout is exceeded, the live-store becomes ready anyway.
+	// Only used if ReadinessTargetLag > 0. Default: 30m.
```
javiermolinar, line 58: Can this creates a read outage if both zones have lag?

modules/livestore/config.go
```diff
@@ -83,13 +93,19 @@ func (cfg *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet)
...
+	cfg.ReadinessMaxWait = 30 * time.Minute
+
 	// Initialize block config with defaults
 	cfg.BlockConfig.RegisterFlagsAndApplyDefaults(prefix+".block", f)
 
 	// Register flags for existing fields
 	f.DurationVar(&cfg.CompleteBlockTimeout, prefix+".complete-block-timeout", cfg.CompleteBlockTimeout, "Duration to keep blocks in the live store after they have been flushed.")
 	f.UintVar(&cfg.QueryBlockConcurrency, prefix+".concurrent-blocks", cfg.QueryBlockConcurrency, "Number of concurrent blocks to query for metrics.")
 	f.Float64Var(&cfg.Metrics.TimeOverlapCutoff, prefix+".metrics.time-overlap-cutoff", cfg.Metrics.TimeOverlapCutoff, "Time overlap cutoff ratio for metrics queries (0.0-1.0).")
+	f.DurationVar(&cfg.ReadinessTargetLag, prefix+".readiness-target-lag", cfg.ReadinessTargetLag, "Target consumer lag threshold before live-store is ready. 0 disables waiting (backward compatible).")
```
javiermolinar: should we use Tempo terminology here? "Target livestore lag..."

modules/livestore/live_store.go
```diff
@@ -294,6 +318,11 @@ func (s *LiveStore) starting(ctx context.Context) error {
 		return fmt.Errorf("failed to start partition reader: %w", err)
 	}
 
+	// Wait for catch-up before marking ready (if enabled)
+	if err := s.waitForCatchUp(ctx); err != nil {
```
javiermolinar, line 322: Should this not be in the **starting** method?

modules/livestore/live_store.go
```diff
@@ -316,6 +345,11 @@ func (s *LiveStore) starting(ctx context.Context) error {
 	// allow background processes to start
 	s.startAllBackgroundProcesses()
 
+	// Mark as ready at end of starting()
+	s.readyErr.Store(nil)
+	metricReady.Set(1)
```
javiermolinar, line 350: This should be Set to 0 on stop

modules/livestore/live_store.go
```diff
@@ -366,6 +400,110 @@ func (s *LiveStore) stopping(error) error {
...
+					"max_wait", s.cfg.ReadinessMaxWait,
+				)
+				metricCatchUpDuration.Set(elapsed.Seconds())
+				return nil
+			}
+
+			// Calculate current lag
+			lag := s.calculateTimeLag()
+			if lag == nil {
+				level.Debug(s.logger).Log("msg", "catch-up lag could not be determined, waiting")
```
javiermolinar, line 444: I think this can block forever. If there is no lag (nil) we should set the status to ready to serve

modules/livestore/partition_reader.go
```diff
@@ -101,6 +119,18 @@ func (r *PartitionReader) running(ctx context.Context) error {
 			continue
 		}
 
+		// Calculate and store maximum lag in entries
+		// Technically shaped as a nested loop, but we only have one partition (at the moment)
+		for _, fetch := range fetches {
+			for _, topic := range fetch.Topics {
+				for _, partition := range topic.Partitions {
```
javiermolinar, line 126: Since we only handle one partition is not more precise to do a direct lookup like this?

`lag = fetch[topic][partition]`

modules/livestore/partition_reader.go
```diff
@@ -101,6 +119,18 @@ func (r *PartitionReader) running(ctx context.Context) error {
...
+					// HighWatermark guaranteed to be >= LastStableOffset
+					lag := partition.HighWatermark - partition.LastStableOffset
+					r.lag.Store(max(r.lag.Load(), lag))
```
javiermolinar: Does this not creates a monotonic non-decresing issue?

- Poll 1: lag = 1000 → store 1000
- Poll 2: lag = 50 → max(1000, 50) = 1000 → stored lag stays 1000
- Poll 3: lag = 0 → still 1000

modules/livestore/config.go
oleg-kozlyuk-grafana: It should not. The wait is bounded by `ReadinessMaxWait`, so if warpstream is very slow, live store will simply fall back to old behavior after a while.

EDIT: also, this behavior is only triggered at start - before live-store is marked `ready`, i.e. any behavior after startup sequence is unchanged

modules/livestore/partition_reader.go
oleg-kozlyuk-grafana: Good catch, there was a line above the loop set lag to -1, but I've managed to lose it in iterations. Fixed

modules/livestore/partition_reader.go
oleg-kozlyuk-grafana: The effect would be the same, unless there is a specific functional concern?

modules/livestore/config.go
oleg-kozlyuk-grafana: Removed `consumer` to avoid misunderstandings

modules/livestore/live_store.go
oleg-kozlyuk-grafana: Wait time is bounded by `ReadinessMaxWait`. Nil does not mean "no lag", it means "no data" - if there is actually no lag (empty partition or caught up), the function returns 0

modules/livestore/live_store.go
oleg-kozlyuk-grafana: Good point, added

modules/livestore/live_store.go
oleg-kozlyuk-grafana: `starting` should be finished before `running`, and a lot of live-store init is done in `running`. I would have to move a lot of other logic there as well. I'm not confident yet I can pull this off without breaking some subtle behavior

oleg-kozlyuk-grafana:
I still don't like the integration tests and will give them another pass - but I think this is the right step in the direction I'd like Tempo services to go - to have consistent predictable behavior that is easy to operate both for us and for customers

cmd/tempo/app/app.go
```diff
@@ -376,6 +376,15 @@ func (t *App) readyHandler(sm *services.Manager, shutdownRequested *atomic.Bool)
 			}
 		}
 
+		// LiveStore has a special check that makes sure it has caught up with Kafka
+		// before serving queries.
+		if t.liveStore != nil {
+			if err := t.liveStore.CheckReady(r.Context()); err != nil {
+				http.Error(w, "LiveStore not ready: "+err.Error(), http.StatusServiceUnavailable)
+				return
+			}
```
oleg-kozlyuk-grafana, line 383: Annotations don't catch e2e tests yet

oleg-kozlyuk-grafana:
@codex can you review this for me as well?

modules/livestore/live_store.go
```diff
@@ -366,6 +401,110 @@ func (s *LiveStore) stopping(error) error {
...
+	}
+
+	// Check if we are near end or partition is empty
+	// Arbitrary value picked to shortcut calculations
+	if lag <= 1000 {
+		level.Debug(s.logger).Log(
+			"msg", "At or close to partition end",
+			"lag", lag)
+		return &zero
+	}
```
mapno, line 492: This can be any amount of time, right? Should it be based on the time since the last record's ts?

oleg-kozlyuk-grafana: This is a heuristic to shortcut the wait in cases when lag may be seen as higher than it is for different reasons - e.g. clock skew. The logic here is that live-store should be able to process this amount of traces in roughly a second (I am basing this off of observations of our prod metrics, where lag of ~300k records corresponds to ~200s of timestamp lag). Hence, when we are within ~1000 records, we are already within the target that this change pursues.

mapno: Okay, sounds reasonable.

modules/livestore/partition_reader.go
```diff
@@ -101,6 +119,18 @@ func (r *PartitionReader) running(ctx context.Context) error {
...
 		}
 
+		// Calculate and store maximum lag in entries
+		// Technically shaped as a nested loop, but we only have one partition (at the moment)
+		for _, fetch := range fetches {
+			for _, topic := range fetch.Topics {
+				for _, partition := range topic.Partitions {
+					// HighWatermark guaranteed to be >= LastStableOffset
+					lag := partition.HighWatermark - partition.LastStableOffset
+					r.lag.Store(lag)
```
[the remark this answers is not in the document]
oleg-kozlyuk-grafana: @codex create a PR fixing this to correct calculation

PR #6239, opened 2026-01-16
traceql: canonicalize logical expressions before hashing

### LINKED ISSUE #6214: TraceQL string canonicalization doesn't sort
**Describe the bug**
We format/canonicalize a traceql query [here](https://github.com/grafana/tempo/blob/6bbbb618bba3da4fb7243655679a691b9193a6cc/modules/frontend/search_sharder.go#L361-L367), to get a consistent hash of the query for caching purposes. This is so that small/trivial changes in whitespace and formatting that don't affect the output have better hit rates.

Maybe it should also recursively sort sub-clauses to catch more cases where the query output would be the same, and we could reuse cache. 

Example, these 2 queries are the same, but currently treated as different:

`{ span.foo = "bar" || span.bar = "foo" }`
`{ span.bar = "foo" || span.foo = "bar" }`

Syedowais312:
### What does this PR do?

TraceQL queries that are semantically equivalent but differ in operand order
(e.g. `a || b` vs `b || a`) currently produce different cache keys because the
AST is stringified without canonicalization.

This PR introduces a normalization pass over the TraceQL AST that canonicalizes
commutative logical expressions (`AND` / `OR`) before stringification and hashing.

### Why is this needed?

- Improves cache hit rate for equivalent TraceQL queries
- Prevents redundant backend work
- Preserves existing query semantics and execution behavior

### What changed?

- Added an AST normalization pass for TraceQL
- Canonicalized `AND` / `OR` binary operations recursively
- Invoked normalization before query hashing
- Added unit tests covering:
  - `AND` / `OR` commutativity
  - Nested logical expressions
  - Non-commutative safety

### Testing

- `go test ./pkg/traceql -p 1`
- `go test ./modules/frontend -p 1`

fix: #6214
Full test suite is covered by CI.

### DIFF
--- a/modules/frontend/search_sharder.go
+++ b/modules/frontend/search_sharder.go
@@ -366,6 +366,7 @@ func hashForSearchRequest(searchRequest *tempopb.SearchRequest) uint64 {
 	}
 
 	// forces the query into a canonical form
+	traceql.Normalize(ast)
 	query := ast.String()
 
 	// add the query, limit and spss to the hash
--- a/pkg/traceql/normalize.go
+++ b/pkg/traceql/normalize.go
@@ -30,20 +30,28 @@ func normalizePipelineElement(e PipelineElement) {
 	case *SpansetFilter:
 		normalizeFieldExpr(n.Expression)
 
-	case SpansetOperation:
+	case *SpansetOperation:
 		normalizePipelineElement(n.LHS)
 		normalizePipelineElement(n.RHS)
 
-	case ScalarFilter:
+		if n.Op != OpAnd && n.Op != OpOr {
+			return
+		}
+
+		if n.LHS.String() > n.RHS.String() {
+			n.LHS, n.RHS = n.RHS, n.LHS
+		}
+
+	case *ScalarFilter:
 		normalizeScalarExpr(n.LHS)
 		normalizeScalarExpr(n.RHS)
 
-	case Aggregate:
+	case *Aggregate:
 		if n.e != nil {
 			normalizeFieldExpr(n.e)
 		}
 
-	case GroupOperation:
+	case *GroupOperation:
 		normalizeFieldExpr(n.Expression)
 	}
 }
--- a/pkg/traceql/normalize.go
+++ b/pkg/traceql/normalize.go
@@ -1,0 +1,100 @@
+package traceql
+
+// Normalize rewrites the AST into a canonical form suitable for hashing.
+// It preserves semantics and only reorders commutative logical expressions.
+func Normalize(e Element) {
+	if e == nil {
+		return
+	}
+	normalizeElement(e)
+}
+
+func normalizeElement(e Element) {
+	switch n := e.(type) {
+
+	case *RootExpr:
+		for _, elem := range n.Pipeline.Elements {
+			normalizePipelineElement(elem)
+		}
+
+	case Pipeline:
+		for _, elem := range n.Elements {
+			normalizePipelineElement(elem)
+		}
+	}
+}
+
+func normalizePipelineElement(e PipelineElement) {
+	switch n := e.(type) {
+
+	case *SpansetFilter:
+		normalizeFieldExpr(n.Expression)
+
+	case SpansetOperation:
+		normalizePipelineElement(n.LHS)
+		normalizePipelineElement(n.RHS)
+
+	case ScalarFilter:
+		normalizeScalarExpr(n.LHS)
+		normalizeScalarExpr(n.RHS)
+
+	case Aggregate:
+		if n.e != nil {
+			normalizeFieldExpr(n.e)
+		}
+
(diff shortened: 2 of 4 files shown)


Syedowais312:
Hi @mdisibio,
I’ve opened a draft PR addressing this. Please let me know if you’d like any changes or adjustments
Happy to iterate

modules/frontend/search_sharder.go
```diff
@@ -364,6 +364,7 @@ func hashForSearchRequest(searchRequest *tempopb.SearchRequest) uint64 {
 	}
 
 	// forces the query into a canonical form
+	traceql.Normalize(ast)
```
javiermolinar, line 369: What about adding the normalize step inside the String() method so we can take advantage of it everywhere?

Syedowais312: Good question. I considered that but I was hesitate to put normalization inside Sting() since Normalize() mutates the AST and String is currently Side-effect free and used in a few non hashing contexts (debugging, errors, tests, etc.) 

Keeping normalization expliciy at the  call site makes it clear that we're opting into canonicalization for hashing purpose, without changing the semantics of String() globally or losing the original ordering in logs/debud output. 

That said, if the consensus is that String () should always represent a canonical form, I'm happy to move it there or explore a separate CanonicalString() helper. Let me know what you think

pkg/traceql/normalize.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+func normalizePipelineElement(e PipelineElement) {
+	switch n := e.(type) {
+
+	case *SpansetFilter:
+		normalizeFieldExpr(n.Expression)
+
+	case SpansetOperation:
```
oleg-kozlyuk-grafana: Can we also normalize `SpansetOperation`? `{ a } && { b }`   vs   `{ b } && { a }`

pkg/traceql/normalize.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+	case Pipeline:
+		for _, elem := range n.Elements {
+			normalizePipelineElement(elem)
+		}
+	}
+}
+
+func normalizePipelineElement(e PipelineElement) {
```
oleg-kozlyuk-grafana: nit: Methods receiving structs everywhere which I expect causes a fair bit of copying. Pointer arg or pointer receiver would be better (and save from debugging subtle bugs)

pkg/traceql/normalize.go
```diff
@@ -0,0 +1,100 @@
...
+	if op.RHS != nil {
+		normalizeFieldExpr(op.RHS)
+	}
+
+	// Only reorder commutative logical operators
+	if op.Op != OpAnd && op.Op != OpOr {
+		return
+	}
+
+	leftKey := op.LHS.String()
```
oleg-kozlyuk-grafana, line 129: nitpick: For N operands, this will result in N*log(N) string allocations - which is not necessarily terrible, but might be worth having a thought about.

pkg/traceql/normalize_test.go
oleg-kozlyuk-grafana, line 1: Needs more test cases, refer to `pkg/traceql/parse_test.go` for examples

modules/frontend/search_sharder.go
oleg-kozlyuk-grafana: I think that having an explicit mutating function `Normalize` is fine. `String` should be a pure (side-effect-free) function.

oleg-kozlyuk-grafana (changes requested):

pkg/traceql/normalize.go
Syedowais312: Agree
I’ll update this in a follow-up commit.

At the moment, normalization only mutates referenced fields rather than the pipeline element structs themselves, so value receivers are safe here. That said, I agree pointer receivers would make the mutation intent clearer and avoid unnecessary copying.

pkg/traceql/normalize.go
Syedowais312: Yes, You are right

 Spanset-level logical composition should also be normlize for commutative operators. I'll add deterministic ordering for `Spansetoperation` with AND /OR and cover it with tests

pkg/traceql/normalize.go
Syedowais312: Good point. This does introduce additional `String()` calls and allocations, but given the typical size of TraceQL expressions and the fact that normalization runs once per request for hashing, it felt acceptable for now.

If this ever shows up in profiling, we could revisit this with a cheaper structural key or a cached canonical representation.

Curious if you’d suggest a preferred approach here, @oleg-kozlyuk-grafana

Syedowais312:
Hey @oleg-kozlyuk-grafana,
PR is updated and ready for another review.

PR #6250, opened 2026-01-19
feat(distributor): add support for trace push middlewares

zeitlinger:
**What this PR does**:

Adds support for OTLP push middleware similar to https://github.com/grafana/mimir/pull/12666

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/distributor.go
+++ b/modules/distributor/distributor.go
@@ -188,7 +188,7 @@ type Distributor struct {
 
 	// TracePushMiddlewares are hooks called when a trace push request is received.
 	// Middleware errors are logged but don't fail the push (fail open behavior).
-	tracePushMiddlewares []receiver.TracePushMiddleware
+	tracePushMiddlewares []TracePushMiddleware
 
 	// For testing functionality that relies on timing without having to sleep in unit tests.
 	sleep func(time.Duration)
@@ -1070,3 +1070,8 @@ func (d *Distributor) RetryInfoEnabled(ctx context.Context) (bool, error) {
 	// cluster level is enabled, check per-tenant override and respect that.
 	return d.overrides.IngestionRetryInfoEnabled(userID), nil
 }
+
+// TracePushMiddleware is a hook called when a trace push request is received.
+// Middlewares are invoked after the request is decoded but before it's processed.
+// Errors returned by middleware are logged but don't fail the push (fail open behavior).
+type TracePushMiddleware func(ctx context.Context, td ptrace.Traces) error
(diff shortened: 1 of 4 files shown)


mapno:
Hi @zeitlinger, is this added with some use-cases already in mind?

modules/distributor/receiver/shim.go
```diff
@@ -349,6 +351,14 @@ func (r *receiversShim) ConsumeTraces(ctx context.Context, td ptrace.Traces) err
 	ctx, span := tracer.Start(ctx, "distributor.ConsumeTraces")
 	defer span.End()
 
+	// Call trace push middlewares
+	for _, mw := range r.tracePushMiddlewares {
```
joe-elliott: why put this in the shim and not just the `distributor.PushTraces` function?

zeitlinger: fixed

modules/distributor/distributor.go
```diff
@@ -411,6 +415,13 @@ func (d *Distributor) PushTraces(ctx context.Context, traces ptrace.Traces) (*te
 	ctx, span := tracer.Start(ctx, "distributor.PushBytes")
 	defer span.End()
 
+	// Call trace push middlewares
+	for _, mw := range d.tracePushMiddlewares {
+		if err := mw(ctx, traces); err != nil {
```
joe-elliott, line 421: should we allow push middleware to fail the write? for the given use case i think we should fail open. 

maybe for future use cases?

modules/distributor/receiver/middleware.go
```diff
@@ -83,3 +83,7 @@ func (m *multiTenancyMiddleware) Wrap(next consumer.Traces) consumer.Traces {
 		return next.ConsumeTraces(ctx, td)
 	})
 }
+
+// TracePushMiddleware is a hook called when a trace push request is received.
+// Middlewares are invoked after the request is decoded but before it's processed.
+type TracePushMiddleware func(ctx context.Context, td ptrace.Traces) error
```
joe-elliott: move up a folder to distributor.go?

modules/distributor/distributor.go
zeitlinger: fail open is a good idea - did that

modules/distributor/receiver/middleware.go
zeitlinger: done

zeitlinger:
@joe-elliott can you check again?

PR #6261, opened 2026-01-20
feat: add a new metric to measure batch weight

javiermolinar:
**What this PR does**:

It exposes a new metric histogram: `tempo_query_frontend_batch_weight_bucket`

Currently we have a histogram metric that tracks the number of elements of a batch. But it doesn't tell us the weight of that batch. Since now every job is weighted we need to add a second histogram to calculate the batch density.

ie: 

batch lenght 2, weight 5 (2 and 3)
batch lenght 6, weight 6 (1,1,1,1,2)

We want to maximize the ratio of batch_weight / MaxBatchSize. For that we need more data

**Checklist**
- [X] Tests updated
- [ ] Documentation added

### DIFF
--- a/std/null
+++ b/std/null
@@ -1,3 +1,0 @@
-modules/frontend/queue/user_queues.go:108:3: assignOp: replace `uid = uid + 1` with `uid++` (gocritic)
-		uid = uid + 1
-		^


std/null
```diff
@@ -0,0 +1,3 @@
+modules/frontend/queue/user_queues.go:108:3: assignOp: replace `uid = uid + 1` with `uid++` (gocritic)
```
joe-elliott: did you pipe a linter to `./std/null` :)?

javiermolinar: I though I deleted that thing

PR #6268, opened 2026-01-20
[jsonnet] Update memcached to latest

joe-elliott:
**What this PR does**:

Updates memcached and memcached-exporter to latest to catch any CVEs.


joe-elliott:
rollout-operator updates appeared when i ran `make jsonnet`. so apparently they are coming along for the ride.

PR #6271, opened 2026-01-21
Introduce PDB for several components

zalegrala:
**What this PR does**:

Here we include PDB resources to allow configuration of the Pod Disruption Budgets for several components, enabled through configuration.  This follows the approach which was added in #6224.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


operations/jsonnet/microservices/pod-disruption-budget.libsonnet
```diff
@@ -0,0 +1,93 @@
...
+  _config+:: {
+
+    backend_scheduler+: {
+      pdb: {
+        enabled: false,
+        max_unavailable: 1,
+      },
+    },
+
+    backend_worker+: {
```
javiermolinar, line 15: Is this not handled by the rollout operator?

zalegrala: Its a little nuance, but the PDB is a safety net in cases where the eviction API is used, which is not what is happening with the rollouts, for example.  If nodes are evicting pods, I don't think the rollout operator is helpful in maintaining a healthy number of replicas.  I think we need/want both.  The config here will allow us to enable/disable per component.  Additionally, not all components use the rollout-operator.

PR #6273, opened 2026-01-21
[Tempo 3.0] Remove compactor and v2 block encoding code

joe-elliott:
**What this PR does**:

In preparation for the 3.0 release this PR removes the compactor component and the v2 block encoding. This PR purposefully does not address docs b/c I believe those are being handled separately.

Note the jsonnet changes. This will impact our internal environments and we may require some cleanup after revendoring.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/backendworker/backendworker_test.go
+++ b/modules/backendworker/backendworker_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/grafana/dskit/flagext"
 	"github.com/grafana/dskit/kv"
+	"github.com/grafana/dskit/services"
 	backendscheduler_client "github.com/grafana/tempo/modules/backendscheduler/client"
 	"github.com/grafana/tempo/modules/overrides"
 	"github.com/grafana/tempo/modules/storage"
@@ -57,6 +58,9 @@ func TestWorker(t *testing.T) {
 	}
 
 	err = w.processJobs(ctx)
+	require.NoError(t, err)
+
+	err = services.StopAndAwaitTerminated(ctx, w)
 	require.NoError(t, err)
 }
 


modules/backendworker/backendworker.go
```diff
@@ -118,7 +118,7 @@ func New(cfg Config, schedulerClientCfg backendscheduler_client.Config, store st
 
 		w.Ring, err = ring.New(cfg.Ring.ToLifecyclerConfig().RingConfig, backendWorkerRingKey, cfg.OverrideRingKey, log.Logger, reg)
 		if err != nil {
-			return nil, fmt.Errorf("unable to initialize compactor ring: %w", err)
+			return nil, fmt.Errorf("unable to initialize backend-worker ring: %w", err)
```
zalegrala, line 119: 🙏

cmd/tempo/app/modules.go
```diff
@@ -885,16 +863,15 @@ func (t *App) setupModuleManager() error {
...
 		Querier:                       {Common, Store, IngesterRing, MetricsGeneratorRing, SecondaryIngesterRing, PartitionRing},
-		Compactor:                     {Common, Store, MemberlistKV},
 		BlockBuilder:                  {Common, Store, MemberlistKV, PartitionRing},
 		BackendScheduler:              {Common, Store},
 		BackendWorker:                 {Common, Store, MemberlistKV},
 		LiveStore:                     {Common, MemberlistKV, PartitionRing},
 
 		// composite targets
-		SingleBinary:         {Compactor, QueryFrontend, Querier, Ingester, Distributor, MetricsGenerator},
+		SingleBinary:         {QueryFrontend, Querier, Ingester, Distributor, MetricsGenerator},                                                 // jpe - remove
```
zalegrala: I think we will need the scheduler and the worker for this to work.

joe-elliott: this will technically break the `all` target in tip of main, but as long as we don't cut a release it's ok. as we continue with the cleanup, this will be renamed to be the new `all` target:

https://github.com/grafana/tempo/blob/main/cmd/tempo/app/modules.go#L898

onimisionipe:
Could you please update your docs as this change being part of the latest release broke tempo for us with `field compactor not found in type app.Config`

joe-elliott:
We are currently making large changes in preparation for 3.0 so it will be very difficult to follow tip of main. my advice would be to use 2.10 until we cut 3.0 at which point all the docs will be updated.

PR #6282, opened 2026-01-22
[vParquet5] Allow up to 20 dedicated string columns and add new 3% target heuristic

mdisibio:
**What this PR does**:
Doubles the max number of dedicated string columns in vParquet5 from 10 to 20.  And adds target percent heuristic to tempo-cli analysis and suggest commands, similar to how we did for integers.  It's not always beneficial to use all 20, so this is how we determine the optimum number. However the math works out differently such that the default cutoff is 3% for strings (whereas ints were 5%).

Because of how vParquet5 introduced dynamic columns, this is a backwards compatible change.  Tempo will gracefully read/compact blocks that were previously configured for only 10.

TODO for a further improvement:  right now we are taking the top N by "size" (count*length), but I suspect after all of this analysis, that top N by percent of rows could be better. 

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/overrides/runtime_config_overrides.go
+++ b/modules/overrides/runtime_config_overrides.go
@@ -26,7 +26,7 @@ import (
 )
 
 type Validator interface {
-	Validate(config *Overrides) (err error, warnings []error)
+	Validate(config *Overrides) (warnings []error, err error)
 }
 
 // perTenantOverrides represents the overrides config file
@@ -104,7 +104,7 @@ func loadPerTenantOverrides(validator Validator, typ ConfigType, expandEnv bool)
 
 		if validator != nil {
 			for tenant, tenantOverrides := range overrides.TenantLimits {
-				err, warnings := validator.Validate(tenantOverrides)
+				warnings, err := validator.Validate(tenantOverrides)
 				if err != nil {
 					return nil, fmt.Errorf("validating overrides for %s failed: %w", tenant, err)
 				}
--- a/cmd/tempo/app/config.go
+++ b/cmd/tempo/app/config.go
@@ -194,19 +194,18 @@ func (c *Config) CheckConfig() []ConfigWarning {
 		warnings = append(warnings, warnMCPServerEnabled)
 	}
 
-	if err, dcWarnings := c.StorageConfig.Trace.Block.DedicatedColumns.Validate(); err != nil || len(dcWarnings) > 0 {
-		if err != nil {
-			warnings = append(warnings, ConfigWarning{
-				Message: err.Error(),
-				Explain: "Tempo will not start with an invalid dedicated attribute column configuration",
-			})
-		}
-		for _, warning := range dcWarnings {
-			warnings = append(warnings, ConfigWarning{
-				Message: warning.Error(),
-				Explain: "Dedicated attribute column configuration contains an invalid configuration that will be ignored",
-			})
-		}
+	dcWarnings, dcErr := c.StorageConfig.Trace.Block.DedicatedColumns.Validate()
+	if dcErr != nil {
+		warnings = append(warnings, ConfigWarning{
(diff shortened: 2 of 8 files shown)


tempodb/encoding/vparquet5/schema.go
```diff
@@ -136,11 +136,26 @@ type DedicatedAttributes struct {
...
 	Int01    []int64  `parquet:",snappy,optional"`
 	Int02    []int64  `parquet:",snappy,optional"`
 	Int03    []int64  `parquet:",snappy,optional"`
 	Int04    []int64  `parquet:",snappy,optional"`
 	Int05    []int64  `parquet:",snappy,optional"`
+	Int06    []int64  `parquet:",snappy,optional"`
+	Int07    []int64  `parquet:",snappy,optional"`
+	Int08    []int64  `parquet:",snappy,optional"`
+	Int09    []int64  `parquet:",snappy,optional"`
+	Int10    []int64  `parquet:",snappy,optional"`
```
stoewer: The `DedicatedAttributes` contains 10 integer columns but in other places only 5 integer columns are handled. For example:
* `DedicatedResourceColumnPaths`
* `switch` statement in [`readValue()`](https://github.com/mdisibio/tempo/blob/8851542fc9aeee0fa91b3194b455563e20936ddb/tempodb/encoding/vparquet5/dedicated_columns.go#L163)
* `switch` statement in [`writeValue()`](https://github.com/mdisibio/tempo/blob/8851542fc9aeee0fa91b3194b455563e20936ddb/tempodb/encoding/vparquet5/dedicated_columns.go#L228)

mdisibio:
> The method `DedicatedColumns.Validate()` relies on the global variable `maxSupportedColumns` defined in [`block_meta.go`](https://github.com/grafana/tempo/blob/b5d7cc2bc351c4a7def5bf102d38a68aae92053c/tempodb/backend/block_meta.go#L44). This variable needs to be adjusted, otherwise dedicated columns with more than 10 string columns will not pass the validation

As discussed offline, this was working because we were doing validation only against the global/static dedicated columns, not the per-tenant overrides.  Added validation of the per-tenant overrides, but we decided that it would be better to treat too many columns as a warning and not a hard error. Pushed all of the plumbing to return the warning up the chain and log it.

modules/overrides/runtime_config_overrides.go
```diff
@@ -26,7 +26,7 @@ import (
 )
 
 type Validator interface {
-	Validate(config *Overrides) error
+	Validate(config *Overrides) (warnings []error, err error)
```
stoewer, line 29: `nitpick` dskit has a package [multierror](https://pkg.go.dev/github.com/grafana/dskit@v0.0.0-20260123032120-fb7d1aad51e6/multierror) that can be used to return an error that consists of multiple errors. Using this would allow us to have a slightly more idiomatic signature:

```suggestion
	Validate(config *Overrides) (warn error, err error)
```

stoewer (approved):
LGTM

modules/overrides/runtime_config_overrides.go
mdisibio: I like the current ability to inspect and log each warning independently.  But no strong opinions either way. What do you think about returning a multierror directly `warn MultiError, err error`, then we could simplify the signature and also still inspect the individual warnings?

mdisibio: Thinking about this more, MultiError helps be idiomatic with returning only one error, because functions should fail fast and early.  But bubbling multiple warnings up the chain seems helpful, actually I hope this is something we can do more of.  So let's keep it?  I was looking around to see if there was a common style for warnings, we can revisit it if we want.

PR #6285, opened 2026-01-23
change: increase default max duration for metrics queries to one day

javiermolinar:
**What this PR does**:
This PR raises the default max duration for TraceQL metrics from 3 h to 24 h while keeping a safety cap in place. The previous 3 h limit was chosen to avoid OOMs and client-side timeouts on large data sets.

Tempo 3.0’s new architecture makes TraceQL metrics non-experimental and eliminates data-loss risk, and recent traceql + parquet performance gains let us raise the default without removing the ceiling.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


std/null
oleg-kozlyuk-grafana, line 1: This looks strange

oleg-kozlyuk-grafana:
The change looks good, except for strange `std/null` file

std/null
javiermolinar: removed

PR #6286, opened 2026-01-23
feat(frontend): add traceID to query range response logs

### LINKED ISSUE #6280: Include trace ID in query-frontend response logs
The query frontend logs helpful messages for every API response like this:

```go
	level.Info(logger).Log(
		"msg", "query range response",
		"tenant", tenantID,
		"query", req.Query,
		"range_nanos", req.End-req.Start,
		"max_series", req.MaxSeries,
		"duration_seconds", durationSeconds,
		"request_throughput", float64(resp.Metrics.InspectedBytes)/durationSeconds,
		"total_requests", resp.Metrics.TotalJobs,
		"total_blockBytes", resp.Metrics.TotalBlockBytes,
		"total_blocks", resp.Metrics.TotalBlocks,
		"completed_requests", resp.Metrics.CompletedJobs,
		"inspected_bytes", resp.Metrics.InspectedBytes,
		"inspected_traces", resp.Metrics.InspectedTraces,
		"inspected_spans", resp.Metrics.InspectedSpans,
		"partial_status", resp.Status,
		"partial_message", resp.Message,
		"num_response_series", len(resp.Series),
		"error", err)
```

We need to include trace ID, so we can easily navigate from the log to the trace for that request.  This is probably a straight-forward fix because we already have traces along the read path, but I'd like to double check - was this a simple miss, or is there a reason it wasn't included in the first place?

FaqihMSY:
**What this PR does**:
Adds `traceID` field to all query range response log statements in the query frontend. This enables easy navigation from logs to traces for that request, following the existing pattern in `handler.go`.

Fixes #6280

**Verification**:
Verified manually via local compilation on Windows:
-  `gofmt -w` formatting applied
-  `go build ./modules/frontend/` -> PASSED
-  `go test ./modules/frontend/` -> PASSED (no tests to run, 0.192s)
-   No changes to dependencies (go.mod/vendor)

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 * [CHANGE] Set traceQL query metrics checks by default in Vulture [#6275](https://github.com/grafana/tempo/pull/6275) (@javiermolinar)
 * [FEATURE] Add span_multiplier_key to overrides. This allows tenants to specify the attribute key used for span multiplier values to compensate for head-based sampling. [#6260](https://github.com/grafana/tempo/pull/6260) (@carles-grafana)
-* [ENHANCEMENT] Add traceID to query range response logs for easier navigation from logs to traces [#<PR_NUMBER>](https://github.com/grafana/tempo/pull/<PR_NUMBER>) (@FaqihMSY)
 * [BUGFIX] Correct avg_over_time calculation [#6252](https://github.com/grafana/tempo/pull/6252) (@ruslan-mikhailov)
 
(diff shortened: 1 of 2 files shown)


javiermolinar:
@FaqihMSY this looks good. I think we should include it on the other handlers too for consistency:

- metrics_query_handler
- search_handler
- tags_handler
- traceid_handler


What do you think?

PR #6288, opened 2026-01-23
Remove duplicate dimension validation

carles-grafana:
**What this PR does**:

we have identified some valid use cases such as using multiple SDKs using different conventions,
such as {"deployment_environment", "deployment.environment"}

it still makes sense to keep it for dimension mappings as these can cause confusion

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## main / unreleased
 
+* [CHANGE] Allow duplicate dimensions for span metrics and service graphs. This is a valid use case if using different instrumentation libraries, with spans having "deployment.environment" and others "deployment_environment", for example. [#6288](https://github.com/grafana/tempo/pull/6288) (@carles-grafana)
 * [CHANGE] Updade default max duration for traceql metrics queries up to one day [#6285](https://github.com/grafana/tempo/pull/6285) (@javiermolinar)
 * [CHANGE] Set traceQL query metrics checks by default in Vulture [#6275](https://github.com/grafana/tempo/pull/6275) (@javiermolinar)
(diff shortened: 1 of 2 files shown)


mdisibio:
To make sure I understand, this config is no longer logged as a warning nor an error.  Therefore we can remove the plumbing to return `warn` because there are no other warnings.  

And then for this config, it sums the 2 inputs together.  e.g. `deployment_environment` and `deployment.environment` are both counted into the final metric `deployment_environment` ?

carles-grafana:
> To make sure I understand, this config is no longer logged as a warning nor an error. Therefore we can remove the plumbing to return `warn` because there are no other warnings.

correct
 
> And then for this config, it sums the 2 inputs together. e.g. `deployment_environment` and `deployment.environment` are both counted into the final metric `deployment_environment` ?

yes, but if a span contains both, one will overwrite the other. This matches the previous behavior. This would be an issue in the user's instrumentation. But having both attributes but in different spans can be a valid use case if instrumented differently.

mdisibio:
+1 This is the approach we take for cost attribution, and I think it's good because it unlocks more use cases.

Good catch on the case when a span contains both - when this happens will it be deterministic which label is used?  First or last both ok, as long as it is deterministic and explainable.  Then it also allows for a tenant to change the behavior by swapping the order of the labels to whichever they prefer.  But if it is random from a map and the data becomes unreliable and there is no fix except change their instrumentation, maybe let's think about if we can fix that.  

Changelog?

carles-grafana:
@mdisibio 

it's deterministic: if there is a collision, for example deployment.environment and deployment_environment, both sanitized to deployment_environment, we just take the last one configured.
see https://github.com/carles-grafana/tempo/blob/remove-duplicate-dimension-validation/modules/generator/processor/spanmetrics/spanmetrics.go#L170

updated changelog

PR #6313, opened 2026-01-27
feature: add automemlimit flag to automatically tune Go GC

oleg-kozlyuk-grafana:
**What this PR does**:

Introduces dynamic Go GC reconfiguration based on CGroups values (as set by Kubernetes). Has same effect as GOMEMLIMIT, but is updated dynamically as CGroups change. Should allow to disable ballast and allow go's GC optimizations to work

Disabled by default

**Checklist**
- [ ] Tests updated
- [x] Documentation added

### DIFF
--- a/docs/sources/tempo/configuration/_index.md
+++ b/docs/sources/tempo/configuration/_index.md
@@ -141,7 +141,9 @@ server:
 ## Memory
 
 Tempo supports automatic GOMEMLIMIT configuration using the [automemlimit](https://github.com/KimMachineGun/automemlimit) library.
-When enabled, it automatically sets Go's memory limit based on available container or system memory.
+When enabled, it automatically sets Go's memory limit based on available container (via CGroups) or system memory every 15 seconds.
+
+NOTE: enabling this will override value set in GOMEMLIMIT environment variable
 
 ```yaml
 memory:
--- a/cmd/tempo/main.go
+++ b/cmd/tempo/main.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"reflect"
 	"runtime"
+	"time"
 
 	"github.com/KimMachineGun/automemlimit/memlimit"
 	"github.com/drone/envsubst"
@@ -237,6 +238,7 @@ func initAutoMemLimit(config *app.Config) {
 				memlimit.FromSystem,
 			),
 		),
+		memlimit.WithRefreshInterval(15*time.Second),
 		memlimit.WithLogger(logger),
 	)
 	if err != nil {


pkg/parquetquery/predicates.gen.go
oleg-kozlyuk-grafana, line 1: `make fmt` touched this file as usual, will revert

cmd/tempo/app/config.go
```diff
@@ -81,6 +88,12 @@ func (c *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet) {
 	c.StreamOverHTTPEnabled = false
 	c.PartitionRingLiveStore = false
 
+	// Memory settings
+	c.Memory = MemoryConfig{
+		AutoMemLimitEnabled: false,
```
mattdurham, line 93: Can we run it with true by default? Or if we cant yet can we add a todo/issue in 3.0 to make it the default?

docs/sources/tempo/configuration/_index.md
```diff
@@ -137,6 +138,23 @@ server:
...
 ```
 
+## Memory
+
+Tempo supports automatic GOMEMLIMIT configuration using the [automemlimit](https://github.com/KimMachineGun/automemlimit) library.
+When enabled, it automatically sets Go's memory limit based on available container or system memory.
+
+```yaml
+memory:
+    # Enable automatic GOMEMLIMIT configuration based on cgroup/system memory.
```
mattdurham, line 150: Whats the precedence for this vs setting GOMEMLIMIT via env? Should we document that?

oleg-kozlyuk-grafana: This performs same function as GOMEMLIMIT but dynamically based on available CGroups limits. I.e. it's "the same", but automatic and is compatible with in-place VPA (https://kubernetes.io/blog/2025/12/19/kubernetes-v1-35-in-place-pod-resize-ga/) - which GOMEMLIMIT, being static, is not

cmd/tempo/app/config.go
oleg-kozlyuk-grafana: I do want to, but I want to test it in our envs first

docs/sources/tempo/configuration/_index.md
mattdurham: IE lets say I have both set, which one wins?

oleg-kozlyuk-grafana:
Hey @codex, review this for me

docs/sources/tempo/configuration/_index.md
oleg-kozlyuk-grafana: GOMEMLIMIT initially, then this will take over for the lifetime of the process. I see your point, worth adding this detail in the docs.

oleg-kozlyuk-grafana: Added to doc

PR #6338, opened 2026-01-29
[bugfix] Fix live-store deadlock after storage failure

ruslan-mikhailov:
**What this PR does**:
**tl;dr** Tempo did not enqueue complete block job for WAL if it failed once

The PR fixes deadlock in live-store after storage failure. When WAL block flushing errored, live-store queues it back with a back off. However, as the key is not cleared, it drops it:
```
// Enqueue adds the op to the next queue and prevents any other items to be added with this key
func (f *ExclusiveQueues[T]) Enqueue(op T) error {
	_, ok := f.activeKeys.Load(op.Key())
	if ok {
		return nil // <-------- here
	}

	f.activeKeys.Store(op.Key(), struct{}{})
	return f.Requeue(op)
}
```
After some time, second WAL block can be created that will stuck on backpressure blocking whole partition queue until old WAL block is cleared.

In order to correctly enqueue failed WAL block, need to call `Requeue` method.

See `TestExclusiveQueueLocks` for an example of the queue behaviour.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/livestore/live_store_background.go
+++ b/modules/livestore/live_store_background.go
@@ -15,20 +15,20 @@ import (
 	"github.com/grafana/tempo/tempodb/encoding"
 )
 
-var initialBackoff = 30 * time.Second
-
 const (
-	maxBackoff       = 120 * time.Second
-	maxFlushAttempts = 10
+	defaultInitialBackoff = 30 * time.Second
+	defaultMaxBackoff     = 120 * time.Second
+	maxFlushAttempts      = 10
 )
 
 type completeOp struct {
 	tenantID string
 	blockID  uuid.UUID
 
-	at       time.Time
-	attempts int
-	bo       time.Duration
+	at         time.Time
+	attempts   int
+	bo         time.Duration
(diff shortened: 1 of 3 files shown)


modules/livestore/live_store_background.go
```diff
@@ -15,9 +15,10 @@ import (
 	"github.com/grafana/tempo/tempodb/encoding"
 )
 
+var initialBackoff = 30 * time.Second
```
mdisibio: What do you think about making this an option, rather than expose global vars for tests?  It doesn't have to be exposed to yaml.  And then rename this const to `defaultInitialBackoff`.

mdisibio: To add, there is precedent for this approach to have internal settings, they look like this:

```go
Poll                 bool          `yaml:"-"`
```

ruslan-mikhailov: Good idea! Added in a separate commit.

Issue #6279, opened 2026-01-22
Use proto for dedicated columns in proto tenant index


ruslan-mikhailov:
For tenant index proto, we marshal/unmarshal dedicated columns as JSONs:

https://github.com/ruslan-mikhailov/tempo/blob/6e50317d8a67828d5b4b112d3cd795ec1275e445/tempodb/backend/block_meta.go#L405-L413

According to cpu profiles unmarshalling dedicated column consumes 10-15% of cpu time in querier and query-frontend. This PR optimises it by using a faster JSON library: https://github.com/grafana/tempo/pull/6262

We need to try to use proto instead of JSON and compare benchmarks. The code should be backward-compatible: if Tempo reads an index with JSON dedicated columns, it still should unmarshal them as JSON.

mdisibio:
The proto field is a plain string (byte array).  We could migrate to proto seamlessly by checking the first char in DedicatedColumns.Unmarshal().  If `{` or `[` then it's json, else it's proto.

rofaidamosa:
Hi @ruslan-mikhailov   I'd like to work on this issue if it's still available.

Syedowais312:
Hi @mdisibio @ruslin-mikhailov 

I’ve been working on refactoring `DedicatedColumns` to avoid storing JSON blobs
inside protobuf fields and instead encode them via `tempopb.DedicatedColumn`.

Most of the changes live in `backend/block_meta.go`. However, due to the use of
`gogoproto.customtype`, the generated `v1.pb.go` code expects
`DedicatedColumns` to implement `Size`, `MarshalTo`, and `Unmarshal`.

Reverting any manual changes to `v1.pb.go` causes CI failures, but editing
generated files directly is obviously not desirable.

Before proceeding further, I wanted to ask for guidance on the preferred
approach here. Would you recommend:

1. Introducing a proper protobuf wrapper message for `DedicatedColumns`
2. Regenerating protobufs after updating the `.proto` definition

Happy to align with whichever direction you prefer.

PR #6347, opened 2026-01-29
Draft: Refactor DedicatedColumns encoding 

Syedowais312:
## Context

This draft PR explores refactoring `DedicatedColumns` to avoid storing JSON
blobs inside protobuf fields and instead encode them via
`tempopb.DedicatedColumn`.

Most of the changes live in `tempodb/backend/block_meta.go`.

## Problem

Because `DedicatedColumns` is used via `gogoproto.customtype`, the generated
`v1.pb.go` code expects the type to implement `Size`, `MarshalTo`, and
`Unmarshal`.

Reverting any manual edits to generated files causes CI failures, but editing
generated protobuf files directly is not desirable.

issue: #6279


oleg-kozlyuk-grafana:
Greetings,

Thank you for your contribution proposal. I'd like to clarify two questions before further discussion:
* What is the intent of this change? What improvements do you expect to achieve (performance improvements, storage savings, ...?)
* What is your strategy w.r.t. backward compatibility here?

Syedowais312:
> Greetings,
> 
> Thank you for your contribution proposal. I'd like to clarify two questions before further discussion:
> 
>     * What is the intent of this change? What improvements do you expect to achieve (performance improvements, storage savings, ...?)
> 
>     * What is your strategy w.r.t. backward compatibility here?


The intent aligns with #6279:

- Replace JSON encoding of `DedicatedColumns` with proto-based encoding
- Reduce CPU overhead during unmarshalling (currently ~10–15% in some paths)
- Keep full backward compatibility by detecting `{` / `[` and falling back to JSON

New blocks write proto.
Old blocks (JSON) are still readable.

PR #6350, opened 2026-01-30
tempodb: record bloom filter size in block metadata

### LINKED ISSUE #6330: Add tempodb backend metric for size of bloom filters
Tempo maintains several metrics about the total size of blocks and traces in object storage.  Let's add a new one that records the total number of bloom filter bytes. This will be helpful for sizing the bloom filter cache.

We record the metrics in the poller [here](https://github.com/grafana/tempo/blob/main/tempodb/blocklist/poller.go#L604), where we have all of the block metas in memory.  We can easily access the [number of bloom shards in each block](https://github.com/grafana/tempo/blob/main/tempodb/backend/v1/v1.proto#L23).

However we don't have the exact bloom filter shard size in bytes recorded per block. This field has been skipped because it isn't necessary for the read path.  It's tempting to record the metric as an estimate, using the current config of bloom shard bytes * count.  But this config exists in various places (compaction, backend-worker, block-builder). So let's add it as a new field in the proto and them sum it for the metric.

A possible good metric is `tempodb_backend_bloom_bytes_total`.

**Additional context**

Syedowais312:
### What this PR does
- Records the exact serialized size of bloom filters in `BlockMeta` at write time
- Persists this value for all parquet encodings (v3–v5)
- Preserves backward compatibility (older blocks report `0`)

### Why
- Bloom filter storage usage could not be measured accurately
- Config-based estimation is error-prone and version-dependent
- `BlockMeta` is the authoritative source for backend storage metrics

### Implementation details
- Computes bloom size by summing serialized shard sizes at block creation/compaction
- Stores the value as `BloomFilterBytes` in block metadata
- Extends poller aggregation to expose total bloom bytes as a backend metric

### Testing
- `make test-with-cover-tempodb` passes
- Local `covdata` error occurs after successful test execution and is environment-specific

Fixes #6330


Syedowais312:
Hey @mdisibio, can you please review this PR.

oleg-kozlyuk-grafana:
Greetings.

Thank you for submitting your contribution.

This PR description states that it will close #6330. However, #6330 requests addition of a metric, not (just) a field in metadata. I.e. if we merge this PR, it will not achieve the target of the issue.

Can you clarify your intent, please?

javiermolinar (commented):
BlockMetas is already quite big, what's the use case of this?

Syedowais312:
> Greetings.
> 
> Thank you for submitting your contribution.
> 
> This PR description states that it will close #6330. However, #6330 requests addition of a metric, not (just) a field in metadata. I.e. if we merge this PR, it will not achieve the target of the issue.
> 
> Can you clarify your intent, please?

Thanks for the review and pointing this out @oleg-kozlyuk-grafana.

You're right, this PR currently adds the missing ground-truth bloom filter size to BLockMeta, but it doesn't not yet expose the backend metric as requested in #6330.

My intention for to implement this into 2 steps:
1. Record the exact bloom filter size at the block crreation/compaction time (current changes)
2. Add a poller-level meric that aggregates this fied across blocks

I agree as this PR alone does not fully satisfy #6330, im current working on to add the `tempodb_backend_bloom_bytes_total` metric to adress the remaining part.

Syedowais312:
> BlockMetas is already quite big, what's the use case of this?

I agree `BlockMeta` should not grow arbitrarily.

But the motivation is that the pollar already operates entirely on inmemory `blockMeta` objects to compute backend-level storage metrics (block bytes, trace counts, etc..)

From my understanding Bloom filter size is only known precisely at the block creation/compaction time and cannot be reconstructed later without re-reading bloom objects or estimating from config (which varies across versions and components).

Storing the exact bloom byte sizze in `blockMeta` allows to
- accurate backend level aggregation in the pollar
- no additional object storage reads
- backward compatibility (older blocks simply report zero)

The intent is not to grow the BlockMeta arbitrarily, but to support concrete operational metric for sizing the bloom cache

tempodb/encoding/common/bloom.go
```diff
@@ -70,6 +70,21 @@ func (b *ShardedBloomFilter) Marshal() ([][]byte, error) {
 	return bloomBytes, nil
 }
 
+func (b *ShardedBloomFilter) SerializedSizeBytes() (uint64, error) {
+	var total uint64
+
+	for _, f := range b.blooms {
+		buf := &bytes.Buffer{}
+		_, err := f.WriteTo(buf)
```
Syedowais312: This cost is paid only once at block creation/compaction time and not during polling or query paths.

Syedowais312:
@oleg-kozlyuk-grafana 
Updated the PR description and addressed the feedback. Ready for another review

tempodb/encoding/common/bloom.go
oleg-kozlyuk-grafana: I am not a fan of doing this at all, given performance implications.

In `copy.go` the length is taken directly, can we just use that approach everywhere? 
```
	for _, b := range blooms {
		bloomBytes += uint64(len(b))
	}
```

oleg-kozlyuk-grafana:
Thanks for addressing the comments @Syedowais312. I have one more regarding `SerializedSizeBytes` - I'm not confident in that approach. Other than that, this is much closer to mergeable form now

tempodb/backend/v1.pb.go
```diff
@@ -45,6 +45,7 @@ type BlockMeta struct {
 	DedicatedColumns DedicatedColumns `protobuf:"bytes,17,opt,name=dedicated_columns,json=dedicatedColumns,proto3,customtype=DedicatedColumns" json:"dedicatedColumns,omitempty"`
 	// repeated bytes dedicated_columns = 17 [(gogoproto.customtype) = "DedicatedColumn", (gogoproto.jsontag) = "dedicatedColumns,omitempty", (gogoproto.nullable) = false];
 	ReplicationFactor uint32 `protobuf:"varint,18,opt,name=replication_factor,json=replicationFactor,proto3" json:"replicationFactor,omitempty"`
+	BloomFilterBytes  uint64 `protobuf:"varint,19,opt,name=bloom_filter_bytes,json=bloomFilterBytes,proto3" json:"bloomBytes"`
```
zhxiaogg: Just came across this PR, thanks for the great work!

One minor thing, I feel it would be great if the namings could be more consistent.

Syedowais312: Done [f42cc92](https://github.com/grafana/tempo/pull/6350/commits/f42cc92330093a9785d2dfb1375c9942eb684715)

tempodb/encoding/common/bloom.go
Syedowais312: That makes sense, i have updated it

Syedowais312:
Hi @oleg-kozlyuk-grafana, hope you're doing well! It's been a while since the last review, and the PR was automatically closed by the stale GitHub Action. Could you let me know if this change is still something the team would like to have? If so, I'm happy to update the PR based on the remaining feedback. Thanks!

PR #6360, opened 2026-02-02
[bugfix] Correct end time adjustment for query range

ruslan-mikhailov:
**What this PR does**: fixes `QueryEndCutoff` for query range. Query frontend initialises query range handler with combiners and alignment first, which leads to two bugs: empty buckets after cutoff and ignoring the param if end is not aligned.

Before:
image
After:
Pasted image 20260202071617

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added


modules/frontend/frontend.go
```diff
@@ -229,7 +229,7 @@ func New(cfg Config, next pipeline.RoundTripper, o overrides.Interface, reader t
 	queryRangePipeline := pipeline.Build(
 		[]pipeline.AsyncMiddleware[combiner.PipelineResponse]{
 			headerStripWare,
-			adjustEndWareNanos,
+			// adjustEndWareNanos, // due to alignments and combiner, it needs to be done in handler
```
javiermolinar: This is not just for clamping the end, it also provides a default start and end. The idea is to simplify the existing logic by relaying on it. Is there a way to keep it?

modules/frontend/metrics_query_range_handler.go
```diff
@@ -50,8 +50,20 @@ func newQueryRangeStreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTripp
...
+
 		traceql.AlignRequest(req)
 
+		// the end time cutoff is applied here because it has to be done before combiner creation
+		if cfg.QueryEndCutoff > 0 {
+			now := time.Now()
+			maxEnd := now.Add(-cfg.QueryEndCutoff)
+			reqEnd := time.Unix(0, int64(req.End))
+			if reqEnd.After(maxEnd) {
+				req.End = uint64(maxEnd.UnixNano())
```
javiermolinar, line 63: What's wrong with the existing reqEnd? The code does something similar, substracting the EndCutOff to time.Now()

modules/frontend/frontend.go
ruslan-mikhailov: Summary of the discussion: currently, it is not possible to do. In order to support the middleware is to force combiner initialisation only after all handlers are applied. There are a couple of ways to do it, but it's not a minor refactoring, so we decided to do it separately.

ruslan-mikhailov: Added ToDos

PR #6363, opened 2026-02-02
[enhancement] Fail if live store cannot guarantee correctness due to lag

ruslan-mikhailov:
**What this PR does**: it introduces new config paramater `fail_on_high_lag` for live-store. If it is enabled and live-store receives query range or search request with end parameter later than lag, it fails the request. It gives preference to consistency over availability. If an instance in one zone returns an error, while second is not lagged and fine, the final response won't fail. But if at least one partition has all live-store zones lagged, the request will fail. 

It **requires** https://github.com/grafana/tempo/pull/6360 to be merged, because due to the bug, live-store receives recent queries without the cutoff.

**How has it been tested:** unit tests and manual tests.
For manual tests, I added a deadlock inside consumption loop for one zone and waited ~10 minutes.
```
+var (
+       startup    = time.Now()
+       laggyAfter = 5 * time.Minute
+)
+
 func (s *LiveStore) consume(ctx context.Context, rs recordIter, now time.Time) (*kadm.Offset, error) {
        defer s.decoder.Reset()
        _, span := tracer.Start(ctx, "LiveStore.consume")
@@ -516,6 +521,10 @@ func (s *LiveStore) consume(ctx context.Context, rs recordIter, now time.Time) (
        cutoff := now.Add(-s.cfg.CompleteBlockTimeout)
        // Process records by tenant
        for !rs.Done() {
+               if s.reader.consumerGroup == "live-store-zone-a" && time.Since(startup) > laggyAfter {
+                       println("DEBUG Went to forever sleep mode")
+                       time.Sleep(100 * time.Hour)
+               }
```
Then run TraceQL Metrics query (e.g. `{} | count_over_time()`) multiple times. If `fail_on_high_lag` is disabled, half of the times it should return incorrect results (for `count_over_time`, drop to zero). If `fail_on_high_lag` is enabled, always correct results, `cannot guarantee complete results` errors in logs for the lagged zone.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/livestore/live_store.go
+++ b/modules/livestore/live_store.go
@@ -473,6 +473,7 @@ func (s *LiveStore) waitForCatchUp(ctx context.Context) error {
 // - we know the watermark but nothing has been consumed yet = indeterminate
 //
 // It takes lagShortcutThreshold to shortcut calculations if the lag is close to the end of the partition.
+// To disable the shortcut, set lagShortcutThreshold to a negative value.
 func (s *LiveStore) calculateTimeLag(lagShortcutThreshold int64) *time.Duration {
 	// Use cached high watermark from fetch responses (avoids extra API call)
 	lag := s.reader.lag.Load()
@@ -486,7 +487,7 @@ func (s *LiveStore) calculateTimeLag(lagShortcutThreshold int64) *time.Duration 
 
 	// Check if we are near end or partition is empty
 	// Arbitrary value picked to shortcut calculations
-	if lagShortcutThreshold > 0 && lag <= lagShortcutThreshold {
+	if lagShortcutThreshold >= 0 && lag <= lagShortcutThreshold {
 		level.Debug(s.logger).Log(
 			"msg", "At or close to partition end",
 			"lag", lag)


modules/livestore/live_store.go
ruslan-mikhailov, line 495: This change is a bit controversial and I would like to hear your opinion. This shortcut helps in case of clock skew. However, if live-store is stuck for any reason (cannot connect to Kafka or a deadlock), the lag value and `lastRecordTimeNanos` won't be updated. That's why I prefer to use `lastRecordTimeNanos`

docs/sources/tempo/configuration/manifest.md
```diff
@@ -1421,4 +1418,5 @@ live_store:
               options: []
     readiness_target_lag: 0s
     readiness_max_wait: 30m0s
+    fail_on_high_lag: false
```
ruslan-mikhailov, line 1424: I think, in 3.0 it should be always true. But the change is dangerous, and I would like to rollout it carefully.

ruslan-mikhailov, line 1424: Should it be time.Duration or we will rely on `QueryEndCutoff`, that should be always set correctly in query-frontend?

modules/livestore/live_store.go
```diff
@@ -734,11 +739,27 @@ func (s *LiveStore) GetMetrics(_ context.Context, _ *tempopb.SpanMetricsRequest)
 
 // QueryRange implements tempopb.MetricsGeneratorServer
 func (s *LiveStore) QueryRange(ctx context.Context, req *tempopb.QueryRangeRequest) (*tempopb.QueryRangeResponse, error) {
+	if s.isLagged(int64(req.End)) {
```
ruslan-mikhailov: Note that I added the check only for two endpoints (+instant query) where correctness is most important. Should it be for all or some other methods too?

modules/livestore/live_store.go
```diff
@@ -484,7 +486,7 @@ func (s *LiveStore) calculateTimeLag() *time.Duration {
 
 	// Check if we are near end or partition is empty
 	// Arbitrary value picked to shortcut calculations
-	if lag <= 1000 {
+	if lagShortcut && lag <= 1000 {
```
mattdurham: I know this existed before, but I hate the unknown time scale on 1000, what does that actually mean and can we move it to a constant?

modules/livestore/live_store.go
```diff
@@ -471,7 +471,9 @@ func (s *LiveStore) waitForCatchUp(ctx context.Context) error {
 // - empty partition = no lag
 // - nothing has been fetched yet = indeterminate
 // - we know the watermark but nothing has been consumed yet = indeterminate
-func (s *LiveStore) calculateTimeLag() *time.Duration {
+//
+// It takes lagShortcut to shortcut calculations if the lag is close to the end of the partition.
+func (s *LiveStore) calculateTimeLag(lagShortcut bool) *time.Duration {
```
mattdurham: Instead of a bool, how about passing in the lag allowance you are comfortable with?

modules/livestore/live_store.go
```diff
@@ -684,6 +686,9 @@ func (s *LiveStore) FindTraceByID(ctx context.Context, req *tempopb.TraceByIDReq
 
 // SearchRecent implements tempopb.Querier
 func (s *LiveStore) SearchRecent(ctx context.Context, req *tempopb.SearchRequest) (*tempopb.SearchResponse, error) {
+	if s.isLagged(int64(req.End) * 1e9) {
```
mattdurham: 1e9 means what? Probably its converting to seconds?

modules/livestore/live_store.go
mattdurham: Does this not need 1e9?!

modules/livestore/live_store.go
ruslan-mikhailov: > what does that actually mean
If a consumer is less than 1000 messages (not seconds!) behind, we automatically assume it's not laggy.

modules/livestore/live_store.go
ruslan-mikhailov: Good idea, changed

modules/livestore/live_store.go
ruslan-mikhailov: Vice versa: from seconds to nanos. Search accepts seconds, while query range - nanos. I've added comments in code. it looks confusing, I agree 😅

modules/livestore/live_store.go
mattdurham: Do we know in a low volume environment how far 1000 messages of backlog is? IE is that 10 minutes in our smallest environment or 30s?

modules/livestore/live_store.go
ruslan-mikhailov: Answered here: https://github.com/grafana/tempo/pull/6363#discussion_r2759065441

modules/livestore/live_store.go
mapno: Another possibility is that there are no new records to consume—for example if a live-store restarts after an inactive partition has already been drained but needs to remain inactive for more time. That's why depending on `lastRecordTimeNanos` can be problematic.

mapno: I have to admit I'm no fan of this heuristic, but it helps identify when lag is "basically zero"

modules/livestore/live_store.go
oleg-kozlyuk-grafana: I've added that, this is about ~1s in our environments and the purpose is to avoid deadlocking on startup in case of Warpstream time skew. Particular number is not very important, but ideally should be > normal operation lag (which is ~5-10 I think)

ruslan-mikhailov:
There is an ongoing discussion around `calculateTimeLag` behind the curtain. I convert the PR to Draft until we reach consensus.

ruslan-mikhailov:
\+ rebase from latest main to include latest fixes for lag calculation

PR #6371, opened 2026-02-03
livestore: make trace_too_large log line an insight

carles-grafana:
**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/livestore/instance_test.go
+++ b/modules/livestore/instance_test.go
@@ -1,10 +1,12 @@
 package livestore
 
 import (
+	"bytes"
 	"context"
 	"testing"
 	"time"
 
+	"github.com/go-kit/log"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


modules/livestore/instance.go
```diff
@@ -224,7 +224,7 @@ func (i *instance) pushBytes(ctx context.Context, ts time.Time, req *tempopb.Pus
 			traceSz := trace.Size()
 			allowResult := i.traceSizes.Allow(traceID, traceSz, maxBytes)
 			if !allowResult.IsAllowed {
-				i.maxTraceLogger.Log("msg", overrides.ErrorPrefixTraceTooLarge, "max", maxBytes, "traceSz", traceSz, "totalSize", allowResult.CurrentTotalSize, "trace", hex.EncodeToString(traceID))
+				i.maxTraceLogger.Log("msg", overrides.ErrorPrefixTraceTooLarge, "max", maxBytes, "traceSz", traceSz, "totalSize", allowResult.CurrentTotalSize, "trace", hex.EncodeToString(traceID), "insight", true)
```
electron0zero, line 227: can we add a test to check for the `insight=true` in the log line.

carles-grafana: we could but it's a trivial change, it's overkill in my opinion. The test would be much more complex than the change

PR #6380, opened 2026-02-04
enhancement: Remove runtime.GC calls from hot path

oleg-kozlyuk-grafana:
**What this PR does**:

Removes `runtime.GC` calls from hot path, as we already have `GOMEMLIMIT` set

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/encoding/vparquet4/compactor.go
+++ b/tempodb/encoding/vparquet4/compactor.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"runtime"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


mdisibio:
I'm ok with trying to remove this, but can you test this out in ops?    It was helpful at the time.  To clarify this is calling GC each time we flush a row group, roughly every 100MB, so it's not too often.

joe-elliott:
Thank you for removing these! I added these as a total desperation move when compactor's were super unstable in the early parquet days.

oleg-kozlyuk-grafana:
The reason I'm looking at these now is because both places are on the hot path of live-store. I'm still gathering data on the effects of the change, but I would expect at least some latency reduction (if not CPU).

mdisibio:
> hot path of live-store

+1 agree, that is much more latency sensitive.  Interested to see the results. There is a chance we will need to remove this from the live-store, but leave it for compaction (backend-worker).

PR #6392, opened 2026-02-05
feat: add new include_any filter policy for spanmetrics filter

javiermolinar:
**What this PR does**:
This PR introduces a new filter policy type, `include_any`. This allows for a any span matching an `include_any` rule to be included.

**This new filter policy is retrocompatible with the existing logic.** So existing filters must work in the same way as today

**The problem** 
The existing `include` policy is very rigid. Multiple include policies act as a logical AND. This means all conditions must be fulfilled for a span to be accepted. For example:

```yaml
 filter_policies:
        - include:
            match_type: regex
            attributes:
              - key: resource.location
                value: eu-.*
         - include:
            match_type: strict
            attributes:
              - key: service.name
                value: foo

```

This configuration only matches spans where the service.name is "foo" and the resource.location starts with "eu". Any span failing either condition is dropped.

This makes certainscenarios impossible, such as filtering by span kind to reduce cardinality while wanting to include INTERNAL spans for a specific service:

```yaml
 filter_policies:
        - include:
            match_type: regex
            attributes:
              - key: kind
              value: SPAN_KIND_(SERVER|CONSUMER)
         - include:
            match_type: strict
            attributes:
              - key: kind
                value: SPAN_KIND_(INTERNAL)
              - key: service.name
                value: foo

```

A span cannot be SERVER and INTERNAL at the same time, so this configuration would drop 100% of the spans.

**The solution** 

The new include_any policy acts as a logical OR relative to other inclusion rules. If a span matches any `include_any` block, its accepted (as long as the exclude policy doesn't have a match).

```yaml
 filter_policies:
        - include:
            match_type: regex
            attributes:
              - key: kind
              value: SPAN_KIND_(SERVER|CONSUMER)
         - include_any:
            match_type: stric
            attributes:
              - key: kind
                value: SPAN_KIND_(INTERNAL)
              - key: service.name
                value: foo
        - include_only:
            match_type: strict
            attributes:
              - key: kind
                value: SPAN_KIND_(CLIENT)
              - key: k8s.namespace.name
                value: bar

```

Under this logic, a span is included if it meets any of these criteria:

- kind is SERVER or CONSUMER 
- kind is INTERNAL and service.name is foo 
- kind is CLIENT and k8s.namespace.name is bar 

**Benchmarks**

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/tempo/pkg/spanfilter
cpu: Apple M3 Pro
                                         │ old_results.txt │          new_results.txt           │
                                         │     sec/op      │   sec/op     vs base               │
SpanFilter_applyFilterPolicyNone-11            4.010m ± 4%   3.887m ± 0%  -3.07% (p=0.000 n=10)
SpanFilter_applyFilterPolicySmall-11           50.03m ± 1%   48.00m ± 1%  -4.05% (p=0.000 n=10)
SpanFilter_applyFilterPolicyMedium-11          50.00m ± 1%   48.02m ± 1%  -3.97% (p=0.000 n=10)
SpanFilter_applyFilterPolicyRegex-11           50.12m ± 0%   47.98m ± 0%  -4.25% (p=0.000 n=10)
SpanFilter_applyFilterPolicyIntrinsic-11       21.62m ± 8%   21.77m ± 1%       ~ (p=0.436 n=10)
geomean                                        25.54m        24.79m       -2.95%

```

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/pkg/spanfilter/spanfilter.go
+++ b/pkg/spanfilter/spanfilter.go
@@ -69,7 +69,7 @@ func (f *SpanFilter) ApplyFilterPolicy(rs *v1.Resource, span *tracev1.Span) bool
 		return false
 	}
 
-	if f.hasIncludeAny && f.isIncludedOnly(rs, span) {
+	if f.hasIncludeAny && f.isIncludedAny(rs, span) {
 		return true
 	}
 
@@ -85,7 +85,7 @@ func (f *SpanFilter) ApplyFilterPolicy(rs *v1.Resource, span *tracev1.Span) bool
 
 // This is different than the isIncluded. It's a VIP pass,working as an OR expression.
 // if ANY policy matches the span is included
-func (f *SpanFilter) isIncludedOnly(rs *v1.Resource, span *tracev1.Span) bool {
+func (f *SpanFilter) isIncludedAny(rs *v1.Resource, span *tracev1.Span) bool {
 	for _, policy := range f.includeAny {
 		if policy.Match(rs, span) {
 			return true


docs/sources/tempo/metrics-from-traces/span-metrics/span-metrics-metrics-generator.md
```diff
@@ -156,8 +156,13 @@ func (ds RatioBasedSampler) Description() string {
 ### Filtering
 
 In some cases, you may want to reduce the number of metrics produced by the `spanmetrics` processor.
-You can configure the processor to use an `include` filter to match criteria that must be present in the span in order to be included.
-Following the include filter, you can use an `exclude` filter to reject portions of what was previously included by the filter policy.
+To do so you can configure any of the following processors, in any order or combination:
+
+- `include`: Defines a matching criteria that all spans must meet. If multiple include policies are defined, a span must match all of them to be included (logical AND).
+
+- `includeOnly`: If a span matches any includeOnly policy, it is immediately included, bypassing the stricter `include` requirements (logical OR). This is ideal for capturing specific internal spans without opening the floodgates for all internal telemetry.
```
electron0zero: is it `include_only` or `includeOnly`?, I would prefer calling it `include_only` everywhere.

javiermolinar: it's include_only. Good catch

pkg/spanfilter/spanfilter_test.go
```diff
@@ -461,6 +462,154 @@ func TestSpanMetrics_applyFilterPolicy(t *testing.T) {
...
+			resource: &v1.Resource{
+				Attributes: []*commonv1.KeyValue{
+					{Key: "service.name", Value: &commonv1.AnyValue{Value: &commonv1.AnyValue_StringValue{StringValue: "any-service"}}},
+				},
+			},
+			span: &tracev1.Span{
+				Kind: tracev1.Span_SPAN_KIND_SERVER,
+				Name: "get-user",
+			},
+		},
```
carles-grafana, line 706: can we add a test with multiple include_only?

docs/sources/tempo/metrics-from-traces/span-metrics/span-metrics-metrics-generator.md
```diff
@@ -308,6 +313,36 @@ metrics_generator:
...
+        # Drop any spans from development tiers
+        - exclude:
+            match_type: regex
+            attributes:
+              - key: resource.tier
+                value: dev-.*
+```
+
+In the above, we want to capture metrics for all production spans in the EU, but we also want to explicitly allow INTERNAL spans from the auth-service, which would otherwise be ignored by the include filter.
+
```
electron0zero, line 345: we should also update `docs/sources/tempo/configuration/_index.md` to include the new policy type.

pkg/spanfilter/config/config.go
```diff
@@ -35,8 +36,8 @@ type MatchPolicyAttribute struct {
 }
 
 func ValidateFilterPolicy(policy FilterPolicy) error {
-	if policy.Include == nil && policy.Exclude == nil {
-		return fmt.Errorf("invalid filter policy; policies must have at least an `include` or `exclude`: %v", policy)
+	if policy.Include == nil && policy.IncludeOnly == nil && policy.Exclude == nil {
+		return fmt.Errorf("invalid filter policy; policies must have at least an `include`, `includeOnly` or `exclude`: %v", policy)
```
electron0zero:
```suggestion
		return fmt.Errorf("invalid filter policy; policies must have at least an `include`, `include_only` or `exclude`: %v", policy)
```

pkg/spanfilter/config/config.go
```diff
@@ -45,6 +46,12 @@ func ValidateFilterPolicy(policy FilterPolicy) error {
 		}
 	}
 
+	if policy.IncludeOnly != nil {
+		if err := ValidatePolicyMatch(policy.IncludeOnly); err != nil {
+			return fmt.Errorf("invalid includeOnly policy: %w", err)
```
electron0zero:
```suggestion
			return fmt.Errorf("invalid include_only policy: %w", err)
```

docs/sources/tempo/metrics-from-traces/span-metrics/span-metrics-metrics-generator.md
```diff
@@ -308,6 +313,36 @@ metrics_generator:
...
+    span_metrics:
+      filter_policies:
+        # Only process spans from EU production environments
+        - include:
+            match_type: regex
+            attributes:
+              - key: resource.location
+                value: eu-.*
+        # Exception Rule: Allow INTERNAL spans for auth-service specifically
+        - include_only:
```
electron0zero: we also need to update the schema in the `Supported fields` section on the `docs/sources/tempo/operations/manage-advanced-systems/user-configurable-overrides.md`

docs/sources/tempo/metrics-from-traces/span-metrics/span-metrics-metrics-generator.md
```diff
@@ -226,8 +226,13 @@ func (ds RatioBasedSampler) Description() string {
...
 In some cases, you may want to reduce the number of metrics produced by the `spanmetrics` processor.
-You can configure the processor to use an `include` filter to match criteria that must be present in the span in order to be included.
-Following the include filter, you can use an `exclude` filter to reject portions of what was previously included by the filter policy.
+To do so you can configure any of the following processors, in any order or combination:
+
+- `include`: Defines a matching criteria that all spans must meet. If multiple include policies are defined, a span must match all of them to be included (logical AND).
+
+- `include_only`: If a span matches any include_only policy, it is immediately included, bypassing the stricter `include` requirements (logical OR). This is ideal for capturing specific internal spans without opening the floodgates for all internal telemetry.
+
+- `exclude`: If a span matches any exclude policy, it is rejected, even if it matched an inclusion rule.
```
electron0zero, line 236: I used Claude to better understand the flow and decision making of these polices, might be useful to include something like this in docs to make it easy for users to understand how these polices work.


```text
Span Filter Policy Evaluation
                           ════════════════════════════

                                ┌─────────────┐
                                │ Incoming Span│
                                └──────┬───────┘
                                       │
                           ┌───────────▼───────────┐
                           │  Any filter policies   │── NO ──→ ACCEPT
                           │      configured?       │         (all spans pass)
                           └───────────┬────────────┘
                                       │ YES
                           ┌───────────▼────────────┐
                           │  Matches any `exclude`  │── YES ─→ REJECT
                           │       policy? (OR)      │
                           └───────────┬─────────────┘
                                       │ NO
                           ┌───────────▼─────────────┐
                           │ Matches any `include_only`│── YES ─→ ACCEPT
                           │       policy? (OR)       │
                           └───────────┬──────────────┘
                                       │ NO
                           ┌───────────▼──────────────┐
                           │  `include` policies       │── YES ─→ All match? ─┐
                           │      configured?          │         │ YES  │ NO  │
                           └───────────┬───────────────┘       ACCEPT  REJECT
                                       │ NO
                           ┌───────────▼───────────────┐
                           │ `include_only` policies   │── YES ─→ REJECT
                           │      configured?          │     (didn't match above)
                           └───────────┬───────────────┘
                                       │ NO
                                       ▼
                                    ACCEPT
                                (no inclusion rules)

    ┌──────────────────────────────────────────────────────────────────┐
    │ Policy Semantics:                                                │
    │                                                                  │
    │  exclude      → OR : any match rejects (highest priority)        │
    │  include_only → OR : any match accepts (bypasses include rules)  │
    │  include      → AND: all must match for acceptance               │
    └──────────────────────────────────────────────────────────────────┘

```

pkg/spanfilter/spanfilter_test.go
```diff
@@ -273,12 +273,13 @@ func TestSpanMetrics_applyFilterPolicy(t *testing.T) {
 			filterPolicies: nil,
 		},
 		{
-			name:   "non nil policy with nil include/exclude fails",
-			err:    fmt.Errorf("invalid filter policy; policies must have at least an `include` or `exclude`: {<nil> <nil>}"),
+			name:   "non nil policy with nil include/includeOnly/exclude fails",
```
electron0zero:
```suggestion
			name:   "non nil policy with nil include/include_only/exclude fails",
```

pkg/spanfilter/spanfilter_test.go
```diff
@@ -273,12 +273,13 @@ func TestSpanMetrics_applyFilterPolicy(t *testing.T) {
...
+			err:    fmt.Errorf("invalid filter policy; policies must have at least an `include`, `includeOnly` or `exclude`: {<nil> <nil> <nil>}"),
```
electron0zero:
```suggestion
			err:    fmt.Errorf("invalid filter policy; policies must have at least an `include`, `include_only` or `exclude`: {<nil> <nil> <nil>}"),
```

pkg/spanfilter/spanfilter_test.go
```diff
@@ -461,6 +462,154 @@ func TestSpanMetrics_applyFilterPolicy(t *testing.T) {
 				Name: "test",
 			},
 		},
+
+		{
+			name:   "a matching includeOnly policy for internal span kind for allowed service",
```
electron0zero:
```suggestion
			name:   "a matching include_only policy for internal span kind for allowed service",
```

pkg/spanfilter/spanfilter_test.go
```diff
@@ -461,6 +462,154 @@ func TestSpanMetrics_applyFilterPolicy(t *testing.T) {
...
+					IncludeOnly: &config.PolicyMatch{
+						MatchType: config.Strict,
+						Attributes: []config.MatchPolicyAttribute{
+							{Key: "kind", Value: "SPAN_KIND_INTERNAL"},
+							{Key: "resource.service.name", Value: "auth-service"},
+						},
+					},
+				},
+			},
+			resource: &v1.Resource{
```
electron0zero, line 697: can we get few more cases that test following things:
- include_only with an exclude policy
- include_only with regex match_type
- a test with exclude, include_only and include polices

electron0zero:
> I'm wondering if the include_only name is clear enough. Maybe include_any is better?

+1, I think Include_any makes it more clear then include_only.

docs/sources/tempo/metrics-from-traces/span-metrics/span-metrics-metrics-generator.md
javiermolinar: I think is useful but too detailed. I think the new name is descriptive enough

docs/sources/tempo/metrics-from-traces/span-metrics/span-metrics-metrics-generator.md
javiermolinar: good catch

docs/sources/tempo/metrics-from-traces/span-metrics/span-metrics-metrics-generator.md
javiermolinar: Good catch!

pkg/spanfilter/spanfilter.go
```diff
@@ -30,45 +30,88 @@ func NewSpanFilter(filterPolicies []config.FilterPolicy) (*SpanFilter, error) {
...
+
+	// If we have an include_any but NO standard include, and we reached
+	// here, it means include_any didn't match. -> return false.
+	// IF NO inclusion rules exist at all -> return true.
+	return !f.hasIncludeAny
+}
+
+// This is different than the isIncluded. It's a VIP pass,working as an OR expression.
+// if ANY policy matches the span is included
+func (f *SpanFilter) isIncludedOnly(rs *v1.Resource, span *tracev1.Span) bool {
```
carles-grafana: still called "only"

PR #6400, opened 2026-02-06
enhancement: improve attribute truncating observability

javiermolinar:
**What this PR does**:
The current `distributor_attributes_truncated_total` counter is too coarse. With this change we can see which scopes are the most truncated. 

**Checklist**
- [x] Tests updated
- [ ] Documentation added


carles-grafana (commented):
should we update the recording rule as well?

modules/distributor/distributor.go
```diff
@@ -114,8 +114,8 @@ var (
 	metricAttributesTruncated = promauto.NewCounterVec(prometheus.CounterOpts{
 		Namespace: "tempo",
 		Name:      "distributor_attributes_truncated_total",
-		Help:      "The total number of attribute keys or values truncated per tenant",
```
electron0zero, line 117: nit: does it make sense to update the "Set max attribute size to help control out of memory errors" section in `docs/sources/tempo/configuration/_index.md` to mention that we haver per scope info?

feels like not, but I your call.

javiermolinar: I don't think so either, that could be a rabbit hole if we do the same for the rest of metrics :P

javiermolinar:
> should we update the recording rule as well?

Not right now. This won't break it

PR #6405, opened 2026-02-06
[bugfix] correct lag calculation

ruslan-mikhailov:
**What this PR does**: fixes lag calculation. This PR contains couple of bugfixes:
1. correct lag calculation as diff between HWM and latest committed offset.
    a. If there was an error on consumption right after the first iteration, we cannot get lag and set it to -1. Note: current code does not return an error, it is a no-op change.
    b. if no error and no recorded offset - means no new records, lag is zero.
2. record lag only after `consume` function call
3. If all records in a batch could not be consumed (e.g. `record.Timestamp.Before(cutoff)`), we still need to commit these changes. Currently, they are committed only if at least one message was pushed to `inst.pushBytes`.
4. Wait for catch up only after all processes started. This fixes a deadlock when live-store is not yet started block processing, while it is already consuming and blocked by `waitForCatchUp`.

**How it has been tested**: unit tests (in the PR) and manual tests.
Manual test case.
1. add long sleep on consumption to be sure live-store is lagging.
2. run it for at least 10 minutes (ideally, 30+ mins)
3. remove long sleep and enable readiness probe.

Expected result: it is running, but in non-ready state. After it catches up, it becomes ready.

Actual result matches expected results: 
image

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 * [ENHANCEMENT] Add new alerts and runbooks entries [#6276](https://github.com/grafana/tempo/pull/6276) (@javiermolinar)
 * [ENHANCEMENT] Double the maximum number of dedicated string columns in vParquet5 and update tempo-cli to determine the optimum number for the data [#6282](https://github.com/grafana/tempo/pull/6282) (@mdisibio)
 * [ENHANCEMENT] Improve attribute truncating observability [#6400](https://github.com/grafana/tempo/pull/6400) (@javiermolinar)
-* [ENHANCEMENT] Improved live store readiness check and added `readiness_target_lag` and `readiness_max_wait` config parameters. Live store will now - if `readiness_target_lag` is set - not report `/ready` until Kafka lag is brought under the specified value [#6238](https://github.com/grafana/tempo/pull/6238) (@oleg-kozlyuk-grafana)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -10,12 +10,12 @@
 * [ENHANCEMENT] Double the maximum number of dedicated string columns in vParquet5 and update tempo-cli to determine the optimum number for the data [#6282](https://github.com/grafana/tempo/pull/6282) (@mdisibio)
 * [ENHANCEMENT] Improve attribute truncating observability [#6400](https://github.com/grafana/tempo/pull/6400) (@javiermolinar)
 * [ENHANCEMENT] Improved live store readiness check and added `readiness_target_lag` and `readiness_max_wait` config parameters. Live store will now - if `readiness_target_lag` is set - not report `/ready` until Kafka lag is brought under the specified value [#6238](https://github.com/grafana/tempo/pull/6238) (@oleg-kozlyuk-grafana)
+* [ENHANCEMENT] Improved live store readiness check and added `readiness_target_lag` and `readiness_max_wait` config parameters. Live store will now - if `readiness_target_lag` is set - not report `/ready` until Kafka lag is brought under the specified value [#6238](https://github.com/grafana/tempo/pull/6238) [#6405](https://github.com/grafana/tempo/pull/6405) (@oleg-kozlyuk-grafana, @ruslan-mikhailov)
```
oleg-kozlyuk-grafana, line 12: I think this can be rolled out into previous entry (one line with 2 PRs and 2 authors)

oleg-kozlyuk-grafana:
Excellent job, thank you for fixing these issues!

CHANGELOG.md
ruslan-mikhailov: oops, I accidentally copied the line. Fixed, thanks!

PR #6414, opened 2026-02-09
metrics-generator: Add per-label limiter to control cardinality

electron0zero:
**What this PR does**:

Adds a per-label limiter that caps the number of distinct values (cardinality) any single label can have. When a label exceeds the configured threshold, only that label's value is replaced with `__cardinality_overflow__` value, and all other labels are preserved. 

This prevents a single high-cardinality label (e.g. `http.url`, `trace.id`) from consuming the entire active series/entity budget and adding an upper bound on the maximum active series generated by a single metrics-generator pod.

This is the first line of limiter before the existing `max_active_series` / `max_active_entities` limiters, which remain as the last-resort safety net.

**Pipeline order:** Drain Sanitizer (if enabled) -> per-label limiter (if enabled) -> entity/series limiter (if enabled)

**How it works:**

- Tracks per-label cardinality using HyperLogLog sketches (configured for ~3.25% standard error), with a status that's updated every 15s
- Always inserts original value hash into HyperLogLog sketches to capture accurate cardinality demand estimate, and prevent estimate oscillation.
- Metadata labels (`__name__`, `__type__`, `__unit__`) are never overflowed.
- If a label's cardinality drops below the threshold, the limiter automatically recovers after the label ages out.
- New per-tenant runtime override: `max_cardinality_per_label` (0 means disabled, default = 0).

**New metrics:**
- `tempo_metrics_generator_registry_label_values_limited_total`, (counter, labels: `tenant`, `label_name`) - to capture the label_values that were limited by the counter.
- `tempo_metrics_generator_registry_label_cardinality_demand_estimate`, (gauge, labels: `tenant`, `label_name`) - to capture the cardinality estimate with HLL.

**Benchmark**

Benchmarked PerLabelLimiter.Limit() and it looks pretty good, here is tl;dr:
- disabled, ~1.6 ns/op, 0 B/op, 0 allocs/op
- under_limit, ~98 ns/op, 0 B/op, 0 allocs/op
- over_limit, ~278 ns/op, 400 B/op, 4 allocs/op
- many_labels_under_limit, ~185 ns/op, 0 B/op, 0 allocs/op
- many_labels_over_limit, ~760 ns/op, 944 B/op, 5 allocs/op
- parallel, ~260 to 290 ns/op, 0 B/op, 0 allocs/op

<details>
<summary>BenchmarkPerLabelLimiter_Limit</summary>

```shell
goos: darwin
goarch: arm64
pkg: github.com/grafana/tempo/modules/generator/registry
cpu: Apple M3 Pro
BenchmarkPerLabelLimiter_Limit/disabled-11         	732658879	         1.575 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/disabled-11         	729314253	         1.625 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/disabled-11         	764622409	         1.611 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/disabled-11         	756254059	         1.583 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/disabled-11         	709476858	         1.609 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/disabled-11         	714076378	         1.631 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/disabled-11         	741946402	         1.611 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/disabled-11         	717511491	         1.617 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/disabled-11         	709165894	         1.596 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/disabled-11         	748806589	         1.623 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/under_limit-11      	11938209	        97.76 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/under_limit-11      	12223900	        98.09 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/under_limit-11      	12321429	        97.87 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/under_limit-11      	11994697	        96.94 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/under_limit-11      	12165490	        97.68 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/under_limit-11      	12081058	        97.90 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/under_limit-11      	12173836	        98.40 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/under_limit-11      	12062518	        97.98 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/under_limit-11      	12162567	        97.92 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/under_limit-11      	11818942	        98.63 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/over_limit-11       	 4324588	       277.6 ns/op	     400 B/op	       4 allocs/op
BenchmarkPerLabelLimiter_Limit/over_limit-11       	 4319218	       276.8 ns/op	     400 B/op	       4 allocs/op
BenchmarkPerLabelLimiter_Limit/over_limit-11       	 4325790	       277.5 ns/op	     400 B/op	       4 allocs/op
BenchmarkPerLabelLimiter_Limit/over_limit-11       	 4293369	       277.9 ns/op	     400 B/op	       4 allocs/op
BenchmarkPerLabelLimiter_Limit/over_limit-11       	 4325608	       278.9 ns/op	     400 B/op	       4 allocs/op
BenchmarkPerLabelLimiter_Limit/over_limit-11       	 4256241	       285.7 ns/op	     400 B/op	       4 allocs/op
BenchmarkPerLabelLimiter_Limit/over_limit-11       	 4311650	       295.6 ns/op	     400 B/op	       4 allocs/op
BenchmarkPerLabelLimiter_Limit/over_limit-11       	 4250446	       282.6 ns/op	     400 B/op	       4 allocs/op
BenchmarkPerLabelLimiter_Limit/over_limit-11       	 4326865	       278.2 ns/op	     400 B/op	       4 allocs/op
BenchmarkPerLabelLimiter_Limit/over_limit-11       	 4279266	       279.3 ns/op	     400 B/op	       4 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_under_limit-11         	 6445549	       184.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_under_limit-11         	 6558026	       184.2 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_under_limit-11         	 6314889	       184.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_under_limit-11         	 6531934	       184.2 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_under_limit-11         	 6401626	       185.5 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_under_limit-11         	 6454704	       184.8 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_under_limit-11         	 6492103	       185.6 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_under_limit-11         	 6460478	       184.3 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_under_limit-11         	 6454285	       184.8 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_under_limit-11         	 6370269	       184.6 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_over_limit-11          	 1594917	       758.9 ns/op	     944 B/op	       5 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_over_limit-11          	 1586004	       758.9 ns/op	     944 B/op	       5 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_over_limit-11          	 1589581	       761.4 ns/op	     944 B/op	       5 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_over_limit-11          	 1583690	       761.8 ns/op	     944 B/op	       5 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_over_limit-11          	 1579005	       764.2 ns/op	     944 B/op	       5 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_over_limit-11          	 1574115	       764.0 ns/op	     944 B/op	       5 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_over_limit-11          	 1570197	       767.0 ns/op	     944 B/op	       5 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_over_limit-11          	 1574158	       766.2 ns/op	     944 B/op	       5 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_over_limit-11          	 1569938	       764.9 ns/op	     944 B/op	       5 allocs/op
BenchmarkPerLabelLimiter_Limit/many_labels_over_limit-11          	 1575445	       763.7 ns/op	     944 B/op	       5 allocs/op
BenchmarkPerLabelLimiter_Limit/parallel-11                        	 4228664	       261.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/parallel-11                        	 4147978	       285.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/parallel-11                        	 3932316	       277.7 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/parallel-11                        	 4242210	       261.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/parallel-11                        	 4348903	       282.0 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/parallel-11                        	 4199176	       278.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/parallel-11                        	 4622820	       264.6 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/parallel-11                        	 4198548	       285.2 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/parallel-11                        	 4451056	       260.6 ns/op	       0 B/op	       0 allocs/op
BenchmarkPerLabelLimiter_Limit/parallel-11                        	 4498274	       288.5 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/grafana/tempo/modules/generator/registry	90.105s
```

</details>

Fixes #

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/modules/generator/registry/per_label_limiter.go
+++ b/modules/generator/registry/per_label_limiter.go
@@ -79,12 +79,15 @@ func NewPerLabelLimiter(tenant string, maxCardinalityF maxCardinalityFunc, stale
 // Labels whose estimated cardinality exceeds the configured max have their
 // value replaced with __cardinality_overflow__.
 func (s *PerLabelLimiter) Limit(lbls labels.Labels) labels.Labels {
+	// do maintenance check as the first thing to ensure maxCardinality
+	// is refreshed from runtime overrides. without this,
+	// a limiter that starts disabled would never be enabled without restart.
+	s.doPeriodicMaintenance()
+
 	// maxCardinality is zero, so limiter is disabled, return labels as is
 	if s.maxCardinality.Load() == 0 {
 		return lbls
 	}
-
-	s.doPeriodicMaintenance()
 
 	s.mtx.Lock()
 	defer s.mtx.Unlock()
@@ -148,10 +151,18 @@ func (s *PerLabelLimiter) getOrCreateState(labelName string) *labelCardinalitySt
 func (s *PerLabelLimiter) doPeriodicMaintenance() {
 	select {
 	case <-s.demandUpdateChan:
+		// step 1: refresh maxCardinality config from override
 		// fetch once per tick and cache atomically, the limit is the same for all labels in a tenant
 		maxCardinality := s.maxCardinalityFunc(s.tenant)
 		s.maxCardinality.Store(maxCardinality)
 
+		// if the check is disabled, skip the demand update and, exit early.
+		// no data is being inserted into the sketch, so nothing to estimate or publish
+		if maxCardinality == 0 {
+			return
+		}
+
+		// step 2: update estimate and publish demand estimate metric
 		s.mtx.Lock()
 		for labelName, state := range s.labelsState {
 			estimate := state.sketch.Estimate()
--- a/modules/generator/registry/per_label_limiter_test.go
+++ b/modules/generator/registry/per_label_limiter_test.go
@@ -25,6 +25,71 @@ func TestPerLabelLimiter_Disabled(t *testing.T) {
 	require.Equal(t, lbls, result)
 }
 
+// TestPerLabelLimiter_RuntimeEnableDisable verifies that toggling max_cardinality_per_label at runtime
+// via overrides takes effect without restarting the generator.
+func TestPerLabelLimiter_RuntimeEnableDisable(t *testing.T) {
+	tenant := "test-runtime-toggle"
+	var maxCardinality atomic.Uint64
+	maxCardinality.Store(0) // start disabled
+
+	s := NewPerLabelLimiter(tenant, func(string) uint64 {
+		return maxCardinality.Load()
+	}, 15*time.Minute)
+
+	// Phase 1: Disabled - all labels pass through
+	for i := 0; i < 10; i++ {
+		lbls := labels.FromStrings("__name__", "m", "url", fmt.Sprintf("/path/%d", i))
+		result := s.Limit(lbls)
+		require.Equal(t, fmt.Sprintf("/path/%d", i), result.Get("url"), "should pass through when disabled")
+	}
+
+	// when disabled, Limit() returns before inserting into sketches, so labelsState
+	// is empty, which means no demand gauge was published as well.
+	triggerDemandUpdate(s)
+	s.mtx.Lock()
+	labelsStateLen := len(s.labelsState)
+	s.mtx.Unlock()
+	require.Equal(t, 0, labelsStateLen, "no label state should exist when disabled")
+
+	// Phase 2: Enable at runtime by changing the override
+	maxCardinality.Store(5)
+	triggerDemandUpdate(s)
+
+	// Push more distinct values - should overflow now
+	for i := 0; i < 10; i++ {
+		lbls := labels.FromStrings("__name__", "m", "url", fmt.Sprintf("/path/%d", i))
+		s.Limit(lbls)
+	}
+	triggerDemandUpdate(s)
+
+	// demand gauge should be published when enabled
+	var g io_prometheus_client.Metric
+	require.NoError(t, metricLabelCardinalityDemand.WithLabelValues(tenant, "url").Write(&g))
+	require.Greater(t, g.GetGauge().GetValue(), float64(5), "demand gauge should be published when enabled")
+
+	lbls := labels.FromStrings("__name__", "m", "url", "/path/999")
+	result := s.Limit(lbls)
+	require.Equal(t, overflowValue, result.Get("url"), "should overflow after runtime enable")
+
+	// capture the demand gauge value before disabling
+	var before io_prometheus_client.Metric
+	require.NoError(t, metricLabelCardinalityDemand.WithLabelValues(tenant, "url").Write(&before))
+	demandBefore := before.GetGauge().GetValue()
+
+	// Phase 3: Disable again at runtime
+	maxCardinality.Store(0)
+	triggerDemandUpdate(s)
+
+	lbls = labels.FromStrings("__name__", "m", "url", "/path/new/10000")
+	result = s.Limit(lbls)
+	require.Equal(t, "/path/new/10000", result.Get("url"), "should pass through after runtime disable")
+
+	// demand gauge should be stale (not updated) after disabling
+	var after io_prometheus_client.Metric
+	require.NoError(t, metricLabelCardinalityDemand.WithLabelValues(tenant, "url").Write(&after))
+	require.Equal(t, demandBefore, after.GetGauge().GetValue(), "demand gauge should not be updated when disabled")
+}
+
 func TestPerLabelLimiter_UnderLimit(t *testing.T) {
 	s := NewPerLabelLimiter("test", testMaxCardinality(100), 15*time.Minute)
 


modules/generator/registry/cardinality_sanitizer.go
```diff
@@ -0,0 +1,126 @@
...
+
+var metricCardinalityLimitOverflows = promauto.NewCounterVec(prometheus.CounterOpts{
+	Namespace: "tempo",
+	Name:      "metrics_generator_registry_cardinality_limit_overflows_total",
+	Help:      "Total number of label values replaced with __cardinality_overflow__ by the per-label cardinality sanitizer",
+}, []string{"tenant"})
+
+type labelCardinalityState struct {
+	sketch    *Cardinality
+	overLimit bool // cached flag, updated periodically in maintenance tick
```
oleg-kozlyuk-grafana: The 15 second interval makes it possible to exceed the limit (in addition to HLL being imprecise at base). Is it possible to calculate the cardinality and break on the hot path, or would that be too high of performance impact?

modules/generator/registry/cardinality_sanitizer.go
```diff
@@ -0,0 +1,126 @@
...
+			sketch: NewCardinality(s.staleDuration, removeStaleSeriesInterval),
+		}
+		s.labelsState[labelName] = state
+	}
+	return state
+}
+
+// doPeriodicMaintenance holds s.mtx while iterating labelsState for both
+// demand updates (every 15s) and pruning (every 5m). This blocks Sanitize()
+// callers for the duration. In practice this is fast, so it's acceptable.
```
oleg-kozlyuk-grafana: Can the lock be per-sketch rather than global?

docs/sources/tempo/troubleshooting/metrics-generator.md
```diff
@@ -121,6 +121,39 @@ overrides:
...
+For example, a high-cardinality `url` label is overflowed while `method` is kept:
+
+```
+{service="foo", method="GET", url="/users/123"} → {service="foo", method="GET", url="__cardinality_overflow__"}
+```
+
+To detect if per-label cardinality limiting is active:
+
+```promql
+sum(rate(tempo_metrics_generator_registry_cardinality_limit_overflows_total{}[1m]))
```
oleg-kozlyuk-grafana: nit: ideally this should feed into existing metric with different label

modules/generator/registry/cardinality_sanitizer.go
```diff
@@ -0,0 +1,126 @@
...
+type labelCardinalityState struct {
+	sketch    *Cardinality
+	overLimit bool // cached flag, updated periodically in maintenance tick
+}
+
+// CardinalitySanitizer replaces individual label values with __cardinality_overflow__
+// when the estimated cardinality of that label exceeds maxCardinality.
+// Unlike the global series limiter which collapses all labels into a single overflow
+// entity, this preserves all other labels and only overflows the problematic one.
+type CardinalitySanitizer struct {
```
oleg-kozlyuk-grafana: I think this sanitizer will have the most value if it is not "on timer", i.e. with delayed reaction - because in 15 seconds, damage would already be done. Ideally, this should be "synchronous", I.e. not require time to pass to react

modules/generator/registry/cardinality_sanitizer.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/prometheus/prometheus/schema"
+)
+
+const overflowValue = "__cardinality_overflow__"
+
+var metricCardinalityLimitOverflows = promauto.NewCounterVec(prometheus.CounterOpts{
+	Namespace: "tempo",
+	Name:      "metrics_generator_registry_cardinality_limit_overflows_total",
+	Help:      "Total number of label values replaced with __cardinality_overflow__ by the per-label cardinality sanitizer",
+}, []string{"tenant"})
```
electron0zero: we can also add `label_name` here so we know which label is overflowing, and we can add another estimate metric where we publish per label Cardinality estimate from the sketch we are building for each label.

modules/generator/registry/cardinality_sanitizer.go
electron0zero: check for cardinality in hot path would have perf. impact because we would need to call `Estimate()` on the HLL on each call to `Sanitize()`.

we can reduce the interval if we want to refresh this flag faster.

modules/generator/registry/cardinality_sanitizer.go
electron0zero: we are okay going little over the limit for a short window, as long as we catchup. we can play around with the 15 seconds timer to see if we can be faster.

docs/sources/tempo/troubleshooting/metrics-generator.md
electron0zero: since this is a different  limiter from existing limiters (max AS / max Entity) so imo, it makes sense to have it's own counter.

renamed it to `tempo_metrics_generator_registry_label_values_limited_total` and added `label_name` as well so we know which labels are limited for which tenant

modules/generator/registry/builder.go
```diff
@@ -63,16 +65,20 @@ func (b *labelBuilder) Add(name, value string) {
...
 	// it's no longer safe to use the builder after this point, so we drop our
 	// reference to it. this may cause a nil panic if the builder is used after
 	// this point, but it's better than memory corruption.
 	builderPool.Put(b.builder)
 	b.builder = nil
 
-	if !labels.IsValid(model.UTF8Validation) {
-		return labels, false
+	if !lbls.IsValid(model.UTF8Validation) {
+		return lbls, false
```
oleg-kozlyuk-grafana, line 80: What's the purpose of renaming the variable?

modules/generator/registry/per_label_limiter.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/cespare/xxhash/v2"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/schema"
+)
+
+const (
+	overflowValue        = "__cardinality_overflow__"
+	demandUpdateInterval = 15 * time.Second
```
oleg-kozlyuk-grafana, line 20: I would use a significantly lower value here, 1-5 seconds, so it's reasonably reactive, but up to you

modules/generator/registry/per_label_limiter.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+var metricLabelValuesLimited = promauto.NewCounterVec(prometheus.CounterOpts{
+	Namespace: "tempo",
+	Name:      "metrics_generator_registry_label_values_limited_total",
+	Help:      "Total number of times a label value was limited due to exceeding the per-label cardinality limit",
+}, []string{"tenant", "label_name"})
```
oleg-kozlyuk-grafana, line 27: Do we have other metrics that have `label_name` as dimension? Would be ironic to cause a cardinality issue by adding a cardinality limiter

modules/generator/registry/per_label_limiter.go
```diff
@@ -0,0 +1,169 @@
...
+// doPeriodicMaintenance holds s.mtx while iterating labelsState for both
+// demand updates (every 15s) and pruning (every 5m). This blocks Limit()
+// callers for the duration. In practice this is fast, so it's acceptable.
+// If it becomes a problem, snapshot the labelsState slice under the lock
+// and do sketch operations outside it.
+func (s *PerLabelLimiter) doPeriodicMaintenance() {
+	select {
+	case <-s.demandUpdateChan:
+		// fetch once per tick and cache atomically, the limit is the same for all labels in a tenant
+		maxCardinality := s.maxCardinalityFunc(s.tenant)
```
oleg-kozlyuk-grafana, line 156: Maybe worth updating when pruned, not on every tick?

oleg-kozlyuk-grafana:
@codex review this as well

modules/generator/registry/builder.go
electron0zero: renamed because the variable name `labels` was [shadowing](https://www.jetbrains.com/guide/go/tips/variable-shadowing-display/) the package `labels` (imported as `github.com/prometheus/prometheus/model/labels`)

modules/generator/registry/per_label_limiter.go
electron0zero: I picked 15 because we have `collection_interval: 15s` so it would refresh the demand atleast once per collection_interval.

faster demand doesn't hurt and it will help us stop early, I will update to 5s

modules/generator/registry/per_label_limiter.go
electron0zero: I looked at it, and the worst case cardinality here would be "numbers of tenants in a clusters" * "avg number of labels rate limited".

for example: in a 1000 tenant cluster with 10 labels, it would multiply the cardinality by `10,000` to the baseline active series  (excluding the things like pod, instance, job labels that exists).

I am inclined to leave `lable_name` in the code and then drop other labels like pod, instance, etc in the collection pipeline if this metric becomes an cardinality issue.

I am more worried about the cardinality of the `tempo_metrics_generator_registry_label_cardinality_demand_estimate` metric then this because this counter will only be active for the labels that are being limited.

modules/generator/registry/per_label_limiter.go
```diff
@@ -0,0 +1,169 @@
...
+
+// Limit applies the per-label cardinality limit to the given labels.
+// Labels whose estimated cardinality exceeds the configured max have their
+// value replaced with __cardinality_overflow__.
+func (s *PerLabelLimiter) Limit(lbls labels.Labels) labels.Labels {
+	// maxCardinality is zero, so limiter is disabled, return labels as is
+	if s.maxCardinality.Load() == 0 {
+		return lbls
+	}
+
```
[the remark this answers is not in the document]
electron0zero: good catch.

modules/generator/registry/cardinality_sanitizer.go
electron0zero: decided to keep it simple for now, we can come back and tune the locking if we see issues in production.

modules/generator/registry/per_label_limiter.go
electron0zero: Reverted back to 15s. 

The demand update holds the same mutex as Limit(), so running it more frequently increases risk of lock contention on the hot path.

modules/generator/registry/per_label_limiter.go
electron0zero: prune tick is every 15m so config updates will take upto 15 mins to be refreshed.

since this method call `maxCardinalityFunc` is just looking up the value in the overrides map (in memory map), it's pretty fast so keeping it as is for now.

modules/generator/registry/per_label_limiter.go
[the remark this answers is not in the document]
electron0zero: @codex review again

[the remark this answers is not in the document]
electron0zero: @codex review again

PR #6420, opened 2026-02-10
chore: Fix flaky backend scheduler test

zeitlinger:
- Fix flaky TestBackendScheduler test by using the correct loop variable tenantID instead of tenant when tracking per-tenant expectations in the accumulant map                                                                        
  - Move accumulation logic inside the per-tenant loop so each tenant's expected blocks and outstanding counts are tracked independently                                                                                                 
                                                                                                                                                                                                                                         
**Checklist**
- [ ] Tests updated
- [ ] Documentation added


oleg-kozlyuk-grafana:
Hi Gregor,

Can you rebase this PR please so the policy check (hopefully) passes?

PR #6422, opened 2026-02-10
live-store: allow zone removal for active partitions

zalegrala:
**What this PR does**:

Currently, removal of a zone is not possible without scaling the partitions to 0.  Here we modify the prepare-downscale logic to alloy an owner to leave without scaling down the partitions at the same time.  This allows a new (temporary) zone to be created to maintain redundancy for a period for maintenance while also allowing its later removal.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/livestore/downscale.go
+++ b/modules/livestore/downscale.go
@@ -180,8 +180,8 @@ func (s *LiveStore) PrepareDownscaleHandler(w http.ResponseWriter, r *http.Reque
 			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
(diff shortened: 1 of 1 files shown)


mattdurham:
What do you mean by temporary zone?

zalegrala:
In our case, we want to replace the existing zones for a storageclass change, which is an immutable field, so we're thinking create a new zone while we perform maintenance on the others to keep our reliability.  I can also imagine scenarios where folks deployed in one way might want to move a zone to a different fault-domain.

PR #6425, opened 2026-02-11
[bugfix] Fix flaky TestLogClusterCache_MaxSizeEviction

Logiraptor:
Fixes flaky test caused by incorrect assumption about otter's eviction policy.

The test assumed LRU behavior where recently added items survive. Otter uses S3-FIFO, which puts new items in a probationary segment—they can be evicted immediately if not accessed again.

Removed the non-deterministic assertion and corrected comments.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pkg/drain/log_cluster_cache_test.go
+++ b/pkg/drain/log_cluster_cache_test.go
@@ -120,21 +120,19 @@ func TestLogClusterCache_MaxSizeEviction(t *testing.T) {
 	// but we can verify the cache size is bounded reasonably
 	clusterCount := 0
 	for i := 1; i <= maxSize*3; i++ {
-		if cache.Get(i) != nil {
+		if cache.GetQuietly(i) != nil {
 			clusterCount++
 		}
 	}
-	// Otter cache eviction behavior is complex and may not evict immediately.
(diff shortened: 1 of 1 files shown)


pkg/drain/log_cluster_cache_test.go
```diff
@@ -120,21 +120,19 @@ func TestLogClusterCache_MaxSizeEviction(t *testing.T) {
...
-		if cache.Get(i) != nil {
+		if cache.GetQuietly(i) != nil {
 			clusterCount++
 		}
 	}
-	// Otter cache eviction behavior is complex and may not evict immediately.
+	// Otter cache uses S3-FIFO eviction which is frequency-based, not strictly LRU.
 	// The important thing is that the cache doesn't grow unbounded and handles Put operations correctly.
 	// Verify cache is bounded (not all items are present, or at least cache operations work)
 	assert.LessOrEqual(t, clusterCount, maxSize*3, "cache should not exceed reasonable bounds")
```
oleg-kozlyuk-grafana, line 130: Can we make any assumption about otter's cache? At this point, we are essentially testing that if we add 3N elements, the cache will contain 3N or less elements (which is obviously true). Not sure if this test is worth leaving in.

PR #6435, opened 2026-02-11
fix: reload span_name_sanitization overrides during runtime

electron0zero:
**What this PR does**:

Before this fix, we were just loading the MetricsGeneratorSpanNameSanitization overrides just at startup, and never checking or reloading after that.

If the overrides were updated (which they can without restart), we won't reload them and the MetricsGeneratorSpanNameSanitization config change in the overrides won't take effact without restart

we missed this detail and the reload in the https://github.com/grafana/tempo/pull/6098, so fixing it now in a follow up PR

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/registry/drain_sanitizer_test.go
+++ b/modules/generator/registry/drain_sanitizer_test.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/prometheus/prometheus/model/labels"
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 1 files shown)


modules/generator/registry/registry.go
```diff
@@ -127,10 +119,7 @@ func New(cfg *Config, overrides Overrides, tenant string, appendable storage.App
 		externalLabels[cfg.InjectTenantIDAs] = tenant
 	}
 
-	var sanitizer Sanitizer = &noopSanitizer{}
-	if sanitizationMode := overrides.MetricsGeneratorSpanNameSanitization(tenant); sanitizationMode != SpanNameSanitizationDisabled {
-		sanitizer = NewDrainSanitizer(tenant, sanitizationMode == SpanNameSanitizationDryRun, cfg.StaleDuration)
-	}
+	sanitizer := NewDrainSanitizer(tenant, overrides.MetricsGeneratorSpanNameSanitization, cfg.StaleDuration)
```
electron0zero, line 122: this is following the same pattern we follow in the `NewTracker` in the `modules/distributor/usage/tracker.go`

another way we can do it pass overrides down and call MetricsGeneratorSpanNameSanitization instead of taking func as argument.

ie-pham (approved):

PR #6440, opened 2026-02-12
Lenient parser for tag values search

### LINKED ISSUE #6144: Improve tag values filtered search brittleness
Autocomplete or filtered tag values search in Tempo is the capability to return values for attributes filtered down by an existent query. 

Due to the performance requirements (ie. only a subset of queries are supported) and the fact that filtering queries are incomplete and fail to parse (eg. `{.foo = "foo" && .bar = }`), this feature has depended on two regexes since it was [first introduced](https://github.com/grafana/tempo/pull/2253).

These two regexes validate that the query is supported and extracts the valid matchers (eg. `.foo = "foo"` from the query above) for filtering. But they haven't come without their fair share of problems—since introduced, a myriad of PRs have been merge to fix or extend the regexes: 
* Extend: https://github.com/grafana/tempo/pull/3846, https://github.com/grafana/tempo/pull/3458
* Fix: https://github.com/grafana/tempo/pull/3865, https://github.com/grafana/tempo/pull/3641, https://github.com/grafana/tempo/pull/3426, https://github.com/grafana/tempo/pull/3277, https://github.com/grafana/tempo/pull/3339

And last but not least, another fix for queries with whitespaces in the attribute values: https://github.com/grafana/tempo/pull/6124. There is a real need to design and implement a long lasting solution that is not so error prone.

Some ideas we could explore:

### Lenient parser

A custom [lexer](https://github.com/grafana/tempo/blob/main/pkg/traceql/lexer.go#L118) that uses substitutes parts of our current [yacc parser](https://github.com/grafana/tempo/blob/main/pkg/traceql/parse.go#L34) while still using the traceql semantics and variables and allows it to recover from parsing errors while still extracting the valid tokens.

This is the approach I've been exploring and have a working prototype for.

### Support incomplete matchers

Add support for incomplete matchers `{.bar=}` to our yacc parser so parsing such queries don't fail. Later, such matchers are removed from the extracted expression.

### Other

This is in exploratory phase at this point. There might be other approaches that are better. Proposals are welcome :)

mapno:
**What this PR does**:

### Problem

Autocomplete (filtered tag values search) has relied on two regexes since it was first introduced — [one to validate](https://github.com/grafana/tempo/blob/533c17ccb170224b70c0be33152e5f2ffa422e5a/pkg/traceql/extractmatcher.go#L37) that the query is supported, [another to extract](https://github.com/grafana/tempo/blob/533c17ccb170224b70c0be33152e5f2ffa422e5a/pkg/traceql/extractmatcher.go#L21) valid matchers from incomplete queries like `{.foo = "foo" && .bar = }`. These regexes have been a persistent source of bugs and maintenance burden.

### Solution

This PR replaces the regexes with a _lenient_ parser that reuses the existing TraceQL lexer and yacc parser. It works in two passes:

  1. Strict parse first: if the query is already valid, return as-is (zero overhead for complete queries)
  2. Token cleanup: `removeIncompleteMatchers` tokenizes the query using the TraceQL lexer, identifies attribute op patterns missing a value (e.g. .foo =), removes the dangling operator, and cleans up orphaned connectors (&&, ||) and empty parentheses. Only the portion before the first | is cleaned — pipeline syntax (rate(), count(), by()) is preserved and re-appended after cleanup.
  3. Re-parse the cleaned string with the strict yacc parser
  4. AST post-processing — `removeBareAttributes` replaces leftover bare attributes (from the cleanup) with true, keeping the AST structurally valid while preserving the original query structure (ORs, ANDs, structural operators)

This handles the full TraceQL grammar — scoped attributes, structural operators (>>, ~), pipelines, nested parentheses — without maintaining fragile regex patterns. The `AllConditions` flag on the returned `FetchSpansRequest` tells callers whether the extracted conditions can be used for filtered search (all AND) or require unfiltered search (contains OR or structural operators).

Fixes #6144

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/lenient_extract.go
+++ b/pkg/traceql/lenient_extract.go
@@ -46,11 +46,11 @@ func ExtractFetchRequest(query string) *FetchSpansRequest {
 	return req
 }
 
-// CanonicalQuery parses a query string using the lenient parser and returns
+// NormalizeQuery parses a query string using the lenient parser and returns
 // a normalized string representation. Used for cache key generation.
 // Match-all queries (e.g. "{}", "{ }", "{ true }") are normalized to "{}"
 // to avoid cache fragmentation.
-func CanonicalQuery(query string) string {
+func NormalizeQuery(query string) string {
 	query = strings.TrimSpace(query)
 	if len(query) == 0 {
 		return emptyQuery
--- a/modules/frontend/tag_sharder.go
+++ b/modules/frontend/tag_sharder.go
@@ -87,7 +87,7 @@ func (r *tagValueSearchRequest) end() uint32 {
 
 func (r *tagValueSearchRequest) hash() uint64 {
 	hash := fnv1a.HashString64(r.request.TagName)
-	hash = fnv1a.AddString64(hash, traceql.CanonicalQuery(r.request.Query))
+	hash = fnv1a.AddString64(hash, traceql.NormalizeQuery(r.request.Query))
 
 	return hash
 }
--- a/modules/ingester/instance_search.go
+++ b/modules/ingester/instance_search.go
@@ -616,7 +616,7 @@ func includeBlock(b *backend.BlockMeta, req *tempopb.SearchRequest) bool {
 func searchTagValuesV2CacheKey(req *tempopb.SearchTagValuesRequest, limit int, prefix string) string {
 	var cacheKey string
 	if req.Query != "" {
-		q := traceql.CanonicalQuery(req.Query)
+		q := traceql.NormalizeQuery(req.Query)
 		if ast, err := traceql.Parse(q); err == nil {
 			// forces the query into a canonical form
 			cacheKey = ast.String()
--- a/modules/livestore/instance_search.go
+++ b/modules/livestore/instance_search.go
@@ -899,7 +899,7 @@ func includeBlock(b *backend.BlockMeta, start, end time.Time) bool {
 func searchTagValuesV2CacheKey(req *tempopb.SearchTagValuesRequest, limit int, prefix string) string {
 	var cacheKey string
 	if req.Query != "" {
-		q := traceql.CanonicalQuery(req.Query)
+		q := traceql.NormalizeQuery(req.Query)
 		if ast, err := traceql.Parse(q); err == nil {
 			// forces the query into a canonical form
 			cacheKey = ast.String()
--- a/pkg/traceql/lenient_extract_test.go
+++ b/pkg/traceql/lenient_extract_test.go
@@ -170,7 +170,7 @@ func TestCanonicalQuery(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			expected := tc.expected
 			expected = strings.ReplaceAll(expected, " ", "")
-			actual := CanonicalQuery(tc.query)
+			actual := NormalizeQuery(tc.query)
 			actual = strings.ReplaceAll(actual, " ", "")
 			actual = strings.ReplaceAll(actual, "`", `"`)
 			assert.Equal(t, expected, actual)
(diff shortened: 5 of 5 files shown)


pkg/traceql/extractmatcher.go
```diff
@@ -1,72 +1,128 @@
...
+	return filter.String()
+}
 
-	var q strings.Builder
-	q.WriteString("{")
-	for i, m := range matchers {
-		m = strings.TrimSpace(m)
-		if i > 0 {
-			q.WriteString(" && ")
+func RemoveUnnecessaryParentheses(query string) string {
```
mdisibio: nit: if this is just for testing, we could move it to that location and unexport it.

pkg/traceql/extractmatcher_test.go
```diff
@@ -31,14 +32,14 @@ func TestExtractMatchers(t *testing.T) {
...
+			name:     "reversed operands with missing closing bracket",
 			query:    "{ 2 = .b ",
-			expected: "{}",
+			expected: "{ 2 = .b}",
 		},
 		{
 			name:     "long query",
-			query:    `{.service_name = "foo" && .http.status_code = 200 && .http.method = "GET" && .cluster = }`,
-			expected: `{.service_name = "foo" && .http.status_code = 200 && .http.method = "GET"}`,
+			query:    `{(.service_name = "foo" && .http.status_code = 200) && .http.method = "GET" && .cluster = }`,
```
mdisibio: Is this maybe testing the handling of parentheses?  If so let's add it to the test case name, else I feel it should work without them

pkg/traceql/lenient_parser.go
```diff
@@ -0,0 +1,332 @@
...
+
+func isAttributeToken(typ int) bool {
+	return typ == DOT || typ == IDENTIFIER || isIntrinsicToken(typ) ||
+		typ == PARENT_DOT || typ == RESOURCE_DOT || typ == SPAN_DOT ||
+		typ == EVENT_DOT || typ == LINK_DOT || typ == INSTRUMENTATION_DOT ||
+		typ == EVENT_COLON || typ == LINK_COLON || typ == TRACE_COLON ||
+		typ == SPAN_COLON || typ == INSTRUMENTATION_COLON
+}
+
+func isIntrinsicToken(typ int) bool {
```
mdisibio, line 352: These functions seem a little brittle, in that they must be updated for every scope/intrinsic/operator. Is there another way to do this?  i.e. is there a way to handle `span:name = "bar"` without caring what `span:name` is, since we pass the cleaned string to the real Parse() anyway.

pkg/traceql/extractmatcher.go
ie-pham: I was planning on using this in cloud-traces as well

pkg/traceql/extractmatcher_test.go
ie-pham: oh interesting. i removed the parenthesis and it still works fine. maybe added during testing and forgot about it

pkg/traceql/lenient_parser.go
ie-pham: I've refactored some but we need these functions in order to know which token we should drop. If we want to parse `{ .attr = "foo" && <attribute> = }` into `{ .attr = "foo" }`, we need to know how many tokens to the left to drop. The original intrinsics are one token, the colon formatted ones are two, and then if it's a parent.span.attr, that's three tokens.

pkg/traceql/lenient_extract.go
```diff
@@ -0,0 +1,73 @@
...
+		return emptyQuery
+	}
+
+	expr, err := ParseLenient(query)
+	if err != nil {
+		return emptyQuery
+	}
+
+	if err := expr.validate(); err != nil {
+		return emptyQuery
```
mdisibio, line 65: The intent of this method is fine, but I probably would choose a different word than Canonical.  For example on this line if invalid we silently revert to the empty query, and the replacement of `{ true }` at the bottom. In my mind, `{ true }` is the actual canonical form for the empty query because the inner spanset really is an AST node returning true (`StaticTrue`).    I would expect canonical to parse/sort/rectify whitespace.

mapno: Good point. Renamed to `NormalizeQuery`.

PR #6453, opened 2026-02-13
feat: allow span filtering in service graph processor

javiermolinar:
**What this PR does**:
This PR introduces span filtering for the Service Graph processor to prevent the creation of unwanted edges. 
Ideally the filters should operate symetrically, discarding the clients and server spans. But since this is heavily depdendent on the intrumentation, it implements a best-effort cache to maintain graph integrity and prevent data skew.

**Why**:
Currently, the processor accepts every span, which can lead to cluttered or uncontrolled graphs. Some users generate service graphs from external sources (e.g., Grafana Beyla) and need the ability to disable graph generation for specific services within Tempo.

**How**:
The implementation reuses the spanfilter logic from the spanmetrics processor. To address the risk of creating "orphan edges" (where one side of a connection is filtered but the other isn't), a bounded cache tracks dropped spans:

1. Every incoming span is evaluated against the filter.
2. If a span is rejected, its key and side (client/server) is added to a "dropped spans" cache (a map with a TTL and a maximum element limit).
3. Before creating a new edge, the processor checks this cache. If the peer span exists in the cache, the edge is discarded.
4. TTL expiration and eviction are handled within the existing Expiration goroutine to minimize overhead.

Two new metrics are added:

**metrics_generator_processor_service_graphs_dropped_edges_total**: increments when an edge is dropped 
**metrics_generator_processor_service_graphs_dropped_span_side_cache_overflow_total**: it increments every time it tries to add a new span to the cache when the max items is reached

Some caveats:
- For simplicity, this cache reuses the existing edge store settings for TTL and `max_items`. We can monitor the new metrics to determine if independent configuration is required later.

- If the cache is full, new "dropped" markers won't be added until existing items expire. This happens every two seconds. This is a deliberate trade-off to prioritize performance over perfect accuracy.

**How it looks in action**:

Before filtering
Screenshot 2026-02-17 at 12 48 57

After adding the following filter:

```yaml
processor:
 service_graphs:
   filter_policies:
     - exclude:
       match_type: strict
       attributes:
         - key: resource.service.name
           value: shop-backend        
```
Screenshot 2026-02-17 at 12 39 54

The filter only removes the spans specifically tagged with that service name.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/sources/tempo/operations/manage-advanced-systems/user-configurable-overrides.md
+++ b/docs/sources/tempo/operations/manage-advanced-systems/user-configurable-overrides.md
@@ -92,7 +92,7 @@ metrics_generator:
               - key: <string>
                 value: <any>
         ]
-      ]
+      ]]
     span_metrics:
       [histogram_buckets: <list of float>]
       [dimensions: <list of string>]


javiermolinar:
@codex review

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -165,6 +188,12 @@ func (p *Processor) consume(resourceSpans []*v1_trace.ResourceSpans) (err error)
 
 		for _, ils := range rs.ScopeSpans {
 			for _, span := range ils.Spans {
+				if !p.filter.ApplyFilterPolicy(rs.Resource, span) {
+					p.addDroppedSpanSide(span)
+					p.filteredSpansCounter.Inc()
+					continue
```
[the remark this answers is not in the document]
javiermolinar: This is a valid point in theory. But it would require traverse the whole linked list in the store to drop existing unmatched edges for all the spans in the hot path. We can simply watch the dropped-edges and virtual nodes creation and optimize only if this became a problem

docs/sources/tempo/configuration/manifest.md
```diff
@@ -595,6 +595,7 @@ metrics_generator:
                 - db.namespace
                 - db.name
                 - db.system
+            filter_policies: []
```
electron0zero, line 598: also update config docs in `docs/sources/tempo/configuration/_index.md` and `docs/sources/tempo/operations/manage-advanced-systems/user-configurable-overrides.md` page

modules/generator/processor/servicegraphs/config.go
```diff
@@ -50,6 +51,9 @@ type Config struct {
 
 	// DatabaseNameAttributes the list of attribute names used to identify the database name from span attributes.
 	DatabaseNameAttributes []string `yaml:"database_name_attributes"`
+
+	// FilterPolicies is a list of policies that will be applied to spans for inclusion or exclusion.
```
electron0zero, line 55: this also has the new `include_any` policy as well? right?

modules/generator/processor/servicegraphs/servicegraphs.go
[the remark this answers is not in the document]
carles-grafana: what about filtering them here:
```
unc (s *store) tryEvictHead() bool {
        head := s.l.Front()
        if head == nil {
                return false
        }

        headEdge := head.Value.(*Edge)
        if !headEdge.isExpired() {
                return false
        }

        // If the missing side was intentionally filtered, silently evict
        // without calling onExpire to prevent phantom virtual-node edges.
        key := headEdge.key
        if len(headEdge.ServerService) == 0 && s.hasDroppedSpanSideLocked(key, Server) {
                s.deleteEdge(head)
                return true
        }
        if len(headEdge.ClientService) == 0 && s.hasDroppedSpanSideLocked(key, Client) {
                s.deleteEdge(head)
                return true
        }

        s.onExpire(headEdge)
        s.deleteEdge(head)
        return true
  }

  // hasDroppedSpanSideLocked checks if a dropped span side exists for the given key.
  // Must be called holding lock.
  func (s *store) hasDroppedSpanSideLocked(key string, side Side) bool {
        _, ok := s.d[droppedSpanSideKey{key: key, side: side}]
        return ok
  }

```

modules/generator/processor/servicegraphs/store/store.go
```diff
@@ -85,14 +109,16 @@ func (s *store) UpsertEdge(key string, update Callback) (isNew bool, err error)
...
-			delete(s.m, key)
-			s.l.Remove(storedEdge)
+			s.deleteEdge(storedEdge)
 		}
 
 		return false, nil
 	}
 
+	if s.hasDroppedCounterpart(key, side) {
+		return true, ErrDroppedSpanSide
```
carles-grafana, line 118: shouldn't this be false?

javiermolinar: No, the edge is new since we haven't found it in the map. Nevertheless that value is useless here since we are returning an error

modules/generator/processor/servicegraphs/servicegraphs.go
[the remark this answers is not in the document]
javiermolinar: This is not a bad idea, it's happens out of the hot-path and it solves the problem of the expired edges for filtered counterparts

modules/overrides/userconfigurable/api/limits_test.go
```diff
@@ -103,6 +104,19 @@ func Test_limitsFromOverrides(t *testing.T) {
         "peer_attributes": [
           "db.name"
         ],
+        "filter_policies": [
```
electron0zero, line 107: can you add a test where `filter_policies` are empty or nil

modules/overrides/userconfigurable/api/limits.go
electron0zero, line 102: I think this should do a similar check `if len(p) == 0` and return nil? wdyt?

modules/generator/processor/servicegraphs/config.go
javiermolinar: Yes, but the wording seems ok, include_any is inclusion

modules/generator/processor/servicegraphs/servicegraphs.go
```diff
@@ -165,6 +188,12 @@ func (p *Processor) consume(resourceSpans []*v1_trace.ResourceSpans) (err error)
 
 		for _, ils := range rs.ScopeSpans {
 			for _, span := range ils.Spans {
+				if !p.filter.ApplyFilterPolicy(rs.Resource, span) {
```
electron0zero, line 196: The filter runs on all span kinds, IIUC, `INTERNAL` spans get filtered but counted in the filteredSpansCounter.

I see that `addDroppedSpanSide` doesn't consider INTERNAL spans, and we are only building edges on the server and client spans.

does it make sense to gate this `ApplyFilterPolicy` and the metric by a check to make sure we only run it on the span kinds that can build server graph edges.

maybe we avoid the extra work of running `ApplyFilterPolicy` on the spans that can never build an edge??

docs/sources/tempo/operations/manage-advanced-systems/user-configurable-overrides.md
```diff
@@ -84,8 +84,16 @@ metrics_generator:
...
+      [filter_policies: [
+        [
+          include/include_any/exclude:
+            match_type: <string> # options: strict, regexp
+            attributes:
+              - key: <string>
+                value: <any>
+        ]
+      ]
+      span_metrics:
```
electron0zero:
```suggestion
    span_metrics:
```

fix indent.

docs/sources/tempo/configuration/_index.md
```diff
@@ -2013,7 +2013,15 @@ overrides:
...
-
+          [filter_policies: [
+            [
+              include/include_any/exclude:
+                match_type: <string> # options: strict, regexp
+                attributes:
+                  - key: <string>
+                    value: <any>
+            ]
+          ]
```
electron0zero:
```suggestion
          ]]
```

nit

docs/sources/tempo/configuration/_index.md
electron0zero:
```suggestion
          ]]
```

nit

docs/sources/tempo/configuration/_index.md
electron0zero: missing CHANGELOG.md entry.

modules/generator/processor/servicegraphs/servicegraphs.go
javiermolinar: Agree, especially for the correctness of the dropped edges metric

modules/overrides/userconfigurable/api/limits.go
javiermolinar: I think that would be a behavioral change. Currently

nil slice -> filter_policies: null
empty slice -> filter_policies: []

If we return nil when len(p) == 0 we lose the distinction

electron0zero: cool, let's leave it as is then, and we can come back if we need to,  dimension mapping was doing this check so asked.

docs/sources/tempo/operations/manage-advanced-systems/user-configurable-overrides.md
```diff
@@ -84,7 +84,15 @@ metrics_generator:
...
-
+      [filter_policies: [
+        [
+          include/include_any/exclude:
+            match_type: <string> # options: strict, regexp
+            attributes:
+              - key: <string>
+                value: <any>
+        ]
+      ]
```
electron0zero:
```suggestion
      ]]
```

oops, I missed it.

modules/generator/processor/servicegraphs/store/store_test.go
```diff
@@ -167,14 +170,202 @@ func TestStore_concurrency(t *testing.T) {
...
 
+func TestStore_AddDroppedSpanSide(t *testing.T) {
+	s := NewStore(time.Hour, 10, noopCallback, noopCallback, newTestCounter())
+
+	const key = "trace-span"
+	const side = Client
+	assert.False(t, s.HasDroppedSpanSide(key, side))
+
+	droppedCounterpart := s.AddDroppedSpanSide(key, side)
+	assert.False(t, droppedCounterpart)
```
electron0zero, line 181: nit: we usually use `require` over `assert` because require fails fast.

any reasons to use assert over require in these tests? if not, let's replace it with require so we fail fast.

modules/generator/processor/servicegraphs/servicegraphs_test.go
```diff
@@ -209,13 +219,120 @@ func TestServiceGraphs_failedRequests(t *testing.T) {
...
+				"server":          "postgres",
+				"connection_type": "database",
+			})
+			requesterToRecorderLabels := labels.FromMap(map[string]string{
+				"client":          "mythical-requester",
+				"server":          "mythical-recorder",
+				"connection_type": "messaging_system",
+			})
+
+			assert.Equal(t, tc.expectedRequesterServer, testRegistry.Query(`traces_service_graph_request_total`, requesterToServerLabels))
```
electron0zero, line 321: same here, let's use require over assert.

modules/generator/processor/servicegraphs/servicegraphs_test.go
```diff
@@ -355,6 +475,292 @@ func TestServiceGraphs_expiredEdges(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	}
+
+	p.PushSpans(context.Background(), request)
+
+	emptyParentKey := buildKey(hex.EncodeToString(traceID), "")
+	assert.False(t, p.(*Processor).store.HasDroppedSpanSide(emptyParentKey, store.Server))
```
electron0zero, line 761: same here, require over assert unless we need to NOT fail fast.

modules/generator/processor/servicegraphs/store/store_test.go
javiermolinar: Previous tests followed that convention I just did the same

PR #6458, opened 2026-02-13
tempo-cli: support relative time strings for start/end args

electron0zero:
**What this PR does**:

Add support for relative time strings across all CLI commands that accept start/end arguments, matching Grafana's time string format.

example relative time: now-1h, now, now-30m, etc.

example usage: `tempo-cli query api search 127.0.0.1:3100 '{status = error}' now-12d5m now-12d`

Affected commands:
- query api search
- query api search-tags
- query api search-tag-values
- query api metrics query-range
- query search

Help text previously said "ISO8601 format" or showed a ISO 8601 format but the code was parsing with `time.RFC3339` (2006-01-02T15:04:05Z07:00). ISO 8601 is a broad standard that allows many representations including timestamps without timezone offsets.

RFC 3339 is a strict profile of ISO 8601 that always requires a timezone offset (Z or +HH:MM).

Help text is update to say RFC3339 with an example.

We now accept a uniform time format across all commands: either RFC3339 absolute timestamps or relative time strings.

query search: previously accepted timestamps without timezone (2024-03-15T10:30:00) using the layout "2006-01-02T15:04:05". This is valid ISO 8601 but not valid RFC 3339. These timestamps are now rejected. Users must include a timezone offset (e.g. 2024-03-15T10:30:00Z) or use relative time (e.g. now-1h).

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/cmd/tempo-cli/shared.go
+++ b/cmd/tempo-cli/shared.go
@@ -169,6 +169,8 @@ func grpcTransportCredentials(secure bool) (opt grpc.DialOption, err error) {
 // - relative: "now", "now-1h", "now-30m", "now-3h30m"
 // - RFC3339: "2024-01-01T00:00:00Z"
 func parseTime(s string) (time.Time, error) {
+	s = strings.TrimSpace(s)
+
 	if strings.HasPrefix(s, "now") {
 		return parseRelativeTime(s)
 	}
--- a/cmd/tempo-cli/shared_test.go
+++ b/cmd/tempo-cli/shared_test.go
@@ -76,6 +76,31 @@ func TestParseTime(t *testing.T) {
 			input:   "now-abc",
 			wantErr: true,
 		},
+		{
+			name:     "leading spaces are trimmed",
+			input:    "  now-1h",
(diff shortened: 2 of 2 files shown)


cmd/tempo-cli/shared.go
```diff
@@ -162,3 +164,40 @@ func grpcTransportCredentials(secure bool) (opt grpc.DialOption, err error) {
 
 	return grpc.WithTransportCredentials(creds), nil
 }
+
+// parseTime parses a time string that can be:
+// - relative: "now", "now-1h", "now-30m", "now-3h30m"
+// - RFC3339: "2024-01-01T00:00:00Z"
+func parseTime(s string) (time.Time, error) {
+	if strings.HasPrefix(s, "now") {
```
javiermolinar, line 174: We should trim the spaces here

```go
s := " now"
if strings.HasPrefix(s, "now") {
	fmt.Println("foo")
} else {
	fmt.Println("bar")
}
```

electron0zero: done

PR #6459, opened 2026-02-13
[backend-scheduler] initial redaction provider

zalegrala:
**What this PR does**:

Here we implement the redaction pipeline for the backend scheduler and worker. It introduces a `SubmitRedaction` gRPC API through which an operator provides a tenant ID and a list of trace IDs to remove. The scheduler fans out one pending job per block, workers redact each block independently, and the batch is cleaned up automatically when all jobs complete.

* add proto definitions for redaction batch and job, and the `SubmitRedaction` API
* add a redaction provider to the scheduler, and a redaction worker to the worker

Updates to `work`:
This change pushes a small evolution of the `work` package to support two new states.  Previously, jobs were unknown until they were assigned to a worker and added.  Adding the new provider here for redaction requires information from the compaction provider, and the same is true in reverse.  To avoid dependencies between the providers, we introduce a pending, and registered state.  Note that previously, the compaction provider need to track this internally to avoid creating jobs operating on duplicate blocks, but now we centralize this same approach on `work`.

* two new states are added to `work`:
    * registered: a job which has been created by a provider, but not yet recorded through the channels.  Not persisted.
    * pending: the job is known to the scheduler, but not yet assigned to a worker. Persisted

* persist redaction batches to local disk
* interface extensions to serve:
    * early job registration to allow providers to make better decisions, earlier
    * batch registration to keep a single copy of the trace IDs
    * pending jobs to allow a snapshot of the blocklist to inform the total work for a provider before assignment to workers.

* include indexes for some of the above to maintain performance.

Updates to existing providers:
* compaction: halt compaction as soon as a redaction batch is active
* retention: move the retention provider job to per-tenant
* retention: implement tenant skipping when compaction is disabled (previously handled at the worker) or a redaction is running.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/backendscheduler/work/batch.go
+++ b/modules/backendscheduler/work/batch.go
@@ -102,11 +102,6 @@ func (b *batchStore) setRescan(tenantID string, ids []string, afterNano int64) {
 	}
 }
 
-// clearRescan zeroes the rescan fields on the batch for tenantID under the write lock.
-func (b *batchStore) clearRescan(tenantID string) {
-	b.setRescan(tenantID, nil, 0)
-}
-
 // load reads batches.pb from localPath. Missing file is not an error (clean start).
 func (b *batchStore) load(localPath string) error {
 	path := filepath.Join(localPath, batchesFileName)


modules/backendscheduler/provider/compaction.go
```diff
@@ -83,13 +78,12 @@ func NewCompactionProvider(
 	scheduler Scheduler,
 ) *CompactionProvider {
 	return &CompactionProvider{
-		cfg:             cfg,
-		logger:          logger,
-		store:           store,
-		overrides:       overrides,
-		curPriority:     tenantselector.NewPriorityQueue(),
-		sched:           scheduler,
-		outstandingJobs: make(map[string][]backend.UUID),
```
zalegrala, line 92: What `outstandingJobs` was doing has been moved to `work` to allow sharing this information to other providers.

modules/backendscheduler/provider/compaction.go
```diff
@@ -169,8 +163,19 @@ func (p *CompactionProvider) Start(ctx context.Context) <-chan *work.Job {
 				continue
 			}
 
-			// Job successfully created, add to recent jobs cache before we send it.
-			p.addToRecentJobs(ctx, job)
+			// Re-check after job creation: a batch may have been submitted while we were
+			// draining the selector that was built before the submission. Discard the job
+			// before it enters the recent-jobs cache or the channel.
+			if p.sched.HasJobsForTenant(p.curTenant.Value(), tempopb.JobType_JOB_TYPE_REDACTION) || p.sched.TenantPending(p.curTenant.Value()) {
```
stoewer: `question` why does this only apply to `JOB_TYPE_REDACTION`?

modules/backendscheduler/provider/types.go
```diff
@@ -15,4 +16,25 @@ type Provider interface {
 // Scheduler interface defines the methods providers need from the scheduler
 type Scheduler interface {
 	ListJobs() []*work.Job
+	PopNextPendingJob(jobType tempopb.JobType) *work.Job
```
stoewer: `nitpick` PopNext sounds like a duplicate. I'd rather use pop or next as a prefix (probably prefer next, as pop implies LIFO)

modules/backendscheduler/work/work.go
```diff
@@ -23,16 +23,52 @@ const (
...
 }
 
 type Work struct {
 	Shards [ShardCount]*Shard `json:"shards"`
 	cfg    Config
 	mtx    sync.Mutex // Protects the entire Work structure during Marshal/Unmarshal
+
+	// pendingBlocks indexes (tenantID, blockID) -> jobID for fast pending-block lookup.
+	// Not persisted; rebuilt on LoadFromLocal and in Unmarshal from Shard.Pending.
+	pendingBlocks map[string]string `json:"-"`
```
stoewer: I don't think this `json:"-"` tag is needed (applies to other new fields as well)

stoewer:
### Code review

Found 1 issue:

1. **Data race: `performRescan` mutates batch fields after releasing the lock** — `GetBatch` returns a raw pointer to the live `*tempopb.RedactionBatch` inside `batchStore.byTenant`, but `performRescan` then writes to `SkippedCompactionJobIds` and `RescanAfterUnixNano` without holding any lock. Concurrently, the background flush ticker calls `FlushBatchesToLocal → batchStore.flush`, which holds only a read lock while `proto.Marshal` reads those same fields — exactly the race the comment in `flush` warns against.

https://github.com/grafana/tempo/blob/01dce78b213d3d0069d0c4436830577dc547dfa1/modules/backendscheduler/backendscheduler.go#L690-L696

The intent documented in `batch.go` is that the read lock must be held through marshalling to prevent concurrent field mutation:

https://github.com/grafana/tempo/blob/01dce78b213d3d0069d0c4436830577dc547dfa1/modules/backendscheduler/work/batch.go#L59-L70

The fix would be to perform the re-arm write via a dedicated method on `batchStore` (like `clearRescan`) that holds the write lock for the entire read-modify-write on `SkippedCompactionJobIds` and `RescanAfterUnixNano`, rather than returning a raw pointer for the caller to mutate.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

modules/backendscheduler/provider/compaction.go
zalegrala: This is a good call-out.  `TenantPending` is enough here, since active REDACTION jobs will only ever be present when there is a batch running with an unknown state, which is covered by the `TenantPending`.  I'll push a commit.

modules/backendscheduler/provider/types.go
zalegrala: I've pushed a commit for this.

zalegrala: I've pushed a commit for this.

modules/backendscheduler/work/work.go
zalegrala: Good call.  I've pushed a commit for this.

modules/backendscheduler/provider/compaction.go
zalegrala: I've pushed a commit for this.  Let me know if there is more to discuss.

zalegrala:
Thanks for the review.  I pushed a few commits to address your comments.  Let me know what else stands out.

PR #6467, opened 2026-02-16
distributor: log truncated oversized attributes with rate limited logger

carles-grafana:
**What this PR does**:

global rate-limited logger for truncation warnings

The truncation log line is a supplementary diagnostic: it tells operators which attribute was truncated, at what scope, and how large it was. The primary observability channel is metricAttributesTruncated, which fires unconditionally per-tenant on every truncation.

This approach guarantees the first truncation event in any time window is always logged.

Rate-limited logger over sampling because at peak load (e.g. 20k push req/s per tenant), sampling at 0.1% still produces 20 log lines/s per pod. The rate-limited logger caps it at 1/s regardless of load.

The log lines are most useful for the noisiest tenant: the one generating the most truncations is the one operators most need diagnostic detail for. If tenant A dominates at 100k truncations/s, that's the tenant we want to see in logs. A quieter tenant B (10/s) is suppressed from logs but fully visible via per-tenant metrics.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/distributor/distributor_test.go
+++ b/modules/distributor/distributor_test.go
@@ -1191,32 +1191,6 @@ func TestProcessAttributesDetail(t *testing.T) {
 	assert.Equal(t, 3000, detail.origSize)
 }
 
-func BenchmarkProcessAttributes(b *testing.B) {
-	const maxAttrSize = 2048
-	const numAttrs = 100
-
-	makeAttrs := func() []*v1_common.KeyValue {
-		attrs := make([]*v1_common.KeyValue, numAttrs)
-		for i := range attrs {
-			attrs[i] = test.MakeAttribute("key", strings.Repeat("v", 5000))
-		}
-		return attrs
-	}
-
-	b.Run("truncation_no_detail", func(b *testing.B) {
-		for i := 0; i < b.N; i++ {
-			processAttributes(makeAttrs(), maxAttrSize, nil, "span")
-		}
-	})
-
-	b.Run("truncation_with_detail", func(b *testing.B) {
-		for i := 0; i < b.N; i++ {
-			detail := truncatedAttrInfo{}
-			processAttributes(makeAttrs(), maxAttrSize, &detail, "span")
-		}
-	})
-}
-
 func BenchmarkTestsByRequestID(b *testing.B) {
 	spansPer := 5000
 	batches := 100


modules/distributor/distributor.go
```diff
@@ -849,19 +851,32 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount, ma
...
+// processAttributes finds and truncates attribute keys/values that exceed maxAttrSize.
+// Truncated attributes are logged via the rate-limited logger when provided.
+func processAttributes(attributes []*v1_common.KeyValue, maxAttrSize int, logger *tempolog.RateLimitedLogger, tenant, scope string) int {
 	count := 0
 	for _, attr := range attributes {
 		if len(attr.Key) > maxAttrSize {
+			originalSize := len(attr.Key)
 			attr.Key = attr.Key[:maxAttrSize]
+			if logger != nil {
+				logger.Log("msg", "attribute truncated", "tenant", tenant, "scope", scope,
```
javiermolinar: There are two risks with this approach:
- Logging has a performance impact in the hot-path, this rate limit logger introduces checks that under bad traffic can result in degradation
- Logs volume can be massive for customers that hit these limits, even with a hard limit of 10

carles-grafana:
benchmark comparison with count 8, no significant difference

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/tempo/modules/distributor
cpu: Apple M3 Pro
                                            │             │
                                            │   sec/op    │
ProcessAttributes/truncation_no_detail-11     54.61µ ± 2%
ProcessAttributes/truncation_with_detail-11   52.16µ ± 2%
geomean                                       53.37µ

                                            │              │
                                            │     B/op     │
ProcessAttributes/truncation_no_detail-11     530.5Ki ± 0%
ProcessAttributes/truncation_with_detail-11   530.5Ki ± 0%
geomean                                       530.5Ki

                                            │            │
                                            │ allocs/op  │
ProcessAttributes/truncation_no_detail-11     400.0 ± 0%
ProcessAttributes/truncation_with_detail-11   400.0 ± 0%
geomean                                       400.0
```

modules/distributor/distributor.go
carles-grafana: I've implemented the sampling approach, the benchmark show the overhead is negligible

modules/distributor/distributor.go
```diff
@@ -729,51 +753,54 @@ func (d *Distributor) sendToKafka(ctx context.Context, userID string, keys []uin
...
-func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount, maxSpanAttrSize int) ([]uint32, []*rebatchedTrace, truncatedAttributesCount, error) {
+func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount, maxSpanAttrSize int) ([]uint32, []*rebatchedTrace, truncatedAttributesCount, *truncatedAttrInfo, error) {
 	const tracesPerBatch = 20 // p50 of internal env
 	tracesByID := make(map[uint64]*rebatchedTrace, tracesPerBatch)
 	truncatedCount := truncatedAttributesCount{}
+
+	// Probabilistically sample requests for detailed truncation logging.
+	// nil means not sampled — zero overhead in processAttributes.
+	var detail *truncatedAttrInfo
+	if maxSpanAttrSize > 0 && rand.Float64() < truncationLoggingSamplingRate { // nolint:gosec // G404: cryptographic randomness not needed for sampling
```
javiermolinar: If we do the sampling here we are going to lose most of the truncated attributes logs. We need to do the sampling at the moment an attribute is truncated

modules/distributor/distributor.go
```diff
@@ -729,51 +753,54 @@ func (d *Distributor) sendToKafka(ctx context.Context, userID string, keys []uin
 
 // requestsByTraceID takes an incoming tempodb.PushRequest and creates a set of keys for the hash ring
 // and traces to pass onto the ingesters.
-func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount, maxSpanAttrSize int) ([]uint32, []*rebatchedTrace, truncatedAttributesCount, error) {
+func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount, maxSpanAttrSize int) ([]uint32, []*rebatchedTrace, truncatedAttributesCount, *truncatedAttrInfo, error) {
```
javiermolinar, line 757: Instead of returning the pointer to the struct we could pass it from the outside.

modules/distributor/distributor.go
```diff
@@ -506,6 +519,17 @@ func (d *Distributor) PushTraces(ctx context.Context, traces ptrace.Traces) (*te
 		metricAttributesTruncated.WithLabelValues(userID, "span").Add(float64(truncatedAttributesCount.Span))
 		metricAttributesTruncated.WithLabelValues(userID, "event").Add(float64(truncatedAttributesCount.Event))
 		metricAttributesTruncated.WithLabelValues(userID, "link").Add(float64(truncatedAttributesCount.Link))
+
+		if sampledDetail != nil {
+			level.Warn(d.logger).Log("msg", "attributes truncated",
+				"tenant", userID,
+				"total_truncated", truncatedAttributesCount.Total(),
+				"max_size_bytes", maxAttributeBytes,
```
javiermolinar, line 527: I think we don't need to add this on every single log line since is a per tenant config.

modules/distributor/distributor.go
```diff
@@ -846,21 +873,29 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount, ma
...
+// processAttributes finds and truncates attribute keys/values that exceed maxAttrSize.
+// When detail is non-nil and unfilled (sampled request), the first truncation is recorded for logging.
+func processAttributes(attributes []*v1_common.KeyValue, maxAttrSize int, detail *truncatedAttrInfo, scope string) int {
 	count := 0
 	for _, attr := range attributes {
 		if len(attr.Key) > maxAttrSize {
+			origSize := len(attr.Key)
 			attr.Key = attr.Key[:maxAttrSize]
+			if detail != nil && detail.origSize == 0 {
+				*detail = truncatedAttrInfo{scope: scope, name: attr.Key, field: "key", origSize: origSize}
```
javiermolinar: attr.Key is truncated in the previous line.

carles-grafana: this is on purpose, if the attribute was huge it could be problematic to log it

modules/distributor/distributor.go
carles-grafana: I think it's good to have the context though, to quickly compare it with the original size

modules/distributor/distributor.go
carles-grafana: I was under the impression we wanted to do sampling, that's why we put it before the truncation. But if we want to do it only when we are sure there will be truncation, the rate-limited logger approach might be simpler.

carles-grafana: In the end I've implemented a rate-limited logger approach after the truncation decision, I believe it's the simplest. See the PR description for the rationale.

modules/distributor/distributor.go
```diff
@@ -284,6 +296,7 @@ func New(
 		overrides:            o,
 		traceEncoder:         model.MustNewSegmentDecoder(model.CurrentEncoding),
 		tracePushMiddlewares: cfg.TracePushMiddlewares,
+		truncationLogger:     tempo_log.NewRateLimitedLogger(truncationLogsPerSecond, level.Warn(logger)),
```
electron0zero, line 299: is it better served to be at Info or debug level instead of warn?

if it's a warn, operators who who don't want this log line will have to give up other warnings logs to stop logging this?

I am mainly thinking about the log volume from this log line in the deployments where operators expect truncation as normal mode of operation and don't want to see these log lines?

maybe this concern can be solved by making `truncationLogsPerSecond` configurable? so operators can control the volume of these log lines??

modules/distributor/distributor_test.go
```diff
@@ -1092,6 +1092,131 @@ func TestProcessAttributes(t *testing.T) {
...
+	assert.Equal(t, 1, count)
+	assert.Equal(t, "first", detail.name)
+	assert.Equal(t, 3000, detail.origSize)
+}
+
+func BenchmarkProcessAttributes(b *testing.B) {
+	const maxAttrSize = 2048
+	const numAttrs = 100
+
+	makeAttrs := func() []*v1_common.KeyValue {
```
javiermolinar: This is not reliable since most of the time of the benchmark is crafting the test data. We should do it just once and reuse it between iterations

modules/distributor/distributor.go
```diff
@@ -846,21 +870,32 @@ func requestsByTraceID(batches []*v1.ResourceSpans, userID string, spanCount, ma
...
-// find and truncate the span attributes that are too large
-func processAttributes(attributes []*v1_common.KeyValue, maxAttrSize int) int {
+// processAttributes finds and truncates attribute keys/values that exceed maxAttrSize.
+func processAttributes(attributes []*v1_common.KeyValue, maxAttrSize int, truncationExample *truncatedAttrInfo, scope string) int {
 	count := 0
 	for _, attr := range attributes {
 		if len(attr.Key) > maxAttrSize {
+			origSize := len(attr.Key)
 			attr.Key = attr.Key[:maxAttrSize]
+			if truncationExample != nil && truncationExample.origSize == 0 { // only capture the first truncation
```
javiermolinar, line 886: I'm ok with this. We could also keep a example per scope but we should measure the impact

modules/distributor/distributor.go
```diff
@@ -506,6 +519,17 @@ func (d *Distributor) PushTraces(ctx context.Context, traces ptrace.Traces) (*te
 		metricAttributesTruncated.WithLabelValues(userID, "span").Add(float64(truncatedAttributesCount.Span))
 		metricAttributesTruncated.WithLabelValues(userID, "event").Add(float64(truncatedAttributesCount.Event))
 		metricAttributesTruncated.WithLabelValues(userID, "link").Add(float64(truncatedAttributesCount.Link))
+
+		if truncationExample != nil {
+			d.truncationLogger.Log("msg", "attributes truncated",
```
javiermolinar, line 524: Since we want to expose this to tenants should we include the insight=true?

modules/distributor/distributor.go
carles-grafana: I would avoid making this configurable unless we have a clear need. It's only 1 per second, so not a big deal.

> where operators expect truncation as normal mode of operation and don't want to see these log lines

Is this a known use case?

modules/distributor/distributor.go
carles-grafana: We can do this later once we see it's good from our side

modules/distributor/distributor.go
electron0zero: I would say yes, there might be OSS users who expect traces to contain things like long stacktraces or LM prompts and want to truncate them to ensure stability of the system.

> It's only 1 per second, so not a big deal.

true, but that's still 86400 warning log lines per day

> I would avoid making this configurable unless we have a clear need

agree, but if we make it at `info` or `debug` level, folks who don't want to see it can drop this logline by changing the log level to `warn`.

carles-grafana: I still think we should start with warning but not a strong opinion. @javiermolinar thoughts, since you requested the feature?

modules/distributor/distributor_test.go
carles-grafana: It is reliable in relative terms to show no extra allocations and a tiny overhead of nanoseconds. But you are right that ideally it would be done outside, but the problem is that processAttributes truncates the slice values in-place, so we would need to pre-generate millions of slices. Not doable. So I'm going do remove this once, since it's no longer useful and I pasted the benchstat results in the PR.

modules/distributor/distributor.go
javiermolinar: We can use warning since we are actually truncating the data. If they query back that attribute can be useless. Since it’s rate-limited to 1/sec, the volume shouldn’t be excessive

carles-grafana: We are also using warn for trace too large errors:  https://github.com/grafana/tempo/blob/52e6cc95b/modules/livestore/instance.go#L128

electron0zero: cool, let's leave it as as, and we can tweak it if we see issues.

Issue #6417, opened 2026-02-10
[Epic] TraceQL Math


ruslan-mikhailov:
This epic tracks work related to supporting TraceQL Math queries.

**Scope:**
- Simple arithmetic operations (+, -, /, *) between queries: A + B (e.g. `({status=error} | rate()) / ({} | rate())`)
- Simple arithmetic operations (+, -, /, *) between a query and a number: A * 100 (e.g. `({} | rate()) * 100`)
- Comparison operations: A > 5 (e.g. `({} | rate()) by (resource.service.name) > 5`)

ruslan-mikhailov:
Comparison operations (https://github.com/grafana/tempo/pull/6474) are included in 3.0.
Arithmetic operations are merged and highly likely will be included in 3.1 (https://github.com/grafana/tempo/pull/6866 and https://github.com/grafana/tempo/pull/7199).

There are issues for Grafana UI aiming improving user experience for Math, not blockers:

- https://github.com/grafana/grafana-tempo-datasource/issues/158
- https://github.com/grafana/grafana-tempo-datasource/issues/154 
- https://github.com/grafana/lezer-traceql/pull/80

PR #6474, opened 2026-02-16
[Feature] traceql comparison operators

ruslan-mikhailov:
**What this PR does**: it introduces comparison queries to TraceQL Metrics such as `{} | rate() > 10`. It is sensitive to step variable, especially to queries like count_over_time().
**Examples:**
No filters
image
More than 1
image
More than 3. In this example you can see that Grafana connects dots even when NaN is between them. This should be addressed in a separate issue (proposal is here https://github.com/grafana/tempo/pull/6486).
image
How it works if Tempo returns NaN (**not** included in this PR!)
image

Duration
image
Combination with topk
image

Fixes #<issue number> https://github.com/grafana/tempo/issues/6417

**Checklist**
- [x] Tests updated
- [ ] Documentation added


pkg/traceql/expr.y
```diff
@@ -331,6 +335,28 @@ metricsSecondStage:
     | BOTTOMK OPEN_PARENS INTEGER CLOSE_PARENS                   { $$ = newTopKBottomK(OpBottomK, $3) }
   ;
 
+// **********************
+// Metrics Filter (comparison operators applied to metrics results)
+// Uses == for equality instead of = to distinguish from field-level equality
+// **********************
+metricsFilterOperation:
+    EQ         { $$ = OpEqual        }
```
ruslan-mikhailov, line 341: I picked `=` and not `==` to align with how we compare attributes

pkg/traceql/expr.y
```diff
@@ -67,6 +67,8 @@ import (
 %type <scalarFilterOperation> scalarFilterOperation
 %type <metricsAggregation> metricsAggregation
 %type <metricsSecondStage> metricsSecondStage
+%type <scalarFilterOperation> metricsFilterOperation
```
ruslan-mikhailov, line 72: After we agree on syntax, we can add the feature to documentation

pkg/traceql/parse_test.go
```diff
@@ -1679,6 +1679,42 @@ func TestMetricsSecondStage(t *testing.T) {
 				newHint("foo", NewStaticString("bar")),
 			})),
 		},
+		// Combined second stage tests (topk/bottomk with filter)
+		{
+			in: `{ } | rate() | topk(5) > 10`,
```
javiermolinar, line 1798: Can we invert the order?

```
{ } | rate() > 10 | topk(5) 
```

pkg/traceql/ast_metrics.go
```diff
@@ -421,3 +421,129 @@ func (m *TopKBottomK) process(input SeriesSet) SeriesSet {
...
+			Labels:    series.Labels,
+			Values:    newValues,
+			Exemplars: exemplars,
+		}
+	}
+
+	return result
+}
+
+func (m *MetricsFilter) compare(v float64) bool {
```
javiermolinar, line 497: This is calculated for every value and every exemplar. We can pre-calculate the comparation function in the init in the same way we do here https://github.com/grafana/tempo/blob/b9f3aa30585f1400db0ffd2c07d4d9a70411eace/pkg/traceql/engine_metrics.go#L489

It's worth to do a benchmark

pkg/traceql/ast_metrics.go
```diff
@@ -421,3 +421,129 @@ func (m *TopKBottomK) process(input SeriesSet) SeriesSet {
...
+			hasValue = true
+		}
+
+		if !hasValue {
+			continue
+		}
+
+		exemplars := make([]Exemplar, 0, len(series.Exemplars))
+		for i, e := range series.Exemplars {
+			if m.compare(e.Value) {
```
javiermolinar: We need to guard here against NaN too since NaN != 10 is true

pkg/traceql/expr.y
```diff
@@ -331,6 +335,28 @@ metricsSecondStage:
     | BOTTOMK OPEN_PARENS INTEGER CLOSE_PARENS                   { $$ = newTopKBottomK(OpBottomK, $3) }
   ;
 
+// **********************
+// Metrics Filter (comparison operators applied to metrics results)
+// Uses == for equality instead of = to distinguish from field-level equality
```
javiermolinar: Comment is wrong?

javiermolinar (commented):
Nice feature! I missed some benchmarks

pkg/traceql/parse_test.go
ruslan-mikhailov: In the current solution, it is not possible. I can implement, but do we actually need this feature? 🤔

pkg/traceql/ast_metrics.go
ruslan-mikhailov: I tried it:
```
func newMetricsFilter(op Operator, value float64) *MetricsFilter {
	return &MetricsFilter{op: op, compare: compareFunc(op, value), value: value}
}

func compareFunc(op Operator, value float64) func(float64) bool {
	switch op {
	case OpGreater:
		return func(v float64) bool { return v > value }
	case OpGreaterEqual:
		return func(v float64) bool { return v >= value }
	case OpLess:
		return func(v float64) bool { return v < value }
	case OpLessEqual:
		return func(v float64) bool { return v <= value }
	case OpEqual:
		return func(v float64) bool { return v == value }
	case OpNotEqual:
		return func(v float64) bool { return v != value }
	default:
		return func(v float64) bool { return false }
	}
}
```

Funny that it is actually slower!

```
goos: darwin
goarch: arm64
cpu: Apple M4 Pro
                         │ /tmp/old.txt │             /tmp/new.txt             │            /tmp/third.txt            │
                         │    sec/op    │    sec/op     vs base                │    sec/op     vs base                │
MetricsFilter_compare-12   0.4036n ± 6%   0.7220n ± 5%  +78.92% (p=0.000 n=10)   0.7292n ± 5%  +80.68% (p=0.000 n=10)

                         │ /tmp/old.txt │          /tmp/new.txt          │         /tmp/third.txt         │
                         │     B/op     │    B/op     vs base            │    B/op     vs base            │
MetricsFilter_compare-12     0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=10) ¹   0.000 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal

                         │ /tmp/old.txt │          /tmp/new.txt          │         /tmp/third.txt         │
                         │  allocs/op   │ allocs/op   vs base            │ allocs/op   vs base            │
MetricsFilter_compare-12     0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=10) ¹   0.000 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```
old is the version in the PR, new is the optimisation attempt, third is a variant with no variable closure


I think the reason is that compare was inlined in the first solution, but not in "optimisation"-solutions:
```
pkg/traceql/ast_metrics.go:467:34: inlining call to (*MetricsFilter).compare
pkg/traceql/ast_metrics.go:482:16: inlining call to (*MetricsFilter).compare
```
(checked by `go build -gcflags="-m" ./pkg/traceql/ 2>&1 | grep -i compare | grep ast`)

Therefore, I leave current solution + benchmarks

pkg/traceql/ast_metrics.go
ruslan-mikhailov: Nice, catch! Fixed by dropping such exemplars

pkg/traceql/expr.y
ruslan-mikhailov: Oh, forgot to delete it. I used == in the first version.

Fixed

ruslan-mikhailov:
\+ review fixes
\+ tests for `.String()`

pkg/traceql/parse_test.go
javiermolinar: I guess so? I think allowing both is the best user experience. Right now it will be confussing why we allow one form and not the other

mdisibio: Yes agree, I would expect to be able to use the scalar filter on any part of the pipeline which is metrics.  Also the queries `{ } | rate() | topk(5) > 10` and `{ } | rate() > 10 | topk(5)`  are meaningfully different.   

I think we could add the scalar filter as another second stage function, so that `{ } | rate() > 10 | topk(5) < 11` is parsed as:

spanpipeine:  { }
first stage: rate()
second stages: > 10, topk(5), < 11

An argument could be made to add the scalar filter inside a stage.  Depending on if that has very large implementation benefits.

If this helps, here is a query with multiple metrics stages and regrouping that I would like to support eventually:

`{ } | rate() by (pod) > 10 | topk(5) by (namespace) | sum() by (cluster) > 20`

pkg/traceql/ast_metrics.go
mdisibio: This is likely because the function pointers are not inline-able.

pkg/traceql/parse_test.go
```diff
@@ -1679,6 +1679,42 @@ func TestMetricsSecondStage(t *testing.T) {
...
+					NewIntrinsic(IntrinsicName),
+				}),
+				newChainedSecondStage(
+					newTopKBottomK(OpBottomK, 3),
+					newMetricsFilter(OpGreaterEqual, 5.5),
+				),
+			),
+		},
+		{
+			in: `{ } | count_over_time() | topk(10) != 0`,
```
electron0zero, line 1824: can we add a test case that has flags ` with(sample=0.1)` or other flags ? to ensure this works with the query options.

pkg/traceql/test_examples.yaml
```diff
@@ -171,6 +171,15 @@ valid:
   - '{} | count_over_time() by (name) | bottomk(10) with(sample=0.1)'
   - '{} | quantile_over_time(duration, 0, 0.9, 1) by (span.http.path) | bottomk(10)'
   - '{} | avg_over_time(duration) by (span:name) | bottomk(10)'
+  # metrics filter (comparison on metrics results)
+  - '{} | rate() > 10'
+  - '{} | rate() by (span.client_ip) >= 5.5'
+  - '{} | count_over_time() by (name) < 100'
+  - '{} | rate() by (span.client_ip) <= 0.5'
+  - '{} | rate() = 42'
```
electron0zero, line 179: can we test each `op` with query flags like ` with(sample=0.1)`??

pkg/traceql/test_examples.yaml
```diff
@@ -326,6 +339,7 @@ validate_fails:
   # compare function with second stage functions is not valid
   - '{} | compare({status=error}) | topk(10)'
   - '{} | compare({status=error}) | bottomk(10)'
+  - '{} | compare({status=error}) > 10'
```
electron0zero, line 371: can we also add all `op` here to the full covergae.

tempodb/tempodb_metrics_test.go
```diff
@@ -601,6 +601,156 @@ var queryRangeTestCases = []struct {
...
+					{TimestampMs: 30_000, Value: 30},
+					{TimestampMs: 45_000, Value: 45},
+					{TimestampMs: 60_000, Value: 50},
+				},
+			},
+		},
+		expectedL2: nil,
+		expectedL3: []*tempopb.TimeSeries{
+			{
+				Labels: []common_v1.KeyValue{tempopb.MakeKeyValueString("__name__", "max_over_time")},
```
electron0zero, line 773: can we have a test with query flags like ` with(sample=0.1)`

pkg/traceql/engine_metrics_test.go
```diff
@@ -2300,6 +2300,68 @@ func TestSecondStageBottomKInstant(t *testing.T) {
...
+			expectedValues: []float64{5, 5, 5, 5, 5, 5, 5, 5},
+		},
+		{
+			name:           "less than or equal",
+			query:          "{ } | rate() by (span.foo) <= 3",
+			expectedLabel:  "quax",
+			expectedValues: []float64{3, 3, 3, 3, 3, 3, 3, 3},
+		},
+		{
+			name:           "less than",
```
electron0zero, line 2347: add a test case to ensure queries like ` with(sample=0.1)` work as expected.

pkg/traceql/ast_metrics.go
```diff
@@ -421,3 +421,129 @@ func (m *TopKBottomK) process(input SeriesSet) SeriesSet {
...
+		return false
+	}
+}
+
+var _ secondStageElement = (*MetricsFilter)(nil)
+
+// ChainedSecondStage chains two second stage elements together.
+// The first element is processed, then its output is passed to the second.
+// Example: {status=error} | rate() | topk(5) > 10
+type ChainedSecondStage struct {
```
electron0zero, line 521: nice, I like how you are using the second stage to add support for these 💯

pkg/traceql/ast_metrics.go
electron0zero: possible, maybe it makes sense to leave this finding as a comment so we are aware :)

pkg/traceql/parse_test.go
oleg-kozlyuk-grafana: I concur with the above. This feature would benefit from more time on drawing board to ensure more cases are supported, or there is a concrete plan on how to achieve that

pkg/traceql/test_examples.yaml
ruslan-mikhailov: Added

pkg/traceql/parse_test.go
ruslan-mikhailov: Added a separate test case

tempodb/tempodb_metrics_test.go
ruslan-mikhailov: The test has too few spans and have too inaccurate results with `0.1`. I added with (sample=0.75)

pkg/traceql/engine_metrics_test.go
ruslan-mikhailov: The test has too few spans to check correctly, but anyway, added this test case

pkg/traceql/ast_metrics.go
ruslan-mikhailov: Added a comment about it

ruslan-mikhailov:
\+ review fixes

pkg/traceql/parse_test.go
ruslan-mikhailov: Added support for `{ } | rate() > 10 | topk(5)` and minor refactoring for better support

mdisibio: This is so close to being generalized for N stages:

```
// ChainedSecondStage chains two second stage elements together.
// The first element is processed, then its output is passed to the second.
// Example: {status=error} | rate() | topk(5) > 10
type ChainedSecondStage struct {
	first  secondStageElement
	second secondStageElement
}
```

Encountered this somewhat quickly by trying to do a query with `rate() > X | topk(5) < Y`.   I do think the PR as-is could be confusing to users, and it is worth generalizing.   For example it is not clear why a comparison operator works with topk in the first query but not the second, without understanding the current parsing:
```
rate() | topk(5) < 100
rate() > 0 | topk(5) < 100
```

ruslan-mikhailov: Oh, I misunderstood you. Added, now it's even more clean

pkg/traceql/expr.y
mdisibio: I think starting with consistency is good.  If for some reason it's not fitting well, we could change it or support both.   Fwiw this operator is likely to not be used much.

knylander-grafana:
Do we need docs for this?

pkg/traceql/ast_metrics.go
```diff
@@ -421,3 +426,154 @@ func (m *TopKBottomK) process(input SeriesSet) SeriesSet {
...
+}
+
+func (m *MetricsFilter) init(_ *tempopb.QueryRangeRequest) {}
+
+func (m *MetricsFilter) process(input SeriesSet) SeriesSet {
+	result := make(SeriesSet, len(input))
+
+	for key, series := range input {
+		hasValue := false
+		newValues := make([]float64, len(series.Values))
```
mdisibio: For performance, I think it would be ok to edit the input series here.  Is there any case where the same series is reused elsewhere?

pkg/traceql/ast_metrics.go
```diff
@@ -421,3 +426,154 @@ func (m *TopKBottomK) process(input SeriesSet) SeriesSet {
...
+func (c ChainedSecondStage) String() string {
+	s := c[0].String()
+	for _, e := range c[1:] {
+		s += e.separator() + e.String()
+	}
+	return s
+}
+
+func (c ChainedSecondStage) separator() string {
+	return c[0].separator()
```
mdisibio: This isn't used, which makes me consider if separator() shouldn't be part of the interface of individual elements.  But part of the second stage, to be used in String() function above only.

i.e. instead of appending directly in expr.y with `append($1, $3)`, we could add a helper that appends with a given separator.

pkg/traceql/expr.y
```diff
@@ -122,13 +126,12 @@ import (
...
-  | scalarPipelineExpressionFilter              { yylex.(*lexer).expr = newRootExpr($1) } 
-  | spansetPipeline PIPE metricsAggregation     { yylex.(*lexer).expr = newRootExprWithMetrics($1, $3) }
-  // note: would only work for single metrics pipeline and not for multiple metrics pipelines before the fucntions
-  | spansetPipeline PIPE metricsAggregation PIPE metricsSecondStage  { yylex.(*lexer).expr = newRootExprWithMetricsTwoStage($1, $3, $5) }
-  | root hints                                  { yylex.(*lexer).expr.withHints($2) }
+    spansetPipeline                                                       { yylex.(*lexer).expr = newRootExpr($1) }
+  | spansetPipelineExpression                                             { yylex.(*lexer).expr = newRootExpr($1) }
+  | scalarPipelineExpressionFilter                                        { yylex.(*lexer).expr = newRootExpr($1) } 
+  | spansetPipeline PIPE metricsAggregation                               { yylex.(*lexer).expr = newRootExprWithMetrics($1, $3) }
+  | spansetPipeline PIPE metricsAggregation metricsSecondStagePipeline    { yylex.(*lexer).expr = newRootExprWithMetricsTwoStage($1, $3, secondStagePipeline($4)) }
```
mdisibio: This is fantastic 👍   One of these days we could rename metricsAggregation to metricsFirstStagePipeline and I think it would be very easy to understand.

pkg/traceql/test_examples.yaml
```diff
@@ -171,6 +171,42 @@ valid:
...
+  # metrics filter (comparison on metrics results)
+  - '{} | rate() > 10'
+  - '{} | rate() by (span.client_ip) >= 5.5'
+  - '{} | count_over_time() by (name) < 100'
+  - '{} | rate() by (span.client_ip) <= 0.5'
+  - '{} | rate() = 42'
+  - '{} | rate() != 0'
+  - '{} | rate() > -3'
+  - '{} | rate() by (span.client_ip) > 10 with(sample=0.1)'
+  # second stage combined
```
mdisibio, line 183: Do we need to add some examples to test the new N stages?  i.e. rate() > X | topk(1) < Y

mdisibio (commented):
This is looking great. 95% lgtm, just want to see if we can update a few small places.

pkg/traceql/test_examples.yaml
ruslan-mikhailov: There are plenty of test cases for 2 elements, and 1 for 3 elements. Added a bit more cases for N=3.

pkg/traceql/ast_metrics.go
ruslan-mikhailov: I think they should not? Calling Result twice with MetricsFilter will also have the same result due to IsNaN check.

Pushed in a separate commit

pkg/traceql/test_examples.yaml
mdisibio: Ah ok I see it now:  `- '{} | rate() > 0 | topk(5) <= 10'`, then it is good.

mdisibio (approved):
I'm loving this PR. BTW this works too `{} | rate() by (resource.service.name) > 20 < 60` 🚀  

LGTM

ruslan-mikhailov:
No-op push to retrigger CI pipeline

PR #6480, opened 2026-02-17
feat: log x-plugin-id header

javiermolinar:
Grafana plugings like the assistant injects the X-Plugin-ID header in its requests. Logging in we can measure the assistant usage.

### DIFF
--- a/modules/frontend/grpc_util.go
+++ b/modules/frontend/grpc_util.go
@@ -3,13 +3,16 @@ package frontend
 import (
 	"context"
 	"net/http"
+	"strings"
 
+	"github.com/grafana/tempo/pkg/api"
 	"google.golang.org/grpc/metadata"
 )
 
 var copyHeaders = []string{
 	"Authorization",
 	"X-Scope-OrgID",
+	api.HeaderPluginID,
 }
 
 func headersFromGrpcContext(ctx context.Context) (hs http.Header) {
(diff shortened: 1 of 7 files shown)


modules/querier/http.go
```diff
@@ -31,6 +31,16 @@ const (
 	QueryModeExternal  = "external"
 )
 
+func setRequestSpanAttributes(span oteltrace.Span, r *http.Request) {
```
electron0zero, line 34: nit: setPluginId would be a more appropriate name here?

modules/frontend/handler.go
```diff
@@ -135,6 +135,17 @@ func (f *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	level.Info(f.logger).Log(logMessage...)
 }
 
+func setRequestSpanAttributes(span trace.Span, orgID, pluginID string) {
+	if span == nil {
+		return
+	}
+
+	span.SetAttributes(attribute.String("orgID", orgID))
+	if pluginID != "" {
```
electron0zero, line 144: any reason to do this check? 

it should be fine if we set the attribute value to "", and it will tell us that this header value is not being sent from upstream?

modules/frontend/grpc_util.go
```diff
@@ -22,8 +25,8 @@ func headersFromGrpcContext(ctx context.Context) (hs http.Header) {
 	}
 
 	for _, h := range copyHeaders {
-		if v := md.Get(h); len(v) > 0 {
-			hs[h] = v
+		if v := md.Get(strings.ToLower(h)); len(v) > 0 {
```
electron0zero, line 28: I am not sure, why we need this strings.ToLower call here? also it would be nice to document why we do this.

PR #6481, opened 2026-02-17
fix: normalize allowlist headers when building the allowlist map

javiermolinar:
**What this PR does**:
The async_strip_headers_middleware was case-sensitive. Meaning that an allowlist value like x-plugin-id can fail against X-Plugin-Id.

In this PR we first normalize the headers using http.CanonicalHeaderKey when the allowlist is build and normalize again when matching them against the list.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/pipeline/async_strip_headers_middleware_test.go
+++ b/modules/frontend/pipeline/async_strip_headers_middleware_test.go
@@ -35,6 +35,18 @@ func TestStripHeaders(t *testing.T) {
 			headers:  map[string][]string{"X-Plugin-Id": {"grafana-assistant"}, "authorization": {"Bearer abc"}, "X-Drop-Me": {"nope"}},
 			expected: map[string][]string{"X-Plugin-Id": {"grafana-assistant"}, "authorization": {"Bearer abc"}},
 		},
+		{
+			name:     "allow list keeps uppercase non-canonical incoming header",
+			allow:    []string{"x-plugin-id", "AUTHORIZATION"},
+			headers:  map[string][]string{"X-PLUGIN-ID": {"grafana-assistant"}, "authorization": {"Bearer abc"}, "X-Drop-Me": {"nope"}},
+			expected: map[string][]string{"X-PLUGIN-ID": {"grafana-assistant"}, "authorization": {"Bearer abc"}},
+		},
+		{
+			name:     "allow list keeps lowercase non-canonical incoming header",
+			allow:    []string{"x-plugin-id", "AUTHORIZATION"},
+			headers:  map[string][]string{"x-plugin-id": {"grafana-assistant"}, "authorization": {"Bearer abc"}, "X-Drop-Me": {"nope"}},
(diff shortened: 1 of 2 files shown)


modules/frontend/pipeline/async_strip_headers_middleware.go
```diff
@@ -17,7 +19,7 @@ func NewStripHeadersWare(allowList []string) AsyncMiddleware[combiner.PipelineRe
 	// build allowed map
 	allowed := make(map[string]struct{}, len(allowList))
 	for _, header := range allowList {
```
electron0zero, line 21: nit: can we add a comment around why we called `CanonicalHeaderKey` here.

modules/frontend/pipeline/async_strip_headers_middleware.go
```diff
@@ -36,7 +38,7 @@ func (c stripHeadersWare) RoundTrip(req Request) (Responses[combiner.PipelineRes
 	} else {
 		// clear out headers not in allow list
 		for header := range httpReq.Header {
-			if _, ok := c.allowed[header]; !ok {
+			if _, ok := c.allowed[http.CanonicalHeaderKey(header)]; !ok {
```
electron0zero, line 41: nit: can also use a comment here?

modules/frontend/pipeline/async_strip_headers_middleware_test.go
```diff
@@ -29,6 +29,12 @@ func TestStripHeaders(t *testing.T) {
 			headers:  map[string][]string{"header1": {"value1"}, "header2": {"value2"}},
 			expected: map[string][]string{"header1": {"value1"}},
 		},
+		{
+			name:     "allow list is case-insensitive",
+			allow:    []string{"x-plugin-id", "AUTHORIZATION"},
+			headers:  map[string][]string{"X-Plugin-Id": {"grafana-assistant"}, "authorization": {"Bearer abc"}, "X-Drop-Me": {"nope"}},
```
electron0zero, line 35: can we add a test where incoming header is not in the CanonicalHeader form, and was allowed because of the `CanonicalHeaderKey` conversion we do in the RoundTrip.

basically same test as this but `"X-Plugin-Id": {"grafana-assistant"}` is changed to `"X-PLUGIN-ID": {"grafana-assistant"}`, and another case where it's `"x-plugin-id": {"grafana-assistant"}`.

`allow` config can be same across these cases.

modules/frontend/pipeline/async_strip_headers_middleware.go
javiermolinar: done

modules/frontend/pipeline/async_strip_headers_middleware.go
javiermolinar: done

modules/frontend/pipeline/async_strip_headers_middleware_test.go
javiermolinar: Good idea, new tests added

PR #6486, opened 2026-02-17
Proposal: [Enhancement] TraceQL Metrics: return NaN values

ruslan-mikhailov:
**What this PR does**: returns NaN values instead of skipping bucket if returned value is NaN

**How has it been tested:** the easiest way to check is to run in docker compose version of this branch on top of https://github.com/grafana/tempo/pull/6474 and request traceql math. Results:
image
image

Another example with avg_over_time(duration)
Before
image

After
image

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/engine_metrics.go
+++ b/pkg/traceql/engine_metrics.go
@@ -1486,8 +1486,17 @@ func NewSimpleCombiner(req *tempopb.QueryRangeRequest, op SimpleAggregationOp, e
 		initWithNaN = true
 	default:
 		// Simple addition aggregator. It adds existing values with the new sample.
-		f = func(existingValue float64, newValue float64) float64 { return existingValue + newValue }
-		initWithNaN = false
+		// NaN is used to represent "no data" and should be preserved if no actual data comes in.
+		f = func(existingValue float64, newValue float64) float64 {
+			if math.IsNaN(newValue) {
+				return existingValue // Skip NaN values during aggregation
+			}
+			if math.IsNaN(existingValue) {
+				return newValue // Replace initial NaN with actual data
+			}
+			return existingValue + newValue
+		}
+		initWithNaN = true
 
 	}
 	return &SimpleAggregator{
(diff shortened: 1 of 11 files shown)


pkg/traceql/engine_metrics_test.go
```diff
@@ -1935,14 +1937,14 @@ func TestMaxOverTimeForSpanAttribute(t *testing.T) {
...
 	ts := result.ToProto(req)
-	fooBarSamples := []tempopb.Sample{{TimestampMs: 1000, Value: 404}, {TimestampMs: 2000, Value: 403}}
-	fooBazSamples := []tempopb.Sample{{TimestampMs: 1000, Value: 204}, {TimestampMs: 3000, Value: 500}}
+	fooBarSamples := []tempopb.Sample{{TimestampMs: 1000, Value: 404}, {TimestampMs: 2000, Value: 403}, {TimestampMs: 3000, Value: math.NaN()}}
+	fooBazSamples := []tempopb.Sample{{TimestampMs: 1000, Value: 204}, {TimestampMs: 2000, Value: math.NaN()}, {TimestampMs: 3000, Value: 500}}
 
 	for _, s := range ts {
 		if LabelsFromProto(s.Labels).String() == "{\"span.foo\"=\"bar\"}" {
-			assert.Equal(t, fooBarSamples, s.Samples)
+			assert.True(t, cmp.Equal(fooBarSamples, s.Samples, cmpopts.EquateNaNs()), "expected %v, got %v", fooBarSamples, s.Samples)
```
ruslan-mikhailov, line 1945: `assert.Equal` cannot check NaN values, because math.NaN() != math.NaN(). I used cmp lib instead here.

pkg/traceql/engine_metrics.go
ruslan-mikhailov, line 346: Alternative approach is to pass "withNaN bool" to ToProto and set it true only for level 3 (query-frontend). It will be much safer, but clunky. Wdyt?

I did lots of changes to avoid "NaN+1=NaN" problem, but never can be sure I did not miss anything, and I personally would prefer safety over perfection

ruslan-mikhailov: PR with the alternative approach: https://github.com/grafana/tempo/pull/6500

ruslan-mikhailov:
It should be done in grafana/grafana, closing the proposal

PR #6509, opened 2026-02-19
Update live store to honor local and global block configs

mdisibio:
**What this PR does**:
The live store currently hardcodes the completeBlock encoding to DefaultEncoding(), which means it can be configured to use vParquet5.  This fixes it to honor the block config, but also fixes the fallbacks for it and the wal versions, to use the global storage.trace.block.version if nothing else overrides it.   

This also moves the encoding check up a level, to the LiveStore from the instance, so that an invalid version will ensure a startup error.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/livestore/instance.go
+++ b/modules/livestore/instance.go
@@ -115,19 +115,19 @@ func newInstance(instanceID string, cfg Config, wal *wal.WAL, completeBlockEncod
 	logger = log.With(logger, "tenant", instanceID)
 
 	i := &instance{
-		tenantID:               instanceID,
-		logger:                 logger,
-		Cfg:                    cfg,
-		wal:                    wal,
+		tenantID:              instanceID,
+		logger:                logger,
+		Cfg:                   cfg,
+		wal:                   wal,
 		completeBlockEncoding: completeBlockEncoding,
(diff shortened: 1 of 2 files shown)


modules/livestore/config.go
```diff
@@ -177,5 +180,49 @@ func (cfg *Config) Validate() error {
...
+	)
+
+	if cfg.GlobalBlockConfig != nil && cfg.GlobalBlockConfig.Version != "" {
+		blockVer = cfg.GlobalBlockConfig.Version
+		walVer = cfg.GlobalBlockConfig.Version
+	}
+
+	if cfg.BlockConfig.Version != "" {
+		blockVer = cfg.BlockConfig.Version
+		walVer = cfg.BlockConfig.Version
```
mapno, line 210:
```suggestion
```

mdisibio: Are you thinking to remove to this line?  This is the fallback where the live store WAL will use the main block version unless it is overridden in the WAL section, which I think we want.

mapno: Makes sense 👍

PR #6511, opened 2026-02-19
enhancement: Do not engage backpressure on WAL replay

oleg-kozlyuk-grafana:
**What this PR does**:

When live store starts with WAL files left over from previous launch, it adds them to completion queue, and delays ingestion of new records from Kafka, until open wal blocks is reduced to 1. This change attempts to circumvent the problem by subtracting the count of 'inherited' WAL blocks when calculating backpressure

Risks:

* This trades slower startup for slower searches for some time until all WAL blocks are replayed, as well as IO load

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/livestore/instance.go
+++ b/modules/livestore/instance.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"sort"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/go-kit/log"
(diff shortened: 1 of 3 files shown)


modules/livestore/instance.go
```diff
@@ -480,6 +483,10 @@ func (i *instance) completeBlock(ctx context.Context, id uuid.UUID) error {
 	}
 	delete(i.walBlocks, (uuid.UUID)(walBlock.BlockMeta().BlockID))
 
+	if newVal := i.replayedWALBlockCount.Add(-1); newVal < 0 {
```
javiermolinar, line 486: If there is an error the counter is not decremented. Is that expected?

oleg-kozlyuk-grafana: Maybe I am lacking Go language understanding here - how can a counter be not incremented if a previous line (`delete(i.walBlocks, (uuid.UUID)(walBlock.BlockMeta().BlockID))`) is executed?

PR #6124, opened 2026-01-05
Correct handle whitespace in tag value search

mapno:
**What this PR does**:

Whitespaces and invisible separators were not being correctly handled in the tag value search filtered queries, thus dropping those conditions. This PR adds the `\p{Z}` Unicode class to add support for those characters.

Contributes to https://github.com/grafana/tempo/issues/6144

**Checklist**
- [x] Tests updated
- [ ] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #6513, opened 2026-02-19
refactor: replace regex-based matcher extraction with token parser

### LINKED ISSUE #6144: Improve tag values filtered search brittleness

rohansood10:
## Description

Replaces the two fragile regular expressions (`matchersRegexp` and `singleFilterRegexp`) in `ExtractMatchers` with a token-based parser that handles TraceQL query strings more robustly.

### Problem

As documented in #6144, the regex-based approach for extracting matchers from autocomplete/filtered tag value queries has required a continuous stream of bug-fix PRs:
- Extend: #3846, #3458
- Fix: #3865, #3641, #3426, #3277, #3339, #6124

Each new edge case (whitespace in values, special characters, regex matchers, scoped intrinsics) required modifying complex regular expressions.

### Solution

Token-based parsing that:
- Parses attribute names, operators, and values at the character level
- Properly handles quoted strings with whitespace and special characters
- Splits on `&&` while respecting string boundaries
- Validates each matcher is complete (attribute + operator + value) before including it
- Rejects queries with OR conditions, multiple spanset filters, or pipe operators

### Testing

- All 24 existing `TestExtractMatchers` test cases pass unchanged
- Benchmark performance is comparable (microsecond-level per call)
- No regex dependencies remain in the file

### What this enables

New syntax support can be added by modifying token-level rules rather than complex regex patterns. The code is more readable and easier to reason about for edge cases.

Fixes #6144

### DIFF
--- a/pkg/traceql/extractmatcher.go
+++ b/pkg/traceql/extractmatcher.go
@@ -1,44 +1,22 @@
 package traceql
 
 import (
-	"regexp"
 	"strings"
+	"unicode"
 )
-
-// TODO: Support spaces, quotes
-//  See: https://github.com/grafana/grafana/issues/77394
-
-// Regex to extract matchers from a query string
-// This regular expression matches a string that contains three groups separated by operators.
-// The first group matches one or more Unicode letters, digits, underscores, or periods. It essentially matches variable names or identifiers
-// The second group is a comparison operator, which can be one of several possibilities, including =, >, <, and !=.
-// The third group is one of several possible values:
-//  1. A double-quoted string consisting of one or more Unicode characters, including letters, digits, punctuation, diacritical marks, and symbols,
-//  2. A sequence of one or more digits, which can represent numeric values, possibly with units like 's', 'm', or 'h'.
-//  3. The boolean values "true" or "false".
-//
-// Example: "http.status_code = 200" from the query "{ .http.status_code = 200 && .http.method = }"
-var matchersRegexp = regexp.MustCompile(`[\p{L}\p{N}._\-:" ]+\s*(=|<=|>=|=~|!=|>|<|!~)\s*(?:"[\p{L}\p{N}\p{P}\p{M}\p{S}\p{Z}]+"|true|false|[a-z]+|[0-9smh]+)`)
-
-// TODO: Merge into a single regular expression
-
-// Regex to extract selectors from a query string
-// This regular expression matches a string that contains a single spanset filter and no OR `||` conditions.
-// Examples
-//
-//	Query                                    |  Match
-//
-// { .bar = "foo" }                          |   Yes
-// { .bar =~ "foo|bar" }                     |   Yes
-// { .bar = "foo" && .foo = "bar" }          |   Yes
-// { .bar = "foo" || .foo = "bar" }          |   No
-// { .bar = "foo" } && { .foo = "bar" }      |   No
-// { .bar = "foo" } || { .foo = "bar" }      |   No
-var singleFilterRegexp = regexp.MustCompile(`^(\{[^|{}]*[^|{}]}?|\{[^|{}]*=~[^{}]*})$`)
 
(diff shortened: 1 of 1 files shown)


rohansood10:
Friendly ping — any chance to review this when you get a moment? Happy to make changes if needed.

mapno:
Hey @rohansood10, thanks for putting this together — I really appreciate you picking up #6144 and the time you spent on it.

We want to be upfront: we also have an open PR for this (#6502) that we'd been working on in parallel. That's entirely on us for not assigning the issue — I'm very sorry about that collision.

Some context on why we're currently preferring the approach in #6502: a hand-rolled parser ends up duplicating a lot of TraceQL syntax knowledge — operators, value types, status/kind enums, etc — that already exists in the lexer and parser. When TraceQL changes, those two copies get out of sync, which is kind of the same problem we had with the regex. The lenient parser in #6502 reuses the existing lexer, so new syntax works automatically. It also changes the engine to take `[]Condition` instead of a string, which gets rid of the re-parse and fallback logic in `engine.go`.

That said, your approach has some real strengths, and has made us re-think the approach — it more well-contained, didn't touch callers, and all existing tests passed as-is.

We don't want this experience to be discouraging. If you're interested in continuing to contribute to Tempo, we'd welcome that. Feel free to browse the open issues — and if something catches your eye, drop a comment so we can assign it to avoid this happening again.

Thanks again for the contribution, and sorry for the overlap.

rohansood10:
Hi maintainers! This PR has been idle for a few days. Could someone please take a look when you have a chance? Thank you!

rohansood10:
Hi @mapno, thanks for the thoughtful response and explanation! No worries about the overlap — I completely understand, and I appreciate you taking the time to explain the reasoning behind preferring #6502.

The lenient parser approach using the existing lexer definitely makes sense from a maintenance perspective. I learned a lot working on this, and I'm glad it helped validate some aspects of the solution.

I'll close this PR and look for other issues to contribute to. Thanks for the welcoming response — I'll definitely browse the open issues and comment before starting work next time!

Best regards and thanks again for maintaining such a great project.

PR #6515, opened 2026-02-19
enhancement: use frontend MaxExemplars config as single source of truth for exemplar limits

### LINKED ISSUE #5166: Use query_frontend.metrics.max_exemplars instead of hardcoded limit

zhxiaogg:
**What this PR does**:

Removed a hardcoded `maxExemplars = 100` from `engine_metrics.go`. Instead use the configurable limit from frontend for metrics range queries.

Without the fix, the clients could never receive more than 100 exemplars regardless of server configuration.

**Test:**
- Set `MaxExemplars = 300` in frontend configuration
- Verified previous logic returns ~90 exemplars while the new logic returns ~270 exemplars, see the following screenshots for details.

Before:
Screenshot 2026-02-19 at 8 46 04 AM

After:
Screenshot 2026-02-19 at 8 50 29 AM

Fixes #5166 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/traceql/util.go
+++ b/pkg/traceql/util.go
@@ -93,7 +93,7 @@ func newBucketSet(exemplars uint32, start, end uint64) *limitedBucketSet {
 	end /= uint64(time.Millisecond.Nanoseconds())   //nolint: gosec // G115
 
 	interval := end - start
-	bucketWidth := interval / uint64(buckets)
+	bucketWidth := max(interval/uint64(buckets), 1)
 
 	return &limitedBucketSet{
 		sz:          int(buckets),
--- a/pkg/traceql/util_test.go
+++ b/pkg/traceql/util_test.go
@@ -116,6 +116,16 @@ func TestBucketSetSingleExemplar(t *testing.T) {
 	assert.True(t, s.addAndTest(tsMilli), "ts=%d should not be added to bucket", 100)
 }
 
+func TestBucketSetLargeExemplarsShortRange(t *testing.T) {
+	// exemplars=10000 → buckets=5000, but the range is only 1 second (1000ms interval).
+	// Without the guard, bucketWidth=0 causes a divide-by-zero in bucket().
+	s := newBucketSet(10000, 0, uint64(time.Second.Nanoseconds())) //nolint: gosec // G115
+	assert.NotPanics(t, func() {
+		s.addAndTest(500) // 500ms into the range
+	})
+	assert.False(t, s.testTotal(), "should not be full after one exemplar")
+}
+
 func TestBucketSetZeroExemplars(t *testing.T) {
 	// exemplars=0 means collection is disabled: testTotal() should always return true
 	// and no exemplars should ever be accepted.


modules/frontend/metrics_query_range_handler.go
```diff
@@ -148,6 +156,14 @@ func newMetricsQueryRangeHTTPHandler(cfg Config, next pipeline.AsyncRoundTripper
 		}
 		req = api.BuildQueryRangeRequest(req, queryRangeReq, "")
 
+		// Normalize exemplars before combiner creation: 0 (unspecified) or above the configured
```
javiermolinar: We can refactor this

javiermolinar: Maybe we can change the type to uint32 here
https://github.com/grafana/tempo/blob/9a86b201376f68606efaa0fe8f9cca10cc39c349/modules/frontend/metrics_query_range_sharder.go#L51

It will catch negative numbers unmarshalling

modules/frontend/metrics_query_range_handler.go
```diff
@@ -75,6 +75,14 @@ func newQueryRangeStreamingGRPCHandler(cfg Config, next pipeline.AsyncRoundTripp
 		tenant, _ := user.ExtractOrgID(ctx)
 		start := time.Now()
 
+		// Normalize exemplars before combiner creation: 0 (unspecified) or above the configured
+		// maximum defaults to MaxExemplars. The sharder applies the same normalization to sub-requests,
+		// but runs after the combiner is created, so we must mirror it here.
+		if cfg.Metrics.Sharder.MaxExemplars > 0 {
```
javiermolinar: A very similar code is in the sharder too as you noted.
https://github.com/grafana/tempo/blob/9a86b201376f68606efaa0fe8f9cca10cc39c349/modules/frontend/metrics_query_range_sharder.go#L121C2-L125C4

If we clamp before the  api.BuildQueryRangeRequest the proper clamping will be propagated downstream and we won't need to do it again in the sharder

pkg/traceql/engine_metrics_average.go
```diff
@@ -581,8 +583,8 @@ func (g *avgOverTimeSpanAggregator[F, S]) getSeries(span Span) avgOverTimeSeries
 		intervals := g.intervalMapper.IntervalCount()
 		s = avgOverTimeSeries[S]{
 			vals:            g.buf.vals,
-			average:         newAverageSeries(intervals, maxExemplars, nil),
-			exemplarBuckets: newExemplarBucketSet(maxExemplars, g.start, g.end, g.step, g.instant),
+			average:         newAverageSeries(intervals, int(g.exemplars), nil),
```
javiermolinar: Should we use uint32 everywhere?

pkg/traceql/util.go
```diff
@@ -83,9 +83,6 @@ type limitedBucketSet struct {
 // newBucketSet creates a new bucket set for the given time range
 // start and end are in nanoseconds
 func newBucketSet(exemplars uint32, start, end uint64) *limitedBucketSet {
-	if exemplars > maxExemplars || exemplars == 0 {
```
javiermolinar, line 86: Removing this can be dangerous. This code can be reached from the queriers without passing by the clamping code in the frontend.

zhxiaogg: Previously I assume all requests sent to queriers are from frontend. What else requests are you referring to here?

My thinking is to cap the exemplars at the earliest stage, e.g. frontend, before it being used by downstream codes.

modules/frontend/metrics_query_range_handler.go
zhxiaogg: Yeah, that's a much cleaner way, thanks!

modules/frontend/metrics_query_range_handler.go
zhxiaogg: updated accordingly, thanks!

pkg/traceql/util.go
zhxiaogg: Discussed in person. There are queries sent to querier directly.

Updated the PR to cap the `req.Exemplars` at the entrance of the traceQL engine, which are `CompileMetricsQueryRange` and `CompileMetricsQueryRangeNonRaw`.

pkg/traceql/util.go
javiermolinar: This can lead to a division by zero panic now.

For a very small interval and a big number of buckets bucketWidth can be zero.

Later on in the bucket() method we use bucketWidth as the division denominator

Example:

interval = 1000ms
buckets = 1500
bucketWidth = 1000/1500 = 0 since we are using integers.

We need to handle that case

pkg/traceql/engine_metrics.go
javiermolinar, line 1017: This is overwritting the maxExemplars cap when the hint is used

zhxiaogg: This is a separate known issue, will fix in a following up PR if that's possible.

pkg/traceql/util.go
zhxiaogg: Wow, thanks for catching this!

javiermolinar (approved):
LGTM!

PR #6532, opened 2026-02-21
fix: block builder honor the global storage block config for block and WAL versions

### LINKED ISSUE #6451: Tempo ignores block config in storage section
**Describe the bug**
Tempo ignores the [block config](https://grafana.com/docs/tempo/latest/configuration/#block) in the [storage section](https://grafana.com/docs/tempo/latest/configuration/#storage-block-configuration-example) of the tempo configuration file.

This configuration should change the encoding from vParquet4 to vParquet5, but does not:

```yaml
storage:
  trace:
    block:
      version: vParquet5
```

When the block section is added to the block builder like this:

```yaml
block_builder:
  block:
    version: vParquet5
    parquet_dedicated_columns:
      - scope: resource
        name: k8s.cluster.name
        type: string
        options: [ ]
```

It respects the block version and the block builder builds vParquet5 blocks, but the resulting blocks don't have dedicated columns.

**To Reproduce**
1. Compile docker image from `main` using `make docker-tempo`
2. In `example/docker-compose/single-binary` change the storage section of the `tempo.yaml` such that the beginning of the section looks like this
```yaml
storage:
  trace:
    block:
      version: vParquet5
      ...
```
3. Run the single-binary example with docker-compose
4. Navigate to the minio console and check the version in the `meta.json` of created blocks
5. Version is vParquet4

**Expected behavior**
Expected block version is vParquet5

**Environment:**
 - Infrastructure: laptop
 - Deployment tool: docker-compose

Harry-kp:
**What this PR does**:
The block builder was hardcoding the default encoding version in `BlockConfig.RegisterFlagsAndApplyDefaults`, ignoring the global `storage.trace.block.version` config. This replicates the approach from #6509 (live store fix) for the block builder:

- Added `GlobalBlockConfig *common.BlockConfig` to `Config`, injected by the application at startup
- Created `coalesceBlockVersion()` that resolves the encoding with proper fallback: default → global `storage.trace.block` → local `block_config`
- Removed hardcoded `encoding.DefaultEncoding().Version()` from `BlockConfig.RegisterFlagsAndApplyDefaults`
- Updated `Validate()` to use `coalesceBlockVersion()` for early startup error on invalid version
- WAL version always follows the resolved block version

Fixes #6451

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/blockbuilder/config.go
+++ b/modules/blockbuilder/config.go
@@ -108,28 +108,22 @@ func (c *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet) {
 	f.StringVar(&c.WAL.Filepath, prefix+".wal.path", "/var/tempo/block-builder/traces", "Path at which store WAL blocks.")
 }
 
-// coalesceBlockVersion resolves the block encoding from configs.
-// Starts with the default version and overrides as each layer is checked (global block config, then block_config).
-// The WAL version always follows the block version.
-// Returns an error if the resolved version isn't writable.
+// coalesceBlockVersion resolves the block encoding version using the shared
+// encoding.CoalesceVersion helper. Priority: default < storage.trace.block < block_builder.block.
+// The WAL version always follows the resolved block version.
 func coalesceBlockVersion(cfg *Config) (encoding.VersionedEncoding, error) {
-	ver := encoding.DefaultEncoding().Version()
-
-	if cfg.GlobalBlockConfig != nil && cfg.GlobalBlockConfig.Version != "" {
-		ver = cfg.GlobalBlockConfig.Version
+	globalVer := ""
+	if cfg.GlobalBlockConfig != nil {
+		globalVer = cfg.GlobalBlockConfig.Version
 	}
 
-	if cfg.BlockConfig.BlockCfg.Version != "" {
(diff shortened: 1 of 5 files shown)


modules/blockbuilder/config.go
```diff
@@ -106,10 +105,35 @@ func (c *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet) {
...
 	c.WAL.RegisterFlags(f)
-	c.WAL.Version = c.BlockConfig.BlockCfg.Version
 	f.StringVar(&c.WAL.Filepath, prefix+".wal.path", "/var/tempo/block-builder/traces", "Path at which store WAL blocks.")
 }
 
+// coalesceBlockVersion resolves the block encoding from configs.
+// Starts with the default version and overrides as each layer is checked (global block config, then block_config).
+// The WAL version always follows the block version.
+// Returns an error if the resolved version isn't writable.
+func coalesceBlockVersion(cfg *Config) (encoding.VersionedEncoding, error) {
```
javiermolinar, line 114: Can we extract this into a shared helper? So we can use it in the LiveStore as well keeping module-specific WAL handling

PR #6539, opened 2026-02-23
enhancement: supported deduping spans within block builder

### LINKED ISSUE #6516: Need to deduplicate spans in the block builder
**Describe the bug**
Due to network, software, or config issues, it's possible for Tempo to receive the same data multiple times.  This is handled by deduplication in several places including search, trace by ID lookup, and compaction.  However a large gap is traceql metrics queries, which do not dedupe and expect storage to be clean.

We can add dedupe to block-builder and improve things.  

I would add it in [func (i *liveTracesIter) iter(ctx context.Context)](https://github.com/grafana/tempo/blob/main/modules/blockbuilder/live_traces_iter.go#L113)  That is the point at which we unmarshal the proto and reconstruct the full trace, before writing to the block.

Let's also add a metric similar to the compactor for [deduped spans](https://github.com/grafana/tempo/blob/main/tempodb/compactor.go#L66), which would be good for troubleshooting and monitoring the effectiveness of this dedupe over time.

zhxiaogg:
**What this PR does**:
- Block builder: deduplicate spans within traces during block creation and track removed duplicates via `tempo_block_builder_spans_deduped_total` metric

Fixes #6516

**Test**
- can verify the metrics are generated and showing in prometheus ui when testing with singlebinary example.

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/blockbuilder/live_traces_iter.go
+++ b/modules/blockbuilder/live_traces_iter.go
@@ -3,6 +3,7 @@ package blockbuilder
 import (
 	"bytes"
 	"context"
+	"errors"
 	"slices"
 	"sync"
 
@@ -36,6 +37,7 @@ type liveTracesIter struct {
 	cancel       func()
 	start, end   uint64
 	dedupedSpans uint32
+	exhausted    bool
 }
 
 func newLiveTracesIter(liveTraces *livetraces.LiveTraces[[]byte]) *liveTracesIter {
@@ -157,6 +159,8 @@ func (i *liveTracesIter) iter(ctx context.Context) {
 			return
 		}
 	}
+
+	i.exhausted = true
 }
 
 // MinMaxTimestamps returns the earliest start, and latest end span timestamps,
@@ -170,12 +174,15 @@ func (i *liveTracesIter) MinMaxTimestamps() (uint64, uint64) {
 }
 
 // DedupedSpans returns the total number of duplicate spans that were removed
-// across all traces. The iterator must be exhausted before this can be accessed.
-func (i *liveTracesIter) DedupedSpans() uint32 {
+// across all traces. Returns an error if the iterator has not been fully exhausted.
+func (i *liveTracesIter) DedupedSpans() (uint32, error) {
 	i.mtx.Lock()
 	defer i.mtx.Unlock()
 
-	return i.dedupedSpans
+	if !i.exhausted {
+		return 0, errors.New("iterator must be exhausted before calling DedupedSpans")
+	}
+	return i.dedupedSpans, nil
 }
 
 func (i *liveTracesIter) Close() {
--- a/modules/blockbuilder/live_traces_iter_test.go
+++ b/modules/blockbuilder/live_traces_iter_test.go
@@ -43,5 +43,28 @@ func TestLiveTracesIter_DedupSpans(t *testing.T) {
 		}
 	}
 	require.Equal(t, spanCount, total)
-	require.Equal(t, uint32(spanCount), iter.DedupedSpans())
+	n, err := iter.DedupedSpans()
+	require.NoError(t, err)
+	require.Equal(t, uint32(spanCount), n)
+}
+
+func TestLiveTracesIter_DedupedSpans_ErrorWhenNotExhausted(t *testing.T) {
+	// Push 11 traces so iter() produces 2 chunks (chunk size is 10).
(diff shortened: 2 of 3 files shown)


modules/blockbuilder/live_traces_iter.go
```diff
@@ -151,8 +156,40 @@ func (i *liveTracesIter) MinMaxTimestamps() (uint64, uint64) {
...
+// dedupeTrace removes duplicate spans in-place from tr, deduplicating by span ID and kind.
+// Returns the number of removed duplicate spans.
+func dedupeTrace(tr *tempopb.Trace) uint32 {
+	seen := make(map[uint64]struct{})
+	var deduped uint32
+	for _, rs := range tr.ResourceSpans {
+		for _, ss := range rs.ScopeSpans {
+			unique := ss.Spans[:0]
+			for _, s := range ss.Spans {
+				token := util.SpanIDAndKindToToken(s.SpanId, int(s.Kind))
```
javiermolinar: I wonder if we should use this hashing algorithm instead:

https://github.com/grafana/tempo/blob/2ad734b16785d8be4a52c319265a2e523a1c05c8/pkg/model/trace/combine.go#L27

It seems more correct, with less chances of collisions

modules/blockbuilder/live_traces_iter.go
```diff
@@ -151,8 +156,40 @@ func (i *liveTracesIter) MinMaxTimestamps() (uint64, uint64) {
...
 }
 
+// dedupeTrace removes duplicate spans in-place from tr, deduplicating by span ID and kind.
+// Returns the number of removed duplicate spans.
+func dedupeTrace(tr *tempopb.Trace) uint32 {
+	seen := make(map[uint64]struct{})
+	var deduped uint32
+	for _, rs := range tr.ResourceSpans {
+		for _, ss := range rs.ScopeSpans {
+			unique := ss.Spans[:0]
```
stoewer: Nice in-place dedup 👍

modules/blockbuilder/live_traces_iter.go
```diff
@@ -151,8 +156,40 @@ func (i *liveTracesIter) MinMaxTimestamps() (uint64, uint64) {
...
+}
+
 func (i *liveTracesIter) Close() {
 	i.cancel()
 }
 
+// dedupeTrace removes duplicate spans in-place from tr, deduplicating by span ID and kind.
+// Returns the number of removed duplicate spans.
+func dedupeTrace(tr *tempopb.Trace) uint32 {
+	seen := make(map[uint64]struct{})
```
stoewer: There is an opportunity here to save lots of smaller heap allocations:
* Initialize `seen` with a reasonable start size in to avoid allocations and rehashing when the map grows
* Reuse the `seen` in multiple deduplications (and use `clear(seen)` before reuse)

modules/blockbuilder/live_traces_iter.go
```diff
@@ -109,6 +111,9 @@ func (i *liveTracesIter) iter(ctx context.Context) {
 				}
 			}
 
+			// Deduplicate spans within the trace
+			i.dedupedSpans += dedupeTrace(tr)
```
stoewer: Each trace is traversed twice: one time in `dedupeTrace()` and another time in L118 to update timestamp bounds. Maybe those can be combined

modules/blockbuilder/live_traces_iter.go
```diff
@@ -151,8 +156,40 @@ func (i *liveTracesIter) MinMaxTimestamps() (uint64, uint64) {
 	return i.start, i.end
 }
 
+// DedupedSpans returns the total number of duplicate spans that were removed
+// across all traces. The iterator must be exhausted before this can be accessed.
```
stoewer: > The iterator must be exhausted before this can be accessed

Just for the sake of being a bit more defensive, what do you think about enforcing this by returning an error when `liveTracesIter.DedupedSpans()` is called before it's exhausted?

modules/blockbuilder/live_traces_iter.go
zhxiaogg: thanks, changed to use the alternative one you suggested!

modules/blockbuilder/live_traces_iter.go
zhxiaogg: thanks, all changed including the following comments.

PR #6559, opened 2026-02-25
bugfix: apply exemplars hint end-to-end and fix safety cap bypass

zhxiaogg:
**What this PR does**:

Two bugs in exemplars hint handling (`with(exemplars=true|false|N)`):

1. Safety cap bypass: hint value in CompileMetricsQueryRange was not checked against maxExemplars, allowing users to exceed the max limit.

2. Hint only affected per-shard collection: the frontend combiner always used the config-level MaxExemplars, so the hint never bounded the total exemplars in the final result.

Fix by moving hint resolution into normalizeRequestExemplars at the frontend handler, where req.Exemplars is set once and flows through the sharder, combiner, and backend uniformly.

Fixes CVE-2026-27878

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/frontend/metrics_query_range_handler_test.go
+++ b/modules/frontend/metrics_query_range_handler_test.go
@@ -537,10 +537,10 @@ func TestQueryRangeHandlerExemplarNormalization(t *testing.T) {
 		},
 	}
 
-	makeRequest := func(exemplars uint32) *http.Request {
+	makeRequestWithQuery := func(query string, exemplars uint32) *http.Request {
 		httpReq := httptest.NewRequest("GET", api.PathMetricsQueryRange, nil)
 		httpReq = api.BuildQueryRangeRequest(httpReq, &tempopb.QueryRangeRequest{
-			Query:     "{} | rate()",
+			Query:     query,
 			Start:     start,
 			End:       end,
 			Step:      step,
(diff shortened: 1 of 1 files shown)


modules/frontend/metrics_query_range_handler.go
```diff
@@ -139,10 +137,8 @@ func newMetricsQueryRangeHTTPHandler(cfg Config, next pipeline.AsyncRoundTripper
 			return httpInvalidRequest(err), nil
 		}
 
-		// Normalize exemplars before combiner creation: 0 (unspecified) or above the configured
-		// maximum defaults to MaxExemplars.
-		if queryRangeReq.Exemplars == 0 || queryRangeReq.Exemplars > cfg.Metrics.Sharder.MaxExemplars {
-			queryRangeReq.Exemplars = cfg.Metrics.Sharder.MaxExemplars
+		if err := normalizeRequestExemplars(queryRangeReq, cfg.Metrics.Sharder.MaxExemplars); err != nil {
+			return httpInvalidRequest(err), nil
```
oleg-kozlyuk-grafana, line 141: Worth testing the exception cases, as per annotations

mdisibio (approved):

PR #6568, opened 2026-02-26
prevent active-series underflow in local series limiter

carles-grafana:
`tempo_metrics_generator_registry_active_series` could wrap due to uint32 underflow
  when stale cleanup deletes overflow series that were never counted as active.

  - ignore `OnDelete` for overflow entity hash
  - clamp delete count to current active-series value before subtracting
  - add regression test for accepted+overflow stale-delete sequence

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@
 * [BUGFIX] fix: block builder honor the global storage block config for block and WAL versions [#6451](https://github.com/grafana/tempo/issues/6451) (@Harry-kp)
 * [BUGFIX] fix: normalize allowlist headers when building the allowlist map [#6481](https://github.com/grafana/tempo/pull/6481) (@javiermolinar)
 * [BUGFIX] fix: compactor deduped spans metric uses wrong type (gauge instead of counter) [#6558](https://github.com/grafana/tempo/issues/6558) (@bejaratommy)
+* [BUGFIX] metrics-generator: Fix active-series counter underflow in local series limiter when overflow series are deleted [#6568](https://github.com/grafana/tempo/pull/6568) (@carles-grafana)
 
 ### 3.0 Cleanup
 


knylander-grafana:
Do we need a changelog entry for this?

modules/generator/localserieslimiter/local_series_limiter.go
```diff
@@ -114,7 +114,19 @@ func (l *LocalSeriesLimiter) OnUpdate(uint64, uint32) {
...
-func (l *LocalSeriesLimiter) OnDelete(_ uint64, seriesCount uint32) {
+func (l *LocalSeriesLimiter) OnDelete(hash uint64, seriesCount uint32) {
+	// Overflow series are never counted in activeSeries, so deleting them must
+	// not decrement the active series counter.
+	if hash == l.overflowEntityHash {
+		return
+	}
+
+	activeSeries := l.activeSeries.Load()
+	// Clamp delete count to avoid uint32 underflow if we receive a mismatched delete.
```
oleg-kozlyuk-grafana, line 125: To be quite honest, I don't think this solves the issue. This basically means that a "mismatched" delete will render the active series counter completely unreliable. Is there a way to solve this by getting accurate counters of active series?

carles-grafana: after the overflow series fix, the count should be accurate. This is a safeguard for future bugs that could cause an underflow

carles-grafana: what solves the issue is the overflow check above

oleg-kozlyuk-grafana: If this is purely defensive, fair enough.

PR #6575, opened 2026-02-26
[metrics-generator]: configure partition reassignment behaviour on shutdown

zalegrala:
**What this PR does**:

Improves partition handoff between metrics-generator pods during rollouts, particularly when migrating from StatefulSet (stable pod names) to Deployment (random pod names).

With static membership (`instance_id`), franz-go intentionally skips the automatic LeaveGroup on `Close()` so that a pod restarting with the same InstanceID can rejoin without waiting for session timeout. For Deployments, pod names change on every restart, so the departing pod's static member slot holds the partition for the full session timeout (~3 min) before the coordinator reassigns it.

- Adds `leave_consumer_group_on_shutdown` (default `false`): when enabled, sends an explicit LeaveGroup on shutdown so the coordinator can rebalance immediately. Enable for Deployment rollouts; leave disabled for StatefulSet deployments where pods restart with the same InstanceID to avoid an unnecessary leave+rejoin rebalance.
- Fixes a cooperative rebalancing bug in `handlePartitionsAssigned`: the callback was replacing the full assignment on each call. In cooperative (incremental) rebalancing, the callback fires with only the *delta* — stable partitions are not re-reported and would be silently lost on any incremental rebalance.
- Adds `tempo_metrics_generator_assigned_partitions` gauge (set on every assign/revoke). A value of 0 for an extended period indicates an idle pod (more replicas than partitions) or a stuck rebalance — the right signal for pod-deletion-cost controllers and scale-down automation.
- Adds `deployment_max_unavailable` Jsonnet knob (default `1`) for controlling how many pods roll simultaneously.
- Adds `TestPartitionHandoff_LeaveGroupTriggersImmediateReassignment`, an end-to-end test using kfake (which accurately models session timeouts). The positive subtest confirms partition transfer completes quickly with LeaveGroup; the negative subtest confirms it does NOT transfer within the same window without LeaveGroup.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


oleg-kozlyuk-grafana:
@codex review this as well

operations/jsonnet/microservices/config.libsonnet
```diff
@@ -82,6 +82,11 @@
       ephemeral_storage_request_size: error 'Must specify a generator ephemeral_storage_request size',
       ephemeral_storage_limit_size: error 'Must specify a metrics generator ephemeral_storage_limit size',
       replicas: 0,
+      // deployment_max_unavailable: max unavailable during rolling update (default 1).
+      // Accepts an integer or a percentage string, e.g. deployment_max_unavailable: '25%'.
+      // Rolling (1): 2 rebalances per pod. All-at-once (replicas): fewer total rebalances;
+      // first new pod briefly holds all partitions until others rejoin.
+      deployment_max_unavailable: 1,
```
oleg-kozlyuk-grafana, line 89: That seems to be a harsh reduction, especially for big deployments with 50-100 replicas

modules/generator/generator_kafka.go
```diff
@@ -22,6 +22,16 @@ var metricEnqueueTime = promauto.NewCounter(prometheus.CounterOpts{
...
 
+// metricAssignedPartitions tracks the number of Kafka partitions currently assigned to this
+// generator instance. A value of 0 for an extended period indicates a misconfiguration
+// (e.g. more generator replicas than topic partitions, or a stuck rebalance).
+var metricAssignedPartitions = promauto.NewGauge(prometheus.GaugeOpts{
+	Namespace: "tempo",
+	Subsystem: "metrics_generator",
+	Name:      "assigned_partitions",
+	Help:      "Number of Kafka partitions currently assigned to this generator instance.",
+})
```
oleg-kozlyuk-grafana, line 34: nice idea

modules/generator/generator_kafka.go
```diff
@@ -43,6 +53,25 @@ func (g *Generator) stopKafka() {
...
 	close(g.kafkaCh)
+	// When enabled, with static membership (InstanceID) franz-go does not send
+	// LeaveGroup on Close(); explicitly leave by instance ID so the coordinator
+	// can rebalance immediately. When disabled, avoid two rebalances (leave then
+	// join) e.g. when all replicas go down then up together.
+	if g.cfg.LeaveConsumerGroupOnShutdown && g.cfg.InstanceID != "" {
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		fn := g.leaveGroupFn
+		if fn == nil {
```
oleg-kozlyuk-grafana: Not a fan of "start" function changing "static" state. Ideally fn should be assigned in constructor, and startKafka should just call the function if != nil

operations/jsonnet/microservices/config.libsonnet
zalegrala: This is only the default in jsonnet, which we can override.

zalegrala: Also note that we're copying the value from what is there currently, but hardcoded and less easy to override.

modules/generator/generator_kafka.go
zalegrala: Good catch.  I'll move it to New().

pkg/ingest/metrics.go
```diff
@@ -139,10 +147,43 @@ func getGroupLag(ctx context.Context, admClient *kadm.Client, partitionClient *P
...
+		}
+		lag := int64(-1)
+		if tcommit != nil {
+			if pcommit, ok := tcommit[p]; ok && pcommit.Err == nil && pcommit.At >= 0 {
+				lag = pend.Offset - pcommit.At
+			}
+		}
+		if lag < 0 && tstart != nil {
+			if pstart, ok := tstart[p]; ok && pstart.Err == nil {
+				lag = pend.Offset - pstart.Offset
```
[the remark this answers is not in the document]
zalegrala: This looks correct, I'll push a fix.

zalegrala:
@codex please review the updates.

CHANGELOG.md
```diff
@@ -38,6 +38,11 @@
...
+* [ENHANCEMENT] metrics-generator: add `leave_consumer_group_on_shutdown` to send LeaveGroup on shutdown for immediate partition reassignment instead of waiting for session timeout [#6575](https://github.com/grafana/tempo/pull/6575) (@zalegrala)
+* [ENHANCEMENT] metrics-generator: improve Kafka partition lag reporting to use live group membership when available and correctly handle rebalancing state [#6575](https://github.com/grafana/tempo/pull/6575) (@zalegrala)
+* [BUGFIX] fix: compactor deduped spans metric uses wrong type (gauge instead of counter) [#6558](https://github.com/grafana/tempo/issues/6558) (@bejaratommy)
+* [ENHANCEMENT] metrics-generator: add `leave_consumer_group_on_shutdown` to send LeaveGroup on shutdown for immediate partition reassignment instead of waiting for session timeout [#6575](https://github.com/grafana/tempo/pull/6575) (@zalegrala)
+* [ENHANCEMENT] metrics-generator: improve Kafka partition lag reporting to use live group membership when available and correctly handle rebalancing state [#6575](https://github.com/grafana/tempo/pull/6575) (@zalegrala)
```
oleg-kozlyuk-grafana: I think the merge went wrong

pkg/ingest/metrics.go
```diff
@@ -119,7 +119,15 @@ func ResetLagMetricsForRevokedPartitions(group string, partitions []int32) {
 // the lag is the difference between the last produced offset and the offset committed in the consumer group.
 // Otherwise, if the block builder didn't commit an offset for a given partition yet (e.g. block builder is
 // running for the first time), then the lag is the difference between the last produced offset and fallbackOffsetMillis.
-func getGroupLag(ctx context.Context, admClient *kadm.Client, partitionClient *PartitionOffsetClient, group string, assignedPartitions []int32) (kadm.GroupLag, error) {
+//
+// When the group has current members (DescribeGroups), we use CalculateGroupLagWithStartOffsets so lag is
+// computed for each member's assignment. When the group has no members (e.g. rebalancing, or block-builder
+// which does not use consumer groups), we compute lag manually for (topic, assignedPartitions) so that
+// newly assigned partitions (e.g. after a rebalance) report correct lag.
+func getGroupLag(ctx context.Context, admClient *kadm.Client, partitionClient *PartitionOffsetClient, group, topic string, assignedPartitions []int32) (kadm.GroupLag, error) {
```
oleg-kozlyuk-grafana: Are you confident that this is covered by e2e tests, given annotations below? While the other functionality is hidden behind a flag, this change we won't be able to disable if something goes wrong.

knylander-grafana:
Do we need docs for this?

modules/generator/generator_kafka_test.go
```diff
@@ -0,0 +1,136 @@
...
+					InstanceID:                   tc.instanceID,
+				},
+				logger:      log.NewNopLogger(),
+				kafkaStop:   func() {},
+				kafkaCh:     make(chan *kgo.Record, 1),
+				kafkaClient: ingest.NewClientForTesting(kgoClient),
+				leaveGroupFn: func(_ context.Context) error {
+					leaveCalled = true
+					return nil
+				},
```
mapno, line 148: Why not use a control key in the kfake cluster instead of this function/field?

pkg/ingest/reader_client.go
```diff
@@ -110,12 +113,48 @@ func (c *Client) monitorPartitions() {
...
+// NewClientForTesting wraps a *kgo.Client in an ingest.Client without starting
+// the partition monitor goroutine. For use in unit tests only.
+func NewClientForTesting(client *kgo.Client) *Client {
+	return &Client{
+		Client: client,
+		stopCh: make(chan struct{}),
+		logger: log.NewNopLogger(),
+	}
+}
+
```
mapno: nit: Move to test file or pkg?

pkg/ingest/metrics.go
zalegrala: I pushed a commit to cover some of this method.

modules/generator/generator_kafka_test.go
zalegrala: It looks like this function the best to use the native group coordinator to simulate the handoff in a kafka environment.  The goal is to verify that a real LeaveGroup sent by a franz-go client causes the coordinator to trigger an immediate rebalance.  If we use a ControlKey, this would let us assert the wire format but would not exercise the coordinator rebalance, which is what we want to validate for this test.

pkg/ingest/reader_client.go
zalegrala: This has been moved to its own file, but can't be moved to _test because its imported elsewhere.

mattdurham:
Are all these intertwined, or could this be 4-6 smaller PRs?

zalegrala:
I think it could be two PRs if that would help the review.

operations/jsonnet/microservices/generator.libsonnet
```diff
@@ -45,6 +45,10 @@
     $.util.readinessProbe +
     (if $._config.variables_expansion then container.withArgsMixin(['-config.expand-env=true']) else {}),
 
+  // Deployment rollout strategy for Kafka ingest.
+  // Rolling (max_unavailable=1): each pod leave+rejoin = 2 rebalances per pod.
+  // All-at-once (max_unavailable=replicas): first pod briefly holds all partitions;
```
mattdurham: Is there a range between 1 and # replicas? What is that behavior?

pkg/ingest/reader_client.go
```diff
@@ -110,6 +113,47 @@ func (c *Client) monitorPartitions() {
 	}
 }
 
+// LeaveConsumerGroupByInstanceID sends a LeaveGroup request for the given
+// instance ID so the coordinator can rebalance without waiting for session
+// timeout. Use this on shutdown when using static membership (InstanceID):
+// franz-go does not send LeaveGroup on Close() when InstanceID is set.
+// Requires Kafka 2.4+. No-op if instanceID is empty.
```
mattdurham: What is the behavior on < 2.4?

pkg/ingest/reader_client.go
```diff
@@ -110,6 +113,47 @@ func (c *Client) monitorPartitions() {
...
+	req.Members = append(req.Members, member)
+	resp, err := req.RequestWith(ctx, client)
+	if err != nil {
+		return err
+	}
+	if err := kerr.ErrorForCode(resp.ErrorCode); err != nil {
+		return err
+	}
+	// v3+ responses carry per-member error codes; check those too.
+	var memberErr error
```
mattdurham: Is this kafka v3? If so does it really need kafka >= v3?

pkg/ingest/reader_client.go
zalegrala: The InstanceID field in LeaveGroupRequest was added in v3 of the protocol with KIP-345 in Kafka 2.4+. Franz-go negotiates the request version with the broker, so on a pre-2.4 broker the InstanceID field would be stripped and the leave would be sent without it  -- meaning the targeted-by-instance-ID semantics would be lost. To your comment below also, the per-member ErrorCode in LeaveGroupResponse was also added as part of `KIP-345` (v3+), so checking it does require a 2.4+ broker for the field to be populated.
**Do we have a version requirement for Tempo?**  I'm not seeing it in the existing docs.  2.4 was released in 2019.  My expectation is that this version requirement is acceptable.

pkg/ingest/reader_client.go
zalegrala: `KIP-345` mentioned above included the per-member ErrorCode, so I think yes its required.

operations/jsonnet/microservices/generator.libsonnet
zalegrala: Yes, its a range up to the number of replicas.  The tunable tunable to allow us to find the right balance between speed and disruption.  Lower values will stagger the rebalance for less disruption, which is less memory requirements moving around, allocating, etc.  If we used the replica count, if all but one pod is deleted, all partitions immediately move to the one remaining pod which will have implications for the pod resources.
Ideally we chose a value which migrates the partitions to new owners in less than the 3m timeout behavior which this PR is addressing.

pkg/ingest/metrics.go
zalegrala: The metrics have been split into a new PR: https://github.com/grafana/tempo/pull/6879

zalegrala: The metrics have been split into a new PR: https://github.com/grafana/tempo/pull/6879

pkg/ingest/reader_client.go
zalegrala: Just double checked, Kafka 2.4+ with the v3 wire protocol is our base requirement I'm seeing.

operations/jsonnet/microservices/generator.libsonnet
zalegrala: Yes -- any integer between 1 and replicas is valid. Updated the comment to explain the full range: each value N means N pods can be down simultaneously, causing N concurrent partition rebalances. The comment previously only described the two extremes.

pkg/ingest/reader_client.go
zalegrala: This uses LeaveGroup protocol version 3, which was introduced in Kafka 2.4 (KIP-345 / static membership) -- not Kafka broker 3.x. The requirement is Kafka 2.4+. Since static membership itself requires Kafka 2.4+, any cluster using instance_id is already on a compatible broker. Updated the docstring to spell this out.

pkg/ingest/reader_client.go
zalegrala: On Kafka < 2.4: franz-go auto-negotiates the LeaveGroup request version to the highest the broker supports. Below version 3, the InstanceID field is absent. With an empty MemberID and no InstanceID, the broker returns UNKNOWN_MEMBER_ID, we log a warning, and partitions fall back to the session timeout (~3 min) -- the same behaviour as if the flag were disabled. No incorrect leave occurs.

CHANGELOG.md
```diff
@@ -66,6 +66,7 @@
 * [ENHANCEMENT] Add metric to track livestore block cut reasons [#6922](https://github.com/grafana/tempo/pull/6922) (@zhxiaogg)
 * [ENHANCEMENT] Enable async parquet read mode for WAL completion path [#6967](https://github.com/grafana/tempo/pull/6967) (@zhxiaogg)
 * [BUGFIX] livestore: check readiness before lag for SearchRecent and QueryRange queries [#6911](https://github.com/grafana/tempo/pull/6911) (@zhxiaogg)
+* [ENHANCEMENT] metrics-generator: add `leave_consumer_group_on_shutdown` to send LeaveGroup on shutdown for immediate partition reassignment instead of waiting for session timeout [#6575](https://github.com/grafana/tempo/pull/6575) (@zalegrala)
 * [BUGFIX] Fix integer overflow in query parameters by using `strconv.ParseUint` instead of `strconv.Atoi`/`strconv.ParseInt` for unsigned integer fields. [#6612](https://github.com/grafana/tempo/pull/6612) (@bejaratommy)
 * [BUGFIX] Fix live-store SearchTagValuesV2 disk cache never being populated on complete blocks [#6858](https://github.com/grafana/tempo/pull/6858) (@mapno)
```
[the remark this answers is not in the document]
zalegrala: Fixed: moved the entry into the ENHANCEMENT block (before the BUGFIX section) and removed the duplicate plus the lag-reporting entry that was split to #6879.

modules/generator/generator_kafka_test.go
```diff
@@ -0,0 +1,306 @@
...
+	t.Run("with LeaveGroup", func(t *testing.T) {
+		fake, err := kfake.NewCluster(kfake.NumBrokers(1), kfake.SeedTopics(1, topic))
+		require.NoError(t, err)
+		t.Cleanup(fake.Close)
+		addr := fake.ListenAddrs()[0]
+
+		client1, gen1Assigned := newConsumer(t, addr, "gen-1")
+		ctx1, cancel1 := context.WithCancel(context.Background())
+		go poll(ctx1, client1)
+
```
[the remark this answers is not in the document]
zalegrala: Good catch. Added t.Cleanup(client1.Close) and t.Cleanup(cancel1) in both subtests right after creation. The explicit cancel1()/client1.Close() calls in the happy path are harmless (cancel on already-cancelled context and Close on already-closed client are both idempotent in kgo).

pkg/ingest/reader_client.go
```diff
@@ -110,12 +113,39 @@ func (c *Client) monitorPartitions() {
...
+// instance ID so the coordinator can rebalance without waiting for session
+// timeout. Use this on shutdown when using static membership (InstanceID):
+// franz-go does not send LeaveGroup on Close() when InstanceID is set.
+// Requires Kafka 2.4+. No-op if instanceID is empty.
+func LeaveConsumerGroupByInstanceID(ctx context.Context, client *kgo.Client, group, instanceID string, logger log.Logger) error {
+	if instanceID == "" {
+		return nil
+	}
+	req := kmsg.NewPtrLeaveGroupRequest()
+	req.Version = 4 // v3+ for Members[].InstanceID (KIP-345); using v4 as the minimum flexible version
```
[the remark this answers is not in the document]
zalegrala: Addressed: the req.Version field is not set in the current code. Franz-go negotiates the version automatically.

pkg/ingest/reader_client.go
```diff
@@ -110,12 +113,38 @@ func (c *Client) monitorPartitions() {
...
+	member := kmsg.NewLeaveGroupRequestMember()
+	member.InstanceID = &instanceID
+	req.Members = append(req.Members, member)
+	resp, err := req.RequestWith(ctx, client)
+	if err != nil {
+		return err
+	}
+	if err := kerr.ErrorForCode(resp.ErrorCode); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
zalegrala: Fixed: consumer_group.go now iterates resp.Members and returns the per-member ErrorCode for the requested instanceID (preferring the exact match), falling back to any member error.

modules/generator/generator_kafka.go
```diff
@@ -161,8 +183,12 @@ func (g *Generator) handlePartitionsAssigned(m map[string][]int32) {
 	g.partitionMtx.Lock()
 	defer g.partitionMtx.Unlock()
 
+	// In cooperative (incremental) rebalancing this callback fires with only
+	// newly added partitions; stable partitions are not re-reported. Append
+	// rather than replace so we don't lose partitions that weren't moved.
 	g.assignedPartitions = append(g.assignedPartitions, assigned...)
 	sort.Slice(g.assignedPartitions, func(i, j int) bool { return g.assignedPartitions[i] < g.assignedPartitions[j] })
+	metricAssignedPartitions.Set(float64(len(g.assignedPartitions)))
```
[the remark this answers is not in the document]
zalegrala: Fixed: getAssignedActivePartitions now returns slices.Clone(g.assignedPartitions) so the caller cannot observe concurrent mutations.

pkg/ingest/testutil.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+// NewClientForTesting wraps a *kgo.Client in an ingest.Client without starting
+// the partition monitor goroutine. For use in unit tests only.
+func NewClientForTesting(client *kgo.Client) *Client {
+	return &Client{
+		Client: client,
+		stopCh: make(chan struct{}),
+		logger: log.NewNopLogger(),
+	}
```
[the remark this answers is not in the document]
zalegrala: Addressed: the helper was moved to pkg/ingest/ingesttest, a dedicated test sub-package. The production ingest package surface is unchanged.

CHANGELOG.md
```diff
@@ -46,6 +46,8 @@
 * [ENHANCEMENT] Add new metric for generator ring size: `tempo_distributor_metrics_generator_tenant_ring_size` [#5686](https://github.com/grafana/tempo/pull/5686) (@zalegrala)
 * [ENHANCEMENT] Remove explicit `runtime.GC()` calls in vParquet5 compactor/block creation and CLI [#6603](https://github.com/grafana/tempo/pull/6603) (@oleg-kozlyuk-grafana)
 * [ENHANCEMENT] Implemented anti-affinity for pods in same livestore zone [#6757](https://github.com/grafana/tempo/pull/6757) (@zhxiaogg)
+* [ENHANCEMENT] metrics-generator: add `leave_consumer_group_on_shutdown` to send LeaveGroup on shutdown for immediate partition reassignment instead of waiting for session timeout [#6575](https://github.com/grafana/tempo/pull/6575) (@zalegrala)
+* [ENHANCEMENT] metrics-generator: improve Kafka partition lag reporting to use live group membership when available and correctly handle rebalancing state [#6575](https://github.com/grafana/tempo/pull/6575) (@zalegrala)
```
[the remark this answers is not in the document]
zalegrala: Fixed: duplicate entries removed.

modules/generator/generator_kafka.go
```diff
@@ -43,6 +54,18 @@ func (g *Generator) stopKafka() {
...
 	close(g.kafkaCh)
+	// When enabled, with static membership (InstanceID) franz-go does not send
+	// LeaveGroup on Close(); explicitly leave by instance ID so the coordinator
+	// can rebalance immediately. When disabled, avoid two rebalances (leave then
+	// join) e.g. when all replicas go down then up together.
+	if g.cfg.LeaveConsumerGroupOnShutdown && g.cfg.InstanceID != "" {
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		if err := g.leaveGroupFn(ctx); err != nil {
+			level.Warn(g.logger).Log("msg", "failed to leave Kafka consumer group by instance ID (partitions may reassign after session timeout)", "err", err)
```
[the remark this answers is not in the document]
zalegrala: Fixed: warning log now includes instance_id and group fields.

pkg/ingest/reader_client.go
```diff
@@ -110,12 +113,54 @@ func (c *Client) monitorPartitions() {
 	}
 }
 
+// LeaveConsumerGroupByInstanceID sends a LeaveGroup request for the given
+// instance ID so the coordinator can rebalance without waiting for session
+// timeout. Use this on shutdown when using static membership (InstanceID):
+// franz-go does not send LeaveGroup on Close() when InstanceID is set.
+// Requires Kafka 2.4+. No-op if instanceID is empty.
+func LeaveConsumerGroupByInstanceID(ctx context.Context, client *kgo.Client, group, instanceID string, logger log.Logger) error {
```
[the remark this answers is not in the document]
zalegrala: Done: LeaveConsumerGroupByInstanceID was moved to pkg/ingest/consumer_group.go, keeping reader_client.go closer to its upstream Loki fork.

pkg/ingest/metrics.go
```diff
@@ -139,10 +147,48 @@ func getGroupLag(ctx context.Context, admClient *kadm.Client, partitionClient *P
...
+	tstart := startOffsets[topic]
+	tend := endOffsets[topic]
+	if tend == nil {
+		return kadm.GroupLag{}, nil
+	}
+	result := make(kadm.GroupLag)
+	result[topic] = make(map[int32]kadm.GroupMemberLag)
+	for _, p := range assignedPartitions {
+		pend, ok := tend[p]
+		if !ok || pend.Err != nil {
```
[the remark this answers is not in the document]
zalegrala: The metrics.go lag improvements (including the stale-metric fallback fix) were split to #6879 to keep this PR focused on partition handoff. This PR no longer touches metrics.go.

modules/generator/generator.go
```diff
@@ -72,6 +72,11 @@ type Generator struct {
 	partitionRing      ring.PartitionRingReader
 	partitionMtx       sync.RWMutex
 	assignedPartitions []int32
+
+	// leaveGroupFn is called by stopKafka when LeaveConsumerGroupOnShutdown is
+	// true. If nil, defaults to ingest.LeaveConsumerGroupByInstanceID. Can be
+	// overridden in tests.
```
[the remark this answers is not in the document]
zalegrala: Fixed: comment updated to reflect that leaveGroupFn is always initialized in New(), not defaulted lazily.

pkg/ingest/metrics.go
```diff
@@ -119,7 +119,15 @@ func ResetLagMetricsForRevokedPartitions(group string, partitions []int32) {
 // the lag is the difference between the last produced offset and the offset committed in the consumer group.
 // Otherwise, if the block builder didn't commit an offset for a given partition yet (e.g. block builder is
 // running for the first time), then the lag is the difference between the last produced offset and fallbackOffsetMillis.
```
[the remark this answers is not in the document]
zalegrala: metrics.go changes were split to #6879. This PR no longer touches that file.

modules/generator/config.go
```diff
@@ -73,6 +73,13 @@ type Config struct {
 	Ingest            ingest.Config `yaml:"-"`
 	IngestConcurrency uint          `yaml:"ingest_concurrency"`
 	InstanceID        string        `yaml:"instance_id" doc:"default=<hostname>" category:"advanced"`
+
+	// LeaveConsumerGroupOnShutdown, when true, sends LeaveGroup to the Kafka coordinator
+	// on shutdown so partitions are reassigned immediately rather than waiting for session
+	// timeout (~3 min). Defaults to false so operators can control rollout behaviour
+	// explicitly; set to true when deploying as a Deployment (random pod names) to avoid
+	// the session-timeout delay on partition reassignment. Requires Kafka ingest enabled.
+	LeaveConsumerGroupOnShutdown bool `yaml:"leave_consumer_group_on_shutdown" category:"advanced"`
```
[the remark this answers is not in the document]
zalegrala: The PR description has been updated; both the description and the code default to false. The description now reads: 'Adds leave_consumer_group_on_shutdown (default false)'.

CHANGELOG.md
```diff
@@ -60,6 +60,8 @@
 * [BUGFIX] fix: correct block-builder fetch metrics to use counters instead of gauges.
 * [BUGFIX] Log tenant on receiver push errors [#6780](https://github.com/grafana/tempo/pull/6780) (@javiermolinar)
 * [BUGFIX] Fix race conditions in WAL block [#6773](https://github.com/grafana/tempo/pull/6773) (@ruslan-mikhailov)
+* [ENHANCEMENT] metrics-generator: add `leave_consumer_group_on_shutdown` to send LeaveGroup on shutdown for immediate partition reassignment instead of waiting for session timeout [#6575](https://github.com/grafana/tempo/pull/6575) (@zalegrala)
+* [ENHANCEMENT] metrics-generator: improve Kafka partition lag reporting to use live group membership when available and correctly handle rebalancing state [#6575](https://github.com/grafana/tempo/pull/6575) (@zalegrala)
```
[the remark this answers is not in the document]
zalegrala: Fixed: entry moved into the [ENHANCEMENT] block.

pkg/ingest/ingesttest/ingesttest.go
```diff
@@ -0,0 +1,15 @@
+// Package ingesttest provides test helpers for the ingest package.
+// It is intended for use in tests only; no production code should import it.
+package ingesttest
```
mattdurham, line 3: should the file name be ingest_test so its only compiled for tests?

pkg/ingest/consumer_group.go
```diff
@@ -0,0 +1,61 @@
...
+		return err
+	}
+	// v3+ responses carry per-member error codes; check those too.
+	var memberErr error
+	for _, m := range resp.Members {
+		if m.ErrorCode == 0 {
+			continue
+		}
+		if m.InstanceID != nil && *m.InstanceID == instanceID {
+			return kerr.ErrorForCode(m.ErrorCode)
```
mattdurham, line 50: I dont understand this if.

pkg/ingest/consumer_group.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+		if memberErr == nil {
+			memberErr = kerr.ErrorForCode(m.ErrorCode)
```
mattdurham, line 53: So we only pull the first memberErr we get? At this point is it safe to return instead of looping and returning the instanceid err code?

mattdurham (approved):
Small request on error handling

pkg/ingest/consumer_group.go
zalegrala: We're checking per-member error codes in case they exist, but if we find our self member has an error, we return here with the error.  We're here because we know that the error code is non zero above but we don't know which member.

pkg/ingest/consumer_group.go
zalegrala: No because we need to collect them all.  I think the results are non-ordered, and we want to prioritize our own error above the rest, but will return the last one we save if not our own in case our self member does not have an error.

pkg/ingest/ingesttest/ingesttest.go
zalegrala: Its used in modules/generator/generator_kafka_test.go and if it was in a _test.go file it would not be importable.

mattdurham (approved):
Lgtm

PR #6577, opened 2026-02-27
tempodb: rename deduped spans metric to counter

### LINKED ISSUE #6558: Incorrect metric types for the deduplicated spans in compactor
**Describe the bug**
The [compactor](https://github.com/grafana/tempo/blob/ad20e59db2b87b4b33c55b45445451d47530ead7/tempodb/compactor.go#L66) is using a gauge to count the deduped spans instead of a counter.

**To Reproduce**
N/A

**Expected behavior**
- the deduped spans should be a counter metric.

**Environment:**
 - all

**Additional Context**

- Probably good to scan all other metrics and make sure we're using the right types.
- Also, the fix might need to introduce(renaming) the existing metric, as prometheus is not supporting change metric type.

WinterCabbage:
**What this PR does**:

- Fixes the compactor deduped spans metric type by replacing a gauge with a counter.
- Renames metric `tempodb_compaction_spans_combined_total` (gauge) to `tempodb_compaction_deduped_spans_total` (counter), since Prometheus does not support changing metric type in-place.
- Adds `TestCompactionDedupedSpansMetric` to verify the deduped spans counter increases after compaction.
- Updates upgrade docs and changelog with the breaking metric rename and migration note (no dual-emission period).

Fixes #6558

**Checklist**
- [x] Tests updated
- [x] Documentation added


javiermolinar:
Thank you for your contribution but the issue has been fixed

PR #6580, opened 2026-02-27
fix(traceql): err on division by zero

### LINKED ISSUE #6579: Division by zero panics AST executor
**Describe the bug**

When executing an AST query of the likes of `{ 1 / 0 = 1 }`, the executor panics on the binary operation. Example in form of an HTTP request:

```
$ curl -sk https://tempo.localhost/api/search \
          -H "X-Scope-OrgID: 1" \
          --url-query "q={ 1 / 0 = 1 }"
invalid TraceQL query: parse error : runtime error: integer divide by zero
```

**To Reproduce**
1. Start Tempo (tested with [this image](https://hub.docker.com/layers/grafana/tempo/main-e15e38a/images/sha256-518be605335e018b0ef5c444bcb2cf08c46fb0d058b986579a682466f8d235bc))
2. Perform a query that executes this AST. See description for example.

**Expected behavior**
The request should be rejected OR the request should simply return no traces as no traces can possible fit this.

**Environment:**
 - Infrastructure: Docker
 - Deployment tool: Docker Compose

**Additional Context**

This finding was LLM-driven!

Proximyst:
Right now, Tempo will panic in the request goroutine if we use a query like `{ 1 / 0 = 1 }`. Instead, an error should be returned such that the query simply cannot return any results.

Note: the bug and fix were both LLM-driven.

Fixes #6579

**Checklist**
- [x] Tests updated
- [ ] Documentation added
  - I don't think we need any?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,10 +39,10 @@
 * [BUGFIX] fix: live store honor the config options for block and WAL versions [#6509](https://github.com/grafana/tempo/pull/6509) (@mdisibio)
 * [BUGFIX] fix: block builder honor the global storage block config for block and WAL versions [#6451](https://github.com/grafana/tempo/issues/6451) (@Harry-kp)
 * [BUGFIX] fix: normalize allowlist headers when building the allowlist map [#6481](https://github.com/grafana/tempo/pull/6481) (@javiermolinar)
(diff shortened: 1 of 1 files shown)


Proximyst:
@codex Please review the PR

PR #6585, opened 2026-02-27
[chore] restore the automated backport process

zhxiaogg:
**What this PR does**:

Restored the automated backport workflow using https://github.com/grafana/grafana-github-actions-go.

Right now the new backport workflow:
1. requires the "backport release-*" label on PRs, and will be triggered when a PR merged to main
2. can also be triggered via github action UI with a PR number as input

Fixes N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,45 +1,112 @@
-# This checks merged PRs for labels like "backport release-x.y"
-# and opens a new PR backporting the same commit to the release branch.
-# This workflow also runs when the PR is labeled or opened, but will
-# will only check a few things and detect that the PR is not yet merged.
-# At this time only squashed PRs are supported since the cherry-pick
-# command does not include "-m <n>" arg required for merge commits.
+# This triggers on every push to main and finds the PR that was merged.
+# If the PR has any "backport release-*" labels, it opens a backport PR
+# targeting each labeled release branch.
+#
+# Manual re-trigger: Go to Actions → Backport PR Creator → Run workflow,
+# provide the PR number. Note: running twice on the same PR will post a
+# failure comment on the original PR (no deduplication).
 name: Backport PR Creator
+
 on:
-  pull_request:
-    types:
-      - closed
-      - labeled
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: "PR number to backport (required for manual trigger)"
+        required: true
+        type: string
 
 permissions: {}
 
 jobs:
-  main:
-    # skip it in all forks, only run in grafana/tempo.
+  find-pr:
+    name: Find PR and check backport labels
+    # skip in forks — only run in grafana/tempo
     if: github.repository == 'grafana/tempo'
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -1,45 +1,112 @@
...
+          GH_TOKEN: ${{ github.token }}
+          EVENT_NAME: ${{ github.event_name }}
+          INPUT_PR_NUMBER: ${{ inputs.pr_number }}
+          REPO: ${{ github.repository }}
+          SHA: ${{ github.sha }}
+        run: |
+          if [ "$EVENT_NAME" = "workflow_dispatch" ]; then
+            PR_NUMBER="$INPUT_PR_NUMBER"
+          else
+            # Retry up to 5 times with 10s delay: the commits/{sha}/pulls API
```
electron0zero, line 48: do we need this? did you see failures or it's just added as a defensive guard?

.github/workflows/backport.yml
```diff
@@ -1,45 +1,112 @@
...
+          COUNT=$(gh pr view "$PR_NUMBER" \
+            --repo "$REPO" \
+            --json labels \
+            --jq '[.labels[].name | select(startswith("backport release-"))] | length')
+
+          if [ "$COUNT" -gt "0" ]; then
+            echo "Found $COUNT backport label(s), will run backport."
+            echo "has_backport_labels=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "No backport labels found, skipping backport."
```
electron0zero, line 85: or can we just check the backport labels as first item and bail out if they are missing?

.github/workflows/backport.yml
```diff
@@ -48,9 +115,23 @@ jobs:
...
+
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          token: ${{ steps.app-token.outputs.token }}
+          fetch-depth: 0
+
+      - name: Configure git identity
+        run: |
+          git config --global user.email "132647405+grafana-delivery-bot[bot]@users.noreply.github.com"
```
electron0zero, line 127: do we need this? this bot is not our bot and it's used by grafana team, so if need to set this, it should be out bot/github app account details instead of the grafana team's bot/github app account.

zhxiaogg: yeah, this is needed, in fact this comes from the previous grafana-github-actions, and the new `grafana-github-actions-go` doesn't set it.

.github/workflows/backport.yml
zhxiaogg: yes, it fails during my test. And retry succeeded. I guess Github uses eventual consistency for this data.

.github/workflows/backport.yml
zhxiaogg: > as first item 

Maybe I'm not fully get your idea. But this is the step right after getting PR number/info. The job would stop immediately when it finds the label is not present in the PR.

PR #6586, opened 2026-02-27
Fix bug related to filtering of dedicated columns

stoewer:
**What this PR does**:
This bug can cause missing attributes when the tempo config or overrides contain dedicated columns that are filtered. For example when vp4 blocks contain dedicated columns that correspond to well-known attributes.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/.github/workflows/publish-technical-documentation-next.yml
+++ b/.github/workflows/publish-technical-documentation-next.yml
@@ -20,11 +20,11 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           persist-credentials: false
-      - uses: grafana/writers-toolkit/publish-technical-documentation@7ed542b937fe3202d7840081c36c7bba1d37f34f
+      - uses: grafana/writers-toolkit/publish-technical-documentation@3a5f900daffc465c2c7b12f6ed6d1f6168ba95c0
         with:
           source_directory: docs/sources/tempo
           website_directory: content/docs/tempo/next
-      - uses: grafana/writers-toolkit/publish-technical-documentation@7ed542b937fe3202d7840081c36c7bba1d37f34f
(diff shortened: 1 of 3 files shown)


tempodb/encoding/vparquet5/create.go
```diff
@@ -37,7 +37,7 @@ func (b *backendWriter) Close() error {
 }
 
 func CreateBlock(ctx context.Context, cfg *common.BlockConfig, meta *backend.BlockMeta, i common.Iterator, r backend.Reader, to backend.Writer) (*backend.BlockMeta, error) {
-	s := newStreamingBlock(ctx, cfg, meta, r, to, tempo_io.NewBufferedWriter)
+	s, newMeta := newStreamingBlock(ctx, cfg, meta, r, to, tempo_io.NewBufferedWriter)
```
mdisibio, line 40: The meta returned here is only partially populated. i.e. final stats like size aren't populated yet.   But things happen to work because it's a pointer and the underlying object is continued to be updated.   Might prefer to put it behind a method like `.Meta()` that implies it's the latest at the time of the call, thoughts?

stoewer: Good point. The main motivation behind the signature change was to make it explicit to the caller that the returned `meta` has been modified and needs to be respected on the caller side. Not surfacing this clearly was likely what caused the earlier bug with filtered vs. unfiltered dedicated columns. Using a `.Meta()` doesn’t communicate as clearly that the caller should update its local `meta`.

I can think of a couple of alternative ways to address this:
1. Return a more fully populated `newMeta`
2. Return the filtered dedicated columns instead, and update `meta` at the call site, e.g (preferred)

```go
s, meta.DedicatedColumns = newStreamingBlock(ctx, cfg, meta, r, to, tempo_io.NewBufferedWriter)
```

What do you think?

mdisibio: Ok yes that it was likely the cause for the error, has convinced me.  This signature is good.

PR #6592, opened 2026-03-02
fix: merge per-tenant overrides with defaults for unset fields

electron0zero:
**What this PR does**:

Fixes a bug where per-tenant and wildcard overrides that only set a few fields cause all other fields to resolve to Go zero values instead of falling back to static defaults.

Bug: When a tenant has an override entry that sets only a subset of fields (e.g., just `max_cardinality_per_label: 100`), all other fields resolve to Go zero values instead of the static defaults. This is because the override struct is returned as-is without merging in the defaults for unset fields.

Fix: Each override entry is now field-level merged with static defaults at config load time using reflection. Unset fields fall back to defaults instead of being zero values. Merged views are pre-computed at load/reload - zero cost per request.

Fields where zero is semantically meaningful (e.g., 0 = "no limit", false = "disabled") are migrated to pointer types so the merge can distinguish "not set" (nil) from "explicitly zero" (*0).

This is needed because the merge uses "is this field zero?" to decide whether a field was set in the override file.

For most fields this works fine - a zero value means "not set". But some fields use zero as a real value: `max_bytes_per_trace: 0` means "no limit", `span_multiplier_key: ""` means "disable multiplier". 

Without pointers, the merge would treat these as "not set" and overwrite them with the default, making it impossible for operators to explicitly set these values.

Pointer types solve this because we can check `nil` and assign it to "not set, fall back to default", non-nil zero = "explicitly set to this value".

I also added a type of the ingestion rate strategy config and updated the default overrides and per tenant overrides metrics.

**Behavior before this PR:**
- Lookup returns the first matching override struct whole (per-tenant OR wildcard OR defaults)
- If a per-tenant override exists, it is returned as-is - fields not set in the override file resolve to Go zero values, not the static defaults
- If a wildcard tenant override exists, it is returned as-is - fields not set in the override file resolve to Go zero values, not the static defaults
- if both per tenant and wildcard tenant override doesn't exist, returns default

**Behavior after this PR:**
- Lookup order is unchanged, it's still per-tenant OR wildcard OR defaults
- The matched override entry is now field-level merged with static defaults - fields not set in the override file correctly fall back to the static defaults

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


electron0zero:
@codex review

electron0zero:
@codex review again

modules/overrides/merge_test.go
```diff
@@ -0,0 +1,317 @@
...
+// but not handled in the corresponding Merge method, this test will fail.
+func TestMergeCoversAllFields(t *testing.T) {
+	full := generateFullOverrides()
+
+	t.Run("full overlay onto empty base", func(t *testing.T) {
+		result := (&Overrides{}).Merge(full)
+		assertAllFieldsNonZero(t, reflect.ValueOf(result).Elem(), "Overrides")
+	})
+
+	t.Run("empty overlay onto full base", func(t *testing.T) {
```
electron0zero, line 254: add another test where full merged with empty overrides should be same as full, each field.

do a string compare with full merged with with empty overrides because they should be same.

mdisibio: Commenting that I think this is likely the most important part of this PR.  It should be letting us know 100% that:  (a) all fields are covered (b) all fields are correct (c) adding a new field incorrectly will be caught.

I mentioned that I'd like to check that `result` and `full` are equal too. That might be covered by other tests, but I kind of prefer the location here, i.e. this test is the comprehensive one.

electron0zero: done, added a new test in https://github.com/grafana/tempo/pull/6592/changes/0a3208e9050141d15290ad1e61c54667bcf16d6c

modules/overrides/runtime_config_overrides.go
```diff
@@ -294,9 +310,29 @@ func (o *runtimeConfigOverridesManager) GetRuntimeOverridesFor(userID string) *O
...
+		}
+		// check if a wildcardTenant override exists
+		if m, ok := tenantOverrides.merged[wildcardTenant]; ok {
+			return m
+		}
+	}
+
+	// return default overrides if no per tenant overrides exist
+	return o.defaultLimits
+}
```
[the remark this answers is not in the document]
electron0zero: not a concern, we always wanted the final configs that are applicable for the user.

mapno (commented):
I find very confusing that parameters that are now nullable also have defaults (eg. `max_blocks_per_tag_values_query`).

I understand default as the value when there isn't an explicit value set, but an unset param is getting a new meaning via inheritance. How is that resolved? If `max_blocks_per_tag_values_query`'s default is `0`, for it to inherit another value it should explicitly set to null?

PR #6607, opened 2026-03-04
Make search and metrics streaming easier

mdisibio:
**What this PR does**:
The goal of this PR is to make streaming as easy and compatible as possible.   By default the Grafana gRPC max packet size is 4MB, and Tempo will quickly try to respond with messages much larger than this.  Rather than require users to configure ever larger gRPC packet sizes everywhere, instead this PR makes the query frontend split streaming responses into smaller packet sizes with the same content. Setting a default of 2MB, so for example a 16MB initial response is split into 8 smaller ones.  The 2MB default is to account for the fact that Grafana transforms the Tempo proto objects into the native data-frame format, which is larger.  

I've been testing this locally with a default install of Grafana 12.3, and it's working well.  However it only works for up to ~5K search results.  This needs to be fixed from the Grafana side because it [repeats sending the entire search results list](https://github.com/grafana/grafana/blob/main/pkg/tsdb/tempo/search_stream.go#L116) on each partial gRPC response from Tempo. This causes a default Grafana to exceed its own packet size, error, and restart the query.  However the current PR is still an improvement on the situation.

This also fixes some issues with metrics streaming upon hitting the series limit:
* Updated all locations to use `GRPCFinal()` for the last response, instead of calling `GRPCDiff()`.
* The diff didn't return data after the series limit was hit (because no new shards completed), so it would appear like empty results in the frontend.
* This also impacted instant query streaming, and works now too.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


modules/frontend/combiner/common.go
```diff
@@ -221,6 +225,17 @@ func (c *genericCombiner[T]) GRPCDiff() (T, error) {
 	return diffClone.(T), nil
 }
 
+func (c *genericCombiner[T]) GRPCSegment(response T, maxSize int) ([]T, error) {
+	c.mu.Lock()
```
joe-elliott, line 240: i don't think this mutex is technically needed based on the impls below. afaict none of them actually access the object, but maybe just nice to have since there's an implication this could access shared memory?

modules/frontend/pipeline/collector_grpc.go
```diff
@@ -77,20 +76,38 @@ func (c GRPCCollector[T]) RoundTrip(req *http.Request) error {
 	span.AddEvent("consumeAndCombineResponses done")
 
 	// send the final diff if there is anything left
-	resp, err := c.combiner.GRPCDiff()
+	resp, err := c.combiner.GRPCFinal()
 	if err != nil {
 		return grpcError(err)
 	}
 	span.AddEvent("final combiner.GRPCDiff() done")
```
joe-elliott: nit: combiner.GRCPFinal() done

modules/frontend/combiner/search_tag_values.go
```diff
@@ -107,3 +109,77 @@ func NewSearchTagValuesV2(maxDataBytes int, maxTagsValues uint32, staleValueThre
...
+			Metrics:   response.Metrics,
+		}
+		currentSz = current.Size()
+		out = append(out, current)
+	}
+
+	startNextPacket()
+
+	for _, v := range response.TagValues {
+		itemSz := stringSize(v)
```
joe-elliott: technically not the proto size, but i think a fair estimate? maybe for a ton of very short strings this may matter?

joe-elliott (commented):
nice fix. i should have thought of this months ago

modules/frontend/combiner/common.go
mdisibio: True about the stateless implementations, but I think it's needed to access the c.segment fn pointer?

modules/frontend/combiner/search_tag_values.go
mdisibio: Good point, updated to use the same calculation as in tempopb, which includes some extra bytes, the only tricky part is the length of the string as a varint.

mdisibio:
This is ready for review.  We can safely make these changes on the Tempo side even without the Grafana-side change to search streaming.  This will still work great for TraceQL metrics streaming.

modules/frontend/combiner/search.go
```diff
@@ -155,3 +156,40 @@ func padTraceIDsInResponse(traces []*tempopb.TraceSearchMetadata) {
...
+
+		// Start a new packet if there isn't room for this entry,
+		// unless it's the first one, that way we always try to fit at least one.
+		if len(current.Traces) > 0 && currentSz+traceSz > maxSize {
+			startNextPacket()
+		}
+
+		current.Traces = append(current.Traces, t)
+		currentSz += traceSz
+	}
```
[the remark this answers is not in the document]
mdisibio: Hmm, this is easier said than done.  The reason we don't call `proto.Size()` on the full response as we build it, is because it's very inefficient.  If there are 1000 entries, we don't want to calculate the size of items 0-999 repeatedly (I tried this first and it was of course a non-starter).

We could account for the fiddly bits as you suggest but there isn't a need to be perfect here. We are setting the default to 2MB but the real limit is 4MB, so there is already some overhead accounted for.

What would the fix look like?  If it's simple I'll try it, but if it's complicated I'd rather not.

modules/frontend/combiner/common.go
```diff
@@ -232,6 +236,17 @@ func (c *genericCombiner[T]) GRPCDiff() (T, error) {
...
+	defer c.mu.Unlock()
+
+	if c.segment == nil {
+		return nil, errors.New("grpc response segmentation not supported for response type: " + reflect.TypeOf(response).String())
+	}
+
+	return c.segment(response, maxSize), nil
```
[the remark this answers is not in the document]
mdisibio: This is probably an over-optimization but sure why not.

modules/frontend/combiner/search.go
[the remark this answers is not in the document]
mdisibio: @copilot ⬆️ what do you think?  (btw I already know but I want to test if you can figure this out)

stoewer (dismissed):

modules/frontend/combiner/metrics_query_range.go
```diff
@@ -163,6 +163,51 @@ func NewTypedQueryRange(req *tempopb.QueryRangeRequest, maxSeries int) (GRPCComb
...
+		seriesSz := s.Size()
+
+		// Start a new packet if there isn't room for this entry,
+		// unless it's the first one, that way we always try to fit at least one.
+		if len(current.Series) > 0 && currentSz+seriesSz > maxSize {
+			startNextPacket()
+		}
+
+		current.Series = append(current.Series, s)
+		currentSz += seriesSz
```
javiermolinar, line 199: If we dont want to do a current.Size() for performance reasons, should we take into account the extra bytes for the tag and the used varint bytes?

https://github.com/grafana/tempo/blob/2f74ea818de1377e6ac9dd15117fa284c9c6ba36/pkg/tempopb/tempo.pb.go#L7086

I guess its a very minor correction since this just implies about 3 or 4 bytes per packet which should be fine for the 4MB max in Grafana

modules/frontend/combiner/common.go
```diff
@@ -232,6 +236,18 @@ func (c *genericCombiner[T]) GRPCDiff() (T, error) {
 	return diffClone.(T), nil
 }
 
+func (c *genericCombiner[T]) GRPCSegment(response T, maxSize int) ([]T, error) {
+	c.mu.Lock()
+	segment := c.segment
+	c.mu.Unlock()
+
+	if segment == nil {
+		return nil, errors.New("grpc response segmentation not supported for response type: " + reflect.TypeOf(response).String())
```
stoewer: `nitpick` Use `%T` instead of `reflect` package
```suggestion
		return nil, fmt.Errorf("grpc response segmentation not supported for response type:  %T", response)
```

modules/frontend/combiner/metrics_query_range.go
```diff
@@ -163,6 +163,51 @@ func NewTypedQueryRange(req *tempopb.QueryRangeRequest, maxSeries int) (GRPCComb
...
+// segmentQueryRangeResponseToMaxPacketSize splits resp into one or more QueryRangeResponse values, each within
+// maxSize bytes (by proto Size()), accounting for the response object itself. Metrics are included
+// in every segment; Status and Message are set on the last segment.
+func segmentQueryRangeResponseToMaxPacketSize(resp *tempopb.QueryRangeResponse, maxSize int) []*tempopb.QueryRangeResponse {
+	// If not configured return as-is.
+	if maxSize <= 0 {
+		return []*tempopb.QueryRangeResponse{resp}
+	}
+
+	var out []*tempopb.QueryRangeResponse
```
javiermolinar, line 175: should we preallocate something here?

`out := make([]*tempopb.QueryRangeResponse, 0, 8)`

modules/frontend/combiner/metrics_query_range.go
mdisibio: Yeah copilot called out the same thing in the comments above.  Not opposed to giving it a try.

modules/frontend/combiner/search_test.go
```diff
@@ -787,3 +787,39 @@ func TestSearchCombinerPadTraceIDs(t *testing.T) {
...
+	const maxSize = 1
+	out := segmentSearchResponse(input, maxSize)
+	require.Len(t, out, 2)
+	require.Len(t, out[0].Traces, 1)
+	require.Len(t, out[1].Traces, 1)
+	require.Equal(t, input.Traces[0], out[0].Traces[0])
+	require.Equal(t, input.Traces[1], out[1].Traces[0])
+	require.Equal(t, input.Metrics, out[0].Metrics)
+	require.Equal(t, input.Metrics, out[1].Metrics)
+}
```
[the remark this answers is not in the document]
mdisibio: Added the second set of tests.

modules/frontend/combiner/metrics_query_range.go
mdisibio: Pushed

modules/frontend/combiner/metrics_query_range.go
```diff
@@ -163,6 +163,51 @@ func NewTypedQueryRange(req *tempopb.QueryRangeRequest, maxSeries int) (GRPCComb
...
+		}
+
+		current.Series = append(current.Series, s)
+		currentSz += seriesSz
+	}
+
+	// Attach real status and message only to the last packet
+	if len(out) > 0 {
+		out[len(out)-1].Status = resp.Status
+		out[len(out)-1].Message = resp.Message
```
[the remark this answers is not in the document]
javiermolinar: I think this is not worth

javiermolinar (approved):

PR #6608, opened 2026-03-04
feat: add -health flag for Docker healthchecks

### LINKED ISSUE #6536: Implement a healthcheck command for docker healthchecks
Ever since busybox was removed, it is now difficult to implement a docker healthcheck for Tempo. The only solution is to use a custom docker image, pulling in the necessary network tools and shell.

The Loki team implemented this neat healthcheck command into the loki binary. Something like this would be incredibly useful to have in Tempo aswell. https://github.com/grafana/loki/pull/20590/changes

Creating a custom image with the necessary binaries from busybox. This is inferior as it likely adds back some CVEs and is not really great for production. We do not need a shell or wget or curl for anything other than the purposes of checking Tempo's health, and not having to include these would allow us to make the container more secure.

antonio-mazzini:
## What this PR does

Adds a `-health` CLI flag to the Tempo binary that checks the `/ready` endpoint, enabling native Docker `HEALTHCHECK` without requiring a shell, curl, or wget in the container.

This is important because Tempo uses `gcr.io/distroless/static-debian12` as its base image, which has no shell or network tools.

## Usage

```dockerfile
HEALTHCHECK CMD ["/tempo", "-health"]
```

Custom URL:
```
tempo -health -health.url=http://localhost:3200/ready
```

## How it works

- The `-health` flag is checked **before** any config parsing, so it works without a config file
- Makes an HTTP GET to `http://localhost:3200/ready` (default) with a 5s timeout
- Returns exit code 0 (healthy) or 1 (unhealthy)
- Includes unit tests for flag parsing, URL extraction, and health check behavior

## Reference

Follows the same pattern adopted in Loki: https://github.com/grafana/loki/pull/20590

Closes #6536

### DIFF
--- a/cmd/tempo/health.go
+++ b/cmd/tempo/health.go
@@ -2,33 +2,23 @@ package main
 
 import (
 	"fmt"
+	"io"
 	"net/http"
 	"os"
-	"regexp"
 	"time"
 )
 
 const (
-	healthFlag       = "health"
 	defaultHealthURL = "http://localhost:3200/ready"
 	healthTimeout    = 5 * time.Second
 )
 
-// CheckHealth checks if args contain the -health flag
-func CheckHealth(args []string) bool {
-	pattern := regexp.MustCompile(`^-+` + healthFlag + `$`)
-	for _, a := range args {
-		if pattern.MatchString(a) {
-			return true
-		}
-	}
-	return false
-}
-
 // RunHealthCheck performs a health check against the /ready endpoint.
 // Returns exit code 0 if healthy, 1 if unhealthy.
-func RunHealthCheck(args []string) int {
-	url := getHealthURL(args)
+func RunHealthCheck(url string) int {
+	if url == "" {
+		url = defaultHealthURL
+	}
 
 	client := &http.Client{
 		Timeout: healthTimeout,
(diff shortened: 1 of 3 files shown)


cmd/tempo/health.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	fmt.Fprintf(os.Stderr, "Tempo is unhealthy: status code %d\n", resp.StatusCode)
+	return 1
+}
+
+// getHealthURL extracts the URL from args or returns default.
+// Looks for -health.url=<url> or -health.url <url>.
+func getHealthURL(args []string) string {
+	urlPattern := regexp.MustCompile(`^-+health\.url[=:]?(.*)$`)
```
mdisibio: I'm not sure about the need for this type of manual parsing, and regex.  Could we use the standard flag package for the params?  Similar to how we print the version and exit.

oleg-kozlyuk-grafana: +1 here, use standard libraries

cmd/tempo/health.go
```diff
@@ -0,0 +1,71 @@
...
+		return 1
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusOK {
+		fmt.Println("Tempo is healthy")
+		return 0
+	}
+
+	fmt.Fprintf(os.Stderr, "Tempo is unhealthy: status code %d\n", resp.StatusCode)
```
oleg-kozlyuk-grafana: It would be useful to print the error text in addition to status code

cmd/tempo/main.go
```diff
@@ -45,6 +45,12 @@ func init() {
 }
 
 func main() {
+	// Health check runs before any config parsing so it works without a config file.
```
oleg-kozlyuk-grafana: Do we actually need this to work without the config file? The use case is to run a `HEALTHCHECK` command within the container as far as I understand, the config will be available

antonio-mazzini:
Good points, switched to the standard flag package — much cleaner. Also printing the response body now on errors.

oleg-kozlyuk-grafana:
LGTM, let's get one more pair of eyes on this and merge if no corrections

oleg-kozlyuk-grafana:
@codex review this

cmd/tempo/main.go
```diff
@@ -59,6 +61,9 @@ func main() {
 		fmt.Println(version.Print(appName))
 		os.Exit(0)
 	}
+	if *healthCheck {
+		os.Exit(RunHealthCheck(*healthURL))
```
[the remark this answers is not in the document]
oleg-kozlyuk-grafana: Generally useful, but caused by outdated PR comment in this case

oleg-kozlyuk-grafana:
Thanks for your contribution @antonio-mazzini!

PR #6618, opened 2026-03-05
chore: remove span-metrics leftovers and lazy-init generator clients

javiermolinar:
**What this PR does**:

Now that we have decoupled the livestores from the MetricsGenerator service we can now complete the cleanup. This pr does:

- Metrics generator gprc path is removed
- Simplifies the initialization, SingleBinary mode use a simple callback for pushing traces to the Metrics generator
- Deprecate unnecesary configs
- Remove all query-path from generator and proto code
- Lazy-init generator clients in the distributor. Only for single binary mode.

**How this has been tested**:
- Compiled locally and ran single-binary and distributed docker compose modes. As expected the single binary do no longer connect to Kafka for the distributor

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo/app/modules.go
+++ b/cmd/tempo/app/modules.go
@@ -286,9 +286,7 @@ func (t *App) initGenerator() (services.Service, error) {
 	t.cfg.Generator.ConsumeFromKafka = !IsSingleBinary(t.cfg.Target)
 
 	t.cfg.Generator.Ingest = t.cfg.Ingest
-	if t.cfg.Generator.ConsumeFromKafka {
-		t.cfg.Generator.Ingest.Kafka.ConsumerGroup = generator.ConsumerGroup
-	}
+	t.cfg.Generator.Ingest.Kafka.ConsumerGroup = generator.ConsumerGroup
 
 	genSvc, err := generator.New(&t.cfg.Generator, t.Overrides, prometheus.DefaultRegisterer, t.partitionRing, log.Logger)
 	if errors.Is(err, generator.ErrUnconfigured) && t.cfg.Target != MetricsGenerator { // just warn if we're not running the metrics-generator


cmd/tempo/app/modules.go
```diff
@@ -248,9 +248,13 @@ func (t *App) initOverridesAPI() (services.Service, error) {
...
 
 func (t *App) initDistributor() (services.Service, error) {
+	singleBinary := IsSingleBinary(t.cfg.Target)
+
 	t.cfg.Distributor.KafkaConfig = t.cfg.Ingest.Kafka
 	t.cfg.Distributor.IngesterWritePathEnabled = false
-	t.cfg.Distributor.KafkaWritePathEnabled = t.cfg.Ingest.Enabled // TODO: Don't mix config params
+	t.cfg.Distributor.KafkaWritePathEnabled = false
+	t.cfg.Distributor.PushSpansToKafka = true
+	t.cfg.Distributor.PushSpansToGenerator = singleBinary
```
javiermolinar: This is necessary until we implement the LiveStore grpc path

javiermolinar:
@codex review

cmd/tempo/app/modules.go
```diff
@@ -248,9 +248,11 @@ func (t *App) initOverridesAPI() (services.Service, error) {
 }
 
 func (t *App) initDistributor() (services.Service, error) {
+	singleBinary := IsSingleBinary(t.cfg.Target)
+
 	t.cfg.Distributor.KafkaConfig = t.cfg.Ingest.Kafka
-	t.cfg.Distributor.IngesterWritePathEnabled = false
-	t.cfg.Distributor.KafkaWritePathEnabled = t.cfg.Ingest.Enabled // TODO: Don't mix config params
+	t.cfg.Distributor.PushSpansToKafka = true
```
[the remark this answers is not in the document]
javiermolinar: Kafka is the only deployment model now unless we are in singlebinary mode. So this is totally right

modules/distributor/distributor.go
```diff
@@ -543,19 +553,39 @@ func (d *Distributor) PushTraces(ctx context.Context, traces ptrace.Traces) (*te
...
-			return nil, err
-		}
-	} else if len(d.overrides.MetricsGeneratorProcessors(userID)) > 0 { // See if we need to send to the generators
-		d.generatorForwarder.SendTraces(ctx, userID, ringTokens, rebatchedTraces)
+	if err := d.pushTracesKafka(ctx, userID, ringTokens, rebatchedTraces); err != nil {
+		level.Error(d.logger).Log("msg", "failed to write to kafka", "err", err, "tenant", userID)
+		return nil, err
+	}
+
+	// In single-binary mode we also push directly to metrics-generator over gRPC,
```
mapno: If this functionality is only for monolithic, let's avoid gRPC and push directly via a func

javiermolinar:
@codex review

cmd/tempo/app/modules.go
```diff
@@ -248,16 +242,26 @@ func (t *App) initOverridesAPI() (services.Service, error) {
...
+
```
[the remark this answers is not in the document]
javiermolinar: There are no ingesters anymore

cmd/tempo/app/app_test.go
```diff
@@ -23,6 +23,7 @@ func TestApp_RunStop(t *testing.T) {
 	}()
 
 	config := NewDefaultConfig()
+	config.Target = BackendScheduler
```
mapno, line 26: What's this change?

modules/distributor/config.go
```diff
@@ -42,10 +42,14 @@ type Config struct {
...
-	KafkaWritePathEnabled    bool               `yaml:"kafka_write_path_enabled"`
-	KafkaConfig              ingest.KafkaConfig `yaml:"kafka_config"`
+	// Deprecated: this field will be removed in a future release. Write path routing is set by deployment model.
+	IngesterWritePathEnabled bool `yaml:"ingester_write_path_enabled"`
+	// Deprecated: this field will be removed in a future release. Write path routing is set by deployment model.
+	KafkaWritePathEnabled bool               `yaml:"kafka_write_path_enabled"`
+	KafkaConfig           ingest.KafkaConfig `yaml:"kafka_config"`
+
+	// Internal routing toggle set by app wiring (not user-configurable).
+	PushSpansToKafka bool `yaml:"-"`
```
mapno, line 52: Isn't this always true? Why have this param?

modules/distributor/distributor.go
```diff
@@ -310,22 +304,10 @@ func New(
 		d.usage = tracker
 	}
 
-	var generatorsPoolFactory ring_client.PoolAddrFunc = func(addr string) (ring_client.PoolClient, error) {
-		return generator_client.New(addr, generatorClientCfg)
+	if d.pushSpansToLocalFunc != nil {
+		d.generatorForwarder = newGeneratorForwarder(logger, d.sendToGenerators, o)
```
mapno, line 308: I don't think there is a strong argument to keep all this complexity. If the push fails we can log the error and continue. No need to keep channels, workers, etc.

javiermolinar: The main advantage of this is the decoupling between the distributor and the generator execution. Not that much for error handling but for e2e latency. We can remove it, or we can simplify it to a single bounded channel to not spawn infinite goroutines

javiermolinar: I think we should follow up this in a different pr because it has some tradeoff we need to discuss

cmd/tempo/app/app_test.go
javiermolinar: Now setting up Kafka is mandatory. If we dont set it and we dont set target we get a validation error:

```go
level=info msg="server listening on addresses" http=[::]:60047 grpc=[::]:60048
--- FAIL: TestApp_RunStop (30.00s)
    app_test.go:44: Checking Tempo is up...
    app_test.go:38:
                Error Trace:    /Users/javimolina/grafana/tempo.fork/cmd/tempo/app/app_test.go:38
                                                        /opt/homebrew/Cellar/go/1.26.0/libexec/src/runtime/asm_arm64.s:1447
                Error:          Received unexpected error:
                                failed to init module services: error initialising module: distributor: failed to create distributor: the Kafka topic has not been configured
                Test:           TestApp_RunStop
    app_test.go:43:
                Error Trace:    /Users/javimolina/grafana/tempo.fork/cmd/tempo/app/app_test.go:43
                Error:          Condition never satisfied
                Test:           TestApp_RunStop
FAIL
exit status 1
FAIL    github.com/grafana/tempo/cmd/tempo/app  30.994s
```

Until we make the whole target --all kafkaless we need to use a non kafka module. It can be the backendscheduler or any other one

modules/distributor/distributor.go
mapno: If the concern is decoupling, a goroutine or a simple buffered channel is sufficient. The current queuing code is over-engineered for an in-process call.

I'm ok to leave it if it gets addressed in another PR.

modules/distributor/config.go
javiermolinar: This is a placeholder. When we make the kafkaless mode this will be set to false instructing the distributor to not push to kafka

cmd/tempo/app/modules.go
```diff
@@ -248,16 +242,26 @@ func (t *App) initOverridesAPI() (services.Service, error) {
...
+	var pushSpansToLocalGenerator distributor.PushSpansFunc
+	if singleBinary {
+		pushSpansToLocalGenerator = func(ctx context.Context, req *tempopb.PushSpansRequest) (*tempopb.PushResponse, error) {
+			if t.generator == nil {
+				return nil, errors.New("metrics-generator not initialized")
+			}
+			return t.generator.PushSpans(ctx, req)
+		}
+	}
```
mapno, line 258: Thinking about more kafkaless changes, it'd be nice to instead have a call in the distributor `.AddDirectPush()` that appends these types of calls, because we're going to need it for the live-store as well. That way it also doesn't change the happy path.

I'm fine if you prefer it this way.

cmd/tempo/app/modules.go
```diff
@@ -279,10 +283,12 @@ func (t *App) initDistributor() (services.Service, error) {
...
 
 func (t *App) initGenerator() (services.Service, error) {
-	t.cfg.Generator.Ring.ListenPort = t.cfg.Server.GRPCListenPort
+	t.cfg.Generator.ConsumeFromKafka = !IsSingleBinary(t.cfg.Target)
 
 	t.cfg.Generator.Ingest = t.cfg.Ingest
-	t.cfg.Generator.Ingest.Kafka.ConsumerGroup = generator.ConsumerGroup
+	if t.cfg.Generator.ConsumeFromKafka {
+		t.cfg.Generator.Ingest.Kafka.ConsumerGroup = generator.ConsumerGroup
+	}
```
mapno: Does this matter? Or can it be kept simple as before.

cmd/tempo/app/modules.go
```diff
@@ -294,32 +300,14 @@ func (t *App) initGenerator() (services.Service, error) {
...
-	t.Server.HTTPRouter().Handle(path.Join(api.PathPrefixGenerator, addHTTPAPIPrefix(&t.cfg, api.PathMetricsQueryRange)), queryRangeHandler)
-
-	if !IsSingleBinary(t.cfg.Target) {
-		tempopb.RegisterMetricsGeneratorServer(t.Server.GRPC(), t.generator) // todo: this can be removed before 3.0 but needs to exist as long as we have any deployments anywhere on the traditional arch
-	}
-
 	return t.generator, nil
 }
 
 func (t *App) initGeneratorNoLocalBlocks() (services.Service, error) {
```
mapno, line 304: I believe there is now no functional difference between the two can could be consolidated.

cmd/tempo/app/modules.go
javiermolinar: Good idea

cmd/tempo/app/modules.go
javiermolinar: Yes, this is something I want to tackle on a different PR. We need first to allow the initGenerator to use a different ring and differect codecs for at least one release since this target (initGeneratorNoLocalBlocks) is already in use

cmd/tempo/app/modules.go
javiermolinar: No, you are right. We can keep it as before

cmd/tempo/app/modules.go
javiermolinar: do you mind if we do it in a different pr? this is wired with the middleware and the shim and it involves some other changes

PR #6630, opened 2026-03-09
return error at startup when no write path is configured

### LINKED ISSUE #6629: distributor silently drops all traces when no write path is configured
**Describe the bug**

After upgrading from 2.8.0 to :latest, everything started up fine with no errors. i sent traces through the otel collector and checked grafana but this is what i saw:

Image

i checked the distributor logs but there was no relevant info. then, i looked at the /metrics endpoint:

```
tempo_receiver_accepted_spans{receiver="otlp/otlp_receiver",transport="http"} 12
# tempo_distributor_ingester_appends_total  <-- not present at all
```

it showed that 12 spans were accepted but none were forwarded. There were no errors logged anywhere. After some digging, I found out that both write paths (`IngesterWritePathEnabled` and `KafkaWritePathEnabled`) default to `false` in the current version. Since Kafka isn't set up, nothing is actually being written anywhere. Also, there’s no check at startup to catch this, so Tempo just accepts the traces and silently drops them.

**To Reproduce**

1. Start `grafana/tempo:latest` in single-binary mode.
2. Run without `ingest.enabled: true` (no Kafka configured).
3. Send some traces via OTLP.
4. Check the `/metrics` for `tempo_receiver_accepted_spans` vs `tempo_distributor_ingester_appends_total`.

**Expected behavior**

Tempo should fail at startup with a clear error, not just silently drop every trace.

**Environment:**

- docker compose

areebahmeddd:
**What this PR does:**

* `IngesterWritePathEnabled` is hardcoded to `false`, and `KafkaWritePathEnabled` is set to `ingest.Enabled` (which defaults to `false`).
* With both flags off, the distributor starts without errors but silently drops all traces.
* This PR adds a check in `Config.Validate()` to throw a clear error if both write paths are disabled, so Tempo fails at startup with a useful error message.

**Which issue(s) this PR fixes:**

Fixes #6629 

**Checklist:**

* [x] Tests updated
* [ ] Documentation added
* [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/cmd/tempo/app/app_test.go
+++ b/cmd/tempo/app/app_test.go
@@ -23,6 +23,7 @@ func TestApp_RunStop(t *testing.T) {
 	}()
 
 	config := NewDefaultConfig()
+	config.Target = BackendScheduler
 	config.Server.HTTPListenPort = util.MustGetFreePort()
 	config.Server.GRPCListenPort = util.MustGetFreePort() // not used in the test; set to ensure conflict-free start
 	config.StorageConfig.Trace.Backend = backend.Local


mdisibio:
Hi thank you for the fix.

@mapno @javiermolinar - Is this related to additional 3.0 cleanup needed?  For example I'm thinking in 3.0 the distributor shouldn't be able to enter into this situation, it would always require the Kafka write-path to be enabled.  But I remember we were considering a non-kafka solution for single binary.

areebahmeddd:
> I'm thinking in 3.0 the distributor shouldn't be able to enter into this situation, it would always require the Kafka write-path to be enabled.

yea i figured the same.

> But I remember we were considering a non-kafka solution for single binary.

made a small change to the pr; switched the test target to fix the failing check. i randomly spent some time tracing this one down so hoping this gets merged 🙂

javiermolinar:
This is going to change soon. All these configs are about to be removed. For the distributed mode kafka will be required. For the singlebinary mode we will implement a way to not require Kafka. 

This is implemented in this pull request @areebahmeddd, @mdisibio 

https://github.com/grafana/tempo/pull/6618/changes#diff-eb95dd8feab0971c6292f706ad1d66d784e64c0e30b74eb4c83b5f70f40bffbe

areebahmeddd:
okay, that makes sense. you may close the pr if it's out of scope 🙂
thanks for the fast response on this!

javiermolinar:
This should not be happening now. We have simplified the config and more changes are incoming. Thank you for the pr @areebahmeddd. hopefully the next one will be merged

PR #6634, opened 2026-03-09
[jsonnet]: set block-builder to follow repliacs of live-store by default

zalegrala:
**What this PR does**:

Here we set the rollout-operator to manage the replicas of the block-builder and follow the live-store, since the replicas should always match.  Rollout-operator is no longer optional for live-store, and so we continue to leverage it here for this purpose.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/operations/jsonnet-compiled/microservices-with-extras/gen/StatefulSet-block-builder.yaml
+++ b/operations/jsonnet-compiled/microservices-with-extras/gen/StatefulSet-block-builder.yaml
@@ -15,6 +15,9 @@ spec:
     metadata:
       annotations:
         config_hash: 42cd6f6d5e54ec2a274f12dc9d7e8d7b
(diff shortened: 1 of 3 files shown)


mdisibio:
When scaling down a cluster - the rollout operator already drains a live store and waits 30-60 minutes for it to go outside the query window, before actually removing the pod, correct?   I think that gives enough time for the block builder to drain and flush it's partition as well.

zalegrala:
Yes, that's correct, which is why we are using the live-store statefulset here (we picked `a`) and not the ReplicaTemplate, since the scaledown of the ReplicaTemplate would reflect the replica reduction immediately.  We allow changing the target here in case `a` or `b` are not always the zones in a cluster.

PR #6641, opened 2026-03-10
fix(storage): support FQDN trailing dot in storage backends via custom transport for Kubernetes ndots=5

### LINKED ISSUE #1726: Use fully qualifed domain name for storage backends when running on Kubernetes
**Problem:**

Kubernetes configures `ndots=5` under `/etc/resolv.conf` options, that means any hostname that has less then 5 dots in it is looked up locally first, and then finally it is resolved.

As per our tests, we have seen 11 local lookups before DNS resolution is done (see https://github.com/grafana/jsonnet-libs/pull/762 for more details). These local lookups add latency, take up space in the conntrack table and add extra load to the kube-dns service. 

Tempo makes a tons of API calls to storage backend, and these extra DNS in storage backend calls adds to read and write latency for tempo.

We started looking into it because we noticed DNS Lookup Errors (see https://github.com/grafana/tempo/issues/1462), which we improved with https://github.com/grafana/tempo/pull/1632, and lowered ndots=3 to deal with it.

There are two ways to solve this: 
1. Use FQDN (hostname tld with dot) in storage backend config, and that will hint kube-dns to skip local search and do a direct lookup
2. lower ndots but it can break kubernetes service discovery [we can't do that]

solution 2 is not ideal so we are only left with solution 1.

This is NOT an issue if you are running tempo on VMs, In most Linux distros ndots is not configured,  and default value of 1 is used. if `cat /etc/resolv.conf  | grep ndots` shows nothing, that means it's set to default.

See more: 
- https://pracucci.com/kubernetes-dns-resolution-ndots-options-and-why-it-may-affect-application-performances.html
- https://github.com/grafana/jsonnet-libs/pull/762

**Solution:**

We start using FQDN in storage config, and suggest our users to use FQDN when running tempo on kubernetes.

**Azure Blob Storage:** If we just add dot at the end of TLD, we get 400 (see more https://github.com/grafana/tempo/issues/1462#issuecomment-1224082681), that happens because Azure doesn't allow trailing dot in `Host` header. We need to trim dot from Host header in Azure Client

**Google Cloud Storage:** not tested yet

**AWS S3:**  not tested yet

**Minio:** not tested yet

priyendang:
## What this PR does

Fixes #1726

Adds a custom `fqdnTransport` (`http.RoundTripper`) to all three storage
backends (Azure, S3, GCS) that strips the trailing dot from the `Host`
header **only**, while keeping the trailing dot in the URL so that DNS
resolution uses the fully qualified domain name.

---

## Problem

Kubernetes sets `ndots=5` in every Pod's `/etc/resolv.conf`. Any hostname
with fewer than 5 dots (e.g. `blob.core.windows.net`) is searched locally
first before a real DNS lookup, causing:

- Up to 11 failed DNS lookups per storage API call
- Added read/write latency on every Tempo operation
- Pressure on the conntrack table and extra load on kube-dns
- DNS lookup failures at scale (see also: #1462)

---

## Solution

A custom `fqdnTransport` wraps the existing HTTP transport stack on all
three backends. On every request it:

1. **Keeps the trailing dot in the URL** → the HTTP client dials using
   `blob.core.windows.net.` → kube-dns recognises the FQDN and skips
   local search entirely
2. **Strips the trailing dot from the `Host` header only** → the cloud
   API receives a valid host header with no trailing dot

This is the correct split — `req.URL.Host` drives DNS resolution while
`req.Host` drives the `Host:` header sent over the wire.

| Backend | API rejects trailing dot in Host with | Config field |
|---|---|---|
| Azure | HTTP 400 | `endpoint_suffix` |
| S3 | HTTP 404 | `endpoint` |
| GCS | HTTP 301 redirect | `bucket_name` |

---

## Changes

| File | What changed |
|---|---|
| `tempodb/backend/azure/azure_helpers.go` | Added `fqdnTransport`; `cfg.Endpoint` passed as-is to URL (dot retained) |
| `tempodb/backend/s3/s3.go` | Added `fqdnTransport`; `cfg.Endpoint` passed as-is to minio (dot retained) |
| `tempodb/backend/gcs/gcs.go` | Added `fqdnTransport`; `cfg.BucketName` passed as-is to GCS client (dot retained) |
| `tempodb/backend/azure/azure_helpers_test.go` | Tests verifying Host header has dot stripped; URL retains dot |
| `tempodb/backend/s3/s3_test.go` | Tests verifying fqdnTransport strips dot from Host header only |
| `tempodb/backend/gcs/gcs_test.go` | Tests verifying fqdnTransport strips dot from Host header only |

---

## Example config
```yaml
# Azure
storage:
  trace:
    backend: azure
    azure:
      endpoint_suffix: blob.core.windows.net.   # trailing dot = FQDN

# S3
storage:
  trace:
    backend: s3
    s3:
      endpoint: s3.amazonaws.com.               # trailing dot = FQDN

# GCS
storage:
  trace:
    backend: gcs
    gcs:
      bucket_name: my-tempo-traces.             # trailing dot = FQDN
```

---

## Backward compatibility

- `fqdnTransport.RoundTrip` is a no-op when the Host header has no trailing dot
- Existing users without trailing dots are completely unaffected
- No config changes required for non-Kubernetes deployments

---

## Testing
```
go test -short ./tempodb/backend/azure/...   # ok
go test -short ./tempodb/backend/s3/...      # ok
go test -short ./tempodb/backend/gcs/...     # ok
```
```

### New Commit Message (amend):
```
fix(storage): support FQDN trailing dot via custom transport for Kubernetes ndots=5

On Kubernetes (ndots=5), storage hostnames trigger up to 11 spurious
DNS lookups per API call. Users can configure a trailing dot on their
storage endpoint/bucket to signal a fully qualified domain name (FQDN),
which causes kube-dns to skip local search entirely.

This adds fqdnTransport (http.RoundTripper) to all three storage backends:
- Keeps the trailing dot in the URL so DNS resolves using the FQDN
- Strips the trailing dot from the Host header only, since Azure (400),
  S3 (404) and GCS (301) all reject a trailing dot in the Host header

The dot is NOT stripped from cfg.Endpoint/cfg.BucketName upfront —
that was the previous (incorrect) approach which lost the DNS benefit.

Fixes grafana#1726

### DIFF
--- a/tempodb/backend/azure/azure_helpers.go
+++ b/tempodb/backend/azure/azure_helpers.go
@@ -23,6 +23,28 @@ const (
 	maxRetries = 1
 )
 
+// fqdnTransport is a custom http.RoundTripper that strips the trailing dot
+// from the Host header before sending the request, while keeping the trailing
+// dot in the URL so that DNS resolution uses the fully qualified domain name.
+//
+// This solves the Kubernetes ndots=5 problem (issue #1726): users can configure
+// a trailing dot in endpoint_suffix (e.g. "blob.core.windows.net.") to hint
+// kube-dns to skip local search and perform a direct DNS lookup, eliminating
+// up to 11 spurious DNS queries per storage API call. The trailing dot is then
+// stripped from the Host header only, because Azure rejects it with HTTP 400.
+type fqdnTransport struct {
+	wrapped http.RoundTripper
+}
+
+func (t *fqdnTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Clone the request to avoid mutating the original
+	r := req.Clone(req.Context())
+	// Strip trailing dot from Host header only.
+	// The URL retains the trailing dot for correct FQDN DNS resolution.
+	r.Host = strings.TrimSuffix(r.Host, ".")
+	return t.wrapped.RoundTrip(r)
+}
+
 func getContainerClient(ctx context.Context, cfg *Config, hedge bool) (*container.Client, error) {
 	var err error
 
@@ -59,6 +81,10 @@ func getContainerClient(ctx context.Context, cfg *Config, hedge bool) (*containe
 		instrumentation.PublishHedgedMetrics(stats)
 	}
 
+	// Wrap with fqdnTransport so the Host header has its trailing dot stripped
+	// on every request, while the URL retains it for FQDN DNS resolution.
+	transport = &fqdnTransport{wrapped: transport}
+
 	opts := azblob.ClientOptions{}
 	opts.Transport = &http.Client{Transport: transport}
 	opts.Retry = retry
@@ -67,6 +93,10 @@ func getContainerClient(ctx context.Context, cfg *Config, hedge bool) (*containe
 	}
 
 	accountName := getStorageAccountName(cfg)
+
+	// Use cfg.Endpoint as-is (trailing dot preserved) so the URL carries the
+	// FQDN for correct DNS resolution. The fqdnTransport above strips the dot
+	// from the Host header before the request is sent to Azure.
 	u, err := url.Parse(fmt.Sprintf("https://%s.%s", accountName, cfg.Endpoint))
 
 	// If the endpoint doesn't start with blob. we can assume Azurite is being used
--- a/tempodb/backend/azure/azure_helpers_test.go
+++ b/tempodb/backend/azure/azure_helpers_test.go
@@ -2,11 +2,14 @@ package azure
 
 import (
 	"context"
+	"net/http"
+	"net/http/httptest"
 	"os"
 	"testing"
 
 	"github.com/grafana/dskit/flagext"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 const (
@@ -97,6 +100,31 @@ func TestGetContainerClient(t *testing.T) {
 			endpoint:    "blob.core.cloudapi.de",
 			expectedURL: "https://devstoreaccount1.blob.core.cloudapi.de/traces",
 		},
+		// FQDN test cases for Kubernetes ndots=5 support (issue #1726).
+		// Users can add a trailing dot to endpoint_suffix in config to hint
+		// kube-dns to skip local search and perform a direct DNS lookup.
+		// The URL must RETAIN the trailing dot (for FQDN DNS resolution),
+		// and the fqdnTransport strips it from the Host header before sending.
+		{
+			name:        "Azure Global FQDN with trailing dot",
+			endpoint:    "blob.core.windows.net.",
+			expectedURL: "https://devstoreaccount1.blob.core.windows.net./traces",
+		},
+		{
+			name:        "Azure China FQDN with trailing dot",
+			endpoint:    "blob.core.chinacloudapi.cn.",
+			expectedURL: "https://devstoreaccount1.blob.core.chinacloudapi.cn./traces",
+		},
+		{
+			name:        "Azure US Government FQDN with trailing dot",
+			endpoint:    "blob.core.usgovcloudapi.net.",
+			expectedURL: "https://devstoreaccount1.blob.core.usgovcloudapi.net./traces",
+		},
+		{
+			name:        "Azure German FQDN with trailing dot",
+			endpoint:    "blob.core.cloudapi.de.",
+			expectedURL: "https://devstoreaccount1.blob.core.cloudapi.de./traces",
+		},
 	}
 
 	for _, tc := range tests {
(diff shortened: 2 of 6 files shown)


priyendang:
CI / RUN Unit Tests (test-with-cover-others) check: - The CI failure in TestBlockbuilder_flushingFails is a flaky temp directory cleanup issue in modules/blockbuilder — unrelated to this PR which only touches tempodb/backend/{azure,s3,gcs}. No files in modules/blockbuilder were modified.

tempodb/backend/azure/azure_helpers.go
```diff
@@ -23,6 +23,28 @@ const (
 	maxRetries = 1
 )
 
+// fqdnTransport is a custom http.RoundTripper that strips the trailing dot
```
electron0zero, line 26: can this comment be less verbose and more focused

tempodb/backend/azure/azure_helpers.go
```diff
@@ -67,6 +93,10 @@ func getContainerClient(ctx context.Context, cfg *Config, hedge bool) (*containe
 	}
 
 	accountName := getStorageAccountName(cfg)
+
+	// Use cfg.Endpoint as-is (trailing dot preserved) so the URL carries the
+	// FQDN for correct DNS resolution. The fqdnTransport above strips the dot
+	// from the Host header before the request is sent to Azure.
```
electron0zero, line 99: not sure if this comment makes sense here? there is not code change here?

tempodb/backend/azure/azure_helpers.go
```diff
@@ -164,4 +194,4 @@ func getStorageAccountKey(cfg *Config) string {
 	}
 
 	return accountKey
-}
+}
```
electron0zero, line 197: no trailing `\n`

tempodb/backend/azure/azure_helpers_test.go
```diff
@@ -97,6 +100,31 @@ func TestGetContainerClient(t *testing.T) {
 			endpoint:    "blob.core.cloudapi.de",
 			expectedURL: "https://devstoreaccount1.blob.core.cloudapi.de/traces",
 		},
+		// FQDN test cases for Kubernetes ndots=5 support (issue #1726).
```
electron0zero, line 103: less verbose comment?

tempodb/backend/azure/azure_helpers_test.go
```diff
@@ -109,3 +137,61 @@ func TestGetContainerClient(t *testing.T) {
...
+			resp, err := transport.RoundTrip(req)
+			require.NoError(t, err)
+			defer resp.Body.Close()
+
+			// The Host header received by the server must never have a trailing dot
+			assert.Equal(t, tc.expectedHostHdr, receivedHost,
+				"Host header must have trailing dot stripped before sending to server")
+		})
+	}
+}
```
electron0zero, line 197: no trailing `\n`

tempodb/backend/gcs/gcs.go
```diff
@@ -51,6 +51,28 @@ var (
...
+// a trailing dot in bucket_name (e.g. "my-traces-bucket.") to hint kube-dns to
+// skip local search and perform a direct DNS lookup, eliminating up to 11
+// spurious DNS queries per storage API call. The trailing dot is then stripped
+// from the Host header only, because GCS rejects it with a 301 redirect.
+type fqdnTransport struct {
+	wrapped http.RoundTripper
+}
+
+func (t *fqdnTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Clone the request to avoid mutating the original
```
electron0zero, line 68: will this break the context propagation? can you double check it to make sure we don't break it

tempodb/backend/gcs/gcs.go
```diff
@@ -51,6 +51,28 @@ var (
...
+	r := req.Clone(req.Context())
+	// Strip trailing dot from Host header only.
+	// The URL retains the trailing dot for correct FQDN DNS resolution.
+	r.Host = strings.TrimSuffix(r.Host, ".")
```
electron0zero, line 72: can this be wrapped in a condition so it only happen when there is a trailing dot in the host

tempodb/backend/gcs/gcs.go
```diff
@@ -249,8 +271,6 @@ func (rw *readerWriter) ListBlocks(ctx context.Context, tenant string) ([]uuid.U
 				id    uuid.UUID
 			)
 
-			// If max is global max, then we don't want to set an end offset to
```
electron0zero, line 252: seems unrelated? not sure why these and other comments were removed below?

tempodb/backend/gcs/gcs.go
```diff
@@ -579,4 +603,4 @@ func createPreconditions(version backend.Version) (preconditions storage.Conditi
 	}
 	preconditions.GenerationMatch = generation
 	return
-}
+}
```
electron0zero, line 606: same here, no `\n`

tempodb/backend/gcs/gcs_test.go
```diff
@@ -66,13 +66,10 @@ func TestHedge(t *testing.T) {
 
 			ctx := context.Background()
 
-			// the first call on each client initiates an extra http request
```
electron0zero, line 69: same here? why were these comments removed

tempodb/backend/gcs/gcs_test.go
```diff
@@ -328,15 +313,9 @@ func TestObjectWithPrefix(t *testing.T) {
 			httpHandler: func(t *testing.T) http.HandlerFunc {
 				return func(w http.ResponseWriter, r *http.Request) {
 					if r.Method == "GET" {
-						_, _ = w.Write([]byte(`
```
electron0zero, line 331: this diff seems to be unrelated? can you explain this diff?

tempodb/backend/gcs/gcs_test.go
```diff
@@ -474,31 +437,21 @@ func TestListBlocksWithPrefix(t *testing.T) {
 				return func(w http.ResponseWriter, r *http.Request) {
 					if r.Method == "GET" {
 						assert.Equal(t, "a/b/c/single-tenant/", r.URL.Query().Get("prefix"))
-
 						_, _ = w.Write([]byte(`
 						{
 							"kind": "storage#objects",
 							"items": [{
-								"kind": "storage#object",
```
electron0zero, line 482: same here and other places across this test.

tempodb/backend/s3/s3.go
```diff
@@ -96,6 +96,28 @@ func (s *overrideSignatureVersion) IsExpired() bool {
...
+// This solves the Kubernetes ndots=5 problem (issue #1726): users can configure
+// a trailing dot in endpoint (e.g. "s3.amazonaws.com.") to hint kube-dns to
+// skip local search and perform a direct DNS lookup, eliminating up to 11
+// spurious DNS queries per storage API call. The trailing dot is then stripped
+// from the Host header only, because S3 rejects it with HTTP 404.
+type fqdnTransport struct {
+	wrapped http.RoundTripper
+}
+
+func (t *fqdnTransport) RoundTrip(req *http.Request) (*http.Response, error) {
```
electron0zero, line 112: this code is duplicated across all the backend? I think it's better to move it into a common place and reuse across the backends.

tempodb/backend/s3/s3.go
```diff
@@ -590,10 +606,6 @@ func (rw *readerWriter) WriteVersioned(ctx context.Context, name string, keypath
 }
 
 func (rw *readerWriter) DeleteVersioned(ctx context.Context, name string, keypath backend.KeyPath, version backend.Version) error {
-	// Note there is a potential data race here because S3 does not support conditional headers. If
```
electron0zero, line 593: also here? why was this comment removed?

tempodb/backend/s3/s3_test.go
```diff
@@ -378,7 +348,7 @@ func TestObjectBlockTags(t *testing.T) {
 				SecretKey: flagext.SecretWithValue("test"),
 				Bucket:    "blerg",
 				Insecure:  true,
-				Endpoint:  server.URL[7:], // [7:] -> strip http://
```
electron0zero, line 381: why was this comment removed???

electron0zero (changes requested):
the core change looks reasonable but there are lots of unrelated and stray changes like comments are removed, etc.

Sidenote: It does look like LLMs were used in producing this PR.  LLM usage is okay but please review the code before you create the PR and ensure that it makes sense. sending unreviewed LLM output creates extra burden on the maintainers.

PR #6648, opened 2026-03-11
jsonnet: Add emptyDir data volume to block-builder StatefulSet

mapno:
**What this PR does**:

The block-builder StatefulSet writes temporary data to the container's root filesystem. This offers no benefit: rootfs writes go through the overlay filesystem (slower), have no resource isolation, and if the node's root disk fills up it affects all pods on the node.

This PR adds an emptyDir volume with a configurable `sizeLimit` (default 5Gi) mounted at `/var/tempo`. Since block-builder data is ephemeral, a PVC is unnecessary.

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


operations/jsonnet/microservices/config.libsonnet
```diff
@@ -95,6 +95,7 @@
     },
     block_builder: {
       replicas: 0,
+      data_volume_size: '5Gi',
```
mapno, line 98: How's this default?

stoewer: Looks reasonable

mdisibio: I checked some clusters internally, and the average is 300MB of block data per 3 minute flush window, but there was a maximum of 1.5GB.

PR #6653, opened 2026-03-11
generator: fix drain old series on metric replacement to prevent limi…

carles-grafana:
…ter leak

When a processor config change (dimensions, histogram buckets, etc.) triggers processor recreation, old metrics are replaced in the registry via registerMetric. Previously, the old metric's series were silently garbage-collected without calling OnDelete on the limiter, permanently inflating LocalSeriesLimiter.activeSeries. Each replacement accumulated more leaked counts, eventually pushing activeSeries above maxActiveSeries and trapping the tenant in permanent overflow.

Drain all old series through removeStaleSeries before replacing the metric, which properly calls OnDelete for each series and decrements the limiter's counter.

**What this PR does**:

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/generator/registry/registry.go
+++ b/modules/generator/registry/registry.go
@@ -2,6 +2,7 @@ package registry
 
 import (
 	"context"
+	"math"
 	"os"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


modules/generator/registry/registry.go
```diff
@@ -208,8 +208,11 @@ func (r *ManagedRegistry) registerMetric(m metric) {
 	r.metricsMtx.Lock()
 	defer r.metricsMtx.Unlock()
 
-	if _, ok := r.metrics[m.name()]; ok {
+	if old, ok := r.metrics[m.name()]; ok {
 		level.Info(r.logger).Log("msg", "replacing metric, counters will be reset", "metric", m.name())
+		// Drain old series so the limiter's active count is properly decremented.
+		// Use a timestamp far in the future to force-remove all series.
+		old.removeStaleSeries(time.Now().Add(time.Hour).UnixMilli())
```
electron0zero: time.Hour seems fine but would it be better to use something like math.MaxInt64 if we want to go far into future?

carles-grafana: stale time is 15 minutes so 1 hour has the same result, but makes sense to use max integer since 1 hour feels like a magic number. fixed

PR #6660, opened 2026-03-12
fix: skip per-label limiter and sanitizer for info metrics

electron0zero:
**What this PR does**:

The PerLabelLimiter was tracking cardinality for ALL labels that passed through it, including resource attributes from target_info and host_info.

This caused the demand estimate metric to report way more label names instead of just the configured dimensions from span-metrics and service-graphs.

info metrics carry all resource attributes as labels and have high cardinality by design so we should not limit and sanitize info metrics.

Made changes to skip the span name drain sanitizer for info metrics as well.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [x] Documentation added


electron0zero:
@codex review

electron0zero:
@codex review again

PR #6670, opened 2026-03-12
Cleanup changelog after 2.10.2 release

zhxiaogg:
**What this PR does**:

- cleanup changelog in mainline after 2.10.2 release

Fixes N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


zhxiaogg:
> I think we usually don't remove it from main since it was commit there before backporting

@ie-pham I agree it's a bit confusing. But probably we did this before as well, see: https://github.com/grafana/tempo/commit/b32432ec964837fe957633a5c241cd0c231e1741 and https://github.com/grafana/tempo/commit/05e3654e36a31602d2def7e740bec610d572c0da

And this is part of the release process. I feel it's more about declare the new version change logs than cleaning up the pendings in main. See: 

https://github.com/grafana/tempo/blob/3f6deffd14f99ccf438e714dc5d55a6d9ca60399/RELEASES.MD#L72

PR #6681, opened 2026-03-13
fix: make live-store lag tests deterministic

javiermolinar:
**What this PR does**:
Makes the live-store lag tests deterministic by avoiding a started PartitionReader in `TestIsLagged` and fixes the `SearchRecent` test request start timestamp.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


CHANGELOG.md
```diff
@@ -36,6 +36,7 @@
 * [ENHANCEMENT] Allow span_name_sanitization to be set via user-configurable overrides API [#6411](https://github.com/grafana/tempo/pull/6411) (@Logiraptor)
 * [ENHANCEMENT] Add `fail_on_high_lag` parameter to allow live-store to fail if it is lagged [#6363](https://github.com/grafana/tempo/pull/6363) [#6567](https://github.com/grafana/tempo/pull/6567) (@ruslan-mikhailov, @carles-grafana)
 * [ENHANCEMENT] Add support for per-tenant left-padding of trace IDs [#6439](https://github.com/grafana/tempo/pull/6489) (@mapno)
+* [BUGFIX] Make live-store lag tests deterministic to avoid flaky failures. [#6681](https://github.com/grafana/tempo/pull/6681) (@javiermolinar)
```
ruslan-mikhailov: I think from changelog reader perspective, it is a no-op change as it does not change the product. I would suggest adding it to this line:
https://github.com/grafana/tempo/pull/6681/changes#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR37

ruslan-mikhailov (commented):
LGTM

CHANGELOG.md
javiermolinar: we can remove it

MukundaKatta:
Test determinism refactor looks solid — the `newMockLiveStore` helper cleanly avoids the real `PartitionReader` init race.

Two small things:

1. The `[BUGFIX] ... [#6428]` CHANGELOG entry in this diff already exists on `main` (I see the same line in `origin/main:CHANGELOG.md` after fetching). Merging this PR as-is would leave a duplicate line. Probably safe to just drop the `+` line from the CHANGELOG hunk.

2. Nice silent bonus fix here:
   ```diff
   -   Start: uint32(now.Add(-5 * time.Hour).Second()),
   +   Start: uint32(now.Add(-5 * time.Hour).Unix()),
   ```
   `time.Time.Second()` returns the 0–59 second offset within the minute, not a Unix timestamp, so the original test was almost certainly not exercising what it looked like it was. Worth calling out in the PR description so reviewers know this is part of the determinism story, not a mechanical change.

PR #6724, opened 2026-03-17
[backend] Add flag to enable ListObjects V1

JaspreetSandhu:
**What this PR does**:

Enables `ListObjects` V1 calls to S3-compatible backends.

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


mdisibio:
Out of curiosity - what s3-compatible backend is this to support?

@zalegrala might be good to get your thoughts on this.  I know that a lot of work has been done for performance in this area, i.e. polling up to 500K-1M blocks in one tenant.   Don't think we would ever run this option ourselves, or be able to troubleshoot it in depth, but perhaps it if unlocks additional backends it could be useful to the community.

zalegrala:
Right, I'm curious which platform this is solving for.  Can you expand on the use case?  I expect the v2 API to be more performant due to the pagination.

PR #6742, opened 2026-03-19
feat: add tempo_cache_requests_total metric

oleg-kozlyuk-grafana:
**What this PR does**:

Adds dedicated hit/miss counters with a `type` label ("memcached" or "redis") to both Memcached and RedisCache, incremented in Fetch and FetchKey methods.

**Checklist**
- [x] Tests updated
- [ ] Documentation added


oleg-kozlyuk-grafana:
@codex review again

javiermolinar (commented):
I like this metric since it comes from Tempo but it partial overlap with the memcached ones and probably redis too.

`memcached_commands_total`

Image

This is non blocking though.

mdisibio:
New metrics are more worthwhile, if we include the `role` as a label.  This is something we lack with the current exporter-based monitoring, since they operate on the cache cluster, which is not always a 1-to-1 mapping with `role`.  i.e. cache cluster could be a shared cluster `memcached-01` but role could be `bloom`, etc.

PR #6746, opened 2026-03-19
enhancement: Use string interning using Go's builtin 'unique' package

oleg-kozlyuk-grafana:
**What this PR does**:

Adds an optimization on reading from Parquet v4/v5 that 'interns' strings. As most of spans have repeated info, this 
should result in reduction of memory footprint

Ref: https://github.com/golang/go/issues/71926 - in recent versions, `unique.Make` should not perform extra allocations, which makes this a viable alternative to `unsafeToString`

TODO: gather metrics

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added


oleg-kozlyuk-grafana:
@codex review

oleg-kozlyuk-grafana:
In case anyone is curious about this: this is fundamentally possible to execute, but to reap the benefits, requires full pipeline integration - in both protobuf serialization/deserialization & in parquet reading. Migration to vtproto would be required first

PR #6766, opened 2026-03-23
fix(s3): paginate ClearBlock deletes and add configurable S3 response header timeout

swood:
Fix retention cleanup against S3-compatible backends where listing under a block
prefix can return truncated results or take longer than a single ListObjects
round-trip allows.

ClearBlock now walks object keys with ListObjectsV2 pagination so all objects
under a block prefix are deleted, matching behavior expected for retention.

Add optional storage.trace.s3.response_header_timeout so deployments on slow
S3 gateways (e.g. long ListObjects) can raise the HTTP client response header
timeout above the minio default (1m). When unset, behavior is unchanged.

Build: use tempo-distributed Helm 2.10.x with image built from v2.10.2-based
branches, not main (config schema differs).

### DIFF
--- a/tempodb/backend/s3/compactor.go
+++ b/tempodb/backend/s3/compactor.go
@@ -55,20 +55,33 @@ func (rw *readerWriter) ClearBlock(blockID uuid.UUID, tenantID string) error {
 		return backend.ErrEmptyBlockID
 	}
 
-	path := backend.RootPath(blockID, tenantID, rw.cfg.Prefix) + "/"
+	path := backend.RootPath(blockID, tenantID, rw.cfg.Prefix)
+	if len(path) > 0 {
+		path += "/"
+	}
 	level.Debug(rw.logger).Log("msg", "deleting block", "block path", path)
 
-	// ListObjects(bucket, prefix, marker, delimiter string, maxKeys int)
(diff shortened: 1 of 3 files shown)


mdisibio:
Hi, this is an interesting change.  Are you seeing these issues in your environment, could you share any details about your setup?  With really long (60+s) header timeouts I would be surprised if Tempo is working at all.  And likewise our largest blocks should max out around 100 files inside the block folder, and it shouldn't be having issues with paginated responses, which I think default to 1000 items?

I think we can proceed with the change, but I'd like to know how it's faring in a real environment, all things considered.


I would like to see if we can change the behavior of the default.  In the case of 0 unconfigured - to me it seems a little misleading because there still is a default.  But both the manifest and /status endpoints would show the zero.  I saw you added the note in the comment, but that has a chance to drift.

Let's see if we can set the default from the actual minio default like we did here:  https://github.com/grafana/tempo/pull/6561/changes#diff-751719400a8d7b5c62a91097d99550d912f2bb41f6e0f93bfb707d01ceef9c99R94   That way the manifest and /status endpoints always show the real value, which is very helpful.   In this case what if we did `def, err := minio.DefaultTransport(false).    cfg.ResponseHeaderTimeout = def.ResponseHeaderTimeout`

Also please run `make generate-manifest` and add a changelog entry.

PR #6778, opened 2026-03-25
fix(blockbuilder): resolve partition lag via mid-cycle streaming flush

mattdurham:
## Problem

`TempoBlockBuilderPartitionLagWarning` and `TempoBlockBuilderPartitionLagCritical` fire . Root cause: `MaxBytesPerCycle` (5GB cap) stops Kafka consumption mid-cycle. High-ingest partitions (37 MB/s = 11GB/cycle) can never catch up. Pod memory at 82% of 7GB limit — cannot simply raise the cap.

## Solution: Mid-Cycle Streaming Flush

Replace stop-reading with flush-and-continue: when `liveTraces` memory exceeds threshold (`MaxBytesInMemory`, default 2.5GB), flush all tenants to object storage, reset liveTraces, and continue reading. Lagging partitions now catch up in one `consumePartition()` call instead of taking 15–20 minutes.

**Memory safety:** threshold (2.5GB) + encoding overhead (~15-30%) + GC lag (~50%) ≈ 4.5GB peak — within 7GB pod limit. Currently peaks at 5.76GB with the 5GB cap. Streaming flush is safer.

**Idempotency:** `DeterministicIDGenerator` is keyed on `(tenantID, partitionID, startOffset)` with an internal counter. Multiple flushes per cycle produce unique sequential IDs. On restart from the same committed offset, the same sequence regenerates.

## Additional Fixes

- **Lag metric timing**: Remove start-of-cycle `SetPartitionLagSeconds` call that caused alerts to fire longer than actual lag window (avg_over_time averages start+end values)
- **Alert annotation**: Fix `TempoBlockBuilderPartitionLagWarning` annotation using wrong threshold variable
- **Stable partition sort**: Add partition ID as tiebreaker for deterministic ordering
- **AllowCompaction multi-block**: Track all flushed block IDs per cycle (not just last) so all no-compact flags are cleared
- **Observability**: `tempo_block_builder_flush_threshold_reached_total` counter lets operators observe when threshold is hit

## Test Coverage

All changes TDD'd with new tests:
- `TestBlockBuilder_StreamingFlush_MultipleBlocksCreated` — multiple blocks in one cycle
- `TestBlockBuilder_FlushThresholdCounter` — counter increments on threshold hit
- `TestBlockBuilder_honor_maxBytesInMemory` — threshold behavior including disabled (0)
- `TestPartitionWriter_TotalSize` / `TestPartitionWriter_FlushAndReset`
- `TestTenantStoreReset` / `TestTenantStoreAllowCompaction_MultipleBlocks`
- `TestPartitionSort_Stable` / `TestLagMetric_NotSetAtStartOfCycle`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/modules/blockbuilder/blockbuilder.go
+++ b/modules/blockbuilder/blockbuilder.go
@@ -304,6 +304,9 @@ func (b *BlockBuilder) consume(ctx context.Context) (time.Duration, error) {
 	// Iterate over the laggiest partition until the lag is less than the cycle duration or none of the partitions has records
 	for {
 		sort.Slice(ps, func(i, j int) bool {
+			if ps[i].lastRecordTs.Equal(ps[j].lastRecordTs) {
+				return ps[i].partition < ps[j].partition
+			}
 			return ps[i].lastRecordTs.Before(ps[j].lastRecordTs)
 		})
 
@@ -413,8 +416,6 @@ outer:
 			// Initialize on first record
 			if !init {
 				end = rec.Timestamp.Add(dur) // When block will be cut
-				// Record lag at the start of the consumption
-				ingest.SetPartitionLagSeconds(group, ps.partition, time.Since(rec.Timestamp))
 				writer = newPartitionSectionWriter(
 					b.logger,
 					uint64(ps.partition),
--- a/modules/blockbuilder/blockbuilder.go
+++ b/modules/blockbuilder/blockbuilder.go
@@ -452,16 +452,16 @@ outer:
 			processedRecords++
 			lastRec = rec
 
-			if maxBytesInMemory > 0 && writer.TotalSize() >= maxBytesInMemory {
+			if totalSize := writer.TotalSize(); maxBytesInMemory > 0 && totalSize >= maxBytesInMemory {
 				level.Debug(b.logger).Log(
 					"msg", "memory threshold reached, flushing mid-cycle",
 					"partition", ps.partition,
 					"timestamp", rec.Timestamp,
-					"total_size", writer.TotalSize(),
+					"total_size", totalSize,
 				)
 				span.AddEvent("mid-cycle flush triggered", trace.WithAttributes(
 					attribute.Int64("maxBytesInMemory", int64(maxBytesInMemory)),
-					attribute.Int64("totalSize", int64(writer.TotalSize())),
+					attribute.Int64("totalSize", int64(totalSize)),
 				))
 				metricFlushThresholdReached.WithLabelValues(partLabel).Inc()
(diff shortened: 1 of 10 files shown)


mattdurham:
### Code review

Found 1 issue:

1. `TestLagMetric_NotSetAtStartOfCycle` tests source code text rather than observable behavior — it reads `blockbuilder.go` as a raw string and asserts the literal call `SetPartitionLagSeconds(group, ps.partition, time.Since(rec.Timestamp))` is absent. This will fail on any rename, reformat, or refactor even when the behavior is correct, and it doesn't actually verify the lag metric is set at the right point in the cycle.

https://github.com/grafana/tempo/blob/23ab040b022c5a9e139132e75d4171e4b4d0b851/modules/blockbuilder/blockbuilder_test.go#L1480-L1492

A behavioral replacement could run `consumePartition()` with a mock metric collector, capture the gauge value immediately after the first record is pushed (the `!init` block), and assert it has not yet been set at that point — then assert it is set after the cycle completes.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

modules/blockbuilder/blockbuilder.go
```diff
@@ -445,19 +451,22 @@ outer:
...
-					attribute.Int64("maxBytesPerCycle", int64(maxBytesPerCycle)),
-					attribute.Int64("consumedBytes", int64(consumedBytes))),
-				)
-				break outer
+				span.AddEvent("mid-cycle flush triggered", trace.WithAttributes(
+					attribute.Int64("maxBytesInMemory", int64(maxBytesInMemory)),
+					attribute.Int64("totalSize", int64(totalSize)),
+				))
+				metricFlushThresholdReached.WithLabelValues(partLabel).Inc()
+				if err := writer.FlushAndReset(ctx, b.reader, b.writer, b.compactor); err != nil {
```
mdisibio, line 467: The intent of the max cycle limit, was to prevent very long/potentially fragile flush cycles from occuring.  I.e. going a long time without committing.  So I think this PR is doubling down on that which is what we don't want, because instead of committing to kafka after reading the default 5GB, it will go even further without a commit?

This loop should be close to what we wanted  - Is the issue is that the final `return lastRec.Timestamp, offset.At, nil` not working like expected?   If we break part way due to the size limit, I would expect that it doesn't wait the full cycle.  It should be just the remainder after the last record?  Or if we need to return something else like `continue=true` to repump the loop without waiting -- I think that would be a better approach than, than the current flush and reset.

PR #6797, opened 2026-03-25
live-store: parallelize WAL block completion during shutdown

mattdurham:
## Problem

Tempo pods with many tenants can take >5 minutes to shut down, exceeding the Kubernetes grace period. The root cause is that `globalCompleteLoop` only runs 2 workers. When there are many tenants each with WAL blocks to complete, these workers cannot drain all outstanding blocks before the pod is forcibly terminated, resulting in unnecessary replay on the next restart.

## Solution

Add a `parallelShutdownComplete()` method to `LiveStore` that is called at the beginning of `stopping()`, before the `holdAllBackgroundProcesses` guard. This method:

- Iterates all tenants and collects every WAL block that still needs to be completed
- Fans out completion using an `errgroup` with a configurable semaphore (default 16 goroutines)
- Logs a summary of how many blocks were completed vs. failed

`completeBlock()` is already idempotent — it re-checks `walBlocks` under `blocksMtx` before doing any work — so concurrent calls are safe. Blocks that fail to complete will be replayed on the next restart, which is the existing behavior.

Two dead `ticker`/`timeout` variable allocations that were present in `stopping()` but never consumed were also removed.

## Configuration

New field: `shutdown_complete_concurrency` (YAML tag `shutdown_complete_concurrency`, CLI flag `--<module>.shutdown-complete-concurrency`)

- Type: `int`
- Default: `16`
- Validation: must be `> 0`

## Changes

- `modules/livestore/config.go` — new `ShutdownCompleteConcurrency` field with default, YAML tag, CLI registration, and validation
- `modules/livestore/config_test.go` — two new validation test cases for the field
- `modules/livestore/live_store.go` — `parallelShutdownComplete()` method, call site in `stopping()`, removed dead allocations, added `errgroup` import
- `modules/livestore/live_store_test.go` — `TestShutdownParallelCompletion`, `TestShutdownParallelCompletionWithEncodingError`; updated replay assertions (`walBlocks` -> `completeBlocks` after clean shutdown); fixed misleading comments

## Test Plan

- [ ] `go test ./modules/livestore/...` passes
- [ ] New tests `TestShutdownParallelCompletion` and `TestShutdownParallelCompletionWithEncodingError` pass
- [ ] Existing replay test assertions still pass with corrected expectations
- [ ] Manual / integration: verify shutdown completes within grace period on a cluster with many tenants
- [ ] Verify `shutdown_complete_concurrency` config field is parsed and validated correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/modules/livestore/live_store.go
+++ b/modules/livestore/live_store.go
@@ -28,6 +28,7 @@ import (
 	"github.com/twmb/franz-go/pkg/kadm"
 	"github.com/twmb/franz-go/pkg/kgo"
 	"go.opentelemetry.io/otel/attribute"
+	"golang.org/x/sync/errgroup"
 )
 
 const (
@@ -399,6 +400,8 @@ func (s *LiveStore) stopping(error) error {
 	err := services.StopAndAwaitTerminated(context.Background(), s.reader)
 	if err != nil {
 		level.Warn(s.logger).Log("msg", "failed to stop reader", "err", err)
+		// If the reader fails to stop cleanly, we skip WAL completion and return.
+		// Pending WAL blocks will be replayed on the next startup.
 		return err
 	}
 
@@ -414,15 +417,16 @@ func (s *LiveStore) stopping(error) error {
 		level.Warn(s.logger).Log("msg", "failed to remove shutdown marker", "path", shutdownMarkerPath, "err", err)
 	}
 
+	// Complete all pending WAL blocks in parallel. This runs unconditionally, including in
+	// test mode with holdAllBackgroundProcesses=true (where it is the *only* completion path).
+	// In production mode, globalCompleteLoop workers may also race to complete some of the
+	// same blocks — this is safe because completeBlock() re-checks walBlocks existence under
+	// blocksMtx before finalizing (idempotent); the second caller no-ops.
+	s.parallelShutdownComplete()
+
 	if s.cfg.holdAllBackgroundProcesses { // nothing to do
 		return nil
 	}
-
-	ticker := time.NewTicker(100 * time.Millisecond)
-	defer ticker.Stop()
-
-	timeout := time.NewTimer(s.cfg.InstanceCleanupPeriod)
-	defer timeout.Stop()
 
 	s.stopAllBackgroundProcesses()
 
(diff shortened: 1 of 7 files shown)


modules/livestore/live_store.go
```diff
@@ -668,6 +672,58 @@ func (s *LiveStore) cutAllInstancesToWal() {
...
+	if len(items) == 0 {
+		return
+	}
+
+	level.Info(s.logger).Log("msg", "parallel shutdown completion starting", "blocks", len(items), "concurrency", s.cfg.ShutdownCompleteConcurrency)
+
+	// s.ctx is guaranteed live here: s.cancel() is called only inside
+	// stopAllBackgroundProcesses(), which runs after this function returns.
+	g, ctx := errgroup.WithContext(s.ctx)
+	g.SetLimit(s.cfg.ShutdownCompleteConcurrency)
```
javiermolinar, line 706: What about not adding a new config and use an heuristic?

min(pendingBlocks, maxParallel)

modules/livestore/live_store.go
```diff
@@ -668,6 +672,58 @@ func (s *LiveStore) cutAllInstancesToWal() {
...
+// the next startup via WAL replay.
+func (s *LiveStore) parallelShutdownComplete() {
+	type work struct {
+		inst    *instance
+		blockID uuid.UUID
+	}
+
+	// Collect all pending WAL block IDs under a read lock per instance.
+	instances := s.getInstances()
+	var items []work
```
javiermolinar, line 688: items is a bit meaningless name

modules/livestore/live_store.go
```diff
@@ -414,16 +417,17 @@ func (s *LiveStore) stopping(error) error {
 		level.Warn(s.logger).Log("msg", "failed to remove shutdown marker", "path", shutdownMarkerPath, "err", err)
 	}
 
+	// Complete all pending WAL blocks in parallel. This runs unconditionally, including in
+	// test mode with holdAllBackgroundProcesses=true (where it is the *only* completion path).
+	// In production mode, globalCompleteLoop workers may also race to complete some of the
+	// same blocks — this is safe because completeBlock() re-checks walBlocks existence under
+	// blocksMtx before finalizing (idempotent); the second caller no-ops.
+	s.parallelShutdownComplete()
```
javiermolinar, line 425: Should the completition workers be stopped before the parallelShutdown starts? Right now we can have a race condition between this method and the instance.CompleteBlock

PR #6812, opened 2026-03-27
enhancement(livestore): dedicated lock for headblock

zhxiaogg:
**What this PR does**:
- added a dedicated lock for the headblock, to reduce potential read/write contentions.

Fixes N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/livestore/instance.go
+++ b/modules/livestore/instance.go
@@ -114,12 +114,15 @@ type instance struct {
 	completeBlockEncoding  encoding.VersionedEncoding
 	completeBlockLifecycle completeBlockLifecycle
 
-	// Block management
+	// headBlockMtx guards headBlock and lastCutTime; blocksMtx guards walBlocks and
+	// completeBlocks. When both are held, acquire headBlockMtx first.
+	headBlockMtx sync.RWMutex
+	headBlock    common.WALBlock
+	lastCutTime  time.Time
+
 	blocksMtx      sync.RWMutex
-	headBlock      common.WALBlock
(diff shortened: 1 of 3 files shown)


modules/livestore/instance.go
```diff
@@ -346,23 +360,11 @@ func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
...
-	if i.headBlock == nil {
-		err := i.resetHeadBlock()
-		if err != nil {
-			return err
-		}
-	}
-
 	tr := &tempopb.Trace{
 		ResourceSpans: liveTrace.Batches,
 	}
```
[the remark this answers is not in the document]
zhxiaogg: My understanding is this is the single only writer per each tenant. And there is no contention from the query side, since they only read flushed pages.

modules/livestore/instance.go
mdisibio: I think what we might need to do is to eliminate this loop and create a new method `i.headblock.WriteAndFlush(slice of traces)`, which could create the page unlocked (new mini parquet file on disk), then only lock the block internally to add the page to the list of flushes.  What's happening now is that writeheadBlock is locking the headblock entirely while serializing the contents to the parquet writer.

It might also be that we just need to split the head block out to its own mutex.  Right now to access the headblock we must take `blocksMtx` which is locking all blocks, right?

zhxiaogg: There is no more lock in writeHeadBlock after this change, and the lock is required for the later flush only. I think it's effectively what you're describing?

Issue #5756, opened 2025-10-15
TraceQL: support division (/) operator for ratio-like queries


ppcano:
I’m not sure if this is unexpected behavior or a missing feature.

According to the [TraceQL Metrics design proposal](https://github.com/grafana/tempo/blob/main/docs/design-proposals/2023-11%20TraceQL%20Metrics.md#arithmetic) , TraceQL is expected to support the division `/` operation:

> Arithmetic operations between time-series can be used to calculate things like error rates.
Find the 5xx error rate for an endpoint:
```
({ span.http.path = "/myapi" && span.http.status_code >= 500 } | rate())  
  /  
({ span.http.path = "/myapi" | rate())  
```

Ratio-like queries (`numerator / denominator`) are used to calculate success/error event ratios.
When I tried this syntax, it appears that the `/` operator is currently not supported.

<img width="1052" height="215" alt="Image" src="https://github.com/user-attachments/assets/56ec2deb-6ae9-427a-b090-8d7290ff3c8f" />

**Describe alternatives you've considered**

I needed this functionality for an alert rule. Not ideal but I was able to work around using Math expressions:

<img width="884" height="923" alt="Image" src="https://github.com/user-attachments/assets/efab37f8-7f20-44b2-b248-40dbe375c5d5" />

**Additional context**

Related issue: [/grafana/grafana/issues/107357-Trace Metric Visualisation: TraceQL metric timeseries don't appear to be a valid source for Math expressions](https://github.com/grafana/grafana/issues/107357)

joe-elliott:
We would like to add math expressions to TraceQL metrics, but we have not yet found the time. We are currently more focused on performance then additional TraceQL features, but I would like to see this eventually as well.

Expressions in Grafana are the best workaround I'm aware of for now.

Syedowais312:
Hi! @joe-elliott  I'd like to work on this issue.

I'm still exploring the Tempo metrics pipeline code and understanding how metric arithmetic is currently handled, but I would love to take this on if you're open to it.

Before I start, could you please confirm whether this feature is something you would like to support now, and if there are any architectural directions or constraints I should follow?

If you give me the green light, I’m happy to start digging deeper into the parser, planner, and execution logic and propose a design before implementing it.

Thanks! 🙌

joe-elliott:
[...]

PR #6866, opened 2026-04-07
[Feature] Support Math in TraceQL Metrics

ruslan-mikhailov:
## What this PR does:
It adds support for binary arithmetic operations (`+`, `-`, `*`, `/`) between TraceQL metrics queries.
This enables queries like error rate ratio:
```
({status=error} | rate()) / ({} | rate())
```

I force using parentheses to have clear boundaries between sub-queries, to not have misunderstandings in queries

This PR does not support operations with number like
```
1000 * ({} | rate())
```
, but lays a foundation to add it in a separate PR. **UPD** branch for scalars is created and waits for current PR's approval.

## Examples
Error rate

image

Share of resource.service.name

image

A query with filtration

image

Error rate by service

image

### Idea
When received a Math query, Tempo passes it through read path as-is without changes. On querier/live-store it then is divided into sub-queries to fetch and process independently. For example, `({status=error} | rate()) / ({} | rate())` becomes `({status=error} | rate())` and `({} | rate())`. Processing goes independently through L1 to L3. On level 3 processing step, it evaluates math expression using all the query fragments. We do not form independent jobs for each sub-query, because keeping them together allows us to better utilise open descriptors, OS caches etc.

Alternative approaches we considered is a single-pass execution. With this approach, we would form a storage request for all the sub-queries and then we could route resulted spans into sub-query pipelines. However, on benchmarks, the first approach showed better results.

To help you to understand data process flow, I recommend starting review with test cases in `pkg/traceql/engine_metrics_test.go`, they explain well how data are processed and transformed.

### Implementation
`firstStageElement` is an interface that combines all the operations in L1, L2 and L3 for metrics. I separated it into 2 logically different interfaces: `spanProcessor` for processing spans (L1) and `seriesProcessor` for series (L2 and L3). While `BatchSpanProcessor` still needs to be a map, because L1 needs to know how to route a span, for `seriesProcessor` it can determine under the hood based on __query_fragment label we provide in L1. `map[string]seriesProcessor` then becomes `seriesProcessor` and L2/L3 processing in engine_metrics.go not affected.

As Math execution can be done only after L1-3 processing, it lays well to secondStage. I introduced `mathExpression` that reflects structure of the math expression and satisfies secondStage interface.

### Rollout problem
If we keep __query_fragment logic for all queries, then during rollout when only part of tempo has new version, it can bring inconsistent results. If all but 1 querier updated, final result would drop results from that querier as it won't have __query_fragment; if frontend is not updated, then it won't drop __query_fragment label, and it will end up in response.

To solve this problem, I changed the logic: when only one sub-query, it won't add __query_fragment. When processed in L2/3 and math, it will takes first and only result without looking at the label. This will also work for queries like `({} | rate()) + ({} | rate())`.

Commit with this change: https://github.com/grafana/tempo/pull/6866/changes/adfde7a13a131f4a5c1ba26d936f3830f1bf5cb4#diff-0a82119641a37adbdfe820b3dd67e18f6c1a6af8a5f72e88022206ae4449409a

## Testing
The main test that was used to verify the feature is `pkg/traceql/engine_metrics_test.go`
Added simple integration tests for e2e smoke regression testing.
Additionally, manually tested that results make sense.

### Semi-automated testing
To verify e2e behaviour, I wrote a python script targeting `example/docker-compose/single-binary`. It validates that server-side arithmetic between metric queries returns correct results for both range and instant queries.

**Approach:** Each test fetches the combined math expression from the API, then independently fetches each sub-query and computes the expected result client-side. Results are compared with 5% tolerance.

**What's tested**

- **All operator × function combinations** — every metric function (rate, count, sum, max, avg, p95, histogram) combined with all four arithmetic operators (+, -, *, /) against count and against itself
- **No-data handling** — when one or both sides return empty results: +/- preserve the non-empty side, */÷ yield zero
- **Different selectors** — math across partially overlapping datasets (filtered left, unfiltered right)
- **Step granularity** — range queries at 1s, 1m, 5m, and 15m steps produce correct results
- **Commutativity** — A+B == B+A and A*B == B*A for multiple function pairs
- **Negative values** — min - max correctly produces negative results
- **Filters** — both post-expression filters (e.g. `> 5s`) and inline per-operand filters (e.g. `> 10ms`) applied to math results
- **Complex expressions** — multi-operand formulas like `(sum+count)/rate` and `(rate+count)*(max-min)`
- **Group-by** — math between per-label grouped series and scalars, and between two grouped series, across all four operators
- **Group-by + filters combined** — grouped math with a post-expression threshold
- **Stability** — repeated identical queries return consistent results (10 runs)

## Which issue(s) this PR fixes
https://github.com/grafana/tempo/issues/6417
https://github.com/grafana/tempo/issues/5756

## Checklist
- [ ] Tests updated
- [ ] Documentation added


pkg/traceql/lenient_extract.go
```diff
@@ -30,8 +30,14 @@ func ExtractFetchRequest(query string) *FetchSpansRequest {
...
-	req := &FetchSpansRequest{AllConditions: true}
-	expr.extractConditions(req)
+	requests := expr.extractConditions(FetchSpansRequest{AllConditions: true})
+
+	// Use the first sub-query request.
+	var req FetchSpansRequest
+	for _, v := range requests {
+		req = v
+		break
+	}
```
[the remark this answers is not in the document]
ruslan-mikhailov: It is used only for tags search and is out of PR's scope

pkg/traceql/engine_metrics_average.go
```diff
@@ -23,7 +23,11 @@ type averageOverTimeAggregator struct {
 	mode       AggregateMode
 }
 
-var _ firstStageElement = (*averageOverTimeAggregator)(nil)
+var (
+	_ firstStageElement = (*averageOverTimeAggregator)(nil)
+	_ spanProcessor     = (*averageOverTimeAggregator)(nil)
+	_ seriesProcessor   = (*averageOverTimeAggregator)(nil)
+)
```
ruslan-mikhailov, line 30: I would also split such structs into spanProcessor and seriesProcessor, but left as is to not overblow the PR.

pkg/traceql/ast_metrics_math.go
```diff
@@ -0,0 +1,290 @@
...
+			continue
+		}
+
+		// Build new key without __query_fragment and __name__
+		key := SeriesMapKey{}
+		j := 0
+		for i := range smk {
+			if smk[i].Name == internalLabelQueryFragment {
+				continue
+			}
```
ruslan-mikhailov, line 138: Ideally, we can pass internal labels in proto inside InternalLabels map[string]Label inside of TimeSeries, and just drop it in L3 (frontend). This would simplify code here and for other features like histogram or avg_over_time.

It will be done in a separate PR

pkg/traceql/ast_metrics.go
```diff
@@ -9,17 +9,82 @@ import (
...
+				fragments[fragmentKey] = append(fragments[fragmentKey], ts)
+				break
+			}
+		}
+	}
+	for k, v := range fragments {
+		if p, ok := b[k]; ok {
+			p.observeSeries(v)
+		}
+	}
```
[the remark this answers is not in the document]
ruslan-mikhailov: Oh, I again completely forgot about it! Added a fix for this issue. Now for single queries it won't return query fragment label

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
   Adds `max_cardinality_per_label` per per-tenant override and new metrics to estimate per label cardinality demand estimate.
 * [FEATURE] Add an extension mechanism for per-tenant overrides [#6758](https://github.com/grafana/tempo/pull/6758) (@stoewer)
 * [FEATURE] Single-binary mode: push distributor local ingest directly to live-store and metrics-generator without Kafka [#6729](https://github.com/grafana/tempo/pull/6729) (@javiermolinar)
+* [FEATURE] Support arithemtic operations in TraceQL Metrics [#6866](https://github.com/grafana/tempo/pull/6866) (@ruslan-mikhailov)
```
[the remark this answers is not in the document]
ruslan-mikhailov: /facepalm
fixed

pkg/traceql/engine_metrics.go
```diff
@@ -953,29 +968,31 @@ func (e *Engine) CompileMetricsQueryRangeNonRaw(req *tempopb.QueryRangeRequest,
...
-	// as we are not sharding them now in lower layers.
-	if metricsSecondStage != nil && mode == AggregateModeFinal {
-		metricsSecondStage.init(req)
-		mfe.metricsSecondStage = metricsSecondStage
+	// Only run second stage in final mode
+	var secondStage secondStageElement
+	if ss := expr.MetricsSecondStage(); mode == AggregateModeFinal {
+		ss.init(req)
+		secondStage = ss
 	}
```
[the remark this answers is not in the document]
ruslan-mikhailov: Ha, good point! Added a shortcut for flat queries to avoid unnecessary logic.

pkg/traceql/ast_metrics_math.go
```diff
@@ -0,0 +1,291 @@
...
+	if len(a) == 0 {
+		return b
+	}
+	if len(b) == 0 {
+		return a
+	}
+	result := make([]Exemplar, 0, len(a)+len(b))
+	result = append(result, a...)
+	result = append(result, b...)
+	return result
```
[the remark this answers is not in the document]
ruslan-mikhailov: Good point! But better to distribute exemplars in engine

pkg/traceql/ast.go
```diff
@@ -65,22 +88,35 @@ func NeedsFullTrace(e ...Element) bool {
 }
 
 func (r *RootExpr) NeedsFullTrace() bool {
```
ruslan-mikhailov: If at least one sub-query has NeedsFullTrace, it will require all sub-queries to request full trace. This should be per sub-query, not overall.

pkg/traceql/ast_metrics_math.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	result := make(SeriesSet, len(target))
+
+	// pre-allocate array once to avoid multiple smaller allocations
+	var valuesLen int
+	for _, t := range target {
+		valuesLen += len(t.Values)
+	}
+	buf := make([]float64, valuesLen)
```
ruslan-mikhailov, line 179: Maybe that's overkill? But benchmark showed nice boost from this optimisation. It allocates one big array instead of allocating multiple smaller arrays.

pkg/traceql/ast_metrics_math.go
```diff
@@ -0,0 +1,291 @@
...
+	return result
+}
+
+func applyArithmeticOp(op Operator, lhs, rhs float64) float64 {
+	if math.IsNaN(lhs) {
+		lhs = 0
+	}
+	if math.IsNaN(rhs) {
+		rhs = 0
+	}
```
ruslan-mikhailov: This violates IEEE-754. I think, we should comply with it. In that case when a query has no spans within bucket's time range, all operations for this bucket will also result in empty bucket (NaN)

wdyt?

pkg/traceql/ast_metrics.go
```diff
@@ -547,16 +629,23 @@ func (c ChainedSecondStage) validate() error {
...
 	return input
 }
 
+func (c ChainedSecondStage) separator() string {
+	if len(c) == 0 {
+		return ""
+	}
+	return c[0].separator()
+}
+
```
ruslan-mikhailov, line 650: I don't quite remember, why I returned separator. I think that was to make String for expression. But I think after refactoring it's not necessary again. 

I need to try to remove it

pkg/traceql/ast.go
ruslan-mikhailov: Fixed

pkg/traceql/engine_metrics.go
```diff
@@ -1036,139 +1048,166 @@ func (e *Engine) CompileMetricsQueryRange(req *tempopb.QueryRangeRequest, opts .
...
+	// distribute exemplars for each sub-query
+	exemplarsCapacity := int(req.Exemplars)
+	if req.Exemplars > maxExemplars {
+		level.Warn(log.Logger).Log("msg", "capping exemplars to safety limit", "requested", req.Exemplars, "cap", maxExemplars)
+		exemplarsCapacity = int(maxExemplars)
+	}
+	exemplars := exemplarsCapacity / len(expr.Pipeline)
+	if exemplarsCapacity > 0 && exemplars == 0 {
+		exemplars = 1 // at least one per sub-query when exemplars are requested
+	}
```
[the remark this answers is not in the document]
ruslan-mikhailov: This is why we have exemplarsCapacity

pkg/traceql/ast.go
```diff
@@ -31,13 +31,36 @@ type typedExpression interface {
...
+	Pipeline           map[string]Pipeline        // spanset filters
+	BatchSpanProcessor map[string]spanProcessor   // span processing
+	SeriesProcessor    map[string]seriesProcessor // time series processing
+	expression         *mathExpression            // query expression structure (second stage)
 	Hints              *Hints
 	OptimizationCount  int
 }
 
+// SinglePipeline returns the pipeline and span processor when there is exactly
+// one sub-query. Returns false for math expressions (multiple pipelines).
```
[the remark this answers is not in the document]
ruslan-mikhailov: Done

pkg/traceql/lenient_parser.go
```diff
@@ -35,8 +35,22 @@ func ParseLenient(s string) (*RootExpr, error) {
...
+		removeBareAttrsInPipeline(p)
+		newKey := p.String()
+		if sp, ok := root.BatchSpanProcessor[k]; ok {
+			newKey += " | " + sp.(Element).String()
+		}
+		newPipelines[newKey] = p
+		keyMap[k] = newKey
+	}
+	root.Pipeline = newPipelines
+	root.expression = root.expression.rewriteKeys(keyMap)
```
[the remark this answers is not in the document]
ruslan-mikhailov: Added just in case

pkg/traceql/test_examples.yaml
```diff
@@ -180,6 +180,12 @@ valid:
   - '{} | rate() != 0'
   - '{} | rate() > -3'
   - '{} | rate() by (span.client_ip) > 10 with(sample=0.1)'
+  - '({} | rate()) > 10'
+  - '({} | max_over_time(duration)) > 30'
+  # math expressions with second stage pipeline
+  - '({} | rate()) + ({} | rate()) | topk(5)'
+  - '({} | rate()) + ({} | rate()) > 10'
+  - '({} | rate()) + ({} | rate()) | topk(5) > 10'
```
ruslan-mikhailov, line 194: I need to add test cases for all operations + invalid like `{} | rate() + {} | rate()` and unsupported like `%`

pkg/traceql/lenient_extract.go
ruslan-mikhailov, line 40: I did not support Math for Tag Values search and in this PR did it only for the first sub-query.

I'm thinking if Grafana could send us not the whole query, but only sub-query where the cursor is. For example, 
```
({ .foo = "bar" } | rate()) + ({ .b = "val" && .v=| } | rate())
                                                  ^ cursor is here
```
it could send `{ .b = "val" && .v= } | rate()`. That would make more sense.

Anyways, support will be added in a separate PR

ruslan-mikhailov: Did it here: https://github.com/grafana/grafana/compare/main...ruslan-mikhailov:grafana:feature/parse-tempo-math-expr

With 
```
({resource.host.name=} | rate()) + ({ resource.k6.1Fn7HcicetM8qkG= } | rate())
                     ^                                            ^
                cursor position 1                             cursor position 2
```
for cursor position 1, it sends `{resource.host.name=} | rate()` into request, and `{ resource.k6.1Fn7HcicetM8qkG= } | rate()` for position 2.

I will open a PR after this PR with math syntax proposal gets merged

pkg/traceql/engine.go
```diff
@@ -27,30 +29,49 @@ func NewEngine() *Engine {
 	return &Engine{}
 }
 
-func Compile(query string) (*RootExpr, SpansetFilterFunc, firstStageElement, secondStageElement, *FetchSpansRequest, error) {
+// Compile compiles given query. It does not support math expressions.
+func Compile(query string) (*RootExpr, Pipeline, SpansetFilterFunc, *FetchSpansRequest, error) {
```
mdisibio: nit: I think I want to rename this function now.  Maybe `CompileSearch()` ?   Based on the current usage: only in tests and inside ExecuteSearch, this matches.  Would also align with `CompileMetricsQueryRange()`.

pkg/traceql/engine_metrics.go
```diff
@@ -1266,7 +1305,61 @@ func lookup(needles []Attribute, haystack Span) Static {
...
+		spansDeduped += sd
+	}
+	return bytes, spansTotal, spansDeduped
+}
+
+func (e batchMetricsEvaluator) Results() SeriesSet {
+	merged := make(SeriesSet)
+	for q, eval := range e {
+		for _, v := range eval.Results() {
+			if len(e) > 1 { // no need for fragment filtration
```
mdisibio: nit:  We could check this first and return early when 1 fragment without allocating the map.

pkg/traceql/engine_metrics.go
```diff
@@ -1266,7 +1305,61 @@ func lookup(needles []Attribute, haystack Span) Static {
...
+				v.Labels = v.Labels.Add(Label{Name: internalLabelQueryFragment, Value: NewStaticString(q)})
+			}
+			merged[v.Labels.MapKey()] = v
+		}
+	}
+	return merged
+}
+
+type metricsEvaluator struct {
```
mdisibio, line 1330: We need an interface assertion for this type now.

pkg/traceql/engine_metrics_average.go
mdisibio: Agree, let's think about that later. We've implemented some things as single and split structs, and neither is a clear winner to me yet.

pkg/traceql/engine_metrics_compare.go
```diff
@@ -392,6 +397,9 @@ func (b *BaselineAggregator) Combine(ss []*tempopb.TimeSeries) {
 				metaType = l.Value.GetStringValue()
 			case internalLabelError:
 				err = l.Value.GetStringValue()
+			case internalLabelQueryFragment:
+				b.queryFragment = l.Value.GetStringValue()
```
mdisibio: Oh my, I wasn't considering how the `compare()` might be involved. If this is working, then seems simple enough. But an argument could be made for making this an invalid query for now, if there are any hurdles or the use case isn't clear (to note: it is not clear to me 😄)

pkg/traceql/expr.y
```diff
@@ -347,22 +351,43 @@ metricsFilterOperation:
...
 // Metrics Second Stage Pipeline (chains of second stage elements)
 // **********************
 metricsSecondStagePipeline:
-    PIPE metricsSecondStage                              { $$.Append($2, " | ") }
-  | metricsFilter                                        { $$.Append($1, " ") }
-  | metricsSecondStagePipeline PIPE metricsSecondStage   { $$ = $1; $$.Append($3, " | ") }
-  | metricsSecondStagePipeline metricsFilter             { $$ = $1; $$.Append($2, " ") }
+    PIPE metricsSecondStage                              { $$ = ChainedSecondStage{$2} }
+  | metricsFilter                                        { $$ = ChainedSecondStage{$1} }
+  | metricsSecondStagePipeline PIPE metricsSecondStage   { $$ = append($1, $3) }
```
mdisibio, line 368: I still don't love how the separators are on the children, and would prefer them to be at these levels in the parents, exactly where the language spec puts them.  Can you walk me through again what the blocker was? We can chat offline.

pkg/traceql/engine_metrics.go
mdisibio, line 1463: There might be a change we need here.  Exemplars are showing on series that seem wrong (but might actually be correct...).  See this screenshot, where the series for `tempo-all` contains an exemplar to trace without it.   It might be correct because technically this trace was included in the second pipeline (`{} | rate()`)...   Let's think about if this is something we can improve.

Image

pkg/traceql/ast_metrics_math.go
```diff
@@ -0,0 +1,291 @@
...
+	// pre-allocate array once to avoid multiple smaller allocations
+	var valuesLen int
+	for _, t := range target {
+		valuesLen += len(t.Values)
+	}
+	buf := make([]float64, valuesLen)
+	var offset int
+
+	for k := range target {
+		l, lOk := getTSMatch(lhs, k)
```
mdisibio: Let's add a note here that we need to think about implementing prometheus-like `on()` and `group_left()` like matching.  I love that the easy case is easy, when RHS is a single series:  rate by service / total rate (no labels).

But I have a feeling we will need to implement the custom matching soon.

pkg/traceql/ast_metrics_math.go
```diff
@@ -0,0 +1,291 @@
...
+		}
+	}
+	return out
+}
+
+func getTSMatch(set SeriesSet, key SeriesMapKey) (TimeSeries, bool) {
+	if s, ok := set[key]; ok {
+		return s, true
+	}
+	if s, ok := set[noLabelsSeriesMapKey]; ok {
```
mdisibio, line 263: Can you think of any issues where the no-labels-series would be the incorrect one?   In my local testing, all seems good.  The case I had in mind was non-existent attributes where they become a label=nil, but they are handled as expected.  A single label=nil is preserved (they are not the "no-labels" series)

BUT here is an edge case when there are multiple labels, stemming from how we drop the nils then.  It seems incorrect, but we match Prometheus...

rate by (service, bar) / rate by (service, foo).    You might expect these to never join because different labels, but since service=non-nil, we will drop the bar=nil and foo=nil. Then they join on service and result is 1.0.  Second screenshot shows it happening in Prom.

Image

Image

ruslan-mikhailov:
\+ tough rebase from latest main

pkg/traceql/engine_metrics.go
ruslan-mikhailov: Good idea! Fixed

pkg/traceql/engine_metrics.go
ruslan-mikhailov: Oh, I missed it, yes. Added

pkg/traceql/engine_metrics_compare.go
ruslan-mikhailov: Oh! Math with compare is actually an invalid query. The changes here are from version when query fragment label was passing for every query. Now, when it sends them only for Math, it is not neccessary.

Removed, added math with compare queries to `test_examples.yaml`

pkg/traceql/ast_metrics_math.go
ruslan-mikhailov: \+

Added TODO to getTSMatch's comment

pkg/traceql/engine_metrics.go
ruslan-mikhailov: It confuses indeed. I'm thinking if we can provide what sub-query an exemplar belongs. In grafana UI, if you query two queries, it's the same confusing behaviour:

image

For now, I do just drop exemplars for fan-out sub-query. Fixed in a separate commit. 
What do you think if we prepare a proper fix/improvement in a separate PR to not block the feature?

pkg/traceql/ast_metrics_math.go
```diff
@@ -0,0 +1,304 @@
...
+		case lFanOut:
+			exemplars = r.Exemplars
+		default:
+			exemplars = l.Exemplars
+		}
+
+		n := min(len(r.Values), len(l.Values))
+		values := buf[offset : offset+n]
+		for j := 0; j < n; j++ {
+			values[j] = applyArithmeticOp(op, l.Values[j], r.Values[j])
```
stoewer, line 202: It can be interesting to move the loop inside `applyArithmeticOp` and maybe have separate loops for each operator, as this could make it easier for the compiler to apply optimizations and use SIMD instructions

ruslan-mikhailov:
\+ one more tough rebase from main

pkg/traceql/ast_metrics_math.go
ruslan-mikhailov: > Can you think of any issues where the no-labels-series would be the incorrect one? 

From user's perspective, if we support 
```
({} | rate() by (A))
/
({} | rate())
```

I think it makes sense to also support

```
({} | rate() by (A,B))
/ 
({} | rate() by (A))
```

knylander-grafana:
Docs: https://github.com/grafana/tempo/pull/7190

pkg/traceql/ast_metrics_math.go
ruslan-mikhailov: I've pushed commit that respects `IEEE-754` and handles NaNs correctly

pkg/traceql/ast_metrics_math.go
ruslan-mikhailov: I wrote a benchmark (locally, not pushed to the branch) and compared two options: loop-switch (current solution) and switch-loop. There are no visible differences in bench results.

go build reported that `applyArithmeticOp` here is inlined.

Discussed offline, I leave the current version for code simplicity.

pkg/traceql/engine.go
ruslan-mikhailov: Agreed to change in a follow-up PR

pkg/traceql/ast_metrics.go
mdisibio: We will revisit this in a separate PR to keep this one moving.

pkg/traceql/ast_metrics_math.go
mdisibio: I think we can revisit this if it becomes a bottleneck, which we haven't seen yet.  I bet in the querier side of things on the raw spans, it would benefit more from that.

pkg/traceql/ast_metrics_math.go
mdisibio: I think we will need to add the `on(x, y, z)` clause like prometheus to go to that next step.  Because of ambiguity where the series for these look the same after dropping nils:
```
({} | rate() by (A,B))    ----> {A non-nil, B == nil}
({} | rate() by (A))      ----> {A non-nil}
```

And we can get this in followup work, not part of this PR.

pkg/traceql/ast_metrics_math.go
```diff
@@ -280,12 +280,6 @@ func mergeExemplars(a, b []Exemplar) []Exemplar {
 }
 
 func applyArithmeticOp(op Operator, lhs, rhs float64) float64 {
-	if math.IsNaN(lhs) {
```
mdisibio: +1 I think this is the safest way to start, honoring NaNs, since it will guarantee correct results. And then we can prioritize work-arounds like `max(<query>, 0)` or something like prometheus `or vector(0)` to let it be fixed in the query if that behavior is desired.

pkg/traceql/test_examples.yaml
ruslan-mikhailov: Added

mdisibio (approved):
This PR is a huge upgrade to TraceQL, and I am very excited for it.  Nice work 👍

PR #6879, opened 2026-04-08
[ingest]: fix partition lag reporting for live groups and rebalance state

zalegrala:
**What this PR does**:

Fixes three bugs in `getGroupLag` / `ExportPartitionLagMetrics` in `pkg/ingest`:

1. **Newly assigned partitions silently omitted.** The original code always passed `DescribedGroup{State: "Empty"}` (no members) to `CalculateGroupLagWithStartOffsets`. That function's empty-group fallback only iterates over previously committed partitions — a newly assigned partition with no committed offset is never included, so its lag gauge stays stale or is never set. Fix: call `DescribeGroups` first; use the real group description when members are present, fall back to manual per-partition computation over `assignedPartitions` (including uncommitted partitions, using start offset) when there are none.

2. **Spuriously large lag after log truncation.** The `lag == -1` sentinel used to gate the start-offset fallback would incorrectly trigger when a valid commit existed but `endOffset - committedOffset < 0`. Fix: replace with an explicit `hasCommit` bool; fall back to start offset only when no commit exists, clamp to zero otherwise.

3. **Stale gauge on missing end offset.** When a partition's end offset was absent from the broker response the old code `continue`d silently, leaving the gauge frozen. Fix: emit `Lag=-1` with `Err` set so callers can distinguish zero lag from a failed measurement.

Note: neither old nor new code reports lag for genuinely unowned partitions (those absent from every instance's `assignedPartitions`). This pre-existing observability gap is tracked as follow-up work.

N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/ingest/metrics.go
+++ b/pkg/ingest/metrics.go
@@ -69,7 +69,7 @@ func ExportPartitionLagMetrics(ctx context.Context, kclient *kgo.Client, log log
 				assignedPartitions := getAssignedActivePartitions()
 				boff.Reset()
 				for boff.Ongoing() {
-					lag, err = getGroupLag(ctx, admClient, partitionClient, group, assignedPartitions)
+					lag, err = getGroupLag(ctx, admClient, partitionClient, group, topic, assignedPartitions)
 					if err == nil {
 						break
 					}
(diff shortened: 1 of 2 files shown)


zalegrala:
Stale.

PR #6903, opened 2026-04-08
Fixes # 6902 Add flag for MCP server

tiffanyfay:
**What this PR does**:
Adds a flag to enable the MCP server instead of just by config.

Fixes #6902 for the 2.10 branch

The OTel LGTM Docker image [currently uses v2.10.3](https://github.com/grafana/docker-otel-lgtm/blob/d16b3991fe51cdfacce4d9ad8e400ced4f216037/docker/Dockerfile#L7). Currently there's no clean way to enable the Tempo MCP server. This PR adds the flag `--frontend.mcp-server.enabled=true`. It also adds the flag to the command line flags doc and the MCP server doc.

This makes it so we can therefore do the following with a future image build of `grafana/otel-lgtm:latest`

From the run command, I could then use the env var `TEMPO_EXTRA_ARGS: "--frontend.mcp-server.enabled=true"` which is used in `run-tempo.sh`: https://github.com/grafana/docker-otel-lgtm/blob/main/docker/run-tempo.sh
```
#!/bin/bash

source ./logging.sh

extra_args=()
if [[ -n "${TEMPO_EXTRA_ARGS:-}" ]]; then
	read -ra extra_args <<<"${TEMPO_EXTRA_ARGS}"
fi
run_with_logging "Tempo ${TEMPO_VERSION}" "${ENABLE_LOGS_TEMPO:-false}" \
	./tempo/tempo --config.file=./tempo-config.yaml "${extra_args[@]}"
```

This was from my own build and run:
```
  lgtm:
    image: tiffanyfay/otel-lgtm:tempo-mcp-2.10.3-apr8-1511
    ports:
      - "3000:3000" # Grafana UI
      - "4317:4317" # OTLP gRPC
      - "4318:4318" # OTLP HTTP
      - "3200:3200"
    volumes:
      - ./grafana/custom-dashboard.json:/otel-lgtm/grafana/conf/provisioning/dashboards/custom/custom-custom-dashboard.json:ro
      - ./grafana/dashboards-provisioning.yaml:/otel-lgtm/grafana/conf/provisioning/dashboards/custom.yaml:ro
    environment:
      GF_DASHBOARDS_DEFAULT_HOME_DASHBOARD_PATH: /otel-lgtm/grafana/conf/provisioning/dashboards/custom/custom-custom-dashboard.json
      TEMPO_EXTRA_ARGS: "--frontend.mcp-server.enabled=true"
```
image

I need to look at what I need to do to add this to the `main` branch as well...

**Checklist**
- [x] Tests updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## main / unreleased
 
+* [ENHANCEMENT] Add flag for enabling MCP server. [#6903](https://github.com/grafana/tempo/pull/6903) (@tiffanyfay)
+
 # v2.10.3
 
 * [SECURITY] S3 SSE-C `encryption_key` is now treated as a secret to prevent it from being exposed in plaintext. Resolves CVE-2026-28377. [#6711](https://github.com/grafana/tempo/pull/6711) (@mattdurham)
--- a/cmd/tempo/app/config.go
+++ b/cmd/tempo/app/config.go
@@ -148,7 +148,7 @@ func (c *Config) RegisterFlagsAndApplyDefaults(prefix string, f *flag.FlagSet) {
 	c.Ingest.RegisterFlagsAndApplyDefaults(util.PrefixConfig(prefix, "ingest"), f)
 	c.BlockBuilder.RegisterFlagsAndApplyDefaults(util.PrefixConfig(prefix, "block-builder"), f)
 	c.Querier.RegisterFlagsAndApplyDefaults(util.PrefixConfig(prefix, "querier"), f)
-	c.Frontend.RegisterFlagsAndApplyDefaults(util.PrefixConfig(prefix, "frontend"), f)
+	c.Frontend.RegisterFlagsAndApplyDefaults(util.PrefixConfig(prefix, "query-frontend"), f)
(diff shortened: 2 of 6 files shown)


javiermolinar (commented):
I believe this is not needed. To enable the MCP server you just need to to add these lines to the tempo.yaml config


https://github.com/grafana/tempo/blob/88c7a4f7ff53f26ae76cdd19b70b3fb4f7920c61/example/docker-compose/single-binary/tempo.yaml#L49-L52


It is defined here:

https://github.com/grafana/tempo/blob/88c7a4f7ff53f26ae76cdd19b70b3fb4f7920c61/modules/frontend/config.go#L51

tiffanyfay:
@javiermolinar  So I suppose I should have removed 
```
	cfg.MCPServer = MCPServerConfig{
		Enabled: false,
	}
```
To not have some redundancy.


Given the reasons not to enable the MCP server by default in Tempo, it also makes sense to me not to enable it by default in the Docker LGTM OTel image, which is what adding it to the image's config would do.

joe-elliott (approved):

PR #6932, opened 2026-04-10
chore: improve observability with the livestore startup/shutdown

zhxiaogg:
**What this PR does**:
- added more fine grained logs over the livestore shutdown/startup process
- record the partition lag regardless of record state

Fixes N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/livestore/live_store_background.go
+++ b/modules/livestore/live_store_background.go
@@ -76,6 +76,10 @@ func (s *LiveStore) runInBackground(fn func()) {
 }
 
 func (s *LiveStore) globalCompleteLoop(idx int) {
+	level.Info(s.logger).Log("msg", "starting completing loop", "index", idx)
+	defer func() {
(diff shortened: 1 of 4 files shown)


modules/livestore/live_store_background.go
```diff
@@ -76,6 +76,10 @@ func (s *LiveStore) runInBackground(fn func()) {
 }
 
 func (s *LiveStore) globalCompleteLoop(idx int) {
+	level.Info(s.logger).Log("msg", "starting completing loop", "index", idx)
```
zalegrala, line 79: Should this be debug?  I expect this will be quite noisy.

zhxiaogg: oh, those will only print at the livestore startup/shutdown, not for each completion job if that's what you mean.

zalegrala: Yes, that was my concern.

zalegrala (approved):
Looks reasonable to me.

PR #6947, opened 2026-04-13
fix: metrics duration check by moving it before request aligment

javiermolinar:
**What this PR does**:

The validation of the max metrics duration is currently broken because is done before the request aligment:

1. User select a max valid range in Grafana ie 48h
2. The request is aligned. If the step is 15m then the duration can becomes either 48h:15m or 48h:30m
3. An error is returned

Since this config is user facing its missleading to return an error here. To address it I have moved the validation to the handler before any aligment is done. It also care about multitenant queries as it was done in the sharder (always happens before the multitenant fan out middleware)

Fixes #<issue number>

**Checklist**
- [X] Tests updated
- [ ] Documentation added


javiermolinar:
Closing this to favor https://github.com/grafana/tempo/pull/7170

PR #6967, opened 2026-04-14
enhancement: enable async parquet read mode for WAL completion path

zhxiaogg:
**What this PR does**:
- Enabled parquet ReadModeAsync when opening WAL files during block completion. This allows parquet-go to issue concurrent ReadAt calls, improving I/O throughput. OS page cache handles buffering for local disk so no userspace BufferedReaderAt is needed.

Based on the test result, we saw slightly improvements on the WAL completion duration. Didn't find major issues with cpu/meme/query etc.

Fixes N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/encoding/vparquet4/wal_block.go
+++ b/tempodb/encoding/vparquet4/wal_block.go
@@ -33,6 +33,8 @@ var _ common.WALBlock = (*walBlock)(nil)
 // likely the best candidate is some fraction of max trace size per tenant.
 const defaultRowPoolSize = 100000
 
+const EnvVarWALAsyncIO = "TEMPO_WAL_ASYNC_IO"
+
 // completeBlockRowPool is used by the wal iterators and complete block logic to pool rows
 var completeBlockRowPool = newRowPool(defaultRowPoolSize)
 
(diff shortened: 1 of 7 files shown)


tempodb/encoding/vparquet4/wal_block.go
```diff
@@ -240,6 +242,9 @@ func (w *walBlockFlush) file(ctx context.Context) (*pageFile, error) {
 		parquet.SkipPageIndex(true),
 		parquet.FileSchema(parquetSchema),
 	}
+	if os.Getenv(EnvVarWALAsyncIO) == "1" {
+		o = append(o, parquet.FileReadMode(parquet.ReadModeAsync))
```
mattdurham: Any idea on the concurrency of this?

zhxiaogg: ah good point, this is a existing pattern at the read side I believe, let me do some investigation.

zhxiaogg: there will be goroutines for each column. I don't see ways to limit that so far.

mattdurham: So likely has a reasonable cap then.

mapno (approved):

PR #6971, opened 2026-04-15
[Bugfix] Live-store: Cut head block if it exceeds MaxBlockBytes

ruslan-mikhailov:
**What this PR does**: when cutting live traces in `cutIdleTraces` it could be that block can exceed MaxBlockBytes, especially when there are big live traces, which can result in ignoring the config. This PR preliminary cuts head block for such case.

About this check in tests:
```go
		// block size estimation can be x5 off, so we check that that block size at least makes sense
		assert.LessOrEqual(t, blk.DataLength(), inst.Cfg.MaxBlockBytes*5,
			"WAL block %s exceeds MaxBlockBytes: %d > %d", id, blk.DataLength(), inst.Cfg.MaxBlockBytes)
```

Apparently `estimateMarshalledSizeFromTrace` can be off x5 times from real size. I checked for vparquet4 and 5 for a testing trace generated by `test.MakeTraceWithSpanCount(50, 100, id)`:

vp5:
actual size: 21'133'126
estimated size: 5'242'880
vp4:
actual size: 21'133'669
estimated size: 5'242'880

Therefore, in the test I just do a sanity check. Without the fix the test fails and catches the bug.
The problem with `estimateMarshalledSizeFromTrace` will be fixed separatly.

Fixes #<issue number>

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/modules/livestore/instance.go
+++ b/modules/livestore/instance.go
@@ -331,7 +331,7 @@ func countSpans(trace *tempopb.Trace) int {
 	return count
 }
 
-func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
+func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) ([]uuid.UUID, error) {
 	_, span := tracer.Start(ctx, "instance.cutIdleTraces",
 		oteltrace.WithAttributes(attribute.String("tenant", i.tenantID)))
 	defer span.End()
@@ -350,23 +350,45 @@ func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
 	span.SetAttributes(attribute.Int("traces_to_cut", len(tracesToCut)))
 
 	if len(tracesToCut) == 0 {
-		return nil
+		return nil, nil
 	}
 	// Sort by ID
 	sort.Slice(tracesToCut, func(i, j int) bool {
 		return bytes.Compare(tracesToCut[i].ID, tracesToCut[j].ID) == -1
 	})
-	// Collect the trace IDs that will be flushed
+
+	var cutBlockIDs []uuid.UUID
+
+	// Write traces to head block, cutting when MaxBlockBytes is reached.
 	span.AddEvent("writing traces to head block")
 	for _, t := range tracesToCut {
 		err := i.writeHeadBlock(t.ID, t)
 		if err != nil {
 			span.RecordError(err)
 			span.SetStatus(codes.Error, err.Error())
-			return err
+			return cutBlockIDs, err
 		}
 
 		i.tracesCreatedTotal.Inc()
+
+		// if the head block has reached max block bytes,
+		// cut it and enqueue for completion later
+		i.blocksMtx.Lock()
+		reason := i.shouldCutHead(false)
+		i.blocksMtx.Unlock()
+		if reason&cutReasonMaxBlockBytes != 0 {
+			id, err := i.cutBlocks(ctx, false)
+			if err != nil {
+				span.RecordError(err)
+				return cutBlockIDs, err
+			}
+			if id != uuid.Nil {
+				span.AddEvent("cut oversized block", oteltrace.WithAttributes(
+					attribute.String("blockID", id.String()),
+				))
+				cutBlockIDs = append(cutBlockIDs, id)
+			}
+		}
 	}
 	span.AddEvent("wrote traces to head block")
 
@@ -381,12 +403,10 @@ func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
 		if err != nil {
 			span.SetStatus(codes.Error, err.Error())
 			span.RecordError(err)
-			return err
+			return cutBlockIDs, err
 		}
-
-		return nil
 	}
-	return nil
+	return cutBlockIDs, nil
 }
 
 func (i *instance) writeHeadBlock(id []byte, liveTrace *livetraces.LiveTrace[*v1.ResourceSpans]) error {
@@ -532,11 +552,28 @@ func (i *instance) cutBlocks(ctx context.Context, immediate bool) (uuid.UUID, er
 
 	i.traceSizes.ClearIdle(i.lastCutTime)
 
-	// Final flush
-	err := i.headBlock.Flush()
+	id, err := i.cutHeadLocked()
 	if err != nil {
 		span.SetStatus(codes.Error, err.Error())
 		span.RecordError(err)
+		return uuid.Nil, err
+	}
+
+	span.SetAttributes(
+		attribute.String("blockID", id.String()),
+		attribute.Int64("block_size", int64(i.walBlocks[id].DataLength())),
+	)
+
+	recordBlockCutMetric(reason)
+
+	return id, nil
+}
+
+// cutHeadLocked flushes the head block, moves it to walBlocks, and resets the
+// head block. Caller must hold blocksMtx.
+func (i *instance) cutHeadLocked() (uuid.UUID, error) {
+	err := i.headBlock.Flush()
+	if err != nil {
 		return uuid.Nil, err
 	}
 
@@ -544,21 +581,12 @@ func (i *instance) cutBlocks(ctx context.Context, immediate bool) (uuid.UUID, er
 	blockSize := i.headBlock.DataLength()
 	i.walBlocks[id] = i.headBlock
 
-	span.SetAttributes(
-		attribute.String("blockID", id.String()),
-		attribute.Int64("block_size", int64(blockSize)),
-	)
-
 	level.Info(i.logger).Log("msg", "queueing wal block for completion", "block", id.String(), "size", blockSize)
 
 	err = i.resetHeadBlock()
 	if err != nil {
-		span.RecordError(err)
-		span.SetStatus(codes.Error, err.Error())
 		return uuid.Nil, err
 	}
-
-	recordBlockCutMetric(reason)
 
 	return id, nil
 }
--- a/modules/livestore/instance.go
+++ b/modules/livestore/instance.go
@@ -330,7 +330,7 @@ func countSpans(trace *tempopb.Trace) int {
 	return count
 }
 
-func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
+func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) ([]uuid.UUID, error) {
 	_, span := tracer.Start(ctx, "instance.cutIdleTraces",
 		oteltrace.WithAttributes(attribute.String("tenant", i.tenantID)))
 	defer span.End()
@@ -349,22 +349,45 @@ func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
 	span.SetAttributes(attribute.Int("traces_to_cut", len(tracesToCut)))
 
 	if len(tracesToCut) == 0 {
-		return nil
+		return nil, nil
 	}
 	// Sort by ID
 	sort.Slice(tracesToCut, func(i, j int) bool {
 		return bytes.Compare(tracesToCut[i].ID, tracesToCut[j].ID) == -1
 	})
-	// Collect the trace IDs that will be flushed
+
+	var cutBlockIDs []uuid.UUID
+
+	// Write traces to head block, cutting when MaxBlockBytes is reached.
 	span.AddEvent("writing traces to head block")
 	for _, t := range tracesToCut {
 		err := i.writeHeadBlock(t.ID, t)
 		if err != nil {
 			span.RecordError(err)
-			return err
+			return cutBlockIDs, err
 		}
 
 		i.tracesCreatedTotal.Inc()
+
+		// if the head block has reached max block bytes,
+		// cut it and enqueue for completion later
+		i.blocksMtx.Lock()
+		if reason := i.shouldCutHead(false); reason&cutReasonMaxBlockBytes != 0 {
+			i.traceSizes.ClearIdle(i.lastCutTime)
+			id, err := i.cutHeadLocked()
+			i.blocksMtx.Unlock()
+			if err != nil {
+				span.RecordError(err)
+				return cutBlockIDs, err
+			}
+			cutBlockIDs = append(cutBlockIDs, id)
+			span.AddEvent("cut oversized block", oteltrace.WithAttributes(
+				attribute.String("blockID", id.String()),
+			))
+			recordBlockCutMetric(cutReasonMaxBlockBytes)
+		} else {
+			i.blocksMtx.Unlock()
+		}
 	}
 	span.AddEvent("wrote traces to head block")
 
@@ -378,12 +401,10 @@ func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
 		err := i.headBlock.Flush()
 		if err != nil {
 			span.RecordError(err)
-			return err
+			return cutBlockIDs, err
 		}
-
-		return nil
 	}
-	return nil
+	return cutBlockIDs, nil
 }
 
 func (i *instance) writeHeadBlock(id []byte, liveTrace *livetraces.LiveTrace[*v1.ResourceSpans]) error {
@@ -529,10 +550,27 @@ func (i *instance) cutBlocks(ctx context.Context, immediate bool) (uuid.UUID, er
 
 	i.traceSizes.ClearIdle(i.lastCutTime)
 
-	// Final flush
-	err := i.headBlock.Flush()
+	id, err := i.cutHeadLocked()
 	if err != nil {
 		span.RecordError(err)
+		return uuid.Nil, err
+	}
+
+	span.SetAttributes(
+		attribute.String("blockID", id.String()),
+		attribute.Int64("block_size", int64(i.walBlocks[id].DataLength())),
+	)
+
+	recordBlockCutMetric(reason)
+
+	return id, nil
+}
+
+// cutHeadLocked flushes the head block, moves it to walBlocks, and resets the
+// head block. Caller must hold blocksMtx.
+func (i *instance) cutHeadLocked() (uuid.UUID, error) {
+	err := i.headBlock.Flush()
+	if err != nil {
 		return uuid.Nil, err
 	}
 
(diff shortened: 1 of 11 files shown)


modules/livestore/instance.go
```diff
@@ -341,22 +342,42 @@ func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
...
 		}
 
 		i.tracesCreatedTotal.Inc()
+
+		// if the head block has reached max block bytes,
+		// cut it and enqueue for completion later
+		if reason := i.shouldCutHead(false); reason&cutReasonMaxBlockBytes != 0 {
+			i.blocksMtx.Lock()
+			id, err := i.cutHeadLocked()
+			i.blocksMtx.Unlock()
```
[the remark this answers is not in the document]
ruslan-mikhailov: Good catch!

modules/livestore/instance_test.go
```diff
@@ -276,3 +279,52 @@ func TestInstanceWALBackpressure(t *testing.T) {
...
+
+	walUUID, err := inst.cutBlocks(t.Context(), true)
+	require.NoError(t, err)
+
+	walBlocks = append(walBlocks, walUUID)
+	actualWALBlocks := make([]uuid.UUID, 0, len(inst.walBlocks))
+	for wb := range inst.walBlocks {
+		actualWALBlocks = append(actualWALBlocks, wb)
+	}
+	require.ElementsMatch(t, walBlocks, actualWALBlocks, "some WAL blocks where not registered")
```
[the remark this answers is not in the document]
ruslan-mikhailov: Just a test, and it was stable for multiple runs, but anyway, valid point. Fixed

modules/livestore/instance.go
```diff
@@ -349,22 +349,44 @@ func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
...
+			if err != nil {
+				span.RecordError(err)
+				return cutBlockIDs, err
+			}
+			cutBlockIDs = append(cutBlockIDs, id)
+			span.AddEvent("cut oversized block", oteltrace.WithAttributes(
+				attribute.String("blockID", id.String()),
+			))
+			recordBlockCutMetric(cutReasonMaxBlockBytes)
+		} else {
```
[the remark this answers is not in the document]
ruslan-mikhailov: Awesome catch! Fixed

modules/livestore/instance.go
```diff
@@ -349,22 +349,45 @@ func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
...
-		return nil
+		return nil, nil
 	}
 	// Sort by ID
 	sort.Slice(tracesToCut, func(i, j int) bool {
 		return bytes.Compare(tracesToCut[i].ID, tracesToCut[j].ID) == -1
 	})
-	// Collect the trace IDs that will be flushed
+
+	var cutBlockIDs []uuid.UUID
```
zhxiaogg, line 359: nit, can move the var declaration closer to its usage.

modules/livestore/instance.go
```diff
@@ -349,22 +349,45 @@ func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
...
-			return err
+			return cutBlockIDs, err
 		}
 
 		i.tracesCreatedTotal.Inc()
+
+		// if the head block has reached max block bytes,
+		// cut it and enqueue for completion later
+		i.blocksMtx.Lock()
+		if reason := i.shouldCutHead(false); reason&cutReasonMaxBlockBytes != 0 {
```
zhxiaogg, line 375: This might not be the ideal place to check and cut head block, since it's effectively duplicate with another similar check after this method call:

The `cutOneInstanceToWal` calls `cutIdleTraces` (this method) then `cutBlocks`. And there is a `shouldCutHead` in `cutBlocks` which is duplicate with this one.

modules/livestore/instance.go
```diff
@@ -349,22 +349,45 @@ func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
...
 		return bytes.Compare(tracesToCut[i].ID, tracesToCut[j].ID) == -1
 	})
-	// Collect the trace IDs that will be flushed
+
+	var cutBlockIDs []uuid.UUID
+
+	// Write traces to head block, cutting when MaxBlockBytes is reached.
 	span.AddEvent("writing traces to head block")
 	for _, t := range tracesToCut {
 		err := i.writeHeadBlock(t.ID, t)
```
zhxiaogg, line 364: I'm wondering why the WAL block size could be exceeded? Maybe we should check here for each write? (need a way to reduce the overhead for sure)

Also, I think this is definitely great to have, but still want to get a clear understanding: what's the impact looks like if the WAL size exceeded max limit?

zhxiaogg (commented):
Left some questions.

I think it's better to look at the `cutOneInstanceToWal` method, right now it's 2 step: write head block then cut it and create new wal. If the assumption is large trace could lead to wal size limit exceeding, then we might need refactor this method to make the live trace flush more continuously: write head block and cut wal whenever it exceeds the limit during the writes.

modules/livestore/instance.go
ruslan-mikhailov: > I'm wondering why the WAL block size could be exceeded?

It is because on shutdown/startup it flushes all live traces and by default max size can be 250mb, which can make a terribly huge WAL block. That's why when we cut idle traces, per each tracesToCut loop iteration, we check if the block is huge enough to cut and queue the WAL block for competition.

modules/livestore/instance.go
ruslan-mikhailov: My idea is that within the loop, when we already know that the block is huge enough to be cut due to its size, we cut right now. If not, of the new head block, we proceed with the old logic.

For example, we have limit per block 100Mb, we had traces 50Mb, 80Mb, 50Mb.
Before: we iterate and have one head block with estimated WAL block size 180Mb. cutOneInstanceToWal calls `reason := i.shouldCutHead(immediate)` and decides to cut it. It will result in 1 WAL block.
Branch version: on the second iteration, our block size is already 130Mb and we decide to cut it right now. So, after the function we have one WAL block and head block with size 50Mb. shouldCutHead for the head block will make a decision if needs to be cut (in our case, it will be no). 

Without the fix the test fails with `"119083047" is not less than or equal to "26214400"`

zhxiaogg: > on the second iteration, our block size is already 130Mb and we decide to cut it right now. So, after the function we have one WAL block and head block with size 50Mb.

I agree with the approach. I think I misread the code before. The logic looks good to me.

The code however duplicates with the other `shouldCutHead&cutHeadLocked&recordBlockCutMetric` method calls in `cutBlocks`. Do you think we can make it simpler to reduce the redundancy? I think the logic can be simplified by refactoring the `cutOneInstanceToWal`, basically loop through the livetraces and `cutBlocks` during the loop or at the end of the loop.

ruslan-mikhailov:
\+ rebase from latest main to resolve conflicts in changelog

modules/livestore/instance_test.go
```diff
@@ -276,3 +284,54 @@ func TestInstanceWALBackpressure(t *testing.T) {
...
+	}
+
+	walBlocks, err := inst.cutIdleTraces(t.Context(), true)
+	require.NoError(t, err)
+	for i := 0; i < traceCount; i += 5 { // sample, otherwise the test is slow
+		requireTraceInLiveStore(t, ls, traceIDs[i], traces[i])
+	}
+
+	walUUID, err := inst.cutBlocks(t.Context(), true)
+	require.NoError(t, err)
```
ruslan-mikhailov, line 313: I measured time for these lines before and after the fix. Results:

with fix
```
8.743576439s
5.613254582s
5.156454138s
5.08559323s
4.745346512s
8.381864035s
8.006707048s
7.868523095s
7.161819977s
5.708332955s
6.752840802s
7.608246308s
7.672004999s
9.679174377s
```

without fix

```
3.889355929s
3.265260414s
3.175263626s
4.778848507s
4.223715103s
3.657919071s
3.806618754s
3.990337718s
5.857796883s
4.35515728s
```

modules/livestore/instance.go
```diff
@@ -349,22 +349,39 @@ func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
...
+		id, err := i.cutBlocks(ctx, false)
+		if err != nil {
+			span.RecordError(err)
+			return cutBlockIDs, err
+		}
+		if id != uuid.Nil { // the block was cut
+			span.AddEvent("cut oversized block", oteltrace.WithAttributes(
+				attribute.String("blockID", id.String()),
+			))
+			cutBlockIDs = append(cutBlockIDs, id)
```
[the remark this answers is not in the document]
ruslan-mikhailov: Valid point about spans. Yeah, I might need to get back to the previous version

modules/livestore/instance.go
mapno: Also, the head block is cut only every `flush_check_period` seconds. If block size was N-1byte at the check, then by the next time it checks it can be `flush_check_period` * throughput off. Based on our operational knowledge, +100MB wouldn't be unreasonable.

modules/livestore/instance.go
```diff
@@ -350,23 +350,45 @@ func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
...
+			return cutBlockIDs, err
 		}
 
 		i.tracesCreatedTotal.Inc()
+
+		// if the head block has reached max block bytes,
+		// cut it and enqueue for completion later
+		i.blocksMtx.Lock()
+		reason := i.shouldCutHead(false)
+		i.blocksMtx.Unlock()
```
mapno: I think this contention can be avoided by returning the block size when we're writing to the headblock
```go
  func (i *instance) writeHeadBlock(id []byte, lt *livetraces.LiveTrace[*v1.ResourceSpans]) (uint64, error) {
      i.blocksMtx.Lock()
      defer i.blocksMtx.Unlock()
      // existing logic
      if err := i.headBlock.AppendTrace(id, tr, startSeconds, endSeconds, false); err != nil {
          return 0, err
      }
      return i.headBlock.DataLength(), nil
  }
```

mapno: That would avoid another check right after to know if the reason is block size 
```go
if reason&cutReasonMaxBlockBytes != 0 {
```

modules/livestore/instance.go
```diff
@@ -331,7 +331,7 @@ func countSpans(trace *tempopb.Trace) int {
 	return count
 }
 
-func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
+func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) ([]uuid.UUID, error) {
```
mapno: Nit: I find confusing that `cutIdleTraces` now cuts blocks too, and it doesn't enqueue them for completion. It's an unexpected combination of responsibilities.

modules/livestore/instance.go
ruslan-mikhailov: Nice! Yes, that will be significantly better. Applied

modules/livestore/instance.go
ruslan-mikhailov: I came up to a better idea. Instead of cutting and iterating all live traces, we instead can use iter.Pull and store its state. cutIdleTraces will be running in a loop until live traces are cut. This will separate responsibilities as it was before.

ruslan-mikhailov: It is also possible to remove a trace from live traces only after it actually cut, so if an error happens we will not lose traces. But for that, we need to hold a lock. This commit returns current behaviour: https://github.com/grafana/tempo/pull/6971/changes/c113e17ad5a37ea979ddad5c7d7d0a91dbbfa185

modules/livestore/instance.go
```diff
@@ -342,32 +343,49 @@ func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
...
 		}
 
+		tracesCut++
 		i.tracesCreatedTotal.Inc()
+
+		// if the head block has reached max block bytes,
+		// we exit earlier in order to cut the block
+		if blockSize >= i.Cfg.MaxBlockBytes {
+			return false, nil
+		}
```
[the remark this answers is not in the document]
ruslan-mikhailov: Yes! This is current behaviour, for this PR I do not change it. fwd: https://github.com/grafana/tempo/pull/6971#discussion_r3129500447

modules/livestore/instance.go
```diff
@@ -349,22 +349,45 @@ func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
...
+				return cutBlockIDs, err
+			}
+			cutBlockIDs = append(cutBlockIDs, id)
+			span.AddEvent("cut oversized block", oteltrace.WithAttributes(
+				attribute.String("blockID", id.String()),
+			))
+			recordBlockCutMetric(cutReasonMaxBlockBytes)
+		} else {
+			i.blocksMtx.Unlock()
+		}
```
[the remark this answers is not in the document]
ruslan-mikhailov: Stale comment

modules/livestore/instance.go
```diff
@@ -529,32 +550,41 @@ func (i *instance) cutBlocks(ctx context.Context, immediate bool) (uuid.UUID, er
...
+	err := i.headBlock.Flush()
+	if err != nil {
+		return uuid.Nil, err
+	}
+
+	id := (uuid.UUID)(i.headBlock.BlockMeta().BlockID)
+	blockSize := i.headBlock.DataLength()
+	i.walBlocks[id] = i.headBlock
+
 	level.Info(i.logger).Log("msg", "queueing wal block for completion", "block", id.String(), "size", blockSize)
```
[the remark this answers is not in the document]
ruslan-mikhailov: There should not be much WAL blocks per call.

pkg/livetraces/livetraces.go
```diff
@@ -115,19 +118,33 @@ func (l *LiveTraces[T]) PushWithTimestampAndLimits(ts time.Time, traceID []byte,
...
+	})
+
+	return func(yield func(*LiveTrace[T]) bool) {
+		for _, tt := range toCut {
+			l.sz -= tt.tr.sz
+			delete(l.Traces, tt.token)
+			if !yield(tt.tr) {
+				return
+			}
+		}
```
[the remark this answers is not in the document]
ruslan-mikhailov: Stale comment.

modules/livestore/live_store.go
```diff
@@ -732,11 +732,29 @@ func (s *LiveStore) cutOneInstanceToWal(ctx context.Context, inst *instance, imm
...
+		if err != nil {
+			level.Error(s.logger).Log("msg", "failed to cut idle traces", "tenant", inst.tenantID, "err", err)
+			span.RecordError(err)
+			break
+		}
+		if !liveTracesDrained { // if live traces not drained and we had no errors
+			id, err := inst.cutBlocks(ctx, false)
+			if err != nil {
+				span.RecordError(err)
+				continue
```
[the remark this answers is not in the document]
ruslan-mikhailov: Stale comment

modules/livestore/live_store.go
```diff
@@ -732,29 +732,28 @@ func (s *LiveStore) cutOneInstanceToWal(ctx context.Context, inst *instance, imm
...
-			level.Error(s.logger).Log("msg", "failed to enqueue complete operation", "tenant", inst.tenantID, "err", err)
+			level.Error(s.logger).Log("msg", "failed to cut idle traces", "tenant", inst.tenantID, "err", err)
 			span.RecordError(err)
-			return
+			break
+		}
+		id, err := inst.cutBlocks(ctx, immediate)
+		if err != nil {
+			span.RecordError(err)
+			continue
```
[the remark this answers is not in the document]
ruslan-mikhailov: Stale comment

modules/livestore/instance.go
```diff
@@ -342,32 +343,49 @@ func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
...
+		i.liveTracesIterNext, i.liveTracesIterStop = iter.Pull(i.liveTraces.CutIdle(time.Now(), immediate))
+	}
 
-	tracesToCut := i.liveTraces.CutIdle(time.Now(), immediate)
 	i.liveTracesMtx.Unlock()
 	span.AddEvent("released liveTracesMtx")
 
-	span.SetAttributes(attribute.Int("traces_to_cut", len(tracesToCut)))
+	var tracesCut int
+	defer span.SetAttributes(attribute.Int("traces_to_cut", tracesCut))
```
mapno: The `defer` here doesn't work as intended. `tracesCut` is passed when its value is 0, so that's what's recorded.

pkg/livetraces/livetraces.go
```diff
@@ -115,19 +118,33 @@ func (l *LiveTraces[T]) PushWithTimestampAndLimits(ts time.Time, traceID []byte,
...
 
-func (l *LiveTraces[T]) CutIdle(now time.Time, immediate bool) []*LiveTrace[T] {
-	res := []*LiveTrace[T]{}
-
+func (l *LiveTraces[T]) CutIdle(now time.Time, immediate bool) iter.Seq[*LiveTrace[T]] {
 	idleSince := now.Add(-l.maxIdleTime)
 	liveSince := now.Add(-l.maxLiveTime)
 
+	type tokenTrace struct {
+		token uint64
```
mapno: Is the token used?

modules/livestore/instance.go
```diff
@@ -342,32 +343,49 @@ func (i *instance) cutIdleTraces(ctx context.Context, immediate bool) error {
 	// Set metrics before cutting (similar to ingester)
 	metricLiveTraces.WithLabelValues(i.tenantID).Set(float64(i.liveTraces.Len()))
 	metricLiveTraceBytes.WithLabelValues(i.tenantID).Set(float64(i.liveTraces.Size()))
+	if i.liveTracesIterNext == nil {
+		i.liveTracesIterNext, i.liveTracesIterStop = iter.Pull(i.liveTraces.CutIdle(time.Now(), immediate))
```
mapno, line 347: IMO, using `iter.Pull` defeats the purpose of returning `iter.Seq`. It's no different that having a slice and a pointer.

I'm okay to address this in a different PR.

mapno (commented):
A few comments. I think the iterator approach is more complex than necessary, but I'm okay to address in a different PR.

pkg/livetraces/livetraces.go
ruslan-mikhailov: Oh, it left from the version where it deletes only after yield
/facepalm

Cleaned

modules/livestore/instance.go
ruslan-mikhailov: Oh no
Shame on me

Fixed. Thanks for finding it!

mapno (approved):

PR #7005, opened 2026-04-20
chore(deps): replace stale dependency github.com/golang/snappy

stoewer:
**What this PR does**:
Replaces the archived `github.com/golang/snappy` with `github.com/klauspost/compress/snappy`, which is already a direct dependency via `parquet-go`. The two libraries expose identical API signatures so the swap is drop-in. The `golang/snappy` package remains as an indirect dependency because it is still used by vendored third-party libraries (dskit, prometheus, OTel).

N/A

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,6 @@ require (
 	github.com/gogo/protobuf v1.3.2
 	github.com/gogo/status v1.1.1
 	github.com/golang/protobuf v1.5.4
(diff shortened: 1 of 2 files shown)


javiermolinar (approved):
LGTM

oleg-kozlyuk-grafana (approved):

PR #7012, opened 2026-04-20
Make AST transformations configurable

stoewer:
**What this PR does**:
AST optimizations can be skipped via hints or global config

We likely need two mechanisms, since they serve different use cases mentioned above

Disabling transformations for a single query using a hint:

```
{ service.name = "api" } with(skip_ast_transformations=[<transformations>])
```

Global opt-out via configuration allows disabling transformations globally in the query-frontend config:

```yaml
query_frontend:
  skip_ast_transformations: [<transformations>]
```

Available transformations names are:
- `or_to_in`: specifically disables the the transformation `{a="b" || a="c"}` to `{a IN ["b", "c"]}`
- `all`: disable all AST stransformation

Fixes grafana/tempo-squad#1054

**Checklist**
- [x] Tests updated
- [x] Documentation added


modules/frontend/pipeline/async_weight_middleware.go
```diff
@@ -103,7 +103,7 @@ func (c weightRequestWare) setTraceQLWeight(req Request) {
 		return
 	}
 
-	rootExpr, _, _, _, spanRequest, err := traceql.Compile(traceQLQuery)
+	rootExpr, _, _, _, spanRequest, err := traceql.Compile(traceQLQuery, traceql.WithSkipOptimization(traceql.TransformationAll))
```
stoewer, line 106: I'm unsure about this one: I think it actually makes more sense to calculate the weight based on the optimized query. 
On the other hand: this compilation is not configurable via `skip_ast_transformations` and `unsafe_query_hints`

modules/livestore/instance_search.go
```diff
@@ -675,16 +682,20 @@ func (i *instance) QueryRange(ctx context.Context, req *tempopb.QueryRangeReques
...
+	var compileOpts []traceql.CompileOption
 
 	unsafe := i.overrides.UnsafeQueryHints(i.tenantID)
 	if unsafe {
-		compileOpts = append(compileOpts, traceql.WithUnsafeQueryHints(true))
+		compileOpts = append(compileOpts, traceql.WithUnsafeHints(true))
+	}
+	for _, name := range req.SkipASTTransformations {
+		compileOpts = append(compileOpts, traceql.WithSkipOptimization(name))
 	}
```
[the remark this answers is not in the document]
stoewer: Done

pkg/api/http.go
```diff
@@ -266,6 +268,14 @@ func ParseSearchRequestWithDefault(r *http.Request, defaultSpansPerSpanSet uint3
 		req.SpansPerSpanSet = uint32(spansPerSpanSet)
 	}
 
+	if s, ok := extractQueryParam(vals, urlParamSkipASTTransformations); ok {
+		for _, name := range strings.Split(s, ",") {
+			if name = strings.TrimSpace(name); name != "" {
+				req.SkipASTTransformations = append(req.SkipASTTransformations, name)
+			}
```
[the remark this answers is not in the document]
stoewer: This is actually fine

modules/frontend/config.go
```diff
@@ -45,6 +45,10 @@ type Config struct {
 
 	// QueryEndCutoff prevents querying incomplete recent data.
 	QueryEndCutoff time.Duration `yaml:"query_end_cutoff,omitempty"`
+
+	// SkipASTTransformations is a list of AST transformation names to disable globally.
+	// Valid names: "or_to_in". Use "all" to disable all transformations.
```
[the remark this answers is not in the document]
stoewer: Docs were added

modules/frontend/pipeline/async_weight_middleware.go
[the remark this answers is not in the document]
stoewer: Same as https://github.com/grafana/tempo/pull/7012#discussion_r3122164708

mdisibio: I think this can probably wait for the cleanup round we talked about, after the language syntax is updated, so that this part doesn't need to be aware.    It's probably not critical to the query's performance either way, what do you think?

modules/frontend/metrics_query_range_sharder.go
```diff
@@ -417,6 +422,9 @@ func hashForQueryRangeRequest(req *tempopb.QueryRangeRequest) uint64 {
 	hash = fnv1a.AddUint64(hash, req.Step)
 	hash = fnv1a.AddUint64(hash, uint64(req.MaxSeries))
 	hash = fnv1a.AddUint64(hash, uint64(req.Exemplars))
+	for _, name := range req.SkipASTTransformations {
+		hash = fnv1a.AddString64(hash, name)
```
mdisibio, line 429: Leaving a note that we can probably remove this after the cleanup round of making all optimizations in the language syntax, so this caching part doesn't need to be aware (the query string itself would be sufficient).

pkg/traceql/options.go
```diff
@@ -0,0 +1,55 @@
+package traceql
+
+// CompileOption is a functional option for Parse, Compile, Engine.CompileMetricsQueryRange, and Engine.CompileMetricsQueryRangeNonRaw.
+// Parse and Compile silently ignore options specific to metrics queries like WithSpanOnlyFetch and WithTimeOverlapCutoff.
+type CompileOption func(*compileOptions)
+
+type compileOptions struct {
```
mdisibio, line 7: 👍 We talked about this not being query-type specific and I think it's fine, this is already an improvement and we can revisit it later if we want.

pkg/traceql/enum_hints.go
```diff
@@ -91,6 +92,16 @@ func (h *Hints) GetBool(k string, allowUnsafe bool) (v, ok bool) {
 	return
 }
 
+func (h *Hints) GetStringSlice(k string, allowUnsafe bool) (v []string, ok bool) {
```
mdisibio, line 95: Interesting, yeah this is likely another area we want to revisit after updating the language syntax to add static array support, e.g. `with(skip_ast_transformations=["a","b",...])` but comma-separated if of course fine.

pkg/traceql/fake_data_test.go
```diff
@@ -24,7 +24,7 @@ func TestPerformanceTestingHints_Search(t *testing.T) {
 		}
 
 		start := time.Now()
-		resp, err := engine.ExecuteSearch(context.Background(), req, nil, true)
+		resp, err := engine.ExecuteSearch(context.Background(), req, nil, WithUnsafeHints(true))
```
mdisibio, line 27: +1 this kind of stuff is much needed cleanup.

mdisibio (dismissed):
We discussed offline a bit. The main thing to note about this change is that it includes more impact than you might expect.  It's because the optimizations create an AST that can't be represented in the query string yet, i.e. the `IN()`  operator exists in the code but not the language.    So it means we can't optimize the query at the very top and pass it down, instead each downstream component (cache, querier, etc) must be aware and have both the original string and the options.

So once we add it to the language, we would be able to simplify this quite a bit, and also some of the outstanding questions like should weight calculation be done on the optimized or unoptimized version are answered.

We are going to explore a next round of language updates soon, so will revisit and cleanup then (later this year), and not hold up this feature because it's needed so we have some mechanisms to restore query functionality in case any issues are found.

mdisibio (dismissed):

modules/frontend/pipeline/async_weight_middleware.go
```diff
@@ -103,6 +103,10 @@ func (c weightRequestWare) setTraceQLWeight(req Request) {
 		return
 	}
 
+	// Calculate the weight based on the optimized TraceQL query, unless an optimization is skipped via query hints.
+	// This will deliver an accurate weight for most queries but also has two caveats:
+	// - Does not respect the users unsafe_query_hints option (most users can't skip optimizations, but they can affect the weight calculation)
+	// - Does not take query_frontend.skip_ast_transformations into account (which is likely not set anyway)
```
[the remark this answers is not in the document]
stoewer: Fixed in [bb51d7d](https://github.com/grafana/tempo/pull/7012/commits/bb51d7d5f4a9a2d090f71f7c8b3412859e5511da)

mdisibio (approved):

PR #7016, opened 2026-04-21
chore: simplify traceid padding

ariesdevil:
**What this PR does**:
Simplify traceid padding by using `fmt.Sprintf()`

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/pkg/util/traceid.go
+++ b/pkg/util/traceid.go
@@ -16,16 +16,14 @@ func HexStringToTraceID(id string) ([]byte, error) {
 	return hexStringToID(id, false)
 }
 
-// traceIDPadding is a string of 32 zeros used for left-padding trace IDs.
-const traceIDPadding = "00000000000000000000000000000000"
-
 // PadTraceIDString left-pads a trace ID hex string to 32 characters with zeros.
 // This produces W3C/OpenTelemetry compliant 32-character lowercase hex trace IDs.
 func PadTraceIDString(id string) string {
 	if len(id) >= 32 {
 		return id
 	}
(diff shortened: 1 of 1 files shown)


javiermolinar (commented):
I like the simplification but I fear this adds a performance regression. The previous code was allocation free in the fast path.

pkg/util/traceid.go
```diff
@@ -16,16 +16,11 @@ func HexStringToTraceID(id string) ([]byte, error) {
...
-const traceIDPadding = "00000000000000000000000000000000"
 
 // PadTraceIDString left-pads a trace ID hex string to 32 characters with zeros.
 // This produces W3C/OpenTelemetry compliant 32-character lowercase hex trace IDs.
 func PadTraceIDString(id string) string {
-	if len(id) >= 32 {
-		return id
-	}
-	return traceIDPadding[:32-len(id)] + id
+	return fmt.Sprintf("%0*s", 32, id)
```
ariesdevil, line 26: @javiermolinar, how about this?
```go
if len(id) >= 32 {
    return id
}

return fmt.Sprintf("%0*s", 32, id)
```

javiermolinar: Personally speaking I find more understable the form:

`traceIDPadding[:32-len(id)] + id`

than

`fmt.Sprintf("%0*s", 32, id)`

What do you think @mapno?

I still believe the is a performance regression using Sprintf. A benchmark would be a nice test

PR #7018, opened 2026-04-21
Vulture: fix for query_end_cutoff

ruslan-mikhailov:
**What this PR does**: it fixes vulture metrics check when query_end_cutoff is enabled. With step=1m, for traces that are close to `longWriteBackoff` (default is 1 minute), request will hit the cutoff and remove the last bucket. In order to solve the problem, I just reduce step duration to 10 seconds.

I ran it for some time to confirm it solves the problem:
image

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/cmd/tempo-vulture/main.go
+++ b/cmd/tempo-vulture/main.go
@@ -694,7 +694,7 @@ func queryMetrics(client httpclient.TempoHTTPClient, seed time.Time, config vult
 
 	resp, err := client.MetricsQueryRange(
 		fmt.Sprintf(`{.%s = "%s"} | count_over_time()`, attr.Key, util.StringifyAnyValue(attr.Value)),
-		start, end, "1m", 0,
+		start, end, "10s", 0,
 	)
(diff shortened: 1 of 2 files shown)


cmd/tempo-vulture/main.go
```diff
@@ -694,7 +694,7 @@ func queryMetrics(client httpclient.TempoHTTPClient, seed time.Time, config vult
 
 	resp, err := client.MetricsQueryRange(
 		fmt.Sprintf(`{.%s = "%s"} | count_over_time()`, attr.Key, util.StringifyAnyValue(attr.Value)),
-		start, end, "1m", 0,
+		start, end, "10s", 0,
```
ruslan-mikhailov, line 697: I was thinking of a config parameter, but won't it make configuration too complex? 10 seconds I think a good middle ground.

mapno (approved):

stoewer (approved):

PR #7033, opened 2026-04-22
chore: make tempo singlebinary example to use the local backend

javiermolinar:
**What this PR does**:
We don't have any example using local backend. Since Tempo singlebinary can be run with a persistent volume I think its beneficial to have a simpler example 

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/example/docker-compose/single-binary/tempo.yaml
+++ b/example/docker-compose/single-binary/tempo.yaml
@@ -34,15 +34,11 @@ query_frontend:
 
 storage:
   trace:
-    backend: s3
-    s3:
-      bucket: tempo
-      endpoint: minio:9000
-      access_key: tempo
-      secret_key: supersecret
-      insecure: true
+    backend: local
     wal:
       path: /var/tempo/wal # where to store the wal locally
+    local:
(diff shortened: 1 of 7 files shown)


example/docker-compose/single-binary/tempo.yaml
```diff
@@ -34,15 +34,11 @@ query_frontend:
...
-      bucket: tempo
-      endpoint: minio:9000
-      access_key: tempo
-      secret_key: supersecret
-      insecure: true
+    backend: local
     wal:
       path: /var/tempo/wal # where to store the wal locally
+    local:
+      path: /var/tempo/blocks
```
oleg-kozlyuk-grafana, line 41: Worth making it a named volume then - so it survives container update?

javiermolinar: Yes, good idea

oleg-kozlyuk-grafana (approved):
LGTM (pending conflict in CHANGELOG.md)

PR #7053, opened 2026-04-23
chore: updated tempo image tag in docs and jsonnets

zhxiaogg:
**What this PR does**:
- pined tempo docker image versions everywhere to next patching release 2.10.5
- cherry-picked make file target to help automate the work for this PR

Fixes #<issue number>

**Checklist**
- [ ] Tests updated
- [ ] Documentation added

### DIFF
--- a/docs/sources/tempo/set-up-for-tracing/setup-tempo/deploy/kubernetes/tanka.md
+++ b/docs/sources/tempo/set-up-for-tracing/setup-tempo/deploy/kubernetes/tanka.md
@@ -227,7 +227,7 @@ Install the `k.libsonnet`, Jsonnet, and Memcachd libraries.
 
    tempo {
        _images+:: {
-           tempo: 'grafana/tempo:latest',
+           tempo: 'grafana/tempo:2.10.5',
            tempo_query: 'grafana/tempo-query:latest',
        },
 
--- a/Makefile
+++ b/Makefile
@@ -377,6 +377,26 @@ docs-test:
 generate-manifest:  ## Generate manifest.md file
 	GO111MODULE=on CGO_ENABLED=0 go run -v pkg/docsgen/generate_manifest.go
 
+##@ Release
+
+TEMPO_IMAGE_TAG ?=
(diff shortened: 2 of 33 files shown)


docs/sources/tempo/set-up-for-tracing/setup-tempo/deploy/kubernetes/tanka.md
```diff
@@ -227,7 +227,7 @@ Install the `k.libsonnet`, Jsonnet, and Memcachd libraries.
 
    tempo {
        _images+:: {
-           tempo: 'grafana/tempo:latest',
+           tempo: 'grafana/tempo:2.10.5',
```
knylander-grafana, line 230: Why did you change this from latest? This should remain latest so we dont' have to update it every release. Unless there was a specific reason for your change? Is this because of the 3.0 release and not wanting these images to go to latest (because that would be 3.0)?

ie-pham (approved):

knylander-grafana (changes requested):
Why are you changing all of the versions from "latest" to 2.10.5? This is going to mean that every time we have a release we'll have to update every occurrence. 

Or is this to prevent the 2.10.x branch from going to the 3.0 image when that's available? If that's the case, I'll change my PR comments.

docs/sources/tempo/set-up-for-tracing/setup-tempo/deploy/kubernetes/tanka.md
zhxiaogg: This is based on recent release process update: https://github.com/grafana/tempo/blob/main/RELEASES.MD#4-update-the-tempo-image-tag

The versions are changed in patching release branch, not in main. So that for each released version, users can follow the docs with corresponding versions.

knylander-grafana: Thank you! I wasn't aware of this change. Learned something. :)

knylander-grafana (approved):
Approving.

PR #7056, opened 2026-04-23
enhancement: stale marker when series expire

### LINKED ISSUE #6494: metrics-generator: send stale markers via remote write when series are removed from registry

When the metrics-generator's `registry.stale_duration` expires for a series, the series is deleted from the internal registry and stops being emitted. However, no [stale marker](https://prometheus.io/docs/specs/prw/remote_write_spec/#stale-markers) (the special NaN value `0x7ff0000000000002`) is sent to the remote write target.

This means Prometheus has no way to know the series is gone. It continues returning the last written sample for up to 5 minutes (the hardcoded lookback delta), producing misleading query results. For example, if a service processes one request and then stops, `traces_spanmetrics_calls_total{service="my-service"}` continues to return `1` for 5 minutes after the last span was received.

This violates the [Prometheus Remote-Write 1.0 specification](https://prometheus.io/docs/specs/prw/remote_write_spec/), which states:

> Senders MUST send stale markers when a time series will no longer be appended to.

Using `rate()` mitigates the issue for counters, but gauges and instant queries remain affected. Dashboards and alerting rules that rely on the presence/absence of a series (e.g., `absent()`, `up`-style checks) are also impacted.

**Environment:**
- Tempo 2.9.0, single-binary mode on Kubernetes (Helm chart 1.24.4)
- Prometheus via kube-prometheus-stack, remote write receiver enabled
- `registry.collection_interval: 1s`, `registry.stale_duration: 3s`

When a series is deleted from the registry after `stale_duration`, the metrics-generator should emit one final remote write request containing a stale marker for that series. This would allow Prometheus (and any remote-write-compatible TSDB) to immediately mark the series as stale, matching the behavior of scrape-based ingestion.

The implementation would likely be in the registry's collection loop (`modules/generator/registry`), at the point where stale series are pruned. Before removing a series, write a sample with the stale NaN value and include it in the next remote write batch.

- **Relying on `rate()`/`increase()` for all queries**: Works for counters but not for gauges or presence-based alerting.
- **Shorter `stale_duration`**: Reduces the window but doesn't eliminate it — Prometheus still shows the last value for up to 5 minutes.
- **Reducing Prometheus lookback delta**: Not configurable per-series; changing it globally affects all queries and can break scrape-based series with irregular intervals.

**Additional context**

- Related: #1303 (metrics-generator production readiness) mentions stale series cleanup but focuses on registry memory, not remote write behavior.
- The Prometheus remote write spec explicitly requires stale markers: https://prometheus.io/docs/specs/prw/remote_write_spec/#stale-markers
- Happy to contribute an implementation if the maintainers agree with the approach.

LucasMRC:
I worked only inside the `/module/generator/registry` folder and I'm not familiar with the code outside of it, so I might be missing something. Please let me know if there's anything else I should do as part of this task!

Also, I updated the tests to cover for most of the code added in this PR, but didn't go all the way through to cover 100% of it yet. If it's required, I can work on further improving coverage.

---

**What this PR does**:
sends a stale marker to remote write when a series is removed as stale

closes https://github.com/grafana/tempo/issues/6494

**Test chages**
* Start single-binary docker compose example 
* Navigate to `localhost:9090` and query `traces_spanmetrics_calls_total{service=~"((auth|article|cart)-service|shop-backend|tempo-vulture|postgres)"}` (metrics coming from vulture and k6). Check results
* Stop k6 and vulture containers `docker compose stop vulture k6-tracing`
* Wait `metrics_generator.registry.stale_duration` + `removeStaleSeriesInterval` and query again. Should return no results.

For testing purposes, I set
```yaml
...
metrics_generator:
  registry:
    stale_duration: 20s
    collection_interval: 5s
...
```
in [the tempo config](example/docker-compose/single-binary/tempo.yaml), and updated `removeStaleSeriesInterval` to 10 seconds in [the registry file](https://github.com/grafana/tempo/blob/main/modules/generator/registry/registry.go#L84)

**Checklist**
- [x] Tests updated
- [ ] ~Documentation added~

### DIFF
--- a/modules/generator/registry/counter.go
+++ b/modules/generator/registry/counter.go
@@ -1,10 +1,12 @@
 package registry
 
 import (
+	"math"
 	"sync"
 	"time"
 
 	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/model/value"
 	"github.com/prometheus/prometheus/storage"
 	"go.uber.org/atomic"
 )
@@ -177,6 +179,12 @@ func (c *counter) removeStaleSeries(appender storage.Appender, timeMs, staleTime
 
 	for hash, s := range c.series {
 		if s.lastUpdated.Load() < staleTimeMs {
+			if appender != nil {
+				_, err := appender.Append(0, s.labels, timeMs, math.Float64frombits(value.StaleNaN))
+				if err != nil {
+					// handle
+				}
+			}
 			delete(c.series, hash)
 			c.lifecycler.OnDelete(hash, 1)
 		}
--- a/modules/generator/registry/counter.go
+++ b/modules/generator/registry/counter.go
@@ -9,6 +9,7 @@ import (
 	"github.com/prometheus/prometheus/model/value"
 	"github.com/prometheus/prometheus/storage"
 	"go.uber.org/atomic"
+	"go.uber.org/multierr"
 )
 
 type counter struct {
@@ -173,16 +174,18 @@ func (c *counter) countSeriesDemand() int {
 	return int(c.seriesDemand.Estimate())
 }
 
-func (c *counter) removeStaleSeries(appender storage.Appender, timeMs, staleTimeMs int64) {
+func (c *counter) removeStaleSeries(appender storage.Appender, timeMs, staleTimeMs int64) error {
 	c.seriesMtx.Lock()
 	defer c.seriesMtx.Unlock()
+
+	errs := []error{}
 
 	for hash, s := range c.series {
 		if s.lastUpdated.Load() < staleTimeMs {
 			if appender != nil {
 				_, err := appender.Append(0, s.labels, timeMs, math.Float64frombits(value.StaleNaN))
 				if err != nil {
-					// handle
+					errs = append(errs, err)
 				}
 			}
 			delete(c.series, hash)
@@ -190,4 +193,9 @@ func (c *counter) removeStaleSeries(appender storage.Appender, timeMs, staleTime
 		}
 	}
 	c.seriesDemand.Advance()
+
+	if len(errs) > 0 {
+		return multierr.Combine(errs...)
+	}
+	return nil
 }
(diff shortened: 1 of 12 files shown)


modules/generator/registry/registry_test.go
```diff
@@ -471,7 +473,12 @@ func collectRegistryMetricsAndAssert(t *testing.T, r *ManagedRegistry, appender
 		actual := actualSamples[i]
 
 		assert.Equal(t, expected.t, actual.t)
-		assert.Equal(t, expected.v, actual.v)
+		if math.IsNaN(expected.v) {
+			// Check for stale marker
+			assert.True(t, math.IsNaN(actual.v))
+		} else {
+			assert.Equal(t, expected.v, actual.v)
+		}
```
LucasMRC, line 481: Now that the stale marker is being sent, the metric is not removed from the samples. I updated this so that I can test for the stale marker, but there's also the option to filter out stale markers in the `capturingAppender.Append` so that the sample would be removed and we don't have to change the test result. It would depend mostly on the behavior we are trying to mimic here.

modules/generator/registry/counter.go
```diff
@@ -171,15 +174,28 @@ func (c *counter) countSeriesDemand() int {
...
 
+	errs := []error{}
+
 	for hash, s := range c.series {
 		if s.lastUpdated.Load() < staleTimeMs {
+			if appender != nil {
+				_, err := appender.Append(0, s.labels, timeMs, math.Float64frombits(value.StaleNaN))
+				if err != nil {
+					errs = append(errs, err)
+				}
```
LucasMRC, line 189: I wasn't sure on how to handle errors in the updated `removeStaleSeries` functions, I copied a pattern I saw in [modules/generator/config.go](https://github.com/grafana/tempo/blob/main/modules/generator/config.go#L163-L176).

LucasMRC:
I'm planning to update this PR this week to have it ready for merge again

PR #7087, opened 2026-04-29
feat(backend): add Tencent Cloud COS object storage backend

garudaen:
Add support for Tencent Cloud Object Storage (COS) as a new backend option
alongside S3, GCS, Azure, and Local. Uses the official COS Go SDK
(github.com/tencentyun/cos-go-sdk-v5) with virtual-hosted style bucket
URLs, multipart upload, range reads, hedged requests, and versioned
operations for user-configurable overrides.
Configuration flags are registered under the 'trace.cos.*' prefix.

**What this PR does**:
Adds a new `cos` backend to TempoDB that integrates with Tencent Cloud
Object Storage using the official COS Go SDK v5. The implementation follows
the same pattern as the existing `azure`, `s3`, `gcs`, and `local` backends.
New files:
- `tempodb/backend/cos/config.go` — Backend-specific configuration with YAML
  struct tags and CLI flag registration (`trace.cos.*`)
- `tempodb/backend/cos/cos.go` — Implements `RawReader`, `RawWriter`,
  `Compactor`, and `VersionedReaderWriter` interfaces
- `tempodb/backend/cos/compactor.go` — Block compaction lifecycle
  (mark compacted, clear block, read compacted meta)
Modified files:
- `tempodb/backend/backend.go` — Added `COS = "cos"` constant
- `tempodb/config.go` — Added `COS *cos.Config` field
- `tempodb/tempodb.go` — Added COS case in backend selection switch
- `modules/storage/config.go` — Registered COS flags and defaults
- `cmd/tempo/app/modules.go` — Added COS in usage report and scheduler init
- `cmd/tempo/app/config.go` — Added COS in storage conflict detection
- `cmd/tempo-cli/main.go` — CLI support for COS backend
- `modules/overrides/userconfigurable/client/client.go` — COS backend for
  user-configurable overrides
- `go.mod` / `go.sum` — Added `github.com/tencentyun/cos-go-sdk-v5`
Key features:
- Virtual-hosted style bucket URLs (standard COS format): `https://<bucket>-<appid>.cos.<region>.myqcloud.com`
- Custom endpoint override with bucket prefix: `https://<bucket>-<appid>.<endpoint>`
- Multipart upload via `InitiateMultipartUpload` / `UploadPart` / `CompleteMultipartUpload`
- Range reads for partial object retrieval
- ETag-based versioning for user-configurable overrides
- Hedged (redundant) HTTP requests support with Prometheus metrics
- Config validation for required fields (bucket, secret_id, secret_key, region, app_id)
Fixes #<issue number>
**Checklist**
- [ ] Tests updated
- [ ] Documentation added
CHANGELOG entry suggestion (before main / unreleased section):
* [FEATURE] Add Tencent Cloud COS object storage backend. [#<PR>] (@<user>)


electron0zero:
I was reading the Tencent Cloud Object Storage [Docs](https://www.tencentcloud.com/document/product/436/34688#prerequisites) , and it says it works with "S3 Compatible" applications, and show how to configure it.

have you tried using the S3 config to use Tencent Cloud Object Storage with Tempo and ran into any issues?

garudaen:
> I was reading the Tencent Cloud Object Storage [Docs](https://www.tencentcloud.com/document/product/436/34688#prerequisites) , and it says it works with "S3 Compatible" applications, and show how to configure it.
> 
> have you tried using the S3 config to use Tencent Cloud Object Storage with Tempo and ran into any issues?

I’ve tried it, but it didn’t work. According to the Tencent Cloud documentation: Buckets created after January 1, 2024 no longer support path-style domain names and only support virtual-hosted-style domain names. Existing buckets created before this date remain unaffected, but it is recommended to use virtual-hosted-style domain names by default.

electron0zero:
> Existing buckets created before this date remain unaffected, but it is recommended to use virtual-hosted-style domain names by default.

can you share more details and docs for this?

garudaen:
> > Existing buckets created before this date remain unaffected, but it is recommended to use virtual-hosted-style domain names by default.
> 
> can you share more details and docs for this?



> > Existing buckets created before this date remain unaffected, but it is recommended to use virtual-hosted-style domain names by default.
> 
> can you share more details and docs for this?

here, [Docs](https://www.tencentcloud.com/document/product/436/34688#prerequisites)   , at the bottom

mdisibio:
Have you run this?  I think it would be difficult for us to support or test it ourselves, so it would require some work from the community long-term.  At the minimum I would feel better know at what scale you have ran this already, and how it's going.

Not sure what is up with Copilot, but Claude found a few things that are worth mentioning:

**1. MarkBlockCompacted double-prefixes the delete path — tempodb/backend/cos/compactor.go:39**

  metaFileName := backend.MetaFileName(blockID, tenantID, rw.cfg.Prefix)  // already prefixed
  ...
  return rw.Delete(ctx, metaFileName, []string{}, nil)

  rw.Delete then does KeyPathWithPrefix([]string{}, rw.cfg.Prefix) → ["myprefix"] and joins it onto metaFileName (which already starts with myprefix/...), so the delete is sent to
  myprefix/myprefix/<tenant>/<id>/meta.json and the real meta.json is never deleted. S3 avoids this by calling rw.core.RemoveObject(...) directly. Fix: call rw.client.Object.Delete(ctx, metaFileName) instead, like
  ClearBlock does.

  **2. bucketURL() produces an invalid host when only Endpoint is set — cos.go:114**

  return fmt.Sprintf("%s://%s-%s.%s", scheme, cfg.Bucket, cfg.AppID, host)

  Validation only requires AppID when Endpoint is empty (cos.go:81). So a user setting just Endpoint (a common case for COS-on-private-VPC or bucket names that already include the AppID suffix like
  mybucket-1250000000) gets https://mybucket-.<endpoint> (trailing dash) or https://mybucket-1250000000-.<endpoint> (extra dash). Either reject the config, or skip the -AppID portion when AppID == "".

  **3. Find() and ListBlocks() may not paginate correctly for >1000 objects**

  opt := &cos.BucketGetOptions{Prefix: prefix, Marker: marker, MaxKeys: 1000} // no Delimiter
  ...
  marker = result.NextMarker

  S3/COS semantics: when no Delimiter is set, the server is not required to populate NextMarker. The conventional fix is to fall back to the last key in Contents. The S3 backend sidesteps this by using ListObjectsV2
  (continuation tokens). On large tenants Find/ListBlocks can either loop forever on the same page or terminate early. (List() is fine because it sets Delimiter: "/".)

  **5. LastModified parsing silently swallows errors and uses two different formats**

  - Find() parses with time.RFC3339. COS returns ISO‑8601 with milliseconds (2025-04-29T08:30:00.000Z) which Go's RFC3339 constant does not accept (it has no fractional seconds) → lastMod is zero, no error logged. Use
   time.RFC3339Nano or check err.
  - CompactedBlockMeta() parses the HTTP Last-Modified header with time.RFC1123 (HTTP dates use GMT, so this often works but not always — time.Parse(http.TimeFormat, ...) is the safe choice).

  The s3 backend doesn't have either issue because the minio SDK returns time.Time directly.

electron0zero:
> here, [Docs](https://www.tencentcloud.com/document/product/436/34688#prerequisites) , at the bottom

I see this at the bottom of the docs page: 
> COS does not guarantee full compatibility with S3. If you have any questions when using COS in your application, [contact us](https://intl.cloud.tencent.com/contact-sales) for assistance, and be sure to indicate that you followed the steps in this document and provide information such as the application name and screenshots.

That answers my question about COS's compatibility with S3, and with this information, I think it makes sense to accept this PR.

However I do share the same concerns as @mdisibio about scale, our ability to test, and support it without COS account.

Is there a mock/fake COS implementation that can be used to test COS in our end to end tests? if yes, that would help us test this better and ensure we don't break it by accident in future.

PR #7101, opened 2026-05-04
bugfix(livestore): skip lookback replay when partition is Inactive

zhxiaogg:
**What this PR does**:

- If a Livestore partition is marked as Inactive, due to `/prepare-partition-downscale`, then the Livestore instance should not replay from the beginning of the partition for any intermediate restarts.

**Why**:
This is to avoid partition lag due to concurrent scaling down and rolling restarts:
1. when downscaling starts, the `/prepare-partition-downscale` would set target partitions to be inactive
2. the corresponding livestore pod is turning into readonly until all the WAL blocks are expired
3. if there happen to be any rolling restarts and all WAL blocks are expired, the livestore might replay from the beginning of the partition. This is **unexpected**.
4. subsequent downscaling procedures still could succeed, but there could be partition lag due to the unexpected lookback replaying. This PR is to avoid the unexpected replaying for any intermediate rolling restarts.

Fixes N/A

**Checklist**
- [x] Tests updated
- [ ] Documentation added


mattdurham (approved):
lgtm

PR #7105, opened 2026-05-05
Better trace by id sharding

mdisibio:
**What this PR does**:
Trace by ID lookup has a tunable `query_shards` which is how many jobs to slice the block range into.  It has several problems, which stem from the fact that it's fixed, and not responsive to changes in volume, or tenants of varying size in the same cell.

This PR introduces a different way of slicing the block range into jobs:  which is to target a number of blocks per job with a different setting `blocks_per_shard`.  

The reasons this is better:
1 - It's responsive to changes in volume.  As a cell grows from 1MB/s to 1GB/s, it's one less tunable that needs to be touched.  The same value will work well at each volume.
2 - It's responsive to tenants of different sizes within the same cell.  The current fixed tunable must be set to the largest tenant, i.e. they have 100K blocks and a target number of shards is 4000 == 25 blocks per shard.  But it becomes poorly tuned for smaller tenants, i.e. they only have 1000 blocks and now execute 1 block per shard (not efficient).

A good number for this is in the range of 15-50. I've set the default to `30`, which is found from surveying a variety of workloads from <1MB/s to 1GB/s, and it works well.  (Organically we had tuned things to this range already).

To ensure that this doesn't impact existing cells, it also respects `max_outstanding_per_tenant` and will never issue more jobs than can fit in the queue.

**Note** - There is a lot of churn changing `startTime, endTime` from `int` to `time.Time`, but I think it's worth it.  There is always some ambiguity of the unit with `int`.

Fixes n/a

**Checklist**
- [x] Tests updated
- [x] Documentation added


zalegrala (dismissed):
Nice improvement.  Lots of mechanical change due to the int64 -> time.Time conversion, but looks good to me.

mdisibio:
> Lots of mechanical change due to the int64 -> time.Time conversion

Yeah, I usually don't like to do this, but I've not been liking the lack of distinction between seconds/millis/nanos, and this is at least one cleanup.

zalegrala (dismissed):

electron0zero (approved):

PR #7115, opened 2026-05-06
Support partial nested set data on broken traces

### LINKED ISSUE #6393: Improve structural and root/parent queries on incomplete traces (nested set model improvement)
Tempo has query language features for structural analysis like `{ parent } >> { descendant }`, and finding root spans with `{nestedSetParent<0}`, but it's brittle and only works on complete traces.  This means if just 1 span is missing from the trace, none of these features work.  This can be quite frustrating as dropped spans are common, caused by rate limiting in various places (server-side and/or client-side), instrumentation memory buffer limits, transient network errors, or even hitting Tempo's trace size limit.  I think we can update our algorithm to fail more gracefully and improve this situation significantly.

The internal data structure is a nested set model.  Here is where we assign it:  https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet5/nested_set_model.go#L28

Imagine we have a trace like this:
A -> B -> C -> D -> E

We start at root A, and recurse the tree, and assign the nested set bounds to:
Node: Left Right
A: 1 10
B: 2 9
C: 3 8
D: 4 7
E: 5 6

Then we can easily check A >> B, A >> E etc.

But what if span C in the middle is lost?  We get partial subtrees like:
A -> B
D -> E

Currently we don't assign any bounds here, but we can, by treating them as "disconnected sub trees".  Like this:

Node: Left Right
A: 1 4
B: 2 3
D: 5 8
E: 6 7

A+B are an internally consistent nested set model, and so are D+E.  With no changes to the read path, every query within them still works, we can find A>>B and D>>E, etc.  Only the things that required C like A >> E are still failing.  So we have preserved as much data/functionality as possible.

**Additional context**

mdisibio:
**What this PR does**:
Instead of assigning nested set data all-or-nothing, this updates it with the logic as discussed in the issue.  It keeps assigning whatever it can, including disconnected sub-trees.

This introduces a potential breaking change when searching for broken traces.  I might have previously guided people to use `{nestedSetLeft=0}`, but this no longer works. Every span will get Left assigned and be non-zero.   What does work now is using `{nestedSetParent=0}`.  This will identify the spans that are known to be broken:  they have a parent span ID, but it wasn't present in the data.  

Drilldown Traces uses `{nestedSetParent<0}` to identify root spans, and that is preserved.

A further subtlety to the change is that previously you could tell a trace was broken via `any span` in the trace, because they were all left=0.   Now you can only tell by the specific ones which lost their parent.  From inside a sub-tree, everything looks good.

Honestly I think this is a good change and impact will be minor but I wanted to highlight it.  If you are querying nested set data in-depth enough to be impacted by this, this improvement will probably be welcome.

Fixes #6393 

**Checklist**
- [x] Tests updated
- [ ] Documentation added

### DIFF
--- a/tempodb/encoding/vparquet4/nested_set_model_test.go
+++ b/tempodb/encoding/vparquet4/nested_set_model_test.go
@@ -267,6 +267,9 @@ func TestAssignNestedSetModelBounds(t *testing.T) {
 			expectedConnected: false,
 		},
 		{
+			// A->B->C chain where A is dropped. With no true root present, B becomes
+			// an orphan sub-root and still receives valid nested set bounds so structural
+			// queries work within the connected B->C segment.
 			name: "no roots",
 			trace: [][]Span{
 				{
@@ -276,8 +279,8 @@ func TestAssignNestedSetModelBounds(t *testing.T) {
 			},
 			expected: [][]Span{
 				{
-					{SpanID: []byte("cccccccc"), ParentSpanID: []byte("bbbbbbbb")},
-					{SpanID: []byte("bbbbbbbb"), ParentSpanID: []byte("aaaaaaaa")},
+					{SpanID: []byte("cccccccc"), ParentSpanID: []byte("bbbbbbbb"), NestedSetLeft: 2, NestedSetRight: 3, ParentID: 1},
+					{SpanID: []byte("bbbbbbbb"), ParentSpanID: []byte("aaaaaaaa"), NestedSetLeft: 1, NestedSetRight: 4, ParentID: 0},
 				},
 			},
 			expectedConnected: false,
--- a/tempodb/encoding/vparquet4/nested_set_model_test.go
+++ b/tempodb/encoding/vparquet4/nested_set_model_test.go
@@ -187,6 +187,36 @@ func TestAssignNestedSetModelBounds(t *testing.T) {
 			expectedConnected: false,
 		},
 		{
+			// Linear chain A->B->C->D->E->F->G->H with C and F dropped (two gaps).
+			// Produces three disjoint connected segments: A->B, D->E, and G->H. Each
+			// segment gets its own nested-set interval range so the tail subtree G->H
+			// does not overlap [1,4] (A->B) or [5,8] (D->E).
+			name: "two gaps",
+			trace: [][]Span{
+				{
+					{SpanID: []byte("aaaaaaaa")},
+					{SpanID: []byte("bbbbbbbb"), ParentSpanID: []byte("aaaaaaaa")},
+					// cccccccc missing (dropped)
+					{SpanID: []byte("dddddddd"), ParentSpanID: []byte("cccccccc")},
+					{SpanID: []byte("eeeeeeee"), ParentSpanID: []byte("dddddddd")},
+					// ffffffff missing (dropped)
+					{SpanID: []byte("gggggggg"), ParentSpanID: []byte("ffffffff")},
+					{SpanID: []byte("hhhhhhhh"), ParentSpanID: []byte("gggggggg")},
+				},
+			},
+			expected: [][]Span{
+				{
+					{SpanID: []byte("aaaaaaaa"), NestedSetLeft: 1, NestedSetRight: 4, ParentID: -1},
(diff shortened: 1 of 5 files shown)


tempodb/encoding/vparquet4/nested_set_model_test.go
```diff
@@ -152,8 +153,35 @@ func TestAssignNestedSetModelBounds(t *testing.T) {
...
+					{SpanID: []byte("eeeeeeee"), ParentSpanID: []byte("dddddddd")},
+				},
+			},
+			expected: [][]Span{
+				{
+					{SpanID: []byte("aaaaaaaa"), NestedSetLeft: 1, NestedSetRight: 4, ParentID: -1},
+					{SpanID: []byte("bbbbbbbb"), ParentSpanID: []byte("aaaaaaaa"), NestedSetLeft: 2, NestedSetRight: 3, ParentID: 1},
+					// dddddddd's parent (cccccccc) is missing — becomes a sub-root with ParentID=0
+					{SpanID: []byte("dddddddd"), ParentSpanID: []byte("cccccccc"), NestedSetLeft: 5, NestedSetRight: 8, ParentID: 0},
+					{SpanID: []byte("eeeeeeee"), ParentSpanID: []byte("dddddddd"), NestedSetLeft: 6, NestedSetRight: 7, ParentID: 5},
```
stoewer, line 184: All test have traces with a single gap.
Maybe also add a case for A->B->C->D->E->F->G->H where C and F are dropped, just to test that the sub-tree G->H has no overlapping bounds with the other two sub trees?

tempodb/encoding/vparquet4/nested_set_model_test.go
```diff
@@ -248,8 +279,8 @@ func TestAssignNestedSetModelBounds(t *testing.T) {
 			},
 			expected: [][]Span{
 				{
-					{SpanID: []byte("cccccccc"), ParentSpanID: []byte("bbbbbbbb")},
-					{SpanID: []byte("bbbbbbbb"), ParentSpanID: []byte("aaaaaaaa")},
+					{SpanID: []byte("cccccccc"), ParentSpanID: []byte("bbbbbbbb"), NestedSetLeft: 2, NestedSetRight: 3, ParentID: 1},
+					{SpanID: []byte("bbbbbbbb"), ParentSpanID: []byte("aaaaaaaa"), NestedSetLeft: 1, NestedSetRight: 4, ParentID: 0},
```
stoewer, line 313: What happens if we have a trace:
```
    A
   / \
  B   D
 /     \
C       E
         \
          F
```
The trace is splitt across blocks A, B, C reside in block 1 and D, E, F in block 2. Prior to this change D, E, F would have no nested set bounds and would not show up as results in structured queries.
Now they would get the following nested set bounds assigned:
```
A {L=1, R=6, P=-1}
B {L=2, R=5, P=1}
C {L=3, R=4, P=2}
```
And
```
D {L=1, R=6, P=0}
E {L=2, R=5, P=1}
F {L=3, R=4, P=2}
```
Judging by their trace ID and nested set bounds F is a descendant of B. Is this a correctness problem or will F not appear as descendant of B in TraceQL because they are in different blocks?

mdisibio: Right, I was thinking about this as well. Since they are in separate blocks, and we evaluate the structural operations per-block, it would be ok.   The numbers would be overlapping if dumped directly.   Which might be an issue for Drilldown Traces trace structure view (I think it queries the nested data directly), but I'm not sure.

One idea I had was to not continue the nested data numbers incrementally for the orphaned sub-trees.  Instead start at a left value using the span ID or span start time, to help distinguish them.  Wondering if this is more of a theoretical problem than a practical one.

mdisibio: The nested set columns are only int32s, so using span ID is not feasible.  Start time even would be awkward, since it is signed, then unix seconds would also not work.  But there could be some options choosing something else for the offset. The only thing is that the sub-trees in different blocks are unique. Will think about it.

tempodb/encoding/vparquet4/nested_set_model_test.go
mdisibio: Sounds good, added.

